PR #8949, opened 2017-01-13
JIT: Remove empty try regions

AndyAyersMS:
In runtimes that do not support thread abort, a try-finally with
an empty try can be optimized to simply the content of the finally.

See documentation update for more details.

Closes #8924.

PR #8952, opened 2017-01-14
JIT: fix bad assumption in non-funclet EH models

AndyAyersMS:
The newly added finally optimizations mistakenly assumed that in
non-fuclet EH models the GT_END_LFIN in the continuation block would
be the last statement. The test case added below provides an example
where this is not so.

Relax the assumption and instead search the continuation for the
GT_END_LFIN. Assert that there is exactly one.


AndyAyersMS:
@BruceForstall PTAL; cc @dotnet/jit-contrib 

#8949 will need this too.

BruceForstall:
:shipit:

PR #8955, opened 2017-01-14
Cleaning up the extra argument in throw new ArgumentException(Argument_InvalidFlag) 

SedarG:
The resource string for Argument_InvalidFlag doesn't have any parameter.
see https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System.Private.CoreLib.txt#L145

@jkotas, PTAL

### DIFF
--- a/src/mscorlib/corefx/System/Globalization/CompareInfo.Unix.cs
+++ b/src/mscorlib/corefx/System/Globalization/CompareInfo.Unix.cs
@@ -256,7 +256,7 @@ namespace System.Globalization
 
             if ((options & ValidSortkeyCtorMaskOffFlags) != 0)
             {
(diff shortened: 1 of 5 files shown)


jkotas:
> The resource string for Argument_InvalidFlag doesn't have any parameter

I do not think that this matters. The parameter name is included in the message separately for argument exceptions. It is not meant to be part of the string.

For example, look at what the following produces:

`new ArgumentException("My message", "myParameter").ToString()`
`new ArgumentException("My message").ToString()`

SedarG:
Wow. I should have tried that. It didn't occur to me that the parameter name would be handled within https://github.com/dotnet/coreclr/blob/345e201f027eb35a7a01e1714afe36b831b6f1e3/src/mscorlib/src/System/ArgumentException.cs#L72-L74. 
Thanks!

PR #8963, opened 2017-01-17
Alternative System.Lazy implementation

manofstick:
_2017-21-01: I have modified the implementation a little, so updated the text to reflect the changes_
_2017-28-01: Updated text to represent current state of implementation_

The current version of System.Lazy<T> isn't particularly performant. @mrange had some analysis in an article called [On the cost of being lazy](https://gist.github.com/mrange/aa82d33e94ad76d0f33ed86e704d7492).

Ultimately this stemmed back to FSharp's implementations of `Seq.init(Infinite)?` which wrapped each call to `IEnumerator<T>.MoveNext()` in a `lazy` that was only evaluated on `Current`. (Which is one of the issues that I have address when composing `Seq`s in my (incomplete) PR [Seq Composer](https://github.com/Microsoft/visualfsharp/pull/1570) - although it still remains in there under some circumstances)

This version:
- creates less objects for `ExecutionAndPublish` (2 objects created, vs 3), `None` (1 vs 2) & `PublicationOnly` (1 vs 2)
- Faster access after initialization (just a null check, no casting)
- Faster creation step - just one virtual call away
- Arguably a more straight-forward implementation that is easier to follow as logic is partitioned.
- Much less generics exposure, so as to improve JIT compilation

Still to do:
- [x] Add and merge your extra tests to for `Lazy<T>` to corefx repo. (https://github.com/dotnet/corefx/pull/15577)
- [x] Make sure that all corefx tests for `Lazy<T>` are passing
- [x] Incorporate any feedback/bug fixes
- [ ] Merge to CoreCLR repo
- [ ] Port the change to CoreRT repo
- [x] External (not @manofstick) validation of threading correctness
- [x] Solution for how to deal with Mode on complete item for Debug object (I suggest Nullable, if that is acceptable)


mrange:
@manofstick -  I have updated the charts to include your updated `Lazy<'T>` class.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -206,25 +157,140 @@ public Lazy(Func<T> valueFactory, LazyThreadSafetyMode mode)
...
-            m_threadSafeObj = GetObjectFromMode(mode);
-            m_valueFactory = valueFactory;
+            m_implementation = CreateInitializerFromMode(mode, this, valueFactory);
         }
 
+#region constructor helpers
+
+        private static class CreateInstance
+        {
+            private static T construct()
```
mrange: Minor comment: `Construct`?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -206,25 +157,140 @@ public Lazy(Func<T> valueFactory, LazyThreadSafetyMode mode)
...
+#endregion
+
         /// <summary>
-        /// Static helper function that returns an object based on the given mode. it also throws an exception if the mode is invalid
+        /// Creates the underlying value using the factory object, placing the result inside this
+        /// object, and then used the Lazy objects ILazyItem implemenation.to refer to it.
         /// </summary>
-        private static object GetObjectFromMode(LazyThreadSafetyMode mode)
+        /// <param name="factory">The object factory used to create the underlying object</param>
+        /// <param name="forceMemoryBarrier">true when called with ExecutionAndPublication, false
```
mrange: Minor: No parameter named `forceMemoryBarrier`?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -206,25 +157,140 @@ public Lazy(Func<T> valueFactory, LazyThreadSafetyMode mode)
...
+        /// Creates the underlying value using the factory object into a helper object and, for the
+        /// first object to complete its factory, uses that objects ILazyItem implementation
+        /// </summary>
+        /// <param name="factory">The object factory used to create the underlying object</param>
+        /// <param name="comparand">The publication object, used for synchronisation and comparison</param>
+        /// <returns>The underlying object</returns>
+        private T CreateValuePublicationOnly(Func<T> factory, PublicationOnly comparand)
+        {
+            var possibleValue = factory();
+            lock (comparand)
```
mrange: Personally I find a bit confusing that the lock for `PublicationOnly` happens inside the `Create*` method but for `ExecutionAndPublication` the caller holds the lock. Is there a reason the lock can't be moved into `CreateValueExecutionAndPublication`?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -206,25 +157,140 @@ public Lazy(Func<T> valueFactory, LazyThreadSafetyMode mode)
...
+        /// <summary>
+        /// Creates the underlying value using the factory object into a helper object and, for the
+        /// first object to complete its factory, uses that objects ILazyItem implementation
+        /// </summary>
+        /// <param name="factory">The object factory used to create the underlying object</param>
+        /// <param name="comparand">The publication object, used for synchronisation and comparison</param>
+        /// <returns>The underlying object</returns>
+        private T CreateValuePublicationOnly(Func<T> factory, PublicationOnly comparand)
+        {
+            var possibleValue = factory();
```
mrange: I was hoping that `PublicationOnly` only could get away for with an `Interlock.CompareAndExchange` over a full lock. What are the reasons a full lock is needed?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -322,147 +398,116 @@ public T Value
...
-                    }
-
-                    LazyInternalExceptionHolder exc = m_boxed as LazyInternalExceptionHolder;
-                    Debug.Assert(exc != null);
-                    exc.m_edi.Throw();
-                }
-
-                // Fall through to the slow path.
-                return LazyInitValue();
+                var implementation = m_implementation;
```
mrange: Would you think it's worth to add aggressive inlining to the performance critical property getters?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -322,147 +398,116 @@ public T Value
...
+
+        abstract private class LazyInitializer : ILazyItem
+        {
+            protected Lazy<T> Owner { get; }
+            protected Func<T> Factory { get; private set; }
+
+            protected Func<T> TakeFactory()
+            {
+                var factory = Factory;
+                if (!ReferenceEquals(CreateInstance.Factory, factory))
```
mrange: Why is this special handling needed?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -322,147 +398,116 @@ public T Value
...
                 {
-                    // We successfully created and stored the value.  At this point, the value factory delegate is
-                    // no longer needed, and we don't want to hold onto its resources.
-                    m_valueFactory = ALREADY_INVOKED_SENTINEL;
+                    if (factory == null)
+                        throw new InvalidOperationException(Environment.GetResourceString("Lazy_Value_RecursiveCallsToValue"));
+                    Factory = null;
```
mrange: Am I right in that this handling is only correct if only one thread is working or we have execution and publication lock?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -322,147 +398,116 @@ public T Value
...
-                        m_boxed = new LazyInternalExceptionHolder(ex);
-                    throw ex;
-                }
+        private sealed class PublicationOnly : LazyInitializer
+        {
+            internal PublicationOnly(Lazy<T> owner, Func<T> factory) : base(owner, factory) { }
+
+            public override T Value
+            {
+                get { return Owner.CreateValuePublicationOnly(Factory, this); }
```
mrange: No `TakeFactory` here I guess because it's not safe for concurrent calls? How is reentrancy detection handled for `PublicationOnly`

src/mscorlib/src/System/Lazy.cs
```diff
@@ -206,25 +157,140 @@ public Lazy(Func<T> valueFactory, LazyThreadSafetyMode mode)
...
+        /// </summary>
+        /// <param name="factory">The object factory used to create the underlying object</param>
+        /// <param name="comparand">The publication object, used for synchronisation and comparison</param>
+        /// <returns>The underlying object</returns>
+        private T CreateValuePublicationOnly(Func<T> factory, PublicationOnly comparand)
+        {
+            var possibleValue = factory();
+            lock (comparand)
+            {
+                if (ReferenceEquals(m_implementation, comparand))
```
mrange: The reason we are comparing against a comparand is that during evaluation of the lazy value we might get an exception? That is captured in an object that is referenced by `m_implementation`. If we had no need to capture exceptions we could have compared against null?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -50,63 +42,22 @@ internal static class LazyHelpers
...
-        [NonSerialized]
-        private Func<T> m_valueFactory;
 
-        // null if it is not thread safe mode
-        // LazyHelpers.PUBLICATION_ONLY_SENTINEL if PublicationOnly mode
-        // object if ExecutionAndPublication mode (may be ALREADY_INVOKED_SENTINEL if the value is already initialized)
-        [NonSerialized]
-        private object m_threadSafeObj;
+        // m_implementation, a volatile reference, is set to null after m_value has been set
+        private volatile ILazyItem m_implementation;
```
mrange: As the current implementation either holds a lock when updating `m_implementation` or we have no protection ==> `Lazy<'T>` can't be used concurrently then the only place where `volatile` matters is in the `.ctor` to ensure all writes to `ILazyItem` objects are visible to other threaads when they read it.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -50,63 +42,22 @@ internal static class LazyHelpers
     [ComVisible(false)]
     [DebuggerTypeProxy(typeof(System_LazyDebugView<>))]
     [DebuggerDisplay("ThreadSafetyMode={Mode}, IsValueCreated={IsValueCreated}, IsValueFaulted={IsValueFaulted}, Value={ValueForDebugDisplay}")]
-    public class Lazy<T>
+    public class Lazy<T> :
```
mrange: Hi @manofstick, I have checked the code and added a few observations/questions. Overall I see no obvious weakness. You said there was no test-coverage for `Lazy<'T>` today. Can that be added (damn difficult for these kinds of classes but at least to cover the functionality)?

mrange:
Hi @manofstick, I have checked the code and added a few observations/questions. Overall I see no obvious weakness. You said there was no test-coverage for Lazy<'T> today. Can that be added (damn difficult for these kinds of classes but at least to cover the functionality)?

mrange:
I think the reason the `Lazy<'T>` implementation above performs worse for `PublicationOnly` then `Lazzzy<'T>` is because of the publication only lock.

mrange:
From my performance testing the read overhead of a `Lazy<'T>` object that is initialized is 3x lower with the new implementation. That is pretty good!

stephentoub:
I've not reviewed these changes yet, but there are functional tests for Lazy; as with most of the types in mscorlib, the full set of tests is in corefx, e.g.
https://github.com/dotnet/corefx/blob/master/src/System.Runtime/tests/System/LazyTests.cs

src/mscorlib/src/System/Lazy.cs
manofstick: Yeah I agree with the asymmetry with the lock, but I need to call TakeFactory to null out the factory object in the case of ExecutionAndPublication. Anyway I have a *plan* to remove the lock from PublicationOnly, so that should make it better!

src/mscorlib/src/System/Lazy.cs
manofstick: Not sure of your comment there? (Or is it a comment that is meant to be a comment? even then I don't think it reads very well?)

The volatile matters in all places where m_implementation is accessed. i.e. it is the gatekeeper to ensure that m_value has a valid value before it is accessed. As reference type writes are atomic, we're guaranteed that we'll either have a valid ILazyItem implementation (from the .ctor) or null, after m_value has been written. (I'm about to make that a *little* more complex...)

src/mscorlib/src/System/Lazy.cs
manofstick: You've shamed me into it! I've thought of a cunning plan... soon, soon...

src/mscorlib/src/System/Lazy.cs
manofstick: Hmmm... that ones above my pay grade :-) I'm guessing that the JIT is probably inlining the calls to Value anyway? Maybe you could check?

src/mscorlib/src/System/Lazy.cs
manofstick: TakeFactory is required to handle re-enterency. In the case of recursive initialization it throws an exception. Needed to replicate existing functionality.

src/mscorlib/src/System/Lazy.cs
manofstick: that's on the money.

src/mscorlib/src/System/Lazy.cs
manofstick: It isn't. It hangs! But same functionality as previous implementation.

manofstick:
@mrange 

OK; with a bit of trickery I restored the CompareExchange in PublicaitonOnly without having to create a new object. In my testing with lots of threads this performed very well. Can you update your numbers to see if you get the previous performance?

manofstick:
Hiya @stephentoub !

Are you happy for me to dump all of [my tests](https://gist.github.com/manofstick/2be555566581c5057b460e53e113cdcd) into the existing LazyTest file?

stephentoub:
> Are you happy for me to dump all of my tests into the existing LazyTest file?

Extra tests would be excellent, thanks!  I'd ask, though, that you de-dup with the existing ones so that we're only adding new tests for behaviors that don't already have them.

manofstick:
@mrange sorry dude, I've modified things again, but I think I'm now at absolute minimum. I think None and PublicationOnly should show some improvements, as they no longer create any other objects other than the Lazy object itself (and as per previous comment, PublicationOnly's use of CompareExchange was restored)

@stephentoub hopefully I'll get a bit of chance to review tests tests within the next week (and hopefully no world war will begin within this time... hmmm...) Anyway, I think I'm pretty happy with the whole thing now, I've minimized object creation and all paths seem pretty optimal. I do hope serialization doesn't prove the show stopper...

jkotas:
Before this optimization, there are 3 generic types: `Lazy<T>`, `Boxed<T>`, `LazyInternalExceptionHolder<T>`. (The last one does not actually have to be generic - it can be moved out to LazyHelpers.)

After this throughput optimization, there are 9 generic types that will be created per `T`: `Lazy<T>`, `ILazyItem`, `CreateInstance<T>`, `LazyException`, `None`, `ExecutionAndPublication`, `PublicationOnly`, `PublicationOnlyWaiter`, `Boxed`. And they are more expensive than the earlier ones because of they implemented interfaces.

The generic types take time and memory to create too, and if T is valuetype there is code JITed per T too. It would interesting to look at whether the ~5x+ increase in startup cost and footprint per T pays off in typical programs. For example, if I run https://github.com/aspnet/MusicStore app, how many different Lazy<T> types are there and how many times is each one created?

manofstick:
Hi @jkotas 

 I must admit I am ignorant in the ways of the JIT. but is it really creating all those types, of are they just potentially created (dare I say it, lazily created)? And if they are created eagerly is that just because they are inner classes, in which case they could be moved out and just the each other with internal scoped members?

I.e. following the happy path of creation would be, 'Lazy', 'ILazyItem', 'ExecutionAndPublication' and that's it. So still just 3. That would be most uses.

You should need 4 for 'PublicationOnly' and 3 for ’None'.

In the case of serialisation of exceptions an extra one each is required, but they are heavy operations anyway.

Anyway, let the knowledgeable people speak, I'll move back to my corner of ignorance!

jkotas:
> because they are inner classes

Inner class relationship does not matter for performance. What matters is how the implementation is interconnected. For example, execution of `private T CreateValue(LazyThreadSafetyMode mode)` will load `LazyException<T>` type too because of it is referenced within the method. It is even worse problem with AOT compilation because of the AOT compilers cannot typically tell which types referenced in the implementation are the useful ones and so they have to compile all of them. You end up paying for it in the AOT binary size and compilation time.

The idea about making the value a direct field and use auxiliary state object is good, ie this is the best part of the change:

```
     public T Value
     {
            get
            {
                var implementation = m_implementation;
                if (implementation == null)
                    return m_value;
```

However the number of extra generic plumbing types makes it hard to tell whether it is obvious good trade-off for programs out there. Could you please try to refactor the implementation to avoid the generic types, and use non-generic types instead? The existing implementation is not perfect in this regard either - you should be able to beat it on the startup performance (=number of generic types) as well. Feel free to bring up any issues you may run into.

I have created a small perf test that creates a lot of different Lazy<T> instantiations that you can use to measure: https://gist.github.com/jkotas/9c492e62c59cb5e7220daeed587d254a . On my machines, it gives 3953ms without your changes, 5734ms with your current changes. You can compile it directly against corelib for convenience:

```
cd \coreclr\bin\Product\Windows_NT.x64.Release
csc /noconfig /nostdlib /r:System.Private.CoreLib.dll /o+ test.cs)
```

Thank you for your work on this so far!

manofstick:
Howdy @jkotas,

OK, well after a little playing around I have found that the main performance issue that raised its head in regards to your concerns was not so much the creation of generics, but rather that a `Func<T>` was being created for the default constructor factory.

Anyway, I have made some modification to remove this, and the numbers appear a fair bit better to me. I ran some tests with 32 & 64-bit and `class MyG<T, U>` and `struct MyG<T, U>` (where I cut the depth by 1) and the new version wins some and loses some.

Let me know what you think.

jkotas:
`CreateInstance<T>` and `Func<T>` 

>  the new version wins some and loses some

Could you please still share some data about where it wins and loses?

I still believe it would be benefical to make all the plumbing types non-generic. I think you will need to add extra cast from `Object` to `Func<T>` to make that work, but that should cost next to nothing.

manofstick:
@jkotas 

I don't think it's as simple as that. Maybe with a completely different implementation, but here I need to be able to call back into the 'Lazy<T>' object from the ILazyItem implementation functions so they therefore need to be generic.

Anyway, I'll have a think. Not sure if I'll get much time to play this week, but I'll try to do a quick run to print timings from your testb at some stage.

manofstick:
@jkotas 

OK; well I found some time to play today, and I think I found the right combination of pieces for pretty much all goals to be achieved (and all without a single cast! hooray!)

So lets start with your test (I'm only running these once, so values may be dubious, but ballpark...

E&P = `LazyThreadSafetyMode.ExecutionAndPublication`
PO = `LazyThreadSafetyMode.PublicationOnly`
None = `LazyThreadSafetyMode.None`

Reference Type - depth = 7
Value Type - depth 6

Construct Args | Bittage | Type | GC | Current Implementation | New Implementation | % 
--- | --- | ---- | ---- | ---- | ---- | ----
`()` | 64 | Reference | WorkStation | 3719 | 1625 | 44%
E&P | 64 | Reference | WorkStation | 4532 | 2140 | 47%
PO | 64 | Reference | WorkStation | 4297 | 2531 | 59%
None | 64 | Reference | WorkStation | 1843 | 1656 | 90%
`()` | 32 | Reference | WorkStation | 5157 | 2094 | 41%
E&P | 32 | Reference | WorkStation | 5156 | 2078 | 40%
PO | 32 | Reference | WorkStation | 5125 | 2468 | 49%
None | 32 | Reference | WorkStation | 2656 | 2734 | 103%
`()` | 64 | Value | WorkStation | 9843 | 9281 | 94%
E&P | 64 | Value | WorkStation | 9860 | 7531 | 76%
PO | 64 | Value | WorkStation | 13468 | 9172 | 68%
None | 64 | Value | WorkStation | 11516 | 8500 | 74%
`()` | 32 | Value | WorkStation | 9062 | 5594 | 62%
E&P | 32 | Value | WorkStation | 10015 | 5610 | 56%
PO | 32 | Value | WorkStation | 9781 | 5453 | 58%
None | 32 | Value | WorkStation | 7594 | 5531 | 73%
`()` | 64 | Reference | Server | 5125 | 2078 | 41%

...and then I got bored, as I hadn't automated this...

Anyway, looking pretty good.

The post JITing times are also pretty good. Not quite as good as they were, but from the original @mrange tests, they are about half that.

So I think we're finally, maybe, getting there?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -27,12 +27,81 @@
...
     {
-        // Dummy object used as the value of m_threadSafeObj if in PublicationOnly mode.
-        internal static readonly object PUBLICATION_ONLY_SENTINEL = new object();
+        public readonly static LazyImplementation NoneViaConstructor            = new LazyImplementation(LazyThreadSafetyMode.None, true);
+        public readonly static LazyImplementation NoneViaFactory                = new LazyImplementation(LazyThreadSafetyMode.None, false);
+        public readonly static LazyImplementation PublicationOnlyViaConstructor = new LazyImplementation(LazyThreadSafetyMode.PublicationOnly, true);
+        public readonly static LazyImplementation PublicationOnlyViaFactory     = new LazyImplementation(LazyThreadSafetyMode.PublicationOnly, false);
+        public readonly static LazyImplementation PublicationOnlySpinWait       = new LazyImplementation();
+
+        private readonly LazyThreadSafetyMode m_mode;
```
jkotas: We try to follow the [corefx coding style](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md) for new or significantly modified code. You can use just `_` as instance variable prefix.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -202,29 +239,171 @@ public Lazy(Func<T> valueFactory, bool isThreadSafe)
...
+        }
+
+        internal void WaitForPublicationOnly()
+        {
+            while (!ReferenceEquals(m_implementation, null))
+            {
+                // CreateValuePublicationOnly temporarily sets m_implementation to "this". The Lazy implementation
+                // has an explicit iplementation of LazyWorker which just waits for the m_value to be set, which is
+                // signalled by m_implemenation then being set to null.
+                Thread.Sleep(0);
```
jkotas: `SpinWait` would be better - it handles the different corner cases for which `Thread.Sleep(0)` does not work well.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -322,147 +505,13 @@ public T Value
...
-                    Debug.Assert(exc != null);
-                    exc.m_edi.Throw();
-                }
+                var implementation = m_implementation;
+                if (implementation == null)
+                    return m_value;
 
-                // Fall through to the slow path.
-                return LazyInitValue();
+                return implementation.GetValue(this);
```
jkotas: Have you tried moving `implementation.GetValue(this)` into separate method, like the LazyInitValue that was there before? It may make the already-initialized path faster.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -202,29 +239,171 @@ public Lazy(Func<T> valueFactory, bool isThreadSafe)
...
+            internal Boxed(T value)
+            {
+                m_value = value;
+            }
+            internal T m_value;
+        }
+
+        public Lazy(SerializationInfo information, StreamingContext context)
+        {
+            var boxed = (Boxed)information.GetValue("m_boxed", typeof(Boxed));
```
jkotas: We are not trying to maintain binary serialization compatibility with full framework. The Boxed type should not be necessary.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -202,29 +239,171 @@ public Lazy(Func<T> valueFactory, bool isThreadSafe)
...
+        class Boxed
+        {
+            internal Boxed(T value)
+            {
+                m_value = value;
+            }
+            internal T m_value;
+        }
+
+        public Lazy(SerializationInfo information, StreamingContext context)
```
jkotas: I do not see test coverage for `Lazy<T>` serialization in corefx. It should be added to make sure that this works.

manofstick:
Yet another round of changes; sorry for the churn, but your test case has led me down a different path of optimizations, but I *think* this will be it, as I have now removed as much of the generic plumbing as I could (I believe)... And the results are significantly better:

Construct Args | Bittage | Type | GC | Current Implementation | New Implementation | % 
--- | --- | ---- | ---- | ---- | ---- | ----
`()` | 64 | Reference | WorkStation | 3719 | 1047 | 28%
E&P | 64 | Reference | WorkStation | 4532 | 1047 | 23%
PO | 64 | Reference | WorkStation | 4297 | 1062 | 25%
None | 64 | Reference | WorkStation | 1843 | 1047  | 57%
`()` | 32 | Reference | WorkStation | 5157 | 1532 | 30%
E&P | 32 | Reference | WorkStation | 5156 | 1578 | 31%
PO | 32 | Reference | WorkStation | 5125 | 1531 | 30%
None | 32 | Reference | WorkStation | 2656 | 1531 | 58%
`()` | 64 | Value | WorkStation | 9843 | 4202 | 43%
E&P | 64 | Value | WorkStation | 9860 | 4188 | 42%
PO | 64 | Value | WorkStation | 13468 | 4218 | 31%
None | 64 | Value | WorkStation | 11516 | 3828 | 33%
`()` | 32 | Value | WorkStation | 9062 | 4047 | 45%
E&P | 32 | Value | WorkStation | 10015 | 4031 | 40%
PO | 32 | Value | WorkStation | 9781 | 3937 | 40%
None | 32 | Value | WorkStation | 7594 | 3750 | 49%
`()` | 64 | Reference | Server | 5125 | 1062 | 21%

I'll hopefully get onto the tests soon; although I have probably already burnt my computer time budget for this week!

manofstick:
Hi @jkotas 

So I'm slowly working through the list of items to get this in! One of the items you mentioned was "Merge to CoreCLR repo" - is this something that I should initiate, or are you just saying this is something that will occur as part of your normal dev processes?

jkotas:
> Merge to CoreCLR repo

It is something that one of the CoreCLR maintainers with the write access to the repo has to do. Just add a comment on this PR when you think it is ready to merge.

danmosemsft:
@manofstick how is this one going? Would love to see this improvement committed.

manofstick:
@danmosemsft

Been waiting on https://github.com/dotnet/corefx/pull/15577 but that looks like it now has a LGTM from all relevant parties (barring some nit picking which I'll fix this weekend)

manofstick:
@jkotas / @danmosemsft 

The regression tests have been merged into corefx, so this can be progressed.

There has been some chatter re: [use of volatile field](https://github.com/dotnet/corefx/pull/16216#issuecomment-280984103) so I would like someone with some knowledge on that too comment though.

And still need confirmation on debugging Mode field (making it Nullable seems to me the right thing to do) and someone to validate that my use of volatile is all valid. I think most of this is over the fence with you guys, but if you want me to do anything, let me know.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -27,12 +27,158 @@
...
+                case LazyState.PublicationOnlyException:
+                    return LazyThreadSafetyMode.PublicationOnly;
+
+                case LazyState.ExecutionAndPublicationViaConstructor:
+                case LazyState.ExecutionAndPublicationViaFactory:
+                case LazyState.ExecutionAndPublicationException:
+                    return LazyThreadSafetyMode.ExecutionAndPublication;
+
+                default:
+                    throw InvalidLogic();
```
jkotas: We use `Debug.Assert(false, "...");` for situations like this that cannot happen.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -27,12 +27,158 @@
...
+        internal static bool GetIsValueFaulted(LazyNonGeneric state)
+        {
+            if (state == null)
+                return false;
+            return state.GetIsValueFaulted();
+        }
+
+        internal static LazyNonGeneric Create(LazyThreadSafetyMode mode, bool useDefaultConstructor)
+        {
+            if (mode == LazyThreadSafetyMode.None)
```
jkotas: Use switch statement?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -27,12 +27,158 @@
...
+        }
+
+        /// <summary>
+        /// Constructor used for exceptions
+        /// </summary>
+        /// <param name="mode"></param>
+        /// <param name="exception"></param>
+        internal LazyNonGeneric(LazyThreadSafetyMode mode, Exception exception)
+        {
+            if (mode == LazyThreadSafetyMode.ExecutionAndPublication)
```
jkotas: Use switch statement?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -27,12 +27,158 @@
...
+            {
+                return Activator.CreateInstance(type);
+            }
+            catch (MissingMethodException)
+            {
+                throw new MissingMemberException(Environment.GetResourceString("Lazy_CreateValue_NoParameterlessCtorForT"));
+            }
+        }
+
+        internal static LazyThreadSafetyMode GetModeFromIsThreadSafe(bool isThreadSafe)
```
jkotas: This is tiny method that will get inlined into the generic code anyway. You can keep it as it is used to be if you would like.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -27,12 +27,158 @@
...
+
+                default:
+                    throw InvalidLogic();
+            }
+        }
+
+        internal static LazyThreadSafetyMode GetMode(LazyNonGeneric state)
+        {
+            if (state == null)
+                return LazyThreadSafetyMode.None; // we don't know the mode anymore
```
jkotas: > making it Nullable seems to me the right thing to do

Agree. Could you please check how it will display in the debugger?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -27,12 +27,158 @@
...
+        PublicationOnlyViaFactory     = 4,
+        PublicationOnlyWait           = 5,
+        PublicationOnlyException      = 6,
+
+        ExecutionAndPublicationViaConstructor = 7,
+        ExecutionAndPublicationViaFactory     = 8,
+        ExecutionAndPublicationException      = 9,
+    }
+
+    internal class LazyNonGeneric
```
jkotas: Nit: Similar types are typically called something like `LazyHelper`. I do not think we have any SomethingNonGeneric type anywhere.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -27,12 +27,158 @@
...
+        PublicationOnlyViaConstructor = 3,
+        PublicationOnlyViaFactory     = 4,
+        PublicationOnlyWait           = 5,
+        PublicationOnlyException      = 6,
+
+        ExecutionAndPublicationViaConstructor = 7,
+        ExecutionAndPublicationViaFactory     = 8,
+        ExecutionAndPublicationException      = 9,
+    }
+
```
jkotas: It would be useful to add a comment here that the purpose of this type is to reduce the generic instantiation code bloat.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -202,27 +308,123 @@ public Lazy(Func<T> valueFactory, bool isThreadSafe)
...
+                case LazyState.PublicationOnlyViaFactory:             return PublicationOnlyViaFactory(state);
+                case LazyState.PublicationOnlyWait:                   return PublicationOnlySpinWait();
+
+                case LazyState.ExecutionAndPublicationViaConstructor: return ExecutionAndPublication(state, true);
+                case LazyState.ExecutionAndPublicationViaFactory:     return ExecutionAndPublication(state, false);
+
+                default:
+                    state.ThrowException();
+                    return default(T);
+            }
```
jkotas: Majority of the initialization paths end with `return Value`. It may be better to make the initialization methods return void; and just have one `return Value` here - smaller code.

jkotas:
> External (not @manofstick) validation of threading correctness

I have reviewed it and it looks good to me. @stephentoub Could you please take a look as well?

jkotas:
> The regression tests have been merged into corefx

Were you able to run the corefx tests (both the new ones you have added, and the existing ones) against this? People were reporting troubles running corefx tests against private CoreCLR builds recently.

jkotas:
LGTM modulo comments. Thanks!

src/mscorlib/src/System/Lazy.cs
```diff
@@ -27,12 +27,158 @@
...
+        internal readonly static LazyNonGeneric PublicationOnlySpinWait       = new LazyNonGeneric(LazyState.PublicationOnlyWait);
+
+        internal LazyState State { get; }
+
+        private readonly System.Runtime.ExceptionServices.ExceptionDispatchInfo _exceptionDispatch;
+
+        /// <summary>
+        /// Constructor that defines the state
+        /// </summary>
+        /// <param name="state"></param>
```
stephentoub: Nit: you can remove empty XML comments

src/mscorlib/src/System/Lazy.cs
```diff
@@ -27,12 +27,158 @@
...
+        internal LazyNonGeneric(LazyState state)
+        {
+            State = state;
+        }
+
+        private Exception InvalidLogic()
```
stephentoub: Is this really needed in production?  Could the call sites just use Debug.Fail instead?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -27,12 +27,158 @@
...
+    {
+        internal readonly static LazyNonGeneric NoneViaConstructor            = new LazyNonGeneric(LazyState.NoneViaConstructor);
+        internal readonly static LazyNonGeneric NoneViaFactory                = new LazyNonGeneric(LazyState.NoneViaFactory);
+        internal readonly static LazyNonGeneric PublicationOnlyViaConstructor = new LazyNonGeneric(LazyState.PublicationOnlyViaConstructor);
+        internal readonly static LazyNonGeneric PublicationOnlyViaFactory     = new LazyNonGeneric(LazyState.PublicationOnlyViaFactory);
+        internal readonly static LazyNonGeneric PublicationOnlySpinWait       = new LazyNonGeneric(LazyState.PublicationOnlyWait);
+
+        internal LazyState State { get; }
+
+        private readonly System.Runtime.ExceptionServices.ExceptionDispatchInfo _exceptionDispatch;
```
stephentoub: ```using System.Runtime.ExceptionServices;```

src/mscorlib/src/System/Lazy.cs
```diff
@@ -27,12 +27,158 @@
...
+        {
+            return _exceptionDispatch != null;
+        }
+
+        internal static bool GetIsValueFaulted(LazyNonGeneric state)
+        {
+            if (state == null)
+                return false;
+            return state.GetIsValueFaulted();
+        }
```
stephentoub: There's no need for the first method above to be separated out; the second member can just be:
```C#
internal static bool GetIsValueFaulted() => state?._exceptionDispatch != null;
```

src/mscorlib/src/System/Lazy.cs
```diff
@@ -52,61 +198,22 @@ internal static class LazyHelpers
...
-        // null if it is not thread safe mode
-        // LazyHelpers.PUBLICATION_ONLY_SENTINEL if PublicationOnly mode
-        // object if ExecutionAndPublication mode (may be ALREADY_INVOKED_SENTINEL if the value is already initialized)
+        // we ensure that m_factory when finished is set to null to allow garbage collector to clean up
+        // any referenced items
         [NonSerialized]
-        private object m_threadSafeObj;
+        private Func<T> _factory;
 
+        // m_value eventually stores the lazily created value. It is ready when _state = null.
```
stephentoub: Nit: "ready" => "valid"?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -116,7 +223,7 @@ internal LazyInternalExceptionHolder(Exception ex)
         /// An instance created with this constructor may be used concurrently from multiple threads.
         /// </remarks>
         public Lazy()
-            : this(LazyThreadSafetyMode.ExecutionAndPublication)
+            : this(null, LazyThreadSafetyMode.ExecutionAndPublication, true)
```
stephentoub: Nit: for readability, can you name the Boolean arg at the call site?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -52,61 +198,22 @@ internal static class LazyHelpers
...
-        //m_value is Boxed --> the value is created, and m_value holds the value
-        //m_value is LazyExceptionHolder --> it holds an exception
-        private object m_boxed;
 
-        // The factory delegate that returns the value.
-        // In None and ExecutionAndPublication modes, this will be set to ALREADY_INVOKED_SENTINEL as a flag to avoid recursive calls
+        // _state, a volatile reference, is set to null after m_value has been set
         [NonSerialized]
-        private Func<T> m_valueFactory;
+        private volatile LazyNonGeneric _state;
```
stephentoub: @jkotas, is the JIT able to remove the volatility of writes to fields like this when performed in the ctor?  If not, might be better to use Volatile.Read/Write on call sites that need it rather than making the field volatile.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -202,27 +308,123 @@ public Lazy(Func<T> valueFactory, bool isThreadSafe)
...
+            _state = null;
+
+            return _value;
+        }
+
+        private T ViaFactory(LazyThreadSafetyMode mode)
+        {
+            try
+            {
+                var factory = _factory;
```
stephentoub: Nit: please use the type name instead of `var` when the type isn't obvious from the RHS of the expression

src/mscorlib/src/System/Lazy.cs
```diff
@@ -202,27 +308,123 @@ public Lazy(Func<T> valueFactory, bool isThreadSafe)
...
+        private T ExecutionAndPublication(LazyNonGeneric executionAndPublication, bool useDefaultConstructor)
+        {
+            lock (executionAndPublication)
+            {
+                // it's possible for multiple calls to have piled up behind the lock, so we need to check
+                // to see if the ExecutionAndPublication object is still the current implementation.
+                if (ReferenceEquals(_state, executionAndPublication))
+                    return useDefaultConstructor ? ViaConstructor() : ViaFactory(LazyThreadSafetyMode.ExecutionAndPublication);
+
+                return Value;
```
stephentoub: Can the `return Value;` be moved out of the lock?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -202,27 +308,123 @@ public Lazy(Func<T> valueFactory, bool isThreadSafe)
...
+            }
+        }
+
+        private T PublicationOnly(LazyNonGeneric publicationOnly, T possibleValue)
+        {
+            try { }
+            finally
+            {
+                // we run this in a finally block to ensure that we don't get a partial completion due
+                // to a Thread.Abort, which could mean that other threads might be left in infinite loops
```
stephentoub: We should not need this try { } finally or the comment in CoreCLR.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -202,27 +308,123 @@ public Lazy(Func<T> valueFactory, bool isThreadSafe)
...
+                var previous = Interlocked.CompareExchange(ref _state, LazyNonGeneric.PublicationOnlySpinWait, publicationOnly);
```
stephentoub: Nit: same var comment as earlier

src/mscorlib/src/System/Lazy.cs
```diff
@@ -202,27 +308,123 @@ public Lazy(Func<T> valueFactory, bool isThreadSafe)
...
+        }
+
+        private T PublicationOnlyViaConstructor(LazyNonGeneric initializer)
+        {
+            return PublicationOnly(initializer, CreateViaDefaultConstructor());
+        }
+
+        private T PublicationOnlyViaFactory(LazyNonGeneric initializer)
+        {
+            var factory = _factory;
```
stephentoub: Nit: var... I'll stop commenting on these :smile:

src/mscorlib/src/System/Lazy.cs
```diff
@@ -202,27 +308,123 @@ public Lazy(Func<T> valueFactory, bool isThreadSafe)
...
+            if (factory == null)
+                return PublicationOnlySpinWait();
+            return PublicationOnly(initializer, factory());
+        }
+
+        private T PublicationOnlySpinWait()
```
stephentoub: Nit: maybe rename this to something like `PublicationOnlyWaitForOtherThreadToPublish`?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -202,27 +308,123 @@ public Lazy(Func<T> valueFactory, bool isThreadSafe)
...
+            {
+                case LazyState.NoneViaConstructor:                    return ViaConstructor();
+                case LazyState.NoneViaFactory:                        return ViaFactory(LazyThreadSafetyMode.None);
+
+                case LazyState.PublicationOnlyViaConstructor:         return PublicationOnlyViaConstructor(state);
+                case LazyState.PublicationOnlyViaFactory:             return PublicationOnlyViaFactory(state);
+                case LazyState.PublicationOnlyWait:                   return PublicationOnlySpinWait();
+
+                case LazyState.ExecutionAndPublicationViaConstructor: return ExecutionAndPublication(state, true);
+                case LazyState.ExecutionAndPublicationViaFactory:     return ExecutionAndPublication(state, false);
```
stephentoub: Nit: same comment as earlier about naming bool args

src/mscorlib/src/System/Lazy.cs
```diff
@@ -254,7 +456,7 @@ internal T ValueForDebugDisplay
                 {
                     return default(T);
                 }
-                return ((Boxed)m_boxed).m_value;
+                return Value;
```
stephentoub: When using locking, if another thread is currently initializing the value while holding the lock, this will end up blocking, right?  I believe ValueForDebugDisplay was explicit coded to avoid the lock because the debugger evaluation could hang if it tries to take the lock that another thread (frozen by the debugger) is currently holding.

(I thought we were using NotifyOfCrossThreadDependency in Value, but it looks like that was removed for CoreCLR: https://referencesource.microsoft.com/#mscorlib/system/Lazy.cs,337. We should probably put that back to avoid this kind of issue; then your change here would be ok.)

src/mscorlib/src/System/Lazy.cs
```diff
@@ -322,147 +516,12 @@ public T Value
...
-                    Debug.Assert(exc != null);
-                    exc.m_edi.Throw();
-                }
-
-                // Fall through to the slow path.
-                return LazyInitValue();
+                var state = _state;
+                if (state == null)
+                    return _value;
+                return LazyGetValue(state);
```
stephentoub: We want to optimize for the initialized path.  Would the compiled code be smaller/fast/more-inlineable/etc. if this were instead written as:
```C#
public T Value => _state == null ? _value : LazyGetValue();
```
and have LazyGetValue access _state again?  I realize it'd be another volatile read in the non-initialized path, but in that path we're going to do a whole lot of other work as well, e.g. execute a delegate.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -52,61 +198,22 @@ internal static class LazyHelpers
...
-        // In None and ExecutionAndPublication modes, this will be set to ALREADY_INVOKED_SENTINEL as a flag to avoid recursive calls
+        // _state, a volatile reference, is set to null after m_value has been set
         [NonSerialized]
-        private Func<T> m_valueFactory;
+        private volatile LazyNonGeneric _state;
 
-        // null if it is not thread safe mode
-        // LazyHelpers.PUBLICATION_ONLY_SENTINEL if PublicationOnly mode
-        // object if ExecutionAndPublication mode (may be ALREADY_INVOKED_SENTINEL if the value is already initialized)
+        // we ensure that m_factory when finished is set to null to allow garbage collector to clean up
```
jkotas: `m_factory` -> `_factory`

src/mscorlib/src/System/Lazy.cs
jkotas: JIT does not have any options that would eliminate volatility of writes in ctors. However, I think `volatile` is fine here. Volatile.Read/Write would not improve anything. x86/x64 do not need memory barriers for volatile. And assignments of reference fields have implicit memory barrier on ARM/ARM64.

Volatile.Read/Write are generic methods with address taken argument. Both these factors may cause worse code to be produced with the current JIT.

stephentoub: Ok.  FWIW, I have a small preference for Volatile.Read/Write at the call site as it makes it more clear where volatile read/writes are being performed (e.g. the assignments to _state elsewhere in the code need to be done after assignments to _value, which is more clear IMO if Volatile.Write is used than if the _state field is marked volatile at its declaration), and I'd hope the JIT could appropriately optimize Volatile.* as intrinsics.

jkotas: I agree about the readability aspect. The innocently looking `_state = null;` do more that what they appear to do. It would be useful to add comment next to them at least, or just convert them to Volatile.Read/Write.

src/mscorlib/src/System/Lazy.cs
manofstick: Um, I don't have Visual Studio configured (nor have I looked how to) Not sure when I would be able to get that up and running; but I doubt it would be soon...

manofstick:
@jkotas 

> Were you able to run the corefx tests (both the new ones you have added, and the existing ones) against this? People were reporting troubles running corefx tests against private CoreCLR builds recently.

I've limited time, so I've just got a cut down test project with just the Lazy testing in it that I've been using... Relying on github to do full test suite...

src/mscorlib/src/System/Lazy.cs
manofstick: I changed `ValudForDebugDisplay`; do we still need the `NotifyOfCrossThreadDependency`?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -27,12 +27,160 @@
...
+                    Debug.Assert(false, "internal constructor, this should never occur");
+                    break;
+            }
+
+            _exceptionDispatch = ExceptionDispatchInfo.Capture(exception);
+        }
+
+        internal void ThrowException()
+        {
+            if (_exceptionDispatch == null)
```
jkotas: This can be just: `Debug.Assert((_exceptionDispatch != null, "execution path is invalid");`

src/mscorlib/src/System/Lazy.cs
```diff
@@ -27,12 +27,160 @@
...
+                    return LazyThreadSafetyMode.PublicationOnly;
+
+                case LazyState.ExecutionAndPublicationViaConstructor:
+                case LazyState.ExecutionAndPublicationViaFactory:
+                case LazyState.ExecutionAndPublicationException:
+                    return LazyThreadSafetyMode.ExecutionAndPublication;
+
+                default:
+                    Debug.Assert(false, "Invalid logic; State should always have a valid value");
+                    throw new ArgumentOutOfRangeException();
```
jkotas: You can just return `return default(LazyThreadSafetyMode)` to make the compiler happy - it is smaller/faster code.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -202,27 +310,143 @@ public Lazy(Func<T> valueFactory, bool isThreadSafe)
...
+                    ExecutionAndPublication(state, useDefaultConstructor:true);
+                    break;
+
+                case LazyState.ExecutionAndPublicationViaFactory:
+                    ExecutionAndPublication(state, useDefaultConstructor:false);
+                    break;
+
+                default:
+                    state.ThrowException();
+                    return default(T);
```
jkotas: `return default(T);` -> `break;`

src/mscorlib/src/System/Lazy.cs
jkotas: Could you please add comment about this?

src/mscorlib/src/System/Lazy.cs
```diff
@@ -254,29 +478,24 @@ internal T ValueForDebugDisplay
...
-                return ((Boxed)m_boxed).m_value;
+                return _value;
             }
         }
 
         /// <summary>
         /// Gets a value indicating whether this instance may be used concurrently from multiple threads.
         /// </summary>
-        internal LazyThreadSafetyMode Mode
+        internal LazyThreadSafetyMode? Mode
```
jkotas: Nit: All these forwarders can be written as one-liners:

`internal LazyThreadSafetyMode? Mode => LazyHelper.GetMode(_state);`
`internal bool IsValueFaulted => LazyHelper.GetIsValueFaulted(_state);`
`public bool IsValueCreated => _state == null;`

src/mscorlib/src/System/Lazy.cs
jkotas: I have checked - it works fine (displays as `null` in the debugger).

src/mscorlib/src/System/Lazy.cs
manofstick: I implemented this, but it isn't right I don't think.

Hmmm. There is a race condition here (well in the previous as well I think) with the interaction that this on in PublicationOnly and ExecutionAndPublication who the state variable. Anyway I need to think through the implications. but don't have time at the moment. Maybe tonight.

manofstick: Actually I think I'm worried about nothing. This is what thinking on public transport does to you! Anyway I'll still contemplate it properly when I have time!

manofstick:
@jkotas / @stephentoub 

OK, well I think that things are good to progress.

There had been a race condition introduced when I have removed the state from being passed to LazyGetValue, but I believe we're all good again now.

I'm getting "GitHub pull request #8963 of commit 77d3a7f746ab533fa56991d791a4c05083924d1c, has merge conflicts." so I think that's why the build is failing? Anyway, all over to you guys.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -322,147 +528,16 @@ public T Value
         {
             get
             {
-                Boxed boxed = null;
-                if (m_boxed != null )
+                while (true)
```
jkotas: Having a while loop here will make the Value property to be a bad inlining candidate - bad for steady state perf. Could you please change CreateValue to return the value instead so that it can be `return (_state == null) ? _value : CreateValue();`

jkotas: Or it can be - without changing `CreateValue` to return the value:

```
if (_state != null)
     CreateValue();
return _value;
```

This only probably slightly better because it has smaller IL, and thus it will make it better inlining candidate.

manofstick: Can't just return _value without check if _state (volatile semantics). I'm not sure I can guarantee that it would always be valid (although I'm sure it would be in tests, but that's not the same thing is it 😀) I'll return it to how it was.

src/mscorlib/src/System/Lazy.cs
```diff
@@ -18,6 +18,7 @@
 using System.Runtime;
 using System.Runtime.InteropServices;
 using System.Security;
+using System.Security.Permissions;
```
jkotas: You should not need to add `using System.Security.Permissions` back

manofstick: @jkotas 

Yeah, sorry about that I'm bashing around trying to control git (I'm used to mercurial at work, and I must say, beyond rudimentary commands git drives me insane!)

Anyway, I'll get there...

jkotas:
Very well done - thanks a lot!

I will take care of porting this to CoreRT repo for you.

manofstick:
Sorry @jkotas I just noticed that in my shameful use of git merging I splatted over your previous task which removed the ComVisible attribute. Can you just remove it; or do you want me to create a new PR?

jkotas:
Create a new PR for it please

jkotas:
I have just moved Lazy.cs to a different directory, so make sure to sync to latest master before creating the PR.

manofstick:
@jkotas do you never sleep/watch a movie/go to pub/etc  :-)  :-)

Anyway, I'll do it tomorrow. Have a good weekend!

jkotas:
I went to the pub twice this week, and watched iMax movie today :-)

Thanks for submitting the PR.

manofstick:
@jkotas obviously you're just superhuman then! :-) No worries for PR; anyway, I can pack up my bags and return to the F# playground, I've left it rotting for a couple of months with an alternative linq-to-objects (F#'s `Seq` module) implementation. I just need to be a superhuman too to find a few weeks of time...

PR #8964, opened 2017-01-17
[x86/Linux] Port 'TransitionFrame::UpdateRegDisplay'

parjong:
This commit allows Core CLR to show "Unhandled Exception" message on x86/Linux.

### DIFF
--- a/src/vm/i386/cgenx86.cpp
+++ b/src/vm/i386/cgenx86.cpp
@@ -302,14 +302,13 @@ void TransitionFrame::UpdateRegDisplayHelper(const PREGDISPLAY pRD, UINT cbStack
     ENUM_CALLEE_SAVED_REGISTERS();
 #undef CALLEE_SAVED_REGISTER
     pRD->PCTAddr = GetReturnAddressPtr();
-    pRD->ControlPC = *PTR_PCODE(pRD->PCTAddr);
-    pRD->Esp  = (DWORD)(pRD->PCTAddr + sizeof(TADDR) + cbStackPop);
 
 #ifdef WIN64EXCEPTIONS
+
     pRD->IsCallerContextValid = FALSE;
     pRD->IsCallerSPValid      = FALSE;
 
-    pRD->pCurrentContext->Eip = pRD->ControlPC;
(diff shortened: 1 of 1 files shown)


parjong:
\CC @seanshpark

parjong:
@janvorli Please take a look.

src/vm/i386/cgenx86.cpp
```diff
@@ -303,6 +305,26 @@ void TransitionFrame::UpdateRegDisplayHelper(const PREGDISPLAY pRD, UINT cbStack
     pRD->ControlPC = *PTR_PCODE(pRD->PCTAddr);
     pRD->Esp  = (DWORD)(pRD->PCTAddr + sizeof(TADDR) + cbStackPop);
 
+#ifdef WIN64EXCEPTIONS
+    pRD->IsCallerContextValid = FALSE;
+    pRD->IsCallerSPValid      = FALSE;
+
+    pRD->pCurrentContext->Eip = pRD->ControlPC;
+    pRD->pCurrentContext->Esp = GetSP();
```
parjong: It seems that stack walker find the target frame based on the address of established frame (which is just Esp at that point).

When I set ``pRD->pCurrentContext->Esp`` as the above adjusted value, stack walker failed to recognize the target frame (which results in assertion failure).

janvorli: A have looked into how this should work in detail. The difference between the value returned by GetSP and the value set to pRD->Esp is that the pRD->Esp = GetSP() + cbStackPop. In other words, one represent the SP before pushing the stack arguments at the caller and the other SP after pushing them. 
So the pRD->Esp should be set for WIN64EXCEPTIONS to the GetSP too. the pEsp and the pRD->pCurrentContext->Esp need to match.
And in fact, it happens later in the SyncRegDisplayToCurrentContext. So the lines 305 and 306 are not needed for WIN64EXCEPTIONS if you change the line 312 to
```C++
pRD->pCurrentContext->Eip = *PTR_PCODE(pRD->PCTAddr);
```
Could you please make this change and change the lines 305 and 306 to be for !WIN64EXCEPTIONS only?

PR #8968, opened 2017-01-17
Use nameof in place of hardcoded variable names

### LINKED ISSUE #8967: Use nameof in place of hardcoded variable names
To keep CoreCLR consistent with changes made in dotnet/corert#2483, some code needs to be updated to use `nameof` instead of passing hardcoded parameter names

tsolarin:
Replaces hardcoded variable names with call to `nameof` to keep code consistent with dotnet/corert#2483

Fixes #8967

/cc @jkotas

### DIFF
--- a/src/mscorlib/src/System/DateTime.cs
+++ b/src/mscorlib/src/System/DateTime.cs
@@ -461,7 +461,7 @@ namespace System {
         // parts of the result are the same as those of this DateTime.
         //
         public DateTime AddYears(int value) {
-            if (value < -10000 || value > 10000) throw new ArgumentOutOfRangeException("years", Environment.GetResourceString("ArgumentOutOfRange_DateTimeBadYears"));
+            if (value < -10000 || value > 10000) throw new ArgumentOutOfRangeException(nameof(value), Environment.GetResourceString("ArgumentOutOfRange_DateTimeBadYears"));
             Contract.EndContractBlock();
             return AddMonths(value * 12);
         }
--- a/src/mscorlib/src/System/Decimal.cs
+++ b/src/mscorlib/src/System/Decimal.cs
@@ -378,7 +378,7 @@ namespace System {
             if (value == null)
                 return 1;
             if (!(value is Decimal))
-                throw new ArgumentException(Environment.GetResourceString("Arg_MustBeDecimal"), nameof(value));
+                throw new ArgumentException(Environment.GetResourceString("Arg_MustBeDecimal"));
     
             Decimal other = (Decimal)value;    
             return FCallCompare(ref this, ref other);
--- a/src/mscorlib/src/System/Decimal.cs
+++ b/src/mscorlib/src/System/Decimal.cs
@@ -378,7 +378,7 @@ namespace System {
             if (value == null)
                 return 1;
             if (!(value is Decimal))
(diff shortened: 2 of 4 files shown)


src/mscorlib/src/System/Decimal.cs
```diff
@@ -378,7 +378,7 @@ public int CompareTo(Object value)
             if (value == null)
                 return 1;
             if (!(value is Decimal))
-                throw new ArgumentException(Environment.GetResourceString("Arg_MustBeDecimal"));
+                throw new ArgumentException(Environment.GetResourceString("Arg_MustBeDecimal"), nameof(value));
```
jkotas: There are many similar `CompareTo` methods on primitive types. This improvement should be either applied to all of them or none of them. @tarekgh @AlexGhiondea Do you think it is a good idea to apply similar fix to all `CompareTo` methods?

jkotas: If it is a good idea, maybe worth doing it in a separate PR.

jkotas:
LGTM modulo one comment

src/mscorlib/src/System/Decimal.cs
tsolarin: Perhaps I revert my change in this PR and open a separate one addressing all `CompareTo` methods?

jkotas: Sounds like a plan to me.

tsolarin:
Change reverted

tsolarin:
@jkotas change on `Decimal` has been reverted in this PR

src/mscorlib/src/System/DateTime.cs
```diff
@@ -461,7 +461,7 @@ public DateTime(int year, int month, int day, Calendar calendar)
         // parts of the result are the same as those of this DateTime.
         //
         public DateTime AddYears(int value) {
-            if (value < -10000 || value > 10000) throw new ArgumentOutOfRangeException("years", Environment.GetResourceString("ArgumentOutOfRange_DateTimeBadYears"));
+            if (value < -10000 || value > 10000) throw new ArgumentOutOfRangeException(nameof(value), Environment.GetResourceString("ArgumentOutOfRange_DateTimeBadYears"));
```
stephentoub: Note that DateTimeOffset.AddYears(int years) is implemented on top of DateTime.AddYears(int value).  By changing this from "years" to "value", it fixes DateTime.AddYears to have the ArgumentOutOfRangeException.ParamName match the parameter name, but in turn it breaks DateTimeOffset.AddYears.  This is highlighted by the two tests failing in corefx due to this line change:
https://github.com/dotnet/corefx/pull/15262#issuecomment-273467307

tsolarin: @stephentoub one possible fix that comes to mind will be to change the name of the argument here to `years` and still use the `nameof(years)` so that way the tests pass and the `nameof` enhancement still remains. What do you think?

jkotas: @tsolarin Changing argument name in signature is breaking change. This change should be rolled back - I will submit rollback in a minute.

tsolarin: Ok. Got it

jkotas: https://github.com/dotnet/coreclr/pull/8985/

PR #8970, opened 2017-01-17
Add ConditionalWeakTable.Clear and IEnumerable implementation

stephentoub:
Contributes to https://github.com/dotnet/corefx/issues/14387 (a separate PR in corefx will update the refs and tests)

Exposes the existing internal Clear implementation, and adds an ```IEnumerable<KeyValuePair<TKey,TValue>>``` implementation.  Right now, as in the associated issue, the GetEnumerator method is implemented explicitly, but we could change that now or in the future to be implemented implicitly.

This enumerator will not include any items added after GetEnumerator has been called, nor any items removed or collected during enumeration that haven't yet been enumerated (the enumerator doesn't keep items alive artificially).  It's almost entirely pay-for-play, with the exception of an int field on ConditionalWeakTable and some if checks in a few other methods.

cc: @jkotas, @weshaggard

### DIFF
--- a/src/mscorlib/src/System/Runtime/CompilerServices/ConditionalWeakTable.cs
+++ b/src/mscorlib/src/System/Runtime/CompilerServices/ConditionalWeakTable.cs
@@ -46,7 +46,6 @@
 **    expose the full public surface area.
 **
 **
-**
 ** Thread safety guarantees:
 **
 **    ConditionalWeakTable is fully thread-safe and requires no
@@ -60,6 +59,7 @@
 **    may be delayed until appdomain shutdown.
 ===========================================================*/
 
+using System.Collections;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using System.Threading;
@@ -68,7 +68,7 @@ namespace System.Runtime.CompilerServices
 {
     #region ConditionalWeakTable
     [ComVisible(false)]
-    public sealed class ConditionalWeakTable<TKey, TValue>
+    public sealed class ConditionalWeakTable<TKey, TValue> : IEnumerable<KeyValuePair<TKey, TValue>>
         where TKey : class
         where TValue : class
     {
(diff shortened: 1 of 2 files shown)


src/mscorlib/src/System/Runtime/CompilerServices/ConditionalWeakTable.cs
```diff
@@ -556,25 +761,31 @@ internal void UpdateValue(int entryIndex, TValue newValue)
...
+                    // If any expired or removed keys exist, we won't resize.
+                    // If there any active enumerators, though, we don't want
+                    // to compact and thus have no expired entries.
+                    for (int entriesIndex = 0; entriesIndex < _entries.Length; entriesIndex++)
                     {
-                        // the entry was removed
-                        hasExpiredEntries = true;
-                        break;
-                    }
+                        if (_entries[entriesIndex].HashCode == -1)
```
jkotas: Can use `ref Entry entry` here to avoid redundant array lookups? (in case the JIT is not smart enough to eliminate them).

src/mscorlib/src/System/Runtime/CompilerServices/ConditionalWeakTable.cs
```diff
@@ -600,29 +812,52 @@ internal Container Resize(int newSize)
...
-                            newEntries[newEntriesIndex].depHnd = depHnd;
-                            int bucket = hashCode & (newBuckets.Length - 1);
-                            newEntries[newEntriesIndex].Next = newBuckets[bucket];
-                            newBuckets[bucket] = newEntriesIndex;
-                            newEntriesIndex++;
-                        }
-                        else
+                        int hashCode = _entries[newEntriesIndex].HashCode;
+
+                        newEntries[newEntriesIndex].HashCode = hashCode;
```
jkotas: Same here

src/mscorlib/src/System/Runtime/CompilerServices/ConditionalWeakTable.cs
stephentoub: Fixed

src/mscorlib/src/System/Runtime/CompilerServices/ConditionalWeakTable.cs
stephentoub: Fixed. Thanks.

stephentoub:
@jkotas, I'll port this to corert once I have the tests in place in corefx.

PR #8972, opened 2017-01-17
GS checkes: only set lVExactSize for SIMD vars.

pgavlin:
Copying this field for non-SIMD variables causes a later call to
`lvaSetStruct` to skip critical initialization steps. Skipping these
steps led to failures in GC stress due to uninitialized frame variables
that contained GC pointers.


pgavlin:
@adiaaida @CarolEidt @dotnet/jit-contrib PTAL

sivarv:
Looks good.

adiaaida:
Thanks for fixing that. Looks good.

PR #8976, opened 2017-01-18
[Local GC] Provide an implementation of GCToOSInterface for Unix-like platforms

swgillespie:
This PR introduces a (first draft work-in-progress) implementation of GCToOSInterface that is complete enough to pass tests both on OSX and Ubuntu. It consists of three major parts:

1. The first commit consists of changes to build.sh to enable building CoreCLR with `FEATURE_STANDALONE_GC` defined (this already exists for Windows)
2. The second commit consists of changes to CMake files and directory structure to facilitate the construction of a "GC PAL" and the actual implementation of the GC PAL.

The "GC PAL" in this case consists of `gcenv.unix.cpp`, which builds against system headers (i.e. not the CoreCLR PAL) and provides an implementation of `GCToOSInterface` using those system headers. It does its own feature tests, independent of CoreCLR's CMake feature tests. 

Many of the implementations of methods on `GCToOSInterface` are taken from CoreRT (and modified if needed). 

There is still one iffy bit of code left in this PR, which is the redefinition of `uint64_t` on line 23 of `gcenv.unix.cpp`. The PAL typedefs `uint64_t` to be `unsigned long`, while clang's `stdint.h` defines it to be `unsigned long long`, which causes clang in turn to mangle symbols that take `uint64_t` as a parameter differently based on whether or not the symbol is compiled against PAL headers or system headers.


swgillespie:
@dotnet-bot help

swgillespie:
@dotnet-bot test OSX Checked standalone_gc
@dotnet-bot test Ubuntu Checked standalone_gc

src/gc/pal/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+add_compile_options(-fPIC)
+include_directories("../env")
+
+check_include_files(sys/time.h HAVE_SYS_TIME_H)
+check_include_files(sys/mman.h HAVE_SYS_MMAN_H)
```
janvorli: Could you please move these checks into a separate file named configure.cmake to be in line with how we do that for PAL and System.Globalization.Native.so?

swgillespie: will do!

build.cmd
```diff
@@ -587,6 +587,7 @@ echo skipmscorlib: skip building System.Private.CoreLib ^(default: System.Privat
 echo skipnative: skip building native components ^(default: native components are built^).
 echo skiptests: skip building tests ^(default: tests are built^).
 echo skipbuildpackages: skip building nuget packages ^(default: packages are built^).
+echo buildstandalone: builds the GC in a standalone mode.
```
swgillespie: you're right, will do - the command is actually "buildstandalonegc", I think I just typo'd it here.

CMakeLists.txt
```diff
@@ -572,26 +572,30 @@ if(WIN32)
   add_compile_options(/Zl) # omit default library name in .OBJ
 endif(WIN32)
 
+#--------------------------------
+# Definition directives
+#  - all clr specific compile definitions should be included in this file
+#  - all clr specific feature variable should also be added in this file
+#----------------------------------
+include(clrdefinitions.cmake)
```
swgillespie: looks like I can - will do!

swgillespie: looks like I have to build the GC pal before the CoreCLR PAL, but I can move the definitions back to their original place.

janvorli: Why does it need to be built before the CoreCLR PAL? Looks like there is some problem elsewhere. I don't see a reason for this ordering.  The defined symbols, compiler setting etc should be the same whether you build it before or after the CoreCLR PAL.

swgillespie: Looking at this again I think I misunderstood the CMake file - I am building the GC PAL after the CoreCLR PAL, but before we are adding the CoreCLR PAL header directories to the inclusion paths. Adding the PAL headers to the inclusion path is what causes problems for the GC PAL, since the GC PAL is trying to use system headers.

src/gc/pal/gcenv.unix.cpp
```diff
@@ -0,0 +1,627 @@
...
+// This isn't something we want, because we're totally fine using non-posix functions.
+#if defined(__APPLE__)
+ #define _DARWIN_C_SOURCE
+#endif // definfed(__APPLE__)
+
+#include <pthread.h>
+#include <signal.h>
+#include "config.h"
+
+// clang typedefs uint64_t to be unsigned long long, which clashes with
```
jkotas: These hacks are CoreCLR PAL specific cruft. It should not be under the GC directory. Instead, it should be on the CoreCLR side (src\vm\...) around where the GC headers are included.

src/gc/pal/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
```
jkotas: The Windows and Unix specific files should be symmetric. If you have Unix specific files under pal directory, the Windows specific files should be there too.

jkotas: It may be better to call these directories windows and unix, not call it PAL.

src/gc/gcenv.ee.standalone.inl
```diff
@@ -11,166 +11,196 @@
 // will be fowarded to this interface instance.
 extern IGCToCLR* g_theGCToCLR;
 
+// A note about this:
+// In general, we don't want to pretend to be smarter than the compiler
+// and force it to inline things. However, inlining is here is required
+// for correctness as it stands today (though it will not always be required).
```
jkotas: When it is not going to be required?

This does not make sense ... there are two build configurations:

- Statically linked GC: This file should not be compiled in for statically linked GC at all.
- Standalone GC: The GCToEEInterface* symbols should be private in standalone GC .so, so I do not see why is the linked mixing them with the ones from VM.

What am I missing?

src/gc/pal/gcenv.unix.cpp
swgillespie: What purpose would this serve by being on the CoreCLR side? This particular hack is only necessary because the GC still builds using PAL headers, which it won't always be doing. It's also only needed for this one file because it defines functions (using system headers) that are called in an environment with PAL headers. Or am I misunderstanding?

src/gc/gcenv.ee.standalone.inl
swgillespie: Not sure I follow - are you saying that it won't be a problem? When I wrote this comment I figured that `dlopen` will do the right thing with regards to conflicting symbols like this, but I'm not sure and haven't tested it.

jkotas: It should not be a problem as long as you build the GC so correctly and only export the few symbols that should be exported from it.

src/gc/pal/gcenv.unix.cpp
jkotas: ok, if it just a temporary hack, then it does not matter.

swgillespie: ok!

src/gc/pal/CMakeLists.txt
swgillespie: ok!

src/gc/gcenv.ee.standalone.inl
swgillespie: Okay, that's what I figured - so then this will not be required at that time, when the GC is loaded dynamically?

jkotas: Right. It may be useful to actually start building the standalone GC as .dll / .so instead of the approximation used today (where you build it as standalone, but still link statically). The benefits are:

- You do not need hacks like these.
- You can add it to the build by default. No need to have a separate build flavor to make sure that it builds. The shipping CoreCLR .dll/.so would still be building with statically linked flavor of the GC.
- I expect that you would need to add a bunch of stubbed out methods to get the standalone .dll / .so to link. It would give you a clear picture and burn down rate about what all you still need to deal with to get to a fully working one.

swgillespie: That does sound useful, will do!

src/gc/pal/CMakeLists.txt
swgillespie: I moved the contents of "src/gc/pal" to "src/gc/unix" and moved the Windows-specific code to "src/gc/windows" in the last commit, is that what you had in mind?

jkotas: Yes! (BTW: The CoreRT platform-specific stuff is done some way https://github.com/dotnet/corert/tree/master/src/Native/Runtime)

swgillespie:

src/gc/pal/CMakeLists.txt
swgillespie: Ok, great!

swgillespie:
I just fixed the merge conflict so this should be ready to merge, if there aren't any objections. I used the GitHub UI interface for resolving the merge conflict which created a merge commit - if we want to do a rebase instead I'd be happy to do that.

PR #8979, opened 2017-01-18
Detach CoreLib from ResourceManager implementation

morganbr:
In preparation for moving ResourceManager to CoreFX, this removes the (few) direct usages of ResourceManager from CoreLib. For the most part, code already uses Environment.GetResourceString. This change makes that and EventSource (which gets resources from other assemblies) use ResourceManager via a private interface and exposes the private contract needed for CoreFX ResourceManager to build against CoreLib.

### DIFF
--- a/src/mscorlib/src/System/Resources/IResourceManager.cs
+++ b/src/mscorlib/src/System/Resources/IResourceManager.cs
@@ -1,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Globalization;
+using System.Runtime.CompilerServices;
+
+namespace System.Resources
+{
+    /// <summary>
+    /// Private interface between CoreLib and ResourceManager so that CoreLib can access resources
+    /// </summary>
+    [FriendAccessAllowed]
+    public interface IResourceManager
+    {
+        String GetString(String name);
+        String GetString(String name, CultureInfo culture);
+    }
+}
--- a/src/mscorlib/src/mscorlib.Friends.cs
+++ b/src/mscorlib/src/mscorlib.Friends.cs
@@ -16,6 +16,9 @@ using System.Runtime.CompilerServices;
 // Depends on WindowsRuntimeImportAttribute
 [assembly: InternalsVisibleTo("System.Runtime.WindowsRuntime.UI.Xaml, PublicKey=" + _InternalsVisibleToKeys.EcmaPublicKeyFull, AllInternalsVisible=false)]
 
+// Depends on internals of Assembly and IResourceManager
+[assembly: InternalsVisibleTo("System.Resources.ResourceManager, PublicKey=" + _InternalsVisibleToKeys.EcmaPublicKeyFull, AllInternalsVisible = false)]
+
 internal class _InternalsVisibleToKeys
 {
   // Token = b77a5c561934e089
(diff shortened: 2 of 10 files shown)


morganbr:
@tarekgh please take a look
CC @dotnet/corert-contrib

src/mscorlib/src/mscorlib.Friends.cs
```diff
@@ -16,6 +16,9 @@
 // Depends on WindowsRuntimeImportAttribute
 [assembly: InternalsVisibleTo("System.Runtime.WindowsRuntime.UI.Xaml, PublicKey=" + _InternalsVisibleToKeys.EcmaPublicKeyFull, AllInternalsVisible=false)]
 
+// Depends on internals of Assembly and IResourceManager
+[assembly: InternalsVisibleTo("System.Resources.ResourceManager, PublicKey=002400000480000094000000060200000024000052534131000400000100010007d1fa57c4aed9f0a32e84aa0faefd0de9e8fd6aec8f87fb03766c834c99921eb23be79ad9d5dcc1dd9ad236132102900b723cf980957fc4e177108fc607774f29e8320e92ea05ece4e821c0a5efe8f1645c4c0c93c1ab99285d622caa652c1dfad63d745d6f2de5f17e5eaf0fc4963d261c8a12436518206dc093344d5ad293", AllInternalsVisible = false)]
```
jkotas: We do not use InternalsVisibleTo in .NET Core framework.

jkotas: (These InternalVisible are left overs from CoreCLR for Windows Phone days that we have not cleaned up yet.)

jkotas:
For the record, here is my earlier email feedback about this change: I do not think that trying to do unnatural things (like setting callbacks) to get ~5 ResourceManager related types moved to corefx is a good idea. ResourceManager is pretty central; and this scheme is different from the plan used by majority of similar types like I/O where we just moved the whole implementation to CoreLib.

src/mscorlib/src/System/Resources/IResourceManager.cs
```diff
@@ -0,0 +1,20 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Globalization;
+using System.Runtime.CompilerServices;
+
+namespace System.Resources
+{
+    /// <summary>
+    /// Private interface between CoreLib and ResourceManager so that CoreLib can access resources
```
jkotas: This won't be just CoreLib, right? All assemblies that are lower than ResourceManager (e.g. System.Private.Uri) will need to be on the same plan.

tarekgh:
@morganbr looks @jkotas not preferring moving the resource manager out of corelib. can you schedule like 30 minutes meeting we can discuss this before we proceed so everyone will be in same page? I'll hold the review till we all agree. 

CC: @ramarag @weshaggard @danmosemsft

src/mscorlib/src/mscorlib.Friends.cs
morganbr: Good to know. I can change this to go through the augments pattern.

src/mscorlib/src/System/Resources/IResourceManager.cs
morganbr: My testing on CoreFX shows I generally don't need to do that (I believe because this layering already exists on ProjectN). I did need to switch System.Collections to use the existing SR.Core indirection we use elsewehre, but the new build system may make that unnecessary..

jkotas: https://github.com/dotnet/corefx/blob/master/src/Common/src/System/SR.Core.cs indirection is ProjectN specific hack. Does this mean that we will have this hack in more assemblies and in CoreCLR as well now? It sounds pretty unfortunate.

This hack is using WinRTInterop.Callbacks. It does not exist in CoreCLR. What is it going to use in CoreCLR instead?

morganbr: Mistake reading my own code :). That indirection is only used for UAP targeting. No change was necessary for CoreCLR.

jkotas: So how does CoreCLR's System.Private.Uri gets access to ResourceManager?

morganbr:
We're going to leave this alone and add a separate copy to CoreRT instead.

PR #8981, opened 2017-01-18
[x86/Linux] Fix segmentation fault during GSCookie check (WIP)

### LINKED ISSUE #8980: [x86/Linux] Catch exceptions thrown from unsafe code

```
using System;

public class BringUpTest
{
    {
        byte* a = stackalloc byte[n];

        *a = 0;

        byte i;

    }

    public static void Main()
    {
        try
        {
            Console.WriteLine("Before");
            Localloc(0);
            Console.WriteLine("After");
        }
        catch (Exception)
        {
        }

        Console.WriteLine("Done");
    }
}
```

```
Before
Enter Localloc(0)
Here - 1
Caught (Enter)
Caught (Leave)
Done
```
```
Before
Enter Localloc(0)
Here - 1
```

parjong:
This commit attempts to fix segfault due to incorrect GSCookie address.

The current implementation computes GSCookie address via dereferencing pEbp field in REGDISPLAY, and adding some offsets, but pEbp is not initialized correctly.

This results in segmentation fault discussed in #8980.

In addition, the currenct implementation of FaultingExceptionFrame::UpdateRegDisplay does not updates pXXX fields at all, which incurs segmentation fault while checking GSCookie inside ExceptionTracker::FindNonvolatileRegisterPointers .

This commit attempts to update pEbp to fix #8980.

### DIFF
--- a/src/vm/exceptionhandling.cpp
+++ b/src/vm/exceptionhandling.cpp
@@ -4488,9 +4488,12 @@ VOID DECLSPEC_NORETURN UnwindManagedExceptionPass1(PAL_SEHException& ex, CONTEXT
 
     do
     {
+        CONTEXT currentContext = *frameContext;
+
         codeInfo.Init(controlPc);
         dispatcherContext.FunctionEntry = codeInfo.GetFunctionEntry();
         dispatcherContext.ControlPc = controlPc;
+        dispatcherContext.ContextRecord = &currentContext;
         dispatcherContext.ImageBase = codeInfo.GetModuleBase();
 #if defined(_TARGET_ARM_) 
         dispatcherContext.ControlPcIsUnwound = !!(frameContext->ContextFlags & CONTEXT_UNWOUND_TO_CALL);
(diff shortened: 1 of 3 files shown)


src/vm/exceptionhandling.cpp
```diff
@@ -4488,9 +4488,12 @@ VOID DECLSPEC_NORETURN UnwindManagedExceptionPass1(PAL_SEHException& ex, CONTEXT
 
     do
     {
+        CONTEXT currentContext = *frameContext;
```
janvorli: We should not change anything in this function, but rather fix it for x86 elsewhere. This function works for amd64 and arm, so it should be possible to ensure that it works for x86 too.
One of the problems here is that you set the dispatcherContext.ContextRecord to a local variable address, but the ProcessCLRException stores it in the ExceptionTracker that outlives this function. 
This is the reason why 65 tests have failed in the CI.

janvorli:
@parjong could you please dump me the call stack when the cookie check has seg faulted?

parjong:
@janvorli As you pointed out, it seems that this PR is not appropriate. I'll close this PR, and re-submit later with right fix.

PR #8982, opened 2017-01-18
Remove PLATFORM_UNIX and FEATURE_PAL checks in PAL

janvorli:
This change removes all ifdefs for PLATFORM_UNIX and FEATURE_PAL
from PAL and also removes dead code that was never compiled in PAL
due to both of them being always defined for PAL.


janvorli:

adityamandaleeka:
LGTM. Thanks for doing this cleanup!

PR #8984, opened 2017-01-18
Fix DateTimeFormatInfo serialization on Unix

stephentoub:
The Unix version of the DateTimeFormatInfo src is missing some [NonSerialized] attributes on a few of its fields, one of which is of type TokenHashValue, which isn't [Serializable].  This is causing DateTimeFormatInfo to itself fail to serialize.

Fixes failures in https://github.com/dotnet/corefx/pull/15068
cc: @jkotas, @danmosemsft


stephentoub:
@dotnet-bot test Linux ARM Emulator Cross Debug Build please

PR #8985, opened 2017-01-18
Rollback parameter name change to fix failing corefx tests

jkotas:


jkotas:
cc @tsolarin @stephentoub

src/mscorlib/src/System/DateTime.cs
```diff
@@ -461,7 +461,12 @@ public DateTime(int year, int month, int day, Calendar calendar)
         // parts of the result are the same as those of this DateTime.
         //
         public DateTime AddYears(int value) {
-            if (value < -10000 || value > 10000) throw new ArgumentOutOfRangeException(nameof(value), Environment.GetResourceString("ArgumentOutOfRange_DateTimeBadYears"));
+            if (value < -10000 || value > 10000)
+            {
+                // DateTimeOffset.AddYears(int years) is implemented on top of DateTime.AddYears(int value). Use the more appropriate
+                // parameter name out of the two for the exception.
```
stephentoub: Another option would be to leave this as ```nameof(value)```, and then add duplicate arg checking in DateTimeOffset.AddYears using ```"years"```, such that the right argument name shows up for each.  That could be done subsequently, though, if desired.

PR #8986, opened 2017-01-18
Refactor GT_IND value-numbering

JosephTremoulet:
Method `fgValueNumberTree` had two pieces of code each attempting to
assume responsibility for `GT_IND` and some block ops.  The one later in
the function wasn't reachable because those ops would always take the
first path.  This change moves the special logic for `GTF_IND_ARR_LEN`,
which was only present in the unreachable code, up to the reachable code,
and removes the unreachable code.

### DIFF
--- a/src/jit/valuenum.cpp
+++ b/src/jit/valuenum.cpp
@@ -6226,6 +6226,39 @@ void Compiler::fgValueNumberTree(GenTreePtr tree, bool evalAsgLhsInd)
                     fgValueNumberArrIndexVal(tree, elemTypeEq, arrVN, inxVN, addrXvnp.GetLiberal(), fldSeq);
                 }
             }
+            else if (tree->gtFlags & GTF_IND_ARR_LEN)
+            {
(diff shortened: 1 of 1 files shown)


JosephTremoulet:
@dotnet/jit-contrib PTAL.

While this change simply removes the switch cases that are unreachable, for testing I used a change that instead replaced those case arms with `unreachable()`, and verified that it passed DDR.

Even though this makes the `GTF_IND_ARR_LEN` processing code reachable within `fgValueNumberTree`, it's still unreachable dynamically (and I verified no SPMI asm diffs), because `GTF_IND_ARR_LEN` only gets set in MinOpts or `compDbgCode`, and when those are set we don't run value-numbering.  So I was on the fence as to whether to make the bit-rotting code nominally reachable or simply remove it, but chose the former so it's still there if scenarios change in the future and we want to value-number such nodes.

JosephTremoulet:

briansull:
LGTM as well

PR #8996, opened 2017-01-19
Fix Linux perf run script

DrewScoggins:
We rely on the build of CoreFX that we grab from the Corefx runs on the
main Jenkins instance.  With the recent changes the name of one of the
folders as been changed from "Native" to "native".  The checking that we
were doing was not case insensitive and as a result we started failing.
I have made that check now look for both the upper case and lower case
version of the word.


DrewScoggins:
PTAL @dpodder @jorive

DrewScoggins:
This is currently blocking all Linux runs.

PR #8997, opened 2017-01-19
Change Http to Https

DrewScoggins:
When we download things from Jenkins we used to use http for the
address we need to change this to https.


DrewScoggins:
PTAL @dpodder @jorive

PR #9001, opened 2017-01-19
[x86/Linux] Allow push/pop in funclet prolog and epilog

parjong:
The current implementation updates stack level even for funclet prolog and epilog, which results in assert failure.

This commit revises emitter not to update stack level while emitting funclet prolog and epilog in order to allow the use of push/pop instruction.

### DIFF
--- a/src/jit/emitxarch.cpp
+++ b/src/jit/emitxarch.cpp
@@ -11196,21 +11196,14 @@ size_t emitter::emitOutputInstr(insGroup* ig, instrDesc* id, BYTE** dp)
     assert(sz == emitSizeOfInsDsc(id));
 
 #if !FEATURE_FIXED_OUT_ARGS
-    bool bUpdateStackLv = true;
+    bool updateStackLevel = !emitIGisInProlog(ig) && !emitIGisInEpilog(ig);
 
-    if (emitIGisInProlog(ig))
-        bUpdateStackLv = false;
-    if (emitIGisInEpilog(ig))
-        bUpdateStackLv = false;
 #if FEATURE_EH_FUNCLETS
-    if (emitIGisInFuncletProlog(ig))
-        bUpdateStackLv = false;
-    if (emitIGisInFuncletEpilog(ig))
-        bUpdateStackLv = false;
+    updateStackLevel = updateStackLevel && !emitIGisInFuncletProlog(ig) && !emitIGisInFuncletEpilog(ig);
 #endif // FEATURE_EH_FUNCLETS
 
     // Make sure we keep the current stack level up to date
-    if (bUpdateStackLv)
+    if (updateStackLevel)
     {
         switch (ins)
         {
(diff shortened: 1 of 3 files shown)


parjong:
\CC @seanshpark

parjong:

parjong:
@dotnet/jit-contrib Please take a look.

src/jit/emitxarch.cpp
```diff
@@ -11196,9 +11196,21 @@ size_t emitter::emitOutputInstr(insGroup* ig, instrDesc* id, BYTE** dp)
...
+    if (emitIGisInProlog(ig))
+        bUpdateStackLv = false;
+    if (emitIGisInEpilog(ig))
+        bUpdateStackLv = false;
+#if FEATURE_EH_FUNCLETS
+    if (emitIGisInFuncletProlog(ig))
+        bUpdateStackLv = false;
+    if (emitIGisInFuncletEpilog(ig))
+        bUpdateStackLv = false;
+#endif // FEATURE_EH_FUNCLETS
```
janvorli: This is really long piece of code for a simple condition and it seems hard to read. Could you please change it to something like:
```C++
bool bUpdateStackLv =!emitIGisInProlog(ig) && !emitIGisInEpilog(ig);
#if FEATURE_EH_FUNCLETS
bUpdateStackLv &= !emitIGisInFuncletProlog(ig) && !emitIGisInFuncletEpilog(ig);
#endif // FEATURE_EH_FUNCLETS
```
Also, should it be FEATURE_EH_FUNCLETS && defined(_TARGET_X86_) to keep the behavior unchanged for Windows x86?

BruceForstall: @janvorli I think you mean to use `FEATURE_EH_FUNCLETS && defined(_TARGET_X86_)` to keep the behavior unchanged for Windows x64. I agree with this. It's possible (probable even?) that this wouldn't change x64 behavior, but it might be better to be safe.

IMO, the variable name should be `updateStackLevel` (don't use Hungarian, don't abbreviate). I like @janvorli's suggestion, except I don't like bitwise operators on bool, so I suggest:
```C++
bool updateStackLevel =!emitIGisInProlog(ig) && !emitIGisInEpilog(ig);
#if FEATURE_EH_FUNCLETS && defined(_TARGET_X86_) // This might also be applicable to x64.
updateStackLevel = updateStackLevel && !emitIGisInFuncletProlog(ig) && !emitIGisInFuncletEpilog(ig);
#endif // FEATURE_EH_FUNCLETS && defined(_TARGET_X86_)
```

---
In reply to: [96857183](https://github.com/dotnet/coreclr/pull/9001#discussion_r96857183) [](ancestors = 96857183)

janvorli: @BruceForstall yes, that's what I meant.

parjong: @janvorli @BruceForstall

Is it required to add ``_TARGET_X86_``? This path affects only x86 ports as ``!FEATURE_FIXED_OUT_ARGS`` is true only for x86. In addition, ``FEATURE_EH_FUNCLETS`` is true only for x86/Linux with RyuJIT. Thus, there is no effect on Windows x86 even in the current form.

In addition, this patch seems to be not related with architecture. The current implementation of ``emitter::emitBegProlog`` does not allow the stack level adjustment in prolog/epilog, and thus the same issue will occur when any windows port takes this code path.

I would like not to append ``_TARGET_X86_`` if it has no issue, but please let me know your opinion.

For the code style, I'll revise as you suggested.

parjong:

PR #9002, opened 2017-01-19
[ARM/CI] Fix arm32 CI scripts

wateret:
The name of CoreFX Native build directory has changed.

### DIFF
--- a/tests/runtest.sh
+++ b/tests/runtest.sh
@@ -356,8 +356,8 @@ function create_core_overlay {
     if [ -z "$coreFxNativeBinDir" ]; then
         exit_with_error "$errorSource" "One of --coreOverlayDir or --coreFxBinDir must be specified." "$printUsage"
     fi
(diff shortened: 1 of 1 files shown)


wateret:
These changes must be merged after https://github.com/dotnet/corefx/pull/15311 landed.
Once CoreFX arm CI works again, all CoreCLR arm PRtests will fail so we need these changes.

cc: @hseok-oh

wateret:
Or maybe it would be better to revert the directory in CoreFX. The directory name changed only case of a letter. It wouldn't matter on Windows anyway.

hseok-oh:
@wateret CoreFX managed code bin also changed. Here is my work.
https://github.com/hseok-oh/coreclr/tree/fix_runtest

wateret:
@hseok-oh I was not aware of that. I'd would be nicer if you create a new PR with your work. Then I'll close this.

hseok-oh:
@wateret I've create a new PR. Thanks.

PR #8757, opened 2016-12-30
Fix heap live-in calculation

JosephTremoulet:
Update the code in liveness that sets the `fgCurHeapUse` flag (which needs
to identify whether the current block has an upwards-exposed use of the
heap) to set this unconditionally for opcodes that use the heap.
Previously, this code was avoiding setting the flag if the block has a
prior def of the heap, but in the absence of a guarantee to the contrary,
we must assume that the heap use may not alias the heap def, and so still
be upwards-exposed and cause the heap to be live-in to the block.

Also remove the OptRepeat workaround for the lack of this, in
`ResetOptAnnotations`.

Fixes #7846.

PR #8905, opened 2017-01-11
Remove bogus fgCurHeapDef guard

JosephTremoulet:
We should set fgCurHeapUse when visiting a heap use regardless of whether
fgCurHeapDef is set, because we haven't proven that the def writes all
memory that the use might read.  This should have been included in #8757
but somehow slipped through.

Issue #7903, opened 2016-10-31
JIT: Value numbering overly conservative around byrefs


JosephTremoulet:
Value numbering is (rightly) conservative w.r.t. indirections off byrefs, but should at least be able to remove back-to-back redundant loads.  For example, consider method `Detail` below:

```c#
using System;
using System.Runtime.CompilerServices;

public class Example
{
    public string color;
    public string make;
    public string model;
    public string year;

    public struct Car
    {
        Example ex;
        public Car(Example e) { ex = e; }

        [MethodImpl(MethodImplOptions.NoInlining)]        
        public void Detail()
        {
            Describe(ex.color, ex.year, ex.make, ex.model);
        }

        [MethodImpl(MethodImplOptions.NoInlining)]
        public void Describe(string color, string make, string model, string year)
        {
            Console.WriteLine("This is a {0} {1} {2} {3}\n",
                color, make, model, year);
        }
    }

    public static int Main(string[] args)
    {
        Example e = new Example();
        e.year = "1977";
        e.make = "GMC";
        e.model = "K1500";
        e.color = "Rusty";
        
        Car c  = new Car(e);
        c.Detail();
        return 100;
    }
}
```

We generate redundant loads of `this.ex` for each of the arguments:

```asm
; Assembly listing for method Car:Detail():this
; Emitting BLENDED_CODE for X64 CPU with AVX
; optimized code
; rsp based frame
; partially interruptible
; Final local variable assignments
;
;  V00 this         [V00,T00] (  7,   7  )   byref  ->  rcx         this
;  V01 OutArgs      [V01    ] (  1,   1  )  lclBlk (40) [rsp+0x00]  
;
; Lcl frame size = 40


G_M19041_IG01:
       4883EC28             sub      rsp, 40

G_M19041_IG02:
       488B11               mov      rdx, gword ptr [rcx]
       488B5218             mov      rdx, gword ptr [rdx+24]
       4889542420           mov      gword ptr [rsp+20H], rdx
       488B11               mov      rdx, gword ptr [rcx]
       488B5208             mov      rdx, gword ptr [rdx+8]
       4C8B01               mov      r8, gword ptr [rcx]
       4D8B4020             mov      r8, gword ptr [r8+32]
       4C8B09               mov      r9, gword ptr [rcx]
       4D8B4910             mov      r9, gword ptr [r9+16]
       E856FBFFFF           call     Car:Describe(ref,ref,ref,ref):this
       90                   nop      

[...]

PR #9004, opened 2017-01-19
More heap liveness fixes

JosephTremoulet:
Two more fixes in the vein of #8757:
  - The `fgCurHeapUse` calculation in codegenlegacy.cpp
  - The cross-block liveness propagation in `LiveVarAnalysis`

### DIFF
--- a/src/jit/liveness.cpp
+++ b/src/jit/liveness.cpp
@@ -1129,7 +1129,9 @@ class LiveVarAnalysis
         VarSetOps::DiffD(m_compiler, m_liveIn, block->bbVarDef);
         VarSetOps::UnionD(m_compiler, m_liveIn, block->bbVarUse);
 
-        m_heapLiveIn = (m_heapLiveOut && !block->bbHeapDef) || block->bbHeapUse;
+        // Even if m_heapDef is set, we must assume that it doesn't kill heap if m_heapLiveOut, since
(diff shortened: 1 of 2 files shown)


JosephTremoulet:
@pgavlin/@briansull PTAL.
/cc @dotnet/jit-contrib 

@pgavlin, I reconsidered your suggestion from #8905 to extend these changes to codegenlegacy; as I'm continuing to work on addressing #7903, it's become clear that trying to avoid changing the legacy path would require some awkward ifdefs and, worse, make some of this code a lot harder to reason about.

I verified no asm diffs, including x86 legacy backend.

src/jit/liveness.cpp
```diff
@@ -1129,7 +1129,9 @@ class LiveVarAnalysis
         VarSetOps::DiffD(m_compiler, m_liveIn, block->bbVarDef);
         VarSetOps::UnionD(m_compiler, m_liveIn, block->bbVarUse);
 
-        m_heapLiveIn = (m_heapLiveOut && !block->bbHeapDef) || block->bbHeapUse;
+        // Even if m_heapDef is set, we must assume that it doesn't kill heap if m_heapLiveOut, since
+        // (without proof otherwise) the use and def may touch different heap memory at run-time.
```
briansull: There is nothing named m_heapDef, so this comment needs to be corrected.
You probably meant block->bbHeapDef

PR #9010, opened 2017-01-19
Cleanup uses of string.Split

justinvp:
Now that there's an overload of Split that takes a single char, use it instead of the array-based overload.

### DIFF
--- a/src/mscorlib/src/System/Security/Util/StringExpressionSet.cs
+++ b/src/mscorlib/src/System/Security/Util/StringExpressionSet.cs
@@ -33,8 +33,8 @@ namespace System.Security.Util {
         protected String[] m_expressionsArray;
 
         protected bool m_throwOnRelative;
-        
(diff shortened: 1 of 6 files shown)


src/mscorlib/src/System/Security/Util/StringExpressionSet.cs
```diff
@@ -33,8 +33,8 @@ internal class StringExpressionSet
         protected String[] m_expressionsArray;
 
         protected bool m_throwOnRelative;
-        
-        protected static readonly char[] m_separators = { ';' };
+
+        private const char Separator = ';';
```
jkotas: BTW: `src/mscorlib/src/System/Security/Util/...` is dead code - it will be deleted as part of the planned cleanups.

jkotas:
Thanks! Could you please port the applicable changes to CoreRT?

PR #9013, opened 2017-01-19
Fix arm32 build break

jashook:
fgClearFinallyTargetBit requires computePreds to have happened in order
to sucessfully leave the bit unset for cases where there are multiple
CALLFINALLY / BBJ_ALWAYS pairs that point to the same finally target on
ARM.

The change adds an assertion to fgClearFinallyTargetBit in addition
to disabling the optimization on ARM.

/cc @AndyAyersMS

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -16910,6 +16910,8 @@ void Compiler::fgMorph()
     fgDebugCheckBBlist(false, false);
 #endif // DEBUG
 
+// RemoveEmptyFinally is disabled on ARM due to github issue #9013
+#ifndef _TARGET_ARM_
     fgRemoveEmptyFinally();
 
     EndPhase(PHASE_EMPTY_FINALLY);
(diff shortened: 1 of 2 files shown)


BruceForstall:
LGTM. Please open an issue, assigned to Andy, to re-enable the optimizations for ARM.

src/jit/morph.cpp
```diff
@@ -16910,13 +16910,15 @@ void Compiler::fgMorph()
     fgDebugCheckBBlist(false, false);
 #endif // DEBUG
 
+#ifndef _TARGET_ARM_
```
CarolEidt: It would be good to have a comment here about why this is not done on ARM.
Is this a legacy-only issue?

jashook: I will add a comment to see #9015

jashook:
The update has no changes that would effect validation. Merging.

src/jit/morph.cpp
```diff
@@ -16910,13 +16910,16 @@ void Compiler::fgMorph()
     fgDebugCheckBBlist(false, false);
 #endif // DEBUG
 
+// RemoveEmptyFinally is disabled on ARM due to github issue #9013
```
CarolEidt: Thanks!

PR #9020, opened 2017-01-20
Add logging for jumpsStubs and ClrVirtualAllocWithinRange

briansull:
Document jumpStub usage in Codeman.cpp
    Added new STRESS_LOG calls that do the following:
     logging for JumpStubs
          log the total count of jump stub lookups: JumpStubLookup
          log the number of actual jump stub thunks created: JumpStubUnique
          log the number of jump stub blocks allocated: JumpStubBlockAllocCount
          log the number jump stub blocks that are completely filled: JumpStubBlockFullCount
     logging for ClrVirtualAllocWithinRange
          log the total number of calls to ClrVirtualAllocWithinRange
          log the number of VirtualQuery operations used by the current call
          log the range that was requested
          log the return value
          log any reasons for failure.

### DIFF
--- a/src/utilcode/util.cpp
+++ b/src/utilcode/util.cpp
@@ -562,6 +562,17 @@ LPVOID ClrVirtualAllocAligned(LPVOID lpAddress, SIZE_T dwSize, DWORD flAllocatio
 #endif // !FEATURE_PAL
 }
 
+#ifdef _DEBUG
+static DWORD ShouldInjectFaultInRange()
+{
+    static DWORD fInjectFaultInRange = 99;
+
+    if (fInjectFaultInRange == 99)
+        fInjectFaultInRange = (CLRConfig::GetConfigValue(CLRConfig::INTERNAL_InjectFault) & 0x40);
+    return fInjectFaultInRange;
+}
+#endif
+
 // Reserves free memory within the range [pMinAddr..pMaxAddr] using
 // ClrVirtualQuery to find free memory and ClrVirtualAlloc to reserve it.
 //
@@ -575,17 +586,6 @@ LPVOID ClrVirtualAllocAligned(LPVOID lpAddress, SIZE_T dwSize, DWORD flAllocatio
 // It returns NULL when it fails to find any memory that satisfies
 // the range.
 //
-
-#ifdef _DEBUG
-static DWORD ShouldInjectFaultInRange()
-{
-    static DWORD fInjectFaultInRange = 99;
-
-    if (fInjectFaultInRange == 99)
-        fInjectFaultInRange = (CLRConfig::GetConfigValue(CLRConfig::INTERNAL_InjectFault) & 0x40);
-    return fInjectFaultInRange;
-}
-#endif
 
 BYTE * ClrVirtualAllocWithinRange(const BYTE *pMinAddr,
                                   const BYTE *pMaxAddr,
@@ -601,7 +601,11 @@ BYTE * ClrVirtualAllocWithinRange(const BYTE *pMinAddr,
     }
     CONTRACTL_END;
 
-    BYTE *pResult = NULL;
+    BYTE *pResult = nullptr;  // our return value;
+
+    static unsigned countOfCalls = 0;  // We log the number of tims we call this method
+    countOfCalls++;                    // increment the call counter
+
     //
     // First lets normalize the pMinAddr and pMaxAddr values
     //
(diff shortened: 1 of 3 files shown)


briansull:
@jkotas PTAL

src/utilcode/util.cpp
```diff
@@ -576,22 +587,11 @@ LPVOID ClrVirtualAllocAligned(LPVOID lpAddress, SIZE_T dwSize, DWORD flAllocatio
...
-    return fInjectFaultInRange;
-}
-#endif
-
 BYTE * ClrVirtualAllocWithinRange(const BYTE *pMinAddr,
-                                  const BYTE *pMaxAddr,
-                                  SIZE_T dwSize, 
-                                  DWORD flAllocationType,
-                                  DWORD flProtect)
+                                   const BYTE *pMaxAddr,
```
jkotas: Nit - this indentation change does not look like an improvement

src/utilcode/util.cpp
```diff
@@ -630,55 +631,88 @@ BYTE * ClrVirtualAllocWithinRange(const BYTE *pMinAddr,
...
-
+        }
+            
         // Is there enough memory free from this start location?
-        // The PAL version of VirtualQuery sets RegionSize to 0 for free
-        // memory regions, in which case we go just ahead and try
-        // VirtualAlloc without checking the size, and see if it succeeds.
-        if (mbInfo.State == MEM_FREE &&
-            (mbInfo.RegionSize >= (SIZE_T) dwSize || mbInfo.RegionSize == 0))
+        if ((mbInfo.State == MEM_FREE) && (mbInfo.RegionSize >= (SIZE_T)dwSize))
```
jkotas: This will break Unix. RegionSize is typically 0 there, and so we never try to allocate...

src/vm/codeman.cpp
```diff
@@ -5062,8 +5143,54 @@ PCODE ExecutionManager::getNextJumpStub(MethodDesc* pMD, PCODE target,
...
+    // Is the 'curBlock' now completely full?
+    if (curBlock->m_used == curBlock->m_allocated)
+    {
+        if (isLCG)
+        {
+            // Increment counter of LCG jump stub blocks that are full
+            m_LCG_JumpStubBlockFullCount++;
+
+            // Log this "LCG JumpStubBlock filled" along with the four counter values
+            STRESS_LOG4(LF_JIT, LL_INFO100, "LCG JumpStubBlock filled - (%u, %u, %u, %u)\n",
```
jkotas: I think this logging level is too low. JITing is level `LL_INFO10000`. I think this should be higher than that.

briansull: Fill a jump stub block is fairly rare. 
But I can increase it a by one digit.
You suggest LL_INFO1000, right?

jkotas: Sounds good.

src/utilcode/util.cpp
briansull: Yeah, I originally made my changes on my Windows TFS enlistment then I copied the change for  this method over.
I will go back and double check for any Linux cahnges to this method.

briansull: Yeah, I originally made my changes on my Windows TFS enlistment then I copied the change for  this method over.
I will go back and double check for any Linux changes to this method.

briansull: I believe that you mean that on UNIX mbInfo.RegionSize is always 0.
Which means that we will always call VirtialAlloc when MEM_FREE is returned.

How do we handle a large request (i.e when dwSize is multiple 64K regions) ??
I guess on UNIX if you end up asking to go into an area that is already reserved  then the call to VirtrualAlloc would fail.

Then we would increase by 64K and keep trying VirtualQuery/VirtualAlloc until we finally got to the area where the memory was allocated.  Could be a bit slow

jkotas: > Which means that we will always call VirtialAlloc when MEM_FREE is returned.

We will never call VirtualAlloc because of `0 >= (SIZE_T)dwSize` will be always false.

briansull: Yes, with the new check we always call VirtualAlloc, without knowing if the size is large enough.

I'm planning on adding a couple of comments to highlight what happens here on UNIX:

```
        // Is there enough memory free from this start location?
        // Note that for UNIX the mbInfo.RegionSize returned will alwys be 0
        if ((mbInfo.State == MEM_FREE) && 
            (mbInfo.RegionSize >= (SIZE_T) dwSize || mdInfo.RegionSize == 0)) 
        {
            // Try reserving the memory using VirtualAlloc now
            pResult = (BYTE*)ClrVirtualAlloc(tryAddr, dwSize, MEM_RESERVE, flProtect);

            // Normally this will be successful
            //
            if (pResult != nullptr)
            {
                success = true;
                // return pResult
                break;
            }

#ifdef _DEBUG
            if (ShouldInjectFaultInRange())
            {
                // return nullptr (failure)
                faultInjected = true;
                break;
            }
#endif // _DEBUG

            // On UNIX we will also fail if our request size 'dwSize' is larger than 64K and
            // and our tryAddr is pointing at a small MEM_FREE region (smaller than 'dwSize')
            // However we can't distinguish between this and the race case.

            // We might fail in a race.  So just move on to next region and continue trying
            tryAddr = tryAddr + VIRTUAL_ALLOC_RESERVE_GRANULARITY;
        }
```

briansull:

briansull:

PR #9022, opened 2017-01-20
[GDB-JIT][Linux] Fix incorrect displaying of (s)byte and char in lldb

lucenticus:
This PR adds typedef for `byte`, `sbyte` and `char` C# types to fix incorrect displaying of these types in lldb.
Without this fix lldb trying to convert DWARF types to C++ basic types and shows `byte` as `unsigned char`, `sbyte` as `char`, and `char` as `short`.
@janvorli, @mikem8361 PTAL

\CC: @Dmitri-Botcharnikov @chunseoklee

### DIFF
--- a/src/vm/gdbjit.cpp
+++ b/src/vm/gdbjit.cpp
@@ -32,6 +32,7 @@ GetTypeInfoFromTypeHandle(TypeHandle typeHandle, NotifyGdb::PTK_TypeInfoMap pTyp
     {
         case ELEMENT_TYPE_I1:
         case ELEMENT_TYPE_U1:
+        case ELEMENT_TYPE_CHAR:
             typeInfo = new (nothrow) ByteTypeInfo(typeHandle, CorElementTypeToDWEncoding[corType]);
             if (typeInfo == nullptr)
                 return nullptr;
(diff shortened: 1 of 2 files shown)


lucenticus:
@dotnet-bot  test Ubuntu x64 Checked Build and Test please

src/vm/gdbjit.cpp
```diff
@@ -964,6 +975,56 @@ TypeKey* TypeInfoBase::GetTypeKey()
...
+
+void ByteTypeInfo::DumpStrings(char* ptr, int& offset)
+{
+    PrimitiveTypeInfo::DumpStrings(ptr, offset);
+    m_typedef_info->m_typedef_name = new (nothrow) char[strlen(m_type_name) + 1];
+    if (strcmp(m_type_name, "System.Byte") == 0)
+        strcpy(m_typedef_info->m_typedef_name, "byte");
+    else if (strcmp(m_type_name, "System.SByte") == 0)
+        strcpy(m_typedef_info->m_typedef_name, "sbyte");
+    else if (strcmp(m_type_name, "char16_t") == 0)
```
janvorli: @lucenticus Could you please explain to me these mappings? First it seemed like it is mapping the CLR type names to C# names, but then I don't understand this char16_t mapping. char16_t is not a CLR type, right?

lucenticus: Yes, you're right. We use `char16_t` to show unicode characters in lldb correctly. You might see that we renamed `System.Char` to `char16_t` here  [gdbjit.cpp:205](https://github.com/lucenticus/coreclr/blob/a6266a598d3893bb76acbb544ff092ea3da8363a/src/vm/gdbjit.cpp#L205). These hacks are needed because DWARF standard doesn't support C# language, and lldb doesn't support C# type system.

janvorli: Ok, thank you for the explanation.

lucenticus:
Thank you!

Issue #9011, opened 2017-01-19
Exponential slowdown in task execution when using WhenAny


yosefd:
@kouvel @alexperovich @stephentoub

The bug we found causes exponential slowdown, so it is pretty severe. I wouldn’t classify it as a performance problem. This practically kills process for all practical purposes, which what happened to our product.  As far as I can see in sources, the bug is in all versions of .net.

This is the repro (run release version, not attached to debugger):

            // Simulate slow operation.
            var completionForOperationThatWillTakeLotsOfTime = new TaskCompletionSource<object>();

            Func<Task> main = async () =>
            {
                for (int i = 0; i < 100; i++)
                {
                    var startBatch = DateTime.UtcNow;
                    for (int j = 0; j < 10000; j++)
                    {
                        // Execute fast operation in a loop, but wait in parallel if slow operation is eventually completed (which btw will never happen)
                        var taskFastOperation = Task.Run(() => { });
                        var taskAny = await Task.WhenAny(taskFastOperation, completionForOperationThatWillTakeLotsOfTime.Task);
                        if (taskAny != taskFastOperation)
                        {
                            // Slow operation should never complete as the task completion source for it is never completed.
                            Console.WriteLine("Unexpected");
                            Console.Read();
                        }
                        // Fast operation completed. Can take the result now.
                        await taskFastOperation;
                    }
                    Console.WriteLine("Batch execution time: {0} ms", (DateTime.UtcNow - startBatch).TotalMilliseconds);
                }
            };

            var start = DateTime.UtcNow;

            main().Wait();

            Console.WriteLine("Execution time: {0} ms", (DateTime.UtcNow - start).TotalMilliseconds);
            Console.Read();

Each batch runs 10000 iterations. Each iteration awaits one of  2 tasks, one of which completes fast, since it do almost nothing (Task.Run(()=>{})). Upon completion of the operation, new iteration starts.

You would expect each batch to run the same time. In reality first few batches are very fast, under 50ms, the following batches become very slow and can run long minutes, eventually. This is output of the code above after a few minutes of execution:

Batch execution time: 30.0168 ms
Batch execution time: 21.8036 ms
Batch execution time: 70.5011 ms
Batch execution time: 359.5303 ms
Batch execution time: 520.5175 ms
Batch execution time: 610.5054 ms
Batch execution time: 1188.5082 ms
Batch execution time: 1477.0104 ms
Batch execution time: 2279.5154 ms
Batch execution time: 2013.0131 ms
[...]

PR #9024, opened 2017-01-20
Fix race condition in Task.WhenAny

stephentoub:
A race condition in the implementation of Task.WhenAny can result in a continuation being left registered with one of the constituent tasks even after the WhenAny operation has completed.  If the same task is then repeatedly used in a WhenAny, each time this manifests it'll result in another continuation being leaked into that task, which can add up over time.  That leak can then also manifest as throughput hits, as removing subsequent continuations results in searching the longer and longer list for the target continuation to remove.

WhenAny creates a promise continuation object which it then registers as a continuation for each involved task.  The first task to complete will invoke that continuation, which will remove the registered continuation from all of the tasks it's been registered with thus far.  And the main loop doing the registering will stop registering when any task has completed.  But if there's a race between adding the continuation to a task and the continuation removing itself from all previous tasks, we can end up not removing it from the last task.  The fix is simply to check after adding a continuation whether the promise has now completed, and if it has then manually try to remove the continuation.

Fixes https://github.com/dotnet/coreclr/issues/9011
cc: @kouvel, @yosefd, @AlfredoMS

### DIFF
--- a/src/mscorlib/src/System/Threading/Tasks/TaskFactory.cs
+++ b/src/mscorlib/src/System/Threading/Tasks/TaskFactory.cs
@@ -2505,7 +2505,23 @@ namespace System.Threading.Tasks
                     checkArgsOnly = true;
                 }
                 // Otherwise, add the completion action and keep going.
-                else task.AddCompletionAction(promise);
+                else
+                {
+                    task.AddCompletionAction(promise);
+                    if (promise.IsCompleted)
+                    {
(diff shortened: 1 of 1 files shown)


AlfredoMS:
LGTM
cc: @alexperovich

stephentoub:
@dotnet-bot test Windows_NT x64 Release Priority 1 Build and Test please (https://github.com/dotnet/coreclr/issues/9107)

PR #9027, opened 2017-01-20
Fix incorrect comment

JosephTremoulet:
Rewrite bogus reference to nonexistsent `m_HeapDef` to correctly refer to
`block->bbHeapDef`.  Also change "doesn't kill heap" to "doesn't kill heap
liveness" (since it still kills availability).

### DIFF
--- a/src/jit/block.h
+++ b/src/jit/block.h
@@ -809,8 +809,8 @@ struct BasicBlock : private LIR::Range
     VARSET_TP bbLiveOut; // variables live on exit
 
     // Use, def, live in/out information for the implicit "Heap" variable.
-    unsigned bbHeapUse : 1;
-    unsigned bbHeapDef : 1;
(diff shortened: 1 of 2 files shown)


JosephTremoulet:
@briansull PTAL; incorporated your feedback from #9004.

briansull:
The definitions for these one bit flags should probably be updated:

    // Use, def, live in/out information for the implicit "Heap" variable.
    unsigned bbHeapUse : 1;      // must be set to true for any block that references the global Heap
    unsigned bbHeapDef : 1;      // must be set to true for any block that Mutates the global Heap
    unsigned bbHeapLiveIn : 1;
    unsigned bbHeapLiveOut : 1;

JosephTremoulet:
Makes sense.  Updated.

briansull:
LGTM

PR #9028, opened 2017-01-20
Fixup Span<T> implementation to be more similar to the corefx one

jkotas:
- Reorder methods, adjust comments and formatting to make it easier to see the diff with corefx
- Delete extra implicit conversion operator

### DIFF
--- a/src/mscorlib/model.xml
+++ b/src/mscorlib/model.xml
@@ -12677,7 +12677,6 @@
       <Member Name="#ctor(T[],System.Int32,System.Int32)" />
       <Member Name="#ctor(System.Void*,System.Int32)" />
       <Member Name="DangerousGetPinnableReference" />
-      <Member Name="op_Implicit(System.Span&lt;T&gt;)" ReturnType="System.ReadOnlySpan&lt;T&gt;" />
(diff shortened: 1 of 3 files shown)


jkotas:
@ahsonkhan @AtsushiKan @KrzysztofCwalina

AtsushiKan:
The Span.Span(T[], int, int) overload still has the K&R-style "if (default(T)==null)" check that doesn't diff with the corefx one.

AtsushiKan:
Span.TryCopyTo()
 if ((uint)_length > (uint)destination.Length)

Changing "destination.Length" to "destination._length" would make it diff perfectly.

AtsushiKan:
Span.DangerousGetPinnableReference() is in a pretty different place in the two files.

AtsushiKan:
LGTM modulo comments. (Which probably apply to ReadOnlySpan, though I didn't check each other.)

jkotas:
> The Span.Span(T[], int, int) overload still has the K&R-style

Fixed.

> if ((uint)_length > (uint)destination.Length)

All other places in the implementation are using `_length` or `span._length` directly, so I have left this one as is for internal consistency. There are a few other similar nits that did not look like an improvement, so I have left them as is. It may be nice to fix them in the slow span instead.

PR #9029, opened 2017-01-20
[Local GC] Add three finalization-related operations to GCToEEInterface

swgillespie:
This PR removes three references to EE-private symbols from the GC and replaces them with calls to `GCToEEInterface`. The three symbols are:

* `FinalizerThread::HaveExtraWorkForFinalizer`
* `FinalizerThread::EnableFinalization`
* `g_fFinalizerRunOnShutDown`

These three symbols are defined on the EE side of the interface and so it's appropriate to interact with them using `GCToEEInterface` instead of referring to them directly.


swgillespie:

src/gc/gcinterface.ee.h
```diff
@@ -128,6 +128,19 @@ class IGCToCLR {
...
+    // be finalized.
+    virtual
+    void EnableFinalization() = 0;
+
+    // Returns true if there is additional work that the finalizer thread needs
+    // to do, outside of its duties of finalization.
+    virtual
+    bool HaveExtraWorkForFinalizer() = 0;
+
+    virtual
```
swgillespie: I'm not familiar with this symbol I'm not sure of an appropriate comment for this method - does "g_fFinalizeRunOnShutDown" refer to finalizing objects during EE shutdown?

swgillespie:

src/gc/gcinterface.ee.h
jkotas: This flag says whether the GC should be reporting all objects with finalizers; not just the collected ones. The `g_fFinalizeRunOnShutDown` check is somewhat warm paths - `CFinalize::RegisterForFinalization`, `CFinalize::GetNextFinalizableObject`. I do not think it should be callback to VM. Instead, the flag should stay on the GC side so that the check continues to be fast and there should be a method on JIT/EE interface to set it. 

Note that you should not ever see this flag set in CoreCLR. It is only used in full Framework.

swgillespie: Oh, ok! Will do!

> there should be a method on JIT/EE interface to set it.

Is this the GC/EE interface? The only place I see this being set in the full framework codebase is in the FinalizerThreadWatchDog (https://github.com/dotnet/coreclr/blob/master/src/vm/finalizerthread.cpp#L1181). Is there somewhere else I'm missing?

src/gc/gcinterface.ee.h
```diff
@@ -128,6 +128,19 @@ class IGCToCLR {
...
+
+    // Signals to the finalizer thread that there are objects ready to
+    // be finalized.
+    virtual
+    void EnableFinalization() = 0;
+
+    // Returns true if there is additional work that the finalizer thread needs
+    // to do, outside of its duties of finalization.
+    virtual
+    bool HaveExtraWorkForFinalizer() = 0;
```
jkotas: The GC is not managing the finalizer thread, and so it is odd to have method on the GC/EE interface that returns whether there is more work for the finalizer thread. It also makes the interface more chatty than necessary, though it does not really matter in this case because of it is not a hot path.

Instead, I would introduce `bool` argument on the EnableFinalization method. The VM implementation of this method would be:

```
void EnableFinalization(bool foundFinalizers)
{
     if (foundFinalizers || HaveExtraWorkForFinalizer()) 
        hEventFinalizer->Set();
}
```

swgillespie: Will do!

jkotas:
LGTM otherwise. Thanks!

src/gc/gcinterface.ee.h
jkotas: Right, it is the only place

jkotas: > Is this the GC/EE interface?

I meant `IGCHeap` in `gcinterface.h`.

swgillespie: Ok, will do!

swgillespie:
Thanks!

cc @Maoni0 @sergiy-k @adityamandaleeka , PTAL also? I'll be addressing Jan's feedback in just a minute.

Maoni0:
LGTM (agree with Jan's comments)

swgillespie:

jkotas:
Could you please delete g_fFinalizerRunOnShutDown from the VM side because of it is no longer used there?

```
\coreclr\src\vm\crst.h(108):extern bool g_fFinalizerRunOnShutDown;
\coreclr\src\vm\vars.cpp(244):bool g_fFinalizerRunOnShutDown = false;
\coreclr\src\vm\vars.hpp(595):EXTERN bool g_fFinalizerRunOnShutDown
```

src/vm/gcenv.ee.cpp
```diff
@@ -1308,3 +1308,11 @@ void GCToEEInterface::StompWriteBarrier(WriteBarrierParameters* args)
...
     }
 }
+
+void GCToEEInterface::EnableFinalization(bool foundFinalizers)
+{
+    if (foundFinalizers || FinalizerThread::HaveExtraWorkForFinalizer())
+    {
+        FinalizerThread::EnableFinalization();
+    }
+}
```
jkotas: Nit - end the file with new line

swgillespie:

swgillespie:

swgillespie:
Not sure where these test failures came from - will try and repro them locally.

swgillespie:

swgillespie:

swgillespie:
I couldn't repro any of the failures locally and they passed on a re-run, so I'm going to merge this.

Issue #9033, opened 2017-01-21
Span: VM side of GC reporting for `Span<T>` - investigate struct passing in regs on Unix


kouvel:
Investigate if there is an issue with structs being passed in registers on Unix:
- Passing a span as an argument in registers (potentially fix ReportPointersFromStructInRegisters to cover ByRefLike types)
- Returning a span in registers during GC suspension

kouvel:
Looks like returning in registers during GC suspension is not an issue

PR #9034, opened 2017-01-21
Report by-ref-like types to the GC (SpanOfT)

### LINKED ISSUE #8517: Span: VM side of GC reporting for `Span<T>`
Fix the following places on the VM side to do proper GC reporting for `Span<T>`:
- ReportPointersFromValueType

kouvel:
Fixes #8517

### DIFF
--- a/src/vm/siginfo.cpp
+++ b/src/vm/siginfo.cpp
@@ -4959,7 +4959,7 @@ void ReportByRefPointersFromByRefLikeObject(promote_func *fn, ScanContext *sc, P
 
     _ASSERTE(pMT->IsByRefLike());
 
-    // TODO: TypedReference should ideally be implemneted as a by-ref-like struct containing a ByReference<T> field,
+    // TODO: TypedReference should ideally be implemented as a by-ref-like struct containing a ByReference<T> field,
     // in which case the check for g_TypedReferenceMT below would not be necessary
     if (pMT == g_TypedReferenceMT || pMT->HasSameTypeDefAs(g_pByReferenceClass))
     {
--- a/src/vm/siginfo.cpp
+++ b/src/vm/siginfo.cpp
@@ -4959,12 +4959,9 @@ void ReportByRefPointersFromByRefLikeObject(promote_func *fn, ScanContext *sc, P
 
     _ASSERTE(pMT->IsByRefLike());
 
-    if (pMT == g_TypedReferenceMT)
-    {
-        return;
-    }
-
-    if (pMT->HasSameTypeDefAs(g_pByReferenceClass))
+    // TODO: TypedReference should ideally be implemneted as a by-ref-like struct containing a ByReference<T> field,
+    // in which case the check for g_TypedReferenceMT below would not be necessary
+    if (pMT == g_TypedReferenceMT || pMT->HasSameTypeDefAs(g_pByReferenceClass))
     {
         (*fn)(dac_cast<PTR_PTR_Object>(pSrc), sc, GC_CALL_INTERIOR);
         return;
--- a/src/vm/siginfo.cpp
+++ b/src/vm/siginfo.cpp
@@ -4953,11 +4953,54 @@ void PromoteCarefully(promote_func   fn,
     (*fn) (ppObj, sc, flags);
 }
 
+void ReportByRefPointersFromByRefLikeObject(promote_func *fn, ScanContext *sc, PTR_MethodTable pMT, PTR_VOID pSrc)
+{
+    WRAPPER_NO_CONTRACT;
+
+    _ASSERTE(pMT->IsByRefLike());
+
+    if (pMT == g_TypedReferenceMT)
+    {
+        return;
+    }
+
+    if (pMT->HasSameTypeDefAs(g_pByReferenceClass))
+    {
+        (*fn)(dac_cast<PTR_PTR_Object>(pSrc), sc, GC_CALL_INTERIOR);
+        return;
+    }
+
+    ApproxFieldDescIterator fieldIterator(pMT, ApproxFieldDescIterator::INSTANCE_FIELDS);
+    for (FieldDesc *pFD = fieldIterator.Next(); pFD != NULL; pFD = fieldIterator.Next())
+    {
+        if (pFD->GetFieldType() != ELEMENT_TYPE_VALUETYPE)
+        {
+            continue;
+        }
+
+        // TODO: GetApproxFieldTypeHandleThrowing may throw. This is a potential stress problem for fragile NGen of non-CoreLib
+        // assemblies. It won’t ever throw for CoreCLR with R2R. Figure out if anything needs to be done to deal with the
+        // exception.
+        PTR_MethodTable pFieldMT = pFD->GetApproxFieldTypeHandleThrowing().AsMethodTable();
+        if (!pFieldMT->IsByRefLike())
+        {
+            continue;
+        }
+
+        int fieldStartIndex = pFD->GetOffset() / sizeof(void *);
+        PTR_PTR_Object fieldRef = dac_cast<PTR_PTR_Object>(PTR_BYTE(pSrc) + fieldStartIndex);
+        ReportByRefPointersFromByRefLikeObject(fn, sc, pFieldMT, fieldRef);
+    }
+}
+
 void ReportPointersFromValueType(promote_func *fn, ScanContext *sc, PTR_MethodTable pMT, PTR_VOID pSrc)
 {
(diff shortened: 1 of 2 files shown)


kouvel:
@jkotas

src/vm/siginfo.cpp
```diff
@@ -4953,11 +4953,54 @@ void PromoteCarefully(promote_func   fn,
     (*fn) (ppObj, sc, flags);
 }
 
+void ReportByRefPointersFromByRefLikeObject(promote_func *fn, ScanContext *sc, PTR_MethodTable pMT, PTR_VOID pSrc)
+{
+    WRAPPER_NO_CONTRACT;
+
+    _ASSERTE(pMT->IsByRefLike());
+
+    if (pMT == g_TypedReferenceMT)
```
jkotas: This should do the same thing as `g_pByReferenceClass`:

`if (pMT->HasSameTypeDefAs(g_pByReferenceClass) || pMT == g_TypedReferenceMT)`

src/vm/siginfo.cpp
```diff
@@ -4987,16 +5030,21 @@ void ReportPointersFromValueType(promote_func *fn, ScanContext *sc, PTR_MethodTa
 void ReportPointersFromValueTypeArg(promote_func *fn, ScanContext *sc, PTR_MethodTable pMT, ArgDestination *pSrc)
 {
     WRAPPER_NO_CONTRACT;
-    
-    if (!pMT->ContainsPointers())
-        return;
+
+    if(pMT->ContainsPointers())
```
jkotas: Why can't this be just

```
   if (!pMT->ContainsPointers() && !pMT->IsByRefLike())
       return;

   ... existing code...
```

?

jkotas: (I would be useful to add some comments if it needs to be what you have written.)

src/vm/siginfo.cpp
kouvel: I don't understand, could you please elaborate?

src/vm/siginfo.cpp
kouvel: As I currently do not have a complete understanding of GC holes with respect to Unix and structs passed in registers, I'm conservatively preserving the previous behavior by calling `ReportPointersFromStructInRegisters` only under the previous conditions. #9033 is intended to track potential issues on Unix, a fix for which should hopefully resolve the question.

src/vm/siginfo.cpp
jkotas: Once the rules for byref-like type in C# compiler are relaxed, it should be possible to write:

```
struct MyStruct
{
    TypedReference field1;
    Span<byte> field2;
    TypedReference field3;
}
```

jkotas: `field1` and `field3` won't be reported correctly because of the early out.

jkotas: `CEEInfo::getClassGClayout` has the same bug

kouvel: It looks to me like outermost call to `ReportByRefPointersFromByRefLikeObject` would iterate over the first field and recursively call itself recursively for `field1`, where it would would return because it's a `TypedReference`. The outermost call to `ReportByRefPointersFromByRefLikeObject` would then recursively call itself for `field2`, where it would report the `ByReference` inside the `Span<T>`. Then the outermost call to `ReportByRefPointersFromByRefLikeObject` would recursively call itself for `field3`, which would return because it's a TypedReference. Did I miss something?

kouvel: It looks to me like `CEEInfo::getClassGClayout` would only return early if the topmost object is a `TypedReference`, which seems acceptable. If a by-ref-like object contains a `TypedReference` as in your example, it would call `ComputeGCLayout`, which would iterate over all fields in the by-ref-like type. I'm not seeing an issue there either.

jkotas: The first field of TypedReference is actually byref, but it is not recorded as byref in metadata. It has been special cased for GC reporting in multiple places. This is yet another places that should be special cased for it.

Ideally, we would change the implementation of TypedReference to:

```
struct TypedReference
{
    ByReference<byte> _value;
    IntPtr _type;
...
}
```

and get rid of the special casing. But until/if that happens, we should continue special casing it.

kouvel: If `TypedReference` is not recorded as byref, then it seems to me like it would have already had an issue before this change - it would not have been reported properly, as there are direct calls to `ReportPointersFromValueType` that bypasses `ReportPointersFromValueTypeArg`. Are you suggesting that there is an issue with the reporting for `TypedReference` as well? If so, I'll need to test with `TypedReference` as well to see the problem before fixing it.

jkotas: The TypedReference has not been allowed as a field and so there is no problem today. They have been only allowed as regular arguments. We should not come for regular TypedReference arguments at all because of they are special cased above ReportPointersFromValueType.

Since you have added the special case for TypedReference here, I thought you want to prepare this code for case where TypedReference are allowed as fields as other byref-like types. There is a good chance that TypedReferences are going to be allowed as fields as natural fallout from the relaxations of rules for byref-like types in C#.

Either of these two options would sound reasonable to me:
- Do not prepare this for TypedReference fields - do not handle TypedReference at all here
- Prepare this for TypedReference fields - handle TypedReference same way as `ByReference<T>` here

kouvel: Ok how does 797ec397c5b1db784c137bb4a0938e17be449df7 look? I kept the check considering TypedReference is currently not implemented the way in which you suggest, but the new code treats it similarly to ByReferece in the sense the first field of both is the pointer, which should conform to the way in which TypedReference is currently implemented.

src/vm/siginfo.cpp
```diff
@@ -4953,11 +4953,51 @@ void PromoteCarefully(promote_func   fn,
     (*fn) (ppObj, sc, flags);
 }
 
+void ReportByRefPointersFromByRefLikeObject(promote_func *fn, ScanContext *sc, PTR_MethodTable pMT, PTR_VOID pSrc)
+{
+    WRAPPER_NO_CONTRACT;
+
+    _ASSERTE(pMT->IsByRefLike());
+
+    // TODO: TypedReference should ideally be implemneted as a by-ref-like struct containing a ByReference<T> field,
```
jkotas: Nit: typo implemneted

jkotas:
It would be nice to make a matching change in `CEEInfo::getClassGClayout` in vm\jitinterface.cpp so that it is sync. Move the special case for `g_TypedReferenceMT` from `CEEInfo::getClassGClayout` to `ComputeGCLayout`. Something like:

```
static unsigned ComputeGCLayout(MethodTable * pMT, BYTE* gcPtrs)
{
    STANDARD_VM_CONTRACT;

    unsigned result = 0;

    _ASSERTE(pMT->IsValueType());

    // TODO: TypedReference should ideally be implemented as a by-ref-like struct containing a ByReference<T> field,  
    // in which case the check for g_TypedReferenceMT below would not be necessary     if (pMT == g_TypedReferenceMT)
    {
        if (gcPtrs[0] == TYPE_GC_NONE)
        {
            gcPtrs[0] = TYPE_GC_BYREF;
            result++;
        }
        else if (gcPtrs[fieldStartIndex] != TYPE_GC_BYREF)
        {
            COMPlusThrowHR(COR_E_BADIMAGEFORMAT);
        }
        return 1;
    }

    ... existing code ...
}
```

jkotas:
LGTM otherwise. Thanks!

src/vm/siginfo.cpp
kouvel: Fixed

kouvel:
> `if (gcPtrs[0] == TYPE_GC_NONE)`

It looks like `getClassGClayout` memsets it to TYPE_GC_NONE but the recursive call from `ComputeGCLayout` does not, so I have omitted this part.

> `else if (gcPtrs[fieldStartIndex] != TYPE_GC_BYREF)`

I don't see where this array is initialized to some default value, so I have omitted this part as well. Simplified to:

```c++
    if (pMT == g_TypedReferenceMT)
    {
        gcPtrs[0] = TYPE_GC_BYREF;
        gcPtrs[1] = TYPE_GC_NONE;
        return 1;
    }
```

And added the comment to `getClassGClayout` as well, which if/when changed, would make the addition to `ComputeGCLayout` redundant.

kouvel:
@dotnet-bot test this please

kouvel:
@dotnet-bot test this please

PR #6088, opened 2016-07-01
FIX #6074 : cannot find exception handler on arm

chunseoklee:
This patch fix issue #6074.

PR #4503, opened 2016-04-22
Fix #4496

parjong:
This commit revises legecy codegen to insert a bkpt instruction if the current BB
and next BB is not in the same exception handling region.

This code is backported from ARM64 codegen to fix #4496.

PR #9043, opened 2017-01-23
[x86/Linux] Fix EH region mismatch

### LINKED ISSUE #9042: [x86/Linux] EH Region Mismatch
The following program is expected to print "Hello, World":
```
using System;

namespace ConsoleApplication { public class Program {

  public static int Throw(int n) {
     return 16 / n;
  }

  public static void Main(string[] args) {
    try
    {
      Throw(0);
    }
    catch (Exception)
    {
      Console.WriteLine("Hello World!");
    }
  }

} }
```

In x86/Linux, however, Core CLR shows the following message:
```
Unhandled Exception: System.DivideByZeroException: Attempted to divide by zero.
   at ConsoleApplication.Program.Throw(Int32 n) in /home/parjong/projects/dotnet/coreclr-analyze/eh-region/src/Program.cs:line 8
   at ConsoleApplication.Program.Main(String[] args) in /home/parjong/projects/dotnet/coreclr-analyze/eh-region/src/Program.cs:line 14
```

parjong:
This commit ports ARM EH region mismatch fix (#6088) into x86 in order to fix #9042.

### DIFF
--- a/src/pal/inc/pal.h
+++ b/src/pal/inc/pal.h
@@ -1726,6 +1726,15 @@ QueueUserAPC(
 #define CONTEXT_EXCEPTION_REQUEST 0x40000000L
 #define CONTEXT_EXCEPTION_REPORTING 0x80000000L
 
+//
+// This flag is set by the unwinder if it has unwound to a call
+// site, and cleared whenever it unwinds through a trap frame.
+// It is used by language-specific exception handlers to help
+// differentiate exception scopes during dispatching.
+//
+
+#define CONTEXT_UNWOUND_TO_CALL 0x20000000
+
 typedef struct _FLOATING_SAVE_AREA {
     DWORD   ControlWord;
     DWORD   StatusWord;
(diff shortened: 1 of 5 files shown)


parjong:
\CC @seanshpark @wateret

parjong:

parjong:
@janvorli Please take a look.

jkotas:
@parjong This one has a conflict now...

parjong:

janvorli:
@jkotas I wonder why we don't have to do the same thing for AMD64. Is the jitter codegen different in this respect?

parjong:
@janvorli In #4503, @BruceForstall mentioned that OS Unwinder (available in Windows) subtracts 2 when it unwinds to call frame. As I remember, ARM64 (and AMD64?) jitter appends a dummy instruction if call is the last instruction of a basic block and that basic block and the next basic block belong to different eh regions.

Issue #8883, opened 2017-01-11
Remove unnecessary defines from CoreCLR CoreLib


danmosemsft:
The following are always defined when building the CoreCLR Corelib. Since we are not maintaining multiple builds, we should not need the defines in the code (they are always true).

- [x] FEATURE_APPDOMAIN_RESOURCEMONITORING
- [x] FEATURE_COLLECTIBLE_TYPES
- [x] FEATURE_CORECLR
- [x] FEATURE_CORESYSTEM
- [x] FEATURE_CORRUPTING_EXCEPTIONS
- [x] FEATURE_EXCEPTIONDISPATCHINFO
- [x] FEATURE_EXCEPTION_NOTIFICATIONS
- [x] FEATURE_HOST_ASSEMBLY_RESOLVER
- [x] FEATURE_LOADER_OPTIMIZATION
- [x] FEATURE_MAIN_CLR_MODULE_USES_CORE_NAME
- [x] FEATURE_MANAGED_ETW_CHANNELS
- [x] FEATURE_MULTICOREJIT
- [x] FEATURE_SPAN_OF_T
- [x] FEATURE_STRONGNAME_MIGRATION
- [x] FEATURE_SYNCHRONIZATIONCONTEXT_WAIT
- [x] FEATURE_VERSIONING
- [x] FEATURE_WINDOWSPHONE

Partially keep - break out into another issue:
- [ ] FEATURE_RANDOMIZED_STRING_HASHING

@jkotas why do we define the COMINTEROP symbols, when we don't support COM interop? 

Also, are there any that are valuable to retain for potential future use, eg., private builds or other work?

Their entries can also be removed from 

- [x] https://github.com/dotnet/coreclr/blob/master/clr.coreclr.props

As a subseqnet workitem, they could be then removed from the native side also, both code and their several cmake files.

danmosemsft:
I need to double check this list on a Unix build.

danmosemsft:
These ones are also present in CoreRT so presumably could be cut from there

```
FEATURE_CORECLR
FEATURE_RANDOMIZED_STRING_HASHING
```

jkotas:
> when we don't support COM interop

We do support a good chunk of COM interop on Windows (except COM interop activation and IDispatch).

jkotas:
> Also, are there any that are valuable to retain for potential future use, eg., private builds or other work?

I do not think so. 

[...]

PR #9044, opened 2017-01-23
Remove always-defined symbols

danmosemsft:
These are always defined in the CoreCLR build, and now we no longer want to preserve mergeability with desktop, we can remove them.

Fixes https://github.com/dotnet/coreclr/issues/8883

Verification in progress: diff decompilation for each architecture.


danmosemsft:
Verified by building before and after and decompiling then diffing the code across Windows debug, Windows x86, Linux debug, Linux release. The only diff in each case is the dead string I removed.

danmosemsft:
@jkotas could you or someone review?

jkotas:
FeatureAppdomainResourceMonitoring should be deleted & disabled instead.

jkotas:
LGTM. Thanks for getting this started!

BTW: I have opened the following issues related to the NetStandard2.0 work as I was reviewing this:

https://github.com/dotnet/coreclr/issues/9045
https://github.com/dotnet/coreclr/issues/9047

cc @gkhanna79 @rahku

hughbe:
Can you clarify why we've abandoned mergeability with Desktop? Thanks

jkotas:
It is not completely abandoned, e.g. it is still maintained for some parts like JIT or GC. The corelib is on the port the selected fixes plan - same as corefx, for the same reasons as corefx - the .NET Core corelib is too different from the desktop one.

danmosemsft:
OK, removed and disabled FEATURE_APPDOMAIN_RESOURCE_MONITORING. This is the only change affecting the final binary (save removing a string)

danmosemsft:
@jkotas merged, but just a note to confirm that you meant me to remove FEATURE_APPDOMAIN_RESOURCE_MONITORING (it was previously enabled)

jkotas:
@danmosemsft Yes, looks good.

PR #9046, opened 2017-01-23
[x86/Linux] Port two assertions in stackwalk.cpp

parjong:
There two assertions are ifdefed by _TARGET_X86_, but seems to be related with WIN64EXCEPTIONS.

This PR ports these assertions into x86/Linux.


parjong:
\CC @seanshpark @wateret

PR #9048, opened 2017-01-23
Clean up #else in EECodeManager::FixContext

parjong:
It seems that only x86 ports uses EECodeManage::FixContext, but EECodeManage::FixContext contains unnecessary #else region in it.

This commit cleans up unnecessary #else region from EECodeManage::FixContext.

### DIFF
--- a/src/inc/eetwain.h
+++ b/src/inc/eetwain.h
@@ -161,7 +161,7 @@ enum
 };
 
 #ifndef DACCESS_COMPILE
-#ifdef _TARGET_X86_
+#ifndef WIN64EXCEPTIONS
 virtual void FixContext(ContextType     ctxType,
                         EHContext      *ctx,
                         EECodeInfo     *pCodeInfo,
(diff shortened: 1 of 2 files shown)


parjong:

src/inc/eetwain.h
```diff
@@ -161,7 +161,7 @@ enum
 };
 
 #ifndef DACCESS_COMPILE
-
+#ifdef _TARGET_X86_
```
jkotas: Should this rather be WIN64EXCEPTIONS like in the other change?

parjong: @jkotas The existing implementation uses ``_TARGET_X86_``, and thus I uses ``_TARGET_X86_`` here. Is this related with ``WIN64EXCEPTIONS``?? If so, it would be better to use ``WIN64EXCEPTIONS`` as you mentioned.

jkotas: This method does not make sense for !WIN64EXCEPTIONS . I think it would more appropriate to put it under `#ifndef WIN64EXCEPTIONS`.

parjong: Do you mean that this method is related with ``WIN64EXCEPTIONS`` (not ``_TARGET_X86_``)?

jkotas: Correct. It is only called in src/vm/i386/excepx86.cpp that is only included for !WIN64EXCEPTIONS. You should not need it for Linux/x86.

parjong:

PR #9049, opened 2017-01-23
Clean up #else in EECodeManager::GetAmbientSP

parjong:
Only x86 ports uses EECodeManager::GetAmbientSP, but it contains unnecessary #else region for other architectures.

This commit cleans up unnecessary #else region from EECodeManager::GetAmbientSP.


parjong:

PR #9050, opened 2017-01-23
Clean up LeaveFinally/LeaveCatch

parjong:
LeaveFinally and LeaveCatch are used only when WIN64EXCEPTIONS is not set ( x86/Windows is currently the only port that uses these methods).

This commit hides the declarations of these methods when WIN64EXCEPTIONS is set, and removes unnecessary #else region from the implementation of these methods.

### DIFF
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -5729,6 +5729,7 @@ BOOL EECodeManager::IsInFilter(GCInfoToken gcInfoToken,
 }
 
 
+#if defined(_TARGET_X86_) && !defined(WIN64EXCEPTIONS)
 BOOL EECodeManager::LeaveFinally(GCInfoToken gcInfoToken,
                                 unsigned offset,
                                 PCONTEXT pCtx)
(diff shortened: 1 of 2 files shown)


parjong:

src/vm/eetwain.cpp
```diff
@@ -5729,6 +5729,7 @@ BOOL EECodeManager::IsInFilter(GCInfoToken gcInfoToken,
 }
 
 
+#if defined(_TARGET_X86_) && !defined(WIN64EXCEPTIONS)
```
jkotas: Shouldn't this be just `#ifndef WIN64EXCEPTIONS` to match what's in the header file?

parjong: As Esp is x86-specific, and thus I used _TARGET_X86_ && !WIN64EXCEPTIONS here.

Is it better to revise this as follows?
```
#ifndef WIN64EXCEPTIONS
#ifdef _TARGET_X86_
~~~
#else
#error "Unsupported"
#endif
#else
~~~
#endif
```

jkotas: These methods are specific to !WIN64EXCEPTIONS exception handling scheme. So I would just make the ifdef consistent with header file here.

In the unlikely case somebody tries to build say `<some new platform> + !WIN64EXCEPTIONS` combination, they will get a ton of build breaks. I do not think we need to worry about it.

parjong:

PR #9052, opened 2017-01-23
Simplify #ifdef for UnwindStackFrame

parjong:
The current implementation has the following #ifdef structure:
```
#ifdef _TARGET_X86_
~~~ 
#elseif WIN64EXCEPTIONS && !CROSSGEN_COMPILE
~~~
#else 
~~~
#endif 
```

This commit attempts to simplify the above structure as follows:
```
#ifndef CROSSGEN_COMPILE
#ifndef WIN64EXCEPTIONS
~~~
#else
~~~
#endif 
#endif 
```


parjong:

PR #9054, opened 2017-01-23
Fix GC info for `ByReference<T>`

kouvel:
When `ByRefernece<T>` is directly on the stack, not inside a `Span<T>`, ComputeGCLayout was missing to include in the GC info. This was causing many span tests to fail in GCStress=c mode.

### DIFF
--- a/src/jit/gcinfo.cpp
+++ b/src/jit/gcinfo.cpp
@@ -265,6 +265,12 @@ GCInfo::WriteBarrierForm GCInfo::gcIsWriteBarrierCandidate(GenTreePtr tgt, GenTr
         case GT_STOREIND:
 #endif               // !LEGACY_BACKEND
         case GT_IND: /* Could be the managed heap */
+            if (tgt->TypeGet() == TYP_BYREF)
+            {
+                // Byref values cannot be in managed heap.
+                // This case occurs for Span<T>.
+                return WBF_NoBarrier;
+            }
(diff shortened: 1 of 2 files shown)


kouvel:
@jkotas, this fixes deterministic GCStress issues I've seen on x64 and x86, but there are still nondeterminitic GCStress=c issues on x86 which I haven't figured out yet.

jkotas:
failing to crossgen mscorlib on Windows arm - I guess the arm JIT has a problem with handling the correct GC info

jkotas:
LGTM otherwise. Thanks!

kouvel:
Hmm I can't find the log file saved anywhere. I'll try a rerun to make sure it's consistent.

kouvel:
@sivarv, please take a look at the second commit and let me know if it tests ok

kouvel:
@sivarv, I have changed the fix to what you have suggested. This one also works on cross arm crossgen, and looks like it does not generate write barrier on x64.

sivarv:
Looks good to me.
CC @briansull and @dotnet/jit-contrib  - please review gcinfo.cpp change.

sivarv:
I have verified that gcinfo.cpp change no SPMI diffs on desktop framework assemblies.

kouvel:

kouvel:
@sivarv, should I wait for more reviews, or can this be satisfactorily be merged once the tests pass?

Issue #8857, opened 2017-01-09
Remove desktop-only code from corelib


danmosemsft:
We do not need to retain the ability for coreclr's corelib to build for desktop. It has probably rotted anyway.

Diffing the defines I see in \src\mscorlib with the defines I see going to the compiler, code marked with these defines seems eligible for removal (either the code, or the define from code)

- [x] FEATURE_ADVANCED_MANAGED_ETW_CHANNELS
- [x] FEATURE_CASE_SENSITIVE_FILESYSTEM
- [x] FEATURE_CODEPAGES_FILE
- [x] FEATURE_COMINTEROP_MANAGED_ACTIVATION
- [x] FEATURE_COMINTEROP_WINRT_DESKTOP_HOST
- [x] FEATURE_COMPRESSEDSTACK
- [x] FEATURE_CRYPTO
- [x] FEATURE_LEAK_CULTURE_INFO
- [x] FEATURE_LEGACYSURFACE
- [x] FEATURE_LINK_DEMAND
- [x] FEATURE_METHOD_RENTAL
- [x] FEATURE_NORM_IDNA_ONLY
- [x] FEATURE_PLS
- [x] FEATURE_RWLOCK
- [x] FEATURE_SERIALIZATION
- [x] FEATURE_UNSAFE_CONTRACTS
- [x] FEATURE_X509
- [x] FEATURE_X509_SECURESTRINGS

This is just the `FEATURE_` ones, perhaps some others, here are the files:
[alldiffsincode.txt](https://github.com/dotnet/coreclr/files/694509/alldiffsincode.txt)
[alldiffsincompiler.txt](https://github.com/dotnet/coreclr/files/694510/alldiffsincompiler.txt)

Desktop specific build functionality including 

- [ ] https://github.com/dotnet/coreclr/blob/master/clr.desktop.props

 can presumably go also.

Care to not trample on concurrent https://github.com/dotnet/coreclr/issues/4651 work by @jamesqo although many of these will be in code he's not removing, also a delete of some code in a file ought to merge fine with a delete of the file itself.

@jkotas is the above correct?

danmosemsft:
Removed the reference to https://github.com/dotnet/coreclr/blob/master/clr.desktop.props above, that's relevant to the native side as well so let's ignore it for this issue.

jkotas:
> @jkotas is the above correct?

Some of the above ifdefs are set differently for Windows/Unix - they should stay:

```
[...]

PR #9058, opened 2017-01-23
Remove redundant symbols disabled in CoreCLR

danmosemsft:
Contributes to https://github.com/dotnet/coreclr/issues/8857


danmosemsft:
@jkotas can you have a quick look?

danmosemsft:
As far as further cleanup (on the managed side) there are a lot of symbols that are in clrdefines.targets that do not appear in the code. (https://gist.github.com/danmosemsft/81b28745c0886e81e307d8339a21c286)

It wouldn't be hard to remove those from the props/targets, but it's just clutter in those, not in the code.

danmosemsft:
Verified that linux debug and windows x64 debug disassembly is unaffected by above changes.

danmosemsft:
@tarekgh could you okay this commit: https://github.com/dotnet/coreclr/pull/9058/commits/4c81ba582c4e9a9128c917ccb25ecac19824b1f9 (it removes only dead code)

jkotas:
@danmosemsft Looks great! Thanks

tarekgh:
@danmosemsft 

we may need to do more clean up as I am seeing FEATURE_CODEPAGES_FILE used in the following files 

```
src\classlibnative\inc\nlsinfo.h:#if FEATURE_CODEPAGES_FILE
src\classlibnative\inc\nlsinfo.h:#endif // FEATURE_CODEPAGES_FILE
src\classlibnative\nls\nlsinfo.cpp:#if FEATURE_CODEPAGES_FILE
src\classlibnative\nls\nlsinfo.cpp:#endif // FEATURE_CODEPAGES_FILE
src\classlibnative\nls\nlstable.cpp:#if FEATURE_CODEPAGES_FILE
```

danmosemsft:
Yep for sure, I opened https://github.com/dotnet/coreclr/issues/9067 to do the native side. I"ll take a look when I have time but since that code isn't going to sync with CoreRT it's no hurry. Thanks

tarekgh:
@danmosemsft thanks for the clean up. this really nice to see it.

PR #8951, opened 2017-01-13
Fix debugger launch race hitting breakpoints in startup code.

mikem8361:
The attached flag was been set asynchronously relative to the DebugActiveProcess
returning.  This could cause a race where the initial module load notification being
missed/not sent to the debugger.

This fix sets the attached flag before any notifications sent during launch if the runtime was
launched/attached using the startup handshake after dbgshim tells the runtime to "continue"
when the runtime startup API callback returns.

Issue: https://github.com/dotnet/coreclr/issues/8606

PR #9060, opened 2017-01-23
Fix debugger launch race hitting breakpoints in startup code. (#8951)

mikem8361:
The attached flag was been set asynchronously relative to the DebugActiveProcess
returning.  This could cause a race where the initial module load notification being
missed/not sent to the debugger.

This fix sets the attached flag before any notifications sent during launch if the runtime was
launched/attached using the startup handshake after dbgshim tells the runtime to "continue"
when the runtime startup API callback returns.

Also fixes another race condition in dbgshim where EnumerateCLRs returns a NULL continue event
handle because the coreclr module was loaded but the g_hContinueStartupEvent wasn't initialized
on the runtime side yet. Changed the static initialization of g_hContinueStartupEvent to
INVALID_HANDLE_VALUE and the InternalEnumerateCLRs sleep/retry loop to retry when any of the
handles are INVALID_HANDLE_VALUE. This fixes the race only when you have the latest dbgshim
and coreclr binaries and the old/new mixes still function but don't fix the race.

### DIFF
--- a/src/debug/ee/debugger.cpp
+++ b/src/debug/ee/debugger.cpp
@@ -1895,7 +1895,7 @@ void Debugger::SendCreateProcess(DebuggerLockHolder * pDbgLockHolder)
 
 #if defined(FEATURE_CORECLR) && !defined(FEATURE_PAL)
 
-HANDLE g_hContinueStartupEvent = NULL;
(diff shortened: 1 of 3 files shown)


mikem8361:

Issue #6574, opened 2016-08-02
GC/Scenarios/DoublinkList/doublinkgen fails occasionally in CI


adityamandaleeka:
Example output:

```
GC/Scenarios/DoublinkList/doublinkgen/doublinkgen.sh
11:24:48                BEGIN EXECUTION
11:24:48                /mnt/j/workspace/dotnet_coreclr/master/checked_ubuntu_tst_prtest/bin/tests/Windows_NT.x64.Checked/Tests/coreoverlay/corerun doublinkgen.exe
11:24:48                Test should return with ExitCode 100 ...
11:24:48                999 DLinkNodes finalized
11:24:48                Test Failed
11:24:48                Expected: 100
11:24:48                Actual: 1
11:24:48                END EXECUTION - FAILED

```

Examples of failing jobs:

http://dotnet-ci.cloudapp.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_tst_prtest/3728/
http://dotnet-ci.cloudapp.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_tst_prtest/3747/
http://dotnet-ci.cloudapp.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_tst_prtest/3768/

adityamandaleeka:
@swgillespie Can you take a look at this please?

swgillespie:
Seems the same as https://github.com/dotnet/coreclr/issues/4093 . This one hasn't shown up in a while, but it looks to be the same as when I observed it a few months ago.

It's not obvious what's happening here. I've been unable to repro this locally and it does only seem to repro on checked Ubuntu builds.

adityamandaleeka:
Added another failure to the list above.

adityamandaleeka:
Another: http://dotnet-ci.cloudapp.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_tst_prtest/3898/

jkotas:
Another: http://dotnet-ci.cloudapp.net/job/dotnet_coreclr/job/master/job/x86_ryujit_checked_windows_nt_prtest/3306/

stephentoub:
Another:
https://ci.dot.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_tst_prtest/82/consoleText

swgillespie:
@stephentoub got a crash dump from that, thanks!

mikedn:
[...]

PR #9063, opened 2017-01-23
Disable GC/Scenarios/DoublinkList/doublinkgen

swgillespie:
This test fails often and blocks people's PRs. See https://github.com/dotnet/coreclr/issues/6574 for the ongoing investigation.


adityamandaleeka:
I don't think this will work on non-Windows platforms, since msbuild isn't involved in running tests there. You will probably want to modify one of the txt files (such as testsFailingOutsideWindows.txt) to accomplish that.

swgillespie:
done, thanks!

Issue #9064, opened 2017-01-23
GC/Coverage/delete_next_card_table/delete_next_card_table failure in CI


swgillespie:
From https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_windows_nt_prtest/2004/console:

```
10:43:44 D:\j\workspace\debug_windows---17180f1b\bin\tests\Windows_NT.x64.Debug\TestWrappers\GC\GC.XUnitWrapper.cs(4866): error : GC._Coverage_delete_next_card_table_delete_next_card_table_._Coverage_delete_next_card_table_delete_next_card_table_cmd [FAIL] [D:\j\workspace\debug_windows---17180f1b\tests\runtest.proj]
10:43:44         
10:43:44         Assert failure(PID 4168 [0x00001048], Thread: 5604 [0x15e4]): !CREATE_CHECK_STRING(!"Detected use of a corrupted OBJECTREF. Possible GC hole.")
10:43:44         
10:43:44         CORECLR! `Object::ValidateInner'::`1'::catch$5 + 0xE9 (0x00007ffa`401c91ce)
10:43:44         CORECLR! CallSettingFrame + 0x20 (0x00007ffa`40014c10)
10:43:44         CORECLR! _CxxCallCatchBlock + 0x15A (0x00007ffa`4000f3ea)
10:43:44         NTDLL! RtlCaptureContext + 0x3C3 (0x00007ffa`56052a63)
10:43:44         CORECLR! Object::ValidateInner + 0x10F (0x00007ffa`3f465b4f)
10:43:44         CORECLR! Object::Validate + 0xE8 (0x00007ffa`3f465488)
10:43:44         CORECLR! WKS::gc_heap::background_promote + 0x117 (0x00007ffa`3fdc7437)
10:43:44         CORECLR! PinObject + 0x195 (0x00007ffa`3fda4e45)
10:43:44         CORECLR! ScanConsecutiveHandlesWithoutUserData + 0xB3 (0x00007ffa`3fe33be3)
10:43:44         CORECLR! BlockScanBlocksWithoutUserData + 0x5C (0x00007ffa`3fe3301c)
10:43:44             File: d:\j\workspace\debug_windows---17180f1b\src\vm\object.cpp Line: 1827
10:43:44             Image: D:\j\workspace\debug_windows---17180f1b\bin\tests\Windows_NT.x64.Debug\Tests\Core_Root\CoreRun.exe
10:43:44         
10:43:44         
10:43:44   
[...]

PR #9065, opened 2017-01-23
Fix the ordering of phases in compphases.h.

pgavlin:
PHASE_EMPTY_FINALLY and PHASE_CLONE_FINALLY were out-of-order in the
phase list with respect to their execution order, which caused the total
cycle count accounting to report incorrect data. This change places
these phases in the correct position in the phase list.


pgavlin:
@AndyAyersMS @dotnet/jit-contrib PTAL

BruceForstall:
LGTM

AndyAyersMS:
Seems like it would not be too hard to future-proof this. For instance, add a do-nothing phase that is listed last and called last, and a comment indicating that it must be the last phase listed, and an assertion check at jit startup or similar, that the max phase ID is the ID for that empty last phase.

AndyAyersMS:
Added note to #9064 for the failing GC test [(log)](https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_windows_nt_prtest/2022/)

Issue #8643, opened 2016-12-15
Introduce a base class for REGDISPLAY


parjong:
While bring up x86/Linux, I found that the current definition of ``REGDISPLAY`` has several duplicated code  (especially code related with ``WIN64EXCEPTIONS``).

It would be better to introduce a base class to remove such code duplication.

parjong:
It would be better to rename Esp as SP in x86 REGDISPLAY to make it more easy to maintain.

gkhanna79:
CC @rahku

PR #9070, opened 2017-01-23
Rename Esp in x86 REGDISPLAY as SP

parjong:
This commit renames Esp in x86 REGDISPLAY as SP to make it consistent with other architetures (as discussed in #8643).


parjong:
@janvorli Please take a look.

PR #8672, opened 2016-12-17
Refactor TreeNodeInfoInitCmp

mikedn:
IMO `TreeNodeInfoInitCmp` tries to do too many things for its own good. The code is difficult to follow, difficult to change and buggy (#8599, #7480). It requires workarounds such as #7561. It requires special casing `GT_AND` in [`genConsumeRegs`](https://github.com/dotnet/coreclr/blob/f305035f29220d2673623d61f663b96f4fb5f56e/src/jit/codegenlinear.cpp#L1223).

So, I moved a bunch of logic to `LowerCompare` so that `TreeNodeInfoInitCmp` now only deals with operand containment. Basically, `LowerCompare` attempts to put the IR in a form that is suitable for register allocation and code generation:
* Narrow the compare when possible - this allows memory operand containment and the use of a smaller instruction encoding. This is done by changing the compare operand type as needed, `GTF_RELOP_SMALL` is no longer needed.
* Transform and-cmp to test when possible. This required adding `GT_TEST_EQ` and `GT_TEST_NE` which is IMO a bit invasive but much preferable to having the transform split between `TreeNodeInfoInitCmp` and `genCompareInt` (not to mention the `genConsumeRegs` special case). Anyway, I think that in long term it may be better to lower all relops and `GT_JTRUE` to `GT_CMP`, `GT_TEST`, `GT_SETCC`, `GT_JCC`.

PR #9072, opened 2017-01-24
Add a test case for a test optimization

CarolEidt:
This was a test case for an old bug. It shows a regression with PR #8672, so it will be useful to have this test case available to reference in an issue (even though it is not expected to be a common scenario and therefore may be low priority).


CarolEidt:
@sivarv PTAL

sivarv:
Looks good.

PR #9074, opened 2017-01-24
Fix GenTree hash function for commutative operators

BruceForstall:
gtHashValue() computes a hash value for a tree that, at least now,
is only used for determining whether to display the tree in the
JitDump after morphing. The hash function for commutative binary
operators was designed to return the same value if the operands
were switched (possibly this was used for CSE previously?). However,
this means that in some cases, nearly symmetric operands before and
after morphing would cause the "after morph" tree not to be displayed,
which is very confusing, since it implies that morphing had no
effect on the tree.

Since it appears we have no need for this special handling of
commutative binary operators, remove the special hashing done for them.

### DIFF
--- a/src/jit/compiler.h
+++ b/src/jit/compiler.h
@@ -2070,13 +2070,13 @@ public:
 
     bool gtHasLocalsWithAddrOp(GenTreePtr tree);
 
-    unsigned gtHashValue(GenTree* tree);
-
     unsigned gtSetListOrder(GenTree* list, bool regs, bool isListCallArgs);
 
     void gtWalkOp(GenTree** op1, GenTree** op2, GenTree* adr, bool constOnly);
 
 #ifdef DEBUG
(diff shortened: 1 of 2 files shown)


BruceForstall:
@briansull @JosephTremoulet PTAL
cc @dotnet/jit-contrib

briansull:
LGTM

BruceForstall:
Wow, this looks like a crappy test:
```
..............Expected thread to sleep for 2000 ms (and not be interrupted).
Average delta: 1094 ms,  acceptable delta: 150


FAILED: threading/SleepEx/test2/paltest_sleepex_test2. Exit code: 1
```
@janvorli : this is a PAL tests. Does it make sense to run in the lab? Not sure we can guarantee anything about machine load when the test is running.
cc @mmitche

BruceForstall:
@dotnet-bot test FreeBSD x64 Checked Build

PR #9081, opened 2017-01-24
[x86/Linux] Port CheckRegDisplaySP

parjong:
This commit removes duplicated declarations of CheckRegDisplaySP, and enables it for x86.

### DIFF
--- a/src/inc/regdisp.h
+++ b/src/inc/regdisp.h
@@ -8,6 +8,8 @@
 
 #ifdef DEBUG_REGDISPLAY
 class Thread;
+
+void CheckRegDisplaySP (REGDISPLAY *pRD);
 #endif // DEBUG_REGDISPLAY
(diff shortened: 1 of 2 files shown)


src/inc/regdisp.h
```diff
@@ -369,11 +364,6 @@ inline void SyncRegDisplayToCurrentContext(REGDISPLAY* pRD)
 
 #if defined(_WIN64)
     pRD->SP         = (INT_PTR)GetSP(pRD->pCurrentContext);
-
-#ifdef DEBUG_REGDISPLAY
```
janvorli: @jkotas I wonder if there was some reason for using the CheckRegDisplaySP only for 64 bit processors.

jkotas: I do not see any good reason.

PR #9086, opened 2017-01-24
Add Vector.op_Explicit tests

### LINKED ISSUE #5135: Vector<T> op_Explicit does not seem to be tested
This is defined as an intrinsic (in simdintrinsiclist.h) but does not seem to be tested by any of the tests in the JIT/SIMD test directory. Such a test should be added, and it should use the `JitLog` class in VectorUtil.cs to ensure that the intrinsic is recognized appropriately.

mikedn:
Fixes #5135

### DIFF
--- a/tests/src/JIT/SIMD/VectorCast.cs
+++ b/tests/src/JIT/SIMD/VectorCast.cs
@@ -1,0 +1,204 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+//
+
+using System;
+using System.Numerics;
+using System.Runtime.CompilerServices;
+
+internal partial class VectorTest
+{
+    const int Pass = 100;
+    const int Fail = -1;
+
+    // An array containing 1,2,3,4...
+    static readonly byte[] bytePattern;
+
+    static VectorTest()
+    {
+        bytePattern = new byte[64]; // enough to test up to 512 bit SIMD vectors
+
+        for (int i = 0; i < bytePattern.Length; i++)
+            bytePattern[i] = (byte)(i + 1);
+    }
+
+    static T[] GetPatternAs<T>() where T : struct
+    {
+        T[] pattern = new T[bytePattern.Length]; // should be bytes.Length / sizeof(T) but sizeof(T) doesn't work 
+                                                 // so we'll settle for a larger than needed array
+        Buffer.BlockCopy(bytePattern, 0, pattern, 0, bytePattern.Length);
+        return pattern;
+    }
+
+    // Get a Vector<T> containing a specific bit pattern
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    static Vector<T> GetVector<T>() where T : struct => new Vector<T>(GetPatternAs<T>());
+
+    // Cast a Vector<T> to specific vector type (we can cast from a generic value but the cast itself can't
+    // be generic so we need one method for each possible vector element type).
+    static Vector<Byte>   To_Byte   <T> (Vector<T> from) where T : struct => (Vector<Byte>  ) from;
+    static Vector<SByte>  To_SByte  <T> (Vector<T> from) where T : struct => (Vector<SByte> ) from;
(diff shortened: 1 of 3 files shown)


mikedn:
It's not clear to me how reliable using the JIT log is. If a method is not inlined then it won't appear in the log and we end up thinking that it was treated as an intrinsic when it was not. And `Vector<T>.op_Explicit` methods are good inline candidates...

cc @CarolEidt

danmosemsft:
@CarolEidt can we merge this old one?

danmosemsft:
ping @CarolEidt ?

danmosemsft:
@BruceForstall can you please take a look?

BruceForstall:
@danmosemsft it's on my to-do list

mikedn:
@dotnet-bot Test OSX10.12 x64 Checked Build and Test please

BruceForstall:
@mikedn you write:

> It's not clear to me how reliable using the JIT log is. If a method is not inlined then it won't appear in the log and we end up thinking that it was treated as an intrinsic when it was not. And Vector<T>.op_Explicit methods are good inline candidates...

I assume you mean "If a method _is_ inlined"?

tests/src/JIT/SIMD/VectorCast.cs
```diff
@@ -0,0 +1,193 @@
...
+        succeeded &= Test_Byte   <Double> ();
+        succeeded &= Test_SByte  <Double> ();
+        succeeded &= Test_UInt16 <Double> ();
+        succeeded &= Test_Int16  <Double> ();
+        succeeded &= Test_UInt32 <Double> ();
+        succeeded &= Test_Int32  <Double> ();
+        succeeded &= Test_UInt64 <Double> ();
+        succeeded &= Test_Int64  <Double> ();
+        succeeded &= Test_Single <Double> ();
+        succeeded &= Test_Double <Double> ();
```
BruceForstall: One problem with this test is that if it fails, you have no idea what failed without running the test in a debugger and (hopefully) seeing the same failure as whoever saw the failure in the first place.

Would it be possible to print out a message when the test fails that indicate which tests fails? This could be as simple as numbering all the test cases, possibly by passing a unique int to the `Test_*` function, and outputting that number if the condition fails, e.g., `Test 23 failed`.

BruceForstall:
I was going to suggest you could force no inlining by adding something like this to the csproj:
```
  <PropertyGroup>
    <CLRTestBatchPreCommands><![CDATA[
$(CLRTestBatchPreCommands)
set COMPlus_JitNoInline=1
]]></CLRTestBatchPreCommands>
  <BashCLRTestPreCommands><![CDATA[
$(BashCLRTestPreCommands)
export COMPlus_JitNoInline=1
]]></BashCLRTestPreCommands>
  </PropertyGroup>
```

but unfortunately COMPlus_JitNoInline is not available in release builds :-(

tests/src/JIT/SIMD/VectorCast.cs
mikedn: Yep, that's a common test problem. I've changed the code to report the line number. Seems to be a good compromise:
* the test author doesn't have to generate a bunch of numbers manually
* it's trivial from a codegen point of view compared to other possible solutions (e.g. using reflection to print out test method names).

mikedn:
> I assume you mean "If a method is inlined"?

Yep.

> but unfortunately COMPlus_JitNoInline is not available in release builds :-(

Hmm, but these tests run in checked builds as well so maybe that should be enough?

BruceForstall:
> Hmm, but these tests run in checked builds as well so maybe that should be enough?

True; if it fails in release by passes in checked, we have bigger problems :-)

tests/src/JIT/SIMD/VectorCast.cs
BruceForstall: > CallerLineNumber

Cool! Obviously, I need to get out (of native code) more often...

BruceForstall:
@mikedn I went ahead and merged this. You can decide if you think it's worthwhile to add the `COMPlus_JitNoInline=1` change as a follow-up.

PR #9090, opened 2017-01-24
Fix arm64 build break

rahku:
Fix arm64 build break caused by commit https://github.com/dotnet/coreclr/commit/285c9f3002627135a8a152fea8ef97c93c8a27e0#diff-0a77138fa545bcae1099ae8734fea0f1

### DIFF
--- a/src/inc/regdisp.h
+++ b/src/inc/regdisp.h
@@ -8,7 +8,7 @@
 
 #ifdef DEBUG_REGDISPLAY
 class Thread;
-
+struct REGDISPLAY;
 void CheckRegDisplaySP (REGDISPLAY *pRD);
 #endif // DEBUG_REGDISPLAY
 


rahku:
@dotnet-bot test Windows_NT arm64 checked

rahku:
PTAL @gkhanna79

src/inc/regdisp.h
```diff
@@ -8,7 +8,7 @@
 
 #ifdef DEBUG_REGDISPLAY
 class Thread;
-
+struct REGDISPLAY;
```
gkhanna79: Can you confirm that x64 was also broken due to this change?

janvorli: x64 was not broken, the CI tests have passed for x64.

gkhanna79: Why was this change not required for x64 in that case? Is the structure defined in some manner for x64 differently than for arm64?

janvorli: My guess is that for other targets, the REGDISPLAY is actually always forward declared before including this file and ARM64 is different. Since ARM32, x64 and x86 have all passed the CI.

gkhanna79: Since Arm64 also defines WIN64EXCEPTIONS like x64, I think we should understand how it gets declared for x64 and see if we can unify the declaration for our 64bit platforms (or for WIN64EXCEPTIONS based approach) instead of explicitly declaring it here.

janvorli: It doesn't have to be declared here if we move the function prototype (wrapped in the same ifdef) to line 340, that means after REGDISPLAY is defined for all the targets.

rahku: Should be fine to redeclare forward declarations

janvorli: I agree, I doesn't matter whether we move the prototype after the definitions or keep it here and use the forward declaration.

rahku: arm64 build was successful. error is for something else. Checking this in to unblock nuget package publish. Will investigate later where x64 seems to be getting forward declaration from.

PR #9092, opened 2017-01-24
Fix runtest.sh build of Coreoverlay to account for new directory structure in CoreFX

wtgodbe:
@weshaggard @gkhanna79 @janvorli 

I expect this will fail in the CI here - it will be tested against the version of netci.groovy that's currently in master, which doesn't pass arguments to runtest.sh in the way I expect here (see my changes to netci.groovy in the PR for what they should look like)


wtgodbe:
@ellismg PTAL as well

weshaggard:
cc @RussKeldorph

PR #9095, opened 2017-01-24
Introduce REGDISPLAY_BASE

### LINKED ISSUE #8643: Introduce a base class for REGDISPLAY

parjong:
This PR introduces REGDISPLAY_BASE as a common base for REGDISPLAY in order to fix #8643.

### DIFF
--- a/src/inc/regdisp.h
+++ b/src/inc/regdisp.h
@@ -152,6 +152,12 @@ struct REGDISPLAY : public REGDISPLAY_BASE {
 #ifdef _TARGET_ARM64_
     Arm64VolatileContextPointer     volatileCurrContextPointers;
 #endif
+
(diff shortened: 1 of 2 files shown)


parjong:

parjong:
@janvorli Could you let me know if there is a problem with this PR?

janvorli:
@parjong no, there is no problem. I just went to bed yesterday when the CI was still running and today it went out of my radar.

parjong:
@janvorli Thank you! I wish that Github will introduce merge if test passes...

PR #9096, opened 2017-01-24
Add BenchmarkGames revcomp benchmark

AndyAyersMS:
This is based on version 1 of the C# variants. Am not including
the largest input file since it is ~250MB. Instead we iterate
on a smaller file.

Note the text input files are opened as binary and the app looks for
Unix line endings, hence the update to .gitattributes.

The benchmark is setup to look for its input next to the benchmark exe
as run in both normal test and xunit-perf test modes, keying off the
CORE_ROOT environment variable. In the xunit-perf mode the input files
are subsequently copied into the sandbox.

Also updated the ending copyright date on the Benchmark Games
license page.


AndyAyersMS:
cc @dotnet/jit-contrib @DrewScoggins 

Looks to be properly CPU bound...
image

sivarv:
Looks good.

PR #9099, opened 2017-01-25
Reintroduce building Core_Root to runtest.cmd

wtgodbe:
https://github.com/dotnet/coreclr/pull/8973 took the step of building Core_Root out of runtest.cmd - this is a problem if a dev wants to build the tests once, make incremental product changes, rebuild the product, then run the tests against that new product. This PR adds that step back to runtest.cmd, while leaving it in build-test.cmd as well in case we're running tests in Helix (without runtest.cmd), or a dev wants to run an individual test without having ever run runtest.cmd.

CC @ramarag


ramarag:
LGTM

PR #9100, opened 2017-01-25
Update isUsedFromMemory to include CNS_DBLs

adiaaida:
Contained GT_CNS_DBLs should also be considered as isUsedFromMemory.

Fixes VSO 370233.

### DIFF
--- a/tests/src/JIT/Regression/JitBlue/DevDiv_370233/DevDiv_370233.il
+++ b/tests/src/JIT/Regression/JitBlue/DevDiv_370233/DevDiv_370233.il
@@ -1,0 +1,136 @@
+.assembly extern mscorlib{}
+.assembly DevDiv_370233{}
+.class C extends [mscorlib]System.Object
+{
+   .method static unsigned int32 M(bool, bool)
+   {
+      .maxstack  65535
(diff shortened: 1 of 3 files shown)


adiaaida:
@CarolEidt PTAL
/cc: @dotnet/jit-contrib

adiaaida:
I believe the failures in OSX and Ubuntu are known failures.

tests/src/JIT/Regression/JitBlue/DevDiv_370233/DevDiv_370233.il
```diff
@@ -0,0 +1,139 @@
+// Reading from '\\jitci\automation\Official\CodegenMirror\JitStress.2.ILGen.amd64chk\D2017-01-18T00.21.50\codegenxarch_cpp__1721____Assertion_failed__regOp__isUsedFromReg__.mc' dumping raw IL for MC Indexes to console
```
pgavlin: Probably want to remove this comment :)

adiaaida: Good point. fixed.

PR #9111, opened 2017-01-25
Fix x86compatjit and x86lb legs

adiaaida:
Change 7eacdd7 moved Exclude from runtest.cmd to build-test.cmd, which
broke the x86compatjit and x86lb legs in the lab. This change splits the
builds for these jobs into build.cmd skiptests and build-tests.cmd Exclude
<exclusion file> and removes the Exclude option from the runtest command.

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -1783,13 +1783,15 @@ def static calculateBuildCommands(def newJob, def scenario, def branch, def isPR
                 case 'x86lb':
                     def arch = architecture
                     def buildOpts = ''
+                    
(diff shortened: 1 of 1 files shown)


adiaaida:
@dotnet-bot test ci please

adiaaida:
@pgavlin @jashook PTAL
/cc: @dotnet/jit-contrib

pgavlin:
LGTM. Do we know why exclusions were changed from a run-time to a build-time option? This seems suboptimal...

adiaaida:
@pgavlin Their commit message said something about helix compatibility.

Issue #972, opened 2015-05-11
Consider running the CodeFormatter over mscorlib source


weshaggard:
There has been a number of questions about making the code formatting consistent in mscorlib and it is something we would like to consider doing in the future once we fully understand the codeflow between the sources on github and our internal source that ships in the full .NET Framework. The reason we are hesitant to run the formatter right now is there is a lot of code and it would add a lot of risk to the full desktop framework source code. A couple of the risks factors stem from things like incorrect merges which are always going to happen when you have a large codebase with code churning on multiple forks and the code formatter will definitely change almost every line of source code in mscorlib. Other things like renaming fields we simply cannot even do because it will break compatibility with serializers for the full framework. 

At any rate I'm opening this issue to track the desire to run the code formatter but it will likely not happen for some time.

jamesqo:
I'm in favor of this! Not too familiar with Roslyn/the formatter, but would it be possible to configure the formatter not to make breaking changes (like renaming fields etc)?

jaredpar:
@jamesqo every rule on code formatter can be disabled including field renames.

jamesqo:
@jaredpar 

> every rule on code formatter can be disabled including field renames.

That's great!

@weshaggard It's been almost a year now, is there any progress on this? I think it would make it easier for people to contribute to mscorlib if the code there were better-formatted.

weshaggard:
There has been no progress. Right now I'm not even sure we will do it at all because of various constraints where this code is shared and will cause nightmare merges which adds more risk then it actually adds in value. I'm keeping the item open because I do want to try and do it at some point in the future if the opportunity arises but I will only do it if I can also bast all the formatting changes to all the places this code is shared at the same time.

jamesqo:
@weshaggard Sounds awesome :grinning: Sorry for the late reply, but please let us know here if/when you're able to do this. I'm sure other contributors to mscorlib would appreciate this, as well.

GSPP:
Maybe it is more feasible to format subsections of the codebase that are not being worked on in branches at the time? It's not necessary to format everything at once.

In general it is important to format by fixing bad things instead of reformatting everything. There is no formatting engine on earth which can produce a nice layout for everything. Formatting should be constrained to obviously correct fixed.

jamesqo:
[...]

PR #9114, opened 2017-01-25
Remove System.Security.Permissions.SecurityPermissionAttribute from System.Private.CoreLib.

erozenfeld:
When compiling with /unsafe+ Roslyn emits
  .permissionset reqmin
             = {class 'System.Security.Permissions.SecurityPermissionAttribute, System.Private.CoreLib,
             Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e' =
             {property bool 'SkipVerification' = bool(true)}}
if System.Security.Permissions.SecurityPermissionAttribute is present.

In particular, System.Private.CoreLib.dll is compiled with /unsafe+ and has the above in the manifest.

https://github.com/dotnet/coreclr/commit/e516211b2d3a3ae2dbd80f2c94b129910a9e280d
made BclRewriter remove System.Security.Permissions.SecurityPermissionAttribute
from System.Private.CoreLib.dll but the above reference remained.

This change removes System.Security.Permissions.SecurityPermissionAttribute completely
to avoid confusion when tools are analyzing msil and try to resolve the manifest
of System.Private.CoreLib.dll.


erozenfeld:
@rahku PTAL

rahku:
LGTM. I would just confirm that type is actually private in the generated assembly.

erozenfeld:
Yes, I confirmed that the type is private in the generated assembly.

erozenfeld:

erozenfeld:

erozenfeld:
1. I don't know if anyone is compiling against CoreLib. I'm working on a tool that's analyzing assemblies in order to reduce their size and it stumbles on this reference. I can hard-code this case in the tool but I think we shouldn't be producing assemblies with references that can't be resolved.
2. I agree, it shouldn't. But even after we fix Roslyn, older versions of csc may be compiling System.Private.CoreLib on developers' machines and producing this bogus reference.

In general, what's our process for removing/refactoring classes in System.Private.CoreLib that Roslyn assumes are always present. I assume System.Security.Permissions.SecurityPermissionAttribute is not the only such class.

jkotas:
>  analyzing assemblies in order to reduce their size and it stumbles on this reference

What is this assembly? Where did this assembly came from?

> System.Private.CoreLib that Roslyn assumes are always present

I do not think that Roslyn assumes that this type is present in CoreLib.

erozenfeld:
The assembly is System.Private.Corelib.dll itself. Just build coreclr normally from the repo and look at the manifest. The assembly is compiled with /unsafe+ and that causes Roslyn to emit this attribute.

jkotas:
>  what's our process for removing/refactoring classes in System.Private.CoreLib 

System.Private.CoreLib is our internal implementation detail. We reserve the right to remove/refactor any types in or out of it.

erozenfeld:
Actually, here is what happens. System.Security.Permissions.SecurityPermissionAttribute is present in System.Private.CoreLib.dll initially. Roslyn emits the reference (because of /unsafe) but then BclRewriter removes the class from System.Private.CoreLib.dll. Should the right fix be to modify BclRewriter to remove the reference once  System.Security.Permissions.SecurityPermissionAttribute is removed?

jkotas:
This is bug in the BCL rewriter tool that is used to post-process the sources: It removes SecurityPermissionAttribute definition, but leaves a reference to it in .permissionset directive.

The right fix for this is to delete the source for SecurityPermissionAttribute from CoreLib. It is dead code. We are working on cleaning up the dead code from CoreLib (https://github.com/dotnet/coreclr/issues/972). If this one is preventing from making progress, I would suggest deleting SecurityPermissionAttribute from CoreLib instead of adding more references to it.

erozenfeld:
@jkotas I removed System.Security.Permissions.SecurityPermissionAttribute. PTAL.

PR #9116, opened 2017-01-25
Fix `Span<T>` GCStress issue on x86

kouvel:
x86 by default treats the type of `ByReference<T>` as the actual type of its IntPtr field as an optimization. `ByReference<T>` is a special case where the struct needs to be treated as a value type so that its field can be considered as a by-ref pointer. This fixes `GCStress=f` issues on span tests on x86.


kouvel:
@jkotas @sivarv

sivarv:
Looks good to me based on my limited understanding.

jkotas:
@dotnet-bot test OSX x64 Checked Build and Test  please

Issue #8776, opened 2017-01-03
Improve parameter names of Array.Copy


jamesqo:
`Array.Copy` accepts the following parameter names: `sourceArray`, `sourceIndex`, `destinationArray`, `destinationIndex`, `length`. However, the parameter names in the exceptions do not match up: https://github.com/dotnet/coreclr/blob/master/src/classlibnative/bcltype/arraynative.cpp#L1024-L1054 They should be changed to match the names in C#.

Blocked on https://github.com/dotnet/corefx/pull/14798, which I am working on. Fixing this right now and changing the expected names in the corefx tests will create a merge conflict with me [here](https://github.com/dotnet/corefx/pull/14798/files#diff-b5f51582b094b16996afaa39afabf9ebR1594). However, I intend to pick this up once I am finished with that PR.

PR #9117, opened 2017-01-25
Fix parameter names for Array.Copy

jamesqo:
Fixes https://github.com/dotnet/coreclr/issues/8776. Going to amend tests in corefx soon.

/cc @jkotas


jamesqo:
Hold on-- before this is merged, I realized there's an issue with the call sites just below because they're throwing ArgumentExceptions with empty string params.

```cpp
FCThrowResVoid(kArgumentException, W("Arg_LongerThanSrcArray"));
FCThrowResVoid(kArgumentException, W("Arg_LongerThanDestArray"));
```

jamesqo:
@jkotas Sorry, one question before this is merged. I see [this line](https://github.com/dotnet/corefx/blob/71014b32aead46a2aa908d8834e2d6d60b59bba3/src/System.Runtime/tests/System/ArrayTests.cs#L1306) in the corefx tests

```cs
Assert.Throws<ArgumentException>(null, () => new int[3].CopyTo(new int[10, 10], 0));
```

Do you see how this line can result in an `ArgumentException`? I have looked over the `Array.Copy` code ~3 times but AFAICS those parameters should reach [this line](https://github.com/dotnet/coreclr/blob/master/src/classlibnative/bcltype/arraynative.cpp#L1036) which throws a `RankException`, not an `ArgumentException`. (I would like to find the place that's throwing an `ArgumentException` with a null name and change it to pass in the name of a param.)

jkotas:
> Do you see how this line can result in an ArgumentException?

It is calling Array.CopyTo. It is the first line of that method.

jkotas:
> Going to amend tests in corefx soon.

Are the existing corefx tests passing with this change, or they need to be fixed to be compatible with this change?

jamesqo:
> It is calling Array.CopyTo. It is the first line of that method.

Ah. OK in that case I think it would be nice to have `Array.CopyTo` include a parameter name as well, so temporarily marking this no merge.

> Are the existing corefx tests passing with this change, or do they need to be fixed to be compatible with this change?

There are quite a few corefx tests that need to be changed because many of them just defer to `Array.Copy` for validation, and also check the param names. See https://github.com/dotnet/corefx/pull/15489

jkotas:
> OK in that case I think it would be nice to have Array.CopyTo include a parameter name as well

I would do that in separate change. The pattern used in Array.CopyTo is used in number of places. It is unclear tp me whether adding the parameter name to all these places is a good idea.

jamesqo:
@jkotas OK. In that case feel free to merge.

Test OSX x64 Checked Build and Test
Test Ubuntu x64 Checked Build and Test

jkotas:
The corefx PR with the test updates has to be ready to go before this is merged. It is not yet - check the last comment on it.

jamesqo:
@jkotas OK, it was a little painful, but I finally managed to finish up the corefx tests. I rebuilt the runtime and ran the corefx tests against it to make sure that all tests related to this change pass against it.

jkotas:
@ianhays @karelz Is System.Array considered to be part of area-System.Collections, or is this tagged incorrectly? (Just asking so that I can tag and route things correctly.)

karelz:
Array was decided to be part of area-System.Runtime (I have updated [areas table](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/issue-guide.md#areas)).

ArraySegment should probably tag along, but I just found 2 ArraySegment issues in CoreFX under area-System.Collections.
@ianhays @AlexGhiondea @joperezr @jkotas any preference / suggestion based on code base closeness where to include it? (area-Runtime vs. area-Collections)

jkotas:
For better or worse, the areas are defined based on contracts. Both Array and ArraySegment are in https://github.com/dotnet/corefx/blob/master/src/System.Runtime/ref/System.Runtime.cs contract, and so they fall into area-Runtime by default.

danmosemsft:
Merging since test PR was also approved.

PR #9119, opened 2017-01-25
Fix CoreFX Binary copying in netci.groovy

wtgodbe:
CoreFX no longer archives their binaries from the root bin folder, but rather just packages up everything in bin/runtime/netcoreapp-{OS}-{Config}-{Arch} into build.tar.gz. This should pick up binaries from the right place now.


wtgodbe:
CC @AndyAyersMS @jkotas

PR #9121, opened 2017-01-26
[x86/Linux] Do NOT use pXXX fields

parjong:
This commit removes pXXX fields in REGDISPLAY, and revises all the references on these fields (as discussed in #8998) to refers corresponding fields in pCurrentContext and pCurrentContetPointers.

### DIFF
--- a/src/inc/regdisp.h
+++ b/src/inc/regdisp.h
@@ -89,7 +89,7 @@ struct REGDISPLAY : public REGDISPLAY_BASE {
 
 #define NONVOLATILE_REG_METHODS(reg) \
     inline PDWORD Get##reg##Location(void) { return (pCurrentContextPointers) ? pCurrentContextPointers->reg : &pCurrentContext->reg; } \
-    inline void   Set##reg##Location(PDWORD p##reg) { if (pCurrentContextPointers) { pCurrentContextPointers->reg = p##reg; } pCurrentContext->reg = *p##reg; }
+    inline void   Set##reg##Location(PDWORD p##reg) { pCurrentContextPointers->reg = p##reg; }
 
 #endif // WIN64EXCEPTIONS
 
--- a/src/inc/regdisp.h
+++ b/src/inc/regdisp.h
@@ -71,28 +71,26 @@ struct REGDISPLAY : public REGDISPLAY_BASE {
     DWORD * pEsi;
     DWORD * pEdi;
     DWORD * pEbp;
+#endif // !WIN64EXCEPTIONS
+
+#ifndef WIN64EXCEPTIONS
 
 #define VOLATILE_REG_METHODS(reg) \
-    inline DWORD  Read##reg(void)   { return *p##reg; } \
-    inline PDWORD Locate##reg(void) { return p##reg;  } \
-    inline void   Restore##reg(PDWORD p##reg) { this->p##reg = p##reg; } \
-    inline void   Trash##reg(PDWORD p##reg)   { this->p##reg = p##reg; }
+    inline PDWORD Get##reg##Location(void) { return p##reg;  } \
+    inline void   Set##reg##Location(PDWORD p##reg) { this->p##reg = p##reg; }
 
 #define NONVOLATILE_REG_METHODS(reg) VOLATILE_REG_METHODS(reg)
 
 #else // !WIN64EXCEPTIONS
 
 #define VOLATILE_REG_METHODS(reg) \
-    inline DWORD  Read##reg(void)   { return pCurrentContext->reg; } \
-    inline PDWORD Locate##reg(void) { return NULL; } \
-    inline void   Restore##reg(PDWORD p##reg) { pCurrentContext->reg = *p##reg; } \
-    inline void   Trash##reg(PDWORD p##reg)   { pCurrentContext->reg = *p##reg; }
+    inline PDWORD Get##reg##Location(void) { return &pCurrentContext->reg; } \
+    inline void   Set##reg##Location(PDWORD p##reg) { pCurrentContext->reg = *p##reg; }
 
 #define NONVOLATILE_REG_METHODS(reg) \
-    inline DWORD  Read##reg(void)   { return pCurrentContext->reg; } \
-    inline PDWORD Locate##reg(void) { return (pCurrentContextPointers) ? pCurrentContextPointers->reg : NULL; } \
-    inline void   Restore##reg(PDWORD p##reg) { if (pCurrentContextPointers) { pCurrentContextPointers->reg = p##reg; } pCurrentContext->reg = *p##reg; } \
-    inline void   Trash##reg(PDWORD p##reg)   { if (pCurrentContextPointers) { pCurrentContextPointers->reg = NULL; } pCurrentContext->reg = *p##reg; }
+    inline PDWORD Get##reg##Location(void) { return (pCurrentContextPointers) ? pCurrentContextPointers->reg : &pCurrentContext->reg; } \
+    inline void   Set##reg##Location(PDWORD p##reg) { if (pCurrentContextPointers) { pCurrentContextPointers->reg = p##reg; } pCurrentContext->reg = *p##reg; }
+
 #endif // WIN64EXCEPTIONS
 
     VOLATILE_REG_METHODS(Eax)
@@ -117,21 +115,21 @@ struct REGDISPLAY : public REGDISPLAY_BASE {
       TAG_EBP
     };
 
-    inline void Restore(TAG tag, PDWORD pReg)
+    inline void SetLocation(TAG tag, PDWORD pReg)
     {
         switch (tag)
         {
         case TAG_EBX:
-            RestoreEbx(pReg);
+            SetEbxLocation(pReg);
             break;
         case TAG_ESI:
-            RestoreEsi(pReg);
+            SetEsiLocation(pReg);
             break;
         case TAG_EDI:
-            RestoreEdi(pReg);
+            SetEdiLocation(pReg);
             break;
         case TAG_EBP:
-            RestoreEbp(pReg);
+            SetEbpLocation(pReg);
             break;
         default:
             _ASSERTE(!"Invalid register");
@@ -155,13 +153,13 @@ inline void SetRegdisplaySP(REGDISPLAY *display, LPVOID sp ) {
 inline TADDR GetRegdisplayFP(REGDISPLAY *display) {
     LIMITED_METHOD_DAC_CONTRACT;
 
-    return (TADDR)display->ReadEbp();
+    return (TADDR)*display->GetEbpLocation();
 }
 
 inline LPVOID GetRegdisplayFPAddress(REGDISPLAY *display) {
     LIMITED_METHOD_CONTRACT;
     
-    return (LPVOID)display->LocateEbp();
+    return (LPVOID)display->GetEbpLocation();
 }
 
 inline PCODE GetControlPC(REGDISPLAY *display) {
--- a/src/inc/regdisp.h
+++ b/src/inc/regdisp.h
@@ -78,7 +78,7 @@ struct REGDISPLAY : public REGDISPLAY_BASE {
     inline void   Restore##reg(PDWORD p##reg) { this->p##reg = p##reg; } \
     inline void   Trash##reg(PDWORD p##reg)   { this->p##reg = p##reg; }
 
-#define NONVOLATILE_REG_METHODS(reg) VOLATILE_REG_METHOD(reg)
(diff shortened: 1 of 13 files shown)


parjong:

parjong:

janvorli:
@parjong I think this refactoring went too far. It seems that instead of defining all the four methods per each register, it would be enough to define just GetRegisterLocation(rd, reg) macro or inline functions GetXXXLocation() where XXX is Eax, Ebx, Ecx, Edx, ... Then you would replace all usages of pXXX by GetXXXLocation(). It seems easier to read that way with using C++ dereferencing to read and write the values instead of introducing new functions.
Also the RestoreXXX has different semantics for WIN64EXCEPTIONS than for !WIN64EXCEPTIONS, which is really confusing.
The TrashXX functions are completely unnecessary, since they are just setters.
Could you please simplify it this way?

janvorli:
Also, I've missed before that you haven't removed the pXXX members for WIN64EXCEPTIONS, which I have expected to happen.

parjong:
@janvorli Sorry for late response. I was out of office in last 4 days. 

Do you mean that the value of ``*pCurrentContextPointers->(Reg)`` may differ from that of ``pCurrentContext->(Reg)``? From your first comment, ``GetXXXLocation`` may work when we read the value, but I'm not sure how we could update the value only with this method. Could you let me know some details?

For your second comment, this PR removes pXXX members for WIN64EXCEPTIONS. These members are inside ``#ifndef WIN64EXCEPTIONS ~ #else`` region, and thus do not exist for WINEXCEPTIONS. (I'll revise this PR to make it easy to read).

janvorli:
@parjong than you for your response. As for my last comment, I am sorry, I can see now that you are right and the pXXX members are no longer there for WIN64EXCEPTIONS. I somehow confused myself.

As for the GetXXXLocation, you are right, we would also need a SetXXXLocation function. But these two should be sufficient and it would be very easy to understand their intent.

parjong:
@janvorli Thank you for comment. Could you let me know some details about ``pCurrentContextPointers``? (I'm trying to revise ``TrashXXX`` methods)

``UnwindEbpDoubleAlignFrame`` updates  all the registers with some garbage data. To revise that code, I introduced ``TrashXXX` methods because I though that these updates should not be reflected in ``pCurrentContextPointers``.

Is it enough to use ``SetXXX`` even for this code?

janvorli:
@parjong I think that basically everything that was done through pXXX before should now be done through pCurrentContextPointers->XXX. I see this change as a pure replacement of pXXX with the other ones.

parjong:

parjong:
@janvorli PR is revised per feedback. Please take a look.

src/vm/eetwain.cpp
```diff
@@ -3207,12 +3211,12 @@ const RegMask CALLEE_SAVED_REGISTERS_MASK[] =
     RM_EBP  // last register to be pushed
 };
 
-const SIZE_T REGDISPLAY_OFFSET_OF_CALLEE_SAVED_REGISTERS[] =
+const REGDISPLAY::TAG CALLEE_SAVED_REGISTERS_TAG[] =
```
janvorli: I would prefer to keep using offsets instead of a tag and switch in the SetLocation. It should work if we let the offsets array for WIN64EXCEPTIONS to keep offsets in the context pointers structure and the code accessing it would use pContext->pCurrentContextPointers as the base.
Since there are five places where we use the offset array the same way, it would then make sense to factor this out into a separate function, e.g. `SetRegisterLocation(i, addr)` so that the WIN64EXCEPTIONS changes would be localized.

src/inc/regdisp.h
```diff
@@ -72,7 +71,71 @@ struct REGDISPLAY : public REGDISPLAY_BASE {
...
+    inline PDWORD Get##reg##Location(void) { return p##reg;  } \
+    inline void   Set##reg##Location(PDWORD p##reg) { this->p##reg = p##reg; }
+
+#define NONVOLATILE_REG_METHODS(reg) VOLATILE_REG_METHODS(reg)
+
+#else // !WIN64EXCEPTIONS
+
+#define VOLATILE_REG_METHODS(reg) \
+    inline PDWORD Get##reg##Location(void) { return &pCurrentContext->reg; } \
+    inline void   Set##reg##Location(PDWORD p##reg) { pCurrentContext->reg = *p##reg; }
```
parjong: @janvorli Could you let me know the role of ``pCurrentContext``? It seems that ``pCurrentContextPointers`` is enough for the purpose of unwind.

janvorli: The pCurrentContext->XXX contain values that the pCurrentContextPointers->XXX point to. Please note that there are actually two context and two context pointer structures. To one of them points the pCurrentContext / pCurrentContextPointers. To the other points the pParentContext / pParentContextPointers. That allows unwinding at some places to eliminate some context / context pointers copying by just switching the the pCurrentXXX and pParentXXX pointers.

While it might theoretically be possible to modify the existing code to not to use pCurrentContext for x86, it would be error prone and make the x86 Linux code even more different from the other targets using WIN64EXCEPTIONS.

src/inc/regdisp.h
```diff
@@ -430,6 +423,10 @@ inline void FillRegDisplay(const PREGDISPLAY pRD, PT_CONTEXT pctx, PT_CONTEXT pC
     pRD->ctxPtrsOne.Esi = &pctx->Esi;
     pRD->ctxPtrsOne.Edi = &pctx->Edi;
     pRD->ctxPtrsOne.Ebp = &pctx->Ebp;
+
```
janvorli: A nit - for consistency with other target, you could now revert back to using a loop for copying the registers if you reordered them in the KNONVOLATILE_CONTEXT_POINTERS to match their order in the CONTEXT.

src/inc/regdisp.h
```diff
@@ -75,35 +75,28 @@ struct REGDISPLAY : public REGDISPLAY_BASE {
...
-
 #else // !WIN64EXCEPTIONS
 
-#define VOLATILE_REG_METHODS(reg) \
-    inline PDWORD Get##reg##Location(void) { return &pCurrentContext->reg; } \
-    inline void   Set##reg##Location(PDWORD p##reg) { pCurrentContext->reg = *p##reg; }
-
-#define NONVOLATILE_REG_METHODS(reg) \
+#define REG_METHODS(reg) \
     inline PDWORD Get##reg##Location(void) { return (pCurrentContextPointers) ? pCurrentContextPointers->reg : &pCurrentContext->reg; } \
```
janvorli: The GetXXXLocation should also assume the pCurrentContextPointers is not NULL.

src/unwinder/i386/unwinder_i386.cpp
```diff
@@ -97,6 +97,10 @@ OOPStackUnwinderX86::VirtualUnwind(
 
     ContextRecord->ContextFlags |= CONTEXT_UNWOUND_TO_CALL;
 
+#define CALLER_SAVED_REGISTER(reg) if (rd.pCurrentContextPointers->reg) ContextRecord->reg = *rd.pCurrentContextPointers->reg;
```
janvorli: This is not needed. The UnwindStackFrame never sets context pointers of the caller saved registers.

src/vm/i386/cgenx86.cpp
```diff
@@ -572,29 +594,30 @@ void FaultingExceptionFrame::UpdateRegDisplay(const PREGDISPLAY pRD)
...
-#undef CALLEE_SAVED_REGISTER
+    memcpy(pRD->pCurrentContext, &m_ctx, sizeof(CONTEXT));
 
-    pRD->SP = m_Esp;
-    pRD->ControlPC = *PTR_PCODE(pRD->PCTAddr);
+    pRD->SP = m_ctx.Esp;
+    pRD->ControlPC = m_ctx.Eip;
 
-#else // WIN64EXCEPTIONS
+#define CALLEE_SAVED_REGISTER(regname) pRD->pCurrentContextPointers->regname = &m_ctx.regname;
```
janvorli: Looking at the amd64 version, we should also set the context pointers for caller saved registers here.

src/vm/i386/cgenx86.cpp
```diff
@@ -757,20 +804,47 @@ void HijackFrame::UpdateRegDisplay(const PREGDISPLAY pRD)
...
+
+#define CALLEE_SAVED_REGISTER(reg) pRD->pCurrentContext->reg = m_Args->reg;
+    ENUM_CALLEE_SAVED_REGISTERS();
+#undef CALLEE_SAVED_REGISTER
+
+#define CALLEE_SAVED_REGISTER(reg) pRD->pCurrentContextPointers->reg = &m_Args->reg;
+    ENUM_CALLEE_SAVED_REGISTERS();
+#undef CALLEE_SAVED_REGISTER
+
+#define CALLER_SAVED_REGISTER(reg) pRD->pCurrentContextPointers->reg = &m_Args->reg;
```
janvorli: Looking at the amd64 version, context pointers to argument registers (ecx, edx) should be NULLed and only the pointer to the return register (eax) should be set.

src/vm/i386/cgenx86.cpp
```diff
@@ -757,20 +804,47 @@ void HijackFrame::UpdateRegDisplay(const PREGDISPLAY pRD)
...
+
+#ifdef WIN64EXCEPTIONS
+
+    pRD->IsCallerContextValid = FALSE;
+    pRD->IsCallerSPValid      = FALSE;        // Don't add usage of this field.  This is only temporary.
+
+    pRD->pCurrentContext->Eip = *PTR_PCODE(pRD->PCTAddr);
+    pRD->pCurrentContext->Esp = (DWORD)(pRD->PCTAddr + sizeof(TADDR));
+
+#define CALLEE_SAVED_REGISTER(reg) pRD->pCurrentContext->reg = m_Args->reg;
```
janvorli: It seems we can use the UpdateRegDisplayFromCalleeSavedRegisters that you've added now.

src/vm/i386/cgencpu.h
```diff
@@ -154,6 +154,11 @@ typedef INT32 StackElemType;
 // This represents some of the FramedMethodFrame fields that are
 // stored at negative offsets.
 //--------------------------------------------------------------------
+#define ENUM_CALLER_SAVED_REGISTERS() \
```
janvorli: A nit. These are argument and scratch registers, so it seems it might be better to call it ENUM_ARGUMENT_AND_SCRATCH_REGISTERS. We have ENUM_ARGUMENT_REGISTERS for amd64, so that would make it similar.

src/unwinder/i386/unwinder_i386.cpp
parjong: ``UnwindEbpDoubleAlignFrame`` sets Eax, Ecx, Edx for debugging purpose:
```
3832 #ifdef _DEBUG
3833             /* The filter has to be called by the VM. So we dont need to
3834                update callee-saved registers.
3835              */
3836
3837             if (flags & UpdateAllRegs)
3838             {
3839                 static DWORD s_badData = 0xDEADBEEF;
3840
3841                 pContext->SetEaxLocation(&s_badData);
3842                 pContext->SetEcxLocation(&s_badData);
3843                 pContext->SetEdxLocation(&s_badData);
3844
3845                 pContext->SetEbxLocation(&s_badData);
3846                 pContext->SetEsiLocation(&s_badData);
3847                 pContext->SetEdiLocation(&s_badData);
3848             }
3849 #endif
```
I'm not sure which one is better: enclosing this region with _DEBUG, or simply removing. Please let me know your opinion.

src/vm/i386/cgenx86.cpp
parjong: That was the my first trial, but ``UpdateRegDisplayFromCalleeSavedRegisters`` cannot be used here as ``m_Args`` is of type ``HijackArgs`` (not ``CalleeSavedRegisters``).

src/unwinder/i386/unwinder_i386.cpp
janvorli: I am sorry, I have missed this. So please leave your change as it is.

src/vm/i386/cgenx86.cpp
janvorli: Ah, I see, the HijackArgs differ from the amd64 version.

janvorli:
@parjong I'll be OOF on Thursday and Friday, so if you create PRs during that time, please ask @jkotas to look at them instead.

parjong:
@janvorli Oh, I see. Thanks you for your consideration.

parjong:

PR #9124, opened 2017-01-26
Fix non-deterministic AVX2 detection

BruceForstall:
The caller required ECX to be zero when calling CPUID, but wasn't calling
the getextcpuid() function to cause that to happen.


BruceForstall:
This was causing problems in the x86 SIMD bring-up.

@CarolEidt PTAL
cc @dotnet/jit-contrib

Issue #8496, opened 2016-12-07
Enable RyuJIT Backend for ARM32


YongseopKim:
We( @YongseopKim; @chunseoklee ; @hseok-oh ; @hqueue ) are going to work for enabling RyuJIT backend for arm32.

Any information that we should know? (like that somebodies are working some feature of RyuJIT on arm32)

jkotas:
cc @dotnet/jit-contrib

RussKeldorph:
Awesome!  It should be much easier to do it now that x86 paved the 32-bit path.  I'm not aware of anyone else working on RyuJIT/ARM32.  dotnet/jit-contrib have certainly talked about it a great deal, so expect feedback and feel free to ask questions.

mskvortsov:
@hqueue I'm working on `CodeGen::genSetRegToCond` right now.
And I've put my head into the `!isRegPairType(tree->TypeGet())` issue already.

(BTW, broadly speaking, there are still quite a few overlapping areas in the bring up process, so at this stage we probably should not try too hard not to repeat each other's efforts).

hqueue:
@mskvortsov Cool. For` !isRegPairType(tree->TypeGet())` and `assignedInterval != nullptr` , I think they are all related to types which requires two registers , e.g. long and double type in all phases including LSRA. 
BTW I will look into other assertions.

hqueue:
Summarize current status with PR #8710 and #8724 (cited from #8724)

# Status
Test result from 139 tests from `JIT/CodeGenBringUpTests`  with `COMPlus_JitDisasm=BringUpTest:* COMPlus_AltJit=BringUpTest:*`

## with PR #8710
```
=======================
Test Results
=======================
# CoreCLR Bin Dir :
# Tests Discovered : 139
# Passed : 61
# Failed : 78
# Skipped : 0
=======================
```
All 78 failures are `Assert failure` as follows.
- 54 of 'NYI: float argument' (gone)
- 7 of 'NYI: genSetRegToCond'
- 6 of 'NYI: Lowering of long register argument'
- 4 of 'NYI_ARM: TreeNodeInfoInit default case'
- 2 of 'NYI_ARM: ARM IsContainableImmed for floating point type' (gone)
- 2 of 'NYI_ARM: fgMorphMultiregStructArgs'
- 1 of 'NYI: Cast'
- 1 of 'NYI: ARM genCallFinally'
- 1 of '!isRegPairType(tree->TypeGet())'


[...]

PR #9131, opened 2017-01-26
[WIP] Change LSRA to update intervals correctly for arm32

hqueue:
Change LSRA to update interval correctly for arm32

We should update `assignedInterval` of a pair of `RegRecord` properly for `TYP_DOUBLE` in `arm32`.
Unless we experience various faulty scenarios with assertion.

## Before
For arm32, one of faulty scenario in `JIT/CodeGenBringUpTests/DblAdd` is as follows.
- We set and clear `assignedInterval` in a pair of `RecRecord` at `unassignPhysReg()` and `assignPhysReg()` for TYP_DOUBLE.
- However we only clear `assignedInterval` for 1st `RecRecord` and don't for 2nd `RecRecord` during LSRA.
- Later we visit `unassignPhysReg()` for the 2nd `RecRecord` while LSRA, because `assignedInterval` is not cleared properly before. And this time, a new pair of `RecRecord` is consist of the 2nd `RecRecord` and another 3rd `RecRecord`.
- After clearing `assignedInterval` of 2nd `RecRecord` with `checkAndClearInterval()` and we also try to clear `assignedInterval` of 3rd `RecRecord` when type is `TYP_DOUBLE`.
- However `assignedInterval` is already `nullptr` and an assertion is raised, because 3rd `RecRecord` is not being used.

You can find a similar logic at `LinearScan::unassignPhysReg()` and `LinearScan::assignPhysReg()` with `__TARGET_ARM__` defined.

For example,
```
FAILED   - [   0][   8s]JIT/CodeGenBringUpTests/DblAdd/DblAdd.sh
               BEGIN EXECUTION
               /mnt/albireo/mywork/device/rpi3/unittests/coreoverlays/coreoverlay.test20170125/corerun DblAdd.exe
               
               Assert failure(PID 12033 [0x00002f01], Thread: 12033 [0x2f01]): Assertion failed 'assignedInterval != nullptr' in 'BringUpTest:Main():int' (IL size 62)
               
                   File: /home/hk0110/work/dxl/dotnet/github/hqueue/coreclr/src/jit/lsra.cpp Line: 6197
                   Image: /mnt/albireo/mywork/device/rpi3/unittests/coreoverlays/coreoverlay.test20170125/corerun
               
               ./DblAdd.sh: line 130: 12033 Aborted                 (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" DblAdd.exe $CLRTestExecutionArguments
               Expected: 100
               Actual: 134
               END EXECUTION - FAILED
```

For `JIT/CodeGenBringUpTests/` with `COMPlus_AltJit=BringUpTest:*` for arm32 ryujit.
```
=======================
     Test Results
=======================
# CoreCLR Bin Dir  : 
# Tests Discovered : 139
# Passed           : 94
# Failed           : 45
# Skipped          : 0
=======================
```
About 29 assertions are related to this issue in `JIT/CodeGenBringUpTests/`.

## Proposed fix
Let's update both `assignedInterval` in a pair of `RecRecord` when neccessary for arm32.

## After
Update both `assignedInterval` in a pair of `RecRecord` when neccessary.

```
PASSED   - [   0][  10s]JIT/CodeGenBringUpTests/DblAdd/DblAdd.sh
               BEGIN EXECUTION
               /mnt/albireo/mywork/device/rpi3/unittests/coreoverlays/coreoverlay.test20170125debugging/corerun DblAdd.exe
               2
               Expected: 100
               Actual: 100
               END EXECUTION - PASSED
```

For `JIT/CodeGenBringUpTests/` with `COMPlus_AltJit=BringUpTest:*` for arm32 ryujit.
About 21 assertions are gone.
```
=======================
     Test Results
=======================
# CoreCLR Bin Dir  : 
# Tests Discovered : 139
# Passed           : 115
# Failed           : 24
# Skipped          : 0
=======================
```
## Status
Still there are remaining assertions related to `assignedInterval` for arm32. 
I'm looking into other places where a pair of `assignedInterval` is not updated correctly.

Looking for your help. Thanks!

(This is a part of #8496 "Enable RyuJIT Backend for ARM32".)

### DIFF
--- a/src/jit/lsra.cpp
+++ b/src/jit/lsra.cpp
@@ -7047,6 +7047,15 @@ void LinearScan::allocateRegisters()
             if (regRecord->assignedInterval != nullptr && !regRecord->assignedInterval->isActive &&
                 regRecord->assignedInterval->isConstant)
             {
+#ifdef _TARGET_ARM_
+                if ((regRecord->assignedInterval->registerType == TYP_DOUBLE) &&
+                    isFloatRegType(regRecord->registerType))
+                {
+                    regNumber  nextRegNum        = REG_NEXT(regRecord->regNum);
+                    RegRecord* nextRegRec        = getRegisterRecord(nextRegNum);
+                    nextRegRec->assignedInterval = nullptr;
+                }
+#endif // _TARGET_ARM_
                 regRecord->assignedInterval = nullptr;
             }
             INDEBUG(dumpLsraAllocationEvent(LSRA_EVENT_FIXED_REG, nullptr, currentRefPosition->assignedReg()));


hqueue:
@dotnet-bot please test Ubuntu x64 Checked Build and Test
@dotnet-bot please test OSX x64 Checked Build and Test

hqueue:
@dotnet/jit-contrib PTAL. Is this a right approach to add code for arm32 at every-point where assignedInterval and prevousInterval are updated ? I was continuing adding code and found out that there are lots of place to be changed. 

Looking for an advice and any comment will be appreciated :)

sivarv:
@hqueue - AFAIK, LSRA needs work to support floating-point register allocation on Arm32 because its Double register D0 overlaps and could also be used as two single-precision floating-point registers S0 and S1.  Note that this is different from Arm64 floating-point model where the same D0 register could be used either as S0 or D0 register.

I think the existing LSRA support for Arm32 is very old code and not updated to keep it functional.  LSRA might need more work than what you have discovered here.  Probably a holistic approach needs to be taken to uncover all the work to support floating-point register allocation on Arm32.

@CarolEidt - might have more inputs on how to approach this.

hqueue:
@sivarv Thank you for the advice :) and sorry for late response due to lunar new years day here.
I was asking for help because I also got a similar conclusion on this issue, i.e. we need quite a lot of works for FP registers in arm32. 
I also aware of irregular layout in arm32 FP register file and concluded that we have to maintain intervals correctly throughout LSAR. It seems that current LSRA make two interval for a variable with double type but it is not considered correctly at every place as you said.

Do you have any plan for this or other huge changes is being made for LSRA ?
If not, I think I can work on this issues based on my previous experience in other linear scan allocator with your help. I will try to minimize changes to other architectures and isolate arm32 implementation with code duplication although it might be messy.

Looking for any further advice :)

hqueue:
Open a issue #9214.

CarolEidt:
@sivarv is right that there are probably holistic changes required to support this. There are no current plans to add support for these types of overlapping registers. I am currently working on refactoring the TreeNodeInfoInit phase, and plan to fold it into LSRA, but I *think* that can be done in parallel with other work (albeit with some occasional merge pain).

danmosemsft:
@hqueue @CarolEidt this PR is now 6 weeks old, should we close it?

hqueue:
@danmosemsft If you think this is too old, then I can close this and will make a new PR later.

BruceForstall:
@hqueue Are you continuing to work on LSRA support, in the general case, for the double registers overlapping the float registers? If so, it probably makes sense to close this PR and open a new one when you have a better idea of all the changes that might be required for LSRA to support this ARM32 concept.

I wonder it would be worthwhile to consider a temporary workaround of splitting the ARM32 register set such that d0-d7 are used for TYP_DOUBLE, and s16-s31 are used for TYP_FLOAT, so you don't have to worry about overlap?

Note that @CarolEidt is the expert here, but she's out for a while.

hqueue:
@BruceForstall Right I'm going to work on LSRA support for general cases. I will close this issue soon after following discussion is done.

> I wonder it would be worthwhile to consider a temporary workaround of splitting the ARM32 register set such that d0-d7 are used for TYP_DOUBLE, and s16-s31 are used for TYP_FLOAT, so you don't have to worry about overlap?

Right this could be a simple workaround and I also once had in mind similar idea of allocating dedicated FP registers for double. But I've dropped this plan because (1) I don't think this approach can be accepted during the review and (2) I've to implement additional interface between floating-point ABI of ARM and this temporary approach which can introduce unnecessary complexity.

Do you have any idea of simple approach to interface between this workaround and FP ABI of ARM,  e.g. floating point argument and return registers ?

BruceForstall:
@hqueue You're right, the ABI issues make my suggested workaround difficult or impossible.

Another idea is to only allocate floats to even registers, to "pretend" floats and doubles are the same, e.g. d0=s0, d1=s2, d2=s4, etc. I suppose ABI issues would be just as challenging here. E.g., if we need to pass a reg in s1, we'd need to move it there in call setup. Maybe that wouldn't be so bad, but I'm not sure. Certainly, a full solution would be best.

hqueue:
@BruceForstall Thank you for the suggestion and feedback.

I'm closing this issue  and will look into this for a full solution and possible workarounds disccussed here.
This discussion will be continued at issue #9214.

PR #9133, opened 2017-01-26
Improve SIMD tests

BruceForstall:
1. Since the F1/F2 tests of VectorReturn found an issue Vector2
on x86, I added variants of the same test for Vector3 and Vector4
(by basically copying F1/F2 and altering appropriately).
2. I noticed VectorMin for Vector4 wasn't checking the 'W' member,
so I fixed that.


BruceForstall:
@sivarv @CarolEidt PTAL
cc @dotnet/jit-contrib

sivarv:
Looks good.

PR #9136, opened 2017-01-26
Enable IsInPrologOrEpilog only if GC Info Decoder is not used

parjong:
IsInPrologOrEpilog is implemented only when USE_GC_INFO_DECODER is not set.

This commit hides its declarations when USE_GC_INFO_DECODER is not set, and removes unnecessary assert in it.


parjong:

parjong:

PR #9137, opened 2017-01-26
RFI - Support cross-building CoreCLR for aarch64 Android

qmfrederik:
This PR enables cross-building CoreCLR to aarch64 Android using Ubuntu.

It's a first stab, building on the efforts that @cydhaselton and @janvorli have undertaken over the holidays.

The PR breaks down into:
- A build system for Android. It's based on the Android NDK, uses clang, and allows to cross-compile for Android from Linux using the usual rootfs setup. I've added 
  * An `./init-android.sh` shell script which will download the NDK, LLDB (for remote debugging) and required third part packages and set up a rootfs for cross compile
  * An `cross/android/arm64` folder which contains the CMake configuration files required to cross configure
- Various fixes to make the code compatible with Android:
  * Use `getc` and `putc` instead of `getw` and `putc`
  * `gcc_s`, `pthread` and `rt` don't need to be linked on Android
  * `int8_t`, `wcsnlen`,  `INT_MIN` and  `PAGE_SIZE` are already defined
  * Fix the thread-related code to use `<semaphore.h>` on Android
  * Use `/data/local/tmp` for temporary file storage, not `/tmp`
  * Fix an issue with the declaration of `__reserved`

With these changes applied, all except 4 PAL tests pass; two related to aarch64 and two related to tests using `/etc/password` which doesn't exist on Android.

It's a rather large PR, so if it's better for the review process, happy to split up the PR on a commit-by-commit basis. The idea is to get feedback early so we make sure we get this right.


src/inc/sortversioning.h
```diff
@@ -165,8 +201,8 @@ namespace SortVersioning
                         __in int cchValue,
                         __out_opt LPINT pcchFound,
                         __in_opt CONST NLSVERSIONINFO * lpVersionInformation,
-                        __reserved LPVOID lpReserved,
-                        __reserved LPARAM lParam);
+                        __dbg_reserved LPVOID lpReserved,
+                        __dbg_reserved LPARAM lParam);
```
am11: This needs to be under `#ifdef __ANDROID__` (breaking Windows builds)

src/pal/src/include/pal/threadsusp.hpp
```diff
@@ -73,6 +73,8 @@ Module Name:
 #include <sys/sem.h>
 #include <sys/types.h>
 
+#elif __ANDROID__
+#include <semaphore.h>
```
am11: nit: `#elif HAVE_SEMAPHORE_H` (in contrast to `HAVE_SYS_SEMAPHORE_H` above), that way it will also take care of future platforms.

src/inc/sortversioning.h
qmfrederik: Thx, I've fixed it.

janvorli:
@qmfrederik is it so that the gcc_s.so is not present on Android? There is a reason for linking it explicitly. The problem is that we have clang stack unwinder and libunwind8 in memory together. And if the gcc_s.so is not in the list of dependencies before the libunwind8, then the high level C++ exception handling from libunwind8 would kick in and exception handling would be broken.

src/pal/inc/pal_mstypes.h
```diff
@@ -218,7 +218,11 @@ typedef __int32 int32_t;
 typedef unsigned __int32 uint32_t;
 typedef __int16 int16_t;
 typedef unsigned __int16 uint16_t;
+
+#if !defined(__ANDROID__)
+// Android already defines these types
 typedef __int8 int8_t;
```
janvorli: It is kind of strange that int8_t is defined on android, but none of the others (int32_t etc) are. Do you know where the typedef comes from on Android?

src/pal/src/include/pal/threadsusp.hpp
qmfrederik: @am11 Would love to do this, but HAVE_SEMAPHORE_H is not defined by default.
I added a check to `src/pal/src/configure.cmake` like this:

```
check_include_files(semaphore.h HAVE_SEMAPHORE_H)
```

but somehow this is never evaluated.

I feel I'm missing something obvious, so if you can help :)

src/pal/src/include/pal/sharedmemory.h
```diff
@@ -27,18 +27,30 @@
...
+#ifndef __ANDROID__
 #define SHARED_MEMORY_TEMP_DIRECTORY_PATH "/tmp"
 #define SHARED_MEMORY_RUNTIME_TEMP_DIRECTORY_PATH "/tmp/.dotnet"
 
 #define SHARED_MEMORY_SHARED_MEMORY_DIRECTORY_PATH "/tmp/.dotnet/shm"
 #define SHARED_MEMORY_LOCK_FILES_DIRECTORY_PATH "/tmp/.dotnet/lockfiles"
+
+#define SHARED_MEMORY_UNIQUE_TEMP_NAME_TEMPLATE "/tmp/.coreclr.XXXXXX"
+#else
+#define SHARED_MEMORY_TEMP_DIRECTORY_PATH "/data/local/tmp"
```
janvorli: Maybe we can define somewhere a symbol TEMP_DIRECTORY_PATH conditionally to /tmp or /data/local/tmp based on whether we are on Android or not and then use it everywhere. So e.g.
```C++
#define SHARED_MEMORY_RUNTIME_TEMP_DIRECTORY_PATH TEMP_DIRECTORY_PATH "/.dotnet"
```
Then the #ifdef would be just at one place.

src/pal/src/thread/thread.cpp
```diff
@@ -2633,6 +2633,8 @@ CPalThread::GetStackBase()
     status = pthread_attr_get_np(thread, &attr);
 #elif HAVE_PTHREAD_GETATTR_NP
     status = pthread_getattr_np(thread, &attr);
+#elif __ANDROID__
```
janvorli: We should rather fix the detection of HAVE_PTHREAD_GETATTR_NP to work on Android too.

src/pal/tests/palsuite/c_runtime/__iscsym/test1/CMakeLists.txt
```diff
@@ -13,7 +13,6 @@ add_executable(paltest_iscsym_test1
 add_dependencies(paltest_iscsym_test1 coreclrpal)
 
 target_link_libraries(paltest_iscsym_test1
-  pthread
```
janvorli: I don't think we can remove pthread unconditionally. On Linux, if it is not linked with, some synchronization primitives like mutexes don't work, they just become dummy.

janvorli: So I would rather recommend creating a cmake variable COMMON_TEST_LIBRARIES in the src/pal/tests/CMakeLists.txt that would contain pthread, m and coreclrpal for non-Android and just the last two for Android. Then all tests would have ${COMMON_TEST_LIBRARIES} here instead of the three.

qmfrederik:
@janvorli Thanks for the (ongoing) comments! I'll review them & get back to you.

Retarding `_getw` and `_putw`, you are right, they are unused, so I removed them altogether in b8e725ccccb7546e7cdacc2c4d1c85f1162a8e4a.

If the tests pass, can we already split off that commit in a separate PR or do you want to process all the changes at once?

janvorli:
@qmfrederik yes, it would be nice to split it into a separate cleanup PR.

src/pal/src/include/pal/threadsusp.hpp
am11: Going by the NDK's `semaphore.h` file, seems like it should be  `#ifdef _SEMAPHORE_H`. :)

Regarding the cmake check, I *think* (haven't tested myself) the following patch would work in `src/pal/src/configure.cmake`:

```cmake
include(CheckIncludeFileCXX)
check_include_file_cxx(semaphore.h HAVE_SEMAPHORE_H) # C++ instead of C
```

src/pal/inc/mbusafecrt.h
```diff
@@ -56,7 +56,7 @@ extern WCHAR* wcstok_s( WCHAR* inString, const WCHAR* inControl, WCHAR** ioConte
 // strnlen is not required unless the source string is completely untrusted (e.g. anonymous input on a website)
 #ifndef SUPPRESS_STRNLEN
     extern size_t PAL_strnlen( const char* inString, size_t inMaxSize );
-    extern size_t wcsnlen( const WCHAR* inString, size_t inMaxSize );
+    extern size_t PAL_wcsnlen( const WCHAR* inString, size_t inMaxSize );
```
janvorli: You are missing two additional changes:
In pal.h, next to `#define strnlen       PAL_strnlen` you need to add `#define wcsnlen       PAL_wcsnlen`
In palinternal.h, next to `#undef strnlen` you need to add `#undef wcsnlen`. 
Otherwise wcsnlen would not be visible in the coreclr sources.

src/pal/inc/pal_mstypes.h
qmfrederik: There's a couple of locations where they are declared, such as `stdint`, but it appears this is no longer causing any issues when cross-compiling on Android. So I'm going to drop this commit from the PR.

src/pal/src/include/pal/sharedmemory.h
qmfrederik: OK - see #9150 for that change.

src/pal/inc/mbusafecrt.h
qmfrederik: Thx! #9149 contains the changes.

src/pal/tests/palsuite/c_runtime/__iscsym/test1/CMakeLists.txt
qmfrederik: I can make that change, but I think it is not necessary because [according to the documentation](https://cmake.org/cmake/help/v3.0/command/target_link_libraries.html) `target_link_libraries` is transitive.

All these tests depend on `coreclrpal` and `corepal` has a link dependency on `pthread`, so it should work. (I guess the CI build still passing is kind of a confirmation of that theory)

That said, I like the idea of `COMMON_TEST_LIBRARIES` because it removes a lot of common declarations. Let me see if I can make that work.

janvorli: As for the transitivity, I have a feeling that it didn't work for some reason in the past and I had to add it explicitly, but it might be that I am just daydreaming.

src/pal/src/thread/thread.cpp
qmfrederik: That was an easy one:
```
check_library_exists(pthread pthread_getattr_np "" HAVE_PTHREAD_GETATTR_NP)
```

`pthread` doesn't exist as a separate library on Android but as part of bionic; I'll update the CMakeFile accordingly.

cydhaselton:
@janvorli 

> is it so that the gcc_s.so is not present on Android? There is a reason for linking it explicitly. The problem is that we have clang stack unwinder and libunwind8 in memory together. And if the gcc_s.so is not in the list of dependencies before the libunwind8, then the high level C++ exception handling from libunwind8 would kick in and exception handling would be broken.

gcc_s.so should be available for Android.

qmfrederik:
@cydhaselton 

> gcc_s.so should be available for Android.

That's great news! How do you link with gcc_s.so on Android?

cydhaselton:
@qmfrederik 

> That's great news! How do you link with gcc_s.so on Android?

There should be a termux package available in the Termux repo. Ping me if you can't find it.

janvorli:
@qmfrederik @cydhaselton as for the gcc_s.so, I wonder if it is the right library to link to if it needs to be installed as a separate package. On Linux, clang automatically links all binaries with that one, so we explicitly reference it just to change the order of this dependency. If it is not used on Android, then adding that dependency as a new thing doesn't sound right.
Is there something like `ldd` on Android that would dump dependencies of a binary? I would like to see what are the dependencies. I guess that we would need to explicitly link to something else than the gcc_s.so.

qmfrederik:
I'm closing this PR as it has been split up in smaller PRs and has largely landed in CoreCLR by now.

PR #9141, opened 2017-01-26
Remove dead code in the JIT.

pgavlin:
All code under `LONG_ASG_OPS` and `FANCY_ARRAY_OPT` is statically dead
(these symbols are never defined as anything besides 0). This change
removes this cruft.


pgavlin:
@dotnet/jit-contrib PTAL

BruceForstall:
:shipit:

adiaaida:
LGTM

PR #9142, opened 2017-01-26
JIT: fix some bad asserts

AndyAyersMS:
Fix a few asserts that use = instead of ==.


AndyAyersMS:
@JosephTremoulet PTAL
cc @dotnet/jit-contrib

BruceForstall:
Yikes!

LGTM

briansull:
The compilers really should be flagging this and the other case (possibly with warning as errors)

if (i = 0)

briansull:
LGTM

adityamandaleeka:
😮 LGTM

PR #6638, opened 2016-08-06
[no merge] Further tweak Buffer.MemoryCopy performance

jamesqo:
This is a follow-up to #6627, since that was accidentally merged. I've experimented with @tannergooding's idea of doing word writes at the beginning/end of the buffer before actually aligning `dest`, which shaves off a couple of branches and seems to improve performance.

This is the performance data I have for now (the Gist will be updated continually): https://gist.github.com/jamesqo/18b61a17a65489b5dd8eaf0617b9099d You need to press Ctrl+F and search for `---BASELINE---` to see the times for the existing version (the one that just got merged).

During the other PR I was unable to get consistent timings between benchmark runs (and BenchmarkDotNet takes waaay too long), so what I've done instead is run the tests multiple times and see which numbers keep popping up. Here are the 'average' times for each configuration:

**# of Bytes** | 30 | 40 | 50
---|---|---|---
**Baseline** | ~2.0 | ~2.7 | ~3.3
**Experimental** | ~1.7/8 | ~2.4 | ~2.0

Note that this should not be merged yet, as I haven't tested for i386 and have not collected data for when dest is unaligned. Also I'm thinking of possibly raising the threshold for when we do a QCall into the native `memmove`, since copying 511 bytes seems to be significantly faster than 512 (see [my comment](https://github.com/dotnet/coreclr/pull/6627#issuecomment-237999609) in the other PR).

cc @jkotas @tannergooding @GSPP @benaadams

PR #7198, opened 2016-09-14
update JIT_MemSet/MemCpy, Buffer::BlockCopy and Buffer::InternalBlock…

helloguo:
This PR provides an optimized version of JIT_MemSet/JIT_MemCpy assembly helper functions on 64-bit Windows. JIT_MemSet gets invoked when bytecode Initblk is executed while JIT_MemCpy gets invoked when bytecode Cpblk is executed. JIT_MemCpy takes care of both overlap and non-overlap scenarios. The use of this optimized JIT_MemCpy is extended to Buffer::BlockCopy and Buffer::InternalBlockCopy by replacing the CRT memmove. This PR addresses the issue #7146. 

The unit test https://github.com/dotnet/coreclr/blob/master/tests/src/performance/perflab/BlockCopyPerf.cs is used as reference and modified so that the copy length varies from 0 byte to 520 bytes. This micro benchmark tests Buffer::BlockCopy (JIT_MemCpy). The following chart and table show the result on Intel(R) Core(TM) i7-5960X CPU @ 3.00GHz with 32GB RAM. OS is Windows 10 Enterprise. 

image

image

We also measure the performance of TechEmpower (JSON serialization and Plaintext tests). The experimental configuration for the app server is: Intel(R) Xeon(R) CPU E5-2697 v3 @2.60GHz with 32GB RAM running Windows Server 2012 R2 64 bits. The difference between baseline and the prototype is within the run to run variation. For JSON serialization test, VTune profiling shows the default JIT_MemSet takes 0.6% of overall cycles while the optimized JIT_MemSet takes 0.4% cycles for our prototype. Same thing happens to Plaintext test: the default JIT_MemSet takes 1.2% of overall cycles while the optimized JIT_MemSet takes 0.8% of overall cycles. For both JSON serialization and Plaintext tests, JIT_MemSet improves by about 33%. Both JSON serialization and Plaintext tests do not use JIT_MemCpy. In terms of Buffer::BlockCopy, the profiling shows most of the copy lengths are from 16 bytes to 64 bytes, for which CRT memmove and the optimized JIT_MemCpy have similar performance. So the improvement is negligible for TechEmpower.

These optimized routines may need to be revisited when VC/universal CRTs get updated from both performance and maintenance points of view.

PR #9143, opened 2017-01-26
[Revived PR] Faster Buffer.MemoryCopy for AMD64

jamesqo:
This is a revival of the PR at https://github.com/dotnet/coreclr/pull/6638 I had to close a few months earlier because it had too many comments. I basically re-wrote the code I had written earlier, except with a few changes:

- I removed the optimizations for ARM64. I know very little about that platform, so someone more qualified should probably optimize it.
- I kept things mostly the way they were for non-AMD64 platforms. The only major thing I changed was replacing branching at the beginning/end of the copy with unaligned writes.
- I very conservatively kept the P/Invoke threshold at 1024 bytes, because even though my benchmarks said that the managed implementation was faster up to 8192 bytes, @nietras' benchmarks were saying 1024.
  - After this is merged, I'll open up a new PR where I increase it back to 8192 and we can discuss more about the threshold size.
- I also renamed some things to make the code cleaner, e.g. so that variables have self-explanatory names.

**Please note:** I am looking to get this merged mostly as-is because it's a clear win over what we currently have. Once this gets in, then we can open the door for further discussion/experimentation/etc. over optimizations.

/cc @jkotas, @nietras @tannergooding @benaadams FYI

### DIFF
--- a/src/mscorlib/src/System/Buffer.cs
+++ b/src/mscorlib/src/System/Buffer.cs
@@ -258,11 +258,12 @@ namespace System {
         [ReliabilityContract(Consistency.WillNotCorruptState, Cer.Success)]
         internal unsafe static void MemoryCopyCore(byte* destination, byte* source, nuint length)
         {
-            const nuint PInvokeThreshold = 512;
 #if AMD64
+            const nuint PInvokeThreshold = 1024;
             const nuint CopyAlignment = 16; // SIMD is enabled for AMD64, so align on a 16-byte boundary
             const nuint BytesPerIteration = 64;
 #else
+            const nuint PInvokeThreshold = 512;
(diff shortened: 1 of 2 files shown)


jamesqo:
Again this is largely a cleaned-up version of the same code I had written earlier, but I'll post benchmarks regardless. I have to re-build CoreCLR and write the benchmarks however, so I won't be able to do so straightaway. I'll update this PR when I have them.

benaadams:
What's the impact with https://github.com/dotnet/coreclr/pull/7198

jamesqo:
Going to get back to this sometime later-- have been busy lately

jkotas:
Superseded by https://github.com/dotnet/coreclr/pull/9786

PR #9149, opened 2017-01-26
Clean up PAL functions

qmfrederik:
Clean up PAL functions to enable cross-compiling for Android
* Rename `wcsnlen` to `PAL_wcsnlen`, avoids a name collision on Android
* Remove unused `_getw` and `_putw` functions


qmfrederik:
@janvorli As discussed on #9137

PR #9150, opened 2017-01-26
Use a constant to define the location of '/tmp'

qmfrederik:
On Android, the temp folder is in `/data/local/tmp` on Android, not `/tmp`.

This PR:
- Defines `TEMP_DIRECTORY_PATH` in `file.h`
- Updates locations where `/tmp` was hardcoded to use that define.


qmfrederik:
@janvorli as discussed in #9137

src/pal/src/include/pal/sharedmemory.h
```diff
@@ -27,18 +28,18 @@
 #define SHARED_MEMORY_MAX_FILE_NAME_CHAR_COUNT (_MAX_FNAME - 1)
```
janvorli: Could you please also update the comments above? I would probably just start with saying the the following paths are relative to the temporary folder and then put in just the relative parts of the path or use TMP in the paths instead and explain that TMP is platform specific and corresponds to the TEMP_DIRECTORY_PATH or something along these lines.

src/pal/src/include/pal/file.h
```diff
@@ -42,6 +42,14 @@ typedef struct _find_handle
     char   **next;
 } find_obj;
 
+#ifndef __ANDROID__
```
janvorli: I would prefer putting it into the palinternal.h somewhere close to where PROCESS_PIPE_NAME_PREFIX is defined. The file.h is a header file for file.cpp, and this is more general. Then you don't need to add any additional includes.

qmfrederik:
@janvorli Thx, I updated this PR to incorporate your feedback.

adityamandaleeka:
Looks like you'll have to update some PAL tests that depended on the hard-coded path being "/tmp/".

adityamandaleeka:
LGTM pending successful CI.

qmfrederik:
@adityamandaleeka Thx! Fixed it the other way around and made sure path to the temp dir includes a trailing slash.

Issue #9148, opened 2017-01-26
Improve sync times when downloading test native bins from Azure


wtgodbe:
Currently sync.cmd -ab takes a very long time, because it downloads test native bins for all Linux distros + OSX. We should add an option to download only test native bins, and allow the user to specify which distro they want to download binaries for.

wtgodbe:
https://github.com/dotnet/coreclr/pull/9152

PR #9152, opened 2017-01-26
Speed up test native binary syncing in pipeline

wtgodbe:
This will make publish-packages.sh publish test native binaries into an Azure container named {build-label}-{Runtime ID}, instead of in a folder named {Runtime-ID} inside of a container named {build-label}-test-native-bins. That way, you only have to download from the container with binaries for the OS you're building for, rather than downloading binaries for all non-windows OS's. The workflow for this is:

`sync.cmd -n -AzureAccount=xxx -AzureToken=yyy -Container=zzz`

Where 'zzz' is the name of the container with the native binaries you desire (i.e. {build-label}-{runtime ID}. Once this is merged, it will require pipeline changes.

### DIFF
--- a/src/publish.proj
+++ b/src/publish.proj
@@ -14,7 +14,7 @@
     <ItemGroup>
       <ForPublishing>
         <RelativeBlobPath Condition="'$(PublishTestNativeBins)' != 'true'">$(__BuildType)/%(RecursiveDir)%(Filename)%(Extension)</RelativeBlobPath>
-        <RelativeBlobPath Condition="'$(PublishTestNativeBins)' == 'true'">$(__DistroRid)-$(__BuildArch)/$(__BuildType)/%(RecursiveDir)%(Filename)%(Extension)</RelativeBlobPath>    
+        <RelativeBlobPath Condition="'$(PublishTestNativeBins)' == 'true'">$(__BuildType)/%(RecursiveDir)%(Filename)%(Extension)</RelativeBlobPath>    
       </ForPublishing>
     </ItemGroup>
     <Error Condition="'@(ForPublishing)' == ''" Text="No items were found matching pattern '$(PublishPattern)'." />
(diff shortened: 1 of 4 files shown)


wtgodbe:
CC @MattGal @gkhanna79 @maririos

wtgodbe:
Addresses https://github.com/dotnet/coreclr/issues/9148

src/publish.proj
```diff
@@ -30,9 +30,9 @@
           Condition="'$(ContainerName)' == '' or '$(PublishTestNativeBins)' == 'true'">
     <PropertyGroup>
       <ContainerName Condition="'$(__Container)' == '' and '$(PublishTestNativeBins)' != 'true'">coreclr-$(PreReleaseLabel)-$(BuildNumberMajor)-$(BuildNumberMinor)</ContainerName>
-      <ContainerName Condition="'$(__Container)' == '' and '$(PublishTestNativeBins)' == 'true'">coreclr-$(PreReleaseLabel)-$(BuildNumberMajor)-$(BuildNumberMinor)-test-native-bins</ContainerName>
+      <ContainerName Condition="'$(__Container)' == '' and '$(PublishTestNativeBins)' == 'true'">coreclr-$(PreReleaseLabel)-$(BuildNumberMajor)-$(BuildNumberMinor)-$(__DistroRid)-$(__BuildArch)</ContainerName>
```
gkhanna79: Test Binaries no longer go into their subfolder (test-native-bins)?

wtgodbe: This defines where we're publishing things in azure - in this case, I'm publishing test native binaries into a folder whose suffix is the RID those binaries were built on. Before, test native binaries from all OS's were published into the same Azure container (named xxx-test-native-bins) - I'm changing it so they each get their own container.

wtgodbe:
@dotnet-bot test this please

wtgodbe:
Windows job that failed last time just passed, looks like it was transient. Merging.

PR #9153, opened 2017-01-26
Fix another JIT-disasm check

adityamandaleeka:
@sbomer 

I just noticed another broken check introduced in 37139ef1a7631e618ee587287b48d49dad04be94, PTAL.


sbomer:
LGTM, thanks!

PR #9154, opened 2017-01-26
Support platforms where pthread is part of libc

qmfrederik:
This PR adds support for platforms where pthread is part of libc, and `semaphore.h` replaces `sys/semaphore.h`, such as Android.

- The CMake scripts have been updated to poll both libpthread and libc for the various pthread functions
- The `HAS_POSIX_SEMAPHORES` macro now links with libpthread or libc depending on the platform
- `HAVE_SEMAPHORE_H`  can be used to check for `semaphore.h` (without the `sys` prefix)


qmfrederik:
@janvorli In preparation for #9137 / Android support.
The last one for the day ;-)

src/pal/src/configure.cmake
```diff
@@ -48,15 +49,33 @@ check_include_files(gnu/lib-names.h HAVE_GNU_LIBNAMES_H)
...
-check_library_exists(pthread pthread_resume_np "" HAVE_PTHREAD_RESUME_NP)
-check_library_exists(pthread pthread_attr_get_np "" HAVE_PTHREAD_ATTR_GET_NP)
-check_library_exists(pthread pthread_getattr_np "" HAVE_PTHREAD_GETATTR_NP)
-check_library_exists(pthread pthread_getcpuclockid "" HAVE_PTHREAD_GETCPUCLOCKID)
-check_library_exists(pthread pthread_sigqueue "" HAVE_PTHREAD_SIGQUEUE)
+
+check_library_exists(pthread pthread_create "" HAVE_LIBPTHREAD)
+check_library_exists(c pthread_create "" HAVE_PTHREAD_IN_LIBC)
+
+if (HAVE_LIBPTHREAD)
```
janvorli: You can have just a cmake variable containing `c` or `pthread` and then have all of the check_library_exists just once. This double list is error prone for possible future additions of checks here.

qmfrederik: Correct, thanks!

Issue #9161, opened 2017-01-27
Optimize Span Fill, Clear, and CopyTo method implementations


ahsonkhan:
Currently the following Span methods contain trivial implementations and they need to be optimized.
- Fill
- Clear
- CopyTo

gkhanna79:
@kouvel Is this something you will be working upon? If so, can you please assign to your self or reassign as appropriate?

kouvel:
@ahsonkhan, is this something you will be working on later?

ahsonkhan:
@kouvel, I wasn't planning on working on this in the near future.

jkotas:
We will want to add some runtime helpers to make these efficient. It would be nice to match the names and signatures of the similar helpers in CoreRT to ease sharing of the implementation:

e.g. RhBulkMoveWithWriteBarrier https://github.com/dotnet/corert/blob/d84f5310c80708584bb908615a331b5520a28d63/src/System.Private.CoreLib/src/System/Runtime/RuntimeImports.cs#L531

PR #9162, opened 2017-01-27
Adding missing implementations for Span - Fill and Clear 

ahsonkhan:
This is required as part of work to enable fast span in corefx.

Further work related to optimization is needed: Issue #9161


ahsonkhan:
cc @jkotas

PR #9165, opened 2017-01-27
Avoid unnecessary intermediate String.IndexOf calls

justinvp:


src/mscorlib/src/System/String.Searching.cs
```diff
@@ -113,15 +113,6 @@ public partial class String
...
         [Pure]
         public int IndexOf(String value, int startIndex, int count) {
-            if (startIndex < 0 || startIndex > this.Length) {
-                throw new ArgumentOutOfRangeException(nameof(startIndex), Environment.GetResourceString("ArgumentOutOfRange_Index"));
-            }
-
-            if (count < 0 || count > this.Length - startIndex) {
-                throw new ArgumentOutOfRangeException(nameof(count), Environment.GetResourceString("ArgumentOutOfRange_Count"));
-            }
-            Contract.EndContractBlock();
```
justinvp: Note that removing the validation here does introduce a subtle behavior difference: if `value` is null, `ArgumentNullException` will be thrown now before the other arguments are validated. I'm guessing this is an acceptable change, but let me know if not and I'll revert this part of the change.

justinvp:

src/mscorlib/src/System/String.Searching.cs
danmosemsft: I would rather revert that, as this is such a heavily used type

justinvp: Reverted

danmosemsft:
LGTM

stephentoub:
What's the value of this change?

justinvp:
> What's the value of this change? Are the intermediates not getting inlined? We're duplicating some code, so I'm wondering what it buys us.

I haven't checked the inlining yet... The reason I submitted this change was because I was looking at cleaning up the formatting of this file and noticed all the other overloads in this file are all calling the largest overload directly instead of chaining. This change makes these three methods consistent with the rest of the file in that regard. If there's concern over some duplicated code, should we change the rest of the file to call the next largest overload (chain), instead of directly calling the largest overload?

justinvp:
Going to close this as there doesn't seem to be a clear benefit.

Issue #9036, opened 2017-01-21
RyuJIt: Loop hoist invariant struct field accesses


sivarv:
Knucleotide benchmark heavily uses Dictionary lookups.  It uses the following struct as key into hash table

```
struct ByteString
{
   public byte[] Array
   public int Start;
   public int Length;

    public override int GetHashCode()
    {
        int hc = 0;
        for (int i = 0; i < Length; i++)
            hc = hc * 31 + Array[Start + i];
        return hc;
    }

    public bool Equals(ByteString other)
    {
        if (Length != other.Length) return false;
        for (int i = 0; i < Length; i++)
            if (Array[Start+i] != other.Array[other.Start+i]) return false;
        return true;
    }
}
```

Call Chain:
`Dictionary.TryGetValue() --> FindEntry() --> calls GetHashCode() and Equals() methods on ByteString.`

In case of Equals() or GetHashCode() methods, ByteString accessed is either "thisptr" or argument passed to the method.  Struct args on Windows are considered implicitly by-ref and hence struct fields get accessed through an indirection on a byref.  Legacy JIT64 was able to loop hoist this.Array, this.Array.Length, this.Start,  other.Array, other.Array.Length and other.Start by recognizing them as loop invariant.    This shows up as execution perf difference between RyuJIT vs Legacy Jit64.

sivarv:
The reason RyuJIT doesn't hoist the field accesses is that these reads from By-ref through indir are given unique value numbers.  Note that these methods (Equals and GetHashCode) are pure methods in the sense that they only read from heap and no writes to it.

sivarv:
Please note that ByteString struct used by knucleotide benchmark is similar in concept to Span<T>.  Getting this right for ByteString would is also likely to benefit Span<T> scenarios where Span<T> accessed is an argument.

JosephTremoulet:
The [changes](https://github.com/dotnet/coreclr/compare/master...JosephTremoulet:UntrackedMem) I have in flight to address #7903 currently produce [these diffs](https://gist.github.com/JosephTremoulet/1dce85b79961b78766b780b661827f10) on these methods.  Looks like the only loads we're leaving in the loops are the variant ones and the array lengths in the bounds-checks.  I'll look into why we're not getting those array lengths...

JosephTremoulet:
[...]

Issue #7914, opened 2016-10-31
JIT: opportunities for improving async state machine codegen


AndyAyersMS:
There are some idiomatic patterns in the async state machines created by the C# compiler that cause the jit to generate sub-optimal code.

A motivating example is the method `<RequestProcessingAsync>d__2[Context][Microsoft.AspNetCore.Hosting.Internal.HostingApplication+Context]:MoveNext():this` from Kestrel ([source code](https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs#L30)).

- The `MoveNext` method is a value class method that holds a pointer to the associated `Frame` class. Frame field references in `MoveNext` thus must go through one extra level of indirection. The JIT also spills the `this` pointer in `MoveNext`, leading to code sequences like the following:
```asm
G_M55243_IG09:
       488B5510             mov      rdx, bword ptr [rbp+10H]   // RDX=this
       488B12               mov      rdx, gword ptr [rdx]       // RDX=this.<>4__this (frame)
       8B0A                 mov      ecx, dword ptr [rdx]       // null check (?)
       488B4A58             mov      rcx, gword ptr [rdx+88]    // RCX = this.<>_this.somefield
       488B5510             mov      rdx, bword ptr [rbp+10H]   // RDX = this
       488B12               mov      rdx, gword ptr [rdx]       // RDX = this.<>4_this (frame)
       488B9290010000       mov      rdx, qword ptr [rdx+400]
       488B4920             mov      rcx, gword ptr [rcx+32]
```
The jit should be able to CSE some of these redundant loads. See #7903 for a small repro case. The C# compiler could help the jit by caching `<>4__this` in a local, and perhaps by inserting an early null check.

- The method is (roughly speaking) an outermost switch. Such methods can incur relatively high prolog costs from zero initializations in the prolog, and this method is a good example:
```asm
       488DBD58FFFFFF       lea      rdi, [rbp-A8H]
       B922000000           mov      ecx, 34
       33C0                 xor      rax, rax
       F3AB                 rep stosd 
```
Most of the zeroing comes from structs with GC references. Possible mitigation routes: more aggressive copy (and reverse-copy) prop of structs, tracking struct field lifetimes, or stack packing.

[...]

PR #9169, opened 2017-01-27
Value-number byref loads

JosephTremoulet:
This change teaches value numbering to track states of "byref exposed" memory,
like it currently tracks states of heap memory; the default value number for a load
can then become a function of the memory state and pointer VN.  This enables
CSE and loop hoisting to kick in when there are no intervening stores.
It also clears the way for us to chain "byref exposed" memory VNs similar to how
heap VNs are chained (via `VNF_MapStore`), to enable value propagation through
byref params and/or address-taken locals when there are no intervening conflicting
stores.

There are a few constituent changes:
  1. A minor refactoring to facilitate the subsequent changes
  2. Abstracting the current "heap" tracking into a slightly more general
    "memory" tracking that allows for multiple memory kinds, of which
    "heap" is one.
  3. Adding the "byref exposed" memory kind
  4. Value-numbering "byref exposed" memory states and loads

The first three changes are no-diff.


JosephTremoulet:
@briansull @dotnet/jit-contrib PTAL.

I expect it would be easiest to review the constituent commits individually.

I've confirmed this gets the desired loop hoisting from #9036, and the desired CSEs from #7903, and recognizes the motivating CSE from #7914 as legal, though CSE's heuristics decide not to fire in that case (though many similar methods improve as expected).

Diffs generally consist of improved redundant load elimination, invariant code motion, and bounds-check elimination.

Desktop SPMI diffs report a codesize win of 2% across 35,000 affected methods.

Lab runs appear to show a 7% improvement in the Matinv4 benchmark, in which a number of redundant loads and bounds checks have been removed.

JosephTremoulet:

src/jit/block.h
```diff
@@ -144,6 +144,85 @@ struct EntryState
...
+{
+    return 0;
+}
+
+// Bitmask for a MemoryKindSet containing the specified MemoryKinds
+template <typename... MemoryKinds>
+inline MemoryKindSet memoryKindSet(MemoryKind id, MemoryKinds... ids)
+{
+    return (1U << id) | memoryKindSet(ids...);
+}
```
briansull: I'm not familiar with this syntax:
`typename...`  `MemoryKinds...' and `ids...`
What is this called?
And is it supported across our compilers?

src/jit/block.h
```diff
@@ -144,6 +144,85 @@ struct EntryState
...
+const char* const memoryKindNames[] = {"ByrefExposed", "GcHeap"};
+#endif // DEBUG
+
+// Bitmask describing a set of memory kinds (usable in bitfields)
+typedef unsigned int MemoryKindSet;
+
+// Bitmask for an empty MemoryKindSet
+inline MemoryKindSet memoryKindSet()
+{
+    return 0;
```
briansull: Should this be called emptyMemoryKindSet?

src/jit/block.h
```diff
@@ -144,6 +144,85 @@ struct EntryState
...
 
+// Enumeration of the kinds of memory whose state changes the compiler tracks
+enum MemoryKind
+{
+    ByrefExposed = 0, // Includes anything byrefs can read/write (everything in GcHeap, address-taken locals,
+                      //                                          unmanaged heap, callers' locals, etc.)
+    GcHeap,           // Includes actual GC heap, and also static fields
+    MemoryKindCount,  // Number of MemoryKinds
+};
+#ifdef DEBUG
```
briansull: You might want to mention that we really only support two kinds of memory and that if a third kind is every added then you need to examine the use of the variable byrefStatesMatchGcHeapStates.

src/jit/compiler.cpp
```diff
@@ -1869,8 +1869,11 @@ void Compiler::compInit(ArenaAllocator* pAlloc, InlineInfo* inlineInfo)
     m_fieldSeqStore      = nullptr;
     m_zeroOffsetFieldMap = nullptr;
     m_arrayInfoMap       = nullptr;
-    m_heapSsaMap         = nullptr;
-    m_refAnyClass        = nullptr;
+    for (MemoryKind memoryKind : allMemoryKinds())
+    {
+        m_memorySsaMap[memoryKind] = nullptr;
+    }
+    m_refAnyClass = nullptr;
```
briansull: I would be willing to reorder m_refAnyClass to come before the loop

src/jit/optimizer.cpp
```diff
@@ -6837,7 +6840,8 @@ void Compiler::optComputeLoopSideEffectsOfBlock(BasicBlock* blk)
 
     AddVariableLivenessAllContainingLoops(mostNestedLoop, blk);
 
-    bool heapHavoc = false; // True ==> there's a call or a memory store that has arbitrary heap effects.
+    // MemoryKinds for which an in-loop call or store has arbitrary effects.
+    MemoryKindSet memoryHavoc = memoryKindSet();
```
briansull: This would read better as 
`memoryHavoc = emptyMemoryKindSet(); `

src/jit/block.h
```diff
@@ -144,6 +144,85 @@ struct EntryState
...
+// Bitmask for a MemoryKindSet containing the specified MemoryKinds
+template <typename... MemoryKinds>
+inline MemoryKindSet memoryKindSet(MemoryKind id, MemoryKinds... ids)
+{
+    return (1U << id) | memoryKindSet(ids...);
+}
+
+// Bitmask containing all the MemoryKinds
+const MemoryKindSet fullMemoryKindSet = (1 << MemoryKindCount) - 1;
+
```
briansull: Also add `emptyMemoryKindSet`

src/jit/liveness.cpp
```diff
@@ -225,38 +253,39 @@ void Compiler::fgPerNodeLocalVarLiveness(GenTree* tree)
             break;
 
         case GT_CLS_VAR:
-            // For Volatile indirection, first mutate the global heap
+            // For Volatile indirection, first mutate the GC heap
```
briansull: GC andf Byref

src/jit/liveness.cpp
```diff
@@ -225,38 +253,39 @@ void Compiler::fgPerNodeLocalVarLiveness(GenTree* tree)
...
             if ((tree->gtFlags & GTF_CLS_VAR_ASG_LHS) == 0)
             {
-                fgCurHeapUse = true;
+                fgCurMemoryUse |= memoryKindSet(GcHeap, ByrefExposed);
             }
             break;
 
         case GT_IND:
-            // For Volatile indirection, first mutate the global heap
+            // For Volatile indirection, first mutate GcHeap
```
briansull: GC and Byref

src/jit/liveness.cpp
```diff
@@ -425,6 +466,10 @@ void Compiler::fgPerBlockLocalVarLiveness()
                     break;
             }
         }
+
+        // In minopts, all stores conservatively affect both GcHeap and ByrefExposed memory
+        byrefStatesMatchGcHeapStates = true;
```
briansull: We track these during MinOpts?

src/jit/liveness.cpp
```diff
@@ -501,35 +550,49 @@ void Compiler::fgPerBlockLocalVarLiveness()
...
-            {
-                printf("*");
+                if ((fgCurMemoryDef & memoryKindSet(memoryKind)) != 0)
+                {
+                    printf(" + %s", memoryKindNames[memoryKind]);
+                }
+                if ((fgCurMemoryHavoc & memoryKindSet(memoryKind)) != 0)
+                {
+                    printf("*");
+                }
```
briansull: You could reduce what we print in the dumps by adding:

`if (byrefStatesMatchGcHeapStates) { break; }`

src/jit/liveness.cpp
```diff
@@ -501,35 +550,49 @@ void Compiler::fgPerBlockLocalVarLiveness()
...
             printf(" USE(%d)=", VarSetOps::Count(this, fgCurUseSet));
             lvaDispVarSet(fgCurUseSet, allVars);
-            if (fgCurHeapUse)
+            for (MemoryKind memoryKind : allMemoryKinds())
             {
-                printf(" + HEAP");
+                if ((fgCurMemoryUse & memoryKindSet(memoryKind)) != 0)
+                {
+                    printf(" + %s", memoryKindNames[memoryKind]);
+                }
```
briansull: You could reduce what we print in the dumps by adding:

if (byrefStatesMatchGcHeapStates) { break; }

src/jit/liveness.cpp
```diff
@@ -1209,8 +1272,8 @@ class LiveVarAnalysis
             VarSetOps::ClearD(m_compiler, m_liveIn);
             VarSetOps::ClearD(m_compiler, m_liveOut);
 
-            m_heapLiveIn  = false;
-            m_heapLiveOut = false;
+            m_memoryLiveIn  = memoryKindSet();
+            m_memoryLiveOut = memoryKindSet();
```
briansull: `emptyMemoryKindSet`

src/jit/liveness.cpp
```diff
@@ -2961,15 +3024,21 @@ void Compiler::fgDispBBLiveness(BasicBlock* block)
...
     printf(" IN (%d)=", VarSetOps::Count(this, block->bbLiveIn));
     lvaDispVarSet(block->bbLiveIn, allVars);
-    if (block->bbHeapLiveIn)
+    for (MemoryKind memoryKind : allMemoryKinds())
     {
-        printf(" + HEAP");
+        if ((block->bbMemoryLiveIn & memoryKindSet(memoryKind)) != 0)
+        {
+            printf(" + %s", memoryKindNames[memoryKind]);
+        }
```
briansull: if (byrefStatesMatchGcHeapStates) { break; }

src/jit/liveness.cpp
```diff
@@ -2961,15 +3024,21 @@ void Compiler::fgDispBBLiveness(BasicBlock* block)
...
     printf("\n     OUT(%d)=", VarSetOps::Count(this, block->bbLiveOut));
     lvaDispVarSet(block->bbLiveOut, allVars);
-    if (block->bbHeapLiveOut)
+    for (MemoryKind memoryKind : allMemoryKinds())
     {
-        printf(" + HEAP");
+        if ((block->bbMemoryLiveOut & memoryKindSet(memoryKind)) != 0)
+        {
+            printf(" + %s", memoryKindNames[memoryKind]);
+        }
```
briansull: if (byrefStatesMatchGcHeapStates) { break; }

src/jit/optimizer.cpp
```diff
@@ -7045,33 +7066,39 @@ void Compiler::optComputeLoopSideEffectsOfBlock(BasicBlock* blk)
...
-                        // All other gtOper node kinds, leave 'heapHavoc' unchanged (i.e. false)
+                        // All other gtOper node kinds, leave 'memoryHavoc' unchanged (i.e. false)
                         break;
                 }
             }
         }
     }
 
-    if (heapHavoc)
+    if (memoryHavoc != 0)
```
briansull: Can memoryHavoc actually be different for GcHeap vs ByrefExposed?

If so it would be useful to have a test case in this area

src/jit/valuenum.cpp
```diff
@@ -2478,7 +2479,7 @@ ValueNum ValueNumStore::VNApplySelectorsAssignTypeCoerce(ValueNum elem, var_type
 //    block - Block where the assignment occurs
 //
 // Return Value:
-//    The value number corresopnding to the heap after the assignment.
+//    The value number corresopnding to memory after the assignment.
```
briansull: typo corresopnding

src/jit/valuenum.cpp
```diff
@@ -2809,14 +2810,11 @@ void Compiler::fgValueNumberArrIndexAssign(CORINFO_CLASS_HANDLE elemTypeEq,
...
-        printf("  fgCurHeapVN assigned:\n");
+        printf("  fgCurMemoryVN assigned:\n");
     }
 #endif // DEBUG
 
-    // bbHeapDef must be set to true for any block that Mutates the global Heap
-    assert(compCurBB->bbHeapDef);
-
-    fgCurHeapVN = vnStore->VNForMapStore(TYP_REF, fgCurHeapVN, elemTypeEqVN, newValAtArrType);
+    return vnStore->VNForMapStore(TYP_REF, fgCurMemoryVN[GcHeap], elemTypeEqVN, newValAtArrType);
```
briansull: Should we try to keep the old assert?
`assert(compCurBB->bbHeapDef);`

src/jit/valuenum.cpp
```diff
@@ -4488,75 +4498,93 @@ void Compiler::fgValueNumberBlock(BasicBlock* blk)
...
 
-    // Now do the same for "Heap".
-    // Is there a phi for this block?
-    if (blk->bbHeapSsaPhiFunc == nullptr)
-    {
-        fgCurHeapVN = GetHeapPerSsaData(blk->bbHeapSsaNumIn)->m_vnPair.GetLiberal();
-        assert(fgCurHeapVN != ValueNumStore::NoVN);
-    }
-    else
+    // Now do the same for memory.
```
briansull: `Now do the same for both kinds of memory.`

src/jit/valuenum.cpp
```diff
@@ -4488,75 +4498,93 @@ void Compiler::fgValueNumberBlock(BasicBlock* blk)
...
+    for (MemoryKind memoryKind : allMemoryKinds())
```
briansull: Do we have to twice the work here, or should we check the bool byrefStatesMatchGcHeapStates

src/jit/block.h
JosephTremoulet: It's called a variadic template, and yes it's supported across our compilers (and already used in a  [couple](https://github.com/dotnet/coreclr/blob/93cb39e3c1bbd4407261926a7365949f288ebc37/src/jit/jitstd/list.h#L476-L480) other [places](https://github.com/dotnet/coreclr/blob/93cb39e3c1bbd4407261926a7365949f288ebc37/src/jit/gentree.h#L1085-L1088) in the jit)

src/jit/block.h
JosephTremoulet: Updated, thanks for the suggestion; having `emptyMemoryKindSet` instead of `memoryKindSet()` at those uses reads much cleaner I think.

src/jit/codegenlegacy.cpp
```diff
@@ -20743,39 +20743,39 @@ GenTreePtr Compiler::fgLegacyPerStatementLocalVarLiveness(GenTreePtr startNode,
...
             case GT_IND:
-                // For Volatile indirection, first mutate the global heap
+                // For Volatile indirection, first mutate GcHeap/ByrefExposed
                 // see comments in ValueNum.cpp (under case GT_CLS_VAR)
                 // This models Volatile reads as def-then-use of the heap.
                 // and allows for a CSE of a subsequent non-volatile read
                 if ((tree->gtFlags & GTF_IND_VOLATILE) != 0)
                 {
                     // For any Volatile indirection, we must handle it as a
                     // definition of the global heap
```
JosephTremoulet: Fixed (and likewise for the rest of these)

src/jit/block.h
JosephTremoulet: I think I avoided baking that assumption into the places that use `byrefStatesMatchGcHeapStates`.  I used a few patterns:

1 - places where `output[GcHeap]` and `output[ByrefExposed]` reference the same data structure:
```c++
  for (MemoryKind memoryKind : allMemoryKinds()) {
    if (memoryKind == GcHeap && statesMatch) {
      // already updated
      assert(memoryKind > ByrefExposed);
      assert(input[memoryKind] == input[ByrefExposed]); // and/or assert(output[memoryKind] == f(input[memoryKind]))
      continue;
    }
    output[memoryKind] = f(input[memoryKind]);
  }
```

2a - places where `output[GcHeap]` and `output[ByrefExposed]` need to be copies of each other:
```c++
  for (MemoryKind memoryKind : allMemoryKinds()) {
    if (memoryKind == GcHeap && statesMatch) {
      // copy result from ByrefExposed
      assert(memoryKind > ByrefExposed);
      output[memoryKind] = output[ByrefExposed];
      continue;
    }
    output[memoryKind] = f(input[memoryKind]);
  }
```

2b - other places where `output[GcHeap]` and `output[ByrefExposed]` need to be copies of each other:
```c++
  for (MemoryKind memoryKind : allMemoryKinds()) {
    if (memoryKind == GcHeap && statesMatch) {
      // already updated
      assert(memoryKind > ByrefExposed);
      assert(input[memoryKind] == input[ByrefExposed]); // and/or assert(output[memoryKind] == f(input[memoryKind]))
      continue;
    }
    output[memoryKind] = f(input[memoryKind]);
    if (memoryKind == GcHeap && statesMatch) {
      // copy over to ByrefExposed
      output[ByrefExposed] = output[memoryKind];
    }
  }
```

(2a was more convenient for some of the particular updates, 2b for others).

In all cases, if a third kind were added the loop would just process it independently of GcHeap and ByrefExposed, which I think is a good default position.

Does that address your concern, or am I misunderstanding the feedback?

src/jit/liveness.cpp
JosephTremoulet: No, perhaps the comment was misleading.  I don't expect `byrefStatesMatchGcHeapStates` to be checked in minOpts, but leaving it unset didn't seem right.  I've updated the comment to

```c++
        // In minopts, we don't explicitly build SSA or value-number; GcHeap and
        // ByrefExposed implicitly (conservatively) change state at each instr.
```

Is that better?

src/jit/liveness.cpp
JosephTremoulet: I actually had it like that at first, but regretted it as soon as I started debugging this stuff.  Perhaps if that meant that `ByrefExposed` (or `GcHeap`) were just never mentioned in the dumps of `byrefStatesMatchGcHeapStates` methods, it would be useful, but since some of the liveness code that runs before `statesMatch` has been computed will end up dumping both, it just turns out confusing; I think better a more understandable dump than a smaller one, if we have to choose (and it seems in this case we have to choose).

src/jit/optimizer.cpp
JosephTremoulet: Since we assume all object field stores, static field stores, and array element stores alias all byref loads, we [treat them](https://github.com/dotnet/coreclr/pull/9169/commits/c0b13389b3cb5fe9d16127ae0eebf135589b8760#diff-15cd21de3f1ffaa677e7da145576d60cR6952) as `ByrefExposed`-havoc (but not `GcHeap`-havoc).  So every loop that has these but not `GcHeap`-havoc will have one kind of havoc but not the other.

src/jit/valuenum.cpp
JosephTremoulet: This code used to assign `fgCurHeapVN` but doesn't anymore.  Now it just returns the new value number, which then gets passed to `recordGcHeapStore` to do the update to `fgCurMemoryVN[GcHeap]`, and that function does have the updated assert.

src/jit/valuenum.cpp
JosephTremoulet: We do check `byrefStatesMatchGcHeapStates`, on [line 4512](https://github.com/dotnet/coreclr/pull/9169/files#diff-dec96d75e3fa9427319bc0a7fd0213a9R4512) and [line 4572](https://github.com/dotnet/coreclr/pull/9169/files#diff-dec96d75e3fa9427319bc0a7fd0213a9R4572).  It made more sense to push them into the `else` branch of the conditional in the loop, because the `if` branch needs to do the same processing for both `memoryKind`s regardless.

src/jit/block.h
briansull: Yes,  thanks for the explaination

PR #9170, opened 2017-01-27
Fix GT_LCL_FLD case in gtGetStructHandleIfPresent

### LINKED ISSUE #9159: RayTracer fails with COMPlus_JitStress=2 and COMPlus_EnableAVX=0
A previous failure in RayTracer/JitStress2, #9108, was caused by the issues fixed in #9123. However, on a non-AVX2 system a different failure is exposed, which is due to `gtGetStructHandleIfPresent` returning the parent struct handle for a `GT_LCL_FLD`. I am working on a fix.

CarolEidt:
This method was incorrectly returning the parent struct handle instead of the handle for the field. This is showing up in remorphing (due to JitStress) of a SIMD field that was assigned in a field-by-field assignment.

Fix #9159

### DIFF
--- a/src/jit/gentree.cpp
+++ b/src/jit/gentree.cpp
@@ -16323,8 +16323,15 @@ CORINFO_CLASS_HANDLE Compiler::gtGetStructHandleIfPresent(GenTree* tree)
             case GT_ASG:
                 structHnd = gtGetStructHandleIfPresent(tree->gtGetOp1());
                 break;
-            case GT_LCL_VAR:
             case GT_LCL_FLD:
+#ifdef FEATURE_SIMD
+                if (varTypeIsSIMD(tree))
+                {
+                    structHnd = gtGetStructHandleForSIMD(tree->gtType, TYP_FLOAT);
+                }
+#endif
+                break;
+            case GT_LCL_VAR:
                 structHnd = lvaTable[tree->AsLclVarCommon()->gtLclNum].lvVerTypeInfo.GetClassHandle();
                 break;
             case GT_RETURN:


CarolEidt:
@dotnet/jit-contrib PTAL

sivarv:
Looks good.

sivarv:
Wondering whether this fix is needed even on desktop?

CarolEidt:
@sivarv Yes, but I don't think it needs to be ported for immediate release, because it is only exposed because of the recent addition of support for promoting fields of SIMD type.

src/jit/gentree.cpp
```diff
@@ -16323,8 +16323,15 @@ CORINFO_CLASS_HANDLE Compiler::gtGetStructHandleIfPresent(GenTree* tree)
...
                 break;
-            case GT_LCL_VAR:
             case GT_LCL_FLD:
+#ifdef FEATURE_SIMD
+                if (varTypeIsSIMD(tree))
+                {
+                    structHnd = gtGetStructHandleForSIMD(tree->gtType, TYP_FLOAT);
+                }
+#endif
+                break;
```
BruceForstall: In the non-SIMD case, does it need to continue to execute the GT_LCL_VAR code as it did before (i.e., fall through)?

CarolEidt: No; in fact we don't want it to as it is not the right struct handle for this case. In that case it should return the `NO_CLASS_HANDLE` value which `structHnd` has already been set to.

BruceForstall:
LGTM

CarolEidt:
@dotnet-bot test Windows_NT minopts

CarolEidt:

Issue #8989, opened 2017-01-18
Win Arm64 and Arm32 crossgen fails due to JIT asserts


gkhanna79:
https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=525330&tab=details (arm)
 
```
Assert failure(PID 6352 [0x000018d0], Thread: 7604 [0x1db4]): Assertion failed '(block->bbFlags & BBF_FINALLY_TARGET) != 0' in 'System.Reflection.LoaderAllocatorScout:Finalize():this' (IL size 63)
 
    File: e:\a\_work\35\s\src\jit\flowgraph.cpp Line: 11661
    Image: E:\A\_work\35\s\bin\Product\Windows_NT.arm.Checked\x86\crossgen.exe
```
 
https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=525329&tab=details (arm64)
 
```
Assert failure(PID 780 [0x0000030c], Thread: 5808 [0x16b0]): Assertion failed '(lastConsumedNode == nullptr) || (node->gtUseNum == -1) || (node->gtUseNum > lastConsumedNode->gtUseNum)' in 'System.Globalization.HebrewCalendar:GetDatePart(long,int):int:this' (IL size 467)
 
    File: e:\a\_work\12\s\src\jit\codegenlinear.cpp Line: 1101
    Image: E:\A\_work\12\s\bin\Product\Windows_NT.arm64.Checked\x64\crossgen.exe
```

Also reported in today's (1/18) build - https://devdiv.visualstudio.com/DevDiv/_build?buildId=526476.

CarolEidt:
The second assert (arm64) has been fixed with #8991. @jashook is looking at the arm one.

gkhanna79:
@CarolEidt Thank you. Today's build confirms the issue is addressed and the only outstanding issue is that for WinArm32. @jashook How is that looking since this build break is preventing publishing of nuget packages from the official builds?

jashook:
#9013 should fix the crossgen assertion, but it does not address the root cause. Will open a new issue and assign it to Andy after #9013 is in.

jashook:
Build break is fixed with #9013. #9015 is opened to address the root issue. /cc @AndyAyersMS

PR #9171, opened 2017-01-27
JIT: fixup finally target flag after finally opts run

### LINKED ISSUE #9015: fgRemoveEmptyFinally disabled on ARM32
The EH model on ARM requires requires predecessors to have been computed in order to sucessfully leave the bit unset for cases where there are multiple CALLFINALLY / BBJ_ALWAYS pairs that point to the same finally target.

The following example demonstrates the problem.

<pre>
-------------------------------------------------------------------------------------------------------------------------------------
BBnum descAddr ref try hnd                 weight   [IL range]      [jump]      [EH region]         [flags]
-------------------------------------------------------------------------------------------------------------------------------------
BB01 [06172A40]  1  0                           1   [000..00D)-> BB03 ( cond ) T0      try {       keep i try label
BB02 [06172AF0]  1  0                           1   [00D..00F)-> BB08 (callf ) T0                  i
BB11 [06175168]  0  0                           1   [???..???)-> BB09 (ALWAYS) T0                  i internal KEEP
BB03 [06172BA0]  1  0                           1   [00F..016)-> BB07 ( cond ) T0                  i
BB04 [06172C50]  1  0                           1   [016..022)-> BB07 ( cond ) T0                  i
BB05 [06172D00]  1  0                           1   [022..02F)-> BB07 ( cond ) T0                  i
BB06 [06172DB0]  1  0                           1   [02F..035)                 T0                  i
BB07 [06172E60]  4  0                           1   [035..037)-> BB08 (callf ) T0                  i
BB10 [06173D94]  0  0                           1   [???..???)-> BB09 (ALWAYS) T0      }           i internal KEEP
BB08 [06172F10]  1     0                        1   [037..03E)        (finret)    H0   finally { } keep i label target
BB09 [06172FC0]  2                              1   [03E..03F)        (return)                     i ftarget
-------------------------------------------------------------------------------------------------------------------------------------

***************  Exception Handling table
index  eTry, eHnd
  0  ::            - Try at BB01..BB10 [000..037), Finally at BB08..BB08 [037..03E)

EH#0 has empty finally, removing the region.
fgRemoveBlock BB11

Removing unreachable BB11
fgRemoveBlock BB10

Removing unreachable BB10
EH#0: New last block of try: BB07
</pre>

AndyAyersMS:
The finally target flag can't be incrementally maintaned during finally
opts, so flag if any of these opts run and then run a post-opts fixup pass
to properly determine the flag value.

Only impacts arm targets.

Closes #9015.

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -22703,6 +22703,7 @@ void Compiler::fgRemoveEmptyFinally()
     if (emptyCount > 0)
     {
         JITDUMP("fgRemoveEmptyFinally() removed %u try-finally clauses from %u finallys\n", emptyCount, finallyCount);
+        fgOptimizedFinally = true;
 
 #ifdef DEBUG
         if (verbose)
@@ -23010,6 +23011,7 @@ void Compiler::fgRemoveEmptyTry()
     if (emptyCount > 0)
     {
         JITDUMP("fgRemoveEmptyTry() optimized %u empty-try try-finally clauses\n", emptyCount);
+        fgOptimizedFinally = true;
 
 #ifdef DEBUG
         if (verbose)
@@ -23575,6 +23577,7 @@ void Compiler::fgCloneFinally()
     if (cloneCount > 0)
     {
         JITDUMP("fgCloneFinally() cloned %u finally handlers\n", cloneCount);
+        fgOptimizedFinally = true;
 
 #ifdef DEBUG
         if (verbose)
@@ -23796,7 +23799,11 @@ void Compiler::fgDebugCheckTryFinallyExits()
 //    a callfinally/always pair.
 //
 //    Used by finally cloning, empty try removal, and empty
-//    finally removal. Resets flags and removes finally artifacts.
+//    finally removal.
+//
+//    BBF_FINALLY_TARGET bbFlag is left unchanged by this method
+//    since it cannot be incrementally updated. Proper updates happen
+//    when fgUpdateFinallyTargetFlags runs after all finally optimizations.
 
 void Compiler::fgCleanupContinuation(BasicBlock* continuation)
 {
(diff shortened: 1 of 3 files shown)


AndyAyersMS:
@BruceForstall PTAL
cc @dotnet/jit-contrib

src/jit/flowgraph.cpp
```diff
@@ -23827,3 +23829,47 @@ void Compiler::fgCleanupContinuation(BasicBlock* continuation)
...
+            BasicBlock* const continuation = leave->bbJumpDest;
+
+            if ((continuation->bbFlags & BBF_FINALLY_TARGET) == 0)
+            {
+                JITDUMP("Found callfinally BB%02u; setting finally target bit on BB%02u\n", block->bbNum,
+                        continuation->bbNum);
+
+                continuation->bbFlags |= BBF_FINALLY_TARGET;
+            }
+        }
```
BruceForstall: You could just set it unconditionally, but make the DEBUG-only JITDUMP conditional, if you only want to print once per set bit.

BruceForstall:
LGTM

I looked at whether we could simply not mark finally targets until just prior to codegen -- after adding/removing blocks is no longer allowed -- but it's not clear we can do that, as the finally target bit is used to inhibit various optimizations that might break unwind behavior. The semantics are subtle so it's not obvious if there are other ways to achieve the same goals (easily) without the finally target bit.

AndyAyersMS:
Yeah, I looked into late marking, but it seemed like it would require a prepass (since finally targets BBs might appear before any of the corresponding callfinallys) and also we'd have to sort through anything sensitive in between. So it seemed simpler to just fix things immediately.

AndyAyersMS:
Unexpected failure [(log)](https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_windows_nt_prtest/2221/) that seems to be GC related:
```
11:14:23          "D:\j\workspace\debug_windows---17180f1b\tests\..\bin\tests\Windows_NT.x64.Debug\Tests\Core_Root\corerun.exe" HandleCopy.exe 
11:14:23         Expected: 100
11:14:23         Actual: -532462766
```

AndyAyersMS:
@gkhanna79 any way to test this more completely before merging? Trying to avoid similar issues to the ones flagged in #8989, where the builds that broke don't seem to run in CI.

jashook:
@dotnet-bot test Windows_Arm64 debug

jashook:
@dotnet-bot test Windows_NT Arm64 debug

AndyAyersMS:
@dotnet-bot test Windows_NT Arm64 checked

AndyAyersMS:
Looks like there is no checked arm64 test? Hmm. Not clear if debug will expose issues since in such builds the core library probably doesn't get optimized, and likewise, release won't have the asserts.

AndyAyersMS:
Am going to merge this and keep an eye on things that don't seem well tested in CI.

PR #9172, opened 2017-01-27
Android - Fix duplicate definitions of INT_MIN and PAGE_SIZE in unit tests

qmfrederik:
Both `INT_MIN` and `PAGE_SIZE` already exist on Android, but are redefined for the unit tests. This PR fixes that conflict.

### DIFF
--- a/src/pal/tests/palsuite/threading/NamedMutex/test1/nopal.cpp
+++ b/src/pal/tests/palsuite/threading/NamedMutex/test1/nopal.cpp
@@ -16,7 +16,9 @@
 
 #define _countof(a) (sizeof(a) / sizeof(a[0]))
 
+#ifndef __ANDROID__
 #define PAGE_SIZE (4096)
+#endif
 
 auto test_strcpy = strcpy;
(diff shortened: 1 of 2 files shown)


qmfrederik:
@janvorli Another Android PR. Alternatives are:

```
#ifndef INT_MIN
#define INT_MIN (-2147483647 - 1)
#endif
```

This one has the advantage of being working on other platforms where INT_MIN is defined, but re-using the platform definition of INT_MIN

or
```
#undef INT_MIN
#define INT_MIN (-2147483647 - 1)
```

This one is perhaps the most portable one?

janvorli:
@qmfrederik Since it is in the tests, I don't really care which way we choose. But I would slightly prefer the the latter.

qmfrederik:
@janvorli OK, done

qmfrederik:
@janvorli Is this one good to go? Anything else you need from my side?

janvorli:
@qmfrederik I am sorry, I've just lost track of this one.

PR #9175, opened 2017-01-27
Add scripts & docs which support creating an Android toolchain on Linux

qmfrederik:
This PR kind of wraps up the changes required to get CoreCLR to cross compile for Android.

I've included documentation and a shell scripts which help you construct a toolchain to enable cross-building for Android.

It includes
- Downloading & extracting the NDK
- Downloading & extracting additional libraries which are required to cross compile from Termux. Termux seemed to be a convenient source for precompiled versions of most of the libraries required; happy to discuss alternatives if there are any
- Building libunwind from source. libunwind is built form the Android fork of libunwind. Although Android ships with libunwind, the libunwind headers and libraries are not available in the NDK so this step is requried.
- Downloading the lldb server which enables remote debugging on Android. Although not required to build CoreCLR for Android, it probably helps to have it available
- Instructions on how to copy the output to an Android device and run the PAL test.

The script tries to avoid downloading the same binaries twice, so you can re-run it fairly easily if required.


qmfrederik:
@cydhaselton You had some feedback on the script so feel free to comment on this PR :)

cross/android/.gitignore
```diff
@@ -0,0 +1,8 @@
+deb/
+tmp/
+toolchain/
+lldb/
+android-ndk-*/
+libunwind/
+*.zip
+
```
janvorli: The guidance for coreclr is to have just a single .gitignore in the root. 
I think that we should add just the cross/android/* in there, like we do for the cross/rootfs/*

qmfrederik:
@janvorli Thanks for the feedback!

I moved & renamed the init-android script as you suggested, and I've also taken the time to make it a bit more flexible.

The script writes out how you could start a CoreCLR build but does not actually start one.

The `.gitignore` structure has also been updated. I've kept `cross/android/` for the CMake files, and used cross/android-rootfs/ for everything related to the `rootfs/toolchains`. It also serves as a cache for files that are downloaded (NDK, Termux packages,...). The actual toolchain ends up in `cross/android-rootfs/toolchain/arm64`.

Let me know if you have additional feedback!

qmfrederik:
@cydhaselton You can now also specify the API level; it's set to 21 by default.

cydhaselton:
@qmfrederik and @janvorli There were some problems in the `init-android.sh` script I tested; notably with the unpacking of Android-targeted dependencies; I'll re-test with these changes and report back.

The script generating the cross-toolchain should target the latest Android API level by default; Google tends to introduce breaking changes that cause binaries compiled for earlier levels to fail on later levels.

cydhaselton:
@qmfrederik Unpacking Termux packages still fail; which will cause the build to fail as well.

qmfrederik:
@cydhaselton On which OS are you running those scripts? The scripts use `dpkg`. Is that available on your Linux distribution? What's the error message you are seeing?

cydhaselton:
@qmfrederik I'm running them on Ubuntu 14.04 and dpkg is installed. You can see the error in the build log I linked to in the cross build thread; sample line:
```
dpkg-deb: error: archive '/build/core2/coreclr/cross/android-rootfs/deb/libandroid-support_13.10_aarch64.deb' has premature member 'control.tar.xz' before 'control.tar.gz', giving up
```

cydhaselton:
@qmfrederik Additionally, the script doesn't download the NDK. Originally the `wget` commands errored out because `--show-progress` isn't a valid parameter; I removed them from the script.

qmfrederik:
@cydhaselton `--show-progress` is a wget command, but it has [been introduced in version 1.16](http://savannah.gnu.org/forum/forum.php?forum_id=8133). That version is available in [Ubuntu 16.04 and above](http://packages.ubuntu.com/search?keywords=wget&searchon=names&suite=all&section=all).

Can you run the `dpkg` command on the `.deb` packages if you download them directly from the Termux repository using another mechanism? If that also fails, perhaps `dpkg` in Ubuntu 14.04 has an issue with the Termux packages? Is there a reason you're on Ubuntu 14.04? The current LTS version is 16.04.

cydhaselton:
@qmfrederik The Linux build instructions still cite 14.04 as preferred version

> These instructions are written assuming the Ubuntu 14.04 LTS, since that's the distro the team uses. Pull Requests are welcome to address other environments as long as they don't break the ability to use Ubuntu 14.04 LTS.
> 
> There have been reports of issues when using other distros or versions of Ubuntu (e.g. Issue 95). If you're on another distribution, consider using docker's ubuntu:14.04 image.

If that's changed, i.e. the build instructions are out of date, please confirm and I'll upgrade before continuing work.

qmfrederik:
@cydhaselton Those instructions are for building CoreCLR for Ubuntu; in this case, we're building CoreCLR for Android. To get CoreCLR running on Android, you can safely ignore these instructions.

qmfrederik:
@cydhaselton Regarding the standard API level, I don't think there is a problem as you can pass the desired API level to the script. So if you want to cross build for the exact same API level as you have on your phone, this PR allows you to do that.

That said, the [Android changes for NDK developers](https://android-developers.googleblog.com/2016/06/android-changes-for-ndk-developers.html) document lays out which rules you need to respect so you can compile for one API level and still have your program run at another API level. It is correct that Android has implemented additional restrictions (no use of private API, textrels,...). The build of CoreCLR for Android as currently implemented respect those rules.

Termux does the same thing: they also [generate a toolchain for API Level 21](https://github.com/termux/termux-packages/blob/master/build-package.sh#L174) and distribute the same binaries for all versions of Android.

cydhaselton:
@qmfrederik: Will re-test script post-upgrade of build system and post results.

cydhaselton:
@qmfrederik Post-upgrade script runs successfully. The next error occurs when running the post-script command:
```
Invoking "/build/core2/coreclr/src/pal/tools/gen-buildsys-clang.sh" "/build/core2/coreclr" 3 6 arm64 Debug  Include_Tests  -DCLR_CMAKE_TARGET_OS=Linux -DCLR_CMAKE_PACKAGES_DIR=/build/core2/coreclr/packages -DCLR_CMAKE_PGO_INSTRUMENT=0 -DENABLE_LLDBPLUGIN=0
Including tests directory in build.
loading initial cache file /build/core2/coreclr/cross/android/arm64/tryrun.cmake
CMake Error: Error processing file: /build/core2/coreclr/cross/android/arm64/tryrun.cmake
CMake Error at /usr/share/cmake-3.5/Modules/CMakeDetermineSystem.cmake:104 (message):
  Could not find toolchain file:
  /build/core2/coreclr/cross/android/arm64/toolchain.cmake
Call Stack (most recent call first):
  CMakeLists.txt:9 (project)
 
 
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
```
I'm working with the toolchain-scripts branch.

qmfrederik:
@janvorli Is this one good to go?

cydhaselton:
@qmfrederik Still seeing errors with Android  cross-compile build scripts; specifically results when building with the prompted command. See http://pastebin.com/eEwrWxhm for errors.

cydhaselton:
@qmfrederik Ping for above issue. Let me know if I should open this as a separate issue.

qmfrederik:
@cydhaselton Sorry I didn't have the chance to look at this until now. I just tried with the latest master from coreclr (one last PR got merged in), I didn't see this error.

Can you try to check out the latest coreclr/master, remove the bin folder (or just run git clean -xfd) and try again?

cydhaselton:
@qmfrederik Done and same errors.  Ran `git fetch upstream && git merge upstream master`. Created a new branch, ran `./cross/build-android-rootfs.sh` then the ./build.sh command and got the same errors. Then ran `git clean -xfd` and started again with the build-android-rootfs.sh script. Same error again.

qmfrederik:
@cydhaselton I'm sorry for all the issues you're hitting trying to get the cross compile working; I wish things were easier.

I spent some time today making sure I have a clear and reproducible path to get the cross build working. I'm using a Vagrant box (yes, yet another technology) this time to basically create a VM from scratch and cross-build CoreCLR and CoreFX from Android.

I've seen the errors you've reported a couple of times, but they always were because either the `CONFIG_DIR` or `ROOTFS_DIR` variable was not set correctly. (In your case, it looks like at least `ROOTFS_DIR` is set correctly because it's picking up clang from the NDK folder).

If it helps, [here's the Vagrant file I'm using](https://github.com/qmfrederik/coredroid/blob/master/box/Vagrantfile). It basically starts from a clean Ubuntu 16.04 box, clones the repositories and then starts the build process.

Net, what is being executed is this, on Ubuntu 16.04:

```
sudo apt-get update
sudo apt-get install -y unzip python build-essential autoconf libtool cmake git wget libicu55 libunwind8 clang-3.6

mkdir -p ~/git
cd ~/git
git clone https://github.com/dotnet/coreclr
git clone https://github.com/dotnet/corefx
git clone https://github.com/qmfrederik/coredroid

cd ~/git/coreclr
cross/build-android-rootfs.sh
cd ~/git/corefx
cross/build-android-rootfs.sh
cd ~/git/coreclr
./build.sh
CONFIG_DIR=`realpath cross/android/arm64` ROOTFS_DIR=`realpath cross/android-rootfs/toolchain/arm64/sysroot` ./build.sh cross arm64 skipgenerateversion skipnuget
cd ~/git/corefx
CONFIG_DIR=`realpath cross/android/arm64` ROOTFS_DIR=`realpath cross/android-rootfs/toolchain/arm64/sysroot` ./build-native.sh -debug -buildArch=arm64 -- verbose cross

```

cydhaselton:
@qmfrederik:  Two questions:

1. Should `CONFIG_DIR` and `ROOTFS_DIR`  be exported as env variables?
2. You have an extra `./build.sh` in the coreclr build instructions above:
```
cd ~/git/coreclr
./build.sh
CONFIG_DIR=`realpath cross/android/arm64` ROOTFS_DIR=`realpath cross/android-rootfs/toolchain/arm64/sysroot` ./build.sh cross arm64 skipgenerateversion skipnuget cmakeargs -DENABLE_LLDBPLUGIN=0
cd ~/git/corefx
CONFIG_DIR=`realpath cross/android/arm64` ROOTFS_DIR=`realpath cross/android-rootfs/toolchain/arm64/sysroot` ./build-native.sh -debug -buildArch=arm64 -- verbose cross
```
Was this intentional?

qmfrederik:
@cydhaselton You can either export them as env variables, or just add them before the call to the `./build.sh` script like I've done in the code I posted. Both should work.

Regarding the double `./build.sh`, I though I needed it at one point in time, but I'm pretty sure you don't need to do it - if you look at the Vagrant box, the double `./build.sh` is not there and the build succeeds.

cydhaselton:
@qmfrederik One last question...obvious but I want to make absolutely sure I'm repro-ing this correctly...`realpath` is the root path to the various directories...i.e. `CONFIG_DIR=/some/path/to/cross/android/arm64`, correct?

qmfrederik:
`realpath` is an Ubuntu program that will echo the "real" (or full) path back. So if you type:
`realpath cross/android/arm64` the output should be something like `/home/me/git/coreclr/cross/android/arm64`. Because the ticks are around the command, bash will evaluate the command for you.

So the script should work as is, but it should be equivalent to:

```
export CONFIG_DIR=/home/me/git/coreclr/cross/android/arm64
export ROOTFS_DIR=/home/me/git/coreclr/cross/android-rootfs/toolchain/arm64
./build.sh cross arm64 skipgenerateversion skipnuget cmakeargs -DENABLE_LLDBPLUGIN=0
```

If you do it this way, make sure update the env variable CONFIG_DIR and ROOTFS_DIR as you switch between corefx and coreclr.

cydhaselton:
@qmfrederik:  Can't say why this worked over the explicit path, but running `CONFIG_DIR=`realpath cross/android/arm64` ROOTFS_DIR=`realpath cross/android-rootfs/toolchain/arm64/sysroot` ./build.sh cross arm64 skipgenerateversion skipnuget cmakeargs -DENABLE_LLDBPLUGIN=0` worked this time; coreclr build is progressing

qmfrederik:
Cool, let me know how it goes!

cydhaselton:
@qmfrederik The build failed to generate Nuget packages, and threw an error that I assumed was related
(An error occured in _pathspec_ _pathspec_/run 11.  Check 11 logs under _pathspec_) but otherwise succeeded.

qmfrederik:
Sounds good! I think you can safely ignore the NuGet package failure.

Did you manage to cross-build CoreFX?

Out of curiosity, did you have `skipnuget` in your command line arguments? That should have caused the build to skip creating the NuGet packages.

cydhaselton:
I did do the corefx cross-build
How did you handle the mscorlib build for coreclr?

qmfrederik:
@cydhaselton If you did not specify `skipcorlib` (I think you did not) when cross-building coreclr, you should already have a cross-built mscorlib.

cydhaselton:
@qmfrederik Will test hello world on termux then; though I suspect I'll still get the initialization 80004005 error. I'll look into disabling named pipes, re [this](https://github.com/dotnet/coreclr/issues/9780#issuecomment-282480543)

cydhaselton:
@qmfrederik backing up a few posts; your build script instructs to use the `skipmscorlib` param, so I did use it when cross building for Android.  Have you successfully cross-built without specifying `skipmscorlib`?

cydhaselton:
mscorlib does build, although with several MSB3644 warnings for NETPortable 4.0 and NETFramework 4.5

janvorli:
@cydhaselton The warnings are benign, they happen for x64 build as well.

Issue #7894, opened 2016-10-30
Span<T>.Slice method is hitting asserts in JIT on Unix


jkotas:
`Assertion failed '((regMask & emitThisGCrefRegs) && (ins == INS_add)) || ((regMask & emitThisByrefRegs) && (ins == INS_add || ins == INS_sub))' in 'System.Span`1[Byte][System.Byte]:Slice(int,int):struct:this' (IL size 44)`

jkotas:
I am adding `[MethodImpl(MethodImplOptions.NoOptimization)]` workaround in #7886 that should be removed once this is fixed.

jkotas:
Note that even though this was hit by the new Span code, there is nothing Span-specific in the repro. This looks like a general codegen bug.

To repro:

- Revert https://github.com/dotnet/coreclr/commit/4c5e9791dcc5b082ae01abc437dcb7456ae3e806
- crossgen System.Private.CoreLib on Unix

pgavlin:
cc @russellhadley

sivarv:
The issue here is Jit-EE interface call `getFieldType()` is returning `CORINFO_TYPE_NATIVEINT` (instead of `CORINFO_TYPE_BYREF`) as the type of  field `ByReference<T>._value`.  As a result, when JIT struct promotes a var of `ByReference<T>`, it uses type int as the type of field of `_value`.  This leads to incorrect gc-reporting and surfaces as an assert in emitter.

sivarv:
Another related issue is that JIT-EE call getClassAttribs() on ByReference<T> struct is not returning CORINFO_FLG_CONTAINS_GC_PTR in the set of flags returned.

jkotas:
> `getClassAttribs()` on `ByReference<T>` struct is not returning `CORINFO_FLG_CONTAINS_GC_PTR` in the set of flags returned.

The contract of `getClassAttribs()` that we agreed on is that `CORINFO_FLG_CONTAINS_STACK_PTR` will signal to the JIT that the struct may contain byrefs. `CORINFO_FLG_CONTAINS_GC_PTR` is only set for structs that do contain regular references.

There is comment about it here: https://github.com/dotnet/coreclr/blob/master/src/jit/importer.cpp#L1492

sivarv:
Ok. JIT has one more assert here: https://github.com/dotnet/coreclr/blob/master/src/jit/lclvars.cpp#L1595
that too needs to fixed to account for `ByReference<T>`

[...]

PR #9182, opened 2017-01-28
Renaming Span/ReadOnlySpan Extension classes to avoid conflict with CoreFX.

ahsonkhan:
This change enables the following Corefx PR to succeed:
https://github.com/dotnet/corefx/pull/15569


ahsonkhan:
@dotnet-bot Test Windows_NT x86 Checked Build and Test please.

ahsonkhan:
cc @jkotas @kouvel

ahsonkhan:
@dotnet-bot Test Windows_NT x86 Checked Build and Test please.

ahsonkhan:
@dotnet-bot Test Ubuntu x64 Checked Build and Test please.

ahsonkhan:
@dotnet-bot Test OSX x64 Checked Build and Test please.

ahsonkhan:
@dotnet-bot Test Windows_NT x86 Checked Build and Test please.
@dotnet-bot Test Ubuntu x64 Checked Build and Test please.
@dotnet-bot Test OSX x64 Checked Build and Test please.

jkotas:
crossgen is crashing with JIT assert:

```
Assert failure(PID 11680 [0x00002da0], Thread: 37004 [0x908c]): Assertion failed '((regMask & emitThisGCrefRegs) && (ins == INS_add)) || ((regMask & emitThisByrefRegs) && (ins == INS_add || ins == INS_sub))' in 'System.Span`1[Byte][System.Byte]:GetItem(int):byref:this' (IL size 35)
```

src/mscorlib/src/System/ReadOnlySpan.cs
```diff
@@ -328,7 +328,7 @@ public T[] ToArray()
         public static ReadOnlySpan<T> Empty => default(ReadOnlySpan<T>);
     }
 
-    public static class ReadOnlySpanExtensions
+    public static class ReadOnlySpanExtensionMethods
```
jkotas: This does not feel right. I think we need a better plan for factoring the Span extension methods than this as I have mentioned over email.

kouvel:
@russellhadley / @sivarv, could one of you please take a look at the JIT assertion failure mentioned by @jkotas above?

src/mscorlib/src/System/ReadOnlySpan.cs
kouvel: How about removing the extensions classes from CoreCLR altogether and using the CoreFX implementations (with whichever appropriate refactoring of the classes we decide upon in CoreFX).

ahsonkhan: @jkotas, does this work? Shall I move all these extension methods over to corefx?

jkotas: There are two kinds of extension methods:

- Low-level extension methods that need access to runtime internals. Moving these to corefx is a hassle because of them you need to expose the runtime internals instead.
- A large number of high-level extension methods. We want to have them in corefx to allow duplicated sources.

I think we want to have several different types for different kinds of extension methods, as it is being discussed in the email thread parallel to this issue. Krzysztof and Immo should make a call on the naming.

ahsonkhan: cc @KrzysztofCwalina, @terrajobst

ahsonkhan: @jkotas, can this discussion/change be part of a separate PR. I want to get fast span enabled.

sivarv:
@ahsonkhan , @kouvel - Can you please provide me binaries for repro'ing JIT assert failure?

jkotas:
@sivarv Run this to repro the JIT assert:

```
git remote add ahsonkhan
git fetch ahsonkhan
git checkout ResolvingSpanImpls
build.cmd x86 skiptests
```

jkotas:
Scratch that ... it won't work because of the offending piece of code was removed from the PR.

src/mscorlib/src/System/ReadOnlySpan.cs
jkotas: Once you get the fast span enabled, changing the name will be pain. I would recommend getting this resolved before enabling fast span.

kouvel:
@sivarv you should be able to sync to commit 27b01ed9c3e4084ab18a281e23a069036739a590 from this branch, which caused the failure

KrzysztofCwalina:
Should we call the non-portable extensions holder simply "Span" and then the actual extensions (portable) would be called SpanExtensions? SpanExtensionMethods is introducing a new naming pattern, and I don't think we should do it. 

cc: @terrajobst

ahsonkhan:
@dotnet-bot Test Windows_NT x64 Debug Build and Test please.

ahsonkhan:
@KrzysztofCwalina, what about the Slice ReadOnlySpan Extension methods?

sivarv:
@ahsonkhan @kouvel, @jkotas - I have syned to the specific commit on ResolvingSpanImpl branch and tried building x86 and x64 both check/debug versions and each time crossgen of System.Private.Corelib.dll succeeded.

Can any of you could repro this on your local machines?

ahsonkhan:
> @ahsonkhan @kouvel, @jkotas - I have syned to the specific commit on ResolvingSpanImpl branch and tried building x86 and x64 both check/debug versions and each time crossgen of System.Private.Corelib.dll succeeded.

Can any of you could repro this on your local machines?

I had tried to build locally a couple of times (just ran build.cmd on coreclr directory) and didn't see any build failures, but the CI kept failing.

jkotas:
Sorry, forgot that you have to do checked build to see the problem: `build.cmd x86 checked skiptests`

jkotas:
The JIT crash maybe dup of https://github.com/dotnet/coreclr/issues/7894. You can try adding NoOptimization on GetItem to workaround it - the same workaround is used in other places.

ahsonkhan:
@jkotas, can this be merged in once the CI passes?

sivarv:
@ahsonkhan @jkotas - I have repro'ed the assert failure on my machine and looking into it.

sivarv:
The issue here is Jit-EE interface call `getFieldType()` is returning `CORINFO_TYPE_NATIVEINT` (instead of `CORINFO_TYPE_BYREF`) as the type of  field `ByReference<T>._value`.  As a result, when JIT struct promotes a var of `ByReference<T>`, it uses type int as the type of field of `_value`.  This leads to incorrect gc-reporting and surfaces as an assert in emitter.

Since already have issue #7894 tracking this, I have updated and assigned it to @kouvel

PR #9184, opened 2017-01-28
Add BenchmarkGames regexdna test

AndyAyersMS:
Based on the "version 6" test base. As with revcomp, keeping the input
files as unix line ending, and relatively small.


AndyAyersMS:
@sivarv PTAL
cc @dotnet/jit-contrib

sivarv:
Looks good.

Issue #9167, opened 2017-01-27
Make EnsureSufficientExecutionStack less conservative


jkotas:
> @jcouv 

The compiler is using EnsureSufficientExecutionStack. From what I can tell, that method throws when the stack is only a bit more than half full. In the repro I’m investigating, the stack when the exception is thrown is about 520kB long. When I remove the EnsureSufficientExecutionStack check, the overflow occurs when the stack is full (at 1MB).

Is there any way to do something like EnsureSufficientExecutionStack but not give up on half of the potential stack usage?

> @jkotas

I agree that it may be too conservative and it may be worth it to make it less conservative, e.g. return true if there is 128k (on 64-bit platforms) / 64k (on 32-bit platforms) available.

jcouv:
I didn't follow how https://github.com/dotnet/corefx/pull/15600 fixed this. That PR only has test changes. Was the problem resolved earlier and the new tests just verify it?

kouvel:
Yes these are just tests for the functional fix in PR dotnet/coreclr#9186

Issue #9158, opened 2017-01-27
BCL Assert(maxStackSize == 0) in System.Threading.Thread.SetStartHelper should not be an assertion


RussKeldorph:
Found due to [this](https://github.com/dotnet/corefx/blob/88d8bbf7dbf5869b7f239737779e75295e43ef95/src/System.Linq.Expressions/tests/CompilerTests.cs#L45) line.

Per [this](https://github.com/dotnet/coreclr/blob/a928a37a1503871b5da5c80f74fe54b9f937ddc3/src/mscorlib/src/System/Threading/Thread.cs#L431) line, CoreCLR only supports the default stack size.  However, it seems wrong to enforce this via assertion.  Shouldn't [this](https://github.com/dotnet/coreclr/blob/a928a37a1503871b5da5c80f74fe54b9f937ddc3/src/mscorlib/src/System/Threading/Thread.cs#L202) and [this](https://github.com/dotnet/coreclr/blob/a928a37a1503871b5da5c80f74fe54b9f937ddc3/src/mscorlib/src/System/Threading/Thread.cs#L219) be changed to catch this programming error?

JonHanna:
Is the assertion valid? That test was succeeding in its purpose of reaching the point where `EnsureSufficientExecutionStack` throws very quickly, compared with the outerloop `CompileDeepTree_NoStackOverflow` which does the same thing in the test-runner's thread's stack, and takes much longer to do so. That would suggest it was indeed getting a smaller than default stack.

RussKeldorph:
@JonHanna It looks like the Thread constructors offering the maxStackSize parameter aren't documented for .NET Core and therefore shouldn't be used.  If so, then an assertion may be appropriate.  Unfortunately, I don't know how the restricted Core API is supposed to be enforced and whether the test in dotnet/corefx/issues/15533 may be exploiting a back door.  Seems like ideally the test would fail to compile.

/cc @dotnet/coreclr-contrib @adityamandaleeka @janvorli

jkotas:
>  It looks like the Thread constructors offering the maxStackSize parameter aren't documented for .NET Core

Thread constructors that take maxStackSize were added back as part of NetStandard2.0 work as official public documented APIs:

https://github.com/dotnet/corefx/blob/ddd1e5ce0f81f7091e544c65f99c37bf52a4ded9/src/System.Threading.Thread/ref/System.Threading.Thread.cs#L37

[...]

PR #9186, opened 2017-01-28
Fix Thread constructor with stack size argument

kouvel:
Functional fix for #9158 and #9167

### DIFF
--- a/src/inc/switches.h
+++ b/src/inc/switches.h
@@ -102,9 +102,20 @@
 #define OS_PAGE_SIZE PAGE_SIZE
 #endif
 
+#ifndef THREAD_MIN_EXECUTION_STACK_SIZE
+// This stack size is sufficient to allow the average Framework function to run, and to allow us to throw and dispatch an
+// exception up a reasonable call chain
+#ifdef BIT64
+#define THREAD_MIN_EXECUTION_STACK_SIZE (128 * 1024)
+#else // !BIT64
+#define THREAD_MIN_EXECUTION_STACK_SIZE (64 * 1024)
+#endif // BIT64
+#endif // !THREAD_MIN_EXECUTION_STACK_SIZE
+
 #ifndef THREAD_MIN_STACK_SIZE
-#define THREAD_MIN_STACK_SIZE (64 * 1024) // 64 KB is a reasonably portable minimum between Windows and Unix
-#endif
+// Make sure a thread has some extra stack space prior to reaching the minimum execution stack size
+#define THREAD_MIN_STACK_SIZE (THREAD_MIN_EXECUTION_STACK_SIZE * 2)
+#endif // !THREAD_MIN_STACK_SIZE
 
 #if defined(_WIN64)
 #define JIT_IS_ALIGNED
--- a/src/inc/switches.h
+++ b/src/inc/switches.h
@@ -102,21 +102,6 @@
 #define OS_PAGE_SIZE PAGE_SIZE
 #endif
 
-#ifndef THREAD_MIN_EXECUTION_STACK_SIZE
-// This stack size is sufficient to allow the average Framework function to run, and to allow us to throw and dispatch an
-// exception up a reasonable call chain
-#ifdef BIT64
-#define THREAD_MIN_EXECUTION_STACK_SIZE (128 * 1024)
-#else // !BIT64
-#define THREAD_MIN_EXECUTION_STACK_SIZE (64 * 1024)
-#endif // BIT64
-#endif // !THREAD_MIN_EXECUTION_STACK_SIZE
-
-#ifndef THREAD_MIN_STACK_SIZE
-// Make sure a thread has some extra stack space prior to reaching the minimum execution stack size
-#define THREAD_MIN_STACK_SIZE (THREAD_MIN_EXECUTION_STACK_SIZE * 2)
-#endif // !THREAD_MIN_STACK_SIZE
-
 #if defined(_WIN64)
 #define JIT_IS_ALIGNED
 #endif
--- a/src/inc/switches.h
+++ b/src/inc/switches.h
@@ -102,6 +102,10 @@
 #define OS_PAGE_SIZE PAGE_SIZE
 #endif
 
+#ifndef THREAD_MIN_STACK_SIZE
+#define THREAD_MIN_STACK_SIZE (64 * 1024) // 64 KB is a reasonably portable minimum between Windows and Unix
+#endif
+
 #if defined(_WIN64)
 #define JIT_IS_ALIGNED
 #endif
--- a/src/pal/src/thread/thread.cpp
+++ b/src/pal/src/thread/thread.cpp
@@ -35,6 +35,7 @@ SET_DEFAULT_DEBUG_CHANNEL(THREAD); // some headers have code with asserts, so do
 #include "pal/environ.h"
 #include "pal/init.h"
 #include "pal/utils.h"
+#include "pal/virtual.h"
 
 #if defined(__NetBSD__) && !HAVE_PTHREAD_GETCPUCLOCKID
 #include <sys/cdefs.h>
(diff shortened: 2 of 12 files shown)


kouvel:
@janvorli @jkotas

src/inc/switches.h
```diff
@@ -102,6 +102,10 @@
 #define OS_PAGE_SIZE PAGE_SIZE
 #endif
 
+#ifndef THREAD_MIN_STACK_SIZE
+#define THREAD_MIN_STACK_SIZE (64 * 1024) // 64 KB is a reasonably portable minimum between Windows and Unix
+#endif
```
stephentoub: @jkotas mentions 128k in https://github.com/dotnet/coreclr/issues/9167 with regards to sufficient stack size on 64-bit... ?

kouvel: Thanks, fixed

kouvel:
@dotnet-bot test this please

src/pal/src/thread/thread.cpp
```diff
@@ -626,20 +644,28 @@ CorUnix::InternalCreateThread(
...
         WARN("default stack size is reported as %d, but PTHREAD_STACK_MIN is "
              "%d\n", pthreadStackSize, PTHREAD_STACK_MIN);
     }
 #endif
-    
-    if (pthreadStackSize < dwStackSize)
+
+    if (dwStackSize == 0)
+    {
+        dwStackSize = CPalThread::s_dwDefaultThreadStackSize;
```
jkotas: Why do we have our own default instead of just using pthread default?

src/vm/threads.cpp
```diff
@@ -7317,10 +7322,13 @@ BOOL Thread::SetStackLimits(SetStackLimitScope scope)
...
         // Compute the limit used by EnsureSufficientExecutionStack and cache it on the thread. The limit
-        // is currently set at 50% of the stack, which should be sufficient to allow the average Framework
-        // function to run, and to allow us to throw and dispatch an exception up a reasonable call chain.
-        m_CacheStackSufficientExecutionLimit = reinterpret_cast<UINT_PTR>(m_CacheStackBase) - 
-            (reinterpret_cast<UINT_PTR>(m_CacheStackBase) - reinterpret_cast<UINT_PTR>(m_CacheStackLimit)) / 2;
+        // is should be sufficient to allow the average Framework function to run, and to allow us to throw
+        // and dispatch an exception up a reasonable call chain.
+        _ASSERTE(m_CacheStackBase > m_CacheStackLimit);
+        _ASSERTE(
+            (reinterpret_cast<UINT_PTR>(m_CacheStackBase) - reinterpret_cast<UINT_PTR>(m_CacheStackLimit)) >
```
jkotas: This assert is not correct. It can fire for external threads that entered the runtime - they can be created with arbitrary small stacks.

src/vm/threads.cpp
```diff
@@ -7317,10 +7322,13 @@ BOOL Thread::SetStackLimits(SetStackLimitScope scope)
         }
 
         // Compute the limit used by EnsureSufficientExecutionStack and cache it on the thread. The limit
-        // is currently set at 50% of the stack, which should be sufficient to allow the average Framework
-        // function to run, and to allow us to throw and dispatch an exception up a reasonable call chain.
-        m_CacheStackSufficientExecutionLimit = reinterpret_cast<UINT_PTR>(m_CacheStackBase) - 
-            (reinterpret_cast<UINT_PTR>(m_CacheStackBase) - reinterpret_cast<UINT_PTR>(m_CacheStackLimit)) / 2;
+        // is should be sufficient to allow the average Framework function to run, and to allow us to throw
```
jkotas: I know that this is the description for EnsureSufficientExecutionStack from MSDN, but I think it is confusing and it does not match what EnsureSufficientExecutionStack is used for in practice. What it should actually do is to ensure that there is enough stack for typical non-recursive call chain to execute, including potential exception handling or garbage collection.

src/vm/threads.cpp
```diff
@@ -3060,7 +3060,12 @@ BOOL Thread::CreateNewOSThread(SIZE_T sizeToCommitOrReserve, LPTHREAD_START_ROUT
     DWORD dwCreationFlags = CREATE_SUSPENDED;
 
 #ifdef FEATURE_CORECLR
-    dwCreationFlags |=  STACK_SIZE_PARAM_IS_A_RESERVATION;
+    dwCreationFlags |= STACK_SIZE_PARAM_IS_A_RESERVATION;
+
+    if (sizeToCommitOrReserve != 0 && sizeToCommitOrReserve < THREAD_MIN_STACK_SIZE)
+    {
+        sizeToCommitOrReserve = THREAD_MIN_STACK_SIZE;
```
jkotas: I do not think that this minimum should be coupled with the amount used by EnsureSufficientExecutionStack. The idea about this minimum is just to prevent guaranteed crash when the specific stack size is too small. It should be smaller than what EnsureSufficientExecutionStack ensures . Thinking about this some more ... I am not sure whether we should have one here. It implies policy that always hurt us because of there is always somebody out there who does not like it. It may be better to just pass the value down (align it up to make the OS API happy) and crash if the value is too small.

jkotas: Either way, it would be useful to add some comments about why we are doing things in certain way.

src/pal/src/thread/thread.cpp
kouvel: I believe it is in case the pthread default is smaller than what CoreCLR intends to use for a default. As the pthread default is configurable by the user and probably appropriate for the system, I suppose it would be better to just use that. I'll remove the hard-coded default.

src/vm/threads.cpp
kouvel: Nice catch, thanks

kouvel:
Made all of the above changes

jkotas:
Build error on Linux ARM:

```
22:40:47 In file included from /mnt/j/workspace/dotnet_coreclr/master/arm_emulator_cross_debug_ubuntu_prtest/src/pal/src/thread/thread.cpp:24:
22:40:47 /mnt/j/workspace/dotnet_coreclr/master/arm_emulator_cross_debug_ubuntu_prtest/src/pal/../inc/stdmacros.h:21:10: fatal error: 'specstrings.h' file not found
```

src/pal/src/thread/thread.cpp
```diff
@@ -21,6 +21,8 @@ Module Name:
 #include "pal/dbgmsg.h"
 SET_DEFAULT_DEBUG_CHANNEL(THREAD); // some headers have code with asserts, so do this first
 
+#include "stdmacros.h"
```
jkotas: Do you still need it to be included here with the most recent changes? I would rather delete this `#include` than try to make it compile. It is generally a bad idea to include headers from src\inc in the PAL.

src/pal/src/thread/thread.cpp
```diff
@@ -573,7 +568,22 @@ CorUnix::InternalCreateThread(
...
+
+    if (dwStackSize != 0)
+    {
+        // Some systems require the stack size to be aligned to the page size
+        DWORD alignedStackSize = static_cast<DWORD>(ALIGN_UP(dwStackSize, OS_PAGE_SIZE));
+        if (alignedStackSize < dwStackSize)
+        {
+            // When coming here from the public API surface, the incoming value is originally a nonnegative signed int32, so
+            // this shouldn't happen
+            ASSERT("Couldn't align the requested stack size (%u) to the page size\n", dwStackSize);
```
janvorli: Since the pthread_attr_setstacksize accepts size_t parameter, I would do this slightly differently so that the problem you assert for cannot happen. I would create size_t alignedStackSize first, then compute it from dwStackSize cast to size_t and then use it in all of the rest of the function code instead of dwStackSize. That way, the condition you are asserting for cannot happen.

src/pal/src/thread/thread.cpp
janvorli: I agree. And if we need something from the stdmacros.h, I would rather redefine it for PAL usage instead.
The change in the stdmacros.h should also be reverted then.

src/pal/src/thread/thread.cpp
kouvel: On 32-bit platforms, DWORD and size_t have the same size and both are unsigned, so it could still overflow even after casting to size_t. I could make the check specific for 32-bit.

janvorli: Oops, right. So please keep it as it is. But in that case, it seems it would make sense to change the message to say that the problem was the stack size being too large, since it is the only case when the attempt to align would fail.

kouvel: There was another issue, ALIGN_UP would fail an assert if it overflows. I have made some fixes.

kouvel:
Made changes from above

kouvel:
Rebased

jcouv:
Thanks!

PR #8969, opened 2017-01-17
Fast-path for ASCII & UTF8 Encoding ASCII data

benaadams:
Up to x4 faster on `Encoding.GetBytes(String s)` for UTF8 and ASCII when string is ascii.

Have `#if !BIGENDIAN` the changes as don't have have a good way of confirming the pattern for little endian.

Call chains changed mostly to Virtual call->validate->concrete call; previously it would do virtual call chains with validations on each call.

The string overload has a wrinkle. It will optimistically take `string.length` as the length of they output bytes; if it encounters a non-ascii char it will then count from that point forward, reallocate the array and copy the bytes to the correct sized array; then pass over to the current implementation to complete - so you maybe get double `byte[]` allocations. The array versions don't exhibit this behaviour.

Following up on https://github.com/dotnet/coreclr/issues/6759 /cc @jamesqo

PR #9187, opened 2017-01-28
Fast-path for ASCII & UTF8 Encoding ASCII data

benaadams:
Fixed https://github.com/dotnet/coreclr/pull/8969

### DIFF
--- a/src/mscorlib/src/System/Text/ASCIIEncoding.cs
+++ b/src/mscorlib/src/System/Text/ASCIIEncoding.cs
@@ -2,11 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+
 namespace System.Text
 {
     using System;
-    using System.Runtime.Serialization;
-    using System.Diagnostics;
     using System.Diagnostics.Contracts;
 
     // ASCIIEncoding
@@ -64,7 +63,7 @@ namespace System.Text
             return EncodingForwarder.GetByteCount(this, chars, index, count);
         }
 
-        public override int GetByteCount(String chars)
+        public override int GetByteCount(string chars)
         {
             return EncodingForwarder.GetByteCount(this, chars);
         }
@@ -75,12 +74,161 @@ namespace System.Text
             return EncodingForwarder.GetByteCount(this, chars, count);
         }
 
-        public override int GetBytes(String chars, int charIndex, int charCount,
-                                              byte[] bytes, int byteIndex)
+        public unsafe override byte[] GetBytes(string s)
         {
-            return EncodingForwarder.GetBytes(this, chars, charIndex, charCount, bytes, byteIndex);
+            if (s == null)
+                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.s, ExceptionResource.ArgumentNull_String);
+            Contract.EndContractBlock();
+
+            int charCount = s.Length;
+
+            byte[] bytes;
+            if (charCount > 0) 
+            {
+                fixed (char* input = s)
+                    bytes = GetBytesValidated(input, charCount);
+            } 
+            else 
+            {
+                bytes = Array.Empty<byte>();
+            }
+
+            return bytes;
         }
 
+        public unsafe override int GetBytes(string chars, int charIndex, int charCount, byte[] bytes, int byteIndex)
+        {
+            if ((chars == null) ||
+                (bytes == null) ||
+                (charIndex < 0) ||
+                (charCount < 0) ||
+                (chars.Length - charIndex < charCount) ||
+                (byteIndex < 0 || byteIndex > bytes.Length))
+            {
+                EncodingForwarder.ThrowValidationFailed(this, chars, charIndex, charCount, bytes);
+            }
+            Contract.EndContractBlock();
+
+            // Note that byteCount is the # of bytes to decode, not the size of the array
+            int byteCount = bytes.Length - byteIndex;
+            int bytesWritten;
+            if (charCount > 0) 
+            {
+                if (byteCount == 0)
+                {
+                    // Definitely not enough space, early bail
+                    EncodingForwarder.ThrowBytesOverflow(this);
+                }
+                fixed (char* pInput = chars)
+                fixed (byte* pOutput = &bytes[0]) 
+                {
+                    char* input = pInput + charIndex;
+                    byte* output = pOutput + byteIndex;
+                    int charactersConsumed;
+                    if (!EncodingForwarder.TryEncode(input, charCount, output, byteCount, out charactersConsumed, out bytesWritten)) 
+                    {
+                        // Not all ASCII, GetBytesFallback for remaining conversion
+                        bytesWritten += GetBytesFallback(input + charactersConsumed, charCount - charactersConsumed, output + bytesWritten, byteCount - bytesWritten, null);
+                    }
+                }
+            } 
+            else 
+            {
+                // Nothing to encode
+                bytesWritten = 0;
+            }
+
+            return bytesWritten;
+        }
+
+        public override byte[] GetBytes(char[] chars)
+        {
+            if (chars == null)
+                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.chars, ExceptionResource.ArgumentNull_Array);
+            Contract.EndContractBlock();
+            
+            return GetBytesValidated(chars, 0, chars.Length);
+        }
+
+        public override byte[] GetBytes(char[] chars, int index, int count)
+        {
+            if ((chars == null) ||
+                (index < 0) ||
+                (count < 0) ||
+                (chars.Length - index < count))
+            {
+                EncodingForwarder.ThrowValidationFailedException(chars, index, count);
+            }
+            Contract.EndContractBlock();
+            
+            return GetBytesValidated(chars, index, count);
+        }
+
+        private unsafe byte[] GetBytesValidated(char[] chars, int index, int count)
+        {
+            byte[] bytes;
+            if (count > 0)
+            {
+                fixed (char* input = chars)
+                {
+                    bytes = GetBytesValidated(input + index, count);
+                }
+            }
+            else
+            {
+                bytes = Array.Empty<byte>();
+            }
+
+            return bytes;
+        }
+
+        private unsafe byte[] GetBytesValidated(char* input, int charCount)
+        {
+            int remaining = 0;
+            // Assume string is all ASCII and size array for that
+            byte[] bytes = new byte[charCount];
+
+            int bytesWritten;
+            fixed (byte* output = &bytes[0]) 
+            {
+                int charactersConsumed;
+                if (!EncodingForwarder.TryEncode(input, charCount, output, charCount, out charactersConsumed, out bytesWritten)) 
+                {
+                    // Not all ASCII, get the byte count for the remaining encoded conversion
+                    remaining = GetByteCount(input + charactersConsumed, charCount - charactersConsumed, null);
+                }
+            }
+
+            if (remaining > 0) 
+            {
+                // Not all ASCII, fallback to slower path for remaining encoding
+                var encoded = ResizeGetRemainingBytes(input, charCount, ref bytes, bytesWritten, remaining);
+                Debug.Assert(encoded == remaining);
+            }
+
+            return bytes;
+        }
+
+        private unsafe int ResizeGetRemainingBytes(char* chars, int charCount, ref byte[] bytes, int alreadyEncoded, int remaining)
+        {
+            if (bytes.Length - remaining != alreadyEncoded)
+            {
+                // Resize the array to the correct size
+                byte[] oldArray = bytes;
+                bytes = new byte[alreadyEncoded + remaining];
+                // Copy already encoded bytes
+                Array.Copy(oldArray, 0, bytes, 0, alreadyEncoded);
+            }
+            int encoded;
+            fixed (byte* output = &bytes[0]) 
+            {
+                // Use GetBytesFallback for remaining conversion
+                encoded = GetBytesFallback(chars + alreadyEncoded, charCount - alreadyEncoded, output + alreadyEncoded, remaining, null);
+            }
+
+            return encoded;
+        }
+        
         // Encodes a range of characters in a character array into a range of bytes
         // in a byte array. An exception occurs if the byte array is not large
         // enough to hold the complete encoding of the characters. The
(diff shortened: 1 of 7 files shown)


benaadams:

benaadams:
@tarekgh @krwq  PTAL

jkotas:
I have noticed while working on https://github.com/dotnet/corefxlab/pull/1139 that there is low hanging performance problem in the current UTF8 encoder implementation: the loop control variables are not getting enregistered. This problem was introduced by adding the fallback handling that was not in the original implementation. I am wondering whether it would be better to fix this problem instead (e.g. by moving the rare fallback cruft to separate methods so that it does not get into way) instead of adding yet another fast path.

The loop control variable enregistration problem does not exist in https://github.com/dotnet/corefxlab/pull/1139, and I am seeing results on par with this fast path.

jkotas:
Or even better: Change the workhorse method to have same (for now just take the pointers; change it to Span once it is production quality) signature as what the workhorse method in corefxlab is; and wrap it with the glue that provides the fallbacks, etc. We will be able to easily share improvements between corefxlab and the existing encoding that way.

benaadams:
This PR started as a clean up for ASCII encoding; as the implementation doesn't perform very well so if you want to do a pure ASCII (rather than UTF8) encode you roll your own.

Additionally cleaning up the error checking and call stacks for the various methods as they dived though various virtual methods and did rechecking depending on caller.

I brought UTF8 into it as up until a non-ascii char is hit the ASCII & UTF8 are the same implementation; so the patterns are:

ASCII => non-virt `ASCII encoder` -> `nonascii` -> virt `ASCII fallback encoder`
UTF8 => non-virt `ASCII encoder` -> `nonascii` -> virt `UTF8 encoder`

I didn't want to change the non-ascii paths as part of this PR are they are hefty functions. 

Happy to change the approach?

benaadams:
@jkotas is the method you are referring to

`TryEncode(ReadOnlySpan<char> utf16, Span<byte> utf8, out int charactersConsumed, out int bytesWritten)`

jkotas:
Yes.

benaadams:
@jkotas made those changes; have marked the calls in Utf8Encoding to change with TODOs

src/mscorlib/src/System/Text/ASCIIEncoding.cs
```diff
@@ -78,12 +80,157 @@ public override unsafe int GetByteCount(char* chars, int count)
             return EncodingForwarder.GetByteCount(this, chars, count);
         }
 
-        public override int GetBytes(String chars, int charIndex, int charCount,
-                                              byte[] bytes, int byteIndex)
+        public unsafe override byte[] GetBytes(String s)
+        {
```
jkotas: The encoding classes are not sealed. These changes have potential to break people who inherit from them. @tarekgh Have you evaluated the breaking impact of this change?

src/mscorlib/src/System/Text/ASCIIEncoding.cs
```diff
@@ -78,12 +80,157 @@ public override unsafe int GetByteCount(char* chars, int count)
...
+            }
+
+            return bytes;
         }
 
+        private unsafe int ResizeGetRemainingBytes(char* chars, int charCount, ref byte[] bytes, int alreadyEncoded, int remaining)
+        {
+            // Resize the array to the correct size
+            byte[] oldArray = bytes;
+            bytes = new byte[alreadyEncoded + remaining];
```
jkotas: This changes the allocation profiles. It makes the change win some lose some, unclear whether it is net benefit for  programs out there. These allocating convenience methods are unlikely to be used on performance critical paths. It may be better to not touch them.

src/mscorlib/src/System/Text/EncodingForwarder.cs
```diff
@@ -179,19 +165,169 @@ public unsafe static int GetBytes(Encoding encoding, char[] chars, int charIndex
...
         }
 
+        internal unsafe static bool TryEncode(char* input, int charCount, byte* output, int byteCount, out int charactersConsumed, out int bytesWritten)
+        {
+            const int Shift16Shift24 = (1 << 16) | (1 << 24);
+            const int Shift8Identity = (1 <<  8) | (1);
+
+            int charsToEncode = Math.Min(charCount, byteCount);
+
+            // Encode as bytes upto the first non-ASCII byte and return count encoded
```
jkotas: I meant:
- Use the complete implementation of this method from https://github.com/dotnet/corefxlab/pull/1156. Its performance for the ASCII case is within 5% of this implementation (varies with alignment of various things), it is pretty much same as what `int GetBytes(char* chars, int charCount, byte* bytes, int byteCount, EncoderNLS baseEncoder)` in this file is today, and thus it handles everything - except for the EncodingFallback that kills the performance.
- Leave refactored handling of EncodingFallback `int GetBytes(char* chars, int charCount, byte* bytes, int byteCount, EncoderNLS baseEncoder)` and call TryEncode to do the actual work.

src/mscorlib/src/System/Text/ASCIIEncoding.cs
benaadams: Added `if` block to check for size changed for Ascii; which shouldn't change size for 1 or 2 byte characters; so would be uncommon case that a sting contains 4 byte characters and is being output as ascii?

Reverted utf8 to count the length first before trying to encode.

src/mscorlib/src/System/Text/ASCIIEncoding.cs
benaadams: In reference to implementing the String only override; or the current implementations having long chains of virtual method calls so inheriting from `ASCIIEncoding` or `UTF8Encoding` only needs to override `GetBytes(char* chars, int charCount, byte* bytes, int byteCount)` to override all the calls; except going via an encoder? (am guessing both)

src/mscorlib/src/System/Text/EncodingForwarder.cs
benaadams: When I do that https://github.com/benaadams/coreclr/commit/252998502caea98e507a57cde51c3190b71d873b 5 utf8 tests in corefx fail; looking into it.

Still not sure 100% if its what you mean?

src/mscorlib/src/System/Text/ASCIIEncoding.cs
benaadams: Ah, full framework all go via the internal virtual `GetBytes(char* chars, int charCount, byte* bytes, int byteCount, EncoderNLS encoder)` so is only adding the `GetBytes(String s)` overload

benaadams:
> STACK TRACE:
at Loader._classloader_generics_Instantiation_Positive_MultipleInterface06_MultipleInterface06_._classloader_generics_Instantiation_Positive_MultipleInterface06_MultipleInterface06_cmd() in D:\j\workspace\debug_windows---17180f1b\bin\tests\Windows_NT.x64.Debug\TestWrappers\Loader\Loader.XUnitWrapper.cs:line 6827

src/mscorlib/src/System/Text/ASCIIEncoding.cs
tarekgh: Looks the changes here is different than previous changes you had before. so we need to evaluate the breaking the changes here. we need to list the scenarios that can be breaking. sub-classing ASCIIEncoding is unlikely but sub-classing UTF8Encoding could be. I'll need some time to look at that.

---
In reply to: [98441711](https://github.com/dotnet/coreclr/pull/9187#discussion_r98441711) [](ancestors = 98441711)

src/mscorlib/src/System/Text/ASCIIEncoding.cs
```diff
@@ -78,12 +80,158 @@ public override unsafe int GetByteCount(char* chars, int count)
...
+        }
+
+        public unsafe override int GetBytes(String chars, int charIndex, int charCount, byte[] bytes, int byteIndex)
+        {
+            if ((chars == null) ||
+                (bytes == null) ||
+                (charIndex < 0) ||
+                (charCount < 0) ||
+                (chars.Length - charIndex < charCount) ||
+                (byteIndex < 0 || byteIndex > bytes.Length))
```
tarekgh: >byteIndex > bytes.Length [](start = 34, length = 24)

didn't we test before if byteIndex >= bytes.Length?

tarekgh: it looks we didn't do this check before so you are fine here.

---
In reply to: [98484325](https://github.com/dotnet/coreclr/pull/9187#discussion_r98484325) [](ancestors = 98484325)

src/mscorlib/src/System/Text/ASCIIEncoding.cs
```diff
@@ -78,12 +80,158 @@ public override unsafe int GetByteCount(char* chars, int count)
...
+            else 
+            {
+                // Nothing to encode
+                bytesWritten = 0;
+            }
+
+            return bytesWritten;
+        }
+
+        public override byte[] GetBytes(char[] chars) {
```
tarekgh: >{ [](start = 54, length = 1)

please move this to the next line

src/mscorlib/src/System/Text/ASCIIEncoding.cs
```diff
@@ -78,12 +80,158 @@ public override unsafe int GetByteCount(char* chars, int count)
...
-            return EncodingForwarder.GetBytes(this, chars, charIndex, charCount, bytes, byteIndex);
+            int remaining = 0;
+            // Assume string is all ASCII and size array for that
+            byte[] bytes = new byte[charCount];
+
+            int bytesWritten;
+            fixed (byte* output = &bytes[0]) 
+            {
+                int charactersConsumed;
+                if (!EncodingForwarder.TryEncode(input, charCount, output, charCount, out charactersConsumed, out bytesWritten))
```
tarekgh: >charactersConsumed, out bytesWritten [](start = 90, length = 36)

do you think we'll have a case that charactersConsumed will be different than byteswritten?

tarekgh:
@jkotas @benaadams I have evaluated the possibility of the breaking changes for providing the overrides methods in ASCIIEncoding and UTF8Encoding classes. I believe we are ok here because if some one sub-classed these encoded classes and was overriding one of these methods we still be ok. the only thing may concern a little if someone was calling base.GetBytes but even in this case we end up calling the version of GetBytes which takes unsafe char* and the behavior should be same at the end. so I don't have any concern providing such overriding methods. 

@jkotas do you think I missed anything here?

src/mscorlib/src/System/Text/EncodingForwarder.cs
jkotas: > 5 utf8 tests in corefx fail; looking into it.

Let me know what you find out. I can take a look as well. It is possible that the corefxlab implementation that I sent the PR for yesterday has bugs because of there is not enough test coverage in corefx for it yet.

> Not sure 100% its the implementation you are after though?

Yes, it is working towards what I have in mind. In addition to that, also delete the redundant implementation of the encoding work horse algorithm in the internal `GetBytes(char* chars, int charCount, ...` method.

The idea is that all GetBytes implementations should follow this pattern:

```
GetBytes(...)
{
    validate arguments

    if there is partial character pending, call method to deal with partial characters/fallbacks/error - this method can be slow and simple

    while (!TryEncode(...))
    {
        call method to deal with partial characters/fallbacks/errors - this method can be slow and simple
    }
}
```

It would allow us to have the exact same code for the fast core encoding algorithm with corefxlab that both I and @KrzysztofCwalina would love to see (https://github.com/dotnet/corefxlab/pull/1156#issuecomment-276141253).

It may make sense to split the change for ASCII and UTF8: Perfect the desired pattern on one first, and then apply it on the other.

jkotas:
> @jkotas do you think I missed anything here?

Make sense. Thanks for checking.

src/mscorlib/src/System/Text/EncodingForwarder.cs
benaadams: > It may make sense to split the change for ASCII and UTF8

I think this is a good idea, especially now they will have different workhorse functions. I will simplify and make the ASCII changes; then follow up with the Utf8 after tracking down the issues.

src/mscorlib/src/System/Text/ASCIIEncoding.cs
benaadams: Not for ASCII - will fix in standalone ASCII version

src/mscorlib/src/System/Text/ASCIIEncoding.cs
benaadams: runs if `charCount == 0 && byteCount == 0` and throws a different error if `charCount > 0 && byteCount == 0` I could probably roll it into the same Throw call...

tarekgh: we need to keep the old behavior so we'll be consistent with the full framework. we just need to get the exact same exception with the exact same condition.

tarekgh:
@benaadams please resolve the conflict with your new changes. thanks.

benaadams:
@tarekgh @jkotas question; for compat do I have to throw the exceptions in exactly the same order?

e.g. if encoder and decoder throw exactly the same set of exceptions; but one throws
`bytes == null` then `chars == null` and the other `chars == null` then `bytes == null` if they both threw
`bytes == null` then `chars == null` would that be a compat break? (i.e. they could share the same throw block)

tarekgh:
It is a slight difference which will break only who check for the parameter names. I think we can tolerate such case if it is really necessary. But i think it is better to keep the old behavior if there is no any complication with it.

jkotas:
> would that be a compat break?

I do not see a problem with it.

tarekgh:
@benaadams could you please resolve the conflict and get it ready to merge? thanks.

tarekgh:
@benaadams could you please respond? Thanks.

tarekgh:
@benaadams please respond. this PR has been for 9 days without any activities. thank you.

benaadams:
Sorry, yes will resolve

tarekgh:
@benaadams please resolve the conflict. thanks.

tarekgh:
@benaadams please resolve the conflicts. This PR has been opened for 19 days now.

benaadams:
Working on this now

benaadams:
Rebased; however still need to test with corefx

Will raise tracking bug for unifying with corefxlabs

tarekgh:
Excellent. please let's know when you finish testing on corefx. also please link the new issue unifying with corefxlab to this PR. thanks a lot.

benaadams:
Taking a little while; having trouble overlaying coreclr to corefx

stephentoub:
> having trouble overlaying CoreCLR to CoreFx

What issues are you hitting?
https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/developer-guide.md#testing-with-private-coreclr-bits

benaadams:
> Unhandled Exception: System.NotSupportedException: FileStream was asked to open a device that was not a file. For support for devices like 'com1:' or 'lpt1:', call Crea
  teFile, then use the FileStream constructors that take an OS handle as an IntPtr.
     at Interop.Kernel32.SafeCreateFile(String lpFileName, Int32 dwDesiredAccess, FileShare dwShareMode, SECURITY_ATTRIBUTES& securityAttrs, FileMode dwCreationDispositio
  n, Int32 dwFlagsAndAttributes, IntPtr hTemplateFile) in D:\Work\GitHub\coreclr\src\mscorlib\corefx\Interop\Windows\kernel32\Interop.SafeCreateFile.cs:line 40
     at System.IO.FileStream.OpenHandle(FileMode mode, FileShare share, FileOptions options) in D:\Work\GitHub\coreclr\src\mscorlib\corefx\System\IO\FileStream.Win32.cs:l
  ine 83
     at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options) in D:\Work\GitHub\coreclr\src\ms
  corlib\corefx\System\IO\FileStream.cs:line 192
     at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, Boolean useAsync) in D:\Work\GitHub\coreclr\src\mscor
  lib\corefx\System\IO\FileStream.cs:line 139
     at Xunit.ConsoleClient.TransformFactory.Handler_DirectWrite(XElement xml, String outputFileName)
     at System.Collections.Generic.List`1.ForEach(Action`1 action) in D:\Work\GitHub\coreclr\src\mscorlib\src\System\Collections\Generic\List.cs:line 553
     at Xunit.ConsoleClient.Program.RunProject(String defaultDirectory, XunitProject project, Boolean teamcity, Boolean appVeyor, Boolean showProgress, Nullable`1 paralle
  lizeAssemblies, Nullable`1 parallelizeTestCollections, Nullable`1 maxThreadCount)
     at Xunit.ConsoleClient.Program.Main(String[] args)
  Finished running tests.  End time=18:00:39.83, Exit code = -532462766

Normally takes me a few tries; then I get something stable, till there is a major shake up again

benaadams:
Actually all my corefx tests are failing for all projects without trying to apply coreclr
```
clean
build
build-tests
```

```
  Using D:\Work\GitHub\corefx\bin\testhost\netcoreapp-Windows_NT-Debug-x64\ as the test runtime folder.
  Executing in D:\Work\GitHub\corefx\bin\AnyOS.AnyCPU.Debug\System.Linq.Parallel.Tests\netcoreapp\
  Running tests... Start time: 13:47:57.65
  Command(s):
  call D:\Work\GitHub\corefx\bin\testhost\netcoreapp-Windows_NT-Debug-x64\\dotnet.exe xunit.console.netcore.exe System.Linq.Parallel.Tests.dll  -xml testResults.xml -notrait Benchmark=true -notrait cat
  egory=nonnetcoreapp1.1tests  -notrait category=OuterLoop -notrait category=failing -notrait category=nonwindowstests
  Using D:\Work\GitHub\corefx\bin\testhost\netcoreapp-Windows_NT-Debug-x64\ as the test runtime folder.
  Executing in D:\Work\GitHub\corefx\bin\AnyOS.AnyCPU.Debug\System.Net.Requests.Tests\netstandard\
  Running tests... Start time: 13:47:57.68
  Command(s):
  call D:\Work\GitHub\corefx\bin\testhost\netcoreapp-Windows_NT-Debug-x64\\dotnet.exe xunit.console.netcore.exe System.Net.Requests.Tests.dll  -xml testResults.xml -notrait Benchmark=true -notrait cate
  gory=nonnetcoreapp1.1tests  -notrait category=OuterLoop -notrait category=failing -notrait category=nonwindowstests
  xUnit.net console test runner (64-bit .NET Core)

  xUnit.net console test runner (64-bit .NET Core)

  Discovering: System.Linq.Parallel.Tests
  Discovering: System.Net.Requests.Tests
  Discovered:  System.Net.Requests.Tests
  Discovered:  System.Linq.Parallel.Tests
  Info:        System.Net.Requests.Tests has no tests to run
  Info:        System.Linq.Parallel.Tests has no tests to run


  Unhandled Exception: System.NotSupportedException: FileStream was asked to open a device that was not a file. For support for devices like 'com1:' or 'lpt1:', call CreateFile, then use the FileStream
   constructors that take an OS handle as an IntPtr.
  Unhandled Exception: System.NotSupportedException: FileStream was asked to open a device that was not a file. For support for devices like 'com1:' or 'lpt1:', call CreateFile, then use the FileStream
   constructors that take an OS handle as an IntPtr.
     at Interop.Kernel32.SafeCreateFile(String lpFileName, Int32 dwDesiredAccess, FileShare dwShareMode, SECURITY_ATTRIBUTES& securityAttrs, FileMode dwCreationDisposition, Int32 dwFlagsAndAttributes,
  IntPtr hTemplateFile)
     at Interop.Kernel32.SafeCreateFile(String lpFileName, Int32 dwDesiredAccess, FileShare dwShareMode, SECURITY_ATTRIBUTES& securityAttrs, FileMode dwCreationDisposition, Int32 dwFlagsAndAttributes,
  IntPtr hTemplateFile)
     at System.IO.FileStream.OpenHandle(FileMode mode, FileShare share, FileOptions options)
     at System.IO.FileStream.OpenHandle(FileMode mode, FileShare share, FileOptions options)
     at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
     at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
  Finished running tests.  End time=13:47:58.40, Exit code = -532462766
     at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, Boolean useAsync)
     at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, Boolean useAsync)
     at Xunit.ConsoleClient.TransformFactory.Handler_DirectWrite(XElement xml, String outputFileName)
     at Xunit.ConsoleClient.TransformFactory.Handler_DirectWrite(XElement xml, String outputFileName)
     at System.Collections.Generic.List`1.ForEach(Action`1 action)
     at System.Collections.Generic.List`1.ForEach(Action`1 action)
     at Xunit.ConsoleClient.Program.RunProject(String defaultDirectory, XunitProject project, Boolean teamcity, Boolean appVeyor, Boolean showProgress, Nullable`1 parallelizeAssemblies, Nullable`1 para
  llelizeTestCollections, Nullable`1 maxThreadCount)
     at Xunit.ConsoleClient.Program.RunProject(String defaultDirectory, XunitProject project, Boolean teamcity, Boolean appVeyor, Boolean showProgress, Nullable`1 parallelizeAssemblies, Nullable`1 para
  llelizeTestCollections, Nullable`1 maxThreadCount)
     at Xunit.ConsoleClient.Program.Main(String[] args)
D:\Work\GitHub\corefx\Tools\tests.targets(230,5): warning MSB3073: The command "D:\Work\GitHub\corefx\bin/AnyOS.AnyCPU.Debug/System.Linq.Expressions.Tests/netcoreapp//RunTests.cmd D:\Work\GitHub\corefx
\bin/testhost/netcoreapp-Windows_NT-Debug-x64/" exited with code -532462766. [D:\Work\GitHub\corefx\src\System.Linq.Expressions\tests\System.Linq.Expressions.Tests.csproj]
D:\Work\GitHub\corefx\Tools\tests.targets(238,5): error : One or more tests failed while running tests from 'System.Linq.Expressions.Tests' please check D:\Work\GitHub\corefx\bin/AnyOS.AnyCPU.Debug/Sys
tem.Linq.Expressions.Tests/netcoreapp/testResults.xml for details! [D:\Work\GitHub\corefx\src\System.Linq.Expressions\tests\System.Linq.Expressions.Tests.csproj]
```

benaadams:
Going a bit better with ye'olde delete repo and re clone

benaadams:
@tarekgh all tests pass in both debug and release

benaadams:
Added issue https://github.com/dotnet/coreclr/issues/9668

tarekgh:
@benaadams thanks a lot.

PR #9115, opened 2017-01-25
Reduce number of branches and IL size of string.Trim/TrimEnd/TrimStart methods

justinvp:
At the point where we use `fixed`, we already know the `char[]` is not null or empty, so we can reduce the number of branches and overall size of the method body by using the address of the first element. Based on the discussion here: https://github.com/dotnet/corefx/pull/15432#discussion_r97684916

cc: @stephentoub, @jkotas

PR #9188, opened 2017-01-28
Avoid unnecessary branches when fixing non-null/empty arrays

justinvp:
Follow-up of https://github.com/dotnet/coreclr/pull/9115, applied throughout to places where `fixed` is used with arrays known not to be null or empty. Based on the discussion here: https://github.com/dotnet/corefx/pull/15432#discussion_r97684916

### DIFF
--- a/src/mscorlib/src/System/Array.cs
+++ b/src/mscorlib/src/System/Array.cs
@@ -154,7 +154,7 @@ namespace System {
                 if (lengths[i] < 0)
                     ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.lengths, i, ExceptionResource.ArgumentOutOfRange_NeedNonNegNum);
 
-            fixed (int* pLengths = lengths)
+            fixed (int* pLengths = &lengths[0])
                 return InternalCreate((void*)t.TypeHandle.Value,lengths.Length,pLengths,null);
         }
 
@@ -211,8 +211,8 @@ namespace System {
                 if (lengths[i] < 0)
                     ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.lengths, i, ExceptionResource.ArgumentOutOfRange_NeedNonNegNum);
 
-            fixed (int* pLengths = lengths)
-                fixed(int* pLowerBounds = lowerBounds)
+            fixed (int* pLengths = &lengths[0])
+                fixed(int* pLowerBounds = &lowerBounds[0])
                     return InternalCreate((void*)t.TypeHandle.Value,lengths.Length,pLengths,pLowerBounds);
         }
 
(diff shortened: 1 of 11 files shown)


src/mscorlib/src/System/StubHelpers.cs
```diff
@@ -27,7 +27,8 @@ internal static class AnsiCharMarshaler
         unsafe static internal byte[] DoAnsiConversion(string str, bool fBestFit, bool fThrowOnUnmappableChar, out int cbLength)
         {
             byte[] buffer = new byte[(str.Length + 1) * Marshal.SystemMaxDBCSCharSize];
-            fixed (byte *bufferPtr = buffer)
+            BCLDebug.Assert(buffer.Length != 0);
```
benaadams: What's one of these?

justinvp: Are you asking why the assert or why `BCLDebug.Assert`? I guess the assert isn't strictly necessary here -- I decided to add it in case `Marshal.SystemMaxDBCSCharSize` ever returned 0 (the way it is currently implemented it will always be greater than 0). I used `BCLDebug.Assert` because it was used elsewhere in this file.

benaadams: `BCLDebug` hadn't seen it before

src/mscorlib/src/System/Array.cs
```diff
@@ -328,7 +328,7 @@ public unsafe Object GetValue(params int[] indices)
             Contract.EndContractBlock();
 
             TypedReference elemref = new TypedReference();
-            fixed(int* pIndices = indices)
+            fixed(int* pIndices = &indices[0])
```
stephentoub: Is it impossible for Rank to return 0?  I assume so.

justinvp: I had the same question as I was considering this change... I dug through the code and it doesn't look like a Rank of 0 is possible:

 1. https://github.com/dotnet/coreclr/blob/a9c90ffd84e987a1793a1ed5c5ad7d89b27d493a/src/vm/ecalllist.h#L1523
 2. https://github.com/dotnet/coreclr/blob/c10c1ff8e3237689212606c9aa5153beec8a1778/src/classlibnative/bcltype/arraynative.cpp#L20-L31
 3. https://github.com/dotnet/coreclr/blob/32f0f9721afb584b4a14d69135bea7ddc129f755/src/vm/object.inl#L163-L167
 4. https://github.com/dotnet/coreclr/blob/32f0f9721afb584b4a14d69135bea7ddc129f755/src/vm/methodtable.inl#L604-L613
 5. https://github.com/dotnet/coreclr/blob/dea1b9e5c0393daa420831eb74177b1fd6546181/src/vm/class.h#L2474-L2485
 6. https://github.com/dotnet/coreclr/blob/32f0f9721afb584b4a14d69135bea7ddc129f755/src/vm/clsload.cpp#L3677-L3679
 7. https://github.com/dotnet/coreclr/blob/32f0f9721afb584b4a14d69135bea7ddc129f755/src/vm/array.cpp#L250

PR #9189, opened 2017-01-28
Enable generic attributes

AviAvni:
https://github.com/dotnet/roslyn/issues/953

Succesfully compile and run https://gist.github.com/AviAvni/00831abbddf4ceca198236de362ed449

image

image


src/vm/customattribute.cpp
```diff
@@ -695,10 +694,6 @@ FCIMPL2(Object*, RuntimeTypeHandle::CreateCaInstance, ReflectClassBaseObject* pC
         PRECONDITION(
             (!pCtor && gc.refCaType->GetType().IsValueType() && !gc.refCaType->GetType().GetMethodTable()->HasDefaultConstructor()) || 
             (pCtor == gc.refCaType->GetType().GetMethodTable()->GetDefaultConstructor()));
-
-        // If we relax this, we need to insure custom attributes construct properly for Nullable<T>
```
jkotas: How does this work for Nullable types that this comment is talking about?

AviAvni: If you mean that the generic argument is nullable so I tried with int? and it works great too

AviAvni:
@jkotas Do I need to do more things to get this PR merged?

src/vm/customattribute.cpp
```diff
@@ -671,7 +671,6 @@ FCIMPL2(Object*, RuntimeTypeHandle::CreateCaInstance, ReflectClassBaseObject* pC
     CONTRACTL {
         FCALL_CHECK;
         PRECONDITION(CheckPointer(pCaTypeUNSAFE));
-        PRECONDITION(!pCaTypeUNSAFE->GetType().IsGenericVariable());
```
AtsushiKan: I believe this precondition is still valid. This is guarding against something like this:

     class G<T>
     {
          [T]   // <-- still disallowed, correct?
          class Inner {}
     }

jkotas:
@AtsushiKan Could you please take a look at this?

There will need to be tests added or updated in corefx.

src/vm/customattribute.cpp
AviAvni: @AtsushiKan You right sorry I revert this delete and validated it's still work

AtsushiKan:
What I see here LGTM but the framework has a bunch of overlapping custom-attribute fetching apis  and the code isn't as factored for sharing as it could be.

Attribute.GetCustomAttribute(*), IsDefined, etc.
CustomAttributeExtensions.GetCustomAttributes() (mostly wraps Attribute, but should check.)
ICustomAttributeProvider.GetCustomAttributes() (implemented on MemberInfo, Assembly, Module, etc. - this is largely its own codepath, I recall.)
CustomAttributeData.GetCustomAttributes(),
MemberInfo.CustomAttributes properties (and on Assemblies, Modules)

Can you test some of these as well to make sure all our code paths have been updated?

AviAvni:
@AtsushiKan what is the best place to write this tests? 
I'll check them
@jkotas do you mean the tests of System.Reflection.* on CoreFX repository I'll check them too

AtsushiKan:
System.Runtime\tests is where you'd add the tests, but this assumes that this new language feature is already enabled in the corefx toolset. If not, you should leave an issue open for writing tests once the feature is available. 

For now, I just meant test it privately as you've been doing...

AviAvni:
@AtsushiKan I also made PR for Roslyn https://github.com/dotnet/roslyn/pull/16814 so when this will merge hope also roslyn merge the PR

jkotas:
Should I be able to say `myType.GetCustomAttribute(MyCustomAttribute<>)` and expect it to work for any T?

AviAvni:
@jkotas I'll check this too

AtsushiKan:
>Should I be able to say myType.GetCustomAttribute(MyCustomAttribute<>) and expect it to work for any T?

This sounds like appropriate material for an api proposal. It's a bit iffy - it conflates the concept of assignability with instantiation.

Note that this would be a breaking change if we don't do it in NS2.0 as in some instances (the GetCustomAttribute() apis off MemberInfo (the ones not "added" by CustomAttributeExtensions) don't do any validation on the passed in attribute type, so this kind of call today would return an empty array (as in "Yessir, none of the attributes on this member are assignable to MyCustomAttribute<>")

MichalStrehovsky:
How will this work for serialized custom attribute data? Suppose we have:

```csharp
class FooAttribute<T> : Attribute
{
    public T Field;
}
```

Now, I would expect this to work:

```csharp
[Foo<int>(Field = 123)]
class X { }
```

But this shouldn't:

```csharp
[Foo<int?>(Field = 123)]
class Y { }
```

Will the class library fail appropriately? Will Roslyn be checking for this too?

AviAvni:
@MichalStrehovsky roslyn already check this

MichalStrehovsky:
> @MichalStrehovsky roslyn already check this

But the runtime needs to check for this too. There could be a buggy compiler that emits such metadata blob. We need to make sure it fails as expected (and make sure we have test coverage for that).

It's possible the runtime already does handle it, but since this PR is removing some validation that would previously cover it, we need to ensure the new failure modes the removed validation opens us up to are covered as well.

AviAvni:
@MichalStrehovsky I'm working on tests I'll add this scenario thanks

AviAvni:
@AtsushiKan Please look at my tests https://gist.github.com/AviAvni/3ee62f4c93d33485db880cf01011e0b6


@MichalStrehovsky I tried to make tests that you mentioned by taking the IL from the tests exe with ildasm
https://gist.github.com/AviAvni/388b8a3023181bd78f9d56e85c2a4b31
and change it and compile back with ilasm but I need help what to change I tried to make line 138 value like 187 value not sure if this correct

AtsushiKan:
Be careful about distinguishing beween the instance member GetCustomAttributes() and the extension method CustomAttributes() - they're called using the same C# syntax but for reasons that escape me, aren't the same at all internally.

It's probably safest to write the tests so that the instance member version is always invoked through the ICustomAttributeProvider interface rather directly from the member, while the extension method is invoke without using the extension method syntax, i.e. rather than writing

  memberInfo.GetCustomAttribute(typeof(D), inherit: true),

write

  ((ICustomAttributeProvider)memberInfo).GetCustomAttributes(typeof(D), inherit: true)

and 

   CustomAttributeExtensions.GetCustomAttribute(memberInfo, typeof(D), inherit: true).

That way, it's always clear which one you mean.

AviAvni:
@AtsushiKan thanks I updated the gist
now I'm investigating why https://gist.github.com/AviAvni/3ee62f4c93d33485db880cf01011e0b6#file-genericattributetests-cs-L171 have different behavior

AviAvni:
@AtsushiKan now the difference is in line https://gist.github.com/AviAvni/3ee62f4c93d33485db880cf01011e0b6#file-genericattributetests-cs-L187
and still don't understand why it's happend

AtsushiKan:
CustomAttributeExtensions.GetCustomAttributes() should never return null. If there are no matching attributes, it should return a 0-length enumeration.

AviAvni:
@AtsushiKan I deebugged the cod and because of this line https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Attribute.cs#L477
when "element.GetCustomAttributes(type, inherit)" return "object[0]" then the "as Attribute[]" make it return null

AtsushiKan:
That's... unfortunate. The train as probably sailed on that one, though...

AviAvni:
@jkotas I need help with modifying the IL to reproduce @MichalStrehovsky test
What do you think about the tests?

jkotas:
The format of the custom attribute blob is described in the ECMA-335 spec (https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/dotnet-standards.md).

Also, the attribute types in IL are not guaranteed to inherit from Attribute. You may also want to add a few tests for that. For example:

`.custom instance void class [System.Private.CoreLib]System.Collections.Generic.List`1<int32>::.ctor() = ( 01 00 00 00 )`

AviAvni:
@jkotas I checked the list attribute and it works
Also I tried to read in the specification and in the expert IL book
But still didn't get to encode the @MichaelSimons scenario
Maybe someone can help me with that?

MichalStrehovsky:
> But still didn't get to encode the @MichaelSimons scenario

I assume you mean my scenario. All I'm asking for is to:

Make a generic attribute that has a T-typed parameter in the constructor. Apply the instantiated custom attribute on something in two ways:

* T is `int32`. I assume you already know how to create the custom attribute data blob bytes for that - it's the same as for a non-generic case.
* T is `int32?`. Use the same custom attribute data blob you used for the above case (there's no encoding for nullable types in the format. This should throw a suitable exception at runtime.

AviAvni:
@MichalStrehovsky thanks I tried this
.custom instance void class ConsoleApplication21.MultiAttribute`1<valuetype [System.Private.CoreLib]System.Nullable`1<bool>>::.ctor() = ( 01 00 01 00 54 02 05 56 61 6C 75 65 01 ) 
and there is no exception but also when I'm debugging it's look like that the coreclr ignore this blob

AviAvni:
@MichalStrehovsky In this line https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Reflection/CustomAttribute.cs#L1662
in the debugger I get that the cNamedArgs is 0 and blobLen is some how 4 so with ildasm everything looks ok but maybe ilasm did something that I don't know

MichalStrehovsky:
> .custom instance void class ConsoleApplication21.MultiAttribute1<valuetype [System.Private.CoreLib]System.Nullable1>::.ctor() = ( 01 00 01 00 54 02 05 56 61 6C 75 65 01 )
and there is no exception but also when I'm debugging it's look like that the coreclr ignore this blob

This blob is bad in a different way. You're referring to the parameterless constructor on the left hand side and then pass some values on the right hand side. CoreCLR ignores them because they're not needed.

I'm asking to test a constructor that does take parameters. Also, the `Nullable` should be instantiated. It's not instantiated in your snippet (there's no generic argument passed to the `Nullable`.

AviAvni:
@MichalStrehovsky sorry I fixed the comment so now you can see that I have type argument
and now I understand you wanted to check this
``.custom instance void class ConsoleApplication21.MultiAttribute`1<valuetype [System.Private.CoreLib]System.Nullable`1<bool>>::.ctor(!0) = ( 01 00 01 00 00 )``
this is also have problem the coreclr return that ctorHasParameters is false

danmosemsft:
@AviAvni where do we stand on this PR?

AviAvni:
@danmosemsft I did some of tests https://gist.github.com/AviAvni/3ee62f4c93d33485db880cf01011e0b6
and 2 tests I don't sure they passs
``.custom instance void class ConsoleApplication21.MultiAttribute`1<valuetype [System.Private.CoreLib]System.Nullable`1<bool>>::.ctor(!0) = ( 01 00 01 00 00 )``
``.custom instance void class ConsoleApplication21.MultiAttribute`1<valuetype [System.Private.CoreLib]System.Nullable`1<bool>>::.ctor() = ( 01 00 01 00 54 02 05 56 61 6C 75 65 01 )``
Do you want me to zip my binaries and share them?

danmosemsft:
@MichalStrehovsky can you help move this along?

MichalStrehovsky:
> @MichalStrehovsky can you help move this along?

@danmosemsft From my perspective (as someone who doesn't own any part of this codebase), this PR needs IL tests that cover the newly exposed code path. They need to be part of the checkin, and they need to cover the case `// If we relax this, we need to insure custom attributes construct properly for Nullable<T>` that the PR is deleting. Code changes that light up features by deleting validation pretty much always introduce new bugs into the product, so this needs to be tested.

Also, the change that incorrectly replaces spaces with tabs needs to go away.

AviAvni:
@MichalStrehovsky @danmosemsft I fixed the tab vs space issue sorry about that
Where I can put the tests?

MichalStrehovsky:
> Where I can put the tests?

`\tests\src\reflection` seems like the most obvious spot.

AviAvni:
@MichalStrehovsky thanks I added the tests there is a C# file that I compiled with my custom Roslyn compiler and then I generated the IL file with ildasm.
I hope te test will run in the build didn't understad how to check this

MichalStrehovsky:
> I hope the test will run in the build didn't understad how to check this

@danmosemsft Could you get someone who works on the CoreCLR codebase to drive this from Microsoft side? I commented on this pull request because metadata is in my general area of interest, but I don't have experience in CoreCLR codebase to help with technicalities (or sign off on code, for that matter).

@AviAvni I would also want to see test coverage for more interesting parts of the format (see [ECMA-335](http://www.ecma-international.org/publications/standards/Ecma-335.htm) II.23.3 Custom attributes) such as a constructor parameter/field/property:

* of type `T[]` when `T` is `string`, `object`, `int`, or enum.
* of type `T` when T is `string`, `System.Type`, or enum

e.g.
```csharp
class MyAttribute<T> : Attribute
{
    public MyAttribute(T[] foo) { }
}
```

Is there test coverage for T instantiated as `int?` and a value is passed as `int` we discussed previously?(This should be an error condition).

AviAvni:
@MichalStrehovsky when T is string and the string value passed to the ctor parameter then I have exception System.Reflection.CustomAttributeFormatException: Binary format of the specified custom attribute was invalid.
but when using the property assignment it's ok
I'll work on this tests later and try to investigate why I get the error

danmosemsft:
> @danmosemsft Could you get someone who works on the CoreCLR codebase to drive this from Microsoft side?

@gkhanna79 who on your team can shepherd this one?

karelz:
@AtsushiKan @DnlHarvey @joshfree any update on this? It is 7 months old PR :(

AtsushiKan:
The corresponding csharp lang proposal (https://github.com/dotnet/csharplang/issues/124) is still in the "Discussed" phase, not in the "Decided" phase. Is there a reason to implement the feature on the runtime side before it's been accepted on the language side (especially as without language support, we'll have to check in all the ILASM hackery in order to test the feature?)

karelz:
Won't we need the Runtime changes in **before** C# full implementation? ... I think we will need the ilasm hackary either way. So what is preventing us to finish the work now?

If the overall design in Runtime is unclear / undecided, that would be another story.

AtsushiKan:
>Won't we need the Runtime changes in before C# full implementation?

No - there's no implementation dependency. 

It's likely that the language feature will be of limited real-world use without the library changes but the reverse is true too. The thing is that the language change hasn't even been approved yet. If it doesn't get approved, there's not much point in merging this.

We might still need the ILASM as an intermediate step, but I'd like to avoid that if we can. The proper place for the tests is in corefx as Project N will need test bits too and they can't leverage coreclr tests.

tests/src/reflection/GenericAttribute/GenericAttribute.cs
```diff
@@ -0,0 +1,214 @@
...
+
+            Assert(CustomAttributeExtensions.GetCustomAttribute<SingleAttribute<int>>(deriveTypeInfo, true) != null);
+            Assert(CustomAttributeExtensions.GetCustomAttribute<SingleAttribute<bool>>(deriveTypeInfo, true) != null);
+            Assert(CustomAttributeExtensions.IsDefined(deriveTypeInfo, typeof(SingleAttribute<int>), true));   
+            Assert(((ICustomAttributeProvider)deriveTypeInfo).IsDefined(typeof(SingleAttribute<int>), true));           
+            Assert(CustomAttributeExtensions.IsDefined(deriveTypeInfo, typeof(SingleAttribute<bool>), true));
+            Assert(((ICustomAttributeProvider)deriveTypeInfo).IsDefined(typeof(SingleAttribute<bool>), true));           
+
+            var a1 = CustomAttributeExtensions.GetCustomAttributes(programTypeInfo, true).GetEnumerator();
+            AssertNext<Attribute>(a1, a => a is SingleAttribute<int>);
```
AtsushiKan: The custom attribute apis make no promise as to the order in which they return attributes. The tests need to be doing an order-agnostic compare.

karelz:
Discussed offline with @AtsushiKan - I believe there will be some chicked-and-egg problem anyway - C# compiler likely won't be RTM until some runtime supports the feature. Which means we will need IL tests anyway IMO (I don't think we use in build tools unreleased C# compilers). From that perspective I would suggest to review current change, make sure we have good (IL) tests and get that in.

C# lang. feature finalization will benefit from that. And we will rewrite the tests in C# once C# has official version published and we adopt it in build tools.

If you think I missed something, I'll be happy to change my mind :)
Overall I don't like long-opened PRs, I don't think they are helpful to the ecosystem.

AtsushiKan:
Looks like the whole test is in IL. That'll make it unnecessarily hard to understand and extend the tests. This should be split into an ILASM dll (that contains only the attribute definitions and metadata decorated with the generic attributes) and a C# exe that has the actual tests.

AviAvni:
@karelz @AtsushiKan thanks for your comments
I'll be happy to continue to work on this
As I understand this are the tasks I need to work on now
1. Split the test to il file with the attributes and cs file for the tests
2. Change the tests to be order agnostic compare
3. Investigate the exception I got with string please look at my last comment I need help with this
also I need help with how to make this tests run on build

AviAvni:
@karelz @AtsushiKan I separated the metadata and tests also made the tests order agnostic
now I need help how to make this tests run in the build in my machine it's run ok with my patched roslyn and patched coreclr
and then I continue add more tests

AtsushiKan:
>how to make this tests run in the build in my machine 

I think you can use this CoreCLR test as a template:

https://github.com/dotnet/coreclr/blob/aabfe1617aef41497aa17fe11e4b4f29bd78680f/tests/src/Loader/classloader/TypeForwarding/UnitTest/UnitTest.il

The instructions for building and running tests are here:

https://github.com/dotnet/coreclr/blob/master/Documentation/building/windows-test-instructions.md


Note that while the "correct" place to put the tests is in corefx, the only precedent I can find for ILASM code there involves checking in the actual test assemblies as GIT-tracked binaries, which is a practice that's severely frowned upon. So I think we'll have to settle for CoreCLR for now, unless someone on this thread has a more practical suggestion.

karelz:
@AviAvni do you plan to finish the PR?

AviAvni:
@AtsushiKan I foud why string and some more types makes the bad format exception
in https://github.com/dotnet/coreclr/blob/7590378d8a00d7c29ade23fada2ce79f4495b889/src/vm/methodtable.cpp#L6425
it's not now to return ELEMENT_TYPE_STRING
also added the callstack when this method is called can I get help with that?
image

AtsushiKan:
Can you add a little context to that? I assume you're trying to retrieve a (generic instantiation) custom attribute from something - what does the custom attribute look like on the something? And what did `GetSignatureCorElementType() ` return?

AviAvni:
@AtsushiKan the custom attribute is
``` [MultiAttribute<string>("Ctor")]```
And the GetSignatureCoreElementType method return the value from the default case

**Comment by AtsushiKan** [comment]
I don't understand what "the value from the default case" means. It should be returning some value from this enumeration:

https://github.com/dotnet/coreclr/blob/1f9aeeb7a3685bc7fd1098fc50d91ac81bae4873/src/inc/corhdr.h#L860

**Comment by AviAvni** [comment]
Sorry the value from the line https://github.com/dotnet/coreclr/blob/7590378d8a00d7c29ade23fada2ce79f4495b889/src/vm/methodtable.cpp#L6462
ELEMENT_TYPE_CLASS

**Comment by AtsushiKan** [comment]
Hmm... could it be that Roslyn is generating a non-canonical signature for the custom attribute constructor? (ET_CLASS + <token of System.String>) instead of ET_STRING?

**Comment by AviAvni** [comment]
How I can check this?

**Comment by AtsushiKan** [comment]
On further look, I realized that the expected signature in the metadata is "!0", not "string".

So the signature blob you're actually parsing was probably constructed at runtime by CoreCLR itself. The simplest way to test the theory that the sig blob is ET_CLASS System.String is simply to change the code that's parsing the sig as if the theory was true and see if it works. It's not a "proper" sig but since it's generated internally (and remember, it's never generated a signature from a generic instantiation custom attribute before), it might be that that's the expected format here.

**Comment by AviAvni** [comment]
@AtsushiKan thanks as you can see
https://github.com/dotnet/coreclr/pull/9189/files#diff-771f078434a184014f4d3a6d5626df7fR23
The Roslyn is generating 0! But in case when the generic value is int the method
GetSignatureCoreElementType return ELEMENT_TYPE_VALUETYPE which is fine for int
In case of string and Type it not return the proper value
I understand that this kind of Sig wasn't checked till now
I'll try to see what happens if I'm hack this method to return the proper value

**Comment by jkotas** [comment]
`GetSignatureCoreElementType` on `System.String` is expected to return `ELEMENT_TYPE_CLASS`. `ELEMENT_TYPE_STRING` is a shortcut to make the metadata signature encoding shorter. `ELEMENT_TYPE_STRING` gets normalized to `ELEMENT_TYPE_CLASS` early in the signature parsing code. Most of CoreCLR does not deal with it.

**Comment by AviAvni** [comment]
@jkotas thanks so now the problem I see is in
https://github.com/dotnet/coreclr/blob/46ab1d132c9ad471d79afa20c188c2f9c85e5f20/src/vm/customattribute.cpp#L1449-L1462
here it goes to the bad block instead of string type if I changed it in the debugger it work
I found that the th.GetName return "System.__Canon"
I think maybe  its in this method
https://github.com/dotnet/coreclr/blob/46ab1d132c9ad471d79afa20c188c2f9c85e5f20/src/vm/customattribute.cpp#L801

**Comment by AviAvni** [comment]
@AtsushiKan @jkotas The problem with the TypeHandle that instead of having the type parameter
have the System.__Cannon is https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Reflection/CustomAttribute.cs#L1720
I tried to pass also the typeInstantiationContextHandles but that not help

**Comment by AtsushiKan** [comment]
That's the start of the problem. This callsite:

https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Reflection/CustomAttribute.cs#L1720

needs to be calling this overload of ResolveMethodHandleInternal() and passing it the typeInstantiationContext (that maps the "!0" to "String").

https://github.com/dotnet/coreclr/blob/1f9aeeb7a3685bc7fd1098fc50d91ac81bae4873/src/mscorlib/src/System/RuntimeHandles.cs#L1339

The typeInstantationContext can be obtained by calling `attributeType.GenericTypeArguments` and calling the `TypeHandle` property on each element of the `Type[]` array it returns.

There may be more plumbing to do downstack but this looks like the root of the problem.

**Comment by AviAvni** [comment]
@AtsushiKan ok no the problem is 
https://github.com/dotnet/coreclr/blob/master/src/vm/memberload.cpp#L1107
you can see in the watch that the pMT->debug_m_szClassName is the right type but
the pCurDeclMD->m_pszDebugClassName is the wrong class
image

**Comment by AviAvni** [comment]
Now I understand that this is ok and that this
https://github.com/dotnet/coreclr/blob/master/src/vm/memberload.cpp#L770
is where the generic type need to be resolved but this method don't use the pTypeContext
to resolve the type instansiation

**Comment by AtsushiKan** [comment]
`ResolveMethodHandleInternal` is also called by the `Module.ResolveMethod()` api (https://github.com/dotnet/coreclr/blob/121d9a31e924c321dfb75dd0ffe16f227cf5e290/src/mscorlib/src/System/Reflection/RuntimeModule.cs#L99) You can look to see what it's doing.

Depending on how much this shakes things up, it may be simpler just to call `Module.ResolveMethod()`. That's the api to do exactly what CustomAttributes.cs is trying to do.

**Comment by AviAvni** [comment]
I changed it to use this ResolveMethod it's look like it's good when I'm debugging the System.Private.CoreLib.dll
image
But the CoreCLR is still the same
image

**Comment by AviAvni** [comment]
@AtsushiKan I pushed the changes so you can debug it too

**Comment by AtsushiKan** [comment]
I've built your changes and taking a more in depth look.

**Comment by AtsushiKan** [comment]
The exception is being thrown when parsing this custom attribute:

```
[MultiAttribute<Type>(typeof(Class))]
public class Class {..}
```

from inside this fcall:

https://github.com/dotnet/coreclr/blob/c49bce445a6950bd05d7fd6752691da2aef2883b/src/mscorlib/src/System/Reflection/CustomAttribute.cs#L1857

From what I can, the managed caller is passing in everything correctly:

blob = ( 01 00 05 43 6C 61 73 73 00 00 )   // ...Class..  - what I expect to see.
ctor was obtained from here:
 https://github.com/dotnet/coreclr/blob/c49bce445a6950bd05d7fd6752691da2aef2883b/src/mscorlib/src/System/Reflection/CustomAttribute.cs#L1720

and the Reflection object for `ctor` is the properly instantiated version (i.e. `MultiAttribute<Type>`, not `MultiAttribute<>`).

However, this test inside the FCall
 https://github.com/dotnet/coreclr/blob/c49bce445a6950bd05d7fd6752691da2aef2883b/src/vm/customattribute.cpp#L1458

is failing to detect that "ctor's" first parameter is `System.Type` - instead, the TypeHandle from the sig comes back as one of those Canon things.

It's likely that the signature checking code in `customattribute.cpp` isn't coded with generics in mind and needs to be changed. At this point, though, I'm out of my knowledge zone.

@jkotas, @vancem - can you give some guidance here?

File: tests/src/reflection/GenericAttribute/GenericAttributeTests.cs [hunk 4]
```diff
@@ -0,0 +1,163 @@
...
+        AssertAny(b7, a => (a as MultiAttribute<bool>)?.Value == false);
+        AssertAny(b7, a => (a as MultiAttribute<bool>)?.Value == true);
+
+        var a8 = CustomAttributeExtensions.GetCustomAttributes<MultiAttribute<bool?>>(programTypeInfo, true);
+        AssertAny(a8, a => (a as MultiAttribute<bool?>)?.Value == null);
+
+        var b8 = ((ICustomAttributeProvider)programTypeInfo).GetCustomAttributes(typeof(MultiAttribute<bool?>), true);
+        AssertAny(b8, a => (a as MultiAttribute<bool?>)?.Value == null);
+
+        Assert(CustomAttributeExtensions.GetCustomAttributes(programTypeInfo, typeof(MultiAttribute<>), false) == null);
```
AtsushiKan: I don't think it's reasonable to expect these last three tests to work. These apis are statically declared to return `Attribute` arrays but they actually return arrays of the specific attribute type, i.e. this works:

`     MyAttribute[] a = (MyAttribute[])(CAE.GetCustomAttributes(t, typeof(MyAttribute));
`

But you can't instantiate an array of an open type like `MyAttribute<>` so I would expect this to throw with an exception from `Array.Instance`.

jkotas:
`MethodDesc*` does not always carry a full instantation information for shared generic code. You need to keep track of the type instantiation separately. The problem happens at this line in `COMCustomAttribute::CreateCaObject`:

```
    MethodDesc* pCtorMD = gc.refCtor->GetMethod();
```

This needs to be modified to also fetch type instantiation from `refCtor` and propagate it through: the MetaSig used to parse the blob needs to be created using `MetaSig(MethodDesc *pMD, TypeHandle declaringType);` ctor, etc.

AviAvni:
@jkotas I finally managed to implement the missing part with the shared generic issue

src/vm/callhelpers.h
```diff
@@ -234,6 +249,25 @@ class MethodDescCallSite
...
     
+    MethodDescCallSite(MethodDesc* pMD, TypeHandle th) :
+        m_pMD(pMD),
+        m_methodSig(pMD, th),
+        m_argIt(&m_methodSig)
+    {
+        CONTRACTL
+        {
+            THROWS;
+        GC_TRIGGERS;
```
jkotas: Formatting. Tabs vs. spaces?

tests/src/reflection/GenericAttribute/GenericAttributeMetadata.cs
```diff
@@ -0,0 +1,72 @@
+using System;
```
jkotas: Could you please add license header?

AviAvni:
@jkotas I added the license header and the missing spaces sorry

jkotas:
@AviAvni The change looks good to me now. Thanks!

We are too close to snapping 2.1 release and I do not think there is runway to verify this together with the matching C# changes. This will need to wait after 2.1.

tests/src/reflection/GenericAttribute/GenericAttributeTests.cs
```diff
@@ -0,0 +1,163 @@
...
+        Assert(CustomAttributeExtensions.GetCustomAttributes(programTypeInfo, typeof(MultiAttribute<>), true) == null);
```
AtsushiKan: If you pass in an open attribute type to this api (`typeof(MultiAttribute<)`), I think it should either:

- Throw an ArgumentException saying explicitly that the attribute type cannot be an open type (and for consistency, make Attribute.GetCustomAttributes work the same way) I think this is the preferable option as I can't think of a good use case for passing in an open type here.

or 

- Return a zero-length array of type `Attribute[]`.

We shouldn't be returning `null` - there is no precedent for this api returning `null`.

AviAvni: @AtsushiKan I implemented this on Attribute.GetCustomAttributes and with CustomAttributeExtensions api it throws NotSupportedException as expected
but it not work for ICustomAttributeProvider api is this api need also to throw?

AtsushiKan:
We should implement the compatible behavior. I just checked on .NETFx – the ICustomAttribute api returns a zero-length array of type Object[] when passed an open type. That’s a bit unfortunate – I’d have preferred a consistent response from all three variants, but compatibility is more important than reporting an obvious programmer error the same way everywhere. So let’s have that variant continue to return an empty Object array.

AviAvni:
@AtsushiKan I made the change you asked

src/mscorlib/src/System/Attribute.cs
```diff
@@ -459,6 +459,9 @@ public static Attribute[] GetCustomAttributes(MemberInfo element, Type type, boo
             if (type == null)
                 throw new ArgumentNullException(nameof(type));
 
+            if (type.IsGenericTypeDefinition)
```
AtsushiKan: The api you want is `Type.ContainsGenericParameters`. This will catch open types such as `G<int, T>` in addition to `G<>`.

tests/src/reflection/GenericAttribute/GenericAttributeTests.cs
```diff
@@ -0,0 +1,183 @@
...
+using System.Reflection;
+using System.Collections;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
+class Program
+{
+    static int Main(string[] args)
+    {
+        new MultiAttribute<string>("a");
```
AtsushiKan: This line looks like an accidental leftover.

AviAvni: sorry

AtsushiKan:
Ah - I just checked the behavior of Attribute.GetCA and CustomAttributeExtensions.GetCA on NETFX and it seems they do return <b>null</b> when passed in an open generic type. It was the CoreRT framework that was throwing `NotSupportedException`. As much as returning <b>null</b> feels wrong to me, we should stay compatible with NETFX. Your original version (11c28a88eeb72c9e33dbfd54c39e52d481cda9c3) is fine as is. I'll make the change to CoreRT so it matches.

Sorry about the bother.

jkotas:
@AviAvni Could you please resolve the conflict? This can be merged now.

AviAvni:
@jkotas conflict resolved build is green

jkotas:
@dotnet-bot test Windows_NT x64 Checked corefx_baseline
@dotnet-bot test Ubuntu x64 Checked corefx_baseline

AviAvni:
@jkotas build is green

jkotas:
@AviAvni Thank you for your persistence in pushing this through!

AviAvni:
@jkotas thank you and the team very much for helping me I'll follow if there is more things from me
I there are more things needed on this ping me

AviAvni:
Hiii @jkotas I'm working on the Roslyn PR and testing with .net core sdk preview 6
I found that in case the generic attribute parameter is String and the ctor is not the default one then there is a bug that the attribute generic parameter that return from GetCustomAttribute is System.__Canon and not  String I'll try to debug this on monday and check what is missing in the tests

jkotas:
@AviAvni Thank you! If you won't be able to get to this soon, could you please at least file an issue in CoreCLR repo so that we have it tracked?

PR #9191, opened 2017-01-29
CoreRT fat call transformation

sandreenko:
1. Set class handler for calli.
2. Create new flag and phase in the compiler. 
3. Make the transformation after importer in the separate phase.

### DIFF
--- a/src/jit/compiler.cpp
+++ b/src/jit/compiler.cpp
@@ -4194,9 +4194,15 @@ void Compiler::compCompile(void** methodCodePtr, ULONG* methodCodeSize, JitFlags
     assert(!fgComputePredsDone);
     if (fgCheapPredsValid)
     {
-        // Remove cheap predecessors before inlining; allowing the cheap predecessor lists to be inserted
-        // with inlined blocks causes problems.
+        // Remove cheap predecessors before inlining and fat call transformation;
+        // allowing the cheap predecessor lists to be inserted causes problems
+        // with splitting existing blocks.
         fgRemovePreds();
+    }
+
+    if (IsTargetAbi(CORINFO_CORERT_ABI) && doesMethodHaveFatPointer())
+    {
+        fgTransformFatCalli();
     }
 
     EndPhase(PHASE_IMPORTATION);
--- a/src/jit/compiler.h
+++ b/src/jit/compiler.h
@@ -3494,6 +3494,8 @@ public:
 
     void fgImport();
 
+    void fgTransformFatCalli();
+
     void fgInline();
 
     void fgRemoveEmptyTry();
@@ -4692,6 +4694,9 @@ private:
 
 #ifdef DEBUG
     static fgWalkPreFn fgDebugCheckInlineCandidates;
+
+    void               CheckNoFatPointerCandidatesLeft();
+    static fgWalkPreFn fgDebugCheckFatPointerCandidates;
 #endif
 
     void         fgPromoteStructs();
@@ -5468,11 +5473,27 @@ public:
         }
     };
 
-#define OMF_HAS_NEWARRAY 0x00000001  // Method contains 'new' of an array
-#define OMF_HAS_NEWOBJ 0x00000002    // Method contains 'new' of an object type.
-#define OMF_HAS_ARRAYREF 0x00000004  // Method contains array element loads or stores.
-#define OMF_HAS_VTABLEREF 0x00000008 // Method contains method table reference.
-#define OMF_HAS_NULLCHECK 0x00000010 // Method contains null check.
+#define OMF_HAS_NEWARRAY 0x00000001   // Method contains 'new' of an array
+#define OMF_HAS_NEWOBJ 0x00000002     // Method contains 'new' of an object type.
+#define OMF_HAS_ARRAYREF 0x00000004   // Method contains array element loads or stores.
+#define OMF_HAS_VTABLEREF 0x00000008  // Method contains method table reference.
+#define OMF_HAS_NULLCHECK 0x00000010  // Method contains null check.
+#define OMF_HAS_FATPOINTER 0x00000020 // Method contains call, that needs fat pointer transformation.
+
+    bool doesMethodHaveFatPointer()
+    {
+        return (optMethodFlags & OMF_HAS_FATPOINTER) != 0;
+    }
+
+    void setMethodHasFatPointer()
+    {
+        optMethodFlags |= OMF_HAS_FATPOINTER;
+    }
+
+    void clearMethodHasFatPointer()
+    {
+        optMethodFlags &= ~OMF_HAS_FATPOINTER;
+    }
 
     unsigned optMethodFlags;
 
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -23870,6 +23870,389 @@ void Compiler::fgUpdateFinallyTargetFlags()
             }
         }
     }
-
 #endif // FEATURE_EH_FUNCLETS && defined(_TARGET_ARM_)
+}
+
+// FatCalliTransformer transforms calli that can use fat function pointer.
+// Fat function pointer is pointer with the second least significant bit set,
+// if the bit is set, the pointer (after clearing the bit) actually points to
+// a tuple <method pointer, instantiation argument pointer> where
+// instantiationArgument is a hidden first argument required by method pointer.
+//
+// Fat pointers are used in CoreRT as a replacement for instantiating stubs,
+// because CoreRT can't generate stubs in runtime.
+//
+// Jit is responsible for the checking the bit, do the regular call if it is not set
+// or load hidden argument, fix the pointer and make a call with the fixed pointer and
+// the instantiation argument.
+//
+// before:
+//   current block
+//   {
+//     previous statements
+//     transforming statement
+//     {
+//       call with GTF_CALL_M_FAT_POINTER_CHECK flag set in function ptr
+//     }
+//     subsequent statements
+//   }
+//
+// after:
+//   current block
+//   {
+//     previous statements
+//   } BBJ_NONE check block
+//   check block
+//   {
+//     jump to else if function ptr has GTF_CALL_M_FAT_POINTER_CHECK set.
+//   } BBJ_COND then block, else block
+//   then block
+//   {
+//     original statement
+//   } BBJ_ALWAYS remainder block
+//   else block
+//   {
+//     unset GTF_CALL_M_FAT_POINTER_CHECK
+//     load actual function pointer
+//     load instantiation argument
+//     create newArgList = (instantiation argument, original argList)
+//     call (actual function pointer, newArgList)
+//   } BBJ_NONE remainder block
+//   remainder block
+//   {
+//     subsequent statements
+//   }
+//
+class FatCalliTransformer
+{
+public:
+    FatCalliTransformer(Compiler* compiler) : compiler(compiler)
+    {
+    }
+
+    //------------------------------------------------------------------------
+    // Run: run transformation for each block.
+    //
+    void Run()
+    {
+        for (BasicBlock* block = compiler->fgFirstBB; block != nullptr; block = block->bbNext)
+        {
+            TransformBlock(block);
+        }
+    }
+
+private:
+    //------------------------------------------------------------------------
+    // TransformBlock: look through statements and transform statements with fat pointer calls.
+    //
+    void TransformBlock(BasicBlock* block)
+    {
+        for (GenTreeStmt* stmt = block->firstStmt(); stmt != nullptr; stmt = stmt->gtNextStmt)
+        {
+            if (ContainsFatCalli(stmt))
+            {
+                StatementTransformer stmtTransformer(compiler, block, stmt);
+                stmtTransformer.Run();
+            }
+        }
+    }
+
+    //------------------------------------------------------------------------
+    // ContainsFatCalli: check does this statement contain fat pointer call.
+    //
+    // Checks fatPointerCandidate in form of call() or lclVar = call().
+    //
+    // Return Value:
+    //    true if contains, false otherwise.
+    //
+    bool ContainsFatCalli(GenTreeStmt* stmt)
+    {
+        GenTreePtr fatPointerCandidate = stmt->gtStmtExpr;
+        if (fatPointerCandidate->OperIsAssignment())
+        {
+            fatPointerCandidate = fatPointerCandidate->gtGetOp2();
+        }
+        return fatPointerCandidate->IsCall() && fatPointerCandidate->AsCall()->IsFatPointerCandidate();
+    }
+
+    class StatementTransformer
+    {
+    public:
+        StatementTransformer(Compiler* compiler, BasicBlock* block, GenTreeStmt* stmt)
+            : compiler(compiler), currBlock(block), stmt(stmt)
+        {
+            remainderBlock  = nullptr;
+            checkBlock      = nullptr;
+            thenBlock       = nullptr;
+            elseBlock       = nullptr;
+            doesReturnValue = stmt->gtStmtExpr->OperIsAssignment();
+            origCall        = GetCall(stmt);
+            fptrAddress     = origCall->gtCallAddr;
+            pointerType     = fptrAddress->TypeGet();
+        }
+
+        //------------------------------------------------------------------------
+        // Run: transform the statement as described above.
+        //
+        void Run()
+        {
+            ClearFatFlag();
+            CreateRemainder();
+            CreateCheck();
+            CreateThen();
+            CreateElse();
+
+            RemoveOldStatement();
+            SetWeights();
+            ChainFlow();
+        }
+
+    private:
+        //------------------------------------------------------------------------
+        // GetCall: find a call in a statement.
+        //
+        // Arguments:
+        //    callStmt - the statement with the call inside.
+        //
+        // Return Value:
+        //    call tree node pointer.
+        GenTreeCall* GetCall(GenTreeStmt* callStmt)
+        {
+            GenTreePtr   tree = callStmt->gtStmtExpr;
+            GenTreeCall* call = nullptr;
+            if (doesReturnValue)
+            {
+                assert(tree->OperIsAssignment());
+                call = tree->gtGetOp2()->AsCall();
+            }
+            else
+            {
+                call = tree->AsCall(); // call with void return type.
+            }
+            return call;
+        }
+
+        //------------------------------------------------------------------------
+        // ClearFatFlag: clear fat pointer candidate flag from the original call.
+        //
+        void ClearFatFlag()
+        {
+            origCall->ClearFatPointerCandidate();
+        }
+
+        //------------------------------------------------------------------------
+        // CreateRemainder: split current block at the fat call stmt and
+        // insert statements after the call into remainderBlock.
+        //
+        void CreateRemainder()
+        {
+            remainderBlock          = compiler->fgSplitBlockAfterStatement(currBlock, stmt);
+            unsigned propagateFlags = currBlock->bbFlags & BBF_GC_SAFE_POINT;
+            remainderBlock->bbFlags |= BBF_JMP_TARGET | BBF_HAS_LABEL | propagateFlags;
+        }
+
+        //------------------------------------------------------------------------
+        // CreateCheck: create check block, that checks fat pointer bit set.
+        //
+        void CreateCheck()
+        {
+            checkBlock                 = CreateAndInsertBasicBlock(BBJ_COND, currBlock);
+            GenTreePtr fatPointerMask  = new (compiler, GT_CNS_INT) GenTreeIntCon(TYP_I_IMPL, FAT_POINTER_MASK);
+            GenTreePtr fptrAddressCopy = compiler->gtCloneExpr(fptrAddress);
+            GenTreePtr fatPointerAnd   = compiler->gtNewOperNode(GT_AND, TYP_I_IMPL, fptrAddressCopy, fatPointerMask);
+            GenTreePtr zero            = new (compiler, GT_CNS_INT) GenTreeIntCon(TYP_I_IMPL, 0);
+            GenTreePtr fatPointerCmp   = compiler->gtNewOperNode(GT_NE, TYP_INT, fatPointerAnd, zero);
+            GenTreePtr jmpTree         = compiler->gtNewOperNode(GT_JTRUE, TYP_VOID, fatPointerCmp);
+            GenTreePtr jmpStmt         = compiler->fgNewStmtFromTree(jmpTree, stmt->gtStmt.gtStmtILoffsx);
+            compiler->fgInsertStmtAtEnd(checkBlock, jmpStmt);
+        }
+
+        //------------------------------------------------------------------------
+        // CreateCheck: create then block, that is executed if call address is not fat pointer.
+        //
+        void CreateThen()
+        {
+            thenBlock                 = CreateAndInsertBasicBlock(BBJ_ALWAYS, checkBlock);
+            GenTreePtr nonFatCallStmt = compiler->gtCloneExpr(stmt)->AsStmt();
+            compiler->fgInsertStmtAtEnd(thenBlock, nonFatCallStmt);
+        }
+
+        //------------------------------------------------------------------------
+        // CreateCheck: create else block, that is executed if call address is fat pointer.
+        //
+        void CreateElse()
+        {
+            elseBlock = CreateAndInsertBasicBlock(BBJ_NONE, thenBlock);
+
+            GenTreePtr fixedFptrAddress  = GetFixedFptrAddress();
+            GenTreePtr actualCallAddress = compiler->gtNewOperNode(GT_IND, pointerType, fixedFptrAddress);
+            GenTreePtr hiddenArgument    = GetHiddenArgument(fixedFptrAddress);
+
+            GenTreeStmt* fatStmt = CreateFatCallStmt(actualCallAddress, hiddenArgument);
+            compiler->fgInsertStmtAtEnd(elseBlock, fatStmt);
+        }
+
+        //------------------------------------------------------------------------
+        // CreateAndInsertBasicBlock: ask compiler to create new basic block.
+        // and insert in into the basic block list.
+        //
+        // Arguments:
+        //    jumpKind - jump kind for the new basic block
+        //    insertAfter - basic block, after which compiler has to insert the new one.
+        //
+        // Return Value:
+        //    new basic block.
+        BasicBlock* CreateAndInsertBasicBlock(BBjumpKinds jumpKind, BasicBlock* insertAfter)
+        {
+            BasicBlock* block = compiler->fgNewBBafter(jumpKind, insertAfter, true);
+            if ((insertAfter->bbFlags & BBF_INTERNAL) == 0)
+            {
+                block->bbFlags &= ~BBF_INTERNAL;
+                block->bbFlags |= BBF_IMPORTED;
+            }
+            return block;
+        }
+
+        //------------------------------------------------------------------------
+        // GetFixedFptrAddress: clear fat pointer bit from fat pointer address.
+        //
+        // Return Value:
+        //    address without fat pointer bit set.
+        GenTreePtr GetFixedFptrAddress()
+        {
+            GenTreePtr fptrAddressCopy = compiler->gtCloneExpr(fptrAddress);
+            GenTreePtr fatPointerMask  = new (compiler, GT_CNS_INT) GenTreeIntCon(TYP_I_IMPL, FAT_POINTER_MASK);
+            return compiler->gtNewOperNode(GT_XOR, pointerType, fptrAddressCopy, fatPointerMask);
+        }
+
+        //------------------------------------------------------------------------
+        // GetHiddenArgument: load hidden argument.
+        //
+        // Arguments:
+        //    fixedFptrAddress - pointer to the tuple <methodPointer, instantiationArgumentPointer>
+        //
+        // Return Value:
+        //    loaded hidden argument.
+        GenTreePtr GetHiddenArgument(GenTreePtr fixedFptrAddress)
+        {
+            GenTreePtr fixedFptrAddressCopy = compiler->gtCloneExpr(fixedFptrAddress);
+            GenTreePtr wordSize = new (compiler, GT_CNS_INT) GenTreeIntCon(TYP_I_IMPL, genTypeSize(TYP_I_IMPL));
+            GenTreePtr hiddenArgumentPtrPtr =
+                compiler->gtNewOperNode(GT_ADD, pointerType, fixedFptrAddressCopy, wordSize);
+            GenTreePtr hiddenArgumentPtr = compiler->gtNewOperNode(GT_IND, pointerType, hiddenArgumentPtrPtr);
+            return compiler->gtNewOperNode(GT_IND, fixedFptrAddressCopy->TypeGet(), hiddenArgumentPtr);
+        }
+
+        //------------------------------------------------------------------------
+        // CreateFatCallStmt: create call with fixed call address and hidden argument in the args list.
+        //
+        // Arguments:
+        //    actualCallAddress - fixed call address
+        //    hiddenArgument - loaded hidden argument
+        //
+        // Return Value:
+        //    created call node.
+        GenTreeStmt* CreateFatCallStmt(GenTreePtr actualCallAddress, GenTreePtr hiddenArgument)
+        {
+            GenTreeStmt* fatStmt = compiler->gtCloneExpr(stmt)->AsStmt();
+            GenTreePtr   fatTree = fatStmt->gtStmtExpr;
+            GenTreeCall* fatCall = GetCall(fatStmt);
+            fatCall->gtCallAddr  = actualCallAddress;
+            GenTreeArgList* args = fatCall->gtCallArgs;
+            args                 = compiler->gtNewListNode(hiddenArgument, args);
+            fatCall->gtCallArgs  = args;
+            return fatStmt;
+        }
+
+        //------------------------------------------------------------------------
+        // RemoveOldStatement: remove original stmt from current block.
+        //
+        void RemoveOldStatement()
+        {
+            compiler->fgRemoveStmt(currBlock, stmt);
+        }
+
+        //------------------------------------------------------------------------
+        // SetWeights: set weights for new blocks.
+        //
+        void SetWeights()
+        {
+            remainderBlock->inheritWeight(currBlock);
+            checkBlock->inheritWeight(currBlock);
+            thenBlock->inheritWeightPercentage(currBlock, HIGH_PROBABILITY);
+            elseBlock->inheritWeightPercentage(currBlock, 100 - HIGH_PROBABILITY);
+        }
+
+        //------------------------------------------------------------------------
+        // ChainFlow: link new blocks into correct cfg.
+        //
+        void ChainFlow()
+        {
+            assert(!compiler->fgComputePredsDone);
+            checkBlock->bbJumpDest = elseBlock;
+            thenBlock->bbJumpDest  = remainderBlock;
+        }
+
+        Compiler*    compiler;
+        BasicBlock*  currBlock;
+        BasicBlock*  remainderBlock;
+        BasicBlock*  checkBlock;
+        BasicBlock*  thenBlock;
+        BasicBlock*  elseBlock;
+        GenTreeStmt* stmt;
+        GenTreeCall* origCall;
+        GenTreePtr   fptrAddress;
+        var_types    pointerType;
+        bool         doesReturnValue;
+
+        const int FAT_POINTER_MASK = 0b00000010;
+        const int HIGH_PROBABILITY = 80;
+    };
+
+    Compiler* compiler;
+};
+
+#ifdef DEBUG
+
+//------------------------------------------------------------------------
+// fgDebugCheckFatPointerCandidates: callback to make sure there are no more GTF_CALL_M_FAT_POINTER_CHECK calls.
+//
+Compiler::fgWalkResult Compiler::fgDebugCheckFatPointerCandidates(GenTreePtr* pTree, fgWalkData* data)
+{
+    GenTreePtr tree = *pTree;
+    if (tree->IsCall())
+    {
+        assert(!tree->AsCall()->IsFatPointerCandidate());
+    }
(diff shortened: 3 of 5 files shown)


src/jit/flowgraph.cpp
```diff
@@ -23827,3 +23827,238 @@ void Compiler::fgCleanupContinuation(BasicBlock* continuation)
...
+        const int FAT_POINTER_MASK = 0b00000010;
+        const int HIGH_PROBABILITY = 80;
+
+        StatementTransformer(Compiler* compiler, BasicBlock* block, GenTreeStmt* stmt)
+            : compiler(compiler), currBlock(block), stmt(stmt)
+        {
+            remainderBlock = checkBlock = thenBlock = elseBlock = nullptr;
+            doesReturnValue                                     = stmt->gtStmtExpr->OperIsAssignment();
+            SetFptrAddress();
+            pointerType = fptrAddress->TypeGet(); // FIXME
```
sandreenko: Maybe we should do it differently.

src/jit/flowgraph.cpp
```diff
@@ -23827,3 +23827,237 @@ void Compiler::fgCleanupContinuation(BasicBlock* continuation)
...
+        {
+            GenTreePtr fptrAddressCopy = compiler->gtCloneExpr(fptrAddress);
+            GenTreePtr fatPointerMask  = new (compiler, GT_CNS_INT) GenTreeIntCon(TYP_I_IMPL, FAT_POINTER_MASK);
+            return compiler->gtNewOperNode(GT_XOR, pointerType, fptrAddressCopy, fatPointerMask);
+        }
+
+        GenTreePtr GetHiddenArgument(GenTreePtr fixedFptrAddress)
+        {
+            GenTreePtr fixedFptrAddressCopy = compiler->gtCloneExpr(fixedFptrAddress);
+            GenTreePtr wordSize = new (compiler, GT_CNS_INT) GenTreeIntCon(TYP_I_IMPL, sizeof(int*)); // FIXME
```
sandreenko: Do we have method to get target pointer size?

src/jit/importer.cpp
```diff
@@ -6462,6 +6463,11 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
         call->gtCall.callSig  = new (this, CMK_CorSig) CORINFO_SIG_INFO;
         *call->gtCall.callSig = calliSig;
 #endif // DEBUG
+
+        if (IsTargetAbi(CORINFO_CORERT_ABI))
```
MichalStrehovsky: Should there be a check that `calliSig.callConv` is a managed calling convention?

src/jit/flowgraph.cpp
```diff
@@ -23827,3 +23827,237 @@ void Compiler::fgCleanupContinuation(BasicBlock* continuation)
...
+            {
+                block->bbFlags &= ~BBF_INTERNAL;
+                block->bbFlags |= BBF_IMPORTED;
+            }
+            return block;
+        }
+
+        void CreateRemainder()
+        {
+            remainderBlock = compiler->fgSplitBlockAfterStatement(currBlock, stmt);
```
sandreenko: There is a problem with cheap preds, I was able to reproduce it when didn't disable CoreRT IL transformation. Do we have another function, that works with cheap preds?

src/jit/flowgraph.cpp
mikedn: `genTypeSize(TYP_I_IMPL)` would be one option.

src/jit/flowgraph.cpp
mikedn: Can't this transform be done right after `fgRemovePreds` is called instead of just before it?

sandreenko: fgRemoveRefPred removes an edge between the old block and the block, that we have just created in split.

sandreenko: Fixed

src/jit/flowgraph.cpp
sandreenko: Thank you, fixed.

src/jit/importer.cpp
sandreenko: Is that the only one case when we don't need this transformation?

MichalStrehovsky: That's the only one we shouldn't do that is required for correctness.

There are some other cases of managed `calli` where (depending on the method that has the `calli`) the EE side could say that the expansion is not needed. But it's a perf optimization that we can postpone. On Project N (aka .NET Native for Universal Windows Apps), this applies to a few performance critical helpers in the class library that are known to never call through a fat function pointer. We can just file an issue to track it.

sandreenko: FIxed

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23870,239 @@ void Compiler::fgUpdateFinallyTargetFlags()
             }
         }
     }
-
 #endif // FEATURE_EH_FUNCLETS && defined(_TARGET_ARM_)
 }
+
+class FatCalliTransoformer
```
justinvp: Typo? FatCalliTransoformer => FatCalliTransformer

sandreenko: Yes, thanks

sandreenko:
PTAL @dotnet/jit-contrib 

CoreCLR and desktop could be affected by the first two commits, but 
1 the first doesn't change behavior;
2 the second should not change as well.

superpmi asm diff doesn't show any diffs.

sandreenko:
@dotnet-bot  test
Windows_NT x64 Release Priority 1 Build and Test

src/jit/gentree.h
```diff
@@ -3284,6 +3284,9 @@ struct GenTreeCall final : public GenTree
 #define GTF_CALL_M_TAILCALL_VIA_HELPER 0x0200 // GT_CALL -- call is a tail call dispatched via tail call JIT helper.
 #endif                                        // !LEGACY_BACKEND
 
+#define GTF_CALL_M_FAT_POINTER_CHECK 0x0200 // GT_CALL -- it is calli that needs to be tranformed
+                                            // to check CoreRT fat pointer case.
```
sivarv: This flag has the same value as that of GTF_CALL_M_TAILCALL_VIA_HELPER.  It would mean we are not expecting calli to be never dispatched as a tail call via helper.
This might be true now but it would be a good idea to use a different value for it. (e.g. 0x0400)

sivarv: Please add a couple of lines explaining what is CoreRT fat pointer calli here to serve as some context/documentation.

src/jit/gentree.h
```diff
@@ -3284,6 +3284,9 @@ struct GenTreeCall final : public GenTree
 #define GTF_CALL_M_TAILCALL_VIA_HELPER 0x0200 // GT_CALL -- call is a tail call dispatched via tail call JIT helper.
 #endif                                        // !LEGACY_BACKEND
 
+#define GTF_CALL_M_FAT_POINTER_CHECK 0x0200 // GT_CALL -- it is calli that needs to be tranformed
```
sivarv: >#define GTF_CALL_M_FAT_POINTER_CHECK 0x0200 [](start = 0, length = 43)

As you can see in this file, for every flag we have routines that check whether that flag is set or not.

e.g. IsTailCallViaHelper()

I see in other source files we are referencing GTF_CALL_M_FAT_POINTER_CHECK instead of using routines that abstract flag check.

src/jit/importer.cpp
```diff
@@ -6462,6 +6463,15 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
...
+        {
+            bool managedCall = (calliSig.callConv & GTF_CALL_UNMANAGED) == 0;
+            if (managedCall)
+            {
+                call->gtFlags |= GTF_CALL_M_FAT_POINTER_CHECK;
```
sivarv: >call->gtFlags |= GTF_CALL_M_FAT_POINTER_CHECK; [](start = 16, length = 46)

GTF_CALL_M_FAT_POINTER_CHECK - is to be set on call->gtCallMoreFlags not on gtFlags!

GTF_CALL_M prefix denotes that it is a flag to be set on gtCallMoreFlags.

Note that the value of this flag overlaps with GTF_SPILLED_OP2

src/jit/importer.cpp
```diff
@@ -7655,6 +7663,7 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
             if ((call->gtFlags & GTF_CALL_INLINE_CANDIDATE) != 0)
             {
                 assert(opts.OptEnabled(CLFLG_INLINING));
+                assert(!(call->gtFlags & GTF_CALL_M_FAT_POINTER_CHECK)); // We should not try to inline calli.
```
sivarv: > assert(!(call->gtFlags & GTF_CALL_M_FAT_POINTER_CHECK)); [](start = 15, length = 57)

The check should be on gtCallMoreFlags and we should abstract flag check in a routine.

This comment applies to all places where we are setting and checking GTF_CALL_M_FAT_POINTER_CHECK on call->gtFlags.

src/jit/importer.cpp
fadimounir: We'll also need this transform for the GVM call, and in that case, the opcode will be a CALL_VIRT. For v1 of the GVM implementation, the result returned from the jit interface getCallInfo will have a CORINFO_VIRTUALCALL_LDVIRTFTN callinfo kind. The codegen will contain a lookup into a R2R cell, which will return the GVM calli target (which could be a fat function pointer)

fadimounir: See a few lines below: `case CORINFO_VIRTUALCALL_LDVIRTFTN`
And then: `GenTreePtr fptr = impImportLdvirtftn(thisPtr, pResolvedToken, callInfo)`
fptr is the target address that will be called, and could be a fat function pointer.

sandreenko:
Original issue dotnet/corert#1985

src/jit/importer.cpp
sandreenko: do you have test, that can reproduce it?

MichalStrehovsky: @fadimounir If this is only something that is applicable to your private branches, let's skip this for now. This needs to be specced out. For example, we definitely don't want this for all callvirts - just the generic ones - this might require a JitInterface change, etc.

fadimounir: @sandreenko @MichalStrehovsky 
>do you have test, that can reproduce that?
No... I'll create an issue to track this (there will probably be more work needed than just a simple hookup to the changes in this PR). I'll put in a complete description of the problem there.
>let's skip this for now
Absolutely :). Don't make any GVM related changes to this PR, and let it just be scoped to the calli case.

src/jit/importer.cpp

src/jit/importer.cpp

src/jit/gentree.h
sandreenko: Fixed

src/jit/gentree.h
sandreenko: Fixed

sandreenko: Fixed

sandreenko:

sandreenko:

sandreenko:
@dotnet-bot  test Ubuntu x64 Checked Build and Test

src/jit/gentree.h
```diff
@@ -3264,42 +3264,48 @@ struct GenTreeCall final : public GenTree
...
-#define GTF_CALL_M_DOES_NOT_RETURN 0x4000     // GT_CALL -- call does not return
-#define GTF_CALL_M_SECURE_DELEGATE_INV 0x8000 // GT_CALL -- call is in secure delegate
+#define GTF_CALL_M_R2R_REL_INDIRECT 0x00002000 // GT_CALL -- ready to run call is indirected through a relative address
+#define GTF_CALL_M_DOES_NOT_RETURN 0x00004000  // GT_CALL -- call does not return
+#define GTF_CALL_M_SECURE_DELEGATE_INV 0x00008000 // GT_CALL -- call is in secure delegate
+
+#define GTF_CALL_M_FAT_POINTER_CHECK 0x00010000 // GT_CALL -- CoreRT managed calli needs transformation, that checks
+    // special bit in calli address. If it is set, then it is nessesary to restore real
+    // function address and load hidden argument as the first argument for calli.
+    // It is CoreRT replacement for instantiating stubs, because there is no
```
sivarv: Alignment of these comment lines with the one one on line 3304

src/jit/gentree.h
```diff
@@ -3264,42 +3264,48 @@ struct GenTreeCall final : public GenTree
...
+    // a luxury of being able to generate new executable code at runtime.
```
sivarv: > a luxury [](start = 6, length = 9)

(Minor)  typeo "a"?

I would suggest to simplify the last part of the sentence to "..., because executable code cannot be generated at runtime".

src/jit/gentree.h
```diff
@@ -3264,42 +3264,48 @@ struct GenTreeCall final : public GenTree
...
 #define GTF_CALL_M_EXPLICIT_TAILCALL                                                                                   \
-    0x0001                         // GT_CALL -- the call is "tail" prefixed and importer has performed tail call checks
-#define GTF_CALL_M_TAILCALL 0x0002 // GT_CALL -- the call is a tailcall
-#define GTF_CALL_M_VARARGS 0x0004  // GT_CALL -- the call uses varargs ABI
-#define GTF_CALL_M_RETBUFFARG 0x0008        // GT_CALL -- first parameter is the return buffer argument
-#define GTF_CALL_M_DELEGATE_INV 0x0010      // GT_CALL -- call to Delegate.Invoke
-#define GTF_CALL_M_NOGCCHECK 0x0020         // GT_CALL -- not a call for computing full interruptability
-#define GTF_CALL_M_SPECIAL_INTRINSIC 0x0040 // GT_CALL -- function that could be optimized as an intrinsic
-                                            // in special cases. Used to optimize fast way out in morphing
+    0x00000001 // GT_CALL -- the call is "tail" prefixed and importer has performed tail call checks
```
sivarv: > 0x00000001 // GT_CALL -- the call is "tail" prefixed and importer has performed tail call checks [](start = 3, length = 97)

Since you are editing this, see if you can fix the alignment of constants and comment.

src/jit/importer.cpp
```diff
@@ -7664,6 +7676,21 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
...
+                if (fatPointerCandidate)
+                {
+                    assert(!bIntrinsicImported);
+                    assert(IsTargetAbi(CORINFO_CORERT_ABI));
+                    if (call->OperGet() != GT_LCL_VAR) // can be already converted by impFixupCallStructReturn.
+                    {
+                        unsigned   calliSlot  = lvaGrabTemp(true DEBUGARG("calli"));
+                        LclVarDsc* varDsc     = &lvaTable[calliSlot];
+                        varDsc->lvVerTypeInfo = tiRetVal;
+                        impAssignTempGen(calliSlot, call, clsHnd, (unsigned)CHECK_SPILL_NONE);
```
sivarv: >impAssignTempGen(calliSlot, call, clsHnd, (unsigned)CHECK_SPILL_NONE); [](start = 24, length = 70)

What if the return type of call is TYP_VOID?

sivarv: Why fatpointer calli's need to be transformed to be of the form 
var = call
?

Please add a comment explaining the same here.

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23888,253 @@ void Compiler::fgUpdateFinallyTargetFlags()
             }
         }
     }
-
 #endif // FEATURE_EH_FUNCLETS && defined(_TARGET_ARM_)
 }
+
+class FatCalliTransformer
```
sivarv: Please add a comment block explaining at a high-level the input to output transform its does for a calli

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23888,253 @@ void Compiler::fgUpdateFinallyTargetFlags()
...
+    }
+
+    void Run()
+    {
+        BasicBlock* firstBasicBlock = compiler->fgFirstBB;
+        for (BasicBlock* block = firstBasicBlock; block != nullptr; block = block->bbNext)
+        {
+            TransformBlock(block);
+        }
+    }
```
sivarv: Standard function headers are missing

src/jit/gentree.h
sandreenko: It is jit-format choice, and it is strange because it chooses right for GTF_CALL_M_PINVOKE 0x00001000 several lines above.  @adiaaida can you advise how to fix that?

src/jit/importer.cpp
sivarv: Looks like that callRetType cannot be void because of the check on line 7605.

Also impFixupCallStructReturn() will not modify struct return type of a call.

---
In reply to: [99204735](https://github.com/dotnet/coreclr/pull/9191#discussion_r99204735) [](ancestors = 99204735)

sivarv:
Looks good.
(except for feedback to add comments or function headers).

src/jit/gentree.h
BruceForstall: The alignment here is terrible and unreadable. We should just turn off clang-format for this section.

---
In reply to: [99203381](https://github.com/dotnet/coreclr/pull/9191#discussion_r99203381) [](ancestors = 99203381)

src/jit/gentree.h
```diff
@@ -3264,42 +3264,48 @@ struct GenTreeCall final : public GenTree
...
 
-#define GTF_CALL_M_R2R_REL_INDIRECT 0x2000    // GT_CALL -- ready to run call is indirected through a relative address
-#define GTF_CALL_M_DOES_NOT_RETURN 0x4000     // GT_CALL -- call does not return
-#define GTF_CALL_M_SECURE_DELEGATE_INV 0x8000 // GT_CALL -- call is in secure delegate
+#define GTF_CALL_M_R2R_REL_INDIRECT 0x00002000 // GT_CALL -- ready to run call is indirected through a relative address
+#define GTF_CALL_M_DOES_NOT_RETURN 0x00004000  // GT_CALL -- call does not return
+#define GTF_CALL_M_SECURE_DELEGATE_INV 0x00008000 // GT_CALL -- call is in secure delegate
+
+#define GTF_CALL_M_FAT_POINTER_CHECK 0x00010000 // GT_CALL -- CoreRT managed calli needs transformation, that checks
+    // special bit in calli address. If it is set, then it is nessesary to restore real
```
BruceForstall: >nessesary [](start = 62, length = 9)

typo: nessesary

src/jit/gentree.h
```diff
@@ -3497,9 +3503,26 @@ struct GenTreeCall final : public GenTree
...
+    bool IsFatPointerCandidate() const
+    {
+        return (gtCallMoreFlags & GTF_CALL_M_FAT_POINTER_CHECK) != 0;
+    }
+
     bool IsPure(Compiler* compiler) const;
 
-    unsigned short gtCallMoreFlags; // in addition to gtFlags
+    void SetFatPointerCandidate(bool value)
+    {
```
BruceForstall: Why not have SetFatPointerCandidate() and ClearFatPointerCandidate()?

src/jit/compiler.h
```diff
@@ -5468,11 +5473,29 @@ class Compiler
...
+    void setHasFatPointer(bool value)
+    {
+        if (value)
+        {
+            optMethodFlags |= OMF_HAS_FATPOINTER;
+        }
+        else
+        {
+            optMethodFlags ^= ~OMF_HAS_FATPOINTER;
+        }
```
BruceForstall: shouldn't this be `&=`, not `^=`?

BruceForstall: I'd also prefer `SetHasFatPointer()` and `ClearHasFatPointer()`, in general.

---
In reply to: [99243765](https://github.com/dotnet/coreclr/pull/9191#discussion_r99243765) [](ancestors = 99243765)

src/jit/gentree.h
sandreenko: I didn't find existing setter, maybe I was looking bad. Do we prefer set and clear?

src/jit/compiler.h
sandreenko: Yes, thank you,

src/jit/compiler.cpp
```diff
@@ -4191,6 +4191,16 @@ void Compiler::compCompile(void** methodCodePtr, ULONG* methodCodeSize, JitFlags
...
+    {
+        if (fgCheapPredsValid)
+        {
+            fgRemovePreds();
+        }
+        fgTransformFatCalli();
+        setHasFatPointer(false);
+    }
+
     assert(!fgComputePredsDone);
```
BruceForstall: Couldn't you move this code below the `if (fgCheapPredsValid)` case, below, and then remove the `if (fgCheapPredsValid)` code in this `if`? Also, could you put the `setHasFatPointer(false)` at the end of `fgTransformFatCalli()`? In this function, I'd prefer that as much as possible, it's just a linear listing of phases.

Also, is this going to take long enough (e.g., do an IR pass / tree walk) such that you should add a phase name for it, and thus an `EndPhase()` call?

src/jit/compiler.cpp
```diff
@@ -4191,6 +4191,16 @@ void Compiler::compCompile(void** methodCodePtr, ULONG* methodCodeSize, JitFlags
 
     fgImport();
 
+    if (IsTargetAbi(CORINFO_CORERT_ABI) && hasFatPointer())
+    {
+        if (fgCheapPredsValid)
+        {
```
briansull: Also add a comment explaining what fgCheapPredsValid has to do with any of this.

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23888,253 @@ void Compiler::fgUpdateFinallyTargetFlags()
...
+    bool ContainsFatCalli(GenTreeStmt* stmt)
+    {
+        GenTreePtr fatPointerCandidate = stmt->gtStmtExpr;
+        if (fatPointerCandidate->OperIsAssignment())
+        {
+            fatPointerCandidate = fatPointerCandidate->gtGetOp2();
+        }
+        return fatPointerCandidate->IsCall() && fatPointerCandidate->AsCall()->IsFatPointerCandidate();
+    }
+
```
BruceForstall: So the only "fat pointer candidate" forms are statements of the form: `call()` or `var = call()`? Do the importer changes you make guarantee this? Is there a reason why we can't just handle `call` nodes anywhere in the tree?

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23888,253 @@ void Compiler::fgUpdateFinallyTargetFlags()
...
+{
+public:
+    FatCalliTransformer(Compiler* compiler) : compiler(compiler)
+    {
+    }
+
+    void Run()
+    {
+        BasicBlock* firstBasicBlock = compiler->fgFirstBB;
+        for (BasicBlock* block = firstBasicBlock; block != nullptr; block = block->bbNext)
```
BruceForstall: creating a local var for `firstBasicBlock` seems unnecessary.

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23888,253 @@ void Compiler::fgUpdateFinallyTargetFlags()
...
+
+    Compiler* compiler;
+};
+
+#ifdef DEBUG
+void Compiler::CheckNoFatPointerCandidatesLeft()
+{
+    BasicBlock* block = fgFirstBB;
+    noway_assert(block);
+    do
```
BruceForstall: should be a normal `assert`, not a `noway_assert`. (and should explicitly be `assert(block != nullptr)`)

BruceForstall: seems like you should just use a `for` statement over the blocks; I don't see much value asserting that the first block non-null

---
In reply to: [99247721](https://github.com/dotnet/coreclr/pull/9191#discussion_r99247721) [](ancestors = 99247721)

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23888,253 @@ void Compiler::fgUpdateFinallyTargetFlags()
...
+        const int FAT_POINTER_MASK = 0b00000010;
+        const int HIGH_PROBABILITY = 80;
+    };
+
+    Compiler* compiler;
+};
+
+#ifdef DEBUG
+void Compiler::CheckNoFatPointerCandidatesLeft()
+{
```
BruceForstall: Seems like you should place the code for `fgDebugCheckFatPointerCandidates()` immediately above this function, not hidden somewhere above.

src/jit/flowgraph.cpp
BruceForstall: In particular, I want to see comments that answer the questions: (1) what is a fat pointer? (2) what does one look like? (3) where do they come from? (4) how are they used? (5) what platforms use them, and why? (6) what does the JIT need to do to make them work? (7) what does the generated code look like; show an example.

---
In reply to: [99206797](https://github.com/dotnet/coreclr/pull/9191#discussion_r99206797) [](ancestors = 99206797)

src/jit/compiler.h
BruceForstall: It might be useful to be more verbose, for example, `SetMethodHasFatPointer()`. I realized reading the code that I wasn't sure what "thing" was being marked, so adding the word "method" would be useful.

---
In reply to: [99244432](https://github.com/dotnet/coreclr/pull/9191#discussion_r99244432) [](ancestors = 99244432)

src/jit/flowgraph.cpp
sandreenko: Yes, it is done in the same way as inlining candidates were implemented.
The reason is that we want to find them quickly and don't want to deal with cutting trees.

src/jit/compiler.cpp
sandreenko: > Also, is this going to take long enough (e.g., do an IR pass / tree walk) such that you should add a phase name for it, and thus an EndPhase() call?

No, it doesn't.

sandreenko: > Also, could you put the setHasFatPointer(false) at the end of fgTransformFatCalli()? In this function, I'd prefer that as much as possible, it's just a linear listing of phases.

clearHasFatPointer in compiler.cpp after 
`if (hasFatPointer())`
looks more obviously for me, you don't need to open fgTransformFatCalli to see that this method flag is not set after this check. But I am not sure and will try your variant,

sandreenko:
I hope all comments were fixed.
However I am strongly believe that we should insert function headers for small functions for many reasons. I would be glad to drop the last commit ( and rewrite functions that are not obvious without comments).

sandreenko:
@BruceForstall @sivarv could you take another look?

src/jit/compiler.h
```diff
@@ -5468,11 +5473,27 @@ class Compiler
...
-#define OMF_HAS_NULLCHECK 0x00000010 // Method contains null check.
+#define OMF_HAS_NEWARRAY 0x00000001   // Method contains 'new' of an array
+#define OMF_HAS_NEWOBJ 0x00000002     // Method contains 'new' of an object type.
+#define OMF_HAS_ARRAYREF 0x00000004   // Method contains array element loads or stores.
+#define OMF_HAS_VTABLEREF 0x00000008  // Method contains method table reference.
+#define OMF_HAS_NULLCHECK 0x00000010  // Method contains null check.
+#define OMF_HAS_FATPOINTER 0x00000020 // Method contains call, that needs fat pointer transformation.
+
+    bool doesMethodHaveFatPointer()
+    {
```
BruceForstall: Seems like this should be `MethodHasFatPointer()` since `if (MethodHasFatPointer()) ...` reads better than `if (doesMethodHaveFatPointer()) ...`

src/jit/compiler.cpp
```diff
@@ -4194,11 +4194,17 @@ void Compiler::compCompile(void** methodCodePtr, ULONG* methodCodeSize, JitFlags
     assert(!fgComputePredsDone);
     if (fgCheapPredsValid)
     {
-        // Remove cheap predecessors before inlining; allowing the cheap predecessor lists to be inserted
-        // with inlined blocks causes problems.
+        // Remove cheap predecessors before inlining and fat call tranforming;
```
BruceForstall: >tranforming [](start = 66, length = 11)

typo: tranforming => transformation

src/jit/importer.cpp
```diff
@@ -7664,6 +7676,22 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
             }
             else
             {
+                if (fatPointerCandidate)
+                {
+                    // fatPointer candidates should be in statements of the formL call() or var = call().
```
BruceForstall: >formL [](start = 76, length = 5)

typo: `formL`?

BruceForstall: how about explaining why?

---
In reply to: [99698702](https://github.com/dotnet/coreclr/pull/9191#discussion_r99698702) [](ancestors = 99698702)

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23870,392 @@ void Compiler::fgUpdateFinallyTargetFlags()
...
+//     }
+//     subsequent statements
+//   }
+//
+// after:
+//   current block
+//   {
+//     previous statements
+//   } BBJ_ALWAYS check block
+//   check block
```
BruceForstall: Shouldn't this be BBJ_NONE fall-through?

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23870,392 @@ void Compiler::fgUpdateFinallyTargetFlags()
...
+//   } BBJ_ALWAYS remainder block
+//   else block
+//   {
+//     unset GTF_CALL_M_FAT_POINTER_CHECK
+//     load actual function pointer
+//     load instatiation argument
+//     create newArgList = (instatiation argument, original argList)
+//     call (actual function pointer, newArgList)
+//   } BBJ_ALWAYS remainder block
+//   remainder block
```
BruceForstall: Shouldn't this be BBJ_NONE fall-through?

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23870,392 @@ void Compiler::fgUpdateFinallyTargetFlags()
...
+//   } BBJ_COND then block, else block
+//   then block
+//   {
+//     original statement
+//   } BBJ_ALWAYS remainder block
+//   else block
+//   {
+//     unset GTF_CALL_M_FAT_POINTER_CHECK
+//     load actual function pointer
+//     load instatiation argument
```
BruceForstall: >instatiation [](start = 12, length = 12)

typo: instatiation => instantiation (also, same typo below)

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23870,392 @@ void Compiler::fgUpdateFinallyTargetFlags()
...
+            {
+                StatementTransformer stmtTransformer(compiler, block, stmt);
+                stmtTransformer.Run();
+            }
+        }
+    }
+
+    //------------------------------------------------------------------------
+    // ContainsFatCalli: check does this statement contain fat pointer call.
+    //
```
BruceForstall: IMO, you should state the conditions it intends to check, that is, the IR forms that are expected/allowed.

src/jit/compiler.h
sandreenko: the problem is that we have to write Method in the name. 
hasMethodFatPointer is confusing.
MethodHasFatPointer doesn't start with verb and should be methodHasFatPointer in our CC.

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23870,392 @@ void Compiler::fgUpdateFinallyTargetFlags()
...
+        //    new basic block.
+        BasicBlock* CreateAndInsertBasicBlock(BBjumpKinds jumpKind, BasicBlock* insertAfter)
+        {
+            BasicBlock* block = compiler->fgNewBBafter(jumpKind, insertAfter, true);
+            if ((insertAfter->bbFlags & BBF_INTERNAL) == 0)
+            {
+                block->bbFlags &= ~BBF_INTERNAL;
+                block->bbFlags |= BBF_IMPORTED;
+            }
+            return block;
```
BruceForstall: Please explain (here, and in a comment) why you need to clear the BBF_INTERNAL bit in this case.

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23870,392 @@ void Compiler::fgUpdateFinallyTargetFlags()
...
+//     load instatiation argument
+//     create newArgList = (instatiation argument, original argList)
+//     call (actual function pointer, newArgList)
+//   } BBJ_ALWAYS remainder block
+//   remainder block
+//   {
+//     subsequent statements
+//   }
+//
+class FatCalliTransformer
```
BruceForstall: Have you tested this with debugging? What does the debugging info look like for your newly created blocks? That is, the IL offsets. I'm worried single stepping isn't going to work.

src/jit/flowgraph.cpp
sandreenko: Yes, you are right.

src/jit/flowgraph.cpp
sandreenko: Yes, you are right.

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23870,392 @@ void Compiler::fgUpdateFinallyTargetFlags()
...
+        //
+        void ChainFlow()
+        {
+            compiler->fgRemoveRefPred(remainderBlock, currBlock);
+            compiler->fgAddRefPred(checkBlock, currBlock);
+            compiler->fgAddRefPred(thenBlock, checkBlock);
+            compiler->fgAddRefPred(elseBlock, checkBlock);
+            compiler->fgAddRefPred(remainderBlock, thenBlock);
+            compiler->fgAddRefPred(remainderBlock, elseBlock);
+            checkBlock->bbJumpDest = elseBlock;
```
BruceForstall: This transformation phase is done before preds have been computed. You shouldn't need to do any of this. Instead, you should just assert (someplace): `assert(!fgComputePredsDone)`

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23870,392 @@ void Compiler::fgUpdateFinallyTargetFlags()
...
+        const int HIGH_PROBABILITY = 80;
+    };
+
+    Compiler* compiler;
+};
+
+#ifdef DEBUG
+
+/*****************************************************************************
+* Callback to make sure there is no more GTF_CALL_M_FAT_POINTER_CHECK calls.
```
BruceForstall: >is [](start = 30, length = 2)

grammar: is => are

BruceForstall: Please use standard header, with `//` not `/* */`

---
In reply to: [99704078](https://github.com/dotnet/coreclr/pull/9191#discussion_r99704078) [](ancestors = 99704078)

src/jit/flowgraph.cpp
```diff
@@ -23870,6 +23870,392 @@ void Compiler::fgUpdateFinallyTargetFlags()
...
+//------------------------------------------------------------------------
+// CheckNoFatPointerCandidatesLeft: walk through blocks and check that there are no fat pointer candidates left.
+//
+void Compiler::CheckNoFatPointerCandidatesLeft()
+{
+    assert(!doesMethodHaveFatPointer());
+    for (BasicBlock* block = fgFirstBB; block; block = block->bbNext)
+    {
+        for (GenTreeStmt* stmt = fgFirstBB->firstStmt(); stmt; stmt = stmt->gtNextStmt)
+        {
```
BruceForstall: CC: compare explicitly against nullptr, e.g. `block != nullptr`, `stmt != nullptr`

src/jit/compiler.cpp
```diff
@@ -4194,11 +4194,17 @@ void Compiler::compCompile(void** methodCodePtr, ULONG* methodCodeSize, JitFlags
...
-        // with inlined blocks causes problems.
+        // Remove cheap predecessors before inlining and fat call transformation;
+        // allowing the cheap predecessor lists to be inserted causes problems
+        // with splitting existing blocks.
         fgRemovePreds();
     }
 
+    if (IsTargetAbi(CORINFO_CORERT_ABI) && doesMethodHaveFatPointer())
+    {
+        fgTransformFatCalli();
```
BruceForstall: Why does this phase need to happen so early? It seems like it should run during fgMorph(), after inlining.

sandreenko: Different positions for this transformation were discussed. 
I tried to implement it in morph and lowering, but it was harder.

src/jit/flowgraph.cpp
sandreenko: Yes, it works as expected.

src/jit/importer.cpp
sandreenko: ` Such form allows to find statements with fat calls without walking through whole trees and removes problems with cutting trees.`
I will add this to the comment.

src/jit/flowgraph.cpp
sandreenko: I want to fix it in a separate commit because it is common issue.

sandreenko:
@dotnet-bot test Windows_NT arm64 Cross Checked Build and Test  please

sandreenko:

src/jit/flowgraph.cpp
sandreenko: Fixed

src/jit/compiler.cpp
BruceForstall: Why was it harder? I really think this is the wrong place.

AndyAyersMS: I believe it could fairly easily be moved back after inlining since `fgInsertInlineeBlocks` will merge inlinee `optMethodFlags` into the root method. 

IIRC from discussions the "harder" part was when we were talking about deferring this even further, say to lower.

sandreenko: I don't understand your doubts.

1. It supposed to be part of importer, because it is transformation from IL calli into our IR.
It doesn't have anything with "Massage the trees so that we can generate code out of them".
Are you are agree that we should set the flag in importer?
2. We don't want to transform cfg inside fgImport, it is why this transformation is done as separate phase. 
3. We are trying to make this transformation as close as possible to importer to avoid any collisions with different optimizations. (for example tail calls or struct promotions).

I can move this phase, it will create many new checks in phases between importer and the new place. Do you want it just to avoid setting BBF_IMPORTED on created blocks?

fadimounir: [FYI] We also do it in the importer in UTC for ProjectN, if i remember correctly.

sandreenko:
@dotnet-bot  test Windows_NT arm64 Cross Checked Build and Test

Issue #7670, opened 2016-10-17
InvalidOperationException(Environment.GetResourceString("InvalidOperation_UnknownEnumType") at System_Private_CoreLib_ni!System.Enum.InternalFormattedHexString()+0x13e


RussKeldorph:
Repro steps:

```
Build coreclr Windows x64 checked (build.cmd x64 checked skiptests)
Run corefx tests with above
    build.cmd -Release -- /p:BUILDTOOLS_OVERRIDE_RUNTIME=<coreclr_root>\bin\Product\Windows_NT.x64.Checked
```

Failing test assembly: System.Runtime.Tests.dll

coreclr commit: bc6bd52c44209647bc2541dad55a05cb2058f76f
corefx commit: dotnet/corefx@9b270c7b06751ab24d98d1404e4fd671c63aeb9e

RussKeldorph:
@gkhanna79

RussKeldorph:
[This test](https://github.com/dotnet/corefx/blob/5a361ea9edf9280229841b4e70044c48e476ef23/src/System.Runtime/tests/System/EnumTests.cs#L1483) is hitting [this assertion](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Enum.cs#L73).  In my opinion, assertions should not be used to validate parameters on public interfaces.  I would recommend simply removing the assertions in several places in Enum.cs when an exception is going to be thrown anyway.  There are other Debug.Asserts in this file in paths that simply return a default value instead of throwing an exception.  I'm not sure if those are reachable from public interfaces, but, if they are, simply removing the assertion may not be appropriate.

Another thing I noticed is that the test case will accept a "ContractException" (in addition to the InvalidOperationException that it receives in Release builds).  Did Debug.Assert result in a ContractException at one point?  @jkotas, could this be related to your rename of Contract.Assert to Debug.Assert (the last change to Enum.cs).  Not sure; just speculating as to how the test might have come to look like that.

@dotnet/coreclr-contrib I need to stop this assertion failing our CI jobs.  Feel free to point out an owner if there is one.

RussKeldorph:
@jkotas Just noticed I opened this before you made the Debug.Assert change, so disregard that part.

RussKeldorph:
I should clarify that this test is now failing with an assertion failure rather than an unhandled exception, but I believe the root issue is the fact that the test is intentionally passing an invalid parameter but not recovering properly.

PR #9192, opened 2017-01-29
Remove invalid asserts on unsupported underlying enum type

jkotas:
The underlying enum type can be floating point type or native integer for historic reasons.
However, not all framework methods support these special enum types. They should throw exception
for them instead of asserting.

Fix issue #7670


jkotas:
cc @danmosemsft @AtsushiKan

PR #9202, opened 2017-01-30
Mark RetBuf as SIMD variable

BruceForstall:
I see changes where storeIndir is used now instead of
a BLK copy, which is an IR shape improvement.

This improves codegen for SIMD8, degrades it for SIMD12 (due
to encoding size differences with SSE instructions, though
the IR looks better), and is neutral for SIMD16, in the
VectorReturn test.

### DIFF
--- a/src/jit/lclvars.cpp
+++ b/src/jit/lclvars.cpp
@@ -488,6 +488,16 @@ void Compiler::lvaInitRetBuffArg(InitVarDscInfo* varDscInfo)
                 varDsc->lvType = TYP_I_IMPL;
             }
         }
+#ifdef FEATURE_SIMD
+        else if (featureSIMD && varTypeIsSIMD(info.compRetType))
+        {
+            varDsc->lvSIMDType = true;
+            varDsc->lvBaseType =
+                getBaseTypeAndSizeOfSIMDType(info.compMethodInfo->args.retTypeClass, &varDsc->lvExactSize);
+            assert(varDsc->lvBaseType != TYP_UNKNOWN);
+        }
+#endif // FEATURE_SIMD
+
         assert(isValidIntArgReg(varDsc->lvArgReg));
 
 #ifdef DEBUG


BruceForstall:
@sivarv @CarolEidt PTAL
cc @dotnet/jit-contrib

src/jit/lclvars.cpp
```diff
@@ -488,6 +488,26 @@ void Compiler::lvaInitRetBuffArg(InitVarDscInfo* varDscInfo)
...
+        {
+            CORINFO_SIG_INFO sigInfo;
+            info.compCompHnd->getMethodSig(info.compMethodHnd, &sigInfo);
+
+            // If the return buffer is a SIMD type, we need to note that by setting lvSIMDType.
+            var_types simdBaseType = TYP_UNKNOWN;
+            var_types type         = impNormStructType(sigInfo.retTypeClass, nullptr, nullptr, &simdBaseType);
+            if (simdBaseType != TYP_UNKNOWN)
+            {
+                assert(!"hello!");
```
sivarv: Left over code meant for debugging purpose?

BruceForstall: oops.

src/jit/lclvars.cpp
```diff
@@ -488,6 +488,26 @@ void Compiler::lvaInitRetBuffArg(InitVarDscInfo* varDscInfo)
...
         }
+#ifdef FEATURE_SIMD
+        else if (featureSIMD && varTypeIsSIMD(info.compRetType))
+        {
+            CORINFO_SIG_INFO sigInfo;
+            info.compCompHnd->getMethodSig(info.compMethodHnd, &sigInfo);
+
+            // If the return buffer is a SIMD type, we need to note that by setting lvSIMDType.
+            var_types simdBaseType = TYP_UNKNOWN;
+            var_types type         = impNormStructType(sigInfo.retTypeClass, nullptr, nullptr, &simdBaseType);
```
sivarv: We have info.compMethodInfo->args.retTypeClass gives the class handle of the return type, there is no need to call getMethodSig() VM call?

I see impNormStructType() call for SIMD type struct return in lvaInitTypeRef() and it would be redundant here.  I think here we can call

varDsc->lvSIMDType = true
varDsc->lvBaseType = getBaseTypeAndSizeOfSIMDType(info.compMethodInfo->args.retTypeClass, &varDsc->lvExactSize);


Given info.compRetType is a SIMD type, its base type cannot be TYP_UNKNOWN.  In fact we should assert it here rather than checking.

BruceForstall:
@sivarv nice! Much cleaner now.

sivarv:
:shipit:

BruceForstall:
Fixed formatting.

PR #9203, opened 2017-01-31
Adding some missing APIs on core that where added to N.S 2.0

safern:
This is work needed in coreclr for issue [dotnet/corefx #15362](https://github.com/dotnet/corefx/issues/15362)

ArgIterator apis throw PlatformNotSupported() in implementation since in order to enable Vararg calling convetions there is work needed in the JIT as @jkotas mentions [here](https://github.com/dotnet/standard/issues/20#issuecomment-272775599)

Tests are [dotnet/corefx #15643](https://github.com/dotnet/corefx/pull/15643)

/cc @weshaggard @danmosemsft

### DIFF
--- a/src/mscorlib/src/System/ArgIterator.cs
+++ b/src/mscorlib/src/System/ArgIterator.cs
@@ -153,10 +153,11 @@ namespace System {
         public override bool Equals(Object o) 
         {  
             throw new PlatformNotSupportedException(); //The JIT requires work to enable ArgIterator see: https://github.com/dotnet/coreclr/issues/9204.
+        }
 
         public override int GetHashCode()
         { 
-            throw new PlatformNotSupportedException(); //The JIT requires work to enable ArgIterator see: https://github.com/dotnet/coreclr/issues/9204..
+            throw new PlatformNotSupportedException(); //The JIT requires work to enable ArgIterator see: https://github.com/dotnet/coreclr/issues/9204.
         }
 
         [System.CLSCompliantAttribute(false)]
--- a/src/mscorlib/src/System/ArgIterator.cs
+++ b/src/mscorlib/src/System/ArgIterator.cs
@@ -136,50 +136,49 @@ namespace System {
 #else
         public ArgIterator(RuntimeArgumentHandle arglist)
         {
-            throw new PlatformNotSupportedException(); //The JIT requires work to enable ArgIterator see: https://github.com/dotnet/standard/issues/20#issuecomment-272775599.
+            throw new PlatformNotSupportedException(); //The JIT requires work to enable ArgIterator see: https://github.com/dotnet/coreclr/issues/9204.
         }
 
         [CLSCompliant(false)]
         public unsafe ArgIterator(RuntimeArgumentHandle arglist, void* ptr)
         {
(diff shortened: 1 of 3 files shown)


src/mscorlib/src/System/ArgIterator.cs
```diff
@@ -133,5 +133,54 @@ public override bool Equals(Object o)
...
+
+        [System.CLSCompliantAttribute(false)]
+        public System.TypedReference GetNextArg(System.RuntimeTypeHandle rth)
+        {
+            throw new PlatformNotSupportedException(); //The JIT requires work to enable ArgIterator https://github.com/dotnet/standard/issues/20#issuecomment-272775599.
+        }
+
+        public unsafe System.RuntimeTypeHandle GetNextArgType()
+        {
+            throw new PlatformNotSupportedException(); //The JIT requires work to enable ArgIterator https://github.com/dotnet/standard/issues/20#issuecomment-272775599.
```
danmosemsft: Instead of this link, I would have one to an open (or existing) issue in the CoreCLR repo to track this work.

src/vm/ecalllist.h
```diff
@@ -373,10 +373,8 @@ FCFuncEnd()
 
 FCFuncStart(gTypedReferenceFuncs)
     FCFuncElement("InternalToObject", ReflectionInvocation::TypedReferenceToObject)
-#ifndef FEATURE_CORECLR
     FCFuncElement("InternalSetTypedReference", ReflectionInvocation::SetTypedReference)
```
danmosemsft: These you've now exposed won't actually get called yet right? I suppose there is no harm to un-ifdeffing them in advance.

danmosemsft:
LGTM

src/mscorlib/src/System/ArgIterator.cs
safern: Yup will do that when this is merged

src/vm/ecalllist.h
safern: The TypedReference and FieldInfo ones will actually get called.

These ones will not [src](https://github.com/dotnet/coreclr/pull/9203/files#diff-3667dffbd11675529c85670ef344242eL1815) because these ones are ArgIterator functions.

src/mscorlib/src/System/ArgIterator.cs
```diff
@@ -133,5 +133,54 @@ public override bool Equals(Object o)
             
         private IntPtr ArgPtr;                  // Pointer to remaining args.
         private int    RemainingArgs;           // # of remaining args.
+#else
+        public ArgIterator(RuntimeArgumentHandle arglist)
+        {
+            throw new PlatformNotSupportedException(); //The JIT requires work to enable ArgIterator see: https://github.com/dotnet/coreclr/issues/9204.
```
jkotas: Nit: it would be nice to add the comment above the `throw` to avoid a super long line.

danmosemsft: I asked them to do it this way as periodically we wantto grep for PNSE and this way it appears conveniently in the output, I dont have to go visit the code.

PR #9209, opened 2017-01-31
Simplify #ifdef in EECodeManager::GetGSCookieAddr 

parjong:
This commit simplifies #ifdef inside EECodeManager::GetGSCookieAddr via enabling it only when CROSSGEN_COMPILE is not defined.


parjong:
@janvorli Please take a look.

janvorli:
LGTM, thank you!

PR #9211, opened 2017-01-31
Use -O1 optimization for arm and armel

hqueue:
Because there are bugs for -O3 optimization in clang (<=3.8) for arm, let's use -O1 instead of -O3 for arm.

This PR is made to address a request at https://github.com/dotnet/core-setup/issues/790#issuecomment-275433662

### DIFF
--- a/src/pal/tools/clang-compiler-override-arm.txt
+++ b/src/pal/tools/clang-compiler-override-arm.txt
@@ -1,0 +1,20 @@
+SET (CMAKE_C_FLAGS_INIT                "-Wall -std=c11")
+SET (CMAKE_C_FLAGS_DEBUG_INIT          "-g -O0")
+SET (CLR_C_FLAGS_CHECKED_INIT          "-g -O1")
+# Refer to the below instruction to support __thread with -O2/-O3 on Linux/ARM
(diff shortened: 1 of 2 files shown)


hqueue:
@gkhanna79 PTAL. When building for arm and armel, I've replaced all `-O2` and `-O3` with `-O1 `for `release` and `checked` build.

/cc: @jyoungyun @hseok-oh @chunseoklee

janvorli:
@hqueue I wonder if we should rather require clang >= 3.9 instead. Btw, are we sure that the issue was fixed in 3.9?

gkhanna79:
@janvorli once @hqueue can identify the toolset that is confirmed to work, we can then look transitioning to it.

The change looks good to me. @hqueue Were you able to confirm that, with this change, the issues you ran into when running tests are now addressed?

hqueue:
@janvorli We are not sure yet, because we didn't test core-setup, coreclr and corefx with clang-3.9 altogether yet. At least, coreclr is built and ran well, but we didn't test them together and other featcures including sos plugin. We will also look into them this time and let you know.

@gkhanna79 Actually this patch was what my collegues and I have been used for a long time to build `Linux.arm.Release`  with `clang 3.6` about an year. So I think this PR works correclty and addresses clang -O3 problem.

PR #9212, opened 2017-01-31
[x86/Linux] Use Explicit Field Reference in FillRegDisplay

parjong:
CONTEXT and KNONVOLATILE_CONTEXT_POINTERS has incompatible ordering of Ebx, Esi, Edi, Ebp fields, and thus the update on these fields via for loop inside FillRegDisplay is invalid.

This commit replaces for loop with explicit field update.

### DIFF
--- a/src/inc/regdisp.h
+++ b/src/inc/regdisp.h
@@ -393,10 +393,10 @@ inline void FillRegDisplay(const PREGDISPLAY pRD, PT_CONTEXT pctx, PT_CONTEXT pC
 
     pRD->ctxPtrsOne.Lr = &pctx->Lr;
 #elif defined(_TARGET_X86_) // _TARGET_ARM_
-    for (int i = 0; i < 4; ++i)
-    {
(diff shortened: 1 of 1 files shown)


parjong:
@janvorli Please take a look.

janvorli:
I think we should rather reorder the registers in the KNONVOLATILE_CONTEXT_POINTERS.

parjong:
To fix this issue via reordering, we need to reorder the registers in ``CONTEXT`` instead of ``KNONVOLATILE_CONTEXT_POINTERS``.

IMO, the updates based on for loop is unstable as it depends on both ``CONTEXT`` and ``KNONVOLATILE_CONTEXT_POINTERS``. Could you let me know if there is a problem with explicit reference?

In addition, is it possible changes the order of ``CONTEXT``? It is defined in ``src/pal/inc/pal.h``, and thus it would have corresponding definitions in Windows.

janvorli:
I see, I have not realized the CONTEXT doesn't have the registers next to each other. So let's keep it as you have it.

PR #9216, opened 2017-01-31
Fixes to make the lldb plugin work on ARM64

qmfrederik:
A first set of fixes to make the lldb plugin load correctly on lldb on arm64

### DIFF
--- a/src/ToolBox/SOS/Strike/exts.cpp
+++ b/src/ToolBox/SOS/Strike/exts.cpp
@@ -111,30 +111,29 @@ ArchQuery(void)
     {
         targetMachine = AMD64Machine::GetInstance();
     }
-#endif // SOS_TARGET_AMD64
-#ifdef SOS_TARGET_X86
+#elif defined(SOS_TARGET_X86)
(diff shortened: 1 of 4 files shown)


src/ToolBox/SOS/Strike/exts.cpp
```diff
@@ -111,30 +111,29 @@ ArchQuery(void)
...
-#endif // SOS_TARGET_ARM64
+#else
+#error "Undefined target architecture"
+#endif
 
     if (targetMachine == NULL)
     {
         g_targetMachine = NULL;
-        ExtErr("SOS does not support the current target architecture.\n");
+        ExtErr("SOS does not support the current target architecture %llx.\n", targetArchitecture);
```
janvorli: Just a nit, I would print `0x` in front of the printed value to make it clear it is hex. Otherwise LGTM.

qmfrederik:
@janvorli OK, added the `0x`  and also moved `IMAGE_FILE_MACHINE_ARM64` to `lldbservices.h`

PR #9224, opened 2017-01-31
Make SpinLock.TryEnter(ref bool) fail faster

stephentoub:
One uses SpinLock.TryEnter(ref bool) when they want to do a quick test to either acquire the lock or know it's held by someone else.  We can make the latter path much faster so that it fails faster.

With this change, TryEnter(ref bool) on an already held SpinLock is ~3.2x faster than before.  It looks like there may be a small ~2-3% regression when the lock isn't already held, but it's hard to say for sure.

cc: @jkotas, @benaadams, @kouvel

### DIFF
--- a/src/mscorlib/src/System/Threading/SpinLock.cs
+++ b/src/mscorlib/src/System/Threading/SpinLock.cs
@@ -183,7 +183,22 @@ namespace System.Threading
         /// </exception>
         public void TryEnter(ref bool lockTaken)
         {
-            TryEnter(0, ref lockTaken);
+            int observedOwner = m_owner;
+            if (((observedOwner & LOCK_ID_DISABLE_MASK) == 0) | lockTaken)
+            {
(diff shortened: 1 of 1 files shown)


src/mscorlib/src/System/Threading/SpinLock.cs
```diff
@@ -183,7 +183,22 @@ public void Enter(ref bool lockTaken)
         /// </exception>
         public void TryEnter(ref bool lockTaken)
         {
-            TryEnter(0, ref lockTaken);
+            int observedOwner = m_owner;
+            if ((observedOwner & LOCK_ID_DISABLE_MASK) == 0 | lockTaken)
```
benaadams: Didn't realise you could bitwise a bool 😄

benaadams:
LGTM and the trade off is a good one

PR #9228, opened 2017-01-31
Add/Expose String.EndsWith(char) and String.StartsWith(char)

fujiy:
Add/Expose String.EndsWith(char) and String.StartsWith(char)

Fix https://github.com/dotnet/corefx/issues/4805

### DIFF
--- a/src/mscorlib/src/System/String.Comparison.cs
+++ b/src/mscorlib/src/System/String.Comparison.cs
@@ -780,13 +780,32 @@ namespace System
         }
 
         [Pure]
-        internal bool EndsWith(char value) {
+        internal bool EndsWithOrdinal(char value) {
             int thisLen = this.Length;
             if (thisLen != 0) {
                 if (this[thisLen - 1] == value)
                     return true;
             }
             return false;
+        }
+
+        [Pure]
+        public Boolean EndsWith(Char value)
+        {
+            return EndsWith(new string(value, 1), StringComparison.CurrentCulture);
+        }
+
+        [Pure]
+        [ComVisible(false)]
+        public Boolean EndsWith(Char value, StringComparison comparisonType)
+        {
+            return EndsWith(new string(value, 1), comparisonType);
+        }
+
+        [Pure]
+        public Boolean EndsWith(Char value, Boolean ignoreCase, CultureInfo culture)
+        {
+            return EndsWith(new string(value, 1), ignoreCase, culture);
         }
 
         // Determines whether two strings match.
@@ -1134,6 +1153,25 @@ namespace System
                 referenceCulture = culture;
 
             return referenceCulture.CompareInfo.IsPrefix(this, value, ignoreCase ? CompareOptions.IgnoreCase : CompareOptions.None);
+        }
+
+        [Pure]
+        public Boolean StartsWith(Char value)
+        {
+            return StartsWith(new string(value, 1));
+        }
+
+        [Pure]
+        [ComVisible(false)]
+        public Boolean StartsWith(Char value, StringComparison comparisonType)
+        {
+            return StartsWith(new string(value, 1), comparisonType);
+        }
+
+        [Pure]
+        public Boolean StartsWith(Char value, Boolean ignoreCase, CultureInfo culture)
+        {
+            return StartsWith(new string(value, 1), ignoreCase, culture);
         }
     }
 }
--- a/src/mscorlib/src/System/String.Comparison.cs
+++ b/src/mscorlib/src/System/String.Comparison.cs
@@ -780,32 +780,9 @@ namespace System
         }
 
         [Pure]
-        internal bool EndsWithOrdinal(char value) {
+        public Boolean EndsWith(char value) {
             int thisLen = this.Length;
-            if (thisLen != 0) {
-                if (this[thisLen - 1] == value)
-                    return true;
-            }
-            return false;
-        }
-
-        [Pure]
-        public Boolean EndsWith(Char value)
-        {
-            return EndsWith(new string(value, 1), StringComparison.CurrentCulture);
-        }
-
-        [Pure]
-        [ComVisible(false)]
-        public Boolean EndsWith(Char value, StringComparison comparisonType)
-        {
-            return EndsWith(new string(value, 1), comparisonType);
-        }
-
-        [Pure]
-        public Boolean EndsWith(Char value, Boolean ignoreCase, CultureInfo culture)
-        {
-            return EndsWith(new string(value, 1), ignoreCase, culture);
+            return thisLen != 0 && this[thisLen - 1] == value;
         }
 
         // Determines whether two strings match.
@@ -1156,22 +1133,6 @@ namespace System
         }
 
         [Pure]
-        public Boolean StartsWith(Char value)
-        {
-            return StartsWith(new string(value, 1));
-        }
-
-        [Pure]
-        [ComVisible(false)]
-        public Boolean StartsWith(Char value, StringComparison comparisonType)
-        {
-            return StartsWith(new string(value, 1), comparisonType);
-        }
-
-        [Pure]
-        public Boolean StartsWith(Char value, Boolean ignoreCase, CultureInfo culture)
-        {
-            return StartsWith(new string(value, 1), ignoreCase, culture);
-        }
+        public Boolean StartsWith(char value) => Length != 0 && m_firstChar == value;
     }
 }
--- a/src/mscorlib/src/System/String.Comparison.cs
+++ b/src/mscorlib/src/System/String.Comparison.cs
@@ -780,7 +780,7 @@ namespace System
         }
 
         [Pure]
-        public Boolean EndsWith(char value) {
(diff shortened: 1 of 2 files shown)


src/mscorlib/model.xml
```diff
@@ -7297,6 +7297,7 @@
       <Member Name="EndsWith(System.String)" />
       <Member Name="EndsWith(System.String,System.Boolean,System.Globalization.CultureInfo)" />
       <Member Name="EndsWith(System.String,System.StringComparison)" />
+      <Member Name="EndsWith(System.Char)" />
```
justinvp: Nit: Do we try to keep these sorted? Should the `System.Char` overload be listed above the `System.String` overloads?

src/mscorlib/src/System/String.Comparison.cs
```diff
@@ -1144,5 +1144,16 @@ public override int GetHashCode()
 
             return referenceCulture.CompareInfo.IsPrefix(this, value, ignoreCase ? CompareOptions.IgnoreCase : CompareOptions.None);
         }
+
+        [Pure]
+        public Boolean StartsWith(Char value)
```
justinvp: This can be simplified as:

```c#
public bool StartsWith(char value) => Length != 0 && this[0] == value;
```

src/mscorlib/model.xml
fujiy: Done

src/mscorlib/src/System/String.Comparison.cs
```diff
@@ -783,7 +783,7 @@ public int CompareTo(String strB)
         }
 
         [Pure]
-        internal bool EndsWith(char value) {
+        public Boolean EndsWith(char value) {
```
justinvp: This could be simplified as:

```c#
public bool EndsWith(char value)
{
    int thisLen = this.Length;
    return thisLen != 0 && this[thisLen - 1] == value;
}
```

src/mscorlib/src/System/String.Comparison.cs
fujiy: I thought the same, but I was not sure if I should use that syntax, I didn't see much code wrote like that. I'll simplify it

justinvp: Actually, it may be faster to check `m_firstChar` directly, e.g.:

```c#
public bool StartsWith(char value) => Length != 0 && m_firstChar == value;
```

src/mscorlib/src/System/String.Comparison.cs
fujiy: The same here, I was not sure if I should use the more verbose syntax or not. I'll change it. I also prefer this way. Thanks

src/mscorlib/src/System/String.Comparison.cs
fujiy: Hmm. I didn't know about that field. I'll see how it works

fujiy: Maybe I should just write
public bool StartsWith(char value) => m_firstChar == value;

Since:

        // For empty strings, this will be '\0' since
        // strings are both null-terminated and length prefixed
        [NonSerialized] private char m_firstChar;

fujiy: What about the m_stringLength, would be faster than Length property too?

justinvp: Most other methods on string use `Length`, so I'm guessing the JIT makes it as fast as it can, but it might be worth comparing the code the JIT generates.

src/mscorlib/src/System/String.Comparison.cs
fujiy: Done

src/mscorlib/src/System/String.Comparison.cs
```diff
@@ -783,13 +783,9 @@ public int CompareTo(String strB)
...
```
justinvp: Nit: `Boolean` -> `bool`. I know this file is inconsistent, but at some point soon it's going to be cleaned up to follow the CoreFX coding style where language keywords are preferred over BCL types.

src/mscorlib/src/System/String.Comparison.cs
```diff
@@ -1144,5 +1140,8 @@ public override int GetHashCode()
 
             return referenceCulture.CompareInfo.IsPrefix(this, value, ignoreCase ? CompareOptions.IgnoreCase : CompareOptions.None);
         }
+
+        [Pure]
+        public Boolean StartsWith(char value) => Length != 0 && m_firstChar == value;
```
justinvp: Nit: `Boolean` -> `bool`

src/mscorlib/src/System/String.Comparison.cs
fujiy: I'll try to compare the assembly code generate. But you are right:

        /// This is a EE implemented function so that the JIT can recognise is specially
        /// and eliminate checks on character fetchs in a loop like:
        ///        for(int i = 0; i < str.Length; i++) str[i]
        /// The actually code generated for this will be one instruction and will be inlined.
        //
        // Spec#: Add postcondition in a contract assembly.  Potential perf problem.
        public extern int Length {
            [MethodImplAttribute(MethodImplOptions.InternalCall)]
            get;
        }

fujiy:
Test Windows_NT x64 Debug Build and Test

fujiy:
Test Windows_NT x64 Debug Build and Test

src/mscorlib/src/System/String.Comparison.cs
fujiy: I can't see any difference from a normal property. It gets inlined (as any simple property)

I created the following test to validate it, but couldn't get any conclusion from that: https://gist.github.com/fujiy/284a31a3237720c8f9e895844a426698

I also did the same tests running on .NET 4.6.2. Interesting, the field returning method get a lot more code in .NET Core

src/mscorlib/src/System/String.Comparison.cs
fujiy: Done

src/mscorlib/src/System/String.Comparison.cs
fujiy: Done

danmosemsft:
Is this ready for a final review
@AlexGhiondea

fujiy:
model.xml does not exists anymore?

justinvp:
> model.xml does not exists anymore?

Correct. It was recently deleted as the rewriter is no longer used.

fujiy:
Ok. I rebased my fork, the conflict doesn't happen anymore

fujiy:
Just to be sure, there is no pending action for me?

justinvp:
@danmosemsft, this is ready to be reviewed/merged... who should review?

danmosemsft:
Per https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/issue-guide.md (which applies to corelib also) @AlexGhiondea and @joperezr shoudl reivew.

For any changes to corelib, we have begun asking that there be a corefx change ready with any necessary test changes or API exposing. This is because there were a few cases in the past where the first happened but the second didnt. Not sure if there's a corefx one for this.

danmosemsft:
Incidentally @fujiy not only is model.xml gone but also we are not really maintaining mscorlib.cs anymore. Your edits are fine but not really needed. It will be gone in a couple weeks. There is an issue Jan is working on.

fujiy:
I created a PR in CoreFX to expose them. It is temporarily failing (as expected): https://github.com/dotnet/corefx/pull/16198

danmosemsft:
OK. If you don't get a CR in a couple of days (on both) please @ summon them.

src/mscorlib/src/System/String.Comparison.cs
```diff
@@ -780,13 +780,9 @@ public int CompareTo(String strB)
         }
 
         [Pure]
-        internal bool EndsWith(char value) {
+        public bool EndsWith(char value) {
             int thisLen = this.Length;
```
danmosemsft: Please remove 'this' per our coding guidelines.

fujiy: Done

danmosemsft:
@dotnet-bot test windows_nt x64 release priority 1 build and test (https://github.com/dotnet/coreclr/issues/9630 https://github.com/dotnet/coreclr/issues/9631)

fujiy:
@dotnet-bot test windows_nt x64 release priority 1 build and test

danmosemsft:
@dotnet-ainot test windows_nt x64 release priority 1 build and test (myget again?)

danmosemsft:
@dotnet-bot test windows_nt x64 release priority 1 build and test

fujiy:
@dotnet-bot test windows_nt x64 release priority 1 build and test

danmosemsft:
Thanks @fujiy let's get the tests merged next week!

fujiy:
Thanks @danmosemsft and @brfalcon who committed the firsts implementations

PR #9229, opened 2017-01-31
increase gtCallMoreFlags size

sandreenko:
gtCallMoreFlags was short and all 16 values were already used. We need
one more for CoreRT fat pointers and we wll need more
in the future.

Needs for #9191.
PTAL @dotnet/jit-contrib

### DIFF
--- a/src/jit/gentree.h
+++ b/src/jit/gentree.h
@@ -3264,42 +3264,42 @@ struct GenTreeCall final : public GenTree
     }
 
 #define GTF_CALL_M_EXPLICIT_TAILCALL                                                                                   \
-    0x0001                         // GT_CALL -- the call is "tail" prefixed and importer has performed tail call checks
-#define GTF_CALL_M_TAILCALL 0x0002 // GT_CALL -- the call is a tailcall
-#define GTF_CALL_M_VARARGS 0x0004  // GT_CALL -- the call uses varargs ABI
-#define GTF_CALL_M_RETBUFFARG 0x0008        // GT_CALL -- first parameter is the return buffer argument
-#define GTF_CALL_M_DELEGATE_INV 0x0010      // GT_CALL -- call to Delegate.Invoke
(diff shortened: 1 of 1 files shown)


BruceForstall:
Does this increase the size of the CALL node on any architecture?

@pkukol Any comments on CALL node size implications?

sivarv:
Please measure increase in memory allocated to gentree nodes due to this change.

CarolEidt:
I believe that we are attempting to take a pretty hard line on adding additional bits to the `GenTree` nodes, and therefore we should consider whether there are alternatives. At a very minimum, as @sivarv suggests we should measure additional memory allocated - as well as measuring throughput impact.

sandreenko:
> Does this increase the size of the CALL node on any architecture?
sizeof(GenTreeCall) == 200 before and after change on x64 debug.
but let we get struct layout.

sivarv:
Size on release is interesting than on debug/checked.

sandreenko:
```
Before:
136     | gtCallMoreFlags
138.    | gtCallType (bitstart=0,nbits=3)
138.    | gtReturnType (bitstart=3,nbits=5)
        | <alignment member> (size=5)
144     | gtRetClsHnd

After: 
136     | gtCallMoreFlags
140.    | gtCallType (bitstart=0,nbits=3)
140.    | gtReturnType (bitstart=3,nbits=5)
        | <alignment member> (size=3)
144     | gtRetClsHnd

```

sandreenko:
@adiaaida Can you please confirm that failing tests are not connected with this PR?

CarolEidt:
Nice when you can increase the size of a field and it fits in existing padding space!

adiaaida:
They look like it, but to be safe, I'd rerun the legs. I just checked in the change to disable the test that was causing all the other tests to OOM.

sivarv:
Looks good.

Issue #8772, opened 2017-01-02
Add relevant CHA inside Roslyn and RyuJIT


ciplogic:
RyuJIT could handle it more seriously the sealed contract and devirtualize and inline more aggresively.

[Issue in Roslyn repository](https://github.com/dotnet/roslyn/issues/16187)

benaadams:
Related https://github.com/dotnet/coreclr/issues/1166

jamesqo:
One place this would really help is `StringBuilder.ToString()`.

AndyAyersMS:
Just to clarify: RyuJit proper doesn't have enough context to do CHA. 

However, the runtime can do analyses like these on the jit's behalf, since the runtime implements the "type system" and so understands how classes and method overrides and such interrelate. And if you look at the code for [getCallInfo](https://github.com/dotnet/coreclr/blob/master/src/vm/jitinterface.cpp#L5189-L5339) in the runtime-jit interface you see that sealed is used in some cases to devirtualize, but there are various complications and servicing restrictions.

I don't know if the analysis the runtime does is as general as it can be, but I suspect trying to extend this to consider internal or private attributes will break various currently supported use cases.

morganbr:
.NET Native does this currently (at compile time). It's been shown to save some size on disk in native images and can also help the inliner. It hasn't been tested for meaningful performance improvements, but somebody interested in this could test sealed vs in-devirtualizable methods out on .NET Native.

gkhanna79:
CC @kouvel

AndyAyersMS:
I have a prototype for sealed method/class devirtualization in the works. 

Will post a link to my fork when it's a little bit further along. Does not appear it requires anything new on the runtime side other than one or two jit interface calls.

Issue #8321, opened 2016-11-28
Failing threading.SleepEx.test2


janvorli:
The test has failed in the CI test run below with the following error:
`Expected thread to sleep for 2000 ms (and not be interrupted). Average delta: 858993393 ms, acceptable delta: 150`
https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_centos7.1_prtest/436/
It seems we have a bug in the delta computation. Based on the test start times in the log, there was no delay in test execution.

janvorli:
@kount, can you take a look please?

AndyAyersMS:
Maybe unrelated but https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_centos7.1_prtest/3351 failed on
```
threading.Sleep.test1 (from (empty))
The sleep function slept for 2353 ms when it should have slept for 2000 ms
```

danmosemsft:
https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_centos7.1_prtest/3516/testReport/junit/threading/SleepEx/test2/

danmosemsft:
@kouvel maybe write out the timestamp before and after each sleep, at least in the failure case, so we can see whehter we're getting reasonable numbers?

AndyAyersMS:
Another instance: https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_centos7.1_prtest/3608/

adityamandaleeka:
I just ran into a similar issue, but in this case, the time it slept for was slightly less than expected:
```
The sleep function slept for 1984 ms when it should have slept for 2000 ms
```
https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_centos7.1_prtest/3695/testReport/junit/threading/SleepEx/test1/

pgavlin:
(continuing discussion from #10261)

> This is a general problem with all of the wait tests that verify that a wait completes in reasonable time, especially when the machine is overloaded running other tests in parallel. These tests are useful when bringing up new systems and a sufficient buffer should be included when expecting wait completion. The buffer needs to be adjusted, for that part I'll consider this a duplicate of #8321.

[...]

Issue #1166, opened 2015-06-22
JIT should treat sealed virtual methods as normal ones and inline them where appropriate


omariom:
Currently JIT treats sealed methods as virtual. 

```C#
[MethodImpl(MethodImplOptions.NoInlining)]
private int TestSealedMethodInlining(Derived obj)
{
    return obj.GetValue(3);
}

class Base
{
    public virtual int GetValue(int value)
    {
        return 0x33;
    }
}

class Derived : Base
{
    public sealed override int GetValue(int value)
    {
        return value;
    }
}

```

TestSealedMethodInlining in machine code:
```ASM
 sub         rsp,28h  
 mov         qword ptr [rsp+38h],rdx  
 mov         rcx,rdx  
 mov         edx,3  
 mov         rax,qword ptr [rsp+38h]  
 mov         rax,qword ptr [rax]  
 mov         rax,qword ptr [rax+40h]  
 call        qword ptr [rax+20h]  
 nop  
 add         rsp,28h  
 ret  
```

What I would expect:
```ASM
  mov   rdx, qword ptr [rdx]
  mov   rax, 3 
  ret
```

whoisj:
You know I'd always assumed it did the "right thing". :+1: On getting it fixed.

omariom:
Great :+1:

sharwell:
If I had to guess the reason this happens, it's the fact that the bytecode actually references the base method. From ILSpy; see `IL_0003`:

```
.method private hidebysig 
	instance int32 TestSealedMethodInlining (
		class ClassLibrary1.Class1/Derived obj
	) cil managed noinlining 
{
	// Method begins at RVA 0x2050
	// Code size 13 (0xd)
	.maxstack 2
	.locals init (
		[0] int32
	)

	IL_0000: nop
	IL_0001: ldarg.1
	IL_0002: ldc.i4.3
	IL_0003: callvirt instance int32 ClassLibrary1.Class1/Base::GetValue(int32)
	IL_0008: stloc.0
	IL_0009: br.s IL_000b

	IL_000b: ldloc.0
	IL_000c: ret
} // end of method Class1::TestSealedMethodInlining
```

omariom:
@sharwell 
That's very strange. Why C# compiler emits a call to the base class?

omariom:
@sharwell 
I've checked what JIT does when Derived method referenced - it is the same.

zabulus:
[...]

PR #9230, opened 2017-01-31
Simple devirtualization

AndyAyersMS:
Prototype for simple devirtualization; requires final classes or methods.

The first commit is the jit interface side of things, the second is the jit side.


AndyAyersMS:
Marking no merge for now since this is work in progress.

This impacts ~267 methods in System.Private.CoreLib,. 453 sites are devirtualized. No hits in other assemblys in jit-diff, presumably because of R2R restrictions, though I have yet to dig in deeper. Simple jitted test cases show expected results.

Particular areas for review: implementation of the jit interface call and the fragile way that call related info is updated.

@jkotas, @JosephTremoulet PTAL
cc @dotnet/jit-contrib 

Related issues: #1166, #8772

src/jit/importer.cpp
```diff
@@ -7309,6 +7309,13 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
             /* only true object pointers can be virtual */
 
             assert(obj->gtType == TYP_REF);
+
+            // If we have a virtual call, is the class of 'this' a subtype of
+            // the class of the method? If so perhaps we can devirtualize.
+            if ((call->gtFlags & GTF_CALL_VIRT_KIND_MASK) == GTF_CALL_VIRT_VTABLE)
```
jkotas: You should also do this for `CORINFO_VIRTUALCALL_STUB`. It is used for interface calls under JIT and fragile NGen, and for all virtual calls under R2R. Thus, you should see some hits for R2R once you do that.

src/jit/importer.cpp
```diff
@@ -18269,4 +18276,286 @@ bool Compiler::IsMathIntrinsic(GenTreePtr tree)
...
+    CORINFO_CLASS_HANDLE  baseClass  = info.compCompHnd->getMethodClass(baseMethod);
+
+    // Bail out if there's missing class info.
+    if (baseClass == nullptr)
+    {
+        return;
+    }
+
+    // If the objClass is sealed (final), then we may be able to devirtualize.
+    const DWORD objClassAttribs = info.compCompHnd->getClassAttribs(objClass);
```
jkotas: It should be cheaper (less VM/EE calls) and cleaner to let the VM do all these checks that are pre-reqs for devirtualization.

src/vm/jitinterface.cpp
```diff
@@ -8912,6 +8912,72 @@ void CEEInfo::getMethodVTableOffset (CORINFO_METHOD_HANDLE methodHnd,
 }
 
 /*********************************************************************/
+CORINFO_METHOD_HANDLE CEEInfo::resolveVirtualMethod(CORINFO_METHOD_HANDLE methodHnd,
+                                                    CORINFO_CLASS_HANDLE derivedClass)
+{
+    CORINFO_METHOD_HANDLE result;
+
+    JIT_TO_EE_TRANSITION();
```
jkotas: You do not need this if you just forwarding to other JIT/EE interface method.

src/ToolBox/superpmi/superpmi-shared/icorjitinfoimpl.h
```diff
@@ -125,6 +125,16 @@
             unsigned*                   offsetAfterIndirection  /* OUT */
             );
 
+    CORINFO_METHOD_HANDLE resolveVirtualMethod(
+        CORINFO_METHOD_HANDLE virtualMethod,
+        CORINFO_CLASS_HANDLE implementingClass
+        );
+
+    CORINFO_METHOD_HANDLE resolveVirtualMethodExact(
```
jkotas: It would be useful to have some comments what these methods do. (In particular, what's the difference between them.)

src/vm/jitinterface.cpp
```diff
@@ -8912,6 +8912,72 @@ void CEEInfo::getMethodVTableOffset (CORINFO_METHOD_HANDLE methodHnd,
...
+    _ASSERTE(slot < method->GetMethodTable()->GetNumVirtuals());
+
+    // TODO: Derived class should have base class as parent...
+    TypeHandle DerivedClsHnd(derivedClass);
+    MethodTable* pMT = DerivedClsHnd.GetMethodTable();
+
+    // MethodDescs returned to JIT at runtime are always fully loaded. Verify that it is the case.
+    _ASSERTE(pMT->IsRestored() && pMT->IsFullyLoaded());
+
+    MethodDesc* pMD = pMT->GetMethodDescForSlot(slot);
```
jkotas: For R2R, this will also need the version bubble check.

AndyAyersMS: Are you saying the current approach (where the jit calls canInline on the resulting method handle) is insufficient?

src/ToolBox/superpmi/superpmi-shared/icorjitinfoimpl.h
AndyAyersMS: Sure. This is the shim layer header for SPMI, so I'll put something cursory here and something more detailed over in corinfo.h

src/jit/importer.cpp
AndyAyersMS: Agreed. I'll refactor to consolidate stuff on the VM side after I get a bit more functionality in place.

src/jit/importer.cpp
AndyAyersMS: Have this sort of working, and yes I'm seeing it kick in for R2R code.

src/vm/jitinterface.cpp
jkotas: canInline should work, but it is too pessimistic check and makes the optimization order dependent because of the noinlining hint is cached. I do not think you should be using canInline to gate it.

src/vm/jitinterface.cpp
```diff
@@ -8912,6 +8912,72 @@ void CEEInfo::getMethodVTableOffset (CORINFO_METHOD_HANDLE methodHnd,
...
 /*********************************************************************/
+CORINFO_METHOD_HANDLE CEEInfo::resolveVirtualMethod(CORINFO_METHOD_HANDLE methodHnd,
+                                                    CORINFO_CLASS_HANDLE derivedClass)
+{
+    CORINFO_METHOD_HANDLE result;
+
+    JIT_TO_EE_TRANSITION();
+
+    result = resolveVirtualMethodExact(methodHnd, derivedClass);
+
```
briansull: Why do we have two methods here?

src/vm/jitinterface.cpp
AndyAyersMS: Ah, right. I have comments about the unfortunate impact of the noinline hint; this would be a good way to get past it.

src/vm/jitinterface.cpp
AndyAyersMS: I was matching jit interface changes from .Net Native. It looks like I may not need both versions.

src/vm/jitinterface.cpp
AndyAyersMS: Seems like I can maybe use `getMethodBeingCompiled()` for this? Something like:
```
    // Allow devirtialization if jitting, or if prejitting and the
    // method being jitted and the devirtualized method are in the
    // same versioning bubble.
    if (!IsCompilingForNGen() || IsInSameVersionBubble(getMethodBeingCompiled(), pMD))
    {
        result = (CORINFO_METHOD_HANDLE) pMD;
    }
```
Otherwise I need to pass in another method handle.

jkotas: Yes, this should work.

AndyAyersMS:
Could not repro release crossgen failure at b334a7f; wonder if I need to rebase to hit it...?

AndyAyersMS:
Failures in 56f702a are missing null checks; looks like at least some of the vcall stubs do implicit null checking, which will now need to be made explicit. Not sure if I should enable this for all the stub devirts or just a subset; for now I'll turn it on for all.

AndyAyersMS:
Same tests failed again in 0d17370; serves me right for jumping to the conclusion that it was stub null checks at fault. Will debug locally this time...

src/jit/importer.cpp
```diff
@@ -18269,4 +18271,325 @@ bool Compiler::IsMathIntrinsic(GenTreePtr tree)
...
+#ifdef FEATURE_READYTORUN_COMPILER
+        if (opts.IsReadyToRun())
+        {
+            // We should refetch the lookup value from the VM.
+            call->setEntryPoint(callInfo->codePointerLookup.constLookup);
+            call->gtCallMoreFlags &= ~GTF_CALL_M_VIRTSTUB_REL_INDIRECT;
+        }
+#endif
+
+        // Need to update call info too. This is fragile
```
JosephTremoulet: Do you think we could run these checks upstream, at the point where we're calling `getCallInfo` today?  Then the `callInfo` wouldn't be mutating on us, and `importCall` could set these other flags on the call node as usual.

In fact, since you're already intending to push a number of these checks over to the VM, I'm wondering if we could end up with an interface where `getCallInfo` just takes an additional parameter which is the most-derived "thisPtr" type that the jit can determine...

src/jit/importer.cpp
```diff
@@ -18269,4 +18271,325 @@ bool Compiler::IsMathIntrinsic(GenTreePtr tree)
...
+        return;
+    }
+
+    // Bail if devirt is disabled.
+    if (JitConfig.JitEnableDevirtualization() == 0)
+    {
+        return;
+    }
+
+    // Do we know anything about the type of the 'this'?
```
AndyAyersMS: Yes, we will (always?) see an upstream Box and can work out the type from there. 

However I will probably leave this as future work since it would also be nice to optimize away the boxing at the same time.

src/jit/importer.cpp
```diff
@@ -18269,4 +18271,325 @@ bool Compiler::IsMathIntrinsic(GenTreePtr tree)
...
+    // happen when prejitting, if the devirtualization crosses
+    // servicing bubble boundaries.
+    if (derivedMethod == nullptr)
+    {
+        JITDUMP("--- no derived method, sorry\n");
+        return;
+    }
+
+    // Fetch attributes and do more vetting of the method. In general,
+    // if we can't inline, we won't devirtualize.
```
AndyAyersMS: This is a stale comment. I used to call `canInline` but have since moved the relevant checks to the VM side.

src/jit/importer.cpp
```diff
@@ -18269,4 +18271,325 @@ bool Compiler::IsMathIntrinsic(GenTreePtr tree)
...
+        const char* const derivedMethodName = eeGetMethodName(derivedMethod, &derivedMethodClassName);
+
+        printf("    devirt to %s::%s%s\n", derivedMethodClassName, derivedMethodName, derivedMethodNote);
+        gtDispTree(call);
+    }
+#endif // defined(DEBUG)
+
+    if (objClassIsFinal || derivedMethodIsFinal)
+    {
+        JITDUMP("!!! Inlining ok, no restrictions, final; can devirtualize\n");
```
AndyAyersMS: Also a stale comment.

src/jit/importer.cpp
AndyAyersMS: This is one area I'm still experimenting with.

I don't like the idea of patching the call info or the various context handles, and it seems like the patching may get more complex when extending this to support shared generics. 

But I am not sure we can anticipate all the info we might need early.

For instance, I'd like to call back into the devirtualization code during inlining or even later on during optimization, as we potentially learn more about types.

JosephTremoulet: Right now you have it coded up as

  1. (if we can) get an updated `CORINFO_METHOD_HANDLE` (else abort)
  2. update the `GenTreeCall` node to the new target (returned from 1) and flags (inferred from change to nonvirtual call)
  3. update the `CORINFO_CALL_INFO`

I think you might be happier to switch the second two:

  1. (if we can) get an updated `CORINFO_METHOD_HANDLE` (else abort)
  2. update the `CORINFO_CALL_INFO`
  3. update the `GenTreeCall` node to the new target (returned from 1) and flags (returned from 2)

because

  1. You can code the changes to the `GenTreeCall` as a function of the `CORINFO_CALL_INFO`, so you may be able to share that code with the importer upstream and optimizations downstream
  2. You can separate the concerns of determining what the new info/attributes should be (which would be the part computing the new `CORINFO_CALL_INFO`) and determining which fields on a `GenTreeCall` need to change in response to devirtualization (which could be pretty mechanical).  Currently you have a mix of those things in your second and third steps.
  3. It would line up with the desire to move some of this to the VM side, since we're used to the VM populating the `CORINFO_CALL_INFO`

AndyAyersMS: Wanted to get the post-inline case implemented so whatever approach I finally settle on handles both importer and inliner variants smoothly.

AndyAyersMS:
So with latest updates the jit can devirtualize based on actual return types. For example, in the "sealed default" pattern from #9276:
```C#
using System;

public class Base
{
    public virtual int Foo() { return 33; }
    static BaseSealed s_Default = new BaseSealed();
    public static Base Default => s_Default;
}

sealed class BaseSealed : Base {}

public class Test
{
    public static int Main()
    {
        Base b = Base.Default;
        int x = b.Foo();
        return (x == 33 ? 100 : -1);
    }
}
``` 
the jit will devirtualize the call to `Foo` in `Main`.

Inlining `Foo` is a bit trickier; my current thinking is that we should split trees in the importer at all virtual calls to make some of the logistics simpler later on.

AndyAyersMS:
Thinking those ARM64 failures may be related to #9375.

AndyAyersMS:
Latest change ends up triggering a lot of non-crossgen fatal R2R errors of the form:
```
Unspecified error (Exception from HRESULT: 0x80004005 (E_FAIL)) while compiling method Microsoft.CodeAnalysis.CSharp.CSharpCompilation.GetBinderFactory
```
suspect something is off either with the devirt update to R2R calls or the dependence tracking.

AndyAyersMS:
Still working through this. Was thrown off by the fact that in R2R mode crossgen will throw/catch for "expected" failures.

AndyAyersMS:
With the 68af9f2 changes we lose 68 methods in R2R images, running into the "Method entrypoint cannot be encoded" notimpl in `ZapInfo::getFunctionEntryPoint`. So presumably after devirt the entry point info is not getting updated properly and the jit incorrectly tries to make a direct call.

I haven't been able to find a way to re-invoke `getCallInfo` for the new call target without making things worse. Have tried consing up a resolved token using the class/method handles I already have, as well as trying to recreate the information needed to call `resolveToken`; both just lead to more problems.

@jkotas any ideas?

src/vm/jitinterface.cpp
```diff
@@ -8912,6 +8912,112 @@ void CEEInfo::getMethodVTableOffset (CORINFO_METHOD_HANDLE methodHnd,
...
+    bool allowDevirt = false;
+    if (IsCompilingForNGen())
+    {
+        MethodDesc* pCallerMD = m_pMethodBeingCompiled;
+        Module* pCallerModule = pCallerMD->GetLoaderModule();
+        Module* pDevirtModule = pDevirtMD->GetModule();
+        Assembly* pCallerAssembly = pCallerModule->GetAssembly();
+        Assembly* pDevirtAssembly = pDevirtModule->GetAssembly();
+
+        if (pCallerAssembly == pDevirtAssembly)
```
jkotas: This should be just:

```
if (IsReadyToRunCompilation())
{
    allowDevirt = IsInSameVersionBubble(m_pMethodBeingCompiled->GetModule()->GetAssembly(), pDevirtModule->GetModule()->GetAssembly());
}
else
{
   allowDevirt = true;
}
```

jkotas:
> @jkotas any ideas?

You can remove the TODO+THROW from `ZapInfo::getFunctionEntryPoint`. It is possible

src/vm/jitinterface.cpp
AndyAyersMS: Just to make sure I understand: this extra check is for the case where the class the jit initially identifies as the object type is a subclass of the class that introduces the method, right? 

If so, I though I had that covered over on the jit side by mapping back from the devirtualized method to the introducing class, and passing the implementing class to resolveVirtual.

jkotas: > this extra check is for the case where the class the jit initially identifies as the object type is a subclass of the class that introduces the method, right?

Right.

jkotas: I have not realized that it may be handled in the JIT side. It may be good idea to do the version bubble check defensively here anyway.

AndyAyersMS: Agreed, will update.

AndyAyersMS:
Removing the throw leads to an AV in `ZapMethodEntryPointTable::GetMethodEntryPoint` on since `m_pImage->m_pMethodEntryPoints` is null.

src/vm/jitinterface.cpp
AndyAyersMS: Also just to be clear -- no version bubble checks for ngen?

jkotas: Right. Fragile NGen is like JIT.

AndyAyersMS:
@jkotas looks like for R2R it is happy if I reinvoke `getCallInfo`, this triggers a bunch of stuff on the zap side.
I have to fake up a resolved token but that doesn't seem all that tricky, at least for for non-shared cases.

AndyAyersMS:
Snip from the latest jit-diff stats. It would be really nice if jit-diff would also report how many methods remained unchaged; I'm curious what % of methods this change impacts.

The 41 methods that get lost are still a concern. Looks like most of then now bail out in R2R with: Runtime method access checks not supported. Could be a bug in devirt, since overrides should be just as accessible as the base methods. Will dig in deeper...

```
Total bytes of diff: -30127 (-0.21 % of base)
    diff is an improvement.

Total byte diff includes -18436 bytes from reconciling methods
        Base had   41 unique methods,    18571 unique bytes
        Diff had    1 unique methods,      135 unique bytes
...
2051 total methods with size differences (1695 improved, 356 regressed).

**Comment by AndyAyersMS** [comment]
Rebased and pushed a few updates. Still pretty rough around the edges.

Jit can now spot "exact" types from `newobj` and so devirtualize some cases w/o final class or method.

Fixed issue with context update on the call for cases where the base method is from a non-generic class and the derived class is a shared generic.

R2R dropout issues still hit in 41 methods, mostly in Roslyn, still unsure as to what exactly is going wrong. In those cases, the zapper `getCallInfo` fails with an unsupported access check, so suspect something is amiss with the faked-up resolved token. Note we have to re-call `getCallInfo` for R2R to trigger proper recording. Have been looking for a simpler repro for this but no luck so far.

**Comment by AndyAyersMS** [comment]
@dotnet-bot test Windows_NT_x64 perf

**Comment by AndyAyersMS** [comment]

File: src/jit/importer.cpp [hunk 12]
```diff
@@ -18390,4 +18396,371 @@ bool Compiler::IsMathIntrinsic(GenTreePtr tree)
...
+    //   IL_021d:  ldloc.0
+    //   IL_021e:  callvirt   instance int32 System.Object::GetHashCode()
+    if ((objClassAttribs & CORINFO_FLG_INTERFACE) != 0)
+    {
+        JITDUMP("--- obj class is interface, sorry\n");
+        return;
+    }
+
+    // Bail if obj class is a shared instance, in particular watch for System::__Canon.
+    // Perhaps overly broad...?
```
davidwrighton: I believe this will also fire for things like SomeClass<__Canon> which is definitely overly broad.

src/vm/jitinterface.cpp
davidwrighton: On UTC the resolveVirtualMethod canonicalize's its result in most cases, but resolveVirtualMethodExact does not. This is used to handle the inlining of generic dictionary cell usage into the outer method. Given that RyuJit has a different model for handling shared generics that can't do the inlining anyways, this is probably ok. I would suggest only implementing the resolveVirtualMethod api, and not implementing resolveVirtualMethodExact. (This is simply done, as the implementation below is effectively the same as resolveVirtualMethod should be returning, but isn't quite the same as what resolveVirtualMethodExact would return in ProjectN.)

src/jit/importer.cpp
jkotas: On the VM side, you can just compare the class handle, like `clsHnd == TypeHandle(g_pCanonMethodTableClass)`. I do not think there is a flag that let's you do it on the JIT side right now.

AndyAyersMS:
Rebased, squashed commits down again into two: one for runtime side and one for jit side. 

Incorporate some feedback: remove the unused interface method, add the check for __Canon over to the runtime side and remove the overly general check on the jit side.

AndyAyersMS:
Think I cleared up the problems that were causing loss of some R2R methods. No losses seen in the current jit-diff framework set.

@dotnet-bot test Windows_NT Checked r2r

AndyAyersMS:
Pushed an update so the jit interface should be in final form for merging. New GUID and all.
Am going to remove the no merge tag and ask for one last round of review.

@JosephTremoulet PTAL
cc @dotnet/jit-contrib 

FYI this will require a coordinated update for desktop. I'll get that ready and kick off desktop testing.

AndyAyersMS:
Another forced update to fix formatting, polish the commit text, and remove an unused declaration.

benaadams:
> For example, in the "sealed default" pattern from #9276: ... the jit will devirtualize the call to Foo in Main.

@AndyAyersMS this still true at current iteration? i.e. is it worth reactivating that PR?

AndyAyersMS:
Yes, it will devirtualize (but not inline, yet) -- provided you use the property directly when you make calls.

AndyAyersMS:
CentOS failure [(log)](https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_centos7.1_prtest/3351/) was in a PAL test; added note to possibly related #8321. Will retry.

@dotnet-bot retest CentOS7.1 x64 Debug Build and Test

src/jit/flowgraph.cpp
```diff
@@ -21419,11 +21420,46 @@ void Compiler::fgAttachStructInlineeToAsg(GenTreePtr tree, GenTreePtr child, COR
...
+//    The code doesn't actually know if the corresponding inline
+//    succeeded or not; it relies on the fact that gtInlineCandidate
+//    initially points back at the call and is modified in place to
+//    the inlinee return expression if the inline is successful (see
+//    tail end of fgInsertInlineeBlocks for the update of iciCall).
+//
+//    If the parent of the GT_RET_EXPR is a virtual call,
+//    devirtualization is attempted. This should only succeed in the
+//    successful inline case, when the inlinee's return value
+//    expression provides a better type then the declared type of the
```
JosephTremoulet: typo: then -> than

src/jit/lclvars.cpp
```diff
@@ -247,10 +247,16 @@ void Compiler::lvaInitTypeRef()
         CORINFO_CLASS_HANDLE typeHnd;
         CorInfoTypeWithMod   corInfoType =
             info.compCompHnd->getArgType(&info.compMethodInfo->locals, localsSig, &typeHnd);
+
         lvaInitVarDsc(varDsc, varNum, strip(corInfoType), typeHnd, localsSig, &info.compMethodInfo->locals);
 
         varDsc->lvPinned  = ((corInfoType & CORINFO_TYPE_MOD_PINNED) != 0);
         varDsc->lvOnFrame = true; // The final home for this local variable might be our local stack frame
+
+        if (strip(corInfoType) == CORINFO_TYPE_CLASS)
```
JosephTremoulet: Does this get the same `CORINFO_CLASS_HANDLE` that the call to `getArgType` just above got and passed to `lvaInitVarDsc`?  Is so, maybe move this logic down into `lvaInitVarDsc`?

src/jit/lclvars.cpp
```diff
@@ -550,6 +558,11 @@ void Compiler::lvaInitUserArgs(InitVarDscInfo* varDscInfo)
 
         lvaInitVarDsc(varDsc, varDscInfo->varNum, strip(corInfoType), typeHnd, argLst, &info.compMethodInfo->args);
 
+        if (strip(corInfoType) == CORINFO_TYPE_CLASS)
```
JosephTremoulet: Same as above regarding moving this into `lvaInitVarDsc`

src/jit/importer.cpp
```diff
@@ -18393,4 +18399,366 @@ bool Compiler::IsMathIntrinsic(GenTreePtr tree)
...
+//     call -- the call node to examine/modify
+//     thisObj  -- the value of 'this' for the call
+//     callInfo -- info about the call from the VM
+//     exactContextHnd -- [OUT] updated context handle iff call devirtualized
+//
+// Notes:
+//     Virtual calls in IL will always "invoke" the base class method.
+//
+//     This transformation looks for evidence that the type of 'this'
+//     in the call is a final class or would invoke a final method, and
```
JosephTremoulet: nit: or is exact

src/jit/importer.cpp
```diff
@@ -18393,4 +18399,366 @@ bool Compiler::IsMathIntrinsic(GenTreePtr tree)
...
+//     if that and other safety checks pan out, modifies the call and
+//     the call info to create a direct call.
+//
+//     This transformation is done in the importer and not in some
```
JosephTremoulet: This paragraph reads like we only call this before inline, but the same method gets called for late devirtualization, right?

src/jit/importer.cpp
JosephTremoulet: Are you still planning to do this?

src/jit/importer.cpp
```diff
@@ -18393,4 +18399,366 @@ bool Compiler::IsMathIntrinsic(GenTreePtr tree)
...
+    // Make the updates.
+    call->gtFlags &= ~GTF_CALL_VIRT_VTABLE;
+    call->gtFlags &= ~GTF_CALL_VIRT_STUB;
+    call->gtCallMethHnd = derivedMethod;
+    call->gtCallType    = CT_USER_FUNC;
+
+    // Virtual calls include an implicit null check, which we may
+    // now need to make explicit. Not sure yet if we can restrict
+    // this to just a subset or need to do it for all cases, so
+    // will do it for all unless we know the object is not null.
```
JosephTremoulet: Supposing we could restrict this to a subset, would that be up in the code that's setting `nullCheck`, or down here?  If it would be down here, maybe better to invert the sense of the bool and name it `knownNotNull` or something rather than `nullCheck`.

src/jit/lclvars.cpp
AndyAyersMS: `getArgType` doesn't set the class handle for ref types. It is mainly used to establish the representation of the type and not the identity of a type.

src/jit/lclvars.cpp
AndyAyersMS: Same reason as above here.

src/jit/importer.cpp
AndyAyersMS: When I wrote this I was thinking there might be a locally identifiable subset of calls that would not need checks -- one that could be identified by just looking at the call and its properties. But I no longer believe this.

So, suppressing the check requires some sort of analysis of the 'obj' and doing anything nontrivial here seems risky since we are in the middle of importing.

The string case is trivial so we might as well keep it. I suppose we could also rely on `newobj` not returning null. 

Inverting the sense of the bool looks like a good suggestion since the common case is that we will add a null check.

src/jit/flowgraph.cpp
AndyAyersMS: Fixed and reworded some neighboring text....

AndyAyersMS:
Updated to address feedback. 

Added a 4.6 compat stub (needed for clean desktop builds).

Removed the distribution table showing which operators lead to type loss. The information is stale and arguably should live somewhere else.

src/jit/importer.cpp
AndyAyersMS: Fixed

AndyAyersMS:
Haven't hit any issues in desktop testing, though have only run a subset because the GUID/interface change makes getting the full test of tests running a challenge.

src/jit/importer.cpp
AndyAyersMS: Thinking of leaving things as they are now -- a few checks (eg the one for __Canon) were moved to the VM side, but a number remain on the jit side -- notably the attribute checks for final/sealed methods and classes.

Seemed logical to me to keep the policy decisions on the jit side and have the VM deal with correctness/servicing, which more or less dictates the current setup. Otherwise the jit has to pass in a bit more state (for instance, is the type known exactly) and get back more state (whether the target is known exactly).

Some of the jit-side checks will be removed in future work, eg interface call devirtualization.

AndyAyersMS:
Waiting for builds with #9869 to make their way over to CoreFx before merging this. Believe they should be in the beta25101-02 build of CoreCLR.

AndyAyersMS:
Actually, watching dotnet/corefx#16574

AndyAyersMS:
Local CoreFx changes with these changes look good.

briansull:
LGTM

AndyAyersMS:
Am going to push one more force update that will reorder things so the non-jit changes are back together in one commit and the jit changes in another.

PR #9231, opened 2017-01-31
Simplified MinOpts GC ref tracking.

pkukol:
Not ready for "official" review yet, but feel free to look and comment anytime ;).


pkukol:
@dotnet-bot test CentOS7.1 gcstress0xc_minopts_heapverify1
@dotnet-bot test CentOS7.1 minopts
@dotnet-bot test CentOS7.1 zapdisable
@dotnet-bot test OSX Checked r2r_jitminopts
@dotnet-bot test OSX gcstress0x3
@dotnet-bot test OSX gcstress0xc
@dotnet-bot test OSX gcstress0xc_minopts_heapverify1
@dotnet-bot test OSX minopts
@dotnet-bot test Ubuntu Checked r2r_jitminopts
@dotnet-bot test Ubuntu corefx_minopts
@dotnet-bot test Ubuntu gcstress0x3
@dotnet-bot test Ubuntu gcstress0xc
@dotnet-bot test Ubuntu gcstress0xc_jitstress1
@dotnet-bot test Ubuntu gcstress0xc_jitstress2
@dotnet-bot test Ubuntu gcstress0xc_minopts_heapverify1
@dotnet-bot test Windows_NT Checked r2r_jitminopts
@dotnet-bot test Windows_NT arm64 Checked gcstress0x3
@dotnet-bot test Windows_NT arm64 Checked gcstress0xc
@dotnet-bot test Windows_NT corefx_minopts
@dotnet-bot test Windows_NT gcstress0xc_minopts_heapverify1
@dotnet-bot test Windows_NT minopts
@dotnet-bot test Windows_NT x86 Checked gcstress0xc_minopts_heapverify1
@dotnet-bot test Windows_NT zapdisable

DrewScoggins:
test Windows_NT perf

DrewScoggins:
test Ubuntu14.04 perf

pkukol:
@dotnet-bot test OSX Checked r2r_jitminopts
@dotnet-bot test OSX gcstress0x3
@dotnet-bot test OSX gcstress0xc
@dotnet-bot test OSX gcstress0xc_minopts_heapverify1
@dotnet-bot test OSX minopts
@dotnet-bot test Ubuntu Checked r2r_jitminopts
@dotnet-bot test Ubuntu corefx_minopts
@dotnet-bot test Ubuntu gcstress0x3
@dotnet-bot test Ubuntu gcstress0xc
@dotnet-bot test Ubuntu gcstress0xc_jitstress1
@dotnet-bot test Ubuntu gcstress0xc_jitstress2
@dotnet-bot test Ubuntu gcstress0xc_minopts_heapverify1
@dotnet-bot test Windows_NT Checked r2r_jitminopts
@dotnet-bot test Windows_NT arm64 Checked gcstress0x3
@dotnet-bot test Windows_NT arm64 Checked gcstress0xc
@dotnet-bot test Windows_NT corefx_minopts
@dotnet-bot test Windows_NT gcstress0xc_minopts_heapverify1
@dotnet-bot test Windows_NT minopts
@dotnet-bot test Windows_NT x86 Checked gcstress0xc_minopts_heapverify1
@dotnet-bot test Windows_NT zapdisable

pkukol:
@dotnet-bot test Ubuntu Checked r2r_jitminopts
@dotnet-bot test Ubuntu gcstress0xc_minopts_heapverify1
@dotnet-bot test OSX Checked r2r_jitminopts

pkukol:
@dotnet-bot test Ubuntu Checked r2r_jitminopts
@dotnet-bot test Ubuntu corefx_minopts
@dotnet-bot test Ubuntu gcstress0x3
@dotnet-bot test Ubuntu gcstress0xc
@dotnet-bot test Ubuntu gcstress0xc_jitstress1
@dotnet-bot test Ubuntu gcstress0xc_jitstress2
@dotnet-bot test Ubuntu gcstress0xc_minopts_heapverify1

pkukol:
@dotnet-bot test OSX gcstress0x3
@dotnet-bot test OSX gcstress0xc
@dotnet-bot test OSX gcstress0xc_minopts_heapverify1

pkukol:
@dotnet-bot test OSX Checked r2r_jitminopts

pkukol:

pkukol:
@dotnet-bot test Ubuntu gcstress0xc_minopts_heapverify1

pkukol:
@dotnet-bot test CentOS7.1 gcstress0xc
@dotnet-bot test CentOS7.1 gcstress0xc_minopts_heapverify1
@dotnet-bot test CentOS7.1 minopts
@dotnet-bot test CentOS7.1 zapdisable

pkukol:

pkukol:
@swaroop-sridhar @jkotas @BruceForstall @sivarv PTAL
BTW, I went out of my way to minimize any impact on non-MinOpts paths etc, but this forced some duplication of code etc; if you have any suggestions as far as how to implant the new stuff please let me know - I'd love to share the blame I mean credit with others.

BruceForstall:
@pkukol Looks like there are a bunch of test failures. Do you still want it reviewed now?

pkukol:
@BruceForstall   Yes please, most of the tests appear to be spurious or are errors in the tests themselves; I am tracking two strange failures on desktop (one is COBOL and the other one is a complex physics simulation so it's proving to be a bit difficult) but other than that everything passes. Except for some flavors of Linux but I can't tell whether those are genuine failures or not. Still hoping to hear from someone as far as how to run tests on Linux with my build to close that loop. Sorry about that.

pkukol:
@dotnet-bot test Ubuntu gcstress0xc_minopts_heapverify1
@dotnet-bot test CentOS7.1 minopts
@dotnet-bot test CentOS7.1 gcstress0xc_minopts_heapverify1

pkukol:
@dotnet-bot test OSX minopts
@dotnet-bot test OSX gcstress0xc_minopts_heapverify1

pkukol:
@dotnet-bot test OSX gcstress0x3
@dotnet-bot test OSX gcstress0xc_minopts_heapverify1
@dotnet-bot test OSX gcstress0xc
@dotnet-bot test OSX Checked r2r_jitminopts
@dotnet-bot test OSX Checked r2r
@dotnet-bot test Ubuntu corefx_minopts
@dotnet-bot test Ubuntu Checked r2r_jitminopts
@dotnet-bot test Windows_NT corefx_minopts

pkukol:
@dotnet-bot test CentOS7.1 gcstress0xc_minopts_heapverify1
@dotnet-bot test Ubuntu gcstress0xc_jitstress1
@dotnet-bot test Ubuntu gcstress0xc_jitstress2

pkukol:
@dotnet-bot test Ubuntu gcstress0x3

tests/src/JIT/Methodical/Arrays/misc/arrres.cs
```diff
@@ -62,52 +62,52 @@ private static void Test2()
...
-            GC.WaitForPendingFinalizers();
-            GC.Collect();
         }
         [System.Runtime.CompilerServices.MethodImplAttribute(System.Runtime.CompilerServices.MethodImplOptions.NoInlining)]
         private static void Test3()
         {
+            GC.Collect();
+            GC.WaitForPendingFinalizers();
+            GC.Collect();
             for (int i = 0; i < 50; i++)
```
BruceForstall: Would it make more sense to create a new `DoCollect()` function that does these three lines, is marked NoInlining, and is called between the Test1()/Test2()/etc. calls in Main()?

src/jit/gcencode.cpp
```diff
@@ -3844,13 +3844,18 @@ struct InterruptibleRangeReporter
...
-                               unsigned       codeSize,
-                               unsigned       prologSize,
-                               MakeRegPtrMode mode)
+void GCInfo::gcMakeRegPtrTable(
+    GcInfoEncoder* gcInfoEncoder, unsigned codeSize, unsigned prologSize, MakeRegPtrMode mode, unsigned* callCntRef)
 {
     GCENCODER_WITH_LOGGING(gcInfoEncoderWithLog, gcInfoEncoder);
 
+#if defined(_TARGET_AMD64_) && defined(FEATURE_CORECLR)
+    const bool fastPath = (compiler->opts.MinOpts() && !compiler->opts.jitFlags->IsSet(JitFlags::JIT_FLAG_PREJIT) &&
```
BruceForstall: >fastPath [](start = 15, length = 8)

`fastPath` seems too generic a name. Why not `trackGcRefs` or `MinOptsTrackGcRefs`?

src/jit/gcencode.cpp
```diff
@@ -4204,9 +4221,26 @@ void GCInfo::gcMakeRegPtrTable(GcInfoEncoder* gcInfoEncoder,
...
-                    numCallSites++;
+                    // We have the call count from the previous run.
+                    numCallSites = *callCntRef;
+
+                    // If there are no calls, tell the world and bail.
+                    if (numCallSites == 0)
+                    {
+                        gcInfoEncoderWithLog->DefineCallSites(nullptr, nullptr, 0);
+                        return;
+                    }
```
BruceForstall: It's unfortunate and undesirable to add a "return" in the middle of this enormous method. Isn't there a way to restructure the code to avoid this, and let it fall through (without adding too much complexity in the way of additional nesting, etc.)?

src/jit/jitconfigvalues.h
```diff
@@ -218,6 +218,10 @@ CONFIG_INTEGER(JitEnableNoWayAssert, W("JitEnableNoWayAssert"), 0)
 CONFIG_INTEGER(JitEnableNoWayAssert, W("JitEnableNoWayAssert"), 1)
 #endif // !defined(DEBUG) && !defined(_DEBUG)
 
+#if defined(_TARGET_AMD64_) && defined(FEATURE_CORECLR)
+CONFIG_INTEGER(JitMinOptsTrackGCrefs, W("JitMinOptsTrackGCrefs"), 0) // Allow GC roots to be tracked w/MinOpts
```
BruceForstall: Why is this feature for AMD64 only? And why only CoreCLR? You can set the defaults for that, if you are worried about changing existing behavior. So, shouldn't it be:
```
#if !defined(JIT32_GCENCODER)
#if defined(_TARGET_AMD64_) && defined(FEATURE_CORECLR)
#define JitMinOptsTrackGCrefs_Default 0 // Not tracking GC refs in MinOpts is new behavior
#else
#define JitMinOptsTrackGCrefs_Default 1
#endif
CONFIG_INTEGER(JitMinOptsTrackGCrefs, W("JitMinOptsTrackGCrefs"), JitMinOptsTrackGCrefs_Default) // If non-zero, allow GC roots to be tracked w/MinOpts
#endif // !defined(JIT32_GCENCODER)
```

src/vm/gcinfodecoder.cpp
```diff
@@ -648,7 +651,16 @@ bool GcInfoDecoder::EnumerateLiveSlots(
         //
         if(!executionAborted)
         {
+#if defined(_TARGET_AMD64_) && defined(FEATURE_CORECLR)
+            if(m_NumInterruptibleRanges == 0)
+            {
+                // No ranges and no explicit safepoint - must be MinOpts.
+                noTrackedRefs = true;
+                goto NO_RANGES;
+            }
```
BruceForstall: Can you please figure out a way to avoid adding this `goto`?

src/jit/lclvars.cpp
```diff
@@ -2676,6 +2676,173 @@ int __cdecl Compiler::WtdRefCntCmp(const void* op1, const void* op2)
     return 0;
 }
 
+#if defined(_TARGET_AMD64_) && defined(FEATURE_CORECLR)
+
+/* static */
+int __cdecl Compiler::WtdRefCntCmpSimpleGC(const void* op1, const void* op2)
+{
```
BruceForstall: Please add a new standard-format function header.

You don't care about weighted ref counts in this case, right?, so why is it named "Wtd..."?

In fact, couldn't you assert:
```
assert(!dsc1->lvTracked);
assert(!dsc2->lvTracked);
```
?

So the only thing you are doing is sorting GC refs together?

src/jit/lclvars.cpp
```diff
@@ -2676,6 +2676,173 @@ int __cdecl Compiler::WtdRefCntCmp(const void* op1, const void* op2)
...
+        if (dsc1 > dsc2)
+        {
+            return +1;
+        }
+
+        return 0;
+    }
+
+    // TODO(REVIEW): if we're here it means we have MinOpts and "simplified
+    // GC tracking", so we may be able to prune the stuff that follows.
```
BruceForstall: Seems correct to me, if you add the assert I suggest.

BruceForstall: It seems like you should secondarily sort on weights, after GC-ness, before falling back to lclvar number to break ties.

src/jit/gcencode.cpp
pkukol: The main goal is to skip all the code below here because we know (from the first pass) that we'll find no calls that we'll need to record, but there are usually quite a few calls on the list (they just all get skipped twice in 99.9% of the cases). The only way to avoid wrapping everything below in yet another level of nesting would be to bash the call list to null but that seems extremely dirty (and may not work). If you have a specific idea, though, please let me know - I agree the method is pretty knarly as it is.

src/jit/lclvars.cpp
pkukol: Decided to ditch this thing, it's just too awkward and ugly (and a maintenance danger); thanks for flagging it.

src/vm/gcinfodecoder.cpp
pkukol: That's trivial to do (and I've tried it both ways) but when I asked earlier no feedback resulted so I left it alone; will do it by adding an 'if', then.

src/jit/jitconfigvalues.h
pkukol: Hmm, hold on - are you suggesting we ditch all the #ifdef's as well? That's not as obvious because even if it's normally disabled, it adds a bunch of code (mostly only 'if' statements but still) but more importantly in the GC decoder it weakens an assert that I think is quite important (I wanted to add an explicit bit so we don't have to assume we must have the new behavior but there's not obvious room for a new flag plus nobody else though it was worth a flag anyway). The suggestion to keep it out of the non-CORECLR build came from the VM folks so how about I leave it the way it is, since it's only been tested that way, and once it's proven its worth (which has yet to be verified independently) it can be enabled in more places?

tests/src/JIT/Methodical/Arrays/misc/arrres.cs
pkukol: I'm just making the minimal changes to make the test more correct (and leaving the basic structure intact when it's not impacted by the GC tracking changes); I have no idea why it was written this way and frankly I don't feel comfortable changing it without a specific reason and checking with whoever owns this. There are other tests similar to this one and my suggestion would be to change all of them separately (or leave them) but not conflate such a thing with this PR. Hope that makes sense but if not just let me know.

src/jit/gcencode.cpp
pkukol: It's actually the opposite so I picked **"noTrackedGCSlots* - hope that's OK.

pkukol:
@swaroop-sridhar @jkotas @BruceForstall @sivarv @pgavlin PTanotherL
I think I've addressed all the comments / suggestions; if you happen to spot something I've missed please let me know (and I do apologize).

src/jit/jitconfigvalues.h
BruceForstall: Yes, I was suggesting eliminating all the `#ifdefs`, so it would be possible to enable and test on desktop, or on non-AMD64 platforms, simply by setting a COMPlus variable. We have very few `#ifdef FEATURE_CORECLR` and I'd like to keep it that way.

src/jit/lclvars.cpp
```diff
@@ -2847,6 +2847,15 @@ void Compiler::lvaSortByRefCount()
             lvaSetVarDoNotEnregister(lclNum DEBUGARG(DNER_PinningRef));
 #endif
         }
+#if defined(_TARGET_AMD64_) && defined(FEATURE_CORECLR)
+        else if (opts.MinOpts() && !JitConfig.JitMinOptsTrackGCrefs())
+        {
+            if (varTypeIsGC(varDsc->TypeGet()))
+            {
```
BruceForstall: minor: you may want all the conditions on the "else if" line above. Otherwise, if someone adds yet another "else if" case below, then it wouldn't get hit for the minopts & !JitMinOptsTrackGCrefs() case if the var isn't GC.

pkukol:
@dotnet-bot test OSX minopts
@dotnet-bot test OSX Checked r2r_jitminopts
@dotnet-bot test CentOS7.1 minopts
@dotnet-bot test OSX Checked r2r_jitminopts
@dotnet-bot test Ubuntu Checked r2r_jitminopts
@dotnet-bot test Windows_NT minopts
@dotnet-bot test Windows_NT Checked r2r_jitminopts

src/jit/lclvars.cpp
pkukol: Good point - in my testing version there's actually another statement in there so the two if's need to be separate but in the "real" version it's simple and might as well combine the conditions. Good catch, will include it in my next push.

pkukol:
@dotnet-bot test CentOS7.1 minopts
@dotnet-bot test Windows_NT minopts

pkukol:
@dotnet-bot test CentOS7.1 minopts
@dotnet-bot test Windows_NT minopts

pkukol:
@dotnet-bot test CentOS7.1 minopts
@dotnet-bot test Windows_NT minopts

pkukol:
@dotnet-bot test OSX minopts
@dotnet-bot test OSX Checked r2r_jitminopts
@dotnet-bot test Ubuntu Checked r2r_jitminopts
@dotnet-bot test Windows_NT Checked r2r_jitminopts

src/jit/jitconfigvalues.h
pkukol: Missed your second comment until now; sorry about that. OK, I will change this as you suggest, but it still leaves the issue of the assert in vm\gcinfodecode.cpp - assuming it can see the above flag (the whole CLR vs. JIT config flags thing is still a mystery to me), it will need to test the flag when decoding GC info (at runtime) or we just weaken the decoding assert for every config / all the time? Neither approach seems ideal to me, but there's probably a way to do this better that I'm not seeing; please bear with me.

And, just to make absolutely sure - with this newest change, the new code will always be compiled in and we'll use a runtime test of (pseudo-code) config.JitMinOptsTrackGCrefs in the various spots in the *.cpp files instead of switching the code out completely (using the current "#ifdef AMD64 && blah blah"); correct?

pkukol:
@dotnet-bot test CentOS7.1 minopts
@dotnet-bot test Windows_NT minopts

src/jit/jitconfigvalues.h
BruceForstall: clrconfigvalues.h is for things read by the VM; jitconfigvalues.h is for things read by the VM. Note that clrconfigvalues.h does duplicate many things, *if* those things are used by JIT32/JIT64, which does not have jitconfigvalues.h and that mechanism. That is usually what causes confusion.

Yes, there will be code dynamically executed because of this, but presumably it will be very small/cheap.

pkukol: Should be done now - @BruceForstall kindly TAL.

pkukol:
@dotnet-bot test CentOS7.1 minopts
@dotnet-bot test Windows_NT minopts

BruceForstall:
LGTM. Please squash commits before merging.

pkukol:
@dotnet-bot test CentOS7.1 minopts

pkukol:
@dotnet-bot test OSX Checked r2r_jitminopts
@dotnet-bot test Ubuntu Checked r2r_jitminopts
@dotnet-bot test Windows_NT Checked r2r_jitminopts

PR #9233, opened 2017-01-31
Disable delete_next_card_table

adiaaida:
Since 163983e was checked in, various tests have been failing because
delete_next_card_table is meant to allocate memory until it OOMs. Disable
this test to get the lab in better shape until we can deal with the GC bug
in #9064, then move it to the long gc test leg.


adiaaida:
@swgillespie PTAL

swgillespie:
LGTM, thanks!

PR #9244, opened 2017-02-01
Synchronize SpinLock with CoreRT

jkotas:


src/mscorlib/corefx/SR.cs
```diff
@@ -595,4 +595,29 @@ internal static string ArgumentException_ValueTupleLastArgumentNotATuple
     {
         get { return Environment.GetResourceString("ArgumentException_ValueTupleLastArgumentNotATuple"); }
     }
+
+    internal static string SpinLock_TryEnter_ArgumentOutOfRange
```
stephentoub: Would there be any benefit in preventing these from getting inlined?

jkotas: The inliner should be smart enough to not inline these if it can see that they are called on cold exception throwing paths. I believe that Andy has done some changes recently to make it work even better than before.

jkotas:

PR #9245, opened 2017-02-01
[x86/Linux] Create minimal frame for funclets

parjong:
This commit makes each funclet to have minimal frame.

### DIFF
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -10281,7 +10281,12 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
 
     compiler->unwindBegProlog();
 
-    // TODO Save callee-saved registers
+    // Save frame pointer
+    inst_RV(INS_push, REG_FPBASE, TYP_REF);
+    compiler->unwindPush(REG_FPBASE);
+
+    // Save callee-saved registers
+    genPushCalleeSavedRegisters();
 
     // This is the end of the OS-reported prolog for purposes of unwinding
     compiler->unwindEndProlog();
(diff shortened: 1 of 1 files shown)


parjong:
\CC @seanshpark @wateret

parjong:

parjong:
@dotnet/jit-contrib Please take a look.

parjong:
@jkotas @BruceForstall Please take a look.

jkotas:
Is this needed to make something work, or are you just fixing TODO? 

We are not consistent between platforms ... but the preferred funclet convention is that the funclet does not need to save any calllee saved registers; and that the exception handling subsystem restores them to what they were when the control left the function originally.

parjong:
I currently have no issue related with callee-saved registers. I just want to fix TODO before starting PSPSym-related fixes. 

For that purpose, it seems that EBP should be stored/restored, but I'm not sure about other callee saved registers. I just mimic the corresponding code in AMD64.

BruceForstall:
It would be useful to have a comment like the one above the amd64 version of `CodeGen::genFuncletProlog()` that shows the calling convention and stack layout for linux/x86.

You need to determine what the register convention is for the EH subsystem, as @jkotas says. This will affect register liveness into and out of funclets. Currently, we kill all registers, and local vars need to get reloaded from the stack. We use the PSPSym because we can't trust that the frame pointer is re-established to what the parent frame had. If the EH subsystem can guarantee it gets re-established, you wouldn't need that.

parjong:
@jkotas @BruceForstall Thank you for comments. It would be better to consider this issue when we consider PSPSym-related issues. I'll create another PR when we are ready later.

PR #9247, opened 2017-02-01
Fix GC reporting for Span<T> passed in registers on Unix

kouvel:
Part of fix for #9033

Test coverage is already there in CoreFX's span tests.


kouvel:
@jkotas

PR #9249, opened 2017-02-01
[Wip] ASCII fastpath, System.Text cleanup

benaadams:
Need to change how the workhorse+fallback functions are done

Current jitdiff with the fastpath; should drop more when doing fallback correctly as suggested by @jkotas in https://github.com/dotnet/coreclr/pull/9187#discussion_r98521142
```
Summary:
(Note: Lower is better)

Total bytes of diff: -1771 (-0.05 % of base)
    diff is an improvement.

Total byte diff includes 5517 bytes from reconciling methods
        Diff had   42 unique methods,     5517 unique bytes

       -1771 : System.Private.CoreLib.dasm (-0.05 % of base)

1 total files with size differences.

        1327 : System.Private.CoreLib.dasm - ASCIIEncoding:GetBytesFallback(long,int,long,int,ref):int:this
         704 : System.Private.CoreLib.dasm - ASCIIEncoding:GetBytes(ref,int,int,ref,int):int:this (2 methods)
         366 : System.Private.CoreLib.dasm - ASCIIEncoding:TryEncode(long,int,long,int,byref,byref):bool
         310 : System.Private.CoreLib.dasm - EncodingForwarder:GetValidationFailedException(ref,int,int,ref):ref
         297 : System.Private.CoreLib.dasm - EncodingForwarder:GetValidationFailedException(ref,int,int):ref
         271 : System.Private.CoreLib.dasm - EncodingForwarder:GetValidationFailedException(ref):ref
         244 : System.Private.CoreLib.dasm - EncodingForwarder:GetValidationFailedException(ref,ref,int,int,ref):ref
         236 : System.Private.CoreLib.dasm - ASCIIEncoding:GetBytesValidated(long,int):ref:this
         234 : System.Private.CoreLib.dasm - ASCIIEncoding:GetBytes(long,int,long,int):int:this
         209 : System.Private.CoreLib.dasm - EncodingForwarder:GetValidationFailedException(ref,int,int,ref,int,int):ref

        -985 : System.Private.CoreLib.dasm - ASCIIEncoding:GetBytes(long,int,long,int,ref):int:this
        -634 : System.Private.CoreLib.dasm - EncodingForwarder:GetBytes(ref,ref,int,int,ref,int):int (2 methods)
        -352 : System.Private.CoreLib.dasm - Decoder:Convert(ref,int,int,ref,int,int,bool,byref,byref,byref):this
        -352 : System.Private.CoreLib.dasm - Encoder:Convert(ref,int,int,ref,int,int,bool,byref,byref,byref):this
        -350 : System.Private.CoreLib.dasm - DecoderNLS:Convert(ref,int,int,ref,int,int,bool,byref,byref,byref):this
        -293 : System.Private.CoreLib.dasm - EncodingForwarder:GetChars(ref,ref,int,int,ref,int):int
        -283 : System.Private.CoreLib.dasm - DecoderNLS:GetChars(ref,int,int,ref,int,bool):int:this
        -266 : System.Private.CoreLib.dasm - EncodingForwarder:GetString(ref,ref,int,int):ref
        -260 : System.Private.CoreLib.dasm - Encoding:GetBytes(ref,int,int):ref:this (2 methods)
        -259 : System.Private.CoreLib.dasm - Encoding:GetByteCount(ref,int,int):int:this

132 total methods with size differences.
```

### DIFF
--- a/src/mscorlib/src/System/Text/ASCIIEncoding.cs
+++ b/src/mscorlib/src/System/Text/ASCIIEncoding.cs
@@ -5,10 +5,8 @@
 namespace System.Text
 {
     using System;
-    using System.Runtime.Serialization;
-    using System.Security.Permissions;
-    using System.Diagnostics;
     using System.Diagnostics.Contracts;
+    using System.Runtime.InteropServices;
 
     // ASCIIEncoding
     //
@@ -22,12 +20,14 @@ namespace System.Text
     //
 
     [Serializable]
-[System.Runtime.InteropServices.ComVisible(true)]
+    [ComVisible(true)]
     public class ASCIIEncoding : Encoding
     {
+        // Allow for devirtualization (see https://github.com/dotnet/coreclr/issues/1166#issuecomment-276251559)
+        internal sealed class ASCIIEncodingSealed : ASCIIEncoding { }
         // Used by Encoding.ASCII for lazy initialization
         // The initialization code will not be run until a static member of the class is referenced
-        internal static readonly ASCIIEncoding s_default = new ASCIIEncoding();
+        internal static readonly ASCIIEncodingSealed s_default = new ASCIIEncodingSealed();
 
         public ASCIIEncoding() : base(Encoding.CodePageASCII)
         {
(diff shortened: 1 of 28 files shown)


src/mscorlib/src/System/Text/UTF7Encoding.cs
```diff
@@ -170,20 +168,20 @@ public override int GetByteCount(char[] chars, int index, int count)
...
-        [System.Runtime.InteropServices.ComVisible(false)]
+        [ComVisible(false)]
         public override int GetByteCount(String s)
         {
             return EncodingForwarder.GetByteCount(this, s);
         }
 
         [CLSCompliant(false)]
-        [System.Runtime.InteropServices.ComVisible(false)]
+        [ComVisible(false)]
```
jkotas: These ComVisible annotations should rather be deleted (ideally in separate PR). Similar to what was done in https://github.com/dotnet/corefx/pull/782 for CoreFX.

src/mscorlib/src/System/Text/ASCIIEncoding.cs
```diff
@@ -714,16 +1119,15 @@ public override int GetMaxByteCount(int charCount)
...
             return (int)byteCount;
         }
 
-
         public override int GetMaxCharCount(int byteCount)
         {
             if (byteCount < 0)
-               throw new ArgumentOutOfRangeException(nameof(byteCount),
-                    Environment.GetResourceString("ArgumentOutOfRange_NeedNonNegNum"));
+                ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.byteCount, ExceptionResource.ArgumentOutOfRange_NeedNonNegNum);
```
jkotas: There ThrowHelper optimizations should be done only in generic code where it really matters. I do think it really matters here.

src/mscorlib/src/System/Text/DecoderBestFitFallback.cs
```diff
@@ -34,13 +31,7 @@ public override DecoderFallbackBuffer CreateFallbackBuffer()
...
 
         // Maximum number of characters that this instance of this fallback could return
-        public override int MaxCharCount
-        {
-            get
-            {
-                return 1;
-            }
-        }
+        public override int MaxCharCount => 1;
```
jkotas: We want to get the sources of types like this unified between CoreCLR and CoreRT (https://github.com/dotnet/coreclr/issues/972). There is one set of cleanups applied here and different set of cleanups applied in CoreRT. I would be useful to get the code unified first, and apply the cleanups only after that (or do the cleanups in CoreRT and then copy the cleaned up version over to CoreCLR).

benaadams: System.Text is far prettier in CoreRT; can they be swapped wholesale or is it a partial overlap?

jkotas: It cannot be just copy&paste unfortunately. CoreRT version is missing the .NET Standard 2.0 work; and the CoreCLR version is missing most of the cleanup.

@danmosemsft Do you plan to have somebody start the work to unify the two?

danmosemsft: I'm working on removing all the code that BCLRewriter strips, and that's partly complete. I figured that would make the unification work easier since just about all of that should also be missing from CoreRT.

I expect to have someone work on the unification soon after I've done that so that UWP work can proceed.

PR #9251, opened 2017-02-01
Synchronize src\mscorlib\corefx with CoreRT

jkotas:


src/mscorlib/corefx/System/Globalization/CultureNotFoundException.cs
```diff
@@ -66,14 +65,8 @@ protected CultureNotFoundException(SerializationInfo info, StreamingContext cont
             _invalidCultureName = (string)info.GetValue("InvalidCultureName", typeof(string));
         }
 
-        [System.Security.SecurityCritical]  // auto-generated_required
         public override void GetObjectData(SerializationInfo info, StreamingContext context)
         {
-            if (info == null)
-            {
-                throw new ArgumentNullException(nameof(info));
-            }
```
stephentoub: Is this removal intentional?

jkotas: Yes - the base class does the validation already. Some exceptions types used to have the redundant validation here, some did not. I have done pass in CoreRT to strip it from all of them.

stephentoub: Cool

PR #9252, opened 2017-02-01
Use full set of runtimes for all test project.json files

wtgodbe:
Tests in Helix were failing for Ubuntu16.04, Ubuntu16.10, and Fedora23, because we weren't pulling down packages for those platforms - meaning that the Core_Root that was built from packages didn't have a large set of the test dependencies for those platforms. This should fix that error

### DIFF
--- a/tests/src/Common/test_runtime/project.json
+++ b/tests/src/Common/test_runtime/project.json
@@ -1,5 +1,6 @@
 {
   "dependencies": {
+    "Microsoft.NETCore.Platforms": "2.0.0-beta-25001-02",
     "Microsoft.DotNet.CoreCLR.TestDependencies": "1.0.0-prerelease",
     "jit-dasm": "0.0.1.4",
     "cijobs": "0.0.1.2",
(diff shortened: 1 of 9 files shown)


wtgodbe:
CC @MattGal @ramarag @gkhanna79

gkhanna79:
PTAL at the CI failures.

wtgodbe:
Looks to be working now, but will wait for CI to be fully green to confirm - thanks @dagood for helping me diagnose the issue.

tests/src/Common/test_runtime/project.json
```diff
@@ -1,5 +1,6 @@
 {
   "dependencies": {
+    "Microsoft.NETCore.Platforms": "2.0.0-beta-25001-02",
```
ramarag: is this really necessary ? Can't it be in test_dependencies/project.json ?
This json was meant to be only for native assets [ though quite a few managed assets have creeped in]

wtgodbe: We already have that dependency there - the issue (as @dagood pointed out) is that one of the dependencies in this file (and TestWrappersConfig's project.json) is bringing in an older version of Microsoft.NETCore.Platforms, that didn't support Ubuntu16.10/ Ubuntu16.04/ Fedora23.

ramarag:
@wtgodbe please open up a bug to get the root cause, as Microsoft.NETCore.Platforms already exists in test_dependecy\project.json and some other package seems to be overriding it 

LGTM otherwise

PR #8866, opened 2017-01-10
[Linux/ARM] Fix cross-architecture component build error: disable unused code

hseok-oh:
Fix debug & checked build error
- Disable TakeExceptionPointersOwnership function not used in Linux cross-architecture

Modify build.sh
- Remove duplicate test build in cross-architecture build

PR #9262, opened 2017-02-02
[ARM32/Linux] Fix definition and CMake for cross-architecture build

hseok-oh:
- Fix ABI flag based on target
- Exclude src/corefx for cross-architecture component


hseok-oh:
related: https://github.com/dotnet/coreclr/pull/8866#issuecomment-276843728

hseok-oh:
@janvorli PTAL

src/corefx/System.Globalization.Native/CMakeLists.txt
```diff
@@ -115,5 +115,5 @@ verify_dependencies(
 
 # add the install targets
 install_clr(System.Globalization.Native)
-install(TARGETS System.Globalization.Native_Static DESTINATION .)
+install_clr(System.Globalization.Native_Static)
```
janvorli: This change doesn't seem right. The install_clr strips debug symbols and installs PGO related file. None of this applies to static libraries. Moreover, neither the shared nor the static version of the System.Globalization.Native is needed for the cross components, so I think the whole src/corefx should be excluded for cross components build.

hseok-oh: @janvorli I agree. I'll fix to exclude src/corefx for cross component.

hseok-oh:
`install_clr` and `_install` exclude installation for cross component build, so I change `install_clr` to `_install` for static `System.Globalization.Native` and keep `install_clr` for shared version.

janvorli:
@hseok-oh but why do we build the System.Globalization.Native.so at all in the cross components build? I was thinking that you would exclude it from building by excluding `add_subdirectory(src/corefx)` in the root CMakeLists.txt for cross components build. Is there a problem with that?

hseok-oh:
@janvorli You're right. It's better to exclude it in root CMakeLists.txt.

hseok-oh:
@janvorli Is there anything to do more?

PR #9263, opened 2017-02-02
Refactor coreclr packaging

chcosta:
https://github.com/dotnet/core-eng/issues/21

This aligns how coreclr does packaging more closely with recent changes in corefx packaging which also enables building rid split packages on unsupported distros.

@ericstj, can you review this?  One of the concepts I introduced was `SupportedPackageOSGroups`.  This was required because the Microsoft.NETCore.Native package only builds on *nix platforms, and I needed some way to express which platforms were supported.  Without this concept (or something similar), we would produce a different lineup package when building on Windows for that package than we would on the *nix OS's and that seems problematic to me.  Another angle would be to make this an exclusion list.  I see some benefit to taking the exclusion list approach in case we build on a new os, but if we do build on a new os, I think there's other work that will need to be done to enable the new platform as well, so being explicit about which platforms we build on isn't terrible.

I had to annotate ProjectReferences in the lineup package with an `IsLineupPackage=false` "AdditionalProperties" metadata item.  This was to prevent that project reference from in turn pulling in the rid packages for the reference and adding them to the runtime.json file.  We might want to have some other way to denote this, and it would be nice to move it into buildtools so that it just happens by default, but I didn't want to block on that.

@ellismg, I pushed my changes from the source-build submodule and it looks like I pulled in one of your commits as well, let me know if you want me to exclude that change.  https://github.com/dotnet/coreclr/commit/13593513cbabfc4cbbb50e62bedbe0ce6f93488a

/cc @ellismg @gkhanna79 @ericstj


src/.nuget/Microsoft.NETCore.ILAsm/runtime.Linux.Microsoft.NETCore.ILAsm.props
```diff
@@ -1,11 +1,9 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <PropertyGroup>
-    <SkipPackageFileCheck>true</SkipPackageFileCheck>
-    <PackageTargetRuntime>linux-$(PackagePlatform)</PackageTargetRuntime>
-    <!-- only build for x64 -->
     <PackagePlatforms>x64;</PackagePlatforms>
+    <PackagePlatforms Condition="'$(RuntimeOS)' == 'ubuntu.14.04'">x64;arm;</PackagePlatforms>
```
chcosta: @ericstj It's not clear to me if these types of exceptions are needed,  I figured I'd pick your brain about this and the armel arch builds as well.

chcosta:
@dotnet-bot test this please

build.sh
```diff
@@ -304,42 +304,46 @@ isMSBuildOnNETCoreSupported()
...
-                    ;;
-                "ubuntu.16.10-x64")
-                    __isMSBuildOnNETCoreSupported=1
-                    ;;
-                "alpine.3.4.3-x64")
-                    __isMSBuildOnNETCoreSupported=1
-                    ;;
-                *)
-            esac
+            if [ "$__msbuildonunsupportedplatform" == "1" ]; then
```
ellismg: To save the indentation, you could consider just doing this the `*)` clause of the case (e.g set  __isMSBuildOnNETCoreSupported to __msbuildonunsupportedplatform)

dir.traversal.targets
```diff
@@ -18,4 +18,5 @@
 
   <Target Name="Rebuild" DependsOnTargets="Clean;Build" />
 
+  <Target Name="Dump" />
```
ellismg: Did you intend to remove this.

src/.nuget/Microsoft.NETCore.ILAsm/runtime.Linux.Microsoft.NETCore.ILAsm.props
ericstj: I think we can remove PackagePlatforms.  This was an older version of filtering where we dummy-ed out the BuildTarget when PackagePlatforms didn't contain the build Platform.  We no longer use this in CoreFx, we should remove it here as well.  We're doing more advanced filtering in the builds file.

src/.nuget/Microsoft.NETCore.ILAsm/Microsoft.NETCore.ILAsm.builds
```diff
@@ -1,89 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), packaging.props))\packaging.props" />
```
ericstj: Why the different convention?  Couldn't you have named these dir.props and dir.targets in src/.nuget?

src/.nuget/packaging.props
```diff
@@ -0,0 +1,150 @@
...
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+
+  <PropertyGroup>
+    <!-- Distro rid is passed as runtimeos-arch-->
+    <_distroRidIndex>$(__DistroRid.IndexOfAny("-"))</_distroRidIndex>
+    <_archRidIndex>$([MSBuild]::Add($(_distroRidIndex), 1))</_archRidIndex>
+    <OSRid Condition="'$(OSRid)' == '' and '$(_distroRidIndex)' != '-1'">$(__DistroRid.SubString(0, $(_distroRidIndex)))</OSRid>
+    <OSRid Condition="'$(OSRid)' == ''">win10</OSRid>
+    <ArchGroup Condition="'$(ArchGroup)' == '' and '$(_archRidIndex)' != '0'">$(__DistroRid.SubString($(_archRidIndex)))</ArchGroup>
+  </PropertyGroup>
```
ericstj: Combine these two groups.

src/.nuget/Microsoft.NETCore.Native/Microsoft.NETCore.Native.pkgproj
```diff
@@ -1,50 +1,16 @@
...
-    </ProjectReference>
-    <ProjectReference Include="alpine\3.4.3\Microsoft.NETCore.Native.pkgproj">
-      <Platform>amd64</Platform>
-    </ProjectReference>
-    <ProjectReference Include="osx\Microsoft.NETCore.Native.pkgproj">
-      <Platform>amd64</Platform>
-    </ProjectReference>
-  </ItemGroup>
+
+  <Import Condition="'$(_packageTargetOSGroup)' != ''" Project="$(MSBuildThisFileDirectory)runtime.$(_packageTargetOSGroup).$(MSBuildProjectName).props" />
```
ericstj: Why import here rather than in packaging.props (dir.props)

src/.nuget/packaging.targets
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.traversal.targets))\dir.traversal.targets" />
+
+  <Target Name="FilterProjects" BeforeTargets="Build">
+    <Error Condition="'$(PackageRID)' == ''" Text="'PackageRID' property must be specified."/>
+
+    <!-- Only build packages for current RID -->
```
ericstj: Current RID or non-RID-specific.

src/.nuget/Microsoft.NETCore.ILAsm/Microsoft.NETCore.ILAsm.builds
chcosta: I didn't want to cause a conflict with Microsoft.TargetingPackage.Private.CoreClr, which doesn't build using these props / targets.  Also, there are enough places where we don't want to import packaging.targets, that we would have to substitute `..\..\..\dir.targets` instead, that it seemed ugly and I didn't want to do it.

src/.nuget/Microsoft.NETCore.ILAsm/runtime.Linux.Microsoft.NETCore.ILAsm.props
chcosta: I don't think I can remove PackagePlatforms yet, because the filtering still exists.  That's a slightly more complex fix to remove PackagePlatforms.  I'll leave this as-is for now.

src/.nuget/Microsoft.NETCore.Native/Microsoft.NETCore.Native.pkgproj
chcosta: Not every project has a `runtime.$(_packageTargetOSGroup).$(MSBuildProjectName).props` file.  I could also condition on that existing, but then I was worried that we would hide a valid missing file if we're bringing up a new osgroup support or something.

src/.nuget/Microsoft.NETCore.ILAsm/Microsoft.NETCore.ILAsm.pkgproj
```diff
@@ -16,5 +16,5 @@
   
   <Import Condition="'$(_packageTargetOSGroup)' != ''" Project="$(MSBuildThisFileDirectory)runtime.$(_packageTargetOSGroup).$(MSBuildProjectName).props" />
 
-  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
+  <Import Project="$(MSBuildThisFileDirectory)..\..\..\dir.targets" />
```
ericstj: Why not GetDirectoryNameOfFileAbove?

src/.nuget/Microsoft.NETCore.ILDAsm/Microsoft.NETCore.ILDAsm.pkgproj
```diff
@@ -15,5 +15,5 @@
 
   <Import Condition="'$(_packageTargetOSGroup)' != ''" Project="$(MSBuildThisFileDirectory)runtime.$(_packageTargetOSGroup).$(MSBuildProjectName).props" />
 
-  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
+  <Import Project="$(MSBuildThisFileDirectory)..\..\..\dir.targets" />
```
ericstj: Why not GetDirectoryNameOfFileAbove?

ericstj: Same goes for other instances of the same thing.

src/.nuget/Microsoft.NETCore.ILAsm/runtime.Linux.Microsoft.NETCore.ILAsm.props
ericstj: You should be able to due to this:https://github.com/dotnet/buildtools/blob/08573119630248c2ee1b636ceaf3ca8fc8a2ecda/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets#L129-L130

src/.nuget/Microsoft.NETCore.ILAsm/Microsoft.NETCore.ILAsm.builds
ericstj: If you want to have different modes for packages I think it would be clearer to describe those modes with properties rather than importing different targets with names that don't correspond to how they are different.

src/.nuget/Microsoft.NETCore.ILAsm/runtime.Linux.Microsoft.NETCore.ILAsm.props
chcosta: If we remove `PackagePlatforms`, then we stop producing lineup packages because those packages don't have a `PackageTargetRuntime` .

ericstj: Then that means that CoreCLR is not setting `PackagePlatform` correctly.

src/.nuget/Microsoft.NETCore.ILAsm/Microsoft.NETCore.ILAsm.pkgproj
chcosta: we don't want `..\..\dir.targets`, we want `..\..\..\dir.targets`.  This was part of the reason I had named `..\..\dir.targets` as `..\..\packaging.targets`.

src/.nuget/Microsoft.NETCore.ILAsm/runtime.Linux.Microsoft.NETCore.ILAsm.props
chcosta: I've pushed a change to remove PackagePlatforms

src/.nuget/Microsoft.NETCore.ILDAsm/runtime.OSX.Microsoft.NETCore.ILDAsm.props
```diff
@@ -19,11 +12,10 @@
     <AdditionalLibPackageExcludes Include="%2A%2A\%2A.dwarf" />
     <AdditionalSymbolPackageExcludes Include="%2A%2A\%2A.dylib" />
     <AdditionalSymbolPackageExcludes Include="%2A%2A\%2A.dll" />
-    <ArchitectureSpecificNativeSymbol Include="..\..\_.pdb" />
+    <ArchitectureSpecificNativeSymbol Include="..\_.pdb" />
```
ericstj: Nit: we should have a property for this sort of thing.  I think we did in corefx.  That can be addressed separately.  /cc @dagood

src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/Microsoft.NETCore.Runtime.CoreCLR.pkgproj
```diff
@@ -1,75 +1,18 @@
...
+  <ItemGroup Condition="'$(IsLineupPackage)' == 'true'">
+    <ProjectReference Include="..\Microsoft.NETCore.Jit\Microsoft.NETCore.Jit.pkgproj">
+      <AdditionalProperties>%(ProjectReference.AdditionalProperties);IsLineupPackage=false</AdditionalProperties>
     </ProjectReference>
   </ItemGroup>
-  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
+
+  <Import Condition="'$(_packageTargetOSGroup)' != ''"  Project="$(MSBuildThisFileDirectory)runtime.$(_packageTargetOSGroup).$(MSBuildProjectName).props" />
+  
+  <Import Project="$(MSBuildThisFileDirectory)..\..\..\dir.targets" />
```
ericstj: Did you look into using some sort of property to differentiate behavior, rather than importing different targets?  I really think this type of difference is confusing, especially with no comments.  We should have a consistent pattern followed by all projects.

chcosta: I renamed `src\.nuget\dir.targets` to `src\.nuget\dir.traversal.targets`, which aligns closer with its functionality and allows us to use the `GetDirectoryNameOfFileAbove` syntax to import either `dir.targets` or `dir.traversal.targets`.  This is similar to how imports were handled in CoreFx for pkgproj vs builds imports.

PR #9264, opened 2017-02-02
[x86/Linux] Set Establisher Frame Pointer as Caller SP

parjong:
This commit revises x86/Linux unwinder to set Establisher Frame Pointer as Caller SP (instead of Initial SP).

This commit also updates CLR ABI document accordingly (which is missing before).


parjong:
\CC @seanshpark @wateret

parjong:
@jkotas Please take a look.

Issue #4651, opened 2016-04-28
[corelib] Remove unused types/members


jamesqo:
Ran into this issue while trying to create perf benchmarks for #4540. It looks like the types in the decompiled output of mscorlib:

![mscorlib-io-types](https://cloud.githubusercontent.com/assets/9159214/14900506/d602cec8-0d5e-11e6-8734-f1d4cfacc055.PNG)

are missing a lot of types from what's actually [there in the sources](https://github.com/dotnet/coreclr/tree/master/src/mscorlib/src/System/IO). Can those missing types be removed without causing a build break with the full framework? (Their primary implementations seem to live in corefx, so if I'm not mistaken this should be no different than removing `Console`, right?)

List of some types that don't appear in the decompiled output:
- Directory
- DirectoryInfo
- DriveInfo
- File
- FileAttributes
- FileSystemEnumerable
- FileSystemInfo
- ReadLinesIterator
- SearchOption

cc @jkotas 

**edit:** It also looks like `FileStream` (and probably many other of the `File*` types) could be removed as well (even though they appear in the decompiled output), if the `StreamWriter`, `StreamReader`, and `ResourceReader` constructors which accept a path string were removed (which aren't exposed in the contract), since they're the only ones that depend on `FileStream`. See this snapshot:

![mscorlib-fs](https://cloud.githubusercontent.com/assets/9159214/14900956/1cf68f70-0d61-11e6-95c6-c8f6c042cce4.PNG)

jkotas:
We have older CoreCLR workloads that we still need to support, and that depend on these types being in mscorlib.  We can remove these only once it is cleaned up (no timeline yet).

jkotas:
cc @AlexGhiondea @weshaggard

danmosemsft:
@jkotas what are these older CoreCLR workloads? Is this no longer the case now we have given up on project K mirroring?

jkotas:
Yes, number of these unused types can be deleted now. It is step 2 mentioned in https://github.com/dotnet/coreclr/issues/972.

https://github.com/dotnet/coreclr/issues/8732 or https://github.com/dotnet/corefx/issues/12678 overlaps with this as well.

jamesqo:
@jkotas I will grab this issue.

jamesqo:
[...]

PR #9269, opened 2017-02-02
Remove stripped code

danmosemsft:
Contributes to https://github.com/dotnet/coreclr/issues/4651

This applies all the changes that BCLRewriter does to our actual code, with the aim of making it possible to remove the rewriting step. I did this mostly with a Roslyn rewriter but some hand edits as well. Not ready for detailed review but perhaps someone can glance and see I'm not on the wrong track.

- [x] do the delete
- [x] diff x64 windows debug against regular rewrite output using decompiler and using apireviewer
- [x] diff unix, x86, and release binaries
- [x] review the changes, delete empty files
- [x] ideally, check the tests pass .. although the diff should be enough, this is a big change


danmosemsft:
There are some diffs that I can't avoid, principally the rewriter removes compiler generated default ctor from internal classes, so this is now present (eg., LocalDataStore()). Rewriter also adds [CompilerGenerated] on some private fields. Rewriter also inlines some constants, which I avoided to keep the code clear.

danmosemsft:
@jkotas

src/mscorlib/corefx/SR.cs
```diff
@@ -4,113 +4,14 @@
 // CoreFX creates SR in the System namespace. While putting the CoreCLR SR adapter in the root
 // may be unconventional, it allows us to keep the shared code identical.
 
-internal static class SR
+internal static class SR 
 {
-    public static string Arg_ArrayZeroError
```
jkotas: This will cause build break on Unix

src/mscorlib/src/System/Diagnostics/Eventing/EventSource.cs
```diff
@@ -5368,19 +5322,6 @@ public static void DisableFilter(ref ActivityFilter filterList, EventSource sour
...
-                if (af.m_providerGuid == source.Guid && af.m_samplingFreq != -1)
-                    return af;
-            }
-            return null;
-        }
-
-        /// <summary>
         /// Returns a session mask representing all sessions in which the activity 
         /// associated with the current thread is allowed  through the activity filter. 
         /// If 'triggeringEvent' is true the event MAY be a triggering event. Ideally
```
jkotas: Folks would prefer the EventSource to be left intact.

src/mscorlib/corefx/System/Threading/ClrThreadPoolBoundHandle.cs
```diff
@@ -15,7 +15,7 @@ namespace System.Threading
     ///     Represents an I/O handle that is bound to the system thread pool and enables low-level
     ///     components to receive notifications for asynchronous I/O operations.
     /// </summary>
-    public sealed partial class ThreadPoolBoundHandle : IDisposable
```
jkotas: This should stay as is - this type was moved from corefx, but the redundant copy in corefx was not removed yet. I have pending change to clean it up.

src/mscorlib/corefx/System/Threading/ClrThreadPoolPreAllocatedOverlapped.cs
```diff
@@ -8,7 +8,7 @@ namespace System.Threading
     /// Represents pre-allocated state for native overlapped I/O operations.
     /// </summary>
     /// <seealso cref="ThreadPoolBoundHandle.AllocateNativeOverlapped(PreAllocatedOverlapped)"/>
-    public sealed class PreAllocatedOverlapped : IDisposable, IDeferredDisposable
+    internal sealed class PreAllocatedOverlapped : IDisposable, IDeferredDisposable
```
jkotas: Same here

jkotas:
Looks reasonable to me so far

src/mscorlib/corefx/Interop/Unix/Interop.Libraries.cs
```diff
@@ -4,15 +4,4 @@
 
 internal static partial class Interop
 {
-    private static partial class Libraries
-    {
-        internal const string GlobalizationInterop = "System.Globalization.Native"; // CoreFX wrappers for ICU
-        // Shims
-        internal const string SystemNative = "System.Native";
```
stephentoub: Won't deleting these break the Unix build?

src/mscorlib/corefx/System/Globalization/SortKey.cs
```diff
@@ -52,26 +52,6 @@ internal SortKey(String localeName, String str, CompareOptions options, byte[] k
             _options    = options;
             _string   = str;
         }
-
-        [OnSerializing]
```
stephentoub: Anything on exposed types like this marked as [OnSerializing], [OnDeserializing], etc. should not be removed, as it's used via reflection as part of serialization/formatting.  If it was being removed before, it was a bug and should have been listed in the model.xml to prevent them from being removed.

src/mscorlib/corefx/System/IO/PathInternal.Unix.cs
```diff
@@ -43,22 +43,6 @@ internal static bool IsDirectorySeparator(char c)
...
-            return fullPath.Length >= Interop.Sys.MaxPath;
-        }
-
-        /// <summary>
-        /// Returns true if the directory is too long
-        /// </summary>
-        internal static bool IsDirectoryTooLong(string fullPath)
-        {
-            return fullPath.Length >= Interop.Sys.MaxPath;
-        }
```
stephentoub: Do we have any goal of keeping files consistent between CoreCLR and CoreFx?

src/mscorlib/src/Microsoft/Win32/SafeHandles/SafeFileMappingHandle.cs
```diff
@@ -20,20 +20,5 @@
 
 namespace Microsoft.Win32.SafeHandles
 {
-    internal sealed class SafeFileMappingHandle : SafeHandleZeroOrMinusOneIsInvalid
```
stephentoub: Can you just delete the whole file instead?

src/mscorlib/src/Microsoft/Win32/SafeHandles/SafeLocalAllocHandle.cs
```diff
@@ -8,22 +8,4 @@ namespace Microsoft.Win32.SafeHandles
     using System.Runtime.CompilerServices;
     using System.Runtime.InteropServices;
     using System.Runtime.ConstrainedExecution;
-
-    internal sealed class SafeLocalAllocHandle : SafeBuffer {
```
stephentoub: Ditto

src/mscorlib/src/Microsoft/Win32/SafeHandles/SafeViewOfFileHandle.cs
```diff
@@ -22,25 +22,5 @@
 
 namespace Microsoft.Win32.SafeHandles
 {
-    internal sealed class SafeViewOfFileHandle : SafeHandleZeroOrMinusOneIsInvalid
```
stephentoub: Ditto.  Anywhere all of the code is being removed, we should be able to instead just remove the whole file.

src/mscorlib/src/Microsoft/Win32/SafeHandles/Win32SafeHandles.cs
```diff
@@ -30,7 +30,7 @@ protected SafeHandleZeroOrMinusOneIsInvalid(bool ownsHandle) : base(IntPtr.Zero,
         }
 
         // A default constructor is needed to satisfy CoreCLR inheritence rules. It should not be called at runtime
-        protected SafeHandleZeroOrMinusOneIsInvalid()
```
stephentoub: Why are these being changed from protected to protected internal?

src/mscorlib/src/System/AggregateException.cs
```diff
@@ -474,23 +474,6 @@ public override string ToString()
...
-        /// <summary>
-        /// This helper property is used by the DebuggerDisplay.
-        /// 
-        /// Note that we don't want to remove this property and change the debugger display to {InnerExceptions.Count} 
-        /// because DebuggerDisplay should be a single property access or parameterless method call, so that the debugger 
-        /// can use a fast path without using the expression evaluator.
-        /// 
-        /// See http://msdn.microsoft.com/en-us/library/x810d419.aspx
-        /// </summary>
-        private int InnerExceptionCount
```
stephentoub: Removing this will break the DebuggerDisplayAttribute (it should not have been removed by the rewriter).  See the above comment.

src/mscorlib/src/System/AppDomainUnloadedException.cs
```diff
@@ -17,27 +17,11 @@ namespace System {
...
-    
-        public AppDomainUnloadedException(String message, Exception innerException) 
-            : base(message, innerException) {
-            SetErrorCode(__HResults.COR_E_APPDOMAINUNLOADED);
-        }
-
-        //
-        //This constructor is required for serialization.
-        //
-        protected AppDomainUnloadedException(SerializationInfo info, StreamingContext context) : base(info, context) {
```
stephentoub: I assume that an instance of this exception will never be thrown to user code?  If it could be, removing this ctor will break code if that code tries to serialize and deserialize the exception it caught.

src/mscorlib/src/System/ApplicationId.cs
```diff
@@ -15,127 +15,4 @@ namespace System {
     using System.Security.Util;
     using System.Text;
     using System.Diagnostics.Contracts;
-
```
stephentoub: The whole file can be removed.

src/mscorlib/src/System/CfgParser.cs
```diff
@@ -23,538 +23,6 @@ namespace System
...
-                BCLDebug.Trace("REMOTE","Key = "+de.Key+"   Value = "+de.Value);
-            }
-
-            for (int i=0; i<m_children.Count; i++)
-            {
-                ((ConfigNode)m_children[i]).Trace();
-            }
-        }
-#endif
-    }
```
stephentoub: Whole file can be removed

src/mscorlib/src/System/Collections/CollectionBase.cs
```diff
@@ -181,33 +166,33 @@ public abstract class CollectionBase : IList {
             return InnerList.GetEnumerator();
         }
 
-        protected virtual void OnSet(int index, Object oldValue, Object newValue) {
```
stephentoub: I'm not sure why all of these protecteds are being changed to protected internals.  If this is what the rewriter is doing, seems like a bug in the rewriter.  protected internal is more accessible than protected, not less, as protected internal is an or, not an and.

src/mscorlib/src/System/Collections/Comparer.cs
```diff
@@ -41,18 +41,6 @@ public sealed class Comparer : IComparer , ISerializable
             Contract.EndContractBlock();
             m_compareInfo = culture.CompareInfo;
         }
-        
-        private Comparer(SerializationInfo info, StreamingContext context) {
```
stephentoub: Removing this will break deserialization.  Do CoreFx tests pass with this change?  If so, that's a gap in our tests.

src/mscorlib/src/System/Collections/Concurrent/ConcurrentQueue.cs
```diff
@@ -212,36 +212,6 @@ void ICollection.CopyTo(Array array, int index)
...
-        /// </summary>
-        /// <param name="item">The object to add to the <see
-        /// cref="Concurrent.IProducerConsumerCollection{T}"/>. The value can be a null
-        /// reference (Nothing in Visual Basic) for reference types.
-        /// </param>
-        /// <returns>true if the object was added successfully; otherwise, false.</returns>
-        /// <remarks>For <see cref="ConcurrentQueue{T}"/>, this operation will always add the object to the
-        /// end of the <see cref="ConcurrentQueue{T}"/>
-        /// and return true.</remarks>
-        bool IProducerConsumerCollection<T>.TryAdd(T item)
```
stephentoub: This should break the build: without these members, the type won't properly implement the interface it claims to implement.

stephentoub: (Oh, I see, the interface was modified as well.)

src/mscorlib/src/System/Collections/Concurrent/OrderablePartitioner.cs
```diff
@@ -18,263 +18,4 @@
 namespace System.Collections.Concurrent
 {
 
-    /// <summary>
```
stephentoub: Delete the whole file

src/mscorlib/src/System/Collections/Concurrent/Partitioner.cs
```diff
@@ -17,85 +17,4 @@
 
 namespace System.Collections.Concurrent
 {
-    /// <summary>
```
stephentoub: Delete the whole file

src/mscorlib/src/System/Collections/Generic/Comparer.cs
```diff
@@ -199,9 +199,6 @@ public Int32EnumComparer()
             Debug.Assert(typeof(T).IsEnum, "This type is only intended to be used to compare enums!");
         }
         
-        // Used by the serialization engine.
-        private Int32EnumComparer(SerializationInfo info, StreamingContext context) { }
```
stephentoub: These deserialization ctors should not be removed.  We have tests for this in CoreFx (I know because I recently added them)... they pass with this change?

src/mscorlib/src/System/Collections/Hashtable.cs
```diff
@@ -290,126 +221,31 @@ internal Hashtable( bool trash )
...
         // 2) Because we choose table sizes to be primes, we just need to insure that the increment is 0 < incr < hashSize
         //
         // Thus this function would work: Incr = 1 + (seed % (hashSize-1))
         // 
-        // While this works well for �uniformly distributed� keys, in practice, non-uniformity is common. 
-        // In particular in practice we can see �mostly sequential� where you get long clusters of keys that �pack�. 
-        // To avoid bad behavior you want it to be the case that the increment is �large� even for �small� values (because small 
-        // values tend to happen more in practice). Thus we multiply �seed� by a number that will make these small values
-        // bigger (and not hurt large values). We picked HashPrime (101) because it was prime, and if �hashSize-1� is not a multiple of HashPrime
+        // While this works well for �uniformly distributed� keys, in practice, non-uniformity is common.
```
stephentoub: Some characters are messed up

src/mscorlib/src/System/Collections/KeyValuePairs.cs
```diff
@@ -15,26 +15,4 @@
 
 namespace System.Collections {
     using System.Diagnostics;
-
```
stephentoub: Delete the whole file

src/mscorlib/src/System/Collections/SortedList.cs
```diff
@@ -18,991 +18,4 @@ namespace System.Collections {
     using System.Diagnostics.CodeAnalysis;
     using System.Diagnostics.Contracts;
     using System.Globalization;
-
```
stephentoub: Delete the whole file

src/mscorlib/src/System/Collections/Stack.cs
```diff
@@ -17,363 +17,4 @@ namespace System.Collections {
...
     using System.Diagnostics.CodeAnalysis;
     using System.Diagnostics.Contracts;
-
-    // A simple stack of objects.  Internally it is implemented as an array,
-    // so Push can be O(n).  Pop is O(1).
-    [DebuggerTypeProxy(typeof(System.Collections.Stack.StackDebugView))] 
-    [DebuggerDisplay("Count = {Count}")]
-[System.Runtime.InteropServices.ComVisible(true)]
-    [Serializable]
-    public class Stack : ICollection, ICloneable {
```
stephentoub: Delete the whole file

src/mscorlib/src/System/Configuration/Assemblies/AssemblyHash.cs
```diff
@@ -12,63 +12,5 @@
 ===========================================================*/
 namespace System.Configuration.Assemblies {
     using System;
-    [Serializable]
```
stephentoub: Delete the whole file

src/mscorlib/src/System/ContextMarshalException.cs
```diff
@@ -19,27 +19,5 @@ namespace System {
 	using System.Runtime.Remoting;
 	using System;
 	using System.Runtime.Serialization;
-    [System.Runtime.InteropServices.ComVisible(true)]
```
stephentoub: Delete the whole file

src/mscorlib/src/System/Decimal.cs
```diff
@@ -346,9 +346,6 @@ public static Decimal Add(Decimal d1, Decimal d2)
         [MethodImplAttribute(MethodImplOptions.InternalCall)]
         private static extern void FCallAddSub(ref Decimal d1, ref Decimal d2, byte bSign);
         
-        [MethodImplAttribute(MethodImplOptions.InternalCall)]
-        private static extern void FCallAddSubOverflowed(ref Decimal d1, ref Decimal d2, byte bSign, ref bool overflowed);
```
stephentoub: For these fcalls, is there corresponding code in the runtime that can/should be deleted as well?

src/mscorlib/src/System/Diagnostics/Eventing/EventSource.cs
```diff
@@ -4180,39 +4180,13 @@ public enum EventSourceSettings
...
-        /// as well as for any EventSources that come into existence after the EventListener
-        /// has been created.
-        /// 
-        /// These 'catch up' events are called during the construction of the EventListener.
-        /// Subclasses need to be prepared for that.
-        /// 
-        /// In a multi-threaded environment, it is possible that 'EventSourceEventWrittenCallback' 
-        /// events for a particular eventSource to occur BEFORE the EventSourceCreatedCallback is issued.
-        /// </summary>
-        public event EventHandler<EventSourceCreatedEventArgs> EventSourceCreated
```
stephentoub: I expect @brianrob will want this to remain consistent with the copy in CoreFx.

src/mscorlib/src/System/Environment.cs
```diff
@@ -226,9 +226,6 @@ internal sealed class ResourceHelper
         [MethodImplAttribute(MethodImplOptions.InternalCall)]
         public static extern void FailFast(String message);
 
-        [MethodImplAttribute(MethodImplOptions.InternalCall)]
-        internal static extern void FailFast(String message, uint exitCode);
```
stephentoub: Is there corresponding runtime code to be deleted?

src/mscorlib/src/System/Environment.cs
```diff
@@ -543,31 +485,14 @@ private unsafe static char[] GetEnvironmentCharArray()
...
 
-        
-        /*==================================WorkingSet==================================
-        **Action:
-        **Returns:
-        **Arguments:
-        **Exceptions:
-        ==============================================================================*/
-        [DllImport(JitHelpers.QCall, CharSet = CharSet.Unicode), SuppressUnmanagedCodeSecurity]
-        private static extern long GetWorkingSet();
```
stephentoub: Corresponding runtime code to be deleted?

src/mscorlib/src/System/Globalization/CultureInfo.cs
```diff
@@ -1845,9 +1832,6 @@ internal static bool IsTaiwanSku
         [MethodImplAttribute(MethodImplOptions.InternalCall)]
         internal static extern int nativeGetLocaleInfoExInt(String localeName, uint field);
 
-        [MethodImplAttribute(MethodImplOptions.InternalCall)]
-        internal static extern bool nativeSetThreadLocale(String localeName);
```
stephentoub: Corresponding runtime code to be deleted?

src/mscorlib/src/System/Globalization/EncodingTable.cs
```diff
@@ -203,10 +203,6 @@ internal static int GetCodePageFromName(String name)
 
         [MethodImplAttribute(MethodImplOptions.InternalCall)]
         private unsafe static extern InternalCodePageDataItem* GetCodePageData();
-
-        [MethodImplAttribute(MethodImplOptions.InternalCall)]
-        internal unsafe static extern byte* nativeCreateOpenFileMapping(
-            String inSectionName, int inBytesToAllocate, out IntPtr mappedFileHandle);
```
stephentoub: Corresponding runtime code to be deleted?

src/mscorlib/src/System/Globalization/SortKey.cs
```diff
@@ -55,25 +55,6 @@ internal SortKey(String localeName, String str, CompareOptions options, byte[] k
...
-
-        [OnDeserialized]
-        private void OnDeserialized(StreamingContext context)
-        {
-            //set locale name to proper value after Whidbey deserialization
-            if (String.IsNullOrEmpty(localeName) && win32LCID != 0)
-            {
-                localeName = CultureInfo.GetCultureInfo(win32LCID).Name;
-            }
-        }
```
stephentoub: As per earlier comment, removing these serialization-attributed members will break serialization/deserialization of the type.  We probably have a test gap.

src/mscorlib/src/System/IO/DriveNotFoundException.cs
```diff
@@ -18,7 +18,7 @@ namespace System.IO {
     //Thrown when trying to access a drive that is not availabe.
     [Serializable]
     [System.Runtime.InteropServices.ComVisible(true)]
-    public class DriveNotFoundException : IOException {
+    internal class DriveNotFoundException : IOException {
```
stephentoub: Where is this exception still used?  If it may be thrown to user code, it needs its deserialization ctor put back.  But I would have expected this whole type could be deleted from CoreCLR.

src/mscorlib/src/System/IO/FileSystemEnumerable.cs
```diff
@@ -25,598 +25,5 @@
 
 namespace System.IO
 {
-    // Overview:
```
stephentoub: Delete the whole file

src/mscorlib/src/System/IO/SearchOption.cs
```diff
@@ -19,18 +19,4 @@
 using System;
 
 namespace System.IO {
-    [Serializable]
```
stephentoub: Delete the whole file

src/mscorlib/src/System/Threading/ThreadLocal.cs
```diff
@@ -486,20 +486,6 @@ private List<T> GetValuesAsList()
             return valueList;
         }
 
-        /// <summary>Gets the number of threads that have data in this instance.</summary>
-        private int ValuesCountForDebugDisplay
```
stephentoub: Removing this will break the DebuggerDisplayAttribute.

src/mscorlib/src/System/Threading/Thread.cs
```diff
@@ -212,7 +201,7 @@ static void AsyncLocalSetCurrentUICulture(AsyncLocalValueChangedArgs<CultureInfo
             SetStartHelper((Delegate)start, 0);
         }
 
-        public Thread(ParameterizedThreadStart start, int maxStackSize) {
+        internal Thread(ParameterizedThreadStart start, int maxStackSize) {
```
stephentoub: I'm surprised these public methods are being removed; aren't we exposing them in .NET Standard 2.0?

src/mscorlib/src/System/Threading/Tasks/ConcurrentExclusiveSchedulerPair.cs
```diff
@@ -251,16 +251,6 @@ private void FaultWithTask(Task faultedTask)
...
 
-        /// <summary>Gets the number of tasks waiting to run concurrently.</summary>
-        /// <remarks>This does not take the necessary lock, as it's only called from under the debugger.</remarks>
-        [SuppressMessage("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode")]
-        private int ConcurrentTaskCountForDebugger { get { return m_concurrentTaskScheduler.m_tasks.Count; } }
-
-        /// <summary>Gets the number of tasks waiting to run exclusively.</summary>
-        /// <remarks>This does not take the necessary lock, as it's only called from under the debugger.</remarks>
-        [SuppressMessage("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode")]
-        private int ExclusiveTaskCountForDebugger { get { return m_exclusiveTaskScheduler.m_tasks.Count; } }
```
stephentoub: Removing all of these "ForDebugger" properties breaks the debugger display attributes.

src/mscorlib/src/System/Threading/LockCookie.cs
```diff
@@ -16,42 +16,5 @@
 namespace System.Threading {
 
     using System;
-    [System.Runtime.InteropServices.ComVisible(true)]
```
stephentoub: Delete the whole file

src/mscorlib/src/System/Text/UnicodeEncoding.cs
```diff
@@ -1770,31 +1770,6 @@ public Decoder(UnicodeEncoding encoding) : base(encoding)
                 // base calls reset
             }
 
-            // Constructor called by serialization, have to handle deserializing from Everett
-            internal Decoder(SerializationInfo info, StreamingContext context)
```
stephentoub: Removing this will break deserialization of this type

src/mscorlib/src/System/Text/UTF8Encoding.cs
```diff
@@ -2158,29 +2158,6 @@ public UTF8Encoder(UTF8Encoding encoding) : base(encoding)
                 // base calls reset
             }
 
-            // Constructor called by serialization, have to handle deserializing from Everett
-            internal UTF8Encoder(SerializationInfo info, StreamingContext context)
```
stephentoub: Removing this will break deserialization of this type

src/mscorlib/src/System/Text/UTF7Encoding.cs
```diff
@@ -652,20 +652,6 @@ public Decoder(UTF7Encoding encoding) : base (encoding)
                 // base calls reset
             }
 
-            // Constructor called by serialization, have to handle deserializing from Everett
-            internal Decoder(SerializationInfo info, StreamingContext context)
```
stephentoub: Removing this will break deserialization of this type

src/mscorlib/src/System/Text/Latin1Encoding.cs
```diff
@@ -32,17 +32,6 @@ public Latin1Encoding() : base(Encoding.ISO_8859_1)
         {
         }
 
-        // Constructor called by serialization.
-        // Note:  We use the base GetObjectData however
-        internal Latin1Encoding(SerializationInfo info, StreamingContext context) :
```
stephentoub: Removing this will break deserialization of this type

src/mscorlib/src/System/Text/Encoding.cs
```diff
@@ -1432,25 +1392,6 @@ public DefaultEncoder(Encoding encoding)
                 m_hasInitializedEncoding = true;
             }
 
-            // Constructor called by serialization, have to handle deserializing from Everett
-            internal DefaultEncoder(SerializationInfo info, StreamingContext context)
```
stephentoub: Removing this will break deserialization of this type

src/mscorlib/src/System/Text/Encoding.cs
```diff
@@ -1553,26 +1494,6 @@ public DefaultDecoder(Encoding encoding)
                 m_hasInitializedEncoding = true;
            }
 
-            // Constructor called by serialization, have to handle deserializing from Everett
-            internal DefaultDecoder(SerializationInfo info, StreamingContext context)
```
stephentoub: Removing this will break deserialization of this type

src/mscorlib/src/System/Text/EncoderNLS.cs
```diff
@@ -35,15 +35,6 @@ internal class EncoderNLS : Encoder, ISerializable
 
 #region Serialization
 
-        // Constructor called by serialization. called during deserialization.
-        internal EncoderNLS(SerializationInfo info, StreamingContext context)
```
stephentoub: Removing this will break deserialization of this type

src/mscorlib/src/System/Text/DecoderNLS.cs
```diff
@@ -32,15 +32,6 @@ internal class DecoderNLS : Decoder, ISerializable
 
 #region Serialization
 
-        // Constructor called by serialization. called during deserialization.
-        internal DecoderNLS(SerializationInfo info, StreamingContext context)
```
stephentoub: Removing this will break deserialization of this type (even though I see its implementation is to throw a not supported exception)

src/mscorlib/src/System/Runtime/Serialization/ISurrogateSelector.cs
```diff
@@ -17,18 +17,4 @@ namespace System.Runtime.Serialization {
     using System.Runtime.Remoting;
     using System.Security.Permissions;
     using System;
-[System.Runtime.InteropServices.ComVisible(true)]
```
stephentoub: Delete the whole file

src/mscorlib/src/System/Runtime/Serialization/MemberHolder.cs
```diff
@@ -16,36 +16,4 @@ namespace System.Runtime.Serialization {
     using System.Runtime.Remoting;
     using System;
     using System.Reflection;
-    [Serializable]
```
stephentoub: Delete the whole file (same goes for the rest of these)

src/mscorlib/src/System/Runtime/Loader/AssemblyLoadContext.cs
```diff
@@ -18,9 +18,6 @@ namespace System.Runtime.Loader
 {
     public abstract class AssemblyLoadContext
     {
-        [DllImport(JitHelpers.QCall, CharSet = CharSet.Unicode)]
-        [SuppressUnmanagedCodeSecurity]
-        private static extern bool OverrideDefaultAssemblyLoadContextForCurrentDomain(IntPtr ptrNativeAssemblyLoadContext);
```
stephentoub: Anything corresponding to remove from the runtime? Same below.

src/mscorlib/src/System/Runtime/InteropServices/RuntimeEnvironment.cs
```diff
@@ -31,75 +31,13 @@ static public class RuntimeEnvironment {
         [MethodImplAttribute(MethodImplOptions.InternalCall)]
         internal static extern String GetModuleFileName();
 
-        [MethodImplAttribute(MethodImplOptions.InternalCall)]
```
stephentoub: Anything to remove from the runtime for these?

src/mscorlib/src/System/Runtime/InteropServices/GcHandle.cs
```diff
@@ -296,8 +296,6 @@ internal void SetIsPinned()
         internal static extern IntPtr InternalAddrOfPinnedObject(IntPtr handle);
         [MethodImplAttribute(MethodImplOptions.InternalCall)]
         internal static extern void InternalCheckDomain(IntPtr handle);
-        [MethodImplAttribute(MethodImplOptions.InternalCall)]
-        internal static extern GCHandleType InternalGetHandleType(IntPtr handle);
```
stephentoub: Anything to remove from the runtime?

src/mscorlib/src/System/Runtime/CompilerServices/RuntimeWrappedException.cs
```diff
@@ -41,11 +41,6 @@ private RuntimeWrappedException(Object thrownObject)
             base.GetObjectData(info, context);
             info.AddValue("WrappedException", m_wrappedException, typeof(Object));
         }
-
-        internal RuntimeWrappedException(SerializationInfo info, StreamingContext context)
```
stephentoub: Removing this will break deserialization, but I'm not sure if we ever actually throw these in CoreCLR.

src/mscorlib/src/System/Runtime/CompilerServices/AsyncMethodBuilder.cs
```diff
@@ -385,18 +385,6 @@ public void SetResult()
         public void SetException(Exception exception) { m_builder.SetException(exception); }
 
         /// <summary>
-        /// Called by the debugger to request notification when the first wait operation
-        /// (await, Wait, Result, etc.) on this builder's task completes.
-        /// </summary>
-        /// <param name="enabled">
-        /// true to enable notification; false to disable a previously set notification.
-        /// </param>
-        internal void SetNotificationForWaitCompletion(bool enabled)
```
stephentoub: This must not be deleted.  It will break debugging of async methods.

src/mscorlib/src/System/Runtime/CompilerServices/AsyncMethodBuilder.cs
```diff
@@ -673,23 +661,6 @@ public void SetException(Exception exception)
...
-        /// (await, Wait, Result, etc.) on this builder's task completes.
-        /// </summary>
-        /// <param name="enabled">
-        /// true to enable notification; false to disable a previously set notification.
-        /// </param>
-        /// <remarks>
-        /// This should only be invoked from within an asynchronous method,
-        /// and only by the debugger.
-        /// </remarks>
-        internal void SetNotificationForWaitCompletion(bool enabled)
```
stephentoub: This must not be deleted.  It will break debugging of async methods.

src/mscorlib/src/System/InsufficientMemoryException.cs
```diff
@@ -38,8 +38,5 @@ public InsufficientMemoryException(String message, Exception innerException)
             : base(message, innerException) {
             SetErrorCode(__HResults.COR_E_INSUFFICIENTMEMORY);
         }
-
-        private InsufficientMemoryException(SerializationInfo info, StreamingContext context) : base(info, context) {
```
stephentoub: Removing this will break deserialization of this exception

src/mscorlib/src/System/IO/TextReader.cs
```diff
@@ -24,378 +24,4 @@
 using System.Threading.Tasks;
 
 namespace System.IO {
-    // This abstract base class represents a reader that can read a sequential
```
stephentoub: Delete the whole file

src/mscorlib/Common/PinnableBufferCache.cs
```diff
@@ -21,22 +21,6 @@ namespace System
 {
     internal sealed class PinnableBufferCache
     {
-        /// <summary>
-        /// Create a new cache for pinned byte[] buffers
-        /// </summary>
-        /// <param name="cacheName">A name used in diagnostic messages</param>
-        /// <param name="numberOfElements">The size of byte[] buffers in the cache (they are all the same size)</param>
-        public PinnableBufferCache(string cacheName, int numberOfElements) : this(cacheName, () => new byte[numberOfElements]) { }
```
brianrob: You can remove the entire file.  This is not used anywhere in core.

src/mscorlib/src/System/Diagnostics/Eventing/EventSource.cs
brianrob: Yes, let's leave this one as is please.

src/mscorlib/src/System/Diagnostics/Eventing/EventSource_CoreCLR.cs
```diff
@@ -116,20 +116,10 @@ public static Guid CurrentThreadActivityId
             }
         }
 
-        private int GetParameterCount(EventMetadata eventData)
```
brianrob: Same for this one as well.

src/mscorlib/src/System/Diagnostics/Eventing/StubEnvironment.cs
```diff
@@ -5,51 +5,6 @@
...
-#endif
-    using Microsoft.Reflection;
-    using System.Reflection;
-
-    internal static class Environment
-    {
-        public static readonly string NewLine = System.Environment.NewLine;
-
-        public static int TickCount
-        { get { return System.Environment.TickCount; } }
```
brianrob: Same here.

src/mscorlib/src/System/Diagnostics/Eventing/TraceLogging/DataCollector.cs
```diff
@@ -142,11 +142,6 @@ internal void AddBinary(string value, int size)
             }
         }
 
-        internal void AddBinary(Array value, int size)
-        {
-            this.AddArray(value, size, 1);
```
brianrob: Same here.

src/mscorlib/src/System/Diagnostics/Eventing/TraceLogging/FieldMetadata.cs
```diff
@@ -59,44 +59,6 @@ internal class FieldMetadata
...
-        /// <summary>
-        /// Fixed-length array.
-        /// </summary>
-        public FieldMetadata(
-            string name,
-            TraceLoggingDataType type,
-            EventFieldTags tags,
-            ushort fixedCount)
-            : this(
-                name,
```
brianrob: Same here.

src/mscorlib/src/System/Diagnostics/Eventing/TraceLogging/Statics.cs
```diff
@@ -144,21 +144,6 @@ public static void EncodeTags(int tags, ref int pos, byte[] metadata)
             }
         }
 
-        public static byte Combine(
-            int settingValue1,
-            int settingValue2,
-            byte defaultValue)
-        {
-            unchecked
```
brianrob: Same here.

src/mscorlib/src/System/Diagnostics/Eventing/TraceLogging/TraceLoggingDataCollector.cs
```diff
@@ -49,17 +49,6 @@ public void EndBufferedArray(int bookmark, int count)
             DataCollector.ThreadInstance.EndBufferedArray(bookmark, count);
         }
 
-        /// <summary>
-        /// Adds the start of a group to the event.
-        /// This has no effect on the event payload, but is provided to allow
-        /// WriteMetadata and WriteData implementations to have similar
-        /// sequences of calls, allowing for easier verification of correctness.
```
brianrob: Same here.

src/mscorlib/src/System/Diagnostics/Eventing/TraceLogging/TraceLoggingEventTypes.cs
```diff
@@ -166,30 +166,6 @@ internal string Name
         }
 
         /// <summary>
-        /// Gets the default level that will be used for events with this descriptor.
-        /// </summary>
-        internal EventLevel Level
-        {
-            get { return (EventLevel)this.level; }
-        }
```
brianrob: Same here.

src/mscorlib/src/System/Diagnostics/Eventing/TraceLogging/TraceLoggingMetadataCollector.cs
```diff
@@ -262,30 +262,6 @@ public void EndBufferedArray()
             this.impl.EndBuffered();
         }
 
-        /// <summary>
-        /// Adds a custom-serialized field to an event.
```
brianrob: Same here.

danmosemsft:
Thanks for the detailed comments everyone - I hadn't expected a detailed review yet. I'll work through this and rebase.

src/mscorlib/src/System/Threading/Thread.cs
jkotas: The public System.Threading.Thread is corefx

stephentoub: Ah, right, I forgot it got moved.

src/mscorlib/src/System/Collections/Comparer.cs
jkotas: This whole file should be deleted. The shipping public implementation is in corefx. It is just that something is keeping alive in CoreLib.

src/mscorlib/src/System/Diagnostics/Eventing/EventSource.cs
danmosemsft: Should I introduce an #if CORECLR for these? Eg WriteEvent(..) are quite large. Don't know how much we care about size.

---
In reply to: [99062226](https://github.com/dotnet/coreclr/pull/9269#discussion_r99062226) [](ancestors = 99062226)

src/mscorlib/corefx/Interop/Unix/Interop.Libraries.cs
danmosemsft: Yes, I just hadn't got to that.

---
In reply to: [99095195](https://github.com/dotnet/coreclr/pull/9269#discussion_r99095195) [](ancestors = 99095195)

src/mscorlib/corefx/System/Globalization/SortKey.cs
danmosemsft: Unix only, didn't diff those yet. I"ll revert any.

---
In reply to: [99095564](https://github.com/dotnet/coreclr/pull/9269#discussion_r99095564) [](ancestors = 99095564)

danmosemsft: Looks like yes it was previously stripping it. I'll open a bug in corefx to add tests for serializing those.

---
In reply to: [99196301](https://github.com/dotnet/coreclr/pull/9269#discussion_r99196301) [](ancestors = 99196301,99095564)

danmosemsft: Yes, rewriter was stripping some of them. I found all the ones getting stripped and reverted them, also opened this to add tests https://github.com/dotnet/corefx/issues/15731

src/mscorlib/src/System/Diagnostics/Eventing/EventSource.cs
brianrob: To the extent possible, I'd rather have fewer ifdefs in this code simply because we'd like EventSource to be as close to the same as possible on all platforms.  Are there particular methods that you're concerned about?  I believe that all of the WriteEvent methods should be available on all platforms and should not get removed by the rewriter.  Is this not the case?

src/mscorlib/corefx/System/IO/PathInternal.Unix.cs
danmosemsft: @JeremyKuhne why do we have a copy of this in PathInternal in CoreFX?

src/mscorlib/src/Microsoft/Win32/SafeHandles/SafeFileMappingHandle.cs
danmosemsft: I'll delete empty files

src/mscorlib/src/Microsoft/Win32/SafeHandles/SafeViewOfFileHandle.cs
danmosemsft: Yes I just hadn't done that yet, you're too fast.

src/mscorlib/src/System/AggregateException.cs
danmosemsft: Fixed, and checked all the other such attributes.

src/mscorlib/src/System/AppDomainUnloadedException.cs
danmosemsft: Brought back as part of bringing back all the serialization.

src/mscorlib/src/System/AggregateException.cs
stephentoub: Thanks.  This is also something it'd be good to have some tests in CoreFx for.  We have them in a few libraries, with tests that look at types we expect to have debugger attributes, and then use some basic parsing and reflection to verify that the expected members exist and can be invoked.  But it'd be good to effectively cover all of our types with such attributes; it's so easy to accidentally remove such members, either manually or via automatic tools like this.

src/mscorlib/src/System/Collections/Comparer.cs
danmosemsft: It's used in several places in Array and in some other types so I believe I have to keep it.

I brought back this with the other serialiation stuff.

src/mscorlib/src/System/Collections/Generic/Comparer.cs
danmosemsft: all back

src/mscorlib/src/System/Collections/Hashtable.cs
danmosemsft: fixed

src/mscorlib/src/System/Decimal.cs
danmosemsft: Several. Removed

src/mscorlib/src/System/Diagnostics/Eventing/EventSource.cs
danmosemsft: Reverted all Eventing/*. I promise I was going to!

src/mscorlib/corefx/System/IO/PathInternal.Unix.cs
JeremyKuhne: We're trying to remove the pre-emptive max path checks.

src/mscorlib/src/System/Environment.cs
danmosemsft: Removed

src/mscorlib/src/System/Environment.cs
danmosemsft: Removed. I think there are other things in ecalllist.h that are dead but I didn't go through.

src/mscorlib/src/System/Globalization/CultureInfo.cs
danmosemsft: fixed

src/mscorlib/src/System/Globalization/SortKey.cs
danmosemsft: Fixed. Why do we have two near identical copies of SortKey?

src/mscorlib/src/System/IO/DriveNotFoundException.cs
danmosemsft: Used in __Error. Brought back.

src/mscorlib/src/System/Threading/ThreadLocal.cs
danmosemsft: Thanks, fixed

src/mscorlib/src/System/Threading/Tasks/ConcurrentExclusiveSchedulerPair.cs
danmosemsft: Fixed

src/mscorlib/src/System/IO/DriveNotFoundException.cs
stephentoub: Does _Error actually need it, though?  With DriveInfo being in CoreFx, I'm wondering if anything in CoreCLR would actually generate the error code that would trigger that exception to be created/thrown.

src/mscorlib/src/System/Text/UnicodeEncoding.cs
danmosemsft: fixed

src/mscorlib/src/System/Text/UTF8Encoding.cs
danmosemsft: fixed

src/mscorlib/src/System/Text/UTF7Encoding.cs
danmosemsft: fixed

src/mscorlib/src/System/Text/Latin1Encoding.cs
danmosemsft: fixed

src/mscorlib/src/System/Text/Encoding.cs
danmosemsft: fixed

src/mscorlib/src/System/Text/Encoding.cs
danmosemsft: fixed

src/mscorlib/src/System/Text/EncoderNLS.cs
danmosemsft: fixed

src/mscorlib/src/System/Text/DecoderNLS.cs
danmosemsft: fixed

src/mscorlib/src/System/IO/DriveNotFoundException.cs
jkotas: And if is actually needed by _Error, the copy in corefx should be deleted.

The other problem is that the old `__Error` is duplicate of the new refactored `Win32Marshal.GetExceptionForWin32Error` that was copied from corefx. The mapping for `DriveNotFoundException` was removed from `Win32Marshal.GetExceptionForWin32Error` as part of this refactoring. This duplication and inconsistency should be fixed.

src/mscorlib/src/System/Collections/Comparer.cs
jkotas: It is used as internal implementation detail that cannot leak and thus cannot be serialized. It was renamed to LowLevelComparer (with more cruft stripped) in CoreRT. We should get it eventually as we are unifying the CoreCLR and CoreRT impls.

src/mscorlib/src/System/Runtime/Loader/AssemblyLoadContext.cs
danmosemsft: fixed

src/mscorlib/src/System/Runtime/InteropServices/RuntimeEnvironment.cs
danmosemsft: done

src/mscorlib/src/System/Runtime/InteropServices/GcHandle.cs
danmosemsft: done

src/mscorlib/src/System/Runtime/CompilerServices/RuntimeWrappedException.cs
danmosemsft: brought back

src/mscorlib/src/System/Runtime/CompilerServices/AsyncMethodBuilder.cs
danmosemsft: They would already be broken of course. Brought back.

src/mscorlib/src/System/Runtime/CompilerServices/AsyncMethodBuilder.cs
danmosemsft: fixed

src/mscorlib/src/System/InsufficientMemoryException.cs
danmosemsft: fixed

src/mscorlib/Common/PinnableBufferCache.cs
danmosemsft: It's used by System.Threading.Overlapped, which is used by other classes.

src/mscorlib/src/System/Diagnostics/Eventing/EventSource_CoreCLR.cs
danmosemsft: removed. I don't know why the rewriter didn't do this.

src/mscorlib/src/System/Diagnostics/Eventing/StubEnvironment.cs
danmosemsft: removed class

src/mscorlib/src/System/Diagnostics/Eventing/TraceLogging/DataCollector.cs
danmosemsft: removed method

src/mscorlib/src/System/Runtime/CompilerServices/AsyncMethodBuilder.cs
stephentoub: Thanks. Can you make sure these and others are added to model file so they don't get removed?

jkotas: We should get rid of the BCL rewriter and model.xml as part of this change; or immediately after this change goes in. I would not bother maintaining model.xml.

stephentoub: If we're sure we're going to delete them asap, great.

src/mscorlib/src/System/Globalization/SortKey.cs
jkotas: >  Why do we have two near identical copies of SortKey?

We have two copies of globalization:
- For Windows, we use the old one from desktop
- For Unix, we use the new one from CoreRT.

Ideally, we would use the one from CoreRT on Windows as well, but it needs some work... .

src/mscorlib/Common/PinnableBufferCache.cs
brianrob: Yes, you're right.  I forgot that it is used by Overlapped.  In that case, do you mind keeping the public methods and ifdef them out (or just keep them because they're small and the class is internal)?  There are a handful of copies of this class around and it would be good to have these around in case we need to use them.  That way we don't have to copy them from another source.

src/mscorlib/src/System/Diagnostics/Eventing/TraceLogging/FieldMetadata.cs
danmosemsft: removed method. I'm relying on your implying that desktop doesn't need it either.

src/mscorlib/src/System/Diagnostics/Eventing/TraceLogging/Statics.cs
danmosemsft: removed method.

src/mscorlib/src/System/Diagnostics/Eventing/TraceLogging/TraceLoggingDataCollector.cs
danmosemsft: removed method. I'm removing what you've flagged but it seems better to have a separate issue to remove dead stuff from eventing.

src/mscorlib/src/System/Diagnostics/Eventing/TraceLogging/TraceLoggingEventTypes.cs
danmosemsft: removed property

src/mscorlib/src/System/Diagnostics/Eventing/TraceLogging/TraceLoggingMetadataCollector.cs
danmosemsft: Removed method.

src/mscorlib/corefx/System/IO/PathInternal.Unix.cs
danmosemsft: @JeremyKuhne I interpret that to mean there is a good reason for both, and I can continue to delete what's not used by coreclr.

danmosemsft:
@brianrob can you clarify, you DO want Diagnostics/Eventing/** left untouched?

brianrob:
Yes, untouched.  Sorry for the confusion.

danmosemsft:
OK rolled back Eventing,

src/mscorlib/Common/PinnableBufferCache.cs
danmosemsft: @brianrob can you clarify? I can't ifdef anything out because Overlapped is using it?

src/mscorlib/src/System/Collections/CollectionBase.cs
danmosemsft: I don't know why either -- I rolled the back.

src/mscorlib/src/System/IO/DriveNotFoundException.cs
danmosemsft: I"ll open an issue.

danmosemsft:
I believe I addressed all the feedback, let me repeat all the diffing before I ask for a final CR.

src/mscorlib/model.xml
```diff
@@ -8160,7 +8160,6 @@
       <Member Name="Dispose" />
     </Type>
     <Type Name="System.Threading.ThreadPoolBoundHandle">
-      <Member Name="get_Handle" />
```
jkotas: Undo

src/mscorlib/src/System/IO/DriveNotFoundException.cs
danmosemsft: For the error, do you mean that you happen to know that ERROR_DRIVE_INFO is only relevant to GetVolumeInformation (which is only used in corefx)?

danmosemsft:
I've diffed disassembly and API for linux debug and release, x86/x64 windows and all seems fine.

Any comments on final review? I've not been successful coaxing corefx to run tests on my private binary yet.

jkotas:
Looks great to me!

jkotas:
I guess you have deleted them already. mscorlib.shared.sources.props is showing up as binary file so I have not seen it in the diff.

danmosemsft:
Fixed binary .props file.
All CoreFX tests pass against release x64.

If you get to it first, please squash before commit...

danmosemsft:
@ericstj  --
```
D:\j\workspace\debug_windows---17180f1b\Tools\Packaging.targets(1061,5): error MSB4018: The "GeneratePackageReport" task failed unexpectedly. [D:\j\workspace\debug_windows---17180f1b\src\.nuget\Microsoft.NETCore.Runtime.CoreCLR\Microsoft.NETCore.Runtime.CoreCLR.pkgproj]
21:09:19 D:\j\workspace\debug_windows---17180f1b\Tools\Packaging.targets(1061,5): error MSB4018: System.IO.IOException: The process cannot access the file 'D:\j\workspace\debug_windows---17180f1b\bin\Product\Windows_NT.x64.Debug\mscordaccore_amd64_amd64_4.6.25005.00.dll' because it is being used by another process. [D:\j\workspace\debug_windows---17180f1b\src\.nuget\Microsoft.NETCore.Runtime.CoreCLR\Microsoft.NETCore.Runtime.CoreCLR.pkgproj]
21:09:19 D:\j\workspace\debug_windows---17180f1b\Tools\Packaging.targets(1061,5): error MSB4018:    at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath) [D:\j\workspace\debug_windows---17180f1b\src\.nuget\Microsoft.NETCore.Runtime.CoreCLR\Microsoft.NETCore.Runtime.CoreCLR.pkgproj]
21:09:19 D:\j\workspace\debug_windows---17180f1b\Tools\Packaging.targets(1061,5): error MSB4018:    at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost) [D:\j\workspace\debug_windows---17180f1b\src\.nuget\Microsoft.NETCore.Runtime.CoreCLR\Microsoft.NETCore.Runtime.CoreCLR.pkgproj]
21:09:19 D:\j\workspace\debug_windows---17180f1b\Tools\Packaging.targets(1061,5): error MSB4018:    at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share) [D:\j\workspace\debug_windows---17180f1b\src\.nuget\Microsoft.NETCore.Runtime.CoreCLR\Microsoft.NETCore.Runtime.CoreCLR.pkgproj]
21:09:19 D:\j\workspace\debug_windows---17180f1b\Tools\Packaging.targets(1061,5): error MSB4018:    at Microsoft.DotNet.Build.Tasks.Packaging.VersionUtility.GetAssemblyVersion(String assemblyPath) [D:\j\workspace\debug_windows---17180f1b\src\.nuget\Microsoft.NETCore.Runtime.CoreCLR\Microsoft.NETCore.Runtime.CoreCLR.pkgproj]
```

danmosemsft:
@dotnet-bot test Windows_NT x64 Debug Build and Test please  (as above)

mattwarren:
Just a small thing I noticed, when you made the changes to [SharedStatics.cs (diff from #9269)](https://github.com/dotnet/coreclr/pull/9269/files#diff-b5bed7208c19377439e89a3a16917d87), it stripped out the `public`/`internal` `properties`/`methods`, but left the `private` backing fields behind [you can see that they are still there](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/SharedStatics.cs#L37-L40)

Should they be removed as well?

danmosemsft:
@mattwarren thanks, did you notice anything else? I can remove those.

mattwarren:
No sorry, I was only looking at `SharedStatics.cs` when I noticed the issue.

We're these changes done with a tool and is there a chance that the tool isn't properly stripping out `private` code that was only used by the `public` code it removed?

danmosemsft:
I diffed assemblies before and after BCLRewriter then used Roslyn to remove all those types. I did some manual work as the rewriter had bugs (as did my Roslyn code). Then I diffed the API manually including privates. There's enough noise at the level of privates though that I can imagine some thing slipped through. I do'nt plan to do another pass unless there's some big mistake I made.

PR #9270, opened 2017-02-02
Delete remaining legacy ComInterfaces

jkotas:


jkotas:
@danmosemsft I have noticed that this did not get deleted in your cleanup, so I have done it manually.

danmosemsft:
Looks like the rewriter was not removing them.

jkotas:
There are clusters of types where we should not need anything, but the rewriter has been keeping some of it because of some bogus references.

This was one small example. System.Security is a big cluster like this and there are a few more smaller ones. I can go clean it up the ones I see now; or wait for your change to go in. Doing it now will make your cleanup change smaller, but you will need to resolve some conflicts (keep the deleted file). Let me know what you prefer.

danmosemsft:
It's a bit painful to rebase what I have + repeat validation when I do so it might make sense to wait to commit what you suggest. With interruptions it might be a few days but I will finish this weekend if necessary.

Would definitely like to continue the cleanup momentum though.

Issue #8937, opened 2017-01-13
React to CoreFX layout changes


ellismg:
dotnet/corefx#15135 is introducing changes in how both their build is driven, the layout on disk of their artifacts, and how tests are run.

In CI we have jobs that depend on all of the above which would be broken by this change.  We are fixing the version we depend on in CI to before this change, but we need to go back and then update CI to use the new layouts so we can continue to pick up new versions of CoreFX.

gkhanna79:
@ellismg @wtgodbe merged https://github.com/dotnet/coreclr/pull/9119 and https://github.com/dotnet/coreclr/pull/9113 to use the new CoreFX binary layout for running CoreCLR tests. Is there anything else required to address this issue?

wtgodbe:
@ellismg I believe this has been addressed, the CI has been looking good since the changes Gaurav mentioned above - anything else that needs to be done? Otherwise I'm going to close this at EOD

ellismg:
Nothing else I can think of.  I see @RussKeldorph fixed up the python scripts as well.  Closing.

jkotas:
@wtgodbe @ellismg The CoreFX version is not getting updated in CoreCLR anymore - the CoreCLR tests on my local box are running against January 13 CoreFX. When are we going to get live CoreFX from CoreCLR again? (I need to get more recent CoreFX to fix https://github.com/dotnet/coreclr/pull/9278.)

gkhanna79:
I chatted with @wtgodbe yesterday and he is looking into addressing this.

gkhanna79:
@wtgodbe What is the next step on this?

wtgodbe:
I'm still working on that change, I wasn't able to make much progress yesterday as I spent most of the day working on other things - I'm probably still a few days out from putting a PR up.

gkhanna79:
@wtgodbe Any update on this?

wtgodbe:
I'm going to be focusing on this this week, per our discussion last Tuesday

ramarag:
@wtgodbe i beleive we talked about a solution for this a long time ago to actually not to build corefx for xplat.

I believe the fix is just getting the following in to replace for the build, so that we always get what we want from nuget and do not need to build corefx in CI

```
restore_TestAssets()
{

    if [ $__isMSBuildOnNETCoreSupported == 0 ]; then
        echo "System.Private.CoreLib.dll build unsupported."
        return
    fi

[...]

PR #9278, opened 2017-02-02
Delete mscorlib facade

### LINKED ISSUE #7607: Move mscorlib facade from CoreCLR to CoreFX
cc: @jkotas @weshaggard 

Today we version mscorlib and System.Private.CoreLib as netstandard1.0 which means that we use the same two dlls across different frameworks (i.e. netcoreapp1.0 and netcoreapp1.1). This started being an issue since PR #7578 caused the mscorlib facade to type forward some types to System.IO. The problem with that, is that in netcoreapp1.0 world, System.IO.dll implementation type-forwards System.IO.Stream to mscorlib.dll facade, which used to type-forward it System.Private.CoreLib.dll where the implementation lives, but that is not the case anymore, and it now type-forwards it back to System.IO causing a type not found.

```
Resolution of System.IO.Stream type
netcoreapp1.1: System.IO(v.4.2.0.0) -> System.Runtime(v.4.2.0.0) -> System.Private.CoreLib (type lives here)
netcoreapp1.0: System.IO(v.4.1.0.0) -> mscorlib -> System.IO(v.4.1.0)....
```

This causes:

```
Unhandled Exception: System.TypeLoadException: Could not load type 'System.IO.Stream' from assembly 'System.IO, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
```

This would get fixed if we version the mscorlib facade

jkotas:
Fixes #7607


jkotas:
cc @AlexGhiondea @danmosemsft 

There may still be things that need fixing ...

danmosemsft:
@weshaggard pointed out the CLR tests use it, unless @RussKeldorph made the change that we were talking about to refer to the CoreFX package. Will CI show that if it's the case?

danmosemsft:
LGTM. Is it worth a grep for other references to "mscorlib" in the repo? There are also mscorlib.shared.sources.props, src\mscorlib.Friends.cs etc. Very optional...

jkotas:
> Is it worth a grep for other references to "mscorlib" in the repo?

Agree. Number of these should be renamed ... e.g. the top level directory should be renamed to S.P.CoreLib, but it would make for interesting conflict with your cleanup changes ...

danmosemsft:
`cp: cannot stat ‘./bin/Product/Linux.armel.Debug/mscorlib.dll’: No such file or directory`

you missed
`https://github.com/dotnet/coreclr/blob/165fa5cebf0b937f74855584ca5b658e7e83c41b/tests/scripts/arm32_ci_script.sh#L250`

jkotas:
Waiting for https://github.com/dotnet/corefx/pull/15643 to propagate through the system

danmosemsft:
I believe this is blocked on @wtgodbe https://github.com/dotnet/coreclr/issues/8937

jkotas:
> I believe this is blocked on @wtgodbe #8937

Right.

jkotas:
I will open a new PR once this is unblocked.

wtgodbe:
Working on that right now, hoping to have a PR up today or tomorrow.

PR #9279, opened 2017-02-02
Some more perf tweaks to ThreadPool

stephentoub:
Main changes:
- The work-stealing system was storing all of the local lists in a sparse array.  Any time a thread went looking for work, it would pick a random place to start in the array and iterate through all of the cells.  But the array often was much larger than the number of lists it contained, because it a) grows doubling each time, and b) never shrinks when threads are removed, so all of the iterating needed to check for null on every cell.  And all of the reads were done with Volatile.Read, to account for potentially concurrent writes as threads were added/removed.  However, the addition/removal of thread-local lists is actually relatively rare (even when the thread pool takes threads in and out of service, it's not generally not allocating new threads or destroying old ones, just keeping them in reserve on the side for a while).  Instead, we can just use an immutable array that's always the right size.
- All of the old StackCrawlMark was still cluttering up {Unsafe}QueueUserWorkItem and preventing those functions and helpers from getting inlined.  I've removed that goo.
- Random.Next is used to determine where a thread should start looking for work to steal, and it was showing up in a trace as a few percentage points.  I've replaced it with a faster routine.
- Cleaned up some empty try blocks and the associated code around it, and helped a few other functions that were showing up in traces be more inlineable.

cc: @jkotas, @kouvel, @benaadams

### DIFF
--- a/src/mscorlib/src/System/Threading/ThreadPool.cs
+++ b/src/mscorlib/src/System/Threading/ThreadPool.cs
@@ -469,14 +469,15 @@ namespace System.Threading
                 !workItems.TryDequeue(out callback)) // then try the global queue
             {
                 // finally try to steal from another thread's local queue
-                WorkStealingQueue[] otherQueues = WorkStealingQueueList.Queues;
-                int c = otherQueues.Length;
+                WorkStealingQueue[] queues = WorkStealingQueueList.Queues;
+                int c = queues.Length;
+                Debug.Assert(c > 0, "There must at least be a queue for this thread.");
                 int maxIndex = c - 1;
                 int i = tl.random.Next(c);
                 while (c > 0)
                 {
                     i = (i < maxIndex) ? i + 1 : 0;
-                    WorkStealingQueue otherQueue = otherQueues[i];
+                    WorkStealingQueue otherQueue = queues[i];
                     if (otherQueue != wsq)
                     {
                         callback = otherQueue.TrySteal(ref missedSteal);
@@ -532,10 +533,6 @@ namespace System.Threading
                 //
                 while ((Environment.TickCount - quantumStartTime) < ThreadPoolGlobals.TP_QUANTUM)
                 {
-                    //
-                    // Dequeue and EnsureThreadRequested must be protected from ThreadAbortException.  
-                    // These are fast, so this will not delay aborts/AD-unloads for very long.
-                    //
                     bool missedSteal = false;
                     workItem = workQueue.Dequeue(tl, ref missedSteal);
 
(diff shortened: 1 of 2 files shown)


src/mscorlib/src/System/Threading/ThreadPool.cs
```diff
@@ -534,81 +536,66 @@ internal static bool Dispatch()
                     // Dequeue and EnsureThreadRequested must be protected from ThreadAbortException.
```
jkotas: Comment not applicable anymore

src/mscorlib/src/System/Threading/ThreadPool.cs
```diff
@@ -644,6 +631,24 @@ internal static bool Dispatch()
         }
     }
 
+    // Simple random number generator. We don't need great randomness, we just need a little and for it to be fast.
+    internal sealed class FastRandom // xorshift prng
```
jkotas: struct?

src/mscorlib/src/System/Threading/ThreadPool.cs
stephentoub: Removed

src/mscorlib/src/System/Threading/ThreadPool.cs
stephentoub: Changed to a struct

stephentoub:
@jkotas, I've not seen this error before (in the arm Cross Debug Build leg):
```
error C3859: virtual memory range for PCH exceeded; please recompile with a command line option of '-Zm58' or greater [D:\j\workspace\arm_cross_deb---74b8368e\bin\obj\Windows_NT.arm.Debug\crossgen\src\vm\crossgen\cee_crossgen.vcxproj]
```
I assume this has nothing to do with my PR?

jkotas:
I do not think it has anything to do with your PR. I have not seen it before.

stephentoub:

src/mscorlib/src/System/Threading/ThreadPool.cs
```diff
@@ -467,15 +469,15 @@ public IThreadPoolWorkItem Dequeue(ThreadPoolWorkQueueThreadLocals tl, ref bool
...
             {
                 // finally try to steal from another thread's local queue
-                WorkStealingQueue[] otherQueues = allThreadQueues.Current;
+                WorkStealingQueue[] otherQueues = WorkStealingQueueList.Queues;
                 int c = otherQueues.Length;
                 int maxIndex = c - 1;
                 int i = tl.random.Next(c);
                 while (c > 0)
                 {
                     i = (i < maxIndex) ? i + 1 : 0;
```
benaadams: Will this have a very high chance of starting always at zero?

benaadams: nvm it won't :)

src/mscorlib/src/System/Threading/ThreadPool.cs
```diff
@@ -644,6 +627,30 @@ internal static bool Dispatch()
...
+        }
+
+        public int Next(int maxValue)
+        {
+            uint t = _x ^ (_x << 11);
+            _x = _y; _y = _z; _z = _w;
+            _w = _w ^ (_w >> 19) ^ (t ^ (t >> 8));
+
+            int r = (int)_w & int.MaxValue;
+            return (int)(r * (1.0 / ((double)int.MaxValue + 1)) * maxValue);
```
benaadams: As much as `idiv` bothers me; I wonder what the effect of casting to double doing math and casting back; since vectors are increasingly in play; may be better just to `% maxValue`? /cc @mellinoe

benaadams:
LGTM

src/mscorlib/src/System/Threading/ThreadPool.cs
stephentoub: I'll switch to mod.  Looks like it might be 1-2% faster, but regardless it's simpler.

benaadams: I still wonder if light coordination between the threads using interlocked might be better than random to avoid collisions

As in rather than
`int i = tl.random.Next(c);`
do
`int i = NextCounter.Next(c);`

from https://gist.github.com/benaadams/9a66811bc54e7126a9cd8c181bbe6f53

Maybe something to look at in future.

stephentoub: Thanks, Ben.  Please let me know what you find.

PR #9284, opened 2017-02-02
Fix field type for ByReference<T> and TypedReference

### LINKED ISSUE #7894: Span<T>.Slice method is hitting asserts in JIT on Unix

kouvel:
Fixes #7894

### DIFF
--- a/src/vm/jitinterface.cpp
+++ b/src/vm/jitinterface.cpp
@@ -9108,11 +9108,7 @@ CorInfoType CEEInfo::getFieldTypeInternal (CORINFO_FIELD_HANDLE fieldHnd,
     if (type == ELEMENT_TYPE_I)
     {
         PTR_MethodTable enclosingMethodTable = field->GetApproxEnclosingMethodTable();
-        if (enclosingMethodTable->IsByRefLike() &&
-            (
-                enclosingMethodTable == g_TypedReferenceMT && field->GetOffset() == 0 ||
-                enclosingMethodTable->HasSameTypeDefAs(g_pByReferenceClass)
-            ))
+        if (enclosingMethodTable->IsByRefLike() && enclosingMethodTable->HasSameTypeDefAs(g_pByReferenceClass))
         {
             _ASSERTE(field->GetOffset() == 0);
             return CORINFO_TYPE_BYREF;
--- a/src/vm/jitinterface.cpp
+++ b/src/vm/jitinterface.cpp
@@ -9119,33 +9119,30 @@ CorInfoType CEEInfo::getFieldTypeInternal (CORINFO_FIELD_HANDLE fieldHnd,
         }
     }
 
-    if (type != ELEMENT_TYPE_BYREF)
+    if (type != ELEMENT_TYPE_BYREF && !CorTypeInfo::IsPrimitiveType(type))
     {
+        PCCOR_SIGNATURE sig;
+        DWORD sigCount;
+        CorCallingConvention conv;
+
+        field->GetSig(&sig, &sigCount);
+
+        conv = (CorCallingConvention)CorSigUncompressCallingConv(sig);
+        _ASSERTE(isCallConv(conv, IMAGE_CEE_CS_CALLCONV_FIELD));
+
+        SigPointer ptr(sig, sigCount);
+
         // For verifying code involving generics, use the class instantiation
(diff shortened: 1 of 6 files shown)


kouvel:
@jkotas @sivarv

src/vm/methodtablebuilder.cpp
```diff
@@ -10235,6 +10236,12 @@ void MethodTableBuilder::CheckForSystemTypes()
             pMT->ClearFlag(MethodTable::enum_flag_Category_Mask);
             pMT->SetInternalCorElementType(ELEMENT_TYPE_VALUETYPE);
 #endif
+
+            // Fix the element type of the by-ref field
+            _ASSERTE(pMT->GetNumIntroducedInstanceFields() == 1);
+            PTR_FieldDesc byRefFieldDesc = pMT->GetFieldDescByIndex(0);
+            _ASSERTE(byRefFieldDesc->GetFieldType() == ELEMENT_TYPE_I);
+            byRefFieldDesc->SetFieldType(ELEMENT_TYPE_BYREF);
```
jkotas: I think it would be better to compensate for this on the JIT-EE interface just to make the JIT happy. It is very hard to make this self-consistent in the VM, e.g. the parts of the VM that access metadata directly will still see the IntPtr and not byref.

kouvel: Done

src/vm/jitinterface.cpp
```diff
@@ -9105,30 +9105,47 @@ CorInfoType CEEInfo::getFieldTypeInternal (CORINFO_FIELD_HANDLE fieldHnd,
...
-    if (!CorTypeInfo::IsPrimitiveType(type))
+    if (type != ELEMENT_TYPE_BYREF)
     {
-        PCCOR_SIGNATURE sig;
-        DWORD sigCount;
-        CorCallingConvention conv;
+        // For verifying code involving generics, use the class instantiation
+        // of the optional owner (to provide exact, not representative,
+        // type information)
+        SigTypeContext typeContext(field, (TypeHandle)owner);
```
jkotas: The SigTypeContext constructor does not have any interesting side-effects. You can move it right in front of GetTypeHandleThrowing where it is used.

kouvel: Done

src/vm/jitinterface.cpp
```diff
@@ -9105,30 +9105,47 @@ CorInfoType CEEInfo::getFieldTypeInternal (CORINFO_FIELD_HANDLE fieldHnd,
...
+    {
+        PTR_MethodTable enclosingMethodTable = field->GetApproxEnclosingMethodTable();
+        if (enclosingMethodTable->IsByRefLike() &&
+            (
+                enclosingMethodTable == g_TypedReferenceMT && field->GetOffset() == 0 ||
+                enclosingMethodTable->HasSameTypeDefAs(g_pByReferenceClass)
+            ))
+        {
+            _ASSERTE(field->GetOffset() == 0);
+            type = ELEMENT_TYPE_BYREF;
```
jkotas: I think it would be better to just do early out here - `if (pTypeHnd != NULL) *pTypeHnd = NULL; return CORINFO_TYPE_BYREF`, so that it is obvious what happens in this case. It is not easy to follow it through `asCorInfoType`.

kouvel: At the top of the function it already does `*pTypeHnd = 0`, so I guess it's not null currently. Should it check for null?

kouvel: Since `typeHnd.IsNull()` in `CEEInfo::asCorInfoType` would always be true here, changed to shortcut anyway.

jkotas:
Looks good so far ... crossgen S.P.CoreLib is still crashing though.

kouvel:
It doesn't seem to like returning ByRef for `TypedReference`. Don't know why, and I suppose that must be handled already somehow, but this seems to work.

jkotas:
Make sense. I was kind of surprised that you are able to get away with doing it for TypedReference as well.

PR #9286, opened 2017-02-02
[x86/Linux] Revise QuickUnwindStackFrame

### LINKED ISSUE #9285: [x86/Linux] Assert failure while printing unhandled exception message
In x86/Linux, Core CLR shows the following assert failure while printing unhandled exception message:
```
Unhandled Exception:
Assert failure(PID 3195 [0x00000c7b], Thread: 3195 [0x0c7b]): flag == UnwindCurrentStackFrame
    File: /home/parjong/projects/dotnet/coreclr/src/vm/eetwain.cpp Line: 3124
```

parjong:
This commit revises QuickUnwindStackFrame similarly as the corresponding code for AMD64 to fix #9285.

### DIFF
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -3118,7 +3118,7 @@ void EECodeManager::QuickUnwindStackFrame(PREGDISPLAY pRD, StackwalkCacheEntry *
     _ASSERTE(pCacheEntry);
     _ASSERTE(GetControlPC(pRD) == (PCODE)(pCacheEntry->IP));
 
-#if defined(_TARGET_X86_)
+#if defined(_TARGET_X86_) && !defined(WIN64EXCEPTIONS)
     _ASSERTE(flag == UnwindCurrentStackFrame);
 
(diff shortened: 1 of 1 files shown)


parjong:
\CC @seanshpark @wateret

parjong:
@jkotas Please take a look.

jkotas:
This is old performance optimization that was important for partial trust execution environments. We do not have these partial trust execution environments anymore and this performance optimization is not important. It also explains why this method just returns NULL for arm and arm64, and everything works fine with it.

It would be better to not do anything here and just return NULL. It will save from chasing bugs in this code.

parjong:
@jkotas Could you let me know some details? Do you mean that we do nothing inside ``QuickUnwindStackFrame``, or we should not invoke ``QuickUnwindStackFrame``?

jkotas:
We should not invoke QuickUnwindStackFrame at all, e.g. make `StackwalkCache::s_Enabled` to be unconditionally false to Linux x86.

parjong:
Thank you for answer. If so, it would be better to make another PR. I'll try that approach, and submit another PR.

PR #9290, opened 2017-02-02
[x86/Linux] Do NOT use GSCookie on Funclet

parjong:
This commit suppresses GSCookieAccess for funclets for x86/Linux as in other architectures.


parjong:
\CC @seanshpark @wateret

PR #9298, opened 2017-02-03
Rejit support for R2R - WIP

noahfalk:
Attempting to use Rejit with R2R rapidly hits two issues:
a) The prestub never communicates to the RejitManager that new code from an R2R image is about to be used. This causes pending requests to rejit a method never
to be honored. Fixed in prestub.cpp

b) ICorProfilerInfo6::EnumNgenMethodsInliningThisMethod didn't work with R2R images because R2R images did not carry the inline tracking table. The majority of the changes address this part. Currently only intra-module inlining is supported. Cross module inlining via NonVersionableAttribute won't be tracked by design.

This change isn't ready to be checked in yet, but it does demonstrate the proposed file format, the layout of the code, and works well enough to pass really trivial functional tests on windows. Most of my planned work at this point is going to be subjecting it to broader testing (still on windows), fixing issues, and responding to PR feedback. Linux is on the roadmap and may work as-is, but verifying that is not part of this PR. I'm going to look into putting a better guard in place so that anyone adding cross module inlining in the future will realize the implications for the profiler. I wasn't planning to do perf investigation for size on disk as this format should be strictly smaller than NGEN's and we already accepted the perf overhead in that case.

### DIFF
--- a/src/inc/corprof.idl
+++ b/src/inc/corprof.idl
@@ -3740,15 +3740,20 @@ interface ICorProfilerInfo6 : ICorProfilerInfo5
 {
     /*
     * Returns an enumerator for all methods that 
-    * - belong to a given NGen module (inlinersModuleId) and 
+    * - belong to a given NGen or R2R module (inlinersModuleId) and 
     * - inlined a body of a given method (inlineeModuleId / inlineeMethodId). 
     *
     * If incompleteData is set to TRUE after function is called, it means that the methods enumerator 
     * doesn't contain all methods inlining a given method. 
     * It can happen when one or more direct or indirect dependencies of inliners module haven't been loaded yet.
-    * If profiler needs accurate data it should retry later when more modules are loaded (preferable on each module load).
+    * If profiler needs accurate data it should retry later when more modules are loaded (preferably on each module load).
     *
     * It can be used to lift limitation on inlining for ReJIT.
+    *
+    * NOTE: If the inlinee method is decorated with the System.Runtime.Versioning.NonVersionable attribute then
+    * then some inliners may not ever be reported. If you need to get a full accounting you can avoid the issue
+    * by disabling the use of all native images.
+    * 
     */
     HRESULT EnumNgenModuleMethodsInliningThisMethod(
         [in] ModuleID    inlinersModuleId,
--- a/src/inc/readytorun.h
+++ b/src/inc/readytorun.h
@@ -16,7 +16,8 @@
 #define READYTORUN_SIGNATURE 0x00525452 // 'RTR'
 
 #define READYTORUN_MAJOR_VERSION 0x0002
-#define READYTORUN_MINOR_VERSION 0x0000
+#define READYTORUN_MINOR_VERSION 0x0001
+// R2R Version 2.1 adds the READYTORUN_SECTION_INLINING_INFO section
 
 struct READYTORUN_HEADER
 {
@@ -57,6 +58,13 @@ enum ReadyToRunSectionType
     // 107 used by an older format of READYTORUN_SECTION_AVAILABLE_TYPES
     READYTORUN_SECTION_AVAILABLE_TYPES              = 108,
     READYTORUN_SECTION_INSTANCE_METHOD_ENTRYPOINTS  = 109,
+    READYTORUN_SECTION_INLINING_INFO                = 110  // Added in V2.1
+
+	// If you add a new section consider whether it is a breaking or non-breaking change.
+	// Usually it is non-breaking, but if it is preferable to have older runtimes fail
+	// to load the image vs. ignoring the new section it could be marked breaking.
+	// Increment the READYTORUN_MINOR_VERSION (non-breaking) or READYTORUN_MAJOR_VERSION
+	// (breaking) as appropriate.
 };
 
 //
--- a/src/vm/ceeload.cpp
+++ b/src/vm/ceeload.cpp
@@ -183,10 +183,32 @@ ARRAY_PTR_COR_IL_MAP InstrumentedILOffsetMapping::GetOffsets() const
     return m_rgMap;
 }
 
-PTR_PersistentInlineTrackingMap Module::GetNgenInlineTrackingMap()
+BOOL Module::HasInlineTrackingMap()
 {
     LIMITED_METHOD_DAC_CONTRACT;
-    return m_persistentInlineTrackingMap;
+#ifdef FEATURE_READYTORUN
+	if (IsReadyToRun() && GetReadyToRunInfo()->GetInlineTrackingMap() != NULL)
+	{
+		return TRUE;
+	}
+#endif
+    return (m_pPersistentInlineTrackingMapNGen != NULL);
+}
+
+COUNT_T Module::GetInliners(PTR_Module inlineeOwnerMod, mdMethodDef inlineeTkn, COUNT_T inlinersSize, MethodInModule inliners[], BOOL *incompleteData)
+{
+    WRAPPER_NO_CONTRACT;
+#ifdef FEATURE_READYTORUN
+    if(IsReadyToRun() && GetReadyToRunInfo()->GetInlineTrackingMap() != NULL)
+    {
+        return GetReadyToRunInfo()->GetInlineTrackingMap()->GetInliners(inlineeOwnerMod, inlineeTkn, inlinersSize, inliners, incompleteData);
+    }
+#endif
+    if(m_pPersistentInlineTrackingMapNGen != NULL)
+    {
+        return m_pPersistentInlineTrackingMapNGen->GetInliners(inlineeOwnerMod, inlineeTkn, inlinersSize, inliners, incompleteData);
+    }
+    return 0;
 }
 
 
@@ -10064,8 +10086,8 @@ void Module::Save(DataImage *image)
     InlineTrackingMap *inlineTrackingMap = image->GetInlineTrackingMap();
     if (inlineTrackingMap) 
     {
-        m_persistentInlineTrackingMap = new (image->GetHeap()) PersistentInlineTrackingMap(this);
-        m_persistentInlineTrackingMap->Save(image, inlineTrackingMap);
+        m_pPersistentInlineTrackingMapNGen = new (image->GetHeap()) PersistentInlineTrackingMapNGen(this);
+        m_pPersistentInlineTrackingMapNGen->Save(image, inlineTrackingMap);
     }
 
     if (m_pNgenStats && g_CorCompileVerboseLevel >= CORCOMPILE_STATS)
@@ -11064,14 +11086,14 @@ void Module::Fixup(DataImage *image)
     }
 
     // Fix up inlining data
-    if(m_persistentInlineTrackingMap)
+    if(m_pPersistentInlineTrackingMapNGen)
     {
-        image->FixupPointerField(this, offsetof(Module, m_persistentInlineTrackingMap));
-        m_persistentInlineTrackingMap->Fixup(image);
+        image->FixupPointerField(this, offsetof(Module, m_pPersistentInlineTrackingMapNGen));
+        m_pPersistentInlineTrackingMapNGen->Fixup(image);
     } 
     else
     {
-        image->ZeroPointerField(this, offsetof(Module, m_persistentInlineTrackingMap));
+        image->ZeroPointerField(this, offsetof(Module, m_pPersistentInlineTrackingMapNGen));
     }
 
     SetIsModuleSaved();
--- a/src/vm/inlinetracking.cpp
+++ b/src/vm/inlinetracking.cpp
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 // =============================================================================================
-// Code for tracking method inlinings in NGen images.
+// Code for tracking method inlinings in NGen and R2R images.
 // The only information stored is "who" got inlined "where", no offsets or inlining depth tracking. 
 // (No good for debugger yet.)
 // This information is later exposed to profilers and can be useful for ReJIT.
@@ -11,6 +11,8 @@
 #include "common.h"
 #include "inlinetracking.h"
 #include "ceeload.h"
+
+#ifndef DACCESS_COMPILE
 
 bool MethodInModule::operator <(const MethodInModule& other) const
 {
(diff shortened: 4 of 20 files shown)


noahfalk:
@jkotas @fadimounir - PTAL

@gkhanna79 - Gaurav are there additional people on your team that should be included on this PR?

Thanks!

fadimounir:
>The prestub never communicates to the RejitManager that new code from an R2R image is about to be used. This causes pending requests to rejit a method never
to be honored

@noahfalk could you please give me some context on what this rejit feature is? What do we rejit and why? Is it because we can't use AOT code from the R2R image for some reason? etc...

gkhanna79:
CC @rahku @jkotas

src/inc/readytorun.h
```diff
@@ -57,6 +57,7 @@ enum ReadyToRunSectionType
     // 107 used by an older format of READYTORUN_SECTION_AVAILABLE_TYPES
     READYTORUN_SECTION_AVAILABLE_TYPES              = 108,
     READYTORUN_SECTION_INSTANCE_METHOD_ENTRYPOINTS  = 109,
+    READYTORUN_SECTION_INLINING_INFO                = 110
```
fadimounir: >READYTORUN_SECTION_INLINING_INFO [](start = 4, length = 32)

@jkotas Does this need a change the the R2R minor version?

src/vm/ceeload.cpp
```diff
@@ -187,10 +187,29 @@ ARRAY_PTR_COR_IL_MAP InstrumentedILOffsetMapping::GetOffsets() const
     return m_rgMap;
 }
 
-PTR_PersistentInlineTrackingMap Module::GetNgenInlineTrackingMap()
+BOOL Module::HasInlineTrackingMap()
 {
     LIMITED_METHOD_DAC_CONTRACT;
-    return m_persistentInlineTrackingMap;
+#ifdef FEATURE_READYTORUN
+    if(IsReadyToRun() && GetReadyToRunInfo()->GetInlineTrackingMap() != NULL) return TRUE;
```
fadimounir: >return TRUE; [](start = 78, length = 12)

Nit: new line :)

src/inc/readytorun.h
jkotas: Yes, I think it would be a good idea to bump the R2R minor version for this.

fadimounir:
You should have a test (at least a manual one before merging the PR) where you crossgen an assembly to produce a R2R using an old crossgen without any of your changes, and see how this loads with your runtime changes.
And vice versa, you should ensure that new R2R images with your changes either work with an older runtime, or get rejected by it (but not crash it)

src/vm/inlinetracking.h
```diff
@@ -170,63 +199,181 @@ typedef DPTR(InlineTrackingMap) PTR_InlineTrackingMap;
...
+    Module *GetModuleByIndex(DWORD index);
+
+};
+
+typedef DPTR(PersistentInlineTrackingMapNGen) PTR_PersistentInlineTrackingMapNGen;
+
+
+// This type know how to serialize and deserialize the inline tracking map format within an R2R image. See
+// above for a description of the format.
+class PersistentInlineTrackingMapR2R
```
fadimounir: >PersistentInlineTrackingMapR2R [](start = 6, length = 30)

Maybe this should be under the FEATURE_READYTORUN?

src/vm/readytoruninfo.h
```diff
@@ -40,6 +41,9 @@ class ReadyToRunInfo
     Crst                            m_Crst;
     PtrHashMap                      m_entryPointToMethodDescMap;
 
+    PersistentInlineTrackingMapR2R  m_persistentInlineTrackingMap;
+    PTR_PersistentInlineTrackingMapR2R m_pPersistentInlineTrackingMap;
```
fadimounir: I'm personally not a big fan of this... Maybe you should make the .TryLoad function a static one that returns a pointer to a heap allocated map. Feels a bit like a hack to me

src/vm/inlinetracking.cpp
```diff
@@ -123,9 +125,72 @@ InlineTrackingEntry & InlineTrackingEntry::operator = (const InlineTrackingEntry
...
 
+void InlineTrackingEntry::Add(PTR_MethodDesc inliner)
+{
+    STANDARD_VM_CONTRACT;
+
+    MethodInModule method(inliner->GetModule(), inliner->GetMemberDef());
 
-#ifndef DACCESS_COMPILE
-COUNT_T PersistentInlineTrackingMap::GetInliners(PTR_Module inlineeOwnerMod, mdMethodDef inlineeTkn, COUNT_T inlinersSize, MethodInModule inliners[], BOOL *incompleteData)
+    // Going through last 10 inliners to check if a given inliner has recently been registered.
```
fadimounir: >10 [](start = 26, length = 2)

What is this number based on? just curious :). Any data?

src/vm/inlinetracking.cpp
```diff
@@ -418,12 +424,181 @@ void PersistentInlineTrackingMap::Save(DataImage *image, InlineTrackingMap* runt
...
+#endif //FEATURE_NATIVE_IMAGE_GENERATION
+
+struct InliningHeader
+{
+    short MajorVersion;
+    short MinorVersion;
+    int SizeOfInlineeIndex;
+};
+
+#ifdef FEATURE_NATIVE_IMAGE_GENERATION
```
fadimounir: >FEATURE_NATIVE_IMAGE_GENERATION [](start = 7, length = 31)

shouldn't we also have FEATURE_READYTORUN here?

src/vm/inlinetracking.cpp
```diff
@@ -418,12 +424,181 @@ void PersistentInlineTrackingMap::Save(DataImage *image, InlineTrackingMap* runt
...
+
+void PersistentInlineTrackingMapR2R::ProcessInlineTrackingEntry(SBuffer *inlinersBuffer, SArray<InlineeRecordR2R> *inlineeIndex, InlineTrackingEntry *entry)
+{
+    STANDARD_VM_CONTRACT;
+    // This call removes duplicates from inliners and makes sure they are sorted
+    entry->SortAndDeduplicate();
+    MethodInModule inlinee = entry->m_inlinee;
+    InlineSArray<MethodInModule, 3> &inliners = entry->m_inliners;
+    COUNT_T tatalInlinersCount = inliners.GetCount();
```
fadimounir: >tatalInlinersCount [](start = 12, length = 18)

typo

fadimounir:
>     InlineTrackingEntry **inlinees = new InlineTrackingEntry *[runtimeMapCount];

We should probably check for null allocations (out of memory cases)

---
Refers to: src/vm/inlinetracking.cpp:382 in c761c11. [](commit_id = c761c11bdaea71aaaace73438278067011609af8, deletion_comment = False)

src/vm/inlinetracking.cpp
```diff
@@ -418,12 +424,181 @@ void PersistentInlineTrackingMap::Save(DataImage *image, InlineTrackingMap* runt
...
+    _ASSERTE(pSaveTarget != NULL);
+    _ASSERTE(runtimeMap != NULL);
+
+    SArray<InlineeRecordR2R> inlineeIndex;
+    SBuffer inlinersBuffer;
+
+    // Sort records from runtimeMap, because we need to make sure 
+    // we save everything in deterministic order. Hashtable iteration is not deterministic.
+    COUNT_T runtimeMapCount = runtimeMap->GetCount();
+    InlineTrackingEntry **inlinees = new InlineTrackingEntry *[runtimeMapCount];
```
fadimounir: >new InlineTrackingEntry *[runtimeMapCount]; [](start = 37, length = 43)

check for null allocation (OOM)

src/vm/inlinetracking.cpp
```diff
@@ -418,12 +424,181 @@ void PersistentInlineTrackingMap::Save(DataImage *image, InlineTrackingMap* runt
...
+    // Iterate throught each inlinee record from the InlineTrackingMap
+    // and write corresponding records into inlineeIndex and inlinersBuffer
+    for (COUNT_T i = 0; i < runtimeMapCount; i++)
+    {
+        ProcessInlineTrackingEntry(&inlinersBuffer, &inlineeIndex, inlinees[i]);
+    }
+
+    InliningHeader header;
+    header.MajorVersion = 1;
+    header.MinorVersion = 0;
```
fadimounir: It would be nice to have these versions declared in #define directives so that updating them in the future would be easier (including reading them in TryLoad)

src/vm/inlinetracking.cpp
```diff
@@ -418,12 +424,181 @@ void PersistentInlineTrackingMap::Save(DataImage *image, InlineTrackingMap* runt
...
+    InlineeRecordR2R record(RidFromToken(inlinee.m_methodDef));
+    DWORD inlinersStreamSize;
+    const BYTE *inlinersStreamPtr = (const BYTE *)inlinersStream.GetBlob(&inlinersStreamSize);
+    record.m_offset = inlinersBuffer->GetSize();
+    inlinersBuffer->Insert(inlinersBuffer->End(), SBuffer(SBuffer::Immutable, inlinersStreamPtr, inlinersStreamSize));
+
+    inlineeIndex->Append(record);
+}
+
+void PersistentInlineTrackingMapR2R::Save(SBuffer* pSaveTarget, InlineTrackingMap* runtimeMap)
```
fadimounir: >Save [](start = 37, length = 4)

Nit: Ideally, we should share the pre-saving logic between this R2R and the NGEN persistent inline tracking map, and just have the Save method do the acutal writing (the pre-saving logic is the same in both case)

src/vm/inlinetracking.cpp
```diff
@@ -418,12 +424,181 @@ void PersistentInlineTrackingMap::Save(DataImage *image, InlineTrackingMap* runt
...
+        "PersistentInlineTrackingMap saved. InlineeIndexSize: %d bytes, InlinersBufferSize: %d bytes\n",
+        header.SizeOfInlineeIndex, inlinersBuffer.GetSize()));
 }
+
 #endif //FEATURE_NATIVE_IMAGE_GENERATION
+
+BOOL PersistentInlineTrackingMapR2R::TryLoad(Module* pModule, const BYTE* pBuffer, DWORD cbBuffer)
+{
+    InliningHeader* pHeader = (InliningHeader*)pBuffer;
+    if(pHeader->MajorVersion != 1)
```
fadimounir: >MajorVersion [](start = 16, length = 12)

Have you considered using the R2R major and minor versions instead of declaring new ones? (also seems like MinorVersion is not used)

fadimounir:
The general direction looks good to me.

Do you think it would be better to have the 2 implementations for NGEN and R2R separated by #ifdefs and avoid having the NGEN and R2R suffixes to class names?

noahfalk:
Historical background on rejit:

Rejit is a profiler feature that we introduced in .Net 4.0 (or 4.5 I can't recall?). The intention is to allow a profiler writer to change the implementation of a method at any time during the execution of a process. The profiler does this by indicating which Modules/MethodTokens need to be updated in ICorProfilerInfo4::RequestRejit, then providing a new IL body in IProfilerControl::SetILFunctionBody. The IL is usually a close derivative of the previous IL, but it doesn't have to be. The runtime takes this IL, uses the JIT to compile it and places a jump instruction on the beginning of the old method body to send execution to the new method body. A simple scenario where this might be useful - a developer is running an app in production and they want to log the value of a local variable in the middle of a function without rebuilding and redeploying a new application. Using rejit the developer could ask the profiler to change the implementation so the IL includes a call to a logging statement that was not there before.

In 4.5.1 (I think?) we added support for rejit in NGEN images, previously it was only supported for runtime jitted code. Part of that additional work was exposing the inlining information. This was required because if you wanted to change the implementation of method A, and the IL of method A has already been inlined inside method B, then you need to recompile B in order to get the new behavior.

When R2R was created nobody hooked it up to the rejit mechanism in the same way that NGEN was, so scenario-wise profilers lost the ability to change a bunch of framework code when moving from desktop to .Net Core. This feature aims to restore that ability.

src/inc/readytorun.h
noahfalk: Will do

src/vm/ceeload.cpp
noahfalk: Will do

noahfalk:
Re: testing comment above - yes I plan to do both the forward and back-compat tests (new image/old runtime) + (old image/new runtime)

src/vm/inlinetracking.h
noahfalk: Will do

src/vm/readytoruninfo.h
noahfalk: Mostly I just did it because the machinery to make new allocations looks like it requires some additional arguments to be passed around and the map datastructure is only a few pointers in size, however if I am breaking form I can certainly do a seperate allocation.

src/vm/inlinetracking.cpp
noahfalk: This number didn't actually come from me, I was simply rearranging the file a bit so that methods on different types wouldn't be intermingled. The original author was Eugene and I think he came up with this number and several others empirically based on some ad-hoc experiements/guessing reasonable values.

src/vm/inlinetracking.cpp
noahfalk: will do

src/vm/inlinetracking.cpp
noahfalk: will do

src/vm/inlinetracking.cpp
noahfalk: Will do (though if it alarms anyone this is a pre-existing bug in the NGEN variant of the code that got replicated here)

src/vm/inlinetracking.cpp
noahfalk: will do

src/vm/inlinetracking.cpp
noahfalk: Is there a document or comments somewhere that describe the r2r versioning model? I'm not at all attached to these version checks if I can piggy back on a broader versioning boundary.,

src/vm/inlinetracking.cpp
noahfalk: Sure, I'll confirm that I can merge a bit more than we've got so far.

noahfalk:
> Do you think it would be better to have the 2 implementations for NGEN and R2R separated by #ifdefs and avoid having the NGEN and R2R suffixes to class names?

I don't think we can. We need to support both the mscorlib image which is NGEN'ed and the other framework images which are R2R using the same build of coreclr.

src/vm/inlinetracking.cpp
fadimounir: @gkhanna79 or @jkotas could you please point Noah to the r2r design doc? I can't remember where it was

noahfalk: I made some changes for versioning already guessing that we are using the standard major = breaking, minor = non-breaking and added comments to that effect. If that all looks good then I don't need the doc, but if I guessed incorrectly then yes the doc could be very useful.

src/vm/inlinetracking.cpp
noahfalk: Looking around at numerous other allocation sites in NGEN/R2R paths I am not seeing any that do NULL checks. It appears the common pattern is "new (image->GetHeap()) InlineTrackingEntry *[runtimeMapCount]" that will throw an out of memory exception automatically on failure to allocate.

noahfalk:
@fadimounir - I've taken a stab at addressing all the feedback, but keep it coming as needed. I still have a few small changes of my own planned and testing is continuing. Thanks!

src/inc/corprof.idl
```diff
@@ -3740,15 +3740,20 @@ interface ICorProfilerInfo6 : ICorProfilerInfo5
...
     *
     * If incompleteData is set to TRUE after function is called, it means that the methods enumerator 
     * doesn't contain all methods inlining a given method. 
     * It can happen when one or more direct or indirect dependencies of inliners module haven't been loaded yet.
-    * If profiler needs accurate data it should retry later when more modules are loaded (preferable on each module load).
+    * If profiler needs accurate data it should retry later when more modules are loaded (preferably on each module load).
     *
     * It can be used to lift limitation on inlining for ReJIT.
+    *
+    * NOTE: If the inlinee method is decorated with the System.Runtime.Versioning.NonVersionable attribute then
```
mjsabby: I would more precisely call out that either NonVersionable or the profiler author explicitly disables all native images.

noahfalk: hmm, not sure I understand your suggestion? Are you suggesting adding a comment like this: "Inlining results for methods marked NonVersionable may not be complete. If you need to get an full accounting you may avoid the issue by disabling the use of all native images"

mattwarren:
> Rejit is a profiler feature that we introduced in .Net 4.0 (or 4.5 I can't recall?). 

According to [David Broman's blog](https://blogs.msdn.microsoft.com/davbr/2011/10/12/rejit-a-how-to-guide/) ReJIT was added in .NET 4.5

src/inc/corprof.idl
mjsabby: Correct, since I'm assuming this doc will go on MSDN.

noahfalk: Updated

noahfalk:
Unless I hear otherwise by Monday afternoon I'm going to assume this PR is in good shape. I'll squash and merge at that time. Thanks all!

fadimounir:
LGTM

PR #9300, opened 2017-02-03
[x86/Linux] Initial patch to restore EBP for funclet

seanshpark:
This is a initial patch to restore EBP and Exception object for funclets
With this patch it fixes lots of try-catch fails like PinObj-neg.exe
Further changes maybe needed for nested try-catch.

### DIFF
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -10265,7 +10265,52 @@ void CodeGen::genCaptureFuncletPrologEpilogInfo()
 
 /*****************************************************************************
  *
- *  Generates code for an EH funclet prolog.
+ *  Generates code for an EH funclet prolog. WIP.
+ *
+ *  Funclets have the following incoming arguments:
+ *
+ *      catch/filter-handler: ebp+8 = CallerSP, ebp+12 = the exception object that was caught (see GT_CATCH_ARG)
+ *      filter:               ebp+8 = CallerSP, ebp+12 = the exception object to filter (see GT_CATCH_ARG)
+ *      finally/fault:        ebp+8 = CallerSP
+ *
+ *  Funclets set the following registers on exit:
+ *
+ *      catch/filter-handler: eax = the address at which execution should resume (see BBJ_EHCATCHRET)
+ *      filter:               eax = non-zero if the handler should handle the exception, zero otherwise (see GT_RETFILT)
+ *      finally/fault:        none
+ *
+ *  The X86 funclet prolog sequence is:
+ *
+ *     push     ebp
+ *     mov      ebp, dword ptr [esp+08H]
+ *     lea      ebp, [ebp-08H]
+ *     mov      eax, dword ptr [esp+0CH]
+ *     push     callee-saved regs   ; if necessary
+ *
+ *  The epilog sequence is then:
+ *
+ *     pop callee-saved regs        ; if necessary
+ *     pop ebp
+ *     ret
+ *
+ *  The funclet frame is thus:
+ *
+ *      |                       |
+ *      |-----------------------|
+ *      |       incoming        |
+ *      |       arguments       |
+ *      +=======================+ <---- Caller's SP
+ *      |    Return address     |
+ *      |-----------------------|
+ *      |      Saved EBP        |
+ *      |-----------------------|
+ *      |Callee saved registers |
+ *      |-----------------------|
+ *      |       |               |
+ *      ~       | Stack grows   ~
+ *      |       | downward      |
+ *              V
+ *
  */
 
 void CodeGen::genFuncletProlog(BasicBlock* block)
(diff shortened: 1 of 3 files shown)


seanshpark:
Related #9245 with help and draft code from @parjong / CC @wateret .

src/jit/codegencommon.cpp
```diff
@@ -10279,12 +10326,39 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
...
+            // TODO 16 byte stack alignment
+            // TODO Fix, if needed, for nested try-catch
+
+            // mov ebp, [esp + 8]  ; get ebp from caller argument
+            getEmitter()->emitIns_R_AR(INS_mov, EA_PTRSIZE, REG_FPBASE, REG_SPBASE, 8);
+            // lea ebp, [ebp - 8]  ; load ebp from main function
+            getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_FPBASE, REG_FPBASE, -8);
+            // mov eax, [esp + 12] ; exception object to eax: funclet needs eax having the exception object
+            getEmitter()->emitIns_R_AR(INS_mov, EA_PTRSIZE, REG_EAX, REG_SPBASE, 12);
+
```
parjong: It would be better to use ``REG_EXCEPTION_OBJECT`` instead of ``REG_EAX`` (although they are same).

src/jit/codegencommon.cpp
```diff
@@ -10279,12 +10326,39 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
...
+            regSet.rsRemoveRegsModified(RBM_FPBASE);
+        }
+    }
+
+    // Save callee-saved registers
+    genPushCalleeSavedRegisters();
 }
```
parjong: It would be better to invoke genPushCalleeSavedRegisters in the middle of ``unwindBegProlog`` and ``unwindEndProlog``

src/jit/codegencommon.cpp
```diff
@@ -10265,7 +10265,54 @@ void CodeGen::genCaptureFuncletPrologEpilogInfo()
...
+ *      +=======================+ <---- Caller's SP
+ *      |    Return address     |
+ *      |-----------------------|
+ *      |      Saved EBP        |
+ *      |-----------------------|
+ *      |Callee saved registers |
+ *      |-----------------------|
+ *      |   Outgoing arg space  | // this only exists if the function makes a call
+ *      |-----------------------|
+ *      |       |               |
```
parjong: There may be no outgoing arg space as ``FEATURE_FIXED_OUT_ARGS`` is not set for x86.

src/jit/codegencommon.cpp
seanshpark: OK, thanks!

src/jit/codegencommon.cpp
seanshpark: Problem is that it will change the ESP, now knowing the amount, so EBP restore offset and also for EAX should have calculated offset not fixed 8 and 12. If genPushCalleeSavedRegisters must be in between, I need to find another way to restore EBP and EAX.

src/jit/codegencommon.cpp
seanshpark: It's for spaces when calling other methods, if any, inside funclets with "push" instruction.

seanshpark: I think this block can be removed as you noted.

src/jit/codegencommon.cpp
parjong: Or, we may skip callee-saved registers save/restore (as discussed in #9245).

seanshpark: OK, remove for now :)

src/jit/codegencommon.cpp
```diff
@@ -10265,7 +10265,52 @@ void CodeGen::genCaptureFuncletPrologEpilogInfo()
 
 /*****************************************************************************
  *
- *  Generates code for an EH funclet prolog.
+ *  Generates code for an EH funclet prolog. WIP.
+ *
+ *  Funclets have the following incoming arguments:
+ *
+ *      catch/filter-handler: ebp+8 = CallerSP, ebp+12 = the exception object that was caught (see GT_CATCH_ARG)
```
jkotas: This calling convention for funclets looks odd. Was it inspired by something?

I think the natural calling convention for funclets would be:

```
ecx = established frame
edx = exception object
ebp, esi, edi, ebx = restored to what they were when the control flow left the method last time
```

parjong: It seems that the following code in ``exceptionhandling.cpp`` invokes funclets:
```
   //
    // Invoke the funclet.
    //
    dwResumePC = pfnHandler(sf.SP, OBJECTREFToObject(throwable));
```

``pfnHandler`` is of type ``HandlerFn *``:
```
src/ $ grep -Rn 'pfnHandler'
vm/exceptionhandling.cpp:3241:    HandlerFn*          pfnHandler = (HandlerFn*)uHandlerStartPC;
```

```
src/ $ grep -Rn 'HandlerFn'
...
vm/exceptionhandling.h:35:typedef DWORD_PTR   (HandlerFn)(UINT_PTR uStackFrame, Object* pExceptionObj);
```

There is no calling convention annotation on ``HandlerFn`` which means that ``CDECL``.

jkotas: I think so.

seanshpark: JIT code looks like this. It uses EAX for the Exception object, but the object itself seems to be in the stack.
```
G_M9939_IG08:
       8945EC       mov      gword ptr [ebp-14H], eax
       8B4DEC       mov      ecx, gword ptr [ebp-14H]
       894DF4       mov      gword ptr [ebp-0CH], ecx
       90           nop      
       8B4DF4       mov      ecx, gword ptr [ebp-0CH]
       8B45F4       mov      eax, gword ptr [ebp-0CH]
       8B00         mov      eax, dword ptr [eax]
       8B402C       mov      eax, dword ptr [eax+44]
       FF5010       call     gword ptr [eax+16]System.Exception:get_Message():ref:this
```
source code:
```
            catch (Exception e)
            {
                Console.WriteLine(e.Message);
            }
```

seanshpark:
I'm closing this issue as @wateret has started with #9489

Issue #9317, opened 2017-02-03
Add support for Converts, Widen and Narrow


CarolEidt:
In support of https://github.com/dotnet/corefx/issues/1605, RyuJIT should support the new instrinsics as proposed.

PR #9318, opened 2017-02-03
Support for Vector Conversions

CarolEidt:
See #9317 
This is preliminary, and not yet ready for merging.
It includes most of the support, except:
* Unsigned conversions to/from floating point
* SSE2 support for narrowing 32 bit signed/unsigned integers
* SSE2 support for widening of unsigned types.


CarolEidt:
Notes on building:

- You need to clone the implementation from @mellinoe: https://github.com/mellinoe/corefx/tree/vectort-additions
- Then you need to build both the dll and the package (in corefx\src\System.Numerics.Vectors):
  - msbuild /p:Configuration=Release System.Numerics.Vectors.sln
  - msbuild /p:Configuration=Release pkg\System.Numerics.Vectors.pkgproj
- Get the package version number from corefx\bin\packages\Release\System.Numerics.Vectors.<version>.nupkg
- Edit the various project.json files that include System.Numerics.Vectors to change it to that version.
- In tests\dir.props - add a package source:
			     <DnuSourceList Include="\git\corefx\bin\packages\Release" />
  - Note that I had some difficulty in getting this to work correctly, and each time I would update I would have to work through the issues again. Perhaps someone else knows how to accomplish this more easily.

- Finally, when you build, you have to specify:
  - \build.cmd x64 Checked -- /p:SkipVerifyPackageVersions=true
	OR
  - set SkipVerifyPackageVersions=true
  - .\build.cmd x64 Checked

Notes on the implementation:
* I'm not sure how useful the calls to `getOpForSIMDIntrinsic()`. It might be cleaner and more efficient to do this separately for each intrinsic.
* I iterated a bit on the code sequences - I'm nearly certain they can still be improved. But in the process I added instructions to instrsxarch.h that may not be used, so it might be worth checking (and perhaps leaving them in but adding a comment in case we run out of enum space for the instructions).
I added support for encoding three registers plus immediate, and there are places (some marked with TODO's) in simdcodegenxarch.cpp where we could eliminate a copy with this encoding.

CarolEidt:
@sivarv - here is the current state
cc @dotnet/jit-contrib

jkotas:
superseded by https://github.com/dotnet/coreclr/pull/9920

Issue #9030, opened 2017-01-20
AssemblyLoadContext : issue with dynamic assemblies and custom ALC


davjohna:
There appears to be a problem with using a custom AssemblyLoadContext combined with dynamic assemblies that have a dependency on a custom ALC-loaded assembly. This issue was tested and found in .NET Core 1.0.0, 1.0.3 (LTS), and 1.1.0. A good example is using the XmlSerializer, since it generates a dynamic assembly to serialize types.
 
Steps to reproduce:
 
1. Setup two projects: XmlTest, and BadAlc.
2. For XmlTest in Program.cs, use (https://gist.github.com/davjohna/4d143cf4d73a3830f77fdc3b3ba143b9).
3. In XmlTest for project.json, use (https://gist.github.com/davjohna/281bd75991cc5868253e683b0e8883b3).
4. In BadAlc, in Program.cs, use (https://gist.github.com/davjohna/32b4699e41ebaf931a97a0b92f40afef).
4. In BadAlc, for project.json, use (https://gist.github.com/davjohna/79bb5ce2745fc90eed94257c05c6268d).
5. Do a “dotnet publish” of XmlTest, and copy all output of the bin/Debug/netcoreapp1.0/publish directory to the directory of your choice.
6. In BadAlc Program.cs,  add the location of NETCore.App 1.0.3 to line 31, and the path where you saved the XmlTest publish artifacts to line 32.
7. Do a “dotnet run” in BadAlc.
 
You should see a stack trace like the one below:
 
Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.InvalidOperationException: There is an error in XML document (0, 0). ---> System.MissingMethodException: Method not found: 'System.Xml.XmlReader System.Xml.Serialization.XmlSerializationReader.get_Reader()'.
   at Microsoft.Xml.Serialization.GeneratedAssembly.XmlSerializationReaderPerson.InitIDs()
   at System.Xml.Serialization.TempAssembly.InvokeReader(XmlMapping mapping, XmlReader xmlReader, Object events, String encodingStyle)
   at System.Xml.Serialization.XmlSerializer.Deserialize(XmlReader xmlReader, String encodingStyle, Object events)

jkotas:
cc @gkhanna79 @rahku

gkhanna79:
Why are you trying to load FX assemblies? They are automatically loaded/resolved for you if your Load implementation returns null.

[...]

PR #9320, opened 2017-02-03
Use fallback load context when dynamic assembly triggers load

gkhanna79:
Fixes https://github.com/dotnet/coreclr/issues/9030


gkhanna79:
Linux Arm Emulator build is timing out - unrelated to the change in question.

PR #9323, opened 2017-02-03
Minor perf improvement to List<T>.Add

stephentoub:
Small improvement to List.Add's code quality.  In a microbenchmark that adds lots of integers to a list (without needing to grow), this improves throughput by ~10%.

cc: @jkotas

### DIFF
--- a/src/mscorlib/src/System/Collections/Generic/List.cs
+++ b/src/mscorlib/src/System/Collections/Generic/List.cs
@@ -16,6 +16,7 @@ namespace System.Collections.Generic {
 
     using System;
     using System.Runtime;
+    using System.Runtime.CompilerServices;
     using System.Runtime.Versioning;
     using System.Diagnostics;
     using System.Diagnostics.Contracts;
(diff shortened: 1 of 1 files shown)


src/mscorlib/src/System/Collections/Generic/List.cs
```diff
@@ -210,11 +211,21 @@ public class List<T> : IList<T>, System.Collections.IList, IReadOnlyList<T>
...
+            int size = _size;
+            if (size == _items.Length) GrowCapacityForAdd();
+            _items[size] = item;
+            _size = size + 1;
             _version++;
         }
 
+        // Separated out of List.Add to improve its code quality
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private void GrowCapacityForAdd()
```
jkotas: Caching size in a local makes sense, but I would not expect that factoring out `EnsureCapacity(_size+1)` to separate method to help anything. Is `GrowCapacityForAdd` really needed?

stephentoub: On the machine I was using before, it actually did make a measurable difference, and it resulted in different asm (beyond the expected changes related to args).  I'm now on a different machine, though, and I'm not able to repro the perf difference.  I'll take another look at it over the weekend or on Monday.

stephentoub:
I'm going to close this; I've tried on a few machines now, and in some cases the new version is faster, in others it's slower.

src/mscorlib/src/System/Collections/Generic/List.cs
benaadams: I tried this; but `NoIining` was getting ignored as function was coming up as `[below ALWAYS_INLINE size]` couldn't determine if it was new behaviour or always like it. Was also doing it before the StackCrawl change, where caller would get inlined regardless if it was extremely simple.

jkotas: >  where caller would get inlined regardless if it was extremely simple.

This sounds like a bug that we should fix if you still see it. NoInlining should always mean NoInlining, no exceptions.

PR #9327, opened 2017-02-04
Remove duplicate attributes from CoreLib that are in CoreFX

shrah:
Couldn't remove LCIDConversionAttribute as it is being used in mscorlib.

@jkotas @tijoytom

### DIFF
--- a/src/mscorlib/src/System/Internal.cs
+++ b/src/mscorlib/src/System/Internal.cs
@@ -19,15 +19,7 @@ using System.StubHelpers;
 using System.Threading.Tasks;
 
 #if FEATURE_COMINTEROP
-
 using System.Runtime.InteropServices.WindowsRuntime;
-
-[assembly:Guid("BED7F4EA-1A96-11d2-8F08-00A0C9A6186D")]
-
-// The following attribute are required to ensure COM compatibility.
-[assembly:System.Runtime.InteropServices.ComCompatibleVersion(1, 0, 3300, 0)]
-[assembly:System.Runtime.InteropServices.TypeLibVersion(2, 4)]
-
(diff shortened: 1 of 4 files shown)


src/mscorlib/src/System/Internal.cs
```diff
@@ -25,7 +25,6 @@
 [assembly:Guid("BED7F4EA-1A96-11d2-8F08-00A0C9A6186D")]
 
 // The following attribute are required to ensure COM compatibility.
-[assembly:System.Runtime.InteropServices.ComCompatibleVersion(1, 0, 3300, 0)]
 [assembly:System.Runtime.InteropServices.TypeLibVersion(2, 4)]
```
jkotas: You can delete the Guid and TypeLibVersion too while you are on it,

jkotas:
> Couldn't remove LCIDConversionAttribute as it is being used in mscorlib.

So it should be delete in corefx then. 

While you are on it, I think you can also delete these from here and move them to corefx:

```
TypeLibImportClassAttribute 
ProgIdAttribute 
ImportedFromTypeLibAttribute 
ComEventInterfaceAttribute 
TypeLibVersionAttribute 
```

src/mscorlib/src/System/Internal.cs
shrah: Done.

shrah:
@jkotas, couldn't remove ProgIdAttribute as it is being used in mscorlib. I have removed the rest of the attributes.

jkotas:

stephentoub:
@jkotas, @shrah, it looks like the removal of ComEventInterfaceAttribute is breaking https://github.com/dotnet/corefx/pull/15779

PR #9328, opened 2017-02-04
Remove cross appdomains cultures checks

### LINKED ISSUE #9250: Inconsistent Behavior for Calendars between macOS on Windows
This (bad) code here behaves differently for Windows and macOS:

```csharp
using System;
using System.Globalization;

class Program
{
    private static CultureInfo _parsingCulture;
    
    {
        _parsingCulture = (CultureInfo)CultureInfo.InvariantCulture.Clone();
        _parsingCulture.DateTimeFormat.Calendar = new AsnUTCTimeCalendar();
        
    }

        private class AsnUTCTimeCalendar : GregorianCalendar
        {
            public override int ToFourDigitYear(int year) => (year < 50 ? 2000 : 1900) + year;
        }
}
```

On Windows, this throws an exception:

>System.InvalidOperationException: Cannot set sub-classed Program+UTCTimeCalendar object to System.Globalization.DateTimeFormatInfo object.

On macOS, an exception is not thrown and it "works".

I understand why the Windows version is throwing, but I was surprised that it did not on macOS. I think the correct behavior is that all platforms should throw the exception.

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64

tarekgh:
In the ful framework, CultureInfo and other globalization objects are marked as agile which means it can cross the appdomains. In coreclr we don't have multiple appdomains anymore so these checks are not really needed.
removing these checks will make coreclr consistent when running on Windows and other OS's

Fixes #9250

### DIFF
--- a/src/mscorlib/src/System.Private.CoreLib.txt
+++ b/src/mscorlib/src/System.Private.CoreLib.txt
@@ -1052,7 +1052,6 @@ InvalidOperation_CantCancelCtrlBreak = Applications may not prevent control-brea
 InvalidOperation_CalledTwice = The method cannot be called twice on the same instance.
 InvalidOperation_CollectionCorrupted = A prior operation on this collection was interrupted by an exception. Collection's state is no longer trusted.
 InvalidOperation_CriticalTransparentAreMutuallyExclusive = SecurityTransparent and SecurityCritical attributes cannot be applied to the assembly scope at the same time.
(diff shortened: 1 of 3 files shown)


tarekgh:
@jkotas could you please have a look?

tarekgh:

PR #9329, opened 2017-02-04
Get call arg value numbers from late args

JosephTremoulet:
When an arg has an associated late arg, take its value number from
the late arg (which represents the value) rather than the other one
(which represents argument setup).

This fixes the issue that is blocking #9169.

This change also includes a little cleanup of some related code.


JosephTremoulet:
@briansull PTAL.
/cc @dotnet/jit-contrib 

Only two SPMI asm diffs, both from cases where we have enough locals/temps that the temp used to communicate the argument between the setup arg and late arg doesn't get included in SSA.

russellhadley:
LGTM

PR #9336, opened 2017-02-04
Fix incorrect cast in gtFoldExprConst

### LINKED ISSUE #9282: Assertion failed '!op2->IsCnsIntOrI() || genTypeCanRepresentValue(type, op2->AsIntCon()->IconValue())
After 378a73e, we are seeing this in autogenerated testing. Adding a GitHub test case for this.

```
BEGIN EXECUTION
 "C:\michelm\coreclr\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\corerun.exe" DevDiv_377155.exe

Assert failure(PID 16672 [0x00004120], Thread: 18732 [0x492c]): Assertion failed '!op2->IsCnsIntOrI() || genTypeCanRepresentValue(type, o
p2->AsIntCon()->IconValue())' in 'C:M():float' (IL size 77)

    File: c:\michelm\coreclr\src\jit\codegenxarch.cpp Line: 6199
    Image: C:\michelm\coreclr\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\CoreRun.exe

Expected: 100
Actual: 123456789
END EXECUTION - FAILED
FAILED
```

mikedn:
The fact that an operation is unsigned affects the operation itself but that doesn't mean that the result of the operation is also unsigned. Constants are stored as `ssize_t` and the node type is `TYP_INT` so the result has to be sign extended, not zero extended.

Otherwise code that uses the return of `IconValue()` without first narrowing it to `int` will behave incorrectly. Such code does exists, even `gtFoldExprConst` does this when folding shift operations:

```C#
uint a = uint.MaxValue;
uint b = 0;
int r = (int)checked(a + b);
Console.WriteLine((r >> 2).ToString("X"));
```

The above code prints 3FFFFFFF instead of the expected FFFFFFFF.

This also makes `gtFoldExprConst` consistent with `ValueNumStore::EvalFuncForConstantArgs` which evaluates `TYP_INT` nodes as `int` and then casts to `ssize_t`.

Fixes #9282

### DIFF
--- a/src/jit/gentree.cpp
+++ b/src/jit/gentree.cpp
@@ -13410,15 +13410,10 @@ GenTreePtr Compiler::gtFoldExprConst(GenTreePtr tree)
 #endif
 
 #ifdef _TARGET_64BIT_
-            // we need to properly re-sign-extend or truncate as needed.
-            if (tree->gtFlags & GTF_UNSIGNED)
-            {
-                i1 = UINT32(i1);
-            }
-            else
-            {
-                i1 = INT32(i1);
-            }
+            // Some operations are performed as 64 bit instead of 32 bit so the upper 32 bits
+            // need to be discarded. Since constant values are stored as ssize_t and the node
+            // has TYP_INT the result needs to be sign extended rather than zero extended.
+            i1 = INT32(i1);
 #endif // _TARGET_64BIT_
 
             /* Also all conditional folding jumps here since the node hanging from
--- a/tests/issues.targets
+++ b/tests/issues.targets
@@ -190,9 +190,6 @@
         <ExcludeList Include="$(XunitTestBinBase)\JIT\opt\perf\doublealign\Locals\*">
             <Issue>8418</Issue>
         </ExcludeList>
-        <ExcludeList Include="$(XunitTestBinBase)\JIT\Regression\JitBlue\DevDiv_377155\DevDiv_377155\*">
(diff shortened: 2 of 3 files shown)


mikedn:
Some builds appear to have failed due to a "file in use" error:

https://ci.dot.net/job/dotnet_coreclr/job/master/job/checked_osx_flow_prtest/2496/
https://ci.dot.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_flow_prtest/2539/

@dotnet-bot test OSX x64 Checked Build and Test please
@dotnet-bot test Ubuntu x64 Checked Build and Test please

src/jit/gentree.cpp
```diff
@@ -13412,15 +13412,10 @@ GenTreePtr Compiler::gtFoldExprConst(GenTreePtr tree)
...
-            if (tree->gtFlags & GTF_UNSIGNED)
-            {
-                i1 = UINT32(i1);
-            }
-            else
-            {
-                i1 = INT32(i1);
-            }
+            // Some operations are performed as 64 bit instead of 32 bit so the upper 32 bit
+            // need to be discarded. Since constant values are stored as ssize_t and the node
```
pgavlin: `s/need/needs`

pgavlin: Ah, nvm--the typo is actually `s/bit/bits`.

PR #9340, opened 2017-02-04
Use StringBuilderCache for Enum.InternalFlagsFormat

benaadams:
Greatly reduces the allocations when calling ToString on a Flag enum

Before

After

### DIFF
--- a/src/mscorlib/src/System/Enum.cs
+++ b/src/mscorlib/src/System/Enum.cs
@@ -157,7 +157,7 @@ namespace System
             Debug.Assert(names.Length == values.Length);
 
             int index = values.Length - 1;
-            StringBuilder retval = new StringBuilder();
+            StringBuilder sb = StringBuilderCache.Acquire();
             bool firstTime = true;
             ulong saveResult = result;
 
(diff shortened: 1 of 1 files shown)


benaadams:
I think dotnet bot is sleeping...

src/mscorlib/src/System/Enum.cs
```diff
@@ -173,31 +173,40 @@ private static String InternalFlagsFormat(RuntimeType eT, TypeValuesAndNames ent
...
+                    returnString = "0";
+                }
             }
             else
             {
-                return retval.ToString(); // Return the string representation
+                returnString = sb.ToString(); // Return the string representation
             }
+
+            StringBuilderCache.Release(sb);
```
danmosemsft: why not just StringBuilderCache.GetStringAndRelease(sb) instead

benaadams: 3 paths don't use what is in StringBuilder but it still needs to be returned

danmosemsft: Ah

jkotas:
@dotnet-bot test this please

PR #9341, opened 2017-02-04
Rename internal CoreLib Console to LowLevelConsole

jkotas:
Prep-work for #9278

### DIFF
--- a/src/mscorlib/model.xml
+++ b/src/mscorlib/model.xml
@@ -12588,7 +12588,7 @@
       <Member Name="get_TraceLevel" />
     </Type>
     <!--End of PFX types -->
-    <Type Name="System.Console">
+    <Type Name="System.LowLevelConsole">
       <Member Name="Write(System.String)" />
       <Member Name="WriteLine(System.String)" />
       <Member Name="WriteLine" />
--- a/src/mscorlib/src/System/LowLevelConsole.cs
+++ b/src/mscorlib/src/System/LowLevelConsole.cs
@@ -10,18 +10,14 @@ using Microsoft.Win32.SafeHandles;
 namespace System
 {
     //
-    // Simple limited console class for internal printf-style debugging in mscorlib
-    // and low-level tests that just want to depend on mscorlib.
+    // Simple limited console class for internal printf-style debugging in System.Private.CoreLib
+    // and low-level tests that want to call System.Private.CoreLib directly
     //
 
-    public static class Console
+    public static class LowLevelConsole
     {
-        static SafeFileHandle _outputHandle;
-
-        static Console()
-        {
-            _outputHandle = new SafeFileHandle(Win32Native.GetStdHandle(Win32Native.STD_OUTPUT_HANDLE), false);
-        }
+        private static readonly SafeFileHandle _outputHandle =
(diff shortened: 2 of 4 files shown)


src/mscorlib/model.xml
```diff
@@ -12588,7 +12588,7 @@
       <Member Name="get_TraceLevel" />
     </Type>
     <!--End of PFX types -->
-    <Type Name="System.Console">
+    <Type Name="System.LowLevelConsole">
```
stephentoub: Where do we still use this in System.Private.CoreLib?  Is it just for debugging?

src/mscorlib/src/System/LowLevelConsole.cs
```diff
@@ -10,15 +10,15 @@
 namespace System
 {
     //
-    // Simple limited console class for internal printf-style debugging in mscorlib
-    // and low-level tests that just want to depend on mscorlib.
+    // Simple limited console class for internal printf-style debugging in System.Private.CoreLib
+    // and low-level internal tests that want to call System.Private.CoreLib directly
```
stephentoub: Answers my question.  Does the implementation work on Unix?  I guess it does via the coreclr pal?

src/mscorlib/src/System/LowLevelConsole.cs
```diff
@@ -10,15 +10,15 @@
...
 
-    public static class Console
+    public static class LowLevelConsole
     {
         static SafeFileHandle _outputHandle;
 
-        static Console()
+        static LowLevelConsole()
         {
             _outputHandle = new SafeFileHandle(Win32Native.GetStdHandle(Win32Native.STD_OUTPUT_HANDLE), false);
```
stephentoub: Nit: any reason _outputHandle isn't just initialized at its declaration site?  It could also be readonly.

src/mscorlib/src/System/LowLevelConsole.cs
jkotas: Yes, it works on Unix via CoreCLR PAL today.

src/mscorlib/src/System/LowLevelConsole.cs
jkotas: Fixed.

danmosemsft:
Why is this part of removing the facade?

jkotas:
There are tests that exercise corelib APIs that are not part of any public contract today. I run into problems with duplicate Console type when getting them to build. I choose to deal with this by renaming the primitive Console to LowLevelConsole. The alternatives I have rejected:
- Use aliases to disambiguate the two types when necessary. We would need to do something about it eventually anyway to https://github.com/dotnet/corefx/issues/12678.
- Delete the primitive Console completely or make it internal. I think it is good idea to have this primitive Console around in CoreLib for low-level printf-style debugging or the internal testing when you just want target CoreLib.

So the rename is general goodness towards fixing https://github.com/dotnet/corefx/issues/12678. I am not 100% sure whether I will absolutely need it to get rid of the façade at the end because of I do not have it fully working yet - I am stuck on CoreCLR tests running on month old CoreFX right now https://github.com/dotnet/coreclr/issues/8937#issuecomment-277422923.

jkotas:

PR #9342, opened 2017-02-04
Remove NoInlining/StackCrawlMarks from Tasks

stephentoub:
These crawl marks in Tasks aren't relevant in coreclr (they're passed down to ExecutionContext, which in coreclr just ignores them), yet they required NoInlining, which does impact coreclr.  Removing them all  from tasks.

cc: @jkotas, @kouvel


jkotas:
Delete the method that takes StackCrawlMark from ExecutionContext?

LGTM otherwise.

benaadams:
Has conflicts

stephentoub:
> Has conflicts

Yes, I'm working on this.

stephentoub:
> Delete the method that takes StackCrawlMark from ExecutionContext?

It was still being used by some other calls in Timer, Overlapped, ThreadPool, and Thread, so I went through and cleaned up those as well, and deleted this Capture overload.  As long as I was doing so, I also cleaned up a few other things I noticed along the way.

PR #9343, opened 2017-02-04
Remove internal self-replicating support from Tasks

stephentoub:
In the initial TPL prototype, Tasks had a feature that allowed them to "replicate", meaning to schedule a copy of themselves.  Prior to shipping the initial version, this feature was made internal, as it was deemed unnecessary, but by then the Parallel loops constructs had already been built on top of them.  Now that Parallel no longer depends on this feature (and isn't in CoreLib), we can entirely rip it out.

cc: @kouvel, @AlfredoMS

### DIFF
--- a/src/mscorlib/src/System/Threading/Tasks/future.cs
+++ b/src/mscorlib/src/System/Threading/Tasks/future.cs
@@ -341,16 +341,10 @@ namespace System.Threading.Tasks
         /// <param name="cancellationToken">The CancellationToken for the task.</param>
         /// <param name="creationOptions">Options to control the future's behavior.</param>
         /// <param name="internalOptions">Internal options to control the future's behavior.</param>
-        /// <exception cref="T:System.ArgumentOutOfRangeException">The <paramref name="creationOptions"/> argument specifies
-        /// a SelfReplicating <see cref="Task{TResult}"/>, which is illegal."/>.</exception>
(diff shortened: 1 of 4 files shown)


src/mscorlib/src/System/Threading/Tasks/future.cs
```diff
@@ -420,10 +404,6 @@ public Task(Func<object, TResult> function, object state, CancellationToken canc
             {
                 ThrowHelper.ThrowArgumentNullException(ExceptionArgument.scheduler);
             }
-            if ((internalOptions & InternalTaskOptions.SelfReplicating) != 0)
-            {
-                ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.creationOptions, ExceptionResource.TaskT_ctor_SelfReplicating);
```
benaadams: Can remove these ExceptionResource enums

AlfredoMS:
LGTM. I'm glad the codes will be a little cleaner now.
cc: @alexperovich

src/mscorlib/src/System/Threading/Tasks/future.cs
stephentoub: Fixed.  I removed the ExceptionResource, but not the ExceptionArgument, as it's used elsewhere.

Issue #9344, opened 2017-02-05
Remove BCL rewriter


danmosemsft:
After https://github.com/dotnet/coreclr/pull/9269 it should not be necessary anymore.

@jkotas is there anything necessary to take care of beyond just removing it from the build?

jkotas:
Just removing from the build should be enough (together with model.xml, etc.)

PR #9351, opened 2017-02-05
Remove bclrewriter

danmosemsft:
Fixes https://github.com/dotnet/coreclr/issues/9344

Fixed a few remaining diffs with the rewritten version. It is still about 40KB (1%) larger as there are still diffs for constants that are no longer inlined, and serialization helper functions.

@jkotas @stephentoub


src/mscorlib/src/System/Reflection/Assembly.cs
```diff
@@ -1691,14 +1691,6 @@ public override Assembly GetSatelliteAssembly(CultureInfo culture, Version versi
             return retAssembly;
         }
 
-        [MethodImplAttribute(MethodImplOptions.InternalCall)]
-        static internal extern RuntimeAssembly nLoadImage(byte[] rawAssembly,
```
danmosemsft: I'm not going to edit ecalllist.h right now, I think that can be cleaned up more separately.

src/mscorlib/src/System/Security/Permissions/HostProtectionPermission.cs
```diff
@@ -41,8 +41,8 @@ internal enum HostProtectionResource
     [AttributeUsage(AttributeTargets.Method | AttributeTargets.Constructor | AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Assembly | AttributeTargets.Delegate, AllowMultiple = true, Inherited = false )] 
     [System.Runtime.InteropServices.ComVisible(true)]
     [Serializable]
-    // This needs to be in the asmmeta to enable SecAnnotate to successfully resolve and run the security rules. It gets marked
-    // as internal by BCLRewriter so we are simply marking it as FriendAccessAllowed so it stays in the asmmeta.
+    // This needs to be in the asmmeta to enable SecAnnotate to successfully resolve and run the security rules. 
+    // We are simply marking it as FriendAccessAllowed so it stays in the asmmeta.
     [System.Runtime.CompilerServices.FriendAccessAllowedAttribute]
```
jkotas: You can delete this comment as well as FriendAccessAllowedAttribute. None of it is applicable anymore.

jkotas:
LGTM ... once the CI is green.

src/mscorlib/src/System/Runtime/InteropServices/IException.cs
```diff
@@ -23,7 +23,7 @@ namespace System.Runtime.InteropServices {
     [InterfaceTypeAttribute(ComInterfaceType.InterfaceIsDual)]
     [CLSCompliant(false)]
     [System.Runtime.InteropServices.ComVisible(true)]
-    public interface _Exception
```
jkotas: This whole file should be deleted - I forgot to delete it when I was cleaning up these.

danmosemsft:
@dotnet-bot test windows_nt x64 debug build and test ("The "GeneratePackageReport" task failed unexpectedly." again.. cited separately)

danmosemsft:
This is not going to be finished for a while (later today?) as I need to set up a Unix machine to avoid working through the linux crossgen issues one by one.

src/mscorlib/src/System/AppDomain.cs
```diff
@@ -837,10 +837,17 @@ public Object GetData(string name)
                 }
             }
         }
-        
+       
+#if !PLATFORM_UNIX
```
jkotas: Why do you need to do this? Can we delete it instead?

danmosemsft: I would have to #if !UNIX several pieces of EventSource. I can do that if you prefer.

jkotas: What was the actual problem with it on Unix? We have `GetCurrentThreadId` in the CoreCLR Unix PAL, so it should have worked just fine as is.

danmosemsft: It appears this was a mistake. Reverted. Thank you.

danmosemsft:
@dotnet-bot test osx x64 checked build and test (https://github.com/dotnet/coreclr/issues/9388)

danmosemsft:
Good to merge?

PR #9352, opened 2017-02-05
ConcurrentQueue don't spin for small race loss

benaadams:
@stephentoub  this correct? (As discussed in https://github.com/benaadams/ThreadPoolTaskTesting/pull/3#issuecomment-277535755)

### DIFF
--- a/src/mscorlib/src/System/Collections/Concurrent/ConcurrentQueue.cs
+++ b/src/mscorlib/src/System/Collections/Concurrent/ConcurrentQueue.cs
@@ -954,6 +954,8 @@ namespace System.Collections.Concurrent
                             }
                             return true;
                         }
+                        // Slightly lost race. Loop and try again.
+                        continue;
                     }
                     else if (sequenceNumber < currentHead + 1)
                     {
@@ -1078,6 +1080,8 @@ namespace System.Collections.Concurrent
                             Volatile.Write(ref _slots[slotsIndex].SequenceNumber, currentTail + 1);
                             return true;
                         }
+                        // Slightly lost race. Loop and try again.
+                        continue;
                     }
                     else if (sequenceNumber < currentTail)
                     {


benaadams:
Pre
```
                                                                             Parallelism
QUWI No Queues (TP)             10.291 M    11.050 M    11.189 M    11.230 M    11.141 M
- Depth    2                     8.211 M     8.152 M     8.168 M     8.121 M     8.116 M
- Depth   16                     8.902 M     8.747 M     8.988 M     8.950 M     8.904 M
- Depth   64                     9.021 M     9.114 M     9.135 M     9.075 M     9.142 M
- Depth  512                     9.157 M     9.251 M     9.285 M     9.361 M     9.409 M

QUWI No Queues                  10.471 M    10.628 M    10.343 M    10.459 M     6.620 M
- Depth    2                     8.059 M     7.814 M     7.674 M     7.643 M     4.881 M
- Depth   16                     8.303 M     8.280 M     8.454 M     8.013 M     5.342 M
- Depth   64                     8.709 M     8.696 M     8.557 M     8.113 M     5.394 M
- Depth  512                     8.696 M     8.643 M     8.662 M     8.278 M     4.858 M
```
Post
```
                                                                             Parallelism
QUWI No Queues (TP)             10.517 M    10.162 M    10.632 M    10.697 M    11.650 M
- Depth    2                     8.869 M     7.874 M     8.248 M     8.679 M     8.725 M
- Depth   16                     9.955 M    10.015 M    10.100 M    10.170 M    10.176 M
- Depth   64                    10.322 M    10.472 M    10.587 M    10.530 M    10.503 M
- Depth  512                    10.534 M    10.606 M    10.635 M    10.621 M    10.569 M

QUWI No Queues                  11.136 M    10.677 M     9.908 M     9.941 M     5.780 M
- Depth    2                     9.103 M     8.084 M     7.423 M     7.510 M     5.218 M
- Depth   16                    10.151 M    10.147 M    10.016 M     9.806 M     5.370 M
- Depth   64                    10.464 M    10.446 M    10.505 M    10.250 M     4.747 M
- Depth  512                    10.685 M    10.740 M    10.563 M    10.493 M     5.373 M
```

Some random in  TryDequeue and concurrent so the results are non-deterministic; but generally improved.

stephentoub:
> this correct?

In general, yes, but I'll explain why I suggested it.

I think there are three main cases to consider:
- Case 1: If two dequeuers read the current head and sequence number at the exact same time, then they may both try to increment the head with a compare-exchange: one will win, one will lose, and the loser will need to start over and retry.
- Case 2: One dequeuer reads the current head and gets preempted.  Another dequeuer reads the current head, reads the current sequence number, successfully increments the head, and successfully writes the new sequence number.  At that point, the original dequeuer reads the sequence number for the head slot it already grabbed, and finds that it's not the next one in the sequence, so it needs to start over and retry.

These two cases are by far going to be the most common, and I think there's a good argument to be made that neither of them really needs spinning: very little work is done between interlocked operations, and in most cases these cases are lock-free in the true sense of the definition: one thread getting starved won't prevent other threads from making forward progress... if the winner of the compare-exchange then stalls before it can do its write, the other dequeuer will simply loop around again, find a new head, and be able to dequeuer from there.

However, there is a potential 3rd less likely but more impactful case:
- Case 3: One dequeuer reads the current head and sequence number, successfully increments the head, and then gets preempted before it can write the new sequence number.  Then a bunch of other dequeues happen, such that we wrap around again to this same slot (and the original dequeuer is still stalled).  At this point, we see that the algorithm isn't actually lock-free as far as the definition is concerned, but rather is only obstruction-free: until that original dequeuer wakes up and does its sequence number write, attempts to dequeuer at this point will loop indefinitely, as they'll continually read the head value pointing to this slot, find that the sequence number doesn't line up, and try again.

It's case 3 that safer, less CPU-intensive spinning is critical for, as you want to stall the additional dequeuers in this case so that they don't further prevent the original dequeuer from waking up and completing its sequence number write.  The problem is, I'm not sure there's a good way to differentiate case 2 from case 3.  Hence my suggestion to spin for case 2/3 but not for case 1.

Thinking about it now, it's possible case 2 and case 3 could be differentiated by examining the distance between the read sequence number and the read head, and if it's < the length of the array, you know you're in case 2, otherwise case 3.  I think that works, but I'd need to think through it more.

Assuming we did go ahead with this change, my one code suggestion would be to remove the added continue and instead just move the SpinOnce into an else block off of the existing if/else if chain.

> but generally improved

It looks like the non-depth/TP case regressed across the board from serial to 64x.  Is that repeatable?  Have you tried on a machine larger than 4 core?

stephentoub:
ps And any change we do make here, we should also make to the general version used in corefx.  I don't expect there'd be any use-case difference that would necessitate spinning in one but not the other in the same code paths.

benaadams:
> It looks like the non-depth/TP case regressed across the board from 2x to 64x. Is that repeatable? 

No, it goes back up to the 11.x Ms across runs; but then another area will come down.

Hmm... overflow on the comparisons? Need to cast to `uint`?

I'd prefer if you make the the PR; then I can run it on a bunch of configurations? Can go up to 16 (32 HT) bare metal with 2 NUMA.

benaadams:
Closing as there are a few subtleties I don't completely have a handle on.

benaadams:
Also is it worth spinning if you are say for example `Math.Min(2 * ProcCount, _slots.Length)` away?

stephentoub:
> I'd prefer if you make the the PR

OK. I'll send something tomorrow to test.

> Also is it worth spinning if you are say for example Math.Min(ProcCount, _slots.Length) away?

Can you explain the thinking here?

benaadams:
> Can you explain the thinking here?

If you manage to wrap the queue and the queue is longer than the number of procs, probably should have started spinning earlier?

stephentoub:
Sorry, I'm still not getting it. What does the number of elements in the queue have to do with the number of processors? One thread on one core could have added or removed them all. Can you describe where you're taking about adding this spinning, what the actual condition of the check would be,  and what race condition it would help?

benaadams:
> What does the number of elements in the queue have to do with the number of processors? 

Was Case 3; but have worked it though and the issue only kicks in > segment length; so proc count doesn't matter.

Other confusion I'm have is 
> Volatile.Write(ref _slots[slotsIndex].SequenceNumber, currentHead + _slots.Length);

and

> (sequenceNumber < currentHead + 1)

Is this a test with int overflow?

stephentoub:
> Is this a test valid with int overflow? Or will it cause perf blips?

This is a bug.  Will fix.

Issue #9321, opened 2017-02-03
Remove CAS from CoreLib


danmosemsft:
There is lots of CAS-only code -- CodeAccessPermission and related types and code that uses them. Ideally it would all be removed. I'm guessing one thing to take care of is making sure we continue to do the same parameter validation/throw the same exceptions. That has been a gotcha for IO in the past.

Removing remaining FEATURE_CAS_POLICY on the native side is already tracked by https://github.com/dotnet/coreclr/issues/9067

danmosemsft:
I'm already deleting most of the CAS types/members that aren't actually referenced. (There are a few that are rooted by model.xml that I'm not)

benaadams:
Dotnet just keeps getting better and better 😄

danmosemsft:
I want to be clear this concerns removing the implementations in corelib. We brought back (mostly dummy) CAS types in CoreFX to make existing code work, and those are staying.

jkotas:
I am working on round of cleanup for this: https://github.com/jkotas/coreclr/tree/security-cleanup3

benaadams:
@danmosemsft I know; it just means Execution context etc are definitely staying slim 😄

jkotas:
The work remaining here is to cleanup references to types in https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/CleanupToDoList.cs and delete this file.

danmosemsft:
The last ones involve some sigificant changes in AppDomain that I'm reluctant to do without spending the time to properly understand the code to be confident what is really dead.

Issue #9353, opened 2017-02-05
src\mscorlib\mscorlib.shared.sources.props is messed up in git


jkotas:
Repro (on Windows):

```
git clone https://github.com/dotnet/coreclr
cd coreclr\src\mscorlib
touch mscorlib.shared.sources.props (or open in notepad and save it without changing anything)
git checkout -- *
git status
```

Actual result:
`git status` reports mscorlib.shared.sources.props as modified, but there is no actual diff. It is impossible to get to clean state with nothing checked out.

Expected result:
`git status` reports nothing modified

jkotas:
@danmosemsft It has something to do with changing the file to binary that was part of your change. I am not able to tell what you have done exactly to change it ... could you please take a look?

danmosemsft:
I cannot repro this (on a clean clone)

Maybe time to copy and paste the content into a fresh file named "system.private.corelib.sources.props" (or right into the project itself) ..?

jkotas:
Agree ... I will do it.

jkotas:
Fixed by https://github.com/dotnet/coreclr/pull/9355

PR #9355, opened 2017-02-05
CAS Security cleanup

jkotas:
- Delete most CAS support under mscorlib\src\System\Security. Introduce temporary dummy stubs for a few types that are referenced in many places, to be cleaned up later. https://github.com/dotnet/coreclr/pull/9355/files#diff-7f73d6dbdafd14ee136ff60b85276236
- Delete enough on the unmanaged side to break ties with the removed types on the managed side. There is still quite a bit left though.

Contributes to https://github.com/dotnet/coreclr/issues/9321


jkotas:
cc @danmosemsft

jkotas:
mscorlib.shared.sources.props shows up as all diff because of https://github.com/dotnet/coreclr/issues/9353

jkotas:
I have run coreclr and corefx tests on this.

@danmosemsft @gkhanna79 Could you please take a cursory look?

src/mscorlib/src/System/Reflection/Assembly.cs
```diff
@@ -1268,11 +1239,11 @@ internal bool IsSafeForReflection()
                                              IntPtr pPrivHostBinder,
                                              bool throwOnFileNotFound,
                                              bool forIntrospection,
-                                             bool suppressSecurityChecks, IntPtr ptrLoadContextBinder = default(IntPtr))
+                                             IntPtr ptrLoadContextBinder = default(IntPtr))
         {
             return _nLoad(fileName, codeBase, assemblySecurity, locationHint, ref stackMark,
                 pPrivHostBinder,
-                throwOnFileNotFound, forIntrospection, suppressSecurityChecks, ptrLoadContextBinder);
+                throwOnFileNotFound, forIntrospection, true, ptrLoadContextBinder);
```
danmosemsft: >true [](start = 55, length = 4)

nit, it is nice to prefix bools 
like /* suppressSecurityChecks = */ true
but I see this is not widely done.

danmosemsft:
LGTM in a quick look. I hope we have enough test coverage of all the parameter validation.

Issue #7689, opened 2016-10-18
Provide the appropriate intrinsic implementations for specific single-precision math functions


tannergooding:
PR #5492 implemented dotnet/corefx#1151, by providing scalar single-precision floating-point support for many of the trigonometric, logarithmic, and other common mathematical functions.

However, there is still the additional work that needs to be done where the new intrinsics are properly plumbed through the `ValueNumStore`.

PR #5492 has some details on what needs to be done here, but essentially:

> Pretty much all of the intrinsic hookup work, as I understand it, is in the valueenum.cpp file. However, there does not appear to be much (if any) documentation on the ValueNumStore.

tannergooding:
I am self-assigning these for the time being and am going to work on getting these completed.

tannergooding:
This was resolved with #9363 and #9717

PR #9363, opened 2017-02-06
Updating jit/valuenum to properly handle the single-precision versions of the math intrinsics.

tannergooding:
This resolves https://github.com/dotnet/coreclr/issues/7689

### DIFF
--- a/src/jit/valuenum.cpp
+++ b/src/jit/valuenum.cpp
@@ -230,6 +230,30 @@ double ValueNumStore::EvalOp<double>(VNFunc vnf, double v0, double v1, ValueNum*
     }
 }
 
+// Specialize for float for floating operations, that doesn't involve unsigned.
+template <>
+float ValueNumStore::EvalOp<float>(VNFunc vnf, float v0, float v1, ValueNum* pExcSet)
+{
+    genTreeOps oper = genTreeOps(vnf);
+    // Here we handle those that are the same for floating-point types.
+    switch (oper)
+    {
+        case GT_ADD:
+            return v0 + v1;
+        case GT_SUB:
+            return v0 - v1;
+        case GT_MUL:
+            return v0 * v1;
+        case GT_DIV:
+            return v0 / v1;
+        case GT_MOD:
+            return fmodf(v0, v1);
+
+        default:
+            unreached();
+    }
+}
+
 template <typename T>
 int ValueNumStore::EvalComparison(VNFunc vnf, T v0, T v1)
 {
@@ -961,6 +985,17 @@ ValueNum ValueNumStore::VNForFunc(var_types typ, VNFunc func, ValueNum arg0VN)
     }
 }
 
+// Windows x86 and Windows ARM/ARM64 may not define _isnanf() but they do define _isnan().
+// We will redirect the macros to these other functions if the macro is not defined for the
+// platform. This has the side effect of a possible implicit upcasting for arguments passed.
+#if (defined(_TARGET_X86_) || defined(_TARGET_ARM_) || defined(_TARGET_ARM64_)) && !defined(FEATURE_PAL)
+
+#if !defined(_isnanf)
+#define _isnanf _isnan
+#endif
+
+#endif
+
 ValueNum ValueNumStore::VNForFunc(var_types typ, VNFunc func, ValueNum arg0VN, ValueNum arg1VN)
 {
     assert(arg0VN != NoVN && arg1VN != NoVN);
@@ -988,8 +1023,12 @@ ValueNum ValueNumStore::VNForFunc(var_types typ, VNFunc func, ValueNum arg0VN, V
         // We don't try to fold a binary operation when one of the constant operands
         // is a floating-point constant and the other is not.
         //
-        bool arg0IsFloating = varTypeIsFloating(TypeOfVN(arg0VN));
-        bool arg1IsFloating = varTypeIsFloating(TypeOfVN(arg1VN));
+        var_types arg0VNtyp      = TypeOfVN(arg0VN);
+        bool      arg0IsFloating = varTypeIsFloating(arg0VNtyp);
+
+        var_types arg1VNtyp      = TypeOfVN(arg1VN);
+        bool      arg1IsFloating = varTypeIsFloating(arg1VNtyp);
+
         if (arg0IsFloating != arg1IsFloating)
         {
             canFold = false;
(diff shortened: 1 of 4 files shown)


tannergooding:
FYI. @JosephTremoulet

src/jit/valuenum.cpp
```diff
@@ -1613,6 +1613,7 @@ double ValueNumStore::GetConstantDouble(ValueNum argVN)
 {
     assert(IsVNConstant(argVN));
     var_types argVNtyp = TypeOfVN(argVN);
+    assert(varTypeIsFloating(argVNtyp));
```
tannergooding: `ValueNumStore::GetConstantDouble` currently returns for both `float` and `double` constants. This means in some cases we are upcasting, executing some code, and then downcasting back to `float`.

After the reworking of `EvalMathFuncUnary` there are still a couple of places where this is happening (a simple search for `GetConstantDouble` in this file will show them). These calls could probably be rewritten to no longer up/down cast, but I would like some input before I do that work.

src/jit/valuenum.cpp
```diff
@@ -1630,6 +1631,27 @@ double ValueNumStore::GetConstantDouble(ValueNum argVN)
...
+//
+float ValueNumStore::GetConstantSingle(ValueNum argVN)
+{
+    assert(IsVNConstant(argVN));
+    var_types argVNtyp = TypeOfVN(argVN);
+    assert(argVNtyp == TYP_FLOAT);
+
+    float result = 0;
+
+    switch (argVNtyp)
```
JosephTremoulet: `argVNTyp` is only used in an assert and in a single-case switch with an `unreached` default case, which effectively asserts the same thing again.  I'd suggest getting rid of `argVNTyp` (use `TypeOfVN(argVN)` in the assertion) and simplifying away the switch.

src/jit/valuenum.cpp
```diff
@@ -3249,46 +3271,106 @@ ValueNum ValueNumStore::EvalMathFuncUnary(var_types typ, CorInfoIntrinsics gtMat
...
-            default:
-                unreached(); // the above are the only math intrinsics at the time of this writing.
-        }
         if (typ == TYP_DOUBLE)
         {
+            // Both operand and its result must be of the same floating point type.
+            assert(typ == TypeOfVN(arg0VN));
+
+            // If the math intrinsic is not implemented by target-specific instructions, such as implemented
+            // by user calls, then don't do constant folding on it. This minimizes precision loss.
```
JosephTremoulet: ISTM this comment was explaining why `&& Compiler::IsTargetIntrinsic(gtMathFN)` was included in the now-outer conditional; please move it back out.

src/jit/valuenum.cpp
```diff
@@ -3249,46 +3271,106 @@ ValueNum ValueNumStore::EvalMathFuncUnary(var_types typ, CorInfoIntrinsics gtMat
...
             return VNForDoubleCon(res);
         }
         else if (typ == TYP_FLOAT)
         {
-            return VNForFloatCon(float(res));
+            // Both operand and its result must be of the same floating point type.
+            assert(typ == TypeOfVN(arg0VN));
+
+            // If the math intrinsic is not implemented by target-specific instructions, such as implemented
+            // by user calls, then don't do constant folding on it. This minimizes precision loss.
```
JosephTremoulet: Ditto re: comment placement

src/jit/valuenum.cpp
```diff
@@ -3249,46 +3271,106 @@ ValueNum ValueNumStore::EvalMathFuncUnary(var_types typ, CorInfoIntrinsics gtMat
...
+            return VNForFloatCon(res);
         }
         else
         {
             assert(typ == TYP_INT);
             assert(gtMathFN == CORINFO_INTRINSIC_Round);
 
-            return VNForIntCon(int(res));
+            int res = 0;
+            if (gtMathFN == CORINFO_INTRINSIC_Round)
```
JosephTremoulet: This `if` and the `unreached()` in its `else` branch are redundant with the assertion 3 lines above.  I'd probably let the assertion do the talking (maybe add a comment that Round is the only intrinsic with fp args and a non-fp result) and get rid of the if/else.

src/jit/valuenum.cpp
JosephTremoulet: Yeah... in fact, it looks like we're already (without this change) constant-folding the single-precision intrinsics, just doing it by widening to double and evaluating and truncating back down, is that right?  The language in the [spec](https://www.ecma-international.org/publications/standards/Ecma-335.htm) certainly seems to give us the liberty to do that (I.12.1.3). I'm inclined to say that we should either go ahead and make this change for all constant-folding on fp ops or not make this change, to keep things consistent.

I'm not sure whether it's best to continue folding at double precision or to switch to single.  I'd think we'd want to minimize differences between optimized and unoptimized code's results, and I think the unoptimized code would end up happening at whatever precision our back-end uses for singles, which maybe has a greater chance of matching the C++ compiler's handling of `float` than of `double`, which would argue for continuing this change... but I'm no floating-point expert.  Maybe someone else wants to chime in. @sivarv? @CarolEidt?

tannergooding: > Yeah... in fact, it looks like we're already (without this change) constant-folding the single-precision intrinsics, just doing it by widening to double and evaluating and truncating back down, is that right?

Yes, which can result in lower performance and different computed results for the new System.MathF functions.

> The language in the spec certainly seems to give us the liberty to do that (I.12.1.3).

I believe that most consumers of the single-precision functions would rather have the speed than the additional precision (including any additional precision retained by upcasting, computing, and downcasting).

> I'd think we'd want to minimize differences between optimized and unoptimized code's results

I agree, and my understanding is that 'internal representation' is a platform agnostic way of saying whatever form the CPU accepts (so 80-bits when using the x87 FPU, for example). I think that using the C type names (float and double), calling the appropriate versions of the functions (`sqrtf` and `sqrt`, respectively), and then letting the native runtime handle the internal implementation details is the best way to go here.

JosephTremoulet: @tannergooding, your reasoning sounds good to me; I'd say go ahead with this and the other instances.

src/jit/valuenum.cpp
tannergooding: Fixed.

src/jit/valuenum.cpp
tannergooding: Fixed

src/jit/valuenum.cpp
tannergooding: Fixed.

tannergooding:
Updated to handle all but one case where we are upcasting float to double when performing evaluations.

The remaining case is in Compiler::fgValueNumberTreeConst where it downcasts:
```C++
tree->gtVNPair.SetBoth(vnStore->VNForFloatCon((float)tree->gtDblCon.gtDconVal));
```

I'm taking a closer look at the code in question there to see if that is feasible to include here as well.

tannergooding:
@JosephTremoulet, this should be good for a final review pass now.

tannergooding:
@JosephTremoulet, thanks. Do I need any further sign-off here before merging?

JosephTremoulet:
> Do I need any further sign-off here before merging?

No, go ahead.

Issue #9362, opened 2017-02-06
[ARM32/Linux] Cross-architecture debug/checked build error with clang 3.9 - casting  calling convention


hseok-oh:
Build with #8866, #9262 

`./build.sh checked cross arm crosscomponent clang3.9`

result:

```
/nfs/rbp3/coreclr/src/vm/ceeload.cpp:14454:24: error: cast between incompatible calling conventions 'cdecl' and 'stdcall'; calls through this pointer may abort at runtime
      [-Werror,-Wcast-calling-convention]
                       (ENCODEMODULE_CALLBACK) Module::EncodeModuleHelper,
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/nfs/rbp3/coreclr/src/vm/ceeload.cpp:14224:26: note: consider defining 'EncodeModuleHelper' with the 'stdcall' calling convention
/*static*/ DWORD Module::EncodeModuleHelper(void * pModuleContext, Module *pReferencedModule)
                         ^
                         QCALLTYPE
/nfs/rbp3/coreclr/src/vm/ceeload.cpp:14455:25: error: cast between incompatible calling conventions 'cdecl' and 'stdcall'; calls through this pointer may abort at runtime
      [-Werror,-Wcast-calling-convention]
                        (DEFINETOKEN_CALLBACK) Module::TokenDefinitionHelper);
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/nfs/rbp3/coreclr/src/vm/ceeload.cpp:14249:25: note: consider defining 'TokenDefinitionHelper' with the 'stdcall' calling [ 58%] Building CXX object src/utilcode/staticnohost/CMakeFiles/utilcodestaticnohost.dir/__/iallocator.cpp.o
convention
/*static*/ void Module::TokenDefinitionHelper(void* pModuleContext, Module *pReferencedModule, DWORD index, mdToken* pToken)
                        ^
                        QCALLTYPE
```
`Module::EncodeModuleHelper` uses `cdecl` calling convention in x86 linux (maybe stdcall in x86 windows), but `ENCODEMODULE_CALLBACK` is defined as `stdcall`.

It needs to fix definition of `ENCODEMODULE_CALLBACK` to `cdecl` in linux or `Module::EncodeModuleHelper` to `stdcall`.

hseok-oh:
[...]

PR #9364, opened 2017-02-06
[ARM32/Linux] Fix calling convention - stdcall in crossgen

hseok-oh:
- Fix calling convention:
not use stdcall for ENCODEMODULE_CALLBACK & DEFINETOKEN_CALLBACK in linux/x86
x86-host/arm32-target crossgen & x86 crossgen
build error in clang 3.9: not allowed calling convention casting

### DIFF
--- a/src/inc/corcompile.h
+++ b/src/inc/corcompile.h
@@ -1456,12 +1456,19 @@ class ICorCompilationDomain
 /*********************************************************************************
  * ICorCompileInfo is the interface for a compiler
  *********************************************************************************/
-
+#if defined(_X86_) and !defined(FEATURE_PAL)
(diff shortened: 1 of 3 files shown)


hseok-oh:
related issue: #9362 

I also fix this issue by using stdcall in linux/x86: https://github.com/hseok-oh/coreclr/tree/cac/crossgen_stdcall
But I push this commit because this is better than usgin stdcall in x86/linux. Please check which approach is better.

jkotas:
Could you please fix this without introducing ifdefs? It should be fine to not have the explicit __stdcall here on all platforms.

hseok-oh:

Issue #9357, opened 2017-02-06
[x86/Linux] Unit test Interop FunctionPtrTest failes with SIGSEGV


seanshpark:
Interop.MarshalAPI.FunctionPointer.FunctionPtrTest.FunctionPtrTest fails with Segment fault.
Fail is when Native calls Managed and in returning.
```
extern "C" DLL_EXPORT bool __cdecl CheckFcnPtr(bool(__stdcall *fcnptr)(__int64))
{
...
    else
    {
        return fcnptr(999999999999);
    }
}
```
`fcnptr()` is `__stdcall` so above code is generated as
```
   0xb7fc0db8 <CheckFcnPtr(bool (*)(long long))+152>:	8b 4d f0	mov    -0x10(%ebp),%ecx
   0xb7fc0dbb <CheckFcnPtr(bool (*)(long long))+155>:	89 e3	mov    %esp,%ebx
   0xb7fc0dbd <CheckFcnPtr(bool (*)(long long))+157>:	c7 43 04 e8 00 00 00	movl   $0xe8,0x4(%ebx)
   0xb7fc0dc4 <CheckFcnPtr(bool (*)(long long))+164>:	c7 03 ff 0f a5 d4	movl   $0xd4a50fff,(%ebx)
   0xb7fc0dca <CheckFcnPtr(bool (*)(long long))+170>:	8b 5d e8	mov    -0x18(%ebp),%ebx
=> 0xb7fc0dcd <CheckFcnPtr(bool (*)(long long))+173>:	89 45 d4	mov    %eax,-0x2c(%ebp)
   0xb7fc0dd0 <CheckFcnPtr(bool (*)(long long))+176>:	ff d1	call   *%ecx
   0xb7fc0dd2 <CheckFcnPtr(bool (*)(long long))+178>:	83 ec 08	sub    $0x8,%esp
```
With GDB, after and before `ESP` of `call   *%ecx` has same  value but as it is `__stdcall` it shouldn't be.
`fcnptr` is passed from Managed as
```
fcnptr = Marshal.GetFunctionPointerForDelegate<DelegateWithLong>(del);
```
This code needs some investigation.

seanshpark:
CC @parjong, @wateret

seanshpark:
Function that was last called(or jumped) and returned to `CheckFcnPtr` was `UMThunkStub`. 
Current implementation is just fixed "ret" and without any callee pop.
[...]

PR #9365, opened 2017-02-06
[x86/Linux] Fix UMThunkStub stack alignment

seanshpark:
Fixes out going call in UMThunkStub to be 16 byte stack aligned

### DIFF
--- a/src/vm/i386/umthunkstub.S
+++ b/src/vm/i386/umthunkstub.S
@@ -28,7 +28,7 @@ NESTED_ENTRY UMThunkStub, _TEXT, UnhandledExceptionHandlerUnix
 #define UMThunkStub_UMENTRYTHUNK_OFFSET     (UMThunkStub_SAVEDREG+4)
 #define UMThunkStub_THREAD_OFFSET           (UMThunkStub_UMENTRYTHUNK_OFFSET+4)
 #define UMThunkStub_INT_ARG_OFFSET          (UMThunkStub_THREAD_OFFSET+4)
-#define UMThunkStub_FIXEDALLOCSIZE          (UMThunkStub_LOCALVARS)
+#define UMThunkStub_FIXEDALLOCSIZE          (UMThunkStub_LOCALVARS+4) // extra 4 is for stack alignment
 
 // return address                           <-- entry ESP
 // saved ebp                                <-- EBP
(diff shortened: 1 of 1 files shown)


seanshpark:
Related #9357

seanshpark:
@jkotas , @janvorli , PTAL

seanshpark:
CC @parjong, @lucenticus

src/vm/i386/umthunkstub.S
```diff
@@ -122,11 +123,13 @@ LOCAL_LABEL(InvalidTransition):
 LOCAL_LABEL(DoTrapReturningThreadsTHROW):
 
     // extern "C" VOID STDCALL UMThunkStubRareDisableWorker(Thread *pThread, UMEntryThunk *pUMEntryThunk)
+    sub     esp, 8
```
janvorli: A nit - could you please add a comment that this is to ensure proper stack alignment?

src/vm/i386/umthunkstub.S
```diff
@@ -24,7 +24,7 @@ NESTED_END TheUMEntryPrestub, _TEXT
 NESTED_ENTRY UMThunkStub, _TEXT, UnhandledExceptionHandlerUnix
 
 #define UMThunkStub_SAVEDREG                (3*4)   // ebx, esi, edi
-#define UMThunkStub_LOCALVARS               (2*4)   // UMEntryThunk*, Thread*
+#define UMThunkStub_LOCALVARS               (3*4)   // UMEntryThunk*, Thread*, dummy void*
```
janvorli: A nit - I would keep this unchanged instead of including the alignment in the local variables and modify the UMThunkStub_FIXEDALLOCSIZE definition to include the +4 alignment. Having the alignment a part of the local vars seems a little bit odd to me.

seanshpark: OK, thank you for the explanation. I'll follow.

src/vm/i386/umthunkstub.S
seanshpark: Yes, comment is needed :)

seanshpark:
@janvorli , could you please check this again?

src/vm/i386/umthunkstub.S
```diff
@@ -37,6 +37,7 @@ NESTED_ENTRY UMThunkStub, _TEXT, UnhandledExceptionHandlerUnix
 // saved edi
 // UMEntryThunk*
 // Thread*
+// dummy void* for 16 byte stack alignment
```
janvorli: A nit, the comment that you had there before ("dummy 4 bytes" instead of "void*") seemed better to me.

seanshpark: OK

seanshpark:
@janvori, PTAL?

seanshpark:
Thank you!

Issue #9272, opened 2017-02-02
[x86/Linux] JIT.Directed.coverage.oldtests.ovfldiv1_il_r Test Failure


parjong:
``JIT.Directed.coverage.oldtests.ovfldiv1_il_r`` test ends with segmentation fault:
```
$ ./corerun coreclr-unittest/JIT/Directed/coverage/oldtests/ovfldiv1_il_r/ovfldiv1_il_r.exe
-9223372036854775808
1
-8762203435012037017
PASSED
Segmentation fault (core dumped)
```

parjong:
It seems that there is calling convention issue:
```
(gdb) r
-9223372036854775808
1
-8762203435012037017
PASSED

Program received signal SIGSEGV, Segmentation fault.
0xffffffff in ?? ()
(gdb) bt
#0  0xffffffff in ?? ()
#1  0x00000000 in ?? ()
(gdb)
```

parjong:
I figured out why segfault happens. Here is the JIT disasm for Main.
```
...
G_M56529_IG03:
       6800000080   push     0x80000000
       6A00         push     0
       6AFF         push     -1
       6AFF         push     -1
       E8960A9104   call     CORINFO_HELP_LDIV
       EB13         jmp      SHORT G_M56529_IG06
G_M56529_IG04:
       8B0D401170F3 mov      ecx, gword ptr [F3701140H]
       FF153C9AA7F5 call     [System.Console:WriteLine(ref)]
       B864000000   mov      eax, 100
G_M56529_IG05:
       5D           pop      ebp
       C3           ret
G_M56529_IG06:
       8B0D441170F3 mov      ecx, gword ptr [F3701144H]
       FF153C9AA7F5 call     [System.Console:WriteLine(ref)]
       B801000000   mov      eax, 1
G_M56529_IG07:
       5D           pop      ebp
       C3           ret
G_M56529_IG08:
G_M56529_IG09:
       8D05AC24E1F1 lea      eax, G_M56529_IG04
G_M56529_IG10:
       C3           ret
```

``CORINFO_HELP_LDIV`` called inside ``G_M56529_IG03`` throws an exception, and corresponding fault code is located at  G_M56529_IG09.

After executing fault code, CLR resumes execution from G_M56529_IG04 without popping stack argument:
```
Breakpoint 1, RtlRestoreContext () at /home/parjong/projects/dotnet/coreclr/src/pal/src/arch/i386/context2.S:99
[...]

PR #9366, opened 2017-02-06
[x86/Linux] Intreface-based x86 unwinder (Work in progress)

parjong:
This commit revises x86 unwinder based on two interfaces: IUnwindFrameReader and IUnwindFrameListener.

To be specific, this commit revises x86 unwinder to read initial values from IUnwindFrameReader and notify changes during unwinding via IUnwindFrameListener, which allows flexible and uniform implementation for Linux and Windows.

### DIFF
--- a/src/inc/eetwain.h
+++ b/src/inc/eetwain.h
@@ -641,7 +641,35 @@ HRESULT FixContextForEnC(PCONTEXT        pCtx,
 };
 
 #ifdef _TARGET_X86_
-bool UnwindStackFrame(PREGDISPLAY     pContext,
+struct IUnwindFrameReader
+{
+    IUnwindFrameReader() = default;
+    virtual ~IUnwindFrameReader() = default;
+
+    virtual DWORD GetSP(void) = 0;
+    virtual DWORD GetFP(void) = 0;
+    virtual DWORD GetPC(void) = 0;
+};
+
+struct IUnwindFrameListener
+{
+    IUnwindFrameListener() = default;
+    virtual ~IUnwindFrameListener() = default;
+
+    virtual void NotifyEaxLocation(PDWORD loc) = 0;
+    virtual void NotifyEbxLocation(PDWORD loc) = 0;
+    virtual void NotifyEcxLocation(PDWORD loc) = 0;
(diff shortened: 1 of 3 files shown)


jkotas:
I do not think that introducing interface here is a good idea.

janvorli:
I agree with @jkotas.

parjong:
This PR is an attempt to resolve #9272. 

While investigating #9272, I discovered that ``JIT.Regression.CLR-x86-JIT.V2.0-Beta2.b415164`` test also has a similar issue (which occurs from JITed code).

It currently works well as managed unwinder takes care of callee-poped stack elements.
Unfortunately, we can't leave managed unwinder as its current form as we need to update Esp as SP according to the discussion in #9235.

These findings indicate that we need to maintain SP just before call instruction during unwind steps (as discussed in #9235), but we also need SP just after call instruction in resume step to resolve #9272.

It may be possible to achieve this via modifying unwinder and REGDISPLAY, but I'm afraid side-effect on Windows as x86 unwinder is shared between Windows and Linux.

This PR (interface-based unwinder) is an attempt to isolate that side effect.

Please let me know your opinion. It would be very helpful if we could find a better approach.

jkotas:
> we need to maintain SP just before call instruction for unwind 

It feels pretty unnatural for unwinder to maintain SP before call instruction. I do not think that the fix for the problem you are hitting should be to redo how unwinding works. The SP before call instruction is not even well defined in the presence of tailcalls with callee-pop calling convention.

Instead, the places that do not work with SP after the call should be fixed to work with it instead.

parjong:
@jkotas CLR ABI document states that Caller SP as a stack pointer that call instruction happens. I'm not sure, but it seems that this indicates that the unwinder should maintain SP before call instruction. Do I understand correctly? 

It seems that all the architectures except x86 uses caller-pop calling convention, and thus SP before call instruction should be same as SP after call instruction.

jkotas:
`Caller-SP` is problematic on platforms with caller-pop calling convention.

Can we use Initial-SP for establisher frame like on amd64?

parjong:
@jkotas Sorry for late response. I'm not sure about ``Initial-SP`` as wrote in #9384. It depends on #9300.

PR #9368, opened 2017-02-06
Two fixes to ConcurrentQueue

stephentoub:
- Changed comparisons to use subtraction to avoid overflow issues on sequence numbers.
- Added an exception message where a TODO was left in the code

cc: @benaadams, @kouvel 

(Same fix as https://github.com/dotnet/corefx/pull/15848)


stephentoub:
Merging to keep in sync with corefx

PR #9375, opened 2017-02-06
Arm64 remove assert gtBlkOpKind == BlkOpKindHelper

jashook:
We will unconditionally assign blkNode->gtBlkOpKind to BlkOpKindHelper in lsraarm64.cpp. Which makes the assert seemingly unnecessary. 

@CarolEidt @briansull ptal
/cc @dotnet/arm64-contrib @dotnet/jit-contrib

### DIFF
--- a/src/jit/lsraarm64.cpp
+++ b/src/jit/lsraarm64.cpp
@@ -1252,11 +1252,10 @@ void Lowering::TreeNodeInfoInitBlockStore(GenTreeBlk* blkNode)
         else
 #endif // 0
         {
-            assert(blkNode->gtBlkOpKind == GenTreeBlk::BlkOpKindHelper);
+            blkNode->gtBlkOpKind = GenTreeBlk::BlkOpKindHelper;
             // The helper follows the regular ABI.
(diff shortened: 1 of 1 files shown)


jashook:
@dotnet-bot test Windows_NT Arm64 Debug

CarolEidt:
The fundamental problem is that `Lowering::LowerBlockStore()` isn't being called in the arm64 case. The fact that the `gtBlkOpKind` is being set in the `TreeNodeInfoInitBlockStore` methods is also a (less significant) bug. The proper fix is to add a call to `LowerBlockStore()` before the call to `TreeNodeInfoInitBlockStore()` as is done in lsraxarch.cpp. For now it doesn't really matter, but longer term we will want to ensure consistency (e.g. when we add support for unrolling), and we want to make sure that these kinds of decisions are consistently made in `Lowering`.
BTW - thanks @jashook for picking this up so quickly!

jashook:
@CarolEidt thank you for the insight. I have opened PR #9385 to correctly address this based on your feedback. I have some testing still outstanding; however, it is good so far.

PR #9384, opened 2017-02-07
[x86/Linux] Fix GetCallerSp

parjong:
x86 unwinder takes care of callee-poped stack elements, and thus Caller's Esp is not same as Caller-SP (discussed in #9366).

This commit revises GetCallerSp to return correct Caller-SP for x86.

### DIFF
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -3099,7 +3099,17 @@ size_t EECodeManager::GetCallerSp( PREGDISPLAY  pRD )
     {
         EnsureCallerContextIsValid(pRD, NULL);
     }
-    return (size_t) (GetSP(pRD->pCallerContext));
+
+    size_t res = GetSP(pRD->pCallerContext);
+
+#ifdef _TARGET_X86_
+    EECodeInfo codeInfo;
+    codeInfo.Init((PCODE) pRD->ControlPC);
+
+    res -= codeInfo.GetCodeManager()->GetStackParameterSize(&codeInfo);
+#endif // _TARGET_X86_
+
+    return res;
 }
 
 #endif // WIN64EXCEPTIONS && !CROSSGEN_COMPILE
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -3100,16 +3100,7 @@ size_t EECodeManager::GetCallerSp( PREGDISPLAY  pRD )
         EnsureCallerContextIsValid(pRD, NULL);
     }
 
-    size_t res = GetSP(pRD->pCallerContext);
-
-#ifdef _TARGET_X86_
-    EECodeInfo codeInfo;
-    codeInfo.Init((PCODE) pRD->ControlPC);
-
-    res -= codeInfo.GetCodeManager()->GetStackParameterSize(&codeInfo);
-#endif // _TARGET_X86_
-
-    return res;
+    return GetSP(pRD->pCallerContext);
 }
 
 #endif // WIN64EXCEPTIONS && !CROSSGEN_COMPILE
--- a/src/vm/i386/cgenx86.cpp
+++ b/src/vm/i386/cgenx86.cpp
@@ -299,7 +299,12 @@ void TransitionFrame::UpdateRegDisplay(const PREGDISPLAY pRD)
 
     MethodDesc * pFunc = GetFunction();
     _ASSERTE(pFunc != NULL);
+
+#ifdef WIN64EXCEPTIONS
+    UpdateRegDisplayHelper(pRD, 0);
+#else
     UpdateRegDisplayHelper(pRD, pFunc->CbStackPop());
+#endif // WIN64EXCEPTIONS
 
     LOG((LF_GCROOTS, LL_INFO100000, "STACKWALK    TransitionFrame::UpdateRegDisplay(ip:%p, sp:%p)\n", pRD->ControlPC, pRD->SP));
 
--- a/src/vm/i386/cgenx86.cpp
+++ b/src/vm/i386/cgenx86.cpp
@@ -328,7 +328,7 @@ void TransitionFrame::UpdateRegDisplayHelper(const PREGDISPLAY pRD, UINT cbStack
     pRD->IsCallerSPValid      = FALSE;
 
     pRD->pCurrentContext->Eip = *PTR_PCODE(pRD->PCTAddr);;
-    pRD->pCurrentContext->Esp = GetSP();
+    pRD->pCurrentContext->Esp = (DWORD)(pRD->PCTAddr + sizeof(TADDR) + cbStackPop);
 
     UpdateRegDisplayFromCalleeSavedRegisters(pRD, regs);
     ClearRegDisplayArgumentAndScratchRegisters(pRD);
--- a/src/vm/i386/cgenx86.cpp
+++ b/src/vm/i386/cgenx86.cpp
@@ -300,11 +300,7 @@ void TransitionFrame::UpdateRegDisplay(const PREGDISPLAY pRD)
     MethodDesc * pFunc = GetFunction();
     _ASSERTE(pFunc != NULL);
 
-#ifdef WIN64EXCEPTIONS
-    UpdateRegDisplayHelper(pRD, 0);
-#else
     UpdateRegDisplayHelper(pRD, pFunc->CbStackPop());
-#endif // WIN64EXCEPTIONS
 
     LOG((LF_GCROOTS, LL_INFO100000, "STACKWALK    TransitionFrame::UpdateRegDisplay(ip:%p, sp:%p)\n", pRD->ControlPC, pRD->SP));
 
@@ -329,11 +325,14 @@ void TransitionFrame::UpdateRegDisplayHelper(const PREGDISPLAY pRD, UINT cbStack
 
 #ifdef WIN64EXCEPTIONS
 
+    DWORD CallerSP = (DWORD)(pRD->PCTAddr + sizeof(TADDR));
+
     pRD->IsCallerContextValid = FALSE;
     pRD->IsCallerSPValid      = FALSE;
 
     pRD->pCurrentContext->Eip = *PTR_PCODE(pRD->PCTAddr);;
-    pRD->pCurrentContext->Esp = (DWORD)(pRD->PCTAddr + sizeof(TADDR) + cbStackPop);
+    pRD->pCurrentContext->Esp = CallerSP;
+    pRD->pCurrentContext->ResumeEsp = CallerSP + cbStackPop;
 
     UpdateRegDisplayFromCalleeSavedRegisters(pRD, regs);
     ClearRegDisplayArgumentAndScratchRegisters(pRD);
@@ -386,7 +385,7 @@ void HelperMethodFrame::UpdateRegDisplay(const PREGDISPLAY pRD)
 #endif // DACCESS_COMPILE
 
     pRD->pCurrentContext->Eip = pRD->ControlPC = m_MachState.GetRetAddr();
-    pRD->pCurrentContext->Esp = pRD->SP = (DWORD) m_MachState.esp();
+    pRD->pCurrentContext->Esp = pRD->pCurrentContext->ResumeEsp = pRD->SP = (DWORD) m_MachState.esp();
 
 #define CALLEE_SAVED_REGISTER(regname) pRD->pCurrentContext->regname = *((DWORD*) m_MachState.p##regname());
     ENUM_CALLEE_SAVED_REGISTERS();
@@ -686,7 +685,8 @@ void InlinedCallFrame::UpdateRegDisplay(const PREGDISPLAY pRD)
     pRD->IsCallerSPValid      = FALSE;        // Don't add usage of this field.  This is only temporary.
 
     pRD->pCurrentContext->Eip = *PTR_PCODE(pRD->PCTAddr);
-    pRD->pCurrentContext->Esp = (DWORD) dac_cast<TADDR>(m_pCallSiteSP) + stackArgSize;
+    pRD->pCurrentContext->Esp = (DWORD) dac_cast<TADDR>(m_pCallSiteSP);
+    pRD->pCurrentContext->ResumeEsp = (DWORD) dac_cast<TADDR>(m_pCallSiteSP) + stackArgSize;
     pRD->pCurrentContext->Ebp = (DWORD) m_pCalleeSavedFP;
 
     ClearRegDisplayArgumentAndScratchRegisters(pRD);
@@ -823,7 +823,7 @@ void HijackFrame::UpdateRegDisplay(const PREGDISPLAY pRD)
     pRD->IsCallerSPValid      = FALSE;        // Don't add usage of this field.  This is only temporary.
 
     pRD->pCurrentContext->Eip = *PTR_PCODE(pRD->PCTAddr);
-    pRD->pCurrentContext->Esp = (DWORD)(pRD->PCTAddr + sizeof(TADDR));
+    pRD->pCurrentContext->Esp = pRD->pCurrentContext->ResumeEsp = (DWORD)(pRD->PCTAddr + sizeof(TADDR));
 
 #define CALLEE_SAVED_REGISTER(reg) pRD->pCurrentContext->reg = m_Args->reg;
     ENUM_CALLEE_SAVED_REGISTERS();
@@ -902,7 +902,7 @@ void TailCallFrame::UpdateRegDisplay(const PREGDISPLAY pRD)
     pRD->IsCallerSPValid      = FALSE;        // Don't add usage of this field.  This is only temporary.
 
     pRD->pCurrentContext->Eip = *PTR_PCODE(pRD->PCTAddr);
-    pRD->pCurrentContext->Esp = (DWORD)(pRD->PCTAddr + sizeof(TADDR));
+    pRD->pCurrentContext->Esp = pRD->pCurrentContext->ResumeEsp = (DWORD)(pRD->PCTAddr + sizeof(TADDR));
 
     UpdateRegDisplayFromCalleeSavedRegisters(pRD, &m_regs);
     ClearRegDisplayArgumentAndScratchRegisters(pRD);
--- a/src/pal/inc/pal.h
+++ b/src/pal/inc/pal.h
@@ -1782,6 +1782,7 @@ typedef struct _CONTEXT {
 
     UCHAR   ExtendedRegisters[MAXIMUM_SUPPORTED_EXTENSION];
 
+    ULONG   ResumeEsp;
 } CONTEXT, *PCONTEXT, *LPCONTEXT;
 
 // To support saving and loading xmm register context we need to know the offset in the ExtendedRegisters
--- a/src/pal/src/arch/i386/asmconstants.h
+++ b/src/pal/src/arch/i386/asmconstants.h
@@ -28,3 +28,4 @@
 #define CONTEXT_Xmm5 CONTEXT_Xmm4+16
 #define CONTEXT_Xmm6 CONTEXT_Xmm5+16
 #define CONTEXT_Xmm7 CONTEXT_Xmm6+16
+#define CONTEXT_ResumeEsp CONTEXT_ExtendedRegisters+512
--- a/src/pal/src/arch/i386/context2.S
+++ b/src/pal/src/arch/i386/context2.S
@@ -42,6 +42,7 @@ LEAF_ENTRY CONTEXT_CaptureContext, _TEXT
     mov   [eax + CONTEXT_Ebp], ebp
     lea   ebx, [esp + 12]
     mov   [eax + CONTEXT_Esp], ebx
+    mov   [eax + CONTEXT_ResumeEsp], ebx
     mov   ebx, [esp + 8]
     mov   [eax + CONTEXT_Eip], ebx
 
@@ -114,7 +115,7 @@ LOCAL_LABEL(Done_Restore_CONTEXT_FLOATING_POINT):
 LOCAL_LABEL(Done_Restore_CONTEXT_EXTENDED_REGISTERS):
 
     // Restore Stack
-    mov   esp, [eax + CONTEXT_Esp]
+    mov   esp, [eax + CONTEXT_ResumeEsp]
 
     // Create a minimal frame
     push  DWORD PTR [eax + CONTEXT_Eip]
--- a/src/pal/src/arch/i386/exceptionhelper.S
+++ b/src/pal/src/arch/i386/exceptionhelper.S
@@ -23,7 +23,7 @@ LEAF_ENTRY ThrowExceptionFromContextInternal, _TEXT
         mov   ebx, [esp + 8]  // eax: CONTEXT *
 
         mov   ebp, [ebx + CONTEXT_Ebp]
-        mov   esp, [ebx + CONTEXT_Esp]
+        mov   esp, [ebx + CONTEXT_ResumeEsp]
 
         // The ESP is re-initialized as the target frame's value, so the current function's
         // CFA is now right at the ESP.
--- a/src/pal/src/exception/seh-unwind.cpp
+++ b/src/pal/src/exception/seh-unwind.cpp
@@ -155,6 +155,7 @@ static void UnwindContextToWinContext(unw_cursor_t *cursor, CONTEXT *winContext)
 #elif defined(_X86_)
     unw_get_reg(cursor, UNW_REG_IP, (unw_word_t *) &winContext->Eip);
     unw_get_reg(cursor, UNW_REG_SP, (unw_word_t *) &winContext->Esp);
+    unw_get_reg(cursor, UNW_REG_SP, (unw_word_t *) &winContext->ResumeEsp);
     unw_get_reg(cursor, UNW_X86_EBP, (unw_word_t *) &winContext->Ebp);
     unw_get_reg(cursor, UNW_X86_EBX, (unw_word_t *) &winContext->Ebx);
     unw_get_reg(cursor, UNW_X86_ESI, (unw_word_t *) &winContext->Esi);
--- a/src/pal/src/thread/context.cpp
+++ b/src/pal/src/thread/context.cpp
@@ -499,11 +499,13 @@ void CONTEXTFromNativeContext(const native_context_t *native, LPCONTEXT lpContex
     if ((contextFlags & CONTEXT_CONTROL) == CONTEXT_CONTROL)
     {
         ASSIGN_CONTROL_REGS
-#ifdef _ARM_
+#if defined(_ARM_)
         // WinContext assumes that the least bit of Pc is always 1 (denoting thumb)
         // although the pc value retrived from native context might not have set the least bit.
         // This becomes especially problematic if the context is on the JIT_WRITEBARRIER.
         lpContext->Pc |= 0x1;
+#elif defined(_X86_)
+        lpContext->ResumeEsp = MCREG_Esp(native->uc_mcontext);
 #endif
     }
 
@@ -945,6 +947,7 @@ CONTEXT_GetThreadContextFromThreadState(
                 lpContext->Esi = pState->esi;
                 lpContext->Ebp = pState->ebp;
                 lpContext->Esp = pState->esp;
+                lpContext->ResumeEsp = pState->esp;
                 lpContext->SegSs = pState->ss;
                 lpContext->EFlags = pState->eflags;
                 lpContext->Eip = pState->eip;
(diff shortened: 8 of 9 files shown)


janvorli:
@parjong It looks to me that it would be better to keep the EstablisherFrame and GetCallerSP so that they match what is stored in the REGDISPLAY and instead compensate for the stack arguments when getting the GS cookie address. Would that work?
And have you considered using the Initial-SP instead of Caller-SP as @jkotas has suggested in #9366 for the establisher frame?

parjong:
@janvorli It currently has no issue to use of initial-SP instead of caller-SP as Establish Frame Pointer, but it will have an impact on #9300 (@wateret are currently working on). #9300 currently assumes that the value Establish Frame Pointer is Caller-Sp, and use it to restore Ebp. I currently have no idea whether it is a good idea to use initial-SP instead of caller-SP.

In its current form, ``GetCallerSp`` will return caller's Initial-SP instead of Caller-SP (according to CLR ABI). I could find several places that invokes ``GetCallerSp``, and thus I'm not sure whether this semantic difference affects these code, or not.

\CC @jkotas @seanshpark

janvorli:
@parjong I went through all of the usages of GetCallerSp and there are just two that matter for x86 on Linux. In the EECodeManager::GetGSCookieAddr and in the ExceptionTracker::ProcessOSExceptionNotification. For the latter, the only thing that matters is that it matches the way the establisher frame is reported.

janvorli:
@parjong one more thing. The CLR ABI talks about the caller SP only for amd64 and arm, since on x86 before the changes to use WIN64EXCEPTIONS on Linux, there was no concept of caller SP. So we can actually define the semantics of the caller SP on x86 Linux. We can define it as the SP before the caller started to push parameters for the call.

jkotas:
> We can define it as the SP before the caller started to push parameters for the call.

I think it would be better to just define it same way as it is defined on amd64; to reduce the variety of options.

janvorli:
@jkotas ok, in that case the spirit of the current change is doing exactly that.

src/vm/eetwain.cpp
```diff
@@ -3099,7 +3099,17 @@ size_t EECodeManager::GetCallerSp( PREGDISPLAY  pRD )
...
     }
-    return (size_t) (GetSP(pRD->pCallerContext));
+
+    size_t res = GetSP(pRD->pCallerContext);
+
+#ifdef _TARGET_X86_
+    EECodeInfo codeInfo;
+    codeInfo.Init((PCODE) pRD->ControlPC);
+
+    res -= codeInfo.GetCodeManager()->GetStackParameterSize(&codeInfo);
```
janvorli: What if we stored the callerSP in the REGDISPLAY in the OOPStackUnwinderX86::VirtualUnwind when we set the EstablisherFrame? Then we would not have to initialize the EECodeInfo and call the GetStackParameterSize here again.

src/vm/i386/cgenx86.cpp
```diff
@@ -299,7 +299,12 @@ void TransitionFrame::UpdateRegDisplay(const PREGDISPLAY pRD)
 
     MethodDesc * pFunc = GetFunction();
     _ASSERTE(pFunc != NULL);
+
+#ifdef WIN64EXCEPTIONS
+    UpdateRegDisplayHelper(pRD, 0);
```
janvorli: I am not sure I understand why the behavior here is different for WIN64EXCEPTIONS. This UpdateRegDisplay is essentially a simple unwind to the caller context. So I would expect it to behave as the UnwindStackFrame w.r.t. the semantics of the value it sets to the SP.

src/vm/eetwain.cpp
parjong: @janvorli Then, CLR will get into stuck when resuming (please refer to 4th comment in #9366 for details).

src/vm/i386/cgenx86.cpp
parjong: @janvorli You're right. It would be better to use FEATURE_PAL (instead of WIN64EXCEPTIONS) here.

As discussed in #9272, libunwind does not consider calling convention at all, and stack frames from native method such as ``ThePreStub`` is unwound by libunwind.

As a result, stack walker failed to find the target frame as the esp comes from libuwnind and the esp updated by this method are different.

src/vm/eetwain.cpp
janvorli: @parjong I am confused by your reply. My comment was not suggesting a change in the behavior, but rather "caching" the value we assign to EstablisherFrame in the OOPStackUnwinderX86::VirtualUnwind, since that's exactly the value we would compute here. So the point was that we don't have to recompute it here that way.

parjong:
@janvorli FYI, 

The unittest result with this PR:
```
=======================
     Test Results
=======================
# CoreCLR Bin Dir  : 
# Tests Discovered : 7027
# Passed           : 5920
# Failed           : 828
# Skipped          : 279
=======================
```

The unittest result with this PR revised as you suggested (change only GetGSCookieAddress):
```
=======================
     Test Results
=======================
# CoreCLR Bin Dir  : 
# Tests Discovered : 7027
# Passed           : 5912
# Failed           : 836
# Skipped          : 279
=======================
```

I'm not sure whether there is actual difference as there are some random failures.

janvorli:
@parjong thank you for the rest results. Based on the comment from @jkotas, I have actually changed my mind to keep the CallerSP semantics consistent with AMD64. That's why I have started to add the comments to the specific parts of your change.

src/vm/i386/cgenx86.cpp
janvorli: @parjong I am also not sure I understand your reply here. The TransitionFrame::UpdateRegDisplay effectively unwinds to a managed method. So I was assuming that we would use the `UpdateRegDisplayHelper(pRD, pFunc->CbStackPop())` for WIN64EXCEPTIONS too so that pRD->pCurrentContext->Esp gets ESP before the stack arguments were pushed, like the UnwindStackFrame does.

src/vm/eetwain.cpp
parjong: @janvorli Oh, I'm sorry, I misunderstood your comment. It will work if your suggestion was to add ``DWORD CallerSp`` field in ``REGDISPLAY``. I thought that your suggestion was to update Esp as Caller SP.

janvorli: No problem. Yes, I meant adding `CallerSp` field in `REGDISPLAY`

src/vm/i386/cgenx86.cpp
parjong: @janvorli I'm sorry for unclear comment.

Suppose that a managed method ``A`` calls a  native method, and this native method throws an exception that ``A`` will handle.

``FindNonvolatileRegisterPointers`` attempts to find the frame of ``A`` using ``uOriginalSP`` which comes from ``Thread::VirtualUnwindToFirstManagedCallFrame``.

Note that the last step of ``VirtualUnwindToFirstManagedCallFrame`` will be ``PAL_VirtualUnwind`` (instead of ``UnwindStackFrame``) which does not consider callee-poped stack elements, and thus  ``uOriginalSP`` will be exactly same as Caller-SP of the native method.

(As I understand) stack walker will get the stack frame of ``A`` via calling `TransitionFrame::UpdateRegDisplay``, and the SP of ``A`` will be native method's Caller-SP + the result of CbStackPop() in the previous form.

Due to this mismatch, stack walker failed to recognize ``A``'s frame, which results in an assertion failure inside ``FindNonvolatileRegisterPointers``.

Please let me know if there is something wrong. I find all of these steps via execution log, but there may be something wrong.

parjong:
@janvorli I understand. I did this test before you changed your mind, and I thought that this result may be useful for someone else.

src/vm/eetwain.cpp
parjong: @janvorli Oops, I found a problem. ``rd`` in ``OOPStackUnwinderX86::VirtualUnwind`` is a local variable, and thus the update on ``CallerSp`` here has no effect.

janvorli: Ah, you are right. I've forgotten about it.

src/vm/i386/cgenx86.cpp
janvorli: Hmm, I was looking at it from the amd64 prespective, where the RIP and RSP are extracted in a different way.
The TransitionFrame (resp. the frame types derived from it, since TransitionFrame is an abstract base) is created for transition from managed code. Stack for one of the cases (taken from the PinnedObject test) is below:
```
00007FFFFFFFA000 00007FFFF630460F libcoreclr.so!PreStubWorker + 1935 at /home/janvorli/git/coreclr/src/vm/prestub.cpp:1078
00007FFFFFFFA228                  [PrestubMethodFrame: 00007fffffffa228] System.Reflection.Metadata.dll!System.Reflection.Internal.FileStreamReadLightUp+NativeMethods.ReadFileModern(System.Runtime.InteropServices.SafeHandle, Byte*, Int32, Int32 ByRef, IntPtr)
00007FFFFFFFA270 00007FFFF61EC9BC libcoreclr.so!ThePreStub + 92 at /home/janvorli/git/coreclr/bin/obj/Linux.x64.Debug/src/vm/wks/theprestubamd64.S:2
00007FFFFFFFA360 00007FFF7D7AC7F2 System.Reflection.Metadata.dll!System.Reflection.Internal.FileStreamReadLightUp.TryReadFile(System.IO.Stream, Byte*, Int64, Int32) + 274 [/home/janvorli/git/corefx/src/System.Reflection.Metadata/src/System/Reflection/Internal/Utilities/FileStreamReadLightUp.cs @ 111]
```
`ThePreStub` asm helper creates a data structure called transition block on the stack. This block contains callee saved and argument registers. The transition block is created right below the return address to the managed caller of the `ThePreStub`. `The PrestubMethodFrame` (derived from the `TransitionFrame`) is created in the `PreStubWorker` and it contains a pointer to the transition block. 
On x64, the RIP is set to the return address to the managed caller and the RSP is set to the end of the transition block plus sizeof(void*) (skipping the return address).
So, these values correspond to the frame of the managed caller of the `ThePreStub`. 
That means that on x86 Linux, the ESP should be the one before the arguments were pushed to be consistent with the cases when we unwind to a managed frame from another managed frame directly. Otherwise it seems that the next unwind would fail if the managed function that called the `ThePreStub` had stack arguments.

Now I also wonder why we extract the pRD->pCurrentContext->Eip and pRD->pCurrentContext->Esp below in a different way than on amd64 (using pRD->PCTAddr instead). 
What I mean, why don't we do it like this:
```
pRD->pCurrentContext->Eip = GetReturnAddress();
pRD->pCurrentContext->Esp = GetSP() + cbStackPop;
```

wateret:
@janvorli @jkotas I'm working on a new version of #9300 with PSPSym support. 
Basically I was going to write it same way with ARM but I'm still not sure which to use for establisher frame pointer, CallerSP or InitialSP. As far as I understand, CallerSP could be problematic in x86 so InitialSP would be a safer way.
Am I right? If so, would it be OK to follow ARM ABI but InitialSP as establisher frame pointer?

src/vm/i386/cgenx86.cpp
parjong: I encountered the following assertion failure when I revised the code as you suggested:
```
FAILED   - [ 026 :    5 sec ] GC/API/GC/GetGeneration_fail/GetGeneration_fail.sh
               BEGIN EXECUTION
               /root/dotnet-test/coreclr_unittest_20170208_193938_x86_x86-emu/coreclr-unittest-170208-104248/overlay/corerun GetGeneration_fail.exe
               This test should throw an exception!
               Generation: 0

               Assert failure(PID 1078 [0x00000436], Thread: 1078 [0x0436]): Consistency check failed: FAILED: state.fFound
                   File: /home/parjong/projects/dotnet/coreclr/src/vm/exceptionhandling.cpp Line: 360
                   Image: /root/dotnet-test/coreclr_unittest_20170208_193938_x86_x86-emu/coreclr-unittest-170208-104248/overlay/corerun
```

Here is the state just before that assertion failure happens (inside ``TransitionFrame::UpdateRegDisplayHelper``)
```
TID 1e30: CEHelper::CanMethodHandleException - Processing CorruptionSeverity: 2.
TID 1e30: ....  RESUMEFRAME:  sf is  FFFFA7C8 and  m_sfResumeStackFrame: FFFFA7C8
TID 1e30: ....  RESUMEFRAME:  REACHED initial resume frame: FFFFA7C8
TID 1e30: ....  RESUMEFRAME:  Found last frame to process finallys in, need to process only part of call frame
TID 1e30: ....  RESUMEFRAME:  Limit clause token: EFEB049C
TID 1e30: ....  considering typed clause [b6,e2], ControlPc is inside clause (offset ce)
TID 1e30: ....  found limit clause, stopping clause enumeration
TID 1e30: ....  second pass unwind completed
TID 1e30: ....  calling catch at 0xEFEB5AA2
TID 1e30: ....    calling handler at 0xEFEB5AA2, sp = 0xFFFFA7C8
TID 1e30: ....  resume address should be 0xEFEB5A12
TID 1e30: ....second pass finished
TID 1e30: ....cleaning up ExceptionTracker state
TID 1e30: ....Unlinking ExceptionTracker object 0x08088A14, thread = 0x0808A9F0
TID 1e30: DestroyHandle: *F62F10BC->F2F3F500
TID 1e30: ..TrackerAllocator: freeing tracker 0x08088A14, thread = 0x0808A9F0
TID 1e30: CreateHandle: F62F10BC, type=2
TID 1e30: DestroyHandle: *F62F10C0->F2F3F500
TID 1e30: CEHelper::ResetLastActiveCorruptionSeverityPostCatchHandler - Reset LastActiveException corruption severity to 2.
TID 1e30: ..FinishSecondPass complete, uResumePC = EFEB5A12, current SP = FFFF94BC
TID 1e30: STACKWALK: [000] FILTER  : EXPLICIT : PC= 00000000  SP= 00000000  Frame= ffffa70c  vtbl= f7c5fc98
TID 1e30: STACKWALK: [001] CONSIDER: EXPLICIT : PC= 00000000  SP= 00000000  Frame= ffffa70c  vtbl= f7c5fc98
TID 1e30: STACKWALK: [001] CALLBACK: EXPLICIT : PC= 00000000  SP= 00000000  Frame= ffffa70c  vtbl= f7c5fc98

Breakpoint 1, TransitionFrame::UpdateRegDisplayHelper (this=0xffffa70c, pRD=0xffff8dd8, cbStackPop=12) at /home/parjong/projects/dotnet/coreclr/src/vm/i386/cgenx86.cpp:324
324         CalleeSavedRegisters* regs = GetCalleeSavedRegisters();
(gdb) p/x GetSP()
$1 = 0xffffa758
(gdb) bt 20
#0  TransitionFrame::UpdateRegDisplayHelper (this=0xffffa70c, pRD=0xffff8dd8, cbStackPop=12) at /home/parjong/projects/dotnet/coreclr/src/vm/i386/cgenx86.cpp:324
#1  0xf7126a52 in TransitionFrame::UpdateRegDisplay (this=0xffffa70c, pRD=0xffff8dd8) at /home/parjong/projects/dotnet/coreclr/src/vm/i386/cgenx86.cpp:305
#2  0xf6df7cd6 in StackFrameIterator::NextRaw (this=0xffff89d0) at /home/parjong/projects/dotnet/coreclr/src/vm/stackwalk.cpp:2642
#3  0xf6df32f1 in StackFrameIterator::Next (this=0xffff89d0) at /home/parjong/projects/dotnet/coreclr/src/vm/stackwalk.cpp:1612
#4  0xf6df2a55 in Thread::StackWalkFramesEx (this=0x808a9f0, pRD=0xffff8dd8, pCallback=0xf710e240 <UpdateObjectRefInResumeContextCallback(CrawlFrame*, void*)>, pData=0xffff8d58, flags=0, pStartFrame=0xffffa70c)
    at /home/parjong/projects/dotnet/coreclr/src/vm/stackwalk.cpp:963
#5  0xf710e711 in ExceptionTracker::FindNonvolatileRegisterPointers (pThread=0x808a9f0, uOriginalSP=4294944600, pRegDisplay=0xffff8dd8, uResumeFrameFP=4294944704)
    at /home/parjong/projects/dotnet/coreclr/src/vm/exceptionhandling.cpp:342
#6  0xf7114ee3 in FixNonvolatileRegisters (uOriginalSP=4294944600, pThread=0x808a9f0, pContextRecord=0xffff9f28, fAborting=false) at /home/parjong/projects/dotnet/coreclr/src/vm/exceptionhandling.cpp:1217
#7  0xf7112223 in ProcessCLRException (pExceptionRecord=0x811f52c, MemoryStackFp=4294944712, pContextRecord=0xffff9f28, pDispatcherContext=0xffff9a78)
    at /home/parjong/projects/dotnet/coreclr/src/vm/exceptionhandling.cpp:1136
#8  0xf711c63e in UnwindManagedExceptionPass2 (ex=..., unwindStartContext=0xffff9f28) at /home/parjong/projects/dotnet/coreclr/src/vm/exceptionhandling.cpp:4382
#9  0xf711cfdf in UnwindManagedExceptionPass1 (ex=..., frameContext=0xffffa2b0) at /home/parjong/projects/dotnet/coreclr/src/vm/exceptionhandling.cpp:4536
#10 0xf711da77 in DispatchManagedException (ex=..., isHardwareException=false) at /home/parjong/projects/dotnet/coreclr/src/vm/exceptionhandling.cpp:4643
#11 0xf7298920 in PreStubWorker (pTransitionBlock=0xffffa73c, pMD=0xefec9970) at /home/parjong/projects/dotnet/coreclr/src/vm/prestub.cpp:1078
#12 0xf713440f in ThePreStub () at /home/parjong/projects/dotnet/coreclr/src/vm/i386/asmhelpers.S:921
#13 0xefeb59ff in ?? ()
#14 0xefeb57b3 in ?? ()
#15 0xefeb56b8 in ?? ()
#16 0xefeb553c in ?? ()
#17 0xf267459a in ?? ()
#18 0xf2673b64 in ?? ()
#19 0xf266dfa2 in ?? ()
(More stack frames follow...)
(gdb) info frame 12
Stack frame at 0xffffa758:
 eip = 0xf713440f in ThePreStub (/home/parjong/projects/dotnet/coreclr/src/vm/i386/asmhelpers.S:921); saved eip = 0xefeb59ff
 called by frame at 0xffffa7c8, caller of frame at 0xffffa730
 source language asm.
 Arglist at 0xffffa750, args:
 Locals at 0xffffa750, Previous frame's sp is 0xffffa758
 Saved registers:
  ecx at 0xffffa740, edx at 0xffffa73c, ebx at 0xffffa74c, ebp at 0xffffa750, esi at 0xffffa748, edi at 0xffffa744, eip at 0xffffa754
(gdb) p/x 4294944600
$2 = 0xffffa758
(gdb)
```

janvorli: @parjong It seems that the GetSP() result matches the Previous frame's sp is 0xffffa758 reported by gdb, so it looks correct. Does the pRD->pCurrentContext->Eip match the frame 13 (also reported as "saved eip" in the result of info frame 12?
And, how do these eip / esp values compare to the ones extracted using the former way?

jkotas:
Yes, I think Initial-SP would be the right choice for establisher frame on x86. What are parts of the ARM ABI that you are thinking about following for x86? It may be useful if you could send PR with just edits and clarifications in the relevant chapter of clr-abi.md so that we can get the details nailed.

src/vm/i386/cgenx86.cpp
janvorli: @parjong a little correction to the scenario you have described (managed method A calls a native method B, and B throws a managed exception that A will handle). There doesn't have to be the transition frame. If the native method is a FCall like e.g. IL_Throw, there is the HelperMethodFrame instead. 

I believe the fix to the issue with the result from VirtualUnwindToFirstManagedCallFrame not matching the other one should be to make everything consistent. That means, if we have a context of a managed frame, the ESP should always have the same semantics. No matter whether we got it from the native unwinder that has reached the first managed frame or if we got it from managed unwinder that has unwound from one managed frame to another one.
As we have already discussed here, there are two options. Either modify the OOPStackUnwinderX86::VirtualUnwind so that it returns context with ESP being the ESP at the call instruction or augment the result of the PAL_VirtualUnwind when it reaches a managed frame so that the ESP is the ESP before the stack arguments were pushed.

Having context for managed frames with the ESP semantic sometimes one way and sometimes the other is error prone and hard to reason about.

parjong: @janvorli Here is the value that you mentioned:
```
TID 1e85: ..FinishSecondPass complete, uResumePC = EFEB5A12, current SP = FFFF94BC
TID 1e85: STACKWALK: [000] FILTER  : EXPLICIT : PC= 00000000  SP= 00000000  Frame= ffffa70c  vtbl= f7c5fc98
TID 1e85: STACKWALK: [001] CONSIDER: EXPLICIT : PC= 00000000  SP= 00000000  Frame= ffffa70c  vtbl= f7c5fc98
TID 1e85: STACKWALK: [001] CALLBACK: EXPLICIT : PC= 00000000  SP= 00000000  Frame= ffffa70c  vtbl= f7c5fc98
336         UpdateRegDisplayFromCalleeSavedRegisters(pRD, regs);
(gdb) p/x pRD->pCurrentContext->Eip
$1 = 0xefeb59ff
(gdb) p/x pRD->pCurrentContext->Esp
$2 = 0xffffa764
(gdb) bt
...
#12 0xf713440f in ThePreStub () at /home/parjong/projects/dotnet/coreclr/src/vm/i386/asmhelpers.S:921
#13 0xefeb59ff in ?? ()
(gdb) info frame 12
Stack frame at 0xffffa758:
 eip = 0xf713440f in ThePreStub (/home/parjong/projects/dotnet/coreclr/src/vm/i386/asmhelpers.S:921); saved eip = 0xefeb59ff
 called by frame at 0xffffa7c8, caller of frame at 0xffffa730
 source language asm.
 Arglist at 0xffffa750, args:
 Locals at 0xffffa750, Previous frame's sp is 0xffffa758
 Saved registers:
  ecx at 0xffffa740, edx at 0xffffa73c, ebx at 0xffffa74c, ebp at 0xffffa750, esi at 0xffffa748, edi at 0xffffa744, eip at 0xffffa754
...

parjong: @janvorli The first approach (revise ``OOPStackUnwinderX86::VirtualUnwind`` to update ``ESP`` as ``Caller-SP``) will make CLR get into stuck when resuming. Thus, we need to additionally maintain SP for resume.

I'm not sure whether it is possible to take the second approach. The discussion in #9272 shows that libunwind does not consider calling convention for unwinding. As the issue happens when we unwind native method frame, we could not use the existing infrastructure for managed methods.

wateret:
@jkotas What I meant by ARM ABI, specifically, everything in [The PSPSym and funclet parameters](https://github.com/dotnet/coreclr/blob/master/Documentation/botr/clr-abi.md#the-pspsym-and-funclet-parameters) for ARM/ARM64 and EH funclet convention as discussed in https://github.com/dotnet/coreclr/pull/9300#discussion_r99286009.

What I thought was, there are two different conventions: ARM/ARM64 and AMD64. I thought AMD64 is a special case so it would be better to follow ARM/ARM64 ABI. However we've decided to use Initial-SP  as establisher frame which is like AMD64 ABI so I wanted to clarify the rest of ABI.

jkotas:
Well, the ARM/ARM64 is same because of it was easy choice to copy ARM to ARM64. Both AMD64 and ARM/ARM64 have grown somewhat organically.

There is actually a third variant used by CoreRT on all platforms. It was intentionally designed, taking the best parts from both AMD64 and ARM/ARM64. It would be nice to snap to CoreRT convention for Linux x86 where possible. 

- First argument of all funclets is establisher frame, second argument is exception object (where applicable)
- For filter funclets the VM sets the frame register to be the same as the parent function. For second pass funclets the VM restores all non-volatile registers.
- No PSP sym (it may be hard to not have PSP sym for CoreCLR - it is ok to skip it)

src/vm/i386/cgenx86.cpp
janvorli: @parjong I have spent some time today thinking about all of the aspects of the problem. Here are my notes:
1. We cannot store the esp at the call instruction in the context, unless we change the UnwindStackFrame to work with that semantics of esp directly. I don't see a way how OOPStackUnwinderX86::VirtualUnwind could augment it back to the SP before pushing the stack args. The problem is that the unwind gets the current context, but for getting the size of the stack arguments, we would need the previous context that we've unwound to the current context before. And that's not available at that point. 
2. There are actually two different ways the stack is unwound. 
  a. One is in the stackwalk.cpp, where it doesn't directly walk native code frames, but it uses the explicit stack frames (classes derived from the Frame class) to cross the native area. That's where the XXXFrame::UpdateRegDisplay is used. Frames derived from the TransitionFrame contain MethodDesc of the function that gets called after the transition. There is a managed method A that calls a native code with the TransitionFrame and then a managed method B is called. But logically, you can see it as method A calling method B. And so the stack arguments size from the MethodDesc of method B stored in the transition frame gives you the delta between the caller SP stored in the regdisplay passed to the UpdateRegDisplay and the SP where the stack arguments were pushed. So the TransitionFrame::UpdateRegDisplay should always call the UpdateRegDisplayHelper with the pFunc->CbStackPop(), even for WIN64EXCEPTIONS in the current model when the context of managed frames contains esp before the stack arguments were pushed and not the caller SP. First for the sake of consistency and second  because calling the EECodeManager::GetCallerSp on REGDISPLAY of the method A would otherwise return nonsense.
  b. The other one is in the UnwindManagedExceptionPassX and Thread::VirtualUnwindToFirstManagedCallFrame where we walk every single native frame. In this case, the PAL_VirtualUnwind is used and so when we walk out of native frames to a managed frame, the esp is obviously the caller SP and not the SP before the stack arguments were pushed. So at this place we would need to have a way to lookup the size of stack arguments from an address in the function. It seems we could use the information in the jithelpers.h to build some kind of lookup.

janvorli: @parjong I just want to confirm something. I think you've said before that the UnwindStackFrame didn't work if you have tried to call it with the esp set to the caller SP instead of the SP before the stack arguments were pushed. Am I right? And if yes, do you remember what was the issue?

I am asking since reading through the unwinder code, it seems to me that it could be possible to modify it (using ifdefs only for Linux) so that it would be able to unwind in the same way as the libunwind. That would remove the mismatch between these two unwinders, both would unwind to the caller SP.

janvorli: @parjong please ignore my last comment. Reading through the unwinder code more, I've found that it would be difficult to make such a change in a reasonable way.

parjong: @janvorli When I set ``Esp`` as ``Caller-SP`` inside ``UnwindStackFrame``, ``JIT.Regression.CLR-x86-JIT.V2.0-Beta2.b415164`` get stuck (discussed in #9366). It is almost same as #9272, but the callee is managed one instead of native one.

Actually, #9366 was an attempt to implement libunwind-like unwinder without affecting x86/Windows.

parjong:
@janvorli I currently consider the following changes to store both SP before call and SP after call via adding ``ResumeEsp`` in ``CONTEXT``:
 - Revise ``OOPStackUnwinderX86::VirtualUnwind`` to update both ``ResumeEsp`` and ``Esp`` ( ``OOPStackUnwinderX86::VirtualUnwind``  unwinds a managed frame (a frame created by a managed method), and thus it could know both ``ResumeEsp`` and ``Esp`` after unwind)
 - Revise ``TransitionFrame::UpdateRegDisplay`` to update both ``ResumeEsp`` and ``Esp`` (``cbStackPop`` allows us to update ``ResumeEsp`` correctly)
 - Revise ``PAL_VirtualUnwind`` to update both ``ResumeEsp`` and ``Esp`` as ``Caller-SP`` (libunwind does not know ``ResumeEsp``)
 - Revise ``RtlCaptureContext`` to update both ``ResumeEsp`` and ``Esp`` as current SP)
 - Revise ``RtlRestoreContext`` to use ``ResumeEsp`` instead of ``Esp``

I think that the above changes allows us to get correct ``ResumeEsp`` and ``Esp`` for managed frames except one case: unwound from FCall as you pointed out. As discussed in #9272, we need to introduce additional adjustment after ``PAL_VirtualUnwind`` to deal with such cases.

Could you let me know your opinion on this approach?

janvorli:
@parjong the plan looks reasonable to me. So basically, Context.Esp would always be the caller context and Context.ResumeEsp would be equivalent to the caller context for native frames and it would be the Esp before pushing the arguments for managed frames, right?

parjong:
@janvorli Yes, ``Esp`` should be same as libunwind unwinder's SP, and ``ResumeEsp`` should be same as CLR-internal unwinder's SP.

janvorli:

parjong:

parjong:

parjong:
@janvorli Could you please merge this PR?

PR #9385, opened 2017-02-07
Call LowerBlockStore TreeNodeInfoInitBlockStore

jashook:
For arm64 call LowerBlockStore before TreeNodeInfoInitBlockStore
as done on other architectures. See #9375 for more history.


jashook:
@CarolEidt ptal

RussKeldorph:
@BruceForstall This change should fix the default ARM64 jobs.

@dotnet/jit-contrib

PR #9389, opened 2017-02-07
[x86/Linux] Use SP as a stack marker

### LINKED ISSUE #9378: [x86/Linux] Type.GetType() returns null
Current `Type.GetType(string name)` returns null. `GC/Scenarios/ReflectObj` fail cause of this.

Test code
```
namespace TestApp
{
    public class Program
    {
        public static void Main(string[] args)
        {
            Type type = Type.GetType("TestApp.Program");
            string name = (type != null ? type.ToString() : "(null)");
            Console.WriteLine("Type : " + name);
        }
    }
}
```

Expected:
`Type : TestApp.Program`

Actual:
`Type : (null)`

parjong:
This commit revises Core CLR to use SP as Stack Marker (as in x64) for x86/Linux in order to fix #9378.

### DIFF
--- a/src/inc/regdisp.h
+++ b/src/inc/regdisp.h
@@ -135,12 +135,21 @@ inline PCODE GetControlPC(REGDISPLAY *display) {
 inline BOOL IsInCalleesFrames(REGDISPLAY *display, LPVOID stackPointer) {
     LIMITED_METHOD_CONTRACT;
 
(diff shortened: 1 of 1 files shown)


parjong:
\CC @seanshpark @wateret

parjong:

PR #9390, opened 2017-02-07
Remove more CAS

danmosemsft:
Contributes to https://github.com/dotnet/coreclr/issues/9321

I did not test this yet. The remaining ones are also just follow and delete, right? Nothing special.

@jkotas


src/mscorlib/src/System/Reflection/Emit/AssemblyBuilder.cs
```diff
@@ -396,9 +385,6 @@ private void InitManifestModule()
                                                               AssemblyName name,
                                                               Evidence identity,
                                                               ref StackCrawlMark stackMark,
```
jkotas: There needs to be matching change in C++.

justinvp:
Was adding model.xml back intentional?

danmosemsft:
Nope, thank you @justinvp.

CI was green, and model.xml is no longer in the build, so I'm merging.

PR #9392, opened 2017-02-07
Add gtGetPossibleOp2()

BruceForstall:
It was noticed that many uses of gtGetOp2() fully expect op2 to
exist and be non-nullptr, and the use of gtGetOp2() was simply
for style/readability purposes. However, it has a cost, as it
checks whether the tree is binary first, and returns nullptr if
not. For most cases, this is unecessary and expensive.

Introduce a new gtGetPossibleOp2() function that captures the
previous behavior: checking if the tree node is binary before
returning op2, or returning nullptr otherwise. gtGetOp2() is
changed to simply return gtOp2 directly, without any checking
in non-DEBUG builds. It can be used if you know the op2 exists.

Most uses of gtGetOp2() were left alone (and hence get the new
behavior). The ones that need the old behavior were renamed
gtGetPossibleOp2().

Mostly newer code is affected, as older code would generally
access gtOp2 directly, whereas newer code started using gtGetOp2()
for style and symmetry, especially after gtGetOp1() was introduced.

### DIFF
--- a/src/jit/gentree.h
+++ b/src/jit/gentree.h
@@ -1603,7 +1603,13 @@ public:
 
     inline GenTreePtr gtGetOp1();
 
+    // Directly return op2. Asserts the node is binary. Might return nullptr if the binary node allows
+    // a nullptr op2, such as GT_LIST. This is more efficient than gtGetOp2IfPresent() if you know what
+    // node type you have.
     inline GenTreePtr gtGetOp2();
+
+    // The returned pointer might be nullptr if the node is not binary, or if non-null op2 is not required.
+    inline GenTreePtr gtGetOp2IfPresent();
 
     // Given a tree node, if this is a child of that node, return the pointer to the child node so that it
     // can be modified; otherwise, return null.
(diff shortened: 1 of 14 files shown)


BruceForstall:

src/jit/gentree.h
```diff
@@ -1307,7 +1307,7 @@ struct GenTree
         return OperIsShiftOrRotate(OperGet());
     }
 
-    bool OperIsArithmetic() const
+    bool OperIsArithmetic() const // TODO-Throughput: This seems expensive, and is frequently called.
```
mikedn: Currently this is not a problem, the checked values form a continuous range so the generated code looks something like this (in VC++):
```asm
movzx       eax,byte ptr [rdi+8]  
lea         ecx,[rax-2Eh]  
cmp         cl,0Eh  
ja          Compiler::fgMorphForRegisterFP+0E7h (07FFAF52E3137h)  
```
In general, from what I've seen at least VC++ is pretty good at handling such checks.

BruceForstall: Impressive!

BruceForstall:
@dotnet/jit-contrib PTAL

pgavlin:
LGTM.

My only nit is that I'm not a huge fan of `gtGetPossibleOp2` as a name; I'd prefer something like `gtGetOp2IfPresent`. That said, this is certainly not blocking feedback.

BruceForstall:
I like your name better. Changed.

BruceForstall:

PR #9398, opened 2017-02-07
Change switch value widening

### LINKED ISSUE #8760: ECMA/JIT/Roslyn switch value inconsistency
IL `switch` description from the ECMA spec doesn't clearly state what type the switch value is expected to have:
```
The switch instruction pops value off the stack and compares it, as an unsigned integer, to n. 
```
Note that it says "unsigned integer", not "unsigned int32". Contrast this with the `ldelem` description that states:
```
The ldelem instruction loads the value of the element with index index (of type native int
or int32) in the zero-based one-dimensional array array
```

Now, RyuJIT64 happily accepts any integer type (importer code has `assertImp(genActualTypeIsIntOrI(op1->TypeGet()));`) and if the type is not already `TYP_I_IMPL` (aka `native int`) then it sign extends the switch value to `TYP_I_IMPL`. The fact that it sign extends seems odd considering that after that we have an unsigned compare.

On the other hand Roslyn seems to think that IL `switch` accepts only `unsigned int32` and if you try to switch on a `long` it generates IL code like:
```
    IL_0007:  ldloc.0
    IL_0008:  dup
    IL_0009:  ldc.i4.2
    IL_000a:  conv.i8
    IL_000b:  ble.un.s   IL_0010

    IL_000d:  pop
    IL_000e:  br.s       IL_002d

    IL_0010:  conv.u4
    IL_0011:  switch     (
                          IL_0024,
                          IL_0027,
                          IL_002a)
```
It does a separate check to see if the `long` value is within range and the converts it to `unsigned int32`. This wouldn't be necessary if IL `switch` does in fact accept any integral types.

In the end we get code like:
```asm
G_M57965_IG02:
       488B4228             mov      rax, qword ptr [rdx+40]
       4883F802             cmp      rax, 2
       7732                 ja       SHORT G_M57965_IG10
G_M57965_IG03:
       4863D0               movsxd   rdx, eax
       4883FA02             cmp      rdx, 2
       7729                 ja       SHORT G_M57965_IG10
       488D052E000000       lea      rax, [reloc @RWD00]
       8B0490               mov      eax, dword ptr [rax+4*rdx]
       488D0DDCFFFFFF       lea      rcx, G_M57965_IG02
       4803C1               add      rax, rcx
       FFE0                 jmp      rax
```
I'm not sure what needs to happen here. Maybe the spec needs to be clarified, maybe the JIT simply decided to be more relaxed and IL `switch` indeed accepts only `unsigned int32`.

I'd be happy to hear an explanation for the sign extension that happens in the JIT importer. The code generated for a switch is already pretty large as is and doing sign extend instead of zero extend just wastes a byte. IMO the importer should not do anything here, it's codegen's job to zero extend the switch value before using it for indexing the switch table.

mikedn:
For 64 bit target the importer sign extends the switch value to long but
- the switch value is supposed to be unsigned (it is compared to the switch arm count which is unsigned int32 in the CIL spec)
- this widening is needed to index the jump table, it is not needed when handling the default case

Changing the cast to unsigned and moving it just before jump table indexing usually results in smaller code:
- movsx becomes mov which is shorter
- the default case comparison no longer requires a REX prefix (if R8-R15 aren't used)

Fixes #8760

### DIFF
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -11566,22 +11566,6 @@ void Compiler::impImportBlockCode(BasicBlock* block)
                 op1 = impPopStack().val;
                 assertImp(genActualTypeIsIntOrI(op1->TypeGet()));
 
-#ifdef _TARGET_64BIT_
-                // Widen 'op1' on 64-bit targets
-                if (op1->TypeGet() != TYP_I_IMPL)
-                {
-                    if (op1->OperGet() == GT_CNS_INT)
-                    {
-                        op1->gtType = TYP_I_IMPL;
-                    }
-                    else
-                    {
-                        op1 = gtNewCastNode(TYP_I_IMPL, op1, TYP_I_IMPL);
-                    }
-                }
-#endif // _TARGET_64BIT_
-                assert(genActualType(op1->TypeGet()) == TYP_I_IMPL);
-
                 /* We can create a switch node */
 
                 op1 = gtNewOperNode(GT_SWITCH, TYP_VOID, op1);
@@ -15966,11 +15950,11 @@ SPILLSTACK:
                 }
                 else
                 {
-                    assert(addTree->gtOper == GT_SWITCH && genActualType(addTree->gtOp.gtOp1->gtType) == TYP_I_IMPL);
+                    assert(addTree->gtOper == GT_SWITCH && genActualTypeIsIntOrI(addTree->gtOp.gtOp1->TypeGet()));
 
                     unsigned temp = lvaGrabTemp(true DEBUGARG("spill addStmt SWITCH"));
                     impAssignTempGen(temp, addTree->gtOp.gtOp1, level);
-                    addTree->gtOp.gtOp1 = gtNewLclvNode(temp, TYP_I_IMPL);
+                    addTree->gtOp.gtOp1 = gtNewLclvNode(temp, genActualType(addTree->gtOp.gtOp1->TypeGet()));
                 }
             }
 
--- a/src/jit/lower.cpp
+++ b/src/jit/lower.cpp
@@ -465,7 +465,7 @@ GenTree* Lowering::LowerSwitch(GenTree* node)
     // both GT_SWITCH lowering code paths.
     // This condition is of the form: if (temp > jumpTableLength - 2){ goto jumpTable[jumpTableLength - 1]; }
     GenTreePtr gtDefaultCaseCond = comp->gtNewOperNode(GT_GT, TYP_INT, comp->gtNewLclvNode(tempLclNum, tempLclType),
-                                                       comp->gtNewIconNode(jumpCnt - 2, TYP_INT));
+                                                       comp->gtNewIconNode(jumpCnt - 2, tempLclType));
 
     // Make sure we perform an unsigned comparison, just in case the switch index in 'temp'
     // is now less than zero 0 (that would also hit the default case).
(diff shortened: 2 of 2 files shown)


mikedn:
jit-diff summary:
```
Total bytes of diff: -2831 (-0.02 % of base)
    diff is an improvement.
           8 : System.Runtime.Numerics.dasm (0.01 % of base)
           1 : System.Diagnostics.Tracing.dasm (0.02 % of base)
        -715 : Microsoft.CodeAnalysis.CSharp.dasm (-0.05 % of base)
        -568 : Microsoft.CodeAnalysis.VisualBasic.dasm (-0.04 % of base)
        -290 : System.Private.CoreLib.dasm (-0.01 % of base)
        -242 : System.Linq.Expressions.dasm (-0.05 % of base)
        -197 : Microsoft.CodeAnalysis.dasm (-0.03 % of base)
47 total files with size differences (45 improved, 2 regressed).
           6 : Microsoft.VisualBasic.dasm - Microsoft.VisualBasic.CompilerServices.OverloadResolution:MoreSpecificProcedure(ref,ref,ref,ref,int,byref,bool):ref
           5 : Microsoft.CodeAnalysis.CSharp.dasm - Microsoft.CodeAnalysis.CSharp.Binder:FoldNeverOverflowBinaryOperators(int,ref,ref):ref
           4 : Microsoft.CodeAnalysis.CSharp.dasm - Microsoft.CodeAnalysis.CSharp.SymbolKindExtensions:Localize(int):struct
           4 : Microsoft.CodeAnalysis.dasm - Microsoft.CodeAnalysis.CodeGen.ILBuilder:EmitLoadArgumentOpcode(int):this
           4 : Microsoft.CodeAnalysis.VisualBasic.dasm - Microsoft.CodeAnalysis.VisualBasic.Syntax.InternalSyntax.Parser:ParseDeclarationStatementInternal():ref:this
         -20 : Microsoft.CodeAnalysis.CSharp.dasm - Microsoft.CodeAnalysis.CSharp.Binder:DoUncheckedConversion(byte,ref):ref
         -16 : Microsoft.CodeAnalysis.CSharp.dasm - Microsoft.CodeAnalysis.CSharp.ErrorFacts:GetWarningLevel(int):int
         -16 : System.Linq.Expressions.dasm - System.Dynamic.Utils.TypeUtils:IsImplicitNumericConversion(ref,ref):bool
         -14 : Microsoft.CodeAnalysis.dasm - Microsoft.CodeAnalysis.CodeGen.ILBuilder:EmitNumericConversion(int,int,bool):this
         -12 : Microsoft.CodeAnalysis.dasm - Microsoft.CodeAnalysis.CommonCompiler:GenerateSqmData(ref,struct):this
1308 total methods with size differences (1207 improved, 101 regressed).
```

mikedn:
Before and after example:
```asm
; before
       lea      edx, [rax-43]
       movsxd   rdx, edx
       cmp      rdx, 25
       ja       SHORT G_M50478_IG03
       lea      rax, [reloc @RWD00]
       lea      rcx, G_M50478_IG02
       add      rax, rcx
       jmp      rax
; after
       lea      edx, [rax-43]
       cmp      edx, 25                ; 1 byte shorter than cmp rdx, 25
       ja       SHORT G_M50478_IG03
       mov      eax, edx               ; 1 byte shorter than movsxd
       lea      rdx, [reloc @RWD00]
       mov      edx, dword ptr [rdx+4*rax]
       lea      rcx, G_M50478_IG02
       add      rdx, rcx
       jmp      rdx
```
The few regressions that show up in jit-diff are caused by different register choices and instruction selection (LEA vs. SUB). For example:
```asm
; before
       sub      ecx, ebp
       movsxd   r13, ecx
       cmp      r13, 2
       ja       G_M5885_IG06
       lea      rax, [reloc @RWD00]
       mov      eax, dword ptr [rax+4*r13]
       lea      rdx, G_M5885_IG02
       add      rax, rdx
       jmp      rax
; after
       mov      r13d, ecx       ; this wasn't here before
       sub      r13d, ebp       ; now this uses r13 so it needs a REX prefix
       cmp      r13d, 2
       ja       G_M5885_IG06
       mov      eax, r13d
       lea      rdx, [reloc @RWD00]
       mov      edx, dword ptr [rdx+4*rax]
       lea      rcx, G_M5885_IG02
       add      rdx, rcx
       jmp      rdx
```

On x64 the ideal solution would be to take advantage of the fact that 32 bit instructions zero extend their result and thus uint->pointer casts can be completely eliminated. But ensuring that nothing weird happens in codegen that could result in the upper 32 bit containing garbage seems problematic.

danmosemsft:
@mikedn where are we at, are you waiting for a review from @dotnet/jit-contrib ?

mikedn:
@danmosemsft Yep, ready for review.

russellhadley:
@sivarv Can you look at this?

sivarv:
:shipit:

PR #9403, opened 2017-02-08
Add ComVisible(true) back to several special types

### LINKED ISSUE #9401: Exception related to COM interop thrown/caught first time Tasks are used
I'm seeing an InvalidOperationException get thrown here:
https://github.com/dotnet/coreclr/blob/b638af3a4dd52fa7b1ea1958164136c72096c25c/src/mscorlib/src/System/Threading/Tasks/AsyncCausalityTracer.cs#L154
and caught here:
https://github.com/dotnet/coreclr/blob/b638af3a4dd52fa7b1ea1958164136c72096c25c/src/mscorlib/src/System/Threading/Tasks/AsyncCausalityTracer.cs#L157-L162
the first time tasks are used.
```
System.InvalidOperationException: This type has a ComVisible(false) parent in its hierarchy, therefore QueryInterface calls for IDispatch or class interfaces are disallowed.
   at System.StubHelpers.InterfaceMarshaler.ConvertToNative(Object objSrc, IntPtr itfMT, IntPtr classMT, Int32 flags)
   at Windows.Foundation.Diagnostics.IAsyncCausalityTracerStatics.add_TracingStatusChanged(EventHandler`1 eventHandler)
   at System.Threading.Tasks.AsyncCausalityTracer..cctor()
```

Is there something we can/should do to avoid this?  I noticed it because it triggers ~13K of allocations as part of calling ToString on the exception:
https://github.com/dotnet/coreclr/blob/b638af3a4dd52fa7b1ea1958164136c72096c25c/src/mscorlib/src/System/Threading/Tasks/AsyncCausalityTracer.cs#L283
which in turn causes a bunch of stuff to be initialized.

cc: @jkotas

jkotas:
It is required by WinRT/COM interop. Fixes #9401.

### DIFF
--- a/src/mscorlib/src/System/Delegate.cs
+++ b/src/mscorlib/src/System/Delegate.cs
@@ -16,6 +16,8 @@ namespace System {
     using System.Diagnostics.Contracts;
 
     [Serializable]
+    [ClassInterface(ClassInterfaceType.AutoDual)]
+    [System.Runtime.InteropServices.ComVisible(true)]
     public abstract class Delegate : ICloneable, ISerializable 
     {
         // _target is the object we will invoke on
--- a/src/mscorlib/src/System/MulticastDelegate.cs
+++ b/src/mscorlib/src/System/MulticastDelegate.cs
@@ -14,6 +14,7 @@ namespace System
     using System.Reflection.Emit;
 
     [Serializable]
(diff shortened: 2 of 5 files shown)


jkotas:
Selective rollback of recent cleanup.

cc @stephentoub

jkotas:
Thanks - I have verified that there is no exception anymore.

PR #9404, opened 2017-02-08
Span Slice method AggressiveInlining

ahsonkhan:


ahsonkhan:
cc @jkotas good to go?

jkotas:
There are more methods in span.cs that should be marked:

- `public ref T GetItem(int index)` , also delete the link to the issue that is fixed now
- Constructor that takes `void*`
- ...

ahsonkhan:
> also delete the link to the issue that is fixed now
Has the issue been resolved? public T this[int index] does not return ref T still.

jkotas:
You are right - I thought that it is link to the issue with the GC crash.

PR #9405, opened 2017-02-08
JIT: enable implicit tail calls from inlined code

### LINKED ISSUE #9349: Inlining and implicit tail calls
Recognition of implicit tail calls (calls in tail position w/o the `tail` prefix) is disabled in when importing inlinee IL. So for example in the program below the generated code for `A` contains a normal call to `C`.
```C#
    // Needs to be big enough that we won't inline into caller
    // w/o using the noinline attribute
    static bool C(int x)
    {
        int result = 0;
        for (int i = 0; i < 25; i++)
        {
            result++;
        }
        for (int i = 0; i < 25; i++)
        {
            result++;
        }
        return x == result;
    }

    // Want this to inline
    static bool B(int x)
    {
        return C(x);
    }

    // Needs to be big enough that we won't inline into caller
    // w/o using the noinline attribute
    static bool A()
    {
        int result = 0;
        for (int i = 0; i < 25; i++)
        {
            result++;
        }
        for (int i = 0; i < 25; i++)
        {
            result++;
        }
        return B(result);
    }
```
Such cases arise fairly frequently in real code, for instance in overload delegation chains.

It should be ok to flag calls as implicit tail calls within the inlinee, if the inline call site is itself an implicit tail call. Some care may be required to make sure that the jit correctly models recursion, since recursive tail calls are convert into within-method jumps. If `A` contains implicit tail calls to itself, `B`, and `C`, we should make sure those are properly converted to jumps or left as calls.

Note the jit won't inline at explicit tail call sites, instead favoring the explicit tail call.

AndyAyersMS:
Inlines of calls from implicit tail call sites should allow recognition
of inlinee implicit tail call sites.

The jit recognizes implicit tail call sites during importation,
but the inlinee compiler instance did not have compTailCallOpt set
and so never recognized these instances. Fix this and update the logic
to detect the transitively implicit tail calls.

Now that these sites are recognized, morph needs a fix to tunnel through
repeated casts for tail calls, since each level of inlining might add a
cast. All these casts should be identical.

Note under R2R tail calls are not yet recognized (see ZapInfo::canTailCall).

Closes #9349.

### DIFF
--- a/src/jit/compiler.cpp
+++ b/src/jit/compiler.cpp
@@ -3028,25 +3028,37 @@ void Compiler::compInitOptions(JitFlags* jitFlags)
     setUsesSIMDTypes(false);
 #endif // FEATURE_SIMD
 
-    if (compIsForInlining() || compIsForImportOnly())
+    if (compIsForImportOnly())
     {
         return;
     }
+
+#if FEATURE_TAILCALL_OPT
+    // By default opportunistic tail call optimization is enabled.
+    // Recognition is done in the importer so this must be set for
+    // inlinees as well.
+    opts.compTailCallOpt = true;
+#endif // FEATURE_TAILCALL_OPT
+
+    if (compIsForInlining())
+    {
+        return;
+    }
+
     // The rest of the opts fields that we initialize here
     // should only be used when we generate code for the method
     // They should not be used when importing or inlining
+    CLANG_FORMAT_COMMENT_ANCHOR;
+
+#if FEATURE_TAILCALL_OPT
+    opts.compTailCallLoopOpt = true;
+#endif // FEATURE_TAILCALL_OPT
 
     opts.genFPorder = true;
     opts.genFPopt   = true;
 
     opts.instrCount = 0;
     opts.lvRefCount = 0;
-
-#if FEATURE_TAILCALL_OPT
-    // By default opportunistic tail call optimization is enabled
-    opts.compTailCallOpt     = true;
-    opts.compTailCallLoopOpt = true;
-#endif
 
 #ifdef PROFILING_SUPPORTED
     opts.compJitELTHookEnabled = false;
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -12740,11 +12740,27 @@ void Compiler::impImportBlockCode(BasicBlock* block)
                             prefixFlags |= PREFIX_TAILCALL_EXPLICIT;
                         }
                     }
+                }
 
-                    // Note that when running under tail call stress, a call will be marked as explicit tail prefixed
-                    // hence will not be considered for implicit tail calling.
-                    bool isRecursive = (callInfo.hMethod == info.compMethodHnd);
-                    if (impIsImplicitTailCallCandidate(opcode, codeAddr + sz, codeEndp, prefixFlags, isRecursive))
+                // This is split up to avoid goto flow warnings.
+                bool isRecursive;
+                isRecursive = !compIsForInlining() && (callInfo.hMethod == info.compMethodHnd);
+
+                // Note that when running under tail call stress, a call will be marked as explicit tail prefixed
+                // hence will not be considered for implicit tail calling.
+                if (impIsImplicitTailCallCandidate(opcode, codeAddr + sz, codeEndp, prefixFlags, isRecursive))
+                {
+                    if (compIsForInlining())
+                    {
+                        // Are we inlining at an implicit tail call site? If so we can flag
+                        // implicit tail call sites in the inline body.
+                        if (impInlineInfo->iciCall->IsImplicitTailCall())
+                        {
+                            JITDUMP(" (Inline Implicit Tail call: prefixFlags |= PREFIX_TAILCALL_IMPLICIT)");
+                            prefixFlags |= PREFIX_TAILCALL_IMPLICIT;
+                        }
+                    }
+                    else
                     {
                         JITDUMP(" (Implicit Tail call: prefixFlags |= PREFIX_TAILCALL_IMPLICIT)");
                         prefixFlags |= PREFIX_TAILCALL_IMPLICIT;
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -12752,13 +12752,17 @@ void Compiler::impImportBlockCode(BasicBlock* block)
                 {
                     if (compIsForInlining())
                     {
-                        // Are we inlining at an implicit tail call site? If so we can flag
-                        // implicit tail call sites in the inline body.
+#if FEATURE_TAILCALL_OPT_SHARED_RETURN
+                        // Are we inlining at an implicit tail call site? If so the we can flag
+                        // implicit tail call sites in the inline body. These call sites
+                        // often end up in non BBJ_RETURN blocks, so only flag them when
+                        // we're able to handle shared returns.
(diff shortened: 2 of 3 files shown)


AndyAyersMS:
@JosephTremoulet PTAL
cc @dotnet/jit-contrib 

Jit-diff data (only diffs in corelib because R2R currently inhibits tail calls):
```
Total bytes of diff: 798 (0.01 % of base)
    diff is a regression.
         798 : System.Private.CoreLib.dasm (0.02 % of base)
1 total files with size differences.
          47 : System.Private.CoreLib.dasm - System.Diagnostics.Tracing.ActivityTracker:OnStart(ref,ref,int,byref,byref,int):this
          40 : System.Private.CoreLib.dasm - System.Diagnostics.Tracing.ActivityTracker:OnStop(ref,ref,int,byref):this
          27 : System.Private.CoreLib.dasm - System.StubHelpers.NullableMarshaler:ConvertToNative(byref):long (9 methods)
          20 : System.Private.CoreLib.dasm - System.Security.Permissions.FileIOPermission:AddPathList(int,int,ref,bool,bool,bool):this
          12 : System.Private.CoreLib.dasm - System.Reflection.Emit.TypeBuilder:SetInterfaces(ref):this
         -34 : System.Private.CoreLib.dasm - System.Reflection.Emit.SignatureHelper:AddOneArgTypeHelperWorker(ref,bool):this
         -10 : System.Private.CoreLib.dasm - Decoder:.ctor(ref):this (2 methods)
          -5 : System.Private.CoreLib.dasm - UTF32Decoder:.ctor(ref):this
          -5 : System.Private.CoreLib.dasm - System.Nullable`1[Guid][System.Guid]:ToString():ref:this
          -5 : System.Private.CoreLib.dasm - System.Nullable`1[TimeSpan][System.TimeSpan]:ToString():ref:this
384 total methods with size differences.
```

benaadams:
@dotnet-bot test Windows_NT arm64 Cross Checked Build and Test

benaadams:
@dotnet-bot retest Windows_NT arm64 Cross Checked Build and Test

benaadams:
@dotnet-bot test Windows_NT arm64 Cross Checked Build and Test

AndyAyersMS:
Tail call opts are not enabled in debug builds, so even w/o looking in the logs, the ARM64 failures are almost certainly unrelated. Looking at the logs shows some kind of scripting issue.
```
18:56:27 d:\j\workspace\arm64_cross_c---f41ee988>python tests\scripts\arm64_post_build.py -repo_root d:\j\workspace\arm64_cross_c---f41ee988 -arch arm64 -build_type checked -scenario default -key_location C:\tools\key.txt 
18:56:27 Traceback (most recent call last):
18:56:27   File "tests\scripts\arm64_post_build.py", line 311, in <module>
18:56:27     main(args)
18:56:27   File "tests\scripts\arm64_post_build.py", line 254, in main
18:56:27     repo_root, arch, build_type, scenario, key_location, force_update = validate_args(args)
18:56:27   File "tests\scripts\arm64_post_build.py", line 228, in validate_args
18:56:27     validate_arg(key_location, lambda item: os.path.isfile(item))
18:56:27   File "tests\scripts\arm64_post_build.py", line 218, in validate_arg
18:56:27     raise Exception("Argument: %s is not valid." % (arg))
18:56:27 Exception: Argument: C:\tools\key.txt is not valid.
```

src/jit/compiler.cpp
```diff
@@ -3028,26 +3028,38 @@ void Compiler::compInitOptions(JitFlags* jitFlags)
...
+        return;
+    }
+
     // The rest of the opts fields that we initialize here
     // should only be used when we generate code for the method
     // They should not be used when importing or inlining
+    CLANG_FORMAT_COMMENT_ANCHOR;
+
+#if FEATURE_TAILCALL_OPT
+    opts.compTailCallLoopOpt = true;
```
JosephTremoulet: Would it be possible/interesting (at some point in the future) to recognize tail calls back to an inlinee and make them loops?

sivarv:
Looks good.
Trying ngen diffs on desktop will provide another measure of CQ impact.
On desktop, tail call differences could lead to app compat issues (we had an instance of GetCallingAssembly() failure during 4.6 sign-off time).  We can be conservative and enable this only for CoreCLR or should remember to disable this if there are any app compat failures on desktop due to this.

src/jit/compiler.cpp
AndyAyersMS: Yes, it would be possible. We already track the ancestral inline context for each call site, so we can tell when a call site recursively invokes an inlinee. What we'd need in addition is the following:

* track the logical entry point for the inlinee in the inline context, so we know where to target the loop branch.
* disable direct sub for inlinees in tail position, since the arguments are now potentially starg.

With that I think we could generalize the loop tail call opt to handle inlined cases, and I think a tail call loop opt is potentially much more impactful then a tail call alone.

AndyAyersMS:
@JosephTremoulet I vaguely recall you had a way of getting fragile ngen codegen in jit-diffs. If not, we should potentially add that as an option. Or get SPMI online and ensure it has a proper mix of jit/ngen/r2r instances.

@sivarv good point about desktop. Do you recall what the fix was for the app compat issue?

JosephTremoulet:
> @JosephTremoulet I vaguely recall you had a way of getting fragile ngen codegen in jit-diffs. If not, we should potentially add that as an option

No, I've just used jit-produced asm when I wanted to be extra sure I was looking at diffs matching benchmark execution.  You may be remembering that I opened dotnet/jitutils#75, but I haven't implemented it.  It should be trivial to implement if crossgen has a --fragile flag, but if it has one it seems to be hidden...

> Or get SPMI online and ensure it has a proper mix of jit/ngen/r2r instances.

That would be great!

src/jit/compiler.cpp
AndyAyersMS: One more thing: we currently won't tail call if the root method is noinline, but that block doesn't make sense for these inlinee loop tail call opts, so we should track calls in tail position in addition to calls eligible to be made into tail calls.

AndyAyersMS:
crossgen has a `/FragileNonVersionable` option for this, though it doesn't show up in the help.

I would patch jit-dasm/jit-diff but I'm stuck on an older version....

sivarv:
@AndyAyersMS - desktop appcompat failure was due to a particular class of implicit tail calls (the ones with shared return). See FEATURE_TAILCALL_OPT_SHARED_RETURN in jit.h, it is defined as 1 for CoreCLR and 0 on desktop.

AndyAyersMS:
Tried running desktop SPMI. Unfortunately it fails on exactly the methods we'd like to examine, since there are new calls to `canTailCall` that aren't resolvable by the cached data. So will try ngen or PMI diffs instead.

AndyAyersMS:
Desktop also running into asserts because for desktop `FEATURE_TAILCALL_OPT_SHARED_RETURN` is disabled, and it's not guaranteed that inline implicit tail call sites are in a `BBJ_RETURN` block (they may be in a predecessor block that falls through or jumps to the return). It seems prudent for now to only enable recognition of inline implicit tail calls under this feature.

Which means, they won't be recognized on desktop....

AndyAyersMS:
Updated (& rebased) to only kick in if we support shared returns. Verified no diff on desktop.

sivarv:
Looks good.

PR #9408, opened 2017-02-08
Disable checked arm64 jobs until jobs are stable

jashook:
Jobs currently are expected to run red until a lstFile update goes in. Therefore,
disable the job as a PR trigger until it is clean.

@RussKeldorph @adiaaida ptal


jashook:
@dotnet-bot test CI Please

PR #9174, opened 2017-01-27
CMake configuration for cross building for Android

qmfrederik:
This PR adds CMake files which support cross building for Android.
You can cross-build assuming `ROOTFS_DIR` points to the `sysroot` folder for the toolchain which was generated by Android. That is, we assume the toolchain can be found in `${ROOTFS_DIR}/../bin`.

I've placed them in cross/android/arm64.

You can then use a command like:
```
CONFIG_DIR=`realpath cross/android/arm64` ROOTFS_DIR=`realpath cross/android/toolchain/arm64/sysroot` ./build.sh cross arm64 skipgenerateversion skipmscorlib cmakeargs -DENABLE_LLDBPLUGIN=0
```

to cross-compile for Android.

I'll create a PR for scripts that can generate the toolchain shortly.

PR #9409, opened 2017-02-08
Synchronous toolchain.cmake files with CoreFX

jyoungyun:
Remove unnecessary `TOOLCHAIN_PREFIX` define and synchronous toolchain.cmake files with CoreFX

Related issue: https://github.com/dotnet/corefx/pull/15882

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,7 +103,7 @@ else (WIN32)
       # Ensure that objcopy is present
       if (DEFINED ENV{CROSSCOMPILE} AND NOT DEFINED CLR_CROSS_COMPONENTS_BUILD)
         if (CMAKE_SYSTEM_PROCESSOR STREQUAL armv7l OR CMAKE_SYSTEM_PROCESSOR STREQUAL aarch64)
-          find_program(OBJCOPY ${TOOLCHAIN_PREFIX}objcopy)
+          find_program(OBJCOPY ${TOOLCHAIN}-objcopy)
         elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL i686)
           find_program(OBJCOPY objcopy)
         else()
--- a/cross/arm/toolchain.cmake
+++ b/cross/arm/toolchain.cmake
@@ -4,15 +4,7 @@ set(CMAKE_SYSTEM_NAME Linux)
 set(CMAKE_SYSTEM_VERSION 1)
 set(CMAKE_SYSTEM_PROCESSOR armv7l)
 
-## Specify the toolchain
 set(TOOLCHAIN "arm-linux-gnueabihf")
-set(TOOLCHAIN_PREFIX ${TOOLCHAIN}-)
-#set(CMAKE_C_COMPILER ${TOOLCHAIN_PREFIX}gcc)
-#set(CMAKE_CXX_COMPILER ${TOOLCHAIN_PREFIX}g++)
-#set(CMAKE_ASM_COMPILER ${TOOLCHAIN_PREFIX}as)
-#set(CMAKE_AR ${TOOLCHAIN_PREFIX}ar)
-#set(CMAKE_OBJCOPY ${TOOLCHAIN_PREFIX}objcopy)
-#set(CMAKE_OBJDUMP ${TOOLCHAIN_PREFIX}objdump)
 
 add_compile_options(-target armv7-linux-gnueabihf)
 add_compile_options(-mthumb)
@@ -20,7 +12,7 @@ add_compile_options(-mfpu=vfpv3)
 add_compile_options(--sysroot=${CROSS_ROOTFS})
 
 set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} -target ${TOOLCHAIN}")
-set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} -B ${CROSS_ROOTFS}/usr/lib/gcc/${TOOLCHAIN}")
+set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} -B${CROSS_ROOTFS}/usr/lib/gcc/${TOOLCHAIN}")
 set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} -L${CROSS_ROOTFS}/lib/${TOOLCHAIN}")
 set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} --sysroot=${CROSS_ROOTFS}")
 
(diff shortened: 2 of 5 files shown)


cross/arm/toolchain.cmake
```diff
@@ -4,23 +4,15 @@ set(CMAKE_SYSTEM_NAME Linux)
 set(CMAKE_SYSTEM_VERSION 1)
 set(CMAKE_SYSTEM_PROCESSOR armv7l)
 
-## Specify the toolchain
 set(TOOLCHAIN "arm-linux-gnueabihf")
-set(TOOLCHAIN_PREFIX ${TOOLCHAIN}-)
-#set(CMAKE_C_COMPILER ${TOOLCHAIN_PREFIX}gcc)
```
hqueue: Cool! I wanted it be removed for a long time. Thanks :)

CMakeLists.txt
```diff
@@ -103,7 +103,7 @@ else (WIN32)
       # Ensure that objcopy is present
       if (DEFINED ENV{CROSSCOMPILE} AND NOT DEFINED CLR_CROSS_COMPONENTS_BUILD)
         if (CMAKE_SYSTEM_PROCESSOR STREQUAL armv7l OR CMAKE_SYSTEM_PROCESSOR STREQUAL aarch64)
-          find_program(OBJCOPY ${TOOLCHAIN_PREFIX}objcopy)
+          find_program(OBJCOPY ${TOOLCHAIN}-objcopy)
```
qmfrederik: To cross-compile for Android, we need to use objcopy from the Android toolchain, not the host. On Android, that would be `${CROSS_NDK_TOOLCHAIN}/bin/aarch64-linux-android-objcopy`; previously we set `${TOOLCHAIN_PREFIX}` to `${CROSS_NDK_TOOLCHAIN}/bin/aarch64-linux-android-` so it worked.

If you set it to `${TOOLCHAIN}` this will evaluate to `aarch64-linux-android-objcopy` and from what I get, CMake will look for this file in the bin folders on the host, which doesn't work.

Another option would be to set `${TOOLCHAIN_PREFIX}` to `${CROSS_NDK_TOOLCHAIN}/bin/` on Android, leave it empty for Linux, and change this line to:

```
 find_program(OBJCOPY ${TOOLCHAIN_PREFIX}${TOOLCHAIN}-objcopy)
```

cross/arm/toolchain.cmake
```diff
@@ -4,23 +4,15 @@ set(CMAKE_SYSTEM_NAME Linux)
...
-#set(CMAKE_OBJDUMP ${TOOLCHAIN_PREFIX}objdump)
 
 add_compile_options(-target armv7-linux-gnueabihf)
 add_compile_options(-mthumb)
 add_compile_options(-mfpu=vfpv3)
 add_compile_options(--sysroot=${CROSS_ROOTFS})
 
 set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} -target ${TOOLCHAIN}")
-set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} -B ${CROSS_ROOTFS}/usr/lib/gcc/${TOOLCHAIN}")
+set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} -B${CROSS_ROOTFS}/usr/lib/gcc/${TOOLCHAIN}")
```
qmfrederik: We do the same on Android (to keep the files consistent), hence `${TOOLCHAIN}` must be an absolute path, so we can't prefix `${TOOLCHAIN}` like we could do with `${TOOLCHAIN_PREFIX}`

qmfrederik:
@jyoungyun I left some comments; the gist of it is that for Android, we need to set a full path `objcopy`. We did that in the past by setting `${TOOLCHAIN_PREFIX}` to a full path, but if you remove it, it doesn't work anymore.

Another option may be to agree that `${TOOLCHAIN_PREFIX}` should contain the path in which the toolchain is installed (where "prefix" means "path" like in `./configure --prefix=...`) and change the find command for `objcopy` to `find_program(OBJCOPY ${TOOLCHAIN_PREFIX}${TOOLCHAIN}-objcopy)`.

I think you could then leave `TOOLCHAIN_PREFIX` empty for Linux cross-compile and we would only need to set it for Android.

/cc @janvorli

CMakeLists.txt
jyoungyun: What about replace `${TOOLCHAIN_PREFIX}` to `${TOOLCHAIN}` and redefine other macro for the previous `${TOOLCHAIN}` variable? I can revise the find objcopy logic as you said. But I think this issue can be resolved by modifiying `cross/android/arm64/toolchain.cmake` file. What do you think?

jyoungyun:
@qmfrederik I think we can support android build without revising CMakeLists.txt. Here is my example.
```
set(ANDROID_TOOLCHAIN "aarch64-linux-android")
set(TOOLCHAIN ${CROSS_NDK_TOOLCHAIN}/bin/${ANDROID_TOOLCHAIN})
set(CMAKE_C_COMPILER ${TOOLCHAIN}-clang)
...
set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} -B ${CROSS_ROOTFS}/usr/lib/gcc/${ANDROID_TOOLCHAIN}") 
set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} -L${CROSS_ROOTFS}/lib/${ANDROID_TOOLCHAIN}") 
```
I prefer not to create unnecessary define. So, if possible, I hope it will be implemented in existing structure. Let me know if something is wrong with me. If we need `${TOOLCHAIN_PREFIX}`, I will patch it. :)

qmfrederik:
@jyoungyun Yup, we've done something similar in #9174 to support cross compiling for Android.

When we did that, we hit a problem with `objcopy`, though. Like I described above, the checks you have now would cause the script to either pick up the `objcopy` from the host OS or set it to OBJCOPY-NOTFOUND - see https://github.com/cydhaselton/coreclr/issues/2#issuecomment-273330052 for the discussion around that.

Looking at the documentation for [find_program](https://cmake.org/cmake/help/v3.0/command/find_program.html), it looks like there are some alternatives we can try:
- Include the Android toolchain in the `CMAKE_PREFIX_PATH` varible
- Provide an explicit hint for Android

I can check the CMAKE_PREFIX_PATH option later today my time zone, and I'll get back to you. Would that be OK?

qmfrederik:
@jyoungyun Yup, setting `CMAKE_PREFIX_PATH` to the Android toolchain folder works and has the nice side-effect of simplifying the scripts.

I've updated #9174 

LGTM!

jyoungyun:
@gkhanna79 @janvorli PTAL

PR #9410, opened 2017-02-08
[x86/Linux] Fix calling convertion mismatch inside ilasm/ildasm

parjong:
ilasm/ildasm dynamically load serveral functions, but their signature does not match with that of the actual implementation.

This results in segmentation fault while running.

This commit fixes signature mismatch.


parjong:
\CC @seanshpark @wateret

PR #9412, opened 2017-02-08
Clean up EnumGcRefs

parjong:
This commit cleans up EnumGcRefs.

This commit first declares EnumGcRefs only for !CROSSGEN_COMPILE, and then revises the implementation as follows:
  #ifndef CROSSGEN_COMPILE
  #ifndef USE_GC_INFO_DECODER (use USE_GC_INFO_DECODER instead of _TARGET_X86_)
  ...
  #else 
  ...
  #endif
  #endif

In addition, this commit removes several unnecessary #ifdef inside EnumGcRefs implemenation for USE_GC_INFO_DECODER.


parjong:

PR #9416, opened 2017-02-08
Remove OpenSuse13 from CI

wtgodbe:
We're no longer going to support OpenSuse13.2, so we should remove it from the CI.


wtgodbe:
@gkhanna79 @mmitche @ramarag PTAL

gkhanna79:
CC @weshaggard @RussKeldorph

PR #9417, opened 2017-02-08
Publish test native bins from correct directory

wtgodbe:
$(OutputPath) now points to bin/Product/{os}.{arch}/{config}, where it used to point to bin/obj/{os}.{arch}/{config}. Inside bin/Product is a folder named 'bin' which contains all test native binaries - we should publish from there, instead of from obj. This will fix build breaks in the CoreCLR pipeline.

### DIFF
--- a/src/publish.proj
+++ b/src/publish.proj
@@ -22,7 +22,7 @@
 
   <PropertyGroup>
     <PublishPattern Condition="'$(PublishPattern)' == '' and '$(PublishTestNativeBins)' != 'true'">$(PackagesBinDir)**\*.nupkg</PublishPattern>
-    <PublishPattern Condition="'$(PublishPattern)' == '' and '$(PublishTestNativeBins)' == 'true'">$(OutputPath)\tests\src\**</PublishPattern>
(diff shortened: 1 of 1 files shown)


wtgodbe:
CC @gkhanna79 @dagood

wtgodbe:
@dagood test native binaries live in two places - in `bin/obj/{OsPlatformConfig}/tests/src/......`, and in `bin/Product/{OsPlatformConfig}/bin/.....` . Previously, I was publishing from the first location. After https://github.com/dotnet/coreclr/pull/9351, the variable `$(OutputPath)` changed from pointing to `bin/obj/{OsPlatformConfig}`, to pointing to `bin/Product/{OsPlatformConfig}` - so I was looking for binaries in `bin/Product/{OsPlatformConfig}/tests/src/......`, which doesn't exist. It's probably better to pick up the binaries from `bin/Product` anyways, so this change picks up binaries from the `Product` location, rather than the `obj` location

dagood:
That makes sense, thanks @wtgodbe!

PR #9431, opened 2017-02-09
Generate better code for x86 SIMD get[i] intrinsic

### LINKED ISSUE #9374: RyuJIT/x86 SIMD: generate better code for LONG get[i] intrinsic for constant 'i'
Currently, we always load the index into a temp register.

BruceForstall:
For get[i] intrinsic, with LONG base type and constant index, we can
compute the decomposed indices statically.

Fixes #9374

### DIFF
--- a/src/jit/decomposelongs.cpp
+++ b/src/jit/decomposelongs.cpp
@@ -1642,49 +1642,83 @@ GenTree* DecomposeLongs::DecomposeSimdGetItem(LIR::Use& use)
     assert(varTypeIsSIMD(simdTree->gtOp.gtOp1->gtType));
     assert(simdTree->gtOp.gtOp2->gtType == TYP_INT);
 
+    bool    indexIsConst = simdTree->gtOp.gtOp2->IsCnsIntOrI();
+    ssize_t index        = 0;
+    if (indexIsConst)
+    {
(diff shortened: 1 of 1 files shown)


BruceForstall:
@sivarv PTAL
cc @dotnet/jit-contrib

sivarv:
:shipit:

src/jit/decomposelongs.cpp
```diff
@@ -1642,49 +1642,83 @@ GenTree* DecomposeLongs::DecomposeSimdGetItem(LIR::Use& use)
...
+    {
+        LIR::Use op2(Range(), &simdTree->gtOp.gtOp2, simdTree);
+        indexTmpVarNum = op2.ReplaceWithLclVar(m_compiler, m_blockWeight);
+        JITDUMP("[DecomposeSimdGetItem]: Saving op2 tree to a temp var:\n");
+        DISPTREERANGE(Range(), op2.Def());
+    }
 
     // Create:
-    //      loResult = GT_SIMD{get_item}[int](tmp_simd_var, tmp_index * 2)
+    //      loResult = GT_SIMD{get_item}[int](tmp_simd_var, index * 2)
```
CarolEidt: It would be nice to update this comment since we're not using the tmp_simd_var inthe case where `indexIsConst`.

PR #9433, opened 2017-02-09
Inlinable ArraySegment ctor

benaadams:
`ArraySegment``1:.ctor(ref,int,int):this` currently is too big to be inlinable and shows up on plaintext traces

This change shrinks it so it inlines

```
Successfully inlined ArraySegment`1:.ctor(ref,int,int):this (47 IL bytes) (depth 1) [profitable inline]
```

/cc @jkotas @stephentoub @AndyAyersMS

### DIFF
--- a/src/mscorlib/src/System/ArraySegment.cs
+++ b/src/mscorlib/src/System/ArraySegment.cs
@@ -45,7 +45,7 @@ namespace System
         public ArraySegment(T[] array, int offset, int count)
         {
             // Validate input arguments
-            if (array == null || offset < 0 || count < 0|| (array.Length - offset < count))
+            if (array == null || (offset | count) < 0 || (array.Length - offset < count))
                 ThrowConstructorValidationFailedExceptions(array, offset, count);
             Contract.EndContractBlock();
 
--- a/src/mscorlib/src/System/ArraySegment.cs
+++ b/src/mscorlib/src/System/ArraySegment.cs
@@ -44,8 +44,9 @@ namespace System
 
         public ArraySegment(T[] array, int offset, int count)
         {
-            // Validate input arguments
-            if (array == null || (offset | count) < 0 || (array.Length - offset < count))
+            // Validate arguments, check is minimal instructions with reduced branching for inlinable fast-path
+            // Failure should be rare and location determination and message is delegated to failure functions
+            if (array == null | ((offset | count) < 0) || (array.Length - offset < count))
                 ThrowConstructorValidationFailedExceptions(array, offset, count);
             Contract.EndContractBlock();
 
--- a/src/mscorlib/src/System/ArraySegment.cs
+++ b/src/mscorlib/src/System/ArraySegment.cs
@@ -47,7 +47,7 @@ namespace System
             // Validate arguments, check is minimal instructions with reduced branching for inlinable fast-path
             // Failure should be rare and location determination and message is delegated to failure functions
             if (array == null | ((offset | count) < 0) || (array.Length - offset < count))
-                ThrowConstructorValidationFailedExceptions(array, offset, count);
+                ThrowHelper.ThrowArraySegmentCtorValidationFailedExceptions(array, offset, count);
             Contract.EndContractBlock();
 
             _array = array;
@@ -152,24 +152,6 @@ namespace System
         public static bool operator !=(ArraySegment<T> a, ArraySegment<T> b)
         {
             return !(a == b);
-        }
-
-        private static void ThrowConstructorValidationFailedExceptions(T[] array, int offset, int count)
-        {
-            throw GetConstructorValidationFailedException(array, offset, count);
-        }
-
-        private static Exception GetConstructorValidationFailedException(T[] array, int offset, int count)
-        {
-            if (array == null)
-                return ThrowHelper.GetArgumentNullException(ExceptionArgument.array);
-            if (offset < 0)
-                return ThrowHelper.GetArgumentOutOfRangeException(ExceptionArgument.offset, ExceptionResource.ArgumentOutOfRange_NeedNonNegNum);
-            if (count < 0)
-                return ThrowHelper.GetArgumentOutOfRangeException(ExceptionArgument.count, ExceptionResource.ArgumentOutOfRange_NeedNonNegNum);
-
-            Debug.Assert(array.Length - offset < count);
-            return ThrowHelper.GetArgumentException(ExceptionResource.Argument_InvalidOffLen);
         }
 
         #region IList<T>
--- a/src/mscorlib/src/System/ArraySegment.cs
+++ b/src/mscorlib/src/System/ArraySegment.cs
@@ -46,7 +46,7 @@ namespace System
         {
             // Validate arguments, check is minimal instructions with reduced branching for inlinable fast-path
             // Failure should be rare and location determination and message is delegated to failure functions
-            if ((array == null | (offset | count) < 0) || (array.Length - offset < count))
+            if (array == null || (offset | count) < 0 || (array.Length - offset < count))
                 ThrowHelper.ThrowArraySegmentCtorValidationFailedExceptions(array, offset, count);
             Contract.EndContractBlock();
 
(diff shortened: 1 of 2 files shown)


src/mscorlib/src/System/ArraySegment.cs
```diff
@@ -160,6 +154,24 @@ public bool Equals(ArraySegment<T> obj)
             return !(a == b);
         }
 
+        private static void ThrowConstructorValidationFailedExceptions(T[] array, int offset, int count)
```
benaadams: Could take an `object` and move to ThrowHelper to reduce generic load? Throw functions could also probably be shared with `Span<T>`?

benaadams: Oh Span does weird stuff...

jkotas: > Could take an object and move to ThrowHelper 

Yes, it may be nice.

jkotas:
LGTM. Thanks!

src/mscorlib/src/System/ArraySegment.cs
benaadams: Changed; think the Asserts are valid?

src/mscorlib/src/System/ThrowHelper.cs
```diff
@@ -209,6 +212,23 @@ internal static class ThrowHelper {
             throw GetInvalidOperationException(ExceptionResource.InvalidOperation_EnumOpCantHappen);
         }
 
+        internal static void ThrowArraySegmentCtorValidationFailedExceptions(object array, int offset, int count) {
```
justinvp: Why not `Array` instead of `object`?

src/mscorlib/src/System/ThrowHelper.cs
```diff
@@ -209,6 +212,23 @@ internal static class ThrowHelper {
...
+            Debug.Assert(array is Array);
+            throw GetArraySegmentCtorValidationFailedException(array, offset, count);
+        }
+
+        private static Exception GetArraySegmentCtorValidationFailedException(object array, int offset, int count) {
```
justinvp: Why not `Array` instead of `object`?

src/mscorlib/src/System/ArraySegment.cs
```diff
@@ -46,14 +44,10 @@ public ArraySegment(T[] array)
...
-                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.array);
-            if (offset < 0)
-                ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.offset, ExceptionResource.ArgumentOutOfRange_NeedNonNegNum);
-            if (count < 0)
-                ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.count, ExceptionResource.ArgumentOutOfRange_NeedNonNegNum);
-            if (array.Length - offset < count)
-                ThrowHelper.ThrowArgumentException(ExceptionResource.Argument_InvalidOffLen);
+            // Validate arguments, check is minimal instructions with reduced branching for inlinable fast-path
+            // Failure should be rare and location determination and message is delegated to failure functions
+            if (array == null | ((offset | count) < 0) || (array.Length - offset < count))
```
AndyAyersMS: Shouldn's this be `array == null || ` ? You want to short-circuit and not continue on to possibly look at `array.Length`... right?

src/mscorlib/src/System/ThrowHelper.cs
benaadams: I over thought it with the Debug.Assert casts. It should be `Array`

benaadams:
> Shouldn't this be `array == null ||` ? You want to short-circuit and not continue on to possibly look at `array.Length`... right?

Short circuit is still there before the length test; but in the common case all the tests will be executed; so it bitwise `|`s the two (three) independent tests and eliminates a branch. However the asm is slightly longer.

With `array == null || `
```
488B4E18             mov      rcx, gword ptr [rsi+24]
4885FF               test     rdi, rdi
744D                 je       SHORT G_M24074_IG06
8BD3                 mov      edx, ebx
0BD5                 or       edx, ebp
85D2                 test     edx, edx
7C45                 jl       SHORT G_M24074_IG06
8B5708               mov      edx, dword ptr [rdi+8]
2BD3                 sub      edx, ebx
3BD5                 cmp      edx, ebp
7C3C                 jl       SHORT G_M24074_IG06
```

With `array == null | `
```
488B4E18             mov      rcx, gword ptr [rsi+24]
8BD3                 mov      edx, ebx
0BD5                 or       edx, ebp
85D2                 test     edx, edx
0F9CC2               setl     dl
0FB6D2               movzx    rdx, dl
4885FF               test     rdi, rdi
410F94C0             sete     r8b
450FB6C0             movzx    r8, r8b
410BD0               or       edx, r8d
7545                 jne      SHORT G_M24074_IG06
8B5708               mov      edx, dword ptr [rdi+8]
2BD3                 sub      edx, ebx
3BD5                 cmp      edx, ebp
7C3C                 jl       SHORT G_M24074_IG06
```
@AndyAyersMS Not sure which is better, the shorter asm or one branch less? (Hadn't thought it would make it that much longer)

mikedn:
> ArraySegment``1:.ctor(ref,int,int):this currently is too big to be inlinable and shows up on plaintext traces

Can you measure any kind of performance improvement after inlining? It's not like what shows up in traces is only the call code, it includes the actual code too.

> Not sure which is better, the shorter asm or one branch less?

I tried this kind of trick before and usually it doesn't work. The only way to tell for sure is to measure, preferably by using a real workload and not by using micro-benchmarking. You are turning a predictable branch into more instructions/code bytes and slightly longer dependency chains.

benaadams:
Reverted to logical or

@mikedn inlining aside before its

```asm
; Lcl frame size = 32
G_M41510_IG01:
       push     rdi
       push     rsi
       push     rbx
       sub      rsp, 32
       mov      rsi, rcx
       mov      edi, r9d
       mov      ebx, dword ptr [rsp+60H]
G_M41510_IG02:
       test     r8, r8
       je       G_M41510_IG08
G_M41510_IG03:
       test     edi, edi
       jl       G_M41510_IG09
G_M41510_IG04:
       test     ebx, ebx
       jl       G_M41510_IG10
G_M41510_IG05:
       mov      edx, dword ptr [r8+8]
       sub      edx, edi
       cmp      edx, ebx
       jl       G_M41510_IG11
G_M41510_IG06:
       mov      rcx, rsi
       mov      rdx, r8
       call     CORINFO_HELP_CHECKED_ASSIGN_REF
       mov      dword ptr [rsi+8], edi
       mov      dword ptr [rsi+12], ebx
G_M41510_IG07:
       add      rsp, 32
       pop      rbx
       pop      rsi
       pop      rdi
       ret      
```
And after
```asm
       488B4E18             mov      rcx, gword ptr [rsi+24]
       4885FF               test     rdi, rdi
       744D                 je       SHORT G_M24074_IG06
       8BD3                 mov      edx, ebx
       0BD5                 or       edx, ebp
       85D2                 test     edx, edx
       7C45                 jl       SHORT G_M24074_IG06
       8B5708               mov      edx, dword ptr [rdi+8]
       2BD3                 sub      edx, ebx
       3BD5                 cmp      edx, ebp
       7C3C                 jl       SHORT G_M24074_IG06

G_M24074_IG03:
       488D542420           lea      rdx, bword ptr [rsp+20H]
       48893A               mov      gword ptr [rdx], rdi
       895A08               mov      dword ptr [rdx+8], ebx
       896A0C               mov      dword ptr [rdx+12], ebp
```
So still shorter

src/mscorlib/src/System/ArraySegment.cs
benaadams: Changed to logical or

benaadams:
For the round [13 TechEmpower plaintext](https://www.techempower.com/benchmarks/#section=data-r13&hw=ph&test=plaintext) this method gets called 3,644,732 times per second

benaadams:
@mikedn also I completely blame you for making its transition from its original version https://referencesource.microsoft.com/#mscorlib/system/arraysegment.cs,48 to this one possible 😁

jamesqo:
Nice!

mikedn:
> For the round 13 TechEmpower plaintext this method gets called 3,644,732 times per second

Yes but I was curious if you have any numbers that show any improvements in performance.

mikedn:
Side note:
```asm
0BD5                 or       edx, ebp
85D2                 test     edx, edx
7C45                 jl       SHORT G_M24074_IG06
```
#7566 should eliminate the unnecessary `test` in the above code

PR #9436, opened 2017-02-09
[x86/Linux] Port SWCB_GetExecutionState

### LINKED ISSUE #9435: [x86/Linux] Random GC.Scenarios.BinTree.thdtree test failure
``GC.Scenarios.BinTree.thdtree`` test randomly fails with the following assert:
```
Assert failure(PID 7963 [0x00001f1b], Thread: 7968 [0x1f20]): IsAddressInStack(esb->m_ppvRetAddrPtr)
    File: /home/parjong/projects/dotnet/coreclr/src/vm/threadsuspend.cpp Line: 6912
```

parjong:
SWCB_GetExecutionState currently updates m_ppvRetAddrPtr using PCTAddr.

Unfortunately, however, there is no guarantee that PCTAddr is in actual stack range (for x86/Linux), which  results in #9435.

This commit use Caller SP based routine (which is used for other architectures) for x86/Linux in order to fix #9435.

### DIFF
--- a/src/vm/threadsuspend.cpp
+++ b/src/vm/threadsuspend.cpp
@@ -7018,7 +7018,7 @@ StackWalkAction SWCB_GetExecutionState(CrawlFrame *pCF, VOID *pData)
                 // return address for hijacking.
                 if (!pES->m_IsInterruptible)
                 {
-#if defined(_TARGET_AMD64_) || defined(_TARGET_ARM_) || defined(_TARGET_ARM64_) 
+#ifdef WIN64EXCEPTIONS
                     PREGDISPLAY pRDT = pCF->GetRegisterSet();
                     _ASSERTE(pRDT != NULL);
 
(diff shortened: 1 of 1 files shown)


parjong:
\CC @seanshpark @wateret

parjong:

src/vm/eetwain.cpp
```diff
@@ -3085,6 +3085,17 @@ void EECodeManager::EnsureCallerContextIsValid( PREGDISPLAY  pRD, StackwalkCache
     _ASSERTE( pRD->IsCallerContextValid );
 }
 
+size_t EECodeManager::GetCallerSpUnsafe( PREGDISPLAY  pRD )
```
janvorli: I don't think it is worth extracting this simple expression into a function, especially when it is obvious what it does and it is used at two places only. Could you please revert this part of the change? Otherwise LGTM.

src/vm/threadsuspend.cpp
```diff
@@ -7075,15 +7075,17 @@ StackWalkAction SWCB_GetExecutionState(CrawlFrame *pCF, VOID *pData)
                             // in the caller of the current non-interruptible frame.
                             pES->m_ppvRetAddrPtr = (void **) pRDT->pCallerContextPointers->Lr;
                         }
-#else
-                        pES->m_ppvRetAddrPtr = (void **) ((size_t)GetSP(pRDT->pCallerContext) - sizeof(void*));
-#endif
+#elif defined(_TARGET_X86_) || defined(_TARGET_AMD64_)
+                        pES->m_ppvRetAddrPtr = (void **) (EECodeManager::GetCallerSp(pRDT) - sizeof(void*));
```
janvorli: @parjong I meant just use the original form without calling the EECodeManager method. Is there a problem with it?

parjong: As we discussed in #9384, it is possible that caller's esp is not same as ``Caller-SP``. I would like to express that we need ``Caller-SP`` here.

janvorli: @parjong I am sorry, I have not realized that it was the reason for introducing the unsafe method before. But that means that this change depends on the CallerSP change and so it should not be merged before the other one, right?

parjong: @janvorli This change requires Caller SP change to be correct, but it seems that this change could be merged first as the current implementation (based on ``PCTAddr``) already has an issue.

PR #9441, opened 2017-02-09
Optimize Span.Fill

ahsonkhan:
For issue #9161 

~~Missing Unsafe.InitBlockUnaligned for when:~~
```C#
Unsafe.SizeOf<T>() == 1
```

### DIFF
--- a/src/mscorlib/src/System/Span.cs
+++ b/src/mscorlib/src/System/Span.cs
@@ -231,6 +231,7 @@ namespace System
         /// <summary>
         /// Clears the contents of this span.
         /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public void Clear()
         {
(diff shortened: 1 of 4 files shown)


src/mscorlib/src/System/Span.cs
```diff
@@ -243,12 +244,42 @@ public void Clear()
         /// <summary>
         /// Fills the contents of this span with the given value.
         /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
jkotas: This should not be marked aggressive inlining - it is too big of a method to inline.

ahsonkhan: Done. Thanks.

jkotas:
@dotnet-bot test 
Windows_NT x86 Checked Build and Test please

PR #9444, opened 2017-02-09
Unify GetControlPC/GetRegdisplaySP/SetRegdisplaySP

parjong:
This commit unifies similar implementations of GetControlPC into one.

### DIFF
--- a/src/inc/regdisp.h
+++ b/src/inc/regdisp.h
@@ -44,18 +44,25 @@ struct REGDISPLAY_BASE {
     Thread *_pThread;
 #endif // DEBUG_REGDISPLAY
 
-#if defined(_TARGET_X86_) || defined(_TARGET_ARM_)
-    DWORD   SP;                // Stack Pointer
-    PCODE   ControlPC;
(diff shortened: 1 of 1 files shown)


src/inc/regdisp.h
```diff
@@ -53,6 +53,11 @@ struct REGDISPLAY_BASE {
 #endif // !_TARGET_X86_ && !_TARGET_ARM
 };
 
+inline PCODE GetControlPC(REGDISPLAY_BASE *pRD) {
```
janvorli: It seems it would be nice to unify also the GetRegdisplaySP and SetRegdisplaySP. I wonder if we can even unify the ControlPC and SP in the REGDISPLAY_BASE to be of size_t type for all targets.

parjong: I'm not sure. It may block cross-architecture crossgen (which may be required for ARM pipeline bring up).

parjong: Anyway, I'll take a look whether it is possible to unify GetRegdisplaySP and SetRegdisplaySP.

PR #9448, opened 2017-02-09
[x86/Linux] Fix incorrect update in HijackFrame::UpdateRegDisplay

### LINKED ISSUE #9435: [x86/Linux] Random GC.Scenarios.BinTree.thdtree test failure

parjong:
While x86/Linux bring up, HijackFrame::UpdateRegDisplay is revised similarly as the corresponding x64 implementation.

It turns out that this implementation is incorrect. The implementation of HijackFrame::UpdateRegDisplay should be aligned with OnHijackTripThread, but x64 and x86 has different OnHijackTripThread implementation.

This results which results in segfault discussed in #9435.

This commit revises HijackFrame::UpdateRegDisplay in order to fix #9435.

### DIFF
--- a/src/vm/i386/cgenx86.cpp
+++ b/src/vm/i386/cgenx86.cpp
@@ -802,6 +802,7 @@ void ResumableFrame::UpdateRegDisplay(const PREGDISPLAY pRD)
 }
 
 // The HijackFrame has to know the registers that are pushed by OnHijackTripThread
+//  -> HijackFrame::UpdateRegDisplay should restore all the registers pushed by OnHijackTripThread
 void HijackFrame::UpdateRegDisplay(const PREGDISPLAY pRD)
 {
     CONTRACTL {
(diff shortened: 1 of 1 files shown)


parjong:
\CC @seanshpark @wateret @YongseopKim

src/vm/i386/cgenx86.cpp
```diff
@@ -822,19 +823,15 @@ void HijackFrame::UpdateRegDisplay(const PREGDISPLAY pRD)
...
-#define CALLEE_SAVED_REGISTER(reg) pRD->pCurrentContextPointers->reg = NULL;
-    ENUM_CALLEE_SAVED_REGISTERS();
-#undef CALLEE_SAVED_REGISTER
-
-#define ARGUMENT_AND_SCRATCH_REGISTER(reg) pRD->pCurrentContextPointers->reg = NULL;
-    ENUM_ARGUMENT_AND_SCRATCH_REGISTERS();
-#undef ARGUMENT_AND_SCRATCH_REGISTER
-
-    pRD->pCurrentContextPointers->Eax = (PDWORD) &m_Args->Eax;
+#define RESTORE_REG(reg) { pRD->pCurrentContext->reg = m_Args->reg; pRD->pCurrentContextPointers->reg = &m_Args->reg; }
```
janvorli: I wonder why you haven't just modify the previous code to set the context pointers correctly for both callee saved and argument registers and  haven't added setting the argument register values. It would be nice to keep using the enumeration of registers unified everywhere.

parjong: I thought that it would be better to make it consistent with the below x86/Windows implementation, but it would be better to use register enumeration macro as before (as you suggest)

I'll revise this using ``ENUM_ARGUMENT_AND_SCRATCH_REGISTERS`` and ``ENUM_CALLEE_SAVED_REGISTERS``

parjong:
@janvorli Could you merge this PR?

Issue #9097, opened 2017-01-24
RyuJIT: redundant array bounds checks are not eliminated in fannkuch-redux bechmark hot method


sivarv:
Here is the hot method of fannkuch-redux

```
    int CountFlips()
    {
        int flips = 1;
        int first = p[0];
        if (p[first] != 0)
        {
            Array.Copy(p, 0, pp, 0, pp.Length);
            do
            {
                ++flips;
                for (int lo = 1, hi = first - 1; lo < hi; ++lo, --hi)
                {
                    int t = pp[lo];

                    // Bounds check could be eliminated while storing to pp[lo] because
                    // prior to this pp[low] was read.
                    pp[lo] = pp[hi];

                    // Bounds check could be eliminated while storing to pp[hi] because
                   // prior to thi pp[hi] was read
                    pp[hi] = t;
                }

                int tp = pp[first];
              
                // Bounds check could be eliminated while storing to pp[first] because
                // prior to this pp[first] was read.
                pp[first] = first;
                first = tp;
            } while (pp[first] != 0);
        }
        return flips;
    }

```

In the above method, there are three places (as indicated above in code) where array bounds check could be eliminated but RyuJIT doesn't eliminate them. Looking at the jit dump, this.pp is cse'd into a local and Value Number has assigned the same VN to array-bounds-check and its operands in each of the case where bounds check is redundant.

sivarv:
CC @JosephTremoulet

sivarv:
With optRepeatCount=3, RyuJIt eliminates the above mentioned redundant bounds checks.

sivarv:
Issue #https://github.com/dotnet/coreclr/issues/5371 is related to this.

sivarv:
Another hot method of fannkuch-redux benchmark NextPermutation suffers from redundant bounds checks and not loop hoisting invariant field accesses https://github.com/dotnet/coreclr/issues/9036

```
    bool NextPermutation()
    {
        int first = p[1];

        // bounds check on p[1] is not eliminated
       // bounds check on p[0] is not eliminated
        p[1] = p[0];

       // bounds check on p[0] is not eliminated
        p[0] = first;

        int i = 1;

        // Invariant field accesses this.count and this.count.Length could be loop hoisted
        while (++count[i] > i)
        {
            count[i++] = 0;

            // bounds check on p[0] and p[1] are not eliminated.
            int next = p[0] = p[1];           
            for (int j = 1; j < i; ++j)
            {
[...]

Issue #5371, opened 2016-06-01
RyuJIT bounds checks not eliminated for readonly arrays


benaadams:
Range analysis doesn't take into account readonly arrays and you need to make a function local reference to eliminate the range check.

Analysis https://gist.github.com/benaadams/11123e9785481e07d8ca483bef798689

| Method | Median | StdDev | Scaled |
| --- | --- | --- | --- |
| ForLoopArray | 2.5021 us | 0.0591 us | 1.00 |
| ForLoopOneBoundsCheck | 2.3430 us | 0.0296 us | 0.94 |
| ForLoopReadOnlyArray | 2.5047 us | 0.0334 us | 1.00 |
| ForLoopAddPointer | 2.3360 us | 0.0059 us | 0.93 |
| ForEachArray | 2.3432 us | 0.0051 us | 0.94 |

omariom:
Range check can't be eliminated for  readonly instance fields because ctor's can call other methods.
But   if a static readonly field  has already been initialized JIT could  skip generatin range checks.

GSPP:
What's the CLR's policy on changing readonly fields through reflection? Is this undefined behavior? I believe so for the static case. I have seen the JIT inline static readonly constants at JIT time.

omariom:
> I have seen the JIT inline static readonly constants at JIT time.

It currently works for fields of primitive types. 
would be interesting to extend it further to arrays and  [dynamic dispatch](https://github.com/dotnet/coreclr/issues/4108)

GSPP:
I was more interested in what format guarantees there are. I guess for static readonly there are none. But what about:

    var x = this.readonlyField;
    SetXThroughReflection();
    Assert(x == this.readonlyField);


Can this ever throw? The JIT might assume the field cannot change.

For a static readonly field this would always fail.

omariom:
wow! I has  been sure Reflection doesn't allow setting readonly fields.
But it does!

```C#
void Main()
{
// Static
	var staticField = typeof(TestClass).GetField("StaticField", BindingFlags.Static | BindingFlags.Public);
	
	staticField.SetValue(null, 6);
	
	Console.WriteLine(staticField.GetValue(null));

// Instance
	var field = typeof(TestClass).GetField("Field", BindingFlags.Instance | BindingFlags.Public);
	var instance = new TestClass();
	field.SetValue(instance, 6);

	Console.WriteLine(field.GetValue(instance));
}

class TestClass
{
[...]

PR #9451, opened 2017-02-09
Update conservative value numbers during CSE

JosephTremoulet:
When a CSE candidate's defs all share the same conservative value number,
its uses can be updated to share that conservative value number as well
when CSE is performed, because we are removing any reloads that may have
been the cause of the divergence.  Performing this update can improve
subsequent range check elimination when the CSE use is array length or
index in a bounds check.

### DIFF
--- a/src/jit/compiler.h
+++ b/src/jit/compiler.h
@@ -5388,6 +5388,9 @@ protected:
 
         treeStmtLstPtr csdTreeList; // list of matching tree nodes: head
         treeStmtLstPtr csdTreeLast; // list of matching tree nodes: tail
+
+        ValueNum defConservativeVN; // if all def occurrences share the same conservative value
+                                    // number, this will reflect it; otherwise, NoVN.
     };
 
     static const size_t s_optCSEhashSize;
--- a/src/jit/optcse.cpp
+++ b/src/jit/optcse.cpp
@@ -996,6 +996,17 @@ void Compiler::optValnumCSE_Availablity()
 
                         /* This is a CSE def */
 
+                        if (desc->csdDefCount == 0)
+                        {
+                            // This is the first def visited, so copy its conservative VN
+                            desc->defConservativeVN = tree->gtVNPair.GetConservative();
+                        }
+                        else if (tree->gtVNPair.GetConservative() != desc->defConservativeVN)
+                        {
+                            // This candidate has defs with differing conservative VNs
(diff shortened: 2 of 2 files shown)


JosephTremoulet:
@dotnet/jit-contrib PTAL; I implemented @mikedn's suggestion from #5371.  It gets the related (straight-line) case in #9097; it doesn't get the original (loop) case from #5371, but is an incremental step in the right direction nonetheless.

Diffs are a bunch of removed bounds checks and null checks.

SPMI asm diffs report 0.83% codesize improvement across 6474 affected methods, jit-analyze summary as follows:

```
Summary:
(Note: Lower is better)

Total bytes of diff: -4882 (0.00 % of base)
    diff is an improvement.

       -1196 : dasmset_1\diff\Methodical\fp\exgen\10w5d_cs_do\10w5d_cs_do.dasm (-0.13 % of base)
       -1195 : dasmset_1\diff\Methodical\fp\exgen\10w5d_cs_ro\10w5d_cs_ro.dasm (-0.14 % of base)
        -925 : dasmset_1\diff\System.Private.CoreLib.dasm (-0.03 % of base)
        -167 : dasmset_1\diff\System.Linq.Parallel.dasm (-0.03 % of base)
        -162 : dasmset_1\diff\Microsoft.CSharp.dasm (-0.04 % of base)

65 total files with size differences (65 improved, 0 regressed).

         -63 : dasmset_1\diff\System.Private.CoreLib.dasm - System.Globalization.CultureData:get_SENGDISPLAYNAME():ref:this
         -61 : dasmset_1\diff\System.Reflection.Metadata.dasm - System.Reflection.PortableExecutable.PEReader:GetPESectionBlock(int):ref:this
         -56 : dasmset_1\diff\System.Linq.Parallel.dasm - System.Linq.Parallel.FixedMaxHeap`1[Int32][System.Int32]:HeapifyRoot():this
         -52 : dasmset_1\diff\System.Private.CoreLib.dasm - System.Globalization.DateTimeFormatInfo:get_DateTimeOffsetPattern():ref:this
         -49 : dasmset_1\diff\Methodical\fp\exgen\10w5d_cs_do\10w5d_cs_do.dasm - testout1:Func_0_1_1_2_1():double

466 total methods with size differences (466 improved, 0 regressed).
```

briansull:
LGTM, 
 I think these are some of the same range checks that we get if we re-run value numbering a second time after we do the CSE phase...

Issue #7641, opened 2016-10-16
RyuJIT bit testing issues


mikedn:
* `return (i4 & 16) == 0;` works fine:
```
       F7C110000000         test     ecx, 16
       0F94C0               sete     al
       0FB6C0               movzx    rax, al
```
A `shr`/`and`/`not` sequence would be shorter but probably a bit slower so this is fine as it is now.

* `return (i4 & 16) != 0;` fails to eliminate the `and`:
```
       83E110               and      ecx, 16
       85C9                 test     ecx, ecx
       0F97C0               seta     al
       0FB6C0               movzx    rax, al
```
This happens because there's no `cne` instrunction in IL so the C# compiler emits `cgt.un`. The JIT fails to recognize GT_GT when attempting to eliminate the `and` in `TreeNodeInfoInitCmp`.

In this case a `shr`/`and` sequence would likely be beneficial as it is shorter and should have the same speed. It's also useful on x86 as it doesn't require byte regs. This should be trivial to implement.

* `return (i8 & (1L << 42)) == 0;` fails to eliminate the `and`:
```
       48B80000000000040000 mov      rax, 0x40000000000
       4823C8               and      rcx, rax
       4885C9               test     rcx, rcx
       0F94C0               sete     al
       0FB6C0               movzx    rax, al
```
This happens because the test constant cannot be contained. This is simply a side effect of the peculiar split of this optimization between lower and codegen. I'm not yet sure why codegen expects the test constant to be contained, if it really needs that then we might have to add `GT_TEST_EQ` to get this to work properly.

A `shr`/`and`/`not` might be useful in this case as it avoids the need to load the large test constant into a register.

* `return (x & y) == 0;` fails to eliminate the `and`:
```
       23CA                 and      ecx, edx
       85C9                 test     ecx, ecx
       0F94C0               sete     al
       0FB6C0               movzx    rax, al
```
Likely a variation of the containment issue mentioned above.

This also applies to testing groups of bits (except that `shr`/`and` is usually not an option anymore).

mikedn:
#7787 partially took care of `return (x & y) == 0;` and `return (i8 & (1L << 42)) == 0;`. "Partially" because instead of eliminating the `and` the `test` gets eliminated.

mikedn:
`return (i4 & 16) != 0;` was fixed in #9454:
```asm
F7C110000000         test     ecx, 16
0F95C0               setne    al
0FB6C0               movzx    rax, al
```
[...]

PR #9454, opened 2017-02-09
Morph (x GT_GT.unsigned 0) into (x GT_NE 0)

### LINKED ISSUE #9419: Optimize default(T) != null at compile time
`default(T) != null` in generic code is a popular* pattern to determine if `T` is a nullable or a reference type.

Example:

```csharp
using System;

class Program
{
    static bool IsNotNullableValueType<T>()
    {
        return default(T) != null;
    }

    {
        Console.WriteLine(IsNotNullableValueType<string>());
        Console.WriteLine(IsNotNullableValueType<int>());
        Console.WriteLine(IsNotNullableValueType<int?>());
    }
}
```

Will print `False True False`.

While the codegen for the "it's a reference type" case is pretty efficient (we can do better though...):

```nasm
xor         eax,eax  
test        rax,rax  
setb        al  
movzx       eax,al  
ret  
```

For the "it's a valuetype" and "it's a nullable" case we get a useless allocation:

```nasm
sub         rsp,28h  
call        __NewHelper_System_Private_CoreLib_System_Int32
xor         edx,edx  
mov         dword ptr [rax+8],edx  
test        rax,rax  
seta        al  
movzx       eax,al  
add         rsp,28h  
ret  
```

The places that use this pattern would likely benefit from being able to statically evaluate this because it could lead to elimination of entire branches and smaller code (aside from removing a useless allocation).

\* There are 17 hits for the `default\(.*\) == null` regular expression in the CoreFX repo and a couple more for the "not equal to null" case.

mikedn:
IL doesn't have a cne instruction so compilers use cgt.un instead. The JIT recognizes certain patterns that involve `GT_NE` (e.g `(x & 4) != 0`) and fails if `GT_GT` is used instead. Transform `x GT_GT.unsigned 0` into `x GT_NE 0` and `x GT_LE.unsigned 0` into `x GT_EQ 0`. The later case is rare, it sometimes occurs as a result of branch inversion.

Fixes #9419 
Contributes to #7641

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -12213,6 +12213,23 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
                         }
                     }
                 }
+                else // we have an unsigned comparison
+                {
+                    if (op2->IsIntegralConst(0))
+                    {
+                        if ((oper == GT_GT) || (oper == GT_LE))
+                        {
+                            // IL doesn't have a cne instruction so compilers use cgt.un instead. The JIT
+                            // recognizes certain patterns that involve GT_NE (e.g (x & 4) != 0) and fails
+                            // if GT_GT is used instead. Transform (x GT_GT.unsigned 0) into (x GT_NE 0)
+                            // and (x GT_LE.unsigned 0) into (x GT_EQ 0). The later case is rare, it sometimes
+                            // occurs as a result of branch inversion.
+                            oper = (oper == GT_LE) ? GT_EQ : GT_NE;
+                            tree->SetOper(oper, GenTree::PRESERVE_VN);
+                            tree->gtFlags &= ~GTF_UNSIGNED;
+                        }
+                    }
+                }
 
             COMPARE:
 


mikedn:
jit-diff summary:
```
Total bytes of diff: -9909 (-0.08 % of base)
    diff is an improvement.
           1 : System.Collections.Concurrent.dasm (0.00 % of base)
       -3569 : Microsoft.CodeAnalysis.VisualBasic.dasm (-0.24 % of base)
       -2213 : Microsoft.CodeAnalysis.CSharp.dasm (-0.16 % of base)
       -1233 : System.Private.CoreLib.dasm (-0.03 % of base)
        -479 : System.Xml.ReaderWriter.dasm (-0.12 % of base)
        -384 : Microsoft.CSharp.dasm (-0.10 % of base)
30 total files with size differences (29 improved, 1 regressed).
           6 : Microsoft.CodeAnalysis.VisualBasic.dasm - Microsoft.CodeAnalysis.VisualBasic.CodeGen.CodeGenerator:StackMergeType(ref):ref:this
           5 : Microsoft.CodeAnalysis.VisualBasic.dasm - Microsoft.CodeAnalysis.VisualBasic.OverloadResolution:DetermineMostSpecificNarrowingConversion(ref,ref,ref,ref,byref,byref,bool,byref):bool
           2 : Microsoft.CodeAnalysis.VisualBasic.dasm - Microsoft.CodeAnalysis.VisualBasic.OverloadResolution:DetermineMostSpecificWideningConversion(ref,ref,ref,ref,byref,byref,bool,byref):bool
           1 : Microsoft.CodeAnalysis.CSharp.dasm - Microsoft.CodeAnalysis.CSharp.Symbols.SourceTypeParameterSymbol:get_HasReferenceTypeConstraint():bool:this
           1 : Microsoft.CodeAnalysis.CSharp.dasm - Microsoft.CodeAnalysis.CSharp.Symbols.SourceTypeParameterSymbol:get_HasValueTypeConstraint():bool:this
         -74 : System.Private.CoreLib.dasm - System.Resources.ResourceManager:.ctor(ref,ref,ref):this (2 methods)
         -74 : System.Private.CoreLib.dasm - System.Resources.ResourceManager:OnDeserialized(struct):this
         -66 : Microsoft.CodeAnalysis.VisualBasic.dasm - Microsoft.CodeAnalysis.VisualBasic.Symbols.SynthesizedStringSwitchHashMethod:GetBoundMethodBody(ref,byref):ref:this
         -66 : Microsoft.CodeAnalysis.VisualBasic.dasm - Microsoft.CodeAnalysis.VisualBasic.BoundForToStatement:.ctor(ref,ref,ref,ref,bool,ref,ref,ref,ref,struct,ref,ref,bool):this
         -58 : System.Reflection.Metadata.dasm - System.Reflection.Metadata.MetadataReader:InitializeTableReaders(struct,ubyte,ref,ref):this
1411 total methods with size differences (1385 improved, 26 regressed).
```

mikedn:
Before and after examples:
```asm
; before (somewhere in Microsoft.CodeAnalysis.CSharp)
       movzx    rdx, byte  ptr [rbx+17]
       and      edx, 1
       test     edx, edx
       seta     dl
; after
       test     byte  ptr [rbx+17], 1
       setne    dl

; before (somewhere in System.Linq.Parallel)
       mov      rsi, rax
       call     [CORINFO_HELP_READYTORUN_NEW]
       mov      r12, rax
       lea      rcx, bword ptr [rsi+8]
       xorpd    xmm0, xmm0
       movdqu   qword ptr [rcx], xmm0
       test     rsi, rsi
       seta     r13b
       movzx    r13, r13b
; after
       mov      r13d, 1
```
The occasional size regressions are due to instruction encoding differences:
```asm
; before
       and      edx, 1
       test     edx, edx
       seta     al
; after
       test     edx, 1
       setne    al
```
Unlike `and` `test` doesn't have a r/m32,imm8 encoding so it needs 6 bytes instead of 3. `LowerCompare` could be extended to narrow down the instruction to `test dl, 1` to avoid this issue.

mikedn:
@dotnet-bot test Ubuntu x64 Formatting please

briansull:
LGTM

Issue #8102, opened 2016-11-12
Enable callers of S.P.CoreLib NoInlining methods to inline.


jamesqo:
Currently, methods that are marked with `NoInlining` within S.P.CoreLib are assumed to contain a StackCrawlMark local variable. Because of this, it is assumed it is not safe to inline the caller, which can unnecessarily degrade performance under certain circumstances where StackCrawlMark is not being used. There needs to be a better way for identifying `StackCrawlMark` methods so we can overcome this limitation.

Moved from #7449

PR #9470, opened 2017-02-09
JIT: fix confusing inline failure reason

AndyAyersMS:
When attempting to inline a method from the core library, the inline
will fail if the method being inlined has a noinline callee; the
presumption being that the callee expects to be able to find its
immediate caller on the stack (this may or not may be the case, see #8102).

Update the failure message for this case to hopefully better explain
why the inline doesn't happen.


AndyAyersMS:
@stephentoub PTAL
cc @dotnet/jit-contrib

PR #9471, opened 2017-02-09
Reduce Task and async method overheads

stephentoub:
This PR attempts to reduce the overhead associated with executing Tasks and invoking async methods.

Microbenchmarks look pretty good:
- Repeatedly creating an ```AsyncTaskMethodBuilder<bool>``` and completing it "synchronously" (before accessing its Task) with a known cached value: *1.95x*
- Repeatedly creating an ```AsyncTaskMethodBuilder<int>``` and completing it "synchronously" (before accessing its Task) with a known non-cached value:  *1.53x*
- Repeatedly creating an ```AsyncTaskMethodBuilder<int>``` and completing it "asynchronously" (after accessing its Task):  *1.38x*
- Repeatedly invoking an empty async ```Task<bool>``` method: *1.26x*
- Repeatedly creating and completing a TaskCompletionSource<int>(): *1.37x*
- Repeatedly creating and completing a TaskCompletionSource<int>(object): *1.29x*
- Having ProcessorCount worker threads continually queueing a task to its local queue, popping/executing it: *1.06x*
- Having ProcessorCount worker threads continually queueing a task to the global queue, dequeueing/executing one: Fluctuated within noise +/- 1-2%
- Long chain of synchronous continuations: Fluctuated within noise +/- 1-2%
- Long chain of asynchronous continuations: Fluctuated within noise +/- 1-2%

(I'd like to hold off on merging though until doing some more verification on ASP.NET workloads to make sure they're at least not negatively affected.)

cc: @jkotas, @kouvel, @benaadams

### DIFF
--- a/src/mscorlib/src/System/Runtime/CompilerServices/AsyncMethodBuilder.cs
+++ b/src/mscorlib/src/System/Runtime/CompilerServices/AsyncMethodBuilder.cs
@@ -63,7 +63,7 @@ namespace System.Runtime.CompilerServices
             // See comment on AsyncMethodBuilderCore.Start
             // AsyncMethodBuilderCore.Start(ref stateMachine);
 
-            if (stateMachine == null) throw new ArgumentNullException(nameof(stateMachine));
+            if (stateMachine == null) ThrowHelper.ThrowArgumentNullException(ExceptionArgument.stateMachine);
             Contract.EndContractBlock();
 
             // Run the MoveNext method within a copy-on-write ExecutionContext scope.
@@ -72,7 +72,6 @@ namespace System.Runtime.CompilerServices
 
             Thread currentThread = Thread.CurrentThread;
             ExecutionContextSwitcher ecs = default(ExecutionContextSwitcher);
-            RuntimeHelpers.PrepareConstrainedRegions();
             try
             {
                 ExecutionContext.EstablishCopyOnWriteScope(currentThread, ref ecs);
@@ -240,15 +239,8 @@ namespace System.Runtime.CompilerServices
             }
         }
 
-        // This property lazily instantiates the Task in a non-thread-safe manner.  
-        private Task Task 
-        {
-            get
-            {
-                if (m_task == null) m_task = new Task();
-                return m_task;
-            }
-        }
+        /// <summary>Lazily instantiate the Task in a non-thread-safe manner.</summary>
+        private Task Task => m_task ?? (m_task = new Task());
 
         /// <summary>
         /// Gets an object that may be used to uniquely identify this builder to the debugger.
@@ -295,7 +287,7 @@ namespace System.Runtime.CompilerServices
             // See comment on AsyncMethodBuilderCore.Start
             // AsyncMethodBuilderCore.Start(ref stateMachine);
 
-            if (stateMachine == null) throw new ArgumentNullException(nameof(stateMachine));
+            if (stateMachine == null) ThrowHelper.ThrowArgumentNullException(ExceptionArgument.stateMachine);
             Contract.EndContractBlock();
 
             // Run the MoveNext method within a copy-on-write ExecutionContext scope.
@@ -304,7 +296,6 @@ namespace System.Runtime.CompilerServices
 
             Thread currentThread = Thread.CurrentThread;
             ExecutionContextSwitcher ecs = default(ExecutionContextSwitcher);
-            RuntimeHelpers.PrepareConstrainedRegions();
             try
             {
                 ExecutionContext.EstablishCopyOnWriteScope(currentThread, ref ecs);
@@ -358,7 +349,11 @@ namespace System.Runtime.CompilerServices
         /// <summary>Gets the <see cref="System.Threading.Tasks.Task"/> for this builder.</summary>
         /// <returns>The <see cref="System.Threading.Tasks.Task"/> representing the builder's asynchronous operation.</returns>
         /// <exception cref="System.InvalidOperationException">The builder is not initialized.</exception>
-        public Task Task { get { return m_builder.Task; } }
+        public Task Task
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            get { return m_builder.Task; }
+        }
 
         /// <summary>
         /// Completes the <see cref="System.Threading.Tasks.Task"/> in the 
@@ -449,7 +444,7 @@ namespace System.Runtime.CompilerServices
             // See comment on AsyncMethodBuilderCore.Start
             // AsyncMethodBuilderCore.Start(ref stateMachine);
 
-            if (stateMachine == null) throw new ArgumentNullException(nameof(stateMachine));
+            if (stateMachine == null) ThrowHelper.ThrowArgumentNullException(ExceptionArgument.stateMachine);
             Contract.EndContractBlock();
 
             // Run the MoveNext method within a copy-on-write ExecutionContext scope.
@@ -458,7 +453,6 @@ namespace System.Runtime.CompilerServices
 
             Thread currentThread = Thread.CurrentThread;
             ExecutionContextSwitcher ecs = default(ExecutionContextSwitcher);
-            RuntimeHelpers.PrepareConstrainedRegions();
             try
             {
                 ExecutionContext.EstablishCopyOnWriteScope(currentThread, ref ecs);
@@ -502,7 +496,7 @@ namespace System.Runtime.CompilerServices
                 {
                     // Force the Task to be initialized prior to the first suspending await so 
                     // that the original stack-based builder has a reference to the right Task.
-                    var builtTask = this.Task;
+                    Task builtTask = this.Task;
 
                     // Box the state machine, then tell the boxed instance to call back into its own builder,
                     // so we can cache the boxed reference. NOTE: The language compiler may choose to use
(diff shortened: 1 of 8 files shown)


stephentoub:
@benaadams, would you mind giving this a go with your various test suites (microbenchmarks and ASP.NET) to see how well this does for you?

benaadams:
Good timing; have test-bed already setup and running 😄

stephentoub:
> Good timing; have test-bed already setup and running 

:smile:

benaadams:
Don't know how much its used but line 2330 in TaskFactory.cs could change from
```csharp
if (Interlocked.CompareExchange(ref m_firstTaskAlreadyCompleted, 1, 0) == 0)
```
to?
```csharp
if (Interlocked.Exchange(ref m_firstTaskAlreadyCompleted, 1) == 0)
```

benaadams:
Definitely hasn't regressed; peaking up to 3.3GiB/s
```
Requests/sec: 2,478,571.74
Requests/sec: 2,561,367.13
```
I get the odd run that locks at ~1.9GiB/s but was getting that before this change. Is probably an entire server rack complaining about their noisy neighbour 😄 

Will pulls some traces

benaadams:
Plaintext Traces https://aoa.blob.core.windows.net/aspnet/HttpTest-Task.zip (97MB) 
All dlls should also be there

Sampling trace in `HttpTest-TaskSampling.dtp` (209,135,165 requests)
Call count trace in `HttpTest-TaskTracing.dtp` (1,458,706 requests)
Code in `HttpTest.zip`
Exe is `HttpTest.exe` listens on port 80

Will try the task and treadpool tests in morning.

benaadams:
We'll likely merge clash in ThrowHelper https://github.com/dotnet/coreclr/pull/9473

benaadams:
AsyncMethodBuilder.cs line 888

`Debugger.NotifyOfCrossThreadDependency();` can't be inlined because `Debugger:NotifyOfCrossThreadDependencySlow` is marked as NoInlining; which in mscorlib means it has a StackCrawMark which means its caller also won't be inlined. This bail is then baked into the NGen image.



Have been handling this with an extra call to separate NoInline one more call away. (e.g. [in ThrowHelper](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/ThrowHelper.cs#L269-L286))

As there are already two calls might be worth removing the `IsAttached` test from `NotifyOfCrossThreadDependency` adding a comment about `NoInining` being weird and then move the `Debugger.IsAttached` test into `AsyncMethodBuilderCore.GetCompletionAction`.

e.g.
```csharp
if (Debugger.IsAttached)
    Debugger.NotifyOfCrossThreadDependency();
```

benaadams:
Hmm... is public, might have to leave the inner `Debugger.IsAttached` test in and just add the if to `GetCompletionAction`

benaadams:
Raised PR for Debugger https://github.com/dotnet/coreclr/pull/9490

stephentoub:
Thanks, @benaadams and @kouvel.

benaadams:
@stephentoub will do some longer runs today with the latest merges; and the Task tests

stephentoub:
Thanks, @benaadams.

benaadams:
Running tests on updated libs https://aoa.blob.core.windows.net/aspnet/HttpTest-Task2.zip (106MB)
`HttpTest.exe` plaintext server
`ThreadPoolTaskTesting.exe` task tests

Will post traces when finished; using the `.ni` from release

benaadams:
Taking a little while; made the iterations a little high I think. But better for sampling.

benaadams:
`ThreadPoolTaskTesting2.dtp` https://aoa.blob.core.windows.net/aspnet/ThreadPoolTaskTesting2.zip (34MB)

benaadams:
`HttpTest-Task2.dtp` https://aoa.blob.core.windows.net/aspnet/HttpTest-Task2.zip (1.2MB)
```
  40 threads and 1000 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     5.45ms    9.55ms 499.24ms   95.16%

  2,822,861,095 requests in 20.00m, 347.03GB read
  Socket errors: connect 19, read 698, write 0, timeout 0
```

benaadams:
`HttpTest-Task2a.dtp` https://aoa.blob.core.windows.net/aspnet/HttpTest-Task2a.zip (1.2MB)
```
  40 threads and 1000 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     5.67ms    6.98ms 488.99ms   93.15%

  7,524,482,848 requests in 60.00m, 0.90TB read
  Socket errors: connect 19, read 1876, write 0, timeout 0
```

stephentoub:
Thanks for running these, Ben. What am I looking at here?  Are these numbers better? worse? same?

benaadams:
Long running execution samples to see what's being executed. The workloads are definitely better :)

There are some slight regressions when very many threads end up queuing batches to the global threadpool; but they aren't consistent. As discussed previously that looks like increased contention from the essentially empty tasks going even faster, rather than a practical issue. One thread creating single threadpool/Task work on 20 core VM was always an outliner (1/4 speed of rest) and that seems to have gone.

ASP.NET workload looks good.

benaadams:
LGTM

davidfowl:
Ideally we'd run our benchmarks as well before the ASP.NET workload is considered "good". 

@benaadams it would be good to understand what kinds of tests you're running and what spectrum of things are being covered. If you have some that we're missing, we should get them into our benchmarks.

stephentoub:
> Ideally we'd run our benchmarks as well before the ASP.NET workload is considered "good".

I've been running my own that test a spectrum of things that hit these code paths, in addition to the ASP.NET plaintext and JSON workloads. Ben's been helpful in providing additional verification. What other scenarios are you concerned this will regress, David?

davidfowl:
I don't have any problems with this change in particular. It be cool if we were all running the same set of tests. Also I'd like it if somebody from our team also ran benchmarks for changes where you think it may affect our scenarios. More eyes on it is a good thing. 

PS: Ben I do trust you 😁

stephentoub:
> Also I'd like it if somebody from our team also ran benchmarks for changes where you think it may affect our scenarios

I would like that, too.

benaadams:
>  it would be good to understand what kinds of tests you're running 

Am [testing 12 different ways to call and use Tasks](https://github.com/benaadams/ThreadPoolTaskTesting), with awaiting, continuations, fanout, chains and direct returns. Via both the threadpool and dedicated threads. Also calling QUWI from both dedicated threads and threadpool. These are tested being called serially and in parallel (x1, x2, x16, x64, x512) and at different recursion depths (x1, x16, x64, x512).

In total 650 different tests ((12+1) x 2 x 5 x 5); which are then run in batches of 262,144 iterations (then pause for full GC)

The [`ThreadPoolTaskTesting2.dtp`](https://aoa.blob.core.windows.net/aspnet/ThreadPoolTaskTesting2.zip) sampling profile above was run on a 20 core 2 Numa node D15 v2, and is 132MB unzipped for 17,039,360,000 total test cycles taking about 6 hrs.

The [`HttpTest-Task2a.dtp`](https://aoa.blob.core.windows.net/aspnet/HttpTest-Task2a.zip) sampling profile above is a simple plaintext and was run between two of the same VMs for an hour for 7,524,482,848 requests, throughput did decline slowly over the profiling.

Though the traces are mostly for identifying the hot methods than the actual performance.

I also run the Task tests without profiling on 4 core, 8(16HT) core and 20 core VM.  The plaintext test on the VM which is where it is bursting up to 3.3.GiB/s for requests/sec: 2,561,367.13; though I am testing over shared networking between two VMs in two different Azure subscriptions and it does vary between runs.

The ASP.NET test is particularly important because as the code paths for the global threadpool have become more efficient; it has increased contention on it for the zero workload tests, with the contention becoming stronger as the core count increases and across Numa nodes. 

Adding a workload to the threadpool items decreases this contention and is more real world; with the ASP.NET being a very relevant workload to use to do this. 

Plaintext is the shortest task workload to check the contention "relief" on; the others would mostly be in their own code and decrease the contention much more significantly.

The simple plaintext server code was included in earlier bundle in thread though is https://aoa.blob.core.windows.net/aspnet/HttpTest-Code.zip
The Task tests are on github https://github.com/benaadams/ThreadPoolTaskTesting

The binaries for coreclr and corefx from current source, aspnet from myget and the two executables are in earlier comment https://github.com/dotnet/coreclr/pull/9471#issuecomment-279153946

benaadams:
> > Also I'd like it if somebody from our team also ran benchmarks for changes where you think it may affect our scenarios

> I would like that, too.

I'd like that too 😄

mikeharder:
I will run Plaintext before and after this change.  Are we expecting any change in throughput?

stephentoub:
> Are we expecting any change in throughput?

I wouldn't expect any significant impact; plaintext doesn't use tasks much, and the impact here is mostly on the overhead associated with task execution.  It's possible that the combination of this PR, https://github.com/dotnet/coreclr/pull/9562, and https://github.com/dotnet/coreclr/pull/9560 might move the needle slightly, as together they'll end up reducing the overhead associated with async methods, but I wouldn't expect anything huge.

mikeharder:
Before: 2.0.0-beta-001559-00 (2/10/17 21:38 UTC)
Commit: 2/12/17 00:58 UTC
After: 2.0.0-beta-001577-00 (2/13/17 19:09 UTC)

benaadams:
There are a few other related changes like the no-line vs stackcrawl mark,
it may be easier to compare before to current rather than try to pickout
the individual commits

mikeharder:
1577 is currently the latest build at https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.App

mikeharder:
No significant difference between 1559 and 1577 for Plaintext.

PR #9472, opened 2017-02-09
RM localization parity with desktop

ramarag:
Fix for https://github.com/dotnet/corefx/issues/12625
comment: https://github.com/dotnet/corefx/issues/12625#issuecomment-257447037


jkotas:
cc @morganbr

src/classlibnative/inc/nlsinfo.h
```diff
@@ -84,11 +84,6 @@ class COMNlsInfo {
     static BOOL QCALLTYPE InternalGetUserDefaultUILanguage(QCall::StringHandleOnStack userDefaultUiLanguage);
     static BOOL QCALLTYPE InternalGetSystemDefaultUILanguage(QCall::StringHandleOnStack systemDefaultUiLanguage);
 
-// Added but disabled from desktop in .NET 4.0, stayed disabled in .NET 4.5
```
gkhanna79: Since this will be a breaking change from 1.0.X and 1.1.x, do we want to enable this under a compat switch?

CC @tarekgh

tarekgh:
> Since this will be a breaking change from 1.0.X and 1.1.x, do we want to enable this under a compat switch?

I think this was the plan. @ramarag please talk to @AlexGhiondea for how you can use AppContext to add this switch. ASP.NET team requested that. Thanks.

morganbr:
@ramarag, can you please port this change to ProjectN/CoreRT as well?

ramarag:
@gkhanna79 @tarekgh I discussed this with @AlexGhiondea.

We should not be adding AppContext switches for all breaking changes. 
First a summary of history on this behavior for the record, For the complete detail [see](https://github.com/dotnet/coreclr/blob/e2d70ff4c0a30c947d572c0604f7cd1a44f258b7/src/mscorlib/src/System/Resources/ResourceFallbackManager.cs#L184):

This behavior was introduced in silverlight and never made it to desktop due to app-compat reasons. This behavior makes more sense for UI stack as there are multiple ways\reasons for app or user to override the culture. Since coreclr has pretty much carried over the code base from silverlight. We see the behavior from the above issue.

I think we should just be compatible with desktop, as this feature does not make sense from the server or console perspective.

@weshaggard comments ?

jkotas:
> I think we should just be compatible with desktop

I agree. I do not think we need to add a proactive quirk for this.

gkhanna79:
Thanks for the explanation @ramarag  - it sounds reasonable to me.

weshaggard:
>I think we should just be compatible with desktop

I agree with this as well. We should only introduce a quirk if we are aware of scenarios that will be broken and will need to opt-in to the old behavior.

tarekgh:
@gkhanna79 @ramarag we need to communicate this to the ASP.NET so they will be aware about it. Thanks.

tarekgh:
Chanes LGTM. thanks @ramarag

gkhanna79:
Adding @DamianEdwards and @Eilon so that they are aware of this as well.

PR #9473, opened 2017-02-09
GcHandle Perf Tweaks

benaadams:
GCHandle is either mostly a thin wrapper around VM functions; or simple bitshifts and casts. However its current implementation has significant overhead as it is called a huge amount and its functions are marked by NGen as permanently uninlinable.

Previously 2,028.497 GCHandle calls for 158,475 libuv write calls (in Marshal)

After 814 GCHandle calls for 2,760,965 libuv write calls (in Marshal)

Mostly they are fairly simple, so for example while GCHandle:SetIsPinned is unprofitable and permanently no inline; is asm ends up extremely simple

```asm
Marking GCHandle:SetIsPinned():this as NOINLINE because of unprofitable inline
Successfully inlined IntPtr:op_Explicit(long):long (9 IL bytes) (depth 1) [below ALWAYS_INLINE size]
Successfully inlined IntPtr:.ctor(long):this (9 IL bytes) (depth 1) [below ALWAYS_INLINE size]
**************** Inline Tree
Inlines into 0600369E GCHandle:SetIsPinned():this
  [1 IL=0007 TR=000004 06000C06] [below ALWAYS_INLINE size] IntPtr:op_Explicit(long):long
  [2 IL=0015 TR=000017 06000BF6] [below ALWAYS_INLINE size] IntPtr:.ctor(long):this
Budget: initialTime=138, finalTime=146, initialBudget=1380, currentBudget=1380
Budget: initialSize=724, finalSize=724
; Assembly listing for method GCHandle:SetIsPinned():this
; optimized code
; rsp based frame
; partially interruptible
;
;  V00 this         [V00,T00] (  4,   4  )   byref  ->  rcx         this
;* V01 tmp0         [V01    ] (  0,   0  )    long  ->  zero-ref   
;  V02 tmp1         [V02,T01] (  2,   4  )    long  ->  rax         ld-addr-op
;  V03 tmp2         [V03,T02] (  2,   4  )    long  ->  rax        
;# V04 OutArgs      [V04    ] (  1,   1  )  lclBlk ( 0) [rsp+0x00]  
;
; Lcl frame size = 0

G_M19266_IG01:
       0F1F440000           nop      

G_M19266_IG02:
       488B01               mov      rax, qword ptr [rcx]
       4883C801             or       rax, 1
       488901               mov      qword ptr [rcx], rax

G_M19266_IG03:
       C3                   ret      

; Total bytes of code 16, prolog size 5 for method GCHandle:SetIsPinned():this
```

### DIFF
--- a/src/mscorlib/src/System/Runtime/InteropServices/GcHandle.cs
+++ b/src/mscorlib/src/System/Runtime/InteropServices/GcHandle.cs
@@ -7,7 +7,6 @@ namespace System.Runtime.InteropServices
     using System;
     using System.Runtime.CompilerServices;
     using System.Threading;
-    using System.Runtime.Versioning;
     using System.Diagnostics.Contracts;
 
     // These are the types of handles used by the EE.  
@@ -57,7 +56,7 @@ namespace System.Runtime.InteropServices
         {
             // Make sure the type parameter is within the valid range for the enum.
             if ((uint)type > (uint)MaxHandleType)
-                throw new ArgumentOutOfRangeException(nameof(type), Environment.GetResourceString("ArgumentOutOfRange_Enum"));
+                ThrowArgumentOutOfRangeException_ArgumentOutOfRange_Enum();
             Contract.EndContractBlock();
 
             m_handle = InternalAlloc(value, type);
@@ -65,7 +64,7 @@ namespace System.Runtime.InteropServices
             // Record if the handle is pinned.
             if (type == GCHandleType.Pinned)
                 SetIsPinned();
-        }  
+        }
 
         // Used in the conversion functions below.
         internal GCHandle(IntPtr handle)
@@ -89,36 +88,21 @@ namespace System.Runtime.InteropServices
             return new GCHandle(value, type);
         }
 
-
         // Frees a GC handle.
         public void Free()
         {
-            // Copy the handle instance member to a local variable. This is required to prevent
-            // race conditions releasing the handle.
-            IntPtr handle = m_handle;
-
             // Free the handle if it hasn't already been freed.
-            if (handle != IntPtr.Zero && Interlocked.CompareExchange(ref m_handle, IntPtr.Zero, handle) == handle)
-            {
+            IntPtr handle = Interlocked.Exchange(ref m_handle, IntPtr.Zero);
+            ValidateHandle(handle);
 #if MDA_SUPPORTED
-                // If this handle was passed out to unmanaged code, we need to remove it
-                // from the cookie table.
-                // NOTE: the entry in the cookie table must be released before the
-                // internal handle is freed to prevent a race with reusing GC handles.
-                if (s_probeIsActive)
-                    s_cookieTable.RemoveHandleIfPresent(handle);
+            // If this handle was passed out to unmanaged code, we need to remove it
+            // from the cookie table.
+            // NOTE: the entry in the cookie table must be released before the
+            // internal handle is freed to prevent a race with reusing GC handles.
+            if (s_probeIsActive)
+                s_cookieTable.RemoveHandleIfPresent(handle);
 #endif
-
-#if BIT64
-                InternalFree((IntPtr)(((long)handle) & ~1L));
-#else // BIT64 (32)
-                InternalFree((IntPtr)(((int)handle) & ~1));
-#endif
-            }
-            else
-            {
-                throw new InvalidOperationException(Environment.GetResourceString("InvalidOperation_HandleIsNotInitialized"));
-            }
+            InternalFree(GetHandleValue(handle));
         }
         
         // Target property - allows getting / updating of the handle's referent.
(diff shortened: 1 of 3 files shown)


src/mscorlib/src/System/Runtime/InteropServices/GcHandle.cs
```diff
@@ -89,70 +88,47 @@ public static GCHandle Alloc(Object value, GCHandleType type)
...
         public void Free()
         {
-            // Copy the handle instance member to a local variable. This is required to prevent
-            // race conditions releasing the handle.
-            IntPtr handle = m_handle;
-
             // Free the handle if it hasn't already been freed.
-            if (handle != IntPtr.Zero && Interlocked.CompareExchange(ref m_handle, IntPtr.Zero, handle) == handle)
-            {
+            IntPtr handle = Interlocked.Exchange(ref m_handle, IntPtr.Zero);
```
benaadams: Doesn't need `CompareExchange` as `m_handle` always ends up as `IntPtr.Zero`; so can just `Exchange` it with `IntPtr.Zero` and the perform the tests on the result

src/mscorlib/src/System/Runtime/InteropServices/GcHandle.cs
```diff
@@ -197,10 +166,16 @@ public static GCHandle FromIntPtr(IntPtr value)
...
                     return new GCHandle(IntPtr.Zero);
                 }
+                return new GCHandle(handle);
             }
 #endif
+            return new GCHandle(value);
+        }
 
-            return new GCHandle(handle);
+        internal static object GetTargetFromIntPtr(IntPtr value)
```
jkotas: Why do we need this?

src/mscorlib/src/System/Runtime/InteropServices/GcHandle.cs
jkotas: Yes, agree.

src/mscorlib/src/System/Runtime/InteropServices/GcHandle.cs
benaadams: AssemblyLoadContext does a set of handle double verifies with the target property: `GCHandle.FromIntPtr(...).Target` as seemed to be a particular area where some of the functions where getting marked as no-inline. So I collapsed the two calls and gave its own internal function.

However, I think the most significant change was `m_handle == IntPtr.Zero` to `m_handle.IsNull()` so I could probably revert this and retest?

benaadams:
Updated summary

benaadams:
`GCHandle:SetIsPinned` is probably a bad example as it still doesn't inline :)

src/mscorlib/src/System/Runtime/InteropServices/GcHandle.cs
jkotas: The AssemblyLoadContext is not performance critical. The double verification is not a problem there.

jkotas:
> SetIsPinned 

This is how you can get fix it:

```
            IntPtr handle = InternalAlloc(value, type);

            // Record if the handle is pinned.
            if (type == GCHandleType.Pinned)
            {
#if BIT64
                handle = new IntPtr(((long)handle) | 1L);
#else // !BIT64 (32)
                handle = new IntPtr(((int)handle) | 1);
#endif
            }

            handle = handle;
```

jkotas:
Or even better - add this at the top:

```
#if BIT64
using nuint = System.UInt64;
#else
using nuint = System.UInt32;
#endif
```

And then you can write it as `handle = (IntPtr)((nuint) | 1)` without ifdefs.

benaadams:
Yep, that seems to do the trick, also removed `GetTargetFromIntPtr`

src/mscorlib/src/System/Runtime/InteropServices/GcHandle.cs
```diff
@@ -7,8 +7,14 @@ namespace System.Runtime.InteropServices
     using System;
     using System.Runtime.CompilerServices;
     using System.Threading;
-    using System.Runtime.Versioning;
     using System.Diagnostics.Contracts;
+#if BIT64
+    using nuint = System.UInt64;
+    using nint = System.Int64;
+#else
+    using nuint = System.UInt32;
```
jkotas: Can we use signed nint consistently everywhere - since the handle is signed IntPtr?

src/mscorlib/src/System/Runtime/InteropServices/GcHandle.cs
```diff
@@ -40,8 +46,13 @@ public enum GCHandleType
     [StructLayout(LayoutKind.Sequential)]
     public struct GCHandle
     {
-        // IMPORTANT: This must be kept in sync with the GCHandleType enum.
-        private const GCHandleType MaxHandleType = GCHandleType.Pinned;
+#if BIT64
+    private const long NoFlag = ~1L;
```
jkotas: Can this be just `~(nint)1` without ifdef ? Also, since it is used in just one place, it may not be worth a constant.

jkotas:
LGTM modulo few nits.

src/mscorlib/src/System/Runtime/InteropServices/GcHandle.cs
benaadams: Changed

src/mscorlib/src/System/Runtime/InteropServices/GcHandle.cs
benaadams: Changed

Issue #9469, opened 2017-02-09
Properly #if CAS changes made under TraceLogging


danmosemsft:
Reverse recent changes and instead put #if NETFX around them.

PR #9478, opened 2017-02-10
Revert earlier CAS changes to Eventing/** and put under !if !CORECLR

danmosemsft:
Hopefully it's correct this time. Fixes https://github.com/dotnet/coreclr/issues/9469

`git diff   f6e1fc~..head` (ie from before the change) shows only #if !CORECLR changes plus the changes made by @benaadams. This will allow us to move the attribute type definitions out of corelib in due course.

@jkotas @brianrob

### DIFF
--- a/src/mscorlib/src/System/Diagnostics/Eventing/XplatEventLogger.cs
+++ b/src/mscorlib/src/System/Diagnostics/Eventing/XplatEventLogger.cs
@@ -23,9 +23,6 @@ namespace System.Diagnostics.Tracing
 
         private static bool initializedPersistentListener = false;
 
-#if !CORECLR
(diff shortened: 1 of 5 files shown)


src/mscorlib/src/System/Diagnostics/Eventing/XplatEventLogger.cs
```diff
@@ -23,6 +23,9 @@ internal  class XplatEventLogger : EventListener
 
         private static bool initializedPersistentListener = false;
 
+#if !CORECLR
```
jkotas: XplatEventLogger is CoreCLR-specific file. The ifdef should not be needed here.

Issue #9484, opened 2017-02-10
[x86/Linux] Exception Handling ABI


wateret:
Currently there is no EH ABI defined for x86/Linux so we need one. I'm going to start with @jkotas suggested in https://github.com/dotnet/coreclr/pull/9384#issuecomment-278543484

Which is:
> There is actually a third variant used by CoreRT on all platforms. It was intentionally designed, taking the best parts from both AMD64 and ARM/ARM64. It would be nice to snap to CoreRT convention for Linux x86 where possible.
> * First argument of all funclets is establisher frame, second argument is exception object (where applicable). Same as AMD64.
> * For filter funclets the VM sets the frame register to be the same as the parent function. For second pass funclets the VM restores all non-volatile registers. Similar to ARM/ARM64.
> * No PSP sym (it may be hard to not have PSP sym for CoreCLR - it is ok to skip it)

wateret:
cc: @parjong @seanshpark

wateret:
@jkotas @parjong
About using Caller-SP or Initial-SP for establisher frame and funclets, currently x86 does not support `FEATURE_FIXED_OUT_ARGS`. Wouldn't it be problem with using Initial-SP?

parjong:
@wateret When ``UNIX_X86_ABI`` is used, the jitter inserts padding to align stack, and thus it is not so trivial to obtain ``Initial-SP`` (Please see #9602 for details).

danmosemsft:
@parjong @wateret with only a few days remaining, would it make sense to set an owner on this issue?

parjong:
@danmosemsft Sorry for late response. I think that we may close this issue. Currently, there are no further actions related with this issue. EH subsystem now works for x86/Linux (although there are several GC-related issues), and https://github.com/dotnet/coreclr/pull/10023 already updated CLR ABI document. 

@wateret  Could you let me know your opinion?

wateret:
@parjong I agree. I think it's good to close this issue.

danmosemsft:
thanks!

PR #9489, opened 2017-02-10
[WIP] [x86/Linux] Call EH Funclet with CallEHFunclet

wateret:
Implement CallEHFunclet same as ARM/ARM64.'

This follows ARM/ARM64 way for calling non-filter funclets which is actually not the way discussed in #9484. This can handle simple `try-catch` statements but fails for nested exceptions. (e.g. throw in catch block)

### DIFF
--- a/cross/x86/toolchain.cmake
+++ b/cross/x86/toolchain.cmake
@@ -7,6 +7,7 @@ set(CMAKE_SYSTEM_PROCESSOR i686)
 add_compile_options("-m32")
 add_compile_options("--sysroot=${CROSS_ROOTFS}")
 add_compile_options("-Wno-error=unused-command-line-argument")
+add_compile_options("-mstack-alignment=4")
 
 set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} --sysroot=${CROSS_ROOTFS}")
 set(CROSS_LINK_FLAGS "${CROSS_LINK_FLAGS} -B ${CROSS_ROOTFS}/usr/lib/gcc/i686-linux-gnu")
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -10284,12 +10284,21 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
 
     ScopedSetVariable<bool> _setGeneratingProlog(&compiler->compGeneratingProlog, true);
 
+    gcInfo.gcResetForBB();
+
     compiler->unwindBegProlog();
 
-    // TODO Save callee-saved registers
+    inst_RV(INS_push, REG_FPBASE, TYP_REF);
+    compiler->unwindPush(REG_FPBASE);
+
+    inst_RV(INS_push, REG_EBX, TYP_REF);
+    inst_RV(INS_push, REG_ESI, TYP_REF);
+    inst_RV(INS_push, REG_EDI, TYP_REF);
 
     // This is the end of the OS-reported prolog for purposes of unwinding
     compiler->unwindEndProlog();
+
+    getEmitter()->emitIns_R_R(INS_mov, EA_PTRSIZE, REG_EXCEPTION_OBJECT, REG_ECX);
 }
 
 /*****************************************************************************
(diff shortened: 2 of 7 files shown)


wateret:
cc: @seanshpark @parjong @YongseopKim

cross/x86/toolchain.cmake
```diff
@@ -7,6 +7,7 @@ set(CMAKE_SYSTEM_PROCESSOR i686)
 add_compile_options("-m32")
 add_compile_options("--sysroot=${CROSS_ROOTFS}")
 add_compile_options("-Wno-error=unused-command-line-argument")
+add_compile_options("-mstack-alignment=4")
```
parjong: This line should be excluded.

src/jit/codegencommon.cpp
```diff
@@ -10284,12 +10284,21 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
...
+
     compiler->unwindBegProlog();
 
-    // TODO Save callee-saved registers
+    inst_RV(INS_push, REG_FPBASE, TYP_REF);
+    compiler->unwindPush(REG_FPBASE);
+
+    inst_RV(INS_push, REG_EBX, TYP_REF);
+    inst_RV(INS_push, REG_ESI, TYP_REF);
+    inst_RV(INS_push, REG_EDI, TYP_REF);
```
parjong: It would be better to use ``genPushCalleeSavedRegisters`` instead of explicitly adding push instructions.

src/jit/codegencommon.cpp
```diff
@@ -10308,7 +10317,11 @@ void CodeGen::genFuncletEpilog()
 
     ScopedSetVariable<bool> _setGeneratingEpilog(&compiler->compGeneratingEpilog, true);
 
-    // TODO Restore callee-saved registers
+    inst_RV(INS_pop, REG_EDI, TYP_REF);
+    inst_RV(INS_pop, REG_ESI, TYP_REF);
+    inst_RV(INS_pop, REG_EBX, TYP_REF);
```
parjong: Similarly before, ``genPopCalleeSavedRegisters`` would be better.

src/jit/codegencommon.cpp
jkotas: Is there a good reason for saving these registers here? The asm helper that calls the funclet will save them, so the funclet does not have to ... unless there is some other reason.

wateret:
About nested catch failure, for the test case like below:

```
try {
  try {
    throw A;
  } catch { // catches A
    throw B;
  }
} catch { // catches B
}
```

When excuting the funclet that catches `A`, the funclet holds `ebp` that is restored by `CallEHFunclet`. Maybe it makes the unwider starts stack-walking from the main function frame, not from the funclet frame. (Actually I'm not sure how exactly unwinder works.)
If this is the case, I guess there would be no problem if it was AMD64 style funclet. That is: there is no `CallEHFunclet` and the `ebp` is saved and restored by the funclet itself.

wateret:
@jkotas @parjong 
I think the problem with nested catch failure is, like I said above, wrong `ebp` value. However now I don't think just switching to AMD64 style funclet would resolve this issue because `ebp` value is still not valid for unwinder.
For arm it seems like `CallEHFunclet` stores frame pointer in `r7` and unwinder uses it as frame pointer. How could we make x86 work? Maybe knowing how AMD64 works would also help.

cross/x86/toolchain.cmake
wateret: OK thanks!

src/jit/codegencommon.cpp
wateret: I'll remove them, thanks.

parjong:
@wateret While digging ``JIT.Methodical.casts.SEH._il_dbgcastclass_catch_neg`` test failure, I conclude that we need to revise CLR's managed unwinder to treat funclet frame as ESP frame.

I made a draft patch in https://github.com/parjong/coreclr/commit/a992b93ec85cf70c54558ff63a3b11ee0809b4db, it seems to work for ``JIT.Methodical.casts.SEH._il_dbgcastclass_catch_neg`` case. (we need more test).

I'm not sure whether that patch could be aligned with your current patch as it does not use PSPSym at all.

Anyway, it would be worthwhile to try.

wateret:
Closing as #9601 includes this.

PR #9490, opened 2017-02-10
NoInlining Indirect for NotifyOfCrossThreadDependencySlow

benaadams:
To avoid the stackcrawlmark interpretation side effect

`Debugger.NotifyOfCrossThreadDependency();` can't be inlined because `Debugger:NotifyOfCrossThreadDependencySlow` is marked as NoInlining; which in mscorlib means it has a StackCrawMark which means its caller also won't be inlined. This bail is then baked into the NGen image.

So currently AsyncTaskMethodBuilder has to call into it, rather than inline the check

### DIFF
--- a/src/mscorlib/src/System/Diagnostics/Debugger.cs
+++ b/src/mscorlib/src/System/Diagnostics/Debugger.cs
@@ -8,12 +8,7 @@
 namespace System.Diagnostics
 {
     using System;
-    using System.IO;
-    using System.Collections;
-    using System.Reflection;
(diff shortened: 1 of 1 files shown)


jkotas:
I am fixing this for good in https://github.com/dotnet/coreclr/pull/9491

stephentoub:
@jkotas, this can be closed then with your change, right?

benaadams:
I hope so 😄

jkotas:
Right

PR #9491, opened 2017-02-10
Use DynamicSecurityMethod attribute to mark methods with StackCrawlMark

### LINKED ISSUE #8102: Enable callers of S.P.CoreLib NoInlining methods to inline.

jkotas:
CoreCLR does not have CAS, and so we can conveniently use it to mark methods with StackCrawlMark to
decouple it from NoInlining. The original purpose of DynamicSecurityMethod was to disable inlining
of the caller and to insert CAS security checks, so we are basically just keeping the first part.

Fixes #8102


benaadams:

stephentoub:
This error message that was fixed yesterday by @AndyAyersMS is now wrong, again, right?
https://github.com/dotnet/coreclr/pull/9470/files#diff-edc46b80f57431489a82311948a8234dR6522

benaadams:
I feel I should ask if the Jit changes will break on desktop; but have no idea about the relationship between the code bases to know if that's a valid question.

jkotas:
The part shared with desktop is https://github.com/dotnet/coreclr/tree/master/src/jit . There are no changes in this part.

jkotas:
> This error message that was fixed yesterday by @AndyAyersMS is now wrong, again, right?

Yes, the original message was more accurate. I will revert it.

AndyAyersMS:
LGTM. Nice to have this finally ironed out.

jamesqo:
Nice! This will be useful in a bunch of places, e.g. `Char` to separate out the uncommon path so the more used one can be inlined.

PR #9493, opened 2017-02-10
Report bytes allocated in the JIT time log.

pgavlin:
This adds a new column, "Total Bytes Allocated", to the JIT time log.
This column reports the total number of bytes requested from the host by
the JIT's arena allocator.

This change also enables `FEATURE_JIT_TIMER` by default (which only
affects JIT32).

### DIFF
--- a/src/jit/compiler.cpp
+++ b/src/jit/compiler.cpp
@@ -7863,7 +7863,7 @@ void JitTimer::PrintCsvMethodStats(Compiler* comp)
 
     comp->m_inlineStrategy->DumpCsvData(fp);
 
-    fprintf(fp, "%I64u,", (unsigned __int64)comp->compGetAllocator()->getTotalBytesAllocated());
+    fprintf(fp, "%Iu,", comp->compGetAllocator()->getTotalBytesAllocated());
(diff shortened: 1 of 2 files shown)


pgavlin:
@AndyAyersMS @dotnet/jit-contrib PTAL

src/jit/compiler.cpp
```diff
@@ -7862,6 +7863,7 @@ void JitTimer::PrintCsvMethodStats(Compiler* comp)
 
     comp->m_inlineStrategy->DumpCsvData(fp);
 
+    fprintf(fp, "%I64u,", (unsigned __int64)comp->compGetAllocator()->getTotalBytesAllocated());
```
AndyAyersMS: Can we use `%zu` here and avoid the cast?

pgavlin: Sadly, no. From [MSDN](https://msdn.microsoft.com/en-us/library/tcxf1dw6.aspx): "The hh, j, z, and t length prefixes are not supported." We can use the VC-specific `Iu`, however, and avoid the cast.

PR #9495, opened 2017-02-10
Add a flag to enable fine-grained optimize control.

pgavlin:
Defining this flag, `OPT_CONFIG`, enables a number of optimizer-related
configuration values. Changing these config values allows for some
interesting experiments without rebuilding the JIT (e.g. disabling the
SSA-based optimizer without forcing minopts, repeating optimizations,
etc.).

This flag is defined by default in debug builds and undefined by default
in release builds.

### DIFF
--- a/src/jit/jitconfigvalues.h
+++ b/src/jit/jitconfigvalues.h
@@ -6,6 +6,10 @@
 #error CONFIG_INTEGER, CONFIG_STRING, and CONFIG_METHODSET must be defined before including this file.
 #endif // !defined(CONFIG_INTEGER) || !defined(CONFIG_STRING) || !defined(CONFIG_METHODSET)
 
+#ifdef DEBUG
+#define OPT_CONFIG // Enable optimization level configuration.
+#endif
+
 #if defined(DEBUG)
 CONFIG_INTEGER(AltJitLimit, W("AltJitLimit"), 0)               // Max number of functions to use altjit for (decimal)
 CONFIG_INTEGER(AltJitSkipOnAssert, W("AltJitSkipOnAssert"), 0) // If AltJit hits an assert, fall back to the fallback
@@ -36,13 +40,6 @@ CONFIG_INTEGER(JitDebugLogLoopCloning, W("JitDebugLogLoopCloning"), 0) // In deb
 CONFIG_INTEGER(JitDefaultFill, W("JitDefaultFill"), 0xff) // In debug builds, initialize the memory allocated by the nra
                                                           // with this byte.
 CONFIG_INTEGER(JitDirectAlloc, W("JitDirectAlloc"), 0)
-CONFIG_INTEGER(JitDoAssertionProp, W("JitDoAssertionProp"), 1) // Perform assertion propagation optimization
-CONFIG_INTEGER(JitDoCopyProp, W("JitDoCopyProp"), 1)   // Perform copy propagation on variables that appear redundant
-CONFIG_INTEGER(JitDoEarlyProp, W("JitDoEarlyProp"), 1) // Perform Early Value Propagataion
-CONFIG_INTEGER(JitDoLoopHoisting, W("JitDoLoopHoisting"), 1)   // Perform loop hoisting on loop invariant values
-CONFIG_INTEGER(JitDoRangeAnalysis, W("JitDoRangeAnalysis"), 1) // Perform range check analysis
-CONFIG_INTEGER(JitDoSsa, W("JitDoSsa"), 1) // Perform Static Single Assignment (SSA) numbering on the variables
-CONFIG_INTEGER(JitDoValueNumber, W("JitDoValueNumber"), 1) // Perform value numbering on method expressions
 CONFIG_INTEGER(JitDoubleAlign, W("JitDoubleAlign"), 1)
 CONFIG_INTEGER(JitDumpASCII, W("JitDumpASCII"), 1)         // Uses only ASCII characters in tree dumps
 CONFIG_INTEGER(JitDumpFgDot, W("JitDumpFgDot"), 0)         // Set to non-zero to emit Dot instead of Xml Flowgraph dump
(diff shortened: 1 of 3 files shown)


pgavlin:
@JosephTremoulet @dotnet/jit-contrib PTAL

src/jit/jitconfigvalues.h
```diff
@@ -236,6 +230,21 @@ CONFIG_INTEGER(JitInlineSIMDMultiplier, W("JitInlineSIMDMultiplier"), 3)
...
+CONFIG_INTEGER(JitDoLoopHoisting, W("JitDoLoopHoisting"), 1)   // Perform loop hoisting on loop invariant values
+CONFIG_INTEGER(JitDoRangeAnalysis, W("JitDoRangeAnalysis"), 1) // Perform range check analysis
+CONFIG_INTEGER(JitDoSsa, W("JitDoSsa"), 1) // Perform Static Single Assignment (SSA) numbering on the variables
+CONFIG_INTEGER(JitDoValueNumber, W("JitDoValueNumber"), 1) // Perform value numbering on method expressions
+
+CONFIG_INTEGER(JitMinOpts, W("JITMinOpts"), 0) // Forces MinOpts
+
+CONFIG_METHODSET(JitOptRepeat, W("JitOptRepeat"))            // Runs optimizer multiple times on the method
+CONFIG_INTEGER(JitOptRepeatCount, W("JitOptRepeatCount"), 2) // Number of times to repeat opts when repeating
+#endif                                                       // defined(OPT_CONFIG)
```
JosephTremoulet: Unfortunate that the formatter wants to align this comment with the previous.  Does adding a blank line preceding make it pull the `endif` comment back to the left?

BruceForstall:
Another confusing thing: `COMPlus_JITMinOpts=1` should already work in retail, because it is processed by the VM (from clrconfigvalues.h) and passed through to the JIT.

pgavlin:
> Another confusing thing: COMPlus_JITMinOpts=1 should already work in retail, because it is processed by the VM (from clrconfigvalues.h) and passed through to the JIT.

Do you mean that the VM passes `CORJIT_FLG_MIN_OPT` when `COMPlus_JitMinOpts` is set? If it doesn't, then setting that variable will have no effect in release builds, since the relevant JIT config value isn't even defined.

pgavlin:
> I find it confusing to have the same flag mean different things in debug vs release; what would you think of putting all that `JitMinOpts()` processing under `OPT_CONFIG`? IMO the various `JitMinOptsFooCounts` up there might as well be available in `OPT_CONFIG` too.

I can do that. IIUC it requires a bit more refactoring due to the logging that is present in the usual processing.

BruceForstall:
> Do you mean that the VM passes CORJIT_FLG_MIN_OPT when COMPlus_JitMinOpts is set?
yes.

pgavlin:
Okay, I see it now--interesting!

In light of that, @JosephTremoulet @BruceForstall: do you think I should still enable the full set of MinOpts knobs under `OPT_CONFIG`, or should I remove the minopts-related changes entirely?

JosephTremoulet:
If I'm reading this right, the vm is setting `CORJIT_FLG_MIN_OPT` in debug as well as release, and the JitMinOpt processing that lets you pick method sets in interesting ways therefore has no effect (the flag will already be true and the various cases only set it to true, never false), is that right?  If so, and you feel like removing some confusing/useless code, go ahead...

Anyway, yeah, if the flag already works one way in release then I'd say keep it working that same way in release, rather than change it now to suddenly start matching debug if they're different.

BruceForstall:
I say leave MinOpts as it is. Don't enable all the crazy binary search stuff in release; just leave it as-is (only for debug/checked).

pgavlin:
> Anyway, yeah, if the flag already works one way in release then I'd say keep it working that same way in release, rather than change it now to suddenly start matching debug if they're different.
> ...
> I say leave MinOpts as it is. Don't enable all the crazy binary search stuff in release; just leave it as-is (only for debug/checked).

Sounds good to me; I'll make the change.

briansull:
_I say leave MinOpts as it is. Don't enable all the crazy binary search stuff in release; just leave it as-is (only for debug/checked)._

I also agree with this

pgavlin:
I have addressed all of the feedback; PTAL

BruceForstall:
:shipit:

src/jit/jitconfigvalues.h
```diff
@@ -236,6 +231,19 @@ CONFIG_INTEGER(JitInlineSIMDMultiplier, W("JitInlineSIMDMultiplier"), 3)
...
+CONFIG_INTEGER(JitDoCopyProp, W("JitDoCopyProp"), 1)   // Perform copy propagation on variables that appear redundant
+CONFIG_INTEGER(JitDoEarlyProp, W("JitDoEarlyProp"), 1) // Perform Early Value Propagataion
+CONFIG_INTEGER(JitDoLoopHoisting, W("JitDoLoopHoisting"), 1)   // Perform loop hoisting on loop invariant values
+CONFIG_INTEGER(JitDoRangeAnalysis, W("JitDoRangeAnalysis"), 1) // Perform range check analysis
+CONFIG_INTEGER(JitDoSsa, W("JitDoSsa"), 1) // Perform Static Single Assignment (SSA) numbering on the variables
+CONFIG_INTEGER(JitDoValueNumber, W("JitDoValueNumber"), 1) // Perform value numbering on method expressions
+
+CONFIG_METHODSET(JitOptRepeat, W("JitOptRepeat"))            // Runs optimizer multiple times on the method
+CONFIG_INTEGER(JitOptRepeatCount, W("JitOptRepeatCount"), 2) // Number of times to repeat opts when repeating
+#endif                                                       // defined(OPT_CONFIG)
```
JosephTremoulet: Still curious if a blank line before the `#endif` would fix this (that comment shouldn't be aligned with the ones above)

PR #9496, opened 2017-02-10
Fix System.Numerics.Vectors CoreFx test failures

CarolEidt:
CoreFx Issue 15713 is due to a case of an Indir(Addr(Field(Vector3 local))) which for some reason has a MorphAddrContext of MACK_Ind. Although I haven't fully identified why that is the case, we should be conservative in this case and mark the address as do-not-enregister.
In addition, when attempting to debug this with a Checked JIT, I encountered an AV due to `GetJitTls` returning null in the `JITDUMP` calls in `getMaxIntrinsicSIMDVectorLength`. Again, I'm not sure why this would be the case but I have added guarding conditions.

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -6163,6 +6163,14 @@ GenTreePtr Compiler::fgMorphField(GenTreePtr tree, MorphAddrContext* mac)
             return newTree;
         }
     }
+    else if ((objRef != nullptr) && (objRef->OperGet() == GT_ADDR) && varTypeIsSIMD(objRef->gtGetOp1()))
+    {
+        GenTreeLclVarCommon* lcl = objRef->IsLocalAddrExpr();
+        if (lcl != nullptr)
+        {
+            lvaSetVarDoNotEnregister(lcl->gtLclNum DEBUGARG(DNER_LocalField));
+        }
+    }
 #endif
 
     /* Is this an instance data member? */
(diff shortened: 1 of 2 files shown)


CarolEidt:
@sivarv @BruceForstall PTAL
@jashook It looks like the arm64 failures are pre-existing
cc @dotnet/jit-contrib

CarolEidt:
This is a fix for https://github.com/dotnet/corefx/issues/15713

sivarv:
:shipit:

src/jit/morph.cpp
```diff
@@ -6163,6 +6163,14 @@ GenTreePtr Compiler::fgMorphField(GenTreePtr tree, MorphAddrContext* mac)
             return newTree;
         }
     }
+    else if ((objRef != nullptr) && (objRef->OperGet() == GT_ADDR) && varTypeIsSIMD(objRef->gtGetOp1()))
```
BruceForstall: I hit cases before where we were getting into MACK_Ind contexts for Vector3 field access unexpectedly, and I fixed those reasons, so I'm worried this is treating the symptom, not the cause. However, if you believe this is a good, conservative fix, then I'm ok with it.

CarolEidt:
I can't reproduce the Windows_NT x64 Release failure, and it's difficult to imagine how these changes could have broken that test. Will retry.

CarolEidt:

src/jit/morph.cpp
CarolEidt: I'm still working on trying to figure out why we wind up in the `MACK_Ind` context. However, given that we do, I believe this is a conservative fix.

Issue #9492, opened 2017-02-10
JIT Assert: Arm64 Pipeline Buildbreak


gkhanna79:
```

2017-02-10T08:28:11.0070683Z BUILD: Generating native image of System.Private.CoreLib for Windows_NT.arm64.Debug
2017-02-10T08:28:11.0070683Z "E:\A\_work\1\s\bin\Product\Windows_NT.arm64.Debug\x64\crossgen.exe" /Platform_Assemblies_Paths "E:\A\_work\1\s\bin\Product\Windows_NT.arm64.Debug" /out "E:\A\_work\1\s\bin\Product\Windows_NT.arm64.Debug\System.Private.CoreLib.ni.dll" "E:\A\_work\1\s\bin\Product\Windows_NT.arm64.Debug\System.Private.CoreLib.dll"
2017-02-10T08:28:32.7263614Z BUILD: Error: CrossGen System.Private.CoreLib build failed. Refer to E:\A\_work\1\s\bin\Logs\CrossgenCoreLib_Windows_NT__arm64__Debug.log
2017-02-10T08:28:32.7263614Z Microsoft (R) CoreCLR Native Image Generator - Version 4.5.22220.0
2017-02-10T08:28:32.7263614Z Copyright (c) Microsoft Corporation.  All rights reserved.
2017-02-10T08:28:32.7263614Z 
2017-02-10T08:28:32.7263614Z Vararg calling convention not supported. while compiling method System.String.Concat
2017-02-10T08:28:32.7263614Z 
2017-02-10T08:28:32.7263614Z Assert failure(PID 6972 [0x00001b3c], Thread: 6520 [0x1978]): Assertion failed 'initBlkNode->gtRsvdRegs == RBM_ARG_2' in 'System.Runtime.CompilerServices.Unsafe:InitBlockUnaligned(byref,ubyte,int)' (IL size 9)
2017-02-10T08:28:32.7263614Z 
2017-02-10T08:28:32.7263614Z     File: e:\a\_work\1\s\src\jit\codegenarm64.cpp Line: 3381
2017-02-10T08:28:32.7263614Z     Image: E:\A\_work\1\s\bin\Product\Windows_NT.arm64.Debug\x64\crossgen.exe
```

gkhanna79:
Broke https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=565212

jashook:
This started failing after #9441. It is an example of #9458.

[...]

PR #9499, opened 2017-02-10
ARM64 Modify genCodeForInitBlk assert

jashook:
Assert in codegenarm64 was incorrect for the GT_STORE_DYN_BLK oper.
If it is that oper then there will not be a temp reg used and instead
the size is computed and stored into REG_2. Add an assert to assert
this is done.

@briansull @CarolEidt ptal
/cc @dotnet/jit-contrib


jashook:
Fixes #9492

briansull:
LGTM

jashook:
Arm64 Testing passed, the rest is unaffected.

PR #9522, opened 2017-02-11
Generic Virtual methods  for CoreRT

sandreenko:
Initial commit to support GVM on jit side.


sandreenko:
Can you take a look please @jkotas @davidwrighton?
 I am confused with our defines.
Every time, when we add something CoreRT specific to the interface we should surround it with 
`#if COR_JIT_EE_VERSION > 460``
when we add code undo IsTargetAbi(CORINFO_CORERT_ABI) we should surround it with 
```
#ifdef FEATURE_READYTORUN_COMPILER
    if (opts.IsReadyToRun())
    {
        if (IsTargetAbi(CORINFO_CORERT_ABI)) {}
    }
#endif
```
Is it correct?

jkotas:
Everytime we add anything new to the JIT/EE interface, it should be surrounded by `#if COR_JIT_EE_VERSION > 460` currently.

The JIT code specific to CoreRT ABI should be surrounded with `IsTargetAbi(CORINFO_CORERT_ABI)`. It may also need to be surrounded by `#if COR_JIT_EE_VERSION > 460` if it using constants or structures that are under the same ifdef in the JIT/EE interface.

The ReadyToRun ifdefs or conditions are not necessary for CoreRT code.

cc @BruceForstall

sandreenko:
Is it correct to surround them with ReadyToRun defines for CoreCLR throughput purposes?

jkotas:
No.

That ReadyToRun defines are always defined - it does not buy you anything with the current set of build configs.

If we wanted to strip CoreRT-specific code from CoreCLR builds, we would create a config where IsTargetAbi(CORINFO_CORERT_ABI) returns false statically, and let the C++ compiler dead code elination.

BruceForstall:
@sandreenko note that if you change the JIT-EE interface, you must also check in the changed interface files to our internal JIT desktop branch. It *may* also invalidate and require new SPMI desktop collection.

sandreenko:
Yes, it will invalidate the collection. We are adding the new field not on the last slot and enum_id for old values will be changed.

Is it necessary to add new CORINFO_CALL_KIND::CORINFO_VIRTUALCALL_LDVIRTFTN_GENERIC_DESCRIPTOR? cc @fadimounir

jkotas:
> We are adding the new field not on the last slot 

As far as I can tell everything you are adding so far is at the end, and so none of the existing value will be changed. It should not invalidate the collection.

sandreenko:
> As far as I can tell everything you are adding so far is at the end, and so none of the existing value will be changed. It should not invalidate the collection.

CORINFO_HELP_GVMLOOKUP_FOR_SLOT is not at the end, is it?
There is no problem to make a new collection.
I will clean this PR on Monday and delete NO-Merge label.
Do we have corert specific label?

jkotas:
> CORINFO_HELP_GVMLOOKUP_FOR_SLOT is not at the end, is it?

It is right before `CORINFO_HELP_COUNT` that is the last one. Changing `CORINFO_HELP_COUNT` should be fine.

sandreenko:
I deleted new call kind and used CORINFO_CALLCONV_GENERIC to distinguish generic and non genetic virtual calls. Please attention that this flag was not used before and maybe we should set it in CoreCLR for consistency. 
@fadimounir [my change for corert](https://github.com/dotnet/corert/commit/c21b23af9da717e4a4c744660d466aaf3e979d07)

sandreenko:
PTAL @dotnet/jit-contrib

src/jit/importer.cpp
```diff
@@ -6772,7 +6771,30 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
                 thisPtr = impCloneExpr(thisPtr, &thisPtrCopy, NO_CLASS_HANDLE, (unsigned)CHECK_SPILL_ALL,
                                        nullptr DEBUGARG("LDVIRTFTN this pointer"));
 
-                GenTreePtr fptr = impImportLdvirtftn(thisPtr, pResolvedToken, callInfo);
+                GenTreePtr fptr = nullptr;
+#if COR_JIT_EE_VERSION > 460
+                if (sig->callConv & CORINFO_CALLCONV_GENERIC && IsTargetAbi(CORINFO_CORERT_ABI))
+                {
```
BruceForstall: By our coding convention, this should be:
```
                if (((sig->callConv & CORINFO_CALLCONV_GENERIC) != 0) && IsTargetAbi(CORINFO_CORERT_ABI))
```
similarly below.

src/jit/importer.cpp
```diff
@@ -6772,7 +6771,30 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
...
+                    assert(IsTargetAbi(CORINFO_CORERT_ABI));
+                    GenTreePtr runtimeMethodHandle = nullptr;
```
BruceForstall: seems a bit overkill to assert something you just checked two lines above.

BruceForstall:
LGTM

BruceForstall:
@russellhadley @sandreenko @jkotas not related to this change, but I see more and more of these checks:
```
if (IsTargetAbi(CORINFO_CORERT_ABI))
```
These checks are dynamic, so each adds a small cost. Theoretically, it's just a global field access and comparison. But are there any build combinations where we can simply `#ifdef` the answer to a static `false`? E.g., CoreRT isn't out there yet, so until it is, can we `#ifdef` the .NET Core build to return `false`? Whatever JIT we ship for .NET Core 2.0 will presumably never be used by CoreRT.

jkotas:
It is trade-off between number of build flavors and tiny performance loss. We should be data driven to device - if this ever becomes anything measurable, it would make sense to have different builds. I do not think we are anywhere near this point.

Using same .dll for both JIT and crossgen has the same problem. There is number of dynamic checks for crossgen and bookkeeping that is crossgen specific. I believe it is much larger hit compared to CoreRT one; yet still pretty small. We used to have different builds in CoreCLR for these, but as you may remember they got unified into one not too long ago.

Or compare this with the hit from dynamic checks in no way asserts. It is multiple orders of magnitude higher hit compared to any of this...

src/jit/importer.cpp

src/jit/importer.cpp
sandreenko: Thanks, fixed

fadimounir:
@sandreenko @BruceForstall Is CORINFO_CALLCONV_GENERIC only used for generic methods, and never for methods on generic types? Can you confirm that?
If so, then it looks good to me. We'll just need to make sure we set it on the CoreRT side (nothing in the code sets this flag today).

src/jit/importer.cpp
```diff
@@ -6792,6 +6813,10 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
                 call->gtCall.gtCallObjp = thisPtrCopy;
                 call->gtFlags |= GTF_EXCEPT | (fptr->gtFlags & GTF_GLOB_EFFECT);
 
+                if (((sig->callConv & CORINFO_CALLCONV_GENERIC) != 0) && IsTargetAbi(CORINFO_CORERT_ABI))
```
MichalStrehovsky: This is the exact same `if` check as above - would it make sense to set a bool flag in the first if check and key off it here? Duplicated if checks are the exact thing that people will forget to update if they ever tweak the first check in the future.

sandreenko:
> @sandreenko @BruceForstall Is CORINFO_CALLCONV_GENERIC only used for generic methods, and never for methods on generic types? Can you confirm that?
> If so, then it looks good to me. We'll just need to make sure we set it on the CoreRT side (nothing in the code sets this flag today).

I set it in my commit for CoreRT. It wasn't used before and I believe there is no knowledge about the old contract. We can create a comment in CoreRT or CoreCLR that this flag from this moment is used only in generic methods.

src/jit/importer.cpp

PR #9524, opened 2017-02-11
Tweak SZGenericArrayEnumerator implementation

jamesqo:
- Remove the `endIndex` field. I believe accessing the Length may be slow for non-generic arrays, but it's just a simple ldlen -> pointer access for generic arrays right?
  - I just tested [with this program](https://gist.github.com/jamesqo/540268a41a54214033b24ee047a2093f) and unfortunately there's no decrease in memory usage on 64-bit, maybe the story is different for 32-bit>

- Do an unsigned cast in Current so there's only 1 branch check

- Use `Array.Empty<T>()` for the empty enumerator

- Format code

- Behavior of MoveNext is undefined after it returns false, so avoid a branch and unconditionally increment the index.
  - Can cause a breaking change if someone calls the method 2B times and then does something based on MoveNext after that?

/cc @jkotas, @benaadams

### DIFF
--- a/src/mscorlib/src/System/Array.cs
+++ b/src/mscorlib/src/System/Array.cs
@@ -2507,7 +2507,9 @@ namespace System {
             private T[] _array;
             private int _index;
 
-            internal static readonly SZGenericArrayEnumerator<T> Empty = new SZGenericArrayEnumerator<T>(Array.Empty<T>());
+            // Array.Empty is intentionally not used here, so we don't have to pay for generic instantiation overhead
+            // if it wouldn't have otherwise been used for that type.
+            internal static readonly SZGenericArrayEnumerator<T> Empty = new SZGenericArrayEnumerator<T>(new T[0]);
 
             internal SZGenericArrayEnumerator(T[] array)
             {
--- a/src/mscorlib/src/System/Array.cs
+++ b/src/mscorlib/src/System/Array.cs
@@ -2516,21 +2516,21 @@ namespace System {
                 _index = -1;
             }
     
-            public bool MoveNext() => ++_index < _array.Length;
+            public bool MoveNext() => _index < _array.Length && ++_index < _array.Length;
     
             public T Current
             {
                 get
                 {
-                    if ((uint)_index >= (uint)_array.Length)
+                    int index = _index;
+                    T[] array = _array;
+
+                    if ((uint)index >= (uint)array.Length)
                     {
-                        var exceptionResource = _index < 0 ?
-                            ExceptionResource.InvalidOperation_EnumNotStarted :
-                            ExceptionResource.InvalidOperation_EnumEnded;
-                        ThrowHelper.ThrowInvalidOperationException(exceptionResource);
+                        ThrowHelper.ThrowInvalidOperationException_EnumCurrent(index);
                     }
 
-                    return _array[_index];
+                    return array[index];
                 }
             }
     
--- a/src/mscorlib/src/System/Array.cs
+++ b/src/mscorlib/src/System/Array.cs
@@ -2516,7 +2516,18 @@ namespace System {
                 _index = -1;
             }
     
-            public bool MoveNext() => _index < _array.Length && ++_index < _array.Length;
+            public bool MoveNext()
+            {
+                int index = _index + 1;
+                bool result = index < _array.Length;
+                
+                if (result)
+                {
+                    _index = index;
+                }
+
+                return result;
+            }
     
             public T Current
             {
--- a/src/mscorlib/src/System/Array.cs
+++ b/src/mscorlib/src/System/Array.cs
@@ -2406,8 +2406,7 @@ namespace System {
             //! Warning: "this" is an array, not an SZArrayHelper. See comments above
             //! or you may introduce a security hole!
             T[] _this = JitHelpers.UnsafeCast<T[]>(this);
-            int length = _this.Length;
-            return length == 0 ? SZGenericArrayEnumerator<T>.Empty : new SZGenericArrayEnumerator<T>(_this, length);
+            return _this.Length == 0 ? SZGenericArrayEnumerator<T>.Empty : new SZGenericArrayEnumerator<T>(_this);
         }
 
         // -----------------------------------------------------------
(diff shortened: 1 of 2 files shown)


src/mscorlib/src/System/Array.cs
```diff
@@ -2585,45 +2584,43 @@ sealed class SZArrayHelper {
...
+                {
+                    if ((uint)_index >= (uint)_array.Length)
+                    {
+                        var exceptionResource = _index < 0 ?
+                            ExceptionResource.InvalidOperation_EnumNotStarted :
+                            ExceptionResource.InvalidOperation_EnumEnded;
+                        ThrowHelper.ThrowInvalidOperationException(exceptionResource);
+                    }
+
                     return _array[_index];
```
jamesqo: As another micro-optimization, I wonder if it would be possible to use the Unsafe APIs to avoid a bounds check here? I just looked in Span, would `return Unsafe.Add(ref JitHelpers.GetArrayData(_array), _index)` be correct?

jkotas: It would not be ok to remove the bounds check here. We want to guarantee memory safety even with races - when somebody uses the enumerator by accident on multiple threads. The boundscheck is important to guarantee the memory safety in this case.

jkotas:
> Behavior of MoveNext is undefined after it returns false
 
It is not what the documentation says https://msdn.microsoft.com/en-us/library/system.collections.ienumerator.movenext(v=vs.110).aspx

When the enumerator is at this position, subsequent calls to MoveNext also return false until Reset is called.

jamesqo:
@jkotas Hm, this is how ImmutableArray's enumerator is implemented. https://github.com/dotnet/corefx/blob/master/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray_1.Enumerator.cs#L65 Although technically, it doesn't implement `IEnumerator<T>`.

src/mscorlib/src/System/Array.cs
jamesqo: @jkotas What if `_index` (and `_array` if necessary) were cached into a local?

jkotas: Doesn't help

jkotas: Oh, I see what you mean - the range check should be eliminated by the JIT. If it is not, the JIT should be fixed to get it eliminated.

benaadams: Still two branches in Current. Ternary operator is emitted as branch; then you are calling `ThrowInvalidOperationException` with var rather than const so might be some additional reg shuffling.

Push the resource choice into the call and pass in index?
```csharp
var index = _index;
var array = _array;
if ((uint)index >= (uint)array.Length)
{
    ThrowHelper.ThrowInvalidOperationException_EnumCurrent(index);
}
return array[index];
```

```csharp
Exception GetInvalidOperationException_EnumCurrent(int index) {
    if (index < 0)
        return GetInvalidOperationException(ExceptionResource.InvalidOperation_EnumNotStarted);
    return GetInvalidOperationException(ExceptionResource.InvalidOperation_EnumEnded);
}

void ThrowInvalidOperationException_EnumCurrent(int index) {
    throw GetInvalidOperationException_EnumCurrent(index);
}
```
You want to go through an exception getter then throw the return rather than do the branch in the Throw as else it has too many blocks and won't be considered for the "does not return" optimisation.

benaadams:
@jkotas reset can throw NotSupportedException https://msdn.microsoft.com/en-us/library/system.collections.ienumerator.reset(v=vs.110).aspx

> The Reset method is provided for COM interoperability. It does not necessarily need to be implemented; instead, the implementer can simply throw a NotSupportedException.

Though is probably breaking change if it did?

jkotas:
Right. There is no good reason to change any observable behavior here.

jamesqo:
@jkotas, @benaadams Thanks both of you for your feedback. I took @benaadams' advice and put the additional branch in `ThrowHelper` so it doesn't bloat the main codepath.

src/mscorlib/src/System/Array.cs
```diff
@@ -2502,45 +2501,43 @@ sealed class SZArrayHelper {
...
             }
     
-            public bool MoveNext() {
-                if (_index < _endIndex) {
-                    _index++;
-                    return (_index < _endIndex);
-                }
-                return false;
-            }
+            public bool MoveNext() => _index < _array.Length && ++_index < _array.Length;
```
stephentoub: Would this work and be more efficient:
```C#
int index = _index + 1;
if (index < _array.Length)
{
    _index = index;
    return true;
}
return false;
```
?

src/mscorlib/src/System/Array.cs
```diff
@@ -2502,45 +2501,43 @@ sealed class SZArrayHelper {
         // This is a normal generic Enumerator for SZ arrays. It doesn't have any of the "this" stuff
         // that SZArrayHelper does.
         //
-        [Serializable] private sealed class SZGenericArrayEnumerator<T> : IEnumerator<T> {
+        [Serializable]
+        private sealed class SZGenericArrayEnumerator<T> : IEnumerator<T>
+        {
             private T[] _array;
             private int _index;
-            private int _endIndex; // cache array length, since it's a little slow.
```
stephentoub: The author is this code seems to have measured and found this to be impactful.  Is that not / no longer the case?

jamesqo: @stephentoub The code was likely copied from here: https://github.com/jamesqo/coreclr/blob/d613bf15ff7b8771125db49841868333f6b4bf5b/src/mscorlib/src/System/Array.cs#L2238 For non-generic arrays `Length` seems to compile into an actual call rather than a pointer access:

```asm
; Assembly listing for method Program:Foo(ref)
; optimized code
; rsp based frame
; partially interruptible
;
;  V00 arg0         [V00,T00] (  5,   4  )     ref  ->  rcx
;  V01 OutArgs      [V01    ] (  1,   1  )  lclBlk (32) [rsp+0x00]
;
; Lcl frame size = 40

G_M17832_IG01:

G_M17832_IG02:
       4885C9               test     rcx, rcx
       740C                 je       SHORT G_M17832_IG03
       E84AC0CB5F           call     Array:get_Length():int:this
       8BC8                 mov      ecx, eax
       E8ABFBFFFF           call     Console:WriteLine(int)

G_M17832_IG03:
       90                   nop

G_M17832_IG04:
       C3                   ret

; Total bytes of code 27, prolog size 4 for method Program:Foo(ref)
; Assembly listing for method Program:Bar(ref)
; optimized code
; rsp based frame
; partially interruptible
;
;  V00 arg0         [V00,T00] (  4,   3.5)     ref  ->  rcx
;  V01 OutArgs      [V01    ] (  1,   1  )  lclBlk (32) [rsp+0x00]
;
; Lcl frame size = 40

G_M17187_IG01:

G_M17187_IG02:
       4885C9               test     rcx, rcx
       7408                 je       SHORT G_M17187_IG03
       8B4908               mov      ecx, dword ptr [rcx+8]
       E87FFBFFFF           call     Console:WriteLine(int)

G_M17187_IG03:
       90                   nop

G_M17187_IG04:
       C3                   ret

; Total bytes of code 23, prolog size 4 for method Program:Bar(ref)

```

src/mscorlib/src/System/Array.cs
jamesqo: Sure

src/mscorlib/src/System/Array.cs
```diff
@@ -2502,45 +2501,54 @@ sealed class SZArrayHelper {
...
 
-            // Passing -1 for endIndex so that MoveNext always returns false without mutating _index
-            internal static readonly SZGenericArrayEnumerator<T> Empty = new SZGenericArrayEnumerator<T>(null, -1);
+            internal static readonly SZGenericArrayEnumerator<T> Empty = new SZGenericArrayEnumerator<T>(Array.Empty<T>());
```
jkotas: It would be likely better to allocate a fresh empty array instance here. `Array.Empty<T>()` is not free... .

stephentoub:
@jamesqo, can you please share your perf measurements?

src/mscorlib/src/System/Array.cs
danmosemsft: When should one use Array.Empty<T>() ?

stephentoub: ```Array.Empty<T>``` incurs generic instantiation cost per T, so if you don't know that the T[] is going to be very common, in a situation like this where you're at most creating one, you're likely better off just allocating the T[0].  Worst case, you end up with at most one more T[0] allocated; best case, you save all of the generic instantiation overhead for a ```Array.Empty<T>``` you won't otherwise need.

justinvp: Here, we could check if `T` is a common type like `byte` or `char` (or `int`?) and use `Array.Empty<T>` in such cases, otherwise allocate an empty `T[0]`. The JIT should elide the type checks. Probably overkill.

jamesqo:
I'm running into this error again when trying to use the coreclr binaries in a console app for benchmarking:

```
Unhandled Exception: System.TypeLoadException: Could not load type 'System.IO.TextWriter' from assembly 'System.IO, Version=4.1.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
   at System.Console.WriteLine(String value)
   at ConsoleApplication.Program.Main()
```

No idea why it happens, but it seems like every other week it will intermittently fail and rebasing/doing a clean build of everything/etc. will have no effect. Then I wait a week or so, retry everything, then it works. So I will wait some time and then try to post perf measurements again.

benaadams:
Had this periodically also; however... saw an answer that may work; though haven't tried it https://github.com/dotnet/corefx/issues/16206#issuecomment-280172165

Try adding a NETStandard.Library package reference?

jamesqo:
> Try adding a NETStandard.Library package reference?

Doesn't help, unfortunately :( I am already including `Microsoft.NETCore.App 1.1.0` in my package list, so that was already being pulled in I believe.

benaadams:
Other thing is to build corefx + tests in release; then copy all dlls from
```
\corefx\bin\testhost\netcoreapp-Windows_NT-Release-x64\shared\Microsoft.NETCore.App\9.9.9
```
Over your standalone published dlls in exe folder; then copy the compiled and crossgened coreclr `System.Private.CoreLib.xxxx` files - but its a bit long winded.

Haven't worked out what the simple way is.

jamesqo:
@benaadams Thanks! Seems to work now after copying the corefx binaries before the coreclr ones. It's not much more tedious either bc I wrote scripts to automatically copy over the binaries.

jamesqo:
Unfortunately, could not find significant perf benefits from the change... closing for now.

jamesqo:
Note: I have not tested with a 32-bit version of CoreCLR, where I'm unsure if removing the field will cause the size of the enumerator to shrink bc of different alignment requirements. If I get time to test that in the future and it results in a decrease in memory usage, I will reopen this PR.

Issue #9067, opened 2017-01-23
Remove redundant defines from native part of CoreCLR


danmosemsft:
From the native code, remove all defines that are always defined (across platforms, architectures, and flavors) both in the build infrastructure and from the code itself. 
Also remove all defines that are never defined again from the build infrastructure and the code itself.

See these two equivalent PR's already complete for the corelib. 
https://github.com/dotnet/coreclr/pull/9044
https://github.com/dotnet/coreclr/pull/9058

For managed code there was a tool to edit the code easily (see https://github.com/dotnet/coreclr/issues/8857). For native code such tools exist such as http://coan2.sourceforge.net/ -- it would be worth trying some before giving up and doing the edits manually.

danmosemsft:
These are always defined in native (ie unconditionally set in clrdefinitions.cmake and not undefined elsewhere in makefile or code)

- [ ] _BLD_CLR
- [ ] DEBUGGING_SUPPORTED
- [ ] FEATURE_APPDOMAIN_RESOURCE_MONITORING
- [ ] FEATURE_ASYNC_IO
- [ ] FEATURE_COLLECTIBLE_TYPES
~~- [ ] FEATURE_CORECLR~~
~~- [ ] FEATURE_CORESYSTEM~~
- [ ] FEATURE_CORRUPTING_EXCEPTIONS
- [ ] FEATURE_EXCEPTION_NOTIFICATIONS
- [ ] FEATURE_EXCEPTIONDISPATCHINFO
- [ ] FEATURE_HOST_ASSEMBLY_RESOLVER
- [ ] FEATURE_ICASTABLE
- [ ] FEATURE_ISYM_READER
- [ ] FEATURE_LEGACYNETCF_DBG_HOST_CONTROL
- [ ] FEATURE_LOADER_OPTIMIZATION
- [ ] FEATURE_MAIN_CLR_MODULE_USES_CORE_NAME
- [ ] FEATURE_MANAGED_ETW_CHANNELS
- [ ] FEATURE_MERGE_CULTURE_SUPPORT_AND_ENGINE
- [ ] FEATURE_MULTICOREJIT
- [ ] FEATURE_NORM_IDNA_ONLY
- [ ] FEATURE_PREJIT
- [ ] FEATURE_RANDOMIZED_STRING_HASHING
- [ ] FEATURE_READYTORUN
- [ ] FEATURE_SPAN_OF_T
- [ ] FEATURE_STANDALONE_SN
[...]

Issue #9045, opened 2017-01-23
Strip corrupted state exceptions handling


jkotas:
We should never handle corrupted state exceptions in .NET Core. Corrupted state exception should always fail fast.

Some of the code to handle corrupted state exceptions seems to be reachable after the NS2.0 work. We should make sure to not ship it.

jkotas:
cc @gkhanna79

gkhanna79:
@rahku PTAL at this issue at the earliest.

rahku:
is the work to disable FEATURE_CORRUPTING_EXCEPTIONS for coreclr build?

jkotas:
We should make sure that the "state corrupting exceptions" are not converted to managed exceptions at all. They should fail fast or remain unhandled.

Stripping the code for `FEATURE_CORRUPTING_EXCEPTIONS` would be nice too, but it is just a nice-to-have cleanup.

rahku:
There are two instances in corelib where HandleProcessCorruptedStateExceptions attribute is used
https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Threading/ExecutionContext.cs#L150 
https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Threading/ExecutionContext.cs#L201

I believe there might be other scenarios where people might use this attribute on desktop like catch exceptions coming from pinvoked native code and perform some logging before exiting. Reading from this [article](https://msdn.microsoft.com/en-us/magazine/dd419661.aspx) it does not seem like the attribute was added for compat reasons. Because for compat we added config switch legacyCorruptedState­­ExceptionsPolicy=true . So in my opinion we should keep the functionality of this attribute in .net core. Let me know your thoughts?

jkotas:
> There are two instances in corelib where HandleProcessCorruptedStateExceptions attribute is 

It is left over from the hardening against corrupted state exceptions that was never done completely. 

I still think we should fail fast whenever process corrupting exception reaches managed code, and not do any processing or handling of them in .NET Core. In particular, any processing of dirty AVs is security liability. It is best to fail fast for them right away.

gkhanna79:
> In particular, any processing of dirty AVs is security liability. It is best to fail fast for them right away.

@jkotas Are you suggesting FX usage of catching such exceptions should issue FailFast or are you suggesting the runtime should just failfast (like we do when there is an AV in the runtime)?

jkotas:
[...]

PR #9525, opened 2017-02-11
Remove redundant defines from CoreCLR native code

danmosemsft:
Towards https://github.com/dotnet/coreclr/issues/9067, removes the following 

- always defined: ~~FEATURE_CORESYSTEM~~, FEATURE_CORECLR, ~~FEATURE_CORRUPTING_EXCEPTIONS~~
- always undefined: FEATURE_APTCA, FEATURE_CAS_POLICY
- a few redundant related to ASSERTE

Also some dead files. Does not modify gc/** nor jit/** code. Does not modify the `clrdefinitions.cmake` because that code still requires the defines.

This was done with commands like
`coan-6.0.1-x86_64.exe source -K -R -DFEATURE_CORECLR -Fcpp,h,rc,inl,hpp,src,txt,idl .`

There are more that can be easily handled this way.

The primary goal is to make the code easier to work on.

@jkotas


jkotas:
Could you please keep `FEATURE_CORRUPTING_EXCEPTIONS` for now? 

We may want to flip it and then strip -  https://github.com/dotnet/coreclr/issues/9045

danmosemsft:
OK kept it.

src/pal/src/locale/utf8.cpp
```diff
@@ -1399,7 +1399,6 @@ class UTF8Encoding
 
         EncodeChar:
 
-#ifdef FASTLOOP
```
jkotas: Why did these files under src\pal changed?

src/inc/switches.h
```diff
@@ -184,15 +169,9 @@
 
 // Windows desktop supports the profiling API attach / detach feature.
```
jkotas: This comment should be deleted too.

src/inc/switches.h
```diff
@@ -184,15 +169,9 @@
...
 // This will eventually be supported on coreclr as well. 
-#if defined(PROFILING_SUPPORTED) && !defined(FEATURE_CORECLR)
-#define FEATURE_PROFAPI_ATTACH_DETACH
-#endif
 
 // Windows desktop DAC builds need to see some of the data used in the profiling API
 // attach / detach feature, particularly Thread::m_dwProfilerEvacuationCounter
```
jkotas: Same here

jkotas:
Could you please also rollback `FEATURE_CORESYSTEM`? I see a bunch of code there that we should probably keep (performance optimizations, etc. - there were not possible because of limited API surface of Windows Phone).

jkotas:
LGTM otherwise. Thanks!

jkotas:
cc @gkhanna79

src/inc/switches.h
danmosemsft: fixed

src/inc/switches.h
danmosemsft: fixed

src/pal/src/locale/utf8.cpp
danmosemsft: It is because it has `#define FASTLOOP` at the top. The tool sometimes applies such defines. I thought I had reverted them all. Some future pass could look at such things, and `#if (0)`, and decide whether to remove any. Maybe they are debugging code.
Thanks for catching.

danmosemsft: It looks like some removes of `#if (0) ` crept in in ilasm/**. I'll leave those removes.

danmosemsft:
@gkhanna79 will you want to review this? No hurry but when it's in I can remove some more on top of it.

src/ilasm/asmman.cpp
```diff
@@ -154,76 +151,6 @@ void    AsmMan::SetModuleName(__inout_opt __nullterminated char* szName)
 }
 //==============================================================================================================
 // Borrowed from VM\assembly.cpp
```
gkhanna79: Should this comment be deleted as well?

danmosemsft:
> extremely painful to integrate with Desktop source (if not impossible) if we were to ever do that.

I'm taking  other folks' word that this will never be a goal. Perhaps we will be all using CoreRT then......

src/ilasm/asmman.cpp
danmosemsft: I'll do that in follwup.

danmosemsft:
Thanks both.

PR #9537, opened 2017-02-11
Fix checks for methods that use StackCrawlMark

jkotas:
My recent changed these methods to be marked using IsMdRequireSecObject instead, but some
places that have to check for it were not updated correctly.

### DIFF
--- a/src/vm/jitinterface.cpp
+++ b/src/vm/jitinterface.cpp
@@ -6732,23 +6732,21 @@ DWORD CEEInfo::getMethodAttribsInternal (CORINFO_METHOD_HANDLE ftn)
         result |= CORINFO_FLG_NOSECURITYWRAP;
     }
 
+    if (IsMdRequireSecObject(attribs))
+    {
+        // Assume all methods marked as DynamicSecurity are
+        // marked that way because they use StackCrawlMark to identify
+        // the caller.
+        // See comments in canInline or canTailCall
+        result |= CORINFO_FLG_DONT_INLINE_CALLER;
+    }
 
     // Check for an inlining directive.
     if (pMD->IsNotInline())
     {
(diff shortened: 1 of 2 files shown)


jkotas:
Fix for failures encountered in https://github.com/dotnet/corefx/pull/16064

cc @stephentoub @gkhanna79

src/vm/jitinterface.cpp
```diff
@@ -6732,23 +6732,21 @@ DWORD CEEInfo::getMethodAttribsInternal (CORINFO_METHOD_HANDLE ftn)
         result |= CORINFO_FLG_NOSECURITYWRAP;
     }
 
+    if (IsMdRequireSecObject(attribs))
+    {
+        // Assume all methods marked as DynamicSecurity are
```
gkhanna79: Can you please also add a similar comment at https://github.com/dotnet/coreclr/pull/9491/files#diff-6291818ea5a59bda1055bcd3e1d3dcddL12 so that it becomes obvious at the definition about its semantics?

jkotas: Done.

danmosemsft:
Briefly what are StackCrawlMarks for if not partial trust scenarios?

jkotas:
Number of APIs look at who called them and change behavior based on that. 

`MethodBase.GetCurrentMethod()` is

jkotas:
Number of APIs look at who called them and change behavior based on that. StackCrawlMarks are part of the puzzle that lets the API find who the caller was - using stackwalking.

The caller is basically passed in an implicit argument. Alternative way to implement this would be codegen magic that recognizes these and APIs; and morphs them into calls into a different internal method that takes explicit caller argument. It is how it is done in CoreRT - it avoids the need for stackwalks.

PR #9539, opened 2017-02-12
Faster List Add

benaadams:
List Add and Clear are warmspots in Kestrel

This is a mild tweak to `Add` however as what's being cleared is a list of `GCHandle` structs `Clear` is a significant win.

@jkotas is this a valid use of `JitHelpers.ContainsReferences<T>()`?

/cc @stephentoub

### DIFF
--- a/src/mscorlib/src/System/Collections/Generic/List.cs
+++ b/src/mscorlib/src/System/Collections/Generic/List.cs
@@ -207,11 +207,31 @@ namespace System.Collections.Generic
         // Adds the given object to the end of this list. The size of the list is
         // increased by one. If required, the capacity of the list is doubled
         // before adding the new element.
-        //
-        public void Add(T item) {
-            if (_size == _items.Length) EnsureCapacity(_size + 1);
-            _items[_size++] = item;
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Add(T item)
+        {
+            var array = _items;
+            var size = _size;
             _version++;
+            if ((uint)size < (uint)array.Length)
+            {
+                _size = size + 1;
+                array[size] = item;
+            }
+            else
+            {
+                AddWithResize(item);
+            }
+        }
+
+        // Non-inline from List.Add to improve its code quality as uncommon path
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private void AddWithResize(T item)
+        {
+            var size = _size;
+            EnsureCapacity(size + 1);
+            _size = size + 1;
+            _items[size] = item;
         }
 
         int System.Collections.IList.Add(Object item)


jkotas:
> @jkotas is this a valid use of `JitHelpers.ContainsReferences<T>()`?

Yes, it is what it is meant for.

stephentoub:
I'd made a similar change to Add in https://github.com/dotnet/coreclr/pull/9323 (the PR shows separating out the EnsureCapacity, but I'd reverted that part locally and just never pushed it up, so it looked basically identically to this PR), as it showed benefits on one machine, but on another it actually showed a slowdown.  We should just make sure it's consistently better before changing it.

src/mscorlib/src/System/Collections/Generic/List.cs
```diff
@@ -12,14 +12,16 @@
 **
 ** 
 ===========================================================*/
-namespace System.Collections.Generic {
 
+using System.Reflection;
```
stephentoub: The usings should all be together.

src/mscorlib/src/System/Collections/Generic/List.cs
```diff
@@ -208,9 +210,15 @@ public class List<T> : IList<T>, System.Collections.IList, IReadOnlyList<T>
         // increased by one. If required, the capacity of the list is doubled
         // before adding the new element.
         //
-        public void Add(T item) {
-            if (_size == _items.Length) EnsureCapacity(_size + 1);
-            _items[_size++] = item;
+        public void Add(T item)
+        {
+            var size = _size;
```
stephentoub: Nit: var => int

src/mscorlib/src/System/Collections/Generic/List.cs
```diff
@@ -291,10 +299,15 @@ public int BinarySearch(T item, IComparer<T> comparer)
 
     
         // Clears the contents of List.
-        public void Clear() {
+        public void Clear()
+        {
             if (_size > 0)
             {
-                Array.Clear(_items, 0, _size); // Don't need to doc this but we clear the elements so that the gc can reclaim the references.
+                if (JitHelpers.ContainsReferences<T>())
```
stephentoub: Assuming the call completely evaporates, should we use this in List (and maybe elsewhere in other data structures) anywhere we null out a field, e.g. not just Clear but all of the Remove methods?

src/mscorlib/src/System/Collections/Generic/List.cs
```diff
@@ -208,9 +210,15 @@ public class List<T> : IList<T>, System.Collections.IList, IReadOnlyList<T>
...
         //
-        public void Add(T item) {
-            if (_size == _items.Length) EnsureCapacity(_size + 1);
-            _items[_size++] = item;
+        public void Add(T item)
+        {
+            var size = _size;
+            if (size == _items.Length)
+            {
+                EnsureCapacity(size + 1);
```
benaadams: Tried something like @stephentoub's https://github.com/dotnet/coreclr/pull/9323 where the addition was also factored out to try and get `Add` to be inlinable as its not; but the simple function was `[below ALWAYS_INLINE size]` and seemed to ignore the `[MethodImpl(MethodImplAttributes.NoInlining)]` attribute.

Not sure if is expected or a  change? /cc @AndyAyersMS

benaadams: Also had to test in before after in coreclr due to the stackcrawlmark stuff

src/mscorlib/src/System/Collections/Generic/List.cs
```diff
@@ -390,15 +402,38 @@ bool System.Collections.IList.Contains(Object item)
...
-   
+
+        [MethodImpl(MethodImplAttributes.NoInlining)]
+        private void IncreaseCapacity()
+        {
+            int length = _items.Length;
+            if (_size == Array.MaxArrayLength)
+            {
+                // New length would be too large for array
+                ThrowHelper.ThrowLengthArgumentOutOfRange_ArgumentOutOfRange_NeedNonNegNum();
```
jkotas: Why do we need this exception when it was not there before? It seems to be changing the behavior for this case - was OutOfMemoryException thrown before in this case?

src/mscorlib/src/System/Collections/Generic/List.cs
jkotas: Yes, the call should completely evaporate. It should be good idea to use - in particular before bulk clearing (Array.Clear).

src/mscorlib/src/System/Collections/Generic/List.cs
benaadams: Couldn't find docs saying any exception could be thrown; it would call `set_Capacity` with a negative number, which would call the array .ctor with negative number ` new T[value]`. 

Should throw OutOfMemoryExecption instead?

benaadams: Actually... just need to remove the casts from `Math.Min`

src/mscorlib/src/System/Collections/Generic/List.cs
AndyAyersMS: Noinlining should always be honored, it's one of the first things the jit looks for.

benaadams:
Going to submit separate PR for Clear and Remove as they are quite simple. Add ends up with a lot going on in the asm so will see if can take care of it seperately

src/mscorlib/src/System/Collections/Generic/List.cs
```diff
@@ -390,15 +402,38 @@ bool System.Collections.IList.Contains(Object item)
...
                 // Allow the list to grow to maximum possible capacity (~2G elements) before encountering overflow.
                 // Note that this check works even when _items.Length overflowed thanks to the (uint) cast
                 if ((uint)newCapacity > Array.MaxArrayLength) newCapacity = Array.MaxArrayLength;
                 if (newCapacity < min) newCapacity = min;
                 Capacity = newCapacity;
             }
         }
-   
+
+        [MethodImpl(MethodImplAttributes.NoInlining)]
```
danmosemsft: maybe a comment? // Separated out of List.Add to improve its code quality

jamesqo:
@benaadams Awesome, nice work!

benaadams:
Clear+Remove were picked up and merged in https://github.com/dotnet/coreclr/pull/9540 so this will just become Add

src/mscorlib/src/System/Collections/Generic/List.cs
```diff
@@ -208,9 +209,15 @@ public class List<T> : IList<T>, System.Collections.IList, IReadOnlyList<T>
...
-        public void Add(T item) {
-            if (_size == _items.Length) EnsureCapacity(_size + 1);
-            _items[_size++] = item;
+        public void Add(T item)
+        {
+            int size = _size;
+            if (size == _items.Length)
+            {
+                IncreaseCapacity();
+            }
```
jamesqo: Maybe avoid a field access in the common case by doing:

```cs
T[] items = _items;
if (size == items.Length)
{
    IncreaseCapacity();
    items = _items;
}
_size = size + 1;
items[size] = item;
```

?

benaadams:
Willc lose and reopen different PR for Add

benaadams:
Doesn't hugely change the size of the asm; but doesn't get it perminantely branded as no inline.

Pre
```asm
Marking List`1:Add(long):this as NOINLINE because of unprofitable inline
**************** Inline Tree
Inlines into 060034D4 List`1:Add(long):this
  [0 IL=0025 TR=000050 060034E3] [FAILED: noinline per IL/cached result] List`1:EnsureCapacity(int):this
Budget: initialTime=282, finalTime=282, initialBudget=2820, currentBudget=2820
Budget: initialSize=1818, finalSize=1818
; Assembly listing for method List`1:Add(long):this
; optimized code
; rsp based frame
; partially interruptible
;
;  V00 this         [V00,T00] ( 20,  18  )     ref  ->  rsi         this
;  V01 arg1         [V01,T03] (  3,   3  )    long  ->  rdi        
;  V02 loc0         [V02,T02] (  6,   6  )     int  ->  rdx        
;  V03 tmp0         [V03,T01] (  4,   8  )     ref  ->  rax        
;
; Lcl frame size = 40

G_M46198_IG01:
       57                   push     rdi
       488BFA               mov      rdi, rdx

G_M46198_IG02:
       8B5618               mov      edx, dword ptr [rsi+24]
       488B4E08             mov      rcx, gword ptr [rsi+8]
       3B5108               cmp      edx, dword ptr [rcx+8]
       750D                 jne      SHORT G_M46198_IG03
       8B5618               mov      edx, dword ptr [rsi+24]
       FFC2                 inc      edx
       488BCE               mov      rcx, rsi
       E800000000           call     List`1:EnsureCapacity(int):this

G_M46198_IG03:
       488B4608             mov      rax, gword ptr [rsi+8]
       8B5618               mov      edx, dword ptr [rsi+24]
       8D4A01               lea      ecx, [rdx+1]
       894E18               mov      dword ptr [rsi+24], ecx
       3B5008               cmp      edx, dword ptr [rax+8]
       7312                 jae      SHORT G_M46198_IG05
       4863D2               movsxd   rdx, edx
       48897CD010           mov      qword ptr [rax+8*rdx+16], rdi
       FF461C               inc      dword ptr [rsi+28]

G_M46198_IG04:
       5F                   pop      rdi
       C3                   ret      

G_M46198_IG05:
       E800000000           call     CORINFO_HELP_RNGCHKFAIL
       CC                   int3     

; Total bytes of code 79, prolog size 6 for method List`1:Add(long):this
```

Post
```asm
Successfully inlined List`1:AddWithoutResize(long):this (44 IL bytes) (depth 1) [profitable inline]
**************** Inline Tree
Inlines into 060034D4 List`1:Add(long):this
  [1 IL=0018 TR=000021 060034D5] [profitable inline] List`1:AddWithoutResize(long):this
  [0 IL=0026 TR=000013 060034D6] [FAILED: noinline per IL/cached result] List`1:AddWithResize(long):this
Budget: initialTime=156, finalTime=230, initialBudget=1560, currentBudget=1560
Budget: initialSize=860, finalSize=1082
; Assembly listing for method List`1:Add(long):this
; optimized code
; rsp based frame
; fully interruptible
;
;  V00 this         [V00,T00] ( 20,  14  )     ref  ->  rcx         this
;  V01 arg1         [V01,T03] (  6,   4  )    long  ->  rdx        
;  V02 tmp0         [V02,T04] ( 10,   5  )     int  ->  rax        
;  V03 tmp1         [V03,T02] (  6,   6  )     ref  ->  rcx        
;  V05 cse0         [V05,T01] (  9,   7.5)     int  ->   r8        
;
; Lcl frame size = 40

G_M46199_IG01:
       90                   nop      

G_M46199_IG02:
       8B4118               mov      eax, dword ptr [rcx+24]
       4C8B4108             mov      r8, gword ptr [rcx+8]
       458B4008             mov      r8d, dword ptr [r8+8]
       413BC0               cmp      eax, r8d
       7D24                 jge      SHORT G_M46199_IG04
       8B4118               mov      eax, dword ptr [rcx+24]
       448D4801             lea      r9d, [rax+1]
       44894918             mov      dword ptr [rcx+24], r9d
       413BC0               cmp      eax, r8d
       731B                 jae      SHORT G_M46199_IG06
       4863C0               movsxd   rax, eax
       488954C110           mov      qword ptr [rcx+8*rax+16], rdx

G_M46199_IG03:
       C3                   ret      

G_M46199_IG04:

G_M46199_IG05:

G_M46199_IG06:
       E800000000           call     CORINFO_HELP_RNGCHKFAIL
       CC                   int3     

; Total bytes of code 77, prolog size 5 for method List`1:Add(long):this
```

PTAL @stephentoub @jkotas

benaadams:
Can't eliminate the range check so there is a double check on length https://github.com/dotnet/coreclr/issues/9707

src/mscorlib/src/System/Collections/Generic/List.cs
benaadams: Added

src/mscorlib/src/System/Collections/Generic/List.cs
```diff
@@ -208,10 +208,36 @@ public class List<T> : IList<T>, System.Collections.IList, IReadOnlyList<T>
...
+                AddWithoutResize(item);
+            }
+            else
+            {
+                AddWithResize(item);
+            }
+        }
+
+        // Separated out of List.Add to improve inlinability of both functions
+        private void AddWithoutResize(T item)
```
jkotas: Is this different from just marking the public method as ForceInline?

benaadams:
Think I broke it :P

src/mscorlib/src/System/Collections/Generic/List.cs
benaadams: Jit still chooses based on context? Whereas aggressive inline considers context less?

jkotas: @AndyAyersMS Would you prefer MethodImplOptions.AggressiveInlining here; or splitting the method into two to trick the JIT to inline it more often?

AndyAyersMS: Would prefer `AggressiveInlining` since it makes the intent clear.

benaadams: K, can make the asm better by keeping them together

benaadams:
Updated to aggressive inlined method; asm reduced and cleaner though still has both
```
cmp      dword ptr [rax+8], r8d
```
and
```
```

src/mscorlib/src/System/Collections/Generic/List.cs
benaadams: Updated the asm from keeping them in same method with aggressive inline; is reduced by 10 bytes; with potential for the range check to be eliminated.

benaadams:
Second check now elided by https://github.com/dotnet/coreclr/pull/9773

jamesqo:
@benaadams Great, then other collections can take advantage of this then.

jamesqo:
Opened dotnet/corefx#17318

omariom:
@benaadams 
If you change the order..
```C#
array[size] = item;
_size = size + 1;
```
it may reuse r8d.
Like 
```ASM
inc      r8d
mov      dword ptr [rcx+24], r8d
```
A couple of bytes less :) 
Though not sure about perf.

benaadams:
@omariom for struct containing refs and classes `array[size] = item;` becomes a memory barrier assign, so was approaching it as getting the int assign in the cpu pipeline prior to the memory barrier. Might not make much difference tho

PR #9540, opened 2017-02-12
List Clear+Remove

benaadams:
Struct based Clear trims to
```asm
Marking List`1:Clear():this as NOINLINE because of unprofitable inline
Successfully inlined JitHelpers:ContainsReferences():bool (2 IL bytes) (depth 1) [below ALWAYS_INLINE size]
**************** Inline Tree
Inlines into 060034D8 List`1:Clear():this
  [1 IL=0000 TR=000001 06003B02] [below ALWAYS_INLINE size] JitHelpers:ContainsReferences():bool
  [0 IL=0029 TR=000037 06000248] [FAILED: noinline per IL/cached result] Array:Clear(ref,int,int)
Budget: initialTime=228, finalTime=218, initialBudget=2280, currentBudget=2280
Budget: initialSize=1408, finalSize=1408
; Assembly listing for method List`1:Clear():this
; optimized code
; rsp based frame
; partially interruptible
;
;  V00 this         [V00,T00] (  7,   7  )     ref  ->  rcx         this
;  V01 OutArgs      [V01    ] (  1,   1  )  lclBlk (32) [rsp+0x00]  
;
; Lcl frame size = 40

G_M33881_IG01:
       90                   nop      

G_M33881_IG02:
       894118               mov      dword ptr [rcx+24], eax

G_M33881_IG03:
       C3                   ret      

; Total bytes of code 18, prolog size 5 for method List`1:Clear():this
```
Shame about the no-inline...

For remove I was looking to swap `Array.Copy` for `Buffer.Memmove` or similar, however the Span code for getting a pointer to a `T[]` element in an array is a little exotic

### DIFF
--- a/src/mscorlib/src/System/Collections/Generic/List.cs
+++ b/src/mscorlib/src/System/Collections/Generic/List.cs
@@ -292,13 +292,21 @@ namespace System.Collections.Generic
         // Clears the contents of List.
         public void Clear()
         {
-            if (JitHelpers.ContainsReferences<T>() && _size > 0)
+            if (JitHelpers.ContainsReferences<T>())
             {
-                Array.Clear(_items, 0, _size); // Clear the elements so that the gc can reclaim the references.
+                int size = _size;
+                _size = 0;
+                _version++;
+                if (size > 0)
+                {
+                    Array.Clear(_items, 0, size); // Clear the elements so that the gc can reclaim the references.
+                }
             }
-
-            _size = 0;
-            _version++;
+            else
+            {
+                _size = 0;
+                _version++;
+            }
         }
     
         // Contains returns true if the specified element is in the List.
--- a/src/mscorlib/src/System/Collections/Generic/List.cs
+++ b/src/mscorlib/src/System/Collections/Generic/List.cs
@@ -861,9 +861,13 @@ namespace System.Collections.Generic
                     // copy item to the free slot.
                     _items[freeIndex++] = _items[current++];
                 }
-            }                       
-            
-            Array.Clear(_items, freeIndex, _size - freeIndex);
+            }
+
+            if (JitHelpers.ContainsReferences<T>())
+            {
+                Array.Clear(_items, freeIndex, _size - freeIndex); // Clear the elements so that the gc can reclaim the references.
+            }
+
             int result = _size - freeIndex;
             _size = freeIndex;
             _version++;
@@ -911,8 +915,12 @@ namespace System.Collections.Generic
                 if (index < _size) {
                     Array.Copy(_items, index + count, _items, index, _size - index);
                 }
-                Array.Clear(_items, _size, count);
+
                 _version++;
+                if (JitHelpers.ContainsReferences<T>())
+                {
+                    Array.Clear(_items, _size, count);
+                }
             }
         }
     
--- a/src/mscorlib/src/System/Collections/Generic/List.cs
+++ b/src/mscorlib/src/System/Collections/Generic/List.cs
@@ -12,14 +12,13 @@
 **
 ** 
 ===========================================================*/
-namespace System.Collections.Generic {
-
+namespace System.Collections.Generic
+{
     using System;
-    using System.Runtime;
-    using System.Runtime.Versioning;
     using System.Diagnostics;
     using System.Diagnostics.Contracts;
     using System.Collections.ObjectModel;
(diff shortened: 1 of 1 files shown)


src/mscorlib/src/System/Collections/Generic/List.cs
```diff
@@ -291,12 +290,14 @@ public int BinarySearch(T item, IComparer<T> comparer)
 
     
         // Clears the contents of List.
-        public void Clear() {
-            if (_size > 0)
+        public void Clear()
+        {
+            if (JitHelpers.ContainsReferences<T>() && _size > 0)
```
jkotas: There are other places in this file that call Array.Clear. We should fix them all.

jkotas:
>For remove I was looking to swap Array.Copy for Buffer.Memmove or similar, however the Span code for getting a pointer to a T[] element in an array is a little exotic

Right. Array.Copy is just fine for these.

benaadams:
Tail calls into Array.Copy for reference based Clears
```asm
Marking List`1:Clear():this as NOINLINE because of unprofitable inline
Successfully inlined JitHelpers:ContainsReferences():bool (2 IL bytes) (depth 1) [below ALWAYS_INLINE size]
**************** Inline Tree
Inlines into 060034D8 List`1:Clear():this
  [1 IL=0000 TR=000001 06003B02] [below ALWAYS_INLINE size] JitHelpers:ContainsReferences():bool
  [0 IL=0047 TR=000087 06000248] [FAILED: noinline per IL/cached result] Array:Clear(ref,int,int)
Budget: initialTime=285, finalTime=275, initialBudget=2850, currentBudget=2850
Budget: initialSize=1841, finalSize=1841
; Assembly listing for method List`1:Clear():this
; optimized code
; rsp based frame
; fully interruptible
;
;  V00 this         [V00,T00] ( 13,  12  )     ref  ->  rsi         this
;  V01 loc0         [V01,T01] (  5,   4  )     int  ->   r8        
;* V02 tmp0         [V02,T02] (  0,   0  )    long  ->  zero-ref   
;* V03 tmp1         [V03,T03] (  0,   0  )    long  ->  zero-ref   
;
; Lcl frame size = 48

G_M58440_IG01:
       4883EC30             sub      rsp, 48
       48894C2428           mov      qword ptr [rsp+28H], rcx

G_M58440_IG02:
       488B0E               mov      rcx, qword ptr [rsi]
       448B4618             mov      r8d, dword ptr [rsi+24]
       33C9                 xor      ecx, ecx
       894E18               mov      dword ptr [rsi+24], ecx
       FF461C               inc      dword ptr [rsi+28]
       4585C0               test     r8d, r8d
       7E15                 jle      SHORT G_M58440_IG04
       488B4E08             mov      rcx, gword ptr [rsi+8]
       33D2                 xor      edx, edx

G_M58440_IG03:
       4883C430             add      rsp, 48

G_M58440_IG04:
       4883C430             add      rsp, 48
       C3                   ret      

; Total bytes of code 60, prolog size 13 for method List`1:Clear():this
```

benaadams:
@stephentoub @jkotas should arraypool clear using this test when `clearArray == false` as it doesn't have a `: struct` constraint? i.e. to free the GC refs?

stephentoub:
> should arraypool clear anyway using this test when clearArray == false as it doesn't have a : struct constraint? i.e. to free the GC refs?

You mean, if T is or contains refs, clear the array even if it wasn't requested, so as to avoid keeping objects alive? On the one hand, I can see the benefit in some cases; however I know there are cases (because I've done it) where code explicitly clears the used portion itself and then specifies clearArray==false, in which case the change would add cost for no benefit.

benaadams:
> where code explicitly clears the used portion itself

Ah, makes sense

benaadams:
@jkotas added the other Array.Clears

The remaining ones in coreclr are ArrayPool; definite reference types (ArrayList); definite value types (Streams) so assume is conscious decision.

There is also Dictionary; but that has an array of `Entry`s and checks hashcode for zero; so would need to split to multi-arrays to benefit; but that has a high chance of being determental to perf elsewhere.

Looking for `= deafult(T)`

benaadams:
Remaining `= deafult(T)` clears are in `ConcurrentQueue`, `ProducerConsumerQueues` and `ThreadLocal` so I'll stay away from them 😄

omariom:
👍 👍 👍

PR #9548, opened 2017-02-13
SpinWait - always spin once

benaadams:
When `m_count == 0` `SpinOnce()` calls through to [`ThreadNative::SpinWait`](https://github.com/dotnet/coreclr/blob/master/src/vm/comsynchronizable.cpp#L1886-L1915) but doesn't actually issue a pause as iterations are zero.

/cc @stephentoub PTAL

### DIFF
--- a/src/mscorlib/src/System/Threading/SpinWait.cs
+++ b/src/mscorlib/src/System/Threading/SpinWait.cs
@@ -146,6 +146,11 @@ namespace System.Threading
                     Thread.Yield();
                 }
             }
+            else if (m_count == 0)
+            {
+                // Not spun yet, single cpu PAUSE, bitshift below would result in 0 PAUSEs
+                Thread.SpinWait(1);
+            }
             else
             {
                 //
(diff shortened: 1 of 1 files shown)


benaadams:
Not sure about the test in the `else` of `m_count == int.MaxValue`

benaadams:
Instant retry (as now) might be desirable rather than a single pause? If so does it need to call into vm for that?

stephentoub:
What effect does this have on various benchmarks? It's adding a yield where there wasn't one before... could be good, could be bad. I expect it's the right thing to do, but...

benaadams:
Will try, also with adding a single pause e.g. 0, 1, 16...

benaadams:
Ahh... read the bit shift wrong. Spin sequence is PAUSE x 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, Yields

Was wondering if it should start around 4 as `LOCK CMPXCHG8B` is 19 cycles and `PAUSE` is 5 cycles on broadwell... 😁

benaadams:
Basically was reading code too late at night 😆

benaadams:
Increased some of the timeouts on spinwait for higher numbers of cores and could get a x2-3 improvement on the global threadpool - so it is pure raw contention at play. 

Couldn't work out a sensible general improvement though; and it shouldn't exhibit much where actual work is being done (where increasing the spinning times would likely reduce overall work done anyway).

Issue #8684, opened 2016-12-20
[ARM32/Linux] libcoreclrtraceptprovider.so build 


hseok-oh:
If we enable building `libcoreclrtraceptprovider.so` by setting `FEATURE_EVENT_TRACE` to 1 (https://github.com/dotnet/coreclr/pull/8421#discussion_r91406080), the build fail with these error messages

```
/nfs/rbp3/coreclr/cross/rootfs/arm/usr/include/urcu/uatomic/generic.h:75:10: error: address argument to atomic builtin must be a pointer to integer or pointer ('void *' invalid)
                return __sync_val_compare_and_swap_4(addr, old, _new);
                       ^                             ~~~~
/nfs/rbp3/coreclr/cross/rootfs/arm/usr/include/urcu/uatomic/generic.h:112:3: error: address argument to atomic builtin must be a pointer to integer or pointer ('void *' invalid)
                __sync_and_and_fetch_4(addr, val);
                ^                      ~~~~
/nfs/rbp3/coreclr/cross/rootfs/arm/usr/include/urcu/uatomic/generic.h:151:3: error: address argument to atomic builtin must be a pointer to integer or pointer ('void *' invalid)
                __sync_or_and_fetch_4(addr, val);
                ^                     ~~~~
/nfs/rbp3/coreclr/cross/rootfs/arm/usr/include/urcu/uatomic/generic.h:190:10: error: address argument to atomic builtin must be a pointer to integer or pointer ('void *' invalid)
                return __sync_add_and_fetch_4(addr, val);
                       ^                      ~~~~
```

We can reproduce this error message by using clang version >= 3.6 and ubuntu 14.04. In other words, we can avoid this fail by using clang < 3.5 or upper version of ubuntu 14.04. (ex. 16.04). But our default build environment for ARM32/Linux is clang-3.6 and ubuntu 14.04.

janvorli:
@hseok-oh if I remember it correctly, there were some problems with pre-3.5 clang, that's why the 3.5 is the minimal supported version.

hseok-oh:
@gkhanna79 It seems better to using patch file in `build-rootfs.sh` script for ubuntu 14.04 rootfs such that tizen does. The script for Tizen rootfs uses https://github.com/dotnet/coreclr/blob/master/cross/armel/tizen/tizen.patch in https://github.com/dotnet/coreclr/blob/master/cross/armel/tizen-build-rootfs.sh.

gkhanna79:
[...]

Issue #8549, opened 2016-12-09
[ARM32/Linux] CoreCLR building issues for ARM32 on Linux


hseok-oh:
We have some issues for ARM 32bit cross building on Linux. Especially, we need to clear some of these tasks to enable core-setup for ARM32. (dotnet/core-setup#725)

Packages build
- [x] Update CoreCLR to generate nuget packages of CoreCLR (PR: #8421)
- [x] Enable CoreCLR Pipeline build support for Linux for Arm32 and publish packages (issue https://github.com/dotnet/coreclr/issues/8639)
- [x] Build `libcoreclrtraceptprovider.so` (issue #8684)
- [ ] Generate *.ni.dll files using x86 hosted crossgen (issue https://github.com/dotnet/coreclr/issues/8647)
- [ ] pass unit test

Fix build script for cross build
- [x] Fix build script for cross-architecture component (Issue: #8495 PR: #8646)

CI support
- [x] Update CI to use Docker images for ARM32 CI (issue: #9273,  #9445 PR: #10185)

hseok-oh:
cc @chunseoklee @jyoungyun @idkiller 
@gkhanna79 @janvorli

gkhanna79:
How does this relate, tracking wise, to https://github.com/dotnet/core-setup/issues/725?

hseok-oh:
@gkhanna79 I also think not all of tasks in this issue are related to dotnet/core-setup#725. 

And we may generate package file for Linux ARM32 by task 1 (Update CoreCLR to generate nuget packages of CoreCLR). But we need additional tasks to generate more complete package. (such as `libcoreclrtraceptprovider.so` , `crossgen` and `*.ni.dll`)

stevedesmond-ca:
I had previously noticed that libcoreclrtraceptprovider.so is explicitly ignored/removed from the things to build for non-Windows ARM...was this based on a previous assumption/requirement that is no longer true? I'm happy to pick that one up if no one else has.

As for the crossgen pieces, [here](https://github.com/stevedesmond-ca/dotnet-arm/blob/master/build) is how I was doing it, quick and dirty, though that requires sudo'ing, which is probably not the best way.

gkhanna79:
[...]

PR #9549, opened 2017-02-13
[ARM32/Linux] build libcoreclrtraceptprovider.so

hseok-oh:
build libcoreclrtraceptprovider.so by adding fix process of rootfs for ubuntu 14.04 armhf
include libcoreclrtraceptprovider.so in nuget package


hseok-oh:
related issue: #8684, #8549

hseok-oh:
It should be merged after update clang version over 3.6 and fix rootfs generation in CI machine.

PR #7923, opened 2016-11-01
Jit: spill stack for non-candidate calls

AndyAyersMS:
Closes #7907.

Spill the evaluation stack at non-candidate calls since the call may
modify locals or args that are currently live on the stack.

In the attached test case I can't get the problematic IL sequence to
happen for the `F0` variant, but it is there for the `F1` variant.

No diffs seen in corlib or any of the framework assemblies that are
run via jit-dasm.

PR #9560, opened 2017-02-13
Value number TypeHandleToRuntimeType helper

### LINKED ISSUE #9552: Branches guarded by type checks not being trimmed in generic types/methods
Consider this code:
```C#
using System;

public class Test
{
    public static void Main() => Bar<bool>.Foo();
}

internal class Bar<T>
{
    public static void Foo()
    {
        if (typeof(T) == typeof(bool))
        {
            Console.WriteLine("bool");
        }
        else if (typeof(T) == typeof(int))
        {
            Console.WriteLine("int");
        }
    }
}
```
Since this code should be specialized per value type, I was expecting Foo to compile down to effectively "Console.WriteLine("bool")", but instead I get this:
```
; Assembly listing for method Bar`1[Boolean][System.Boolean]:Foo()
; optimized code
; rsp based frame
; fully interruptible
;
;  V00 tmp0         [V00,T00] (  2,   4  )     ref  ->  rsi
;  V01 tmp1         [V01,T01] (  2,   2  )     ref  ->  rsi
;  V02 OutArgs      [V02    ] (  1,   1  )  lclBlk (32) [rsp+0x00]
;
; Lcl frame size = 32

G_M20354_IG01:

G_M20354_IG02:
       48B9081BE1B9FB7F0000 mov      rcx, 0x7FFBB9E11B08
       E850BBF65E           call     CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPE
       488BF0               mov      rsi, rax
       48B9081BE1B9FB7F0000 mov      rcx, 0x7FFBB9E11B08
       E83EBBF65E           call     CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPE
       483BC6               cmp      rax, rsi
       751F                 jne      SHORT G_M20354_IG04
       48B9205728EFA5020000 mov      rcx, 0x2A5EF285720
       48B850010CBAFB7F0000 mov      rax, 0x7FFBBA0C0150

G_M20354_IG03:

G_M20354_IG04:
       48B9081BE1B9FB7F0000 mov      rcx, 0x7FFBB9E11B08
       E80BBBF65E           call     CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPE
       488BF0               mov      rsi, rax
       48B98099E1B9FB7F0000 mov      rcx, 0x7FFBB9E19980
       E8F9BAF65E           call     CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPE
       483BC6               cmp      rax, rsi
       751F                 jne      SHORT G_M20354_IG06
       48B9285728EFA5020000 mov      rcx, 0x2A5EF285728
       48B850010CBAFB7F0000 mov      rax, 0x7FFBBA0C0150

G_M20354_IG05:

G_M20354_IG06:
       C3                   ret

; Total bytes of code 149, prolog size 5 for method Bar`1[Boolean][System.Boolean]:Foo()
```
This is a simplification of code we have in the async/await infrastructure, for getting cached tasks for certain return types, and it used to produce the correct code on desktop... has something changed here recently?  Is this not supposed to remove code in the manner I expected?

JosephTremoulet:
This is a pure helper w/o side-effects, so add it to the lists of
tractable helpers in value-numbering; this allows redundant calls to be
CSEd, and fixes #9552 so we can again optimize away type checks on type
parameters in generic code (a not-infrequent pattern).

### DIFF
--- a/src/jit/utils.cpp
+++ b/src/jit/utils.cpp
@@ -1358,6 +1358,7 @@ void HelperCallProperties::init()
             case CORINFO_HELP_ISINSTANCEOFCLASS:
             case CORINFO_HELP_ISINSTANCEOFANY:
             case CORINFO_HELP_READYTORUN_ISINSTANCEOF:
+            case CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPE:
 
                 isPure  = true;
                 noThrow = true; // These return null for a failing cast
--- a/src/jit/valuenum.cpp
+++ b/src/jit/valuenum.cpp
@@ -7466,6 +7466,10 @@ VNFunc Compiler::fgValueNumberHelperMethVNFunc(CorInfoHelpFunc helpFunc)
             vnf = VNF_IsInstanceOf;
             break;
 
+        case CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPE:
+            vnf = VNF_TypeHandleToRuntimeType;
+            break;
+
         case CORINFO_HELP_READYTORUN_ISINSTANCEOF:
             vnf = VNF_ReadyToRunIsInstanceOf;
             break;
--- a/src/jit/valuenumfuncs.h
+++ b/src/jit/valuenumfuncs.h
@@ -34,6 +34,7 @@ ValueNumFuncDef(CastClass, 2, false, false, false)          // Args: 0: Handle o
 ValueNumFuncDef(IsInstanceOf, 2, false, false, false)       // Args: 0: Handle of class being queried, 1: object being queried.
 ValueNumFuncDef(ReadyToRunCastClass, 2, false, false, false)          // Args: 0: Helper stub address, 1: object being cast.
 ValueNumFuncDef(ReadyToRunIsInstanceOf, 2, false, false, false)       // Args: 0: Helper stub address, 1: object being queried.
+ValueNumFuncDef(TypeHandleToRuntimeType, 1, false, false, false)      // Args: 0: TypeHandle to translate
 
 ValueNumFuncDef(LdElemA, 3, false, false, false)            // Args: 0: array value; 1: index value; 2: type handle of element.
 


JosephTremoulet:
@AndyAyersMS @briansull @dotnet/jit-contrib PTAL.

AndyAyersMS:
Do we know for sure it is pure? Or is it only pure for the special cases that the pattern match was getting?

Would like to see if you get any diffs from this in jit-diff.

JosephTremoulet:
> Do we know for sure it is pure? Or is it only pure for the special cases that the pattern match was getting?

It looks to me like the pattern-match is happy with any `CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPE` call it can find, the heuristics are about how it's consumed rather than how it's produced.  I suppose it's possible there's some hidden contract in the pattern-match?

Maybe @jkotas or @davidwrighton or someone can answer more authoritatively: is it safe for the JIT to always assume that `CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPE` is pure and non-throwing?

jkotas:
>  is it safe for the JIT to always assume that CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPE is pure and non-throwing?

Yes. (It can throw out-of-memory - but it should stop you from treating it as pure and side-effect free.)

mikedn:
Should the relevant morph code be deleted?

JosephTremoulet:

pgavlin:
> Should the relevant morph code be deleted?

I would recommend leaving the code in morph if @AndyAyersMS's pending work makes it effective again, as it may be useful during scenarios that do not run with the SSA-based optimizer enabled.

JosephTremoulet:
>Should the relevant morph code be deleted?

@AndyAyersMS [indicated](https://github.com/dotnet/coreclr/issues/9552#issuecomment-279490828) that he's planning to update the importer in a way that the morph code would again be catching most (all?) of these changes.  My thinking is the redundancy isn't much of a maintenance or throughput concern, so may as well have the morph code to snip out most cases early and the value numbering to pick up any stragglers or new/different patterns.

AndyAyersMS:
Yeah, I think we should do both; the space & size savings are significant and the morph pattern kicks in unconditionally, and it's pretty cheap (in terms of jit time) to recognize. The less code the optimizer has to look at the better.

Will be interesting to see if the VN approach finds any additional instances.

JosephTremoulet:
> Would like to see if you get any diffs from this in jit-diff

I do, but if any of them are from optimizing away this sort of compare, they get drowned out because most/all diffs are straight CSEs of redundant calls to this helper, not on two sides of a compare.  Here's the jit-analyze output:

```
Summary:
(Note: Lower is better)

Total bytes of diff: -9248 (-0.07 % of base)
    diff is an improvement.

       -8110 : System.Private.CoreLib.dasm (-0.23 % of base)
        -582 : System.Linq.Parallel.dasm (-0.09 % of base)
        -188 : System.Reflection.Metadata.dasm (-0.05 % of base)
        -178 : System.Linq.Expressions.dasm (-0.03 % of base)
        -113 : Microsoft.CSharp.dasm (-0.03 % of base)

9 total files with size differences (9 improved, 0 regressed).

          37 : System.Private.CoreLib.dasm - System.Diagnostics.Tracing.ManifestBuilder:CreateManifestString():ref:this
          18 : System.Private.CoreLib.dasm - System.Runtime.InteropServices.WindowsRuntime.CLRIPropertyValueImpl:CoerceScalarValue(int,ref):ref
           8 : System.Private.CoreLib.dasm - System.Diagnostics.Tracing.EventSource:WriteMultiMerge(ref,byref,ref,long,long,long):this
           7 : Microsoft.CodeAnalysis.dasm - Roslyn.Utilities.ObjectWriter:WriteValue(ref):this
           7 : Microsoft.CSharp.dasm - Microsoft.CSharp.RuntimeBinder.Semantics.CType:IsCollectionType():bool:this

       -1286 : System.Private.CoreLib.dasm - System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1[Boolean][System.Boolean]:SetResult(bool):this
       -1133 : System.Private.CoreLib.dasm - System.Runtime.InteropServices.WindowsRuntime.CLRIPropertyValueImpl:CoerceScalarValue(int,ref):int (2 methods)
        -935 : System.Private.CoreLib.dasm - System.Runtime.InteropServices.WindowsRuntime.CLRIPropertyValueImpl:CoerceScalarValue(int,ref):ubyte
        -833 : System.Private.CoreLib.dasm - System.Runtime.InteropServices.WindowsRuntime.CLRIPropertyValueImpl:CoerceScalarValue(int,ref):short
        -780 : System.Private.CoreLib.dasm - System.Runtime.InteropServices.WindowsRuntime.CLRIPropertyValueImpl:CoerceScalarValue(int,ref):long (2 methods)

117 total methods with size differences (92 improved, 25 regressed).
```

> Will be interesting to see if the VN approach finds any additional instances

Presumably these CSE cases get past the compare-centric morph optimization.

AndyAyersMS:
Above data is from running on top of #9562... ?

stephentoub:
Do we have any tests to help catch regressions in this kind of thing in the future?

AndyAyersMS:
For code size and such there is manual testing available via jit-diff; in fact I ran it for #7923 and commented as such in the PR.  We usually ask jit devs to post this data. 

Evidently I did something wrong when I ran the tests.

JosephTremoulet:
> Above data is from running on top of #9562... ?

No, running before that got merged.

AndyAyersMS:
The data I get on top of #9562:
```
Summary:
(Note: Lower is better)

Total bytes of diff: -2631 (-0.02 % of base)
    diff is an improvement.

       -2045 : System.Private.CoreLib.dasm (-0.06 % of base)
        -188 : System.Reflection.Metadata.dasm (-0.05 % of base)
        -173 : System.Linq.Expressions.dasm (-0.03 % of base)
        -134 : Microsoft.CSharp.dasm (-0.03 % of base)
         -69 : System.Linq.Queryable.dasm (-0.11 % of base)

7 total files with size differences (7 improved, 0 regressed).

          37 : System.Private.CoreLib.dasm - System.Diagnostics.Tracing.ManifestBuilder:CreateManifestString():ref:this
          35 : System.Private.CoreLib.dasm - System.Runtime.InteropServices.WindowsRuntime.CLRIPropertyValueImpl:CoerceScalarValue(int,ref):ref
          12 : System.Linq.Expressions.dasm - System.Linq.Expressions.Expression:ValidateGoto(ref,byref,ref,ref,ref)
           7 : Microsoft.CodeAnalysis.dasm - Roslyn.Utilities.ObjectWriter:WriteValue(ref):this
           5 : System.Private.CoreLib.dasm - System.Diagnostics.Tracing.EventSource:WriteMultiMerge(ref,byref,ref,long,long,long):this

        -442 : System.Private.CoreLib.dasm - System.Collections.Generic.EqualityComparer`1[Byte][System.Byte]:CreateComparer():ref
        -330 : System.Private.CoreLib.dasm - System.Runtime.CompilerServices.JitHelpers:UnsafeEnumCast(int):int (4 methods)
         -96 : Microsoft.CSharp.dasm - Microsoft.CSharp.RuntimeBinder.SymbolTable:SetParameterAttributes(ref,ref,int):this
         -95 : System.Private.CoreLib.dasm - System.Exception:GetObjectData(ref,struct):this
         -94 : System.Reflection.Metadata.dasm - System.Reflection.Metadata.BlobWriterImpl:WriteConstant(byref,ref)

124 total methods with size differences (99 improved, 25 regressed).
```

AndyAyersMS:
Diffs in `CreateComparer` are a great example why we want the in-depth aspect of both changes:
```C#
            object result = null;
            RuntimeType t = (RuntimeType)typeof(T);
            
            // Specialize type byte for performance reasons
            if (t == typeof(byte)) {
                result = new ByteEqualityComparer();
            }
```
The code author likely thought hosting `typeof(T)` into a local like this was an optimization, but it breaks the tree-based pattern optimization.

BTW the generated code -- while vastly simpler -- still looks overly complex to me...

PR #9561, opened 2017-02-13
Remove OpenSuse13.2 from runtime section in all Project.Jsons

wtgodbe:
Since we are deprecating support for OpenSuse13.2, we should no longer restore packages for it.


wtgodbe:
CC @gkhanna79 @dagood @ramarag

Issue #7907, opened 2016-10-31
JIT: bad codegen when a method is not an inline candidate


AndyAyersMS:
The jit spills the evaluation stack less aggressively at calls that are not considered inline candidates. This difference leads to a codegen bug in the following test case (distilled from Roslyn):

```C#
    [MethodImpl(MethodImplOptions.NoInlining)]
    public static int G(int z, ref int r)
    {
        r -= z;
        return 1;
    }

    public int F(int count)
    {
        int initialCount = count;

        _position += G(_position, ref count);

        if (initialCount == count)
        {
            count--;
        }

        return initialCount - count;
    }
```
The IL for this method shows `arg1` (`count`) being pushed on the stack, and then while this is live, a call that can modify `arg1`. If the stack is not spilled at the call then the JIT will decide that the `dup` produces the same value as the subsequent `ldarg.1` and so deduce that the branch is never taken.
```
  IL_0000:  ldarg.1
  IL_0001:  ldarg.0
  IL_0002:  ldarg.0
  IL_0003:  ldfld      int32 RoslynMinOptsBug::_position
  IL_0008:  ldarg.0
  IL_0009:  ldfld      int32 RoslynMinOptsBug::_position
  IL_000e:  ldarga.s   count
  IL_0010:  call       int32 RoslynMinOptsBug::G(int32, int32&)  // no inline, so not a candidate
  IL_0015:  add
  IL_0016:  stfld      int32 RoslynMinOptsBug::_position
  IL_001b:  dup
  IL_001c:  ldarg.1
  IL_001d:  bne.un.s   IL_0024
```
This IL pattern comes from (recent?) Roslyn-based C# compilers when optimization is enabled.

PR #9562, opened 2017-02-13
Jit: fix for broken type equality optimization

### LINKED ISSUE #9552: Branches guarded by type checks not being trimmed in generic types/methods

AndyAyersMS:
The jit relies on a particular tree remaining intact during importation
to optimize type equality tests commonly found in generics.

In #7907 the jit started aggressively spilling the evaluation stack before
calls, which broke up the trees needed for the optimizations.

In the special cases pertaining to type tests the trees can safely remain
intact because we know the calls can't interact with the computations on
the stack. Recognize these cases and selectively suppress spilling.

Closes #9552.

### DIFF
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -7692,10 +7692,33 @@ DONE_CALL:
                         call           = gtNewLclvNode(calliSlot, type);
                     }
                 }
+
                 // For non-candidates we must also spill, since we
                 // might have locals live on the eval stack that this
                 // call can modify.
-                impSpillSideEffects(true, CHECK_SPILL_ALL DEBUGARG("non-inline candidate call"));
+                //
+                // Suppress this for certain well-known call targets
+                // that we know won't modify locals, eg calls that are
+                // recognized in gtCanOptimizeTypeEquality. Otherwise
+                // we may break key fragile pattern matches later on.
+                bool spillStack = true;
+                if (call->IsCall())
+                {
+                    GenTreeCall* callNode = call->AsCall();
+                    if ((callNode->gtCallType == CT_HELPER) && gtIsTypeHandleToRuntimeTypeHelper(callNode))
+                    {
+                        spillStack = false;
+                    }
+                    else if ((callNode->gtCallMoreFlags & GTF_CALL_M_SPECIAL_INTRINSIC) != 0)
+                    {
+                        spillStack = false;
+                    }
+                }
+
+                if (spillStack)
+                {
+                    impSpillSideEffects(true, CHECK_SPILL_ALL DEBUGARG("non-inline candidate call"));
+                }
             }
         }
 
@@ -12813,7 +12836,7 @@ void Compiler::impImportBlockCode(BasicBlock* block)
                     impHandleAccessAllowed(callInfo.accessAllowed, &callInfo.callsiteCalloutHelper);
 
 #if 0 // DevDiv 410397 - This breaks too many obfuscated apps to do this in an in-place release
-     
+
                 // DevDiv 291703 - we need to check for accessibility between the caller of InitializeArray
                 // and the field it is reading, thus it is now unverifiable to not immediately precede with
                 // ldtoken <filed token>, and we now check accessibility


AndyAyersMS:
@JosephTremoulet PTAL
cc @dotnet/jit-contrib 

Fixes the case in #9552.

Code size diffs:
```
Total bytes of diff: -19089 (-0.14 % of base)
    diff is an improvement.
          18 : Microsoft.VisualBasic.dasm (0.01 % of base)
           3 : System.Diagnostics.DiagnosticSource.dasm (0.04 % of base)
      -16068 : System.Private.CoreLib.dasm (-0.45 % of base)
       -1917 : System.Linq.Parallel.dasm (-0.31 % of base)
        -322 : Microsoft.CodeAnalysis.dasm (-0.04 % of base)
        -278 : System.Runtime.Extensions.dasm (-0.17 % of base)
        -206 : System.Net.Sockets.dasm (-0.12 % of base)
15 total files with size differences (13 improved, 2 regressed).
          10 : Microsoft.VisualBasic.dasm - Microsoft.VisualBasic.CompilerServices.VBCallBinder:FallbackInvokeMember(ref,ref,ref):ref:this
          10 : System.Linq.Expressions.dasm - MetaDynamic:BuildCallMethodWithResult(ref,ref,ref,ref,ref):ref:this
           8 : Microsoft.VisualBasic.dasm - Microsoft.VisualBasic.CompilerServices.VBIndexSetComplexBinder:FallbackSetIndex(ref,ref,ref,ref):ref:this
           8 : System.Private.CoreLib.dasm - System.Exception:GetExceptionMethodFromString():ref:this
           7 : System.Linq.Expressions.dasm - System.Linq.Expressions.Expression:ValidateIndexedProperty(ref,ref,ref,byref)
       -2580 : System.Private.CoreLib.dasm - System.Runtime.InteropServices.WindowsRuntime.CLRIPropertyValueImpl:CoerceScalarValue(int,ref):long (2 methods)
       -2498 : System.Private.CoreLib.dasm - System.Runtime.InteropServices.WindowsRuntime.CLRIPropertyValueImpl:CoerceScalarValue(int,ref):int (2 methods)
       -1917 : System.Linq.Parallel.dasm - System.Linq.Parallel.Util:GetDefaultComparer():ref (13 methods)
       -1325 : System.Private.CoreLib.dasm - System.Runtime.InteropServices.WindowsRuntime.CLRIPropertyValueImpl:CoerceScalarValue(int,ref):double
       -1325 : System.Private.CoreLib.dasm - System.Runtime.InteropServices.WindowsRuntime.CLRIPropertyValueImpl:CoerceScalarValue(int,ref):float
107 total methods with size differences (98 improved, 9 regressed).
```

stephentoub:
Thanks, @JosephTremoulet and @AndyAyersMS!  I applied Joseph's change and looked at the case in AsyncTaskMethodBuilder I cared about.... significant drop in the amount of assembly, though there was still one CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPE call remaining that looked unnecessary.  Then I applied Andy's change as well, and it went away.  Everything looks good to me now. Thanks!

AndyAyersMS:
This isn't the only place in the importer that can spill so perhaps we can use special knowledge of the callee more generally to pare down the side effect flags for GT_CALL nodes and reduce interference in more cases. Will look at that separately.

JosephTremoulet:
> there was still one CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPE call remaining that looked unnecessary. Then I applied Andy's change as well, and it went away.

Probably in general dead-code elimination we're conservative about removing dead calls for fear of side-effects, and the optimization in morph is more aggressive.  I wonder how many unnecessary helper calls get left around...

briansull:
Note that we don't value number blocks that are unreachable at all.

Thus if this pattern occurs in an unreachable block we won't eliminate it via **identical** value numbers.

This leads to a A/V in our new JIT in some customer code that contains a loop that is entirely unreachable,when we try to unpack value numbers to see if we can eliminate a null check.

PR #9563, opened 2017-02-13
Fix `JitTimeLog::PrintCsvHeader`.

pgavlin:
This function attempts to suppress its output if the JIT time log is not
empty. Unfortunately, the method it uses to do this detection gives
incorrect results on Windows: as documented on MSDN, `ftell` will return
`0` for a file opened for appending until the first I/O operation on
that file occurs. This change fixes this by seeking to the end of the
file prior to `ftell`.


pgavlin:

BruceForstall:
LGTM

PR #9564, opened 2017-02-13
Report code and GC info size in the JIT time log.

pgavlin:
Just what it says on the tin.

### DIFF
--- a/src/jit/compiler.cpp
+++ b/src/jit/compiler.cpp
@@ -7824,6 +7824,8 @@ void JitTimer::PrintCsvHeader()
 
             InlineStrategy::DumpCsvHeader(fp);
 
+            fprintf(fp, "\"Executable Code Bytes\",");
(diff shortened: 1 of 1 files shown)


pgavlin:
@BruceForstall @AndyAyersMS @dotnet/jit-contrib PTAL.

Is it also worth reporting unwind info size? There isn't a convenient way to do that at the moment--the value we pass to `allocUnwindUnfo` is not saved anywhere--but it would be pretty straightforward to fix that.

AndyAyersMS:
I suppose it would be nice to have, though there's not a lot we can do in the jit to directly influence unwind size.

BruceForstall:
@pgavlin Unfortunately, `compInfoBlkSize` only has interesting data if `JIT32_GCENCODER` is defined (i.e., for x86), otherwise it is currently zero. It would be nice to change the non-x86 GC encoder to surface this information so it would be correct for all architectures.

PR #9566, opened 2017-02-13
Add IR size after certain phases to the JIT log.

pgavlin:
This change adds support for measuring and reporting the number of nodes
in a function after a certain subset of the JIT's phases. This reporting
is enabled by setting `COMPlus_JitMeasureIR` and `COMPlus_JtiTimeLogCsv`
in conjunction with each other.

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -24270,10 +24270,12 @@ unsigned Compiler::fgMeasureIR()
         {
             for (GenTreeStmt* stmt = block->firstStmt(); stmt != nullptr; stmt = stmt->getNextStmt())
             {
-                fgWalkTreePre(&stmt->gtStmtExpr, [](GenTree** slot, fgWalkData* data) -> Compiler::fgWalkResult {
-                    (*reinterpret_cast<unsigned*>(data->pCallbackData))++;
-                    return Compiler::WALK_CONTINUE;
-                }, &nodeCount);
+                fgWalkTreePre(&stmt->gtStmtExpr,
+                              [](GenTree** slot, fgWalkData* data) -> Compiler::fgWalkResult {
+                                  (*reinterpret_cast<unsigned*>(data->pCallbackData))++;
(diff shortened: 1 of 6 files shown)


pgavlin:
@BruceForstall @AndyAyersMS @dotnet/jit-contrib PTAL

pgavlin:
@BruceForstall @AndyAyersMS @dotnet/jit-contrib PTAL

pgavlin:
> Would also be interesting to know how many dead IR nodes there are.

Can you file an issue and assign it to me? This is easy to calculate if we're willing to maintain a running total of the # of IR nodes created.

AndyAyersMS:
#9569

BruceForstall:
LGTM

You could add a dump of block count, lclvar/tmp count, etc., on a per-phase basis.

You could imagine a more generic UI to avoid pre-determining the set your plan to collect by listing the phases for which you want the info (using the "short names"):

```
set COMPlus_JitMeasureIR=REGALLOC;LSRA
set COMPlus_JitMeasureIR=* // special: * == all phases
set COMPlus_JitMeasureIR=default or 1 or true // special: some pre-determined "interesting" set, like what you've defined
```

src/jit/flowgraph.cpp
```diff
@@ -24256,3 +24256,36 @@ void Compiler::fgTransformFatCalli()
     CheckNoFatPointerCandidatesLeft();
 #endif
 }
+
+//------------------------------------------------------------------------
+// fgMeasureIR: count and return the number of IR nodes in the function.
+//
+unsigned Compiler::fgMeasureIR()
```
mikedn: 30 more lines in a 24000 lines file 😁

pgavlin:
> You could add a dump of block count, lclvar/tmp count, etc., on a per-phase basis.
> 
> You could imagine a more generic UI to avoid pre-determining the set your plan to collect by listing the phases for which you want the info (using the "short names"):

These are both good suggestions. I've filed #9578 to track.

Issue #9480, opened 2017-02-10
[x86/Linux] Unit test interface_struct01 fails


seanshpark:
Generics MemberAccess has several unit test cases and only `interface_struct01` fails. I've made a simpler test program for this.

```
using System;

interface IGen<T>
{
    void Set(T v);
    T Get();
}

struct Gen<T> : IGen<T>
{
    T Field;

    public void Set(T v) { Field = v;}
    public T Get() { return Field;}
}

namespace ConsoleApplication
{
    public class Program
    {
        public static void Main(string[] args)
        {
            Gen<string> GenStringStruct = new Gen<string>();
            IGen<string> GenString = GenStringStruct;

            GenString.Set("Hello World!!!");
            string tt = GenString.Get();
            Console.WriteLine(tt);

        }
    }
}
```

Expected:
```
Hello World!!!
```

Actual: Some broken characters
```
ƃᆨ뇡눍諨떥龴떥
```

seanshpark:
cc @parjong @wateret @YongseopKim

seanshpark:
There is something going on inside `Set()`. With this code,
```
    public void Set(T v)
    {
        Field = v;
        Console.Write("S1: ");
        Console.WriteLine(Field.ToString());
    }
```
Error message is,
```
S1: 
Unhandled Exception: System.AccessViolationException: 
Attempted to read or write protected memory. 
This is often an indication that other memory is corrupt.
   at Gen`1.Set(T v)
   at ConsoleApplication.Program.Main(String[] args)
Aborted (core dumped)
```

@jkotas , can you please help where I could start lookin at? Or ping some person who can add some comments what might the problem? I've looked at `JIT_CheckedWriteBarrier\rg` code, where I think might be the problem but it's quite hard to understand.

seanshpark:
JITted codes...

Program:Main(ref)
```
...
       8B55E4       mov      edx, gword ptr [ebp-1CH]
       8955EC       mov      gword ptr [ebp-14H], edx
       8B15301110B4 mov      edx, gword ptr [B4101130H]
       8B4DEC       mov      ecx, gword ptr [ebp-14H]
       FF1510E028B6 call     [IGen`1[__Canon][System.__Canon]:Set(ref):this]
       90           nop      
       8B4DEC       mov      ecx, gword ptr [ebp-14H]
       FF1514E028B6 call     [IGen`1[__Canon][System.__Canon]:Get():ref:this]
...
```
and
```
[...]

PR #8848, opened 2017-01-08
[x86/Linux] Fix generic context(hidden arg) position

seanshpark:
This fixes how to handle generic context position in x86 by num of arguments

PR #9571, opened 2017-02-14
[x86/Linux] Fix IL Stub for Shared Generic

seanshpark:
Fix parameter order in UnboxingILStub for Shared Generic

### DIFF
--- a/src/vm/prestub.cpp
+++ b/src/vm/prestub.cpp
@@ -731,14 +731,22 @@ Stub * CreateUnboxingILStubForSharedGenericValueTypeMethods(MethodDesc* pTargetM
 
     // 2. Emit the method body
     mdToken tokPinningHelper = pCode->GetToken(MscorlibBinder::GetField(FIELD__PINNING_HELPER__M_DATA));
-    
+
     // 2.1 Push the thisptr
     // We need to skip over the MethodTable*
-    // The trick below will do that. 
+    // The trick below will do that.
     pCode->EmitLoadThis();
     pCode->EmitLDFLDA(tokPinningHelper);
 
-    // 2.2 Push the hidden context param 
+#if defined(_TARGET_X86_)
+    // 2.2 Push the rest of the arguments for x86
+    for (unsigned i = 0; i < msig.NumFixedArgs();i++)
+    {
+        pCode->EmitLDARG(i);
+    }
+#endif
+
+    // 2.3 Push the hidden context param
     // The context is going to be captured from the thisptr
     pCode->EmitLoadThis();
     pCode->EmitLDFLDA(tokPinningHelper);
@@ -746,16 +754,18 @@ Stub * CreateUnboxingILStubForSharedGenericValueTypeMethods(MethodDesc* pTargetM
     pCode->EmitSUB();
     pCode->EmitLDIND_I();
 
-    // 2.3 Push the rest of the arguments
+#if !defined(_TARGET_X86_)
+    // 2.4 Push the rest of the arguments for not x86
     for (unsigned i = 0; i < msig.NumFixedArgs();i++)
     {
         pCode->EmitLDARG(i);
     }
+#endif
 
-    // 2.4 Push the target address
+    // 2.5 Push the target address
     pCode->EmitLDC((TADDR)pTargetMD->GetMultiCallableAddrOfCode(CORINFO_ACCESS_ANY));
 
-    // 2.5 Do the calli
+    // 2.6 Do the calli
     pCode->EmitCALLI(TOKEN_ILSTUB_TARGET_SIG, msig.NumFixedArgs() + 1, msig.IsReturnTypeVoid() ? 0 : 1);
     pCode->EmitRET();
 


seanshpark:
Similar patch to #8848

seanshpark:
Fixes #9480

seanshpark:
CC @parjong, @wateret , @YongseopKim

seanshpark:
@jkotas , @janvorli , PTAL

src/vm/prestub.cpp
```diff
@@ -731,31 +731,52 @@ Stub * CreateUnboxingILStubForSharedGenericValueTypeMethods(MethodDesc* pTargetM
...
 
-    // 2.2 Push the hidden context param 
-    // The context is going to be captured from the thisptr
-    pCode->EmitLoadThis();
-    pCode->EmitLDFLDA(tokPinningHelper);
-    pCode->EmitLDC(Object::GetOffsetOfFirstField());
-    pCode->EmitSUB();
-    pCode->EmitLDIND_I();
+#if defined(_TARGET_X86_)
+    if (numArgs < NUM_ARGUMENT_REGISTERS)
```
jkotas: I do not think you should be checking the number of arguments here. The generic context should be always the last argument on x86, no exceptions.

E.g. see https://github.com/dotnet/coreclr/blob/4822563366cec72e3e269e56a14a15bb238ba8b2/src/vm/callingconvention.h#L1515

jkotas: I see that a similar fix was done in CreateInstantiatingILStub some time ago as well. It should be fixed there as well to not check NUM_ARGUMENT_REGISTERS - if you agree that it is the right fix.

seanshpark: Then I think this can be a little bit simple. As `CreateInstantiatingILStub` change was from @janvorli , is it OK?

seanshpark:
If don't have to check num of arguments it could be something like this.
```
diff --git a/src/vm/prestub.cpp b/src/vm/prestub.cpp
index 723e6b6..7862101 100644
--- a/src/vm/prestub.cpp
+++ b/src/vm/prestub.cpp
@@ -738,7 +738,15 @@ Stub * CreateUnboxingILStubForSharedGenericValueTypeMethods(MethodDesc* pTargetM
     pCode->EmitLoadThis();
     pCode->EmitLDFLDA(tokPinningHelper);
 
-    // 2.2 Push the hidden context param 
+#if defined(_TARGET_X86_)
+    // 2.2 Push the rest of the arguments
+    for (unsigned i = 0; i < msig.NumFixedArgs();i++)
+    {
+        pCode->EmitLDARG(i);
+    }
+#endif
+
+    // 2.3 Push the hidden context param 
     // The context is going to be captured from the thisptr
     pCode->EmitLoadThis();
     pCode->EmitLDFLDA(tokPinningHelper);
@@ -746,11 +754,13 @@ Stub * CreateUnboxingILStubForSharedGenericValueTypeMethods(MethodDesc* pTargetM
     pCode->EmitSUB();
     pCode->EmitLDIND_I();
 
-    // 2.3 Push the rest of the arguments
+#if !defined(_TARGET_X86_)
+    // 2.4 Push the rest of the arguments
     for (unsigned i = 0; i < msig.NumFixedArgs();i++)
     {
         pCode->EmitLDARG(i);
     }
+#endif
 
     // 2.4 Push the target address
     pCode->EmitLDC((TADDR)pTargetMD->GetMultiCallableAddrOfCode(CORINFO_ACCESS_ANY));
```

jkotas:
Right - this looks better to me.

src/vm/prestub.cpp
janvorli: @jkotas so you are saying we should make it different from Windows x86?

seanshpark:
@jkotas, @janvoli, PTAL once more, thanks!

janvorli:
I am confused by how this could work without corresponding change in the JIT. As I have recently added to the ABI doc (based on looking into where the JITted code expects the arguments while you were fixing the  CreateInstantiatingILStub):
> On x86, if all arguments of the function including "this" pointer fit into argument registers (ECX and EDX) and we still have argument registers available, we store the hidden argument in the next available argument register. Otherwise it is passed as the last stack argument.

@jkotas did you mean to change the JIT accordingly or am I missing something?

jkotas:
I do not think that the ABI doc describes what we do on Windows accurately. Take this example:

```
struct BigStruct
{
    int a, b, c, d, e, f;
}

class Program
{
    static void MyG<T>(BigStruct s1, BigStruct s2)
    {
    }

    {
        MyG<string>(new BigStruct(), new BigStruct());
    }
}
```

The callsite in the main method looks like this:

```
... push s1 and s2 on the stack ...
mov         ecx,49A0F90h ... hidden argument
call        dword ptr ds:[49A0F64h]
```

As you can see the hidden argument is passed in register because of we have still register available. But according to the ABI doc, it should have been passed on the stack because of all arguments of the function do not fit into argument registers.

Instead, it should simply say that the hidden argument is passed in as last argument.

jkotas:
@seanshpark LGTM. Could you please also fix `CreateInstantiatingILStub` to match, and also fix the incorrect description in the ABI doc ?

janvorli:
@jkotas I can see now my description was a bit confusing, I really meant  "if all arguments of the function **that can be passed in registers** including "this" pointer fit into argument registers". And I haven't realized it is in fact equivalent to saying that the hidden argument is passed as the last argument and that the rules of passing parameters in registers / on stack take care of the actual location.

jkotas:
Ok. I have opened https://github.com/dotnet/coreclr/issues/9577 to make the same fix in CreateInstantiatingILStub as well.

seanshpark:
@jkotas , @janvorli , Thank you both. I'll make a PR to fix CreateInstantiatingILStub like this.

PR #9574, opened 2017-02-14
Rewrite GetRegisterAddressByIndex using getRegAddr

parjong:
GetRegisterAddressByIndex in exceptionhandling.cpp is a subset of getRegAddr in regdisp.h.

This commit rewrites GetRegisterAddressByIndex using getRegAddr.

### DIFF
--- a/src/vm/exceptionhandling.cpp
+++ b/src/vm/exceptionhandling.cpp
@@ -4698,29 +4698,7 @@ Return value :
 --*/
 VOID* GetRegisterAddressByIndex(PCONTEXT pContext, UINT index)
 {
-#if defined(_TARGET_AMD64_)
-    _ASSERTE(index < 16);
-    return &((&pContext->Rax)[index]);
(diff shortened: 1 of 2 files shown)


src/vm/exceptionhandling.cpp
```diff
@@ -4698,29 +4698,7 @@ Return value :
...
-        offsetof(CONTEXT, Esi),
-        offsetof(CONTEXT, Edi),
-    };
-
-    return (VOID*)(PBYTE(pContext) + OFFSET_OF_REGISTERS[index]);
-#else
-    PORTABILITY_ASSERT("GetRegisterAddressByIndex");
-    return NULL;
-#endif
+    return getRegAddr(index, pContext);
```
janvorli: Could you please change the implementation of the getRegAddr for x86 to be the more effective one that was in the GetRegisterAddressByIndex?

PR #9582, opened 2017-02-14
Do not report FP restores in x86 epilogs.

### LINKED ISSUE #9452: SIMD tests failing under GCStress=0xC on x86
See e.g. https://ci.dot.net/job/dotnet_coreclr/job/master/job/jitstress/job/x86_checked_windows_nt_gcstress0xc/11

@BruceForstall, can you take a look?

pgavlin:
The x86 unwinder neither needs nor expects to see these restores (which
right now consist solely of a `vzeroupper` instruction that is emitted
to eliminate AVX <-> SSE transition penalties), which was causing unwind
failures under GC stress. This change stops reporting these restores as
part of a function epilog.

Fixes #9452.

### DIFF
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -9451,9 +9451,16 @@ void CodeGen::genFnEpilog(BasicBlock* block)
 #endif // !FEATURE_STACK_FP_X87
 
 #ifdef JIT32_GCENCODER
-    // On IA32, we start the OS-reported portion of the epilog after restoring any callee-saved
-    // floating-point registers. This avoids the need to update the x86 unwinder and retains binary
-    // compatibility between later versions of the JIT and earlier versions of the runtime.
+    // When using the JIT32 GC encoder, we do not start the OS-reported portion of the epilog until after
+    // the above call to `genRestoreCalleeSavedFltRegs` because that function
+    //   a) does not actually restore any registers: there are none when targeting the Windows x86 ABI,
+    //      which is the only target that uses the JIT32 GC encoder
+    //   b) may issue a `vzeroupper` instruction to eliminate AVX -> SSE transition penalties.
+    // Because the `vzeroupper` instruction is not recognized by the VM's unwinder and there are no
+    // callee-save FP restores that the unwinder would need to see, we can avoid the need to change the
+    // unwinder (and break binary compat with older versions of the runtime) by starting the epilog
+    // after any `vzeroupper` instruction has been emitted. If either of the above conditions changes,
+    // we will need to rethink this.
     getEmitter()->emitStartEpilog();
 #endif
 
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -9450,6 +9450,13 @@ void CodeGen::genFnEpilog(BasicBlock* block)
     genRestoreCalleeSavedFltRegs(compiler->compLclFrameSize);
 #endif // !FEATURE_STACK_FP_X87
 
+#ifdef JIT32_GCENCODER
+    // On IA32, we start the OS-reported portion of the epilog after restoring any callee-saved
+    // floating-point registers. This avoids the need to update the x86 unwinder and retains binary
+    // compatibility between later versions of the JIT and earlier versions of the runtime.
+    getEmitter()->emitStartEpilog();
+#endif
+
     /* Compute the size in bytes we've pushed/popped */
 
     if (!doubleAlignOrFramePointerUsed())
--- a/src/jit/emit.cpp
+++ b/src/jit/emit.cpp
@@ -1027,7 +1027,7 @@ void emitter::emitBegFN(bool hasFramePtr
     emitPlaceholderList = emitPlaceholderLast = nullptr;
 
 #ifdef JIT32_GCENCODER
-    emitEpilogList = emitEpilogLast = NULL;
+    emitEpilogList = emitEpilogLast = nullptr;
 #endif // JIT32_GCENCODER
 
     /* We don't have any jumps */
@@ -1215,14 +1215,12 @@ size_t emitter::emitGenEpilogLst(size_t (*fp)(void*, unsigned), void* cp)
     EpilogList* el;
     size_t      sz;
 
-    for (el = emitEpilogList, sz = 0; el; el = el->elNext)
+    for (el = emitEpilogList, sz = 0; el != nullptr; el = el->elNext)
     {
-        assert(el->elIG->igFlags & IGF_EPILOG);
+        assert(el->elLoc.GetIG()->igFlags & IGF_EPILOG);
 
-        UNATIVE_OFFSET ofs =
-            el->elIG->igOffs; // The epilog starts at the beginning of the IG, so the IG offset is correct
-
-        sz += fp(cp, ofs);
+        // The epilog starts at the location recorded in the epilog list.
+        sz += fp(cp, el->elLoc.CodeOffset(this));
     }
 
     return sz;
@@ -1957,22 +1955,20 @@ void emitter::emitBegFnEpilog(insGroup* igPh)
 
 #ifdef JIT32_GCENCODER
 
-    EpilogList* el = new (emitComp, CMK_GC) EpilogList;
-    el->elNext     = NULL;
-    el->elIG       = emitCurIG;
+    EpilogList* el = new (emitComp, CMK_GC) EpilogList();
 
-    if (emitEpilogLast)
+    if (emitEpilogLast != nullptr)
+    {
         emitEpilogLast->elNext = el;
+    }
     else
+    {
         emitEpilogList = el;
+    }
 
     emitEpilogLast = el;
 
 #endif // JIT32_GCENCODER
-
-    /* Remember current position so that we can compute total epilog size */
-
-    emitEpilogBegLoc.CaptureLocation(this);
 }
 
 /*****************************************************************************
@@ -1984,8 +1980,11 @@ void emitter::emitEndFnEpilog()
 {
     emitEndPrologEpilog();
 
+#ifdef JIT32_GCENCODER
+    assert(emitEpilogLast != nullptr);
+
     UNATIVE_OFFSET newSize;
-    UNATIVE_OFFSET epilogBegCodeOffset = emitEpilogBegLoc.CodeOffset(this);
+    UNATIVE_OFFSET epilogBegCodeOffset = emitEpilogLast->elLoc.CodeOffset(this);
 #ifdef _TARGET_XARCH_
     UNATIVE_OFFSET epilogExitSeqStartCodeOffset = emitExitSeqBegLoc.CodeOffset(this);
 #else
@@ -2021,6 +2020,7 @@ void emitter::emitEndFnEpilog()
     }
 
 #endif // _TARGET_X86_
+#endif // JIT32_GCENCODER
 }
 
 #if FEATURE_EH_FUNCLETS
@@ -2067,7 +2067,18 @@ void emitter::emitEndFuncletEpilog()
 
 #endif // FEATURE_EH_FUNCLETS
 
+
 #ifdef JIT32_GCENCODER
+
+//
+// emitter::emitStartEpilog:
+//   Mark the current position so that we can later compute the total epilog size.
+//
+void emitter::emitStartEpilog()
+{
+    assert(emitEpilogLast != nullptr);
+    emitEpilogLast->elLoc.CaptureLocation(this);
+}
 
 /*****************************************************************************
  *
--- a/src/jit/emit.cpp
+++ b/src/jit/emit.cpp
@@ -1983,9 +1983,9 @@ void emitter::emitEndFnEpilog()
 #ifdef JIT32_GCENCODER
     assert(emitEpilogLast != nullptr);
 
-    UNATIVE_OFFSET epilogBegCodeOffset = emitEpilogLast->elLoc.CodeOffset(this);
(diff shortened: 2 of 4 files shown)


pgavlin:
@jkotas @BruceForstall @dotnet/jit-contrib PTAL

pgavlin:
@dotnet-bot help

pgavlin:

BruceForstall:
@pgavlin It looks like we'll generate a vzeroupper *after* all the callee-saved float regs are restored. Does that count as "the first instruction in an x86 epilog"?

BruceForstall:
@janvorli PTAL

pgavlin:
> @pgavlin It looks like we'll generate a vzeroupper after all the callee-saved float regs are restored. Does that count as "the first instruction in an x86 epilog"?

Yes: there are no callee-saved float regs on x86.

pgavlin:

jkotas:
Would it be better to not include it in the epilog size that is reported in GC/Unwind info instead?

pgavlin:
> Would it be better to not include it in the epilog size that is reported in GC/Unwind info instead?

I looked at that avenue; that change is not as straightforward as I'd like. We might need to go down that road if I can't get this working, however (looks like the current change is running into issues and I'm still hunting down an AVX machine for a live repro).

BruceForstall:
I think the first thing that should be fixed in #9452 is that the reported epilog size that includes a vzeroupper is wrong, because the vzeroupper size is incorrectly estimated. I suppose that's independent.

src/vm/eetwain.cpp
```diff
@@ -2919,6 +2919,22 @@ bool InstructionAlreadyExecuted(unsigned walkOffset, unsigned actualHaltOffset)
...
 
+// Skips past a "vzeroupper" instruction if one is present. Returns offset unchanged
+// otherwise.
+inline unsigned SKIP_VZEROUPPER_IF_PRESENT(PTR_CBYTE base, unsigned offset)
+{
+    LIMITED_METHOD_CONTRACT;
+
+    if (CheckInstrByte(base[offset], 0xC5) &&
+        CheckInstrByte(base[offset + 1], 0xF8) &&
+        CheckInstrByte(base[offset + 2], 0x77))
```
janvorli: It would be nice to define a constant for the instruction code like we do for other instructions around line 20 in this file to match handling of other instructions.

src/vm/eetwain.cpp
```diff
@@ -3251,7 +3267,7 @@ void UnwindEspFrameEpilog(
     _ASSERTE(!info->ebpFrame && !info->doubleAlign);
     _ASSERTE(info->epilogOffs > 0);
     
-    int offset = 0;
+    int offset = SKIP_VZEROUPPER_IF_PRESENT(epilogBase, 0);
```
janvorli: It looks like the code at line 3284 below relies on the fact that the pop ecx is at epilogBase[0] which seems it would not be the case with your change when there can be the vzeroupper at epilogBase[0..2].

pgavlin:
> I think the first thing that should be fixed in #9452 is that the reported epilog size that includes a vzeroupper is wrong, because the vzeroupper size is incorrectly estimated. I suppose that's independent.

It turned out that this was not independent; both fixes were necessary. The latest changes I've posted also address some PR feedback and fix an incorrect assertion in `SKIP_LEA_ESP_EBP`.

PTAL

pgavlin:

BruceForstall:
LGTM

src/vm/eetwain.cpp
janvorli: @pgavlin have you considered making this change?

pgavlin:

pgavlin:
I cannot reproduce the GC stress failures that I'm now hitting, but the failures corresponding to the presence of the `vzeroupper` instruction in the SIMD tests have ceased. I will continue to investigate.

jkotas:
> > Would it be better to not include it in the epilog size that is reported in GC/Unwind info instead?

> I looked at that avenue; that change is not as straightforward as I'd like

Why not? I would expect that you just need to subtract the size of vzeroupper when recording the epilog size in gcinfo.

I understand that it may need to be part of what JIT considers epilog internally because of implementation difficulties. But it would be nice to contain the damage within the JIT, not spread it through the system.

pgavlin:
> Why not? I would expect that you just need to subtract the size of vzeroupper when recording the epilog size in gcinfo.
>
> I understand that it may need to be part of what JIT considers epilog internally because of implementation difficulties. But it would be nice to contain the damage within the JIT, not spread it through the system.

I think we would need to adjust both the size and the start location of the epilog.

FWIW, I'm not sure why changing the VM's unwinder is "damage": is there something especially risky about doing so?

Furthermore, I've confirmed that the remaining GC stress failures in the PR job are not due to this change; they are also present in a lab run without this change: https://ci.dot.net/job/dotnet_coreclr/job/master/job/jitstress/job/x86_checked_windows_nt_gcstress0xc/13

Given that this change is simple and functional, I'd like to check it in.

src/vm/eetwain.cpp
pgavlin: Done.

pgavlin:
Failures are all of the form
```
error MSB3073: The command " "D:\j\workspace\x64_release_w---0575cb46\tests\..\Tools\dotnetcli/dotnet.exe" restore --packages "D:\j\workspace\x64_release_w---0575cb46\tests\..\packages" --source https://dotnet.myget.org/F/dotnet-corefxlab/api/v3/index.json --source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json --source https://api.nuget.org/v3/index.json "D:\j\workspace\x64_release_w---0575cb46\tests\src"" exited with code 1.
```

I've seen this locally as well--looks like something in the NuGet/MyGet feeds is horked.

@mmitche is this a known issue?

pgavlin:
@dotnet-bot test Windows_NT arm Cross Release Build

pgavlin:

jkotas:
> I'm not sure why changing the VM's unwinder is "damage"

The handling of this extra instruction does not belong to the unwinder because of it has nothing to do with unwinding. You are only able to do this because of we have a custom unwind format for x86. If there was a fixed standard x86 unwind format like on other architectures, you would not be able to do this as easily.

This change has some risk because of the custom x86 unwinder is tricky, but it is not my top concern. I am ok with risk if it is taken for good reason - I do not think there is a good reason in this case.

Since this change would rev the binary contract, you may also need:
- Bump the R2R format version number because of images generated by the new JIT would not work on old runtimes anymore.
- If there is x86 build flavor of the JIT that can be swapped into the shipping .NET Framework, it should be disabled because of it would not work well anymore.

pgavlin:
> Since this change would rev the binary contract, you may also need:
>
> Bump the R2R format version number because of images generated by the new JIT would not work on old runtimes anymore.
> If there is x86 build flavor of the JIT that can be swapped into the shipping .NET Framework, it should be disabled because of it would not work well anymore.

These are excellent reasons to avoid reporting this instruction as part of the unwind epilog. Thank you for the explanation!

I will go ahead and move the relevant changes to the JIT.

pgavlin:
@jkotas @janvorli @BruceForstall I've reworked this change to avoid reporting the `vzeroupper` instruction as part of the epilog on x86 (along with any other FP restores). PTAL.

pgavlin:

BruceForstall:
emitStartEpilog() doesn't need to be all-platform: it should be x86 only. And thus, you shouldn't even create a new function, just inline the CaptureLocation() call for x86 only.

There should be some serious cleanup done here, e.g.:
1. emitEpilogBegLoc, emitExitSeqBegLoc, emitExitSeqSize, emitEpilogSize should all be under `JIT32_GCENCODER`
2. emitEndFnEpilog() should all be x86-specific except for the first line (emitEndPrologEpilog()).
3. emitStartExitSeq() should be under `JIT32_GCENCODER`
4. emitEndCodeGen() should set `*epilogSize = 0` for non-`JIT32_GCENCODER` (it is passed on to genCreateAndStoreGCInfo() and then not used in that case. Thus, the the non-`JIT32_GCENCODER` genCreateAndStoreGCInfo() version shouldn't even take an epilogSize.

This would clean it up considerably, making it easier to understand, and make the requirement for calculating the epilog size more clear and targeted.

pgavlin:
> emitStartEpilog() doesn't need to be all-platform: it should be x86 only. And thus, you shouldn't even create a new function, just inline the CaptureLocation() call for x86 only.

Personally, I'd prefer to maintain as consistent an interface between the code generator and the emitter as possible across all platforms, hence the new function. Similarly, I'd prefer to insulate the emitter's internals as much as possible, hence the function call (note that this approach is already taken by `emitStartExitSeq`).

For similar reasons, I'd rather not move more functions (and therefore more calls to said functions) under `JIT32_GCENCODER`.

Regardless, we can do these cleanups later if you feel they're warranted. For now I'd prefer to get this fix in so that we can get a clearer picture of our performance under GC stress.

pgavlin:
As it turns out, this change isn't sufficient in any case. `emitEpilogBegLoc` is only used to calculate the size of the epilog, not the starting point.

pgavlin:

I've adjusted the change s.t. the emitter reports the end offset of the epilog to the GC info encoder, which then calculates the start offset as `end_offset - epilog_size`.

BruceForstall:
> Personally, I'd prefer to maintain as consistent an interface between the code generator and the emitter as possible across all platforms, hence the new function. Similarly, I'd prefer to insulate the emitter's internals as much as possible, hence the function call (note that this approach is already taken by emitStartExitSeq).

I very much disagree. The emitStartEpilog() concept you introduced is _only_ relevant to `JIT32_GCENCODER` targets. Adding a function call for other targets is simply confusing, and unnecessary.

Also, we already have emitBegFnEpilog() / emitBegFuncletEpilog(), so the naming of emitStartEpilog() is very confusing. And even if you did keep it, it's not called for funclet epilogs. This doesn't matter for current targets, but soon we'll have Linux/x86 which is currently executing on a plan to set `FEATURE_EH_FUNCLETS`.

src/jit/emit.cpp
```diff
@@ -1219,8 +1219,9 @@ size_t emitter::emitGenEpilogLst(size_t (*fp)(void*, unsigned), void* cp)
     {
         assert(el->elIG->igFlags & IGF_EPILOG);
 
-        UNATIVE_OFFSET ofs =
-            el->elIG->igOffs; // The epilog starts at the beginning of the IG, so the IG offset is correct
+        // The epilog starts epilog_size bytes before the end of the IG. The encoder already has
+        // the epilog size, so hand it the end offset of the IG.
```
BruceForstall: This creates an implicit requirement that an epilog completely lives within a single IG. This is undesirable, and is not required today. (It is required of function prologs; I worked hard to remove the requirement for all other prolog/epilogs).

pgavlin:

pgavlin:
@BruceForstall: I believe I've addressed the main points of your feedback (and produced a working change :))

src/jit/emit.cpp
```diff
@@ -1984,8 +1980,11 @@ void emitter::emitEndFnEpilog()
 {
     emitEndPrologEpilog();
 
+#ifdef JIT32_GCENCODER
```
BruceForstall: I think you should get rid of the other `#ifdefs` in this function, since it will always be `defined(_TARGET_XARCH_)` and `defined(_TARGET_X86_)`.

src/jit/codegencommon.cpp
```diff
@@ -9450,6 +9450,13 @@ void CodeGen::genFnEpilog(BasicBlock* block)
     genRestoreCalleeSavedFltRegs(compiler->compLclFrameSize);
 #endif // !FEATURE_STACK_FP_X87
 
+#ifdef JIT32_GCENCODER
+    // On IA32, we start the OS-reported portion of the epilog after restoring any callee-saved
+    // floating-point registers. This avoids the need to update the x86 unwinder and retains binary
+    // compatibility between later versions of the JIT and earlier versions of the runtime.
+    getEmitter()->emitStartEpilog();
```
BruceForstall: IMO, this comment is confusing. If there _were_ any callee-saved registers, then of course the unwinder would have to know about them. So the real, and only, reason is that we need to avoid including in the VM-known epilog, instructions that the unwinder doesn't understand, and which don't contribute to unwinding, namely vzeroupper. If we were to change the ABI convention to include callee-saved xmm/ymm registers (which perhaps we should do, for CQ/performance reasons), we'd need to update the unwinder, and move this line of code.

BruceForstall:
A couple relatively small comments, but I like how the code looks.

src/jit/codegencommon.cpp
pgavlin: > If we were to change the ABI convention to include callee-saved xmm/ymm registers (which perhaps we should do, for CQ/performance reasons), we'd need to update the unwinder, and move this line of code.

Yeah, this gets tricky. Avoiding the AVX -> SSE transition penalty at the end of a function requires that the `vzeroupper` instruction executes after all AVX instructions in the function, including those that may be used to restore callee-save FP registers. In other words, unless we ensured that the restores only used SSE instructions, we could not move the `vzeroupper` instruction before the restores and it would therefore end up as part of the epilog. At that point we'd need to update the unwinder to recognize `vzeroupper` which would obivate the need for this change.

I agree that the comment is not terribly clear--I'll see if I can think of a better wording.

src/jit/emit.cpp
pgavlin: Sure. I wasn't sure whether or not we used the JIT32 GC encoder for ARM32, so I elected not to change this earlier.

pgavlin:

PR #9583, opened 2017-02-14
Value number casts and zeros of TYP_BYREF

### LINKED ISSUE #9453: `Assertion failed 'unreached' in 'System.SpanTests.ReadOnlySpanTests:CtorPointerNull()` during value numbering in System.Memory.Tests
This is causing a number of failures in the corefx jobs; see e.g. https://ci.dot.net/job/dotnet_coreclr/job/master/job/jitstress/job/x64_checked_windows_nt_corefx_baseline/82/consoleFull

@JosephTremoulet can you take a look?

JosephTremoulet:
Value numbering already has a representation for constants of byref type.
Hook up both `VNZeroForType` and `EvalCastForConstantArgs` to return a
value number using this representation (rather than fail an assertion)
when the requested result type is byref.

Fixes #9453

### DIFF
--- a/src/jit/valuenum.cpp
+++ b/src/jit/valuenum.cpp
@@ -833,7 +833,7 @@ ValueNum ValueNumStore::VNForHandle(ssize_t cnsVal, unsigned handleFlags)
 }
 
 // Returns the value number for zero of the given "typ".
-// It has an unreached() for a "typ" that has no zero value, such as TYP_BYREF.
+// It has an unreached() for a "typ" that has no zero value, such as TYP_VOID.
 ValueNum ValueNumStore::VNZeroForType(var_types typ)
 {
     switch (typ)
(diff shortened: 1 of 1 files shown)


JosephTremoulet:
@briansull @pgavlin PTAL.
/cc @dotnet/jit-contrib

pgavlin:
The mechanics of the change look good to me. Have we verified that we are allowed to encounter a null byref?

JosephTremoulet:
> Have we verified that we are allowed to encounter a null byref?

The code being compiled explicitly generates one, we need to at least avoid crashing when we see it, and I think letting it through sets us up best for good codegen for these "fast span"s.

briansull:
LGTM, 
it makes sense that we would have to add better support for a null byref after introducing Span<T>

PR #9587, opened 2017-02-14
Enable delay signing in ilasm

### LINKED ISSUE #9292: .NET Core ILAsm does not correctly sign binaries
I'm not sure what exactly the root problem is here, but I am seeing some bad / strange behavior when attempting to use the .NET Core version of ilasm from the latest nuget packages. We use the "full framework ILAsm" in our Windows build in corefx in order to build the System.Runtime.CompilerServices.Unsafe library, built from [an IL file](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.CompilerServices.Unsafe/src/System.Runtime.CompilerServices.Unsafe.il). When switching our build targets to use the .NET Core version of ILAsm, we get an error from our signing targets complaining about the assembly produced:

```
"D:\OSS\corefx\src\System.Runtime.CompilerServices.Unsafe\src\System.Runtime.CompilerServices.Unsafe.ilproj" (rebuild target) (1) ->
(OpenSourceSign target) ->
  D:\OSS\corefx\Tools\sign.targets(70,5): error : D:\OSS\corefx\bin/obj/Windows_NT.AnyCPU.Debug/System.Runtime.CompilerServices.Unsafe/netcoreapp/Sys
tem.Runtime.CompilerServices.Unsafe.dll: PE file is not a delay-signed assembly. [D:\OSS\corefx\src\System.Runtime.CompilerServices.Unsafe\src\System
.Runtime.CompilerServices.Unsafe.ilproj]
```

> System.Runtime.CompilerServices.Unsafe.dll: PE file is not a delay-signed assembly

I've used ILDasm (full framework version) to inspect the differences between the Unsafe.dll produced from both ILAsm's (full framework and .NET Core), and there are only two small differences.

[Here is the IL produced from the full framework ILAsm](https://gist.github.com/mellinoe/c3a6c21fd43e143063f86e1069a6347e)

[Here is the IL produced from the .NET Core ILAsm](https://gist.github.com/mellinoe/59ff321e002597b1cc259a86310e8f41)

The obvious difference is the .NET Core version has:
> // Metadata version: v4.0.22220

instead of 
> // Metadata version: v4.0.30319

The .NET Core version also has:
> // Image base: 0x01590000

instead of 
> // Image base: 0x03060000

The MVIDs are different, but that seems normal to me.

Do any of the above differences explain why we would be seeing this error?

adiaaida:
Corefx needs the .Net Core version of ilasm to create delay signable
assemblies. This change brings back AllocateStrongNameSignature that was
deleted in 2efbb92 and modifies it to use similar logic from Roslyn to
determine the size of the signature that we need to allocate space for.

Fixes #9292.

### DIFF
--- a/src/ilasm/writer.cpp
+++ b/src/ilasm/writer.cpp
@@ -982,6 +982,62 @@ BOOL Assembler::EmitEventsProps(Class* pClass)
     return ret;
 }
 
+HRESULT Assembler::AllocateStrongNameSignature()
+{
+    HRESULT             hr = S_OK;
+    HCEESECTION         hSection;
+    DWORD               dwDataLength;
+    DWORD               dwDataOffset;
+    DWORD               dwDataRVA;
+    VOID               *pvBuffer;
+    AsmManStrongName   *pSN = &m_pManifest->m_sStrongName;
+
+    // pSN->m_cbPublicKey is the length of the m_pbPublicKey
+    dwDataLength = ((int)pSN->m_cbPublicKey < 128 + 32) ? 128 : (int)pSN->m_cbPublicKey - 32;
+
+    // Grab memory in the section for our stuff.
+    if (FAILED(hr = m_pCeeFileGen->GetIlSection(m_pCeeFile,
+                                                &hSection)))
+    {
+        return hr;
+    }
+
+    if (FAILED(hr = m_pCeeFileGen->GetSectionBlock(hSection,
+                                                   dwDataLength,
+                                                   4,
+                                                   &pvBuffer)))
+    {
+        return hr;
+    }
+
+    // Where did we get that memory?
+    if (FAILED(hr = m_pCeeFileGen->GetSectionDataLen(hSection,
+                                                     &dwDataOffset)))
+    {
+        return hr;
+    }
+
+    dwDataOffset -= dwDataLength;
+
+    // Convert to an RVA.
+    if (FAILED(hr = m_pCeeFileGen->GetMethodRVA(m_pCeeFile,
+                                                dwDataOffset,
+                                                &dwDataRVA)))
+    {
+        return hr;
+    }
+
+    // Emit the directory entry.
+    if (FAILED(hr = m_pCeeFileGen->SetStrongNameEntry(m_pCeeFile,
+                                                      dwDataLength,
+                                                      dwDataRVA)))
+    {
+        return hr;
+    }
+
+    return S_OK;
+}
+
 #ifdef _PREFAST_
 #pragma warning(push)
 #pragma warning(disable:21000) // Suppress PREFast warning about overly large function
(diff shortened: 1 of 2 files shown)


adiaaida:
@mellinoe @jkotas @RussKeldorph PTAL
/cc: @dotnet/jit-contrib

adiaaida:
@agocke also PTAL

src/ilasm/writer.cpp
```diff
@@ -982,6 +982,59 @@ BOOL Assembler::EmitEventsProps(Class* pClass)
...
+{
+    HRESULT             hr = S_OK;
+    HCEESECTION         hSection;
+    DWORD               dwDataLength;
+    DWORD               dwDataOffset;
+    DWORD               dwDataRVA;
+    VOID               *pvBuffer;
+    AsmManStrongName   *pSN = &m_pManifest->m_sStrongName;
+
+    int keySize = sizeof(pSN->m_pbPublicKey)/sizeof(BYTE);
```
jkotas: This does not look right: `m_cbPublicKey` is pointer. `sizeof(pSN->m_cbPublicKey)` will be size of pointer...

adiaaida: I used m_pbPublicKey, which is an array of BYTES.

adiaaida: I thought m_pbPublicKey was an array of BYTES?

adiaaida: Oh, sorry. My mistake. I see what you're saying.

adiaaida:

src/ilasm/writer.cpp
```diff
@@ -982,6 +982,61 @@ BOOL Assembler::EmitEventsProps(Class* pClass)
...
+        dwDataLength = 0;
+    }
+    else
+    {
+        dwDataLength = (keySize < 128 + 32) ? 128 : keySize - 32;
+    }
+
+    // Grab memory in the section for our stuff.
+    if (FAILED(hr = m_pCeeFileGen->GetIlSection(m_pCeeFile,
+                                                &hSection)))
```
RussKeldorph: Only thing I would ask is that you stop eliding braces.  I realize the formatting of this code is awful, but please just add braces in all of the code you added.

RussKeldorph:
:shipit:

src/ilasm/writer.cpp
adiaaida: I can add them in. Except for calculating the keySize, all of the code that I've added here originally was in writer.cpp, and was removed last week.

src/ilasm/writer.cpp
```diff
@@ -982,6 +982,61 @@ BOOL Assembler::EmitEventsProps(Class* pClass)
...
+    DWORD               dwDataRVA;
+    VOID               *pvBuffer;
+    AsmManStrongName   *pSN = &m_pManifest->m_sStrongName;
+
+    // pSN->m_cbPublicKey is the length of the m_pbPublicKey
+    int keySize = (int)pSN->m_cbPublicKey;
+
+    if (keySize == 0)
+    {
+        dwDataLength = 0;
```
jkotas: Can the `m_cbPublicKey == 0` condition ever happen? This method is only called if `m_pbPublicKey != NULL`. How can we end up in situation where `(m_pbPublicKey != NULL) && (m_cbPublicKey == 0)` ?

adiaaida: I didn't look very hard into that, but I think that it likely cannot. We had that condition in the Roslyn code, so I brought it over.

PR #9588, opened 2017-02-14
Add perf job for legacy jit

adiaaida:
Add performance testing jobs for legacy jit (compatjit). This change adds
x86compatjit as an architecture to perf.groovy. It does something similar
to what we do in netci.groovy to specify arch (x86, x64, x86compatjit) and
architecture (x86, x64) so that for x86compatjit, we use x86 as the arch,
but x86compatjit as the signifier for job names, etc. This change adds
-legacy as a flag to run-xunit-perf.cmd which then sets USE_LEGACY. When
USE_LEGACY is set, we will call the compatjit testenv file, which sets the
appropriate environment variables. This allows us more flexibility if, in
the future, how we run the compat jit changes.

This change also cleans up some of the calls in perf.groovy to use
variables that we've already set, rather than baking in the contents of
those variables (changing Release to ${configuration}).

### DIFF
--- a/perf.groovy
+++ b/perf.groovy
@@ -27,8 +27,21 @@ def static getOSGroup(def os) {
 // Setup perflab tests runs
 [true, false].each { isPR ->
     ['Windows_NT'].each { os ->
-		['x64', 'x86'].each { architecture ->
-			def newJob = job(Utilities.getFullJobName(project, "perf_perflab_${os}_${architecture}", isPR)) {
+		['x64', 'x86', 'x86jit32'].each { arch ->
+            architecture = arch
+            testEnv = ''
+
+            if (arch == 'x86jit32')
+            {
+                architecture = 'x86'
+                testEnv = '-testEnv %WORKSPACE%\\tests\\x86\\compatjit_x86_testenv.cmd'
+            }
+            else if (arch == 'x86')
+            {
+                testEnv = '-testEnv %WORKSPACE%\\tests\\x86\\ryujit_x86_testenv.cmd'
+            }
+
+			def newJob = job(Utilities.getFullJobName(project, "perf_perflab_${os}_${arch}", isPR)) {
 				// Set the label.
 				label('windows_clr_perf')
 				wrappers {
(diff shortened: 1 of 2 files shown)


adiaaida:
@RussKeldorph @DrewScoggins PTAL

adiaaida:
@dotnet-bot test ci please

adiaaida:
@dotnet-bot test ci please

perf.groovy
```diff
@@ -27,8 +27,22 @@ def static getOSGroup(def os) {
...
+		['x64', 'x86ryujit', 'jit32'].each { arch ->
+            architecture = arch
+            testEnv = ''
+
+            if (arch == 'jit32')
+            {
+                architecture = 'x86'
+                testEnv = '-testEnv %WORKSPACE%\\tests\\x86\\compatjit_x86_testenv.cmd'
+            }
+            else if (arch == 'x86ryujit')
```
DrewScoggins: I think we can get rid of this entire branch.  The ryujit_x86 command that we call does not actually do anything so it does not seem like we need to call it.

DrewScoggins:
Other than the one comment LGTM.

perf.groovy
```diff
@@ -27,8 +27,22 @@ def static getOSGroup(def os) {
 // Setup perflab tests runs
 [true, false].each { isPR ->
     ['Windows_NT'].each { os ->
-		['x64', 'x86'].each { architecture ->
-			def newJob = job(Utilities.getFullJobName(project, "perf_perflab_${os}_${architecture}", isPR)) {
+		['x64', 'x86ryujit', 'jit32'].each { arch ->
```
RussKeldorph: Let's use 'x86' = RyuJIT and x86jit32 = JIT32.

perf.groovy
RussKeldorph: @DrewScoggins, I asked for this to make it easier for us to swap out the JITs.  We'd like to be able to (more) atomically make either JIT the default and not affect these perf jobs.  By putting the logic that determines which JIT is used in external scripts instead of the groovy, it is easier to validate that we did that correctly.

tests/scripts/run-xunit-perf.cmd
```diff
@@ -78,6 +78,11 @@ xcopy /s %BENCHDIR%*.txt . >> %RUNLOG%
 
 set CORE_ROOT=%CORECLR_REPO%\sandbox
 
+@rem setup additional environment variables
+if EXIST %TEST_ENV% (
```
RussKeldorph: I know we are generally super lazy about this, but the lack of quotes here makes me nervous.  My understanding is that the preferred pattern is to refer to command line arguments by %~2 instead of %2, which will remove enclosing quotes if they exist.  Then you should be free to add quotes from from that point forward.  Not a big deal in this case, but feel free to change if you like.

adiaaida:
@dotnet-bot test ci please

adiaaida:
@RussKeldorph I've addressed your comment. PTAL

Issue #9577, opened 2017-02-14
[x86/Linux] Fix x86 ABI handling in CreateInstantiatingILStub


jkotas:
https://github.com/dotnet/coreclr/blob/c055f7ba7d85347e1cb88937964f39d96b12d79e/src/vm/prestub.cpp#L850

The hidden argument should be always passed last for x86, similar to what was done in https://github.com/dotnet/coreclr/pull/9571

The current version won't handle signatures that contain large structures correctly.

seanshpark:
Thank you for the issue. PR posted.

PR #9590, opened 2017-02-14
[x86/Linux] Fix CreateInstantiatingILStub

seanshpark:
The hidden argument should be always passed last for x86


seanshpark:
Fix for #9577, Related with #8848

PR #9593, opened 2017-02-14
Return exit code 0 for long-running/GCSimulator tests in Helix

wtgodbe:
Today we return '2' for skipped GC tests - this is fine in the CI, as we can interpret '2' to mean 'skipped' there. But in our xunit runs, anything non-zero means failure. I couldn't figure out an easy way to indicate to runtest.proj which tests should have their xunit facts marked for skipping, as the generation of the xunit wrappers happens after the test build, and has no knowledge of the contents of the source code for any of these tests (I could have added all 500+ to issues.targets, but that seemed like overkill). Previous to https://github.com/dotnet/coreclr/pull/9510/files, these tests were returning 0 when skipped, and still do on Windows.

I've also filed https://github.com/dotnet/coreclr/pull/9510/files against myself to fix this/issues.targets, and make it so our xunit wrappers actually find all of these tests & mark them as 'Skipped', rather than not running them at all/marking them as "succeeded"

### DIFF
--- a/build-test.cmd
+++ b/build-test.cmd
@@ -261,7 +261,7 @@ if defined __UpdateInvalidPackagesArg (
   set __up=-updateinvalidpackageversions
 )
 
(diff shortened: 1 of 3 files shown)


wtgodbe:
CC @MattGal @gkhanna79 

@sdmaclea @jkotas this will not change the behavior of skipped tests returning '2' instead of '0' when you run locally, or in the CI - this only changes that behavior in Helix, during our official builds, since we run non-windows tests through our xunit wrappers there.

PR #9598, opened 2017-02-15
Optimize span clear

ahsonkhan:
For issue #9161

Basic performance test (similar to smoke tests) just to get a general idea on performance comparing the two:
image

The test is just measuring:
```C#
        const int Iterations = 100000;
        
        {
            Stopwatch stopWatch = new Stopwatch();
            byte[] byteArray = new byte[1000000];
            Span<byte> byteSpan = new Span<byte>(byteArray);
            Console.WriteLine("Iterations: " + Iterations);
            Console.WriteLine("Span Length: " + byteSpan.Length);

            stopWatch.Restart();
            PerfTestClear(byteSpan);
            stopWatch.Stop();
            Output(stopWatch.Elapsed);
        }


        private static void PerfTestClear(Span<byte> byteSpan)
        {
            int count = 0;
            for (int j = 0; j < Iterations; j++)
            {
                count = 0;
                byteSpan.Clear();
            }
        }
```

### DIFF
--- a/src/mscorlib/src/System/Runtime/RuntimeImports.cs
+++ b/src/mscorlib/src/System/Runtime/RuntimeImports.cs
@@ -27,6 +27,6 @@ namespace System.Runtime
         }
 
         [DllImport(JitHelpers.QCall, CharSet = CharSet.Unicode)]
-        extern internal unsafe static void RhZeroMemory(byte* b, nuint byteLength);
+        extern private unsafe static void RhZeroMemory(byte* b, nuint byteLength);
     }
 }
--- a/src/mscorlib/src/System/Runtime/RuntimeImports.cs
+++ b/src/mscorlib/src/System/Runtime/RuntimeImports.cs
@@ -1,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+#if BIT64
+using nuint = System.UInt64;
+#else
+    using nuint = System.UInt32;
+#endif
+
+namespace System.Runtime
+{
+    public class RuntimeImports
+    {
+        // Non-inlinable wrapper around the QCall that avoids poluting the fast path
+        // with P/Invoke prolog/epilog.
+        [MethodImplAttribute(MethodImplOptions.NoInlining)]
+        internal unsafe static void RhZeroMemory(ref byte b, nuint byteLength)
+        {
+            fixed (byte* bytePointer = &b)
+            {
+                RhZeroMemory(bytePointer, byteLength);
+            }
+        }
+
+        [DllImport(JitHelpers.QCall, CharSet = CharSet.Unicode)]
+        extern internal unsafe static void RhZeroMemory(byte* b, nuint byteLength);
+    }
+}
--- a/src/mscorlib/src/System/Span.cs
+++ b/src/mscorlib/src/System/Span.cs
@@ -849,12 +849,12 @@ namespace System
             {
                 if ((Unsafe.As<byte, int>(ref b) & 1) != 0)
                 {
-                    Unsafe.AddNative<byte>(ref b, i) = 0;
+                    Unsafe.AddByteOffset<byte>(ref b, i) = 0;
                     i += 1;
                     if ((Unsafe.As<byte, int>(ref b) & 2) != 0)
                         goto IntAligned;
                 }
-                Unsafe.As<byte, short>(ref Unsafe.AddNative<byte>(ref b, i)) = 0;
+                Unsafe.As<byte, short>(ref Unsafe.AddByteOffset<byte>(ref b, i)) = 0;
                 i += 2;
             }
 
@@ -870,7 +870,7 @@ namespace System
 
             if (((Unsafe.As<byte, int>(ref b) - 1) & 4) == 0)
             {
-                Unsafe.As<byte, int>(ref Unsafe.AddNative<byte>(ref b, i)) = 0;
+                Unsafe.As<byte, int>(ref Unsafe.AddByteOffset<byte>(ref b, i)) = 0;
                 i += 4;
             }
 
@@ -899,13 +899,13 @@ namespace System
                 // we save on writes to b.
 
 #if BIT64
-                Unsafe.As<byte, long>(ref Unsafe.AddNative<byte>(ref b, i)) = 0;
-                Unsafe.As<byte, long>(ref Unsafe.AddNative<byte>(ref b, i + 8)) = 0;
+                Unsafe.As<byte, long>(ref Unsafe.AddByteOffset<byte>(ref b, i)) = 0;
+                Unsafe.As<byte, long>(ref Unsafe.AddByteOffset<byte>(ref b, i + 8)) = 0;
 #else
-                Unsafe.As<byte, int>(ref Unsafe.AddNative<byte>(ref b, i)) = 0;
-                Unsafe.As<byte, int>(ref Unsafe.AddNative<byte>(ref b, i + 4)) = 0;
-                Unsafe.As<byte, int>(ref Unsafe.AddNative<byte>(ref b, i + 8)) = 0;
-                Unsafe.As<byte, int>(ref Unsafe.AddNative<byte>(ref b, i + 12)) = 0;
+                Unsafe.As<byte, int>(ref Unsafe.AddByteOffset<byte>(ref b, i)) = 0;
+                Unsafe.As<byte, int>(ref Unsafe.AddByteOffset<byte>(ref b, i + 4)) = 0;
+                Unsafe.As<byte, int>(ref Unsafe.AddByteOffset<byte>(ref b, i + 8)) = 0;
+                Unsafe.As<byte, int>(ref Unsafe.AddByteOffset<byte>(ref b, i + 12)) = 0;
 #endif
 
                 i = counter;
@@ -918,26 +918,26 @@ namespace System
             if ((byteLength & 8) != 0)
             {
 #if BIT64
-                Unsafe.As<byte, long>(ref Unsafe.AddNative<byte>(ref b, i)) = 0;
+                Unsafe.As<byte, long>(ref Unsafe.AddByteOffset<byte>(ref b, i)) = 0;
 #else
-                Unsafe.As<byte, int>(ref Unsafe.AddNative<byte>(ref b, i)) = 0;
-                Unsafe.As<byte, int>(ref Unsafe.AddNative<byte>(ref b, i + 4)) = 0;
+                Unsafe.As<byte, int>(ref Unsafe.AddByteOffset<byte>(ref b, i)) = 0;
+                Unsafe.As<byte, int>(ref Unsafe.AddByteOffset<byte>(ref b, i + 4)) = 0;
 #endif
                 i += 8;
             }
             if ((byteLength & 4) != 0)
             {
-                Unsafe.As<byte, int>(ref Unsafe.AddNative<byte>(ref b, i)) = 0;
+                Unsafe.As<byte, int>(ref Unsafe.AddByteOffset<byte>(ref b, i)) = 0;
                 i += 4;
             }
             if ((byteLength & 2) != 0)
             {
-                Unsafe.As<byte, short>(ref Unsafe.AddNative<byte>(ref b, i)) = 0;
+                Unsafe.As<byte, short>(ref Unsafe.AddByteOffset<byte>(ref b, i)) = 0;
                 i += 2;
             }
             if ((byteLength & 1) != 0)
             {
-                Unsafe.AddNative<byte>(ref b, i) = 0;
+                Unsafe.AddByteOffset<byte>(ref b, i) = 0;
                 // We're not using i after this, so not needed
                 // i += 1;
             }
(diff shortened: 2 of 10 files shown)


src/mscorlib/src/System/Span.cs
```diff
@@ -625,5 +636,266 @@ internal static unsafe void CopyTo<T>(ref T destination, ref T source, int eleme
...
+
+            fixed (byte* bytePointer = &b)
+            {
+                // Note: It's important that this switch handles lengths at least up to 22.
+                // See notes below near the main loop for why.
+
+                // The switch will be very fast since it can be implemented using a jump
+                // table in assembly. See http://stackoverflow.com/a/449297/4077294 for more info.
+
+                switch (byteLength)
```
morganbr: My general understanding is that memset also does tricks like these. Do you have data showing that it's not faster to just unconditionally call RhZeroMemory?

src/mscorlib/src/System/Span.cs
```diff
@@ -625,5 +636,266 @@ internal static unsafe void CopyTo<T>(ref T destination, ref T source, int eleme
...
+                    case 22:
+                        *(long*)bytePointer = (long)0;
+                        *(long*)(bytePointer + 8) = (long)0;
+                        *(int*)(bytePointer + 16) = (int)0;
+                        *(short*)(bytePointer + 20) = (short)0;
+                        return;
+                }
+
+                // P/Invoke into the native version for large lengths
+                if (byteLength >= 512) goto PInvoke;
```
morganbr: Does this code actually need gotos or would breaking it into methods deliver reasonable performance? It's already a pretty big method.

src/mscorlib/src/System/Runtime/RuntimeImports.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+#if BIT64
+using nuint = System.UInt64;
```
morganbr: As we discussed offline, it might be good to understand why you're using these instead of UIntPtr

src/mscorlib/src/System/Span.cs
```diff
@@ -625,5 +636,266 @@ internal static unsafe void CopyTo<T>(ref T destination, ref T source, int eleme
...
+                        *(int*)bytePointer = (int)0;
+                        *(short*)(bytePointer + 4) = (short)0;
+                        return;
+                    case 7:
+                        *(int*)bytePointer = (int)0;
+                        *(short*)(bytePointer + 4) = (short)0;
+                        *(bytePointer + 6) = (byte)0;
+                        return;
+                    case 8:
+                        *(long*)bytePointer = (long)0;
```
morganbr: Is working with longs still worthwhile on x86/arm32? Their 64-bit handling can be pretty slow

src/mscorlib/src/System/Span.cs
```diff
@@ -625,5 +636,266 @@ internal static unsafe void CopyTo<T>(ref T destination, ref T source, int eleme
...
+                }
+
+                // P/Invoke into the native version for large lengths
+                if (byteLength >= 512) goto PInvoke;
+
+                nuint i = 0; // byte offset at which we're copying
+
+                if (((int)bytePointer & 3) != 0)
+                {
+                    if (((int)bytePointer & 1) != 0)
```
morganbr: Isn't this unconditionally true?

src/mscorlib/src/System/Span.cs
```diff
@@ -625,5 +636,266 @@ internal static unsafe void CopyTo<T>(ref T destination, ref T source, int eleme
...
+                        *(long*)(bytePointer + 8) = (long)0;
+                        *(int*)(bytePointer + 16) = (int)0;
+                        return;
+                    case 21:
+                        *(long*)bytePointer = (long)0;
+                        *(long*)(bytePointer + 8) = (long)0;
+                        *(int*)(bytePointer + 16) = (int)0;
+                        *(bytePointer + 20) = (byte)0;
+                        return;
+                    case 22:
```
morganbr: Why is 22 the right number of cases?

src/mscorlib/src/System/Span.cs
ahsonkhan: I don't think so. This is only checking if the lowest bit is set. Any value XXXX XXX1 would make this false.

src/mscorlib/src/System/Span.cs
ahsonkhan: According to Jan, a lot of tuning went into this: https://github.com/dotnet/coreclr/pull/6638

I borrowed parts of the impl from Buffer Memmove.

src/mscorlib/src/System/Runtime/RuntimeImports.cs
ahsonkhan: cc @jkotas

src/mscorlib/src/System/Span.cs
ahsonkhan: There is overhead in calling RhZeroMemory.

If you notice, for small span sizes (< 100 bytes), it is MUCH faster (almost 2-3x) not to call RhZeroMemory. For span sizes > 1k bytes, the current API is on par with calling RhZeroMemory unconditionally. For between 100-1k bytes, yes, calling RhZeroMemory unconditionally is slightly faster (but no more than 1.5x).

image

src/mscorlib/src/System/Span.cs
jkotas: Agree - if should use BIT64 ifdefs similar to what Buffer.Memmove implementation does.

src/mscorlib/src/System/Runtime/RuntimeImports.cs
jkotas: It is royal pain to do size arithmetic using IntPtr/UIntPtr. We are using the same pattern in number of other places in CoreLib.

jkotas: (There is also discussion about getting better native integer arithmetic support in C#.)

src/mscorlib/src/System/Runtime/RuntimeImports.cs
```diff
@@ -0,0 +1,32 @@
...
+        internal unsafe static void RhZeroMemory(ref byte b, nuint byteLength)
+        {
+            fixed (byte* bytePointer = &b)
+            {
+                RhZeroMemory(bytePointer, byteLength);
+            }
+        }
+
+        [DllImport(JitHelpers.QCall, CharSet = CharSet.Unicode)]
+        extern internal unsafe static void RhZeroMemory(byte* b, nuint byteLength);
```
jkotas: This can be private.

src/mscorlib/src/System/Span.cs
```diff
@@ -625,5 +636,266 @@ internal static unsafe void CopyTo<T>(ref T destination, ref T source, int eleme
                 }
             }
         }
+
+        internal static unsafe void ClearWithoutReferences(ref byte b, nuint byteLength)
+        {
+            if (byteLength == 0)
+                return;
+
+            fixed (byte* bytePointer = &b)
```
jkotas: You should rewrite this using `ref byte`, without pinning. The pinning is unnecessary overhead - especially for small sizes.

src/mscorlib/src/System/Span.cs
```diff
@@ -625,5 +636,266 @@ internal static unsafe void CopyTo<T>(ref T destination, ref T source, int eleme
...
+            if (pointerSizeLength == 0)
+                return;
+
+            // TODO: Perhaps do switch casing to improve small size perf
+
+            nuint i = 0;
+            nuint n = 0;
+            while ((n = i + 8) <= (pointerSizeLength))
+            {
+                Unsafe.Add<IntPtr>(ref ip, (int)i + 0) = default(IntPtr);
```
jkotas: It is not ok to case nuint to int here - it does not work for >4GB sizes. You may need to add internal version of Unsafe.Add that works with nuint.

src/mscorlib/src/System/Span.cs
ahsonkhan: I made the following change and performance degrades severely for when byteLength >= 512.
```C#
if (byteLength >= 512) goto PInvoke;  // From
if (byteLength >= 512) RuntimeImports.RhZeroMemory(ref b, byteLength); // To
```

image

src/mscorlib/src/System/Span.cs
ahsonkhan: How do you increment and cast ref byte to other types? Is it using Unsafe.Add and Unsafe.As<TFrom, TTo>APIs?

For example, if I am not pinning to a byte *, how would the syntax look for let's say:
```C#
*(int*)(bytePointer + 16) = (int)0;
// OR
if ((byteLength & 1) != 0)
```

jkotas: Right. Similar to ClearWithReferences.

src/mscorlib/src/System/Span.cs
jkotas: Didn't you forget `return` after calling RhZeroMemory?

ahsonkhan: Yes, I forgot that.
```C#
if (byteLength >= 512)
{
        RuntimeImports.RhZeroMemory(ref b, byteLength);
        return;
}
```
Calling the method directly doesn't make much difference. We could remove the goto and just call RhZeroMemory.

image

benaadams: Can `JIT_MemSet` be used and does it offer any improvement? (Not sure what path it gets used on) https://github.com/dotnet/coreclr/pull/7198

jkotas: Worth trying. I would call JIT_MemSet from the implementation of the PInvoke if it is available and faster. I would not call it directly from here. The condition for when it is available and faster potentially involves architectures, OS versions, and compiler/C-runtime versions. It is better for such condition to be on C++ side.

KrzysztofCwalina:
It would be good to compare it to Array.Clear too.

jamesqo:
I see you copied some of the code from `Buffer.MemoryCopy`... you may want to take a look at https://github.com/dotnet/coreclr/pull/9143 for the managed part of the implementation. It should be more efficient especially on AMD64 because it writes things out in 16/32/64-byte blocks and does not have the extra branches at the beginning/end to check for alignment. (For lengths >= 1024 it is probably better to stick to the native implementation.)

src/mscorlib/src/System/Span.cs
ahsonkhan: > Can `JIT_MemSet` be used and does it offer any improvement? (Not sure what path it gets used on) https://github.com/dotnet/coreclr/pull/7198

I will file an issue to investigate this optimization outside of this PR.

src/mscorlib/src/System/Span.cs
ahsonkhan: Should I be calling https://github.com/dotnet/corefx/blob/master/src/System.Runtime.CompilerServices.Unsafe/src/System.Runtime.CompilerServices.Unsafe.il#L251?

jkotas: Yes, it would work, but it will require IntPtr casts. Since you need to add it to CoreLib anyway you can as well add versions that just take nuint so that you do not need the casts.

ahsonkhan:
> It would be good to compare it to Array.Clear too.


image

ahsonkhan:
> I see you copied some of the code from `Buffer.MemoryCopy`... you may want to take a look at https://github.com/dotnet/coreclr/pull/9143 for the managed part of the implementation. It should be more efficient especially on AMD64 because it writes things out in 16/32/64-byte blocks and does not have the extra branches at the beginning/end to check for alignment. (For lengths >= 512/1024 it is probably better to stick to the native implementation.)

I will file an issue to investigate this optimization outside of this PR.

src/mscorlib/src/System/Span.cs
ahsonkhan: > You should rewrite this using `ref byte`, without pinning. The pinning is unnecessary overhead - especially for small sizes.

So, removing pinning does improve perf for small sizes, but decreases perf for some larger sizes:
image

I am guessing when you don't pin, the call to RhZeroMemory becomes more expensive:
```C#
            // P/Invoke into the native version for large lengths
            if (byteLength >= 512)
            {
                RuntimeImports.RhZeroMemory(ref b, byteLength);
                return;
            }
```

src/mscorlib/src/System/Span.cs
ahsonkhan: I will file a separate issue to add this, outside of this PR.

jkotas: I do not think you should be checking in functionally incorrect code like this. Some of the corefx tests may actually fail on it.

src/mscorlib/src/System/Span.cs
jkotas: > I am guessing when you don't pin, the call to RhZeroMemory becomes more expensive

This does not make sense. You have bug in your measurement.

jamesqo:
@ahsonkhan Thank you for opening the issue to investigate.

KrzysztofCwalina:
Looks good.

src/mscorlib/src/System/Span.cs
```diff
@@ -625,5 +636,262 @@ internal static unsafe void CopyTo<T>(ref T destination, ref T source, int eleme
...
+                i = n;
+            }
+            if ((i + 1) <= (pointerSizeLength))
+            {
+                Unsafe.Add<IntPtr>(ref ip, (int)i) = default(IntPtr);
+            }
+        }
+
+        [StructLayout(LayoutKind.Sequential, Size = 64)]
+        private struct Reg64 { }
```
jkotas: This is unused

src/mscorlib/src/System/Span.cs
ahsonkhan: Makes sense. Will fix it in here.

src/mscorlib/src/System/Span.cs
ahsonkhan: Done

src/mscorlib/src/System/Span.cs
ahsonkhan: Removed.

src/mscorlib/src/System/Span.cs
ahsonkhan: Yes. I was mistaken.

src/mscorlib/src/System/Span.cs
ahsonkhan: @jkotas, I added a method Unsafe.AddNative
Is this what you had in mind?

jkotas: Yes, it is what I had in mind. It may looks better to call it AddByteOffset, and do all the math in byte offsets.

ahsonkhan:
cc @jkotas good to go?

ahsonkhan:
@dotnet-bote test Windows_NT x64 Release Priority 1 Build and Test

ahsonkhan:

ahsonkhan:

ahsonkhan:

ahsonkhan:

ahsonkhan:

PR #9600, opened 2017-02-15
[x86/Linux] Fix GetModRMOperandValue()

seanshpark:
Getting idiv divisor in GetModRMOperandValue() needs to handle x86 also
This patch will correct address of divisor when it resides in memory
Related unit test case: JIT/Regression/CLR-x86-JIT/V1-M09.5-PDC/b16328/b16328

### DIFF
--- a/src/vm/exceptionhandling.cpp
+++ b/src/vm/exceptionhandling.cpp
@@ -4830,10 +4830,15 @@ DWORD64 GetModRMOperandValue(BYTE rex, BYTE* ip, PCONTEXT pContext, bool is8Bit,
             // Get the value we need from the register.
             //
 
-            // Check for RIP-relative addressing mode.
+            // Check for RIP-relative addressing mode for AMD64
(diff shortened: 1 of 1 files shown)


src/vm/exceptionhandling.cpp
```diff
@@ -4833,7 +4833,11 @@ DWORD64 GetModRMOperandValue(BYTE rex, BYTE* ip, PCONTEXT pContext, bool is8Bit,
             // Check for RIP-relative addressing mode.
```
janvorli: Could you please update the comment to reflect the fact that for x86 it is not RIP-relative addressing mode?

seanshpark: OK

stephentoub:

seanshpark:
@janvorli , could you please take a look again?

seanshpark:
Thank you :)

Issue #9439, opened 2017-02-09
[x86/Linux] Unit test RayTracer fails with segment fault


seanshpark:
RayTracer fails with next message.
```
/RayTracer.sh: line 131:  7878 Segmentation fault      (core dumped) 
```
This unit test passed with temporary 4 byte alignment but failed with new 16 byte alignment.
While walking up the stack, ESP changed to un-anligned value in the JITted code.
Related code was in `System.Threading.TimerQueueTimer:Change()` try/finally block:
```
                try { }
                finally
                {
```

where main block calls `finally` block by `call` instruction but as currenlt catch/finally funclet is not JITted correctly it fautls in some other place with segment fault.

```
G_M59154_IG06:
       90           nop      
       90           nop      
       90           nop      
       8BCC         mov      ecx, esp
       E820000000   call     G_M59154_IG11
       90           nop   
...
...
G_M59154_IG11:

G_M59154_IG12:
       90           nop      
       8B45C4       mov      eax, gword ptr [ebp-3CH]
       8B5508       mov      edx, dword ptr [ebp+08H]
       895024       mov      dword ptr [eax+36], edx
```

seanshpark:
@waterlet, @parjong

seanshpark:
The test passes with this little bit more patch to https://github.com/parjong/coreclr/commit/d205e700da6284851465e6124698a42f6645916e
```
diff --git a/src/jit/codegencommon.cpp b/src/jit/codegencommon.cpp
index b1e474b..a4a530b 100644
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -10286,10 +10286,22 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
 
     compiler->unwindBegProlog();
 
+    inst_RV(INS_push, REG_FPBASE, TYP_REF);
+    compiler->unwindPush(REG_FPBASE);
+    inst_RV_IV(INS_sub, REG_SPBASE, 8, EA_PTRSIZE);
+
     // TODO Save callee-saved registers
 
     // This is the end of the OS-reported prolog for purposes of unwinding
     compiler->unwindEndProlog();
+
+    if (block->bbCatchTyp != BBCT_FINALLY && block->bbCatchTyp != BBCT_FAULT)
+    {
+        getEmitter()->emitIns_R_AR(INS_mov, EA_PTRSIZE, REG_EXCEPTION_OBJECT, REG_SPBASE, 12);
+
[...]

Issue #9265, opened 2017-02-02
x86/Linux progress


parjong:
I'm opening an issue to track x86/Linux progress with respect to the regression tests.

Here is the current status on Ubuntu 14.04 (inside Docker Container Ubuntu 16.04 x64):
```
=======================
     Test Results
=======================
# CoreCLR Bin Dir  : 
# Tests Discovered : 7027
# Passed           : 5592
# Failed           : 1156
# Skipped          : 279
=======================
```

I used 63607d8e657 with https://github.com/parjong/coreclr/tree/fix/x86_4byte_alignment and #9261 to get the above result.

parjong:
\CC @seanshpark @wateret

parjong:
63607d8e657 (with alignment workaround) shows the following result (
[xUnit report](https://github.com/dotnet/coreclr/files/749285/coreclrtests.txt))
:
```
=======================
     Test Results
=======================
# CoreCLR Bin Dir  :
# Tests Discovered : 7027
# Passed           : 5892
# Failed           : 856
# Skipped          : 279
=======================
53 minutes and 59 seconds taken to run CoreCLR tests.
```

#9121 seems to resolve several GC and JIT related failures.

parjong:
Here is the result from 2ecadf5d1ce (without any additional patch):
```
=======================
     Test Results
=======================
# CoreCLR Bin Dir  :
# Tests Discovered : 7027
# Passed           : 5874
# Failed           : 874
# Skipped          : 279
=======================
65 minutes and 29 seconds taken to run CoreCLR tests.
```
Recently merged #8849 eliminates the need for alignment workaround. There is a small increase in the number of failed tests, and most of them are caused by stack smashing. (Incorrect funclet prolog/epilog may cause this stack smashing issue).

janvorli:
@parjong, @seanshpark, @wateret - I was trying to build and run dotnet for x86 Linux today, using the current state of the master branch, but I was unable to make it work. So I was wondering if you could share a list of steps to successfuly build and gather all parts of the dotnet core the way you do it. 
I have tried to build the coreclr and corefx native binaries both using the cross build and a build inside of a docker container with x86 ubuntu 14.04 (running the container on x64 Ubuntu 14.04). I have built the corefx managed assemblies on my x64 ubuntu. 
When I try to run a simple hello world like app inside of the docker container, I get a strange assertion in the native runtime even before the coreclr_initialize completes.

parjong:
[...]

PR #9601, opened 2017-02-15
[x86/Linux] Initial patch for EH funclet

wateret:
Generate a simple EH funclet frame and
Support SP-based stack unwinding for funclets.

### DIFF
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -10302,6 +10302,21 @@ void CodeGen::genCaptureFuncletPrologEpilogInfo()
 /*****************************************************************************
  *
  *  Generates code for an EH funclet prolog.
+ *
+ *  Funclets have the following incoming arguments:
+ *
+ *      catch/filter-handler: ecx = CallerSP, edx = the exception object that was caught (see GT_CATCH_ARG)
+ *      filter:               ecx = CallerSP, edx = the exception object to filter (see GT_CATCH_ARG)
+ *      finally/fault:        ecx = CallerSP
+ *
+ *  Funclets set the following registers on exit:
+ *
+ *      catch/filter-handler: eax = the address at which execution should resume (see BBJ_EHCATCHRET)
+ *      filter:               eax = non-zero if the handler should handle the exception, zero otherwise (see GT_RETFILT)
+ *      finally/fault:        none
+ *
+ *  Funclet prolog/epilog sequence and funclet frame layout are TBD.
+ *
  */
 
 void CodeGen::genFuncletProlog(BasicBlock* block)
@@ -10325,21 +10340,17 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
     // This is the end of the OS-reported prolog for purposes of unwinding
     compiler->unwindEndProlog();
 
-    if (block->bbCatchTyp != BBCT_FINALLY && block->bbCatchTyp != BBCT_FAULT)
+    if (handlerGetsXcptnObj(block->bbCatchTyp))
     {
-        getEmitter()->emitIns_R_AR(INS_mov, EA_PTRSIZE, REG_EXCEPTION_OBJECT, REG_SPBASE, 12);
+        getEmitter()->emitIns_R_R(INS_mov, EA_PTRSIZE, REG_EXCEPTION_OBJECT, REG_ARG_1);
+    }
 
-        getEmitter()->emitIns_R_AR(INS_mov, EA_PTRSIZE, REG_FPBASE, REG_SPBASE, 8);
-        getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_FPBASE, REG_FPBASE, -8);
-    }
-    else if (block->bbCatchTyp == BBCT_FAULT)
-    {
-        getEmitter()->emitIns_R_AR(INS_mov, EA_PTRSIZE, REG_FPBASE, REG_SPBASE, 8);
-        getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_FPBASE, REG_FPBASE, -8);
-    }
+    // TODO We will need changes here when we introduce PSPSym
+    getEmitter()->emitIns_R_R(INS_mov, EA_PTRSIZE, REG_FPBASE, REG_ARG_0);
+    getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_FPBASE, REG_FPBASE, -8);
 
     // Add a padding (for 16-byte alignment)
-    //  See #9439 for details
+    //  See https://github.com/dotnet/coreclr/issues/9439 for details
     inst_RV_IV(INS_sub, REG_SPBASE, 8, EA_PTRSIZE);
 
     // We've modified EBP, but not really. Say that we haven't...
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -10333,9 +10333,6 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
 
     compiler->unwindBegProlog();
 
-    inst_RV(INS_push, REG_FPBASE, TYP_REF);
-    compiler->unwindPush(REG_FPBASE);
-
     // This is the end of the OS-reported prolog for purposes of unwinding
     compiler->unwindEndProlog();
 
@@ -10364,7 +10361,6 @@ void CodeGen::genFuncletEpilog()
     // Revert a padding that was added for 16-byte alignment
     inst_RV_IV(INS_add, REG_SPBASE, 8, EA_PTRSIZE);
 
-    inst_RV(INS_pop, REG_FPBASE, TYP_REF);
     instGen_Return(0);
 }
 
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -10317,10 +10317,33 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
 
     compiler->unwindBegProlog();
 
+    inst_RV(INS_push, REG_FPBASE, TYP_REF);
+    compiler->unwindPush(REG_FPBASE);
+
     // TODO Save callee-saved registers
 
     // This is the end of the OS-reported prolog for purposes of unwinding
     compiler->unwindEndProlog();
+
+    if (block->bbCatchTyp != BBCT_FINALLY && block->bbCatchTyp != BBCT_FAULT)
+    {
+        getEmitter()->emitIns_R_AR(INS_mov, EA_PTRSIZE, REG_EXCEPTION_OBJECT, REG_SPBASE, 12);
+
+        getEmitter()->emitIns_R_AR(INS_mov, EA_PTRSIZE, REG_FPBASE, REG_SPBASE, 8);
+        getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_FPBASE, REG_FPBASE, -8);
+    }
+    else if (block->bbCatchTyp == BBCT_FAULT)
+    {
+        getEmitter()->emitIns_R_AR(INS_mov, EA_PTRSIZE, REG_FPBASE, REG_SPBASE, 8);
+        getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_FPBASE, REG_FPBASE, -8);
+    }
+
+    // Add a padding (for 16-byte alignment)
+    //  See #9439 for details
+    inst_RV_IV(INS_sub, REG_SPBASE, 8, EA_PTRSIZE);
+
+    // We've modified EBP, but not really. Say that we haven't...
+    regSet.rsRemoveRegsModified(RBM_FPBASE);
 }
 
 /*****************************************************************************
@@ -10341,6 +10364,9 @@ void CodeGen::genFuncletEpilog()
 
     // TODO Restore callee-saved registers
 
+    inst_RV_IV(INS_add, REG_SPBASE, 8, EA_PTRSIZE);
+
+    inst_RV(INS_pop, REG_FPBASE, TYP_REF);
     instGen_Return(0);
 }
 
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -10339,7 +10339,7 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
     // TODO We may need EBP restore sequence here if we introduce PSPSym
 
     // Add a padding for 16-byte alignment
-    inst_RV_IV(INS_sub, REG_SPBASE, 8, EA_PTRSIZE);
+    inst_RV_IV(INS_sub, REG_SPBASE, 12, EA_PTRSIZE);
 }
 
 /*****************************************************************************
@@ -10359,7 +10359,7 @@ void CodeGen::genFuncletEpilog()
     ScopedSetVariable<bool> _setGeneratingEpilog(&compiler->compGeneratingEpilog, true);
 
     // Revert a padding that was added for 16-byte alignment
-    inst_RV_IV(INS_add, REG_SPBASE, 8, EA_PTRSIZE);
+    inst_RV_IV(INS_add, REG_SPBASE, 12, EA_PTRSIZE);
 
     instGen_Return(0);
 }
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -10303,11 +10303,12 @@ void CodeGen::genCaptureFuncletPrologEpilogInfo()
  *
  *  Generates code for an EH funclet prolog.
  *
+ *
  *  Funclets have the following incoming arguments:
  *
- *      catch/filter-handler: ecx = CallerSP, edx = the exception object that was caught (see GT_CATCH_ARG)
- *      filter:               ecx = CallerSP, edx = the exception object to filter (see GT_CATCH_ARG)
- *      finally/fault:        ecx = CallerSP
+ *      catch/filter-handler: eax = the exception object that was caught (see GT_CATCH_ARG), ecx = FP
+ *      filter:               eax = the exception object that was caught (see GT_CATCH_ARG), ecx = FP
+ *      finally/fault:        ecx = FP
  *
  *  Funclets set the following registers on exit:
  *
@@ -10340,17 +10341,11 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
     // This is the end of the OS-reported prolog for purposes of unwinding
     compiler->unwindEndProlog();
 
-    if (handlerGetsXcptnObj(block->bbCatchTyp))
-    {
-        getEmitter()->emitIns_R_R(INS_mov, EA_PTRSIZE, REG_EXCEPTION_OBJECT, REG_ARG_1);
-    }
-
+    // Restore frame pointer
     // TODO We will need changes here when we introduce PSPSym
-    getEmitter()->emitIns_R_R(INS_mov, EA_PTRSIZE, REG_FPBASE, REG_ARG_0);
-    getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_FPBASE, REG_FPBASE, -8);
+    inst_RV_RV(INS_mov, REG_FPBASE, REG_ARG_0);
 
-    // Add a padding (for 16-byte alignment)
-    //  See https://github.com/dotnet/coreclr/issues/9439 for details
+    // Add a padding for 16-byte alignment
     inst_RV_IV(INS_sub, REG_SPBASE, 8, EA_PTRSIZE);
 
     // We've modified EBP, but not really. Say that we haven't...
@@ -10375,6 +10370,7 @@ void CodeGen::genFuncletEpilog()
 
     // TODO Restore callee-saved registers
 
+    // Revert a padding that was added for 16-byte alignment
     inst_RV_IV(INS_add, REG_SPBASE, 8, EA_PTRSIZE);
 
     inst_RV(INS_pop, REG_FPBASE, TYP_REF);
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -10306,9 +10306,9 @@ void CodeGen::genCaptureFuncletPrologEpilogInfo()
  *
  *  Funclets have the following incoming arguments:
  *
- *      catch/filter-handler: eax = the exception object that was caught (see GT_CATCH_ARG), ecx = FP
- *      filter:               eax = the exception object that was caught (see GT_CATCH_ARG), ecx = FP
- *      finally/fault:        ecx = FP
+ *      catch/filter-handler: eax = the exception object that was caught (see GT_CATCH_ARG)
+ *      filter:               eax = the exception object that was caught (see GT_CATCH_ARG)
+ *      finally/fault:        none
  *
  *  Funclets set the following registers on exit:
  *
@@ -10336,20 +10336,13 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
     inst_RV(INS_push, REG_FPBASE, TYP_REF);
     compiler->unwindPush(REG_FPBASE);
 
-    // TODO Save callee-saved registers
-
     // This is the end of the OS-reported prolog for purposes of unwinding
     compiler->unwindEndProlog();
 
-    // Restore frame pointer
-    // TODO We will need changes here when we introduce PSPSym
-    inst_RV_RV(INS_mov, REG_FPBASE, REG_ARG_0);
+    // TODO We may need EBP restore sequence here if we introduce PSPSym
 
     // Add a padding for 16-byte alignment
     inst_RV_IV(INS_sub, REG_SPBASE, 8, EA_PTRSIZE);
-
-    // We've modified EBP, but not really. Say that we haven't...
-    regSet.rsRemoveRegsModified(RBM_FPBASE);
 }
 
 /*****************************************************************************
@@ -10367,8 +10360,6 @@ void CodeGen::genFuncletEpilog()
 #endif
 
     ScopedSetVariable<bool> _setGeneratingEpilog(&compiler->compGeneratingEpilog, true);
-
-    // TODO Restore callee-saved registers
 
     // Revert a padding that was added for 16-byte alignment
     inst_RV_IV(INS_add, REG_SPBASE, 8, EA_PTRSIZE);
--- a/src/jit/codegenxarch.cpp
+++ b/src/jit/codegenxarch.cpp
@@ -241,11 +241,9 @@ BasicBlock* CodeGen::genCallFinally(BasicBlock* block)
     if ((compiler->lvaPSPSym == BAD_VAR_NUM) ||
         (!compiler->compLocallocUsed && (compiler->funCurrentFunc()->funKind == FUNC_ROOT)))
     {
-#ifdef UNIX_X86_ABI
-        inst_RV_RV(INS_mov, REG_ARG_0, REG_FPBASE, TYP_I_IMPL); // FP for UNIX_X86
-#else // UNIX_X86_ABI
-        inst_RV_RV(INS_mov, REG_ARG_0, REG_SPBASE, TYP_I_IMPL); // SP for all except UNIX_X86
-#endif
+#ifndef UNIX_X86_ABI
+        inst_RV_RV(INS_mov, REG_ARG_0, REG_SPBASE, TYP_I_IMPL);
+#endif // !UNIX_X86_ABI
     }
     else
     {
--- a/src/jit/codegenxarch.cpp
+++ b/src/jit/codegenxarch.cpp
@@ -241,10 +241,10 @@ BasicBlock* CodeGen::genCallFinally(BasicBlock* block)
     if ((compiler->lvaPSPSym == BAD_VAR_NUM) ||
         (!compiler->compLocallocUsed && (compiler->funCurrentFunc()->funKind == FUNC_ROOT)))
     {
-#ifdef _TARGET_X86_
-        getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_ARG_0, REG_FPBASE, 8); // caller-SP for x86
-#else
-        inst_RV_RV(INS_mov, REG_ARG_0, REG_SPBASE, TYP_I_IMPL);
+#ifdef UNIX_X86_ABI
+        inst_RV_RV(INS_mov, REG_ARG_0, REG_FPBASE, TYP_I_IMPL); // FP for UNIX_X86
+#else // UNIX_X86_ABI
+        inst_RV_RV(INS_mov, REG_ARG_0, REG_SPBASE, TYP_I_IMPL); // SP for all except UNIX_X86
 #endif
     }
     else
--- a/src/jit/codegenxarch.cpp
+++ b/src/jit/codegenxarch.cpp
@@ -241,7 +241,11 @@ BasicBlock* CodeGen::genCallFinally(BasicBlock* block)
     if ((compiler->lvaPSPSym == BAD_VAR_NUM) ||
         (!compiler->compLocallocUsed && (compiler->funCurrentFunc()->funKind == FUNC_ROOT)))
     {
+#ifdef _TARGET_X86_
+        getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_ARG_0, REG_FPBASE, 8); // caller-SP for x86
+#else
         inst_RV_RV(INS_mov, REG_ARG_0, REG_SPBASE, TYP_I_IMPL);
+#endif
     }
     else
     {
--- a/src/pal/inc/unixasmmacrosx86.inc
+++ b/src/pal/inc/unixasmmacrosx86.inc
@@ -67,8 +67,6 @@ C_FUNC(\Name\()_End):
 .endm
 
 .macro ESP_PROLOG_BEG
-        push ebp
-        mov ebp, esp
 .endm
 
 .macro ESP_PROLOG_END
@@ -78,7 +76,6 @@ C_FUNC(\Name\()_End):
 .endm
 
 .macro ESP_EPILOG_END
-        pop ebp
 .endm
 
 .macro PREPARE_EXTERNAL_VAR Name, Reg
--- a/src/pal/inc/unixasmmacrosx86.inc
+++ b/src/pal/inc/unixasmmacrosx86.inc
@@ -66,6 +66,21 @@ C_FUNC(\Name\()_End):
         pop ebp
 .endm
 
+.macro ESP_PROLOG_BEG
+        push ebp
+        mov ebp, esp
+.endm
+
+.macro ESP_PROLOG_END
+.endm
+
+.macro ESP_EPILOG_BEG
+.endm
+
+.macro ESP_EPILOG_END
+        pop ebp
+.endm
+
 .macro PREPARE_EXTERNAL_VAR Name, Reg
 .att_syntax
     call  0f
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -3808,9 +3808,13 @@ bool UnwindEbpDoubleAlignFrame(
         TADDR baseSP;
 
 #ifdef WIN64EXCEPTIONS
+        // Funclets' frame pointers(EBP) are always restored so they can access to main function's local variables.
+        // Therefore the value of EBP is invalid for unwinder so we should use ESP instead.
+        // If funclet frame layout is changed from CodeGen::genFuncletProlog() and genFuncletEpilog(),
+        // we need to change here accordingly.
         if (pCodeInfo->IsFunclet())
         {
-            baseSP = curESP + 8;
+            baseSP = curESP + 8; // padding for 16byte stack alignment allocated in genFuncletProlog()
 
             pContext->SetEbpLocation(PTR_DWORD(baseSP));
 
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -3810,8 +3810,9 @@ bool UnwindEbpDoubleAlignFrame(
 #ifdef WIN64EXCEPTIONS
         // Funclets' frame pointers(EBP) are always restored so they can access to main function's local variables.
         // Therefore the value of EBP is invalid for unwinder so we should use ESP instead.
-        // If funclet frame layout is changed from CodeGen::genFuncletProlog() and genFuncletEpilog(),
-        // we need to change here accordingly.
+        // TODO If funclet frame layout is changed from CodeGen::genFuncletProlog() and genFuncletEpilog(),
+        //      we need to change here accordingly. It is likely to have changes when introducing PSPSym.
+        // TODO Currently we assume that ESP of funclet frames is always fixed but actually it could change.
         if (pCodeInfo->IsFunclet())
         {
             baseSP = curESP + 8; // padding for 16byte stack alignment allocated in genFuncletProlog()
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -3815,11 +3815,7 @@ bool UnwindEbpDoubleAlignFrame(
         // TODO Currently we assume that ESP of funclet frames is always fixed but actually it could change.
         if (pCodeInfo->IsFunclet())
         {
-            baseSP = curESP + 8; // padding for 16byte stack alignment allocated in genFuncletProlog()
-
-            pContext->SetEbpLocation(PTR_DWORD(baseSP));
-
-            baseSP += sizeof(TADDR);
+            baseSP = curESP + 12; // padding for 16byte stack alignment allocated in genFuncletProlog()
 
             pContext->PCTAddr = baseSP;
             pContext->ControlPC = *PTR_PCODE(pContext->PCTAddr);
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -3786,10 +3786,11 @@ void UnwindEbpDoubleAlignFrameProlog(
 /*****************************************************************************/
 
 bool UnwindEbpDoubleAlignFrame(
-        PREGDISPLAY pContext, 
-        hdrInfo * info, 
-        PTR_CBYTE methodStart, 
-        unsigned flags,
+        PREGDISPLAY     pContext,
+        EECodeInfo     *pCodeInfo,
+        hdrInfo        *info,
+        PTR_CBYTE       methodStart,
+        unsigned        flags,
         StackwalkCacheUnwindInfo  *pUnwindInfo) // out-only, perf improvement
 {
     LIMITED_METHOD_CONTRACT;
@@ -3805,6 +3806,24 @@ bool UnwindEbpDoubleAlignFrame(
     if (info->handlers && info->prologOffs == hdrInfo::NOT_IN_PROLOG)
     {
         TADDR baseSP;
+
+#ifdef WIN64EXCEPTIONS
+        if (pCodeInfo->IsFunclet())
+        {
+            baseSP = curESP + 8;
+
+            pContext->SetEbpLocation(PTR_DWORD(baseSP));
+
+            baseSP += sizeof(TADDR);
+
+            pContext->PCTAddr = baseSP;
+            pContext->ControlPC = *PTR_PCODE(pContext->PCTAddr);
+
+            pContext->SP = (DWORD)(baseSP + sizeof(TADDR));
+
+            return true;
+        }
+#else // WIN64EXCEPTIONS
 
         FrameType frameType = GetHandlerFrameInfo(info, curEBP,
                                                   curESP, (DWORD) IGNORE_VAL,
@@ -3860,6 +3879,7 @@ bool UnwindEbpDoubleAlignFrame(
 
             return true;
         }
+#endif // !WIN64EXCEPTIONS
     }
 
     //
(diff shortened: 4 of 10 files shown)


wateret:
This patch is originally from parjong@a992b93. Before doing something like #9489, I thought we need sp-based stack unwinding and a basic funclet frame. At least on x86/linux this makes more unit test success with no regression.

cc: @parjong @jkotas

janvorli:
cc: @dotnet/jit-contrib

src/jit/codegencommon.cpp
```diff
@@ -10317,10 +10317,33 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
...
 
+    inst_RV(INS_push, REG_FPBASE, TYP_REF);
+    compiler->unwindPush(REG_FPBASE);
+
     // TODO Save callee-saved registers
 
     // This is the end of the OS-reported prolog for purposes of unwinding
     compiler->unwindEndProlog();
+
+    if (block->bbCatchTyp != BBCT_FINALLY && block->bbCatchTyp != BBCT_FAULT)
```
janvorli: It would be easier to read if you used a positive condition (listing catch types for which the code is generated). I guess it would be just BBCT_FILTER?

src/jit/codegencommon.cpp
```diff
@@ -10317,10 +10317,33 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
...
+        getEmitter()->emitIns_R_AR(INS_mov, EA_PTRSIZE, REG_EXCEPTION_OBJECT, REG_SPBASE, 12);
+
+        getEmitter()->emitIns_R_AR(INS_mov, EA_PTRSIZE, REG_FPBASE, REG_SPBASE, 8);
+        getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_FPBASE, REG_FPBASE, -8);
+    }
+    else if (block->bbCatchTyp == BBCT_FAULT)
+    {
+        getEmitter()->emitIns_R_AR(INS_mov, EA_PTRSIZE, REG_FPBASE, REG_SPBASE, 8);
+        getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_FPBASE, REG_FPBASE, -8);
+    }
```
janvorli: What if it is BBCT_FINALLY? Is it correct that we don't generate anything here?

src/vm/eetwain.cpp
```diff
@@ -3805,6 +3806,24 @@ bool UnwindEbpDoubleAlignFrame(
     {
         TADDR baseSP;
 
+#ifdef WIN64EXCEPTIONS
+        if (pCodeInfo->IsFunclet())
+        {
+            baseSP = curESP + 8;
```
janvorli: Could you please add comment on what the constant "8" means?

src/jit/codegencommon.cpp
```diff
@@ -10317,10 +10317,33 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
...
+
+    // Add a padding (for 16-byte alignment)
+    //  See #9439 for details
```
janvorli: Please use full URL of the github issue (`https://github.com/dotnet/coreclr/issues/9439`)

janvorli:
@wateret It would be great to add comment to the CodeGen::genFuncletProlog describing the details of what arguments are passed to each filter type, what does the stack frame look like etc. Similar to what we have for this function for all other architectures.

src/jit/codegencommon.cpp
wateret: This `else if` block was for BBCT_FINALLY as well but it somehow crashes. It cannot run any managed binaries and it seems like it crashes from mscorlib(AppDomain or something). I didn't figure out why yet.

src/jit/codegencommon.cpp
wateret: Wouldn't it be both BBCT_FILTER and BBCT_FILTER_HANDLER?

janvorli: The BBCT_FILTER_HANDLER represents the handler that's called when the filter decides to handle the exception. So I would think that the BBCT_FILTER_HANDLER should be handled like the BBCT_CATCH.

wateret: The only difference between if and else block is the presence of exception object. So I think we need it for all types of funclets other than FINALLY and FAULT.

janvorli: Why would the BBCT_FILTER_HANDLER be different from the BBCT_FAULT? It looks like it is essentially the same thing

stephentoub:

src/vm/eetwain.cpp
BruceForstall: I'd also like to see a comment that describes what you expect the stack frame to look like, since you obviously have certain expectations about the stack layout.

---
In reply to: [101234990](https://github.com/dotnet/coreclr/pull/9601#discussion_r101234990) [](ancestors = 101234990)

src/jit/codegencommon.cpp
BruceForstall: The code needs to stand on its own without reference to GitHub. So please include all and any comments required to document this requirement. A temporary reference is ok, but I would expect it to be removed when the issue is fixed, and not remain longer than, say, one month.

---
In reply to: [101235235](https://github.com/dotnet/coreclr/pull/9601#discussion_r101235235) [](ancestors = 101235235)

src/jit/codegencommon.cpp
```diff
@@ -10317,10 +10317,33 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
...
+
     // TODO Save callee-saved registers
 
     // This is the end of the OS-reported prolog for purposes of unwinding
     compiler->unwindEndProlog();
+
+    if (block->bbCatchTyp != BBCT_FINALLY && block->bbCatchTyp != BBCT_FAULT)
+    {
+        getEmitter()->emitIns_R_AR(INS_mov, EA_PTRSIZE, REG_EXCEPTION_OBJECT, REG_SPBASE, 12);
+
```
BruceForstall: This "mov" is the only difference between the funclet prolog types, right? This should be determined by `if handlerGetsXcptnObj(block->bbCatchTyp)`.

parjong: @BruceForstall Thanks you for comment! It works well.
@wateret Please check the updated draft [here](https://github.com/parjong/coreclr/commit/b6e59da1c17ae4eb4f46549a8bca8c5a7e5164ca)

src/jit/codegencommon.cpp
parjong: @janvorli @wateret It turns out that we need to revise genCallFinally accordingly to support ``BBCT_FINALLY``. The current implementation generates the following ``call-finally`` sequence:
```
       90           nop
       90           nop
       8BCC         mov      ecx, esp
       E80B000000   call     [Finally]
       90           nop
```
The current implementation passes ``Initial-SP`` via ``ecx``, but EE passes ``Caller-SP`` via stack. This difference causes segmentation fault.

wateret:
@janvorli @BruceForstall  
I added a commit that can handle BBCT_FINALLY. I had to fix `genCallFinally` as @parjong pointed out.

src/jit/codegencommon.cpp
```diff
@@ -10317,10 +10332,29 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
...
     compiler->unwindEndProlog();
+
+    if (handlerGetsXcptnObj(block->bbCatchTyp))
+    {
+        getEmitter()->emitIns_R_R(INS_mov, EA_PTRSIZE, REG_EXCEPTION_OBJECT, REG_ARG_1);
+    }
+
+    // TODO We will need changes here when we introduce PSPSym
+    getEmitter()->emitIns_R_R(INS_mov, EA_PTRSIZE, REG_FPBASE, REG_ARG_0);
+    getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_FPBASE, REG_FPBASE, -8);
```
BruceForstall: Couldn't you eliminate the first line and make the second:
```
getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_FPBASE, REG_ARG_0, -8);
```
?

BruceForstall: Please add a comment here about the `-8`. I assume this is because you assume the main function prolog is:
```
pushd ebp
mov ebp,esp
```
Thus, CallerSP is exactly 8 bytes above, due to the pushed return address and pushed ebp.

Note that this is the reason we have the funclet frame layout diagram comments. Also, these constants are encapsulated in other architectures with the `CodeGen::genCaptureFuncletPrologEpilogInfo()` function and descriptive names.

src/jit/codegenxarch.cpp
```diff
@@ -241,7 +241,11 @@ BasicBlock* CodeGen::genCallFinally(BasicBlock* block)
     if ((compiler->lvaPSPSym == BAD_VAR_NUM) ||
         (!compiler->compLocallocUsed && (compiler->funCurrentFunc()->funKind == FUNC_ROOT)))
     {
+#ifdef _TARGET_X86_
+        getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_ARG_0, REG_FPBASE, 8); // caller-SP for x86
```
BruceForstall: This needs to use `-genCallerSPtoFPdelta()`, not "8"

src/vm/eetwain.cpp
```diff
@@ -3805,6 +3806,28 @@ bool UnwindEbpDoubleAlignFrame(
...
         TADDR baseSP;
 
+#ifdef WIN64EXCEPTIONS
+        // Funclets' frame pointers(EBP) are always restored so they can access to main function's local variables.
+        // Therefore the value of EBP is invalid for unwinder so we should use ESP instead.
+        // If funclet frame layout is changed from CodeGen::genFuncletProlog() and genFuncletEpilog(),
+        // we need to change here accordingly.
+        if (pCodeInfo->IsFunclet())
+        {
+            baseSP = curESP + 8; // padding for 16byte stack alignment allocated in genFuncletProlog()
```
BruceForstall: What if we are trying to unwind in a funclet and we have pushed a lot of arguments in preparation for making a call, but haven't yet made the call? Then the curESP will not be what you expect.

src/jit/codegencommon.cpp
```diff
@@ -10341,6 +10375,9 @@ void CodeGen::genFuncletEpilog()
 
     // TODO Restore callee-saved registers
 
+    inst_RV_IV(INS_add, REG_SPBASE, 8, EA_PTRSIZE);
```
BruceForstall: Add a comment that this is to revert the alignment adjustment.

src/vm/eetwain.cpp
parjong: @BruceForstall Thank you for pointing it out! The current problem is that we do not have a frame pointer for this funclet frame (``ebp`` should be restored as that of main function as I understand). 

Could you let me know if there is any way to reserve any register (except ebp)? It would resolve issues from incorrect ESP if we could reserve any registers for a frame pointer for funclet frames.

BruceForstall: All the other funclet-based architectures also have fixed sized frames -- they pre-allocate the outgoing argument space for calls, and localloc/_alloca is not allowed in EH handlers / funclets by the CLI spec. So this is new territory.

One possibility is to set up ebp as a normal frame pointer in funclets, and then reserve and use some other register as the base pointer for accessing locals within funclet code generation. That would be a pretty pervasive change.

Another possibility would be to change to fixed frames. That doesn't seem great.

Or, as you suggest, you could reserve another (callee-saved) register for storing the funclet frame pointer. But that then becomes part of the ABI contract, as the VM would need to know which register that is.

There are so few registers already on x86 that reserving (another) one in funclets is going to hurt code quality.

I can't think of any obvious good solution right now.

parjong: @BruceForstall How about ``ecx``? Is it supposed to be available in funclet?

jkotas: > What if we are trying to unwind in a funclet and we have pushed a lot of arguments in preparation for making a call,

The current Windows x86 can handle this situation fine, with just ESP and EBP. What is different about the Linux x86 funclets that it does not work?

parjong: @BruceForstall @jkotas That is also interesting for me. Could you let me know if there is any clue? As I understand, we will encounter the same issue if GC happens during pushing argument.

janvorli: @parjong GC cannot happen at any arbitrary point. It can only happen in interruptible regions of functions. When a thread is suspended for GC, a check if the thread is at GC safe point is made first. If it is not at a GC save point (an interruptible region of a function), then we hijack the return address of the function (replace it with an address of an asm hijack helper), resume the thread and wait until it either returns to the hijack helper, synchronizes with GC or we timeout (e.g. if the function we've hijacked didn't return for a long time since it called another one with a long running loop) and retry the suspension.
Only if it is at a GC safe point, we allow the GC stack walk happen from that place.
I belive prolog and epilog are non-interruptible.

parjong: @janvorli Thank you for explanation. Is funclet body (especially function call sequnece in it) also non-interruptible? If not, it seems that GC would get stuck for the following case (with the above unwind routine):
```
push STK_ARG_2
<< GC >>
push STK_ARG _1
call FUNC
```
As I understand, x86/Windows also seems to get stuck for the above case as unwinder couldn't figure out where saved ``ebp`` is, but it seems to work. That is an interesting point for me.

janvorli: @parjong a function can have multiple regions of interruptibility / non-interruptibility. You can see that from a jitdump of a function.
You can see something like this by the end of the jitdump (COMPlus_JitDump=methodname) output:
Defining interruptible range: [0x1b, 0xdc).
Defining interruptible range: [0xf2, 0x136).
Defining interruptible range: [0x146, 0x17e).
This comes from an old jitdump that I have captured about a year ago when debugging something. See the code of a funclet below that's related to the last of the interruptible ranges. You can see that the whole funclet is interruptible except for its prologue and epilogue.
You can generate a jitdump for a function that has try / catch on x86 to see which regions are interruptible.
```
G_M22228_IG11:        ; func=02, offs=000136H, size=0010H, gcVars=0000000000007F49 {V00 V02 V03 V05 V06 V07 V08 V11 V12 V13}, gcrefRegs=00000040 {rsi}, byrefRegs=00000000 {}, gcvars, byref, funclet prolog, nogc

IN005a: 000136 push     rbp
IN005b: 000137 sub      rsp, 16
IN005c: 00013B mov      rbp, qword ptr [rdi]
IN005d: 00013E mov      qword ptr [rsp], rbp
IN005e: 000142 lea      rbp, [rbp+70H]

G_M22228_IG12:        ; offs=000146H, size=0038H, gcVars=0000000000007F49 {V00 V02 V03 V05 V06 V07 V08 V11 V12 V13}, gcrefRegs=00000040 {rsi}, byrefRegs=00000000 {}, gcvars, byref, isz

IN003c: 000146 mov      gword ptr [V09 rbp-48H], rsi
IN003d: 00014A mov      rsi, gword ptr [V09 rbp-48H]
IN003e: 00014E mov      gword ptr [V03 rbp-20H], rsi
IN003f: 000152 nop      
IN0040: 000153 mov      rsi, 0x7F00100010F0
IN0041: 00015D mov      rsi, gword ptr [rsi]
IN0042: 000160 mov      rdi, gword ptr [V00 rbp-10H]
IN0043: 000164 call     System.String:Concat(ref,ref):ref
IN0044: 000169 mov      gword ptr [V10 rbp-50H], rax
IN0045: 00016D mov      rsi, gword ptr [V10 rbp-50H]
IN0046: 000171 mov      gword ptr [V00 rbp-10H], rsi
IN0047: 000175 nop      
IN0048: 000176 nop      
IN0049: 000177 lea      rax, G_M22228_IG06

G_M22228_IG13:        ; offs=00017EH, size=0006H, funclet epilog, nogc, emitadd

IN005f: 00017E add      rsp, 16
IN0060: 000182 pop      rbp
IN0061: 000183 ret 
```

jkotas: > If not, it seems that GC would get stuck for the following case (with the above unwind routine):
>
> ```
> push STK_ARG_2
> << GC >>
> push STK_ARG _1
> call FUNC
> ```

The x86 gc and unwind info keeps track of how much was pushed onto stack and whether it needs protecting. E.g. look for GetPushedArgSize in eetwain.cpp - it returns how much is pushed on the stack at given offset in the method.

BruceForstall: > The current Windows x86 can handle this situation fine, with just ESP and EBP. What is different about the Linux x86 funclets that it does not work?

`GetPushedArgSize()` is currently only used in ESP-based frames. With funclets enabled for Linux x86, we're generating EBP-based frames. For normal EBP-based frame unwind, we don't need to worry about pushed arguments because we just use the EBP to pop the frame. However, with funclets, we can't do that because the EBP is pointing to the parent function (the funclet sets this up in the prolog to enable stack local variable access via EBP). Maybe it's possible to still use `GetPushedArgSize()` in an EBP-based frame in the funclet case to unwind ESP, and then find the actual EBP that was pushed/saved in the funclet prolog (as opposed to the parent frame EBP that was established)?

BruceForstall: BTW, I don't think we should mark regions where we push call arguments within funclets as non-interruptible, which might be what @janvorli was suggesting.

janvorli: @BruceForstall no, I was not trying to suggest that. I was just explaining how things work.

janvorli: We are trying to solve various issues given the current state of how we decided to do certain things on the fly. Given the knowledge of issues we have accumulated recently, I believe we should step back and reconsider the possibilities we have from higher level point of view. Some of the issues are due to the way we have decided to use for stack aligning. Some might be solved if we changed calling convention of jitted code so that the callee is no longer responsible for reclaiming the stack arguments space.
I am afraid that otherwise we would get lost in solving local issues that we could side step using a  different global approach.

wateret:
I have updated the change. I'm aware that we still have some issue but IMHO we will need something like this anyway. So what about merging this, if there is no other issue except for the one that we discussed? Seems like this patch makes most EH test cases(except FILTERs) pass .

src/vm/exceptionhandling.cpp
```diff
@@ -3194,20 +3198,57 @@ CLRUnwindStatus ExceptionTracker::ProcessManagedCallFrame(
...
+static inline UINT_PTR *GetRegRestoreBase(PCONTEXT pContextRecord)
+{
+#if defined(_TARGET_ARM_)
+    return (UINT_PTR*)&(pContextRecord->R4);
+#elif defined(_TARGET_ARM64_)
+    return (UINT_PTR*)&(pContextRecord->X19);
+#elif defined(_TARGET_X86_)
+    return (UINT_PTR*)&(pContextRecord->Edi);
+#else
+    PORTABILITY_ASSERT("GetFirstNonVolRegLocation");
```
parjong: Could you please change this as ``GetRegRestoreBase``?

src/vm/exceptionhandling.cpp
```diff
@@ -3194,20 +3198,57 @@ CLRUnwindStatus ExceptionTracker::ProcessManagedCallFrame(
...
+}
+
+static inline TADDR GetFrameRestoreBase(PCONTEXT pContextRecord)
+{
+#if defined(_TARGET_ARM_) || defined(_TARGET_ARM64_)
+    return GetSP(pContextRecord);
+#elif defined(_TARGET_X86_)
+    return pContextRecord->Ebp;
+#else
+    PORTABILITY_ASSERT("GetRestoreBase");
```
parjong: Similarly, need to change this as ``GetFrameRestoreBase``.

parjong:
@jkotas @janvorli @BruceForstall Could you please take a look this again? As I already posted in #9265, this PR is necessary step to resolve various issues in x86/Linux. I totally agree with @janvorli's opinion that we need to revisit the overall structure. However, the infrastructure that this PR attempts to introduce will be required even when we made another decision.

src/jit/codegencommon.cpp
```diff
@@ -10302,6 +10302,22 @@ void CodeGen::genCaptureFuncletPrologEpilogInfo()
 /*****************************************************************************
  *
  *  Generates code for an EH funclet prolog.
+ *
+ *
+ *  Funclets have the following incoming arguments:
+ *
+ *      catch/filter-handler: eax = the exception object that was caught (see GT_CATCH_ARG), ecx = FP
```
jkotas: Funclets on all other architectures get the argument via regular argument registers. Would it make more sense to pass the exception object in edx instead?

src/jit/codegencommon.cpp
```diff
@@ -10317,10 +10333,23 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
...
+    compiler->unwindPush(REG_FPBASE);
+
     // TODO Save callee-saved registers
 
     // This is the end of the OS-reported prolog for purposes of unwinding
     compiler->unwindEndProlog();
+
+    // Restore frame pointer
+    // TODO We will need changes here when we introduce PSPSym
+    inst_RV_RV(INS_mov, REG_FPBASE, REG_ARG_0);
```
jkotas: The first thing the funclet code does is to move ECX to EBP. Can the assembly helper set EBP just like all other argument registers instead?

src/jit/codegencommon.cpp
parjong: It is possible. we need ``mov eax, edx`` in prolog if we pass exception object via ``edx``.

src/jit/codegencommon.cpp
parjong: @jkotas I tried that approach first, but it makes ``CallEHFunclet`` frame unwindable.

wateret: I tried that last time. It seems like if EBP is restored in the assembly helper, native unwinder cannot unwind the helper correctly.

src/jit/codegencommon.cpp
wateret: Last time I did so and I had to also put `mov eax, edx` in funclet prolog because of 
```
#define REG_EXCEPTION_OBJECT     REG_EAX
```
 in jit/target.h.
So I thought if we pass the Exception Object directly to EAX then we could save an instruction.

I also thought it would be better if we could set REG_EXCEPTION_OBJECT to REG_EDX.

src/jit/codegencommon.cpp
janvorli: I believe that if you change the assembly helper to use ESP frame instead of EBP one, you would not have this problem. It seems all you'd need to do is to remove the PROLOG_BEG / PROLOG_END and EPILOG_BEG / EPILOG_END (or replace them with dummy ESP_PROLOG_BEG, ... to make it kind of symmetric with EBP based frames) and add `PROLOG_PUSH ebp` at the beginning and `EPILOG_POP ebp` at the end. Then changing the EBP in the middle of the function won't cause trouble, since the unwinder would use ESP for unwinding.

src/jit/codegencommon.cpp
```diff
@@ -10317,10 +10333,23 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
...
+
+    // Add a padding for 16-byte alignment
+    inst_RV_IV(INS_sub, REG_SPBASE, 8, EA_PTRSIZE);
+
+    // We've modified EBP, but not really. Say that we haven't...
```
wateret: The comment and the next line is copied from AMD64 version of this function. What I thought was we need that if we modify EBP in funclet.

src/jit/codegencommon.cpp
wateret: Thanks, I'll try that.

wateret: @janvorli I made a draft (I did not update the PR yet) for your suggestion but it does not work with nested exception handling like:

```cs
main {
  try {
    try {
      Console.WriteLine("try, first throw");
      throw new Exception("a");
    } catch (Exception) {
      Console.WriteLine("first catch, second throw");
      throw new Exception("b");
    }
  } catch (Exception) {
      Console.WriteLine("second catch");
    return 2;
  }
  return 0;
}
```

And the error message is:
```
try, first throw
first catch, second throw
terminate called after throwing an instance of 'PAL_SEHException'
```

It crashes when `PAL_ThrowExceptionFromContext()` throws C++ exception. IMHO it went wrong when unwinding the helper(CallEHFunclet) frame which is the caller of inner catch funclet, since there is no funclet(or its helper) when it handled first one.

Here is my draft https://github.com/wateret/coreclr/commit/632f2a6e03de2d641dcecaf40811ede26b26bf5b . It would be nice if you have a look. Thanks.

janvorli: The code looks ok. Can you please set a breakpoint at the `call    edx` in the CallEHFunclet and when you hit is, see if the `bt` command shows a correct stack trace? 
If that doesn't work, the only thing that I can see as possibly needed in addition to the change would be adding `.cfi_def_cfa_offset 4` as the first instruction of the CallEHFunclet

wateret: It does not get to `CallEHFunclet`. It seems like it crashes when processing some unmanaged frame in `UnwindManagedExceptionPass2()`.

I tried `.cfi_def_cfa_offset 4` here(https://github.com/wateret/coreclr/commit/8e7019bb43726ca59a965761f1222852e2bda4be) but did not make changes unfortunately.

wateret: Here's backtrace when it crashes.

```
Program received signal SIGABRT, Aborted.
0xf7fdb440 in __kernel_vsyscall ()
(gdb) bt
#0  0xf7fdb440 in __kernel_vsyscall ()
#1  0xf7ce5687 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#2  0xf7ce8ab3 in __GI_abort () at abort.c:89
#3  0xf7f15405 in __gnu_cxx::__verbose_terminate_handler() () from /usr/lib/i386-linux-gnu/libstdc++.so.6
#4  0xf7f13063 in ?? () from /usr/lib/i386-linux-gnu/libstdc++.so.6
#5  0xf7f1309f in std::terminate() () from /usr/lib/i386-linux-gnu/libstdc++.so.6
#6  0xf7f13306 in __cxa_throw () from /usr/lib/i386-linux-gnu/libstdc++.so.6
#7  0xf7884399 in ThrowExceptionHelper (ex=0x805a174) at /home/hanjoung/ws/dotnet/coreclr/src/pal/src/exception/seh.cpp:231
#8  0xf719afbd in CallEHFunclet () at ehhelpers.S:157
#9  0xffffba48 in ?? ()
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
```

janvorli: But I can see the CallEHFunclet is on the stack trace above, so it must have been there. Maybe you had an issue setting the breakpoint? You can try to set it at the CallEHFunclet itself and then step through the code until you reach the call to the funclet itself.

wateret: Sorry for the confusion. What I meant was there is no CallEHFunclet for the second throw. For the first throw, the backtrace looks OK.

```
(gdb) x/20i CallEHFunclet                                                                                                             [63/9485]
   0xf719af9c <CallEHFunclet>:  push   %ebp
   0xf719af9d <CallEHFunclet+1>:        mov    %esp,%ebp
   0xf719af9f <CallEHFunclet+3>:        push   %ebx
   0xf719afa0 <CallEHFunclet+4>:        push   %esi
   0xf719afa1 <CallEHFunclet+5>:        push   %edi
   0xf719afa2 <CallEHFunclet+6>:        mov    0x14(%ebp),%eax
   0xf719afa5 <CallEHFunclet+9>:        mov    %esp,(%eax)
   0xf719afa7 <CallEHFunclet+11>:       mov    0xc(%ebp),%edx
   0xf719afaa <CallEHFunclet+14>:       mov    0x8(%ebp),%eax
   0xf719afad <CallEHFunclet+17>:       mov    0x10(%ebp),%ecx
   0xf719afb0 <CallEHFunclet+20>:       mov    (%ecx),%edi
   0xf719afb2 <CallEHFunclet+22>:       mov    0x4(%ecx),%esi
   0xf719afb5 <CallEHFunclet+25>:       mov    0x8(%ecx),%ebx
   0xf719afb8 <CallEHFunclet+28>:       mov    0x18(%ecx),%ebp
=> 0xf719afbb <CallEHFunclet+31>:       call   *%edx
   0xf719afbd <CallEHFunclet+33>:       pop    %edi
   0xf719afbe <CallEHFunclet+34>:       pop    %esi
   0xf719afbf <CallEHFunclet+35>:       pop    %ebx
   0xf719afc0 <CallEHFunclet+36>:       pop    %ebp
   0xf719afc1 <CallEHFunclet+37>:       ret    $0x10
(gdb) bt
#0  CallEHFunclet () at ehhelpers.S:157
#1  0xf7176612 in ExceptionTracker::CallHandler (this=0x8081d40, uHandlerStartPC=4066256066, sf=..., pEHClause=0x8081d94, pMD=0xf6335988, funcletType=Catch, pContextRecord=0xffffc398) at /home/hanjoung/ws/dotnet/coreclr/src/vm/exceptionhandling.cpp:3314
#2  0xf7176095 in ExceptionTracker::CallCatchHandler (this=0x8081d40, pContextRecord=0xffffc398, pfAborting=0xffffbd17) at /home/hanjoung/ws/dotnet/coreclr/src/vm/exceptionhandling.cpp:583
#3  0xf7178833 in ProcessCLRException (pExceptionRecord=0x80e7c20, MemoryStackFp=4294954048, pContextRecord=0xffffc398, pDispatcherContext=0xffffbee8) at /home/hanjoung/ws/dotnet/coreclr/src/vm/exceptionhandling.cpp:1104
#4  0xf71832b8 in UnwindManagedExceptionPass2 (ex=..., unwindStartContext=0xffffc398) at /home/hanjoung/ws/dotnet/coreclr/src/vm/exceptionhandling.cpp:4394
#5  0xf7183c6e in UnwindManagedExceptionPass1 (ex=..., frameContext=0xffffc728) at /home/hanjoung/ws/dotnet/coreclr/src/vm/exceptionhandling.cpp:4548
#6  0xf718472e in DispatchManagedException (ex=..., isHardwareException=false) at /home/hanjoung/ws/dotnet/coreclr/src/vm/exceptionhandling.cpp:4655
#7  0xf700060e in IL_Throw (obj=0xf30149bc) at /home/hanjoung/ws/dotnet/coreclr/src/vm/jithelpers.cpp:5297
#8  0xf25e24b7 in ?? ()
#9  0xf719b113 in CallDescrWorkerInternal () at asmhelpers.S:442
#10 0xf6f17f18 in CallDescrWorker (pCallDescrData=0xffffd2a8) at /home/hanjoung/ws/dotnet/coreclr/src/vm/callhelpers.cpp:135
#11 0xf6f17cbc in CallDescrWorkerWithHandler (pCallDescrData=0xffffd2a8, fCriticalCall=0) at /home/hanjoung/ws/dotnet/coreclr/src/vm/callhelpers.cpp:78
#12 0xf6f19d33 in MethodDescCallSite::CallTargetWorker (this=0xffffd428, pArguments=0xffffd3d0, pReturnValue=0xffffd320, cbReturnValue=8) at /home/hanjoung/ws/dotnet/coreclr/src/vm/callhelpers.cpp:645
#13 0xf6e9c097 in MethodDescCallSite::Call_RetArgSlot (this=0xffffd428, pArguments=0xffffd3d0) at /home/hanjoung/ws/dotnet/coreclr/src/vm/callhelpers.h:433
#14 0xf71f8b42 in RunMain(MethodDesc*, short, int*, REF<PtrArray>*)::$_1::operator()(RunMain(MethodDesc*, short, int*, REF<PtrArray>*)::Param*) const::{lambda(RunMain(MethodDesc*, short, int*, REF<PtrArray>*)::Param*)#1}::operator()(RunMain(MethodDesc*, short, int*, REF<PtrArray>*)::Param*) const (this=0xffffd4f0, pParam=0xffffd5a8) at /home/hanjoung/ws/dotnet/coreclr/src/vm/assembly.cpp:2061
#15 0xf71f5110 in RunMain(MethodDesc*, short, int*, REF<PtrArray>*)::$_1::operator()(RunMain(MethodDesc*, short, int*, REF<PtrArray>*)::Param*) const (this=0xffffd5a0, __EXparam=0xffffd5a8) at /home/hanjoung/ws/dotnet/coreclr/src/vm/assembly.cpp:2075
#16 0xf71f4eed in RunMain (pFD=0xf6335988, numSkipArgs=1, piRetVal=0xffffd664, stringArgs=0xffffd9f8) at /home/hanjoung/ws/dotnet/coreclr/src/vm/assembly.cpp:2075
#17 0xf71f5420 in Assembly::ExecuteMainMethod (this=0x80b7100, stringArgs=0xffffd9f8, waitForOtherThreads=1) at /home/hanjoung/ws/dotnet/coreclr/src/vm/assembly.cpp:2157
#18 0xf6cc6c91 in CorHost2::ExecuteAssembly (this=0x805a190, dwAppDomainId=1, pwzAssemblyPath=0x80a11f0 u"/work/nfs/dotnet/misctest/eh1/eh.dll", argc=0, argv=0x0, pReturnValue=0xffffdbfc) at /home/hanjoung/ws/dotnet/coreclr/src/vm/corhost.cpp:501
#19 0xf6c51094 in coreclr_execute_assembly (hostHandle=0x805a190, domainId=1, argc=0, argv=0x0, managedAssemblyPath=0x805209c "/work/nfs/dotnet/misctest/eh1/eh.dll", exitCode=0xffffdbfc) at /home/hanjoung/ws/dotnet/coreclr/src/dlls/mscoree/unixinterface.cpp:376
#20 0x0804c624 in ExecuteManagedAssembly (currentExeAbsolutePath=0x8052014 "/work/nfs/dotnet/tt/corerun", clrFilesAbsolutePath=0x8052074 "/work/nfs/dotnet/tt", managedAssemblyAbsolutePath=0x805209c "/work/nfs/dotnet/misctest/eh1/eh.dll", managedAssemblyArgc=0, managedAssemblyArgv=0x0) at /home/hanjoung/ws/dotnet/coreclr/src/coreclr/hosts/unixcoreruncommon/coreruncommon.cpp:420
#21 0x0804b1e1 in corerun (argc=2, argv=0xffffddf4) at /home/hanjoung/ws/dotnet/coreclr/src/coreclr/hosts/unixcorerun/corerun.cpp:149
#22 0x0804b2ea in main (argc=2, argv=0xffffddf4) at /home/hanjoung/ws/dotnet/coreclr/src/coreclr/hosts/unixcorerun/corerun.cpp:161
```

janvorli: @wateret That means that the unwinding in the CallEHFunclet works fine. So my guess is that there is a problem in the implementation of the ThrowExceptionFromContextInternal. It probably doesn't restore state for the exception throwing properly. As you can see from the failure stack, the next frame after CallEHFunclet is a garbage while it should be a frame of ExceptionTracker::CallHandler.
Could you please dump register values when you hit the breakpoint at the `call edx` and then after you hit the abort issue, dump register values at frame 8 (the one of the CallEHFunclet)? MY guess is that the esp or ebp would differ.

wateret:
Rebased to current master due to conflict.

src/jit/codegencommon.cpp
wateret: Here are register dumps.

on breakpoint `call edx`
```
eax            0xf30149bc       -218019396
ecx            0xffffc434       -15308
edx            0xf25e24c2       -228711230
ebx            0xffffd2a8       -11608
esp            0xffffb924       0xffffb924
ebp            0xffffcc38       0xffffcc38
esi            0xf301237c       -218029188
edi            0xffffcc2c       -13268
eip            0xf719b23b       0xf719b23b <CallEHFunclet+31>
```
when abortion at frame 8
```
#8  0xf719b23d in CallEHFunclet () at ehhelpers.S:156
156     ehhelpers.S: No such file or directory.
(gdb) i r
eax            0x0      0
ecx            0x8      8
edx            0x6      6
ebx            0xf719b23d       -149310915
esp            0xffffb920       0xffffb920
ebp            0xffffcc38       0xffffcc38
esi            0x0      0
edi            0xf7a1c911       -140392175
eip            0xf719b23d       0xf719b23d <CallEHFunclet+33>
```

Like you said, ESPs differ.

wateret: I set  breakpoint on `push    eax` in `ThrowExceptionFromContextInternal`. At this point the restored EBP and ESP seems correct.

```
(gdb) i r esp ebp
esp            0xffffb924       0xffffb924
ebp            0xffffcc38       0xffffcc38
```

janvorli: @wateret thank you. I think the esp at the push eax is actually not correct and it should be larger by 4. But we cannot do that, since we would overwrite something on the stack of the CallEHFunclet (or any other native method whose context is passed to ThrowExceptionFromContextInternal.
Can you please change the ThrowExceptionFromContextInternal calling convention to __fastcall and pass the PAL_SEHException * in ecx? I believe that should fix the issue.

wateret: I simply tested by manually setting ESP larger by 4 on GDB and it works! and I made a patch for that. https://github.com/wateret/coreclr/commit/b5c4f4502a9eae066103578175d2d4ab622da01e is that what you meant? It passes the test case I gave.

BTW I was looking at `ThrowExceptionFromContextInternal`, is there some meaning for pushing Context_Eip before calling into `ThrowExceptionHelper`? I see that some kind of return address should be pushed since calling `ThrowExceptionHelper` with `jmp` instruction. However it looks like doing nothing since it will never return.

janvorli: @wateret yes, the patch is exactly what I've meant. It is possible that it will fix more issues in the coreclr tests, since the pattern in the test is quite common.
As for pushing the Context_Eip on the stack - that enables the stack unwinder to unwind from ThrowExceptionHelper to CallEHFunclet. It creates a stack frame looking as if the ThrowExceptionHelper was called by the CallEHFunclet, which allows the exception to be propagated from ThrowExceptionHelper to CallEHFunclet and further.

wateret: @janvorli Thank you for all of your response. I'll update the PR after polishing the patch and run unittests

wateret:
@janvorli @jkotas @BruceForstall
I have updated the PR with your comments. Could you take a look and merge this if there is no critical issue? Actually we have some works that are depend on this PR so it'd be great if we can have this merged soon.

BruceForstall:
LGTM

src/vm/exceptionhandling.cpp
```diff
@@ -3198,20 +3199,57 @@ CLRUnwindStatus ExceptionTracker::ProcessManagedCallFrame(
...
+static inline UINT_PTR CastHandlerFn(HandlerFn *pfnHandler)
+{
+#ifdef _TARGET_ARM_
+    return DataPointerToThumbCode<UINT_PTR, HandlerFn *>(pfnHandler);
+#else
+    return (UINT_PTR)pfnHandler;
+#endif
+}
+
+static inline UINT_PTR *GetRegRestoreBase(PCONTEXT pContextRecord)
```
janvorli: A nit - I would rename this to GetFirstNonVolatileRegisterAddress. It seems it better describes the purpose and the parameter to the CallEHFunclet that we pass the result of this function to is called pFirstNonVolReg, so it would match that too.

src/pal/inc/unixasmmacrosx86.inc
```diff
@@ -66,6 +66,21 @@ C_FUNC(\Name\()_End):
         pop ebp
 .endm
 
+.macro ESP_PROLOG_BEG
+        push ebp
+        mov ebp, esp
```
wateret: You're right, thanks.

src/vm/exceptionhandling.cpp
wateret: That sounds better to me too.

src/jit/codegenxarch.cpp
```diff
@@ -241,7 +241,11 @@ BasicBlock* CodeGen::genCallFinally(BasicBlock* block)
     if ((compiler->lvaPSPSym == BAD_VAR_NUM) ||
         (!compiler->compLocallocUsed && (compiler->funCurrentFunc()->funKind == FUNC_ROOT)))
     {
-        inst_RV_RV(INS_mov, REG_ARG_0, REG_SPBASE, TYP_I_IMPL);
+#ifdef UNIX_X86_ABI
```
jkotas: This change should not be needed anymore.

src/jit/codegencommon.cpp
```diff
@@ -10317,10 +10333,16 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
 
     compiler->unwindBegProlog();
 
-    // TODO Save callee-saved registers
+    inst_RV(INS_push, REG_FPBASE, TYP_REF);
```
jkotas: You should not need to push FP for funclets because the funclet does not modify it anymore

BruceForstall: @jkotas In the changes to UnwindEbpDoubleAlignFrame(), there is a call to SetEbpLocation() that appears to point to this saved EBP. Is that not required? (Or maybe I'm reading it wrong?)

jkotas: This should not be needed. When we come to the unwind the funclet, we have to have valid EBP and EBP location from wherever we came from. We should not need to create a duplicate copy here.

janvorli:

janvorli:

src/jit/codegenxarch.cpp
parjong: We may need the following change since no argument is passed when we call finally (as I understand):
```
#ifndef UNIX_X86_ABI
inst_RV_RV(INS_mov, REG_ARG_0, REG_SPBASE, TYP_I_IMPL);
#endif // !UNIX_X86_ABI
```

src/jit/codegencommon.cpp
parjong: @BruceForstall It seems that ``UnwindEbpDoubleAlignFrame`` should be revised along with ``genFuncletProlog``.

src/jit/codegenxarch.cpp
jkotas: Agree. (The extra mov ecx, rsp would be harmless - but it is better to not have it.)

wateret: I agree with @parjong

src/jit/codegencommon.cpp
```diff
@@ -10317,10 +10333,13 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
...
 
-    // TODO Save callee-saved registers
-
     // This is the end of the OS-reported prolog for purposes of unwinding
     compiler->unwindEndProlog();
+
+    // TODO We may need EBP restore sequence here if we introduce PSPSym
+
+    // Add a padding for 16-byte alignment
+    inst_RV_IV(INS_sub, REG_SPBASE, 8, EA_PTRSIZE);
```
jkotas: Does the padding need to be different now?

src/vm/eetwain.cpp
```diff
@@ -3806,6 +3807,29 @@ bool UnwindEbpDoubleAlignFrame(
...
+        // Funclets' frame pointers(EBP) are always restored so they can access to main function's local variables.
+        // Therefore the value of EBP is invalid for unwinder so we should use ESP instead.
+        // TODO If funclet frame layout is changed from CodeGen::genFuncletProlog() and genFuncletEpilog(),
+        //      we need to change here accordingly. It is likely to have changes when introducing PSPSym.
+        // TODO Currently we assume that ESP of funclet frames is always fixed but actually it could change.
+        if (pCodeInfo->IsFunclet())
+        {
+            baseSP = curESP + 8; // padding for 16byte stack alignment allocated in genFuncletProlog()
+
+            pContext->SetEbpLocation(PTR_DWORD(baseSP));
```
jkotas: This should be deleted because of EBP is not stored anymore

src/jit/codegencommon.cpp
wateret: Yes, It should be 12. Sorry about that.

src/vm/eetwain.cpp
wateret: I see that we need to remove that. Would it be OK to not doing `pContext->SetEbpLocation()` for unwinder?

cc: @parjong

src/vm/eetwain.cpp
```diff
@@ -3806,6 +3807,29 @@ bool UnwindEbpDoubleAlignFrame(
...
 
+#ifdef WIN64EXCEPTIONS
+        // Funclets' frame pointers(EBP) are always restored so they can access to main function's local variables.
+        // Therefore the value of EBP is invalid for unwinder so we should use ESP instead.
+        // TODO If funclet frame layout is changed from CodeGen::genFuncletProlog() and genFuncletEpilog(),
+        //      we need to change here accordingly. It is likely to have changes when introducing PSPSym.
+        // TODO Currently we assume that ESP of funclet frames is always fixed but actually it could change.
+        if (pCodeInfo->IsFunclet())
+        {
+            baseSP = curESP + 8; // padding for 16byte stack alignment allocated in genFuncletProlog()
```
parjong: The padding size should be 12 instead of 8.

src/vm/eetwain.cpp
parjong: This line and ``baseSP += sizeof(TADDR)`` at line 3822 should be removed as ``ebp`` is not saved any more.

wateret:
@jkotas Please take a look. I ran unit tests and the results are fine(same as the previous version that restors EBP in funclet).

BruceForstall:
@wateret Can you squash the commits?
@jkotas @janvorli I assume we're ready to merge this

jkotas:
Merged.

@BruceForstall The commits can be squashed as part of merge in the GitHub UI. No need to ask folks to do it.

janvorli:
@wateret I have just found that there is an alignment issue in the CallEHFunclet. You push four registers and there is return address on the stack, so the stack is not aligned to 16 bytes before calling the funclet. I have hit this issue when testing a simple null reference test app.
Can you please fix it? You will need both subtract 12 from the esp and use `.cfi_adjust_cfa_offset 12` to ensure the unwinding works properly. I would recommend creating a macro named ESP_PROLOG_STACK_ALLOC for this purpose.

wateret:
@janvorli Thank you for finding that. I'll fix that next Monday.

Issue #9604, opened 2017-02-15
[x86/Linux] sigsegv not cathed when fired from managed code


seanshpark:
JIT.Regression.CLR-x86-JIT.V1-M09.5-PDC.b16335 fails for this issue.
```
FAILED   - [ 4300 :    2 sec ] JIT/Regression/CLR-x86-JIT/V1-M09.5-PDC/b16335/b16335/b16335.sh
               BEGIN EXECUTION
               /root/dotnet-test/coreclr_unittest_20170214_103305_x86_x86-emu/coreclr-unittest-170214-013615/overlay/corerun b16335.exe
               ./b16335.sh: line 131: 20595 Segmentation fault      (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" b16335.exe $CLRTestExecutionArguments
               Expected: 100
               Actual: 139
               END EXECUTION - FAILED
```

seanshpark:
`IsSafeToHandleHardwareException` returns false for this unit test

seanshpark:
```
    CHECK_RANGE(JIT_WriteBarrier)
    CHECK_RANGE(JIT_CheckedWriteBarrier)
```
this block is not included for `_TARGET_X86_`. I think we need to include for `FEATURE_PAL`

seanshpark:
CC @parjong

janvorli:
@seanshpark yes, we need to check for the write barriers for x86 on Unix. But the check needs to be different since for x86, we have multiple write barriers (JIT_WriteBarrierEAX, JIT_WriteBarrierEBX, ... - the same for the JIT_CheckedWriteBarrier). They are all next to each other in memory and so we need to check against JIT_WriteBarrierStart / JIT_WriteBarrierLast range in the IsIPInMarkedJitHelper for x86 Unix.
It also seems we need to check against JIT_PatchedWriteBarrierStart / JIT_PatchedWriteBarrierLast range too.

PR #9605, opened 2017-02-15
[x86/Linux] Fix IsIPInMarkedJitHelper to handle sigsegv

seanshpark:
Include JIT_WriteBarrier and JIT_CheckedWriteBarrier functions
so that we can handle NullReferenceException for sigsegv from WriteBarriers.

### DIFF
--- a/src/vm/excep.cpp
+++ b/src/vm/excep.cpp
@@ -7190,6 +7190,11 @@ bool IsIPInMarkedJitHelper(UINT_PTR uControlPc)
 
     CHECK_RANGE(JIT_WriteBarrier)
     CHECK_RANGE(JIT_CheckedWriteBarrier)
+#else
+#ifdef FEATURE_PAL
+    CHECK_RANGE(JIT_WriteBarrierGroup)
+    CHECK_RANGE(JIT_PatchedWriteBarrierGroup)
+#endif // FEATURE_PAL
 #endif // _TARGET_X86_
 
 #if defined(_TARGET_AMD64_) && defined(_DEBUG)
@@ -7215,8 +7220,8 @@ AdjustContextForWriteBarrier(
 #if defined(_TARGET_X86_) && !defined(PLATFORM_UNIX)
     void* f_IP = (void *)GetIP(pContext);
 
-    if (((f_IP >= (void *) JIT_WriteBarrierStart) && (f_IP <= (void *) JIT_WriteBarrierLast)) ||
-        ((f_IP >= (void *) JIT_PatchedWriteBarrierStart) && (f_IP <= (void *) JIT_PatchedWriteBarrierLast)))
+    if (((f_IP >= (void *) JIT_WriteBarrierGroup) && (f_IP <= (void *) JIT_WriteBarrierGroup_End)) ||
+        ((f_IP >= (void *) JIT_PatchedWriteBarrierGroup) && (f_IP <= (void *) JIT_PatchedWriteBarrierGroup_End)))
     {
         // set the exception IP to be the instruction that called the write barrier
         void* callsite = (void *)GetAdjustedCallAddress(*dac_cast<PTR_PCODE>(GetSP(pContext)));
--- a/src/vm/i386/jithelp.S
+++ b/src/vm/i386/jithelp.S
@@ -371,12 +371,12 @@ NESTED_END JIT_ByRefWriteBarrier, _TEXT
 
 .endm
 
-// WriteBarrierStart and WriteBarrierEnd are used to determine bounds of
-// WriteBarrier functions so can determine if got AV in them.
+// JIT_WriteBarrierGroup and JIT_WriteBarrierGroup_End are used
+// to determine bounds of WriteBarrier functions so can determine if got AV in them.
 //
-LEAF_ENTRY JIT_WriteBarrierStart, _TEXT
+LEAF_ENTRY JIT_WriteBarrierGroup, _TEXT
     ret
-LEAF_END JIT_WriteBarrierStart, _TEXT
+LEAF_END JIT_WriteBarrierGroup, _TEXT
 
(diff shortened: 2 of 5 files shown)


seanshpark:
@parjon, could you please take a look at the changes in jithelp.S? WriteBarrier and CheckWriteBarrier macro is changed(swapped) to reflect real function names.

seanshpark:
Fix for #9604

src/vm/i386/jithelp.S
```diff
@@ -399,23 +399,23 @@ UniversalWriteBarrierHelper CheckedWriteBarrier
...
     ret
-LEAF_END JIT_WriteBarrierLast, _TEXT
+LEAF_END JIT_CheckedWriteBarrierLast, _TEXT
 
 // This is the first function outside the "keep together range". Used by BBT scripts.
-LEAF_ENTRY JIT_WriteBarrierEnd, _TEXT
+LEAF_ENTRY JIT_CheckedWriteBarrier_End, _TEXT
     ret
-LEAF_END JIT_WriteBarrierEnd, _TEXT
+LEAF_END JIT_CheckedWriteBarrier_End, _TEXT
```
parjong: It would be better to use ``LEAF_END_MARKED`` instead of explicitly appending ``_End``.

seanshpark: Thanks!

seanshpark: I'm not sure I could use `LEAF_END_MARKED` for the macro.

parjong:
@seanshpark I currently have no idea on ``_JIT_WriteBarrierEnd``. It was simply ported from ``jithelp.asm`` while x86/Linux bring up, but I couldn't find any reference on this function.

janvorli:
As I have said in my comment to the issue you have created (I have not noticed there is a PR for it already):
The check needs to be different since for x86, we have multiple write barriers (JIT_WriteBarrierEAX, JIT_WriteBarrierEBX, ... - the same for the JIT_CheckedWriteBarrier). They are all next to each other in memory and so we need to check against JIT_WriteBarrierStart / JIT_WriteBarrierLast range in the IsIPInMarkedJitHelper for x86 Unix.
It also seems we need to check against JIT_PatchedWriteBarrierStart / JIT_PatchedWriteBarrierLast range too.
So you only need to modify the IsIPInMarkedJitHelper.

seanshpark:
@janvorli , thank you for the comments. Current implementation uses macro to repeat for all registers so I think it may go ugly at first when expanding for each. :)

janvorli:
@seanshpark I am not sure I understand. You need to check just a single range to cover all of the JIT_CheckedWriteBarrierXXX and JIT_WriteBarrierXXX - the JIT_WriteBarrierStart / JIT_WriteBarrierLast.

stephentoub:

seanshpark:
@janvorli ,Thank you for your clarification, "just a single range to cover all of the ...". I thought but didn't understand correctly. I'll upload the change shortly.

seanshpark:
@janvori, CI is running but could you please take a look at the patch that it is what you are thinking?

src/vm/excep.cpp
```diff
@@ -7164,6 +7164,13 @@ EXTERN_C void JIT_MemCpy_End();
 
 EXTERN_C void JIT_WriteBarrier_End();
 EXTERN_C void JIT_CheckedWriteBarrier_End();
+#else
+#ifdef FEATURE_PAL
+EXTERN_C void JIT_WriteBarrierGroup();
```
janvorli: I am not sure I understand why you have added these new symbols. There already are JIT_WriteBarrierStart and JIT_WriteBarrierLast  / JIT_PatchedWriteBarrierStart and JIT_PatchedWriteBarrierLast symbols that you can use.

seanshpark: `CHECK_RANGE` macro uses `name` and `name##_End`.  To follow this macro `JIT_WriteBarrier` is already used by `UniversalWriteBarrierHelper` and there are a group of methods so used `Group` prefix.

janvorli: Ah, so you did that just so that you can use the CHECK_RANGE macro. I was thinking that for x86, you would just put in a direct comparison like we have in AdjustContextForWriteBarrier for x86.
It is a bit confusing to have two different sets of symbols for the same address range check, but I like your naming better.
Could you please remove the JIT_WriteBarrierStart and JIT_WriteBarrierLast / JIT_PatchedWriteBarrierStart and JIT_PatchedWriteBarrierLast and replace their usage by your new symbols then? This would include renaming them in jithelp.asm too. There are just few places to make this change.

seanshpark: Thank you you liked it. `JIT_WriteBarrierStart` and `JIT_WriteBarrierLast` is used in jitinterfacex86.cpp; so they seem to be needed.
```
    // All write barrier helpers should fit into one page.
    // If you hit this assert on retail build, there is most likely problem with BBT script.
    _ASSERTE_ALL_BUILDS("clr/src/VM/i386/JITinterfaceX86.cpp", (BYTE*)JIT_WriteBarrierLast - (BYTE*)JIT_WriteBarrierStart < PAGE_SIZE);
    _ASSERTE_ALL_BUILDS("clr/src/VM/i386/JITinterfaceX86.cpp", (BYTE*)JIT_PatchedWriteBarrierLast - (BYTE*)JIT_PatchedWriteBarrierStart < PAGE_SIZE);
```

janvorli: I meant that you would rename those to use the ones you have added - the JIT_WriteBarrierGroup, JIT_WriteBarrierGroup_End etc. In your change, you have added your ones right next to the existing ones, so they represent the same range.

seanshpark: OK

seanshpark:
I hope this is what you're thinking :)

PR #9609, opened 2017-02-15
Handle contained nodes earlier in buildRefPositionsForNode.

pgavlin:
These nodes never produce ref positions. Handling them earlier in this function
avoids a rather large amount of unnecessary work (e.g. this decreases overall
compile time by about 0.5% for `crossgen System.Private.CoreLib`).

### DIFF
--- a/src/jit/lsra.cpp
+++ b/src/jit/lsra.cpp
@@ -3563,6 +3563,39 @@ void LinearScan::buildRefPositionsForNode(GenTree*                  tree,
     }
 #endif // DEBUG
 
+    const bool isContainedNode = !info.isLocalDefUse && consume == 0 && produce == 0 && tree->canBeContained();
+    if (isContainedNode)
+    {
+        assert(info.internalIntCount == 0);
+        assert(info.internalFloatCount == 0);
+
+        // Contained nodes map to the concatenated lists of their operands.
+        LocationInfoList locationInfoList;
+        for (GenTree* op : tree->Operands())
+        {
+            if (!op->gtLsraInfo.definesAnyRegisters)
+            {
+                assert(ComputeOperandDstCount(op) == 0);
+                continue;
+            }
+
+            LocationInfoList operandList;
+            bool             removed = operandToLocationInfoMap.TryRemove(op, &operandList);
+            assert(removed);
+
+            locationInfoList.Append(operandList);
+        }
+
+        if (!locationInfoList.IsEmpty())
+        {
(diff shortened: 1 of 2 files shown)


src/jit/gtlist.h
```diff
@@ -261,7 +261,7 @@ GTNODE(PHI_ARG          , "phiArg"       ,GenTreePhiArg      ,0,GTK_LEAF|GTK_LOC
 //-----------------------------------------------------------------------------
 
 #ifndef LEGACY_BACKEND
-GTNODE(JMPTABLE         , "jumpTable"    ,GenTreeJumpTable   ,0, GTK_LEAF)              // Generates the jump table for switches
+GTNODE(JMPTABLE         , "jumpTable"    ,GenTreeJumpTable   ,0, GTK_LEAF|GTK_NOCONTAIN) // Generates the jump table for switches
```
mikedn: So we have only 5 opers that cannot be contained. I was expecting it to be the other way around - most nodes cannot be contained and only a few (LEA, INT) can. What am I missing?

pgavlin: In practice, I believe that you're correct. Marking these few operators with `GTK_NOCONTAIN` is really just a rephrasing of the switch table that `isContained` used to use.

pgavlin: Just a heads up if you're looking at this PR: I intend to break this down into two to four separate PRs in order to make it easier to review. I've really just been using this PR to run my changes through CI.

pgavlin:
@sivarv @dotnet/jit-contrib PTAL

src/jit/lsra.cpp
```diff
@@ -3563,6 +3563,39 @@ void LinearScan::buildRefPositionsForNode(GenTree*                  tree,
...
 #endif // DEBUG
 
+    const bool isContainedNode = !info.isLocalDefUse && consume == 0 && produce == 0 && tree->canBeContained();
+    if (isContainedNode)
+    {
+        assert(info.internalIntCount == 0);
+        assert(info.internalFloatCount == 0);
+
+        // Contained nodes map to the concatenated lists of their operands.
+        LocationInfoList locationInfoList;
```
sivarv: Looks good. Assuming zero SuperPMI asm diffs.

sivarv:
:shipit:

pgavlin:

src/jit/lsra.cpp
pgavlin: SuperPMI just finished; there are no diffs due to this change. I also checked AMD64 Unix via crossgen.

PR #9618, opened 2017-02-15
[x86/Linux] Emit proper frame for FastGetSharedStaticBase

parjong:
The stub generated by EmitFastGetSharedStaticBase currently does not create a proper frame in slow code path.

C++ unwinder fails to unwind this frame, which results in unwind failure in 2nd pass when the callee in the slow path throws a managed exception.

This commit creates a minimal frame for slow code path.

### DIFF
--- a/src/vm/i386/jitinterfacex86.cpp
+++ b/src/vm/i386/jitinterfacex86.cpp
@@ -1397,6 +1397,10 @@ void EmitFastGetSharedStaticBase(CPUSTUBLINKER *psl, CodeLabel *init, bool bCCto
         // DoInit:
         psl->EmitLabel(DoInit);
 
+#ifdef FEATURE_PAL
+        psl->X86EmitPushEBPframe();
+#endif // FEATURE_PAL
+
 #if defined(UNIX_X86_ABI)
         // sub esp, 8 ; to align the stack
         psl->X86EmitSubEsp(8);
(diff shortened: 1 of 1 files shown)


parjong:

src/vm/i386/jitinterfacex86.cpp
```diff
@@ -1397,10 +1397,16 @@ void EmitFastGetSharedStaticBase(CPUSTUBLINKER *psl, CodeLabel *init, bool bCCto
         // DoInit:
         psl->EmitLabel(DoInit);
 
-#if defined(UNIX_X86_ABI)
-        // sub esp, 8 ; to align the stack
-        psl->X86EmitSubEsp(8);
-#endif
+#ifdef FEATURE_PAL
```
janvorli: Why don't we have everything under UNIX_X86_ABI? What would be the FEATURE_PAL without UNIX_X86_ABI?

parjong: I though that this minimal frame is required even when 16-byte alignment is not required (``UNIX_X86_ABI`` currently means).

jkotas: It depends on how the unwinder works. E.g. it is not required on Windows because of Windows debuggers are happy to unwind it correctly as is.

I would either:
- Remove the `#ifdef FEATURE_PAL` and added the ebp frame everywhere. Even though it is not strictly required on Windows, it does not hurt to have it.
   or 
- Make the whole thing `UNIX_X86_ABI`

parjong: I prefer the former approach if you don't mind. I'll revise PR.

PR #9619, opened 2017-02-15
Update Buildtools to 1.0.27-prerelease-01310-03

wtgodbe:


MichaelSimons:
@wtgodbe - I think the immediate [issue](https://github.com/dotnet/buildtools/issues/1325) that is causing the  linux builds to fail should be fixed with 1.0.27-prerelease-01314-02.

wtgodbe:
@weshaggard Where is https://github.com/dotnet/buildtools/commit/2df0065bbac330a8f56e22b7718859cbdc50fb9b#diff-03c502c8604ae1b4ae2efeb1f3d0c45eR713 coming from in CoreFX? I think we're getting errors in building packages because CoreCLR doesn't supply this property

weshaggard:
It should be getting set in the packaging targets itself. See https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets#L48

weshaggard:
The error looks like it is caused by the fact you don't have a PackageIndex setup. For now I would suggest disabling the SkipApplyMetaPackages=true to by-pass this task as you guys don't really baseline stuff today. Although that might need to change at some point. cc @ericstj (who is current OOF but this is more an FYI for him).

wtgodbe:
Gotcha, thanks @weshaggard - just applied the change.

wtgodbe:
@weshaggard my local build just hit an error at https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets#L1032, looks like it's also a PackageIndex issue (CI builds should hit the same thing soon) - doesn't look like there's a condition for skipping that guy, what do you think is the right course of action here? Do we need to depend on that target executing?

src/.nuget/packages.builds
```diff
@@ -4,6 +4,7 @@
   <PropertyGroup>
     <!-- This property must be set to the same value as $(PackageOutputPath) for the nuspecs and nupkgs to be binplaced to the intended location. -->
     <OutputPath>$(PackageOutputPath)</OutputPath>
+    <PackageIndexFile>$(MSBuildThisFileDirectory)packageIndex.json</PackageIndexFile>
```
wtgodbe: @weshaggard is this all that's required (along with adding packageIndex.json, which is just above)?

weshaggard: You will need to have that set in a props file that gets included in all your package projects and you need to put it into an item group called PackageIndex.  See https://github.com/dotnet/wcf/blob/master/pkg/baseline/baseline.props#L20 as an example

wtgodbe:
@weshaggard what exactly is the purpose of PackageIndex.json? I'm now getting errors from buildtools because that file is empty: 11:15:24 D:\j\workspace\debug_windows---17180f1b\Tools\Packaging.targets(1073,5): error : PackageIndex from D:\j\workspace\debug_windows---17180f1b\src\.nuget/packageIndex.json is missing an entry for package Microsoft.NETCore.TestHost.  Please run /t:UpdatePackageIndex on this project to commit an update. [D:\j\workspace\debug_windows---17180f1b\src\.nuget\Microsoft.NETCore.TestHost\Microsoft.NETCore.TestHost.pkgproj]

weshaggard:
>@weshaggard what exactly is the purpose of PackageIndex.json? I'm now getting errors from buildtools because that file is empty:

The primary purpose is to ensure that dependencies on the packages produced in the repo are aligned to a consistent set of versions. The current failure is a packaging validation step which we could either disable or add an entire for each of the packages you are building in the repo. I think you can automatically generate it from this target https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets#L1201, but I've never used it so I'm not sure if it is setup correctly for the CoreCLR repo or not, you might need to setup some of those item groups to get it to index your package build.

wtgodbe:
@weshaggard is it unsafe to just disable that check (https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Packaging.targets#L1047) ? I'm not having much luck with UpdateRepoPackageIndex, and I'm not sure how the necessary ItemGroups correspond to what we have available in CoreCLR.

weshaggard:
>@weshaggard is it unsafe to just disable that check 

It could potentially hide some issues. To unblock yourself I would suggest disabling it for now and filing an issue to re-enable it later and ping @ericstj when he is back to help figure how to correctly setup the index or opt-out of the index completely.

chcosta:
I sync'd with @ericstj before he left and we came up with a fix for the packageindex stuff, which is in my PR.  I'm looking into the framework reference assemblies issue, this is new to me.

Issue #8348, opened 2016-11-29
 AutoResetEventCtor test fails intermittently on Windows x64


janvorli:
The same issue was already seen in the past (#6105)
```
7:56:56   Raw output:
07:56:56   BEGIN EXECUTION
07:56:56          "D:\j\workspace\x64_release_w---0575cb46\tests\..\bin\tests\Windows_NT.x64.Release\Tests\Core_Root\corerun.exe" AutoResetEventCtor.exe 
07:56:56         Beginning test case AutoResetEventCtor at 11/28/2016 11:56:46 PM
07:56:56         Random seed: 20010415; set environment variable CORECLR_SEED to this value to repro
07:56:56         
07:56:56         [Positive]
07:56:56         Beginning scenario: PosTest1: AutoResetEvent.Ctor(true)
07:56:56         Calling AutoResetEvent.WaitOne()... if the event is not signaled it will hang
07:56:56         Beginning scenario: PosTest1: AutoResetEvent.Ctor(false)
07:56:56         Calling AutoResetEvent.WaitOne()... if the event is signaled it will not wait long enough
07:56:56         ERROR!!!-002: AutoResetEvent did not wait long enough... this implies that the parameter was not respected.
07:56:56         ERROR!!!-002:  WaitTime=86421633 (ticks)
07:56:56         ERROR!!!-002:  Execpted=50000000 (ticks)
07:56:56         ERROR!!!-002:  Acceptable Delta=9999999 (ticks)
07:56:56         
07:56:56         Ending test case at 11/28/2016 11:56:55 PM
07:56:56         FAIL
07:56:56         Expected: 100
07:56:56         Actual: 0
07:56:56         END EXECUTION - FAILED
07:56:56         FAILED
07:56:56         Test Harness Exitcode is : 1
```

gkhanna79:
@janvorli Is there a specific reason you tagged this as diagnostics instead of Threading?

CC @kouvel

alexperovich:
It appears that this is actually a bug with the test. The test measured a wait time of 8 seconds, rather than the expected 5. This was measured with a real time clock however, and that could very well have been changed by 3 seconds during the test. I will switch the test to use a monotonic clock.

PR #9622, opened 2017-02-15
JIT: fix desktop build break

AndyAyersMS:
Desktop builds warn about unreachable code after returns, so modify
how finally chain merging is enabled and disabled in various configs.

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -23965,13 +23965,15 @@ void Compiler::fgMergeFinallyChains()
         return;
     }
 
+    bool enableMergeFinallyChains = true;
+
 #if !FEATURE_EH_FUNCLETS
     // For non-funclet models (x86) the callfinallys may contain
     // statements and the continuations contain GT_END_LFINs.  So no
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@briansull PTAL
cc @dotnet/jit-contrib

src/jit/flowgraph.cpp
```diff
@@ -23965,23 +23965,31 @@ void Compiler::fgMergeFinallyChains()
         return;
     }
 
+    bool enableMergeFinallyChains = true;
+
```
briansull: The  optimizer might still issue the unreachable warning for this way to doing this.
Let me check

briansull: The change passes arm32ret and x86ret desktop builds

briansull:
LGTM

AndyAyersMS:
Win x64 [(log)](https://ci.dot.net/job/dotnet_coreclr/job/master/job/x64_release_windows_nt_pri1_prtest/2957/) had one test failure which is a known issue (#8348). So will ignore and merge.

PR #9627, opened 2017-02-16
[x86/Linux] Fix x86 Release build break

wateret:


wateret:
cc: @jkotas @parjong

janvorli:
@wateret I don't understand how this could break only release build of unix x86. This code was there forever and it is not release build specific.

wateret:
Sorry I did not put up any information about it. Here's what I've got so far.

Build command : `./build.sh x86 cross Release skiptests`

The error message is:
```
In file included from /home/hanjoung/ws/dotnet/coreclr/src/gc/gcwks.cpp:22:
/home/hanjoung/ws/dotnet/coreclr/src/gc/gc.cpp:32177:5: error: invalid symbol redefinition
    __asm {
    ^
<inline asm>:5:2: note: instantiated into assembly here
        .L__MSASMLABEL_.0__gotit:
        ^
In file included from /home/hanjoung/ws/dotnet/coreclr/src/gc/gcwks.cpp:22:
/home/hanjoung/ws/dotnet/coreclr/src/gc/gc.cpp:32177:5: error: invalid symbol redefinition
    __asm {
    ^
<inline asm>:5:2: note: instantiated into assembly here
        .L__MSASMLABEL_.0__gotit:
        ^
In file included from /home/hanjoung/ws/dotnet/coreclr/src/gc/gcsvr.cpp:22:
/home/hanjoung/ws/dotnet/coreclr/src/gc/gc.cpp:32177:5: error: invalid symbol redefinition
    __asm {
    ^
<inline asm>:5:2: note: instantiated into assembly here
        .L__MSASMLABEL_.0__gotit:
        ^
In file included from /home/hanjoung/ws/dotnet/coreclr/src/gc/gcsvr.cpp:22:
/home/hanjoung/ws/dotnet/coreclr/src/gc/gc.cpp:32177:5: error: invalid symbol redefinition
    __asm {
    ^
<inline asm>:5:2: note: instantiated into assembly here
        .L__MSASMLABEL_.0__gotit:
        ^
```

in src/gc/gc.cpp
```
32173 inline
32174 void EnterAllocLock()
32175 {
32176 #if defined(_TARGET_X86_)
32177     __asm {
32178         inc dword ptr m_GCLock
32179         jz gotit
32180         call spin_lock
32181             gotit:
32182     }
32183 #else //_TARGET_X86_
32184     spin_lock();
32185 #endif //_TARGET_X86_
32186 }
```

When release mode, this function is inlined so there are a lot of `gotit:` labels. This syntax is OK in MASM, but not allowed for clang assembler I think. It seems like clang assembler does not fully support MASM syntax.

So I thought this may be an easy fix.

janvorli:
@wateret thank you for the explanation. Clang inline assembler happens to support a special sequence in symbol names that makes the symbol unique in every instance of the __asm block. So if the `gotit` was changed to `gotit%=`, the problem would go away. But since Visual C++ doesn't accept these characters in the symbol name, it needs to be ifdefed for `__clang__`.
So e.g.
```C++
#ifdef __clang__
#define GOTIT_LABEL gotit
#else
#define GOTIT_LABEL gotit%=
#endif

inline
void EnterAllocLock()
{
#if defined(_TARGET_X86_)
    __asm {
        inc dword ptr m_GCLock
        jz GOTIT_LABEL
        call spin_lock
            GOTIT_LABEL:
    }
#else //_TARGET_X86_
    spin_lock();
#endif //_TARGET_X86_
}
```

janvorli:
@Maoni0 I actually wonder why do we have this difference on x86. Can you please share some details on it?

wateret:
@janvorli Updated as you suggested, thanks!

janvorli:
LGTM, but I'd like to get @Maoni0's feedback before merging it in.

Maoni0:
@janvorli, I think the only reason was just because you can't have inline asm in x64 and we (well, people who wrote this code at the time) wanted inline asm so inline asm was used where it could be.

jkotas:
Can we get rid of the inlined asm here? I do not think there is a good reason to have it here anymore. (It made sense in ~1999 where this code was written originally ... compilers were not able to inline interlocked instructions, lock prefix was prohibitively expensive. None of these hold anymore.)

Maoni0:
@jkotas yeh, I have no objections to getting rid of it.

janvorli:
@wateret based on that, could you please revise the change to remove the x86 specific part completely?

wateret:
@janvorli Updated, thanks.

PR #9639, opened 2017-02-17
Fix wrong comment

parjong:
The corresponding ifdef seems to be WIN64EXCEPTIONS (not _WIN64).


danmosemsft:
I see there's a nested `#ifdef WIN64EXCEPTIONS` inside..

parjong:
@danmosemsft I added that code. I'll revise it.

Issue #2591, opened 2016-01-09
RyuJIT: [betternes]  Make JIT time constants from calls to Type methods and props


omariom:
If JIT could treat the calls like  typeof(T).IsValueType, typeof(IInterface).IsAssingnableFrom(typeof(T) etc.
that could allow to write more performant generic code.

mikedn:
Hmm, what would be the use of `typeof(IInterface).IsAssingnableFrom(typeof(T)` in generic code? 

Test if a value type implement a certain interface? But if it does you'll probably need to cast to that interface anyway.

omariom:
@mikedn 
For example, we could use the trick EqualityComparer uses to  dispatch handling to proper methods without casting and boxing. 
Like this:

```C#
if (typeof(IInterface1).IsAssigneableFrom(typeof(T)))
{
      DispatcherForIInterface1.Instance.Dispatch(value);
}
else if (typeof(IInterface2).IsAssigneableFrom(typeof(T)))
{
    DispatcherForIInterface2.Instance.Dispatch(value);
}
else
{
     throw new Exception("Bad type.");
}

```

mikedn:
I [see](https://github.com/dotnet/coreclr/blob/bc146608854d1db9cdbcc0b08029a87754e12b49/src/mscorlib/src/System/Collections/Generic/EqualityComparer.cs#L43). With this type of optimizations the JIT could reduce `EqualityComparer<int>.CreateComparer()` to just `return new GenericEqualityComparer<int>();`. And similar code size reductions could also be obtained for nullable and enum types. Even for reference types some size reductions could be obtained if the JIT knows that reference types cannot be enums nor nullables.

mikedn:
While we're at it I think we can also add the case of
```C#
if (typeof(T) == typeof(int)) {...
```
The if is eliminated as expected when `T` is a value type other than `int` but it is not eliminated when `T` is a reference type (obviously, if T is a reference type it can never be `int`).

benaadams:
:+1:

benaadams:
Although for coreclr you need to do
```csharp
typeof(T).GetTypeInfo().IsValueType
typeof(IInterface).GetTypeInfo().IsAssignableFrom(typeof(T).GetTypeInfo())
```

omariom:
Nice property is that the optimizations won't disappear when Ngen'ed or .NetNative'ed, will they?

mikedn:
> Although for coreclr you need to do...

[...]

PR #9640, opened 2017-02-17
Separate nullable/enum Comparer and EqualityComparer creation logic into new methods

jamesqo:
For `Comparer<T>.Default` or `EqualityComparer<T>.Default`, most `T`s will down the path that returns a comparer based on IComparable/IEquatable respectively. If `T` is not a nullable/enum, all of the stuff to handle nullables/enums in `CreateComparer` is dead code for that instantiation. The JIT does not yet recognize reflection code such as `typeof(T).IsValueType` et al. however (https://github.com/dotnet/coreclr/issues/2591) so it will not be able to eliminate this dead code.

This PR moves the creation of nullable/enum comparers into a new method, and cleans up the code in `CreateComparer` for both classes.

/cc @jkotas, @omariom, @benaadams

### DIFF
--- a/src/mscorlib/src/System/Collections/Generic/Comparer.cs
+++ b/src/mscorlib/src/System/Collections/Generic/Comparer.cs
@@ -45,58 +45,84 @@ namespace System.Collections.Generic
         //
         private static Comparer<T> CreateComparer()
         {
-            object result = null;
-            RuntimeType t = (RuntimeType)typeof(T);
+            var type = (RuntimeType)typeof(T);
 
             // If T implements IComparable<T> return a GenericComparer<T>
-            if (typeof(IComparable<T>).IsAssignableFrom(t))
+            if (typeof(IComparable<T>).IsAssignableFrom(type))
             {
-                result = RuntimeTypeHandle.CreateInstanceForAnotherGenericParameter((RuntimeType)typeof(GenericComparer<int>), t);
+                return (Comparer<T>)RuntimeTypeHandle.CreateInstanceForAnotherGenericParameter((RuntimeType)typeof(GenericComparer<int>), type);
             }
-            else if (default(T) == null)
+
+            if (default(T) == null)
             {
-                // If T is a Nullable<U> where U implements IComparable<U> return a NullableComparer<U>
-                if (t.IsGenericType && t.GetGenericTypeDefinition() == typeof(Nullable<>)) {
-                    RuntimeType u = (RuntimeType)t.GetGenericArguments()[0];
-                    if (typeof(IComparable<>).MakeGenericType(u).IsAssignableFrom(u)) {
-                        result = RuntimeTypeHandle.CreateInstanceForAnotherGenericParameter((RuntimeType)typeof(NullableComparer<int>), u);
-                    }
-                }
-            }
-            else if (t.IsEnum)
-            {
-                // Explicitly call Enum.GetUnderlyingType here. Although GetTypeCode
-                // ends up doing this anyway, we end up avoiding an unnecessary P/Invoke
-                // and virtual method call.
-                TypeCode underlyingTypeCode = Type.GetTypeCode(Enum.GetUnderlyingType(t));
-                
-                // Depending on the enum type, we need to special case the comparers so that we avoid boxing
-                // Specialize differently for signed/unsigned types so we avoid problems with large numbers
-                switch (underlyingTypeCode)
+                // Nullable does not implement IComparable<T?> directly because that would add an extra interface call per comparison.
+                // Instead, it relies on Comparer<T?>.Default to specialize for nullables and do the lifted comparisons if T implements IComparable.
+                if (type.IsValueType)
                 {
-                    case TypeCode.SByte:
-                    case TypeCode.Int16:
-                    case TypeCode.Int32:
-                        result = RuntimeTypeHandle.CreateInstanceForAnotherGenericParameter((RuntimeType)typeof(Int32EnumComparer<int>), t);
-                        break;
-                    case TypeCode.Byte:
-                    case TypeCode.UInt16:
-                    case TypeCode.UInt32:
-                        result = RuntimeTypeHandle.CreateInstanceForAnotherGenericParameter((RuntimeType)typeof(UInt32EnumComparer<uint>), t);
-                        break;
-                    // 64-bit enums: use UnsafeEnumCastLong
-                    case TypeCode.Int64:
-                        result = RuntimeTypeHandle.CreateInstanceForAnotherGenericParameter((RuntimeType)typeof(Int64EnumComparer<long>), t);
-                        break;
-                    case TypeCode.UInt64:
-                        result = RuntimeTypeHandle.CreateInstanceForAnotherGenericParameter((RuntimeType)typeof(UInt64EnumComparer<ulong>), t);
-                        break;
+                    return CreateNullableComparer();
                 }
+            }
+            else if (type.IsEnum)
+            {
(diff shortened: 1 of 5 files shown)


src/mscorlib/src/System/Collections/Generic/Comparer.cs
```diff
@@ -45,58 +45,84 @@ public static Comparer<T> Create(Comparison<T> comparison)
...
             {
-                // If T is a Nullable<U> where U implements IComparable<U> return a NullableComparer<U>
-                if (t.IsGenericType && t.GetGenericTypeDefinition() == typeof(Nullable<>)) {
-                    RuntimeType u = (RuntimeType)t.GetGenericArguments()[0];
-                    if (typeof(IComparable<>).MakeGenericType(u).IsAssignableFrom(u)) {
-                        result = RuntimeTypeHandle.CreateInstanceForAnotherGenericParameter((RuntimeType)typeof(NullableComparer<int>), u);
-                    }
+                // Nullable does not implement IComparable<T?> directly because that would add an extra interface call per comparison.
+                // Instead, it relies on Comparer<T?>.Default to specialize for nullables and do the lifted comparisons if T implements IComparable.
+                if (type.IsValueType)
```
jamesqo: Using `default(T) == null && typeof(T).IsValueType` to detect nullables here, after discussion at https://github.com/dotnet/corefx/issues/16217#issuecomment-280222155

src/mscorlib/src/System/Collections/Generic/Comparer.cs
```diff
@@ -196,7 +222,7 @@ internal sealed class Int32EnumComparer<T> : Comparer<T>, ISerializable where T
     {
         public Int32EnumComparer()
         {
-            Debug.Assert(typeof(T).IsEnum, "This type is only intended to be used to compare enums!");
```
jamesqo: This statement is obvious, so I deleted it (I wrote this in a previous PR)

jkotas:
Since you are refactoring this, could you please move everything possible to non-generic type to reduce the code bloat? Majority of this code is reflection based and thus it does not need to be parametrized over T. You can just pass RuntimeType around.

jamesqo:
@jkotas

> Majority of this code is reflection based and thus it does not need to be parametrized over T. You can just pass RuntimeType around.

It needs it for the return type `Comparer<T>` / `EqualityComparer<T>`, and to create `ObjectComparer<T>` as a fallback. Do you want me to have something like a `bool TryCreateNullableComparer(out object result)` pattern and do an extra cast in the caller?

jkotas:
> bool TryCreateNullableComparer(out object result) 

Yes, something like this. It can be just `object TryCreateNullableComparer()`.

jamesqo:
@jkotas I am going to make the return values the non-generic `IComparer` / `IEqualityComparer` interfaces. Let me know if you think that you sacrifice performance.

src/mscorlib/src/System/Collections/Generic/EqualityComparer.cs
```diff
@@ -36,57 +36,37 @@ private static EqualityComparer<T> CreateComparer()
...
+            else if (type.IsEnum)
+            {
+                // The equality comparer for enums is specialized to avoid boxing.
+                result = RareComparers.TryCreateEnumEqualityComparer(type);
             }
             
-            return result != null ?
-                (EqualityComparer<T>)result :
-                new ObjectEqualityComparer<T>(); // Fallback to ObjectEqualityComparer, which uses boxing
+            return result != null ? (EqualityComparer<T>)result : new ObjectEqualityComparer<T>();
```
jkotas: It is unfortunate that we will create ObjectEqualityComparer type because of it is referenced, even though it will be unused in many cases. We should create the ObjectEqualityComparer using CreateInstanceForAnotherGenericParameter like all other comparers.

src/mscorlib/src/System/Collections/Generic/EqualityComparer.cs
```diff
@@ -36,57 +36,37 @@ private static EqualityComparer<T> CreateComparer()
             Contract.Ensures(Contract.Result<EqualityComparer<T>>() != null);
             
             object result = null;
-            RuntimeType t = (RuntimeType)typeof(T);
+            var type = (RuntimeType)typeof(T);
             
-            // Specialize type byte for performance reasons
-            if (t == typeof(byte)) {
+            // Specialize for byte so Array.IndexOf is faster.
+            if (type == typeof(byte))
```
jkotas: There is still a lot of boiled plate code that will be duplicated for every valuetype instantiation. I think you should be able to actually move all this whole method into non-generic type. The only generic part should be:

`static readonly Comparer<T> defaultComparer = (Comparer<T>)EqualityComparerHelpers.GetComparer(typeof(T), typeof(IEquatable<T>));`

(Getting rid of `IEquatable<T>` is possible but a bit harder - you would need to check whether the type implement `IEquatable<T>` without actually loading the `IEquatable<T>` type. It can be done later.)

jkotas:
It would be nice to measure `EqualityComparer<T>` startup time improvement from this change by a small tests that instantiates a lot of comparers. Here is something that you can use as a start: https://gist.github.com/jkotas/07f1f77505611168c295993073633d71 . It is a small modification of the startup perf test used for #8963.

src/mscorlib/src/System/Collections/Generic/EqualityComparer.cs
jkotas: > Getting rid of IEquatable<T> is possible too but a bit harder. 

Actually, it should not be that hard - it can use the same pattern as what is used on the nulalble path: `typeof(IEquatable<>).MakeGenericType(u).IsAssignableFrom(u)`

jamesqo: @jkotas It's going to be kind of convoluted to create `new ObjectComparer<T>()` in that case because then we'd need to use `Activator.CreateInstance`... what about

```cs
static readonly EqualityComparer<T> defaultComparer = (EqualityComparer<T>)ComparerHelpers.CreateComparer(typeof(T)) ?? new ObjectComparer<T>();
```

?

jkotas: `CreateInstanceForAnotherGenericParameter` creates the instance for you. `Activator.CreateInstance` should not be necessary. I think that the path of last resort that creates `ObjectComparer` should look exactly same as all the other more optimized generic paths.

jamesqo:
> It would be nice to measure `EqualityComparer<T>` startup time improvement from this change by a small tests that instantiates a lot of comparers.

OK. Building coreclr has been problematic for me lately (see https://github.com/dotnet/coreclr/pull/9524#issuecomment-280811108) so I will try in a week or so.

Another thing I should do after measuring perf impact is running the corefx tests against this... there's more type-unsafe code than there was before, so it wouldn't be shocking if I screwed up somewhere.

jkotas:
LGTM so far. Thanks!

jamesqo:
@jkotas I ran your test 10 times with the old/new binaries. Here are the results:

Old:

```
17266
14562
15375
14219
15969
13203
17828
21781
19875
16984
```

New:

```
12610
12454
13500
13453
12407
12391
13000
11125
12656
10938
```

It seems to be generally faster than what was there before. I have to run corefx tests on this first though so don't merge yet.

jamesqo:
@jkotas OK, I ran the corefx tests on this. Feel free to merge.

Test Ubuntu x64 Checked Build and Test
Test Windows_NT x86 Checked Build and Test

PR #9645, opened 2017-02-17
Fix x64 perf run break

DrewScoggins:
After the checkin to add legacy JIT testing support for x86 we broke x64
runs.  THe issue is that we do not pass anything in to the TEST_ENV
variable on x64 and when a variable is empty and you attempt to use the
if exists syntax you fail.  I have added a nonsense file name that we
set this to before we go through the args loop. I was not able to find a
better solution and want to get the lab unblocked, but any batch masters
feel free to solve this the right way.

### DIFF
--- a/tests/scripts/run-xunit-perf.cmd
+++ b/tests/scripts/run-xunit-perf.cmd
@@ -10,6 +10,7 @@ set CORECLR_REPO=%CD%
 set TEST_FILE_EXT=exe
 set TEST_ARCH=x64
 set TEST_CONFIG=Release
+set TEST_ENV=thisfilewillnotexist
 
 goto :ARGLOOP
 


DrewScoggins:
test Windows_NT_x86jit32

DrewScoggins:
test Windows_NT_x86jit32 perf

DrewScoggins:
test Windows_NT_x64 perf

DrewScoggins:
We are now working for x64 and the x86jit32 run seems to be working as well.  PTAL @adiaaida.  I am going to go ahead and merge this though now so that we can get the x64 runs off of the floor.

adiaaida:
Sorry, I was on vacation. I probably would have gone with if DEFINED TEST_ENV AND EXIST TEST_ENV, but this works too.

DrewScoggins:
I tried something like this but with the equality operator and square brackets. The problem is that we would still evaluate the empty variable and end up with illegal syntax. We might be able to work around this with delayed expansion (h/t @ellismg). I'll look tomorrow.

adiaaida:
I think delayed expansion should work.

PR #9657, opened 2017-02-17
JIT: relax assert in fgMergeFinallyChains

### LINKED ISSUE #9651: JIT: finally chain tail merge can fail to merge
In cases where the jit ends up reimporting a `leave` it will create a new callfinally. The callfinally created during original importation then becomes unreachable. This trips up an assertion in finally chain merging, since iit sees a mergeable set of callfinallys but will only attempt merging on reachable callfinallys. So despite there being a merge opportunity, no merge actually happens. This triggers an assert:

```
Assertion failed 'didMerge' in 'FinallyReimp:Main(ref):int' (IL size 45)
    File: c:\repos\coreclr\src\jit\flowgraph.cpp Line: 24117
    Image: C:\repos\coreclr\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\CoreRun.exe
```

This only impacts CHK/DBG builds; the bug is that the assert is too strong.

Likely only happens for IL cases.

AndyAyersMS:
In some rare and likely IL-only cases the jit may create unreachable
callfinallys. This trips up finally chain merging since it spots a
merge opportunity but no merge actually happens.

Relax the assertion checks so that `canMerge && !didMerge` is now ok.

Closes #9651.

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -24114,20 +24114,31 @@ void Compiler::fgMergeFinallyChains()
     }
     else
     {
-        assert(didMerge);
-        JITDUMP("Method had try-finallys and some callfinally merges were performed.\n");
+        if (didMerge)
+        {
+            JITDUMP("Method had mergeable try-finallys and some callfinally merges were performed.\n");
 
 #if DEBUG
-
-        if (verbose)
-        {
-            printf("\n*************** After fgMergeFinallyChains()\n");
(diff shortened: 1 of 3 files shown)


AndyAyersMS:
@briansull PTAL
cc @dotnet/jit-contrib

BruceForstall:
LGTM

briansull:
LGTM

PR #9658, opened 2017-02-17
Exclude ni files from linux arm32 builds

chcosta:
/cc @gkhanna79 @ericstj 

@gkhanna79, is there a plan to add one of the Ubuntu arm builds to CI to prevent this type of break?  It seems useful given that it has unique packaging requirements.

### DIFF
--- a/src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/runtime.Linux.Microsoft.NETCore.Runtime.CoreCLR.props
+++ b/src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/runtime.Linux.Microsoft.NETCore.Runtime.CoreCLR.props
@@ -1,8 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <_PlatformDoesntSupportNiFiles Condition="'$(Platform)' == 'armel'">true</_PlatformDoesntSupportNiFiles>
+    <_PlatformDoesntSupportNiFiles Condition="'$(PackageTargetRuntime)' == 'ubuntu.14.04-arm'">true</_PlatformDoesntSupportNiFiles>
+    <_PlatformDoesntSupportNiFiles Condition="'$(PackageTargetRuntime)' == 'ubuntu.16.04-arm'">true</_PlatformDoesntSupportNiFiles>
+  </PropertyGroup>
   <ItemGroup>
     <NativeSplittableBinary Include="$(BinDir)libcoreclr.so" />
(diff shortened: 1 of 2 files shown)


src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/runtime.Linux.Microsoft.NETCore.Runtime.CoreCLR.props
```diff
@@ -1,17 +1,22 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <_PlatformDoesNotSupportNiFiles Condition="'$(Platform)' == 'armel'">true</_PlatformDoesNotSupportNiFiles>
+    <_PlatformDoesNotSupportNiFiles Condition="'$(PackageTargetRuntime)' == 'ubuntu.14.04-arm'">true</_PlatformDoesNotSupportNiFiles>
```
gkhanna79: We dont support this for any Linux Arm - why do we need to check for RID in one case and arch for the other?

gkhanna79:
Yes, @hqueue and @jyoungyun are working to add CI support for Ubuntu 14.04 and 16.04 arm builds (and later for Tizen as well) to avoid such breaks.

src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/runtime.Linux.Microsoft.NETCore.Runtime.CoreCLR.props
chcosta: It appears to be supported for Ubuntu 16.10 arm.  Is that not true?  https://github.com/dotnet/coreclr/commit/89ba46aa6e113486428ac7528f662e3a0a5a55eb#diff-0656a30adba406c0c1d1b61c78a55025R20)

chcosta: It doesn't look like you coreclr is intended to publish for Ubuntu 16.10 arm (I don't see packages on myget), that was carried over from corefx.  i'll update it

chcosta:
I removed the Ubuntu 16.10 arm entry from the OfficialBuildRid list.  Theoretically, it's fine for it to remain but it did cause me confusion so I removed it.  If there is an intent to add Ubuntu 16.10 arm builds soon, then I could be easily convinced to leave it in there though it is a bit odd to list OfficialBuildRids that we don't actually build.

gkhanna79:
Ubuntu 16.10 for Arm is not supported - we only support x64 on that.

PR #9664, opened 2017-02-18
Add stripSymbols argument; refactor symbol package build

dagood:
This adds a `stripSymbols` build arg to explicitly opt in to symbol stripping and applies a package refactor to avoid breaking when symbols don't exist.

Replaces my previous PR: https://github.com/dotnet/coreclr/pull/9326.

Package content diff between this PR and 25017-03, (`alpine`, `osx`, `win10-arm64`, `win7-x64` builds):
https://gist.github.com/dagood/8590cf9a13bc0c58a77736b068987870

Differences:

 * Cross-arch files are `AnyCPU` rather than `arm64`. This is existing behavior in the master branch right now, unrelated to these changes. @chcosta @ericstj was this intentional?
 * All symbol packages include `dll`/`so`/`dylib`. This follows symbol package convention more closely, and concretely means that indexing infra won't need to first merge the lib/sym packages together.
 * Files in `runtimes/win10-arm64/lib/netstandard1.0` now have their symbols placed in the same folder in the symbol package, rather than `runtimes/win10-arm64/native`.

PTAL, @chcosta @ericstj @gkhanna79 
FYI @ellismg

### DIFF
--- a/src/.nuget/dir.targets
+++ b/src/.nuget/dir.targets
@@ -1,0 +1,87 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!--
+    Finds symbol files and injects them into the package build.
+  -->
+  <Target Name="GetSymbolPackageFiles" BeforeTargets="GetPackageFiles">
+    <ItemGroup Condition="'$(SymbolFileExtension)' != ''">
+      <AdditionalLibPackageExcludes Include="%2A%2A\%2A$(SymbolFileExtension)"/>
+    </ItemGroup>
+
+    <ItemGroup>
+      <AdditionalLibPackageExcludes Include="@(LongNameFile -> '%(TargetPath)')" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <NativeWithSymbolFile Include="@(NativeBinary);
+                                     @(ArchitectureSpecificNativeFile);
+                                     @(ArchitectureSpecificNativeFileAndSymbol)">
+        <TargetPath>runtimes/$(PackageTargetRuntime)/native</TargetPath>
+      </NativeWithSymbolFile>
+      <!-- Using lib/netstandard1.0 here.  There is no TFM for this since it is a runtime itself. -->
+      <NativeWithSymbolFile Include="@(ArchitectureSpecificLibFile)">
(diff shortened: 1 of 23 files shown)


gkhanna79:
CC @janvorli @RussKeldorph

ellismg:
@chcosta @ericstj Could one of you please review?

src/.nuget/dir.targets
```diff
@@ -0,0 +1,87 @@
...
+        <IsSymbolFile>true</IsSymbolFile>
+      </File>
+    </ItemGroup>
+
+    <ItemGroup>
+      <!-- On Windows, trim ".dll" before adding ".pdb". -->
+      <WindowsNativeFile Include="@(NativeWithSymbolFile)"
+                         Condition="'%(NativeWithSymbolFile.Extension)'=='.dll' OR '%(NativeWithSymbolFile.Extension)'=='.exe'" />
+      <NativeDllFileWithoutExtension Include="@(WindowsNativeFile -> '%(RootDir)%(Directory)PDB\%(Filename)')" />
+      <WindowsSymbolFile Include="@(NativeDllFileWithoutExtension -> '%(Identity).pdb')" />
```
chcosta: Why not combine these two lines?

src/.nuget/dir.targets
```diff
@@ -0,0 +1,87 @@
...
+      <AdditionalLibPackageExcludes Include="%2A%2A\%2A$(SymbolFileExtension)"/>
+    </ItemGroup>
+
+    <ItemGroup>
+      <AdditionalLibPackageExcludes Include="@(LongNameFile -> '%(TargetPath)')" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <NativeWithSymbolFile Include="@(NativeBinary);
+                                     @(ArchitectureSpecificNativeFile);
```
chcosta: What's the difference between a "NativeBinary" item and an "ArchitectureSpecificNativeFile"?

dagood: I don't think anything. I can probably collapse `ArchitectureSpecificNativeFileAndSymbol` down, too.

A historical reason is that when there were tons of pkgprojs, I could avoid a bunch of changes by not changing the names. Now this isn't a problem.

src/.nuget/dir.targets
dagood: No reason, was just how I took it from CoreFX's copy. Didn't notice this, thanks.

dagood:

dagood:
@chcosta Does this look ok now?

PR #9667, opened 2017-02-18
Back-out promotion of structs containing fields of struct of single field

sivarv:
Struct promotion change was causing intermittent CoreFx test failure and blocking integrating new CoreCLR bits into CoreFX.  Will be re-enabling struct promotion after fixing the gc-stress issue caused by struct promotion.

CoreFX side of the issue: https://github.com/dotnet/corefx/issues/16279

### DIFF
--- a/src/jit/lclvars.cpp
+++ b/src/jit/lclvars.cpp
@@ -1543,60 +1543,8 @@ void Compiler::lvaCanPromoteStructType(CORINFO_CLASS_HANDLE    typeHnd,
 
             if (pFieldInfo->fldSize == 0)
             {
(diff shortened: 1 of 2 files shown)


sivarv:
@AndyAyersMS  - Please review this.
CC @dotnet/jit-contrib

AndyAyersMS:
LGTM, and verified this fixes the CoreFX test issue locally.

It's not a full revert of F686B5F but it looks like the parts you kept are ok.

sivarv:
Arm64 build failure is getting fixed by PR https://github.com/dotnet/coreclr/pull/9665 which is in flight.

PR #9676, opened 2017-02-20
Fix bogus assert in flowgraph and update related ilproj

### LINKED ISSUE #9576: JIT.Methodical.eh.disconnected.tryfinallyincatchtry_r test failure
The following assertion failure is observed while x86/Linux unit testing:
```
FAILED   - [ 3205 :    3 sec ] JIT/Methodical/eh/disconnected/tryfinallyincatchtry_r/tryfinallyincatchtry_r.sh
               BEGIN EXECUTION
               /root/dotnet-test/coreclr_unittest_20170214_103305_x86_x86-emu/coreclr-unittest-170214-013615/overlay/corerun tryfinallyincatchtry_r.exe

               Assert failure(PID 25283 [0x000062c3], Thread: 25283 [0x62c3]): Assertion failed 'lastTryBlock->getTryIndex() == XTnum' in '<Module>:main():int' (IL size 176)

                   File: /home/parjong/projects/dotnet/coreclr/src/jit/flowgraph.cpp Line: 23237
                   Image: /root/dotnet-test/coreclr_unittest_20170214_103305_x86_x86-emu/coreclr-unittest-170214-013615/overlay/corerun

               ./tryfinallyincatchtry_r.sh: line 131: 25283 Aborted                 (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" tryfinallyincatchtry_r.exe $CLRTestExecutionArguments
               Expected: 100
               Actual: 134
               END EXECUTION - FAILED
```

parjong:
End try block may not be directly contained as discussed in #9576, but there is an assert that considers  only direct cases.

This commit revises the assert to take care of indirect cases, and also revises relevant .ilproj file (to prevent future regression).

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -23286,8 +23286,7 @@ void Compiler::fgCloneFinally()
         BasicBlock* const firstTryBlock = HBtab->ebdTryBeg;
         BasicBlock* const lastTryBlock  = HBtab->ebdTryLast;
         assert(firstTryBlock->getTryIndex() == XTnum);
-        // TODO: rewrite the following assert to support nested try blocks
-        // assert(lastTryBlock->getTryIndex() == XTnum);
+        assert(bbInTryRegions(XTnum, lastTryBlock));
         BasicBlock* const beforeTryBlock = firstTryBlock->bbPrev;
 
         BasicBlock* normalCallFinallyBlock   = nullptr;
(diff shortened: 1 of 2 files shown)


src/jit/flowgraph.cpp
```diff
@@ -23286,7 +23286,8 @@ void Compiler::fgCloneFinally()
         BasicBlock* const firstTryBlock = HBtab->ebdTryBeg;
         BasicBlock* const lastTryBlock  = HBtab->ebdTryLast;
         assert(firstTryBlock->getTryIndex() == XTnum);
-        assert(lastTryBlock->getTryIndex() == XTnum);
+        // TODO: rewrite the following assert to support nested try blocks
+        // assert(lastTryBlock->getTryIndex() == XTnum);
```
parjong: @AndyAyersMS Thank you for comment. It works well.

Issue #9626, opened 2017-02-16
[x86/Linux] Unit test RecursiveTailCall failure with Generics


seanshpark:
JIT/CodeGenBringUpTests/RecursiveTailCall fails with

Unhandled Exception: GenericException`1[GenericClass`1[GenericClass`1[GenericClass`1[GenericClass`1[GenericClass`1[System.Int32]]]]]]: Exception of type 'GenericException`1[GenericClass`1[GenericClass`1[GenericClass`1[GenericClass`1[GenericClass`1[System.Int32]]]]]]' was thrown.

Test code:
```
        public static int TestGenericContext<T>(int x)
        {
            try
            {
                if (x == 1) throw new GenericException<T>();
            }
            catch (GenericException<T>)
            {
                return 1;
            }

            return x * TestGenericContext<GenericClass<T>>(x - 1);
        }
```

Catch argument `GenericClass<T>` seems to accumulate.

seanshpark:
Accumulation is also exist in working x64. Need more investigation.

seanshpark:
Changed test code
```
using System;

namespace Test
{
    class GenericClass<T> {}
    class GenericException<T> : Exception { }

    public class Program
    {
        public static int TestGenericContext<T>(int x)
        {
            try
            {
                throw new GenericException<T>();
            }
            catch (GenericException<T>)
            {
                return 100;
            }
            catch (Exception)
            {
                return 1;
            }
            return 0;
        }

        public static int Main(string[] args)
        {
            return TestGenericContext<GenericClass<int>>(1);
        }
    }
}
```

`Generics::GetExactInstantiationsOfMethodAndItsClassFromCallInformation()` failed.

seanshpark:
While Unwinding in EH, GetParamTypeArg() returns 0 and digging more into this and reason is that for x86/Linux, pCallerContext is valid and pCurrentConext* may not/

janvorli:
@seanshpark can you please explain what do you mean by pCurrentContext not being valid:
> pCallerContext is valid and pCurrentConext* may not

and how / with what call stack that happens in this case? I am reviewing your fix and I want to understand more details on the issue first.

parjong:
@janvorli Here is the stack trace when the execution hits ``EECodeManager::GetParamTypeArg``:
```
(gdb) bt 12
#0  EECodeManager::GetParamTypeArg (this=0x807b670, pContext=0xffffb048, pCodeInfo=0xffffb870)
[...]

PR #9678, opened 2017-02-20
[x86/Linux] Fix framepointer while unwinding

seanshpark:
Use pCallerContext when getting frame pointer

### DIFF
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -5169,7 +5169,11 @@ OBJECTREF EECodeManager::GetInstance( PREGDISPLAY    pContext,
     if (info.ebpFrame)
     {
         _ASSERTE(stackDepth == 0);
-        taArgBase = *pContext->GetEbpLocation();
+#if defined(WIN64EXCEPTIONS)
+        taArgBase = GetCallerSp(pContext) - 2 * sizeof(TADDR);
+#else
+        taArgBase = *pContext->pEbp;
+#endif
     }
     else
     {
@@ -5340,7 +5344,11 @@ PTR_VOID EECodeManager::GetParamTypeArg(PREGDISPLAY     pContext,
         return NULL;
     }
 
+#if defined(WIN64EXCEPTIONS)
+    TADDR fp = GetCallerSp(pContext) - 2 * sizeof(TADDR);
+#else
     TADDR fp = GetRegdisplayFP(pContext);
+#endif
     TADDR taParamTypeArg = *PTR_TADDR(fp - GetParamTypeArgOffset(&info));
     return PTR_VOID(taParamTypeArg);
 


seanshpark:
Fixes issue #9626

seanshpark:
This patch also fixes unit test
- JIT/Generics/Exceptions/specific_class_instance01
- JIT/Generics/Exceptions/specific_class_static01
- JIT/Generics/Exceptions/specific_struct_instance01

seanshpark:
@parjong, I 've found in https://github.com/dotnet/coreclr/blob/966e9b4e182278a4ad887f1546fb06f32db5b113/src/vm/eetwain.cpp#L5481 used `WIN64EXCEPTIONS`. Should I use this also?

src/vm/eetwain.cpp
```diff
@@ -5169,7 +5169,11 @@ OBJECTREF EECodeManager::GetInstance( PREGDISPLAY    pContext,
     if (info.ebpFrame)
     {
         _ASSERTE(stackDepth == 0);
+#if defined(_TARGET_X86_) && defined(FEATURE_PAL)
+        taArgBase = GetCallerSp(pContext) - 2 * sizeof(TADDR);
```
parjong: This code is already in ``_TARGET_X86_`` region, and thus ``_TARGET_X86_`` is unnecessary.

I think that it would be better to use ``WIN64EXCEPTIONS`` instead of ``FEATURE_PAL``.

src/vm/eetwain.cpp
```diff
@@ -5169,7 +5169,11 @@ OBJECTREF EECodeManager::GetInstance( PREGDISPLAY    pContext,
...
+#else
         taArgBase = *pContext->GetEbpLocation();
```
parjong: Now this region is not shared between Windows and Linux, and thus it would be better to use ``*pContext->pEbp`` instead of ``*pContext->GetEbpLocation()``.

src/vm/eetwain.cpp
seanshpark: Thanks!

src/vm/eetwain.cpp
seanshpark: It was originally `*pContext->GetEbpLocation()`. Should I change it?

seanshpark: Below line (5180 ?) has direct access to `pContext->SP`
```
taArgBase =  pContext->SP + stackDepth;
```
Would it be better to use like this?
```
taArgBase = pContext->SP - 2 * sizeof(TADDR);
```

parjong: @seanshpark It was ``*pContext->pEbp;``, but revised as the current form while x86/Linux bring up (#9121).

For 5180, it takes care of ``esp``-based frames (not ``ebp``-based frames), and thus it would be better to leave it as the current form.

parjong: @seanshpark If you intend to use ``pContext->SP`` instead of ``GetCallerSp(pContext)`` in 5173 line, I'm not sure about that. 

As I understand, ``pContext->SP`` stores current stack pointer, and thus it looks strange for me to  infer current frame pointer using ``pContext->SP``.

seanshpark: OK, thank you

src/vm/eetwain.cpp
```diff
@@ -5169,7 +5169,11 @@ OBJECTREF EECodeManager::GetInstance( PREGDISPLAY    pContext,
     if (info.ebpFrame)
     {
         _ASSERTE(stackDepth == 0);
-        taArgBase = *pContext->GetEbpLocation();
+#if defined(WIN64EXCEPTIONS)
+        taArgBase = GetCallerSp(pContext) - 2 * sizeof(TADDR);
+#else
+        taArgBase = *pContext->Ebp;
```
parjong: Use ``pEbp`` instead of ``Ebp`` here (please refer to inc/regdisp.h for details).

parjong: Or, it seems reasonable to use ``GetRegdisplayFP(pContext)`` here (similarly as below)

seanshpark: Thansk, I'll go with pEbp for now.

seanshpark:
@janvorli , PTAL

seanshpark:
Thank you!

Issue #9648, opened 2017-02-17
[RyuJIT/ARM32] GT_ARR_INDEX/GT_ARR_OFFSET support in codegen


alpencolt:
I've made implementation for `GT_ARR_INDEX` and `GT_ARR_OFFSET` instructions but can'not test it since compiler doesn't put them to the IR on tests. All array access are made buy using `GT_ARR_BOUNDS_CHECK` + `GT_LEA`.
Does them exist on lowering and codegen? Could you help with C# code which cover this test case?

mikedn:
Those are used for multidimensional arrays. As in:
```C#
int[,] a = new int[2, 3];
a[x, y] = 42;
```
There should be tests for that, try tests\src\JIT\Methodical\MDArray\ for example.

alpencolt:
Thank you!
But it works only in optimized build, the call `System.Int[,].Set` is being generated on debug version.

alpencolt:
@Dmitri-Botcharnikov

mikedn:
> But it works only in optimized build, the call System.Int[,].Set is being generated on debug version.

It's quite possible. IL only has instructions for SZArrays (single dimensional arrays with a 0 lower bound - `int[]` in C#) - `ldlen`, `ldelem`, `stelem` etc. For any other type of arrays (single dimensional arrays with a non-0 lower bound and multidimensional arrays) one needs to call array specific methods such as `Set`. It's up to the JIT compiler to treat such methods as intrinsics and turn them into special IR nodes. And the JIT may or may not do so depending on compilation options such as debug/release.

BruceForstall:
Fixed by #9679

PR #9679, opened 2017-02-20
[RyuJIT/ARM32] GT_ARR_INDEX/GT_ARR_OFFSET support in codegen.

alpencolt:
Code is mostly the same as for ARM64 except register size and some
emitted instructions.

### DIFF
--- a/src/jit/codegenarm.cpp
+++ b/src/jit/codegenarm.cpp
@@ -1002,6 +1002,14 @@ void CodeGen::genCodeForTreeNode(GenTreePtr treeNode)
             genTableBasedSwitch(treeNode);
             break;
 
+        case GT_ARR_INDEX:
+            genCodeForArrIndex(treeNode->AsArrIndex());
+            break;
+
+        case GT_ARR_OFFSET:
+            genCodeForArrOffset(treeNode->AsArrOffs());
+            break;
+
         case GT_IL_OFFSET:
             // Do nothing; these nodes are simply markers for debug info.
             break;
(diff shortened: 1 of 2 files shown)


alpencolt:
#9648 
@Dmitri-Botcharnikov

danmosemsft:
@dotnet/jit-contrib are you needed for CR on this older PR?

BruceForstall:
@danmosemsft yes. I'll look at it.

@alpencolt I assume you still want and need this PR merged?

alpencolt:
@BruceForstall yes, we are need in it. Also could you take  a look at #9979
Both PRs can be checked with tiny test, current with debug version and the second with optimized:
```C#
public class BringUpTest
{
    const int Pass = 100;
    const int Fail = -1;

    static int test0(int i) {
        int[] a = {1, 2, 3, 4};
        return a[i];
    }

    static void test1()
    {
        int[] a = {1, 2, 3, 4};
        a[1] = 5;
    }

    static int test2()
    {
        int[] a = {1, 2, 3, 4};
        return a[1];
    }

    static int test3()
    {
        int[] a = {1, 2, 3, 4};
        a[1] = 5;
        return a[1];
    }

    static int test4()
    {
        int[,] a = {{1, 2}, {3, 4}};
        return a[0, 1];
    }

    static int test5(int i)
    {
        Dummy[] a = {new Dummy(0), new Dummy(1), new Dummy(2), new Dummy(3), new Dummy(4),
                     new Dummy(5), new Dummy(6), new Dummy(7), new Dummy(8), new Dummy(9)};
        return a[i].field;
    }

    static int test6(int i)
    {
        int[][] a = new int[2][];
        a[0] = new int[2] {0, 1};
        a[1] = new int[2] {2, 3};
        return a[1][i];
    }

    static int test7(int x, int y)
    {
        int[,] a = new int[2, 3];
        a[x, y] = 42;
        return a[x, y];
    }

    static int test_ex()
    {
        int[] a = {1, 2, 3, 4};
        return a[5];
    }

    static int Main()
    {
        test0(1);

        test1();
        if (test2() != 2) return Fail;
        if (test3() != 5) return Fail;
        if (test4() != 2) return Fail;
        if (test5(1) != 1) return Fail;
        if (test6(1) != 3) return Fail;
        if (test7(1, 1) != 42) return Fail;

        try
        {
            test_ex();
            return Fail;
        }
        catch (System.IndexOutOfRangeException)
        {
            // OK
        }

        return Pass;
    }

    class Dummy
    {
        public int field;
        public Dummy(int f)
        {
            field = f;
        }
    }
}
```

BruceForstall:
It would be useful to add these test cases as a set of tests to the tests\src\JIT\CodeGenBringUpTests directory. It looks like that directory has very few array tests.

Note that tests in this directory follow a very simple pattern: a single function named the same as the filename, marked NoInlining, that tests the thing in question (e.g., arrays). The idea is that you set `COMPlus_AltJit=foo` for test foo.cs, with function foo(), and it tests exactly one thing.

BruceForstall:
LGTM

Issue #9556, opened 2017-02-13
[RyuJIT/ARM32] No reload of a trashed register


mskvortsov:
`r12` gets trashed by the helper call target :
<pre>
G_M46028_IG07:
000090  9803           ldr     r0, [sp+0x0c]	// [V00 this]
000092  68C0           ldr     r0, [r0+12]
000094  2800           cmp     r0, 0
000096  D120           bne     SHORT G_M46028_IG08
000098  F247 5000      movw    r0, 0x7500
00009C  F2CB 200E      movt    r0, 0xb20e
0000A0  F646 4371      movw    r3, 0x6c71
0000A4  F2CB 6374      movt    r3, 0xb674
0000A8  4798           blx     r3		// CORINFO_HELP_NEWSFAST
<b>0000AA  4684           mov     r12, r0</b>
0000AC  2100           movs    r1, 0
0000AE  F8CC 102C      str     r1, [r12+44]
0000B2  F241 0144      movw    r1, 0x1044
0000B6  F2CB 3190      movt    r1, 0xb390
0000BA  6809           ldr     r1, [r1]
0000BC  F10C 0014      add     r0, r12, 20
<b>0000C0  F24E 6CC1      movw    r12, 0xe6c1
0000C4  F2CB 6C7B      movt    r12, 0xb67b
0000C8  47E0           blx     r12		// CORINFO_HELP_ASSIGN_REF</b>
0000CA  9803           ldr     r0, [sp+0x0c]	// [V00 this]
<b>0000CC  4661           mov     r1, r12</b>
0000CE  2200           movs    r2, 0
0000D0  F24D 5359      movw    r3, 0xd559
0000D4  F2CB 5342      movt    r3, 0xb542
0000D8  4798           blx     r3		// System.AppDomain:SetupFusionStore(ref,ref):this
</pre>
Full JitDump is [here](https://gist.githubusercontent.com/mskvortsov/339d50ebfddaf9bb4bde05f18799faa1/raw/170a7a42b256540aa0af4f8f3b45239d4d84111e/SetupDomain.txt).
Reproduces with rel or chk builds (i.e. when compiling S.P.CoreLib.dll in the opt mode):
```
$ COMPlus_AltJit=System.AppDomain:SetupDomain corerun Any.exe
```
(This issue is a part of #8496.)

mskvortsov:
The codegen log for that output:
<pre>
							GC regs: 0000 {} => 0001 {r0}
[...]

PR #9681, opened 2017-02-20
[RyuJIT/ARM32] Fix helper kill mask and call target consuming

mskvortsov:
Fix for a couple of issues described in #9556.
(Re-)enables rel/chk altjit compilation of S.P.CoreLib.dll methods for the `JIT/CodeGenBringUpTests`.

(This is a part of #8496.)


mskvortsov:

BruceForstall:
I think you should instead change the target.h code to read:

```
#ifdef LEGACY_BACKEND
  #define RBM_CALLEE_TRASH_NOGC   (RBM_R2|RBM_R3|RBM_LR)
#else
  #define RBM_CALLEE_TRASH_NOGC   RBM_CALLEE_TRASH
#endif
```

And then in getKillSetForNode():
```
        case GT_STOREIND:
            if (compiler->codeGen->gcInfo.gcIsWriteBarrierAsgNode(tree))
            {
                killMask = RBM_CALLEE_TRASH_NOGC;
#if !NOGC_WRITE_BARRIERS && (defined(_TARGET_ARM_) || defined(_TARGET_AMD64_))
                killMask |= (RBM_ARG_0 | RBM_ARG_1);
#endif // !NOGC_WRITE_BARRIERS && (defined(_TARGET_ARM_) || defined(_TARGET_AMD64_))
            }
```

delete the `#if` condition. Note that `NOGC_WRITE_BARRIERS` is zero for both ARM and AMD64 in all cases. AMD64 already has:
```
  #define RBM_CALLEE_TRASH_NOGC   RBM_CALLEE_TRASH
```

Thus, that `#if` would no longer do anything. Thus, the code should simply read:
```
        case GT_STOREIND:
            if (compiler->codeGen->gcInfo.gcIsWriteBarrierAsgNode(tree))
            {
                killMask = RBM_CALLEE_TRASH_NOGC;
            }
```

(fyi, the legacy back end picks a register very late, in code generation, and R4 happens to be available -- its register allocator just makes sure at least one is free. In RyuJIT, we have to choose in advance which registers to use. We can use R12 because it is callee-trash and not an argument register.)

BruceForstall:
The codegenarm.cpp change looks fine.

BruceForstall:
@mskvortsov This looks good to me. Can you squash the commits?

BruceForstall:

Issue #9068, opened 2017-01-23
[RyuJit] Declaring vars causes stackspace spilling


benaadams:
Optimized code reserves stack space for pre-optimized stack

e.g. The following code reserves 136 bytes of stack, and zeros out 24 bytes (which it uses) even though I'd reason its output and output from https://github.com/dotnet/coreclr/issues/9066 should be identical
```
; Lcl frame size = 136

G_M2227_IG01:
       57                   push     rdi
       56                   push     rsi
       4881EC88000000       sub      rsp, 136          // 136 bytes reserved
       488BF1               mov      rsi, rcx
       488D7C2428           lea      rdi, [rsp+28H]
       B918000000           mov      ecx, 24           // 24 bytes zero'd
       33C0                 xor      rax, rax
       F3AB                 rep stosd 
```
Code
```csharp
[MethodImpl(MethodImplOptions.NoInlining)]
private static byte LotsOfSpans(byte[] array)
{
    var span00 = new Span<byte>(array);
    var span01 = span00.Slice(1);
    var span02 = span01.Slice(1);
    var span03 = span02.Slice(1);
    var span04 = span03.Slice(1);
    var span05 = span04.Slice(1);
    var span06 = span05.Slice(1);
    var span07 = span06.Slice(1);
    var span08 = span07.Slice(1);
    var span09 = span08.Slice(1);
    var span10 = span09.Slice(1);
    var span11 = span10.Slice(1);
    var span12 = span11.Slice(1);
    var span13 = span12.Slice(1);
    var span14 = span13.Slice(1);
    var span15 = span14.Slice(1);
    var span16 = span15.Slice(1);
    var span17 = span16.Slice(1);
    var span18 = span17.Slice(1);
    var span19 = span18.Slice(1);
    var span20 = span19.Slice(1);
    var span21 = span20.Slice(1);
    var span22 = span21.Slice(1);
    var span23 = span22.Slice(1);
    var span24 = span23.Slice(1);
    var span25 = span24.Slice(1);
    var span26 = span25.Slice(1);
    var span27 = span26.Slice(1);
    var span28 = span27.Slice(1);
    var span29 = span28.Slice(1);
    var span30 = span29.Slice(1);
    return span30[0];
}
```
Asm produced
```asm
; Lcl frame size = 136

G_M2227_IG01:
       57                   push     rdi
       56                   push     rsi
       4881EC88000000       sub      rsp, 136
       488BF1               mov      rsi, rcx
[...]

Issue #1691, opened 2015-10-06
Inlining in shared generic code leaves unused parts of dictionary lookups around


VSadov:
``` C#

    class Program
    {
        static void Main(string[] args)
        {
            new Generic<long, object>().test();
        }
    }

    class Generic<T, U>
    {
        private struct S1
        {
            public T field;
            public object stuff;
        }

        private S1[] arr = new S1[43];

        private static int GetLength(S1[] arr)
        {
            // last element is special and does not count towards length
            return arr.Length - 1;
        }

        public virtual void test()
        {
            System.Console.WriteLine(GetLength(arr));  // <-- stop here
        }

    }
```

The body of test() is:

```
            System.Console.WriteLine(GetLength(arr));  // <-- stop here
00000000  push        rsi 
00000001  sub         rsp,30h                                        
00000005  mov         qword ptr [rsp+28h],rcx        // ???????
0000000a  mov         rsi,rcx 
0000000d  mov         rcx,qword ptr [rsi+8] 
00000011  mov         ecx,dword ptr [rcx+8] 
00000014  dec         ecx 
00000016  call        0000000052CBEBF0 
0000001b  nop 
        }
0000001c  add         rsp,30h 
00000020  pop         rsi 
00000021  ret 
```

If I make GetLength an instance method, the strange store disappears and stack frame is smaller.

```
            System.Console.WriteLine(GetLength(arr));  // <-- stop here
00000000  sub         rsp,28h 
00000004  mov         rcx,qword ptr [rcx+8] 
00000008  mov         ecx,dword ptr [rcx+8] 
0000000b  dec         ecx 
0000000d  call        0000000052CBEC00 
00000012  nop 
        }
00000013  add         rsp,28h 
00000017  ret 
```

It seems unnatural that I can improve CQ by making static methods into instance methods in scenario like this, but it is very tempting when the pattern is on a hot path.

mikedn:
It looks like lvaGenericsContextUsed is set when the call to GetLength is imported because GetLength is static and it needs the generic context argument (even if it doesn't use it). After inlining the compiler figures out that the whole generic context thing isn't needed and discards the code but it doesn't reset lvaGenericsContextUsed. In the end genFnProlog uses lvaGenericsContextUsed and forces this to be spilled.

AndyAyersMS:
I have a prototype that gets this down to
```asm
G_M16884_IG01:
       4883EC28             sub      rsp, 40

G_M16884_IG02:
[...]

PR #9683, opened 2017-02-20
JIT: defer setting outgoing args size until after optimization

AndyAyersMS:
For fixed out args ABIs, the jit currently computes the size of the
out arg area during morph, as it encounters calls that must pass arguments
via the stack.

Subsequently, the optimizer may delete calls, either because they are
pure and the results are unused, or because they are conditional and the
guarding predicates are resolved by the optimizer in such a way that
the call is no longer reachable.

In particular if all the calls seen by morph are subsequently removed by
the optimizer, the jit may generate a stack frame that it never uses and
doesn't need. If only some calls are removed then the stack frame may end
up larger than necessary.

One motivating example is the inlining of a shared generic method that
ignores its generic context parameter. The caller typically must invoke a pure
helper to determine the proper argument to pass. Post inline the call's result
is unused, and the helper call is deleted by the optimizer.

This change defers the outgoing arg size computation until fgSimpleLowering,
which runs after optimization. The code in morph now simply records the
needed size for each call in the associated fgArgInfo.

As before, the outgoing arg size computation ignores fast tail calls, since
they can use the caller-supplied scratch area for memory arguments.

The jit may introduce helper calls after optimization which could seemingly
invalidate the out args area size. The presumption here is that these calls
are carefully designed not to use the caller-supplied scratch area. The
current code makes the same assumption.

This change introduces some unanticipated diffs. Optcse estimates the
frame size to decide how aggressive it should be about cses of local
variable references. This size estimate included the out args area, which
appears to be something of an oversight; there are no cse opportunities
in this area and the offsets of the other local variables are not impacted
by the size of this area. With this change the out args area size is seen
as zero during optimization and so the cse strategy for a method may be
different. Arguably, the new behavior is more sensible.

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -9040,11 +9040,11 @@ void Compiler::fgSimpleLowering()
                 {
                     GenTreeCall* call = tree->AsCall();
                     // Fast tail calls use the caller-supplied scratch
-                    // space so have no impact on this methods outgong arg size.
+                    // space so have no impact on this method's outgoing arg size.
                     if (!call->IsFastTailCall())
                     {
                         // Update outgoing arg size to handle this call
-                        const unsigned thisCallOutAreaSize = tree->AsCall()->fgArgInfo->GetOutArgSize();
+                        const unsigned thisCallOutAreaSize = call->fgArgInfo->GetOutArgSize();
                         assert(thisCallOutAreaSize >= MIN_ARG_AREA_FOR_CALL);
 
                         if (thisCallOutAreaSize > lvaOutgoingArgSpaceSize)
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -9021,6 +9021,49 @@ void Compiler::fgSimpleLowering()
                     // Add in a call to an error routine.
                     fgSetRngChkTarget(tree, false);
                 }
+
+#if FEATURE_FIXED_OUT_ARGS
+                if (tree->gtOper == GT_CALL)
+                {
+                    GenTreeCall* call = tree->AsCall();
+                    // Fast tail calls use the caller-supplied scratch
+                    // space so have no impact on this methods outgong arg size.
+                    if (!call->IsFastTailCall())
+                    {
+                        // Update outgoing arg size to handle this call
+                        const unsigned thisCallOutAreaSize = tree->AsCall()->fgArgInfo->GetOutArgSize();
+                        assert(thisCallOutAreaSize >= MIN_ARG_AREA_FOR_CALL);
+
+                        if (thisCallOutAreaSize > lvaOutgoingArgSpaceSize)
+                        {
+                            lvaOutgoingArgSpaceSize = thisCallOutAreaSize;
+
+                            // If a function has localloc, we will need to move the outgoing arg space when the
+                            // localloc happens. When we do this, we need to maintain stack alignment. To avoid
+                            // leaving alignment-related holes when doing this move, make sure the outgoing
+                            // argument space size is a multiple of the stack alignment by aligning up to the next
+                            // stack alignment boundary.
+                            if (compLocallocUsed)
+                            {
+                                lvaOutgoingArgSpaceSize = (unsigned)roundUp(lvaOutgoingArgSpaceSize, STACK_ALIGN);
+                            }
+
+                            JITDUMP("Bumping lvaOutgoingArgSpaceSize to %u for call [%06d]\n", lvaOutgoingArgSpaceSize,
+                                    dspTreeID(tree));
+                        }
+                        else
+                        {
+                            JITDUMP("lvaOutgoingArgSpaceSize %u sufficient for call [%06d], which needs %u\n",
+                                    lvaOutgoingArgSpaceSize, dspTreeID(tree), thisCallOutAreaSize);
+                        }
+                    }
+                    else
+                    {
+                        JITDUMP("lvaOutgoingArgSpaceSize not impacted by fast tail call [%06d]\n", dspTreeID(tree))
+                    }
+                }
+
+#endif // FEATURE_FIXED_OUT_ARGS
             }
         }
     }
@@ -12127,7 +12170,8 @@ bool Compiler::fgRelocateEHRegions()
                 }
 
                 // Currently it is not good to move the rarely run handler regions to the end of the method
-                // because fgDetermineFirstColdBlock() must put the start of any handler region in the hot section.
+                // because fgDetermineFirstColdBlock() must put the start of any handler region in the hot
+                // section.
                 CLANG_FORMAT_COMMENT_ANCHOR;
 
 #if 0
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -8954,11 +8954,13 @@ void Compiler::fgFindOperOrder()
 
 void Compiler::fgSimpleLowering()
 {
+#if FEATURE_FIXED_OUT_ARGS
     unsigned outgoingArgSpaceSize = 0;
+#endif // FEATURE_FIXED_OUT_ARGS
 
     for (BasicBlock* block = fgFirstBB; block; block = block->bbNext)
     {
-        // Walk the statement trees in this basic block, converting ArrLength nodes.
+        // Walk the statement trees in this basic block.
         compCurBB = block; // Used in fgRngChkTarget.
 
 #ifdef LEGACY_BACKEND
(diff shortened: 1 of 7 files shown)


AndyAyersMS:
Marking as no merge for now since there are some questions to sort out.

@BruceForstall PTAL
cc @dotnet/jit-contrib

benaadams:
Are there examples of this in these issues? (other things like at play too) https://github.com/dotnet/coreclr/issues/9066 https://github.com/dotnet/coreclr/issues/9068 https://github.com/dotnet/coreclr/issues/9041

AndyAyersMS:
This is prep work for #1691. A few examples:
```asm
;; System.ThrowHelper:IfNullAndNullsAreIllegalThenThrow(ref,int)

;; Before
; Lcl frame size = 40
G_M64977_IG01:
       sub      rsp, 40
       nop      
G_M64977_IG02:
       add      rsp, 40
       ret

;; After
; Lcl frame size = 0
G_M64977_IG01:
       nop      
G_M64977_IG02:
       ret

;; System.StubHelpers.HStringMarshaler:ConvertToManaged(long):ref

;; Before
; Lcl frame size = 40
G_M48743_IG01:
       sub      rsp, 40
       nop      
G_M48743_IG02:
       lea      rax, [(reloc)]
G_M48743_IG03:
       add      rsp, 40
       rex.jmp  rax

;; After
; Lcl frame size = 0
G_M48743_IG01:
       nop      
G_M48743_IG02:
       lea      rax, [(reloc)]
G_M48743_IG03:
       rex.jmp  rax
```
The issues you cited above look like either unreferenced locals occupying stack or (lack of) stack packing.

AndyAyersMS:
I may be overly enthusiastic about the new behavior of optcse here -- while it's true the outgoing arg area size won't impact the offset of the local from from the frame pointer, the jit might be using the stack pointer to access locals, in which case the out area arg size will impact the sp-relative offset. But optcse doesn't take FP/SP based access into account either.

At any rate, this only affects a handful of methods and the code size impact is essentially a wash; jit-diffs gives -167 bytes net diff across 313 methods with size diffs.

Methods that only had decreased stack sizes won't show up in jit-diff stats since the method sizes will likely be identical.

It would be easy enough to restore the old behavior by computing the size in morph and then recomputing it later on.

briansull:
Also take a look at  Compiler::lvaAllocOutgoingArgSpace() 
I believe that it handles cases similar to what you mention with MEMCPY

        if (lvaOutgoingArgSpaceSize == 0)
        {
            if (compUsesThrowHelper || compIsProfilerHookNeeded())
            {
                // Need to make sure the MIN_ARG_AREA_FOR_CALL space is added to the frame if:
                // 1. there are calls to THROW_HEPLPER methods.
                // 2. we are generating profiling Enter/Leave/TailCall hooks. This will ensure
                //    that even methods without any calls will have outgoing arg area space allocated.
                //
                // An example for these two cases is Windows Amd64, where the ABI requires to have 4 slots for
                // the outgoing arg space if the method makes any calls.
                lvaOutgoingArgSpaceSize = MIN_ARG_AREA_FOR_CALL;
            }
        }

AndyAyersMS:
Ran subset of desktop tests too, all look good. Desktop SPMI shows small aggregate code size win as well.

AndyAyersMS:
@briansull yeah I saw that bit of code. Looks like most helpers initially dispatch via assembly and don't assume the scratch area is available. Seems like this has to be the case, otherwise we'd see widespread failures.

BruceForstall:
@AndyAyersMS But we don't have any ELT tests in the open (I think). And those calls aren't via helpers, are they?

src/jit/flowgraph.cpp
```diff
@@ -9021,6 +9021,49 @@ void Compiler::fgSimpleLowering()
                     // Add in a call to an error routine.
                     fgSetRngChkTarget(tree, false);
                 }
+
+#if FEATURE_FIXED_OUT_ARGS
+                if (tree->gtOper == GT_CALL)
```
BruceForstall: Can you please (1) add a standard function header to this function that describes its purpose and components, and (2) write an overall comment for this new code that mentions *why* it is located here (e.g., "this happens to be a convenient place that follows all optimization that does a full IR pass")?

src/jit/flowgraph.cpp
```diff
@@ -9021,6 +9021,49 @@ void Compiler::fgSimpleLowering()
...
+                {
```
BruceForstall: Maybe you should change this function to 'switch' over tree->gtOper.

src/jit/flowgraph.cpp
```diff
@@ -9021,6 +9021,49 @@ void Compiler::fgSimpleLowering()
...
+                    GenTreeCall* call = tree->AsCall();
+                    // Fast tail calls use the caller-supplied scratch
+                    // space so have no impact on this methods outgong arg size.
```
BruceForstall: >methods outgong [](start = 55, length = 15)

typo/grammar: "method's outgoing"

AndyAyersMS:
If ELT is the profiler hook, then its needs are handled specially via `lvaAllocOutgoingArgSpace`; that hasn't changed.

src/jit/flowgraph.cpp
```diff
@@ -9021,6 +9021,49 @@ void Compiler::fgSimpleLowering()
...
+                    if (!call->IsFastTailCall())
+                    {
+                        // Update outgoing arg size to handle this call
+                        const unsigned thisCallOutAreaSize = tree->AsCall()->fgArgInfo->GetOutArgSize();
```
BruceForstall: >tree->AsCall() [](start = 61, length = 14)

`tree->AsCall()` => `call`

BruceForstall:
I wonder if it would be possible, and advisable, to change `lvaOutgoingArgSpaceSize ` to a `PhasedVar<unsigned>`. Then, we would get asserts if the value is attempted to be written after it has been read. The idea of `PhasedVar` is to prevent phases from reading and taking a dependency on a value that will later change.

src/jit/flowgraph.cpp
AndyAyersMS: I though about that worried that calls could be created in `fgSetRngChkTarget` -- though it looks like it doesn't actually add calls despite the comment.

AndyAyersMS:
The `PhasedVar` approach seems sensible. I'll look into it.

briansull:
LGTM

AndyAyersMS:
@BruceForstall Updated per feedback (though no switch yet). See what you think.

When using `PhasedVar`, I ended up consolidating all the out arg area size updates into `fgSimpleLowering` and having optcse avoid looking at the size.

There may be some oddness with having constructed data members on the compiler object; need to look more closely to see if we do anything funny there with the backing memory. Noticed this when I added a tracking string flag to the PhasedVar so I could track down when the reads were happening.

src/jit/flowgraph.cpp
```diff
@@ -9021,10 +9036,76 @@ void Compiler::fgSimpleLowering()
...
+                            // If a function has localloc, we will need to move the outgoing arg space when the
+                            // localloc happens. When we do this, we need to maintain stack alignment. To avoid
+                            // leaving alignment-related holes when doing this move, make sure the outgoing
+                            // argument space size is a multiple of the stack alignment by aligning up to the next
+                            // stack alignment boundary.
+                            if (compLocallocUsed)
+                            {
+                                outgoingArgSpaceSize = (unsigned)roundUp(outgoingArgSpaceSize, STACK_ALIGN);
+                            }
+
```
BruceForstall: You could pull the localloc update/alignment down below, just before setting the actual size.

AndyAyersMS: Yeah, good idea, since it really only needs to happen once.

BruceForstall:
Looks like `Compiler::compInit()` should probably call the PhasedVar constructor explicitly. We don't create Compiler with "new" so member var ctors don't get invoked. (I suppose that's because `Compiler::new` calls `compGetMem` which uses the allocator stored on the compiler instance itself.)

BruceForstall:
I like making new things PhasedVar; it improves our ability to reason about the "global" state change variables.

src/jit/utils.h
```diff
@@ -381,6 +381,15 @@ class PhasedVar
         return m_value;
     }
 
+    // Mark the value as read only
+    void MarkAsReadOnly() const
```
BruceForstall: Let me suggest:
``
    // Mark the value as read only; explicitly change the variable to the "read" phase.
``

AndyAyersMS:
x86 needs more FEATURE_FIXED_OUT_ARGS magic.... will update shortly.

AndyAyersMS:
Maybe set it up over in `lvaInit` along with the other fixed frame vars?

src/jit/utils.h
AndyAyersMS: Sure

AndyAyersMS:
Looks like I need a default case label to appease unix.

BruceForstall:
LGTM

AndyAyersMS:
@dotnet-bot retest OSX x64 Checked Build and Test
@dotnet-bot retest Ubuntu x64 Checked Build and Test

PR #9684, opened 2017-02-21
Fixed filepath error in some tests.

aargor:


aargor:
Hello,

I'm Aaron Gorenstein (aargor), a software engineer on the VC++ compiler. I think you're already familiar with how we build a number of OSS "real-world-code" projects, including CoreCLR.

Recently, we've found that some of your tests failed because of what appears to be a small path error in the test files. We have a stock CoreCLR enlistment, so I think this path (environment) issue is real. Please see the commits that we've found fixed the issue in this PR.

I'm guessing this got around your rolling tests I see on GitHub because I don't think these tests are included in your rolling builds, but we run "tests\runtest.cmd x86 release" (and x64).

Thanks,
Aaron

aargor:
Ah, I still managed to misunderstand the environment variables. My mistake, sorry for the noise.

PR #9685, opened 2017-02-21
Clean up FixupDispatcherContext

parjong:
Currently, the functions in exceptionhanding.cpp related with FixupDispatcherContext are used only in Windows.

This commit encloses Windows-specific functions with FEATURE_PAL ifdef.


parjong:
@janvorli Please take a look.

PR #9686, opened 2017-02-21
Use ADJUST_PC_UNWOUND_TO_CALL instead of _TARGET_XXX_

parjong:
These fragments seem to be related with each other, but it is hard to recognize its relationship.

This commit introduces ADJUST_PC_UNWOUND_TO_CALL, and uses it instead of explicitly enumerating _TARGET_XXX_ on each fragment.


parjong:
\CC @janvorli @BruceForstall @sdmaclea

PR #9688, opened 2017-02-21
[x86/Linux] Emit proper frame for ArrayOpStub

### LINKED ISSUE #9687: [x86/Linux] PAL_SEHException is thrown while running b119538 tests
Here is the detailed logs from the test log:
```
FAILED   - [ 4997 :    3 sec ] JIT/Regression/CLR-x86-JIT/V1-M14-SP1/b119538/b119538a/b119538a.sh
               BEGIN EXECUTION
               /root/dotnet-test/coreclr_unittest_20170217_164016_x86_x86-emu/coreclr-unittest-170217-074046/overlay/corerun b119538a.exe
               terminate called after throwing an instance of 'PAL_SEHException'
               ./b119538a.sh: line 131:  2789 Aborted                 (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" b119538a.exe $CLRTestExecutionArguments
               Expected: 100
               Actual: 134
               END EXECUTION - FAILED
FAILED   - [ 4998 :    3 sec ] JIT/Regression/CLR-x86-JIT/V1-M14-SP1/b119538/b119538b/b119538b.sh
               BEGIN EXECUTION
               /root/dotnet-test/coreclr_unittest_20170217_164016_x86_x86-emu/coreclr-unittest-170217-074046/overlay/corerun b119538b.exe
               terminate called after throwing an instance of 'PAL_SEHException'
               ./b119538b.sh: line 131:  2809 Aborted                 (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" b119538b.exe $CLRTestExecutionArguments
               Expected: 100
               Actual: 134
               END EXECUTION - FAILED
```

parjong:
In x86/Linux, C++ unwinder failed to unwind the frame by ArrayOpStub as the current implementation does not emit a proper frame (discussed in #9687).

This commit revises ArrayOpStub to have a ebp frame to fix #9687.

### DIFF
--- a/src/vm/i386/asmhelpers.S
+++ b/src/vm/i386/asmhelpers.S
@@ -263,6 +263,7 @@ LEAF_ENTRY ArrayOpStubNullException, _TEXT
     // them anyway for consistency and to avoid future bugs.
     pop     esi
     pop     edi
+    pop     ebp
     mov     ecx, CORINFO_NullReferenceException_ASM
     jmp     C_FUNC(InternalExceptionWorker)
 LEAF_END ArrayOpStubNullException, _TEXT
@@ -274,6 +275,7 @@ LEAF_ENTRY ArrayOpStubRangeException, _TEXT
     // them anyway for consistency and to avoid future bugs.
     pop     esi
     pop     edi
+    pop     ebp
     mov     ecx, CORINFO_IndexOutOfRangeException_ASM
     jmp     C_FUNC(InternalExceptionWorker)
 LEAF_END ArrayOpStubRangeException, _TEXT
@@ -285,6 +287,7 @@ LEAF_ENTRY ArrayOpStubTypeMismatchException, _TEXT
     // them anyway for consistency and to avoid future bugs.
     pop     esi
     pop     edi
+    pop     ebp
     mov     ecx, CORINFO_ArrayTypeMismatchException_ASM
     jmp     C_FUNC(InternalExceptionWorker)
 LEAF_END ArrayOpStubTypeMismatchException, _TEXT
--- a/src/vm/i386/stublinkerx86.cpp
+++ b/src/vm/i386/stublinkerx86.cpp
@@ -5079,6 +5079,11 @@ VOID StubLinkerCPU::EmitArrayOpStub(const ArrayOpScript* pArrayOpScript)
         }
     }
 #else
+    // Emit frame
+    X86EmitPushEBPframe();
+
+    ofsadjust += sizeof(void*);
+
     // Preserve the callee-saved registers
     // NOTE: if you change the sequence of these pushes, you must also update:
     //  ArrayOpStubNullException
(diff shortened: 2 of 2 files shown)


parjong:
\CC @wateret @seanshpark

parjong:

janvorli:
@parjong what was the problem in the unwind? Was it just unable to get past the frame of the asm helper (getting wrong EIP, ESP or both)? If that's the case, we should be able to fix it by adding CFI instructions to the asm helpers instead to teach the unwinder how to unwind from these. 
I think that adding `.cfi_def_cfa_offset 12` as the first line of the ArrayOpStubXXX functions would work without adding the ebp pushing / popping.

parjong:
@janvorli An exception is throw inside ``ArrayStoreCheck`` called from generated stub (not one of asm helpers).

Here is the stack trace when we hit the issue:
```
(gdb) bt
#0  0xf7fd8be9 in __kernel_vsyscall ()
#1  0xf7c38ea9 in __GI_raise (sig=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#2  0xf7c3a407 in __GI_abort () at abort.c:89
#3  0xf7ea7d35 in __gnu_cxx::__verbose_terminate_handler() () from /usr/lib/i386-linux-gnu/libstdc++.so.6
#4  0xf7ea5833 in ?? () from /usr/lib/i386-linux-gnu/libstdc++.so.6
#5  0xf7ea58ad in std::terminate() () from /usr/lib/i386-linux-gnu/libstdc++.so.6
#6  0xf7ea5b70 in __cxa_throw () from /usr/lib/i386-linux-gnu/libstdc++.so.6
#7  0xf7137776 in DispatchManagedException (ex=..., isHardwareException=false)
    at /home/parjong/projects/dotnet/coreclr/src/vm/exceptionhandling.cpp:4667
#8  0xf6fadb8f in ArrayStoreCheck (pArray=0xffffc5ac, pElement=0xffffc5c0) at /home/parjong/projects/dotnet/coreclr/src/vm/jithelpers.cpp:3368
#9  0xf631b669 in ?? ()
```

janvorli:
@parjong oh, thanks for the explanation. But then the fixed worries me. The thing is that we unwind the stub that's generated as asm as native code using the libunwind. But that stub has no DWARF unwind info and it seems that the fact that libunwind can unwind through it is kind of luck (it probably has a fallback to use ebp if it doesn't find DWARF unwind info).
I would feel much better if we could fix it by switching x86 on Linux to FEATURE_ARRAYSTUB_AS_IL. @jkotas, do you think it would be feasible or are there some pitfalls?

jkotas:
> switching x86 on Linux to FEATURE_ARRAYSTUB_AS_IL

I do not see any problems with it

parjong:
@janvorli As you pointed out, it seems that libunwind runs in fallback mode (based on minimal ebp frame).

Non-volatile register fixup using helper frame (in 2nd pass) seems to mitigate the issues from incorrect register values (May this explain why this patch works), but I agree with you that it would be better to enable FEATURE_ARRAY_STUBS_AS_IL. 

I'll take a look.

PR #9697, opened 2017-02-21
Use CORINFO_CALLCONV to check for unmanaged calls

MichalStrehovsky:
`GTF_CALL_UNMANAGED` is a member of the wrong enum and it's something
that would be set on `call->gtFlags` except that at this point it isn't.
Check against the actual unmanaged calling conventions.

### DIFF
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -6465,9 +6465,9 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
         if (IsTargetAbi(CORINFO_CORERT_ABI))
         {
             bool managedCall = (((calliSig.callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_STDCALL) &&
-                ((calliSig.callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_C) &&
-                ((calliSig.callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_THISCALL) &&
-                ((calliSig.callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_FASTCALL));
+                                ((calliSig.callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_C) &&
+                                ((calliSig.callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_THISCALL) &&
+                                ((calliSig.callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_FASTCALL));
             if (managedCall)
             {
                 addFatPointerCandidate(call->AsCall());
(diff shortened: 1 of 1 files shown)


MichalStrehovsky:
@sandreenko Please take a look

sandreenko:
Where the old version produce the wrong result? 
Do we have this difference only on CoreRT?

MichalStrehovsky:
> Where the old version produce the wrong result?

We have about 60 test failures in the CoreRT due to fat function pointer transform applying to p/invokes:
https://ci.dot.net/job/dotnet_corert/job/master/job/debug_windows_nt/292/consoleText

> Do we have this difference only on CoreRT?

The code I'm touching is under `IsTargetAbi(CORINFO_CORERT_ABI))`, so I guess this is CoreRT only. You wrote the code :P.

sandreenko:
```
bool managedCallOld = (calliSig.callConv & GTF_CALL_UNMANAGED) == 0;
bool managedCallNew = 
               (((calliSig.callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_STDCALL) &&  
                ((calliSig.callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_C) &&  
                ((calliSig.callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_THISCALL) &&  
                ((calliSig.callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_FASTCALL));

assert(managedCallOld == managedCallNew);

```

@BruceForstall  Do we expect this code to fail in CoreCLR? 
According to [msdn ](https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.callingconvention(v=vs.110).aspx) it shouldn't (where is Winapi enum member?).

MichalStrehovsky:
@sandreenko Maybe I wasn't clear in the first post: `calliSig` is a `CORINFO_SIG_INFO` (an exchange type with the EE). EE doesn't know anything about `GTF_CALL_UNMANAGED`.

`GTF_CALL_UNMANAGED` is something RyuJIT sets on gtFlags. It makes no sense to compare possible values of `gtFlags` with a `CorInfoCallConv`. It's a different enum.

sandreenko:
Yes, GTF_CALL_UNMANAGED is set on line after it was previously checked(7084 impCheckForPInvokeCall).
But looks like after this line this condition is not 100% true, because on line 7102 we check all conditions manually.
```
    else if ((opcode == CEE_CALLI) && (((sig->callConv & CORINFO_CALLCONV_MASK) == CORINFO_CALLCONV_STDCALL) ||
                                       ((sig->callConv & CORINFO_CALLCONV_MASK) == CORINFO_CALLCONV_C) ||
                                       ((sig->callConv & CORINFO_CALLCONV_MASK) == CORINFO_CALLCONV_THISCALL) ||
                                       ((sig->callConv & CORINFO_CALLCONV_MASK) == CORINFO_CALLCONV_FASTCALL)))
```

So, I think it is better to add 
```
        if (IsTargetAbi(CORINFO_CORERT_ABI))
        {
                addFatPointerCandidate(call->AsCall());           
        }
```
undo this condition.

sandreenko:
LGTM

PR #9702, opened 2017-02-21
Fix x86 perf testing

adiaaida:
We did not use a different architecture in run-xunit-perf.cmd, so all the
x86 runs were uploading to the same architecture. This change fixes it so
x86jit32 runs upload to their own architecture. Also fixes the delayed
expansion issue for testenv.

### DIFF
--- a/perf.groovy
+++ b/perf.groovy
@@ -74,8 +74,8 @@ def static getOSGroup(def os) {
 					batchFile("py \"%WORKSPACE%\\Microsoft.BenchView.JSONFormat\\tools\\machinedata.py\"")
 					batchFile("set __TestIntermediateDir=int&&build.cmd ${configuration} ${architecture}")
 					batchFile("tests\\runtest.cmd ${configuration} ${architecture} GenerateLayoutOnly")
-					batchFile("tests\\scripts\\run-xunit-perf.cmd -arch ${architecture} -configuration ${configuration} ${testEnv} -testBinLoc bin\\tests\\Windows_NT.${architecture}.${configuration}\\performance\\perflab\\Perflab -library -uploadToBenchview \"%WORKSPACE%\\Microsoft.Benchview.JSONFormat\\tools\" -runtype ${runType}")
-					batchFile("tests\\scripts\\run-xunit-perf.cmd -arch ${architecture} -configuration ${configuration} ${testEnv} -testBinLoc bin\\tests\\Windows_NT.${architecture}.${configuration}\\Jit\\Performance\\CodeQuality -uploadToBenchview \"%WORKSPACE%\\Microsoft.Benchview.JSONFormat\\tools\" -runtype ${runType}")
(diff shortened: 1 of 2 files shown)


adiaaida:
@DrewScoggins PTAL

adiaaida:
@dotnet-bot test ci please

tests/scripts/run-xunit-perf.cmd
```diff
@@ -4,18 +4,27 @@
...
+
 :SETUP
 
+IF /I [%TEST_ARCHITECTURE%] == [x86jit32] (
+    set TEST_ARCH=x86
+) ELSE (
+    set TEST_ARCH=%TEST_ARCHITECTURE%
+)
+
+echo %TEST_ARCH% %TEST_ARCHITECTURE%
```
DrewScoggins: nit:  Probably want to remove this echo statement?

perf.groovy
```diff
@@ -74,8 +74,8 @@ def static getOSGroup(def os) {
...
 					batchFile("set __TestIntermediateDir=int&&build.cmd ${configuration} ${architecture}")
 					batchFile("tests\\runtest.cmd ${configuration} ${architecture} GenerateLayoutOnly")
-					batchFile("tests\\scripts\\run-xunit-perf.cmd -arch ${architecture} -configuration ${configuration} ${testEnv} -testBinLoc bin\\tests\\Windows_NT.${architecture}.${configuration}\\performance\\perflab\\Perflab -library -uploadToBenchview \"%WORKSPACE%\\Microsoft.Benchview.JSONFormat\\tools\" -runtype ${runType}")
-					batchFile("tests\\scripts\\run-xunit-perf.cmd -arch ${architecture} -configuration ${configuration} ${testEnv} -testBinLoc bin\\tests\\Windows_NT.${architecture}.${configuration}\\Jit\\Performance\\CodeQuality -uploadToBenchview \"%WORKSPACE%\\Microsoft.Benchview.JSONFormat\\tools\" -runtype ${runType}")
+					batchFile("tests\\scripts\\run-xunit-perf.cmd -arch ${arch} -configuration ${configuration} -testenv ${testEnv} -testBinLoc bin\\tests\\Windows_NT.${architecture}.${configuration}\\performance\\perflab\\Perflab -library -uploadToBenchview \"%WORKSPACE%\\Microsoft.Benchview.JSONFormat\\tools\" -runtype ${runType}")
```
DrewScoggins: Shouldn't you need to remove the -testenv that you are adding up on line 37 if you are passing it here?

adiaaida: I got confused. Thanks for pointing that out.

adiaaida:
@DrewScoggins Fixed issues. PTAL

DrewScoggins:
test Windows_NT_x86jit32 perf
test Windows_NT_x86 perf
test Windows_NT_x64 perf

DrewScoggins:
Looking at the results on Benchview it still does not look like we are writing out the architecture correctly.  Jose is doing some work to look at it now.  Do you have any submission.json files that got created out of your testing?  That might give us an idea of what is going on.

DrewScoggins:
My fault.  Those runs were never going to include the changes to the groovy file and that is where the changes that would fix this issue reside.  I ran the code locally and it made the correct JSON so LGTM.

PR #9703, opened 2017-02-21
Have Crossgen Failures exit in a dedicated exit routine

wtgodbe:
VSO builds were swallowing Crossgen failures, terminating those builds, and reporting them as successes - this should fix that (see https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/explorer?_a=summary&tab=artifacts&buildId=580018 for an example)

Addresses https://github.com/dotnet/coreclr/issues/8988#issuecomment-279452777


wtgodbe:
CC @gkhanna79

PR #9704, opened 2017-02-21
Correctly pass 12-byte GT_SIMD nodes on x86.

pgavlin:
During lowering, all TYP_SIMD12 GT_SIMD nodes are retyped to TYP_SIMD16.
This is correct except in the case that the GT_SIMD node is going to be
passed by-value on x86, in which case this retyping causes us to push 16
bytes rather than 12. This change recognizes a GT_SIMD node that has
been retyped by checking the value of gtSIMDSize during LowerArg and
uses TYP_SIMD12 for the putarg instead of TYP_SIMD16 (aside from
differences in detecting retyped nodes, this is the same transformation
we already do when passing TYP_SIMD12 lclVars that are being used as
TYP_SIMD16 on x86).

Fixes https://github.com/dotnet/corefx/issues/15913.

### DIFF
--- a/src/jit/lower.cpp
+++ b/src/jit/lower.cpp
@@ -1062,6 +1062,15 @@ void Lowering::LowerArg(GenTreeCall* call, GenTreePtr* ppArg)
             LclVarDsc* varDsc = &comp->lvaTable[varNum];
             type              = varDsc->lvType;
         }
+        else if (arg->OperGet() == GT_SIMD)
+        {
+            assert((arg->AsSIMD()->gtSIMDSize == 16) || (arg->AsSIMD()->gtSIMDSize == 12));
+
+            if (arg->AsSIMD()->gtSIMDSize == 12)
+            {
+                type = TYP_SIMD12;
+            }
+        }
     }
 #endif // defined(FEATURE_SIMD) && defined(_TARGET_X86_)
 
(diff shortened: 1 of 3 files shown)


pgavlin:
@BruceForstall @sivarv @dotnet/jit-contrib PTAL.

sivarv:
:shipit:

BruceForstall:
It would be nice to add a test case to dotnet/coreclr

pgavlin:
> It would be nice to add a test case to dotnet/coreclr

Done.

BruceForstall:
LGTM!

sivarv:
Please add this test under jit\SIMD tests rather than under regression.

pgavlin:
> Please add this test under jit\SIMD tests rather than under regression.

Why? It is a regression test. Unless there's a very compelling reason to move it I'd prefer to leave it where it is and avoid another trip through CI.

sivarv:
Whenever we found holes in tests we try to put them in the category which it belongs. If there is no specific category then we usually add it under regression.  Right now all SIMD specific tests are in one folder.  Going forward if we intend to run SIMD only tests on AVX2/SSE2/SSE3_4 targets it would come handy.  This would also come handy for doing asm diffs of SIMD only tests for SIMD specific optimizations.

pgavlin:
> Whenever we found holes in tests we try to put them in the category which it belongs. If there is no specific category then we usually add it under regression. 

We already have six other SIMD-related regression tests under `JIT/Regression/JitBlue`, so I'm going to leave the test I've added where it is for now. If we want to move it around later we should move the other tests with it.

Issue #4724, opened 2016-05-02
Tests should always be built the same way, regardless of product build flavor


BruceForstall:
This is related to #4124

There are some tests, e.g. JIT\Regression\CLR-x86-JIT\V1-M09.5-PDC\b12053\b12053\b12053.exe, that are being built different between the Debug and Release build. In the Debug build, they appear to be built with /debug+ and the JIT generates debuggable code. The IL is different, e.g., in Debug there are additional NOP instructions, and the result of conditional expressions gets stored to locals.

These differences are bad for JIT testing, because the code we are testing against is different between a Debug JIT build and a Release JIT build.

AndyAyersMS:
I wrote a parser to look at all the project files under the jit test tree. Roughly 1/3 of the tests will change behavior in Debug/Check vs Release.

For the cases that vary we need to decide whether to fix the tests as optimizable or not, or possibly duplicate them so we test both ways. 

Would also be g ood to cross check the result vs the "suffix conventions" (eg _r or _ro for optimizable code, _d or _do for debuggable code).

Not sure what a debug type of `Impl` means.

Some tests may also have source code conditional on `DEBUG` which can cause behavior to vary between build types. In some cases (eg the perf tests) this is intentional, but we should try and keep this to a minumum.

| DebugType | Condition | Count | WillTestVary | WillJitOptimize |
| --------- | --------- | ----- | ------------ | --------------- |
| Full      | No        | 1559  | No           | No              |
| PdbOnly   | No        | 1504  | No           | Yes             |
| None      | No        | 1006  | No           | Yes             |
| Impl      | No        |    2  | No           | ?               |
| (absent)  | No        | 1637  | Yes          | In Rel          |
| PdbOnly   | Yes       |  195  | Yes          | In Rel          |
| (blank)   | No        |  148  | Yes          | In Rel          |

BruceForstall:
As I understand it, the suffix conventions correspond to C# compiler switches:

|suffix|switch|
|------|------|
| _r | /debug- /optimize- |
| _ro | /debug- /optimize+ |
| _d | /debug+ /optimize- |
| _do | /debug+ /optimize+ |

These were carried over from the JIT self-host (pre-GitHub) testbed. Some tests were (presumably) only ever built "ret" (I believe _ro) because we only ever used the RET build of the QA tree before. So not all tests had all flavors. If we were to continue this decision (and not reconsider all the tests that vary), we should just force "_ro" for all tests without a configuration.

[...]

Issue #4124, opened 2016-04-06
Release IL tests should explicitly specify DebugType as PdbOnly


AndyAyersMS:
Otherwise in CHK/DBG builds, the tests won't enable jit optimizations.

There are about 285 release ilprojs that need fixing. There are another 1800 or so IL tests where fixes may be needed.

AndyAyersMS:
Am going to fix all the _il_rel variants and a few tests I've added.

AndyAyersMS:
Fixed via #9712.

PR #9712, opened 2017-02-22
Fix jit test projects to specify debug type and optimization flags

AndyAyersMS:
This partially addresses the issues raised in #4124 and #4724 by explicitly
and unconditionally setting DebugType and Optimize flags for various test
project files.

The projects that were fixed are those that did not specify any debug
flag settings at all.  The fix is to set DebugType to PdbOnly and to set
Optimization to true, unless the test has an _r or _rel suffix, in which
case Optimization is set false.

This impacts 1497 projects, skipping over 311 other projects flagged as needing
fixes. The projects that remain are the CodeGenBringUp set (for which we plan
to clone projects to test the various flag combinations), projects that
specify flags but do so in ways where the effective flag settings may vary
from DBG to RET, projects that specify the flags redundantly, or projects
where the flags are not consistent with the "test case name suffix" rules.

The edits were done by a tool using linq-to-xml and so some incidental
reformatting has happened.


AndyAyersMS:
@pgavlin PTAL
cc @dotnet/jit-contrib

AndyAyersMS:
Actually there are at least 374 left to fix, I missed out on cases where DebugType is set under either Debug or Release configs but not both.

AndyAyersMS:
Pushed an update to fix the remaining projects that aren't bring up tests.

AndyAyersMS:
Windows test failure [(log)](https://ci.dot.net/job/dotnet_coreclr/job/master/job/x64_release_windows_nt_pri1_prtest/3145/) seems unrelated. I didn't change any test project outside of the JIT dir.
```
Unhandled Exception: D:\j\workspace\x64_release_w---0575cb46\tests\tests.targets(47,5): error : CoreMangLib._cti_system_decimal_DecimalParse3_DecimalParse3_._cti_system_decimal_DecimalParse3_DecimalParse3_cmd [FAIL] [D:\j\workspace\x64_release_w---0575cb46\tests\runtest.proj]
10:16:01         Test Infrastructure Failure: The paging file is too small for this operation to complete
10:16:01 System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
```

CentOS test failure [(log)](https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_centos7.1_prtest/3068/) also seemingly unrelated:
```
threading.SleepEx.test2
MESSAGE:
Expected thread to sleep for 2000 ms (and not be interrupted). Average delta: 858993432 ms, acceptable delta: 150
```

Am going to rerun both...

jashook:
@AndyAyersMS the CentOS failure has happened as well on an Arm64 specific change. Seems like it is flaky see #9653 for the failure.

briansull:
LGTM, Thanks Andy

AndyAyersMS:
Once this goes in, I plan to add the project checker itself as a test case to catch any future project file issues.

If that seems too clever, stop me now.

pgavlin:
> Once this goes in, I plan to add the project checker itself as a test case to catch any future project file issues.
>
> If that seems too clever, stop me now.

As long as this doesn't unduly impact our test times (esp. in CI--I'm not sure what the I/O performance is like on our VMs), I think this seems fine.

AndyAyersMS:
Locally it runs in about 2 seconds. I suppose my file cache is all warmed up by now though.

AndyAyersMS:
@pgavlin any concerns with the second commit?

pgavlin:
Nope, LGTM

PR #9713, opened 2017-02-22
Add test coverage for ldtoken with members

MichalStrehovsky:
The test coverage for `ldtoken field X` and `ldtoken method X` in the
repo is very small - this test tries to improve things a little.

In particular, it tests that we get the right reflection objects, and
that we can do this from shared and unshared code.


MichalStrehovsky:
@fadimounir PTAL

PR #9714, opened 2017-02-22
Fix non-Windows amd64 register mask initialization

BruceForstall:
The arrays were being initialized with register numbers, not
register masks. The arrays are only used on non-Windows amd64
for homing circular incoming argument conflicts. Perhaps we
never saw this happen?


BruceForstall:
@briansull PTAL
cc @dotnet/jit-contrib

briansull:
LGTM

PR #9354, opened 2017-02-05
Adding perf tests for the single precision math functions in System.MathF.

tannergooding:
This resolves https://github.com/dotnet/coreclr/issues/7691

Issue #7690, opened 2016-10-18
Provide a set of unit tests over the new single-precision math APIs


tannergooding:
PR #5492 implemented dotnet/corefx#1151, by providing scalar single-precision floating-point support for many of the trigonometric, logarithmic, and other common mathematical functions.

However, there is still the additional work that needs to be done to provide a full set of unit tests over the new single-precision math APIs.

This work is partially blocked on https://github.com/dotnet/corefx/pull/12183 being merged so that a proper NuGet package is available with the appropriate contract assembly.

tannergooding:
I am self-assigning these for the time being and am going to work on getting these completed.

tannergooding:
This was resolved with https://github.com/dotnet/coreclr/pull/9717 where I ported the `CoreMangLib/cti/system/math` tests over to cover the new single-precision math APIs (`CoreMangLib/cti/system/mathf`)

PR #9717, opened 2017-02-22
Updating codegenxarch to support single-precision CORINFO_INTRINSIC_Sqrt

tannergooding:
This resolves an assert that occurs because `codegenxarch` asserted that `CORINFO_INTRINSIC_Sqrt` could only ever be `TYP_DOUBLE`:
```
Assert failure(PID 58004 [0x0000e294], Thread: 58004 [0xe294]): Assertion failed 'treeNode->TypeGet() == TYP_DOUBLE' in 'Functions.MathTests:SqrtSingleTest()' (IL size 99)
```

### DIFF
--- a/src/jit/codegenxarch.cpp
+++ b/src/jit/codegenxarch.cpp
@@ -7372,11 +7372,16 @@ void CodeGen::genIntrinsic(GenTreePtr treeNode)
     switch (treeNode->gtIntrinsic.gtIntrinsicId)
     {
         case CORINFO_INTRINSIC_Sqrt:
-            noway_assert(treeNode->TypeGet() == TYP_DOUBLE);
+        {
+            // Both operand and its result must be of the same floating point type.
+            GenTreePtr srcNode = treeNode->gtOp.gtOp1;
+            assert(varTypeIsFloating(srcNode));
+            assert(srcNode->TypeGet() == treeNode->TypeGet());
+
             genConsumeOperands(treeNode->AsOp());
-            getEmitter()->emitInsBinary(ins_FloatSqrt(treeNode->TypeGet()), emitTypeSize(treeNode), treeNode,
-                                        treeNode->gtOp.gtOp1);
+            getEmitter()->emitInsBinary(ins_FloatSqrt(treeNode->TypeGet()), emitTypeSize(treeNode), treeNode, srcNode);
             break;
+        }
 
         case CORINFO_INTRINSIC_Abs:
             genSSE2BitwiseOp(treeNode);
--- a/src/jit/instr.cpp
+++ b/src/jit/instr.cpp
@@ -3564,9 +3564,13 @@ instruction CodeGen::ins_FloatSqrt(var_types type)
     {
         ins = INS_sqrtsd;
     }
+    else if (type == TYP_FLOAT)
+    {
+        ins = INS_sqrtss;
+    }
     else
     {
-        // Right now sqrt of scalar single is not needed.
+        assert(!"ins_FloatSqrt: Unsupported type");
         unreached();
     }
 
--- a/src/jit/instrsxarch.h
+++ b/src/jit/instrsxarch.h
@@ -232,9 +232,10 @@ INST3( maxsd,  "maxsd",  0, IUM_WR, 0, 0, BAD_CODE, BAD_CODE, SSEDBL(0x5F))    /
 INST3( xorpd,  "xorpd",  0, IUM_WR, 0, 0, BAD_CODE, BAD_CODE, PCKDBL(0x57))    // XOR packed doubles
 INST3( andps,  "andps",  0, IUM_WR, 0, 0, BAD_CODE, BAD_CODE, PCKFLT(0x54))    // AND packed singles
 INST3( andpd,  "andpd",  0, IUM_WR, 0, 0, BAD_CODE, BAD_CODE, PCKDBL(0x54))    // AND packed doubles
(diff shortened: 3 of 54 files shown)


tannergooding:
FYI. @dotnet/jit-contrib

sivarv:
Looks good to me.
Do we already have a unit test case that exercise single precision sqrt or one needs to be added?
Also, single precision sqrt support getting added only to xarch (x86 and x64) or to Arm64/arm32 as well?  If yes, we also needs to update codegenarm.cpp and codegenarm64.cpp.

tannergooding:
> Do we already have a unit test case that exercise single precision sqrt or one needs to be added?

CoreFX has some unit tests today, but those never run under checked mode (to my knowledge).

The PR that originally hit this issue was https://github.com/dotnet/coreclr/pull/9354 (this commit is actually included as part of that PR, but the PR is currently blocked due the fact that it would break the Perf infrastructure right now -- the perf infrastructure doesn't currently support netcoreapp projects).

That all being said, the work item tracking the creation of unit tests for the single-precision math functions is here: https://github.com/dotnet/coreclr/issues/7690.

> Also, single precision sqrt support getting added only to xarch (x86 and x64) or to Arm64/arm32 as well? If yes, we also needs to update codegenarm.cpp and codegenarm64.cpp.

Arm hasn't been updated yet: `vsqrt` needs to be updated for arm32 and `fsqrt` needs to be updated for arm64, I'll see about fixing those later this evening.

BruceForstall:
We should add regression tests to the dotnet/coreclr repo

tannergooding:
@BruceForstall, I've ported the CoreMangLib/cti/system/math tests to cover the single precision-implementations (projects were copied and only updated where necessary to get them working as expected).

@sivarv, I will take a look at covering the arm/arm64 intrinsics tomorrow if I get a chance.

tannergooding:
@sivarv, could you direct me to the best person for asking a question regarding the ARM instructions.

Basically, the instruction encodings currently specified are explicitly for F32, not for F64 as I expected. I am wondering if this is a known issue and if there is any documentation explaining this...

For ARM32, `vsqrt` is currently defined as follows:
```C++
INST1(vsqrt,    "vsqrt",       1, 0,   IF_T2_VFP2,  0xEEB10AC0)               // opc2 = '001',  zp = 01
```
This has a bit encoding of `1110 11101 0 11 0001 0000 101 0 1100 0000` which is the `VSQRT.F32` instruction. I would have expected it to be `1110 11101 0 11 0001 0000 101 1 1100 0000` for the `VSQRT.F64` instruction (and I would have expected that I needed to add the former).

Similarly for ARM64, `fsqrt` is currently defined as:
```C++
INST2(fsqrt,   "fsqrt",  0, 0, IF_EN2J,   0x2EA1F800,  0x1E21C000)
```
This first encoding is for Vecotrs has a bit encoding of `0 0 1 01110 1 0 10000 11111 10 00000 00000` which is for `FSQRT.2S` (`0 1 1 01110 1 0 10000 11111 10 00000 00000` would be for `FSQRT.4S`). I would have expected it to be `0 1 1 01110 1 1 10000 11111 10 00000 00000` for `FSQRT.2D` (and I would have expected that I needed to add the former two).

sivarv:
@briansull and @BruceForstall - would be the best person to direct your question.

BruceForstall:
It's been too long since I've looked at the ARM32 encoder to know the answer here. @briansull is more likely to know.

@parjong -- maybe someone on your team has thought about the ARM32 encodings more recently and can answer?

tannergooding:
I also don't see anywhere where the instruction encodings are modified to have the correct bit representation, but maybe I am missing something...

briansull:
For the vadd/vsub etc. instructions the bit the controls the size (float vs double) is set here:

    ```
    case IF_T2_VFP3:
            // these are the binary operators
            // d = n - m
            sz   = emitGetInstrDescSize(id);
            code = emitInsCode(ins, fmt);
            code |= insEncodeRegT2_VectorN(id->idReg2(), size, true);
            code |= insEncodeRegT2_VectorM(id->idReg3(), size, true);
            code |= insEncodeRegT2_VectorD(id->idReg1(), size, true);
            **if (size == EA_8BYTE)
                code |= 1 << 8;**
            dst += emitOutput_Thumb2Instr(dst, code);
            break;

```

tannergooding:
@briansull, thanks!

It looks like ARM32 is already handled:
* https://github.com/dotnet/coreclr/blob/5c2459a1b4c31ca467d9a50fe76e37a4412bc872/src/jit/emitarm.cpp#L6052

As are Vector and Scalar instructions for ARM64:
* https://github.com/dotnet/coreclr/blob/5c2459a1b4c31ca467d9a50fe76e37a4412bc872/src/jit/emitarm64.cpp#L9037
* https://github.com/dotnet/coreclr/blob/5c2459a1b4c31ca467d9a50fe76e37a4412bc872/src/jit/emitarm64.cpp#L9118

So this PR should need no further adjustments.

tannergooding:
@sivarv or @BruceForstall. Could you give a final review pass? ARM/ARM64 are covered (as per the above) and a set of tests covering the System.MathF APIs have been added.

parjong:
@BruceForstall I'm not sure about encoder as we currently focus on flowgrap and morph (to bring up ARM/RyuJIT). 

@hqueue @hseok-oh Could you let us know if you have any idea?

sivarv:
Looks good.

hqueue:
@tannergooding I also confirm `emitarm.cpp` already handle this case correctly.

PR #9724, opened 2017-02-22
Reduce code in StubHelpers.SafeHandleAddRef

stephentoub:
If DangerousAddRef returns rather than throws, its argument will be true, so there's no need to branch on it.

Also changed its throws to use ThrowHelper.

### DIFF
--- a/src/mscorlib/src/System/StubHelpers.cs
+++ b/src/mscorlib/src/System/StubHelpers.cs
@@ -1618,13 +1618,11 @@ namespace  System.StubHelpers {
         {
             if (pHandle == null)
             {
-                throw new ArgumentNullException(nameof(pHandle), Environment.GetResourceString("ArgumentNull_SafeHandle"));
+                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.pHandle, ExceptionResource.ArgumentNull_SafeHandle);
             }
-            Contract.EndContractBlock();
 
             pHandle.DangerousAddRef(ref success);
-
-            return (success ? pHandle.DangerousGetHandle() : IntPtr.Zero);
+            return pHandle.DangerousGetHandle();
         }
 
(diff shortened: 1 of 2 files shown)


src/mscorlib/src/System/StubHelpers.cs
```diff
@@ -1618,23 +1618,20 @@ static internal IntPtr SafeHandleAddRef(SafeHandle pHandle, ref bool success)
         {
             if (pHandle == null)
             {
-                throw new ArgumentNullException(nameof(pHandle), Environment.GetResourceString("ArgumentNull_SafeHandle"));
+                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.pHandle, ExceptionResource.ArgumentNull_SafeHandle);
             }
-            Contract.EndContractBlock();
 
             pHandle.DangerousAddRef(ref success);
```
danmosemsft: Given your observation, why does this function even have a success argument? Legacy thing?

danmosemsft: Oh, I see it's for convenience in a finally, as in the other  usage in ThreadPool.

stephentoub: > Given your observation, why does this function even have a success argument? Legacy thing?
> Oh, I see it's for convenience in a finally, as in the other usage in ThreadPool.

Not convenience, but reliability, in the face of asynchronous exceptions like thread aborts.  Taking the ref bool as an arg allows the implementation to ensure that the bool can be set to true iff the ref count was increased.  If it instead returned a bool (or didn't provide a bool at all), you'd be unable to have that guarantee, since, for example, a thread abort could occur after the ref count was increased but before the implementation was able to return the Boolean value.  Thread aborts aren't a factor for coreclr, though.

PR #9725, opened 2017-02-22
Android: workaround absence of getcontext/setcontext

qmfrederik:
#9650 added dependencies on getcontext and setcontext, which are not available on Android.

This PR implements getcontext and setcontext using RtlCaptureContext and RtlRestoreContext.

### DIFF
--- a/src/pal/src/exception/signal.cpp
+++ b/src/pal/src/exception/signal.cpp
@@ -47,6 +47,14 @@ SET_DEFAULT_DEBUG_CHANNEL(EXCEPT); // some headers have code with asserts, so do
 
 #include "pal/context.h"
 
+#ifdef __ANDROID__
+int setcontext(const ucontext_t *ucp)
+{
+    // Not implemented on Android
+    return -1;
+}
+#endif
+
 using namespace CorUnix;
 
 #ifdef SIGRTMIN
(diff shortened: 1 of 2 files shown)


qmfrederik:
@janvorli FYI. I guess there's a better way to do this (either by not using the changes from #9650 on Android or implementing setcontext), but this at least gets CoreCLR building on Android.

janvorli:
@qmfrederik I would prefer implementing the getcontext and setcontext for Android using RtlCaptureContext and RtlRestoreContext. 
```C++
int getcontext(ucontext_t *ucp)
{
    CONTEXT context;
    RtlCaptureContext(&context);
    CONTEXTToNativeContext(&context, ucp);

    return 0;
}

int setcontext(const ucontext_t *ucp)
{
    CONTEXT context;
    ULONG contextFlags = CONTEXT_CONTROL | CONTEXT_INTEGER | CONTEXT_FLOATING_POINT;

#if defined(_AMD64_)
    contextFlags |= CONTEXT_XSTATE;
#endif

    CONTEXTFromNativeContext(ucp, &context, contextFlags);
    RtlRestoreContext(&context, NULL);

    return 0;
}
```

janvorli:
@qmfrederik just FYI, I am going to switch to using this on all platforms. Alpine Linux is yet another platform that doesn't support getcontext / setcontext and so I'd like to have it unified.

qmfrederik:
@janvorli Makes sense, thanks for the heads up!

PR #9727, opened 2017-02-22
Speed up JIT builds

BruceForstall:
We don't need to build the static lib and crossgen static lib
versions of the JIT unless FEATURE_MERGE_JIT_AND_ENGINE is set.
And, since we currently don't set it, this means doing two
fewer JIT builds.


BruceForstall:
@pgavlin PTAL
cc @dotnet/jit-contrib

pgavlin:
LGTM. Thanks for doing this!

BruceForstall:

Issue #8683, opened 2016-12-20
Two tests failing locally for windows x64 checked


AndyAyersMS:
I am seeing two tests fail with windows x64 checked builds. This is @8395eb5 (which is my change, but seemingly unrelated). Not sure why we don't see these failing in the lab.
```
JIT\Directed\nullabletypes\castclassvaluetype_r
JIT\Directed\nullabletypes\castclassvaluetype_d
```
Typical failure is:
```
Assert failure(PID 43884 [0x0000ab6c], Thread: 59336 [0xe7c8]): (uImageBase == uImageBaseFromOS) && (pFunctionEntry == pFunctionEntryFromOS)

CORECLR! Thread::VirtualUnwindCallFrame + 0x49A (0x00007ffa`71bf4306)
CORECLR! EECodeManager::EnsureCallerContextIsValid + 0x1F6 (0x00007ffa`71eaa2a2)
CORECLR! StackFrameIterator::CheckForSkippedFrames + 0x70 (0x00007ffa`71beea00)
CORECLR! StackFrameIterator::ProcessCurrentFrame + 0x160 (0x00007ffa`71bf3470)
CORECLR! StackFrameIterator::NextRaw + 0xE02 (0x00007ffa`71bf2dda)
CORECLR! StackFrameIterator::Next + 0xC1 (0x00007ffa`71bf1f19)
CORECLR! Thread::StackWalkFramesEx + 0x168 (0x00007ffa`71bf3ac8)
CORECLR! ExceptionTracker::FindNonvolatileRegisterPointers + 0x15B (0x00007ffa`71c0a61f)
CORECLR! FixNonvolatileRegisters + 0x155 (0x00007ffa`71c0c9d9)
CORECLR! ProcessCLRException + 0x8DE (0x00007ffa`71c18b2e)
    File: c:\repos\coreclr2\src\vm\stackwalk.cpp Line: 631
    Image: C:\repos\coreclr2\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\CoreRun.exe
```

AndyAyersMS:
This does not repro on a different local machine. The failing machine is newer so tried disabling AVX and SSE3_4 codegen but it did not change things any. Am going to do a clean rebuild on the "bad" machine next.

AndyAyersMS:
Does not repro after a clean rebuild, so will close this...

jkotas:
I am hitting with MusicStore on debug builds. 

[...]

PR #9728, opened 2017-02-23
Ignore binder context for corelib when performing binding cache lookup

rahku:
corelib is always loaded in default context. When performing a lookup for corelib in binding cache no binderid is used to modify the key. However in function AssemblySpecBindingCache::GetAssemblyBindingEntryForAssemblySpec we were using binderid to modify the key even for corelib. This is different from implementation in AssemblySpecBindingCache::StoreAssembly. This was causing the POST condition at https://github.com/dotnet/coreclr/blob/master/src/vm/assemblyspec.cpp#L1509 to fail.

### DIFF
--- a/src/vm/assemblyspec.cpp
+++ b/src/vm/assemblyspec.cpp
@@ -1276,7 +1276,14 @@ AssemblySpecBindingCache::AssemblyBinding* AssemblySpecBindingCache::GetAssembly
     
     // Check if the AssemblySpec already has specified its binding context. This will be set for assemblies that are
     // attempted to be explicitly bound using AssemblyLoadContext LoadFrom* methods.
-    pBinderContextForLookup = pSpec->GetBindingContext();
+    if(!pSpec->IsAssemblySpecForMscorlib())
+        pBinderContextForLookup = pSpec->GetBindingContext();
+    else
+    {
+        // For System.Private.Corelib Binding context is either not set or if set then it should be TPA
+        _ASSERTE(pSpec->GetBindingContext() == NULL || pSpec->GetBindingContext() == pSpecDomain->GetFusionContext());
+    }
+
     if (pBinderContextForLookup != NULL)
     {
         // We are working with the actual binding context in which the assembly was expected to be loaded.


rahku:
Fixes issue #8683

rahku:
PTAL @gkhanna79

gkhanna79:
> AssemblySpecBindingCache::GetAssemblyBindingEntryForAssemblySpec we were using binderid to modify the key even for corelib. 

This should not happen because https://github.com/dotnet/coreclr/blob/master/src/vm/assemblyspec.cpp#L1291 will result in keeping `pBinderContextForLookup` to be `NULL`. Is that not NULL even before this line? If so, then you may want to understand why that is the case first.

rahku:
It is set here https://github.com/dotnet/coreclr/blob/master/src/vm/multicorejitplayer.cpp#L936. This sets bindercontext irrespective of whether it is corelib.

src/vm/assemblyspec.cpp
```diff
@@ -1276,7 +1276,14 @@ AssemblySpecBindingCache::AssemblyBinding* AssemblySpecBindingCache::GetAssembly
     
     // Check if the AssemblySpec already has specified its binding context. This will be set for assemblies that are
     // attempted to be explicitly bound using AssemblyLoadContext LoadFrom* methods.
-    pBinderContextForLookup = pSpec->GetBindingContext();
+    if(!pSpec->IsAssemblySpecForMscorlib())
+        pBinderContextForLookup = pSpec->GetBindingContext();
+    else
+    {
+        // For System.Private.Corelib Binding context is either not set or if set then it should be TPA or Custom ALC
```
gkhanna79: It cannot be Custom ALC and only has to be TPA.

gkhanna79:
Do we know why that is set for CoreLib as well? What if it is fixed to not set for CoreLib? Are there other places that need to have a fix like you are making here?

rahku:
>>Do we know why that is set for CoreLib as well? 
Why shouldn't be set for Corelib. I believe it is internal details for binder that we do not use it to lookup in cache.

>>What if it is fixed to not set for CoreLib? 
I believe that would work as well. But would be a good design distribute inner workings of binder to other places.

>>Are there other places that need to have a fix like you are making here?
I did not hit any other issue. Most likely no other place should require change.

rahku:
>>Do we know why that is set for CoreLib as well?

Why shouldn't be set for Corelib. I believe it is internal details for binder that we do not use it to lookup in cache.

>>What if it is fixed to not set for CoreLib?

I believe that would work as well. But would be a good design distribute inner workings of binder to other places.

>>Are there other places that need to have a fix like you are making here?

I did not hit any other issue. Most likely no other place should require change.

src/vm/assemblyspec.cpp
rahku: It will be default ALC (not just TPA) which I believe can be overridden using AssemblyLoadContext.InitializeDefaultContext

gkhanna79: This is not a supported scenario and that API no longer exists - see https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Runtime/Loader/AssemblyLoadContext.cs.

rahku:

rahku:

rahku:
@gkhanna79 I have reviewed this should not affect other places. Can you approve the changes

Issue #9680, opened 2017-02-20
Segmentation Fault when using ZapImage::PrintStats


sjsinju:
I used ZapImage :: PrintStats in zapimage.cpp like following codes to check the configuration of the Native Image created by crossgen, but a segmentation fault occurred.

```
~~
    HANDLE hFile = GenerateFile(wszOutputFileName, pNativeImageSig);
    PrintStats(wszOutputFileName);
~~
}
```
```
void ZapImage::PrintStats(LPCWSTR wszOutputFileName)
{
    m_stats->m_gcInfoSize = m_pHotTouchedGCSection->GetSize() + m_pHotGCSection->GetSize() + m_pGCSection->GetSize();
#if defined(WIN64EXCEPTIONS)
    m_stats->m_unwindInfoSize = m_pUnwindDataSection->GetSize() +
        m_pHotRuntimeFunctionSection->GetSize() + m_pRuntimeFunctionSection->GetSize() + m_pColdRuntimeFunctionSection->GetSize();
#endif // defined(WIN64EXCEPTIONS)
~~
}
```

Looking at the code, I noticed that the m_stats variable used in PrintStats is the pointer variable of ZapperStats, which is being used without being created by 'new' or anything.

So I just added 'm_stats = new ZapperStats();' for creating instance on ZapImage::InitializeSections() and ZapImage::InitializeSectionsForReadyToRun like below.

```
void ZapImage::InitializeSections()
{
    m_stats = new ZapperStats();
    AllocateVirtualSections();
    ...
}
```
```
void ZapImage::InitializeSectionsForReadyToRun()
{
    m_stats = new ZapperStats();
    AllocateVirtualSections();
    ...
}
```

And then I was able to check the results for PrintStats normally.

sjsinju:
For using with ReadyToRun feature, additional null check is needed.
```
void ZapImage::PrintStats(LPCWSTR wszOutputFileName)
{
    if (m_stats == NULL )
        m_stats = new ZapperStats();

    m_stats->m_gcInfoSize = m_pHotTouchedGCSection->GetSize() + m_pHotGCSection->GetSize() + m_pGCSection->GetSize();
#if defined(WIN64EXCEPTIONS)
    m_stats->m_unwindInfoSize = m_pUnwindDataSection->GetSize() + 
        m_pHotRuntimeFunctionSection->GetSize() + m_pRuntimeFunctionSection->GetSize() + m_pColdRuntimeFunctionSection->GetSize();
[...]

PR #9731, opened 2017-02-23
Fix segmentation fault when using ZapImage::PrintStats()

sjsinju:
Currently, ZapImage::PrintStats is not used anywhere. But I want to get detailed informations about Native Image(ni.dll) components.
So I just used ZapImage::PrintStats when called ZapImage::SaveImage. After that I met the segmentation fault likes #9680.

I found the reason for this issue is accessing the null pointer to get sizes.

### DIFF
--- a/src/zap/zapheaders.cpp
+++ b/src/zap/zapheaders.cpp
@@ -171,11 +171,13 @@ void ZapImage::SaveCodeManagerEntry()
 
     if (m_stats)
     {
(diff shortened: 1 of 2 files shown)


jkotas:
There is binary `core` file in the delta. Could you please remove it?

sjsinju:
Sorry. I fixed that.

jkotas:
Thanks!

Issue #5061, opened 2016-05-18
Use GetSystemTimePreciseAsFileTime for DateTime.UtcNow


GSPP:
The [`GetSystemTimePreciseAsFileTime`](https://msdn.microsoft.com/library/windows/desktop/hh706895.aspx?f=255&MSPPError=-2147217396) API provides more precise timing information on Windows 8+. .NET should use this API if available to make additional precision available to .NET code.

Often, the precision of `UtcNow` is ~15ms. Sometimes the relative order of log items cannot be determined if all log items in close temporal proximity have the same timestamp. Logging has become more important in the cloud. It's very helpful to be able to order events relatively to one another. Also, this sometimes trips people up when trying to benchmark.

This seems like a fairly non-expensive change impacting many applications positively one way or another.

Benchmark results:

```c#
    static void Main(string[] args)
    {
        Process.GetCurrentProcess().PriorityClass = ProcessPriorityClass.High;

        {
            var sw = Stopwatch.StartNew();
            long result = 0;
            for (int i = 0; i < 100000000; i++)
            {
                FILE_TIME ft;
                GetSystemTimeAsFileTime(out ft);
                result += ft.ftTimeLow;
            }
            sw.Stop();
            GC.KeepAlive(result);
            Console.WriteLine(sw.Elapsed.TotalMilliseconds);
        }

        {
            var sw = Stopwatch.StartNew();
            long result = 0;
            for (int i = 0; i < 100000000; i++)
            {
                FILE_TIME ft;
                GetSystemTimePreciseAsFileTime(out ft);
                result += ft.ftTimeLow;
            }
            sw.Stop();
            GC.KeepAlive(result);
            Console.WriteLine(sw.Elapsed.TotalMilliseconds);
        }

        Console.ReadKey();
    }

    [StructLayout(LayoutKind.Sequential)]
    struct FILE_TIME
    {
        public int ftTimeLow;
        public int ftTimeHigh;
    }

    [SuppressUnmanagedCodeSecurity, DllImport("kernel32.dll")]
    static extern void GetSystemTimeAsFileTime(out FILE_TIME lpSystemTimeAsFileTime);

    [SuppressUnmanagedCodeSecurity, DllImport("kernel32.dll")]
    static extern void GetSystemTimePreciseAsFileTime(out FILE_TIME lpSystemTimeAsFileTime);
```

GetSystemTimeAsFileTime: 1341,3188 ms, 70 million calls per second
[...]

PR #9736, opened 2017-02-23
Use GetSystemTimePreciseAsFileTime if available in DateTime.UtcNow

stephentoub:
Makes DateTime.UtcNow much more precise, at the expense of being ~2x slower than it was previously.

For example, prior to the change, this code:
```C#
var list = new List<long>();
for (int i = 0; i < 20; i++) list.Add(DateTime.UtcNow.Ticks);
foreach (long tick in list) Console.WriteLine(tick);
```
produced results like:
```
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
636234535450569888
```
After the change, it produces results like:
```
636234537051614021
636234537051614039
636234537051614039
636234537051614039
636234537051614039
636234537051614042
636234537051614045
636234537051614045
636234537051614045
636234537051614045
636234537051614045
636234537051614045
636234537051614180
636234537051614180
636234537051614183
636234537051614183
636234537051614183
636234537051614183
636234537051614183
636234537051614183
```

Fixes https://github.com/dotnet/coreclr/issues/5061
Fixes https://github.com/dotnet/corefx/issues/16405
Closes https://github.com/dotnet/corefx/issues/15739
cc: @vancem, @jkotas

### DIFF
--- a/src/classlibnative/bcltype/system.cpp
+++ b/src/classlibnative/bcltype/system.cpp
@@ -41,14 +41,37 @@ typedef BOOL (*pfnGetPhoneVersion)(LPOSVERSIONINFO lpVersionInformation);
 pfnGetPhoneVersion g_pfnGetPhoneVersion = NULL;
 #endif
 
+typedef void(WINAPI *pfnGetSystemTimeAsFileTime)(LPFILETIME lpSystemTimeAsFileTime);
+extern pfnGetSystemTimeAsFileTime g_pfnGetSystemTimeAsFileTime;
+
+void WINAPI InitializeGetSystemTimeAsFileTime(LPFILETIME lpSystemTimeAsFileTime)
+{
+    pfnGetSystemTimeAsFileTime func = NULL;
+
+#ifndef FEATURE_PAL
+    HMODULE hKernel32 = WszLoadLibrary(W("kernel32.dll"));
+    if (hKernel32 != NULL)
+    {
+        func = (pfnGetSystemTimeAsFileTime)GetProcAddress(hKernel32, "GetSystemTimePreciseAsFileTime");
+    }
+    if (func == NULL)
+#endif
+    {
+        func = &::GetSystemTimeAsFileTime;
+    }
+    
+    g_pfnGetSystemTimeAsFileTime = func;
+    func(lpSystemTimeAsFileTime);
+}
+
+pfnGetSystemTimeAsFileTime g_pfnGetSystemTimeAsFileTime = &InitializeGetSystemTimeAsFileTime;
 
 FCIMPL0(INT64, SystemNative::__GetSystemTimeAsFileTime)
 {
     FCALL_CONTRACT;
 
     INT64 timestamp;
-
-    ::GetSystemTimeAsFileTime((FILETIME*)&timestamp);
+    g_pfnGetSystemTimeAsFileTime((FILETIME*)&timestamp);
 
 #if BIGENDIAN
     timestamp = (INT64)(((UINT64)timestamp >> 32) | ((UINT64)timestamp << 32));


benaadams:
Can it be marked with port to desktop potential? Don't know what the procedure is on it

src/classlibnative/bcltype/system.cpp
```diff
@@ -41,14 +41,35 @@ typedef BOOL (*pfnGetPhoneVersion)(LPOSVERSIONINFO lpVersionInformation);
...
+typedef void(WINAPI *pfnGetSystemTimeAsFileTime)(LPFILETIME lpSystemTimeAsFileTime);
+pfnGetSystemTimeAsFileTime g_pfnGetSystemTimeAsFileTime = NULL;
+Volatile<bool> g_fGetSystemTimeAsFileTimeInitialized = false;
 
 FCIMPL0(INT64, SystemNative::__GetSystemTimeAsFileTime)
 {
     FCALL_CONTRACT;
 
-    INT64 timestamp;
+	if (!g_fGetSystemTimeAsFileTimeInitialized)
```
jkotas: It will save a few cycles to get rid volatile check out and move the initialization into separate method. You can do something like this:

```
extern pfnGetSystemTimeAsFileTime g_pfnGetSystemTimeAsFileTime;

void WINAPI InitializeGetSystemTimeAsFileTime(LPFILETIME lpSystemTimeAsFileTime)
{
    HMODULE hKernel32 = WszLoadLibrary(W("kernel32.dll")); 
    if (hKernel32 != NULL)
    {
        g_pfnGetSystemTimeAsFileTime = (pfnGetSystemTimeAsFileTime)GetProcAddress(hKernel32, "GetSystemTimePreciseAsFileTime");
    }
    else
    {
        g_pfnGetSystemTimeAsFileTime = &::GetSystemTimeAsFileTime;
    }

    g_pfnGetSystemTimeAsFileTime(lpSystemTimeAsFileTime);
}

pfnGetSystemTimeAsFileTime g_pfnGetSystemTimeAsFileTime = InitializeGetSystemTimeAsFileTime(LPFILETIME lpSystemTimeAsFileTime);
```

The hot path can then be just `g_pfnGetSystemTimeAsFileTime((FILETIME*)&timestamp);`

src/classlibnative/bcltype/system.cpp
```diff
@@ -41,14 +41,35 @@ typedef BOOL (*pfnGetPhoneVersion)(LPOSVERSIONINFO lpVersionInformation);
 pfnGetPhoneVersion g_pfnGetPhoneVersion = NULL;
 #endif
 
+typedef void(WINAPI *pfnGetSystemTimeAsFileTime)(LPFILETIME lpSystemTimeAsFileTime);
+pfnGetSystemTimeAsFileTime g_pfnGetSystemTimeAsFileTime = NULL;
+Volatile<bool> g_fGetSystemTimeAsFileTimeInitialized = false;
 
 FCIMPL0(INT64, SystemNative::__GetSystemTimeAsFileTime)
```
jkotas: Another thing that should save a few cycles is to change the signature of this method to: `void GetSystemTimeAsFileTime(INT64* pTimestamp)`.

It will make the signature of the underlying windows API match exactly the signature of the FCall. It means that the compiler can compile the FCall to just `jmp [g_pfnGetSystemTimeAsFileTime]` vs. method with a frame that it is today.

src/classlibnative/bcltype/system.cpp
```diff
@@ -41,14 +41,35 @@ typedef BOOL (*pfnGetPhoneVersion)(LPOSVERSIONINFO lpVersionInformation);
...
 {
     FCALL_CONTRACT;
 
-    INT64 timestamp;
+	if (!g_fGetSystemTimeAsFileTimeInitialized)
+	{
+		// Try to use GetSystemTimePreciseAsFileTime if it's available (Win8+).
+		// Otherwise fall back to GetSystemTimeAsFileTime.
+		pfnGetSystemTimeAsFileTime func = NULL;
+		HMODULE hKernel32 = WszLoadLibrary(W("kernel32.dll"));
```
jkotas: This should be under `#ifndef FEATURE_PAL` so that we are not trying to load kernel32.dll on Unix.

jkotas:
> Can it be marked with port to desktop potential?

Marked https://github.com/dotnet/corefx/issues/15739 as netfx-port-consider

src/classlibnative/bcltype/system.cpp
```diff
@@ -41,14 +41,35 @@ typedef BOOL (*pfnGetPhoneVersion)(LPOSVERSIONINFO lpVersionInformation);
...
+			func = &::GetSystemTimeAsFileTime;
+		}
+
+		g_pfnGetSystemTimeAsFileTime = func;
+		g_fGetSystemTimeAsFileTimeInitialized = true;
+	}
 
-    ::GetSystemTimeAsFileTime((FILETIME*)&timestamp);
+    INT64 timestamp;
+	g_pfnGetSystemTimeAsFileTime((FILETIME*)&timestamp);
```
stephentoub: Fixed

src/classlibnative/bcltype/system.cpp
stephentoub: Changed to this general approach

src/classlibnative/bcltype/system.cpp
stephentoub: Meaning change the API signature up through to how its called from managed?  Makes sense, but I'd like to leave that for a separate change, and leave this one just as the precision change.

src/classlibnative/bcltype/system.cpp
stephentoub: @janvorli noted this to me offline as well. Fixed.

src/classlibnative/bcltype/system.cpp
jkotas: > Meaning change the API signature up through to how its called from managed? 

Right.

>  this one just as the precision change.

ok

vancem:
LGTM,     I would like to document the precision of DateTime.NowUtc to the degree we can.   We should update the documentation comment of DateTime.NowUtc to indicate the precision you can expect (which I think is ~1us+ on both Win8+ and Linux but may be as little as 16msec on older OSes.

stephentoub:

stephentoub:
@dotnet-bot test Ubuntu x64 Checked Build and Test please (https://github.com/dotnet/coreclr/issues/9750)

mj1856:
+1 Thanks!

mj1856:
What about non-windows implementations?  `clock_gettime(CLOCK_REALTIME)`. Seems appropriate, IMHO.

vancem:
In Issue #5061 @mj1856  collected some useful data

stephentoub:
I'll take a look at it later today and submit a PR if needed.

fuerstrainier:
Not only is it slower that way (as stephentoub pointed out), but I also challenge the "more precise" statement.
Our Windows clock gets updated by a reliable UTC clock, which we can use to properly report on (to the authorities) and monitor, but when using GetSystemTimePreciseAsFileTime we start mixing this UTC clock with the local tick rate, which is anything but reliable. So from our perspective GetSystemTimePreciseAsFileTime yields a RANDOM timestamp between to Windows clock updates.

stephentoub:
@fuerstrainier, you're saying you have that problem with GetSystemTimePreciseAsFileTime but not with GetSystemTimeAsFileTime?

fuerstrainier:
@stephentoub yes that is correct. The windows clock (which is read by GetSystemTimeAsFileTime) is updated by a proper external clock and the tick counter (additionally) used by GetSystemTimePreciseAsFileTime is updated by the local machines tick rate, which will tick faster or slower than our reference clock. If it's faster it could even overtake the windows clock updates...

stephentoub:
Thanks, @fuerstrainier.

mj1856:
@fuerstrainier - Is there a basis for your claim that it is ticking faster or slower than your reference clock?  The considerable documentation on this subject would suggest otherwise.

The [GetSystemTimePreciseAsFileTime](https://msdn.microsoft.com/library/windows/desktop/Hh706895.aspx) docs state:

> The GetSystemTimePreciseAsFileTime function retrieves the current system date and time with the highest possible level of precision (<1us). The retrieved information is in Coordinated Universal Time (UTC) format.

The [Windows Server 2016 Accurate Time](https://docs.microsoft.com/en-us/windows-server/identity/ad-ds/get-started/windows-time-service/accurate-time#windows-application-requiring-accurate-time) docs state:
> Programs which require the greatest accuracy with regards to UTC, and not the passage of time, should use the GetSystemTimePreciseAsFileTime API. This assures your application gets System Time, which is conditioned by the Windows Time service.

Perhaps you were thinking of `QueryPerformanceCounter` such as is used in `System.Diagnostics.Stopwatch`?  The definitive doc on that is [Acquiring high-resolution time stamps](https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408.aspx), which also says to use GetSystemTimePreciseAsFileTime if you want to be bound to reference clocks, which for `DateTime.UtcNow` we do:

> QPC is independent of and isn't synchronized to any external time reference. To retrieve time stamps that can be synchronized to an external time reference, such as, Coordinated Universal Time (UTC) for use in high-resolution time-of-day measurements, use GetSystemTimePreciseAsFileTime.

So, if you feel otherwise, I'd really like to understand why. Thanks.

fuerstrainier:
@mj1856 Thanks for your reply. I have to amend my original statement a bit, since it basically depends on the implementation of the timeservice you are using. If it follows a "stepping" approach (using the SetSystemTime operation), you can get into troubles due to different clock speeds, but if it just "skews" the windows clock (using the SetSystemTimeAdjustment), we have no issue with time continuity. Since I learned today that using the stepping approach is not a good idea anyway, it's not a deal for us anymore. Cheers.

Issue #9733, opened 2017-02-23
JIT.Performance.CodeQuality.Serialization.Serialize.Serialize failing


janvorli:
The JIT.Performance.CodeQuality.Serialization.Deserialize.Deserialize and JIT.Performance.CodeQuality.Serialization.Serialize.Serialize tests are failing with 
`Unhandled Exception: System.IO.FileLoadException: Could not load file or assembly 'xunit.performance.core, Version=1.0.0.40, Culture=neutral, PublicKeyToken=67066efe964d3b03'. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040 (FUSION_E_REF_DEF_MISMATCH))`

janvorli:
They fail on Ubuntu starting with build https://ci.dot.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_tst_prtest/2806/

janvorli:
CC: @gkhanna79

gkhanna79:
@pgavlin Can you PTAL as it appears to have commenced from your PR?

AndyAyersMS:
This is an issue caused by CoreFX update. See #9733.

gkhanna79:
@AndyAyersMS the link you shared is for this issue itself. Which CoreFX update issue are you referring to?

AndyAyersMS:
Ah, sorry, there are notes on this over in #9729.

brianrob:
CI should be unblocked now that #9741 is merged.  I'll follow-up on re-enabling these tests.

wtgodbe:
https://github.com/dotnet/coreclr/pull/9749 fixes this issue

PR #9737, opened 2017-02-23
update xunit-performance to build0047

AndyAyersMS:
Seeing if this will fix recent issues with the serialization tests (#9733)


AndyAyersMS:
Doesn't work for me locally, so not optimistic. Failure was:
```
C:\repos\coreclr2\Tools\VersionTools.targets(47,5): error : Dependency verification errors detected. To automatically f
ix based on dependency rules, run the msbuild target 'UpdateDependencies' [C:\repos\coreclr2\tests\build.proj]
C:\repos\coreclr2\Tools\VersionTools.targets(47,5): error : Dependencies invalid: In 'C:\repos\coreclr2\tests\src\Commo
n\test_dependencies\project.json', 'Microsoft.DotNet.xunit.performance 1.0.0-alpha-build0047' must be '1.0.0-alpha-buil
d0040' (Microsoft.DotNet.xunit.performance)
```
Tried running the UpdateDependencies build target as directed but nothing happened.

Am going to retry after a `git clean -dfx`

AndyAyersMS:
CI hitting the same issue so evidently there's something else that must be changed.

jkotas:
@AndyAyersMS You also need to update it in https://github.com/dotnet/coreclr/blob/master/dependencies.props#L72

AndyAyersMS:
Thanks, trying that locally now...

AndyAyersMS:
Locally still having issues, though am also experiencing some network flakiness.
```
EXEC : error : Unable to resolve 'Microsoft.DotNet.xunit.performance.analysis (>= 1.0.0-alpha-build0047)' for '.NETStan
dard,Version=v1.4'. [C:\repos\coreclr2\tests\build.proj]
```

AndyAyersMS:
Based on dotnet/corefx#15767 it looks like packages were added/dropped and so a simple version bump is not likely to work.

brianrob:
@AndyAyersMS, you're right.  We'd need to publish more packages from xunit.performance to make this work.  I want to see if we can keep the xunit.performance files that come in from corefx from being copied.

Issue #9692, opened 2017-02-21
Possible JIT improvement for loops with return statements


bbowyersmyth:
Loops with a return statement in the body run slower than those without. It would be good if the JIT had a way to optimize this. If tracking this was too complex perhaps moving `return [true|false]` statements could be a starting point.

Test code
https://gist.github.com/bbowyersmyth/9514af463745528d8d290e7cd2492660

The very simple loop runs 85.7ns vs 67.4ns (20% difference). The gap can widen with additional instructions added to the body.
Current theory is that this is due to the CPUs complexity rules for the loop stream detector.

Initial suggestion by @jkotas https://github.com/dotnet/coreclr/pull/2667#r49820503
Recent discussion https://github.com/dotnet/coreclr/pull/9213

cc @mikedn

benaadams:
The inline return will add a bunch of additional `pop` statements that need to be jumped over making the body of the loop larger

bbowyersmyth:
Apart from a branch misprediction are there other factors at play that make skipping over multiple instructions more costly than one for a given iteration?

mikedn:
That branch should be correctly predicted so that's not an issue. One possible explanation for the behavior you are seeing is that predicted and taken branches and slightly more expensive than predicated and not taken branches:
```asm
G_M11558_IG05:
       0FB708               movzx    rcx, word  ptr [rax]
       440FB70A             movzx    r9, word  ptr [rdx]
       413BC9               cmp      ecx, r9d
; predicted and taken every loop iteration
       7407                 je       SHORT G_M11558_IG07
       33C0                 xor      eax, eax
G_M11558_IG06:
       4883C418             add      rsp, 24
       C3                   ret
G_M11558_IG07:
       4883C002             add      rax, 2
       4883C202             add      rdx, 2
       41FFC8               dec      r8d
       4585C0               test     r8d, r8d
       75DD                 jne      SHORT G_M11558_IG05

; versus
G_M50079_IG05:
       440FB701             movzx    r8, word  ptr [rcx]
       440FB70A             movzx    r9, word  ptr [rdx]
       453BC1               cmp      r8d, r9d
; predicted and NOT taken every loop iteration
       7518                 jne      SHORT G_M50079_IG08
       4883C102             add      rcx, 2
       4883C202             add      rdx, 2
       FFC8                 dec      eax
       85C0                 test     eax, eax
       75E5                 jne      SHORT G_M50079_IG05
```
[...]

PR #9738, opened 2017-02-23
Tweak OrdinalComparer for OrdinalIgnoreCase

stephentoub:
- OrdinalComparer was parameterized by ignoreCase, meaning every call to it switched on _ignoreCase and took a path based on whether StringComparer.Ordinal or StringComparer.OrdinalIgnoreCase was used.  As both of these are so commonly used, I've split the type into one sealed class for each, and simplified the code for each.  Even with the split, total lines of C# code decreased.
- https://github.com/dotnet/coreclr/pull/9213 added a special path to string.Equal(string, StringComparison) for OrdinalIgnoreCase.  I switched StringComparer.OrdinalIgnoreCase to use that rather than string.Compare(..., StringComparison.OrdinalIgnoreCase) == 0.  I also then removed a couple of null checks that were optimizing for comparing a non-null string against a null string, which shouldn't be the common case.
- Strongly-typed the various static readonly comparer fields on StringComparer
- Applied the same workaround used in https://github.com/dotnet/coreclr/pull/9213 that moves the return out of the loop.  In some microbenchmarks, it had no effect.  On a few, e.g. comparing "abcdefghijklmno" to "AbCdEfGhIjKlMnz" or "ssssssssss" to "SSSSSSSSSS", it had upwards of a 30% improvement.  I tagged it with a comment with the associated issue number (https://github.com/dotnet/coreclr/issues/9692) to undo if/when the JIT is better able to optimize this.

### DIFF
--- a/src/mscorlib/src/System/StringComparer.cs
+++ b/src/mscorlib/src/System/StringComparer.cs
@@ -249,22 +249,15 @@ namespace System
 
     }
 
-    // Provide a more optimal implementation of ordinal comparison.
     [Serializable]
     internal sealed class OrdinalComparer : StringComparer 
 #if FEATURE_RANDOMIZED_STRING_HASHING           
         , IWellKnownStringEqualityComparer
 #endif
     {
-        public override int Compare(string x, string y) =>
-            object.ReferenceEquals(x, y) ? 0 :
-            x == null ? -1 :
-            y == null ? 1 :
-            string.CompareOrdinal(x, y);
+        public override int Compare(string x, string y) => string.CompareOrdinal(x, y);
 
-        public override bool Equals(string x, string y) =>
-            object.ReferenceEquals(x, y) ||
-            (x != null && x.Equals(y));
+        public override bool Equals(string x, string y) => string.Equals(x, y);
 
         public override int GetHashCode(string obj)
         {
@@ -276,7 +269,7 @@ namespace System
         }
 
         // Equals/GetHashCode methods for the comparer itself. 
-        public override bool Equals(Object obj) => obj is OrdinalComparer;
+        public override bool Equals(object obj) => obj is OrdinalComparer;
         public override int GetHashCode() => nameof(OrdinalComparer).GetHashCode();
 
 #if FEATURE_RANDOMIZED_STRING_HASHING           
@@ -284,22 +277,15 @@ namespace System
 #endif
     }
 
-    // Provide a more optimal implementation of ordinal ignore-case comparison.
     [Serializable]
     internal sealed class OrdinalIgnoreCaseComparer : StringComparer
 #if FEATURE_RANDOMIZED_STRING_HASHING
         , IWellKnownStringEqualityComparer
 #endif
     {
-        public override int Compare(string x, string y) =>
-            object.ReferenceEquals(x, y) ? 0 :
-            x == null ? -1 :
-            y == null ? 1 :
-            string.Compare(x, y, StringComparison.OrdinalIgnoreCase);
+        public override int Compare(string x, string y) => string.Compare(x, y, StringComparison.OrdinalIgnoreCase);
 
-        public override bool Equals(string x, string y) =>
-            object.ReferenceEquals(x, y) ||
-            (x != null && x.Equals(y, StringComparison.OrdinalIgnoreCase));
+        public override bool Equals(string x, string y) => string.Equals(x, y, StringComparison.OrdinalIgnoreCase);
 
         public override int GetHashCode(string obj)
         {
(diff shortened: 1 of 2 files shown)


src/mscorlib/src/System/StringComparer.cs
```diff
@@ -213,83 +213,77 @@ internal sealed class CultureAwareComparer : StringComparer
...
-        }
-                
-        public override bool Equals(string x, string y) {
-            if (Object.ReferenceEquals(x ,y)) return true;
-            if (x == null || y == null) return false;
+        public override int Compare(string x, string y) =>
+            object.ReferenceEquals(x, y) ? 0 :
+            x == null ? -1 :
+            y == null ? 1 :
+            string.CompareOrdinal(x, y);
```
justinvp: `string.CompareOrdinal` itself has the reference equality and null checks. Is there a benefit to duplicating those checks here before calling `string.CompareOrdinal`, or could this method be changed to simply call `string.CompareOrdinal`?

src/mscorlib/src/System/StringComparer.cs
```diff
@@ -213,83 +213,77 @@ internal sealed class CultureAwareComparer : StringComparer
...
+    internal sealed class OrdinalIgnoreCaseComparer : StringComparer
+#if FEATURE_RANDOMIZED_STRING_HASHING
+        , IWellKnownStringEqualityComparer
+#endif
+    {
+        public override int Compare(string x, string y) =>
+            object.ReferenceEquals(x, y) ? 0 :
+            x == null ? -1 :
+            y == null ? 1 :
+            string.Compare(x, y, StringComparison.OrdinalIgnoreCase);
```
justinvp: Same question. `string.Compare(string, string, StringComparison)` also does the reference equality and null checks. Is there a benefit to duplicating those here, or could this be changed to simply call `string.Compare(x, y, StringComparison.OrdinalIgnoreCase)`?

src/mscorlib/src/System/StringComparer.cs
```diff
@@ -213,83 +213,77 @@ internal sealed class CultureAwareComparer : StringComparer
...
-        }               
-                
-        public override int GetHashCode(string obj) {
-            if( obj == null) {
-                throw new ArgumentNullException(nameof(obj));
-            }
-            Contract.EndContractBlock();
+        public override bool Equals(string x, string y) =>
+            object.ReferenceEquals(x, y) ||
+            (x != null && x.Equals(y));
```
justinvp: Similarly, could this be `public override bool Equals(string x, string y) => string.Equals(x, y);`?

src/mscorlib/src/System/StringComparer.cs
```diff
@@ -213,83 +213,77 @@ internal sealed class CultureAwareComparer : StringComparer
...
+    {
+        public override int Compare(string x, string y) =>
+            object.ReferenceEquals(x, y) ? 0 :
+            x == null ? -1 :
+            y == null ? 1 :
+            string.Compare(x, y, StringComparison.OrdinalIgnoreCase);
+
+        public override bool Equals(string x, string y) =>
+            object.ReferenceEquals(x, y) ||
+            (x != null && x.Equals(y, StringComparison.OrdinalIgnoreCase));
```
justinvp: Similarly, could this be `public override bool Equals(string x, string y) => string.Equals(x, y, StringComparison.OrdinalIgnoreCase);`?

src/mscorlib/src/System/StringComparer.cs
stephentoub: Thanks.  Seems ok to trim them down to just be the indirection.  I could see cases where reference equality would prevent the need for the non-inlined static call, but I don't think that should be the common case when going through a comparer like this.

stephentoub: @jkotas, I saw a related comment from you on another PR... what's your opinion, for example, of whether to have:
```C#
public override bool Equals(string x, string y) => string.Equals(x, y);
```
vs:
```C#
public override bool Equals(string x, string y) => ReferenceEquals(x, y) || string.Equals(x, y);
```
vs
```C#
public override bool Equals(string x, string y) => ReferenceEquals(x, y) || (x != null && y != null && string.Equals(x, y));
```
i.e. whether to delegate to string.Equals for everything or special-case reference equality and null to avoid the static call to Equals.

stephentoub:

src/mscorlib/src/System/StringComparer.cs
jkotas: Inlining the common case manually in places like this one where you can tell with reasonable confidence that they are going to be pretty hot should be fine.

Ideally, the profile guided training for corelib would take care of this. The profiling data would say that this is hot method and the codegen would inline more aggresively based on that.

jkotas:
@stephentoub Could you please resolve the conflicts?

stephentoub:
> Could you please resolve the conflicts?

Done, thanks.

PR #9741, opened 2017-02-23
Disable Test Broken By CoreFX PR # 15767

brianrob:
This is a temporary fix for #9733 to unblock the affected CI legs.


brianrob:
cc: @gkhanna79, @wtgodbe

gkhanna79:
LGTM - when can we expect the real fix for this @brianrob ?

brianrob:
@AndyAyersMS, I did not know about tests\issues.targets.  Thanks for the info.

@gkhanna79, I'll work on a real fix for this next, but I tend to agree with @AndyAyersMS that this is not specific to performance testing - we have two different versions of a package required across coreclr and corefx and because the corefx dependency gets copied over to coreclr, things break.  With respect to xunit.performance, I think we likely want to make sure that the files don't get copied over from corefx.  I will follow-up with @wtgodbe to see if there's a way to do this.

PR #9743, opened 2017-02-23
Update UsingYourBuildDocs to use .NET Core 2.0 SDK

weshaggard:
PTAL @jkotas @vancem 

FYI @ericstj @eerhardt 

skip ci please

### DIFF
--- a/Documentation/workflow/UsingYourBuild.md
+++ b/Documentation/workflow/UsingYourBuild.md
@@ -1,43 +1,136 @@
 
 # Using your .NET Core Runtime Build
 
-We assume that you have successfully built CoreCLR repository and thus have file of the form  
+We assume that you have successfully built CoreCLR repository and thus have files of the form
 ```
     bin\Product\<OS>.<arch>.<flavor>\.nuget\pkg\Microsoft.NETCore.Runtime.CoreCLR.<version>.nupkg
 ```
 And now you wish to try it out.  We will be using Windows OS as an example and thus will use \ rather
 than / for directory separators and things like Windows_NT instead of Linux but it should be
-pretty obvious how to adapt these instructions for other operating systems.  
+pretty obvious how to adapt these instructions for other operating systems.
 
 To run your newly built .NET Core Runtime in addition to the application itself, you will need
 a 'host' program that will load the Runtime as well as all the other .NET Core Framework
-code that your application needs.   The easiest way to get all this other stuff is to simply use the 
+code that your application needs. The easiest way to get all this other stuff is to simply use the
 standard 'dotnet' host that installs with .NET Core SDK.
 
 The released version of 'dotnet' tool may not be compatible with the live CoreCLR repository. The following steps
-assume use of unreleased version of 'dotnet' tool that is downloaded as part of the CoreCLR repository
-build at `<repo root>\Tools\dotnetcli`. [Add `Tools\dotnetcli` directory to your path](../building/windows-instructions.md#adding-to-the-default-path-variable)
-and type:
+assume use of a dogfood build of the .NET SDK.
 
-* dotnet -?
+## Acquire the latest nightly .NET Core 2.0 SDK
 
-and it prints some help text, you are ready.
+- [Win 64-bit Latest](https://dotnetcli.blob.core.windows.net/dotnet/Sdk/master/dotnet-dev-win-x64.latest.zip)
+- [macOS 64-bit Latest](https://dotnetcli.blob.core.windows.net/dotnet/Sdk/master/dotnet-dev-osx-x64.latest.tar.gz)
+- [Others](https://github.com/dotnet/cli/blob/master/README.md#installers-and-binaries)
 
-### Step 1: Create a App using the Default Runtime
-At this point you can create a new 'Hello World' program in the standard way. 
+To setup the SDK download the zip and extract it somewhere and add the root folder to your [path](../building/windows-instructions.md#adding-to-the-default-path-variable)
+or always fully qualify the path to dotnet in the root of this folder for all the instructions in this document.
+
+After setting up dotnet you can verify you are using the newer version by:
+
+`dotnet --info` -- the version should be greater than 2.0.0-*
+
+For another small walkthrough see [Dogfooding .NET Core 2.0 SDK](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/dogfooding.md).
+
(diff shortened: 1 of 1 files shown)


Documentation/workflow/UsingYourBuild.md
```diff
@@ -1,43 +1,51 @@
...
 a 'host' program that will load the Runtime as well as all the other .NET Core Framework
-code that your application needs.   The easiest way to get all this other stuff is to simply use the 
+code that your application needs.   The easiest way to get all this other stuff is to simply use the
 standard 'dotnet' host that installs with .NET Core SDK.
 
 The released version of 'dotnet' tool may not be compatible with the live CoreCLR repository. The following steps
-assume use of unreleased version of 'dotnet' tool that is downloaded as part of the CoreCLR repository
-build at `<repo root>\Tools\dotnetcli`. [Add `Tools\dotnetcli` directory to your path](../building/windows-instructions.md#adding-to-the-default-path-variable)
-and type:
+assume use aquire a dogfood build of the .NET SDK
```
jkotas: The following steps assume use aquire a dogfood build of the .NET SDK

Documentation/workflow/UsingYourBuild.md
```diff
@@ -1,43 +1,51 @@
...
 
-* dotnet -?
+## Install the latest nightly .NET Core SDK 2.0
 
-and it prints some help text, you are ready.
+- [Win 64-bit Latest](https://dotnetcli.blob.core.windows.net/dotnet/Sdk/master/dotnet-dev-win-x64.latest.exe)
```
jkotas: I have been always downloading ZIPs that I unpack locally, and recommending others to do the same.

Do we want to be recommending to install this globally? It tends to break up VS and VS Code. For example, https://github.com/dotnet/corert/pull/2618#issuecomment-276894601

mellinoe: Yeah, the zip files are probably a bit safer to use for simple dogfooding. I think you can just replace "exe" with "zip" and get a link from the above, e.g.: https://dotnetcli.blob.core.windows.net/dotnet/Sdk/master/dotnet-dev-win-x64.latest.zip

Documentation/workflow/UsingYourBuild.md
```diff
@@ -46,148 +54,158 @@ this by simply listing the name of the Microsoft.NETCore.Runtime.CoreCLR you bui
...
 * coreclr.dll - Most modifications (with the exception of the JIT compiler and tools) that are C++ code update
-  this DLL. 
-* System.Private.CoreLib.dll - If you modified C# it will end up here.  
+  this DLL.
+* System.Private.CoreLib.dll - If you modified C# it will end up here.
 * System.Private.CoreLib.ni.dll - the native image (code) for System.Private.Corelib.   If you modify C# code
-you will want to update both of these together in the target installation.   
+you will want to update both of these together in the target installation.
 
 Thus after making a change and building, you can simply copy the updated binary from the `bin\Product\<OS>.<arch>.<flavor>`
```
jkotas: I think we should be better to recommend to just copy the whole directory - it is what works best. It is not easy to pick the right subset correctly, and the crashes that results from inconsistent set are hard to debug.

Documentation/workflow/UsingYourBuild.md
```diff
@@ -46,148 +54,158 @@ this by simply listing the name of the Microsoft.NETCore.Runtime.CoreCLR you bui
...
-- Added the 'runtimes' line at the top level.  The runtime name includes the OS name and the architecture name
-you can find the appropriate name for your OS [here](https://github.com/dotnet/core-docs/blob/master/docs/core/rid-catalog.md).
-This tells the tools exactly which flavor of OS and processor architecture you are running on, so it can find the right
-Nuget package for the runtime.
-- Changed netcoreapp1.0 to netcoreapp1.1. This tells the tools that you want to use the latest .NET Core Framework.
-- Expanded Microsoft.NETCore.App metapackage into explicit list of the .NET Core Framework packages because of there is no good published
-build of Microsoft.NETCore.App metapackage for netcoreapp1.1 yet.
 
-### Step 4: Place your build directory and beta .NET Core Framework feed on your Nuget Path
+### Step 5: Place your build directory and beta .NET Core Framework feed on your Nuget Path
```
jkotas: I think that these complex steps with nuget packages should be marked as optional. The recommended default should be just copy over what you have built - the `Quick updates in place` step. It is what I going myself - very simple and hard to make a mistake.

Documentation/workflow/UsingYourBuild.md
weshaggard: Yes I can update the instructions to point at the zip's instead.

Documentation/workflow/UsingYourBuild.md
weshaggard: I agree I will update the instructions accordingly.

weshaggard:
I restructured the document based on feedback.

Documentation/workflow/UsingYourBuild.md
```diff
@@ -1,209 +1,227 @@
...
-use runtime and libraries that came with the dotnet.exe tool but to fetch the dependencies from the Nuget cache.  If you don't do this
-the tools will ignore your request to make the app use an explicitly specified runtime.   
-- Added the 'runtimes' line at the top level.  The runtime name includes the OS name and the architecture name
-you can find the appropriate name for your OS [here](https://github.com/dotnet/core-docs/blob/master/docs/core/rid-catalog.md).
-This tells the tools exactly which flavor of OS and processor architecture you are running on, so it can find the right
-Nuget package for the runtime.
-- Changed netcoreapp1.0 to netcoreapp1.1. This tells the tools that you want to use the latest .NET Core Framework.
-- Expanded Microsoft.NETCore.App metapackage into explicit list of the .NET Core Framework packages because of there is no good published
-build of Microsoft.NETCore.App metapackage for netcoreapp1.1 yet.
+In order to use the runtime nuget package you will need to do some further setup steps.
```
jkotas: Add something here on why somebody may want to use the nuget package? Something like:

Updating CoreCLR from raw binary output is easier for quick one-off testing. Using Nuget package is better for referencing your CoreCLR build in your actual application because of it does not require manual copying of files around each time the application is built.

weshaggard: updated.

PR #5671, opened 2016-06-10
Increase inlineability of String.Equals

jamesqo:
Currently, `string.Equals` isn't getting inlined by the JIT. This means that a new method call has to be made every time someone does something like `"foo" == "blah"`. This is in spite of the fact that the body of `string.Equals` itself is pretty small in terms of code size; all it really does is two pointer comparisons and a length check before calling `EqualsHelper`. This means that for different-length strings, we are wasting a method call every time we call `string.Equals`.

This pull request makes the following changes:

- Omits compiling the `this == null` check on CoreCLR. Since the check and the `throw new NullReferenceException` clause bloated the compiled output significantly, I removed it for .NET Core. This is only going to cause different behavior if someone handwrites IL manually/does a reverse PInvoke (as the comment notes), compiles that app for .NET Core, *and* passes null as a parameter to one of the `Equals` functions. Since for 99.99% percent of cases this won't happen and removing it will speed up apps significantly, I decided to only compile it for the framework.

- Aggregated two `if (...) return false` conditionals into an or statement, so we won't generate duplicate code to return false (and thus decrease method size).

- Added `AggressiveInlining` attributes to the `string.Equals` methods. (I didn't do it for the object-based overload, since it's more code + most of the time it's going to be done as a virtual method call, which can't be inlined anyway.)

## Perf impact

- [Test app 1](https://gist.github.com/jamesqo/90fd6e580f1dbbfe4f8d0c5e4cf98d49) (for getting JIT output)
- [Old disassembly](https://gist.github.com/jamesqo/02b3f73cd5efc105a442e68dcd4adcd6)
- [New disassembly](https://gist.github.com/jamesqo/70488223f34c7fd46e7150b6f6cb1e6e) (the JIT is able to optimize the method down to basically nothing)
- [Test app 2](https://gist.github.com/jamesqo/aa17f8e4b5a0b91a9713aab56a0d713c) (for benchmarking)
- [Old results](https://gist.github.com/jamesqo/8fb59c768683d86010ff5c336389bd3f)
- [New results](https://gist.github.com/jamesqo/85e41481c44e56d394722509f4a25a71)

cc: @jkotas @AlexGhiondea @hughbe @ellismg

PR #9744, opened 2017-02-23
Allow string.Equals to inline for different-length strings

jamesqo:
The most common case is going to be having two non-null, non-reference-equal, different-length strings. So we handle that case directly in `Equals` and move the rest of the logic to `EqualsHelper`.

Surprisingly the JIT did not inline the methods even though they're smaller than in #5671... I used `AggressiveInlining` and there is not that much code bloat. [before and after](https://gist.github.com/jamesqo/f37fd11a32d8bb2cd7dfc13b35c98ba6)

Additional notes:

- Removed `this == null` checks in `Equals(string)` and `Equals(object)`. Are these still needed for .NET Core? It mentions something about reverse P/Invoke. I don't think it will be feasible to inline the methods with the `throw new NullReferenceException()` around.

- Roslyn optimizes `str == null` to do a reference-comparison instead of calling `string.op_Equality`, so this PR will not adversely affect people comparing strings to literal nulls.

/cc @jkotas @benaadams @AndyAyersMS


src/mscorlib/src/System/String.Comparison.cs
```diff
@@ -75,9 +75,17 @@ private unsafe static int CompareOrdinalIgnoreCaseHelper(String strA, String str
 
         private unsafe static bool EqualsHelper(String strA, String strB)
         {
-            Contract.Requires(strA != null);
-            Contract.Requires(strB != null);
-            Contract.Requires(strA.Length == strB.Length);
+            Debug.Assert((strA == null || strB == null) || strA.Length == strB.Length);
+
+            if ((object)strA == strB)
```
jkotas: Why do you think moving it here is a good idea?

IIRC, it was moved up as the first check years ago because of somebody collected data that it is the right tradeoff.

jamesqo: @jkotas 

> IIRC, it was moved up as the first check years ago because of somebody collected data that it is the right tradeoff.

Before, the tradeoff was just that the case where they are non-reference equal would suffer an extra check. Now, the tradeoff is that the case where they are non-reference equal suffer an extra method call.

> Why do you think moving it here is a good idea?

If the two strings are reference equal, the only perf degradation from this PR is two more checks. The number of method calls has not changed.

jkotas: It is code-size tradeoff. You are making the code for all places that compare strings bigger. How do you know that it is a good tradeoff for programs out there?

jamesqo: @jkotas Not disagreeing with you, but how does this differ from [`List.Add`](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Collections/Generic/List.cs#L210) where AggressiveInlining was just used?

jkotas: `List.Add` showed up on the radar in ASP.NET workload, and it was not the first time - I know people were considering doing something about it before.

jamesqo:
@jkotas Through some experimentation just managed to make the instance overload inline naturally without forcing it with AggressiveInlining. The JIT seems to be able to decide whether or not to inline on context, and only does it in the loop method in the below example.

```cs
        private static bool InstanceEquals(string a, string b) => a.Equals(b);

        private static void Loop()
        {
            string left = "", right = "";

            for (int i = 0; i < 100; i++)
            {
                bool _1 = left.Equals(right);
                bool _2 = string.Equals(left, right);
            }
        }
```

```asm
; Assembly listing for method Program:InstanceEquals(ref,ref):bool
; optimized code
; rsp based frame
; partially interruptible
;
;  V00 arg0         [V00,T00] (  5,   4.5)     ref  ->  rsi
;  V01 arg1         [V01,T01] (  4,   3.5)     ref  ->  rdi
;  V02 tmp0         [V02,T02] (  3,   2  )    bool  ->  rax
;  V03 OutArgs      [V03    ] (  1,   1  )  lclBlk (32) [rsp+0x00]
;
; Lcl frame size = 40

G_M7161_IG01:
       57                   push     rdi
       488BFA               mov      rdi, rdx

G_M7161_IG02:
       8B0E                 mov      ecx, dword ptr [rsi]
       488BCE               mov      rcx, rsi
       488BD7               mov      rdx, rdi
       E8171C715E           call     String:ShouldCallEqualsHelper(ref):bool:this
       85C0                 test     eax, eax
       740D                 je       SHORT G_M7161_IG03
       488BCE               mov      rcx, rsi
       488BD7               mov      rdx, rdi
       E8C80A715E           call     String:EqualsHelper(ref,ref):bool
       EB02                 jmp      SHORT G_M7161_IG04

G_M7161_IG03:

G_M7161_IG04:
       5F                   pop      rdi
       C3                   ret

; Total bytes of code 51, prolog size 6 for method Program:InstanceEquals(ref,ref):bool
; Assembly listing for method Program:Loop()
; optimized code
; rsp based frame
; partially interruptible
;
;  V00 loc0         [V00,T00] (  9,  20  )     ref  ->  rsi
;  V01 loc1         [V01,T02] (  7,  16  )     ref  ->  rdi
;  V02 loc2         [V02,T03] (  4,  13  )     int  ->  rbx
;* V03 tmp0         [V03    ] (  0,   0  )    bool  ->  zero-ref
;  V04 tmp1         [V04,T04] (  4,   8  )    bool  ->  rcx
;* V05 tmp2         [V05    ] (  0,   0  )    bool  ->  zero-ref
;  V06 OutArgs      [V06    ] (  1,   1  )  lclBlk (32) [rsp+0x00]
;  V07 cse0         [V07,T05] (  6,   6  )     ref  ->  rcx
;  V08 cse1         [V08,T01] (  8,  16  )     int  ->  rcx
;
; Lcl frame size = 32

G_M7029_IG01:
       57                   push     rdi
       53                   push     rbx

G_M7029_IG02:
       48B9603000903D020000 mov      rcx, 0x23D90003060
       488BF9               mov      rdi, rcx
       33DB                 xor      ebx, ebx

G_M7029_IG03:
       8B0E                 mov      ecx, dword ptr [rsi]
       4885FF               test     rdi, rdi
       740F                 je       SHORT G_M7029_IG04
       8B4E08               mov      ecx, dword ptr [rsi+8]
       3BC9                 cmp      ecx, ecx
       0F94C1               sete     cl
       0FB6C9               movzx    rcx, cl
       85C9                 test     ecx, ecx
       740B                 je       SHORT G_M7029_IG05

G_M7029_IG04:
       488BCE               mov      rcx, rsi
       488BD7               mov      rdx, rdi
       E8630A715E           call     String:EqualsHelper(ref,ref):bool

G_M7029_IG05:
       488BCE               mov      rcx, rsi
       488BD7               mov      rdx, rdi
       E8681E715E           call     String:ShouldCallEqualsHelper(ref,ref):bool
       85C0                 test     eax, eax
       740B                 je       SHORT G_M7029_IG06
       488BCE               mov      rcx, rsi
       488BD7               mov      rdx, rdi
       E8490A715E           call     String:EqualsHelper(ref,ref):bool

G_M7029_IG06:
       FFC3                 inc      ebx
       83FB64               cmp      ebx, 100
       7CBE                 jl       SHORT G_M7029_IG03

G_M7029_IG07:
       5B                   pop      rbx
       5F                   pop      rdi
       C3                   ret

; Total bytes of code 102, prolog size 7 for method Program:Loop()
```

I still have to see if I can get the static one to inline without AI.

jkotas:
> methods inline naturally without 

Well, it is exploiting a hole in the current inlining heuristic. You can trick the current inliner to inline arbitrary large code this way.

@AndyAyersMS Do you have opinion about this change?

AndyAyersMS:
Seems like we really don't know enough to evaluate the impact of this change. There are a lot of factors to take into account:

- Static and dynamic frequency of various calling patterns
- Importance of those calling patterns in key scenarios
- Ability of the jit to partially specialize or simplify
- Code size impact
- Jit time impact
- Likelihood of any of the above changing over time

If all of the above are known then one can work up an analysis that shows that a change is an aggregate improvement w/o unduly harming any (important) cases. But reliably gathering this data is a lot of work.

I have ambitions of changing the jit's discretionary inlining heuristics -- hopefully relatively soonish -- so would not want the motivation for something like this to rely heavily on the inliner behaving like it does now.

jamesqo:
@AndyAyersMS Sounds reasonable. I can close this PR for now.

> I have ambitions of changing the jit's discretionary inlining heuristics -- hopefully relatively soonish

If you remember, please feel free to ping me when you make the changes.

AndyAyersMS:
There is a preliminary version of the new inline heuristics already in the jit, if you want to kick the tires. For details on this, [see my notes](https://github.com/AndyAyersMS/PerformanceExplorer/blob/master/notes/notes-aug-2016.md).

Evaluation of changes like this is not easy (similar to what I noted above). In particular when the inlining heuristic changes almost all code generated by the jit may be impacted. So it is not always easy to contain the "damage" and very broad evaulations are necessary.

I aim to get back to working on this relatively soon. Have been waiting in part for inspiration to strike, for profile feedback mechanisms (aka IBC) to be enabled in CoreCLR, and for the easy ability to run more realistic scenarios. Also I have, so far unsuccessfully, been trying to recruit collaborators with stronger ML backgrounds....

jamesqo:
> There is a preliminary version of the new inline heuristics already in the jit, if you want to kick the tires. For details on this, see my notes.

Pretty interesting read; did not realize that optimizing inlines could be such a complex business. Thanks for sharing.

PR #9749, opened 2017-02-23
Use -f instead of -n when copying from Core_Root in runtest.sh

wtgodbe:
We should force copy everything from Core_Root when creating coreoverlay for a non-windows test run - previously, we were letting CoreFX binaries "win" over CoreCLR binaries, when we can more safely depend on the CoreCLR ones than the CoreFX ones (this caused the failures from https://github.com/dotnet/coreclr/issues/9733).

### DIFF
--- a/tests/runtest.sh
+++ b/tests/runtest.sh
@@ -368,7 +368,8 @@ function create_core_overlay {
     if [ -d "$mscorlibDir/bin" ]; then
         cp -f -v "$mscorlibDir/bin/"* "$coreOverlayDir/" 2>/dev/null
     fi
-    cp -n -v "$testDependenciesDir"/* "$coreOverlayDir/" 2>/dev/null
+    cp -f -v "$testDependenciesDir/"xunit* "$coreOverlayDir/" 2>/dev/null
+    cp -n -v "$testDependenciesDir/"* "$coreOverlayDir/" 2>/dev/null
     if [ -f "$coreOverlayDir/mscorlib.ni.dll" ]; then
         # Test dependencies come from a Windows build, and mscorlib.ni.dll would be the one from Windows
         rm -f "$coreOverlayDir/mscorlib.ni.dll"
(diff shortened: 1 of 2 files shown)


wtgodbe:
CC @brianrob

tests/runtest.sh
```diff
@@ -368,7 +368,7 @@ function create_core_overlay {
     if [ -d "$mscorlibDir/bin" ]; then
         cp -f -v "$mscorlibDir/bin/"* "$coreOverlayDir/" 2>/dev/null
     fi
-    cp -n -v "$testDependenciesDir"/* "$coreOverlayDir/" 2>/dev/null
+    cp -f -v "$testDependenciesDir"/* "$coreOverlayDir/" 2>/dev/null
```
wtgodbe: @brianrob this is actually where the xunit .dll's we want should be coming from, not "CoreClrBinDir" (That refers to the non-windows CoreCLR build). The problem was that we were using -n here, which says "Do NOT overwrite destination files" (hence the message in the logs that said we weren't overwriting those xunit .dlls)

brianrob: From a cursory look, it does look like that's where we'd want the test binaries from, but it looks like there maybe something else in that directory that is causing huge numbers of tests to fail in the CI.

brianrob:
FYI @AndyAyersMS, @valenis

wtgodbe:
Appears we were using "-n" there for a reason - I just changed this so we only force copy xunit binaries from the windows test dependency drop.

AndyAyersMS:
You might look at what happens over in tests\scripts\run-xunit-perf.sh since it does something similar and seemed to mostly work (though notably, it did not work for the serialization tests).

wtgodbe:
Looks like those serialization tests are now passing - @brianrob are you fine with force-copying xunit stuff from CoreCLR, but nothing else?

PR #9755, opened 2017-02-24
[ARM,ARM64/Linux] Fix undefined FixContextHandler reference

parjong:
#9685 cleans up FixContextHandler from Linux, but FixContextHandler is used as a personality routine in ARM/Linux and ARM64/Linux, which results in undefined reference issue.

This commit replaces it with UnhandledExceptionHandlerUnix to fix undefined reference issue (FixContextHandler seems to be Windows-specific personality routine).

### DIFF
--- a/src/vm/arm/ehhelpers.S
+++ b/src/vm/arm/ehhelpers.S
@@ -79,7 +79,7 @@ OFFSET_OF_FRAME=(4 + SIZEOF__GSCookie)
 // Helpers for async (NullRef, AccessViolation) exceptions
 //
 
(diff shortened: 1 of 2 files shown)


parjong:
@janvorli @jkotas Please take a look.

janvorli:
The NakedThrowHelper2 and NakedThrowHelper are not used on Unix. For amd64 and x86, the NakedThrowHelper is defined as dummy asserting stub for now, but we could just remove them both and ifdef out the code that uses them.
For now, I'll merge this change to quickly fix the undefined reference and I'll create a separate github issue for removal of the NakedThrowHelper for Unix.

PR #9756, opened 2017-02-24
JIT: keep ref count for runtime generic context lookups

AndyAyersMS:
Keep a ref count for runtime generic context lookups, and track when
an inlinee requires a runtime lookup but then ignores the resulting
context. Only report/keep the context alive if there are known uses.

This pattern happens frequently in inlined generic code that uses
type tests to extract properties of types.

Addresses #1691.

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -7479,7 +7479,7 @@ GenTreePtr Compiler::fgGetCritSectOfStaticMethod()
         // Collectible types requires that for shared generic code, if we use the generic context paramter
         // that we report it. (This is a conservative approach, we could detect some cases particularly when the
         // context parameter is this that we don't need the eager reporting logic.)
-        lvaGenericsContextUsed = true;
+        lvaGenericsContextUseCount++;
 
         switch (kind.runtimeLookupKind)
         {
@@ -22445,6 +22445,28 @@ GenTreePtr Compiler::fgInlinePrependStatements(InlineInfo* inlineInfo)
 
 void Compiler::fgInlineAppendStatements(InlineInfo* inlineInfo, BasicBlock* block, GenTreePtr stmtAfter)
 {
+    // If this inlinee was passed a generic context but didn't look at
+    // it, we can decrement the "generic context was used" ref count.
+    if ((inlineInfo->inlineCandidateInfo->methInfo.args.callConv & CORINFO_CALLCONV_PARAMTYPE) != 0)
+    {
+        // Did the context require the caller to perform a runtime lookup?
+        if (inlineInfo->inlineCandidateInfo->exactContextNeedsRuntimeLookup)
+        {
+            // Fetch the temp for the generic context
+            const unsigned typeCtxtArg = inlineInfo->typeContextArg;
+            const unsigned tmpNum      = inlineInfo->lclTmpNum[typeCtxtArg];
+
+            // Was it used in the inline body?
+            if (tmpNum == BAD_VAR_NUM)
+            {
+                // No, so the runtime lookup is not needed.
+                JITDUMP("Inlinee ignores runtime lookup generics context\n");
+                assert(lvaGenericsContextUseCount > 0);
+                lvaGenericsContextUseCount--;
+            }
+        }
+    }
+
     // Null out any gc ref locals
     if (!inlineInfo->HasGcRefLocals())
     {
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -22445,21 +22445,31 @@ GenTreePtr Compiler::fgInlinePrependStatements(InlineInfo* inlineInfo)
 
 void Compiler::fgInlineAppendStatements(InlineInfo* inlineInfo, BasicBlock* block, GenTreePtr stmtAfter)
 {
-    // If this inlinee was passed a generic context but didn't look at
-    // it, we can decrement the "generic context was used" ref count.
+    // If this inlinee was passed a runtime lookup generic context and
+    // ignores it, we can decrement the "generic context was used" ref
+    // count, because we created a new lookup tree and incremented the
+    // count when we imported the type parameter argument to pass to
+    // the inlinee. See corresponding logic in impImportCall that
+    // checks the sig for CORINFO_CALLCONV_PARAMTYPE.
+    //
+    // Does this method require a context (type) parameter?
     if ((inlineInfo->inlineCandidateInfo->methInfo.args.callConv & CORINFO_CALLCONV_PARAMTYPE) != 0)
     {
-        // Did the context require the caller to perform a runtime lookup?
+        // Did the computation of that parameter require the
+        // caller to perform a runtime lookup?
         if (inlineInfo->inlineCandidateInfo->exactContextNeedsRuntimeLookup)
         {
-            // Fetch the temp for the generic context
+            // Fetch the temp for the generic context as it would
+            // appear in the inlinee's body.
             const unsigned typeCtxtArg = inlineInfo->typeContextArg;
             const unsigned tmpNum      = inlineInfo->lclTmpNum[typeCtxtArg];
 
             // Was it used in the inline body?
             if (tmpNum == BAD_VAR_NUM)
             {
-                // No, so the runtime lookup is not needed.
+                // No -- so the associated runtime lookup is not needed
+                // and also no longer provides evidence that the generic
+                // context should be kept alive.
                 JITDUMP("Inlinee ignores runtime lookup generics context\n");
                 assert(lvaGenericsContextUseCount > 0);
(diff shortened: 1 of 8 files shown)


AndyAyersMS:
@JosephTremoulet PTAL
cc @dotnet/jit-contrib 

This is somewhat but unapologetically hacky -- because of the aggressive expansion of the context lookup into a series of statements we lose the easy connection between the root of the lookup (the `this` we must keep alive) and the subsequent context use. Would be happy to entertain thoughts on alternative approaches (perhaps deferring expansion until after optimization or something?)

281 methods impacted in corelib, all size wins. Many in Span methods. Wider FX diffs seem broken at the moment.

Typical diff:
```asm
;; Assembly listing for method System.Span`1[__Canon][System.__Canon]:op_Equality(struct,struct):bool

;; Before
; Lcl frame size = 8
G_M27353_IG01:
       push     rax
       mov      qword ptr [rsp], rcx
G_M27353_IG02:
       mov      eax, dword ptr [rdx+8]
       cmp      eax, dword ptr [r8+8]
       jne      SHORT G_M27353_IG04
       mov      rax, bword ptr [rdx]
       mov      rdx, bword ptr [r8]
       cmp      rax, rdx
       sete     al
       movzx    rax, al
G_M27353_IG03:
       add      rsp, 8
       ret      
G_M27353_IG04:
       xor      eax, eax
G_M27353_IG05:
       add      rsp, 8
       ret

;; After
; Lcl frame size = 0
G_M27353_IG01:
       nop      
G_M27353_IG02:
       mov      eax, dword ptr [rdx+8]
       cmp      eax, dword ptr [r8+8]
       jne      SHORT G_M27353_IG04
       mov      rax, bword ptr [rdx]
       mov      rdx, bword ptr [r8]
       cmp      rax, rdx
       sete     al
       movzx    rax, al
G_M27353_IG03:
       ret      
G_M27353_IG04:
       xor      eax, eax
G_M27353_IG05:
       ret

**Comment by AndyAyersMS** [comment]
Assembly for the test method from #1691:
```asm
; Lcl frame size = 0

G_M16884_IG01:

G_M16884_IG02:
       8B4908               mov      ecx, dword ptr [rcx+8]
       FFC9                 dec      ecx
       48B8D801B79DFA7F0000 mov      rax, 0x7FFA9DB701D8

G_M16884_IG03:
```

src/jit/importer.cpp
```diff
@@ -7401,7 +7401,7 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
 #endif // defined(DEBUG) || defined(INLINE_DATA)
 
                 // Is it an inline candidate?
-                impMarkInlineCandidate(call, exactContextHnd, callInfo);
+                impMarkInlineCandidate(call, exactContextHnd, exactContextNeedsRuntimeLookup == TRUE, callInfo);
```
briansull: Can't  this be written without the  `== TRUE`

src/jit/importer.cpp
```diff
@@ -7625,7 +7625,7 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
 #endif // defined(DEBUG) || defined(INLINE_DATA)
 
         // Is it an inline candidate?
-        impMarkInlineCandidate(call, exactContextHnd, callInfo);
+        impMarkInlineCandidate(call, exactContextHnd, exactContextNeedsRuntimeLookup == TRUE, callInfo);
```
briansull: Same

src/jit/importer.cpp
AndyAyersMS: It could be omitted if `BOOL` and `bool` were the same type, but since they're not, I get warnings about possible "performance issues".

briansull: Can you try changing the type of this local to bool (declared on line 6379)?
I don't think that it shoudl be declared a BOOL. 
 (Although the field of exactly the same name in  CORINFO_CALL_INFO is declared a BOOL,
these two aren't assigned into each other)

src/jit/flowgraph.cpp
```diff
@@ -22445,6 +22445,28 @@ GenTreePtr Compiler::fgInlinePrependStatements(InlineInfo* inlineInfo)
...
+            const unsigned typeCtxtArg = inlineInfo->typeContextArg;
+            const unsigned tmpNum      = inlineInfo->lclTmpNum[typeCtxtArg];
+
+            // Was it used in the inline body?
+            if (tmpNum == BAD_VAR_NUM)
+            {
+                // No, so the runtime lookup is not needed.
+                JITDUMP("Inlinee ignores runtime lookup generics context\n");
+                assert(lvaGenericsContextUseCount > 0);
+                lvaGenericsContextUseCount--;
```
JosephTremoulet: I'm having trouble connecting the dots.  Are you expecting the decrement here to offset an increment made in one particular increment site, or it could be any of them?  E.g. I went looking to see if importing a call with `exactContextNeedsRuntimeLookup` calls into `getRuntimeContextTree`, but couldn't find it.  You've mentioned we expand aggressively and so the connection is lost; maybe a more concrete description of what the expansion looks like would be useful as a comment here or in the commit message.

AndyAyersMS: Yeah, I figured this would be confusing.

The current code assumes that the generic context must be kept alive whenever `getRuntimeContextTree` is called, by setting `lvaGenericsContextIsUsed` true. This flag is checked over in `lvaReportParamTypeArg` and `lvaKeepAliveAndReportThis`. Note because we won't inline methods that internally request runtime lookups, in any given method there is at most one generics context in play and it is either the this pointer or the type parameter, and all these runtime lookup trees start from that context.

We don't cache/memoize these lookups so each importer operation that requires on generates a new tree. Given the potential indirect/helper lookups required each "tree" is really 3 statements.

Some subset of these tree-inspiring importations come from calls, where the result of the lookup is passed as an explicit type parameter. For inlinable calls, this happens when importing calls with CORINFO_CALLCONV_PARAMTYPE via `impTokenToHandle`. And some subset of those calls may end up getting inlined. And some subset of those inlined calls will ignore the generic context. And when that happens the results of the tree lookup are unused.

We also don't remove any dead / unreachable code throughout importation and inlining. So at the end of inlining, we have created N lookup trees, N-k of which we know for sure are ignored.

Subsequently the optimizer runs and will delete those k trees as dead code. It may also CSE trees and/or delete even more trees as dead code. So after optimization there are N-k-z trees.

Currently we will always report the context as live if N > 0. This change introduces ref counting to try and spot the case where N == k and if so, avoid keeping the generics context -- the root of all the runtime lookup trees -- live and reportable in the gc info.

A better fix would be to somehow realize what the existence of these trees implies and get the cases where N == k + z too. However I don't know of any examples where z > 0 -- maybe you'd see this from an unused ldvirtfn.

There's a deeper issue here that deserves more consideration -- the question of why the context needs to be kept live in the first place, above and beyond the explicit VM requests (eg for generic catch type filters). 

I don't claim to fully understand this case (the code refers to "collectible classes"). If I had to guess I'd say it was to guard against cases where we have a last use of the generic parameter value at a call that ignores the parameter, so that when the callee is active its class can't be GCd. If so then that contract is still preserved here.

@jkotas is that last bit more or less right?

jkotas: > @jkotas is that last bit more or less right?

Yes.

JosephTremoulet: Got it, thanks.  Something like "// Decrement the ref count, which must have been incremented by getRuntimeContextTree (via impImport[Parent]Handle) when impImportCall manifested the context argument" would have given me the breadcrumbs I was trying to connect, fwiw.

AndyAyersMS: Will do.

briansull: "collectible classes" are also known as dynamic methods and LCG (which in turn mean lightweight codegen)

It happens when a application creates a method dynamically at runtiume by specify it IL and then executing it.

they are "collectable" because when a full GC occurs it also marks each of these dynamic methods as either reachable (and thus alive) or unreachable and thus collected and the JIT code buffer and associated VM datastructures are put onto a free list to be reused (by a future dynamic method)

Reporting the 'this' pointer for such a method insures that the class and any methods that it contains are not collected by the GC.

AndyAyersMS: Yeah, I more or less got that far. The question was more why the normal GC reporting isn't sufficient to keep the collectible classes alive, such that we need this extra bit of logic in the JIT.

briansull:
LGTM

AndyAyersMS:
Ubuntu failure [(log)](https://ci.dot.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_tst_prtest/2924/) looks like an instance of #9750.

Retrying ... 
@dotnet-bot retest Ubuntu x64 Checked Build and Test

PR #9759, opened 2017-02-24
Remove getcontext and setcontext usage

janvorli:
It turns out that the getcontext and setcontext that I have used in my stack
overflow reporting change are not present e.g. on Alpine Linux or on Android.
So I am replacing their usage with RtlCaptureContext and RtlRestoreContext
instead.

I have also found that the addition of the .cfi_adjust_cfa_offset to the
PROLOG_SAVE_REG_PAIR has broken unwinding of all helpers that use
PROLOG_WITH_TRANSITION_BLOCK, because the PROLOG_STACK_ALLOC macro
updates the CFA offset. So I am fixing that by removing the CFA offset
updating from the PROLOG_STACK_ALLOC and adding explicit one to the
ARM64 CallSignalHandlerWrapper.

### DIFF
--- a/src/pal/inc/unixasmmacrosarm64.inc
+++ b/src/pal/inc/unixasmmacrosarm64.inc
@@ -43,7 +43,6 @@ C_FUNC(\Name\()_End):
 
 .macro PROLOG_STACK_ALLOC Size
         sub sp, sp, \Size
-        .cfi_adjust_cfa_offset \Size
 .endm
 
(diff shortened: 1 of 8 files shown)


janvorli:
And please do not merge that in yet -I need to verify it on x86 and arm32 yet.

janvorli:
x86 and arm32 verified

janvorli:

adityamandaleeka:
LGTM

PR #9767, opened 2017-02-24
Avoid unnecessary branch in CultureAwareComparer

justinvp:
Determine the `CompareOptions` up-front instead of during each operation.

### DIFF
--- a/src/mscorlib/src/System/StringComparer.cs
+++ b/src/mscorlib/src/System/StringComparer.cs
@@ -179,29 +179,28 @@ namespace System
         , IWellKnownStringEqualityComparer
 #endif
     {
-        private CompareInfo _compareInfo;
-        private bool _ignoreCase;
+        private readonly CompareInfo _compareInfo;
+        private readonly CompareOptions _options;
 
         internal CultureAwareComparer(CultureInfo culture, bool ignoreCase)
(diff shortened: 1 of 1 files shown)


src/mscorlib/src/System/StringComparer.cs
```diff
@@ -179,29 +179,28 @@ internal sealed class CultureAwareComparer : StringComparer
...
         }
 
         public override bool Equals(string x, string y)
         {
-            if (Object.ReferenceEquals(x, y)) return true;
+            if (object.ReferenceEquals(x, y)) return true;
             if (x == null || y == null) return false;
-
-            return (_compareInfo.Compare(x, y, _ignoreCase ? CompareOptions.IgnoreCase : CompareOptions.None) == 0);
+            return (_compareInfo.Compare(x, y, _options) == 0);
```
stephentoub: Nit: no need for the (existing) parens around the expression

justinvp:
@dotnet-bot test CentOS7.1 x64 Debug Build and Test please (#8321)

src/mscorlib/src/System/StringComparer.cs
justinvp: Fixed

PR #9768, opened 2017-02-24
Remove  `this == null` checks from string.Equals

jamesqo:
Separated from https://github.com/dotnet/coreclr/pull/9744

Are these checks still needed in .NET Core? The comments mention 2 things: 1) people using reverse P/Invoke, and 2) people using call instead of callvirt, the latter of which is emitted by C#/VB compilers for all classes. Both of these seem like they would be pretty rare.

/cc @jkotas

### DIFF
--- a/src/mscorlib/src/System/String.Comparison.cs
+++ b/src/mscorlib/src/System/String.Comparison.cs
@@ -850,9 +850,6 @@ namespace System
         // Determines whether two strings match.
         public override bool Equals(Object obj)
         {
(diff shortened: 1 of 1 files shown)


jkotas:
@jamesqo Could you please port the change to CoreRT when you get a chance?

jamesqo:
@jkotas Just checked. It looks like the checks were already removed from corert https://github.com/dotnet/corert/blob/master/src/System.Private.CoreLib/src/System/String.Comparison.cs#L757

jkotas:
Thanks for checking!

PR #9769, opened 2017-02-24
Remove OSX FileStream Lock/Unlock

ianhays:
OSX doesn't support usage of both fcntl and flock. Since we're already using one in FileShare for the entire file, we cannot enable partial file locking like we do on other Unix platforms. The alternative is to throw a PNSE and suggest using FileShare on the whole file instead.

resolves https://github.com/dotnet/corefx/issues/5964

@stephentoub @JeremyKuhne


src/mscorlib/src/System.Private.CoreLib.txt
```diff
@@ -1719,6 +1719,8 @@ event_Barrier_PhaseFinished=Barrier finishing phase {1}.
 ; Unix threading
 PlatformNotSupported_NamedSynchronizationPrimitives=The named version of this synchronization primitive is not supported on this platform.
 PlatformNotSupported_NamedSyncObjectWaitAnyWaitAll=Wait operations on multiple wait handles including a named synchronization primitive are not supported on this platform.
+; OSX File locking
+PlatformNotSupported_OSXFileLocking=Locking/unlocking file regions is not supported. Use FileShare on the entire file instead.
```
JeremyKuhne: Should have "supported _on this platform_" in the string.

ianhays: Updated.

PR #9771, opened 2017-02-24
Remove NakedThrowHelper and ifdef-out its callers

janvorli:
This change removes NakedThrowHelper function for Unix since it was not used.
It also ifdefs out its upstream callers.


janvorli:
@jkotas can you take a look please?

PR #9772, opened 2017-02-24
Use clock_gettime(CLOCK_REALTIME) in DateTime.UtcNow

stephentoub:
gettimeofday() is currently used in DateTime.UtcNow.  Depending on the system, it may have accuracy and precision in line with either clock_gettime(CLOCK_REALTIME) or clock_gettime(CLOCK_REALTIME_COARSE), but in returning its data it's necessarily limited by struct timeval's tv_usec field (microseconds), whereas clock_gettime uses struct timespec which has tv_nsec (nanoseconds).  This commit switches to use clock_gettime(CLOCK_REALTIME) to gain the additional precision (though the FILETIME through which the data is returned only supports 100-nsec precision).  On my machine, there was no measurable impact to throughput.

cc: @vancem, @mj1856

### DIFF
--- a/src/pal/src/file/filetime.cpp
+++ b/src/pal/src/file/filetime.cpp
@@ -504,23 +504,29 @@ PALAPI
 GetSystemTimeAsFileTime(
             OUT LPFILETIME lpSystemTimeAsFileTime)
 {
-    struct timeval Time;
-
     PERF_ENTRY(GetSystemTimeAsFileTime);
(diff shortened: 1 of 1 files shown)


janvorli:
@stephentoub the CLOCK_REALTIME is not avaiable on OSX.

stephentoub:
Yeah, I'm fixing that now :)

stephentoub:
@janvorli, look ok now?

mj1856:
Thanks!

PR #9773, opened 2017-02-24
Generate OAK_NO_THROW assertions from (uint)i < (uint)a.len checks

mikedn:
If `(uint)i < (uint)a.len` is true then we know that `I` is always within bounds and so `a[i]` does not need a range check.


src/jit/compiler.h
```diff
@@ -5710,6 +5711,7 @@ class Compiler
                 __int64 lconVal;
                 double  dconVal;
                 Range   u2;
+                bool    edge;
```
briansull: Instead on have 'edge' be a bool, how about storing either VNF_GT_UN or VNF_LE_UN as its value.

briansull: Oh, I see there are four cases here.  Still I'm not sure I like the term 'edge' as it is not clear what it means for  it  to be true or false.

mikedn: Would using an enum instead of a bool make it more clear? Say:
```C++
enum optAssertionEdge
{
    OAE_FLOW_INSENSITIVE,
    OAE_NEXT_EDGE,             // assertion holds for bbNext
    OAE_JUMP_EDGE              // assertion holds for bbJumpDest
};
```
Also, it may be better to not use op2 to store this information, there already seems to be some weirdness in the way assertion operands and their `vn` members are handled.

Anyway, I'm still looking for alternatives in this area. The obvious solution would be to add the complementary `OAK_THROW` like all other JTRUE generated assertions do. But such an assertion is unlikely to provide any useful information yet burn an assertion table entry and we don't have too many entries to begin with.

mikedn: Well, this edge stuff is definitely wrong as it fails to solve the problem it's supposed to solve - avoid creating unnecessary assertions. If the edge information is included in the assertion then it must be considered by the `Equals` check so we risk creating up to 3 assertions for the same `vnIdx` `vnLen` pair.

We need to somehow pass this information from `optCreateJTrueBoundsAssertion` to `optComputeAssertionGen` but it has to be "out of band", probably in a GTF flag or in a bit taken from `AssertionIndex`.

mikedn: And speaking of `Equals` - it doesn't work properly in the case of `OAK_NO_THROW` assertions. It doesn't pay attention to `vnIdx` and `vnLen` and the only thing that keeps 2 different assertions from being considered equal is `O2K_INVALID` which always results in inequality. Of course, this means that we generate duplicate assertions. There are 2 large methods in corelib where we fail to eliminate a few range checks because the assertion table fills up.

pgavlin:
cc @JosephTremoulet

mikedn:
jit-diff summary:
```
Total bytes of diff: -1729 (-0.01 % of base)
    diff is an improvement.
       -1729 : System.Private.CoreLib.dasm (-0.05 % of base)
        -244 : System.Private.CoreLib.dasm - System.Internal:SZArrayHelper(ref) (19 methods)
         -59 : System.Private.CoreLib.dasm - System.UnitySerializationHolder:AddElementTypes(ref,ref):ref
         -27 : System.Private.CoreLib.dasm - System.Diagnostics.Tracing.EventSource:InitializeProviderMetadata():this
         -24 : System.Private.CoreLib.dasm - System.Collections.Generic.List`1[Int16][System.Int16]:Add(short):this
         -24 : System.Private.CoreLib.dasm - System.Collections.Generic.List`1[SByte][System.SByte]:Add(byte):this
108 total methods with size differences (108 improved, 0 regressed).
```

mikedn:
`List<int>.Add(int)` example:
```
; before
G_M61899_IG01:
       sub      rsp, 40
       nop      
G_M61899_IG02:
       mov      rax, gword ptr [rcx+8]
       mov      r8d, dword ptr [rcx+24]
       inc      dword ptr [rcx+28]
       cmp      dword ptr [rax+8], r8d
       jbe      SHORT G_M61899_IG04
       lea      r9d, [r8+1]
       mov      dword ptr [rcx+24], r9d
       jae      SHORT G_M61899_IG06
       movsxd   rcx, r8d
       mov      dword ptr [rax+4*rcx+16], edx
G_M61899_IG03:
       add      rsp, 40
       ret      
G_M61899_IG04:
       lea      rax, [(reloc)]
G_M61899_IG05:
       add      rsp, 40
       rex.jmp  rax
G_M61899_IG06:
       call     CORINFO_HELP_RNGCHKFAIL
       int3     
; Total bytes of code 68, prolog size 5 for method System.Collections.Generic.List`1[Int32][System.Int32]:Add(int):this

; after
G_M61899_IG01:
       nop      
G_M61899_IG02:
       mov      rax, gword ptr [rcx+8]
       mov      r8d, dword ptr [rcx+24]
       inc      dword ptr [rcx+28]
       cmp      dword ptr [rax+8], r8d
       jbe      SHORT G_M61899_IG04
       lea      r9d, [r8+1]
       mov      dword ptr [rcx+24], r9d
       movsxd   rcx, r8d
       mov      dword ptr [rax+4*rcx+16], edx
G_M61899_IG03:
       ret      
G_M61899_IG04:
       lea      rax, [(reloc)]
G_M61899_IG05:
       rex.jmp  rax
; Total bytes of code 48, prolog size 5 for method System.Collections.Generic.List`1[Int32][System.Int32]:Add(int):this
```

pgavlin:
@mikedn is this ready for review, then?

mikedn:
@pgavlin Yes (assuming that build passes)

pgavlin:
Awesome.

@JosephTremoulet @briansull @dotnet/jit-contrib PTAL

src/jit/assertionprop.cpp
```diff
@@ -1826,6 +1828,32 @@ Compiler::AssertionIndex Compiler::optCreateJTrueBoundsAssertion(GenTreePtr tree
...
+    else if (vnStore->IsVNArrLenUnsignedBound(relop->gtVNPair.GetConservative(), &arrLenUnsignedBnd))
+    {
+        assert(arrLenUnsignedBnd.vnIdx != ValueNumStore::NoVN);
+        assert((arrLenUnsignedBnd.cmpOper == VNF_LT_UN) || (arrLenUnsignedBnd.cmpOper == VNF_GE_UN));
+        assert(vnStore->IsVNArrLen(arrLenUnsignedBnd.vnLen));
+
+        AssertionDsc dsc;
+        dsc.assertionKind = OAK_NO_THROW;
+        dsc.op1.kind      = O1K_ARR_BND;
+        dsc.op1.vn        = relop->gtVNPair.GetConservative();
```
mikedn: Previously `OAK_NO_THROW` assertions were only created by `optCreateAssertion` and that one sets `op1.vn` to 0. That's likely incorrect as 0 seems to be the VN for `null`. Using the VN of the relop makes more sense here but ultimately it doesn't seem to make any difference for `OAK_NO_THROW` assertions.

src/jit/assertionprop.cpp
```diff
@@ -5081,16 +5132,12 @@ void Compiler::optAssertionPropMain()
...
-                if (tree->HasAssertion())
+                AssertionIndex index = tree->GetAssertion();
+                if (index != NO_ASSERTION_INDEX)
                 {
-                    BitVecOps::AddElemD(apTraits, assertions, tree->GetAssertion() - 1);
-
-                    // Also include any implied assertions for the tree node.
-                    optImpliedAssertions((AssertionIndex)tree->GetAssertion(), assertions);
+                    optImpliedAssertions(index, assertions);
+                    BitVecOps::AddElemD(apTraits, assertions, index - 1);
```
mikedn: The order doesn't really matter but I swapped the above 2 lines to be consistent with similar code in `optComputeAssertionGen`.

src/jit/assertionprop.cpp
```diff
@@ -4446,47 +4469,69 @@ ASSERT_TP* Compiler::optComputeAssertionGen()
...
+                    index &= OAE_INDEX_MASK;
+                    // Currently OAE_NEXT_EDGE is only used with OAK_NO_THROW assertions
+                    assert(optGetAssertion(index)->assertionKind == OAK_NO_THROW);
+                    // Don't bother with implied/complementary assertions, there aren't any for OAK_NO_THROW
+                    BitVecOps::AddElemD(apTraits, valueGen, index - 1);
+                }
+                else
                 {
-                    if (assertionIndex[i] > 0)
+                    optImpliedAssertions(index, jumpDestValueGen);
```
briansull: Comment:
  // If GT_JTRUE, and true part use jumpDestValueGen.

src/jit/assertionprop.cpp
```diff
@@ -4446,47 +4469,69 @@ ASSERT_TP* Compiler::optComputeAssertionGen()
...
+
+                    index = optFindComplementary(index);
+                    if (index != NO_ASSERTION_INDEX)
                     {
-                        // If GT_JTRUE, and true part use jumpDestValueGen.
-                        ASSERT_TP& gen = (i == 0 && tree->OperGet() == GT_JTRUE) ? jumpDestValueGen : valueGen;
-                        optImpliedAssertions(assertionIndex[i], gen);
-                        BitVecOps::AddElemD(apTraits, gen, assertionIndex[i] - 1);
+                        optImpliedAssertions(index, valueGen);
+                        BitVecOps::AddElemD(apTraits, valueGen, index - 1);
```
briansull: Comment:
  // If GT_JTRUE, and false and we have a complementary assertion available use valueGen.

src/jit/assertionprop.cpp
```diff
@@ -4446,47 +4469,69 @@ ASSERT_TP* Compiler::optComputeAssertionGen()
 
             for (GenTreePtr tree = stmt->gtStmt.gtStmtList; tree; tree = tree->gtNext)
             {
-                // Store whatever we have accumulated into jumpDest edge's valueGen.
                 if (tree->gtOper == GT_JTRUE)
                 {
-                    BitVecOps::Assign(apTraits, jumpDestValueGen, valueGen);
+                    assert((tree->gtNext == nullptr) && (stmt->gtNext == nullptr));
+                    jtrue = tree;
+                    break;
```
briansull: Comment:
// A GT_TRUE is always the last node in a tree, so we can break here

src/jit/assertionprop.cpp
```diff
@@ -5070,6 +5115,12 @@ void Compiler::optAssertionPropMain()
                                                        // and thus we must morph, set order, re-link
             for (GenTreePtr tree = stmt->gtStmt.gtStmtList; tree; tree = tree->gtNext)
             {
+                if (tree->OperIs(GT_JTRUE))
+                {
```
briansull: Comment:
// A GT_TRUE is always the last node in a tree, so we can break here

src/jit/assertionprop.cpp
```diff
@@ -4446,47 +4469,69 @@ ASSERT_TP* Compiler::optComputeAssertionGen()
...
-                for (unsigned i = 0; i < 2; ++i)
+            AssertionIndex index = jtrue->GetAssertion();
+            if ((index & OAE_INDEX_MASK) != NO_ASSERTION_INDEX)
+            {
+                if ((index & OAE_NEXT_EDGE) != 0)
+                {
+                    index &= OAE_INDEX_MASK;
+                    // Currently OAE_NEXT_EDGE is only used with OAK_NO_THROW assertions
+                    assert(optGetAssertion(index)->assertionKind == OAK_NO_THROW);
+                    // Don't bother with implied/complementary assertions, there aren't any for OAK_NO_THROW
```
briansull: // Don't bother with implied assertions, there aren't any for OAK_NO_THROW

(The assertion stored here when OAE_NEXT_EDGE is set is essentially a complementary assertion)

briansull:
We should also confirm that there are no Asm diffs (or just the small set of expected Asm diffs) with this change.

src/jit/compiler.h
```diff
@@ -5873,6 +5873,13 @@ class Compiler
 
     typedef unsigned short AssertionIndex;
 
+    enum optAssertionEdge : AssertionIndex
```
JosephTremoulet: Would be nice to have a comment here indicating that the assertion index on a JTRUE is the bitwise-or of the appropriate optAssertionEdge and its normal index.  Nice to have a comment about that over on the definition of the assertion field/accessor, too.

src/jit/assertionprop.cpp
```diff
@@ -4446,47 +4469,69 @@ ASSERT_TP* Compiler::optComputeAssertionGen()
...
-                                                    (tree->OperGet() == GT_JTRUE)
-                                                        ? optFindComplementary((AssertionIndex)tree->GetAssertion())
-                                                        : 0};
+        if (jtrue != nullptr)
+        {
+            // Copy whatever we have accumulated into jumpDest edge's valueGen.
+            ASSERT_TP jumpDestValueGen = BitVecOps::MakeCopy(apTraits, valueGen);
 
-                for (unsigned i = 0; i < 2; ++i)
+            AssertionIndex index = jtrue->GetAssertion();
```
JosephTremoulet: Wouldn't mind a comment here mentioning that the index stored on a `GT_JTRUE` is `index | optAssertionEdge`.  (Actually if I had my druthers there'd be a layer of abstraction here like separate accessors for a node's assertion index and its edge kind that do the masking [or perhaps the plain `GetAssertion()` accessor would assert that it's not called on a jtrue and that the edge bits aren't set, rather than masking, and callers would use a different `GetJTrueAssertion()` to get the masking], but I don't feel too strongly about it)

src/jit/valuenum.cpp
```diff
@@ -3163,6 +3163,51 @@ void ValueNumStore::GetConstantBoundInfo(ValueNum vn, ConstantBoundInfo* info)
...
+//    expression such as "(uint)i < (uint)a.len" that imply that the
+//    array index is valid (0 <= i && i < a.len).
+//
+// Arguments:
+//    vn - Value number to query
+//    info - Pointer to a ArrLenUnsignedBoundInfo object to return
+//           information about the expression. Not populated if the
+//           vn expression isn't suitable (e.g. i <= a.len)
+
+bool ValueNumStore::IsVNArrLenUnsignedBound(ValueNum vn, ArrLenUnsignedBoundInfo* info)
```
JosephTremoulet: It's a bit odd that `IsVNArrLenUnsignedBound` takes an out `info` parameter but its siblings have seprate `IsFoo` and `GetFooInfo` methods.  If I'm following correctly, doing it this way makes more sense for this case because then you have the `ArrLenUnsignedBoundInfo` available in `optCreateJTrueBoundAssertion` and can avoid making the useless complementary assertion.  Would be nice to make sure that's written down somewhere, in the commit message if not in a comment here.

danmosemsft:
@dotnet-bot test centos7.1 x64 debug build and test (https://github.com/dotnet/coreclr/issues/8321)

src/jit/assertionprop.cpp
mikedn: I don't like this bit wizardry myself but it keeps this change smaller. I experimented with something like `struct AssertionData { unsigned short edge : 1; unsigned short index : 15 }` and it looks much better but there's more code to change to get there. I'm more than happy to make such changes but it feels like an improvement for a subsequent PR related to those complementary assertions.

src/jit/valuenum.cpp
mikedn: I went back and forth on this one partly because of this inconsistency. An early version simply returned just `bool` like the others but that isn't enough to create the assertion. Another version returned `VNFuncApp` which the assertion creation code had to parse but it didn't feel quite right to have non VN code do that. So I ended up with a compromise and combined `Is` and `GetInfo` functions into one.

mikedn:
> You've gone to a fair bit of trouble here to avoid creating the useless complementary out-of-bounds assertions. I'd imagine that the out-of-bounds complementary assertions for the other non-constant array bound assertions aren't helping us much either, right? Do you think it would make sense (now or later) to change the way those other assertions are generated so they look more like the new case and avoid the out-of-bounds complement?

In short the answer is yes, I think that we should get rid of at least some complementary assertions. `i < a.len` is clearly useful to have but what good is `i >= a.len`? I suppose it could help eliminate a subsequent redundant `if (i >= a.len)` but it looks like the JIT doesn't do this kind of stuff so such an assertion just waste space in the assertion table.

But it's actually a bit complicated. I took out one of these complementary assertions and to my surprise a few additional range checks were eliminated and it wasn't because the assertion table was filling up. It seems that these complementary assertions can sometime result in rangecheck producing a `<a.len, 0>` range. This range is then merged with `<0, 0>` and produces `<a.len, 0>` and that results in a range check not being eliminated. But of course, all these ranges are actually `<0, 0>` because `a.len` is always >= 0.

mikedn:
There are a few issues I noticed while working on this that may be worth fixing in the future:
* `AssertionDsc::Equals` doesn't handle `OAK_NO_THROW` properly and we end up adding duplicate assertions. As a result we fail to eliminate a couple of range checks in 2 corelib methods because the assertion table fills up.
* The complementary assertion issue and its impact on rangecheck described above.
* The use of `op1.vn` for `OAK_NO_THROW` assertions. Doesn't seem to cause problems as it is but it's confusing.
* Some useless bitvec allocation/copying. I fixed some but there are more issues. In particular there's `EndMerge` that allocates a bitvec every time it is called.

JosephTremoulet:
@mikedn thanks for the analysis, I've created #9936, #9937, and #9938 to track three of the issues you called out (and I agree that use of `op1.vn` for `OAK_NO_THROW` should be cleaned up but didn't think it needed a tracking issue).  Please update if I've missed the mark in any of them

src/jit/assertionprop.cpp
JosephTremoulet: Makes sense.  Seeing something like that in a subsequent PR related to the complementary assertions would be great.

mikedn:
@JosephTremoulet Those issue look good, thank you for creating them

briansull:
LGTM

PR #9774, opened 2017-02-24
coreRT fat pointer hidden argument fix for calls with retBufArg

sandreenko:
in case of 
`class A<T>
{
	BigStruct foo(int arg1);
}`

we expect 
`void foo(this, retBufPointer, hiddenDictionary, arg1)` for amd64 and arm.
`void foo(this, retBufPointer, arg1,  hiddenDictionary)` for x86.
Fix dotnet/corert/issues/2751.

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -24539,7 +24539,7 @@ private:
         void AddArgumentToTail(GenTreeArgList* argList, GenTreePtr hiddenArgument)
         {
             GenTreeArgList* iterator = argList;
-            while (iterator->Rest())
+            while (iterator->Rest() != nullptr)
             {
                 iterator = iterator->Rest();
             }
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -24469,7 +24469,7 @@ private:
         //    fixedFptrAddress - pointer to the tuple <methodPointer, instantiationArgumentPointer>
         //
         // Return Value:
-        //    loaded hidden argument.
+        //    generic context hidden argument.
         GenTreePtr GetHiddenArgument(GenTreePtr fixedFptrAddress)
         {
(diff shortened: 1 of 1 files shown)


sandreenko:
PTAL @BruceForstall 
cc @MichalStrehovsky  @dotnet/jit-contrib

sandreenko:
WIP. I need to check platform and var arg cookie position.

sandreenko:
@BruceForstall please look again.

src/jit/flowgraph.cpp
```diff
@@ -24495,13 +24495,58 @@ class FatCalliTransformer
...
+        //
+        // Arguments:
+        //    argList - fat call node
+        //    hiddenArgument - generic context hidden argument
+        //
+        void AddArgumentToTail(GenTreeArgList* argList, GenTreePtr hiddenArgument)
+        {
+            GenTreeArgList* iterator = argList;
+            while (iterator->Rest())
+            {
```
BruceForstall: nit:
```
            while (iterator->Rest() != nullptr)
```

src/jit/flowgraph.cpp
```diff
@@ -24495,13 +24495,58 @@ class FatCalliTransformer
...
+                newArgs                   = compiler->gtNewListNode(retBuffer, newArgs);
+            }
+            else
+            {
+                newArgs = compiler->gtNewListNode(hiddenArgument, oldArgs);
+            }
+#else
+            newArgs = oldArgs;
+            AddArgumentToTail(newArgs, hiddenArgument);
+#endif
```
BruceForstall: In this case (x86), is it possible to have a var args cookie? If so, the var args cookie comes last, after the generics context. (I can't remember if you can have a generic function that is also var args)

sandreenko: No, we don't have such functions. There are several asserts that check it.

BruceForstall:
LGTM, assuming the answer to my var args question is that it's not a possibility for this case.

PR #9775, opened 2017-02-24
Add Throughput Perf testing

adiaaida:
This change adds scripting to run perf testing in the lab. We perform the
testing by running crossgen on a list of dlls. It then writes the data to a
csv file which is passed to other processing scripts.

### DIFF
--- a/perf.groovy
+++ b/perf.groovy
@@ -28,8 +28,8 @@ def static getOSGroup(def os) {
 [true, false].each { isPR ->
     ['Windows_NT'].each { os ->
 		['x64', 'x86', 'x86jit32'].each { arch ->
-            architecture = arch
-            testEnv = ''
+            def architecture = arch
+            def testEnv = ''
 
             if (arch == 'x86jit32')
             {
@@ -69,13 +69,13 @@ def static getOSGroup(def os) {
 					//Do this here to remove the origin but at the front of the branch name as this is a problem for BenchView
 					//we have to do it all as one statement because cmd is called each time and we lose the set environment variable
 					batchFile("if [%GIT_BRANCH:~0,7%] == [origin/] (set GIT_BRANCH_WITHOUT_ORIGIN=%GIT_BRANCH:origin/=%) else (set GIT_BRANCH_WITHOUT_ORIGIN=%GIT_BRANCH%)\n" +
-					"py \"%WORKSPACE%\\Microsoft.BenchView.JSONFormat\\tools\\submission-metadata.py\" --name " + "\"" + benchViewName + "\"" + " --user " + "\"dotnet-bot@microsoft.com\"\n" +
-					"py \"%WORKSPACE%\\Microsoft.BenchView.JSONFormat\\tools\\build.py\" git --branch %GIT_BRANCH_WITHOUT_ORIGIN% --type " + runType)
+					"py \"%WORKSPACE%\\Microsoft.BenchView.JSONFormat\\tools\\submission-metadata.py\" --name \"${benchViewName}\" --user \"dotnet-bot@microsoft.com\"\n" +
+					"py \"%WORKSPACE%\\Microsoft.BenchView.JSONFormat\\tools\\build.py\" git --branch %GIT_BRANCH_WITHOUT_ORIGIN% --type ${runType}")
 					batchFile("py \"%WORKSPACE%\\Microsoft.BenchView.JSONFormat\\tools\\machinedata.py\"")
 					batchFile("set __TestIntermediateDir=int&&build.cmd ${configuration} ${architecture}")
 					batchFile("tests\\runtest.cmd ${configuration} ${architecture} GenerateLayoutOnly")
-					batchFile("tests\\scripts\\run-xunit-perf.cmd -arch ${arch} -configuration ${configuration} ${testEnv} -testBinLoc bin\\tests\\Windows_NT.${architecture}.${configuration}\\performance\\perflab\\Perflab -library -uploadToBenchview \"%WORKSPACE%\\Microsoft.Benchview.JSONFormat\\tools\" -runtype ${runType}")
-					batchFile("tests\\scripts\\run-xunit-perf.cmd -arch ${arch} -configuration ${configuration} ${testEnv} -testBinLoc bin\\tests\\Windows_NT.${architecture}.${configuration}\\Jit\\Performance\\CodeQuality -uploadToBenchview \"%WORKSPACE%\\Microsoft.Benchview.JSONFormat\\tools\" -runtype ${runType}")
+					batchFile("tests\\scripts\\run-xunit-perf.cmd -arch ${arch} -configuration ${configuration} ${testEnv} -testBinLoc bin\\tests\\${os}.${architecture}.${configuration}\\performance\\perflab\\Perflab -library -uploadToBenchview \"%WORKSPACE%\\Microsoft.Benchview.JSONFormat\\tools\" -runtype ${runType}")
+					batchFile("tests\\scripts\\run-xunit-perf.cmd -arch ${arch} -configuration ${configuration} ${testEnv} -testBinLoc bin\\tests\\${os}.${architecture}.${configuration}\\Jit\\Performance\\CodeQuality -uploadToBenchview \"%WORKSPACE%\\Microsoft.Benchview.JSONFormat\\tools\" -runtype ${runType}")
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


adiaaida:
@dotnet-bot test ci please

adiaaida:
@DrewScoggins PTAL
/cc: @dotnet/jit-contrib

tests/scripts/run-throughput-perf.py
```diff
@@ -0,0 +1,361 @@
...
+        "Microsoft.CSharp",
+        "System",
+        "Microsoft.CodeAnalysis.VisualBasic",
+        "System.Private.DataContractSerialization",
+        "System.Core",
+        "System.Xml",
+        "Microsoft.CodeAnalysis.CSharp",
+        "Microsoft.CodeAnalysis",
+        "System.Linq.Parallel",
+        "System.Private.Xml"
```
briansull: Is the ordering in this list important? If so explain that in a comment.

adiaaida: It is not

adiaaida: Though perhaps I should add a comment as to why these assemblies were chosen, which is that they take at least 1 second or longer to crossgen.

DrewScoggins:
I'll take a look at this, this weekend.

perf.groovy
```diff
@@ -104,6 +104,80 @@ def static getOSGroup(def os) {
...
+					batchFile("C:\\Tools\\nuget.exe install Microsoft.BenchView.JSONFormat -Source http://benchviewtestfeed.azurewebsites.net/nuget -OutputDirectory \"%WORKSPACE%\" -Prerelease -ExcludeVersion")
+					batchFile("C:\\Tools\\nuget.exe install Microsoft.BenchView.ThroughputBenchmarks.${architecture}.${os} -Source https://dotnet.myget.org/F/dotnet-core -OutputDirectory \"%WORKSPACE%\" -Prerelease -ExcludeVersion")
+					//Do this here to remove the origin but at the front of the branch name as this is a problem for BenchView
+					//we have to do it all as one statement because cmd is called each time and we lose the set environment variable
+					batchFile("if [%GIT_BRANCH:~0,7%] == [origin/] (set GIT_BRANCH_WITHOUT_ORIGIN=%GIT_BRANCH:origin/=%) else (set GIT_BRANCH_WITHOUT_ORIGIN=%GIT_BRANCH%)\n" +
+					"py \"%WORKSPACE%\\Microsoft.BenchView.JSONFormat\\tools\\submission-metadata.py\" --name " + "\"" + benchViewName + "\"" + " --user " + "\"dotnet-bot@microsoft.com\"\n" +
+					"py \"%WORKSPACE%\\Microsoft.BenchView.JSONFormat\\tools\\build.py\" git --branch %GIT_BRANCH_WITHOUT_ORIGIN% --type " + runType)
```
DrewScoggins: nit:  Can you change this to use string interpolation.

perf.groovy
```diff
@@ -104,6 +104,80 @@ def static getOSGroup(def os) {
...
+			// Save machinedata.json to /artifact/bin/ Jenkins dir
+			def archiveSettings = new ArchivalSettings()
+			archiveSettings.addFiles('throughput-*.csv')
+			Utilities.addArchival(newJob, archiveSettings)
+
+			Utilities.standardJobSetup(newJob, project, isPR, "*/${branch}")
+
+			if (isPR) {
+				TriggerBuilder builder = TriggerBuilder.triggerOnPullRequest()
+				builder.setGithubContext("${os} ${arch} CoreCLR Perf Tests")
```
DrewScoggins: nit: Mention throughput here.

tests/scripts/run-throughput-perf.py
```diff
@@ -0,0 +1,361 @@
...
+            dll_path
+            ]
+
+    log(" ".join(run_args))
+
+    # Time.clock() returns seconds, with a resolution of 0.4 microseconds, so multiply by the multiplier to get milliseconds
+    multiplier = 1000
+
+    for iteration in range(0,iterations):
+        proc = subprocess.Popen(run_args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
```
DrewScoggins: Reading through some of the Python docs on Popen they recommend not using shell=True, for security reasons.  Since we are not calling any shell specific things on Windows, and I don't believe that we need environment variable expansion we should be able to not do this.

tests/scripts/run-throughput-perf.py
```diff
@@ -0,0 +1,361 @@
...
+
+        start_time = time.clock()
+        (out, err) = proc.communicate()
+        end_time = time.clock()
+
+        if proc.returncode == 0:
+            # Calculate the runtime
+            runtime = end_time * multiplier - start_time * multiplier
```
DrewScoggins: nit:  Would you mind making this (end_time - start_time) * multiplier?  Seems easier to read.

tests/scripts/run-throughput-perf.py
```diff
@@ -0,0 +1,361 @@
...
+    os.chdir(sandbox_path)
+
+    # Set up paths
+    test_root = os.path.join(clr_root, 'bin', 'Tests', operating_system + '.' + arch + '.' + build_type)
+    core_root_path = os.path.join(test_root, 'Tests', 'Core_Root')
+
+    crossgen_path = os.path.join(core_root_path,'crossgen.exe')
+    jit_path = os.path.join(core_root_path, jit)
+
+    runtimes = {}
```
DrewScoggins: nit: Could we change this to something like elapsedTime?  Runtimes is kind of an overloaded term,

tests/scripts/run-throughput-perf.py
```diff
@@ -0,0 +1,361 @@
...
+            runtimes[dll_name] = dll_runtimes
+
+    # Upload the data
+    if not benchview_path is None:
+        for dll_name, dll_runtimes in runtimes.items():
+            # Generate the csv file
+            csv_file_name = generateCSV(dll_name, dll_runtimes)
+
+            # For each benchmark, call measurement.py
+            measurement_args = ["py",
```
DrewScoggins: There is currently a bug in the display part of Benchview around the geomean.  If you do not upload all of the data in a subtree at once we will not correctly calculate the geomean, and since we do this at insertion time and not display time you get the wrong geomean.  This should be easy to fix, by just moving the submission.json creation out of the loop along with the upload.

tests/scripts/run-throughput-perf.py
```diff
@@ -0,0 +1,361 @@
...
+
+    arch = next((a for a in valid_archs if a.lower() == arch.lower()), arch)
+    build_type = next((b for b in valid_build_types if b.lower() == build_type.lower()), build_type)
+
+    validate_arg(arch, lambda item: item in valid_archs)
+    validate_arg(build_type, lambda item: item in valid_build_types)
+    validate_arg(run_type, lambda item: item in valid_run_types)
+    validate_arg(operating_system, lambda item: item in valid_os)
+
+    if clr_root is None:
```
DrewScoggins: I think I would like if if we just gave an error and failed here.  Since this will be running in automation it seems better to explicitly fail if we are not passing the right info rather then trying to intuit what the right place is.

tests/scripts/run-throughput-perf.py
```diff
@@ -0,0 +1,361 @@
...
+        clr_root = nth_dirname(os.path.abspath(sys.argv[0]), 3)
+    else:
+        clr_root = os.path.normpath(clr_root)
+        validate_arg(clr_root, lambda item: os.path.isdir(clr_root))
+
+    if assembly_root is None:
```
DrewScoggins: Same comment as above, but seems worse because this could allow us to accidentally start timing dlls that are not the same version since we are not using the nuget package version.

DrewScoggins:
As a general comment do we have any kind of style guide or coding guidelines for Python that gets checked in?  I don't really know what the best practices are for Python as far as casing and all that goes, but was curious if we have any guide here. @jorive @dpodder @brianrob

tests/scripts/run-throughput-perf.py
adiaaida: I'm a little confused by what you're asking for here. In run-xunit-perf.cmd, we call measurement.py once for each of the benchmarks (from their individual xml files). Here, we generate the csv, which will have all of the elapsed times in it, and call measurement.py on that csv for each benchmark (not multiple times per benchmark). I feel like these are doing the same thing. So are they both wrong? Should I generate only one csv file, which would be a mash up of each of the benchmarks?

dpodder:
@DrewScoggins I'm not sure whether CoreCLR or dotnet in general have adopted any python style guidelines. If there aren't any, I would recommend standard linters like `pylint` or `pep8`--they will highlight places where code diverges from broadly-adopted python conventions (including things like casing, spacing & indentation, etc.)

pylint can be a little too nitpicky, so I don't think there's value in following 100% of the default ruleset ("function too long" and "class: too few public members" are two warnings I've found over-applied), but it's a great start.

tests/scripts/run-throughput-perf.py
dpodder: @adiaaida The CSV generation and the calls to measurement.py are correct as written. The issue with BenchView is that it incorrectly assumes all related data gets uploaded in a single file (a single submission.json).

measurement.py only acts as a converter that reads in various data formats and outputs a "normalized" measurement.json. Multiple "measurement.json" files can get accumulated together and batched into a single submission object (submission.json). The suggestion is to move just the calls to `submission.py` and `upload.py` outside of the for loop. You would just need to add a switch to measurement.py to change the output file names (so that the measurement.jsons don't clobber one another), and accumulate them for passing to submission.py all at once.

CC @jorive as well for visibility... this really needs to get fixed on BenchView soon, IMO. :smile:

adiaaida: Sorry it isn't clear here. They are outside the loop, as python is whitespace sensitive. They are one indentation less indented than the loop (but inside the if not benchview_path is None)

dpodder: D'oh, you're completely right. My bad for missing that (the inline GitHub comments don't exactly help with the scope readability either). You shouldn't have to worry about the aforementioned bug with the code as written.

tests/scripts/run-throughput-perf.py
```diff
@@ -0,0 +1,361 @@
+#!/usr/bin/env python
```
jorive: Is this meant to work with both python 2 and 3? Has it been tested?

tests/scripts/run-throughput-perf.py
```diff
@@ -0,0 +1,361 @@
...
+    """
+
+    csv_file_name = "throughput_%s.csv" % (dll_name)
+    csv_file_path = os.path.join(os.getcwd(), csv_file_name)
+    output_file = open(csv_file_path, "w")
+
+    for iteration in dll_runtimes:
+        output_file.write("%s,%s,%s,%s\n" % ("default", "coreclr-crossgen-tp", dll_name, iteration))
+
+    output_file.close()
```
jorive: Can writing to the csv be done using the "import csv" instead?

jorive: Can this be changed to a with block:
with open(csv_file_path) as csvfile:
    ...

tests/scripts/run-throughput-perf.py
adiaaida: This only works with python 3 right now, due to the use of stat.

jashook: Coreclr has a python 2 dependency. None of the scripts in the repo to my knowledge are compatible with python 3.

tests/scripts/run-throughput-perf.py
adiaaida: done

tests/scripts/run-throughput-perf.py
adiaaida: done

tests/scripts/run-throughput-perf.py
adiaaida: done

tests/scripts/run-throughput-perf.py
adiaaida: done

tests/scripts/run-throughput-perf.py
adiaaida: done

tests/scripts/run-throughput-perf.py
adiaaida: As written, this works with both python 2 and 3, except for the calls to measurement, submission, and upload.py, which have a python 3 dependency.

adiaaida: done.

adiaaida: done

jorive: The csv package works in both python2 and python3. I would rather see a csv writer being used instead of manually generating the string.

adiaaida:
@dotnet-bot test ci please

perf.groovy
adiaaida: I haven't addressed this yet, but would you like me to address it for the regular perf tests too (where I grabbed this code from)?

tests/scripts/run-throughput-perf.py
adiaaida: Sorry, that comment was supposed to go to your question above (and has been deleted here and moved above). I've changed this to use the csv package.

perf.groovy
DrewScoggins: Yeah, I know we have it there as well.  If you are making the change here go ahead and do it in the other bit of code.

tests/scripts/run-throughput-perf.py
```diff
@@ -0,0 +1,362 @@
...
+                os.path.join(benchview_path, "submission.py"),
+                "measurement.json",
+                "--build",
+                os.path.join(clr_root, "build.json"),
+                "--machine-data",
+                os.path.join(clr_root, "machinedata.json"),
+                "--metadata",
+                os.path.join(clr_root, "submission-metadata.json"),
+                "--group",
+                "CoreCLR",
```
DrewScoggins: Do we want to have this uploaded as part of another group?  I am not a Benchview expert, but it looks like if we use the same group we will end up putting extra builds on the trendline for CoreCLR that will only have these Throughput tests.  @jorive @dpodder

DrewScoggins:
@jashook OK that sounds good.

tests/scripts/run-throughput-perf.py
DrewScoggins: Currently, all of our lab machines have both Python 2.7 and Python 3.5 installed.  2.7 is for the CoreCLR build while Python 3.5 is for the Benchview tools as @adiaaida mentioned above.

perf.groovy
adiaaida: Done

DrewScoggins:
test ci please

DrewScoggins:
I just kicked off a run through the CI to make sure that we are working end to end.

adiaaida:
@DrewScoggins thanks for that. To my eye, they look the same pre-string interpolation and post.

adiaaida:
@DrewScoggins argument for benchview_path was wrong (used benchview-path). Fixing and will rerun the testing.

adiaaida:
@dotnet-bot test ci please

adiaaida:
@dotnet-bot test ci please

adiaaida:
@dotnet-bot test ci please

adiaaida:
@dotnet-bot test ci please

adiaaida:
@DrewScoggins I've looked locally, and I'm creating a submission.json, so I'm confused why upload.py says that it doesn't exist. I hesitate to call upload.py locally to test, but can if I need to. Do you have any idea why this is failing:

https://ci2.dot.net/job/dotnet_coreclr/job/perf/job/master/job/GenPRTest/job/perf_throughput_perflab_Windows_NT_x64_prtest/6/console

adiaaida:
@dotnet-bot test ci please

DrewScoggins:
@adiaaida Just from looking at the output I wonder if what is happening is that we are launching the upload script in a location that does not have the submission.json in the same spot.

adiaaida:
Sorry, I figured it out. I forgot () after proc.communicate for both the call to submission and update.

tests/scripts/run-throughput-perf.py
```diff
@@ -0,0 +1,362 @@
...
+                        os.path.join(benchview_path, "measurement.py"),
+                        "csv",
+                        os.path.join(os.getcwd(), csv_file_name),
+                        "--metric",
+                        "execution_time",
+                        "--unit",
+                        "milliseconds",
+                        "--better",
+                        "desc",
+                        "--append"]
```
DrewScoggins: I just noticed this but you either need to add the --drop-first-value flag (check the docs for exact name) or not write out the first iteration.  The first one will be significantly slower as a lot of code will be cold.  I don't have a preference on how you want to do it, but you will need to fix that.

adiaaida: Should I also up the number of iterations?

chrisaut:
Do the perf run results get publish publically somewhere? Would be good to produce a graph or at least raw numbers to see improvements/regressions over time.

tests/scripts/run-throughput-perf.py
DrewScoggins: Yeah, probably adding one more iteration would be good, six, with the first being dropped, leaves us with five.  Which seems good to me.

adiaaida:
@dotnet-bot test ci please

DrewScoggins:
@chrisaut I know we are planning to make it public, but I do not know the details.  @dpodder or @valenis may know better than I.

adiaaida:
@dotnet-bot test ci please

adiaaida:
@dotnet-bot test ci please

adiaaida:
@dotnet-bot test ci please

adiaaida:
@dotnet-bot test ci please

adiaaida:
Throughput run was successful: https://ci2.dot.net/job/dotnet_coreclr/job/perf/job/master/job/GenPRTest/job/perf_throughput_perflab_Windows_NT_x64_prtest/2/console

adiaaida:
@DrewScoggins PTAL

DrewScoggins:
@adiaaida LGTM.  But let me add a few more machines to the lab before you merge this.

adiaaida:
Sounds good

adiaaida:
@DrewScoggins I saw that there are new machines in the lab (or at least machines I never noticed before). Am I good to go?

DrewScoggins:
Yeah go ahead.  This runs are pretty quick all things considered so we should have enough capacity for now.

PR #9778, opened 2017-02-24
Move S.Buffers and S.IO sources shared with corert into shared folder

alexperovich:
This is the first set of changes to move common sources shared with corert into a shared folder.


src/mscorlib/shared/System/IO/FileStream.Unix.cs
```diff
@@ -796,6 +796,22 @@ private Task WriteAsyncInternal(byte[] buffer, int offset, int count, Cancellati
             }
         }
 
+        /// <summary>Prevents other processes from reading from or writing to the FileStream.</summary>
+        /// <param name="position">The beginning of the range to lock.</param>
+        /// <param name="length">The range to be locked.</param>
+        private void LockInternal(long position, long length)
```
jkotas: Do we need this back? I though it got stripped because of it was not used anywhere.

src/mscorlib/shared/System/IO/Win32Marshal.cs
```diff
@@ -23,6 +23,16 @@ internal static Exception GetExceptionForLastWin32Error()
         }
 
         /// <summary>
+        ///     Converts, resetting it, the last Win32 error into a corresponding <see cref="Exception"/> object, optionally 
+        ///     including the specified path in the error message.
+        /// </summary>
+        internal static Exception GetExceptionForLastWin32Error(string path)
```
jkotas: Same here.

src/mscorlib/shared/System.Private.CoreLib.Shared.sources
```diff
@@ -0,0 +1,39 @@
...
+    <TargetsUnix Condition="'$(TargetsUnix)' != 'true'">false</TargetsUnix>
+    <TargetsOSX Condition="'$(TargetsOSX)' != 'true'">false</TargetsOSX>
+  </PropertyGroup>
+  <ItemGroup>
+    <MscorlibSources Include="$(MSBuildThisFileDirectory)\System\Buffers\*.cs"/>
+    <MscorlibSources Include="$(MSBuildThisFileDirectory)\System\IO\Error.cs"/>
+    <MscorlibSources Include="$(MSBuildThisFileDirectory)\System\IO\FileStream.cs"/>
+    <MscorlibSources Include="$(MSBuildThisFileDirectory)\System\IO\Path.cs"/>
+    <MscorlibSources Include="$(MSBuildThisFileDirectory)\System\IO\PathInternal.cs"/>
+    <!--<MscorlibSources Include="$(MSBuildThisFileDirectory)\Microsoft\Win32\SafeHandles\SafeThreadPoolIOHandle.cs"/>-->
```
jkotas: This file should be deleted instead since it is used in CoreCLR. Looks like it was added to CoreCLR by accident.

src/mscorlib/shared/System.Private.CoreLib.Shared.sources
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <TargetsWindows Condition="'$(TargetsWindows)' != 'true'">false</TargetsWindows>
+    <TargetsUnix Condition="'$(TargetsUnix)' != 'true'">false</TargetsUnix>
+    <TargetsOSX Condition="'$(TargetsOSX)' != 'true'">false</TargetsOSX>
+  </PropertyGroup>
+  <ItemGroup>
+    <MscorlibSources Include="$(MSBuildThisFileDirectory)\System\Buffers\*.cs"/>
```
jkotas: It would be nice to keep all files listed explicitly.

src/mscorlib/shared/System/IO/FileStream.Unix.cs
alexperovich: These are in the corert sources. I assume they are still used in corert. If they aren't I will remove.

src/mscorlib/shared/System.Private.CoreLib.Shared.sources
```diff
@@ -0,0 +1,39 @@
...
+    <MscorlibSources Include="$(MSBuildThisFileDirectory)\System\IO\Error.cs"/>
```
jkotas: Can we not call it mscorlibSources because of it is not mscorlib anymore? The regular `<Compile Include=... />` maybe best because of it is likely work the best with tooling.

alexperovich: I was unsure if the MscorlibSources item was treated specially. If that isn't the case then I will just make these Compile items.

src/mscorlib/shared/System/IO/FileStream.Unix.cs
jkotas: Actually, looks like incorrectly resolved conflict with https://github.com/dotnet/coreclr/pull/9769

alexperovich: Okay, will fix.

src/mscorlib/shared/System.Private.CoreLib.Shared.sources
alexperovich: done

src/mscorlib/shared/System/IO/PathInternal.Unix.cs
```diff
@@ -43,6 +43,22 @@ internal static bool IsDirectorySeparator(char c)
         }
 
         /// <summary>
+        /// Returns true if the path is too long
+        /// </summary>
```
jkotas: I do not think we need any of these changes. They are unused code that was deleted in the CoreCLR version. The cleanup should be propagated to the CoreRT copy instead.

src/mscorlib/shared/System/IO/Error.cs
```diff
@@ -16,6 +16,11 @@ namespace System.IO
     [Pure]
     internal static class Error
     {
+        internal static Exception GetStreamIsClosed()
```
jkotas: This one is good. It is not used in CoreCLR right now, but it will be later once we get more files in sync.

src/mscorlib/shared/System/IO/Win32Marshal.cs
```diff
@@ -99,6 +109,21 @@ internal static int MakeHRFromErrorCode(int errorCode)
         }
 
         /// <summary>
+        ///     Returns a Win32 error code for the specified HRESULT if it came from FACILITY_WIN32
+        ///     If not, returns the HRESULT unchanged
+        /// </summary>
+        internal static int TryMakeWin32ErrorCodeFromHR(int hr)
```
jkotas: Undo

src/mscorlib/shared/System/IO/PathInternal.cs
```diff
@@ -42,6 +42,63 @@ internal static void CheckInvalidPathChars(string path)
         }
 
         /// <summary>
+        /// Returns true if the given StringBuilder starts with the given value.
+        /// </summary>
+        /// <param name="value">The string to compare against the start of the StringBuilder.</param>
+        internal static bool StartsWithOrdinal(this StringBuilder builder, string value)
```
jkotas: The changes in this file should be also undone.

jkotas: The rest looks good.

alexperovich:

PR #9779, opened 2017-02-24
Fix incremental build on Unix

janvorli:
This change fixes issue with dactablerva.h being generated even if the
libcoreclr.so hasn't changed. Also, the dependency of the daccess.cpp
on this header was not honored by the build system and so even if the
dactablerva.h was regenerated with a different contents, the daccess.cpp
was not rebuilt.


janvorli:
@jkotas, @BruceForstall can you take a look? 
I've noticed there still remains one incremental issue in the build - the libcorejit.so is always re-linked because the clrjit.exports are always regenerated. The root of the problem seems to be the fact that the original source from which the clrjit.exports stems is created using the cmake FILE WRITE and APPEND commands. It seems that with these commands, there is no way to make them run only when the sources change. I will look into it later.

BruceForstall:
LGTM

janvorli:

PR #9783, opened 2017-02-25
Fix SpanBench

AndyAyersMS:
Make benchmark class public, add namespace. Tweak iteration counts
to very roughly 1 second per xunit-perf iteration.

### DIFF
--- a/tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
+++ b/tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
@@ -17,7 +17,7 @@ using Xunit;
 namespace Span
 {
 
-public class Tests
+public class SpanBench
 {
 
 #if DEBUG
@@ -31,7 +31,6 @@ public class Tests
 #endif
 
     const int Size = 1024;
-
 
     [MethodImpl(MethodImplOptions.NoInlining)]
     static void TestFillAllSpan(Span<byte> span)
@@ -452,7 +451,7 @@ public class Tests
 
     static Action LookupFunc(object o)
     {
-        TypeInfo t = typeof(Tests).GetTypeInfo();
+        TypeInfo t = typeof(SpanBench).GetTypeInfo();
         MethodInfo m = t.GetDeclaredMethod((string) o);
         return m.CreateDelegate(typeof(Action)) as Action;
     }
--- a/tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
+++ b/tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
@@ -17,7 +17,7 @@ using Xunit;
 namespace Span
 {
 
-public class Benchmark
+public class SpanBench
 {
 
 #if DEBUG
@@ -451,7 +451,7 @@ public class Benchmark
 
     static Action LookupFunc(object o)
     {
-        TypeInfo t = typeof(Benchmark).GetTypeInfo();
+        TypeInfo t = typeof(SpanBench).GetTypeInfo();
         MethodInfo m = t.GetDeclaredMethod((string) o);
         return m.CreateDelegate(typeof(Action)) as Action;
     }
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@joperezr @ahsonkhan PTAL
cc @dotnet/jit-contrib 

Resolves the ongoing mystery of why this benchmark hasn't been producing perf results: when I originally wrote it I did not make the benchmark class public.

Local testing using run-xunit-perf shows reasonable results now.

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
```diff
@@ -14,13 +14,20 @@
 [assembly: OptimizeForBenchmarks]
 [assembly: MeasureInstructionsRetired]
 
-class Tests
+namespace Span
+{
+
+public class Tests
```
ahsonkhan: Can you rename this from Tests to something like SpanBench (or more descriptive and unique). Makes it easier to search, especially in benchview.

AndyAyersMS: Sure

ahsonkhan:
Looks good to me :)

AndyAyersMS:
Meant to ping @jorive....

jorive:
@AndyAyersMS Changes look good. Optionally, you could change the benchmarks inner iterations to look more like this:
```
[Benchmark(InnerIterationCount = 10000)]
[MemberData(nameof(InputData))]
public static void TestBenchmark(string[] args)
{
    foreach (BenchmarkIteration iter in Benchmark.Iterations)
    {
        using (iter.StartMeasurement())
        {
            for (int i = 0; i < Benchmark.InnerIterationCount; ++i)
```
As-is, the results will display on BenchView like this: `Span \ SpanBench \ TestFillAllSpan`
I would suggest something more like: `Span \ Benchmarks \ TestFillAll` (remove the repetition)
What do you think?

AndyAyersMS:
The final `Span` suffix is there to distinguish the span cases from the corresponding array cases -- that is there is both a `TestFillAllSpan` and a `TestFillAllArray`.

This test can be run as a perf test two ways -- under xunit perf and also as a standalone exe. I generally haven't generally used inner iteration count for the tests I write to maintain symmetry with the exe-driven portion of the testing.

Happy to change the class name to `Benchmarks` if you think it is an improvement.

ahsonkhan:
@AndyAyersMS, I believe SpanBench is fine.

There is already a method: 
```C#
    static double Bench(Action f)
```

error CS0542: 'Bench': member names cannot be the same as their enclosing type [D:\j\workspace\debug_windows---17180f1b\tests\src\JIT\Performance\CodeQuality\Span\SpanBench.csproj]

ahsonkhan:
@AndyAyersMS, you probably don't want to name your class "Benchmark"

SpanBench.cs(203,45): error CS0117: 'Benchmark' does not contain a definition for 'Iterations' [D:\j\workspace\checked_windo---f6dc6fe4\tests\src\JIT\Performance\CodeQuality\Span\SpanBench.csproj]

It gets confused with Benchmark attribute for the performance tests.
```C#
foreach (var iteration in Benchmark.Iterations)
```

AndyAyersMS:
Odd that I don't see any of these errors building locally... anyways, back to SpanBench it is.

AndyAyersMS:
Arm emulator build failure, seems widespread since roughly midnight.
```
07:23:12 Copied 205 artifacts from "dotnet_coreclr » master » release_windows_nt_bld" build number 1215
07:23:12 ERROR: Unable to find project for artifact copy: dotnet_corefx/master/linuxarmemulator_softfp_cross_release
07:23:12 This may be due to incorrect project name or permission settings; see help for project name in job configuration
```
looks like this artifact comes from a corefx build, last good run has:
```
Copied 1 artifact from "dotnet_corefx » master » linuxarmemulator_softfp_cross_debug" build number 116
```
Root cause is likely this change which broke arm builds in CoreFX:
https://github.com/dotnet/corefx/pull/16378

AndyAyersMS:
@ahsonkhan I'll abandon this since your #9790 has incorporated these changes already.

hqueue:

ahsonkhan:
> I'll abandon this since your #9790 has incorporated these changes already.

Sounds good. Close this PR?

Issue #9707, opened 2017-02-21
Pre-check doesn't eliminate bounds check


benaadams:
I'm experimenting with improvements `List.Add` and noticed that bounds checks are not elimitated where the value is pre-checked e.g.
```csharp
public void Add(T item)
{
    var array = _items;
    var size = _size;
    _version++;
    if ((uint)size < (uint)array.Length)
    {
        _size = size + 1;
        array[size] = item;
    }
    else
    {
        AddWithResize(item);
    }
}
```
Generates
```asm
**************** Inline Tree
Inlines into 060034D4 List`1:Add(long):this
  [0 IL=0054 TR=000029 060034D5] [FAILED: noinline per IL/cached result] List`1:AddWithResize(long):this
Budget: initialTime=240, finalTime=240, initialBudget=2400, currentBudget=2400
Budget: initialSize=1499, finalSize=1499
; Assembly listing for method List`1:Add(long):this
; Emitting BLENDED_CODE for X64 CPU with SSE2
; optimized code
; rsp based frame
; fully interruptible
; Final local variable assignments
;
;  V00 this         [V00,T00] ( 12,  10.5)     ref  ->  rcx         this
;  V01 arg1         [V01,T02] (  4,   3  )    long  ->  rdx        
;  V02 loc0         [V02,T03] (  4,   3  )     ref  ->  rax        
;  V03 loc1         [V03,T01] (  7,   5  )     int  ->   r8        
;  V04 OutArgs      [V04    ] (  1,   1  )  lclBlk (32) [rsp+0x00]  
;
; Lcl frame size = 40

G_M46198_IG01:
       4883EC28             sub      rsp, 40
       90                   nop      

G_M46198_IG02:
       488B4108             mov      rax, gword ptr [rcx+8]
       448B4118             mov      r8d, dword ptr [rcx+24]
       FF411C               inc      dword ptr [rcx+28]
       44394008             cmp      dword ptr [rax+8], r8d  // check
       761C                 jbe      SHORT G_M46198_IG04     // jump alt path
       458D4801             lea      r9d, [r8+1]
       44894918             mov      dword ptr [rcx+24], r9d
       443B4008             cmp      r8d, dword ptr [rax+8]  // check
       731C                 jae      SHORT G_M46198_IG06     // jump bounds check fail (unreachable)
       4963C8               movsxd   rcx, r8d
       488954C810           mov      qword ptr [rax+8*rcx+16], rdx

G_M46198_IG03:
[...]

PR #9784, opened 2017-02-25
Skip some range checks

benaadams:
Requires https://github.com/dotnet/coreclr/pull/9773 to be effective; though it won't decrease perf prior

/cc @mikedn @stephentoub

### DIFF
--- a/src/mscorlib/src/System/Collections/Generic/List.cs
+++ b/src/mscorlib/src/System/Collections/Generic/List.cs
@@ -189,7 +189,6 @@ namespace System.Collections.Generic
         {
             get
             {
-                // Following trick can reduce the range check by one
                 if ((uint)index >= (uint)_size)
                 {
                     ThrowHelper.ThrowArgumentOutOfRange_IndexException();
@@ -247,7 +246,7 @@ namespace System.Collections.Generic
             var array = _items;
             var size = _size;
             _version++;
-            if ((uint)size < (uint)array.Length)
+            if (array.IndexInRange(size))
             {
                 _size = size + 1;
                 array[size] = item;
--- a/src/mscorlib/src/System/Runtime/CompilerServices/jithelpers.cs
+++ b/src/mscorlib/src/System/Runtime/CompilerServices/jithelpers.cs
@@ -230,5 +230,10 @@ namespace System.Runtime.CompilerServices
             typeof(ArrayPinningHelper).ToString(); // Type used by the actual method body
             throw new InvalidOperationException();
         }
+
+        // Allows automatic range checks to be skipped when they are manually checked for other reasons 
+        // https://github.com/dotnet/coreclr/pull/9773
+        static internal bool IndexInRange<T>(this T[] array, int index) => (uint)index < (uint)array.Length;
+        static internal bool IndexOutOfRange<T>(this T[] array, int index) => (uint)index >= (uint)array.Length;
     }
 }
--- a/src/mscorlib/shared/System/Buffers/ConfigurableArrayPool.cs
+++ b/src/mscorlib/shared/System/Buffers/ConfigurableArrayPool.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
 using System.Threading;
 
 namespace System.Buffers
@@ -77,7 +78,8 @@ namespace System.Buffers
             T[] buffer = null;
 
             int index = Utilities.SelectBucketIndex(minimumLength);
-            if (index < _buckets.Length)
+            var buckets = _buckets;
+            if (buckets.IndexInRange(index))
             {
                 // Search for an array starting at the 'index' bucket. If the bucket is empty, bump up to the
                 // next higher bucket and try that one, but only try at most a few buckets.
@@ -86,21 +88,21 @@ namespace System.Buffers
                 do
                 {
                     // Attempt to rent from the bucket.  If we get a buffer from it, return it.
-                    buffer = _buckets[i].Rent();
+                    buffer = buckets[i].Rent();
                     if (buffer != null)
                     {
                         if (log.IsEnabled())
                         {
-                            log.BufferRented(buffer.GetHashCode(), buffer.Length, Id, _buckets[i].Id);
+                            log.BufferRented(buffer.GetHashCode(), buffer.Length, Id, buckets[i].Id);
                         }
                         return buffer;
                     }
                 }
-                while (++i < _buckets.Length && i != index + MaxBucketsToTry);
+                while (buckets.IndexInRange(++i) && i != index + MaxBucketsToTry);
 
                 // The pool was exhausted for this buffer size.  Allocate a new buffer with a size corresponding
                 // to the appropriate bucket.
-                buffer = new T[_buckets[index]._bufferLength];
+                buffer = new T[buckets[index]._bufferLength];
             }
             else
             {
@@ -113,7 +115,7 @@ namespace System.Buffers
             {
                 int bufferId = buffer.GetHashCode(), bucketId = -1; // no bucket for an on-demand allocated buffer
                 log.BufferRented(bufferId, buffer.Length, Id, bucketId);
-                log.BufferAllocated(bufferId, buffer.Length, Id, bucketId, index >= _buckets.Length ?
+                log.BufferAllocated(bufferId, buffer.Length, Id, bucketId, index >= buckets.Length ?
                     ArrayPoolEventSource.BufferAllocatedReason.OverMaximumSize : ArrayPoolEventSource.BufferAllocatedReason.PoolExhausted);
             }
 
@@ -137,7 +139,8 @@ namespace System.Buffers
             int bucket = Utilities.SelectBucketIndex(array.Length);
 
             // If we can tell that the buffer was allocated, drop it. Otherwise, check if we have space in the pool
-            if (bucket < _buckets.Length)
+            var buckets = _buckets;
+            if (buckets.IndexInRange(bucket))
             {
                 // Clear the array if the user requests
                 if (clearArray)
@@ -148,7 +151,7 @@ namespace System.Buffers
                 // Return the buffer to its bucket.  In the future, we might consider having Return return false
                 // instead of dropping a bucket, in which case we could try to return to a lower-sized bucket,
                 // just as how in Rent we allow renting from a higher-sized bucket.
-                _buckets[bucket].Return(array);
+                buckets[bucket].Return(array);
             }
 
             // Log that the buffer was returned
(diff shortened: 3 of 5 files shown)


benaadams:
@dotnet-bot Test Ubuntu x64 Checked Build and Test

mikedn:
Hmm, are all these performance sensitive? Adding these casts all other the place is pretty ugly.

src/mscorlib/src/System/Threading/ThreadPool.cs
```diff
@@ -1387,7 +1387,7 @@ private static object[] ToObjectArray(IEnumerable<IThreadPoolWorkItem> workitems
             i = 0;
             foreach (IThreadPoolWorkItem item in workitems)
             {
-                if (i < result.Length) //just in case someone calls us while the queues are in motion
+                if ((uint)i < (uint)result.Length) //just in case someone calls us while the queues are in motion
```
mikedn: Eh, we do interface calls but we're trying to get rid of a range check, I doubt that this will matter. Granted, it doesn't hurt either, at least the code will be smaller.

src/mscorlib/corefx/System/Buffers/ConfigurableArrayPool.cs
```diff
@@ -77,7 +77,7 @@ public override T[] Rent(int minimumLength)
             T[] buffer = null;
 
             int index = Utilities.SelectBucketIndex(minimumLength);
-            if (index < _buckets.Length)
+            if ((uint)index < (uint)_buckets.Length)
```
mikedn: I'm not sure if this will work as expected due to `_buckets` being a field.

mikedn:
Maybe we should have
```C#
static bool RangeCheck<T>(this T[] array, int index) => (uint)index < (uint)array.Length;
```
Though going through inliner might be a problem...

src/mscorlib/src/System/Threading/ThreadPool.cs
stephentoub: It also won't matter in this particular case, as this code is not particularly performance-sensitive: it's only called from a debugger when it asks for the list of queued work items to show to the developer.

benaadams:
Would it work non-generic?

```csharp
static bool IndexInBounds(this Array array, int index) => (uint)index < (uint)array.Length;
static bool IndexOutOfBounds(this Array array, int index) => (uint)index >= (uint)array.Length;
```

benaadams:
```csharp
static bool OffsetInBounds(this Array array, int offset) => (uint)offset < (uint)array.Length;
static bool OffsetOutOfBounds(this Array array, int offset) => (uint)offset >= (uint)array.Length;
```
Might be more correct; even though every one calls it an index then gets confused why it starts at 0

mikedn:
> Would it work non-generic (for detection/inlining)?

It looks like the non-generic version won't work. The JIT only recognizes `ldlen` instructions, it doesn't deal with calls to `get_Length`.

At least inlining doesn't appear to cause problems, the generic versions work as expected.

benaadams:
`IndexInRange`/`IndexOutOfRange` seem to work best in context.

@mikedn something like that?

hqueue:

mikedn:
> something like that?

Yes. Though the whole thing feels rather unnatural IMO. Oh well, we'll see what other people say.

src/mscorlib/src/System/Runtime/CompilerServices/jithelpers.cs
```diff
@@ -230,5 +230,10 @@ static internal bool ByRefLessThan<T>(ref T refA, ref T refB)
             typeof(ArrayPinningHelper).ToString(); // Type used by the actual method body
             throw new InvalidOperationException();
         }
+
+        // Allows automatic range checks to be skipped when they are manually checked for other reasons 
+        // https://github.com/dotnet/coreclr/pull/9773
+        static internal bool IndexInRange<T>(this T[] array, int index) => (uint)index < (uint)array.Length;
```
mikedn: These aren't true JIT helpers so this is not a good place for them.

src/mscorlib/src/System/Collections/Generic/List.cs
```diff
@@ -247,7 +246,7 @@ public void Add(T item)
             var array = _items;
             var size = _size;
             _version++;
-            if ((uint)size < (uint)array.Length)
+            if (array.IndexInRange(size))
```
benaadams: @jkotas this internal api work or is it a bit ugly? (to avoid casting via `uint` everywhere)

jkotas: It looks a bit ugly and it should not be on JitHelpers helpers. We should design it as if it was meant to be public API.

I am not sure whether it is a good API to add. It is way more readable to just write `size < array.Length` than to call an API to do it in a tricky way. Can we fix the JIT to morph this code into unsigned comparison instead of doing it manually? No need to introduce a new API, and no need to tech everybody to modify their code to use the new API to get the benefit.

benaadams: The issue is negative numbers; what's the cost of `checked`? Could morph array access to cast to `uint` and catch the overflow? Or will that be the same as doing the bounds check?

mikedn: >  Can we fix the JIT to morph this code into unsigned comparison instead of doing it manually?

I suspect that it may be possible in certain situations but in the general case I'm afraid that the answer is no. Let's look at the example that started all this - `List<T>.Add(T)`:
```C#
if (size < array.Length)
{
    _size = size + 1;
    array[size] = item;
 }
 else
 {
     AddWithResize(item);
}
```
The JIT "sees" `size < array.Length` but this is not enough to eliminate the array range check because `size` might be negative. Nor can the JIT change `size < array.Length` to unsigned, doing so would change the meaning of the program. If `size` happens to be negative then the original code would throw an exception while the new code would execute the `AddWithResize` branch.

By using `(uint)size < (uint)array.Length` the developer is adding information that the JIT simply doesn't have and helps it eliminate the range check.

It may be possible for the JIT to recognize the more intuitive `(0 <= size && size < array.Length)` but it will have to work harder for that.

That said, I like clean code and I think that such tricks should only be used when really needed. And I've played enough with this kind of stuff to know that saving a few range checks here and there won't necessarily make a difference for real word applications.

mikedn: > what's the cost of checked? Could morph array access to cast to uint and catch the overflow? Or will that be the same as doing the bounds check?

Not sure what you have in mind. What's the cost of a bounds check anyway?

benaadams: > Not sure what you have in mind. 

If signed test was performed, the range check just has to test its >= 0, rather than comparing to length.

> What's the cost of a bounds check anyway?

The examples I was looking at the generally always ended up as a memory access; so probably dominated by L1 cache hit? (as already checked) 
```asm
jae      SHORT G_M46198_IG06 
```
Maybe if it kept the length in register from earlier lookup and test that would be most of the cost gone.

mikedn: > If signed test was performed, the range check just has to test its >= 0, rather than comparing to length.

Yes, I was thinking about that recently. I'm not sure if it's useful, see below.

> The examples I was looking at the generally always ended up as a memory access; so probably dominated by L1 cache hit? (as already checked). Maybe if it kept the length in register from earlier lookup and test that would be most of the cost gone.

Usually the CSE picks up the duplicate length load so you end up with a single memory access. Something like:
```asm
mov eax, dword ptr [rdx+8]
cmp ecx, eax
jge (throw ArgumentException)
cmp ecx, eax
jae (range_check_failed)
```
What do we gain by eliminating the range check? 

We get rid of a `cmp` and a `jcc` but these are cheap instructions. The `jcc` should be correctly predicted and they may end up being executed in parallel with other instructions, they're not part of a dependency chain.

If there are no other range checks in the method we also get rid of a call to the range check failed helper. Good for overall program size but 0 impact on performance.

We may free up a register because now we can put the memory load in the `cmp` like you show in your example. This is where changing the range check to `i >= 0` could be useful because we no longer have to keep the array length in a register (or reload it from memory). Nice but usually insignificant.

benaadams: In the List`1.Add(long) dasm I was looking at, it never put the length in a register https://github.com/dotnet/coreclr/issues/9707#issue-209303637

```asm
; Assembly listing for method List`1:Add(long):this
; optimized code
; rsp based frame
; fully interruptible
;
;  V00 this         [V00,T00] ( 12,  10.5)     ref  ->  rcx         this
;  V01 arg1         [V01,T02] (  4,   3  )    long  ->  rdx        
;  V02 loc0         [V02,T03] (  4,   3  )     ref  ->  rax        
;  V03 loc1         [V03,T01] (  7,   5  )     int  ->   r8        
;
; Lcl frame size = 40

G_M46198_IG01:
       90                   nop      

G_M46198_IG02:
       488B4108             mov      rax, gword ptr [rcx+8]
       448B4118             mov      r8d, dword ptr [rcx+24]
       44394008             cmp      dword ptr [rax+8], r8d  // check
       761C                 jbe      SHORT G_M46198_IG04     // jump alt path
       458D4801             lea      r9d, [r8+1]
       44894918             mov      dword ptr [rcx+24], r9d
       443B4008             cmp      r8d, dword ptr [rax+8]  // check
       731C                 jae      SHORT G_M46198_IG06     // jump bounds check fail (unreachable)
       4963C8               movsxd   rcx, r8d
       488954C810           mov      qword ptr [rax+8*rcx+16], rdx

G_M46198_IG03:
       C3                   ret      

G_M46198_IG04:

G_M46198_IG05:

G_M46198_IG06:
       E800000000           call     CORINFO_HELP_RNGCHKFAIL
       CC                   int3     

; Total bytes of code 69, prolog size 5 for method List`1:Add(long):this
```
Though it should get picked up by your change https://github.com/dotnet/coreclr/pull/9773, and its not too onerous to do unsigned casts in the rare cases its needed (or to actually use unsigned indexers in the first place) but it doesn't look as pretty as not using them.

mikedn: > In the List`1.Add(long) dasm I was looking at, it never put the length in a register

Hmm, in other cases I looked at this seems to work as expected. Anyway, I think that using this trick in `List<T>.Add` is good now that it inlines. Everyone using it will now pay the cost of increased code size so it make sense to try to minimize that code considering that `Add` is rather common.

benaadams: Might cause a register save and pop to stack in this example? In which case there wouldn't be an advantage to not just using memory?

benaadams:
Will close this as its a bit of a weird change.

@jkotas do you want an issue raised to capture the discussion on alternative detection/morph?

PR #9786, opened 2017-02-25
Optimize Buffer.MemoryCopy

vkvenkat:
This PR optimizes Buffer.MemoryCopy for lengths up to a threshold. Performance measurements were made on a Skylake Desktop - Intel(R) Core(TM)i7-6700K CPU @ 4.00 GHz with 32 GB RAM. We used Windows 10 Enterprise and Ubuntu 14.04. We used a micro-benchmark which exercises multiple buckets of pseudo-random copy sizes from 0 to 4 KB (details below). 

Performance improvements with the optimization:

- 2% to 73% for x64 Windows 
- -4% to 73% for x86 Windows 
- 0% to 70% for x64 Ubuntu

The below charts show the performance speedup:

image

image

The below table has absolute clocks in Ticks for million copy operations:

image

Here's the pseudo-code for the micro-benchmark:

    // 10 Buckets of copy sizes within specified ranges: [0,8], [9,16], [17,32],..., [2049,4096]
    for (int i = 0; i < MAX_BUCKETS; i++)
    {  
        Random rnd = new Random(1);  
        for (int j = 0; j < COPYLEN; j++)
        {
                // Randomize copy length within bounds
                copyLen[j] = (ushort) rnd.Next(0, 9); // Range changes with bucket
        }

        begin = DateTime.Now.Ticks;
    
        for (int k = 0; k < 1000000; k++)
        {
            for (count= 0; count < COPYLEN; count++) 
            {						
                Buffer.MemoryCopy(pSource, pTarget, 8, copyLen[count]); //Dest. max size changes with bucket
            }	
        }
    
        time = DateTime.Now.Ticks - begin; // Time reported for bucket
    }

### DIFF
--- a/src/mscorlib/src/System/Buffer.cs
+++ b/src/mscorlib/src/System/Buffer.cs
@@ -2,6 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#if AMD64 || (BIT32 && !ARM)
+#define HAS_CUSTOM_BLOCKS
+#endif
+
 namespace System
 {
     //Only contains static methods.  Does not require serialization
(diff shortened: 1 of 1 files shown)


vkvenkat:
@jkotas PTAL

src/mscorlib/src/System/Buffer.cs
```diff
@@ -256,325 +256,162 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
-                // We're not using i after this, so not needed
-                // i += 1;
-            }
-
             return;
 
-        PInvoke:
+#if AMD64 && !PLATFORM_UNIX             
+            FCopy:
+            _MemoryCopy(dest, src, len);
```
jkotas: This should go through PInvoke so that it does not starve GC for large copies.

src/mscorlib/src/System/Buffer.cs
```diff
@@ -256,325 +256,162 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
-                    *(long*)dest = *(long*)src;
-                    *(long*)(dest + 8) = *(long*)(src + 8);
+            const nuint CopyThreshold = 256;
+#endif // AMD64
+
+            // FCALL/PInvoke to the native version when the copy length exceeds the threshold or when the buffers are overlapping. This 
+            // check can produce false positives for lengths greater than Int32.MaxInt. But, this is fine because we want to FCALL/PInvoke 
+            // for large lengths anyways.
+            
+            if ((len > CopyThreshold) || ((nuint)dest < (nuint)srcEnd))
```
jkotas: The `(nuint)dest < (nuint)srcEnd` check does not look right. It means we will hit the slow path any time the destination is lower than the source.

jkotas: Why is it profitable to check for the CopyThreshold here, and not later? It adds overhead for small copies.

src/mscorlib/src/System/Buffer.cs
```diff
@@ -618,5 +455,13 @@ public static unsafe void MemoryCopy(void* source, void* destination, ulong dest
             Memmove((byte*)destination, (byte*)source, checked((uint)sourceBytesToCopy));
 #endif // BIT64
         }
+        
+#if BIT64        
+        [StructLayout(LayoutKind.Sequential, Size = 16)]
+        private struct Block16 { }
+
+        [StructLayout(LayoutKind.Sequential, Size = 64)]
+        private struct Block64 { }
```
jkotas: Does it work well / will it work well on ARM64?

jkotas:
cc @jamesqo

src/mscorlib/src/System/Buffer.cs
```diff
@@ -256,325 +256,162 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
-                    return;
-                case 17:
-#if BIT64
-                    *(long*)dest = *(long*)src;
-                    *(long*)(dest + 8) = *(long*)(src + 8);
+            const nuint CopyThreshold = 256;
+#endif // AMD64
+
+            // FCALL/PInvoke to the native version when the copy length exceeds the threshold or when the buffers are overlapping. This 
+            // check can produce false positives for lengths greater than Int32.MaxInt. But, this is fine because we want to FCALL/PInvoke
```
jamesqo: Nit: `int.MaxValue`

jamesqo: Nit: `int.MaxValue`

src/mscorlib/src/System/Buffer.cs
```diff
@@ -256,325 +256,162 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
+                goto PInvoke;
+#endif // AMD64 && !PLATFORM_UNIX                
+            }
+            
+            byte* destEnd = dest + len;            
+            if (len <= 16) goto MCPY02;
+            if (len > 64) goto MCPY05;
+            
+            MCPY00:
+            // len > 16 && len <= 64
```
jamesqo: Nit: Could be a `Debug.Assert` instead of a comment (same for other comments like this)

src/mscorlib/src/System/Buffer.cs
```diff
@@ -256,325 +256,162 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
+            MCPY03:
+            if ((len & 4) == 0) goto MCPY04;
+            // len >= 4 && len < 8
+            *(int*)dest = *(int*)src;
+            *(int*)(destEnd - 4) = *(int*)(srcEnd - 4);
+            return;
+            
+            MCPY04:
+            // len < 4
+            if ((len & len) == 0) return;
```
jamesqo: `len & len` is always going to be 0.

src/mscorlib/src/System/Buffer.cs
```diff
@@ -256,325 +256,162 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
-                    *(dest + 20) = *(src + 20);
-                    return;
-                case 22:
-#if BIT64
-                    *(long*)dest = *(long*)src;
-                    *(long*)(dest + 8) = *(long*)(src + 8);
+            return;
+            
+            MCPY02:
+            if ((len & 24) == 0) goto MCPY03;
```
jamesqo: Interesting that replacing the switch table with all of these branches results in better perf.

src/mscorlib/src/System/Buffer.cs
```diff
@@ -256,325 +256,162 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
-
-                // So the only cost is a bit of code size, which is made up for by the fact that
-                // we save on writes to dest/src.
-
+            *(int*)dest = *(int*)src;
+            *(int*)(dest + 4) = *(int*)(src + 4);
+            *(int*)(destEnd - 8) = *(int*)(srcEnd - 8);
+            *(int*)(destEnd - 4) = *(int*)(srcEnd - 4);
+#endif            
+            return;
```
jamesqo: There are now 2/4 writes being incurred here instead of 1/2 for len == 8. I scanned your test code and it seems to indicate that for random numbers in [0, 8] the change is a win, but have you tried testing consistently with len == 8?

src/mscorlib/src/System/Buffer.cs
```diff
@@ -256,325 +256,162 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
+            
+            MCPY03:
+            if ((len & 4) == 0) goto MCPY04;
+            // len >= 4 && len < 8
+            *(int*)dest = *(int*)src;
+            *(int*)(destEnd - 4) = *(int*)(srcEnd - 4);
+            return;
```
jamesqo: Same here, we're incurring an extra write for len == 4.

src/mscorlib/src/System/Buffer.cs
```diff
@@ -256,325 +256,162 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
+            
+            MCPY05:    
+            // len > 64 && len <= 256
+            nuint n = len >> 6;
+                
+            MCPY06:
 #if BIT64
-                *(long*)(dest + i) = *(long*)(src + i);
-                *(long*)(dest + i + 8) = *(long*)(src + i + 8);
+            *(Block64*)dest = *(Block64*)src;
```
jamesqo: SIMD is not enabled for ARM64 yet, I think. I think all of the `Block*` usages should be under <s>`AMD64`</s> actually, I believe RyuJIT is the default JIT on x86 now so `AMD64 || (BIT32 && !ARM)`? /cc @jkotas

src/mscorlib/src/System/Buffer.cs
```diff
@@ -256,325 +256,162 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
+            *(int*)(dest + 24) = *(int*)(src + 24);
+            *(int*)(dest + 28) = *(int*)(src + 28);
+            *(int*)(dest + 32) = *(int*)(src + 32);
+            *(int*)(dest + 36) = *(int*)(src + 36);
+            *(int*)(dest + 40) = *(int*)(src + 40);
+            *(int*)(dest + 44) = *(int*)(src + 44);
+            *(int*)(dest + 48) = *(int*)(src + 48);
+            *(int*)(dest + 52) = *(int*)(src + 52);
+            *(int*)(dest + 56) = *(int*)(src + 56);
+            *(int*)(dest + 60) = *(int*)(src + 60);
```
jamesqo: Is it beneficial to loop unroll this much? In my experiments last year I recall 64 being slightly worse than doing chunks of 48.

src/mscorlib/src/System/Buffer.cs
```diff
@@ -256,325 +256,162 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
-            while (counter <= end);
-
-            if ((len & 8) != 0)
-            {
+            dest += 64;
+            src += 64;
+            n--;
+            if (n != 0) goto MCPY06;
+            
+            len = len & 0x3f;
```
jamesqo: Nit: `len &= 0x3f`

jamesqo: Nit: Also might be better to just write `%= 64` to make it more understandable for others.

src/mscorlib/src/System/Buffer.cs
Anderman: The switch statement uses a table in memory. And memory is slow. Second it will polluted your memory data cache.

src/mscorlib/src/System/Buffer.cs
vkvenkat: Sure, will do. Buffer::BlockCopy (https://github.com/dotnet/coreclr/blob/da44463552689930c4d4ead5aa01a437fcbea622/src/vm/comutilnative.cpp#L1438) uses a similar FCALL to JIT_MemCpy as well.

src/mscorlib/src/System/Buffer.cs
vkvenkat: Had re-purposed the ```if ((nuint)dest - (nuint)src < len) goto PInvoke``` overlap check from the original. Shall think of an alternate for this.

src/mscorlib/src/System/Buffer.cs
vkvenkat: Shall special-case the usage of custom structs to AMD64 @jamesqo

src/mscorlib/src/System/Buffer.cs
jkotas: Yes, we have places like the use in Buffer::BlockCopy that can starve GC and lead to long GC pause times. We are trying to not introduce more of them, and fix the ones that are there.

src/mscorlib/src/System/Buffer.cs
vkvenkat: Seeing a 10% perf improvement with our version for a million copies with len == 8 on x64-Windows. Looks like eliminating the switch table lead to this, despite the extra write.

src/mscorlib/src/System/Buffer.cs
vkvenkat: Saw an 8% perf drop for len == 4 for a million copies.

src/mscorlib/src/System/Buffer.cs
vkvenkat: Yes, 64 performs better compared to 48. Seeing a regression with 48 for all 3 copy size buckets involved here.

src/mscorlib/src/System/Buffer.cs
mikedn: x86 CoreCLR switched to RyuJIT so there shouldn't be any differences between X86 and X64 here, both should use `Block64`.

I don't know what ARM64 does but it would make sense for it to use the best available method for moving a 64 bytes block, that should be the same or better than moving 4 bytes at a time.

It may be better to handle this by adding some specific define at the top of the file so it's easy to change:
```C#
#if AMD64 || X86
#define HAS_BLOCK64
#endif
```

DrewScoggins:
test Windows_NT_X64 perf
test Ubuntu14.04_x64 perf

DrewScoggins:
test Ubuntu14.04 perf

vkvenkat:
@jkotas  @jamesqo  Addressed all of the CR suggestions and pushed my latest. PTAL.

src/mscorlib/src/System/Buffer.cs
```diff
@@ -256,326 +260,146 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
+            MCPY00:
+            Debug.Assert(len > 16 && len <= 64);
+#if HAS_CUSTOM_BLOCKS            
+            *(Block16*)dest = *(Block16*)src;                   // [0,16]
 #else
-                *(int*)dest = *(int*)src;
-                *(int*)(dest + 4) = *(int*)(src + 4);
-                *(int*)(dest + 8) = *(int*)(src + 8);
-                *(int*)(dest + 12) = *(int*)(src + 12);
+            *(int*)dest = *(int*)src;
```
jkotas: These should be done as `long` for BIT64

src/mscorlib/src/System/Buffer.cs
```diff
@@ -256,326 +260,146 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
+            
+#if AMD64 && !PLATFORM_UNIX 
+            const nuint CopyThreshold = 1024;
+#else
+            const nuint CopyThreshold = 2048;
+#endif // AMD64 && !PLATFORM_UNIX
 
-            switch (len)
+            // PInvoke to the native version when the buffers are overlapping.            
+            if ((((nuint)dest <= (nuint)src) && ((nuint)destEnd > (nuint)src)) ||
```
jkotas: Why are you not using the original condition `if ((nuint)dest - (nuint)src < len) goto PInvoke;` ?

src/mscorlib/src/System/Buffer.cs
```diff
@@ -256,326 +260,146 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
+                
+            MCPY06:
+#if HAS_CUSTOM_BLOCKS
+            *(Block64*)dest = *(Block64*)src;
 #else
-                *(int*)(dest + i) = *(int*)(src + i);
-                *(int*)(dest + i + 4) = *(int*)(src + i + 4);
-                *(int*)(dest + i + 8) = *(int*)(src + i + 8);
-                *(int*)(dest + i + 12) = *(int*)(src + i + 12);
+            *(int*)dest = *(int*)src;
```
jkotas: use long for BIT64

src/mscorlib/src/System/Buffer.cs
vkvenkat: ```src + len``` needs to be re-computed in quite a few places and so created ```srcEnd``` to avoid this. Had optimized the original condition to ```if ((nuint)dest < (nuint)srcEnd)``` and used it in the PR submitted last week. Revised it in favor of a more thorough overlap check as suggested in one of your CR comments.

jkotas: > Had optimized the original condition to if ((nuint)dest < (nuint)srcEnd) 

This optimization was incorrect. The original condition had subtle dependency on integer overflow - the rearranging of left side and right side broke.

> Revised it in favor of a more thorough overlap check 

I do not think we need a more thorough overlap check here. We just need the original one that had a single compare and jump (unless there is some non-obvious reason why the one with two compares is faster...).

jkotas:

jkotas:
@vkvenkat Several tests are consistently failing. Could you please take a look?

```
CoreMangLib._cti_system_text_stringbuilder_StringBuilderReplace4_StringBuilderReplace4_._cti_system_text_stringbuilder_StringBuilderReplace4_StringBuilderReplace4_cmd
CoreMangLib._cti_system_version_VersionToString2_VersionToString2_._cti_system_version_VersionToString2_VersionToString2_cmd
CoreMangLib._cti_system_version_VersionToString1_VersionToString1_._cti_system_version_VersionToString1_VersionToString1_cmd
```

src/mscorlib/src/System/Buffer.cs
```diff
@@ -255,327 +259,169 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
-                *(int*)(dest + 12) = *(int*)(src + 12);
-#endif
-                    *(short*)(dest + 16) = *(short*)(src + 16);
-                    *(dest + 18) = *(src + 18);
-                    return;
-                case 20:
-#if BIT64
-                    *(long*)dest = *(long*)src;
-                    *(long*)(dest + 8) = *(long*)(src + 8);
+            if (len <= 48) goto MCPY01;
```
briansull: Shouldn't this be goto MCPY02

Going to MCPY01 will result in copying 16 bytes unconditionally and returning.
Which will overrun the end of the dest buffer for values less than 48

src/mscorlib/src/System/Buffer.cs
```diff
@@ -255,327 +259,169 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
+            *(int*)(dest + 44) = *(int*)(src + 44);             // [0,48]
 #endif
-                    *(int*)(dest + 16) = *(int*)(src + 16);
-                    return;
-                case 21:
-#if BIT64
-                    *(long*)dest = *(long*)src;
-                    *(long*)(dest + 8) = *(long*)(src + 8);
+            
+            MCPY01:
```
briansull: You should add a comment on each label describing what can be expected when we reach here:

MCPY1:   // Unconditionally copy the last 16 bytes and return.

src/mscorlib/src/System/Buffer.cs
```diff
@@ -255,327 +259,169 @@ internal unsafe static void Memcpy(byte* dest, byte* src, int len)
...
-            // This is separated out into a different variable, so the i + 16 addition can be
-            // performed at the start of the pipeline and the loop condition does not have
-            // a dependency on the writes.
-            nuint counter;
-
-            do
+            *(int*)dest = *(int*)src;
+            *(int*)(dest + 4) = *(int*)(src + 4);
+            *(int*)(destEnd - 8) = *(int*)(srcEnd - 8);
+            *(int*)(destEnd - 4) = *(int*)(srcEnd - 4);
```
briansull: You are using an interesting trick here to prevent the normal overrun that would occur when performing all 8 byte writes when len is not a multiple of 8.

You should explain how this trick works.

Also MCPY01 aqnd MCPY02 perform silimar operations by have a very subtle difference in their input requirements.

src/mscorlib/src/System/Buffer.cs
vkvenkat: As you explained below, MCPY01 copies the last 16 bytes of src to dest, and an overrun doesn't occur here. MCPY01 uses destEnd & srcEnd as opposed to dest & src in other places.

src/mscorlib/src/System/Buffer.cs
vkvenkat: We have debug asserts after labels like ```Debug.Assert(len > 16 && len <= 64)``` after MCPY00. Do you prefer switching these to comments?

src/mscorlib/src/System/Buffer.cs
vkvenkat: Sure, will do.

jkotas: Hmm, this trick may need a full overlap check to work ... it may be explaining the failure on Ubuntu.

vkvenkat: @jkotas Thanks. Investigating the test failures. Shall push a fix for this.

vkvenkat: Updated the overlap check to ```if (((nuint)dest - (nuint)src < len) || ((nuint)src - (nuint)dest < len)) goto PInvoke;```. This should resolve test failures.

vkvenkat: @briansull Added comments after labels to explain copy logic.

src/mscorlib/src/System/Buffer.cs
vkvenkat: @jkotas Yes, all of the results posted are with crossgened corelib

vkvenkat:
@jkotas Addressed latest CR feedback, fixed test failures & updated results. PTAL

jkotas:
Looks great - thank you!

glenn-slayden:
*I posted a comment for c6372c5 [here](https://github.com/dotnet/coreclr/commit/c6372c5bfebd61470ea5d111f224d035b1d2ebdf#commit-comment-27493209), but I don't think that was the right place, I apologize for repeating the remarks here...*

----

Just curious why all these improvements to `Buffer.Memmove` (plus more recent work) aren't deployed in the **IL** `cpblk` instruction instead? It appears that this commit c6372c5 [removes a comment](https://github.com/dotnet/coreclr/commit/c6372c5bfebd61470ea5d111f224d035b1d2ebdf#diff-f3e92259d4c0662a268d2c4b6b3c13c5L266) to the effect that **cpblk** is where all this effort naturally belongs (that is, in case the proliferation of conditional `#if` code here wasn't already somewhat suggestive), and I agree.

I know everyone wants there to be one central nexus for obtaining an optimal **memmove**, with proper alignment, chunking, tuning, and everything else done right; shouldn't `cpblk` be that place?

I also note that this commit [removes all address alignment](https://github.com/dotnet/coreclr/commit/c6372c5bfebd61470ea5d111f224d035b1d2ebdf#diff-f3e92259d4c0662a268d2c4b6b3c13c5L474) behavior. Now while it's true that the alignment code in ea9bee5 was a bit meek in only considering the destination address--at the possible expense of source alignment:

https://github.com/dotnet/coreclr/blob/ea9bee5ac2f96a1ea6b202dc4094b8d418d9209c/src/mscorlib/src/System/Buffer.cs#L474-L503

...at least this was something, and a reasonably motivated approach since misaligned store penalties are typically higher than for read.

But in this commit there's neither. Did the experiments in https://github.com/dotnet/coreclr/pull/9786 empirically determine that physically aligning these operations is always a waste of time? For severe misalignment between source/target, it would be possible to write a split-access loop, albeit complex, that marshals quad-aligned reads to quad-aligned writes. Was this type of thing evaluated before taking the decision to abandon all efforts at alignment?

Finally, note that by insisting on alignment to the _destination_ address exclusively, the alignment code (that was removed in this commit, see above) may have been ruining a naturally ideal _source_ alignment, with the overall result of making things worse. This could have unfairly penalized test results for the existing code, when it would be a trivial matter to only do that adjustment when src/dest are jointly aligned to some appropriate level, e.g., wrap the above code with:

    if ((((int)src ^ (int)dest) & 0x7) == 0)        // (or some other alignment value)
    {
        // ...

As far as I can tell, there has never been code in *buffer.cs* that considers the source and destination physical alignment jointly, in order to then proceed sensibly on the basis of ***3*** broadly defined cases:

 - Input src/dst physical addresses already jointly aligned (or `len` less than max align size);
 - Joint alignment can be improved (to some degree) prior to starting the gang-operation (code removed in c6372c5)
 - Perverse src/dst alignment detected; proceed as-is, or, if worthwhile, use split-access loop to fully remediate.

src/mscorlib/src/System/Buffer.cs
jamesqo: @vkvenkat Great work optimizing this. I didn't have much time to look over the code, but one thing I noticed-- why are there `Block16` / `Block64` structs but no `Block32`?

jamesqo: Lol I just realized you made that comment a year ago. Still asking if the code is still there though

glenn-slayden:
Commit c6372c5bfebd61470ea5d111f224d035b1d2ebdf makes a change to the detection of overlap between the source and destination in `Buffer.Memmove`:

https://github.com/dotnet/coreclr/blob/ea9bee5ac2f96a1ea6b202dc4094b8d418d9209c/src/mscorlib/src/System/Buffer.cs#L262

...becomes...

https://github.com/dotnet/coreclr/blob/c6372c5bfebd61470ea5d111f224d035b1d2ebdf/src/mscorlib/src/System/Buffer.cs#L271

I must admit I puzzled over the original code since the unsigned subtraction and compare "incorrectly" fails to identify source/destination overlap in the case where the destination (address) precedes the source (address) in memory. Indeed c6372c5bfebd61470ea5d111f224d035b1d2ebdf clarifies this and would also seem to fix a bug, but how did the original work anyway?

Then I realized of course that the answer is that the original code is stealthily incorporating a strong assumption about--and dependency upon--the ***copy direction*** of the managed `Buffer.Memmove` implementation which follows it. 

The unfortunate upshot is that, while c6372c5bfebd61470ea5d111f224d035b1d2ebdf certainly makes the code clearer and more readable, it now excludes ***all*** overlap cases and thus *no longer allows the performance benefit of fully-managed operation* (i.e., non-P/Invoke) in the **50% of cases** where the overlap happens to be compatible with the copy direction implemented in [Buffer.cs](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Buffer.cs). I blame the original code for not documenting its entirely non-obvious design and/or operation. 

Anyway, can we revisit this code to restore the optimization that was lost here? Namely, once again allow the managed code path to prevail (still subject to overall size considerations, etc.) instead of P/Invoke for the half of the detected overlaps which comport with the copy direction that's hard-coded into `Buffer.Memmove`?

Issue #9701, opened 2017-02-21
Crossgen treatment of BitConverter.IsLittleEndian


tmat:
Does crossgen recognize BitConverter.IsLittleEndian as an intrinsic today and eliminate dead code if the property is used to conditionally execute code?  

Related discussion: https://github.com/dotnet/corefx/issues/16354

jkotas:
It does not.

nguerrera:
Changing labels to track making it do so.

jkotas:
This is not really specific to crossgen. It won't get optimized with JIT too if you are the first one calling it.

The fix should be to replicate what `CORINFO_FIELD_INTRINSIC_ZERO` does - go through all instances of it, and add  `CORINFO_FIELD_INTRINSIC_ISLITTLEENDIAN` path next to it.

tmat:
Does that mean that in Desktop JIT and NGEN scenarios it won't get optimized either?

jkotas:
Right.

benaadams:
Would changing it to from a [readonly variable](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/BitConverter.cs#L33-L37) to a readonly property be a breaking change? (Guessing so...)

tmat:
@benaadams Yes.

mikedn:
@jkotas Which `BitConverter`? There's one in coreclr and one in corefx.

RussKeldorph:
@dotnet/jit-contrib

jkotas:
I think we should keep the BitConverter in CoreCLR, and get rid of the one on CoreFX. I will take care of cleaning it up. In the meantime, it is fine to do the optimization on CoreCLR BitConverter if you would like to do it.

While it would be possible to do the optimization for BitConverter in corefx, it is more involved; and there are other layering problems with BitConverter in corefx (duplicated in https://github.com/dotnet/corefx/blob/master/src/System.Private.Reflection.Metadata.Ecma335/src/Shims/BitConverter.cs, a bunch of stuff still using it in corelib). So it is easiest to just keep it in corelib.

mikedn:
> So it is easiest to just keep it in corelib.

Agreed. 

Recently I noticed that the ToX methods are pretty horrible CQ wise and I was considering fixing them. They may benefit from using `System.Runtime.CompilerServices.Unsafe` or some arch specific tricks but failing that a better `IsLittleEndian` would be good for them as well. 

[...]

Issue #9474, opened 2017-02-09
Finish bringing CoreRT and CoreCLR's CoreLibs into sync


danmosemsft:
We want to do this to (1) get code formatting from CoreRT (2) get bug fixes and new API to CoreRT so they reach UWP.

- [ ] Renames in CoreCLR to bring directories and files to match CoreRT eg
rename coreclr\src\mscorlib to coreclr\src\System.Private.CoreLib
rename coreclr\src\mscorlib\corefx to coreclr\src\System.Private.CoreLib\corert
- [ ] Update CoreCLR corelib's resource generation: change to .resx and autogenerate the SR.cs
- [ ] Take chunks of CoreRT, merge with CoreCLR and commit to CoreCLR then turn around and commit same to CoreRT.

Some of this was already done -- 

https://github.com/dotnet/coreclr/pull/9259
https://github.com/dotnet/coreclr/pull/9251

weshaggard:
Do we plan to try and share the source via submoduling or some other technique?

jkotas:
> coreclr\src\System.Private.CoreLib\corert

This may rather be coreclr\src\System.Private.CoreLib\shared as we have discussed

danmosemsft:
Added a checkbox for that segregation.

jkotas:
> folders like `core` and 

Yet another core ... we are ~42 core project :-) Maybe call it unshared or exclusive?

alexperovich:
I have the shared portions of coreclr and corert synchronized and my synching tool is now running to keep those directories in sync.

karelz:
Is it rather 2.1 issue than 2.0?

danmosemsft:
Work continues on this and much progress was made; however it is not aligned with 2.1 per se.

danmosemsft:
@jkotas 

> A lot of places in CoreCLR CoreLib use "ThrowHelper" class. ThrowHelper does not work well with ProjectN toolchain because of its use of reflection.

I am looking in C:\git\coreclr\src\mscorlib\src\System\ThrowHelper.cs and I do not see reflection. Where am I missing?

danmosemsft:
@jkotas do you still think it is worth renaming `coreclr\src\mscorlib` to `coreclr\src\System.Private.CoreLib` ? That seems like the only remaining rename.

jkotas:
> I do not see reflection

`Enum.ToString` is reflection call.

danmosemsft:
[...]

PR #9788, opened 2017-02-25
Update BitConverter from CoreFX

jkotas:
Related to #9701

### DIFF
--- a/src/mscorlib/corefx/SR.cs
+++ b/src/mscorlib/corefx/SR.cs
@@ -815,4 +815,19 @@ internal static class SR
     {
         get { return Environment.GetResourceString("Arg_UnauthorizedAccessException"); }
     }
+
+    internal static string ArgumentOutOfRange_GenericPositive
+    {
+        get { return Environment.GetResourceString("ArgumentOutOfRange_GenericPositive"); }
+    }
+
+    internal static string ArgumentOutOfRange_LengthTooLarge
+    {
+        get { return Environment.GetResourceString("ArgumentOutOfRange_LengthTooLarge"); }
+    }
+
+    internal static string Arg_ArrayPlusOffTooSmall
+    {
+        get { return Environment.GetResourceString("Arg_ArrayPlusOffTooSmall"); }
+    }
 }
--- a/src/mscorlib/src/System/BitConverter.cs
+++ b/src/mscorlib/src/System/BitConverter.cs
@@ -2,30 +2,16 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-/*============================================================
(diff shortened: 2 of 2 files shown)


mikedn:
Would it make sense to copy the initialization of `IsLittleEndian` from corefx too? We'd avoid an `#if` and when  `IsLittleEndian` becomes an intrinsic the initialization cost won't be a concern anymore.

jkotas:
I kept it because of there are number of BIGENDIAN ifdefs in corelib today, and so this is following the corelib style for endianness specific code.

jkotas:
Wait for the current CoreCLR drop to stabilize in corefx before pushing this through (see https://github.com/dotnet/corefx/pull/16454)

jkotas:
@stephentoub I see you have commented on the other PR related to this. Could you please take a look?

src/mscorlib/corefx/SR.cs
```diff
@@ -815,4 +815,19 @@ internal static string Arg_UnauthorizedAccessException
     {
         get { return Environment.GetResourceString("Arg_UnauthorizedAccessException"); }
     }
+
+    internal static string ArgumentOutOfRange_GenericPositive
+    {
+        get { return Environment.GetResourceString("ArgumentOutOfRange_GenericPositive"); }
```
stephentoub: Nit: Not for this PR, but at some point subsequently we might want to go through and replace all of this strings with nameof, e.g.
```C#
using static Environment;
...
internal static string ArgumentOutOfRange_GenericPositive => GetResourceString(nameof(ArgumentOutOfRange_GenericPositive);
```

src/mscorlib/src/System/BitConverter.cs
```diff
@@ -495,5 +452,20 @@ public static unsafe float Int32BitsToSingle(int value)
         {
             return *((float*)&value);
         }
+
+        private static void ThrowValueArgumentNull()
+        {
+            throw new ArgumentNullException("value");
```
stephentoub: Did you purposefully decide against ThrowHelper for these exceptions?

src/mscorlib/corefx/SR.cs
jkotas: The plan is switch this file to be auto-generated like everywhere else(https://github.com/dotnet/coreclr/issues/9474)

stephentoub: Even better.

src/mscorlib/src/System/BitConverter.cs
jkotas: Not really ... just copied what was in corefx. I have fixed it up to use ThrowHelper.

hqueue:

PR #9791, opened 2017-02-25
Revert "Simplified MinOpts GC ref tracking."

jkotas:
Reverts dotnet/coreclr#9231


jkotas:
See https://github.com/dotnet/corefx/pull/16454 . Causes massive corefx test failures (segfaults + failing tests).

AndyAyersMS:
dotnet/corefx#16378 seems to have caused CoreFX to stop producing linuxarmemulator_softfp artifacts which leads to the arm emulator build failures. Looks like we should ignore this for now.

PR #9794, opened 2017-02-25
Refactor register arg handling in TreeNodeInfoInit.

pgavlin:
The handling for outgoing register arguments in TreeNodeInfoInitCall was
unnecessarily complex and difficult to follow. In particular, the processing
of FIELD_LIST arguments performed a fair amount of unnecessary work,
contained code to deal with impossible IR patterns, and did not share any
code for handling the component PUTARG_REG nodes with the normal PUTARG_REG
case. This changes resolves all three of these issues and fixes an ordering
issue in lowering that was causing PInvoke prologs to remain unlowered.

### DIFF
--- a/src/jit/lower.cpp
+++ b/src/jit/lower.cpp
@@ -4423,6 +4423,14 @@ void Lowering::DoPhase()
 #endif
 #endif
 
+    // If we have any PInvoke calls, insert the one-time prolog code. We'll inserted the epilog code in the
+    // appropriate spots later. NOTE: there is a minor optimization opportunity here, as we still create p/invoke
+    // data structures and setup/teardown even if we've eliminated all p/invoke calls due to dead code elimination.
+    if (comp->info.compCallUnmanaged)
+    {
+        InsertPInvokeMethodProlog();
+    }
+
 #if !defined(_TARGET_64BIT_)
     DecomposeLongs decomp(comp); // Initialize the long decomposition class.
     decomp.PrepareForDecomposition();
@@ -4438,14 +4446,6 @@ void Lowering::DoPhase()
 #endif //!_TARGET_64BIT_
 
         LowerBlock(block);
-    }
-
-    // If we have any PInvoke calls, insert the one-time prolog code. We've already inserted the epilog code in the
-    // appropriate spots. NOTE: there is a minor optimization opportunity here, as we still create p/invoke data
-    // structures and setup/teardown even if we've eliminated all p/invoke calls due to dead code elimination.
-    if (comp->info.compCallUnmanaged)
-    {
-        InsertPInvokeMethodProlog();
     }
 
 #ifdef DEBUG
@@ -4600,13 +4600,6 @@ void Lowering::CheckCallArg(GenTree* arg)
 
     switch (arg->OperGet())
     {
-#if !defined(_TARGET_64BIT_)
-        case GT_LONG:
-            assert(arg->gtGetOp1()->OperIsPutArg());
-            assert(arg->gtGetOp2()->OperIsPutArg());
-            break;
-#endif
-
         case GT_FIELD_LIST:
         {
             GenTreeFieldList* list = arg->AsFieldList();
--- a/src/jit/lsraarm.cpp
+++ b/src/jit/lsraarm.cpp
@@ -330,6 +330,41 @@ void Lowering::TreeNodeInfoInitReturn(GenTree* tree)
 }
 
 //------------------------------------------------------------------------
+// TreeNodeInfoInitPutArgReg: Set the NodeInfo for a PUTARG_REG.
+//
+// Arguments:
+//    node                - The PUTARG_REG node.
+//    argReg              - The register in which to pass the argument.
+//    info                - The info for the node's using call.
+//    isVarArgs           - True if the call uses a varargs calling convention.
+//    callHasFloatRegArgs - Set to true if this PUTARG_REG uses an FP register.
+//
+// Return Value:
+//    None.
+//
+void Lowering::TreeNodeInfoInitPutArgReg(
+    GenTreeUnOp* node, regNumber argReg, TreeNodeInfo& info, bool isVarArgs, bool* callHasFloatRegArgs)
+{
+    assert(node != nullptr);
+    assert(node->OperIsPutArgReg());
+    assert(argReg != REG_NA);
+
+    // Each register argument corresponds to one source.
+    info.srcCount++;
+
+    // Set the register requirements for the node.
+    const regMaskTP argMask = genRegMask(argReg);
+    node->gtLsraInfo.setDstCandidates(m_lsra, argMask);
+    node->gtLsraInfo.setSrcCandidates(m_lsra, argMask);
+
+    // To avoid redundant moves, have the argument operand computed in the
+    // register in which the argument is passed to the call.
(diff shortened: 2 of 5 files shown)


pgavlin:
@BruceForstall @jashook @dotnet/jit-contrib PTAL

hqueue:

pgavlin:
Ping.

pgavlin:
@sivarv, can you take a look at this?

jashook:

src/jit/lsraarm.cpp
```diff
@@ -454,103 +489,22 @@ void Lowering::TreeNodeInfoInitCall(GenTreeCall* call)
             continue;
         }
 
-        var_types argType    = argNode->TypeGet();
-        bool      argIsFloat = varTypeIsFloating(argType);
-        callHasFloatRegArgs |= argIsFloat;
-
-        regNumber argReg = curArgTabEntry->regNum;
-        // We will setup argMask to the set of all registers that compose this argument
-        regMaskTP argMask = 0;
```
briansull: I note that identical code also appears in lsraarm64.cpp (line 980)

Apparently we compute argMask but it is never used/needed.

src/jit/lsraarm64.cpp
```diff
@@ -971,103 +1006,22 @@ void Lowering::TreeNodeInfoInitCall(GenTreeCall* call)
...
         }
-        else // A scalar argument (not a struct)
+        else
         {
-            // We consume one source
-            info->srcCount++;
-
-            argMask |= genRegMask(argReg);
-            argNode->gtLsraInfo.setDstCandidates(l, argMask);
-            argNode->gtLsraInfo.setSrcCandidates(l, argMask);
```
briansull: Prior to you change the 'argMask' could contain multiple bits (from all of the previous arg regs)
Now it just contains the current arg regmask.
This is probably correct, but I just want you to confirm this.

src/jit/lsraarm.cpp
```diff
@@ -330,6 +330,41 @@ void Lowering::TreeNodeInfoInitReturn(GenTree* tree)
 }
 
 //------------------------------------------------------------------------
+// TreeNodeInfoInitPutArgReg: Set the NodeInfo for a PUTARG_REG.
+//
+// Arguments:
+//    node                - The PUTARG_REG node.
+//    argReg              - The register in which to pass the argument.
+//    info                - The info for the node's using call.
+//    isVarArgs           - True if the call uses a varargs calling convention.
```
briansull: You should add a note here saying that 'isVarArgs' is only used by target xarch

src/jit/lsraarm64.cpp
```diff
@@ -848,6 +848,41 @@ void Lowering::TreeNodeInfoInitReturn(GenTree* tree)
 }
 
 //------------------------------------------------------------------------
+// TreeNodeInfoInitPutArgReg: Set the NodeInfo for a PUTARG_REG.
+//
+// Arguments:
+//    node                - The PUTARG_REG node.
+//    argReg              - The register in which to pass the argument.
+//    info                - The info for the node's using call.
+//    isVarArgs           - True if the call uses a varargs calling convention.
```
briansull: You should add a note here saying that 'isVarArgs' is only used by target xarch

src/jit/lsraxarch.cpp
```diff
@@ -1174,6 +1174,55 @@ void Lowering::TreeNodeInfoInitShiftRotate(GenTree* tree)
...
+    const regMaskTP argMask = genRegMask(argReg);
+    node->gtLsraInfo.setDstCandidates(m_lsra, argMask);
+    node->gtLsraInfo.setSrcCandidates(m_lsra, argMask);
+
+    // To avoid redundant moves, have the argument operand computed in the
+    // register in which the argument is passed to the call.
+    node->gtOp.gtOp1->gtLsraInfo.setSrcCandidates(m_lsra, m_lsra->getUseCandidates(node));
+
+#if FEATURE_VARARG
+    *callHasFloatRegArgs |= varTypeIsFloating(node->TypeGet());
```
briansull: Should this assignment really be under +#if FEATURE_VARARG ?

src/jit/lsraxarch.cpp
```diff
@@ -1174,6 +1174,55 @@ void Lowering::TreeNodeInfoInitShiftRotate(GenTree* tree)
...
+    // we must pass the enregistered arguments in both the integer and floating point registers.
+    // Since the integer register is not associated with this arg node, we will reserve it as
+    // an internal register so that it is not used during the evaluation of the call node
+    // (e.g. for the target).
+    if (isVarArgs && varTypeIsFloating(node))
+    {
+        regNumber targetReg = comp->getCallArgIntRegister(argReg);
+        info.setInternalIntCount(info.internalIntCount + 1);
+        info.addInternalCandidates(m_lsra, genRegMask(targetReg));
+    }
```
briansull: Maybe this varargs stuff could be removed from this common method and you could add a second method and call that just for xarch .

src/jit/lower.cpp
```diff
@@ -4423,6 +4423,14 @@ void Lowering::DoPhase()
 #endif
 #endif
 
+    // If we have any PInvoke calls, insert the one-time prolog code. We'll inserted the epilog code in the
```
sivarv: > inserted [](start = 75, length = 9)

Typo: "insert"

src/jit/lsraxarch.cpp
```diff
@@ -1372,166 +1429,30 @@ void Lowering::TreeNodeInfoInitCall(GenTreeCall* call)
...
-        // an internal register so that it is not used during the evaluation of the call node
-        // (e.g. for the target).
-        if (call->IsVarargs() && varTypeIsFloating(argNode))
-        {
-            regNumber targetReg = compiler->getCallArgIntRegister(argReg);
-            info->setInternalIntCount(info->internalIntCount + 1);
-            info->addInternalCandidates(l, genRegMask(targetReg));
+            TreeNodeInfoInitPutArgReg(argNode->AsUnOp(), curArgTabEntry->regNum, *info, isVarArgs,
+                                      &callHasFloatRegArgs);
         }
```
sivarv: It might be a good idea to generate asm diffs on unix amd64 to make sure there are zero asm diffs.

sivarv:
:shipit:

src/jit/lsraarm.cpp
pgavlin: `argMask` was only ever computed per reg argument. We still compute `argMask`, but this computation is now encapsulated in `TreeNodeInfoInitPutArgReg`.

src/jit/lsraarm64.cpp
pgavlin: Prior to this change `argMask` was guaranteed to be `RBM_NONE` at the point at which `genRegMask(argReg)` was or'd in (i.e. line 1059), so it would also have only contained the current arg reg mask. The behavior is the same in both cases.

src/jit/lsraxarch.cpp
pgavlin: Yes: placing this inside `FEATURE_VARARG` matches the behavior of the original code.

src/jit/lsraxarch.cpp
pgavlin: I have done so; there are no diffs.

src/jit/lsraxarch.cpp
pgavlin: Personally I don't think that bifurcating the interface is worth the extra `ifdefs`.

PR #9795, opened 2017-02-25
Refactor isContained().

pgavlin:
This changes isContained to terminate early if it is not possible for the node in question
to ever be contained. There are many nodes that are statically non-containable: any node
that is not a value may not be contained along with a small number of value nodes. To avoid
the need for a `switch` to identify the latter, their node kinds have been marked with a
new flag, `GTK_NOCONTAIN`. The checks for identifying non-containable nodes are encapsulated
withing a new function, `canBeContained`.

### DIFF
--- a/src/jit/gentree.cpp
+++ b/src/jit/gentree.cpp
@@ -15787,6 +15787,20 @@ unsigned GenTree::IsLclVarUpdateTree(GenTree** pOtherTree, genTreeOps* pOper)
 }
 
 //------------------------------------------------------------------------
+// canBeContained: check whether this tree node may be a subcomponent of its parent for purposes
+//                 of code generation.
+//
+// Return value: returns true if it is possible to contain this node and false otherwise.
+bool GenTree::canBeContained() const
+{
+    assert(IsLIR());
+
+    // It is not possible for nodes that do not produce values or that are not containable values
+    // to be contained.
+    return (OperKind() & (GTK_NOVALUE | GTK_NOCONTAIN)) == 0;
+}
+
+//------------------------------------------------------------------------
 // isContained: check whether this tree node is a subcomponent of its parent for codegen purposes
 //
 // Return Value:
@@ -15801,14 +15815,16 @@ unsigned GenTree::IsLclVarUpdateTree(GenTree** pOtherTree, genTreeOps* pOper)
 //
 bool GenTree::isContained() const
 {
-    if (gtHasReg())
+    assert(IsLIR());
+
+    if (!canBeContained() || gtHasReg())
     {
         return false;
     }
 
     // these actually produce a register (the flags reg, we just don't model it)
     // and are a separate instruction from the branch that consumes the result
-    if (OperKind() & GTK_RELOP)
+    if ((OperKind() & GTK_RELOP) != 0)
     {
         return false;
     }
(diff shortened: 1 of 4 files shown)


pgavlin:
@BruceForstall @sivarv @mikedn @dotnet/jit-contrib PTAL

hqueue:

src/jit/gentree.h
```diff
@@ -120,6 +120,8 @@ enum genTreeKinds
     GTK_NOVALUE = 0x0400, // node does not produce a value
     GTK_NOTLIR  = 0x0800, // node is not allowed in LIR
 
+    GTK_NOCONTAIN = 0x1000, // this node is a value, but may not be contained
```
mikedn: To continue from the other PR:
Wouldn't it make more sense to have `GTK_CONTAIN` and mark only those opers that can actually be contained? 

I think that the complete oper list can be gathered from `genConsumeRegs` - if it doesn't know how to handle an oper then that oper cannot be contained. From a cursory look we have `GT_IND`, `GT_LEA`, `GT_LCL_VAR`, `GT_INIT_VAL`, all leafs and GT_SIMD (only because of SIMD 0). And if it's true that `genConsumeRegs` decides what can be contained then `canBeContained` looks like something that belongs somewhere in codegen, not in `GenTree`.

russellhadley: Could we just add a new GT opcode that means contained mem?  It looks like you could combine IND, LEA, and LCL_VAR into one "contained" mem thing at lower and be done with it.    The reason I'm thinking new opcode is that the register requirements would different in that the operation would not produce a value.  I'm also more inclined to have this at lower so that a more aggressive and concrete job could be done building address modes before RA.

mikedn: > and LCL_VAR into one "contained" mem thing at lower and be done with it

Isn't LCL_VAR either memory or reg depending on RA?

pgavlin: > Wouldn't it make more sense to have GTK_CONTAIN and mark only those opers that can actually be contained?

It might, but the list of such nodes is trickier to gather than it appears. `genConsumeRegs` and friends do not generate a complete list; see e.g. https://github.com/dotnet/coreclr/blob/master/src/jit/codegenxarch.cpp#L3363-L3365. There probably aren't too many more situations like this, but we would need to exhaustively enumerate them.

> I'm also more inclined to have this at lower so that a more aggressive and concrete job could be done building address modes before RA.

The process of choosing which nodes to contain should certainly be handled at lowering, but it does seem reasonable to me to hang `canBeContained` off of `CodeGen` instead of `GenTree`.

> Isn't LCL_VAR either memory or reg depending on RA?

Contained `LCL_VAR` nodes are always memory: they are either register candidate lclVars that are currently live on the stack or non-register-candidate lclVars.

> It looks like you could combine IND, LEA, and LCL_VAR into one "contained" mem thing

Each of these cases has subtly different semantics:
- `GT_IND` denotes a memory load from the address defined by its single operand
- `GT_LEA` does not denote a memory access at all--just a (base, index, scale, offset) address mode, where scale and offset are always constants
- `GT_LCL_VAR` when contained denotes a memory load from an implied address

Now, it may be the case that `GT_LEA` most frequently appears as a contained operand to memory loads/stores (`GT_IND`, `GT_STOREIND`, `GT_NULLCHECK`, block ops), so perhaps the semantic difference isn't as much of a barrier as it seems: I can imagine folding `GT_IND` and contained `GT_LCL_VAR` into e.g. `GT_MEM` which would denote a memory access at some address that is described by either a lclVar number or a (possible target-dependent) address mode and obviate the need for `GT_LEA` in these situations. I don't really see this simplifying much unless we add output operands, however, which might be a rather significant design change.

mikedn: > Contained LCL_VAR nodes are always memory

Yes but contained means that no register has been assigned to the node. So it's not clear how lowering could do the contained mem thing before RA runs.

> I can imagine folding GT_IND and contained GT_LCL_VAR into e.g. GT_MEM which would denote a memory access at some address that is described by either a lclVar number or a (possible target-dependent) address mode and obviate the need for GT_LEA in these situations

It's interesting that we may be able to fold `GT_IND` and `GT_LEA` into a single node but the same cannot be done easily with `GT_STOREIND` and `GT_LEA` because the resulting node would have 3 operands. I suspect that will cause some complications due to `GTF_REVERSE_OPS`.

pgavlin: > I suspect that will cause some complications due to `GTF_REVERSE_OPS`.

FWIW, `GTF_REVERSE_OPS` isn't terribly meaningful in the backend because tree order is no longer relevant in LIR. We still use this flag to decide in which order we consume the operands for a node, but I don't think that usage is really all that useful and it ought to be easily removed.

pgavlin: > Yes but contained means that no register has been assigned to the node. So it's not clear how lowering could do the contained mem thing before RA runs.

The RA does make the final determination re: whether or not most lclVar nodes are contained, yes. In these cases lowering has typically marked the contained node as "register optional". TreeNodeInfoInit does get cute in a couple of places and preemptively mark some LCL_VAR nodes as contained if it thinks they'll never be enregistered, but in all other cases I think we use reg optional.

russellhadley: Leaving the RA to make reg optional decisions is the right thing but in my very general hand wavy idea when the RA makes a decision to use mem rather than reg the IR would be modified for the ref position in resolution to use a MEM node for the spill.  The overarching idea is to minimize the contracts between phases and put the transformation where the decision was made, the contract then could narrowed down to a simpler, unambiguous IR (fingers crossed).  Ideally we would make Lower more responsible for shaping the IR for the target and CodeGen more responsible for encoding the bits.

mikedn: >  Ideally we would make Lower more responsible for shaping the IR for the target and CodeGen more responsible for encoding the bits.

👍

src/jit/gentree.cpp
```diff
@@ -15819,75 +15835,25 @@ bool GenTree::isContained() const
...
-#ifdef FEATURE_SIMD
-        case GT_SIMD_CHK:
-#endif // FEATURE_SIMD
-
-#if !FEATURE_EH_FUNCLETS
-        case GT_END_LFIN:
-#endif
-            return false;
-
 #if !defined(LEGACY_BACKEND) && !defined(_TARGET_64BIT_)
```
sivarv: >#if !defined(LEGACY_BACKEND)  [](start = 0, length = 29)

We don't need to use !defined(LEGACY_BACKEND) because the concept of contained-ness is not used by legacy backend.
In fact this entire routine should be under #ifndef LEGACY_BACKEND.

src/jit/gentree.cpp
```diff
@@ -15819,75 +15835,25 @@ bool GenTree::isContained() const
...
     }
 
-    switch (OperGet())
-    {
-        case GT_STOREIND:
-        case GT_JTRUE:
-        case GT_JCC:
-        case GT_RETURN:
-        case GT_RETFILT:
-        case GT_STORE_LCL_FLD:
```
sivarv: Assuming zero superPMI asm diffs due to this change.

sivarv:
:shipit:

src/jit/gentree.cpp
pgavlin: That is correct.

PR #9802, opened 2017-02-26
Updating the gentree to support directly carrying single-precision floating-point values.

tannergooding:
As per the title, this updates the gentree to support directly carrying single-precision floating-point values. Previously, it would upcast these to `double` and track the original type. All this upcasting/downcasting is non-optimal and blocked several code-paths from executing since `GT_CNS_DBL` could represent either `TYP_DOUBLE` or `TYP_FLOAT`.


tannergooding:
test this please

tannergooding:
FYI. @dotnet/jit-contrib

tannergooding:
FYI. @dotnet/jit-contrib.

russellhadley:
@sivarv Can you take a look here?

src/jit/gentree.h
```diff
@@ -574,11 +574,21 @@ struct GenTree
         return isContained() && IsCnsIntOrI() && !isUsedFromSpillTemp();
     }
 
+#if FEATURE_X87_DOUBLES
```
sivarv: >FEATURE_X87_DOUBLES [](start = 4, length = 19)

Note that FEATURE_X87_DOUBLES is meant for tracking floating point code specific to x87 which only legacy x86 supports.

Using that for float vs dbl constants doesn't seem right to me.

tannergooding: I had done it since there is no concept of a 'float' with regards to x87 FPU, instead everything is 80 bits.

I can certainly drop all the `#if FEATURE_X87_DOUBLES` and either:
* Just use the new code everywhere, or
* Switch to a different define

If a different approach would be more ideal, please let me know.

sivarv: Introducing a  new #ifdef would be a better approach.  Other than x86 Legacy backend, other backends will define it.

tannergooding: How does `FEATURE_CNS_FLOAT` or `FEATURE_GENTREE_FLOAT` sound?

sivarv: Another orthogonal question is why do we need to represent float constant using a separate gentree node?  What optimizations are getting blocked due to current representation using GenTreeDblCon with type being TYP_FLOAT or TYP_DOUBLE?

tannergooding: Here was one example: https://github.com/dotnet/coreclr/pull/9802/files#diff-c2896a82e1a2272c25849727a769712dL2445 and There were a couple other similar instances scattered around.

It isn't strictly required to support float constants (as we could just check the real type and do the appropriate thing). However, all the upcasting and downcasting is not ideal:
* It means we incur the cost of the upcast/downcast whenver we do it
* It means the values computed for floats could be "off" (it may be possible to compute a value that couldn't be computed without upcasting, resulting in a different end result)

sivarv: When you say cost of upcast/downcast done during jitting?  I am not sure that is a significant enough cost to worry about.  

Do you have data to conclude the the two points in your reply?

tannergooding: For the former, just the recommendations specified in Intel Architecture Optimization Manual: http://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-optimization-manual.pdf and in the AMD Software Optimization Guide: https://support.amd.com/TechDocs/47414_15h_sw_opt_guide.pdf.

For the second, a very simple example is `1.70141173E+38f * 3 / 2`. When done using purely single-precision arithmetic, the result is `PositiveInfinity`, however when each constant is first cast to double and then downcast again `(float)((double)1.70141173E+38f * (double)3 / (double)2)` you get a different result: `2.55211755E+38`. This is also where order of operations matters, because `1.70141173E+38f * 1.5f` also returns `2.55211755E+38` rather than `PositiveInfinity`.

Most of the constant folding around float/double appears to be disable today due to the issues that can arise if things are done wrong (and because we don't allow carrying of single-precision constants).

tannergooding: IEEE 754 does allow you to evaluate expressions and perform constant folding, provided you follow all the rules set forth in the spec. This would make much of that easier.

sivarv: Note that we have already shipped in current state on Desktop and CoreCLR.  Now tracking floating-point constants more precisely could lead to an app compat issue for reason two in your reply.

As per my understanding Jit does float const upcast/downcast during jitting not in the generated jitted code.

tannergooding: I think if we are actually violating 2, then it needs to be fixed as it would mean we aren't following the IEEE 754 spec, and by extension the ECMA-335 spec (which states we follow the IEEE 754 spec for all floating point numbers).

If we aren't violating 2, then this change won't break anything and is an optimization and could potentially make future optimizations easier. We would only be taking up 4 less bytes per float constant we stored and would not have to upcast/downcast anytime we had to deal with the value.

danmosemsft:
@sivarv @tannergooding is this PR still in flight?

tannergooding:
At least from my side, yes. I will get to introducing a new `FEATURE_*` flag sometime tonight I believe. If not, I will close the PR and reopen once I do get to fixing it.

BruceForstall:
@tannergooding It doesn't seem like there is consensus here that we should be doing this.

Instead of creating a PR, you should create an issue, motivate the work, tag the JIT team, and make sure we arrive at consensus.

tannergooding:
Can do.

PR #9804, opened 2017-02-26
Timespan division and multiplication.

JonHanna:
https://github.com/dotnet/corefx/issues/7175

### DIFF
--- a/src/mscorlib/src/System/TimeSpan.cs
+++ b/src/mscorlib/src/System/TimeSpan.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
@@ -404,6 +404,48 @@ namespace System
         {
             return t1.Add(t2);
         }
+
+        public static TimeSpan operator *(TimeSpan timeSpan, double factor)
+        {
+            if (double.IsNaN(factor))
+            {
+                throw new ArgumentException(Environment.GetResourceString("Arg_CannotBeNaN"), nameof(factor));
+            }
+
+            // Rounding to the nearest tick is as close to the result we would have with unlimited
+            // precision as possible, and so likely to have the least potential to surprise.
+            double ticks = Math.Round(timeSpan.Ticks * factor);
+            if (ticks > long.MaxValue | ticks < long.MinValue)
+            {
+                throw new OverflowException(Environment.GetResourceString("Overflow_TimeSpanTooLong"));
+            }
+
+            return FromTicks((long)ticks);
+        }
+
+        public static TimeSpan operator *(double factor, TimeSpan timeSpan) => timeSpan * factor;
+
+        public static TimeSpan operator /(TimeSpan timeSpan, double divisor)
+        {
+            if (double.IsNaN(divisor))
+            {
+                throw new ArgumentException(Environment.GetResourceString("Arg_CannotBeNaN"), nameof(divisor));
+            }
+
+            double ticks = Math.Round(timeSpan.Ticks / divisor);
+            if (ticks > long.MaxValue | ticks < long.MinValue || double.IsNaN(ticks))
+            {
+                throw new OverflowException(Environment.GetResourceString("Overflow_TimeSpanTooLong"));
+            }
+
+            return FromTicks((long)ticks);
+        }
+
+        // Using floating-point arithmetic directly means that infinities can be returned, which is reasonable
+        // if we consider TimeSpan.FromHours(1) / TimeSpan.Zero asks how many zero-second intervals there are in
+        // an hour for which ∞ is the mathematic correct answer. Having TimeSpan.Zero / TimeSpan.Zero return NaN
+        // is perhaps less useful, but no less useful than an exception.
+        public static double operator /(TimeSpan t1, TimeSpan t2) => t1.Ticks / (double)t2.Ticks;
 
         public static bool operator ==(TimeSpan t1, TimeSpan t2)
         {


JonHanna:
Tests at dotnet/corefx#16475

JonHanna:
Test CentOS7.1 x64 Debug Build and Test

src/mscorlib/src/System/TimeSpan.cs
```diff
@@ -405,6 +405,42 @@ public String ToString(String format, IFormatProvider formatProvider)
...
+        {
+            double ticks = timeSpan.Ticks / divisor;
+            if (ticks > long.MaxValue || ticks < long.MinValue || double.IsNaN(ticks))
+            {
+                // Assuming exceptions are the rarer case, and since both IsNaN(divisor) and 0/0 can
+                // result in IsNaN(ticks), defer determining which exception until this branch has
+                // been taken.
+                throw double.IsNaN(divisor)
+                    ? new ArgumentException(Environment.GetResourceString("Arg_CannotBeNaN"), nameof(divisor))
+                    : (Exception)new OverflowException(Environment.GetResourceString("Overflow_TimeSpanTooLong"));
```
danmosemsft: Why is this cast to Exception needed?

danmosemsft:
@tarekgh can you review also?

src/mscorlib/src/System/TimeSpan.cs
joelweiss: > [Either the type of first_expression and second_expression must be the same, or an implicit conversion must exist from one type to the other.
> ](https://msdn.microsoft.com/en-us/library/ty67wk28.aspx)

JonHanna: Yep, though it's one of those casts that don't have any effect on the compiled code, but are needed to make the expression valid.

src/mscorlib/src/System/TimeSpan.cs
```diff
@@ -405,6 +405,42 @@ public String ToString(String format, IFormatProvider formatProvider)
...
         }
 
+        public static TimeSpan operator *(TimeSpan timeSpan, double factor)
+        {
+            if (double.IsNaN(factor))
+            {
+                throw new ArgumentException(Environment.GetResourceString("Arg_CannotBeNaN"), nameof(factor));
+            }
+
+            double ticks = timeSpan.Ticks * factor;
```
tarekgh: >timeSpan.Ticks * factor; [](start = 27, length = 24)

should we put this inside checked() to throw for the overflow too?

src/mscorlib/src/System/TimeSpan.cs
```diff
@@ -405,6 +405,42 @@ public String ToString(String format, IFormatProvider formatProvider)
...
+            if (ticks > long.MaxValue | ticks < long.MinValue)
+            {
+                throw new OverflowException(Environment.GetResourceString("Overflow_TimeSpanTooLong"));
+            }
+
+            return FromTicks((long)ticks);
```
tarekgh: >(long)ticks [](start = 29, length = 11)

should we care about rounding?

src/mscorlib/src/System/TimeSpan.cs
```diff
@@ -405,6 +405,42 @@ public String ToString(String format, IFormatProvider formatProvider)
...
+
+        public static TimeSpan operator /(TimeSpan timeSpan, double divisor)
+        {
+            double ticks = timeSpan.Ticks / divisor;
+            if (ticks > long.MaxValue || ticks < long.MinValue || double.IsNaN(ticks))
+            {
+                // Assuming exceptions are the rarer case, and since both IsNaN(divisor) and 0/0 can
+                // result in IsNaN(ticks), defer determining which exception until this branch has
+                // been taken.
+                throw double.IsNaN(divisor)
```
tarekgh: >double.IsNaN(divisor) [](start = 22, length = 21)

it is better to do this check first in the beginning of the method and then make the code cleaner. I understand you want to optimize but I think one check here is not really a big deal.

src/mscorlib/src/System/TimeSpan.cs
```diff
@@ -405,6 +405,42 @@ public String ToString(String format, IFormatProvider formatProvider)
...
+            }
+
+            return FromTicks((long)ticks);
+        }
+
+        public static TimeSpan operator *(double factor, TimeSpan timeSpan) => timeSpan * factor;
+
+        public static TimeSpan operator /(TimeSpan timeSpan, double divisor)
+        {
+            double ticks = timeSpan.Ticks / divisor;
```
tarekgh: >timeSpan.Ticks / divisor; [](start = 27, length = 25)

should we put this inside checked()? for better determination of overflow/underflow?

src/mscorlib/src/System/TimeSpan.cs
```diff
@@ -405,6 +405,42 @@ public String ToString(String format, IFormatProvider formatProvider)
...
+                // been taken.
+                throw double.IsNaN(divisor)
+                    ? new ArgumentException(Environment.GetResourceString("Arg_CannotBeNaN"), nameof(divisor))
+                    : (Exception)new OverflowException(Environment.GetResourceString("Overflow_TimeSpanTooLong"));
+            }
+
+            return FromTicks((long)ticks);
+        }
+
+        public static double operator /(TimeSpan t1, TimeSpan t2) => t1.Ticks / (double)t2.Ticks;
```
tarekgh: >TimeSpan t2 [](start = 53, length = 11)

do we need to allow dividing by TimeSpan(0)? ticks is really long and theoretically we are dividing by long 0 and not double 0

src/mscorlib/src/System/TimeSpan.cs
JonHanna: No effect on double arithmetic.

src/mscorlib/src/System/TimeSpan.cs
JonHanna: I think it's probably best to be consistent with `FromHours()` etc. and they do a direct cast from double to long after explicitly checking for overflow too.

src/mscorlib/src/System/TimeSpan.cs
JonHanna: Sure. Will do.

src/mscorlib/src/System/TimeSpan.cs
JonHanna: Practically we're dividing longs as a matter of representation, but theoretically we're dividing a span of time by another. I think it's reasonable to be able to ask the question "how many zero-second periods in an hour?"

src/mscorlib/src/System/TimeSpan.cs
tarekgh: Actually FromHours rounds to the milliseconds,

JonHanna: Right you are. That seems probably too coarse for this, in that `FromTicks(2) * 2.0` should probably work as one would expect.

src/mscorlib/src/System/TimeSpan.cs
JonHanna: Ignored by the compiler on double arithmetic.

src/mscorlib/src/System/TimeSpan.cs
tarekgh: thanks for reminding me about it :-)

src/mscorlib/src/System/TimeSpan.cs
tarekgh: I am fine with this interpenetration. could you please put some comments around the added APIs describing the design and the issues discussed here?

src/mscorlib/src/System/TimeSpan.cs
tarekgh: but what about FromTicks(2) * 2.9?

---
In reply to: [103531161](https://github.com/dotnet/coreclr/pull/9804#discussion_r103531161) [](ancestors = 103531161)

JonHanna: I think, on balance, that rounding to the nearest tick is probably the best approach, so I'll update to do that. It's only going to matter when half a tick makes a difference, but it's probably what would have the least surprises for users.

tarekgh: thanks make sense. thank you.

tarekgh:
LGTM. thanks @JonHanna for getting this ready. are you willing to enable exposing these APIs in the corefx repo?

stephentoub:
@tarekgh, he already created the PR: https://github.com/dotnet/corefx/pull/16475

tarekgh:
thanks @stephentoub

JonHanna:
Yes, it needs some fixup for uapaot but I should have that ready before the changes here are visible to corefx.

tarekgh:
The UAP-aot changes should go to corert repo and not here.

PR #9808, opened 2017-02-26
Add Unsafe.ReadUnaligned/WriteUnaligned

mikedn:


src/vm/jitinterface.cpp
```diff
@@ -7053,6 +7053,53 @@ bool getILIntrinsicImplementationForUnsafe(MethodDesc * ftn,
         methInfo->options = (CorInfoOptions)0;
         return true;
     }
+    else if (tk == MscorlibBinder::GetMethod(METHOD__UNSAFE__BYREF_READ_UNALIGNED)->GetMemberDef() ||
+             tk == MscorlibBinder::GetMethod(METHOD__UNSAFE__PTR_READ_UNALIGNED)->GetMemberDef())
+    {
+        _ASSERTE(ftn->HasMethodInstantiation());
+        Instantiation inst = ftn->GetMethodInstantiation();
+        _ASSERTE(ftn->GetNumGenericMethodArgs() == 1);
+        mdToken tokGenericArg = FindGenericMethodArgTypeSpec(MscorlibBinder::GetModule()->GetMDImport());
```
jkotas: You may need to add `typeof(T).ToString(); // Type used by the actual method body` to the dummy implementation to make FindGenericMethodArgTypeSpec work well.

mikedn: But how does FindGenericMethodArgTypeSpec work anyway? I just don't see how it knows the method I'm working with, all it gets is a module MD import. And `Add` does use `typeof(T).ToString()` but SizeOf does not. Yet both use the same code to extract the necessary token.

jkotas: `FindGenericMethodArgTypeSpec` finds a typespec token that points to `!!0` signature. This signature is context free. The same typespec token is shared between all methods that reference it, so it will work as long as there is at least one method that references `!!0` in CoreLib.

> but SizeOf does not

SizeOf happens to work because of there is other method that references `!!0` signature. If that other place was changed or deleted, SizeOf would break. It would be nice to add `typeof(T).ToString()` to SizeOf as well.

jkotas:
@mikedn Is this good to go, or is there `[WIP]` in the title for a reason?

mikedn:
The WIP was mostly for `FindGenericMethodArgTypeSpec` since it wasn't clear to me how it works.

Otherwise it's done, if you're ok with the lack of tests :)

jakobbotsch:
What is the difference between this file and the NuGet package of the same name? It looks like they have diverged and will diverge more, for example the NuGet package has a `Read/Write` that only reads from native pointers, and no (easy) support for unaligned reads. Should they support the same things?

mikedn:
@jakobbotsch There's an issue in corefx to add these to the NuGet package: https://github.com/dotnet/corefx/issues/16479

jkotas:
> What is the difference between this file and the NuGet package of the same name?

The type in this file is `internal` to CoreLib. It is subsetted version of Unsafe from the nuget package. There are several other types like this where CoreLib has internal subsetted version of a type that is implemented in corefx. We do this to deal with dependency cycles and growing CoreLib too big.

jkotas:
@mikedn Thank you!

mikedn:
@jkotas Thank you too, I'll update the BitConverter PR in a hour or two.

PR #9607, opened 2017-02-15
[x86/Linux] Fix EnclosingClauseCallerSP

wateret:
Seems like this is missing part of x86/linux port.

PR #9814, opened 2017-02-27
[x86/Linux] Use ESTABLISHER_FRAME_POINTER_IS_CALLER_SP

parjong:
exceptionhandling.cpp contains several code fragments that are valid when Establisher Frame Pointer is same as Caller SP, but these fragments are explicitly ifdefed by _TARGET_XXX_.

This commit introduces ESTABLISHER_FRAME_POINTER_IS_CALLER_SP to annotate such code fragments, and use it to ifdef them (instead of explicitly using _TARGET_XXX_).

### DIFF
--- a/src/vm/exceptionhandling.cpp
+++ b/src/vm/exceptionhandling.cpp
@@ -20,9 +20,10 @@
 #if defined(_TARGET_ARM_) || defined(_TARGET_ARM64_) || defined(_TARGET_X86_)
 #define ADJUST_PC_UNWOUND_TO_CALL
 #define USE_CALLER_SP_IN_FUNCLET
-// EstablisherFrame is Caller-SP (SP just before executing call instruction) for x86, ARM, ARM64.
-//
+// For ARM/ARM64, EstablisherFrame is Caller-SP (SP just before executing call instruction).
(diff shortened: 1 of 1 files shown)


parjong:
This PR includes #9607.

\CC @wateret

parjong:
@janvorli Please take a look.

src/vm/exceptionhandling.cpp
```diff
@@ -20,6 +20,11 @@
 #if defined(_TARGET_ARM_) || defined(_TARGET_ARM64_) || defined(_TARGET_X86_)
 #define ADJUST_PC_UNWOUND_TO_CALL
 #define USE_CALLER_SP_IN_FUNCLET
+// For ARM/ARM64, EstablisherFrame is Caller-SP (SP just before executing call instruction).
+// This has been confirmed by AaronGi from the kernel team for Windows.
+//
+// For x86/Linux, RtlVirtualUnwind sets EstablisherFrame as Caller-SP.
+#define ESTABLISHER_FRAME_POINTER_IS_CALLER_SP
```
parjong: Thanks for comment. I'll update PR as you suggested.

Issue #9818, opened 2017-02-27
[ARM32/Linux] Cross architecture build error in PAL assembly codes


hseok-oh:
build command:
./build.sh release cross arm crosscomponent

Error message
```
[ 25%] Building ASM object src/pal/src/CMakeFiles/coreclrpal.dir/arch/arm/context2.S.o
/nfs/rbp3/coreclr/src/pal/src/arch/arm/context2.S:13:1: error: unknown directive
.syntax unified
^
/nfs/rbp3/coreclr/src/pal/src/arch/arm/context2.S:14:1: error: unknown directive
.thumb
^

<errors in assembly code>

make[2]: *** [src/pal/src/CMakeFiles/coreclrpal.dir/arch/arm/context2.S.o] Error 1
make[1]: *** [src/pal/src/CMakeFiles/coreclrpal.dir/all] Error 2
```

It seems to releated to the change of src/pal/src/CMakeLists.txt in #9650. It tried to compile asm file for arm, but it must have tried to compile asm for x86.

gkhanna79:
@hseok-oh Can this be closed?

hseok-oh:
Fixed by #9819 and remain issue #9902

PR #9819, opened 2017-02-27
Fix cross component builderror - CMakeLists.txt for PAL

hseok-oh:
define & use PAL_ARCH_SOURCES_DIR in PAL's CMakeLists.txt. 
define PAL_ARCH_SOURCES_DIR based on platform architecture, not JIT target architecture.

### DIFF
--- a/src/pal/src/CMakeLists.txt
+++ b/src/pal/src/CMakeLists.txt
@@ -77,13 +77,17 @@ add_definitions(-D_FILE_OFFSET_BITS=64)
 if(PAL_CMAKE_PLATFORM_ARCH_AMD64)
   add_definitions(-DBIT64=1)
   add_definitions(-D_WIN64=1)
+  set(PAL_ARCH_SOURCES_DIR amd64)
 elseif(PAL_CMAKE_PLATFORM_ARCH_ARM)
   add_definitions(-DBIT32=1)
+  set(PAL_ARCH_SOURCES_DIR arm)
 elseif(PAL_CMAKE_PLATFORM_ARCH_ARM64)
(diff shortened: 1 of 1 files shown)


hseok-oh:
related issue: #9818

hseok-oh:

janvorli:
This fix is strange. It seems that rather the ARCH_SOURCES_DIR should be fixed. It doesn't make sense to compile PAL for different arch than the other runtime parts during the crosscomponent build and so the ARCH_SOURCES_DIR should reflect the platform we are targetting with the build.

hseok-oh:
It seems that PAL and other runtime part uses ARCH_SOURCES_DIR as different usage. For example, src/vm/arch is related to code generation. And some debug interfaces in src/debug uses to decode generated code or simulating. (I'm not familiar with debuggin interface). These parts uses ARCH_SOURCES_DIR as target architecture. On the other hand, it seems that ARCH_SOURCES_DIR is used for platform architecture in src/pal/src.

janvorli:
Ah, I have missed the fact that we don't have just asm files but we also have .cpp files in the arch specific directories and that some of those (but not all) are used for crossbuild tools. 
So it seems we should actually split the ARCH_SOURCES_DIR into two ones - one for the files that we use in the crossbuild and one for the ones that are only meant to run on the target (like the .asm files plus some of the cpp ones).
However, that would be a cleanup that's out of scope of this change. So I'll create a separate issue for that and merge your change as is.

hseok-oh:
@janvorli I complete agree with your opinion. Thanks.

PR #9820, opened 2017-02-27
[x86/Linux] Initial filter support

parjong:
Filters in x86/Linux require EE to set frame pointer before EE calls them, but it is impossible to pass correct frame pointer in the current implementation.

This commit revises EH subsystem to pass correct context for the first pass.

### DIFF
--- a/src/pal/inc/rt/palrt.h
+++ b/src/pal/inc/rt/palrt.h
@@ -1502,6 +1502,7 @@ typedef struct _DISPATCHER_CONTEXT {
     PRUNTIME_FUNCTION FunctionEntry;
     DWORD EstablisherFrame;
     DWORD TargetIp;
+    PCONTEXT CurrentRecord;
     PCONTEXT ContextRecord;
     PEXCEPTION_ROUTINE LanguageHandler;
     PVOID HandlerData;
--- a/src/pal/inc/rt/palrt.h
+++ b/src/pal/inc/rt/palrt.h
@@ -1502,7 +1502,7 @@ typedef struct _DISPATCHER_CONTEXT {
     PRUNTIME_FUNCTION FunctionEntry;
     DWORD EstablisherFrame;
     DWORD TargetIp;
-    PCONTEXT CurrentRecord;
+    PKNONVOLATILE_CONTEXT CurrentRecord;
     PCONTEXT ContextRecord;
     PEXCEPTION_ROUTINE LanguageHandler;
     PVOID HandlerData;
--- a/src/pal/inc/rt/palrt.h
+++ b/src/pal/inc/rt/palrt.h
@@ -1502,7 +1502,7 @@ typedef struct _DISPATCHER_CONTEXT {
     PRUNTIME_FUNCTION FunctionEntry;
     DWORD EstablisherFrame;
     DWORD TargetIp;
-    PKNONVOLATILE_CONTEXT CurrentRecord;
+    PKNONVOLATILE_CONTEXT CurrentNonVolatileContextRecord;
     PCONTEXT ContextRecord;
     PEXCEPTION_ROUTINE LanguageHandler;
     PVOID HandlerData;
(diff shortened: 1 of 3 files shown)


parjong:

parjong:

src/pal/inc/rt/palrt.h
```diff
@@ -1502,6 +1502,7 @@ typedef struct _DISPATCHER_CONTEXT {
     PRUNTIME_FUNCTION FunctionEntry;
     DWORD EstablisherFrame;
     DWORD TargetIp;
+    PCONTEXT CurrentRecord;
```
janvorli: It seems the parent function of the filter (the one that contains it) always has EBP frame, otherwise the filter funclet code would not be able to access the locals of the parent. If that's the case, we could just rely on the fact that the EBP is stored at the callerContext->SP - sizeof(TADDR) and we could just modify the GetFrameRestoreBase to return GetCallerSp(pContextRecord) - 2 * sizeof(TADDR) like we already do at several other places for EBP based frames and leave the rest of the code unchanged.

parjong: @janvorli Does it work for nested exceptions? As I understand, Caller SP will points to the frame by ``CallEHFunclet`` when nested exceptions happen.

janvorli: @parjong you are right, my mistake. 
The only downside of your change is that we unnecessarily copy more than 1kB of memory of the context twice (once here and once in the ExceptionTracker::InitializeCurrentContextForCrawlFrame) when we really need just the non-volatile registers. 
Could you please create a new structure called e.g. NonVolatileRegisters for x86 and use it instead of the CurrentRecord? That way, we would copy just a few registers instead of the whole CONTEXT.

src/vm/exceptionhandling.cpp
```diff
@@ -2892,7 +2906,7 @@ CLRUnwindStatus ExceptionTracker::ProcessManagedCallFrame(
                                 {
                                     // CallHandler expects to be in COOP mode.
                                     GCX_COOP();
-                                    dwResult = CallHandler(dwFilterStartPC, sf, &EHClause, pMD, Filter X86_ARG(pCurRegDisplay->pCallerContext) ARM_ARG(pCurRegDisplay->pCallerContext) ARM64_ARG(pCurRegDisplay->pCallerContext));
+                                    dwResult = CallHandler(dwFilterStartPC, sf, &EHClause, pMD, Filter X86_ARG(pContext) ARM_ARG(pContext) ARM64_ARG(pContext));
```
janvorli: I would prefer the former form and just change the X86_ARG(pCurRegDisplay->pCurrentContext). It seems more readable to me.

src/pal/inc/rt/palrt.h
parjong: Thanks for comment. I'll try that approach and update PR.

src/vm/exceptionhandling.cpp
parjong: I introduced this change to isolate all the changes related with ``USE_CURRENT_CONTEXT_IN_FILTER``. 

The former form requires us to take care of this line when there are changes related with ``USE_CURRENT_CONTEXT_IN_FILTER``.

I'll update PR if you prefer the former form, but could you please take a look of this issue?

janvorli: I was thinking about it and I understand your point. So let's keep it this way.

src/pal/inc/rt/palrt.h
```diff
@@ -1502,7 +1502,7 @@ typedef struct _DISPATCHER_CONTEXT {
     PRUNTIME_FUNCTION FunctionEntry;
     DWORD EstablisherFrame;
     DWORD TargetIp;
-    PCONTEXT CurrentRecord;
+    PKNONVOLATILE_CONTEXT CurrentRecord;
```
janvorli: A nit - could you please rename the CurrentRecord, which is a name that doesn't tell what it really represents to CurrentNonVolatileContextRecord or CurrentNonVolatileRegisters? Otherwise LGTM.

PR #9823, opened 2017-02-27
Use flags in exceptionhandling.cpp

### LINKED ISSUE #9822: [x86/Linux] GC.Coverage.delete_next_card_table test failure
The following assert failure is observed during x86/Linux CLR unittest:
```
FAILED   - [ 075 :    7 sec ] GC/Coverage/delete_next_card_table/delete_next_card_table.sh
               BEGIN EXECUTION
               /root/dotnet-test/coreclr_unittest_20170227_155439_x86_x86-emu/coreclr-unittest-170227-065508/overlay/corerun delete_next_card_table.exe
               Churn loop 0...........................OOM while Churning
               
               Assert failure(PID 17634 [0x000044e2], Thread: 17634 [0x44e2]): false && "ExceptionTracker::HasFrameBeenUnwoundByAnyActiveException"
                   File: /home/parjong/projects/dotnet/coreclr/src/vm/exceptionhandling.cpp Line: 6302
                   Image: /root/dotnet-test/coreclr_unittest_20170227_155439_x86_x86-emu/coreclr-unittest-170227-065508/overlay/corerun
               
               ./delete_next_card_table.sh: line 138: 17634 Aborted                 (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" delete_next_card_table.exe $CLRTestExecutionArguments
               Expected: 100
               Actual: 134
               END EXECUTION - FAILED
```

parjong:
This commit introduces two flags in exceptionhandling.cpp:
 - USE_FUNCLET_CALL_HELPER for call helper-related code
 - STACK_RANGE_BOUNDS_ARE_CALLER_SP for m_EHClauseInfo-related checks.

This commit fixes #9822.

### DIFF
--- a/src/vm/exceptionhandling.cpp
+++ b/src/vm/exceptionhandling.cpp
@@ -19,6 +19,7 @@
 
 #if defined(_TARGET_ARM_) || defined(_TARGET_ARM64_) || defined(_TARGET_X86_)
 #define ADJUST_PC_UNWOUND_TO_CALL
+#define STACK_RANGE_BOUNDS_ARE_CALLER_SP
 #define USE_FUNCLET_CALL_HELPER
 #define USE_CALLER_SP_IN_FUNCLET
 #endif // _TARGET_ARM_ || _TARGET_ARM64_ || _TARGET_X86_
@@ -6187,11 +6188,11 @@ bool ExceptionTracker::IsInStackRegionUnwoundBySpecifiedException(CrawlFrame * p
 
     // Remember that sfLowerBound and sfUpperBound are in the "OS format".
     // Refer to the comment for CallerStackFrame for more information.
-#ifndef USE_FUNCLET_CALL_HELPER
+#ifndef STACK_RANGE_BOUNDS_ARE_CALLER_SP
     if ((sfLowerBound < csfToCheck) && (csfToCheck <= sfUpperBound))
-#else // !USE_FUNCLET_CALL_HELPER
+#else // !STACK_RANGE_BOUNDS_ARE_CALLER_SP
     if ((sfLowerBound <= csfToCheck) && (csfToCheck < sfUpperBound))
-#endif // USE_FUNCLET_CALL_HELPER
+#endif // STACK_RANGE_BOUNDS_ARE_CALLER_SP
     {
         return true;
     }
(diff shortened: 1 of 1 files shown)


src/vm/exceptionhandling.cpp
```diff
@@ -6184,11 +6187,11 @@ bool ExceptionTracker::IsInStackRegionUnwoundBySpecifiedException(CrawlFrame * p
 
     // Remember that sfLowerBound and sfUpperBound are in the "OS format".
     // Refer to the comment for CallerStackFrame for more information.
-#if defined(_TARGET_AMD64_)
+#ifndef USE_FUNCLET_CALL_HELPER
```
janvorli: Based on the comments above, this check differs based on whether the bounds are initialized from the caller SP or the SP of the frame itself. So it seems to me that using USE_FUNCLET_CALL_HELPER is a bit misleading. While it is true now that the usage of call helper matches the difference between using caller SP or SP of the frame itself, these things are not necessarily related. 
So I would prefer keeping the bounds checks here and below around likes 6282 and 6308 conditionalized by the target.

parjong: @janvorli While digging this issue, I found that the major difference between AMD64 and ARM/ARM64 is that AMD64 uses the result of ``GetCurrentSP()`` called inside ``CallHandler`` to set ``m_EHClauseInfo``, but ARM/ARM64 updates it inside ``CallEHFunclet`` and ``CallEHFilterFunclet`` later (this is why I used ``USE_FUNCLET_CALL_HELPER`` here).

I agree with you that these things are not necessarily related, but then I would like to introduce another flag for these checks (for me, it was very difficult to find these checks are related when it was ifdefed by targets).

Could you let me know your opinion?

janvorli: I don't feel like the fact where we get the current SP is related to the stack range bounds. We get it inside of the asm helper just because we need the value at the caller of the funclet and the caller is the asm helper, while in the other case the caller is directly the CallHandler. Hence the different place we capture it.
It sounds it would be ok though to introduce a new ifdef, something like STACK_RANGE_BOUNDS_ARE_CALLER_SP if you feel it makes the conditions easier to understand.

parjong: @janvorli Thank you for comment. I'll update PR as suggested.

parjong: @janvorli FYI, I felt like the above due to the current implementation of ``GetCurrentSP`` (for x86). 

I checked the corresponding implementation in AMD64, and found that it is a bit different. 
(I submitted #9838 to fix this).

PR #5939, opened 2016-06-23
Fix RyuJIT/x86 P/Invoke inlining

BruceForstall:
Enable RyuJIT/x86 PInvoke lowering

Fixes #4181 "NYI_X86: Implement PInvoke frame init inlining for x86"

The main work here is to handle the custom calling convention for the
x86 CORINFO_HELP_INIT_PINVOKE_FRAME helper call: it takes EDI as an argument,
trashes only EAX, and returns the TCB in ESI.

The code changes are as follows:
1. Lowering::InsertPInvokeMethodProlog(): don't pass the "secret stub param" for x86.
Also, don't store the InlinedCallFrame.m_pCallSiteSP in the prolog: for x86 this is done
at the call site, due to the floating stack pointer.
2. LinearScan::getKillSetForNode(): for helper calls, call compHelperCallKillSet() to get the killMask,
to account for non-standard kill sets.
3. Morph.cpp::fgMorphArgs(): set non-standard arguments for CORINFO_HELP_INIT_PINVOKE_FRAME.
4. compHelperCallKillSet(): set the correct kill set for CORINFO_HELP_INIT_PINVOKE_FRAME.
5. codegenxarch.cpp::genCallInstruction(): set the ABI return register for CORINFO_HELP_INIT_PINVOKE_FRAME.
6. lowerxarch.cpp::TreeNodeInfoInit(): set the GT_CALL dstCandidates for CORINFO_HELP_INIT_PINVOKE_FRAME.

5 & 6 are both needed to avoid a copy.

With this change, the `#1` NYI with 18415 hits over the tests is gone.
The total number of NYI is now 29516.

PR #9825, opened 2017-02-27
[RyuJIT/ARM32] P/Invoke support

mskvortsov:
This PR enables p/invoke support in lowering, regalloc and codegen. It's almost a copy of #5939.

Tested in both chk/rel modes on the family of `DomainNeutralILStubClass:IL_STUB_PInvoke` methods from CoreLib, as well as on some of the `JIT/Directed/pinvoke` tests. 

(This is a part of #8496.)

### DIFF
--- a/src/jit/codegenarm.cpp
+++ b/src/jit/codegenarm.cpp
@@ -1513,7 +1513,13 @@ void CodeGen::genCallInstruction(GenTreePtr node)
         }
         else
         {
-            if (varTypeIsFloating(returnType))
+            if (call->IsHelperCall(compiler, CORINFO_HELP_INIT_PINVOKE_FRAME))
+            {
+                // The CORINFO_HELP_INIT_PINVOKE_FRAME helper uses a custom calling convention that returns with
+                // TCB in REG_PINVOKE_TCB. fgMorphCall() sets the correct argument registers.
+                returnReg = REG_PINVOKE_TCB;
+            }
+            else if (varTypeIsFloating(returnType))
             {
                 returnReg = REG_FLOATRET;
             }
--- a/src/jit/lower.cpp
+++ b/src/jit/lower.cpp
@@ -2840,11 +2840,11 @@ void Lowering::InsertPInvokeMethodProlog()
     // for x86, don't pass the secretArg.
     CLANG_FORMAT_COMMENT_ANCHOR;
 
-#ifdef _TARGET_X86_
+#if defined(_TARGET_X86_) || defined(_TARGET_ARM_)
     GenTreeArgList* argList = comp->gtNewArgList(frameAddr);
-#else  // !_TARGET_X86_
+#else
     GenTreeArgList*    argList = comp->gtNewArgList(frameAddr, PhysReg(REG_SECRET_STUB_PARAM));
-#endif // !_TARGET_X86_
+#endif
 
     GenTree* call = comp->gtNewHelperCallNode(CORINFO_HELP_INIT_PINVOKE_FRAME, TYP_I_IMPL, 0, argList);
 
@@ -2865,8 +2865,9 @@ void Lowering::InsertPInvokeMethodProlog()
     firstBlockRange.InsertBefore(insertionPoint, LIR::SeqTree(comp, store));
     DISPTREERANGE(firstBlockRange, store);
 
-#ifndef _TARGET_X86_ // For x86, this step is done at the call site (due to stack pointer not being static in the
-                     // function).
+#if !defined(_TARGET_X86_) && !defined(_TARGET_ARM_)
+    // For x86, this step is done at the call site (due to stack pointer not being static in the function).
+    // For arm32, CallSiteSP is set up by the call to CORINFO_HELP_INIT_PINVOKE_FRAME.
 
     // --------------------------------------------------------
     // InlinedCallFrame.m_pCallSiteSP = @RSP;
@@ -2878,7 +2879,10 @@ void Lowering::InsertPInvokeMethodProlog()
     firstBlockRange.InsertBefore(insertionPoint, LIR::SeqTree(comp, storeSP));
     DISPTREERANGE(firstBlockRange, storeSP);
 
-#endif // !_TARGET_X86_
+#endif // !defined(_TARGET_X86_) && !defined(_TARGET_ARM_)
+
+#if !defined(_TARGET_ARM_)
+    // For arm32, CalleeSavedFP is set up by the call to CORINFO_HELP_INIT_PINVOKE_FRAME.
 
     // --------------------------------------------------------
     // InlinedCallFrame.m_pCalleeSavedEBP = @RBP;
@@ -2890,6 +2894,7 @@ void Lowering::InsertPInvokeMethodProlog()
 
     firstBlockRange.InsertBefore(insertionPoint, LIR::SeqTree(comp, storeFP));
     DISPTREERANGE(firstBlockRange, storeFP);
+#endif // !defined(_TARGET_ARM_)
 
     // --------------------------------------------------------
     // On 32-bit targets, CORINFO_HELP_INIT_PINVOKE_FRAME initializes the PInvoke frame and then pushes it onto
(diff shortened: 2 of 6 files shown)


mskvortsov:

BruceForstall:
cc @dotnet/jit-contrib

src/jit/morph.cpp
```diff
@@ -2881,7 +2881,7 @@ GenTreeCall* Compiler::fgMorphArgs(GenTreeCall* callNode)
         CLANG_FORMAT_COMMENT_ANCHOR;
 
 #if !defined(LEGACY_BACKEND)
-#if defined(_TARGET_X86_)
+#if defined(_TARGET_X86_) || defined(_TARGET_ARM_)
         // The x86 CORINFO_HELP_INIT_PINVOKE_FRAME helper has a custom calling convention. Set the argument registers
```
pgavlin: Nit: this comment could use an update to indicate that it also applies to ARM32.

pgavlin:
LGTM aside from one nit.

src/jit/lower.cpp
```diff
@@ -2862,8 +2862,8 @@ void Lowering::InsertPInvokeMethodProlog()
     firstBlockRange.InsertBefore(insertionPoint, LIR::SeqTree(comp, store));
     DISPTREERANGE(firstBlockRange, store);
 
-#ifndef _TARGET_X86_ // For x86, this step is done at the call site (due to stack pointer not being static in the
-                     // function).
+#if !defined(_TARGET_X86_) && !defined(_TARGET_ARM_)
+    // For x86 and arm32, this step is done at the call site (due to stack pointer not being static in the function).
```
BruceForstall: This doesn't apply to arm32, since arm32 has a static SP throughout the function. Thus, offsetOfCallSiteSP for arm32 should be handled the same as amd64 (set in function prolog, not set at pinvoke call site).

src/jit/lower.cpp
```diff
@@ -2875,8 +2875,9 @@ void Lowering::InsertPInvokeMethodProlog()
     firstBlockRange.InsertBefore(insertionPoint, LIR::SeqTree(comp, storeSP));
     DISPTREERANGE(firstBlockRange, storeSP);
 
-#endif // !_TARGET_X86_
+#endif // !defined(_TARGET_X86_) && !defined(_TARGET_ARM_)
 
+#if !defined(_TARGET_ARM_)
     // --------------------------------------------------------
```
BruceForstall: Why is this not for ARM32? It looks like legacy back end sets this.

src/jit/codegenarm.cpp
```diff
@@ -986,7 +986,7 @@ void CodeGen::genCodeForTreeNode(GenTreePtr treeNode)
         case GT_LABEL:
             genPendingCallLabel       = genCreateTempLabel();
             treeNode->gtLabel.gtLabBB = genPendingCallLabel;
-            emit->emitIns_R_L(INS_lea, EA_PTRSIZE, genPendingCallLabel, treeNode->gtRegNum);
+            emit->emitIns_J_R(INS_adr, EA_PTRSIZE, genPendingCallLabel, treeNode->gtRegNum);
             break;
```
BruceForstall: Is this fix unrelated to the pinvoke change? If so, maybe you should submit it separately.

src/jit/lower.cpp
mskvortsov: New output without that `#ifdef` adds a store not present in the original legacy output (which is [here](https://gist.githubusercontent.com/mskvortsov/bcdc1d731c8c7c72ead647267eb9802f/raw/1d0146a8f9fbe21d43d91d67fad73eff74318611/IL_STUB_PInvoke.legacy.txt)):
<pre>
G_M40565_IG02:
000014  AC02           add     r4, sp, 8
000016  F24D 0325      movw    r3, 0xd025
00001A  F2CB 53CA      movt    r3, 0xb5ca
00001E  4798           blx     r3               // CORINFO_HELP_INIT_PINVOKE_FRAME
<b>000020  4658           mov     r0, r11
000022  9007           str     r0, [sp+0x1c]</b>
000024  2000           movs    r0, 0
000026  2F00           cmp     r7, 0
000028  D002           beq     SHORT G_M40565_IG03
00002A  9709           str     r7, [sp+0x24]    // [V03 loc2]
00002C  F107 0008      add     r0, r7, 8

G_M40565_IG03:
000030  9B08           ldr     r3, [sp+0x20]    // [V04 tmp0]
000032  6A9B           ldr     r3, [r3+40]
000034  681B           ldr     r3, [r3]
000036  2400           movs    r4, 0
000038  2200           movs    r2, 0
00003A  9204           str     r2, [sp+0x10]    // [V07 PInvokeFrame+0x0c]
00003C  F20F 0209      adr     r2, G_M40565_IG05
000040  9206           str     r2, [sp+0x18]    // [V07 PInvokeFrame+0x14]
000042  2200           movs    r2, 0
000044  722A           strb    r2, [r5+8]

G_M40565_IG04:
000046  4798           blx     r3
</pre>
`genPInvokeMethodProlog` of `codegenlegacy.cpp` has the following fragment:
```c++
#if defined(_TARGET_X86_)
    if (threadTlsIndex == (DWORD)-1 || pInfo->osType != CORINFO_WINNT)
#else
    if (true)
#endif
    {
        ...
        // We are done, the rest of this function deals with the inlined case.
        return initRegs;
    }
```
therefore the code which sets `offsetOfCalleeSavedFP` near the end of the function doesn't get executed:
```c++
    /* set EBP value in frame */
    getEmitter()->emitIns_S_R(ins_Store(TYP_I_IMPL), EA_PTRSIZE, genFramePointerReg(),
                              compiler->lvaInlinedPInvokeFrameVar, pInfo->inlinedCallFrameInfo.offsetOfCalleeSavedFP);
```

src/jit/lower.cpp
mskvortsov: I've experimented to remove both these my changes about offsetOfCallSiteSP and offsetOfCalleeSavedFP, and it still works. But then I wonder how legacy backend works, because I can't see any analogous stores in it's output:
```
000006  F10D 0B60      add     r11, sp, 96
...
000030  4668           mov     r0, sp
000032  9006           str     r0, [sp+0x18]
000034  4658           mov     r0, r11
000036  9008           str     r0, [sp+0x20]
```

BruceForstall: What test case / function is this? Can you show both legacy arm32 and ryujit arm32 code?

mskvortsov: Any of `DomainNeutralILStubClass:IL_STUB_PInvoke`. Legacy output is here on github in my neighbour comment; full RyuJIT's one I'll be able to show only tomorrow...

It seems that the approach to make ryujit output identical to the legacy one is not the best for every case.

src/jit/lower.cpp
BruceForstall: Ok, I missed that "return initRegs". Yes, looks like the setting of calleeSavedFP is not necessary for ARM.

---
In reply to: [103448440](https://github.com/dotnet/coreclr/pull/9825#discussion_r103448440) [](ancestors = 103448440)

src/jit/lower.cpp
BruceForstall: Ok, I see that your #ifdefs are correct, as this is not needed for ARM32 because it is set up by the call to `CORINFO_HELP_INIT_PINVOKE_FRAME`. However, the comment is confusing: the reason we set RSP here for non-x86 is because RSP is static in the function. Can you update the comment to be correct for ARM32?

---
In reply to: [103531744](https://github.com/dotnet/coreclr/pull/9825#discussion_r103531744) [](ancestors = 103531744)

BruceForstall: > It seems that the approach to make ryujit output identical to the legacy one is not the best for every case.

That may be true for many things, but for many things, especially p/invoke prolog/epilog and call prolog/epilog, the code is so stylized that it is useful to either make it the same or understand the differences.

---
In reply to: [103537865](https://github.com/dotnet/coreclr/pull/9825#discussion_r103537865) [](ancestors = 103537865,103531744)

BruceForstall:
LGTM

mskvortsov:
@BruceForstall Thanks for the thorough review and suggestions!

PR #9831, opened 2017-02-27
CoreFx #16274 Implementation - New overloads of LazyInitializer.EnsureInitialized for reference types

WinCPP:
This PR is for the implementation of new overloads of LazyInitializer.EnsureInitialized for reference types. Please refer to https://github.com/dotnet/corefx/issues/16274 for more information.

@stephentoub @terrajobst @karelz Kindly review the changes.

### DIFF
--- a/src/mscorlib/src/System/Threading/LazyInitializer.cs
+++ b/src/mscorlib/src/System/Threading/LazyInitializer.cs
@@ -82,16 +82,8 @@ namespace System.Threading
         /// if an object was not used and to then dispose of the object appropriately.
         /// </para>
         /// </remarks>
-        public static T EnsureInitialized<T>(ref T target) where T : class
-        {
-            // Fast path.
-            if (Volatile.Read<T>(ref target) != null)
-            {
-                return target;
-            }
-
-            return EnsureInitializedCore<T>(ref target, LazyHelpers<T>.s_activatorFactorySelector);
-        }
+        public static T EnsureInitialized<T>(ref T target) where T : class =>
+            Volatile.Read<T>(ref target) ?? EnsureInitializedCore<T>(ref target, LazyHelpers<T>.s_activatorFactorySelector);
 
         /// <summary>
         /// Initializes a target reference type using the specified function if it has not already been
@@ -121,16 +113,8 @@ namespace System.Threading
         /// if an object was not used and to then dispose of the object appropriately.
         /// </para>
         /// </remarks>
-        public static T EnsureInitialized<T>(ref T target, Func<T> valueFactory) where T : class
-        {
-            // Fast path.
-            if (Volatile.Read<T>(ref target) != null)
-            {
-                return target;
-            }
-
-            return EnsureInitializedCore<T>(ref target, valueFactory);
-        }
+        public static T EnsureInitialized<T>(ref T target, Func<T> valueFactory) where T : class =>
+            Volatile.Read<T>(ref target) ?? EnsureInitializedCore<T>(ref target, valueFactory);
 
         /// <summary>
         /// Initialize the target using the given delegate (slow path).
@@ -151,7 +135,6 @@ namespace System.Threading
             Debug.Assert(target != null);
             return target;
         }
-
 
         /// <summary>
         /// Initializes a target reference or value type with its default constructor if it has not already
@@ -198,9 +181,31 @@ namespace System.Threading
                 return target;
             }
 
-
             return EnsureInitializedCore<T>(ref target, ref initialized, ref syncLock, valueFactory);
         }
+
+        /// <summary>
+        /// Ensure the lock object is intialized.
+        /// </summary>
+        /// <param name="syncLock">A reference to a location containing a mutual exclusive lock. If <paramref name="syncLock"/> is null,
+        /// a new object will be instantiated.</param>
+        /// <returns>Initialized lock object.</returns>
+        private static object EnsureLockInitialized(ref object syncLock) =>
+            syncLock ??
+            Interlocked.CompareExchange(ref syncLock, new object(), null) ??
+            syncLock;
+
+        /// <summary>
+        /// Initializes a target reference type with a specified function if it has not already been initialized.
+        /// </summary>
+        /// <typeparam name="T">The type of the reference to be initialized. Has to be reference type.</typeparam>
+        /// <param name="target">A reference of type <typeparamref name="T"/> to initialize if it has not already been initialized.</param>
+        /// <param name="syncLock">A reference to an object used as the mutually exclusive lock for initializing
+        /// <paramref name="target"/>. If <paramref name="syncLock"/> is null, a new object will be instantiated.</param>
+        /// <param name="valueFactory">The <see cref="T:System.Func{T}"/> invoked to initialize the reference.</param>
+        /// <returns>The initialized value of type <typeparamref name="T"/>.</returns>
+        public static T EnsureInitialized<T>(ref T target, ref object syncLock, Func<T> valueFactory) where T : class =>
+            Volatile.Read(ref target) ?? EnsureInitializedCore<T>(ref target, ref syncLock, valueFactory);
 
         /// <summary>
         /// Ensure the target is initialized and return the value (slow path). This overload permits nulls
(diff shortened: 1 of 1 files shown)


src/mscorlib/src/System/Threading/LazyInitializer.cs
```diff
@@ -198,11 +197,49 @@ public static T EnsureInitialized<T>(ref T target, ref bool initialized, ref obj
...
 
         /// <summary>
+        /// Initializes a target reference with its default constructor if it has not already been initialized.
+        /// </summary>
+        /// <typeparam name="T">The type of the reference to be initialized. Has to be reference type.</typeparam>
+        /// <param name="target">A reference type <typeparamref name="T"/> to initialize if it has not already been initialized.</param>
+        /// <param name="syncLock">A reference to an object used as the mutually exclusive lock for initializing
+        /// <paramref name="target"/>. If <paramref name="syncLock"/> is null, a new object will be instantiated.</param>
+        /// <returns>The initialized value of type <typeparamref name="T"/>.</returns>
+        public static T EnsureInitialized<T>(ref T target, ref object syncLock) where T : class
```
stephentoub: I don't believe this API was discussed or approved, was it?

src/mscorlib/src/System/Threading/LazyInitializer.cs
```diff
@@ -241,6 +278,48 @@ private static T EnsureInitializedCore<T>(ref T target, ref bool initialized, re
...
+                    slock = newLock;
+                }
+            }
+
+            // Now double check that initialization is still required.
+            lock (slock)
+            {
+                if (Volatile.Read(ref target) == null)
+                {
+                    target = valueFactory();
```
stephentoub: This should be: ```Volatile.Write(ref target, valueFactory());```

src/mscorlib/src/System/Threading/LazyInitializer.cs
WinCPP: It was discussed but not concluded ([here](https://github.com/dotnet/corefx/issues/16274#issuecomment-282481439))... that is why I left the comment above ([here](https://github.com/dotnet/coreclr/pull/9831#commitcomment-21051691)). Request you to consider the same as it makes the API symmetric...

karelz: @WinCPP any API change has to be first concluded and approved - it is best done in issues. PRs are bad place to do API review discussions. Please revert it for now and restart the discussion in the issue (or create a new one).

WinCPP: removing...

src/mscorlib/src/System/Threading/LazyInitializer.cs
WinCPP: Changing this. It will force the write of target.

src/mscorlib/src/System/Threading/LazyInitializer.cs
WinCPP: @karelz our comments crossed... yes I realize it is not a good idea. I am removing the API. Build and test is in progress and will push the same.

karelz: @WinCPP perfect, thank you!

WinCPP:
@stephentoub , the changes are in. Sorry.

src/mscorlib/src/System/Threading/LazyInitializer.cs
```diff
@@ -198,11 +197,30 @@ public static T EnsureInitialized<T>(ref T target, ref bool initialized, ref obj
...
+        public static T EnsureInitialized<T>(ref T target, ref object syncLock, Func<T> valueFactory) where T : class
+        {
+            // Fast path.
+            if (Volatile.Read(ref target) != null)
+            {
+                return target;
+            }
+            
+            return EnsureInitializedCore<T>(ref target, ref syncLock, valueFactory);
+        }
```
stephentoub: Nit: I realize you're just copying the style used above; that's fine.  We could subsequently shorten / clean this up as:
```C#
public static T EnsureInitialized<T>(ref T target, ref object syncLock, Func<T> valueFactory) where T : class =>
    Volatile.Read(ref target) ??
    EnsureInitializedCore<T>(ref target, ref syncLock, valueFactory);
```

WinCPP: Done!

src/mscorlib/src/System/Threading/LazyInitializer.cs
```diff
@@ -241,6 +235,48 @@ private static T EnsureInitializedCore<T>(ref T target, ref bool initialized, re
...
+            object slock = syncLock;
+            if (slock == null)
+            {
+                object newLock = new object();
+                slock = Interlocked.CompareExchange(ref syncLock, newLock, null);
+                if (slock == null)
+                {
+                    slock = newLock;
+                }
+            }
```
stephentoub: Since this code is now duplicated in two different methods, could you separate it out into a helper? e.g.
```C#
private static object EnsureLockInitialized(ref object syncLock) =>
    syncLock ??
    Interlocked.CompareExchange(ref syncLock, new object(), null) ??
    syncLock;
```
which you could then use here and in the other function as:
```C#
lock (EnsureLockInitialized(ref syncLock))
{
    ...
}
```
?

WinCPP: Yup. Makes sense. Actually out of habit, I was into the groove of 'don't touch something which already works; follow the style'...

WinCPP:
@stephentoub the changes are ready...

src/mscorlib/src/System/Threading/LazyInitializer.cs
```diff
@@ -217,25 +219,47 @@ public static T EnsureInitialized<T>(ref T target, ref bool initialized, ref obj
...
+        /// <param name="valueFactory">
+        /// The <see cref="T:System.Func{T}"/> to invoke in order to produce the lazily-initialized value.
+        /// </param>
+        /// <returns>The initialized object.</returns>
+        private static T EnsureInitializedCore<T>(ref T target, ref object syncLock, Func<T> valueFactory) where T : class
+        {
+            object slock = syncLock;
+
+            // Lazily initialize the lock if necessary and,  then double check if initialization is still required.
+            lock (EnsureLockInitialized(ref slock))
```
stephentoub: This is not correct.  It needs to be:
```C#
lock (EnsureLockInitialized(ref syncLock))
```
By initializing the local, racing threads will all end up creating and using their own slock object.

src/mscorlib/src/System/Threading/LazyInitializer.cs
```diff
@@ -217,25 +219,47 @@ public static T EnsureInitialized<T>(ref T target, ref bool initialized, ref obj
         /// <returns>The initialized object.</returns>
         private static T EnsureInitializedCore<T>(ref T target, ref bool initialized, ref object syncLock, Func<T> valueFactory)
         {
-            // Lazily initialize the lock if necessary.
             object slock = syncLock;
-            if (slock == null)
+
+            // Lazily initialize the lock if necessary and,  then double check if initialization is still required.
+            lock (EnsureLockInitialized(ref slock))
```
stephentoub: Same problem.  This needs to be ```ref syncLock```, not ```ref slock```.  You can delete the local slock.

src/mscorlib/src/System/Threading/LazyInitializer.cs
```diff
@@ -198,11 +181,30 @@ public static T EnsureInitialized<T>(ref T target, ref bool initialized, ref obj
...
             return EnsureInitializedCore<T>(ref target, ref initialized, ref syncLock, valueFactory);
         }
 
         /// <summary>
+        /// Ensure the lock object is intialized.
+        /// </summary>
+        /// <param name="syncLock">A reference to a location containing a mutual exclusive lock. If <paramref name="syncLock"/> is null,
+        /// a new object will be instantiated.</param>
+        /// <returns>Initialized lock object.</returns>
+        private static object EnsureLockInitialized(ref object syncLock) => syncLock ?? Interlocked.CompareExchange(ref syncLock, new object(), null) ?? syncLock;
```
stephentoub: Nit: please split this across multiple lines to make it easier to read.

src/mscorlib/src/System/Threading/LazyInitializer.cs
WinCPP: Yes, agreed. This is because of oversight. My test cases ran successfully because I added one for non-null syncLock to the new LazyInitializer.EnsureInitialized API and was happy with its success. I will add another test case for scenario that the user sends in a null syncLock which would have set off the race condition.

src/mscorlib/src/System/Threading/LazyInitializer.cs
WinCPP: Done!

stephentoub:
@jkotas, does the mscorlib.cs ref still need to be updated to include new APIs, or is that now defunct?

jkotas:
mscorlib.cs ref is defunct. We can delete it once the CoreCLR tests are fixed to consume live corefx - they are still stuck on corefx before the corefx build changes.

stephentoub:
Thanks.

PR #9833, opened 2017-02-27
Don't generate a runtime lookup for static base if not needed

MichalStrehovsky:
`fgInitThisClass` makes sure that the type initializer for the type that
owns the method being compiled has run. On CoreCLR, the fallthrough code
somehow ends up not doing a runtime lookup if the owning type is not
canonical. We need to avoid it on CoreRT as well.

The other option would be to move the check a bit up (so that we don't
duplicate the call to `fgGetSharedCCtor`, but that would be a change in
behavior on CoreCLR too and I don't know if it's safe.

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -16289,18 +16289,19 @@ GenTreePtr Compiler::fgInitThisClass()
         // Only CoreRT understands CORINFO_HELP_READYTORUN_GENERIC_STATIC_BASE. Don't do this on CoreCLR.
         if (opts.IsReadyToRun() && IsTargetAbi(CORINFO_CORERT_ABI))
         {
-            // We might be in a shared method body, but maybe we don't need a runtime lookup after all.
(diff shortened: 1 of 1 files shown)


MichalStrehovsky:
@dotnet/jit-contrib PTAL

Cc @sergiy-k

AndyAyersMS:
For R2R, `fgGetSharedCCtor` just calls back into `impReadyToRunHelperToTree` with nulls for the two last args ... seems like you could just specialize all this in place?

MichalStrehovsky:
> For R2R, `fgGetSharedCCtor` just calls back into `impReadyToRunHelperToTree` with nulls for the two last args ... seems like you could just specialize all this in place?

True - it looks better inlined.

AndyAyersMS:
Nit: the class attributes for the method being jitted are cached (just like the class handle which you use subsequently) -- so you should be able to use `info.compClassAttr` instead of re-fetching these attributes.

Otherwise LGTM.

PR #9834, opened 2017-02-27
Some tests for devirtualization

AndyAyersMS:


AndyAyersMS:
Requires #9230 before for any devirtualization can happen. 

Within-test comments reflect current state of that PR.

@JosephTremoulet PTAL
cc @dotnet/jit-contrib

tests/src/JIT/opt/Devirtualization/exacttype.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        return new Derived();
+    }
+
+    public static int Main()
+    {
+        Derived d = new Derived();
+        d.Foo();
+
+        M().Foo();
```
JosephTremoulet: Are you counting on M() getting inlined?

tests/src/JIT/opt/Devirtualization/enum.cs
```diff
@@ -0,0 +1,18 @@
...
+enum MyEnum { One, Two, Three, Four, Five, Six, Seven, Eight, Nine, Ten };
+
+// ToString should be devirtualizable
+
+public class Test
+{
+    public static int Main()
+    {
+        string s = (MyEnum.Seven).ToString();
+        return (s.Length == "Seven".Length ? 100 : -1);
```
JosephTremoulet: Why not `s == "Seven"`?

tests/src/JIT/opt/Devirtualization/exacttype.cs
```diff
@@ -0,0 +1,46 @@
...
+// since the exact type is known.
+
+public class Test
+{
+    public static Base M()
+    {
+        return new Derived();
+    }
+
+    public static int Main()
```
JosephTremoulet: Maybe add `new Base().Foo()` for newobj case (or is that in another test already)?

tests/src/JIT/opt/Devirtualization/inlinedevirt.cs
```diff
@@ -0,0 +1,20 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+class ImprovedType
+{
+    // Jit should inline this method and then devirtualize ToString()
```
JosephTremoulet: 👍

tests/src/JIT/opt/Devirtualization/sealedclass.cs
```diff
@@ -0,0 +1,38 @@
...
+    public override int GetValue(int value)
+    {
+        return value;
+    }
+}
+
+public class F
+{
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public static int TestSealedMethodInlining(Derived obj)
```
JosephTremoulet: I take it from the name of this method that you're not just expecting the `GetValue` call to be devirtualized, but also that it subsequently gets inlined?

tests/src/JIT/opt/Devirtualization/sealeddefault.cs
```diff
@@ -0,0 +1,28 @@
...
+    public virtual int Foo() { return 33; }
+    
+    static BaseSealed s_Default = new BaseSealed();
+
+    public static Base Default => s_Default;
+}
+
+sealed class BaseSealed : Base {}
+
+// The jit ought to be able to devirtualize the call to b.Foo below.
```
JosephTremoulet: Does "ought to" mean "in the future should", or "can now"?  If it can now, that's neat, I was expecting the stloc/ldloc through b to block it.  If it can't now but should in the future, then the comment confused me since the comments you've added elsewhere describe the current state.

(where by "current" and "now" I mean with your not-yet-merged devirtualization changes)

tests/src/JIT/opt/Devirtualization/sealedmethod.cs
```diff
@@ -0,0 +1,38 @@
...
+    public sealed override int GetValue(int value)
+    {
+        return value;
+    }
+}
+
+public class F
+{
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public static int TestSealedMethodInlining(Derived obj)
```
JosephTremoulet: Same as above, you're looking for inlining on top of devirtualization?

tests/src/JIT/opt/Devirtualization/sealeddefault.cs
AndyAyersMS: Will update;  `Base.Default.Foo()` will devirtualize, but not, as you note, `x.Foo()` currently won't.

tests/src/JIT/opt/Devirtualization/exacttype.cs
AndyAyersMS: Good idea, will update.

tests/src/JIT/opt/Devirtualization/sealedclass.cs
AndyAyersMS: Yes. If we devirtualize early (in the importer) then we can also inline.

tests/src/JIT/opt/Devirtualization/sealedmethod.cs
AndyAyersMS: Yep.

tests/src/JIT/opt/Devirtualization/enum.cs
AndyAyersMS: This example (and some others here) come from #1166. 

`s.Equals("Seven")` might be more interesting since it should also devirtualize. Will look into this.

tests/src/JIT/opt/Devirtualization/exacttype.cs
AndyAyersMS: Yes, this is the late devirtualization case where substituting in the actual return value expression can improve the return type. This currently (and likely always will) require inlining.

AndyAyersMS:
As far as how to avoid regressions, I don't know of any real satisfactory solutions.

I have thought about enhancing the inline tree to more generally describe the decision making around all calls, for instance, showing which calls end up as tail calls (or loop opt tail calls), which calls got devirtualized, and so on. Then one could imagine potentially using something like that as the basis for a regression test baseline. However this would be a debug/check feature only and it would need correlation artifacts to resolve call site identities that might not be stable with framework/csc changes.

Pattern match driven assembly language checks like the ones LLVM uses might be interesting. It would be more compelling if they were doable in release too.

Or we could try and synthesize perf cases but the feedback loop from a perf test case regression to root cause is currently fairly lengthy.

tests/src/JIT/opt/Devirtualization/enum.cs
JosephTremoulet: I think my comment may have been confusing (or I'm just not understanding your reply).  I was just wondering why you predicated reporting success on getting the length right, instead of the stronger condition of getting the string right.  Not a big concern, as I don't really expect we'll have a bug where we accidentally devirtualize/inline `(MyEnum.Seven).ToString()` into something that returns "Three" or "Eight", but I was just curious if there was a reason you'd written it this way.

AndyAyersMS: All I meant was that I took the code snippet from #1166 and used it directly without thinking much about the second line, since the first line was the interesting part.

I updated the second line to force use of the virtual Equals so it's now interesting too.

PR #9836, opened 2017-02-27
[Local GC] Decouple the EE's usage of the generation table from the GC

swgillespie:
This addresses an issue that arose in my last PR (https://github.com/dotnet/coreclr/pull/9255), in that the EE uses the GC's generation table directly when not using allocation contexts (which, today, means running on a single-proc Windows machine). The offsets that the assembly allocation helpers are using into the generation table are the `alloc_ptr` and `alloc_limit` fields of the Gen 0 allocation context.

The approach of this PR is similar to that of the other PRs that detangle EE and GC globals - upon initialization, the GC calls `stomp_write_barrier_initialize`, which in turn passes the value of the `generation_table` global to the EE, which records it in `g_generation_table` and uses it in allocation helpers.

Of note in this PR is that I named the `GCToEEInterface` function `StompWriteBarrier` to `UpdateEEGlobals`, since the latter name is more indicative of its purpose now that non-write-barrier globals are updated using it as well. If anyone has a better name, I'd be happy to use it instead.

I've tested this locally on a single-proc Windows VM and the tests (that don't timeout) look clean.

### DIFF
--- a/src/gc/env/gcenv.ee.h
+++ b/src/gc/env/gcenv.ee.h
@@ -65,7 +65,9 @@ public:
     static void DiagWalkSurvivors(void* gcContext);
     static void DiagWalkLOHSurvivors(void* gcContext);
     static void DiagWalkBGCSurvivors(void* gcContext);
-    static void StompWriteBarrier(WriteBarrierParameters* args);
+
+    // Inform the EE of changes to GC global variables.
+    static void UpdateEEGlobals(UpdateEEGlobalsParameters* args);
 
     static void EnableFinalization(bool foundFinalizers);
 };
--- a/src/gc/gc.cpp
+++ b/src/gc/gc.cpp
@@ -2177,8 +2177,8 @@ int log2(unsigned int n)
 
 void stomp_write_barrier_resize(bool is_runtime_suspended, bool requires_upper_bounds_check)
 {
-    WriteBarrierParameters args = {};
-    args.operation = WriteBarrierOp::StompResize;
+    UpdateEEGlobalsParameters args = {};
+    args.operation = UpdateEEGlobalsOp::StompResize;
     args.is_runtime_suspended = is_runtime_suspended;
     args.requires_upper_bounds_check = requires_upper_bounds_check;
     args.card_table = g_gc_card_table;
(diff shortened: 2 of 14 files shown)


swgillespie:
@jkotas 

This is the follow-up PR from the discussion we had on my DAC PR - do you still think it's more appropriate for this functionality to be somewhere other than `StompWriteBarrier`/`UpdateEEGlobals`? I've placed it on the `UpdateEEGlobalsOp::Initialize` path for now, since the generation table will not change after initialization.

cc @Maoni0 @adityamandaleeka

adityamandaleeka:
Looks like you missed a few more places that had WriteBarrierParameters.

swgillespie:
Yep, looks like I did - fixed in the latest commit.

swgillespie:
Ouch, thank you for double-checking. I'll address that.

jkotas:
Also, I still think that it is not right to attach this to write barrier geometry. It is fundamentally a different thing.

You should also take care of this together with `m_GCLock` because of the single proc global allocation context and `m_GCLock` are coupled together.

Maoni0:
all the alloc helpers need is to read/modify values at generation_table+0/8. generation_table itself doesn't change. can't you just pass it in some initialization function?

swgillespie:
> Also, I still think that it is not right to attach this to write barrier geometry. It is fundamentally a different thing.

How about separating the `Initialize` operation of `StompWriteBarrier` into two separate functions on the `GCToEEInterface`, then - one for initializing all shared globals, which may include both write-barrier and non-write-barrier constants, and the other for updating write-barrier constants? Does that sound reasonable?

> all the alloc helpers need is to read/modify values at generation_table+0/8. generation_table itself doesn't change. can't you just pass it in some initialization function?

By it you mean generation table? I re-used the `Initialize` operation of `StompWriteBarrier` for this, but I can do something else (like the above) if that's better.

jkotas:
> Does that sound reasonable?

We should be grouping architecturaly related globals together. The global allocation context and write barriers are pretty unrelated. Thus, I think that there should be a separate method on the GC/EE interface to get the information about the global allocation context that is separate from the write barrier one.

This method should provide everything about the global allocation context: both the context and the spin lock (m_Lock). An interesting design question is how the global allocation context should be split between EE and GG: The global allocation is meant to make the allocations on single core machines very fast. The global allocation context is owned by the GC in your current design. You will end up with extra indirections in the assembly allocation helper once you add the missing indirections. The extra indirections will make it slower and thus go against the design point of the global allocation context. I think it would be better to make the global allocation context owned by the EE and just let GC read and update it, similar to how the thread-local allocation contexts are read and updated using FixAllocContext and EnumAllocContexts. It would save the indirections on the hottest path which is the allocation helper.

Another fine answer may be to not support the `UseAllocationContexts() == false` with local GC, at least for now. The `UseAllocationContexts() == false` is pretty narrow already - it is only used on Win x86/x64, only with workstation GC, and only on single core machines. Do we know about any real current workload that it is important for?

swgillespie:
> Do we know about any real current workload that it is important for?

None that I know of. @Maoni0, do you have any thoughts? I don't have a preference about whether or not to support this scenario with local GC, except that the EE won't link unless we do something about `generation_table` and `m_GCLock` by doing one of the things that Jan proposed.

swgillespie:
Going to close this PR and re-open a new one later - the scope of this is expanding a little from the original change.

Issue #9691, opened 2017-02-21
[x86/linux] Loader.classloader.methodoverriding.regressions.549411.exploit test failure


YongseopKim:
result of running the test Loader/classloader/methodoverriding/regressions/549411/exploit/exploit.exe
```
Assert failure(PID 261 [0x00000105], Thread: 261 [0x0105]): false && "BackPatchWorkerAsmStub"
    File: /home/dragon/nfs/coreclr/src/vm/i386/unixstubs.cpp Line: 62
    Image: /nfs/issues/impl-backpatchworkerasmstub/coreoveray/corerun

Program received signal SIGABRT, Aborted.
0xf7fdacd9 in ?? ()
```

YongseopKim:
After https://github.com/dotnet/coreclr/pull/9690, result of running the test Loader/classloader/methodoverriding/regressions/549411/exploit/exploit.exe
```
Assert failure(PID 210 [0x000000d2], Thread: 210 [0x00d2]): isCallRegisterIndirect((BYTE *)returnAddr) || isDelegateCall((BYTE *)siteAddrForRegisterIndirect)
    File: /home/dragon/nfs/coreclr/src/vm/i386/virtualcallstubcpu.hpp Line: 726
    Image: /nfs/issues/impl-backpatchworkerasmstub/coreoveray/corerun

Program received signal SIGABRT, Aborted.
0xf7fdacd9 in ?? ()
```

YongseopKim:
I'm not sure that the Assert failure of isCallRegisterIndirect is related to BackPatchWorkerAsmStub. If this Assert failure is related to other things, I want to make a new issue for it.

YongseopKim:
cc @seanshpark @parjong @wateret

YongseopKim:
Current state with #9690 and the latest commit

```
Program received signal SIGSEGV, Segmentation fault.
0xf63b701f in ?? ()
(gdb) bt
#0  0xf63b701f in ?? ()
#1  0xf2767535 in ?? ()
#2  0xf276744c in ?? ()
#3  0xf71f806b in CallDescrWorkerInternal () at asmhelpers.S:442
#4  0xf6f8ea58 in CallDescrWorker (pCallDescrData=0xffffcc58) at /home/dragon/nfs/coreclr/src/vm/callhelpers.cpp:135
[...]

Issue #5542, opened 2016-06-06
[ARM32] Null exception not caught after two interface calls


chunseoklee:
Code : https://github.com/dotnet/coreclr/blob/master/tests/src/Loader/classloader/methodoverriding/regressions/549411/exploit.cs
Test above fails with segmentation fault with broken stack.
```
Program received signal SIGSEGV, Segmentation fault.
0xb5a3d014 in ?? ()
(gdb) bt
#0  0xb5a3d014 in ?? ()
#1  0xb1cdbc16 in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
(gdb) 
``` 

I found a clue : test fails when any interface function is called more than once. For example, the following code works OK.
```
	public static int DoExploit()
    	{
                // here NUM_ELEM = 4;
		I[] arrI = new I[NUM_ELEM];
		
        	for (int i = 0; i < NUM_ELEM; i++)
        	{
                    
            		if (i % 4 == 0)
            		{
                		arrI[i] = new X();
            		}
            		else if (i%4 == 1)
            		{
                		arrI[i] = new Y();
            		}
                        else if (i%4 == 2)
            		{
                		arrI[i] = new Z();
            		}
                        else {
                            arrI[i] = new Z();
                        }
        	}
        	arrI[NUM_ELEM-1] = null;
        	return TrustedClass.DoWork(arrI);
    	}
}
```
I think the exception handling of interface (cache) could be buggy.

sivarv:
This seems to be a duplicate of #5505
CC @rahku and @gkhanna79

chunseoklee:
@sivarv I think this issue is different from #5505 since CLRVectoredExceptionHandlerShim does not get control from any stubs. We need more investigation. I will keep on following #5505 and update if any progress on arm.

chunseoklee:
[disabling patches](https://github.com/chunseoklee/coreclr/commit/f887088f825d941457c4c7e463e4b79bd0c9010c) 
By disabling stub update(as in patch above), test seems to pass. 
Note that disabling either DISPATCH or RESOLVE stub has no effect.

chunseoklee:
Tested on arm32/linux with the following source code
```
bool DoWork {
........
	try   {                    
                    foreach(I i in arrI)
            		{
                       		i.Foo();                     
            		}                    
        	}
        	catch (NullReferenceException)
        	{
        		counter++;
        	} 
     return counter;
}

////////////////////////////////////////////////////////
[...]

PR #9844, opened 2017-02-28
[x86/linux] Add IsIPinVirtualStub() on x86/linux

YongseopKim:
To pass Loader.classloader.methodoverriding.regressions.549411.exploit
test failure on x86/linux (https://github.com/dotnet/coreclr/issues/9691). This patch is from #5542.

### DIFF
--- a/src/vm/exceptionhandling.cpp
+++ b/src/vm/exceptionhandling.cpp
@@ -5047,7 +5047,7 @@ BOOL IsSafeToHandleHardwareException(PCONTEXT contextRecord, PEXCEPTION_RECORD e
         exceptionRecord->ExceptionCode == STATUS_BREAKPOINT || 
         exceptionRecord->ExceptionCode == STATUS_SINGLE_STEP ||
         (IsSafeToCallExecutionManager() && ExecutionManager::IsManagedCode(controlPc)) ||
-#ifdef _TARGET_ARM_
+#if defined(_TARGET_ARM_) || (defined(_TARGET_X86_) && defined(FEATURE_PAL))
         IsIPinVirtualStub(controlPc) ||  // access violation comes from DispatchStub of Interface call
 #endif
         IsIPInMarkedJitHelper(controlPc));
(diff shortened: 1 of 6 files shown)


YongseopKim:
PTAL @jkotas @janvorli

src/vm/exceptionhandling.cpp
```diff
@@ -5047,7 +5047,7 @@ BOOL IsSafeToHandleHardwareException(PCONTEXT contextRecord, PEXCEPTION_RECORD e
         exceptionRecord->ExceptionCode == STATUS_BREAKPOINT || 
         exceptionRecord->ExceptionCode == STATUS_SINGLE_STEP ||
         (IsSafeToCallExecutionManager() && ExecutionManager::IsManagedCode(controlPc)) ||
-#ifdef _TARGET_ARM_
+#if defined(_TARGET_ARM_) || (defined(_TARGET_X86_) && defined(FEATURE_PAL))
         IsIPinVirtualStub(controlPc) ||  // access violation comes from DispatchStub of Interface call
```
parjong: You may remove FEATURE_PAL as WIN64EXCEPTIONS implies FEATURE_PAL anyway, but it seems optional.

YongseopKim: I'll remove it.

parjong:

src/vm/i386/excepx86.cpp
```diff
@@ -3708,3 +3708,34 @@ AdjustContextForVirtualStub(
 
     return TRUE;
 }
+
+//Return TRUE if pContext->Pc is in VirtualStub
+BOOL IsIPinVirtualStub(PCODE f_IP)
```
janvorli: Instead of having identical copy of this function at two places (one for ARM and one for x86), could you please move it to the exceptionhandling.cpp instead so that we have just one copy?

YongseopKim:
PTAL @janvorli

src/vm/exceptionhandling.cpp
```diff
@@ -7047,5 +7047,38 @@ void ExceptionTracker::ResetThreadAbortStatus(PTR_Thread pThread, CrawlFrame *pC
 }
 #endif //!DACCESS_COMPILE
 
+#if defined(_TARGET_ARM_) || defined(_TARGET_X86_)
```
janvorli: A nit - could you please move this function somewhere before the IsSafeToHandleHardwareException, mark the function as static and remove the function declaration from the exceptionhandling.h? It is not needed anywhere else, so there is no need to expose it.

YongseopKim: Yes, you're right. I didn't consider the point. I've just made the change.

PR #9846, opened 2017-02-28
Clean up x86 unwinders

parjong:
This commit revises x86 unwinder implementations as a static method in relevant classes (EpilogUnwinder, EspFrameUnwinder, EbpDoubleAlignFrameUnwinder) without changing behavior.


parjong:
@janvorli This PR is from my previous attempt to introduce interface-based unwinder, but I'll close this if you feel like that.

Issue #9847, opened 2017-02-28
[x86/Linux] JIT.Methodical.flowgraph.dev10_bug679008.fgloop test failure


parjong:
We observed the following assert failure while running unittest:
```
Assert failure(PID 7295 [0x00001c7f], Thread: 7295 [0x1c7f]): ExceptionTracker::HasFrameBeenUnwoundByAnyActiveException(&m_crawl)
    File: /home/parjong/projects/dotnet/coreclr/src/vm/stackwalk.cpp Line: 1859
    Image: /home/parjong/projects/dotnet/dotnet-overlay/Linux.x86.Debug.Debug/corerun
```

PR #9849, opened 2017-02-28
[x86/Linux] Correct GetStackParameterSize on Funclet (WIP)

parjong:
The current implementation returns the funtion's stack parameter size even for funclet, which results in #9847.

This commit revises GetStackParameterSize to take care of funclets correctly.

### DIFF
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -5913,6 +5913,14 @@ ULONG32 EECodeManager::GetStackParameterSize(EECodeInfo * pCodeInfo)
     } CONTRACTL_END;
 
 #if defined(_TARGET_X86_)
+#if defined(WIN64EXCEPTIONS)
+    if (pCodeInfo->IsFunclet())
+    {
+        // Funclet has no stack argument
+        return 0;
+    }
+#endif // WIN64EXCEPTIONS
+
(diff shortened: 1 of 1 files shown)


src/vm/eetwain.cpp
```diff
@@ -5913,6 +5913,14 @@ ULONG32 EECodeManager::GetStackParameterSize(EECodeInfo * pCodeInfo)
     } CONTRACTL_END;
 
 #if defined(_TARGET_X86_)
+#if defined(WIN64EXCEPTIONS)
+    if (pCodeInfo->IsFunclet())
```
janvorli: I don't understand why funclets should be handled differently. Since the funclet has no stack argument, it seems that the function should return 0 even without this modification. So it seems that we may be fixing a symptom, but not the root of the issue.

parjong: As I understand, it depends on how JITer treats funclet's info. It seems that JIter usually do not distinguish function and funclet in terms of frame layout as far as I know.

parjong: @janvorli Yes, it was. While digging ``JIT.Methodical.flowgraph.dev10_bug679008.fgloop``, I found that there is a mismatch between ``Esp`` and ``ResumeEsp`` and their difference was the stack argument size on the function itself.

PR #9855, opened 2017-02-28
[x86/Linux] Initial fix of arguments passing in FunctionPtrTest (WIP)

lucenticus:
This PR implements initial fix of arguments passing in UMThunkStub.
Fix for issue #9357
This PR also fixes following Interop tests:
Interop/BestFitMapping/BestFitMapping/BestFitMapping.sh
Interop/ArrayMarshalling/BoolArray/MarshalBoolArrayTest/MarshalBoolArrayTest.sh
Interop/RefInt/RefIntTest/RefIntTest.sh
Interop/RefCharArray/RefCharArrayTest/RefCharArrayTest.sh
Interop/StringMarshalling/LPTSTR/LPTSTRTest/LPTSTRTest.sh
@janvorli @parjong @seanshpark PTAL
\CC: @Dmitri-Botcharnikov

### DIFF
--- a/src/vm/dllimportcallback.cpp
+++ b/src/vm/dllimportcallback.cpp
@@ -1414,6 +1414,12 @@ VOID UMThunkMarshInfo::RunTimeInit()
 
         }
     }
+    //
+    // m_cbActualArgSize gets the number of arg bytes for the NATIVE signature
+    //
+    m_cbActualArgSize =
+        (pStubMD != NULL) ? pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize() : pMD->SizeOfArgStack();
+
 #if defined(_TARGET_X86_)
     MetaSig sig(pMD);
     ArgIterator argit(&sig);
@@ -1440,11 +1446,6 @@ VOID UMThunkMarshInfo::RunTimeInit()
             offs += StackElemSize(cbSize);
         }
     }
-    //
-    // m_cbActualArgSize gets the number of arg bytes for the NATIVE signature
-    //
-    m_cbActualArgSize =
-        (pStubMD != NULL) ? pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize() : pMD->SizeOfArgStack();
     PInvokeStaticSigInfo sigInfo;
     if (pMD != NULL)
         new (&sigInfo) PInvokeStaticSigInfo(pMD);
--- a/src/vm/dllimportcallback.cpp
+++ b/src/vm/dllimportcallback.cpp
@@ -1431,11 +1431,14 @@ VOID UMThunkMarshInfo::RunTimeInit()
         {
             if (numRegistersUsed == 1)
                 m_ecxArgOffset = offs;
-            else  if (numRegistersUsed == 2)
+            else if (numRegistersUsed == 2)
                 m_edxArgOffset = offs;
-            offs+= STACK_ELEM_SIZE;
-        } else
-            offs+=StackElemSize(cbSize);
+            offs += STACK_ELEM_SIZE;
+        }
+        else
+        {
+            offs += StackElemSize(cbSize);
+        }
     }
     //
     // m_cbActualArgSize gets the number of arg bytes for the NATIVE signature
--- a/src/vm/dllimportcallback.cpp
+++ b/src/vm/dllimportcallback.cpp
@@ -1411,13 +1411,53 @@ VOID UMThunkMarshInfo::RunTimeInit()
 
             pStubMD = GetILStubMethodDesc(pMD, &sigInfo, dwStubFlags);
             pFinalILStub = JitILStub(pStubMD);
+
         }
     }
+#if defined(_TARGET_X86_)
+    MetaSig sig(pMD);
+    ArgIterator argit(&sig);
+    int numRegistersUsed = 0;
+    m_ecxArgOffset = -1;
+    m_edxArgOffset = -1;
 
+    int offs = 0;
+    for (UINT i = 0 ; i < sig.NumFixedArgs(); i++)
+    {
+        TypeHandle thValueType;
+        CorElementType type = sig.NextArgNormalized(&thValueType);
+        int cbSize = sig.GetElemSize(type, thValueType);
+        if (ArgIterator::IsArgumentInRegister(&numRegistersUsed, type))
+        {
+            if (numRegistersUsed == 1)
+                m_ecxArgOffset = offs;
+            else  if (numRegistersUsed == 2)
+                m_edxArgOffset = offs;
+            offs+= STACK_ELEM_SIZE;
+        } else
+            offs+=StackElemSize(cbSize);
+    }
     //
     // m_cbActualArgSize gets the number of arg bytes for the NATIVE signature
     //
-    m_cbActualArgSize = (pStubMD != NULL) ? pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize() : pMD->SizeOfArgStack();
+    m_cbActualArgSize =
+        (pStubMD != NULL) ? pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize() : pMD->SizeOfArgStack();
+    PInvokeStaticSigInfo sigInfo;
+    if (pMD != NULL)
+        new (&sigInfo) PInvokeStaticSigInfo(pMD);
+    else
+        new (&sigInfo) PInvokeStaticSigInfo(GetSignature(), GetModule());
+    if (sigInfo.GetCallConv() == pmCallConvCdecl)
+    {
+        // caller pop
+        m_cbRetPop = 0;
+    }
+    else
+    {
+        // callee pop
+        m_cbRetPop = static_cast<UINT16>(m_cbActualArgSize);
+    }
+#endif // _TARGET_X86_
 
 #endif // _TARGET_X86_ && !FEATURE_STUBS_AS_IL
 
--- a/src/vm/dllimportcallback.h
+++ b/src/vm/dllimportcallback.h
@@ -211,8 +211,8 @@ private:
                                             // On non-x86, the managed entrypoint for no-delegate no-marshal signatures
     UINT32            m_cbActualArgSize;    // caches m_pSig.SizeOfFrameArgumentArray()
 #if defined(_TARGET_X86_)
-#if defined(FEATURE_STUBS_AS_IL)
     UINT16            m_cbRetPop;           // stack bytes popped by callee (for UpdateRegDisplay)
+#if defined(FEATURE_STUBS_AS_IL)
     UINT32            m_ecxArgOffset;
     UINT32            m_edxArgOffset;
 #else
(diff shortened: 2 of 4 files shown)


lucenticus:
@dotnet-bot test this please

src/vm/i386/umthunkstub.S
```diff
@@ -139,30 +148,40 @@ LOCAL_LABEL(UMThunkStub_CopyStackArgs):
...
     jz      LOCAL_LABEL(UMThunkStub_ArgumentsSetup)
 
-    sub     esp, eax
-    and     esp, -16    // align with 16 byte
-    lea     edi, [esp]
-
-LOCAL_LABEL(CopyLoop):
 
-    // copy rest of the arguments to [esp+08+n], in RTL
+LOCAL_LABEL(CopyStack):
```
janvorli: This doesn't look right. We need to copy all arguments that were not stored in the registers, in other words, the loop needs to skip copying the ones that were in registers to the new location of stack arguments. This loop does something different. EAX at the beginning is equal to the m_cbActualArgSize minus the size of args we pass in registers, so it is the size of the stack arguments. Let's call it K. The loop copies first K arguments to the target location on the stack. But that's not right, the arguments that are passed in registers can be at arbitrary offsets - that's why you have introduced the m_edxArgOffset and m_ecxArgOffset, right? So you could miss some arguments and some could get duplicated.
I think the simplest way to copy the arguments is to just loop through them from the first till the last one and compare the current offset with the m_edxArgOffset and m_ecxArgOffset. If the offset matches, then you load the source argument into a register. If it doesn't match, then you place it on the target stack.

lucenticus: Yes, you are right. I'll change this part, thank you!

lucenticus:
\CC: @sergign60

src/vm/i386/umthunkstub.S
```diff
@@ -154,34 +153,48 @@ LOCAL_LABEL(UMThunkStub_CopyStackArgs):
...
+    push    edx
+    mov     eax, 0  // Offset in edi
+    mov     ebx, 0  // Offset in esi
+LOCAL_LABEL(CopyStack):
+    mov     edx, dword ptr [ebp - UMThunkStub_UMENTRYTHUNK_OFFSET]
+    mov     edx, dword ptr [edx + UMEntryThunk__m_pUMThunkMarshInfo]
+    mov     ecx, dword ptr [edx + UMThunkMarshInfo__m_ecxArgOffset]
+    cmp     ecx, ebx
+    je      LOCAL_LABEL(IncreaseOffset)
+    mov     ecx, dword ptr [edx + UMThunkMarshInfo__m_edxArgOffset]
```
janvorli: You can use directly cmp ebx, dword ptr [edx + UMThunkMarshInfo__m_edxArgOffset] instead of loading the value to ecx first and then comparing it. At all three places in this function.

src/vm/i386/umthunkstub.S
```diff
@@ -154,34 +153,48 @@ LOCAL_LABEL(UMThunkStub_CopyStackArgs):
...
+    cmp     ecx, ebx
+    je      LOCAL_LABEL(IncreaseOffset)
     mov     ecx, dword ptr [esi + ebx]
-    mov     dword ptr [edi + ebx], ecx
+    mov     dword ptr [edi + eax], ecx
+    add     eax, 4
+LOCAL_LABEL(IncreaseOffset):
+    add     ebx, 4
+    mov     ecx, [edx + UMThunkMarshInfo__m_cbActualArgSize]
```
janvorli: If you changed the loop to go backwards as it was before, you will not need to access the [edx + UMThunkMarshInfo__m_cbActualArgSize]. 
You would initialize ebx with the actual arg size, use eax that we have before the loop and use [esi + ebx - 4] and [edi + eax - 4 ] above and here you would just terminate when the ebx became zero.
Or, you can just add eax - 4 to the edi before the loop, use [edi] to store to the target and use sub edi, 4 instead of sub eax, 4. Then you could use eax instead of ecx as a helper register in the loop and you would not have to push ecx.

src/vm/i386/umthunkstub.S
```diff
@@ -173,25 +173,20 @@ LOCAL_LABEL(UMThunkStub_CopyStackArgs):
...
     push    ecx
     push    edx
-    mov     eax, 0  // Offset in edi
-    mov     ebx, 0  // Offset in esi
-LOCAL_LABEL(CopyStack):
     mov     edx, dword ptr [ebp - UMThunkStub_UMENTRYTHUNK_OFFSET]
     mov     edx, dword ptr [edx + UMEntryThunk__m_pUMThunkMarshInfo]
-    mov     ecx, dword ptr [edx + UMThunkMarshInfo__m_ecxArgOffset]
-    cmp     ecx, ebx
+    mov     ebx, [edx + UMThunkMarshInfo__m_cbActualArgSize]
```
janvorli: There is a little bug here, probably induced by my previous comment. You'll need to subtract 4 from ebx before the loop, otherwise the offset comparison below won't work as expected. And change the `mov     ecx, dword ptr [esi + ebx - 4]` to `mov     ecx, dword ptr [esi + ebx]` and the `jnz` at the end to `jnc`.
And if you want, you can move the `add  eax, -4` before the `mov     ecx, dword ptr [esi + ebx]` and then change the `mov     dword ptr [edi + eax - 4], ecx` to `mov     dword ptr [edi + eax], ecx` to make the code look symmetric after the previous change.

lucenticus: I'm not sure whether I understood your comment correctly, FunctionPtrTest and some Interop tests failed with changes below: 
```
LOCAL_LABEL(InitCopyStack):
    push    ecx
    push    edx
    mov     edx, dword ptr [ebp - UMThunkStub_UMENTRYTHUNK_OFFSET]
    mov     edx, dword ptr [edx + UMEntryThunk__m_pUMThunkMarshInfo]
    mov     ebx, [edx + UMThunkMarshInfo__m_cbActualArgSize]
    add     ebx, -4
LOCAL_LABEL(CopyStack):
    cmp     ebx, dword ptr [edx + UMThunkMarshInfo__m_ecxArgOffset]
    je      LOCAL_LABEL(IncreaseOffset)
    cmp     ebx, dword ptr [edx + UMThunkMarshInfo__m_edxArgOffset]
    je      LOCAL_LABEL(IncreaseOffset)
    add     eax, -4
    mov     ecx, dword ptr [esi + ebx]
    mov     dword ptr [edi + eax], ecx
LOCAL_LABEL(IncreaseOffset):
    add     ebx, -4
    jnc     LOCAL_LABEL(CopyStack)

    pop     edx
    pop     ecx
    jmp     LOCAL_LABEL(UMThunkStub_ArgumentsSetup)

```
Without these changes all tests passed, probably I missed something?

janvorli: Hmm, this looks like what I have suggested. But if it doesn't work, then I think there must be some other problem that just hides this problem and makes the test work.
Just look at a simple case - a function with a single parameter passed in ecx. I would expect that the m_ecxArgOffset would be set to zero, right? Then in your loop, you load m_cbActualArgSize to ebx. That would be 4, since that's the size of the single argument, right? And now you compare ebx to m_ecxArgOffset. It doesn't match, since the ebx is 4 and m_ecxArgOffset is 0. So the argument would be plaved to stack. At the end of the loop, we subtract 4 from ebx and since it becomes zero, exit the loop. So it ends up putting the argument to stack.
Can you please step through the code while running the FunctionPtrTest to see what's exactly going on?

lucenticus: > I would expect that the m_ecxArgOffset would be set to zero, right?

I set `m_ecxArgOffset` and `m_edxArgOffset` as -1 if we don't use these registers. At the FunctionPtrTest we use stack only, so `cmp     ebx, dword ptr [edx + UMThunkMarshInfo__m_ecxArgOffset]` sets CF=1, and  `jnc     LOCAL_LABEL(CopyStack)` exit at the first iteration.

janvorli: The CF is set by the add ebx, -4 before the conditional jump. But I have just realized the `jnc` is wrong, since adding -4, which is 0xfffffffc to 0 doesn't set the CF, but rather clears it. And it is set in all other cases (4 + 0xfffffffc sets it, 8 + 0xfffffffc sets it too, etc).
So the polarity should be reverse, it means `jc`.

lucenticus: > The CF is set by the add ebx, -4 before the conditional jump. 

You are right, I missed that. `jc` works fine, thanks a lot!
BTW, 2 tests (Vector3Interop_ro.sh and Vector3Interop_r.sh) failed on x64 checks, so I should investigate the reason of this regression.

src/vm/dllimportcallback.h
```diff
@@ -210,9 +210,11 @@ class UMThunkMarshInfo
                                             // On x86, NULL for no-marshal signatures
                                             // On non-x86, the managed entrypoint for no-delegate no-marshal signatures
     UINT32            m_cbActualArgSize;    // caches m_pSig.SizeOfFrameArgumentArray()
+    UINT16            m_cbRetPop;           // stack bytes popped by callee (for UpdateRegDisplay)
```
janvorli: I've just noticed that these three fields should be present for x86 only (and the offsets for FEATURE_PAL only on top of that), but you have added it for all architectures.

lucenticus: Yes, it could be the reason of this issue, I've updated the PR, thank you!

src/vm/dllimportcallback.h
```diff
@@ -211,14 +211,14 @@ class UMThunkMarshInfo
                                             // On non-x86, the managed entrypoint for no-delegate no-marshal signatures
     UINT32            m_cbActualArgSize;    // caches m_pSig.SizeOfFrameArgumentArray()
 #if defined(_TARGET_X86_)
+#if defined(FEATURE_STUBS_AS_IL)
     UINT16            m_cbRetPop;           // stack bytes popped by callee (for UpdateRegDisplay)
```
janvorli: This one should be out of the FEATURE_STUBS_AS_IL ifdef, only the offsets should be in. The m_cbRetPop is used for x86 on Windows too where the FEATURE_STUBS_AS_IL  is not defined.

lucenticus:
Thank you very much for your review!

src/vm/dllimportcallback.cpp
```diff
@@ -1411,13 +1411,56 @@ VOID UMThunkMarshInfo::RunTimeInit()
...
+        }
+        else
+        {
+            offs += StackElemSize(cbSize);
+        }
+    }
     //
     // m_cbActualArgSize gets the number of arg bytes for the NATIVE signature
     //
-    m_cbActualArgSize = (pStubMD != NULL) ? pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize() : pMD->SizeOfArgStack();
```
janvorli: I can see the AMD64 windows tests are failing. I think the reason is here. This line was used for non-x86 before and now you have put it all under _TARGET_X86_ ifdef and it is missing for other targets.

PR #9857, opened 2017-02-28
Add the correct offset for finding nested structs.

jashook:
As is the base offset of a nested struct that was a lclField was
being dropped.

@pgavlin @briansull ptal
/cc @dotnet/arm64-contrib @dotnet/jit-contrib @sdmaclea


jashook:

jashook:
Note this fixes #8155.

briansull:
LGTM

PR #9859, opened 2017-02-28
Fix mismatch in the cmake script endif condition

janvorli:
This change fixes a mismatch in the cmake script endif condition introduced
by #9809.


janvorli:
I have no idea how come the CI tests succeeded when the #9809 was merged in.
CC: @qmfrederik

gkhanna79:
@janvorli Looks fine to me. Can you please trigger an Arm64 build to confirm? Also, were you able to repro the break locally (and shouldnt it repro for any architecture and not just Arm64)?

qmfrederik:
@janvorli @gkhanna79 Apologies for this! I spent some time reading on the endif() statement and it is not clear to me whether the expression in endif(expression) is still evaluated - the doc says it the expression optional, but not whether the expression is evaluated when it is present.

@janvorli For my information only, did you experience a build break because of this?

*https://cmake.org/Wiki/CMake_FAQ#Isn.27t_the_.22Expression.22_in_the_.22ELSE_.28Expression.29.22_confusing.3F
* http://stackoverflow.com/questions/29959126/why-does-cmake-syntax-have-redundant-parentheses-everywhere/29967305#29967305
* http://stackoverflow.com/questions/29472607/purpose-of-cmake-endif-expression

janvorli:
@gkhanna79, @qmfrederik I am sorry, I've just found that this was just a warning and that there were other errors further in the log. The actual error is different and seems unrelated to this change after all. It looks rather like a misconfiguration of the build machine.
There is plenty of errors like 
`e:\a\_work\4\s\src\coreclr\hosts\coreconsole\logger.cpp(5): fatal error C1083: Cannot open include file: 'stdio.h': No such file or directory`
`e:\a\_work\4\s\src\coreclr\hosts\coreconsole\coreconsole.cpp(10): fatal error C1083: Cannot open include file: 'windows.h': No such file or directory`

PR #9862, opened 2017-02-28
Fix broken GT_GE - VNF_GT_UN mapping

mikedn:
What it says on the tin...

Too bad we don't have GT_UGT and have to do this kind of mapping...


pgavlin:
Yikes; good catch.

LGTM--@JosephTremoulet, can you take a quick peek as well?

pgavlin:
cc @dotnet/jit-contrib

adityamandaleeka:
LGTM

PR #9868, opened 2017-03-01
Allow user to zip tests even when not building against packages

wtgodbe:
Previously, you could only zip tests for a Helix run when building against packages. By passing "ziptests" to buildtest.cmd, you can now be in a state satisfactory for sending jobs to helix, even if doing a normal test build.


wtgodbe:
CC @MattGal @swaroop-sridhar

swaroop-sridhar:
👍

PR #9869, opened 2017-03-01
Streamlined MinOpts GC ref tracking.

pkukol:
When MinOpts is on for a method (and the COMPlus_JitMinOptsTrackGCrefs override is not on), 64-bit GC encoders will generate an abbreviated call site table and mark all GC ref stack slots as 'untracked'. The main purpose is to speed up compilation but this change should also shrink the GC tables (with MinOpts) by an average of about 2x.

### DIFF
--- a/src/vm/gcinfodecoder.cpp
+++ b/src/vm/gcinfodecoder.cpp
@@ -628,6 +628,8 @@ bool GcInfoDecoder::EnumerateLiveSlots(
 
 
 #ifdef PARTIALLY_INTERRUPTIBLE_GC_SUPPORTED
+    bool noTrackedRefs = false;
+
     if(m_SafePointIndex < m_NumSafePoints && !executionAborted)
     {
         // Skip interruptibility information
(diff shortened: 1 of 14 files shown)


src/vm/gcinfodecoder.cpp
```diff
@@ -628,6 +628,8 @@ bool GcInfoDecoder::EnumerateLiveSlots(
 
 
 #ifdef PARTIALLY_INTERRUPTIBLE_GC_SUPPORTED
+    bool noTrackedRefs = false;
+
```
briansull: The CodegenMirror branch only mirrors the JIT directory not the VM (or inc) directories.
So the VM changes don't come over and won't automatically be part of any new Desktop releases.

If the VM changes rae required for correctness then they should be checked in to the CodegenMirror branch shortly after your PR goes through the mirror.

pkukol: Thanks for the heads up, @briansull - hopefully we'll be able to complete this thing quick this time around (it's almost exactly the same as last time, with one simplification that also fixes a weird problem that cropped up in one CoreFX test, BTW - @BruceForstall PTAL when you get a chance) and then I'll jump on the CGmirror. By default it should stay off for the desktop build so it shouldn't break anything.

pkukol:
@dotnet-bot test OSX minopts

pkukol:
@BruceForstall @pgavlin PTAL if you get a chance. I have to merge the CoreFX test fixes before this goes in, BTW, I'm hoping to get that done soon. Thanks!

pgavlin:
LGTM at a glance. What has changed since the previous revision?

pkukol:
This is 99% identical to the first go-around, the only material change is a small edit to the GC decoder where it could mess up earlier under rare conditions (plus the updated code is simpler as well so hopefully a win all around - this is in VM/GCinfoDecoder.cpp around line 730); plus, this time I've also pushed a PR in CoreFX to go in first to fix the invalid tests in that repo. Hope that makes sense; don't hesitate to add more comments or suggestions (anyone), though, I am not trying to rush this through and the CoreFX change has to go in first anyway.

PR #9871, opened 2017-03-01
Guard a use of a possibly-uninitialized BitSet.

pgavlin:
Part of the range check optimization pass was not checking to ensure that a bitset
was initialized before attempting to access it. This was causing one of the new
tests for Vector.Narrow to fail with an AV.

### DIFF
--- a/src/jit/rangecheck.cpp
+++ b/src/jit/rangecheck.cpp
@@ -869,10 +869,13 @@ Range RangeCheck::ComputeRangeForLocalDef(
         case GT_ASG:
         {
             Range range = GetRange(loc->block, loc->stmt, asg->gtGetOp2(), path, monotonic DEBUGARG(indent));
-            JITDUMP("Merge assertions from BB%02d:%s for assignment about %p\n", block->bbNum,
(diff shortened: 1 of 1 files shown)


pgavlin:
@JosephTremoulet @briansull @dotnet/jit-contrib PTAL.

See https://github.com/dotnet/coreclr/pull/9835 for an example of the test failure.

pgavlin:
@dotnet-bot test Windows_NT x86 corefx_baseline

pgavlin:
cc @mellinoe

pgavlin:
> Is it uninitialized because the block is unreachable? It's too bad we spend time trying to compute ranges for unreachable code, but I see the same guard around other calls to MergeEdgeAssertions, so LGTM.

I am not sure why it is initialized. I inferred from the other guards that this was an expected case and copy/pasted :/

At some point it would be good to understand why this is happening, but for the moment I'd simply like to unblock our CoreFX testing.

mikedn:
> I am not sure why it is initialized. I inferred from the other guards that this was an expected case and copy/pasted :/

Eh, I keep running into assertion prop weirdness, maybe I'll run into this too :)

Issue #9306, opened 2017-02-03
Additional BotR Chapters


mattwarren:
Is it possible for the additional BotR chapters to be made available? 

There are a few additional docs that are [tantalising mentioned in ceemain.cpp](https://github.com/dotnet/coreclr/blob/1d03b8fd8d650bd215623a7b035e68db96697e59/src/vm/ceemain.cpp#L40-L127), only some of which appear to match up with what's already available on GitHub. Although I guess that several don't apply to CoreCLR, 'ClickOnce Activation' for instance?

(issue opened as [suggested by](https://twitter.com/ziki_cz/status/826588632064471040) @karelz)

| Mention in cemain.cpp | Publish BOTR doc | 
|-----------------------|------------------|
| Introduction/BOTR Introduction.docx | [Documentation/botr/botr-faq.md](https://github.com/dotnet/coreclr/blob/master/Documentation/botr/botr-faq.md) |
| JIT/JIT Design.doc | [Documentation/botr/ryujit-overview.md](https://github.com/dotnet/coreclr/blob/master/Documentation/botr/ryujit-overview.md) |
| NGEN/NGENDesign.doc | |
| mscorlib/mscorlibDesign.doc | [Documentation/botr/mscorlib.md](https://github.com/dotnet/coreclr/blob/master/Documentation/botr/mscorlib.md) |
| ManagedEH/Design.doc | |
| CER/CERDesign.doc | |
| AssemblyLoader/AssemblyLoader.doc | |
| AssemblyLoader/FusionDesign.doc | |
| BCL/SystemStringDesign.doc | |
| DiagnosticServices/ProfilingAPIDesign.doc | [Documentation/botr/profiling.md](https://github.com/dotnet/coreclr/blob/master/Documentation/botr/profiling.md) |
| EERemotingSupport/RemotingDesign.doc | |
| MDA/MDADesign.doc | |
| Reflection/ReflectionDesign.doc | |
| RuntimeSecurity/RuntimeSecurityDesign.doc | |
| LoadtimeSecurity/DeclarativeSecurity-Design.doc | |
| LoadtimeSecurity/StrongName.doc | |
| RuntimeSecurity/ClickOnce Activation.doc | |
| RuntimeSecurity/Cryptography.doc | |
| RuntimeSecurity/DemandEvalDesign.doc | |

jkotas:
@richlander and I went over all the Botr documents at the time and get all the good ones published as markdown. The rest is either:
[...]

PR #9874, opened 2017-03-01
Update links to point to valid BotR pages

mattwarren:
As per comment from @jkotas, https://github.com/dotnet/coreclr/issues/9306#issuecomment-277282440


jkotas:
Thanks!

PR #9883, opened 2017-03-01
Preallocate jump stubs for dynamic methods

kouvel:
- This eliminates the possibility of running into an out-of-memory situation after compiling the method
- The temporary entry points block containing FixupPrecodes is extended for dynamic methods to include sufficient space for jump stubs
- When the target is too far for the FixupPrecode to encode a short relative jump, it instead does a short relative call or jump to the corresponding jump stub, which does an absolute jump to the target


kouvel:
@jkotas @BruceForstall

src/vm/amd64/cgenamd64.cpp
```diff
@@ -727,6 +727,35 @@ INT32 rel32UsingJumpStub(INT32 UNALIGNED * pRel32, PCODE target, MethodDesc *pMe
     return static_cast<INT32>(offset);
 }
 
+INT32 rel32UsingPreallocatedJumpStub(INT32 UNALIGNED * pRel32, PCODE target, PCODE jumpStubAddr)
+{
+    CONTRACTL
+    {
+        THROWS; // emptBackToBackJump may throw (see emitJump)
```
BruceForstall: >emptBackToBackJump [](start = 19, length = 18)

nit: typo: emptBackToBackJump

kouvel: Fixed

BruceForstall:
LGTM

BruceForstall:
Did you compute/measure the additional size impact of this?

Does NGEN store FixupPrecode? What is the size increase, say, of System.Private.CoreLib.ni.dll? Does nidump still work? Did you test on ngen'ed images?

kouvel:
Since the change only applies to dynamically created methods, my understanding is that it should have no impact on ngen/crossgen. I'll check what you suggested.

kouvel:
In terms of runtime memory impact, since we were allocating an additional 8 bytes per FixupPrecode, which wasn't being used, after this change the additional impact is 4 bytes per FixupPrecode for a dynamic method (12 bytes per jump stub).

kouvel:
> Does NGEN store FixupPrecode?

Looks like it, but it shouldn't be saving any FixupPrecodes for dynamic methods

> What is the size increase, say, of System.Private.CoreLib.ni.dll?

No change

> Does nidump still work?

It should, but while I found the code for it I didn't find info on how to initiate a dump. Could you please point me to some info on this?

> Did you test on ngen'ed images?

No, I can maybe diff the nidump before and after and see if there is anything to test

BruceForstall:
Given that ngen images shouldn't include FixupPrecode, I wouldn't worry about it.

I normally use "nidump /?" or similar to remind myself of the commands. Maybe it's not built in the open? I haven't looked.

jkotas:
Fragile ngen images do include FixupPrecode - a lot of them. 

However, this change have not changed anywhere near NGen or nidump.

PR #9887, opened 2017-03-01
Cleanup the native globalization code

tarekgh:
After we merged the globalization code used for Windows and Linux, this change is cleaning up the native VM globalization code


tarekgh:
@jkotas this is a follow up cleanup after the globalization code merge done. 

cc @danmosemsft

tarekgh:
@dotnet-bot help

tarekgh:

src/classlibnative/inc/nlsinfo.h
```diff
@@ -145,53 +86,7 @@ class COMNlsInfo {
 
     static void QCALLTYPE nativeNormalizationInitNormalization(int NormForm, BYTE* pTableData);
```
jkotas: Why do we need these normalization fcalls? I though that you have ported over the implementation that just PInvokes directly.

tarekgh: my changes was mainly to kill the duplication in the globalization code (i.e. get rid of the files under corefx folder). normalization stuff is not part of this, it is  used by string/System.Text. we can do your suggestion later.

tarekgh: I'll look more at that.

tarekgh: @jkotas we have some special case for normalization when we run on pre-Windows 8 (i.e. Win7). we carry the normalization data and use it when running on Windows 7. the reason is the normalization data on Windows 7 was not that good. if we are willing to scarify that when running on Windows 7, I can get rid of the fcalls along with the data too which will save us some more size.

jkotas: ok, we can leave it as is then ... it may be useful to add this as a comment.

tarekgh: The code using it is kind of clear about that:

https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Text/Normalization.Windows.cs#L48

but I'll add a comment anyway

jkotas: Actually, IsWindows8OrAbove is unconditionally returning true and we are not carrying the resources that would be required to make it work on pre-Win8 machines. We have shipped 1.0/1.1 that way. So all this is unreachable code.

tarekgh: Interesting. I can do the changes in other PR  then. ok?

jkotas: ok

PR #9889, opened 2017-03-01
Fix casing for newly added test

AndyAyersMS:
Fix another `jit` vs `JIT` path case issue that I caused.


AndyAyersMS:
@sdmaclea PTAL. Thanks for spotting this.
cc @dotnet/jit-contrib

PR #9890, opened 2017-03-01
Fix LDVIRTFTN case for GVM target methods

fadimounir:
Fix for issue https://github.com/dotnet/corert/issues/2796
We were handling the GVM case correctly for the callvirt scenarios, but not for the ldvirtftn opcodes.
This fix moves the GVM handling logic for CoreRT into the impImportLdvirtftn helper to fix both cases.

### DIFF
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -4990,22 +4990,22 @@ GenTreePtr Compiler::impImportLdvirtftn(GenTreePtr              thisPtr,
     }
 
 #if COR_JIT_EE_VERSION > 460
-	// CoreRT generic virtual method
-	if (((pCallInfo->sig.callConv & CORINFO_CALLCONV_GENERIC) != 0) && IsTargetAbi(CORINFO_CORERT_ABI))
-	{
-		GenTreePtr runtimeMethodHandle = nullptr;
-		if (pCallInfo->exactContextNeedsRuntimeLookup)
-		{
-			runtimeMethodHandle = impRuntimeLookupToTree(pResolvedToken, &pCallInfo->codePointerLookup, 
-															pCallInfo->hMethod);
-		}
-		else
-		{
-			runtimeMethodHandle = gtNewIconEmbMethHndNode(pResolvedToken->hMethod);
-		}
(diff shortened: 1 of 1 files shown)


fadimounir:
@sandreenko PTAL (feel free to include other folks from the JIT team to this review)

src/jit/importer.cpp
```diff
@@ -4989,6 +4989,25 @@ GenTreePtr Compiler::impImportLdvirtftn(GenTreePtr              thisPtr,
...
 
+#if COR_JIT_EE_VERSION > 460
+	// CoreRT generic virtual method
+	if (((pCallInfo->sig.callConv & CORINFO_CALLCONV_GENERIC) != 0) && IsTargetAbi(CORINFO_CORERT_ABI))
+	{
+		GenTreePtr runtimeMethodHandle = nullptr;
+		if (pCallInfo->exactContextNeedsRuntimeLookup)
+		{
+			runtimeMethodHandle = impRuntimeLookupToTree(pResolvedToken, &pCallInfo->codePointerLookup, 
+															pCallInfo->hMethod);
```
sandreenko: formatting looks strange here.

fadimounir: Hmm... need to change my tabs to spaces. Thanks for spotting this

fadimounir: Fixed

src/jit/importer.cpp
```diff
@@ -4989,6 +4989,25 @@ GenTreePtr Compiler::impImportLdvirtftn(GenTreePtr              thisPtr,
         NO_WAY("Virtual call to a function added via EnC is not supported");
     }
 
+#if COR_JIT_EE_VERSION > 460
```
fadimounir: @BruceForstall @jkotas Could you guys please comment on whether this block should move under the `if (opts.IsReadyToRun())` block a few lines below or not?
CoreRT compilations are also R2R compilations today, but it seems like the CORERT_ABI is declared independently from FEATURE_READYTORUN_COMPILER

jkotas: The CoreRT-specific handling of GVMs and fat functions pointers should be only under `CORERT_ABI`, and not under R2R.

PR #9893, opened 2017-03-02
Use the correct class handle for fat calli temporary

MichalStrehovsky:
`clsHnd` is the class handle of the type that owns the method. We need
the class handle for the method's return type though.

This is ending up with RyuJIT asking weird questions over JitInterface
(`getClassGClayout` of a reference type). I assume other things are
wrong too.

### DIFF
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -6432,7 +6432,6 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
         eeGetSig(pResolvedToken->token, info.compScopeHnd, impTokenLookupContextHandle, &calliSig);
 
         callRetTyp = JITtype2varType(calliSig.retType);
-        clsHnd     = calliSig.retTypeClass;
 
         call = impImportIndirectCall(&calliSig, ilOffset);
 
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -7715,7 +7715,7 @@ DONE_CALL:
                         unsigned   calliSlot  = lvaGrabTemp(true DEBUGARG("calli"));
                         LclVarDsc* varDsc     = &lvaTable[calliSlot];
                         varDsc->lvVerTypeInfo = tiRetVal;
-                        impAssignTempGen(calliSlot, call, clsHnd, (unsigned)CHECK_SPILL_NONE);
+                        impAssignTempGen(calliSlot, call, tiRetVal.GetClassHandle(), (unsigned)CHECK_SPILL_NONE);
                         // impAssignTempGen can change src arg list and return type for call that returns struct.
                         var_types type = genActualType(lvaTable[calliSlot].TypeGet());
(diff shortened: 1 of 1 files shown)


MichalStrehovsky:
@dotnet/jit-contrib @sandreenko Please take a look!

src/jit/importer.cpp
```diff
@@ -7715,7 +7715,7 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
                         unsigned   calliSlot  = lvaGrabTemp(true DEBUGARG("calli"));
                         LclVarDsc* varDsc     = &lvaTable[calliSlot];
                         varDsc->lvVerTypeInfo = tiRetVal;
-                        impAssignTempGen(calliSlot, call, clsHnd, (unsigned)CHECK_SPILL_NONE);
+                        impAssignTempGen(calliSlot, call, tiRetVal.GetClassHandle(), (unsigned)CHECK_SPILL_NONE);
```
MichalStrehovsky: (I don't know if the difference between this and `sig->retTypeClass` is important or which one should be used here. Please review as if a monkey wrote this change.)

AndyAyersMS: I think they are equivalent here. Using tiRetVal seems appropriate.

sandreenko: I think that value of clsHnd was  calliSig.retTypeClass before your fix. 

```
    if (opcode == CEE_CALLI)
    {
        /* Get the call site sig */
        eeGetSig(pResolvedToken->token, info.compScopeHnd, impTokenLookupContextHandle, &calliSig);

        callRetTyp = JITtype2varType(calliSig.retType);
        clsHnd     = calliSig.retTypeClass;
```
So it wasn't the class handle of the type that owns the method.
Am I wrong?

MichalStrehovsky: I think it gets stomped over by `clsHnd = pResolvedToken->hClass;` later.

Plus I'm seeing this on the GVM codepath anyway (so not CEE_CALLI, I think).

sandreenko: Yes, so my code " clsHnd  = calliSig.retTypeClass;" was bad because clsHnd had another meaning before. We can delete that as it was used only for the part of code that you changed.. 
However it doesn't explain the difference between this and sig->retTypeClass.

MichalStrehovsky: Ah, I see. So the `clsHnd = calliSig.retTypeClass;` was only added to support the line I'm changing anyway (in #9191). Line removed.

The `sig->retTypeClass` is only a question of whether `NormaliseForStack`, `SetIsReadonlyByRef`, `SetIsPermanentHomeByRef` et al mess with the class handle in any interesting way. It seems like they don't (so the value will be the same).

It seems reasonable to use `tiRetVal` to obtain the value since that's what we also use to initialize `varDsc->lvVerTypeInfo` line above.

PR #9897, opened 2017-03-02
[ARM32/Linux] Prepare rootfs to build libcoreclrtraceptprovider.so

hseok-oh:
Prepare rootfs to build libcoreclrtraceptprovider.so
by adding fix process of rootfs for ubuntu 14.04 armhf

### DIFF
--- a/cross/arm/trusty.patch
+++ b/cross/arm/trusty.patch
@@ -1,0 +1,97 @@
+diff -u -r a/usr/include/urcu/uatomic/generic.h b/usr/include/urcu/uatomic/generic.h
+--- a/usr/include/urcu/uatomic/generic.h	2014-03-28 06:04:42.000000000 +0900
++++ b/usr/include/urcu/uatomic/generic.h	2017-02-13 10:35:21.189927116 +0900
+@@ -65,17 +65,17 @@
+ 	switch (len) {
+ #ifdef UATOMIC_HAS_ATOMIC_BYTE
+ 	case 1:
+-		return __sync_val_compare_and_swap_1(addr, old, _new);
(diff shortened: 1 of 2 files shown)


hseok-oh:
This commit does not include build `libcoreclrtraceptprovider.so` because of CI failure.
I'll make new commit when CI and rootfs in docker file are updated

related issue: #8684

cc/ @gkhanna79

gkhanna79:
@hseok-oh Sounds like this patch will enable building libcoreclrtraceptprovider.so, right? If so, can you get this patch and fix in core-setup repo as well since we use that to generate the Docker Images with RootFS embedded?

CC @MichaelSimons

hseok-oh:
@gkhanna79 I'll make PR in core-setup, too. Thanks.

gkhanna79:
@hseok-oh I have merged the Core-Setup PR and will merge this as well. Do we need similar change for CoreFX?

hseok-oh:
@gkhanna79 It does not seems need similar change in CoreFX.

gkhanna79:
@hseok-oh My point was to ensure that RootFS is consistent for all cross builds. In that view, shouldn't we have this change in CoreFX and Core-Setup as well? do keep in mind that this change will be present in Docker images and thus, will get used for building CoreFX and Core-Setup as well.

Given that, are you fine with have it different for the repos?

hseok-oh:
@gkhanna79 In that point of view, it may be better to apply this change in CoreFX to ensure RootFS consistancy in coreclr, corefx, and core-setup. 

I'll make new PR in CoreFX, too.

gkhanna79:
Thank you @hseok-oh

PR #9603, opened 2017-02-15
Remove more dead native defines

danmosemsft:
Contributes to https://github.com/dotnet/coreclr/issues/9067

Still a few more to go.

PR #9899, opened 2017-03-02
Remove CLRThreadpoolHosted as it always returns false.

AustinWise:
When commit 1476776a3989817bf5378d583b5cf61dd5957925, as part of #9603, removed all the code under `FEATURE_INCLUDE_ALL_INTERFACES`, it left behind several functions in src/vm/util.hpp that just return `false`. `CLRThreadpoolHosted` is one of them.

While the compiler is probably able to able to see that these methods always return `false` and remove dead code accordingly, some of these `CLR*Hosted` functions make parts of the coreclr more complicated than they need to be and harder for humans to understand. Would it be valuable to have all of these functions removed?

### DIFF
--- a/src/vm/threadpoolrequest.cpp
+++ b/src/vm/threadpoolrequest.cpp
@@ -368,11 +368,8 @@ void UnManagedPerAppDomainTPCount::SetAppDomainRequestsActive()
         LONG prevCount = FastInterlockCompareExchange(&m_outstandingThreadRequestCount, count+1, count);
         if (prevCount == count)
         {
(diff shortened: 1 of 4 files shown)


stephentoub:
Can we also get rid of:
https://github.com/dotnet/coreclr/blob/542ca8e3fe40dcea5c7c69bc20915ca41b8b3f97/src/vm/win32threadpool.h#L557-L560
?

AustinWise:
@stephentoub I will try removing it in a follow-up PR.

PR #9906, opened 2017-03-02
[Linux][GDB-JIT] Fix lldb type displaying issues

ayuckhulk:
This PR introduces fixes in GDB-JIT to help LLDB display primitive types, structs, classes and arrays.

Highlights:

* fix incorrect value type size when it is embedded in other types

* fix reference types duplication

* fix memory leak in ByteTypeInfo::DumpStrings

* use array of 1 element instead of lenght expression in DWARF (because LLDB calculates array length incorrectly)

* use C# type names for built-in types (float instead of System.Single etc.)

* use full names (with namespaces) for classes and structures

* add typedef for reference types (example: make MyClass a typedef for __MyClass *)

* add bounds field in array structure for multi-dimensional arrays

@mikem8361, @janvorli PTAL

cc @Dmitri-Botcharnikov @chunseoklee @seanshpark @lucenticus


ayuckhulk:
@janvorli Thank you for the review!

PR #9915, opened 2017-03-02
Fix AsmConstants.inc to be incrementally rebuilt as necessary

BruceForstall:
There's still an incremental build problem, where changing AsmConstants.h
doesn't cause the assembly files to be rebuilt.

### DIFF
--- a/src/vm/wks/CMakeLists.txt
+++ b/src/vm/wks/CMakeLists.txt
@@ -105,11 +105,21 @@ if (WIN32)
     endif()
 
     add_custom_command(
-        # The AsmConstants.inc will be built in the pre-build phase of the cee_wks build
-        TARGET cee_wks PRE_BUILD 
+        OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/AsmConstants.inc"
+        DEPENDS ${VM_DIR}/${ARCH_SOURCES_DIR}/asmconstants.h
         COMMAND ${POWERSHELL} -NoProfile -ExecutionPolicy Bypass -NonInteractive \"& \"\"${VM_DIR}/h2inc.ps1\"\"\" \"\"\"${VM_DIR}/${ARCH_SOURCES_DIR}/asmconstants.h\"\"\" >"${CMAKE_CURRENT_BINARY_DIR}/AsmConstants.tmp"
         COMMAND ${CMAKE_CXX_COMPILER} ${DEFINITIONS} /EP "${CMAKE_CURRENT_BINARY_DIR}/AsmConstants.tmp" >"${CMAKE_CURRENT_BINARY_DIR}/AsmConstants.inc"
     )
+
+    set_source_files_properties(${CMAKE_CURRENT_BINARY_DIR}/AsmConstants.inc PROPERTIES GENERATED TRUE)
+
+    add_custom_target(
+        asmconstants_inc
+        DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/AsmConstants.inc
+    )
+
+    add_dependencies(cee_wks asmconstants_inc)
+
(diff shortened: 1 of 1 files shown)


src/vm/wks/CMakeLists.txt
```diff
@@ -105,11 +105,15 @@ if (WIN32)
...
-        # The AsmConstants.inc will be built in the pre-build phase of the cee_wks build
-        TARGET cee_wks PRE_BUILD 
+        OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/AsmConstants.inc"
+        DEPENDS ${VM_DIR}/${ARCH_SOURCES_DIR}/asmconstants.h
         COMMAND ${POWERSHELL} -NoProfile -ExecutionPolicy Bypass -NonInteractive \"& \"\"${VM_DIR}/h2inc.ps1\"\"\" \"\"\"${VM_DIR}/${ARCH_SOURCES_DIR}/asmconstants.h\"\"\" >"${CMAKE_CURRENT_BINARY_DIR}/AsmConstants.tmp"
         COMMAND ${CMAKE_CXX_COMPILER} ${DEFINITIONS} /EP "${CMAKE_CURRENT_BINARY_DIR}/AsmConstants.tmp" >"${CMAKE_CURRENT_BINARY_DIR}/AsmConstants.inc"
     )
+
+    set_source_files_properties(${CMAKE_CURRENT_BINARY_DIR}/AsmConstants.inc PROPERTIES GENERATED TRUE)
+    set_property(SOURCE ${VM_SOURCES_WKS_ARCH_ASM} PROPERTY OBJECT_DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/AsmConstants.inc)
```
BruceForstall: It's not clear this last line has any effect for asm files.

BruceForstall:
@janvorli @jkotas PTAL

src/vm/wks/CMakeLists.txt
janvorli: This is not enough to create a proper dependency, unless the object_depends set for all the sources is sufficient, which I am slightly not sure about - the cmake doc says it is a deprecated way. The way we use everywhere else is to add custom target, make it depend on the asmconstants.inc and make the targets that are built out of the asm files that include the asmconstants.inc  depend on that custom target. Unless cmake has problems detecting the dependency of the asm files on the inc file, that should be sufficient.
Like:
```cmake
 add_custom_target(
    asmconstants_inc
    DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/AsmConstants.inc
  )
```

BruceForstall:
@janvorli I implemented the changes you suggested, and they work fine for rebuilding the asmconstants.inc from a changed asmconstants.h. However, they don't cause the .asm files to be rebuilt. In fact, I haven't found any combination of cmake settings that will cause it to create a dependency between the generated file and the .asm files.

Note that even before this change, we rebuilt the asmconstants.inc file every single build, but never rebuilt the .asm files, so this broken behavior is not new.

janvorli:
@BruceForstall I guess it must have something to do with cmake not being able to detect the implicit dependencies of the asm files built by masm. 
The cmake doc at https://cmake.org/Wiki/CMake/Assembler says:
> There may be problems with dependency scanning for assembler files. Let us know if this is the case for you.

BruceForstall:
@janvorli Are you ok merging this as is?

janvorli:
@BruceForstall yes, let's merge this in and then we can try to come back to it later to see if we can find some trick that would fix the problem.

janvorli:
@BruceForstall out of curiosity, I have tried to touch the src/vm/AsmMacros.inc and it also didn't cause asm files rebuild, so it really looks like cmake doesn't know how to recognize dependencies in asm files compiled with masm.

PR #9917, opened 2017-03-02
Build cross-compile altjits for arm32 and arm64

BruceForstall:
On x86, build an x86-host, arm32-target altjit.
On amd64, build an amd64-host, arm64-target altjit.

Each are named protononjit.dll, using our existing naming
convention (proto for "prototype", nonjit meaning the code
won't be executed).

Use these via:
```
set COMPlus_AltJit=*
set COMPlus_AltJitName=protononjit.dll
```
Of course, instead of `*`, you can use any function name.

They are very useful for debugging the JIT in a "nicer" environment
than the full native environment, especially if the full native
environment isn't complete, or if the tools (such as debuggers, editors)
are better in the cross environment than the native environment.

### DIFF
--- a/src/jit/CMakeLists.txt
+++ b/src/jit/CMakeLists.txt
@@ -12,6 +12,8 @@ if (CLR_CMAKE_TARGET_ARCH_AMD64 OR (CLR_CMAKE_TARGET_ARCH_I386 AND NOT CLR_CMAKE
   add_definitions(-DFEATURE_AVX_SUPPORT)
 endif ()
 
+# JIT_BUILD disables certain PAL_TRY debugging features
+add_definitions(-DJIT_BUILD=1)
 
 if(WIN32)
   set(JIT_RESOURCES Native.rc)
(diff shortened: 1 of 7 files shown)


BruceForstall:
@pgavlin @briansull PTAL
cc @dotnet/jit-contrib

BruceForstall:
cc @mskvortsov @sdmaclea

src/jit/legacyjit/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ endif(WIN32)
 add_library_clr(legacyjit
    SHARED
    ${SHARED_LIB_SOURCES}
+   ${JIT_ARCH_SOURCES}
```
pgavlin: We probably want to define `JIT_ARCH_LEGACY_SOURCES` in `jit/CMakeLists.txt` and use that here. As it stands I think the legacyjit build is going to use the RyuJIT backend.

BruceForstall: The legacy sources are included in all builds; it's the `LEGACY_BACKEND` definition here (and not elsewhere) that causes them to actually be used.

This is by design, for simplicity, and to improve IntelliSense.

pgavlin: Ah, I see it now. Thanks for the clarification!

PR #9921, opened 2017-03-02
Fix error in dumpling text file cleanup

adityamandaleeka:
https://github.com/dotnet/coreclr/commit/85358db35ac3f37d5c3eb541984fe3a27b7a1175 added some logic for creating and cleaning up text files which contain dump information.

Both the creation and the clean up code were using the current directory as the location where the files should be. However, because the clean up code was just inserted near the beginning of the script, the current directory at that time is wherever the script is run from, and it can be different from `$testRootDir`, under which the files are created.

This change moves the cleanup code close to the creation code and uses the `$testRootDir` path explicitly rather than just the current directory.

@bryanar

### DIFF
--- a/tests/runtest.sh
+++ b/tests/runtest.sh
@@ -118,14 +118,6 @@ case $OSName in
         ;;
 esac
 
-# clean up any existing dumpling remnants from previous runs.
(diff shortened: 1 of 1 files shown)


bryanar:
Looks good to me.

PR #9922, opened 2017-03-02
Merge shared changes from corert and use C# shared project to enable VS editing

alexperovich:


src/mscorlib/corefx/Interop/Unix/System.Native/Interop.GetRandomBytes.cs
```diff
@@ -0,0 +1,19 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
jkotas: I think you can put this under src\mscorlib\src\Interop. We should get rid of everything under corefx.

src/mscorlib/corefx/Interop/Windows/BCrypt/Interop.BCryptGenRandom.cs
```diff
@@ -0,0 +1,44 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
```
jkotas: Is this different from CoreRT copy? Can this be under shared?

src/mscorlib/corefx/Interop/Unix/System.Native/Interop.GetRandomBytes.cs
jkotas: Same for ArrayPoolEventSource.cs. There are just a few files left under corefx. You can just move all of them so that the directory disappears.

src/mscorlib/shared/System/IO/Path.Unix.cs
```diff
@@ -201,16 +199,17 @@ public static string GetPathRoot(string path)
...
+        internal static bool IsCaseSensitive
         {
-            // We want to avoid dependencies on the Crypto library when compiling in CoreCLR. This
-            // will use the existing PAL implementation.
-            byte[] buffer = new byte[KeyLength];
-            Microsoft.Win32.Win32Native.Random(bStrong: true, buffer: buffer, length: KeyLength);
-            Runtime.InteropServices.Marshal.Copy(buffer, 0, (IntPtr)bytes, KeyLength);
+            get
+            {
+                #if PLATFORM_OSX
```
jkotas: Is this defined for OSX build in CoreCLR repo?

jkotas:
LGTM modulo comments.

src/mscorlib/corefx/Interop/Windows/BCrypt/Interop.BCryptGenRandom.cs
alexperovich: This one can be shared, I had it unshared because the equivalent Unix implementation was different. I will move this one.

src/mscorlib/shared/System/IO/Path.Unix.cs
alexperovich: will check

src/mscorlib/corefx/Interop/Windows/BCrypt/Interop.BCryptGenRandom.cs
jkotas: We will get the Unix implementation on the shared plan too eventually... .

jkotas:
I made a few tweaks to fix the ProjectN TFS build breaks (https://github.com/dotnet/corert/pull/2878) and moved missed file (Interop.MksTemps.cs) to shared. You may want to fold it into this change.

PR #9926, opened 2017-03-03
Implement slicing APIs for ArraySegment<T>

jamesqo:
Progress towards https://github.com/dotnet/corefx/issues/10528.

I let `Slice` accept negative indices so the start of the `ArraySegment<T>` can backtrack to an earlier position in the array. Please let me know if you think this is a good idea or not (checking for it would involve some additional overhead).

/cc @jkotas, @KrzysztofCwalina

### DIFF
--- a/src/mscorlib/src/System/ArraySegment.cs
+++ b/src/mscorlib/src/System/ArraySegment.cs
@@ -27,11 +27,9 @@ namespace System
     [Serializable]
     public struct ArraySegment<T> : IList<T>, IReadOnlyList<T>
     {
-        // Do not replace this with Array.Empty. We don't want to have the overhead of instantiating
-        // another generic type in addition to ArraySegment<T> for new type parameters.
-        private static readonly T[] s_emptyArray = new T[0];
-
-        public static ArraySegment<T> Empty { get; } = new ArraySegment<T>(s_emptyArray);
+        // Do not replace the array allocation with Array.Empty. We don't want to have the overhead of
+        // instantiating another generic type in addition to ArraySegment<T> for new type parameters.
+        public static ArraySegment<T> Empty { get; } = new ArraySegment<T>(new T[0]);
 
         private readonly T[] _array;
         private readonly int _offset;
@@ -112,7 +110,11 @@ namespace System
             }
         }
 
-        public ref T this[int index] => ref _array[_offset + index];
+        public T this[int index]
+        {
+            get { return _array[_offset + index]; }
+            set { _array[_offset + index] = value; }
+        }
 
         public Enumerator GetEnumerator()
         {
@@ -185,7 +187,7 @@ namespace System
 
             if (_count == 0)
             {
-                return s_emptyArray;
+                return Empty._array;
             }
 
             var array = new T[_count];
--- a/src/mscorlib/src/System/ArraySegment.cs
+++ b/src/mscorlib/src/System/ArraySegment.cs
@@ -151,6 +151,12 @@ namespace System
         {
             ThrowInvalidOperationIfDefault();
             destination.ThrowInvalidOperationIfDefault();
+
+            if (_count > destination._count)
+            {
+                ThrowHelper.ThrowArgumentException_DestinationTooShort();
+            }
+
             System.Array.Copy(_array, _offset, destination._array, destination._offset, _count);
         }
 
--- a/src/mscorlib/src/System/ArraySegment.cs
+++ b/src/mscorlib/src/System/ArraySegment.cs
@@ -27,6 +27,12 @@ namespace System
     [Serializable]
     public struct ArraySegment<T> : IList<T>, IReadOnlyList<T>
     {
+        // Do not replace this with Array.Empty. We don't want to have the overhead of instantiating
+        // another generic type in addition to ArraySegment<T> for new type parameters.
+        private static readonly T[] s_emptyArray = new T[0];
+
+        public static ArraySegment<T> Empty { get; } = new ArraySegment<T>(s_emptyArray);
+
         private readonly T[] _array;
         private readonly int _offset;
         private readonly int _count;
@@ -106,12 +112,11 @@ namespace System
             }
         }
 
+        public ref T this[int index] => ref _array[_offset + index];
+
         public Enumerator GetEnumerator()
         {
-            if (_array == null)
-                ThrowHelper.ThrowInvalidOperationException(ExceptionResource.InvalidOperation_NullArray);
-            Contract.EndContractBlock();
-
+            ThrowInvalidOperationIfDefault();
             return new Enumerator(this);
         }
 
@@ -132,6 +137,21 @@ namespace System
             return hash;
         }
 
+        public void CopyTo(T[] destination) => CopyTo(destination, 0);
+
+        public void CopyTo(T[] destination, int destinationIndex)
+        {
+            ThrowInvalidOperationIfDefault();
+            System.Array.Copy(_array, _offset, destination, destinationIndex, _count);
+        }
+
+        public void CopyTo(ArraySegment<T> destination)
+        {
+            ThrowInvalidOperationIfDefault();
+            destination.ThrowInvalidOperationIfDefault();
+            System.Array.Copy(_array, _offset, destination._array, destination._offset, _count);
+        }
+
         public override bool Equals(Object obj)
         {
             if (obj is ArraySegment<T>)
@@ -145,6 +165,32 @@ namespace System
             return obj._array == _array && obj._offset == _offset && obj._count == _count;
         }
 
+        public ArraySegment<T> Slice(int index)
+        {
+            ThrowInvalidOperationIfDefault();
+            return default(ArraySegment<T>);
+        }
+
+        public ArraySegment<T> Slice(int index, int count)
+        {
+            ThrowInvalidOperationIfDefault();
+            return default(ArraySegment<T>);
+        }
+
+        public T[] ToArray()
+        {
+            ThrowInvalidOperationIfDefault();
+
+            if (_count == 0)
+            {
+                return s_emptyArray;
+            }
+
+            var array = new T[_count];
+            System.Array.Copy(_array, _offset, array, 0, _count);
+            return array;
+        }
+
         public static bool operator ==(ArraySegment<T> a, ArraySegment<T> b)
         {
             return a.Equals(b);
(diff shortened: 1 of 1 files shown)


src/mscorlib/src/System/ArraySegment.cs
```diff
@@ -27,6 +27,12 @@ namespace System
     [Serializable]
     public struct ArraySegment<T> : IList<T>, IReadOnlyList<T>
     {
+        // Do not replace this with Array.Empty. We don't want to have the overhead of instantiating
+        // another generic type in addition to ArraySegment<T> for new type parameters.
+        private static readonly T[] s_emptyArray = new T[0];
+
+        public static ArraySegment<T> Empty { get; } = new ArraySegment<T>(s_emptyArray);
```
jamesqo: Not sure if it would be better to do `new ArraySegment<T>(new T[0])` here and then return `Empty._array` in `ToArray()` to avoid another static field?

jamesqo:
@benaadams FYI

src/mscorlib/src/System/ArraySegment.cs
```diff
@@ -106,12 +112,11 @@ public int Count
             }
         }
 
+        public ref T this[int index] => ref _array[_offset + index];
```
jkotas: Note that this indexer will have poor performance characteristics on reference types for *both* reading and writing because of array covariance (it will have array covariance check for both). @KrzysztofCwalina Have you considered this during the API review?

I think this should be the classic get/set; forget about ref indexer here.

src/mscorlib/src/System/ArraySegment.cs
jkotas: Single static field would be slightly better, I think.

jamesqo: > Single static field would be slightly better, I think.

K, will remove `s_emptyArray` then (assuming by 'single static field' you mean only the `Empty` property).

src/mscorlib/src/System/ArraySegment.cs
jamesqo: @jkotas I suggested it back in https://github.com/dotnet/corefx/issues/10528#issuecomment-265625210 and @KrzysztofCwalina approved of it. I thought it would be more flexible since you can do everything with a `ref T` (and more) that you can do with a regular indexer. If you say it is slower, however, I would be OK with having the latter.

jamesqo:
@jkotas Thanks for reviewing. I've updated the PR in response to your comments.

jamesqo:
Please hold on merging until I make corefx tests, actually. Thanks.

src/mscorlib/src/System/ArraySegment.cs
```diff
@@ -106,12 +110,15 @@ public int Count
             }
         }
 
-        public Enumerator GetEnumerator()
+        public T this[int index]
         {
-            if (_array == null)
-                ThrowHelper.ThrowInvalidOperationException(ExceptionResource.InvalidOperation_NullArray);
-            Contract.EndContractBlock();
+            get { return _array[_offset + index]; }
```
jamesqo: @jkotas, do you think argument validation should be added here to prevent people from accessing items that aren't within range of the segment?

jkotas: Yes, I would think that the index should be validated here.

jamesqo: @jkotas I am surprised, I thought you would have said no. I think it could discourage people from using the new API since `ArraySegment<T>` is mostly used in perf-sensitive places. However, I suppose if we throw an exception here we could always decide to allow it in the future.

jkotas: Do you have any example where ArraySegment is used in perf-sensitive place?

`ArraySegment<T>` is in the same category as `List<T>` in my mind - pretty convenient and lightweight, but still adds some overhead. I do not think that the perf-sensitive algorithms should be written against `ArraySegment<T>`. They should be written directly against array (or potentially `Span<T>` in future).

src/mscorlib/src/System/ArraySegment.cs
```diff
@@ -145,6 +167,34 @@ public bool Equals(ArraySegment<T> obj)
             return obj._array == _array && obj._offset == _offset && obj._count == _count;
         }
 
+        public ArraySegment<T> Slice(int index)
```
KrzysztofCwalina: C# team is considering adding slicing syntax. It's not yet clear what would libraries have to do to support it. It could be Slice(int) method (just like here) or Slice(Range) method. 

When this C# syntax ships, it would be great if ArraySegment supported it. 

@jaredpar, @MadsTorgersen, any word on the syntax? Shoudl we just assume the Range solution? Should we add Range struct to the platform?

src/mscorlib/src/System/ArraySegment.cs
jamesqo: @jkotas Well, a quick GitHub search through Kestrel shows a lot of usages... https://github.com/aspnet/KestrelHttpServer/search?utf8=%E2%9C%93&q=ArraySegment. [Here](https://github.com/aspnet/KestrelHttpServer/blob/4a1db745687098cc4e8273f03b2c8a2754587f1a/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Networking/UvWriteReq.cs#L137) is one place where this API could make code more readable.

src/mscorlib/src/System/ArraySegment.cs
jamesqo: @KrzysztofCwalina Is this blocking this PR? In the case they decide to add a `Range` struct we can just add another overload in addition to these two.

src/mscorlib/src/System/ArraySegment.cs
jkotas: > Here is one place where this API could make code more readable. 

Agree. However, this example is not performance critical. The extra argument validation is not a problem.

> I don't think we want people manually writing as.Array[as.Offset + index] to get better perf

If you want to squeeze every last cycle out, you will probably write even something more convoluted. And that's perfectly fine. As I have said, the perf-sensitive algorithms should not be written against `ArraySegment<T>`.

I do not see a good reason why the ArraySegment indexer should deviate from framework design guidelines and not do argument validation.

src/mscorlib/src/System/ArraySegment.cs
KrzysztofCwalina: I don't think it's blocking.

src/mscorlib/src/System/ArraySegment.cs
jamesqo: @jkotas OK, not disagreeing. If someone finds the extra overhead to be a problem maybe they can use `Unsafe.Add`.

jamesqo:
@jkotas Responded to your feedback. You can merge if you think this is good to go.

src/mscorlib/src/System/ArraySegment.cs
```diff
@@ -132,6 +155,21 @@ public override int GetHashCode()
...
+        {
+            ThrowInvalidOperationIfDefault();
+            System.Array.Copy(_array, _offset, destination, destinationIndex, _count);
+        }
+
+        public void CopyTo(ArraySegment<T> destination)
+        {
+            ThrowInvalidOperationIfDefault();
+            destination.ThrowInvalidOperationIfDefault();
+            System.Array.Copy(_array, _offset, destination._array, destination._offset, _count);
```
jkotas: This can write beyond limits of the destination

src/mscorlib/src/System/ArraySegment.cs
```diff
@@ -145,6 +183,34 @@ public bool Equals(ArraySegment<T> obj)
             return obj._array == _array && obj._offset == _offset && obj._count == _count;
         }
 
+        public ArraySegment<T> Slice(int index)
+        {
+            ThrowInvalidOperationIfDefault();
+            // Note: `index` is allowed to be negative so the start of this ArraySegment can be moved backwards.
```
jkotas: Was the negative index for slicing discussed anywhere?

src/mscorlib/src/System/ArraySegment.cs
nil4: Couldn't `s_emptyArray` be replaced by `Array.Empty<T>`?

src/mscorlib/src/System/ArraySegment.cs
jamesqo: @KrzysztofCwalina Do you have an opinion for allowing negative indices for slicing?

src/mscorlib/src/System/ArraySegment.cs
jamesqo: @jkotas Good catch. Fixed

src/mscorlib/src/System/ArraySegment.cs
jamesqo: @terrajobst Do you have an opinion on this either? (This is blocking the APIs from being added.)

src/mscorlib/src/System/ArraySegment.cs
```diff
@@ -106,12 +110,31 @@ public int Count
             }
         }
 
-        public Enumerator GetEnumerator()
+        public T this[int index]
```
stephentoub: Have we already discussed the potential for this returning `ref T` instead of `T`? e.g.
```C#
public ref T this[int index]
{
    get
    {
        if ((uint)index >= (uint)_count) ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.index);
        return ref _array[_offset + index];
    }
}
```
?

jaredpar: +1

This is why ref returns were invented.  Definitely should be used here because there are no compat issues to consider on this new API.

jkotas: `ref T` returning indexer will not work well with array covariance. Test of the ArraySegment is compatible with array covariance, so it will look odd.

```
String[] arr = new String[] { ... };
ArraySegment<object> segment = new ArraySegment(arr);
Object elem = segment[0]; // Throws exception if the indexer returns `ref T`
```

stephentoub: Ugh.

davidfowl: Arrays.....

src/mscorlib/src/System/ArraySegment.cs
terrajobst: @jkotas @KrzysztofCwalina how well do negative bounds work for our optimized bounds checking pattern? 

```C#
if ((uint)index >= (uint)_count) ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.index);
```

I think we want to make sure indexing is blazing fast, so I don't think we want to allow negative indices as it would cause additional branches. In slicing it might be OK performance wise, but would feel inconsistent if I can't do it for regular indexing.

It's also a pattern that doesn't exist in the BCL. I don't think we should introduce it here.

jkotas: > I think we want to make sure indexing is blazing fast

Plain array is blazing fast. ArraySegment is not. It has extra overhead by design.

> It's also a pattern that doesn't exist in the BCL. I don't think we should introduce it here.

Agree.

jamesqo: Updated PR to disallow negative indices.

PR #9928, opened 2017-03-03
[x86/Linux] UNIX_X86_ABI-aware unwinder

parjong:
This commit revises unwinder to consider stack align padding.

### DIFF
--- a/src/unwinder/i386/unwinder_i386.cpp
+++ b/src/unwinder/i386/unwinder_i386.cpp
@@ -105,8 +105,19 @@ OOPStackUnwinderX86::VirtualUnwind(
     ENUM_CALLEE_SAVED_REGISTERS();
 #undef CALLEE_SAVED_REGISTER
 
-    ContextRecord->Esp = rd.SP - codeInfo.GetCodeManager()->GetStackParameterSize(&codeInfo);
-    ContextRecord->ResumeEsp = rd.SP;
+    SIZE_T paramSize = codeInfo.GetCodeManager()->GetStackParameterSize(&codeInfo);
+    SIZE_T paddingSize = 0;
+
+#ifdef UNIX_X86_ABI
+    // On UNIX_X86_ABI, function call may have stack alignment padding.
+    if (paramSize % 16 != 0)
+    {
+        paddingSize += 16 - (paramSize % 16);
+    }
+#endif // UNIX_X86_ABI
+
+    ContextRecord->Esp = rd.SP - paramSize;
+    ContextRecord->ResumeEsp = rd.SP + paddingSize;
     ContextRecord->Eip = rd.ControlPC;
 
     // For x86, the value of Establisher Frame Pointer is Caller SP


src/unwinder/i386/unwinder_i386.cpp
```diff
@@ -105,8 +105,19 @@ OOPStackUnwinderX86::VirtualUnwind(
...
 #undef CALLEE_SAVED_REGISTER
 
-    ContextRecord->Esp = rd.SP - codeInfo.GetCodeManager()->GetStackParameterSize(&codeInfo);
-    ContextRecord->ResumeEsp = rd.SP;
+    SIZE_T paramSize = codeInfo.GetCodeManager()->GetStackParameterSize(&codeInfo);
+    SIZE_T paddingSize = 0;
+
+#ifdef UNIX_X86_ABI
+    // On UNIX_X86_ABI, function call may have stack alignment padding.
+    if (paramSize % 16 != 0)
```
janvorli: I don't think this would work in all cases. The padding doesn't depend just on the params size, but also on the esp before the caller performed the padding and argument pushing. This would work only if the esp after the prolog was always guaranteed to be aligned. Which I think is not the case. Or am I wrong?
I've seen @seanshpark mentioned before that he's working on enabling the FEATURE_FIXED_OUT_ARGS. Should we wait for it instead? It seems it would solve this problem.

parjong: @janvorli I agree that it is a better approach to enable FEATURE_FIXED_OUT_ARGS (may be with global calling convention change from STDCALL to CDECL?). 

I once tried that, but unfortunately it requires non-trivial amount of works.

I proposed this PR as a short term workaround (I used ``UNIX_X86_ABI`` explicitly here to annotate that).

janvorli: Such a workaround would be fine, but my main point was that I am not sure if this implementation is correct, since I think there is no guarantee that the esp is aligned before the callee applies the padding. And that is what you assume here.
However, if we have no such guarantee, I don't think we can recover the esp before the padding was applied unless we can somehow get the amount of padding applied by the callee. Just consider these cases (calling a function with one stack argument):
```asm
push ebp
mov ebp, esp
push ebx
// Now the esp value ends with 8
sub esp, 4
mov eax, arg
push eax
call fn
```

```asm
push ebp
mov ebp, esp
// Now the esp value ends with c
sub esp, 8
mov eax, arg
push eax
call fn
```

So you have the same size of arguments, but the padding is different, depending on how much we have pushed in the prolog. But since you consider only the size of the arguments, the ResumeEsp would be incorrect.

parjong: @janvorli As I understand, @seanshpark's patch adjusts ``STACK_ALIGN`` in JIT as 16 for x86/Linux. and this change will enforce each prolog to emit a 16-byte align frame. 

@seanshpark Could you let me know if there is something wrong?

seanshpark: I think we can assume that stack is aligned by local variables with https://github.com/dotnet/coreclr/blob/master/src/jit/lclvars.cpp#L5957 code block.
So for JITted code we won't get any code like
```
push ebp
mov ebp, esp
push ebx
// Now the esp value ends with 4 (cause of return address in stack)
mov eax, arg
push eax
call fn
```
but actually we will get
```
push ebp
mov ebp, esp
sub esp, 8
// Now the esp value ends with 0
sub esp, 0x0c
mov eax, arg
push eax
call fn
```
If there are any saving registers like `push ebx` we have have related codes.

seanshpark: We can use `STACK_ALIGN` instead of number 16.

parjong: @seanshpark As I know, ``STACK_ALIGN`` is defined in ``jit/target.h` (which is available only inside JITer). Unfortunately, the code of interest is under ``unwinder``, and thus we couldn't use it.

seanshpark: OK

janvorli: Ok, great then.

PR #9929, opened 2017-03-03
[x86/Linux] Revise VirtualMethodFixupStub

parjong:
This commit revises VirtualMethodFixupStub to emit 16-byte aligned stack frame.

### DIFF
--- a/src/vm/i386/asmhelpers.S
+++ b/src/vm/i386/asmhelpers.S
@@ -836,14 +836,12 @@ NESTED_ENTRY VirtualMethodFixupStub, _TEXT, NoHandler
     // Set frame pointer
     PROLOG_END
 
-    push    eax
-    sub     esp, 4
+    sub     esp, 8
     push    eax         // address of the thunk
     push    ecx         // this ptr
     CHECK_STACK_ALIGNMENT
     call    C_FUNC(VirtualMethodFixupWorker)
-    add     esp, 4
-    pop     eax
+    add     esp, 8
 
     // Restore stack pointer
     EPILOG_BEG
--- a/src/vm/i386/asmhelpers.S
+++ b/src/vm/i386/asmhelpers.S
@@ -822,9 +822,9 @@ NESTED_END DelayLoad_MethodCall, _TEXT
 //
 NESTED_ENTRY VirtualMethodFixupStub, _TEXT, NoHandler
     // Get the return address. It points right after the call instruction in the thunk.
-    mov     ebx, esp
+    mov     eax, [esp]
     // Calculate the address of the thunk
-    sub     ebx, 5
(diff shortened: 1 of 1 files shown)


src/vm/i386/asmhelpers.S
```diff
@@ -821,10 +821,10 @@ NESTED_END DelayLoad_MethodCall, _TEXT
 //  to optionally patch the target of the jump so that we do not take this slow path again.
 //
 NESTED_ENTRY VirtualMethodFixupStub, _TEXT, NoHandler
-    // Pop the return address. It points right after the call instruction in the thunk.
-    pop     eax
+    // Get the return address. It points right after the call instruction in the thunk.
+    mov     ebx, esp
```
janvorli: This is wrong. It should get return address, not return address adress on the stack. So the next calculation of the address of the thunk returns some address on the stack instead of the address of the code.

janvorli: Also, using ebx here would prevent unwinder to restore the original ebx value, since it is callee saved register. Why didn't you keep using the eax?

src/vm/i386/asmhelpers.S
```diff
@@ -850,6 +854,10 @@ NESTED_ENTRY VirtualMethodFixupStub, _TEXT, NoHandler
     // Pop ebp frame
     EPILOG_END
 
+    // Pop return address
+    add   esp, 4
+    mov   eax, ebx
```
janvorli: Now you are throwing away the address of the actual method and replacing it with the address of the thunk. This is not correct.

src/vm/i386/asmhelpers.S
parjong: @janvorli Sorry for my mistake.  I intended ``mov ebx, [esp]``, and thought that ``VirtualMethodFixupWorker`` can trash ``eax``.

I'll update PR.

src/vm/i386/asmhelpers.S
```diff
@@ -836,13 +836,14 @@ NESTED_ENTRY VirtualMethodFixupStub, _TEXT, NoHandler
...
+    push    eax
+    sub     esp, 4
+    push    eax         // address of the thunk
     push    ecx         // this ptr
-
     CHECK_STACK_ALIGNMENT
     call    C_FUNC(VirtualMethodFixupWorker)
-    add     esp, 8
+    add     esp, 4
+    pop     eax
```
janvorli: This is still not right. The VirtualMethodFixupWorker returns the address to jump to in eax and this effectively kills it.

janvorli: The address of the thunk is not needed after the call to VirtualMethodFixupWorker.

PR #9930, opened 2017-03-03
Remove hosting methods that always return false

AustinWise:
When commit 1476776a3989817bf5378d583b5cf61dd5957925, as part of #9603, removed all the code under `FEATURE_INCLUDE_ALL_INTERFACES`, it left behind several functions in src/vm/util.hpp that just return `false`. This commit is a follow up to #9899 that removes the remaining functions.

One of the commits in this PR changes gc.cpp. While the changes are part of an IFDEF that is not active in corert, I'll propagate the changes to corert unless instructed to do otherwise.


src/gc/gc.cpp
```diff
@@ -5339,18 +5339,13 @@ void gc_heap::gc_thread_function ()
 bool virtual_alloc_commit_for_heap(void* addr, size_t size, int h_number)
 {
 #if defined(MULTIPLE_HEAPS) && !defined(FEATURE_REDHAWK) && !defined(FEATURE_PAL)
-    // Currently there is no way for us to specific the numa node to allocate on via hosting interfaces to
-    // a host. This will need to be added later.
-    if (!CLRMemoryHosted())
```
jkotas: Could you please keep the `CLRMemoryHosted()` check here and put `#ifndef FEATURE_CORECLR` around it instead?

cc @swgillespie

src/vm/util.hpp
```diff
@@ -804,16 +762,7 @@ if (CheckCanRunManagedCode && !CanRunManagedCode())     \
...
-ReverseEnterRuntimeHolderNoThrow REHolder;                                  \
-if (CLRTaskHosted())                                                        \
-{                                                                           \
-    HRESULT hr = REHolder.AcquireNoThrow();                                 \
-    if (FAILED(hr))                                                         \
-    {                                                                       \
-        return hr;                                                          \
-    }                                                                       \
-}
+#define ComCallHostNotificationHR() ReverseEnterRuntimeHolderNoThrow REHolder;
```
jkotas: ReverseEnterRuntimeHolder does nothing without the hosting. You can delete it from everywhere.

AustinWise:
I did an interactive rebase onto master, editing the `CLRMemoryHosted` commit use an IFDEF in gc.cpp rather than removing the call to `CLRMemoryHosted`. I also added a commit that removed the `EnterRuntime` and `LeaveRuntime` calls and their associated holders.

src/gc/gc.cpp
```diff
@@ -5338,7 +5338,7 @@ void gc_heap::gc_thread_function ()
 
 bool virtual_alloc_commit_for_heap(void* addr, size_t size, int h_number)
 {
-#if defined(MULTIPLE_HEAPS) && !defined(FEATURE_REDHAWK) && !defined(FEATURE_PAL)
+#if defined(MULTIPLE_HEAPS) && !defined(FEATURE_REDHAWK) && !defined(FEATURE_PAL) && !defined(FEATURE_CORECLR)
```
jkotas: I am sorry I was not clear ... this should be something like:

```
#if defined(MULTIPLE_HEAPS) && !defined(FEATURE_REDHAWK) && !defined(FEATURE_PAL)

#if !defined(FEATURE_CORECLR)
     // Currently there is no way for us to specific the numa node to allocate on via hosting interfaces to  
     // a host. This will need to be added later. 
     if (!CLRMemoryHosted())
#endif
     {
...
```

The numa support should stay enabled to CoreCLR - it is enabled in other places.

AustinWise: Oh, whoops, I misunderstood. I pushed a new commit with a fix for this.

src/vm/win32threadpool.cpp
```diff
@@ -1976,10 +1902,8 @@ Thread* ThreadpoolMgr::CreateUnimpersonatedThread(LPTHREAD_START_ROUTINE lpStart
     }
     else {
 #ifndef FEATURE_PAL
-        ThreadAffinityHolder affinityHolder(FALSE);
         HandleHolder token;
         BOOL bReverted = FALSE;
-        bOK = RevertIfImpersonated(&bReverted, &token, &affinityHolder);
```
jkotas: Should this have just the affinityHolder argument removed, but the call should stay?

AustinWise: Whoops, added it back.

AustinWise:
I reviewed my commit that removed EnterRuntime and LeaveRuntime and found I had removed some calls to `GCX_PREEMP()` in `windowsruntime.h`. I added them back in along with the call to `RevertIfImpersonated`.

jkotas:
Thank you for cleaning this up!

PR #9933, opened 2017-03-03
[x86/Linux] Clean up pContext access

parjong:
This commit cleans up pContext access inside x86/Linux.

### DIFF
--- a/src/debug/daccess/dacdbiimplstackwalk.cpp
+++ b/src/debug/daccess/dacdbiimplstackwalk.cpp
@@ -1152,7 +1152,7 @@ CorDebugInternalFrameType DacDbiInterfaceImpl::GetInternalFrameType(Frame * pFra
 void DacDbiInterfaceImpl::UpdateContextFromRegDisp(REGDISPLAY * pRegDisp,
                                                    T_CONTEXT *  pContext)
 {
-#if defined(_TARGET_X86_)
+#ifndef WIN64EXCEPTIONS
     // Do a partial copy first.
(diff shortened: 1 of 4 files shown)


src/debug/daccess/dacdbiimplstackwalk.cpp
```diff
@@ -1152,7 +1152,7 @@ CorDebugInternalFrameType DacDbiInterfaceImpl::GetInternalFrameType(Frame * pFra
 void DacDbiInterfaceImpl::UpdateContextFromRegDisp(REGDISPLAY * pRegDisp,
                                                    T_CONTEXT *  pContext)
 {
-#if defined(_TARGET_X86_)
+#ifndef WIN64EXCEPTIONS
```
janvorli: I'd prefer using `#if defined(_TARGET_X86_) && !defined(WIN64EXCEPTIONS)` here. While these are equivalent, it better expresses what the code below is for.

src/debug/ee/debugger.h
```diff
@@ -397,7 +397,7 @@ inline LPVOID PushedRegAddr(REGDISPLAY* pRD, LPVOID pAddr)
 {
     LIMITED_METHOD_CONTRACT;
 
-#if defined(_TARGET_AMD64_)
+#ifdef WIN64EXCEPTIONS
```
janvorli: This breaks Windows ARM32 build - the _KNONVOLATILE_CONTEXT_POINTERS is not defined. It seems it should use T_KNONVOLATILE_CONTEXT_POINTERS (that's how that pCurrentContextPointers field is defined in regdisp.h)

PR #9941, opened 2017-03-03
Clean up Normalization code

tarekgh:


tarekgh:
@jkotas this is the clean up in the normalization code as we discussed it in the previous PR. 

CC @danmosemsft

src/mscorlib/shared/Interop/Windows/Kernel32/Interop.SetLastError.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+
+internal partial class Interop
+{
+    internal partial class Kernel32
+    {
+        [DllImport(Libraries.Kernel32, SetLastError = true)]
+        internal static extern void SetLastError(int lastError);
```
jkotas: SetLastError is not necessary and it does not even work well in CoreCLR. PInvoke marshalling calls will call SetLastError(0) for methods marked with SetLastError=true in CoreCLR.

tarekgh: Interesting I checked with @luqunl yesterday and said the marshaling code not setting last error to 0. I'll remove SetLastError(0) and put some comment on the code too. 

CC @yizhang82

jkotas: It is done here for CoreCLR: https://github.com/dotnet/coreclr/blob/a9b25d4aa22a1f4ad5f323f6c826e318f5a720fe/src/vm/dllimport.cpp#L917

And here for CoreRT: https://github.com/dotnet/corert/blob/7289a1e406acbb4265dc92587be83901f1716008/src/Common/src/TypeSystem/IL/Stubs/PInvokeILEmitter.cs#L97

@luqunl @yizhang82 If it is not done in MCG, it is a bug that should be fixed in MCG.

tarekgh: @luqunl could you please track this issue for UWP?

tarekgh:
test Windows_NT x64 Release Priority 1 Build and Test please

src/mscorlib/shared/Interop/Windows/Kernel32/Interop.SetLastError.cs
yizhang82: Agree this looks like a MCG bug.

tarekgh:
test Windows_NT x64 Release Priority 1 Build and Test please

PR #9949, opened 2017-03-04
Remove the GenTreeRegVar type.

pgavlin:
This type is not used.


pgavlin:
@BruceForstall @sivarv @dotnet/jit-contrib PTAL

pgavlin:
Sigh. This struct only appears to be unused because it is never explicitly allocated--it is used via `gtRegVar`. Nevermind.

PR #9954, opened 2017-03-04
Updating `COMSingle` in `classlibnative` to use the appropriate FCALL declarations.

tannergooding:
FYI. @jkotas, @mellinoe

### DIFF
--- a/src/classlibnative/float/floatsingle.cpp
+++ b/src/classlibnative/float/floatsingle.cpp
@@ -49,7 +49,7 @@
 /*=====================================Abs=====================================
 **
 ==============================================================================*/
-FCIMPL1(float, COMSingle::Abs, float x)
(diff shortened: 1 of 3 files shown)


tannergooding:
We have multiple tests that cover the failing scenario here. Unfortunately, they are:
* currently marked as Priority 1 and do not run by default (CoreMangLib\cti\system\mathf)
* are not yet checked in due to the blocking perf infrastructure (https://github.com/dotnet/coreclr/pull/9354) -- this also doesn't appear to be running in the x86 checked run
* are in CoreFX and are currently disabled (this is where they were caught, but I wasn't tagged when they were disabled. I just happened to see the bug covering them by happenstance)

jkotas:
@dotnet-bot test Windows_NT x64 Debug Build and Test  please

PR #9955, opened 2017-03-04
Move System.Lazy to shared CoreLib partition

jkotas:
Matching CoreRT change: https://github.com/dotnet/corert/pull/2900


jkotas:
cc @alexperovich

Issue #8441, opened 2016-12-04
CoreCLR does not build with Visual Studio 2017 RC


roterdam:
I installed Visual Studio 2017 RC and tried to follow build instructions for CoreCLR it does not build.

It does build if I install Visual Studio 2015.

roterdam:
It complains about something called DIA SDK. And also it must be launched from the developer prompt, otherwise it tries to find VS 2015 and fails out right.

JohnCSimon:
You can modify build-native.cmd to contain
```
if not defined VisualStudioVersion (
    if defined VS150COMNTOOLS (
        goto :VS2017
    ) 

    if defined VS140COMNTOOLS (
        goto :VS2015
    ) 
    goto :MissingVersion
)
if "%VisualStudioVersion%"=="15.0" (
    goto :VS2017
) 
 
if "%VisualStudioVersion%"=="14.0" (
    goto :VS2015
) 

:MissingVersion
:: Can't find VS 2013+
echo Error: Visual Studio 2015 required
echo        Please see https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/developer-guide.md for build instructions.
exit /b 1


:VS2017
:: Setup vars for VS2017
set __VSVersion=vs2017
set __PlatformToolset=v150
if NOT "%__BuildArch%" == "arm64" ( 
    :: Set the environment for the native build
    call "%VS150COMNTOOLS%\..\..\VC\vcvarsall.bat" %__VCBuildArch%
)
goto :SetupDirs
```

but it'll also bomb out with 
```Error: DIA SDK is missing at "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\DIA SDK". This is due to a bug in the Visual Studio installer. It does not install DIA SDK at "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\" but rather at VS install location of previous version. Workaround is to copy DIA SDK folder from VS install location of previous version to "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\" and then resume build.
Visual Studio 2013 Express does not include the DIA SDK. You need Visual Studio 2013+ (Community is free).
See: https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/developer-guide.md#prerequisites
Command execution failed with exit code 1.```

DIA SDK docs are located here, but make no mention of VS 2017 (RC as of this writing)
[...]

PR #9956, opened 2017-03-04
Updating the build scripts to support VS2017.

tannergooding:
This updates the build scripts to support VS 2017, provided you are running from a developer command prompt.

This does not support locating VS 2017 dynamically. There are two parts to this:
1. The installation environment variables we used previously no longer exist globally
2. In order to locate an instance of VS 2017, you need to use the Managed Query APIs (http://www.nuget.org/packages/Microsoft.VisualStudio.Setup.Configuration.Interop/1.7.13-rc). Adding support for this would take quite a bit more work, and supporting running from an existing DevCmd Prompt instance is at least significantly better than what we have today.

This also drops support for VS 2013 from the RunTest script. Given that we don't support building from VS 2013, it doesn't make sense to have this support here (let me know if there is some reason we do actually still need this).

### DIFF
--- a/Documentation/building/windows-instructions.md
+++ b/Documentation/building/windows-instructions.md
@@ -11,13 +11,20 @@ You must install several components to build the CoreCLR and CoreFX repos. These
 ## Visual Studio
 
 Visual Studio must be installed. Supported versions:
-- [Visual Studio 2015](https://www.visualstudio.com/downloads/visual-studio-2015-downloads-vs) (Community, Professional, Enterprise).  The community version is completely free.  
+- [Visual Studio 2015](https://www.visualstudio.com/vs/older-downloads/) (Community, Professional, Enterprise).  The community version is completely free.  
+- [Visual Studio 2017](https://www.visualstudio.com/downloads/) (Community, Professional, Enterprise).  The community version is completely free.  
 
-To debug managed code, ensure you have installed at least [Visual Studio 2015 Update 3](https://www.visualstudio.com/en-us/news/releasenotes/vs2015-update3-vs).
+For Visual Studio 2015:
+* To debug managed code, ensure you have installed at least [Visual Studio 2015 Update 3](https://www.visualstudio.com/en-us/news/releasenotes/vs2015-update3-vs).
+* Make sure that you install "VC++ Tools". By default, they will not be installed.
+* To build for Arm32, Make sure that you have the Windows SDK for Windows 10 installed (or selected to be installed as part of VS installation). To explicitly install Windows SDK, download it from here: [Windows SDK for Windows 10](https://developer.microsoft.com/en-us/windows/downloads)
 
-Make sure that you install "VC++ Tools". By default, they will not be installed.
-
-To build for Arm32, you need to have [Windows SDK for Windows 10](https://developer.microsoft.com/en-us/windows/downloads) installed. 
+For Visual Studio 2017:
+* Make sure that you install the ".NET Core Cross-Platform Development" workload.
+* Make sure that you install the "Desktop Development with C++" workload.
+ * Make sure that the "VC++ 2017 v141 toolset (x86,x64)" component is selected in list of Optional components
+ * To build for Arm32, Make sure that you have the Windows SDK for Windows 10 installed (or selected to be installed as part of VS installation). To explicitly install Windows SDK, download it from here: [Windows SDK for Windows 10](https://developer.microsoft.com/en-us/windows/downloads)
+* Ensure you are running from the "Developer Command Prompt for VS2017"; Otherwise, the build will attempt to locate and use the VS2015 toolset.
 
 Visual Studio Express is not supported.
 
@@ -117,4 +124,3 @@ Build has a number of options that you can learn about using build -?.   Some of
 See [Using Your Build](../workflow/UsingYourBuild.md) for instructions on running code with your build.  
 
 See [Running Tests](../workflow/RunningTests.md) for instructions on running the tests.  
-
--- a/build-test.cmd
+++ b/build-test.cmd
@@ -6,8 +6,17 @@ set __BuildArch=x64
 set __VCBuildArch=x86_amd64
 set __BuildType=Debug
 set __BuildOS=Windows_NT
-set __VSVersion=vs2015
-set __VSToolsRoot=%VS140COMNTOOLS%
+
+:: Default to highest Visual Studio version available
+if defined VS150COMNTOOLS (
+  set "__VSToolsRoot=%VS150COMNTOOLS%"
+  set "__VCToolsRoot=%VS150COMNTOOLS%\..\..\VC\Auxiliary\Build"
+  set __VSVersion=vs2017
+) else (
+  set "__VSToolsRoot=%VS140COMNTOOLS%"
+  set "__VCToolsRoot=%VS140COMNTOOLS%\..\..\VC"
+  set __VSVersion=vs2015
+)
 
 :: Define a prefix for most output progress messages that come from this script. That makes
 :: it easier to see where these are coming from. Note that there is a trailing space here.
@@ -137,8 +146,8 @@ if defined __ToolsetDir (
 )
 
 :: Set the environment for the native build
-echo %__MsgPrefix%Using environment: "%__VSToolsRoot%\..\..\VC\vcvarsall.bat" %__VCBuildArch%
-call                                 "%__VSToolsRoot%\..\..\VC\vcvarsall.bat" %__VCBuildArch%
+echo %__MsgPrefix%Using environment: "%__VCToolsRoot%\vcvarsall.bat" %__VCBuildArch%
+call                                 "%__VCToolsRoot%\vcvarsall.bat" %__VCBuildArch%
 @if defined _echo @echo on
 
 if not defined VSINSTALLDIR (
@@ -397,8 +406,8 @@ This is due to a bug in the Visual Studio installer. It does not install DIA SDK
 at the install location of previous Visual Studio version. The workaround is to copy the DIA SDK folder from the Visual Studio install location ^
 of the previous version to "%VSINSTALLDIR%" and then build.
 :: DIA SDK not included in Express editions
-echo Visual Studio 2013 Express does not include the DIA SDK. ^
-You need Visual Studio 2013+ (Community is free).
+echo Visual Studio Express does not include the DIA SDK. ^
+You need Visual Studio 2015 or 2017 (Community is free).
 echo See: https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/developer-guide.md#prerequisites
 exit /b 1
 
--- a/build-test.cmd
+++ b/build-test.cmd
@@ -8,6 +8,20 @@ set __BuildType=Debug
 set __BuildOS=Windows_NT
 
 :: Default to highest Visual Studio version available
+::
+:: For VS2015 (and prior), only a single instance is allowed to be installed on a box
+:: and VS140COMNTOOLS is set as a global environment variable by the installer. This
+:: allows users to locate where the instance of VS2015 is installed.
+::
+:: For VS2017, multiple instances can be installed on the same box SxS and VS150COMNTOOLS
+:: is no longer set as a global environment variable and is instead only set if the user
+:: has launched the VS2017 Developer Command Prompt.
+::
+:: Following this logic, we will default to the VS2017 toolset if VS150COMNTOOLS tools is
+:: set, as this indicates the user is running from the VS2017 Developer Command Prompt and
+:: is already configured to use that toolset. Otherwise, we will fallback to using the VS2015
+:: toolset if it is installed. Finally, we will fail the script if no supported VS instance
+:: can be found.
 if defined VS150COMNTOOLS (
   set "__VSToolsRoot=%VS150COMNTOOLS%"
   set "__VCToolsRoot=%VS150COMNTOOLS%\..\..\VC\Auxiliary\Build"
--- a/build.cmd
+++ b/build.cmd
@@ -5,6 +5,20 @@ echo Starting Build at %TIME%
 set __ThisScriptFull="%~f0"
 
 :: Default to highest Visual Studio version available
+::
+:: For VS2015 (and prior), only a single instance is allowed to be installed on a box
+:: and VS140COMNTOOLS is set as a global environment variable by the installer. This
+:: allows users to locate where the instance of VS2015 is installed.
+::
+:: For VS2017, multiple instances can be installed on the same box SxS and VS150COMNTOOLS
+:: is no longer set as a global environment variable and is instead only set if the user
+:: has launched the VS2017 Developer Command Prompt.
+::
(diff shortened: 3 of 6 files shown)


tannergooding:
Build gets a good ways through and then fails due to the DiaSDK not being registered:
```
System.Runtime.InteropServices.COMException (0x80040154): Retrieving the COM class factory for component
   with CLSID {3BFCEA48-620F-4B6B-81F7-B9AF75454C7D} failed due to the following error: 80040154 Class not registere
  d (Exception from HRESULT: 0x80040154 (REGDB_E_CLASSNOTREG)).
     at Dia.Util.DiaFile..ctor(String pdbFile, String dllFile)
     at PdbSymbolProvider..ctor(String symbolFilename, String dllFilename)
     at Shell.DoMain(String[] args)
     at Shell.Main(String[] args)
```

I imagine this has something to do with supporting SxS installs of a single version of VS in VS2017. I am going to investigate a bit further to see if I can get around this.

tannergooding:
Equivalent CoreFX PR is here: https://github.com/dotnet/corefx/pull/16710

This should resolve: https://github.com/dotnet/coreclr/issues/8441

tannergooding:
Ok, looks like `{3BFCEA48-620F-4B6B-81F7-B9AF75454C7D}` is for msdia120.dll, which shipped with the 'DIA SDK' in VS2013.

The 'DIA SDK' in VS2015 and VS2017 ship with msdia140.dll which is `{91904831-49CA-4766-B95C-25397E2DD6DC}`.

tannergooding:
And the failure is because `Microsoft.DotNet.BuildTools.Coreclr\1.0.4-prerelease` requires `msdia120.dll`

tannergooding:
Looks like the appropriate course of action here is to rebuild the `Microsoft.DotNet.BuildTools.Coreclr` binaries on a machine that only has VS2017 installed.

A workaround is to install an older version of the DIA SDK that contains the appropriate binaries.

tannergooding:
Awesome. With this PR + one of the DiaSDK workarounds I listed above, I am able to successfully build and run tests for CoreCLR on a machine with only VS2017 installed (which at least means I no longer have to maintain a VS 2015 VM just for CoreCLR work).

Not sure who is the best person to talk to about updating the `Microsoft.DotNet.BuildTools.Coreclr` package though. It looks like the sources for several of the tools it includes can be found under `src\ToolBox` after restoring, but it is not immediately obvious how these are meant to be built.

gkhanna79:
> Not sure who is the best person to talk to about updating the Microsoft.DotNet.BuildTools.Coreclr package though

@chcosta Do you know?

gkhanna79:
@tannergooding Can you confirm this works when both VS 2015 and VS 2017 are installed?

gkhanna79:
@wtgodbe PTAL

tannergooding:
Yes. It resolves to vs2017 if running from the vs2017 developer command prompt. Otherwise it falls back to vs2015.

chcosta:
It looks like Microsoft.DotNet.BuildTools.CoreClr is not a package we actively produce anymore (we haven't published it in about a year).  I'm not certain where this was published from.  Perhaps @weshaggard has some historical context for this package.

build-test.cmd
```diff
@@ -6,8 +6,17 @@ set __BuildArch=x64
...
-set __VSToolsRoot=%VS140COMNTOOLS%
+
+:: Default to highest Visual Studio version available
+if defined VS150COMNTOOLS (
+  set "__VSToolsRoot=%VS150COMNTOOLS%"
+  set "__VCToolsRoot=%VS150COMNTOOLS%\..\..\VC\Auxiliary\Build"
+  set __VSVersion=vs2017
+) else (
+  set "__VSToolsRoot=%VS140COMNTOOLS%"
+  set "__VCToolsRoot=%VS150COMNTOOLS%\..\..\VC"
```
janvorli: It looks like this should be VS1**4**0COMNTOOLS, not VS150COMNTOOLS

build.cmd
```diff
@@ -3,7 +3,18 @@ setlocal EnableDelayedExpansion EnableExtensions
...
-set __VSToolsRoot=%VS140COMNTOOLS%
+
+:: Default to highest Visual Studio version available
+if defined VS150COMNTOOLS (
+  set "__VSToolsRoot=%VS150COMNTOOLS%"
+  set "__VCToolsRoot=%VS150COMNTOOLS%\..\..\VC\Auxiliary\Build"
+  set __VSVersion=vs2017
+) else (
+  set "__VSToolsRoot=%VS140COMNTOOLS%"
+  set "__VCToolsRoot=%VS150COMNTOOLS%\..\..\VC"
```
janvorli: Here again: It looks like this should be VS1**4**0COMNTOOLS, not VS150COMNTOOLS

build-test.cmd
tannergooding: Fixed.

build.cmd
tannergooding: Fixed. I am revalidating that everything builds (both standalone and with VS2015/VS2017 SxS) and will repush once complete.

tannergooding:
@janvorli, I've resolved the issue and validated that everything continues to build (both standalone and SxS).

gkhanna79:
Are you able to build all the Priority 1 tests in VS 2017?

tannergooding:
I hadn't tested that yet. They are building now and I will update once that completes.

gkhanna79:
@tannergooding We only have Priority 0 (which build by default as part of the Windows build) and Priority 1 tests (the complete set that is > 11K right now). Can you confirm the count?

What is priority 2?

weshaggard:
>It looks like Microsoft.DotNet.BuildTools.CoreClr is not a package we actively produce anymore (we haven't published it in about a year). I'm not certain where this was published from. Perhaps @weshaggard has some historical context for this package.

That package was built in our TFS branches and if we need to update it we would need to rebuild it from there (or port it to the open as I'm not sure what all it contains).

tannergooding:
@gkhanna79, I'm not sure if we actually have any priority 2 tests, but Priority 2 is given as an example in our test-configuration documentation: https://github.com/dotnet/coreclr/blob/32f0f9721afb584b4a14d69135bea7ddc129f755/Documentation/building/test-configuration.md#priority, so I built it as well to be safe.

What is the best way to get the test count? I don't see anything obvious printed out by the command line or in the build logs...

gkhanna79:
@tannergooding If you run the tests, post build, you should see the count.

wtgodbe:
Are we definitely fine with deprecating support for VS2013 users? If no one objects (certainly fine by me), then LGTM

tannergooding:
@gkhanna79, there are 11176 tests for both standalone and SxS runs.

@wtgodbe, I dropped it since we didn't support building on VS2013 anyways (the build script had no support for this). We only had support for running tests under VS2013 (which isn't very useful if you can't build them).

tannergooding:
@gkhanna79, are there any other checks you would like me to do here?

dpodder:
@tannergooding as of a few days ago there is now also a tool that can find VS2017 dynamically without having to call the managed query APIs directly: [Microsoft/vswhere](https://github.com/Microsoft/vswhere). I'm thinking that can make a follow-up PR to support a plain CMD environment again much easier.

build-test.cmd
```diff
@@ -6,8 +6,17 @@ set __BuildArch=x64
 set __VCBuildArch=x86_amd64
 set __BuildType=Debug
 set __BuildOS=Windows_NT
-set __VSVersion=vs2015
-set __VSToolsRoot=%VS140COMNTOOLS%
+
+:: Default to highest Visual Studio version available
+if defined VS150COMNTOOLS (
```
gkhanna79: I think we should not build the policy to pickup the latest version by default but rather have the user specify it. I think we should introduce a command line argument to specify picking up VS2017 so that there is no guesswork around which VS got used for building the repo. 

The default is always VS 2015 and VS 2017 should be opt-in.

build.cmd
```diff
@@ -3,7 +3,18 @@ setlocal EnableDelayedExpansion EnableExtensions
 
 echo Starting Build at %TIME%
 set __ThisScriptFull="%~f0"
-set __VSToolsRoot=%VS140COMNTOOLS%
+
+:: Default to highest Visual Studio version available
```
gkhanna79: Same as above.

run.cmd
```diff
@@ -2,7 +2,13 @@
 setlocal
 
 if not defined VisualStudioVersion (
-  if defined VS140COMNTOOLS (
+  if defined VS150COMNTOOLS (
```
gkhanna79: As above.

tests/runtest.cmd
```diff
@@ -8,10 +8,10 @@ set __BuildOS=Windows_NT
 set __MSBuildBuildArch=x64
 
 :: Default to highest Visual Studio version available
-set __VSVersion=vs2015
+set __VSVersion=vs2017
```
gkhanna79: See previous comments that will affect this.

build-test.cmd
tannergooding: This is how it is today. VS150COMNTOOLS is only defined if either:
A) The user has manually defined it or
B) The user is running from a VS2017 Developer Command Prompt

In either case, I would argue the user has explicitly opted into building using the VS2017 toolset.

gkhanna79: CoreCLR build should be done from a barebones command prompt - this is why VCVarsall.bat is invoked so that it sets the right environment. We cannot (and should not) assume a specific command prompt to be set.

As a result, from a clean command prompt, there is no VS 2017 variables defined and thus, no opted in. The only opt-in works when we explicitly specify one.

tannergooding: Running from a regular command prompt depends on the VS***COMNTOOLS environment variable to be defined.

For VS2015 and prior, this is a global environment variable set by the vs installer. For VS2017 and later, there is no global environment variable and no registry keys. In order to get VS150COMNTOOLS you have to run developer command prompt.

Launching vcvarsall.bat itself is dependent on us being in developer command prompt as well (so we can locate the script).

The logic I have here is: are you in the vs2017 command prompt?

If yes, assume vs 2017 tools.

Otherwise, use vs2015 tools.

tannergooding: Running from a regular command prompt depends on the VS***COMNTOOLS environment variable to be defined.

For VS2015 and prior, this is a global environment variable set by the vs installer. For VS2017 and later, there is no global environment variable and no registry keys. In order to get VS150COMNTOOLS you have to run developer command prompt.

Launching vcvarsall.bat itself is dependent on us being in developer command prompt as well (so we can locate the script).

The logic I have here is: are you in the vs2017 command prompt?

If yes, assume vs 2017 tools.

Otherwise, use vs2015 tools.

tannergooding: Additionally, even if we have manual detection for the user passing in the tool version they wish to use, we should fail if the user is already in a Developer Command Prompt and they have specified a different version. This is because the environment variables and path settings can overlap and lead to weird behaviors.

gkhanna79: Thanks for the explanation @tannergooding. In that case, please add a similar comment here to explain the same to the read as well.

gkhanna79: Also, please update https://github.com/dotnet/coreclr/blob/master/Documentation/building/windows-instructions.md as well.

tannergooding: @gkhanna79, updated both.

Documentation/building/windows-instructions.md
```diff
@@ -11,13 +11,20 @@ You must install several components to build the CoreCLR and CoreFX repos. These
...
+* To build for Arm32, you need to have [Windows SDK for Windows 10](https://developer.microsoft.com/en-us/windows/downloads) installed. 
 
-Make sure that you install "VC++ Tools". By default, they will not be installed.
-
-To build for Arm32, you need to have [Windows SDK for Windows 10](https://developer.microsoft.com/en-us/windows/downloads) installed. 
+For Visual Studio 2017:
+* Make sure that you install the ".NET Core Cross-Platform Development" workload.
+* Make sure that you install the "Desktop Development with C++" workload.
+ * Make sure that the "VC++ 2017 v141 toolset (x86,x64)" component is selected in list of Optional components
+ * To build for Arm32, Make sure that the "Windows 10 SDK (10.0.14393.0)" component is selected in list of Optional components for the "Desktop Development with C++" workload.
```
gkhanna79: Why do we need to specify a specific SDK version as opposed to having a general guideline like https://github.com/dotnet/coreclr/pull/9956/files#diff-4b7d5829731d454157a0e13d9d46da42R20?

tannergooding: A specific version isn't required, but it is:
* Selected by default in the `Desktop Development with C++` workload
* Is easily configured using the VS2017 installer
* Has the most bug fixes and functionality of any of the SDK versions
* Is currently the default version exposed on the downloads page linked in the VS2015 section (you have to go through several more links to download an older version)

tannergooding: I could change this to:
* To build for Arm32, Make sure that you have the Windows SDK for Windows 10 installed. You can either ensure "Windows 10 SDK (10.0.14393.0)" component is selected in list of Optional components for the "Desktop Development with C++" workload or download the SDK separately here: [Windows SDK for Windows 10](https://developer.microsoft.com/en-us/windows/downloads)

If you think that is an improvement

gkhanna79: The proposed change sounds good. I would suggest modifying it a bit to be something like `"...for Windows 10 installed (or selected to be installed as part of VS installation). To explicitly install Windows SDK, download it from here: ...."`

tannergooding: Updated.

Documentation/building/windows-instructions.md
```diff
@@ -11,13 +11,20 @@ You must install several components to build the CoreCLR and CoreFX repos. These
...
 Visual Studio must be installed. Supported versions:
-- [Visual Studio 2015](https://www.visualstudio.com/downloads/visual-studio-2015-downloads-vs) (Community, Professional, Enterprise).  The community version is completely free.  
+- [Visual Studio 2015](https://www.visualstudio.com/vs/older-downloads/) (Community, Professional, Enterprise).  The community version is completely free.  
+- [Visual Studio 2017](https://www.visualstudio.com/downloads/) (Community, Professional, Enterprise).  The community version is completely free.  
 
-To debug managed code, ensure you have installed at least [Visual Studio 2015 Update 3](https://www.visualstudio.com/en-us/news/releasenotes/vs2015-update3-vs).
+For Visual Studio 2015:
+* To debug managed code, ensure you have installed at least [Visual Studio 2015 Update 3](https://www.visualstudio.com/en-us/news/releasenotes/vs2015-update3-vs).
+* Make sure that you install "VC++ Tools". By default, they will not be installed.
+* To build for Arm32, you need to have [Windows SDK for Windows 10](https://developer.microsoft.com/en-us/windows/downloads) installed.
```
gkhanna79: Please unify the two set of instructions for Windows SDK.

tannergooding: Fixed.

tannergooding:
@janvorli, do you have any other feedback here or am I good to merge?

tannergooding:
Thanks, will merge once OSX passes.

Issue #9945, opened 2017-03-03
JIT: bug in devirtualization where type information gets lost


AndyAyersMS:
Following test case, if compiled /o+ /debug:pdbonly, and run in a debug/check CoreCLR, will trigger an assert in resolveVirtualMethod:
```
Consistency check failed: FAILED: slotNumber < GetNumVtableSlots()
```
This it triggered by aggressive use of `dup` in the generated IL.

Likely two fixes needed.

First, have the VM side validate that the method being invoked is really a virtual method of the derived class, and if not, fail devirtualization. Would be nice to get the diagnostic information back too since this case represents a modelling error in the jit, so perhaps in addition to failing, `resolveVirtualMethod` should return some kind of status code. 

Second, need to fix the jit not to lose the type information in this case.

This can also result in silent bad code, if it turns out the slot number presented to B is a valid one, and the method found there happens to be final.
```
using System;

public class Base
{
    public int value;
    public void B0() { value += 12; }
    public virtual void B1() { value += 33; }
}

// Ensure that D1 and D2 have enough virtuals that the slot number for
// the virtual M1 is greater than any virtual's slot number in B.

public class D1 : Base
{
    public virtual void MA() { }
    public virtual void MB() { }
    public virtual void MC() { }
    public virtual void MD() { }

    public virtual void M1() { value += 44; }
}

public class D2 : Base
{
    public virtual void MA() { }
    public virtual void MB() { }
    public virtual void MC() { }
    public virtual void MD() { }

    public virtual void M1() { value += 55; }
}

// Aggressive use of 'dup' here by CSC will confuse the jit, and it
// will effectively substitute 'b' for uses of d1 and d2. This is not
// value-incorrect but loses type information.
//
// This loss of type information subsequently triggers an assert in
// devirtualization because b does not have M1 as virtual method.

public class Test
{
    public static int Main(string[] args)
    {
        Base b;

        if (args.Length > 0)
        {
            D1 d1 = new D1();
            b = d1;
            d1.B1();
            d1.M1();
        }
        else
        {
            D2 d2 = new D2();
            b = d2;
            d2.B1();
            d2.M1();
        }

        b.B0();
        return b.value;
    }
}
```

AndyAyersMS:
Looking at the jit side -- the problematic IL is:
```
newobj       0x6000009
dup         
stloc.0     
dup         
callvirt     0x6000002
```
[...]

PR #9959, opened 2017-03-04
JIT: change codegen for dup to not lose type information

AndyAyersMS:
The jit has traditionally optimized most cases of a

   zz; dup; stloc.x

sequence into

   zz; stloc.x; ldloc.x

with a comment noting that this makes the resulting compiler IR
more amenable to CSE. Here zz is any IL operation that pushes
a value on the stack. However if zz's result is a ref type, the
dup'ed value produced by zz may be a subtype of the type of the
local x, so this transformation can lose type information.

In the past the loss of type information hasn't mattered, but now if
zz subsequently is used as the this object in a virtual call, the jit
may attempt devirtualization and ask an ill-posed question to the VM,
claiming the type of the this object is an impossible supertype.

This changes modifies the transformation by introducing a new local temp y
that has the type of zz. The result is now

   zz; stloc.y; ldloc.y; ldloc.y; stloc.x

with the net result that a ldloc.y is left on the stack with the proper
type. As an optimization, if zz is ldloc.z then the expansion is simply

   zz (ldloc.z); ldloc.z; stloc.y

and if zz is a simple "cheap" constant (say ldnull) then the constant
expression is duplicated:

   zz (ldnull); ldnull; stloc.y

This resolves the jit side of #9945. Further work is needed on the VM
side since in unverifiable IL there may be other cases where the jit
can present an impossible type.

Diffs from this were minor and in many cases small improvements.

### DIFF
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -9508,7 +9508,6 @@ void Compiler::impImportBlockCode(BasicBlock* block)
     int  prefixFlags = 0;
     bool explicitTailCall, constraintCall, readonlyCall;
 
-    bool     insertLdloc = false; // set by CEE_DUP and cleared by following store
     typeInfo tiRetVal;
 
     unsigned numArgs = info.compArgsCount;
@@ -10077,27 +10076,7 @@ void Compiler::impImportBlockCode(BasicBlock* block)
 
                 if (op1->gtOper == GT_LCL_VAR && lclNum == op1->gtLclVarCommon.gtLclNum)
                 {
-                    if (insertLdloc)
-                    {
-                        // This is a sequence of (ldloc, dup, stloc).  Can simplify
-                        // to (ldloc, stloc).  Goto LDVAR to reconstruct the ldloc node.
-                        CLANG_FORMAT_COMMENT_ANCHOR;
-
-#ifdef DEBUG
-                        if (tiVerificationNeeded)
-                        {
-                            assert(
-                                typeInfo::AreEquivalent(tiRetVal, NormaliseForStack(lvaTable[lclNum].lvVerTypeInfo)));
-                        }
-#endif
-
-                        op1         = nullptr;
-                        insertLdloc = false;
-
-                        impLoadVar(lclNum, opcodeOffs + sz + 1);
-                        break;
-                    }
-                    else if (opts.compDbgCode)
+                    if (opts.compDbgCode)
                     {
                         op1 = gtNewNothingNode();
                         goto SPILL_APPEND;
@@ -10152,26 +10131,6 @@ void Compiler::impImportBlockCode(BasicBlock* block)
                         op1->gtType = TYP_BYREF;
                     }
                     op1 = gtNewAssignNode(op2, op1);
-                }
-
-                /* If insertLdloc is true, then we need to insert a ldloc following the
-                   stloc.  This is done when converting a (dup, stloc) sequence into
-                   a (stloc, ldloc) sequence. */
-
-                if (insertLdloc)
-                {
-                    // From SPILL_APPEND
-                    impAppendTree(op1, (unsigned)CHECK_SPILL_ALL, impCurStmtOffs);
-
-#ifdef DEBUG
-                    // From DONE_APPEND
-                    impNoteLastILoffs();
-#endif
-                    op1         = nullptr;
-                    insertLdloc = false;
-
-                    impLoadVar(lclNum, opcodeOffs + sz + 1, tiRetVal);
-                    break;
                 }
 
                 goto SPILL_APPEND;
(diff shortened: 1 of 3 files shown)


AndyAyersMS:
@JosephTremoulet PTAL
cc @dotnet/jit-contrib 

Jit-diff summary
```
Total bytes of diff: 233 (0.00 % of base)
    diff is a regression.

         201 : Microsoft.CodeAnalysis.CSharp.dasm (0.01 % of base)
          81 : Microsoft.CodeAnalysis.VisualBasic.dasm (0.00 % of base)
          53 : System.Linq.Parallel.dasm (0.01 % of base)
          34 : System.Text.RegularExpressions.dasm (0.04 % of base)
          10 : Microsoft.Win32.Registry.dasm (0.07 % of base)

        -122 : System.Private.CoreLib.dasm (0.00 % of base)
         -27 : System.Linq.Expressions.dasm (0.00 % of base)
         -15 : System.Private.Uri.dasm (-0.02 % of base)
          -6 : System.IO.UnmanagedMemoryStream.dasm (-0.03 % of base)
          -3 : Microsoft.CodeAnalysis.dasm (0.00 % of base)
```

mikedn:
Out of curiosity: where do those improvements come from? It seems that the addition of a new lclvar could make things worse, not better.

AndyAyersMS:
I suspect the RA has an easier time with the lifetime of y rather than overlapping lifetimes of x and zz. A few examples:
```asm
; Assembly listing for method Microsoft.CodeAnalysis.CSharp.CSharpDataFlowAnalysis:AnalyzeReadWrite()
; Before
G_M8470_IG06:
       mov      rcx, qword ptr [(reloc)]
       call     [System.Linq.Enumerable:Empty():ref]
       mov      gword ptr [rbp-60H], rax
       mov      rax, gword ptr [rbp-60H]
       mov      gword ptr [rbp-58H], rax
       mov      rax, gword ptr [rbp-58H]
       mov      gword ptr [rbp-50H], rax
       mov      rax, gword ptr [rbp-50H]
       mov      gword ptr [rbp-48H], rax
       mov      rax, gword ptr [rbp-48H]
       mov      gword ptr [rbp-40H], rax
       mov      rax, gword ptr [rbp-40H]
       mov      gword ptr [rbp-38H], rax
       jmp      G_M8470_IG04

; After
G_M8470_IG06:
       mov      rcx, qword ptr [(reloc)]
       call     [System.Linq.Enumerable:Empty():ref]
       mov      gword ptr [rbp-60H], rax
       mov      gword ptr [rbp-58H], rax
       mov      gword ptr [rbp-50H], rax
       mov      gword ptr [rbp-48H], rax
       mov      gword ptr [rbp-40H], rax
       mov      gword ptr [rbp-38H], rax
       jmp      G_M8470_IG04

; Assembly listing for method Microsoft.CodeAnalysis.CSharp.Syntax.InternalSyntax.Lexer:ScanDirectiveToken(byref):bool:this
; Before
       movzx    rbp, ax
       mov      ecx, ebp
       cmp      ecx, 61

; After
       movzx    rbp, ax
       cmp      ebp, 61
```

omariom:
If the type of zz and local x is the same  the new local is not needed, right?

AndyAyersMS:
@omario You mean still produce

    zz; stloc.x; ldloc.x 

for cases where the type of zz is the same as the type of x? That may be worth looking into. It would be an easy change if you want to give it a go.

However there's another consideration -- with zz we may know the type exactly, but if x can be modified at multiple places in the code we may not know the type of x exactly. Because y is a new temp we know it's not modified elsewhere and so it can inherit exactness from zz (say for instance zz is a newobj or is a local z with one definition that is a newobj). 

To be fair the code above does not propagate exactness from zz to y and arguably it should. A more general approach to propagating types and exactness through assignments is a future work item.

src/jit/importer.cpp
```diff
@@ -11991,48 +11950,24 @@ void Compiler::impImportBlockCode(BasicBlock* block)
...
-                //   This helps eliminate a redundant bounds check in cases such as:
-                //       ariba[i+3] += some_value;
-                // - If the top of the stack is a non-leaf that may be expensive to clone.
+                // If the expression to dup is simple, just clone it.
+                // Otherwise spill it to a temp, and reload the temp
+                // twice.
+                op1 = impPopStack(tiRetVal);
 
-                if (codeAddr < codeEndp)
+                if (!opts.compDbgCode && !op1->IsIntegralConst(0) && !op1->IsFPZero() && !op1->IsLocal())
```
JosephTremoulet: Nit: surprisingly conservative definition of what's a "cheap" int constant here, but I have no idea how frequently we see dups of other small constants.

AndyAyersMS: This was essentially the logic that was there before. 

It might be worth experimenting with something that is more willing to clone, though we're in the importer and so don't really have a good handle on the "cost" of general trees.

PR #9961, opened 2017-03-04
Add scripts for building for arm Android

qmfrederik:
This PR adds the ability to create a rootfs for cross-building for arm Android, as well as the CMake files required to get the build process going.


qmfrederik:
I got this PR to a point where I can cross-compile for ARM Android, so anyone willing to try on arm Android should be able to use this PR.

A couple of notes:
* The build currently compiles for soft-float ARM. Compiling for hard float gave linker warnings.
* I had to set `HAVE_PT_REGS=0` in `tryrun.cmake`; `PT_SETREGS` does not appear to be available on Android. I'm not sure why I did not see this for arm64.
* I also had to explicitly add `include/c++/4.9.x/` and `include/c++/4.9.x/arm-linux-androideabi/` to the system includes. Again, this was not required for arm64.
* `cordb.cpp` uses `_aligned_malloc` and `_aligned_free` which I believe are specific to Visual C++. There is `aligned_alloc` in C++ 17, but targetting C++ 17 caused compiler errors in other areas.

src/debug/di/cordb.cpp
```diff
@@ -499,7 +499,7 @@ DbiGetThreadContext(HANDLE hThread,
     DT_CONTEXT *lpContext)
 {
     // if we aren't local debugging this isn't going to work
-#if !defined(_ARM_) || defined(FEATURE_DBGIPC_TRANSPORT_DI)
+#if !defined(_ARM_) || defined(FEATURE_DBGIPC_TRANSPORT_DI) || defined(__clang__)
```
janvorli: I would prefer using `__ANDROID__` instead of `__clang__` here.

src/debug/di/cordb.cpp
```diff
@@ -538,7 +538,7 @@ BOOL
 DbiSetThreadContext(HANDLE hThread,
     const DT_CONTEXT *lpContext)
 {
-#if !defined(_ARM_) || defined(FEATURE_DBGIPC_TRANSPORT_DI)
+#if !defined(_ARM_) || defined(FEATURE_DBGIPC_TRANSPORT_DI) || defined(__clang__)
```
janvorli: And the same here

qmfrederik:
@janvorli I updated the checks to detect `__ANDROID__` instead of `__clang__`.
(I actually amended the PR first for it to detect `_MSC_VER`, as the code will only compile on Visual C++, but forgot to push that).

So `__ANDROID__` it is, let me know if you need anything else.

Issue #9970, opened 2017-03-06
[x86/Linux] Default P/Invoke Calling Convention?


parjong:
I observed stack misalignment while working on [this patch](https://github.com/parjong/coreclr/commit/565d3c22ad3fb68dd47f9a31e683a28143d58465) (which aims to insert stack alignment check inside ``HelperMethodFrameRestoreState``:
```
Program received signal SIGTRAP, Trace/breakpoint trap.
HelperMethodFrameRestoreState () at /home/parjong/projects/dotnet/coreclr/src/vm/i386/asmhelpers.S:377
377         call        C_FUNC(HelperMethodFrameConfirmState)
(gdb) bt 4
#0  HelperMethodFrameRestoreState () at /home/parjong/projects/dotnet/coreclr/src/vm/i386/asmhelpers.S:377
#1  0xf70e0656 in SafeHandle::DangerousRelease (refThisUNSAFE=0xf2e2bf1c) at /home/parjong/projects/dotnet/coreclr/src/vm/safehandle.cpp:429
#2  0xf4f37ed4 in ?? ()
#3  0xf25df4fc in ?? ()
(More stack frames follow...)
(gdb) info frame 2
Stack frame at 0xffffc3dc:
 eip = 0xf4f37ed4; saved eip = 0xf25df4fc
 called by frame at 0xffffc460, caller of frame at 0xffffc3bc
 Arglist at 0xffffc3d4, args:
 Locals at 0xffffc3d4, Previous frame's sp is 0xffffc3dc
 Saved registers:
  ebp at 0xffffc3d4, eip at 0xffffc3d8
(gdb) info frame 3
Stack frame at 0xffffc460:
 eip = 0xf25df4fc; saved eip = 0xf25df2e6
 called by frame at 0xffffc470, caller of frame at 0xffffc3dc
 Arglist at 0xffffc458, args:
 Locals at 0xffffc458, Previous frame's sp is 0xffffc460
 Saved registers:
  ebp at 0xffffc458, eip at 0xffffc45c
```

While digging this issue, I found calling convention mismatch between P/Invoke IL Stub and its actual target. 

Here is the code fragment that make stack incorrectly aligned:
```
       83EC0C       sub      esp, 12
       50           push     eax
       C745B004000000 mov      dword ptr [ebp-50H], 4
       8BC4         mov      eax, esp
       8945B4       mov      dword ptr [ebp-4CH], eax
       8D0589F45DF2 lea      eax, G_M762_IG05
[...]

PR #9974, opened 2017-03-06
[x86/Linux] Explicitly Set P/Invoke Interop Calling Convention

parjong:
All the native Linux interop functions that System.Native and Symtem.Globalzation uses CDECL (not STDCALL). 

This commit revises each P/Invoke declaration to explicitly annotate its calling convention.

### DIFF
--- a/src/mscorlib/shared/Interop/Unix/Interop.Errors.cs
+++ b/src/mscorlib/shared/Interop/Unix/Interop.Errors.cs
@@ -179,13 +179,13 @@ internal static partial class Interop
             return Marshal.PtrToStringAnsi((IntPtr)message);
         }
 
-        [DllImport(Libraries.SystemNative, EntryPoint = "SystemNative_ConvertErrorPlatformToPal")]
+        [DllImport(Libraries.SystemNative, CallingConvention=CallingConvention.Cdecl, EntryPoint = "SystemNative_ConvertErrorPlatformToPal")]
         internal static extern Error ConvertErrorPlatformToPal(int platformErrno);
 
-        [DllImport(Libraries.SystemNative, EntryPoint = "SystemNative_ConvertErrorPalToPlatform")]
+        [DllImport(Libraries.SystemNative, CallingConvention=CallingConvention.Cdecl, EntryPoint = "SystemNative_ConvertErrorPalToPlatform")]
         internal static extern int ConvertErrorPalToPlatform(Error error);
 
-        [DllImport(Libraries.SystemNative, EntryPoint = "SystemNative_StrErrorR")]
+        [DllImport(Libraries.SystemNative, CallingConvention=CallingConvention.Cdecl, EntryPoint = "SystemNative_StrErrorR")]
         private static unsafe extern byte* StrErrorR(int platformErrno, byte* buffer, int bufferSize);
     }
 }
--- a/src/mscorlib/shared/Interop/Unix/System.Native/Interop.Close.cs
+++ b/src/mscorlib/shared/Interop/Unix/System.Native/Interop.Close.cs
@@ -9,7 +9,7 @@ internal static partial class Interop
 {
     internal static partial class Sys
     {
-        [DllImport(Libraries.SystemNative, EntryPoint = "SystemNative_Close", SetLastError = true)]
+        [DllImport(Libraries.SystemNative, CallingConvention=CallingConvention.Cdecl, EntryPoint = "SystemNative_Close", SetLastError = true)]
         internal static extern int Close(IntPtr fd);
     }
 }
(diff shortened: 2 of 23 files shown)


parjong:
This PR is a step to resolve #9970.

jkotas:
Have you checked what the default calling convention is on Mono Linux x86? If it is cdecl, this should be fixed differently - by changing the default.

parjong:
It is also possible to fix this issue via changing the default P/Invoke calling convention, but I think that it would be better to explicitly annotate the corresponding calling convention as those functions will use ``CDECL`` whatever the default calling convention is.

jkotas:
I do not think it is a good idea to be adding unnecessary annotations. They are cruft that makes it hard to see what actually matters.

parjong:
Okay, I'll attempt to change the default P/Invoke calling convention. Could you let me know which code sets default calling convention? I once tried, but it is hard for me to find it without overview of the whole process.

jkotas:
First, we need to find out the default on Mono. .NET Core should be compatible with Mono on Linux where possible.

parjong:
[Mono document](http://www.mono-project.com/docs/advanced/pinvoke/) says that the default calling convention is ``CDCEL`` for  unix:
 > The default CallingConvention is platform-specific. Under Windows, Winapi is the default, as this is used for most Win32 API functions. (Winapi is equivalent to Stdcall for Windows 9x and Windows NT.) Under Unix platforms, Cdecl is the default.

jkotas:
Good. The place to change it is here: https://github.com/dotnet/coreclr/blob/a9b25d4aa22a1f4ad5f323f6c826e318f5a720fe/src/vm/dllimport.cpp#L3009

parjong:
Thanks!! I'll try to revise that.

jkotas:
You will need to deal with PInvokes that depend on the default in mscorlib today. I think it may be best to best to just leave them as default, and change the C++ implementation to be cdecl instead. It should be possible to flip the C++ implementation to cdecl by just changing the defines - ie change `#define PALAPI` and `#define QCALLTYPE` to be cdecl for Unix.

parjong:
@jkotas I am currently testing with the following patch, but it seems not enough:
```
diff --git a/src/vm/dllimport.cpp b/src/vm/dllimport.cpp
index 30b23ea..f27476d 100644
--- a/src/vm/dllimport.cpp
+++ b/src/vm/dllimport.cpp
@@ -3006,7 +3006,11 @@ void PInvokeStaticSigInfo::InitCallConv(CorPinvokeMap callConv, BOOL bIsVarArg)
         SetError(IDS_EE_NDIRECT_BADNATL_CALLCONV);

     if (callConv == 0 && sigCallConv == 0)
+#ifdef PLATFORM_UNIX
+        m_callConv = pmCallConvCdecl;
+#else // PLATFORM_UNIX
         m_callConv = bIsVarArg ? pmCallConvCdecl : pmCallConvStdcall;
+#endif // !PLATFORM_UNIX
     else if (callConv != 0)
         m_callConv = callConv;
     else
```
P/Invoke stub to ``GlobalizationNative_GetDefaultLocaleName`` still uses stdcall calling convention. 
Is there something that I missed?

jkotas:
Set a breakpoint at this method and stepped through it. I guess you may need to change what pmCallConvWinapi converts to a few lines above.

parjong:
@jkotas Could you let me know what is the difference between ``callConv`` and ``sigCallConv``? For ``GlobalizationNative_GetDefaultLocaleName``, it seems that ``callConv`` is ``pmCallConvStdcall`` and ``sigCallConv`` is ``0``.

jkotas:
sigCallConv is calling convention from the metadata signature. It will be always zero for programs written in C# because of C# does not let you control it.

parjong:
It is a bit strange that ``callConv`` is always ``pmCallConvWinapi``:
```
(gdb) bt 4
#0  PInvokeStaticSigInfo::InitCallConv (this=0xffff78d8, callConv=pmCallConvWinapi, bIsVarArg=0) at /home/parjong/projects/dotnet/coreclr/src/vm/dllimport.cpp:2988
#1  0xf728f108 in PInvokeStaticSigInfo::DllImportInit (this=0xffff78d8, pMD=0xf5965a58, ppLibName=0x0, ppEntryPointName=0x0) at /home/parjong/projects/dotnet/coreclr/src/vm/dllimport.cpp:2931
#2  0xf728f455 in PInvokeStaticSigInfo::PInvokeStaticSigInfo (this=0xffff78d8, pMD=0xf5965a58, throwOnError=PInvokeStaticSigInfo::NO_THROW_ON_ERROR)
    at /home/parjong/projects/dotnet/coreclr/src/vm/dllimport.cpp:2783
#3  0xf6d8317b in CEEInfo::getUnmanagedCallConv (this=0xffffb160, method=0xf5965a58) at /home/parjong/projects/dotnet/coreclr/src/vm/jitinterface.cpp:9431
```
This means that the above changes will not be executed.

parjong:
Does ``pmCallConvWinapi`` mean that calling convention is not specified?

jkotas:
I think so.

PR #9977, opened 2017-03-06
[x86/Linux] CDECL as default P/Invoke Calling Convetion

parjong:
This commit revises P/Invoke without explicit annotation to use CDECL instead of STDCALL.


parjong:
This PR is another approach to resolve #9970 (as discussed in #9974).

parjong:
@janvorli Thank you for comment. I forgot about that.. I updated PR.

PR #9980, opened 2017-03-06
Add handling of multiple CMake installations in probe-win

### LINKED ISSUE #8221: Build cannot detect CMake with both x86 and x64 versions installed
The following line in [probe-win](https://github.com/dotnet/coreclr/blob/master/src/pal/tools/probe-win.ps1#L31) will fail to resolve the CMake path when there are multiple versions installed:

```powershell
$inPathPath = (get-command cmake.exe -ErrorAction SilentlyContinue).Path
```

If we have installed both x86 and x64 versions of cmake, there will be 2 command definitions:

```
CommandType     Name                                                Definition
-----------     ----                                                ----------
Application     cmake.exe                                           C:\Program Files\CMake\bin\cmake.exe
Application     cmake.exe                                           C:\Program Files (x86)\CMake\bin\cmake.exe
```

If there is no strict requirement for the cmake version to use, replacing the line by the following code will fix this issue:

```powershell
  $inPathPath = (get-command cmake.exe -ErrorAction SilentlyContinue)
  
  if ($inPathPath.Length -gt 1) {
	$inPathPath = $inPathPath[0].Path
  } else {
	$inPathPath = $inPathPathPath
  }
```

lggomez:
Check if multiple CMake versions are installed, and use by default the first path available

Fixes #8221

### DIFF
--- a/src/pal/tools/probe-win.ps1
+++ b/src/pal/tools/probe-win.ps1
@@ -29,11 +29,12 @@ function LocateCMake
 {
   $errorMsg = "CMake is a pre-requisite to build this repository but it was not found on the path. Please install CMake from http://www.cmake.org/download/ and ensure it is on your path."
   $inPathPath = (get-command cmake.exe -ErrorAction SilentlyContinue)
-  # Resolve the first version of CMake if multiple commands are found
-  if ($inPathPath.Length -gt 1) {
-    $inPathPath = $inPathPath[0].Path
-  } else {
-    $inPathPath = $inPathPath.Path
+  if ($inPathPath -ne $null) {
+    # Resolve the first version of CMake if multiple commands are found
+    if ($inPathPath.Length -gt 1) {
+      return $inPathPath = $inPathPath[0].Path
+    }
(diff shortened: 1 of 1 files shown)


tannergooding:
You can do the following to get the ProductVersion for an executable:
```ps1
function Get-ProductVersion([string[]] $path) {
  if (!(Test-Path -path $path)) {
    return ""
  }

  $item = Get-Item -path $path
  return $item.VersionInfo.ProductVersion
}
```

We could then select the highest version, rather than the first path with minimal changes.

lggomez:
@tannergooding It's not just a matter of number versions and it shouldn't be since CMake doesn't use numbered folders for that that matter, unless there is a broken install that add an extra command entry.

The original repro relies on having both x86 and x64 versions installed, which gives duplicate commands. Another way is checking which installation dir contains x86 and which not, and choose the x64 version by default (or the corresponding version if there is some build flag available from the script)

danmosemsft:
This seems fine to me.

PR #9982, opened 2017-03-06
Fix incremental build of mscordaccore

janvorli:
This change sets up proper dependency for the mscordaccore build and stops
using PRE_LINK command which is executed always.
The way it solves the problem is that it creates an object library from the
mscordac.cpp and then lets the mscordaccore.exp be generated from that object
and finally use both that object library and the mscordaccore.exp while building
mscordaccore.

### DIFF
--- a/src/dlls/mscordac/CMakeLists.txt
+++ b/src/dlls/mscordac/CMakeLists.txt
@@ -102,7 +102,7 @@ if(WIN32)
 
     # Generate export file
     add_custom_command(
-        DEPENDS ${CURRENT_BINARY_DIR_FOR_CONFIG}/mscordac.def ${MSCORDAC_OBJ_PATH} "$<TARGET_FILE:daccess>"
+        DEPENDS mscordaccore_def mscordacobj daccess
         OUTPUT ${CURRENT_BINARY_DIR_FOR_CONFIG}/mscordaccore.exp
         COMMAND lib.exe /OUT:"${CURRENT_BINARY_DIR_FOR_CONFIG}/mscordaccore.lib" /DEF:"${CURRENT_BINARY_DIR_FOR_CONFIG}/mscordac.def" "$<TARGET_FILE:daccess>" $<$<OR:$<CONFIG:Release>,$<CONFIG:Relwithdebinfo>>:/LTCG> ${STATIC_LIBRARY_FLAGS} ${MSCORDAC_OBJ_PATH}
         COMMENT "Generating mscordaccore.exp export file"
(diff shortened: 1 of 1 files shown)


janvorli:
@BruceForstall can you take a look please?

BruceForstall:
cc @noahfalk

BruceForstall:
It looks good to me, although you are well beyond my level of cmake understanding :)

janvorli:
Looks like I have broken Unix build, I am taking a look.

janvorli:
@BruceForstall there was a problem with the dependencies of custom command being files that are byproducts of a target build (the mscordac.obj). Visual Studio build system wasn't seeing that dependency and sometimes built stuff in a wrong order in the CI. Changing the dependencies to the actual target names were the fix.

noahfalk:
I'm also not well versed in cmake, but the principle sounded fine and if it works properly in testing the build I'm happy to assume its correct : )

BruceForstall:
Nice. I presume you've tested: (1) two builds in a row; second build doesn't build dac (or this part), (2) touch minimal dependency (e.g., something to rebuild coreclr.dll, or daccess.cpp); dac rebuilds.

janvorli:
Sigh. There is still a problem. While the dependencies work fine for the mscordac.obj and the daccess target, they don't work for the mscordac.def for some reason. If I touch mscordac.src, then the mscordaccore.exp is not rebuilt.
Let me look into that.

janvorli:
So, it turns out that there were two issues.
First, it turns out that target type dependencies on a custom target in the add_custom_command determine only the order of the build between the custom target and custom command, but they do not cause the command to be re-executed if the custom target that it depends on is rebuilt. The reason is that a custom target doesn't have any output that cmake could see. So the dependencies in the add_custom_command also need to also include the files generated by the custom target.
I have found this described at the end of the page here:
https://samthursfield.wordpress.com/2015/11/21/cmake-dependencies-between-targets-and-files-and-custom-commands/
Second, Visual Studio cannot figure out that it needs to rebuild the mscordaccore target when the .exp file changes **and** it is generated into the build configuration specific folder. In such case, cmake puts the .exp file as $(Configuration)\mscordaccore.exp into the list of link.exe inputs and the $(Configuration) probably prevents Visual Studio from detecting it as a dependency.
If I changed it so that the mscordaccore.exp is not generated into the build specific folder, then then it works. This is totally ok when building using build.cmd, since we always have completely separate cmake generated bin subfolder for each configuration.
When building from VS by opening the CoreCLR.sln generated by cmake and selecting a build configuration, it causes sharing the mscordaccore.exp between all the build configurations, but while that's not ideal, it is also ok, since this file is not configuration specific.

janvorli:

PR #9985, opened 2017-03-06
Port Windows PGO support to master

dpodder:
The bulk of this PR is a cherry-pick of commit fa02660 that shipped in
release/1.1.0, updating the build system support for PGO to support
consuming PGDs properly during release builds on Windows.

Also included are the following new changes:
* Skip attempting to restore opdata if the requisite project.json is
  missing
* If the optdata package restore fails, fail the build.

Note: This change doesn't by itself enable PGO in master yet, because
training data (optdata packages) for master don't exist on myget yet.
However, with these changes, the only step remaining to enable PGO
optimizations is to add a project.json referencing the correct optdata
package.


dpodder:
@janvorli PTAL
/cc @brianrob @valenis

build.cmd
```diff
@@ -204,6 +204,19 @@ call                                 "%__VSToolsRoot%\VsDevCmd.bat"
 
 REM =========================================================================================
 REM ===
+REM === Restore optimization profile data
+REM ===
+REM =========================================================================================
+
+echo %__MsgPrefix%Restoring the OptimizationData Package
```
janvorli: It would be great to add a command line option to build.cmd that would allow skipping this. In development scenarios, it is nice to be able to cut down the build time to the minimum possible and also enable build in disconnected scenarios.

build.sh
```diff
@@ -215,6 +215,13 @@ build_native()
             fi
         fi
 
+        echo "Restoring the OptimizationData package"
```
janvorli: Like I've said for the build.cmd, it would be great to add a command line option to build.cmd that would allow skipping this.

pgosupport.cmake
```diff
@@ -37,6 +37,18 @@ function(add_pgo TargetName)
     endforeach(ConfigType)
 endfunction(add_pgo)
 
+set(CLR_CMAKE_OPTDATA_PACKAGEID "optimization.PGO.CoreCLR")
+set(CLR_CMAKE_OPTDATA_PACKAGEWITHRID "optimization.${CLR_CMAKE_TARGET_OS}-${CLR_CMAKE_TARGET_ARCH}.PGO.CoreCLR")
+
+# Parse optdata package version from project.json
+file(TO_NATIVE_PATH "${CMAKE_SOURCE_DIR}/extract-from-json.py" ExtractFromJsonScript)
+file(TO_NATIVE_PATH "${CMAKE_SOURCE_DIR}/src/.nuget/optdata/project.json" OptDataProjectJsonPath)
+execute_process(
```
janvorli: This would run the command at the time the cmake generates build files (makefiles on Unix, Visual Studio project files on Windows), not at the build time. So if the /src/.nuget/optdata/project.json changes, it would not rerun the python script unless you clear the binary folder.
It seems that you probably want to do this at build time instead, so that it can regenerate the output when either the python script or the the .json file changes. If that's the case, you'll need the add_custom_command and add_custom_target combo instead of execute_process.

build.cmd
dpodder: I like the idea, although I'm hesitant to add to the usage if I can help it. Would it be acceptable to check for the currently-restored-package and only call restore on it if the version has changed? That doesn't fully address the offline scenario (e.g. when doing `git checkout` to an unbuilt but already-fetched branch), but if the right optdata is missing then your release builds won't be optimized the way they are supposed to be anyway.

I can also make it restore conditionally, only when doing a release build, since we don't consume optdata during debug builds. Would this be useful?

build.sh
dpodder: Will do in the same way as for the CMD script.

pgosupport.cmake
dpodder: The problem is that I need to consume the output of this package at configure time. The python script is there simply to parse project.json to extract the package version, which is used to build up the ProfilePath cmake variable, which in turn is passed to the compiler/linker as a flag that tells PGO where to find the profile data.

I thought about putting the version into a text file instead of a project.json, so that I could parse the version with native cmake, but that would just move complexity into the build.cmd/run.cmd/build.proj suite of tools instead (I would have to construct a dummy project.json so that I had something valid to restore).

Is there a better approach to getting the version out of a project.json at configure time?

build.cmd
janvorli: I am not sure I understand what would be the problem with adding that option. By default, it would be on and if I want to skip that, I'd pass an option to do that. We already have a bunch of options to skip various parts of the build (skipconfigure, skipmscorlib, skipnative, skiptests, skipbuildpackages), so this would be just another option of that kind.
Moreover, for the Unix build, we need to be able to run the build.sh on new platforms where we cannot run any managed code yet. So any attempt to restore something needs to be optional over there.

dpodder: Fair enough--I was trying to keep from adding complexity to the usage (having to know the extra command to skip), but you raise a good point about new platform bring-up. I'll just add a skip argument then, and keep things simple.

pgosupport.cmake
janvorli: I see, I have not realized that the result forms a compiler option. Then it seems that if you ran the script out cmake before cmake is executed and passed the resulting path to cmake in a variable (using the -DVARIABLE_NAME="XXXX" command line option, adding it to the __ExtraCmakeArgs), it would work well. CMake can detect that it was passed a different value of the variable and rebuild the VS / make projects in that case. That would mean that any time the config.json changes, the change would be picked up even without cleaning the bin folder.

build.cmd
brianrob: @dpodder, along the same lines, do we want nopgo to be the default or do we want pgo to be the default?  That would likely affect what change you make here.

dpodder: @brianrob Good question.

As a quick aside: I was thinking of following up with a later PR that refactors this a bit, putting PGO into one of four mutually exclusive states: off, on, instrument, or auto (aka try to PGOize but silently fall back to non-PGO if counts are not available).

I think for master, "auto" makes the most sense--that way developers get the same optimizations that we have in release branches automatically, *unless* they opt out. This is in line with the current behavior (just with no way to force either on|off). Either way, the `skiprestoreoptdata` can (and I think should) remain independent.

pgosupport.cmake
dpodder: Thanks @janvorli! I like this approach much better than the current implementation; I'll go ahead and make the change.

build.cmd
dpodder: @janvorli PS thank you for cluing me in on why the FreeBSD CI job was failing too--it's unable to run managed code, just like you hinted. I'll make sure the `skiprestoreoptdata` is passed for now when building on FreeBSD.

AndyAyersMS:
This should only impact release builds, right?

I suspect (for windows anyways) that release builds w/PGO will be as fast or faster than release builds w/LTCG so there should be no extra cost and maybe some improvement in the build time (discounting the time it takes to restore the package, which hopefully amortizes over many builds for most developers).

AndyAyersMS:
Also I think I asked this privately a while back -- historically profile data is fragile and tied to a specific compiler version. So what happens when a developer uses a compiler version that differs from the one used in the "official build"?

dpodder:
@AndyAyersMS Yes, the build should take roughly the same amount of time with or without PGO. The optdata package itself isn't significant yet, either; if it grows to be a burden on the dev loop, we can always revisit the approach.

I confirmed details about the counts' compatibility with the PGO feature team. It is possible for breaking changes to be introduced, but they're rare these days; VS1015 and VS2017 (verified with 2017 RC) has the same exact count format, so they should be compatible. The build doesn't support any other version of VS anymore, either, so we're currently covered. If we need to simul-support incompatible compilers in the future, it would be relatively easy to generate the data both ways and just a small change to pick the right one to pass to the compiler.

dpodder:
@janvorli Thanks for all of your feedback earlier! I've made changes based on your comments; PTAL

build.sh
```diff
@@ -548,6 +561,7 @@ __SkipRestore=""
 __SkipNuget=0
 __SkipCoreCLR=0
 __SkipMSCorLib=0
+__SkipRestoreOptData=0
```
janvorli: Similar to what we do for `__SkipGenerateVersion=1`, we should also skip restoring opt data if `__isMSBuildOnNETCoreSupported` is 0. And you would not need to modify the `netci.groovy` then.

dpodder: Ah, cool, thanks! Fixing now.

PR #9989, opened 2017-03-06
Harden resolveVirtualMethod against invalid types

### LINKED ISSUE #9945: JIT: bug in devirtualization where type information gets lost

AndyAyersMS:
Split resolveVirtualMethod into a wrapper/helper to allow early
returns as failure reasons are discovered.

Before devirtualizing, ensure that the derived class claimed by the jit
is really a subclass of the class that introduces the virtual method.
Fail early if this is not the case.

Also add early out for interface call devirtualization. The jit currently
does not ask for this but asking for it should not trigger an assert.

Closes #9945 (along with #9959).

### DIFF
--- a/src/vm/jitinterface.cpp
+++ b/src/vm/jitinterface.cpp
@@ -8701,73 +8701,99 @@ void CEEInfo::getMethodVTableOffset (CORINFO_METHOD_HANDLE methodHnd,
 }
 
 /*********************************************************************/
+static CORINFO_METHOD_HANDLE resolveVirtualMethodHelper(MethodDesc* callerMethod,
+                                                        CORINFO_METHOD_HANDLE baseMethod,
+                                                        CORINFO_CLASS_HANDLE derivedClass)
+{
+    STANDARD_VM_CONTRACT;
+
+    MethodDesc* pBaseMD = GetMethod(baseMethod);
+    MethodTable* pBaseMT = pBaseMD->GetMethodTable();
+
+    // Method better be from a fully loaded class
+    _ASSERTE(pBaseMD->IsRestored() && pBaseMT->IsFullyLoaded());
+
+    //@GENERICS: shouldn't be doing this for instantiated methods as they live elsewhere
+    _ASSERTE(!pBaseMD->HasMethodInstantiation());
+
+    // Interface call devirtualization is not yet supported.
+    if (pBaseMT->IsInterface())
+    {
+        return nullptr;
+    }
+
+    // Method better be virtual
+    _ASSERTE(pBaseMD->IsVirtual());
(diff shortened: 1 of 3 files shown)


AndyAyersMS:
@jkotas PTAL
cc @davidwrighton @dotnet/jit-contrib 

No impact on codegen in the jit-diff set.

davidwrighton:
This looks much better. Is it possible to add a test to verify that the old bugs don't regress?

AndyAyersMS:
I added a test with the jit side of the fix (see #9959). I think I can work up an IL level test for the cases that can get past that.

AndyAyersMS:
Added another test case that requires this fix.

Issue #7376, opened 2016-09-27
I can't build CoreCLR against ICU 57.1


pkubaj:
I'm about to send a working FreeBSD port. It builds fine against ICU 55.1, but recent upgrade to 57.1 broke it.
It seems CoreCLR is not compatible with new ICU:
`[  0%] Building C object src/inc/CMakeFiles/corguids.dir/__/pal/prebuilt/idl/fusionpriv_i.c.o
/wrkdirs/usr/ports/devel/coreclr/work/coreclr-1.0.4/src/corefx/System.Globalization.Native/collation.cpp:282:9: error: 'ucol_setVariableTop' is deprecated [-Werror,-Wdeprecated-declarations]
        ucol_setVariableTop(pClonedCollator, ignoreSymbolsVariableTop, 1, pErr);
        ^
/usr/local/include/unicode/ucol.h:1254:1: note: 'ucol_setVariableTop' has been explicitly marked deprecated here
ucol_setVariableTop(UCollator *coll, 
^
`

janvorli:
Seems similar to #5160 - maybe the detection of the ucol_setMaxVariable doesn't work properly with the latest ICU again.

janvorli:
@pkubaj can you please check `CMakeFiles/CMakeError.log` in your build output folder (I guess it is in `bin/obj/FreeBSD.x64.Debug`)? Search for the `ucol_setMaxVariable`. That should show why it is not being detected.

pkubaj:
In `CMakeFiles/CMakeError.log` I have:
```
Building C object CMakeFiles/cmTC_de990.dir/CheckSymbolExists.c.o
/usr/local/bin/clang37   -I/usr/local/include -I/usr/local/opt/icu4c/include  -Wall -std=c11    -o CMakeFiles/cmTC_de990.dir/CheckSymbolExists.c.o   -c /usr/local/poudriere/ports/default/devel/coreclr/work/coreclr-1.0.4/bin/obj/FreeBSD.x64.Debug/CMakeFiles/CMakeTmp/CheckSymbolExists.c
Linking C executable cmTC_de990
/usr/local/bin/cmake -E cmake_link_script CMakeFiles/cmTC_de990.dir/link.txt --verbose=1
/usr/local/bin/clang37   -Wall -std=c11     CMakeFiles/cmTC_de990.dir/CheckSymbolExists.c.o  -o cmTC_de990  
[...]

PR #9990, opened 2017-03-06
Fix ICU detection in System.Globalization.Native build

janvorli:
The config.cmake was not getting the ICU library paths it requires to
build a code that detects presence of the ucol_setMaxVariable.
I have excluded the libraries detection from the CMakeLists.txt
some time ago since they are not used in the build itself anymore,
but I have not realized that they are needed for this detection
script that's included from the CMakeLists.txt.


janvorli:
Fixes #7376

janvorli:

tarekgh:
LGTM

PR #9991, opened 2017-03-06
Change PackagePlatform to Platform for x86 jit pkg

adiaaida:
PackagePlatform was removed in 8dcf9665feabee339029c79b9d86dd24c71905d1,
but the x86 jit package props file was never updated to not use that
property. This change updates the props file to use Platform instead of
PackagePlatform to make sure we package compatjit.dll for x86.

### DIFF
--- a/src/.nuget/Microsoft.NETCore.Jit/runtime.Windows_NT.Microsoft.NETCore.Jit.props
+++ b/src/.nuget/Microsoft.NETCore.Jit/runtime.Windows_NT.Microsoft.NETCore.Jit.props
@@ -2,7 +2,7 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup>
     <NativeBinary Include="$(BinDir)clrjit.dll" />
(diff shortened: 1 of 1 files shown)


adiaaida:
@RussKeldorph @gkhanna79  PTAL

adiaaida:
@chcosta @wtgodbe PTAL

wtgodbe:
LGTM, assuming you've tested package generation locally

adiaaida:
@wtgodbe yep, tested locally, and this produces a package with both clrjit and compatjit.

chcosta:
`PackagePlatforms` was removed in that change, but `PackagePlatform` is still a valid property (https://github.com/dotnet/coreclr/blob/371d6087106916ceb9944b8b11f66887f05afff0/dir.props#L166).  In coreclr, there may be some confusion because in some scenarios PackagePlatform defaults to AnyCPU, https://github.com/dotnet/coreclr/blob/61619268805bace6b730de1237fd3c7545eb99b5/src/.nuget/dir.props#L7

I think that PackagePlatform needs to be aligned to better reflect the platform being packaged for because AnyCPU isn't a fully supported concept in CoreClr.  Changing PackagePlatform to Platform is probably a fine interim fix though.

/cc @ericstj

Issue #7845, opened 2016-10-27
[Linux][GDBJIT] CoreCLR can't find usable version of the ICU libraries 


lucenticus:
CoreCLR crashes with error message`No usable version of the ICU libraries was found` on Ubuntu 14.04 with enabled `FEATURE_GDBJIT`. Crash occurred since commit cf0c28faec35090781799df0a2360cf2440d05ce. 
@janvorli, please take a look.  

```
CORECLR_GDBJIT='test-enums.exe' lldb -o "target create corerun" -o "settings set target.run-args ./test-enums.exe" -o "process launch -s" -o "breakpoint set -m false -f test-enums.cs -l 20" -o "c"
(lldb) target create corerun
Current executable set to 'corerun' (x86_64).
(lldb) settings set target.run-args ./test-enums.exe
(lldb) process launch -s
Process 31581 launched: '/home/epavlov/coreclr-demo/runtime/corerun' (x86_64)
(lldb) breakpoint set -m false -f test-enums.cs -l 20
Breakpoint 1: no locations (pending).
WARNING:  Unable to resolve breakpoint to any actual locations.
(lldb) c
No usable version of the ICU libraries was found
Process 31581 resuming
Process 31581 stopped
* thread #1: tid = 31581, 0x00007ffff6e09c37 libc.so.6`__GI_raise(sig=6) + 55 at raise.c:56, name = 'corerun', stop reason = signal SIGABRT
    frame #0: 0x00007ffff6e09c37 libc.so.6`__GI_raise(sig=6) + 55 at raise.c:56

(lldb) bt
* thread #1: tid = 31581, 0x00007ffff6e09c37 libc.so.6`__GI_raise(sig=6) + 55 at raise.c:56, name = 'corerun', stop reason = signal SIGABRT
  * frame #0: 0x00007ffff6e09c37 libc.so.6`__GI_raise(sig=6) + 55 at raise.c:56
    frame #1: 0x00007ffff6e0d028 libc.so.6`__GI_abort + 328 at abort.c:89
    frame #2: 0x00007ffff05971c8 System.Globalization.Native.so`InitializeICUShim() + 184 at icushim.cpp:171
    frame #3: 0x00007ffff7dea10a ld-linux-x86-64.so.2`??? + 154
    frame #4: 0x00007ffff7dea1f3 ld-linux-x86-64.so.2`??? + 115
    frame #5: 0x00007ffff7deec30 ld-linux-x86-64.so.2`??? + 720
[...]

PR #9993, opened 2017-03-06
Enhance ICU version scan

janvorli:
The scanning that tries to find locally installed ICU version was not
considering the case when only ICU library files with major version
component were present (like libicuuc.so.52). This is a rare case that
I've seen so far only when people built and installed ICU manually.
But it is worth to add checks for such case as a fallback, which is
what this change does.
I have also added checks for minorVersion and subVersion being -1 when
trying to detect the symbol versioning used. While the pre-existing code
would still work fine, it would unnecessarily try nonsensical version.


janvorli:
Fixes rare case reported in #7845
@tarekgh can you take a look please?
CC: @gkhanna79

tarekgh:
LGTM.

PR #9996, opened 2017-03-07
Un-clang-format-horrible-ify emitIns_Call() and genEmitCall()

BruceForstall:
Only reformatting. No code changes.

@dotnet/jit-contrib PTAL

### DIFF
--- a/src/jit/codegen.h
+++ b/src/jit/codegen.h
@@ -416,20 +416,30 @@ protected:
 
     void genPrologPadForReJit();
 
+    // clang-format off
     void genEmitCall(int                   callType,
                      CORINFO_METHOD_HANDLE methHnd,
-                     INDEBUG_LDISASM_COMMA(CORINFO_SIG_INFO* sigInfo) void* addr X86_ARG(ssize_t argSize),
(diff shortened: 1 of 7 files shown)


mikedn:
Can't we actually stop clang-format from unwrapping lines? It makes sense for a formatting tool to wrap lines if they are too long but doing the opposite is rather bizarre.

BruceForstall:
@mikedn there may be knobs we could turn off from the clang-format settings we use. I lost the fight over clang-format long ago, and am only fixing the most egregious awfulness when it impacts me.

mikedn:
So we went from battling inconsistent code formatting piled up over the years to battling dubious code formatting imposed by a robot. Oh well, maybe we should rename clang-format to T-1000.

pgavlin:
FWIW, I think that we could probably avoid some of these issues if we used format settings that were shared by other projects (e.g. the [Google style](https://google.github.io/styleguide/cppguide.html), which is a pack-in with `clang-format`).

LGTM, in any case, though examples like this really make me wonder about our tendency to `MACRO(parameters)`, which IMO is the more fundamental readability problem here.

JosephTremoulet:
> this really make me wonder about our tendency to MACRO(parameters), which IMO is the more fundamental readability problem here

+1

PR #9997, opened 2017-03-07
[x86/Linux] 16-byte aligned UMThunkStubs

parjong:
This commit adds alignment padding and check to UMThunkStubs.


parjong:
@janvorli It is not so urgent, and I forgot to mention you. Thank you for your support.

PR #9998, opened 2017-03-07
[x86/Linux] Correctly adjust stack for double aligned frames

parjong:
The current implementation of lvaAlignFrame assume that ebp is pushed only for ebp frames, but ebp is also pushed for double aligned frames.

This commit revises lvaAlignFrame to count ebp when double aligned frame is used.


parjong:

janvorli:
cc: @dotnet/jit-contrib

janvorli:
@BruceForstall can you please take a look?

BruceForstall:
LGTM

Issue #9976, opened 2017-03-06
[x86/Linux] Enable CLR Release Build


ragmani:
undefined reference error
```
dlopen failed to open the libcoreclr.so with error /var/jenkins/workspace/CLR/x86_Rel_RT_Test/running/overlay/libcoreclr.so: undefined symbol: _Z16DecCantStopCountv
```

```
jang@jang:~/DotNet/enable_x86/work/coreclr/src$ echo "_Z16DecCantStopCountv" | c++filt
DecCantStopCount()
```

ragmani:
cc @parjong

ragmani:
after #10000 , A segmentation fault is occurred when running corerun as shown below.
```
Program received signal SIGSEGV, Segmentation fault.
operator() (this=<optimized out>, pParam=<optimized out>) at /home/jang/DotNet/enable_x86/work/coreclr/src/vm/util.cpp:2389
2389                        maxTrueSize = tempSize;
(gdb) bt
#0  operator() (this=<optimized out>, pParam=<optimized out>) at /home/jang/DotNet/enable_x86/work/coreclr/src/vm/util.cpp:2389
#1  GetLargestOnDieCacheSize (bTrueSize=<optimized out>) at /home/jang/DotNet/enable_x86/work/coreclr/src/vm/util.cpp:2473
#2  0xf797feed in GCToOSInterface::GetLargestOnDieCacheSize (trueSize=<optimized out>) at /home/jang/DotNet/enable_x86/work/coreclr/src/vm/gcenv.os.cpp:312
#3  0xf78dba8c in GetValidGen0MaxSize (seg_size=<optimized out>) at /home/jang/DotNet/enable_x86/work/coreclr/src/gc/gc.cpp:35536
#4  WKS::gc_heap::init_semi_shared () at /home/jang/DotNet/enable_x86/work/coreclr/src/gc/gc.cpp:9921
#5  0xf78db9f3 in WKS::gc_heap::initialize_gc (segment_size=<optimized out>, heap_size=<optimized out>) at /home/jang/DotNet/enable_x86/work/coreclr/src/gc/gc.cpp:9901
#6  0xf78fe77a in WKS::GCHeap::Initialize (this=0x807f670) at /home/jang/DotNet/enable_x86/work/coreclr/src/gc/gc.cpp:33401
[...]

PR #10000, opened 2017-03-07
[x86/linux] Fix "Undefined Reference" error in CLR release build.

ragmani:
replace "IncCantStopCount()" and "DecCantStopCount()" functions with nothing(#define IncCantStopCount()) to call in methods of UTSemReadWrite on x86/Linux.


ragmani:
@parjong  This PR is from #9976

janvorli:
@ragmani I am not sure if we can just remove them - they seem to be needed by the debugger for interop debugging.
@noahfalk can you please take a look?

noahfalk:
lgtm

PR #10002, opened 2017-03-07
[x86/Linux] Call JIT_GetRuntimeType via  JIT_GetRuntimeType_Safe [NO MERGE]

### LINKED ISSUE #10001: [x86/Linux] Unaligned Stack Frame during JIT.Generics.Conversions.Reference.gentogen01 test
I found that ``JIT.Generics.Conversions.Reference.gentogen01`` emits unaligned stack frame while running  while  working on [this patch](https://github.com/parjong/coreclr/commit/565d3c22ad3fb68dd47f9a31e683a28143d58465) (which aims to insert stack alignment check inside ``HelperMethodFrameRestoreState``:
```
Program received signal SIGTRAP, Trace/breakpoint trap.
HelperMethodFrameRestoreState () at /home/parjong/projects/dotnet/coreclr/src/vm/i386/asmhelpers.S:377
377         call        C_FUNC(HelperMethodFrameConfirmState)
(gdb) bt
#0  HelperMethodFrameRestoreState () at /home/parjong/projects/dotnet/coreclr/src/vm/i386/asmhelpers.S:377
#1  0xf6fe3dce in JIT_GetRuntimeType_Framed (type=0xf6337d7c) at /home/parjong/projects/dotnet/coreclr/src/vm/jithelpers.cpp:4269
#2  0xf6fe4026 in JIT_GetRuntimeType (type=0xf6337d7c) at /home/parjong/projects/dotnet/coreclr/src/vm/jithelpers.cpp:4294
#3  0xf25de483 in ?? ()
#4  0xf7187f39 in CallDescrWorkerInternal () at /home/parjong/projects/dotnet/coreclr/src/vm/i386/asmhelpers.S:304
```

parjong:
JIT currently invokes JIT_GetRuntimeType while passing arguments, which emits unaligned stack frames as discussed in #10001.

This commit introduces JIT_GetRuntimeType_Safe to fix #10001.


janvorli:
@parjong I would prefer waiting for response from the jit contrib in the #10001 before attempting to workaround it.

parjong:
@janvorli I agree. I submitted this PR to facilitate discussion.

jkotas:
I do think this will work. You would need to write these wrappers for hundreds of FCalls.

parjong:
@jkotas It seems that JIT treats ``JIT_GetRuntimeType`` as a special case. I have never seen FCall like this before. 

@dotnet/jit-contrib Could please take a look this and #10001?

jkotas:
Why do you think `JIT_GetRuntimeType` is special?

I think you were just unlucky to hit it with this one, but the underlying problem can occur with number of other calls.

parjong:
I observed that ``JIT_GetRuntimeType`` is called while passing arguments to another function (discussed in #10001).  Usual FCalls are treated same as other managed calls, and thus it has its own stack align and clean sequences, but  ``JIT_GetRuntimeType`` does not have these setup/cleanup sequence. 

That is why I think ````JIT_GetRuntimeType`` is special.

jkotas:
> I observed that `JIT_GetRuntimeType` is called while passing arguments to another function

This looks like a bug in the JIT that happens for this JIT helpers by accident, not that there is anything special about this JIT helper. The JIT should call this helper like any other JIT helpers. There is not anything special about it.

parjong:
It is possible that this may be a JIT bug. That is why I would like to discuss this with @dotnet/jit-contrib.

janvorli:
@BruceForstall could you please give us your opinion here?

BruceForstall:
Is this a general problem of nested calls on x86? I believe we can push some arguments to one call, then start pushing arguments to another call, invoke that nested call, then continue pushing arguments and call the outer call. Perhaps your code for alignment in the JIT doesn't consider this case?

The non-x86 platforms "hoist" nested calls. Perhaps it would be possible to hoist calls for x86/linux.

parjong:
@BruceForstall Yes, such nested calls on x86/Linux are problematic as the current implementation assumes that the stack is aligned at the beginning of call sequence.

It would be definitely better to hoist nested calls. Could you let me know where the relevant code is?

parjong:
\CC @seanshpark

seanshpark:
I've encountered this issue while enabling `FEATURE_FIXED_OUT_ARGS` in x86/Linux. Could this fix this issue? or do must enable hoist nested calls?
One thing I'm not sure if it's doing something like this;
```
FuncA(10,FuncB(20,30,40),50);

mov [esp+4], 10 // for FuncA
mov [esp+8], 20 // for FuncB
mov [esp+4], 30 
mov [esp+0], 40
call  FuncB
mov [esp+0], 50
call FuncA
```

jkotas:
> or do must enable hoist nested calls?

I think so.

PR #10010, opened 2017-03-07
Add shared changes from corert

alexperovich:


src/mscorlib/shared/System/Collections/Generic/ICollection.cs
```diff
@@ -26,9 +14,8 @@ namespace System.Collections.Generic
     // Note that T[] : IList<T>, and we want to ensure that if you use
```
jkotas: This comment is not relevant for CoreCLR or CoreRT. Could you please delete it?

src/mscorlib/shared/System/Collections/Generic/IEnumerable.cs
```diff
@@ -27,9 +15,8 @@ namespace System.Collections.Generic
     // Note that T[] : IList<T>, and we want to ensure that if you use
     // IList<YourValueType>, we ensure a YourValueType[] can be used
```
jkotas: Irrelevant comment, could you please delete it?

src/mscorlib/shared/System/Collections/Generic/IList.cs
```diff
@@ -28,9 +16,8 @@ namespace System.Collections.Generic
     // Note that T[] : IList<T>, and we want to ensure that if you use
     // IList<YourValueType>, we ensure a YourValueType[] can be used 
     // without jitting.  Hence the TypeDependencyAttribute on SZArrayHelper.
```
jkotas: Same here.

src/mscorlib/shared/System/Lazy.cs
```diff
@@ -183,6 +183,7 @@ internal static LazyThreadSafetyMode GetModeFromIsThreadSafe(bool isThreadSafe)
     /// </para>
     /// </remarks>
     [Serializable]
+    [ComVisible(false)]
```
jkotas: Undo this. It is a recent change made by somebody else - this should be deleted in the CoreRT copy instead.

(It is an example of change that the mirror would take care of propagating if we had one.)

alexperovich: Yep, I am in the process of getting that up and running.

Issue #9969, opened 2017-03-05
Debug.Assert from System.Reflection.Metadata while using clrstack


stephentoub:
I'm using lldb-3.6 on Ubuntu 16.10 to debug an issue, and I just hit an assert from clrstack and its usage of System.Reflection.Metadata:
```
(lldb) clrstack


   at System.Environment.get_StackTrace()
   at System.Diagnostics.Debug.Assert(Boolean condition, String message, String detailMessage)
   at System.Diagnostics.Debug.Assert(Boolean condition)
   at System.Reflection.Metadata.MethodDebugInformation..ctor(MetadataReader reader, MethodDebugInformationHandle handle) in /home/stoub/source/corefx/src/System.Reflection.Metadata/src/System/Reflection/Metadata/PortablePdb/MethodDebugInformation.cs:line 25
   at System.Reflection.Metadata.MetadataReader.GetMethodDebugInformation(MethodDebugInformationHandle handle) in /home/stoub/source/corefx/src/System.Reflection.Metadata/src/System/Reflection/Metadata/MetadataReader.cs:line 1369
   at SOS.SymbolReader.GetSourceLineByILOffset(IntPtr symbolReaderHandle, Int32 methodToken, Int64 ilOffset, Int32& lineNumber, String& fileName)
   at SOS.SymbolReader.GetLineByILOffset(IntPtr symbolReaderHandle, Int32 methodToken, Int64 ilOffset, Int32& lineNumber, IntPtr& fileName)OS Thread Id: 0xc649 (10)
        Child SP               IP Call Site
00007F1E951D5630 00007f1f38d1c6a3 [InlinedCallFrame: 00007f1e951d5630] Interop+Sys.WaitForSocketEvents(IntPtr, SocketEvent*, Int32*)
00007F1E951D5630 00007f1ebefd4aeb [InlinedCallFrame: 00007f1e951d5630] Interop+Sys.WaitForSocketEvents(IntPtr, SocketEvent*, Int32*)
00007F1E951D5620 00007F1EBEFD4AEB DomainBoundILStubClass.IL_STUB_PInvoke(IntPtr, SocketEvent*, Int32*)
00007F1E951D5700 00007F1EBEFD4765 System.Net.Sockets.SocketAsyncEngine.EventLoop() [/home/stoub/source/corefx/src/System.Net.Sockets/src/System/Net/Sockets/SocketAsyncEngine.Unix.cs @ 266]
[...]

PR #10011, opened 2017-03-07
Fix assert in portable pdb reader running the clrstack command

mikem8361:
Issue #9969


mikem8361:

PR #10014, opened 2017-03-08
Revise Interop TC calling convention

parjong:
Recently merged #9977 changes the default P/Invoke calling convention from STDCALL and CDECL, but several Interop testcases assume that the default is STDCALL.

This commit revises Interop TC's calling convention.


parjong:
@janvorli Please take a look.

Issue #10003, opened 2017-03-07
Long CORECLR_GDBJIT env variable 


chunseoklee:
Current gdbjit feature seems not to support very long(>= 1024) CORECLR_GDBJIT variable(https://github.com/dotnet/coreclr/blob/master/src/vm/gdbjit.cpp#L1742). 
Is there any reason to check max length of this env variable ?

chunseoklee:
CC @kvochko @Dmitri-Botcharnikov

Dmitri-Botcharnikov:
@chunseoklee It was assumed that this variable holds the path of assembly to debug. However currently it holds list of assembly file names, so this check can be eliminated.

chunseoklee:
@Dmitri-Botcharnikov Thank you.

PR #10016, opened 2017-03-08
Remove CORECLR_GDBJIT length check in gdbjit

chunseoklee:
issue: https://github.com/dotnet/coreclr/issues/10003

This length check is obsolete since it is based on wrong assumption that this variable holds the path of assembly to debug

### DIFF
--- a/src/vm/gdbjit.cpp
+++ b/src/vm/gdbjit.cpp
@@ -1739,7 +1739,7 @@ void NotifyGdb::MethodCompiled(MethodDesc* MethodDescPtr)
     {
         cCharsNeeded = GetEnvironmentVariableW(W("CORECLR_GDBJIT"), NULL, 0);
 
-        if((cCharsNeeded == 0) || (cCharsNeeded >= MAX_LONGPATH))
+        if(cCharsNeeded == 0)
             return;
         wszModuleNames = new WCHAR[cCharsNeeded+1];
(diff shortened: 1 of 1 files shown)


src/vm/gdbjit.cpp
```diff
@@ -1739,7 +1739,7 @@ void NotifyGdb::MethodCompiled(MethodDesc* MethodDescPtr)
     {
         cCharsNeeded = GetEnvironmentVariableW(W("CORECLR_GDBJIT"), NULL, 0);
 
-        if((cCharsNeeded == 0) || (cCharsNeeded >= MAX_LONGPATH))
+        if(cCharsNeeded == 0)
             return;
         wszModuleNames = new WCHAR[cCharsNeeded+1];
         cCharsNeeded = GetEnvironmentVariableW(W("CORECLR_GDBJIT"), wszModuleNames, cCharsNeeded);
```
janvorli: I've just noticed we are missing a nullptr check here.

chunseoklee: @janvorli  could you clarify which variable you point to ?

chunseoklee: @janvorli  could you clarify which variable you point to ?

janvorli: @chunseoklee I meant the wszModuleNames. But I have not noticed before that this new is not nothrow, so the nullptr check is not missing. I can see that while most of the allocations in this file use `new (nothrow)`, about 14 of them don't. We should unify that.
So this change is ok and I'll create a new issue to fix the allocations.

PR #10023, opened 2017-03-08
Update CLR ABI PSPSym Section

parjong:
This commit updates CLR ABI PSPSym Section according to recent changes on x86/Linux.

### DIFF
--- a/Documentation/botr/clr-abi.md
+++ b/Documentation/botr/clr-abi.md
@@ -273,7 +273,11 @@ Note that JIT64 does not implement this properly. The C# compiler used to always
 
 ## The PSPSym and funclet parameters
 
-The name *PSPSym* stands for Previous Stack Pointer Symbol. It is how a funclet accesses locals from the main function body. This is not used for x86: the frame pointer on x86 is always preserved when the handlers are invoked.
+The *PSPSym* (which stands for Previous Stack Pointer Symbol) is a pointer-sized local variable used to access locals from the main function body.
+
+CoreRT does not use PSPSym. For filter funclets the VM sets the frame register to be the same as the parent function. For second pass funclets the VM restores all non-volatile registers. The same convention is used across all platforms.
+
+CoreCLR uses PSPSym for all platforms except x86: the frame pointer on x86 is always preserved when the handlers are invoked.
 
 First, two definitions.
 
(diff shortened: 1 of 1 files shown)


Documentation/botr/clr-abi.md
```diff
@@ -273,15 +273,19 @@ Note that JIT64 does not implement this properly. The C# compiler used to always
...
+The *PSPSym* (which stands for Previous Stack Pointer Symbol) is a pointer-sized local variable used to access locals from the main function body.
+
+CoreRT does not use PSPSym. For filter funclets the VM sets the frame register to be the same as the parent function. For second pass funclets the VM restores all non-volatile registers. The same convention is used across all platforms.
+
+CoreCLR uses PSPSym for all platforms except x86: the frame pointer on x86 is always preserved when the handlers are invoked.
 
 First, two definitions.
 
 *Caller-SP* is the value of the stack pointer in a function's caller before the call instruction is executed. That is, when function A calls function B, Caller-SP for B is the value of the stack pointer immediately before the call instruction in A (calling B) was executed. Note that this definition holds for both AMD64, which pushes the return value when a call instruction is executed, and for ARM, which doesn't. For AMD64, Caller-SP is the address above the call return address.
```
parjong: > For AMD64, Caller-SP is the address above the call return address.

Is it correct? I found this line while working on this commit.

parjong:

Documentation/botr/clr-abi.md
janvorli: @parjong why doesn't it sound right to you?

parjong: For me, this sentence seems to say that ``Caller-SP`` on AMD64 is ``CFA - 8``. Is that true?

janvorli: No, CFA is the SP before the call is made, so caller SP is CFA.

parjong: Does ``the address above the call return address`` mean that ``Caller-SP`` is equal to the address where ``rip`` saved + 8? If so, it seems correct.

janvorli: Yes, it means that. What was your former understanding?

parjong: I thought that it means ``CFA(call return address) -(above) 8``

parjong:

BruceForstall:
LGTM.

BTW, please tag me for any CLR ABI doc changes.

BruceForstall:
@parjong It seems like the main change here is:
```
CoreCLR uses PSPSym for all platforms except x86: the frame pointer on x86 is always preserved when the handlers are invoked.
```
So the first part is true:
```
CoreCLR uses PSPSym for all platforms except x86
```
but the 2nd part I'm not sure of, or is confusing:
```
the frame pointer on x86 is always preserved when the handlers are invoked.
```
Is this true for both Windows/x86 and Linux/x86? Should we add some text to clarify the behavior of Linux/x86 compared to Windows/x86?

parjong:
@BruceForstall I do not know how x86/Windows works. Is it similar to Core RT? The approach that x86/Linux currently seems to be similar to that of Core RT (VM restores frame pointer for filter, and all the non-volatile resisters for the second pass funclets).

BruceForstall:
@parjong I'm not positive myself. Since this is in the PSPSym section, which does not apply to Windows/x86, I'll go ahead and merge the change.

PR #10024, opened 2017-03-08
Fix wrong GCC inline assembly code.

ragmani:
It was coded as "rbx" instead of "ebx" register in clobber list position.


ragmani:
@parjong This PR is associated with #9976.

adityamandaleeka:
LGTM.

parjong:

Issue #10012, opened 2017-03-07
[x86/Linux] Enable FEATURE_FIXED_OUT_ARGS


seanshpark:
I'd like to add this issue for `FEATURE_FIXED_OUT_ARGS` enabling on x86 Linux.
Currently there exist several stack alignment issues as ESP is changed while making a call.

With discussion with @parjong, we can do as these steps;
1. Use `FEATURE_FIXED_OUT_ARGS` directive to split or insert to current code
2. Turn on `FEATURE_FIXED_OUT_ARGS` locally for development and testing
3. Send PRs one by one and maybe add a reference to this issue.
4. If required, introduce a new definition in cmake or other place to sync GC/VM/JIT
5. When all things goes ok, turn on this directive
6. Optionally, remove(clean) stack alignment codes, current code with `UNIX_X86_ABI`, as it may have failures in unit test and may confuse developers.

janvorli:
@seanshpark thank you for sharing your plan.
As for #3, I would prefer to have one PR that would contain all the logical changes, each change being a separate commit. It would be very difficult to keep track of all the changes that were made if they were separate PRs and it would make my reasoning about them during code reviewing very difficult.
As for #4, I am not sure what that means, can you please give me some details?

parjong:
@janvorli As this is a kind of ABI (contract between VM and JIT), we need to modify both VM and JIT all at once. 

For example, unwinder need to be changed according to ``FEATURE_FIXED_OUT_ARGS``, but ``FEATURE_FIXED_OUT_ARGS`` is defined only in JIT, and thus we currently have to use some indirect approach (such as using _TARGET_X86_).

Item 4 is a plan to introduce a global macro that affects both VM and JIT.

BruceForstall:
cc @BruceForstall @dotnet/jit-contrib

parjong:
\CC @parjong

BruceForstall:
@parjong @seanshpark Can you give me some more context as to *why* you want to enable this for x86?

I can see a couple problems:
1. I don't know if the x86 GC info can handle the outgoing arg space. I think it is built around doing "push" to prepare outgoing arguments. But this requires more investigation.
2. The outgoing arg space will have to be filled by "mov [esp+x], reg" instructions, which are bigger than "push" instructions, so code overall will be bigger. Also, IIRC, you can "push [mem]" but you can't "mov [esp+x], [mem]". Not sure if we generate that pattern, though.

parjong:
[...]

PR #10026, opened 2017-03-08
[x86/Linux] Enforce EBP-frame

### LINKED ISSUE #10025: [x86/Linux] JIT.Regression.CLR-x86-JIT.V1-M12-Beta2.b52840 test failure
``JIT.Regression.CLR-x86-JIT.V1-M12-Beta2.b52840`` test failed as follows:
```
FAILED   - [ 4746 :    4 sec ] JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b52840/b52840/b52840.sh
               BEGIN EXECUTION
               /root/dotnet-test/coreclr_unittest_20170228_101916_x86_x86-emu/coreclr-unittest-170228-011944/overlay/corerun b52840.exe

               Unhandled Exception: System.InvalidCastException: Specified cast is not valid.
                  at Test.App.Main1()
               ./b52840.sh: line 131:  7484 Aborted                 (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" b52840.exe $CLRTestExecutionArguments
               Expected: 100
               Actual: 134
               END EXECUTION - FAILED
```

parjong:
The current implementation of x86/Linux unwinder sometimes failed to restore SP correctly, which results in  #10025.

As a workaround, this commit enforces JIT to emit EBP-frame for all methods on x86/Linux to fix #10025.

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -16270,6 +16270,11 @@ void Compiler::fgSetOptions()
         codeGen->setFramePointerRequiredGCInfo(true);
     }
 
+#ifdef UNIX_X86_ABI
+    // TODO Remove this line once SP restored by unwinder becomes reliable.
+    codeGen->setFramePointerRequired(true);
+#endif // UNIX_X86_ABI
+
     // printf("method will %s be fully interruptible\n", genInterruptible ? "   " : "not");
 }
 


parjong:
@dotnet-bot test OSX x64 Checked Build and Test please.

janvorli:
@parjong can you please investigate and try to fix the issue before adding the workaround? Workarounds often tend to stay in for much longer time than expected since people get distracted by other work and the pressure to fix something that works, even though it is a workaround, becomes very low.

parjong:
@janvorli This issue comes from #9272. Here is the JIT disasm that CLR currently failed to unwind:
```
G_M961_IG01:
       83EC0C       sub      esp, 12

G_M961_IG02:
       83EC08       sub      esp, 8
       684C930CF5   push     0xF50C934C
       6852C1A3F5   push     0xF5A3C152
       B996D403F5   mov      ecx, 0xF503D496
       E816CBA104   call     CORINFO_HELP_GETREFANY
       83C408       add      esp, 8
       8B08         mov      ecx, gword ptr [eax]
       83EC0C       sub      esp, 12
       68383207F5   push     0xF5073238
       E86298A104   call     CORINFO_HELP_LDELEMA_REF
       83C40C       add      esp, 12

G_M961_IG03:
       83C40C       add      esp, 12
       C3           ret
```
If an exception is thrown inside JIT Helpers, ``Esp`` is restored by ``libunwind``, and thus both stack elements and alignment padding are not considered.

We could resolve #9272 via #10012, but it may take some time.

janvorli:
@parjong which of the helpers have thrown an exception?

parjong:
An exception is thrown inside ``JIT_GetRefAny``:
```
(gdb) bt
#0  DispatchManagedException (ex=..., isHardwareException=false) at /home/parjong/projects/dotnet/coreclr/src/vm/exceptionhandling.cpp:4662
#1  0xf6f677ea in __FCThrow (__me=0x0, reKind=kInvalidCastException, resID=0, arg1=0x0, arg2=0x0, arg3=0x0)
    at /home/parjong/projects/dotnet/coreclr/src/vm/fcall.cpp:57
#2  0xf6fb506f in JIT_GetRefAny (type=0xf591c496, typedByRef=...) at /home/parjong/projects/dotnet/coreclr/src/vm/jithelpers.cpp:3624
#3  0xf25a349a in ?? ()
```

janvorli:
@parjong the fact that we restore a wrong esp (which your hotfix doesn't change, just hides, right?) scares me. I understand that the next unwind will fix it thanks to the ebp frame unwinding, but what if the try / catch is right in the caller of the JIT_GetRefAny? Would it still work?
I would rather leave the test failing until we fix it properly with #10012.

parjong:
@janvoli As I understand, exception will be caught properly, but will cause segfault when it returns. Actually, there is another workaround on the above issue (https://github.com/parjong/coreclr/tree/fix/x86_converative_epilog), but you're right.

I'll close this PR as we could use this workaround internally.

PR #10036, opened 2017-03-08
Move System.Security attributes to shared partition.

AtsushiKan:
Since we need to complete this set in CoreRt anyway,
this is a preparation commit so we don't import
all of CoreClr's formatting badness.


AtsushiKan:
@dotnet-bot test this please

AtsushiKan:
cc @jkotas 
cc @alexperovich

PR #10045, opened 2017-03-08
Fix incremental build of resources injected to coreclr.dll

janvorli:
This change fixes the build of resources injected to coreclr.dll so that it
happens only when its dependencies change and not at every build.


janvorli:
@BruceForstall can you take a look please? This was the last incremental build issue on Windows, now the build is fully incremental.
I have tested this change with no changes, change of daccess.cpp, change in one of coreclr sources, change in one of mscordbi sources and change in one of mscordaccore sources. With no changes, nothing is rebuilt, with changes in one of those dependencies, the resources are properly rebuilt and injected.

BruceForstall:
LGTM

I had implemented something similar, but was blocked on handling the more complicated DAC thing, which you've now fixed.

I'm a little worried that "touch" is not a Windows built-in command, and it doesn't look like it's currently used, so possibly some people won't have a "touch" on their PATH. If the CI does, then I suppose that's good enough.

janvorli:
@BruceForstall don't worry about the touch. This is a built-in cmake command. The cmake -E supports various built in commands that cmake implements internaly as platform independent.
See here for list of all commands it supports:
https://cmake.org/cmake/help/v3.8/manual/cmake.1.html#command-line-tool-mode

BruceForstall:
Perfect. I didn't see that the last time I looked.

janvorli:

janvorli:

janvorli:

Issue #9935, opened 2017-03-03
[x86/Linux] Unit test NativeCallableTest fails with SIGSEGV


lucenticus:
Unit test `Interop/NativeCallable/NativeCallableTest/NativeCallableTest.exe` fails with segmentation fault.
It seems that we have some issue with calling convention processing at UMThunkStub, because if i remove `WINAPI` from `CallManagedAdd`  ([NativeCallableDll.cpp:9](https://github.com/dotnet/coreclr/blob/master/tests/src/Interop/NativeCallable/NativeCallableDll.cpp#L9)), then it works fine. It seems we should fix calling stdcall managed function from unmanaged stdcall function in UMThunkStub.
We use `m_cbRetPop` instead of ret in UMThunkStub ([umthunkstub.S:118](https://github.com/dotnet/coreclr/blob/master/src/vm/i386/umthunkstub.S#L118)), but in this test `m_cbRetPop` is equal to zero (argument passed through `ecx`)

gkhanna79:
@hseok-oh Do you know who can help look into this?

hseok-oh:
cc/ @parjong @seanshpark @wateret @jkotas

seanshpark:
Could this be related with https://github.com/dotnet/coreclr/issues/9357#issuecomment-278026518  from @janvorli ?

lucenticus:
>  So, the difference between the m_cbRetPop and the m_cbActualArgSize represents the size of the arguments passed in registers. 

@seanshpark It seems that the behavior described above is not working as expected on x86/Linux, because in this test case we passed one argument (`n`) through ecx register, but  `m_cbActualArgSize` is equal to zero. 

It happens because  we expected cdecl calling convention for unmanaged part, and we have code below for  
`m_cbActualArgSize` calculation:
```
   m_cbActualArgSize =
        (pStubMD != NULL) ? pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize() : pMD->SizeOfArgStack();
```

janvorli:
@seanshpark, @lucenticus I have looked through the sources again and found I was wrong. It is really the case that the m_cbActualArgSize is the size of the stack arguments only. I am sorry for misleading you.

lucenticus:
@janvorli Thanks for clarifying, do you know a way to check whether we have arguments in registers too, or how to check calling convention of caller? If I right understand, at this moment we expected that caller always has `CDECL` calling convention.

janvorli:
[...]

PR #10060, opened 2017-03-09
[x86/Linux] Fix NativeCallableTest

lucenticus:
This PR fixes issue #9935
@janvorli @parjong PTAL

### DIFF
--- a/src/vm/dllimportcallback.cpp
+++ b/src/vm/dllimportcallback.cpp
@@ -1416,6 +1416,7 @@ VOID UMThunkMarshInfo::RunTimeInit()
     {
         // caller pop
         m_cbRetPop = 0;
+        m_cbActualArgSize += numRegistersUsed * STACK_ELEM_SIZE;
     }
     else
     {
--- a/src/vm/dllimportcallback.cpp
+++ b/src/vm/dllimportcallback.cpp
@@ -1375,11 +1375,6 @@ VOID UMThunkMarshInfo::RunTimeInit()
 
         }
     }
-    //
-    // m_cbActualArgSize gets the number of arg bytes for the NATIVE signature
-    //
-    m_cbActualArgSize =
-        (pStubMD != NULL) ? pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize() : pMD->SizeOfArgStack();
 
 #if defined(_TARGET_X86_)
     MetaSig sig(pMD);
@@ -1412,8 +1407,9 @@ VOID UMThunkMarshInfo::RunTimeInit()
         new (&sigInfo) PInvokeStaticSigInfo(pMD);
     else
         new (&sigInfo) PInvokeStaticSigInfo(GetSignature(), GetModule());
-    if (pStubMD == NULL)
-        m_cbActualArgSize += numRegistersUsed * STACK_ELEM_SIZE;
+
+    m_cbActualArgSize = (pStubMD != NULL) ? pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize()
+                                          : (pMD->SizeOfArgStack() + numRegistersUsed * STACK_ELEM_SIZE);
     if (sigInfo.GetCallConv() == pmCallConvCdecl)
     {
         // caller pop
@@ -1424,8 +1420,15 @@ VOID UMThunkMarshInfo::RunTimeInit()
         // callee pop
         m_cbRetPop = static_cast<UINT16>(m_cbActualArgSize);
     }
+#else // _TARGET_X86_
+    //
+    // m_cbActualArgSize gets the number of arg bytes for the NATIVE signature
+    //
+    m_cbActualArgSize =
+        (pStubMD != NULL) ? pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize() : pMD->SizeOfArgStack();
 
 #endif // _TARGET_X86_
+
 
 #endif // _TARGET_X86_ && !FEATURE_STUBS_AS_IL
 
--- a/src/vm/dllimportcallback.cpp
+++ b/src/vm/dllimportcallback.cpp
@@ -1416,13 +1416,14 @@ VOID UMThunkMarshInfo::RunTimeInit()
     {
         // caller pop
         m_cbRetPop = 0;
-        m_cbActualArgSize += numRegistersUsed * STACK_ELEM_SIZE;
     }
     else
     {
         // callee pop
         m_cbRetPop = static_cast<UINT16>(m_cbActualArgSize);
     }
+    m_cbActualArgSize += numRegistersUsed * STACK_ELEM_SIZE;
+
 #endif // _TARGET_X86_
 
 #endif // _TARGET_X86_ && !FEATURE_STUBS_AS_IL
--- a/src/vm/dllimportcallback.cpp
+++ b/src/vm/dllimportcallback.cpp
@@ -1412,6 +1412,8 @@ VOID UMThunkMarshInfo::RunTimeInit()
         new (&sigInfo) PInvokeStaticSigInfo(pMD);
     else
         new (&sigInfo) PInvokeStaticSigInfo(GetSignature(), GetModule());
+    if (pStubMD == NULL)
+        m_cbActualArgSize += numRegistersUsed * STACK_ELEM_SIZE;
     if (sigInfo.GetCallConv() == pmCallConvCdecl)
     {
         // caller pop
@@ -1422,7 +1424,6 @@ VOID UMThunkMarshInfo::RunTimeInit()
         // callee pop
         m_cbRetPop = static_cast<UINT16>(m_cbActualArgSize);
     }
-    m_cbActualArgSize += numRegistersUsed * STACK_ELEM_SIZE;
 
 #endif // _TARGET_X86_
 
--- a/src/vm/dllimportcallback.cpp
+++ b/src/vm/dllimportcallback.cpp
@@ -1378,7 +1378,6 @@ VOID UMThunkMarshInfo::RunTimeInit()
 
 #if defined(_TARGET_X86_)
     MetaSig sig(pMD);
-    ArgIterator argit(&sig);
     int numRegistersUsed = 0;
     m_ecxArgOffset = -1;
     m_edxArgOffset = -1;
@@ -1402,14 +1401,14 @@ VOID UMThunkMarshInfo::RunTimeInit()
             offs += StackElemSize(cbSize);
         }
     }
+    m_cbActualArgSize = (pStubMD != NULL) ? pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize() : offs;
+
     PInvokeStaticSigInfo sigInfo;
     if (pMD != NULL)
         new (&sigInfo) PInvokeStaticSigInfo(pMD);
     else
         new (&sigInfo) PInvokeStaticSigInfo(GetSignature(), GetModule());
 
-    m_cbActualArgSize = (pStubMD != NULL) ? pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize()
-                                          : (pMD->SizeOfArgStack() + numRegistersUsed * STACK_ELEM_SIZE);
     if (sigInfo.GetCallConv() == pmCallConvCdecl)
     {
         // caller pop
@@ -1428,7 +1427,6 @@ VOID UMThunkMarshInfo::RunTimeInit()
         (pStubMD != NULL) ? pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize() : pMD->SizeOfArgStack();
 
 #endif // _TARGET_X86_
-
 
 #endif // _TARGET_X86_ && !FEATURE_STUBS_AS_IL
 
--- a/src/vm/dllimportcallback.h
+++ b/src/vm/dllimportcallback.h
@@ -210,6 +210,7 @@ private:
                                             // On x86, NULL for no-marshal signatures
                                             // On non-x86, the managed entrypoint for no-delegate no-marshal signatures
     UINT32            m_cbActualArgSize;    // caches m_pSig.SizeOfFrameArgumentArray()
+                                            // On x86/Linux we have to augment with numRegistersUsed * STACK_ELEM_SIZE
 #if defined(_TARGET_X86_)
     UINT16            m_cbRetPop;           // stack bytes popped by callee (for UpdateRegDisplay)
 #if defined(FEATURE_STUBS_AS_IL)


src/vm/dllimportcallback.cpp
```diff
@@ -1416,6 +1416,7 @@ VOID UMThunkMarshInfo::RunTimeInit()
     {
         // caller pop
         m_cbRetPop = 0;
+        m_cbActualArgSize += numRegistersUsed * STACK_ELEM_SIZE;
```
janvorli: The should be updated for both calling conventions - the args stack size is not different between them, the only difference is whether we pop them or not in the UMThunkStub.
So this new line should be moved before the `if`.

lucenticus: @janvorli I moved this line after if statement, because we still use `m_cbRetPop`   as result of 
```
m_cbActualArgSize =
        (pStubMD != NULL) ? pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize() : pMD->SizeOfArgStack();

```
If I move this line before `if` statement, then some interop tests fail.

janvorli: Hmm, that's really strange. The m_cbRetPop should be the same as the m_cbActualArgSize, since the native caller calling with stdcall convention expects the function to pop all of the arguments. And since the caller passes everything on stack, the two values should match. So it looks like there is some other problem that we've uncovered. Can you please tell me which tests were failing?

lucenticus: Here is a list of failed tests
```
Interop/BestFitMapping/BestFitMapping/BestFitMapping.sh
Interop/RefInt/RefIntTest/RefIntTest.sh
Interop/RefCharArray/RefCharArrayTest/RefCharArrayTest.sh
Interop/StringMarshalling/LPSTR/LPSTRTest/LPSTRTest.sh
Interop/StringMarshalling/LPSTR/LPSTRTest/LPSTRTest.sh
```

janvorli: The RefIntTest tests look like a simple enough to reason about. I assume the failing test was the DoCallBack_MarshalRefInt_Stdcall, right? Could you please look at what is the m_cbActualArgSize and numRegistersUsed before the if / else when this test is executed?

lucenticus: Yes, the test was failed on DoCallBack_MarshalRefInt_Stdcall. Here is a gdb output for this code:
```
(gdb) n
1417	    if (sigInfo.GetCallConv() == pmCallConvCdecl)
=> 0xf6eba3e3 <UMThunkMarshInfo::RunTimeInit()+1779>:	89 04 24	mov    %eax,(%esp)
   0xf6eba3e6 <UMThunkMarshInfo::RunTimeInit()+1782>:	8b 9d e8 fe ff ff	mov    -0x118(%ebp),%ebx
   0xf6eba3ec <UMThunkMarshInfo::RunTimeInit()+1788>:	e8 ff a9 e4 ff	call   0xf6d04df0 <PInvokeStaticSigInfo::GetCallConv()>
   0xf6eba3f1 <UMThunkMarshInfo::RunTimeInit()+1793>:	3d 00 02 00 00	cmp    $0x200,%eax
   0xf6eba3f6 <UMThunkMarshInfo::RunTimeInit()+1798>:	0f 85 11 00 00 00	jne    0xf6eba40d <UMThunkMarshInfo::RunTimeInit()+1821>
(gdb) p sigInfo.GetCallConv()
$9 = pmCallConvStdcall
(gdb) p numRegistersUsed 
$10 = 1
(gdb) p m_cbActualArgSize 
$11 = 8
(gdb) p pMD->SizeOfArgStack()
$12 = 0
(gdb) p pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize()
$13 = 4

```

janvorli: @lucenticus looking at the code and thinking about it more, it seems the problem is caused by the fact that when we get the MetaSig for iteration over the arguments, we use the pMD. But that's right only if we are not generating any stub, which seems to be the case of the NativeCallable attribute.
However, in other cases, I think we should create the MetaSig from the pStubMD instead. I may be wrong, but I think we can just use `MetaSig sig((pStubMD != NULL )? pStubMD : pMD))`.

janvorli: Or maybe it should be like in the case of !FEATURE_STUBS_AS_IL. That code builds the MetaSig from pStubMD when it is not NULL, but this way if it is NULL:
MetaSig msig(GetSignature(), GetModule(), NULL);

lucenticus: I've tried both ways, but I get the same results. Here is a print of pStubMD and pMD:
```
(gdb) p *pStubMD
$1 = {m_pszDebugMethodName = 0xf7929a45 "IL_STUB_ReversePInvoke", m_pszDebugClassName = 0xf79299fa "DomainBoundILStubClass", m_pszDebugMethodSignature = 0xf244c22c "instance bool *(int32&)", 
  m_pDebugMethodTable = {m_addr = 4064572276}, m_GcCover = 0x0, m_wFlags3AndTokenRemainder = 0, m_chunkIndex = 0 '\000', m_bFlags2 = 8 '\b', m_wSlotNumber = 65535, m_wFlags = 32815, 
  static s_ClassificationSizeTable = 0xf78b4ee0 <MethodDesc::s_ClassificationSizeTable>}
(gdb) p *pMD
$2 = {m_pszDebugMethodName = 0xf4e58da8 "Invoke", m_pszDebugClassName = 0xf62904bc "Cdeclcaller", m_pszDebugMethodSignature = 0xf4e58e2c "instance bool *(int32&)", m_pDebugMethodTable = {m_addr = 4129883552}, 
  m_GcCover = 0x0, m_wFlags3AndTokenRemainder = 16441, m_chunkIndex = 9 '\t', m_bFlags2 = 0 '\000', m_wSlotNumber = 12, m_wFlags = 3, 
  static s_ClassificationSizeTable = 0xf78b4ee0 <MethodDesc::s_ClassificationSizeTable>}

```

janvorli: @lucenticus I have a couple of thoughts:
* When I have briefly discussed this offline with @jkotas, he has said that he thinks the m_cbActualArgSize should actually be the stack arg size of the native caller after all, so no augmenting should be necessary and it is possible that there is some other issue with the [NativeCallable]
* Could you please dump the the disassembly of the IL_STUB_ReversePInvoke stub for the DoCallBack_MarshalRefInt_Stdcall test? That should confirm what arguments the stub really expects.

lucenticus: Here is a dump produced by command `COMPlus_JitDisasm="DomainBoundILStubClass:IL_STUB_ReversePInvoke"` for `DoCallBack_MarshalRefInt_Stdcall` before it crashed: 
```

; Assembly listing for method DomainBoundILStubClass:IL_STUB_ReversePInvoke(int):int
; Emitting BLENDED_CODE for generic X86 CPU
; debuggable code
; ebp based frame
; fully interruptible
;
;  V00 arg0         [V00,T00] (  3,   3  )     int  ->  [ebp-0x04]   do-not-enreg[H]
;  V01 loc0         [V01,T02] (  3,   3  )     int  ->  [ebp-0x08]   do-not-enreg[H]
;  V02 loc1         [V02    ] (  2,   2  )   byref  ->  [ebp-0x0C]   do-not-enreg[H] must-init
;  V03 loc2         [V03,T04] (  2,   2  )    bool  ->  [ebp-0x10]   do-not-enreg[H]
;  V04 loc3         [V04,T03] (  3,   3  )     int  ->  [ebp-0x14]   do-not-enreg[H]
;  V05 loc4         [V05,T05] (  2,   2  )     int  ->  [ebp-0x18]   do-not-enreg[H]
;  V06 tmp0         [V06    ] (  3,   3  )     int  ->  [ebp-0x24]   do-not-enreg[XH] addr-exposed
;  V07 tmp1         [V07,T01] (  2,   4  )     int  ->  [ebp-0x1C]   do-not-enreg[H]
;  V08 tmp2         [V08    ] (  2,   4  )     ref  ->  [ebp-0x20]   do-not-enreg[H] must-init
;
; Lcl frame size = 40

G_M33306_IG01:
       55           push     ebp
       8BEC         mov      ebp, esp
       83EC28       sub      esp, 40
       8945DC       mov      dword ptr [ebp-24H], eax
       8955F4       mov      dword ptr [ebp-0CH], edx
       8955E0       mov      dword ptr [ebp-20H], edx
       894DFC       mov      dword ptr [ebp-04H], ecx

G_M33306_IG02:
       33C9         xor      ecx, ecx
       894DF8       mov      dword ptr [ebp-08H], ecx

G_M33306_IG03:
       90           nop      
       8B4DFC       mov      ecx, bword ptr [ebp-04H]
       894DF4       mov      bword ptr [ebp-0CH], ecx
       90           nop      
       90           nop      
       90           nop      
       8B4DDC       mov      ecx, dword ptr [ebp-24H]
       8B4908       mov      ecx, dword ptr [ecx+8]
       8B09         mov      ecx, gword ptr [ecx]
       8B4904       mov      ecx, gword ptr [ecx+4]
       894DE0       mov      gword ptr [ebp-20H], ecx
       8B4DDC       mov      ecx, dword ptr [ebp-24H]
       8B4908       mov      ecx, dword ptr [ecx+8]
       8B09         mov      ecx, gword ptr [ecx]
       8B490C       mov      ecx, dword ptr [ecx+12]
       894DE4       mov      dword ptr [ebp-1CH], ecx
       90           nop      
       8B4DE0       mov      ecx, gword ptr [ebp-20H]
       8B55F4       mov      edx, bword ptr [ebp-0CH]
       8B45E4       mov      eax, dword ptr [ebp-1CH]
       FFD0         call     eax
       8945F0       mov      dword ptr [ebp-10H], eax
       8B45F0       mov      eax, dword ptr [ebp-10H]
       85C0         test     eax, eax
       0F95C0       setne    al
       0FB6C0       movzx    eax, al
       8945EC       mov      dword ptr [ebp-14H], eax
       90           nop      
       8B45EC       mov      eax, dword ptr [ebp-14H]
       8945E8       mov      dword ptr [ebp-18H], eax
       90           nop      
       90           nop      
       C745F8FFFFFF7F mov      dword ptr [ebp-08H], 0x7FFFFFFF
       E808000000   call     G_M33306_IG06
       90           nop      

G_M33306_IG04:
       8B45E8       mov      eax, dword ptr [ebp-18H]

G_M33306_IG05:
       8BE5         mov      esp, ebp
       5D           pop      ebp
       C3           ret      

G_M33306_IG06:
       83EC0C       sub      esp, 12

G_M33306_IG07:
       8B45F8       mov      eax, dword ptr [ebp-08H]
       3DFEFFFF7F   cmp      eax, 0x7FFFFFFE
       7F05         jg       SHORT G_M33306_IG08
       33C0         xor      eax, eax
       8945EC       mov      dword ptr [ebp-14H], eax

G_M33306_IG08:
       83C40C       add      esp, 12
       C3           ret      

; Total bytes of code 143, prolog size 20 for method DomainBoundILStubClass:IL_STUB_ReversePInvoke(int):int


```

janvorli: Ok, so the stub expects the argument in ecx and it has no stack argument. That means that the original m_cbActualArgSize (the value that came from pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize() in our case) is the correct value for the size of arguments pushed by the native caller and we should not augment it in any way. Also the numRegistersUsed that you've dumped is correct, since the stub has a single argument in ecx.

That means that the problem likely happens only for NativeCallable case and a fix should be made only for that case. I am not sure though if the right fix is to augment the m_cbActualArgSize  only for this case or if the issue is somewhere else and just exposed here. It is possible that the difference is actually not the number of args in registers and it is just a coincidence. That should be  possible to prove by adding one more parameter, e.g. an int, to the callback function signature in the test (and modify the function pointer / native code calling it accordingly). Then you'd see what you get as GetNativeStackArgSize() and numRegistersUsed.

janvorli:
@lucenticus, @parjong, @seanshpark I am leaving for a week of vacation in about 2 hours, so please cc @jkotas with your changes instead of me while I'm away. I'll be back on March 20.

src/vm/dllimportcallback.cpp
lucenticus: @janvorli Ok, thank you, I'll check that with more arguments. Have a nice vacation!

parjong:
@janvorli Have a nice vacation!

lucenticus:
It seems I found the reason of failed tests: in the NativeCallableTest `pStubMD == NULL`, so  for this case `m_cbActualArgSize == pMD->SizeOfArgStack()` and according to comment [here](https://github.com/dotnet/coreclr/blob/master/src/vm/method.hpp#L879) it doesn't include arguments passed in registers and we have to augment the m_cbActualArgSize for this case. For other test cases `pStubMD != NULL` and `m_cbActualArgSize == pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize()`. 
I've added commit, @jkotas please take a look whether my fix is correct.

src/vm/dllimportcallback.cpp
```diff
@@ -1412,6 +1412,8 @@ VOID UMThunkMarshInfo::RunTimeInit()
         new (&sigInfo) PInvokeStaticSigInfo(pMD);
     else
         new (&sigInfo) PInvokeStaticSigInfo(GetSignature(), GetModule());
+    if (pStubMD == NULL)
+        m_cbActualArgSize += numRegistersUsed * STACK_ELEM_SIZE;
```
jkotas: Mutating state is recipe for problems.

It would be better to compute the right number of arguments here and set it here in all cases; and move this:

```
    m_cbActualArgSize =
        (pStubMD != NULL) ? pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize() : pMD->SizeOfArgStack();
```

to the `#else` part of the ifdef.

src/vm/dllimportcallback.cpp
```diff
@@ -1412,6 +1412,8 @@ VOID UMThunkMarshInfo::RunTimeInit()
...
     if (sigInfo.GetCallConv() == pmCallConvCdecl)
```
jkotas: Is this right? Shouldn't the adjustment be applied to just `m_cbRetPop`?

The comment for `m_cbActualArgSize` says that it is cache for `m_pSig.SizeOfFrameArgumentArray`. It won't be the case with this adjustment.

lucenticus: If we apply adjustment to `m_cbRetPop` only, then test failed with incorrect result. I suppose it failed due to [this check](https://github.com/dotnet/coreclr/blob/master/src/vm/i386/umthunkstub.S#L89). Probably, we should change `UMThunkStub` too?

src/vm/dllimportcallback.cpp
```diff
@@ -1412,6 +1407,9 @@ VOID UMThunkMarshInfo::RunTimeInit()
         new (&sigInfo) PInvokeStaticSigInfo(pMD);
     else
         new (&sigInfo) PInvokeStaticSigInfo(GetSignature(), GetModule());
+
+    m_cbActualArgSize = (pStubMD != NULL) ? pStubMD->AsDynamicMethodDesc()->GetNativeStackArgSize()
```
jkotas: You should keep track of the size in the loop above. It is wasteful to get the size by calling these methods - MethodDesc::SizeOfArgStack runs the exact same loop over all arguments.

src/vm/dllimportcallback.cpp
jkotas: Yes, either fix UMThunkStub; or change what m_cbActualArgSize means on Linux x86 (update the comment next to it, and check that there are no other places that depend on the current meaning.

lucenticus:
@jkotas Thanks for review, I've added the commit with suggested changes. I've check `m_cbActualArgSize` and it seems that this meaning is applicable only if  `_TARGET_X86_ && FEATURE_STUBS_AS_IL` is defined, so there are no dependencies with the changes.

lucenticus:
@dotnet-bot test OSX x64 Checked Build and Test  please

jkotas:
Thanks!

lucenticus:
Thank you!

PR #10061, opened 2017-03-09
Add linux throughput testing support

adiaaida:
This change modifies perf-prep.sh, run-throughput-perf.py and perf.groovy
to add support for linux throughput testing. The following changes have
been made:

1. Update run-throughput-perf to specify the jit differently on linux and
windows. Create a new set of dlls to crossgen on linux. The following are
removed: CodeAnalysis, CodeAnalysis.VisualBasic, CodeAnalysis.Csharp
(these do not exist in the corefx runtime dir that we will be using on
linux), System, System.Core, System.XML (these are significantly smaller
on Linux than windows). Make the list of architectures a dictionary so we
can specify different arches for different oses. Change the path to
crossgen/the jit to the Product directory. Change the timer to
timeit.default_timer, which will pick the most accurate timer for each
platform.

2. Modify perf-prep for throughput purposes. In throughput testing, we
need to enlist in corefx and build it. We pick a specific commit to enlist
in so that testing will be consistent. The rest of the work that perf-prep
can be skipped for throughput testing, as we are not running tests.

3. Update perf.groovy with throughput jobs on Linux.

### DIFF
--- a/tests/scripts/perf-prep.sh
+++ b/tests/scripts/perf-prep.sh
@@ -11,6 +11,8 @@ function print_usage {
     echo ''
     echo 'Required arguments:'
     echo '  --branch=<path>             : branch where coreclr/corefx/test bits are copied from (e.g. dotnet_coreclr).'
+    echo 'Optional arguments:'
+    echo '  --throughput                : if we are running setup for a throughput run.'
 }
 
 # Exit code constants
@@ -20,6 +22,7 @@ readonly EXIT_CODE_SUCCESS=0       # Script ran normally.
 perfArch="x64"
 perfConfig="Release"
 perfBranch=
+throughput=0
 
 for i in "$@"
 do
@@ -30,6 +33,9 @@ do
             ;;
         --branch=*)
             perfBranch=${i#*=}
+            ;;
+        -t|--throughput)
+            throughput=1
             ;;
         *)
             echo "Unknown switch: $i"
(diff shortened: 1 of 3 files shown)


adiaaida:
@DrewScoggins @jashook PTAL
/cc @dotnet/rap-team

adiaaida:
@dotnet-bot test ci please
@dotnet-bot test Windows_NT x64 throughput please

adiaaida:
@dotnet-bot test ci please

adiaaida:
@dotnet-bot test Windows_NT x86jit32 throughput please

adiaaida:
@dotnet-bot test ci please

tests/scripts/perf-prep.sh
```diff
@@ -57,22 +63,38 @@ unzip -q -o benchview.zip -d ./tests/scripts/Microsoft.BenchView.JSONFormat
...
+    # Clone corefx
+    if [ -d "_" ]; then
+        rm -r -f _
+    fi
+    mkdir _
+    git clone https://github.com/dotnet/corefx.git _/fx
+    cd _/fx
+
+    # Checkout the specific commit we want
+    git checkout f0b9e238c08f62a1db90ef0378980ac771204d35
```
DrewScoggins: Why are we locking on this particular build of CoreFX, and how will we know when we want to upgrade this?

adiaaida: I locked on it because we don't want a moving testing target in addition to a moving jit. It just happened to be the commit that we were at when I was testing. I'm open to choosing a different commit (maybe some release commit?)

adiaaida:
@dotnet-bot test ci please

adiaaida:
@dotnet-bot test ci please

tests/scripts/perf-prep.sh
DrewScoggins: This makes sense.  I was unsure if the CoreFX binaries were required to run crossgen.  If they are not then we have no worries.

adiaaida:
Ubuntu throughput test run passed.

tests/scripts/run-throughput-perf.py
```diff
@@ -290,7 +329,7 @@ def main(args):
                 shutil.copy(csv_file_name, clr_root)
 
                 # For each benchmark, call measurement.py
-                measurement_args = ["py",
+                measurement_args = [python_exe,
```
DrewScoggins: This is something I guess I did not notice last time, but why do we check to see if benchview_path is None before making the call to upload, but not before using measurement.py?  Seems we should have the same check here as well.

DrewScoggins:
I also kicked off a run for Windows x64 throughput to make sure that the script changes don't break the Windows runs.

tests/scripts/run-throughput-perf.py
adiaaida: Good call. I can't believe I missed that.

adiaaida: It's on line 326

adiaaida:
I had kicked off a x86jit32 run, which passed (I made some additional linux-only changes to perf.groovy, so the build isn't linked anymore). There were no changes to the non-linux perf.groovy code, so the regular call to test runs works.

adiaaida:
@dotnet-bot test Windows_NT x64 throughput please

DrewScoggins:
I also kicked off a regular Ubuntu 14.04 perf run, want to make sure that the changes to perf-prep.sh did not break the regular runs.  If that and the other I fired off succeed LGTM.

adiaaida:
@dotnet-bot test Windows_NT x86jit32 throughput please

adiaaida:
@dotnet-bot test Ubuntu14.04 perf please

adiaaida:
Ubuntu and x64 perf/throughput testing passed. Jit32 testing is failing because I accidentally removed the set of arch = 'x86' in the jit32 case, and I also changed where we look for the jit dll, so that needed to be updated in perf.groovy.

adiaaida:
@dotnet-bot test ci please

PR #10064, opened 2017-03-09
memory limit in containers on linux

rahku:
This adds implementation in pal to get memory limit from cgroups. Memory limit values are written in file <path_to_memory_cgroup>/memory.limit_in_bytes and <path_to_memory_cgroup>/memory.memsw.limit_int_bytes. path_to_memory_cgroup is <path_to_memory_hierarchy>/<cgroup_name>. cgroup_name can be found in file /proc/self/cgroup. <path_to_memory_hierarchy>is normally set to /sys/fs/cgroup/memory. However this is not the standard. It can be changed to any mount point. Standard way of getting that value is from /proc/self/mountinfo. 

Note: CGroup implementation is not available on OSX. Docker does not run natively on OSX. It runs under a light-weigth hypervisor which runs a linux distro and you add container on top of it. So the current implementation so work just fine in this scenario.

### DIFF
--- a/src/gc/unix/cgroup.cpp
+++ b/src/gc/unix/cgroup.cpp
@@ -1,0 +1,342 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+/*++
+
+Module Name:
+
+    cgroup.cpp
+
+Abstract:
+    Read memory limits for the current process
+--*/
+#include <cstdint>
+#include <cstddef>
+#include <cassert>
+#include <unistd.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/resource.h>
+#include <errno.h>
+
+#define SIZE_T_MAX (~(size_t)0)
+#define PROC_MOUNTINFO_FILENAME "/proc/self/mountinfo"
+#define PROC_CGROUP_FILENAME "/proc/self/cgroup"
+#define PROC_STATM_FILENAME "/proc/self/statm"
+#define MEM_LIMIT_FILENAME "/memory.limit_in_bytes"
+
+class CGroup
+{
+    char* m_memory_cgroup_path;
+public:
+    CGroup()
+    {
+        m_memory_cgroup_path = nullptr;
+        char* memoryHierarchyMount = nullptr;
+        char *cgroup_path_relative_to_mount = nullptr;
+        size_t len;
+        memoryHierarchyMount = FindMemoryHierarchyMount();
+        if (memoryHierarchyMount == nullptr)
+            goto done;
+
+        cgroup_path_relative_to_mount = FindCGroupPathForMemorySubsystem();
+        if (cgroup_path_relative_to_mount == nullptr)
+            goto done;
+
+        len = strlen(memoryHierarchyMount);
+        len += strlen(cgroup_path_relative_to_mount);
+        m_memory_cgroup_path = (char*)malloc(len+1);
+        if (m_memory_cgroup_path == nullptr)
+           goto done;
+        
+        strcpy(m_memory_cgroup_path, memoryHierarchyMount);
+        strcat(m_memory_cgroup_path, cgroup_path_relative_to_mount);
+
+    done:
+        free(memoryHierarchyMount);
+        free(cgroup_path_relative_to_mount);
+    }
+
+    ~CGroup()
+    {
+        free(m_memory_cgroup_path);
+    }
+
+    bool GetPhysicalMemoryLimit(size_t *val)
+    {
+        char *mem_limit_filename = nullptr;
+        bool result = false;
+
+        if (m_memory_cgroup_path == nullptr)
+            return result;
+
+        size_t len = strlen(m_memory_cgroup_path);
+        len += strlen(MEM_LIMIT_FILENAME);
+        mem_limit_filename = (char*)malloc(len+1);
+        if (mem_limit_filename == nullptr)
+            return result;
+
+        strcpy(mem_limit_filename, m_memory_cgroup_path);
+        strcat(mem_limit_filename, MEM_LIMIT_FILENAME);
+        result = ReadMemoryValueFromFile(mem_limit_filename, val);
+        free(mem_limit_filename);
+        return result;
+    }
+    
+private:
+    char* FindMemoryHierarchyMount()
+    {
+        char *line = nullptr;
+        size_t lineLen = 0, maxLineLen = 0;
+        char *filesystemType = nullptr;
+        char *options = nullptr;
+        char* mountpath = nullptr;
+
+        FILE *mountinfofile = fopen(PROC_MOUNTINFO_FILENAME, "r");
+        if (mountinfofile == nullptr)
+            goto done;
+    
+        while (getline(&line, &lineLen, mountinfofile) != -1)
+        {
+            if (filesystemType == nullptr || lineLen > maxLineLen)
+            {
+                free(filesystemType);
+                free(options);
+                filesystemType = (char*)malloc(lineLen+1);
+                if (filesystemType == nullptr)
+                    goto done;
+                options = (char*)malloc(lineLen+1);
+                if (options == nullptr)
+                    goto done;
+                maxLineLen = lineLen;
+            }
+
+            char* separatorChar = strchr(line, '-');
+
+            // See man page of proc to get format for /proc/self/mountinfo file
+            int sscanfRet = sscanf(separatorChar, 
+                                   "- %s %*s %s",
+                                   filesystemType,
+                                   options);
+            if (sscanfRet != 2)
+            {
+                assert(!"Failed to parse mount info file contents with sscanf.");
+                goto done;
+            }
+    
+            if (strncmp(filesystemType, "cgroup", 6) == 0)
+            {
+                char* context = nullptr;
+                char* strTok = strtok_r(options, ",", &context); 
+                while (strTok != nullptr)
+                {
+                    if (strncmp("memory", strTok, 6) == 0)
+                    {
+                        mountpath = (char*)malloc(lineLen+1);
+                        if (mountpath == nullptr)
+                            goto done;
+    
+                        sscanfRet = sscanf(line,
+                                           "%*s %*s %*s %*s %s ",
+                                           mountpath);
+                        if (sscanfRet != 1)
+                        {
+                            free(mountpath);
+                            mountpath = nullptr;
+                            assert(!"Failed to parse mount info file contents with sscanf.");
+                        }
+                        goto done;
+                    }
+                    strTok = strtok_r(nullptr, ",", &context);
+                }
+            }
+        }
+    done:
+        free(filesystemType);
+        free(options);
+        free(line);
+        if (mountinfofile)
+            fclose(mountinfofile);
+        return mountpath;
+    }
+    
+    char* FindCGroupPathForMemorySubsystem()
+    {
+        char *line = nullptr;
+        size_t lineLen = 0;
+        size_t maxLineLen = 0;
+        char *subsystem_list = nullptr;
+        char *cgroup_path = nullptr;
+        bool result = false;
+
+        FILE *cgroupfile = fopen(PROC_CGROUP_FILENAME, "r");
+        if (cgroupfile == nullptr)
+            goto done;
+    
+        while (!result && getline(&line, &lineLen, cgroupfile) != -1)
+        {
+            if (subsystem_list == nullptr || lineLen > maxLineLen)
+            {
+                free(subsystem_list);
+                free(cgroup_path);
+                subsystem_list = (char*)malloc(lineLen+1);
+                if (subsystem_list == nullptr)
+                    goto done;
+                cgroup_path = (char*)malloc(lineLen+1);
+                if (cgroup_path == nullptr)
+                    goto done;
+                maxLineLen = lineLen;
+            }
+                   
+            // See man page of proc to get format for /proc/self/cgroup file
+            int sscanfRet = sscanf(line, 
+                                   "%*[^:]:%[^:]:%s",
+                                   subsystem_list,
+                                   cgroup_path);
+            if (sscanfRet != 2)
+            {
+                assert(!"Failed to parse cgroup info file contents with sscanf.");
+                goto done;
+            }
+    
+            char* context = nullptr;
+            char* strTok = strtok_r(subsystem_list, ",", &context); 
+            while (strTok != nullptr)
+            {
+                if (strncmp("memory", strTok, 6) == 0)
+                {
+                    result = true;
+                    break;  
+                }
+                strTok = strtok_r(nullptr, ",", &context);
+            }
+        }
+    done:
+        free(subsystem_list);
+        if (!result)
+        {
+            free(cgroup_path);
+            cgroup_path = nullptr;
+        }
+        free(line);
+        if (cgroupfile)
+            fclose(cgroupfile);
+        return cgroup_path;
+    }
+    
+    bool ReadMemoryValueFromFile(const char* filename, size_t* val)
+    {
+        bool result = false;
+        char *line = nullptr;
+        size_t lineLen = 0;
+        char* endptr = nullptr;
+        size_t num = 0, l, multiplier;
+        FILE* file = nullptr;
+    
+        if (val == nullptr)
+            goto done;
+    
+        file = fopen(filename, "r");
+        if (file == nullptr)
+            goto done;
+        
+        if (getline(&line, &lineLen, file) == -1)
+            goto done;
+    
+        errno = 0;
+        num = strtoull(line, &endptr, 0); 
+        if (errno != 0)
+            goto done;
+    
+        multiplier = 1;
+        switch(*endptr)
+        {
+            case 'g':
+            case 'G': multiplier = 1024;
+            case 'm': 
+            case 'M': multiplier = multiplier*1024;
+            case 'k':
+            case 'K': multiplier = multiplier*1024;
+        }
+    
+        *val = num * multiplier;
+        result = true;
+        if (*val/multiplier != num)
+            result = false;
+    done:
+        if (file)
+            fclose(file);
+        free(line);    
+        return result;
+    }
+};
+   
+size_t GetRestrictedPhysicalMemoryLimit()
+{
+    CGroup cgroup;
+    size_t physical_memory_limit;
+ 
+    if (!cgroup.GetPhysicalMemoryLimit(&physical_memory_limit))
+         physical_memory_limit = SIZE_T_MAX;
+
+    struct rlimit curr_rlimit;
+    size_t rlimit_soft_limit = RLIM_INFINITY;
+    if (getrlimit(RLIMIT_AS, &curr_rlimit) == 0)
+    {
+        rlimit_soft_limit = curr_rlimit.rlim_cur;
+    }
+    physical_memory_limit = (physical_memory_limit < rlimit_soft_limit) ? 
+                            physical_memory_limit : rlimit_soft_limit;
+
+    // Ensure that limit is not greater than real memory size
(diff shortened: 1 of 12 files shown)


rahku:
I have tested this on ubuntu. I am still doing some additional testing.

rahku:
PTAL @janvorli  @gkhanna79 @Maoni0

src/gc/unix/gcenv.unix.cpp
```diff
@@ -78,6 +78,22 @@ static uint8_t g_helperPage[OS_PAGE_SIZE] __attribute__((aligned(OS_PAGE_SIZE)))
 // Mutex to make the FlushProcessWriteBuffersMutex thread safe
 static pthread_mutex_t g_flushProcessWriteBuffersMutex;
 
+static size_t g_RestrictedPhysicalMemoryLimit = (size_t)MAX_PTR;
+
+static size_t GetRestrictedPhysicalMemoryLimit()
```
janvorli: This file is for pure Unix only and so you cannot use any PAL functions here. It is used when the GC is compiled as a standalone library.

janvorli: Also, no contracts and no windowsizms should be here.

rahku: that would imply code duplication..is that ok

janvorli:
Yes, code duplication is ok here.

src/pal/src/misc/cgroup.cpp
```diff
@@ -0,0 +1,275 @@
...
+            char options[MAX_PATH];
+
+            char* separatorChar = strchr(line, '-');
+
+            // See man page of proc to get format for /proc/self/mountinfo file
+            int sscanfRet = sscanf_s(separatorChar, 
+                                     "- %s %*s %s",
+                                     filesystemType, (unsigned)_countof(filesystemType),
+                                     options, (unsigned)_countof(options));
+            if (sscanfRet != 2)
```
janvorli: Could you please unify the spacing style between the if / while and the following `(`? Sometimes you have space there and sometimes not. I prefer the style with space, but I'll leave it up to you.

src/pal/src/misc/cgroup.cpp
```diff
@@ -0,0 +1,275 @@
...
+#define MAX_SIZE_T (~(SIZE_T)0)
+#define PROC_MOUNTINFO_FILENAME "/proc/self/mountinfo"
+#define PROC_CGROUP_FILENAME "/proc/self/cgroup"
+#define PROC_STATM_FILENAME "/proc/self/statm"
+#define MEM_LIMIT_FILENAME "/memory.limit_in_bytes"
+#define MEM_SWAP_LIMIT_FILENAME "/memory.memsw.limit_in_bytes"
+
+class CGroup
+{
+    char _memory_cgroup_path[MAX_PATH];
```
janvorli: Hmm, I have just realized that the StackString is not available out of PAL for the Unix only implementation, so please ignore this.

src/gc/unix/cgroup.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+    while(!retval && getline(&line, &lineLen, mountinfofile) != -1)
+    {
+        char filesystemType[100];
+        char options[MAX_PATH];
+
+        char* separatorChar = strchr(line, '-');
+
+        // See man page of proc to get format for /proc/self/mountinfo file
+        int sscanfRet = sscanf_s(separatorChar,
```
janvorli: the sscanf_s and other _s functions are not available for plain Unix. You'll need to use the plain ones or maybe better add defines that define sscanf_s as sscanf etc so that the rest of the code is the same as for the PAL version.

src/pal/tests/palsuite/miscellaneous/CGroup/test1/test.cpp
```diff
@@ -0,0 +1,56 @@
...
+**  6. echo 4M > /sys/fs/cgroup/memory/mygroup/memory.limit_in_bytes
+**  7. cgexe -g memory:/mygroup --sticky <application>
+**=========================================================*/
+
+#include <palsuite.h>
+
+int __cdecl main(int argc,char *argv[]) 
+{
+    
+    char * AnExampleString = "this is the string";
```
janvorli: These three variables doesn't seem to be used. Can you remove them please?

src/pal/tests/palsuite/miscellaneous/CGroup/test1/test.cpp
```diff
@@ -0,0 +1,56 @@
...
+    {
+      return FAIL;
+    }
+
+    size_t mem_limit = PAL_GetRestrictedPhysicalMemoryLimit();
+
+    FILE* file = fopen("/sys/fs/cgroup/memory/mygroup/memory.limit_in_bytes", "r");
+    if(file != NULL)
+    {
+        if(mem_limit != 4194304)
```
janvorli: I am surprised the test can run at all with just 4MB limit. But I assume you have verified that.

src/gc/unix/cgroup.h
```diff
@@ -0,0 +1,36 @@
...
+
+Module Name:
+
+    cgroup.h
+
+Abstract:
+    Read memory limits for the current process
+--*/
+
+#define MAX_SIZE_T (~(SIZE_T)0)
```
janvorli: It seems that these defines can be in the cgroup.cpp since they are not part of the public interface of the CGroup.

src/vm/gcenv.os.cpp
```diff
@@ -465,11 +480,9 @@ uint64_t GCToOSInterface::GetPhysicalMemoryLimit()
 {
     LIMITED_METHOD_CONTRACT;
 
-#ifndef FEATURE_PAL
     size_t restricted_limit = GetRestrictedPhysicalMemoryLimit();
-    if (restricted_limit != 0)
+    if (restricted_limit != 0 && restricted_limit != (size_t)MAX_PTR)
```
janvorli: Is the added check needed? It looks like the GetRestrictedPhysicalMemoryLimit should return zero if there is no limit to be on par with the Windows version.

src/vm/gcenv.os.cpp
```diff
@@ -489,17 +502,24 @@ void GCToOSInterface::GetMemoryStatus(uint32_t* memory_load, uint64_t* available
 {
     LIMITED_METHOD_CONTRACT;
 
-#ifndef FEATURE_PAL
     uint64_t restricted_limit = GetRestrictedPhysicalMemoryLimit();
-    if (restricted_limit != 0)
+    if (restricted_limit != 0 && restricted_limit != (size_t)MAX_PTR)
```
janvorli: The same question here

src/gc/unix/cgroup.cpp
```diff
@@ -0,0 +1,248 @@
...
+    errno = 0;
+    num = strtoull(line, &endptr, 0); 
+    if(errno != 0)
+        goto done;
+
+    multiplier = 1;
+    switch(*endptr)
+    {
+        case 'g':
+        case 'G': multiplier = 1024;
```
gkhanna79: Are there no breaks expected here?

janvorli:
@rahku have you tried to build it on Linux with the buildstandalonegc option? That compiles all the pure unix parts that are not built otherwise.

rahku:
No i did not know that ...will do

src/gc/unix/cgroup.cpp
rahku: no breaks are not needed it needs to fall though

janvorli: The code as it is would not work. You are multiplying bool (result) by 1024. I guess the intention was to use multiplier = multiplier * 1024 in there.

swgillespie:
There's also a CI job that runs `buildstandalonegc`: https://ci.dot.net/job/dotnet_coreclr/job/master/job/x64_checked_ubuntu_standalone_gc/ . I think it can be triggered on PRs with `test Ubuntu Checked standalone_gc`.

src/gc/unix/cgroup.cpp
rahku: yes. thanks for catching it. I did test it before and then changed variable names but missed here.

src/gc/unix/cgroup.cpp
```diff
@@ -0,0 +1,248 @@
...
+        char* separatorChar = strchr(line, '-');
+
+        // See man page of proc to get format for /proc/self/mountinfo file
+        int sscanfRet = sscanf_s(separatorChar, 
+                                 "- %s %*s %s",
+                                 filesystemType, (unsigned)_countof(filesystemType),
+                                 options, (unsigned)_countof(options));
+        if (sscanfRet != 2)
+        {
+            _ASSERTE(!"Failed to parse mount info file contents with sscanf_s.");
```
jkotas: We prefer vanilla C/C++ for the standalone GC PAL: _ASSERTE -> assert

rahku: yeah i m in process of making these changes.

rahku:
@dotnet-bot test Ubuntu Checked standalone_gc

rahku:
@dotnet-bot test Ubuntu x64 Checked standalone_gc

rahku:
PTAL i have incorprorated all review comments

src/vm/gcenv.os.cpp
```diff
@@ -489,17 +502,24 @@ void GCToOSInterface::GetMemoryStatus(uint32_t* memory_load, uint64_t* available
 {
     LIMITED_METHOD_CONTRACT;
 
-#ifndef FEATURE_PAL
     uint64_t restricted_limit = GetRestrictedPhysicalMemoryLimit();
     if (restricted_limit != 0)
     {
+        size_t workingSetSize;
+        BOOL status = false;
```
Maoni0: >false [](start = 22, length = 5)

nit... FALSE instead of false.

src/gc/unix/cgroup.h
```diff
@@ -0,0 +1,31 @@
...
+
+Abstract:
+    Read memory limits for the current process
+--*/
+
+#define MAX_PATH 1024
+
+class CGroup
+{
+    char m_memory_cgroup_path[MAX_PATH];
```
jkotas: Hungarian notation is Windowsism... . Change it to just `_memory_cgroup_path` ?

src/gc/unix/cgroup.h
```diff
@@ -0,0 +1,31 @@
...
+    CGroup();
+    bool GetPhysicalMemoryLimit(size_t* val);
+    bool GetSwapMemoryLimit(size_t* val);
+private:
+    bool FindMemoryHierarchyMount(char* mountpath);
+    bool FindCGroupPathForMemorySubsystem(char* cgrouppath);
+    bool ReadMemoryValueFromFile(const char* filename, size_t* val);
+};
+
+size_t GetRestrictedPhysicalMemoryLimit();
```
jkotas: It would look better to have these methods as static public methods on CGroup, and have all instance methods in CGroup to be private;  or move `class CGroup` to the .cpp file.

jkotas: Since it is in the .cpp file in the CoreCLR PAL, I think it should be in the .cpp file here as well.

src/gc/unix/cgroup.cpp
```diff
@@ -0,0 +1,271 @@
...
+#define SIZE_T_MAX (~(size_t)0)
+#define PROC_MOUNTINFO_FILENAME "/proc/self/mountinfo"
+#define PROC_CGROUP_FILENAME "/proc/self/cgroup"
+#define PROC_STATM_FILENAME "/proc/self/statm"
+#define MEM_LIMIT_FILENAME "/memory.limit_in_bytes"
+#define MEM_SWAP_LIMIT_FILENAME "/memory.memsw.limit_in_bytes"
+
+class CGroup
+{
+    char _memory_cgroup_path[MAX_PATH];
```
janvorli: @jkotas I have asked @rahku to use m_ convention instead, but I could see you have asked him to change it back However, we use m_ in all coreclr PAL and also in CoreRT native code, so I'd really prefer staying in line with that.

src/gc/unix/cgroup.cpp
```diff
@@ -0,0 +1,271 @@
...
+    bool ReadMemoryValueFromFile(const char* filename, size_t* val)
+    {
+        bool result = false;
+        char *line = nullptr;
+        size_t lineLen = 0;
+        char* endptr = NULL;
+        size_t num = 0, l, multiplier;
+        FILE* file = NULL;
+    
+        if (val == NULL)
```
janvorli: A nit - you have this check only here, but not in the other copy of the function.

src/gc/unix/cgroup.cpp
rahku: I think @jkotas was asking to change in this file which is for standalone gc & linux specific and so should not have any windowism?

jkotas: I do not have a strong opinion. I will leave it up to you what you like best.

src/gc/unix/gcenv.unix.cpp
```diff
@@ -78,6 +78,23 @@ static uint8_t g_helperPage[OS_PAGE_SIZE] __attribute__((aligned(OS_PAGE_SIZE)))
...
+
+static size_t g_RestrictedPhysicalMemoryLimit = 0;
+
+static size_t GetRestrictedPhysicalMemoryLimit()
+{
+    // The limit was cached already
+    if (g_RestrictedPhysicalMemoryLimit != 0)
+        return g_RestrictedPhysicalMemoryLimit;
+
+    size_t memory_limit = GetRestrictedPhysicalMemoryLimit();
```
jkotas: Isn't there an indefinite recursion (GetRestrictedPhysicalMemoryLimit calling GetRestrictedPhysicalMemoryLimit)? I think this can be moved to `GCToOSInterface::GetPhysicalMemoryLimit`.

src/gc/unix/cgroup.cpp
```diff
@@ -0,0 +1,271 @@
...
+            fclose(file);
+        if (line)
+            free(line);    
+        return result;
+    }
+};
+   
+size_t GetRestrictedPhysicalMemoryLimit()
+{
+   CGroup cgroup;
```
jkotas: Nit: This does not seem to be using 4 space indentation

src/gc/unix/cgroup.cpp
```diff
@@ -0,0 +1,271 @@
...
+    
+        while (!retval && getline(&line, &lineLen, cgroupfile) != -1)
+        {
+            char subsystem_list[100];
+            char cgroup_path[MAX_PATH];
+    
+            // See man page of proc to get format for /proc/self/cgroup file
+            int sscanfRet = sscanf(line, 
+                                   "%*[^:]:%[^:]:%s",
+                                   subsystem_list,
```
jkotas: Is this hardened against buffer overruns? What guarantees that the result will fit into subsystem_list?

rahku: There are no secure versions of these in linux. Are you suggesting to not to use these non-secure crt functions.

jkotas: If you are using these non-secure crt functions, it is up to you to guarantee no buffer overruns.

rahku:
unix specific cgroup.cpp is now hardened against buffer overruns. @jkotas does that look fine now?

rahku:
I have tested this on fedora & ubuntu and also done some end-to-end testing and results look good. One of the difference that I found wrt to windows is that on Linux when we go out of memory the process is killed by OS instead of allocation failure. This is with the default setting of the machine...there might be settings to change it.

rahku:
@janvorli I also now do not look at file memory.memsw.limit_in_bytes. This file gives the swap space limit. Swap space limits the virtual address of the process and does not impact real memory that can be used by the process (in our case I am assuming only a single process to be run in a cgroup). But for gc we need real memory pressure and not virtual address space pressure.

rahku:

rahku:

src/gc/unix/cgroup.cpp
```diff
@@ -0,0 +1,292 @@
...
+            m_memory_cgroup_path[0] = '\0';
+            return;
+        }
+        strncat(m_memory_cgroup_path, cgroup_path_relative_to_mount, 
+                MAX_PATH-strlen(m_memory_cgroup_path)-1);
+    }
+    
+    bool GetPhysicalMemoryLimit(size_t *val)
+    {
+        char mem_limit_filename[MAX_PATH];
```
janvorli: Some time ago, we have made a sweeping change over the CoreCLR native code including PAL to get rid of fixed size buffers for file paths unless the path is guaranteed to never exceed that limit. So introducing such a fixed size buffer here makes me nervous. I would prefer using a dynamically allocated buffer at this place. You can use the length of the line from the mount info file that contains the path as an memory size estimate - we would just waste a little memory that's not needed for the path. The FindMemoryHierarchyMount can then return nullptr or the allocated buffer.
I think that similar thing can be done at other places where you use MAX_PATH.

src/pal/src/misc/cgroup.cpp
```diff
@@ -0,0 +1,272 @@
...
+    char* line = NULL;
+
+    if (val == NULL)
+        return FALSE;
+
+    FILE* file = fopen(PROC_STATM_FILENAME, "r");
+    if (file != NULL && getline(&line, &linelen, file) != -1)
+    {
+        int x = sscanf_s(line, "%*s %llu %*s %*s %*s %*s %*s", val);
+        if(x==1)
```
janvorli: A nit - could you please add spaces around the `==` to unify it with the rest of the code?

src/gc/unix/gcenv.unix.cpp
```diff
@@ -471,13 +488,13 @@ void GCToOSInterface::GetMemoryStatus(uint32_t* memory_load, uint64_t* available
...
 
         // Get the physical memory in use - from it, we can get the physical memory available.
         // We do this only when we have the total physical memory available.
-        if (total > 0)
+        if (total > 0 && GetWorkingSetSize(&used))
         {
-            available = sysconf(_SC_PHYS_PAGES) * sysconf(_SC_PAGE_SIZE);
-            uint64_t used = total - available;
+            available = total > used ? total-used : 0; 
             load = (uint32_t)((used * 100) / total);
```
janvorli: This could overflow on 32 bit hardware, you'll need to keep the used type to be uint64_t.

src/gc/unix/cgroup.cpp
```diff
@@ -0,0 +1,339 @@
...
+            {
+                *val = *val * pageSize;
+                result = true;
+            }
+        }
+    }
+
+    if (file)
+        fclose(file);
+    if (line)
```
janvorli: The same case in the PAL version of this code for all the occurences.

src/pal/src/misc/cgroup.cpp
```diff
@@ -0,0 +1,334 @@
...
+    size_t rlimit_soft_limit = (size_t)RLIM_INFINITY;
+    if (getrlimit(RLIMIT_AS, &curr_rlimit) == 0)
+    {
+        rlimit_soft_limit = curr_rlimit.rlim_cur;
+    }
+    physical_memory_limit = min(physical_memory_limit, rlimit_soft_limit);
+
+    // Ensure that limit is not greater than real memory size
+    size_t pages = (size_t) sysconf(_SC_PHYS_PAGES);
+    if (pages != -1)
```
janvorli: size_t is an unsigned type

rahku:

rahku:
@hseok-oh you recently changed the arm test runs. Everything was passing earlier for me before your changes. I do not have a device/emulator to debug this. Could you take a look at failures or point to instructions on how to use arm emulator.

janvorli:
@rahku - The problem is some buffer overrun. I can see in the log the following which indicates that:
`*** stack smashing detected ***`

hseok-oh:
@rahku In fact, our CI ARM emulator didn't test run since test run script changed. That could check build success only. So we removed old ARM emulator build test and reconfigured build and test based on docker.

PR #10065, opened 2017-03-09
[Local GC] Break EE dependency on GC's generation table and alloc lock in single-proc scenarios

swgillespie:
This PR is a follow-up to #9836, with the feedback given in the last PR (hopefully) fully addressed.

The problem: when running on a single-proc x86/x64 Windows machine, a number of assembly allocation helpers directly reference the GC's generation table and alloc lock (`m_GCLock`). The generation table is used because the structure at offset 0 from the start of the generation table is the Gen 0 alloc context. When allocating, the EE acquires the alloc lock and uses the Gen 0 alloc context to allocate a new object in the normal way. If an object can't be allocated in this way (either because the assembly helper failed to acquire the alloc lock or the alloc context's allocation pointer exceeds the alloc limit), the assembly helper takes a slow path that eventually calls into the GC.

This PR modifies this process slightly by moving both of these concepts (a "global" alloc context for all SOH allocations and the alloc lock) to the EE side of the GC/EE interface. Some new functions were added to both `IGCHeap` and `IGCToCLR` to accommodate this:
 * `IGCToCLR::GetGlobalAllocContext` - this new function is called by the GC whenever it needs a pointer to the global alloc context - when fixing allocation contexts.
* `IGCHeap::UseThreadAllocationContexts` - this function is called by the EE whenever it needs to know whether or not the GC intends to use per-thread allocation contexts. Today, per-thread allocation contexts are used on all platforms other than single-proc x86/x64 Windows.

A new class, `GlobalAllocLock`, was also added to facilitate the acquiring and release of the global alloc lock, which guards the global alloc context against concurrent modification and is now also owned by the EE. This lock is acquired by the assembly allocation helpers as well ass `Alloc` and `AllocAlign8` in `gchelpers.cpp`, the slow path for allocations.

This PR also removes two methods from `IGCHeap`, because they are no longer needed: `IGCHeap::Alloc(size_t, size_t)` and `IGCHeap::AllocAlilgn8(size_t, size_t`. Previously, they were used to allocate using the GC's gen 0 alloc context, but they are not necessary anymore since the EE can acquire it's alloc lock and pass a pointer to its global alloc context to the versions of `Alloc` and `AllocAlign8` that accept allocation contexts to achieve the same thing.

I did not remove `m_GCLock` from the GC, because it is still acquired by `GCHeap::StartNoGCRegion` and `GCHeap::EndNoGCRegion`. If this lock is intended to prevent all allocations while these methods are being called, I can ensure that the VM acquires the lock before calling into these functions - if not, and it is intended to synchronize between the two methods, it should work as it is. (Though, perhaps the lock should be renamed if it's only protecting these two methods). I'm not sure about which one of these is true, so I welcome feedback!

### DIFF
--- a/src/gc/env/gcenv.ee.h
+++ b/src/gc/env/gcenv.ee.h
@@ -68,12 +68,6 @@ public:
     static void StompWriteBarrier(WriteBarrierParameters* args);
 
     static void EnableFinalization(bool foundFinalizers);
-
-    // Global alloc context retrieval
-    static gc_alloc_context *GetGlobalAllocContext();
-
-    // Whether the EE is using per-thread allocation contexts.
-    static bool UseThreadAllocationContexts();
 };
 
 #endif // __GCENV_EE_H__
--- a/src/gc/env/gcenv.ee.h
+++ b/src/gc/env/gcenv.ee.h
@@ -71,6 +71,9 @@ public:
 
     // Global alloc context retrieval
     static gc_alloc_context *GetGlobalAllocContext();
+
+    // Whether the EE is using per-thread allocation contexts.
+    static bool UseThreadAllocationContexts();
 };
 
 #endif // __GCENV_EE_H__
--- a/src/gc/env/gcenv.ee.h
+++ b/src/gc/env/gcenv.ee.h
@@ -68,6 +68,9 @@ public:
     static void StompWriteBarrier(WriteBarrierParameters* args);
 
     static void EnableFinalization(bool foundFinalizers);
+
+    // Global alloc context retrieval
+    static gc_alloc_context *GetGlobalAllocContext();
 };
 
 #endif // __GCENV_EE_H__
--- a/src/gc/gc.cpp
+++ b/src/gc/gc.cpp
@@ -68,6 +68,24 @@ int compact_ratio = 0;
 // See comments in reset_memory.
 BOOL reset_mm_p = TRUE;
 
+bool g_fFinalizerRunOnShutDown = false;
+
+#ifdef FEATURE_SVR_GC
+bool g_built_with_svr_gc = true;
+#else
+bool g_built_with_svr_gc = false;
+#endif // FEATURE_SVR_GC
+
+#if defined(BUILDENV_DEBUG)
+uint8_t g_build_variant = 0;
+#elif defined(BUILDENV_CHECKED)
+uint8_t g_build_variant = 1;
+#else
+uint8_t g_build_variant = 2;
+#endif // defined(BUILDENV_DEBUG)
+
+VOLATILE(int32_t) g_no_gc_lock = -1;
+
 #if defined (TRACE_GC) && !defined (DACCESS_COMPILE)
 const char * const allocation_state_str[] = {
     "start",
@@ -92,7 +110,6 @@ const char * const allocation_state_str[] = {
     "check_retry_seg"
 };
 #endif //TRACE_GC && !DACCESS_COMPILE
-
 
 // Keep this in sync with the definition of gc_reason
 #if (defined(DT_LOG) || defined(TRACE_GC)) && !defined (DACCESS_COMPILE)
@@ -150,7 +167,6 @@ size_t GetHighPrecisionTimeStamp()
 }
 #endif
 
-VOLATILE(int32_t) g_no_gc_lock = -1;
 
 #ifdef GC_STATS
 // There is a current and a prior copy of the statistics.  This allows us to display deltas per reporting
@@ -36897,6 +36913,11 @@ BOOL GCHeap::IsConcurrentGCEnabled()
 #else
     return FALSE;
 #endif //BACKGROUND_GC
+}
+
+void GCHeap::SetFinalizeRunOnShutdown(bool value)
+{
+    g_fFinalizerRunOnShutDown = value;
 }
 
 void PopulateDacVars(GcDacVars *gcDacVars)
--- a/src/gc/gc.cpp
+++ b/src/gc/gc.cpp
@@ -2685,10 +2685,6 @@ BOOL        gc_heap::heap_analyze_enabled = FALSE;
 
 #ifndef MULTIPLE_HEAPS
 
-extern "C" {
-    generation generation_table[NUMBERGENERATIONS + 1];
-}
-
 alloc_list gc_heap::loh_alloc_list [NUM_LOH_ALIST-1];
 alloc_list gc_heap::gen2_alloc_list[NUM_GEN2_ALIST-1];
 
@@ -2731,9 +2727,7 @@ int         gc_heap::gen0_must_clear_bricks = 0;
 CFinalize*  gc_heap::finalize_queue = 0;
 #endif // FEATURE_PREMORTEM_FINALIZATION
 
-#ifdef MULTIPLE_HEAPS
 generation gc_heap::generation_table [NUMBERGENERATIONS + 1];
-#endif // MULTIPLE_HEAPS
 
 size_t     gc_heap::interesting_data_per_heap[max_idp_count];
 
@@ -2746,6 +2740,8 @@ size_t     gc_heap::interesting_mechanism_bits_per_heap[max_gc_mechanism_bits_co
 #endif // MULTIPLE_HEAPS
 
 /* end of per heap static initialization */
+
+BOOL gc_heap::using_allocation_contexts = true;
 
 /* end of static initialization */
 
@@ -5716,18 +5712,29 @@ void gc_mechanisms::record (gc_history_global* history)
 //as opposed to concurrent heap verification
 void gc_heap::fix_youngest_allocation_area (BOOL for_gc_p)
 {
-    assert (alloc_allocated);
-    alloc_context* acontext = generation_alloc_context (youngest_generation);
-    dprintf (3, ("generation 0 alloc context: ptr: %Ix, limit %Ix",
-                 (size_t)acontext->alloc_ptr, (size_t)acontext->alloc_limit));
-    fix_allocation_context (acontext, for_gc_p, get_alignment_constant (TRUE));
-    if (for_gc_p)
+    alloc_context *acontext;
+
+    if (!using_allocation_contexts)
     {
-        acontext->alloc_ptr = alloc_allocated;
-        acontext->alloc_limit = acontext->alloc_ptr;
+        acontext = static_cast<alloc_context*>(GCToEEInterface::GetGlobalAllocContext());
+        assert(acontext != nullptr);
+        dprintf (3, ("EE-owned gen 0 alloc context %Ix: ptr %Ix, limit: %Ix",
+              (size_t)acontext, (size_t)acontext->alloc_ptr, (size_t)acontext->alloc_limit));
     }
-    heap_segment_allocated (ephemeral_heap_segment) =
-        alloc_allocated;
+    else
+    {
+        acontext = generation_alloc_context (youngest_generation);
+        assert(acontext != nullptr);
+        dprintf (3, ("generation 0 alloc context: ptr: %Ix, limit %Ix",
+                 (size_t)acontext->alloc_ptr, (size_t)acontext->alloc_limit));
+    }
+
+    fix_allocation_context (acontext, for_gc_p, get_alignment_constant (TRUE));
+    heap_segment_allocated (ephemeral_heap_segment) = alloc_allocated;
+
+
+    dprintf (3, ("youngest allocation context fixed: ptr %Ix, limit %Ix",
+          (size_t)acontext->alloc_ptr, (size_t)acontext->alloc_limit));
 }
 
 void gc_heap::fix_large_allocation_area (BOOL for_gc_p)
@@ -5831,13 +5838,27 @@ void void_allocation (gc_alloc_context* acontext, void*)
 
 void gc_heap::repair_allocation_contexts (BOOL repair_p)
 {
-    GCToEEInterface::GcEnumAllocContexts (repair_p ? repair_allocation : void_allocation, NULL);
-
-    alloc_context* acontext = generation_alloc_context (youngest_generation);
-    if (repair_p)
-        repair_allocation (acontext, NULL);
+    alloc_context *acontext;
+    if (!using_allocation_contexts)
+    {
+        // no need to enumerate alloc contexts if we aren't using per-thread contexts
+        // there's only one alloc context used for allocation, and it's the global one
+        acontext = static_cast<alloc_context*>(GCToEEInterface::GetGlobalAllocContext());
+    }
     else
+    {
+        GCToEEInterface::GcEnumAllocContexts (repair_p ? repair_allocation : void_allocation, NULL);
+        acontext = generation_alloc_context (youngest_generation);
+    }
+
+    if (repair_p)
+    {
+        repair_allocation (acontext, NULL);
+    }
+    else
+    {
         void_allocation (acontext, NULL);
+    }
 }
 
 struct fix_alloc_context_args
@@ -5857,7 +5878,10 @@ void gc_heap::fix_allocation_contexts(BOOL for_gc_p)
     fix_alloc_context_args args;
     args.for_gc_p = for_gc_p;
     args.heap = __this;
-    GCToEEInterface::GcEnumAllocContexts(fix_alloc_context, &args);
+    if (using_allocation_contexts)
+    {
+        GCToEEInterface::GcEnumAllocContexts(fix_alloc_context, &args);
+    }
 
     fix_youngest_allocation_area(for_gc_p);
     fix_large_allocation_area(for_gc_p);
@@ -10016,6 +10040,15 @@ HRESULT gc_heap::initialize_gc (size_t segment_size,
     {
         hres = E_FAIL;
     }
+
+#ifndef MULTIPLE_HEAPS
+    if (!g_theGCHeap->UseThreadAllocationContexts())
+    {
+        dprintf (3, ("Not using allocation contexts b/c workstation GC and %d CPUs",
+              GCToOSInterface::GetCurrentProcessCpuCount()));
+        gc_heap::using_allocation_contexts = false;
+    }
+#endif // MULTIPLE_HEAPS
 
     return hres;
 }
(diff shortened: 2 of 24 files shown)


swgillespie:
Looks like I have problems with x86, investigating now.

Maoni0:
So as we discussed in the meeting, you can create a new lock for the no gc region code.

If you want to make it nicer, you could also get rid of the  IGCToCLR::GetGlobalAllocContext  API altogether and just change the implementation of GCToEEInterface::GcEnumAllocContexts to enumerate the EE alloc context on UP.

swgillespie:
@Maoni0 I like that, will do!

swgillespie:
So the x86 failure is due to a circular dependency between the initialization of the JIT and the GC. On x86, assembly helpers are dynamically generated on JIT startup, and so when the GC initializes itself and attempts to stomp write barriers, the write barriers themselves haven't been written into memory yet and the asserts that we are bashing the right memory locations fail. (And on release builds, we'd overwrite the values we just bashed when the JIT initializes the write barrier).

This is really unfortunate. The cycle is that the JIT needs to know if the GC intends to use per-thread allocation contexts, in order to figure out what allocation helpers to use, but the GC needs the JIT's write barriers to be initialized by the time it initializes.

swgillespie:
cc @adityamandaleeka who's also doing Local GC stuff that has implications on initialization order

Maoni0:
you can move the decision of using the EE allocation contexts (including the global one for UP and per thread for MP) to the EE side. at this point GC really just provides an API that takes an AC and it's up to the EE to give GC AC/ACs efficiently.

src/gc/env/gcenv.ee.h
```diff
@@ -68,6 +68,12 @@ class GCToEEInterface
     static void StompWriteBarrier(WriteBarrierParameters* args);
 
     static void EnableFinalization(bool foundFinalizers);
+
+    // Global alloc context retrieval
+    static gc_alloc_context *GetGlobalAllocContext();
```
jkotas: Do we need two APIs for this? Can this be just one API - that returns null if the EE is not using global alloc context?

jkotas: If you do that, using_allocation_contexts can become the pointer to the global allocation context instead of a flag - it will make the updates a tiny bit cheaper.

swgillespie: Sounds good, will do!

swgillespie:
The last two changes address CR feedback.

Maoni/Jan, do you have any preferences between using `GetGlobalAllocContext` or enumerating the global alloc context with `GCToEEInterface::GcEnumAllocContexts`? I implemented the former one in the last commit but I also have a branch with Maoni's suggestion as well, though I ran into trouble with `gc_heap::fix_youngest_allocation_arena`, since we still need to know whether or not we are using per-thread alloc contexts (if so, we have to call `fix_youngest_allocation_arena` anyway after `GcEnumAllocContexts`). I'm happy to do either solution if there's a preference

jkotas:
I am fine with GetGlobalAllocContext.

src/gc/gcinterface.ee.h
```diff
@@ -133,6 +133,11 @@ class IGCToCLR {
     // be finalized.
     virtual
     void EnableFinalization(bool foundFinalizers) = 0;
+
+    // Retrieves a pointer to the global allocation context which, on single-proc
+    // systems, is used for all SOH allocations.
```
jkotas: The comment should say that this method can return null and what null means

swgillespie: You're right, will fix.

src/gc/gcpriv.h
```diff
@@ -3604,6 +3604,10 @@ class gc_heap
     PER_HEAP
     size_t total_ephemeral_size;
 
+    // True if we are using allocation contexts, false if we
```
jkotas: Would it make sense to this to be under `#ifndef MULTIPLE_HEAPS` ?

jkotas:
LGTM

src/gc/gc.cpp
```diff
@@ -5715,18 +5711,29 @@ void gc_mechanisms::record (gc_history_global* history)
...
+        acontext = global_allocation_context;
+        assert(acontext != nullptr);
+        dprintf (3, ("EE-owned gen 0 alloc context %Ix: ptr %Ix, limit: %Ix",
+              (size_t)acontext, (size_t)acontext->alloc_ptr, (size_t)acontext->alloc_limit));
     }
-    heap_segment_allocated (ephemeral_heap_segment) =
-        alloc_allocated;
+    else
+    {
+        acontext = generation_alloc_context (youngest_generation);
```
Maoni0: >acontext = generation_alloc_context (youngest_generation); [](start = 8, length = 58)

we no longer use the gen0 alloc context in the allocation path so you don't need to fix it. you should assert that gen0's alloc context's limit and ptr are both 0 here. we only need to call GcEnumAllocContexts.

src/gc/gc.cpp
```diff
@@ -5830,13 +5837,27 @@ void void_allocation (gc_alloc_context* acontext, void*)
 
 void gc_heap::repair_allocation_contexts (BOOL repair_p)
 {
-    GCToEEInterface::GcEnumAllocContexts (repair_p ? repair_allocation : void_allocation, NULL);
+    alloc_context *acontext;
+    if (!using_thread_alloc_contexts())
```
Maoni0: >if (!using_thread_alloc_contexts()) [](start = 4, length = 35)

again, you don't need to check here - EnumAllocContexts will enum either the per thread context or the UP context. You do need this line from fix_youngest:

heap_segment_allocated (ephemeral_heap_segment) = alloc_allocated;

src/gc/gcpriv.h
swgillespie: It's true that this field is always null with Server GC but I was hoping to keep `MULTIPLE_HEAPS` defines to a minimum in the hopes of defragmenting `gc.cpp` a little bit.

Maoni0:
we don't need to know whether or  not we are using per thread context with the  GCToEEInterface::GcEnumAllocContexts change, that was the point. GcEnumAllocContexts  would know whether it's using per thread or UP context; GC doesn't need to. And GC no longer uses the gen0 context outside of GC so no need to fix it.

jkotas:
> we don't need to know whether or not we are using per thread context with the GCToEEInterface::GcEnumAllocContexts change, that was the point. 

That sounds better to me - assuming that the code for it looks nice too.

swgillespie:
ok, will do! I'll cherry-pick the branch with those changes + Maoni's feedback onto this one.

src/gc/gc.cpp
swgillespie: Ok!

src/gc/gc.cpp
swgillespie: Ok!

swgillespie:
@Maoni0 done!

src/gc/gc.cpp
```diff
@@ -5715,18 +5709,12 @@ void gc_mechanisms::record (gc_history_global* history)
...
-    if (for_gc_p)
-    {
-        acontext->alloc_ptr = alloc_allocated;
-        acontext->alloc_limit = acontext->alloc_ptr;
-    }
-    heap_segment_allocated (ephemeral_heap_segment) =
-        alloc_allocated;
+    UNREFERENCED_PARAMETER(for_gc_p);
+
+    // The gen 0 alloc context is never used for allocation.
```
Maoni0: >used for allocation. [](start = 40, length = 20)

I would say never used for allocation in the allocator path. it's still used in the allocation path happens during GCs.

src/gc/gccommon.cpp
```diff
@@ -62,7 +62,7 @@ uint8_t g_build_variant = 1;
 uint8_t g_build_variant = 2;
 #endif // defined(BUILDENV_DEBUG)
 
-VOLATILE(int32_t) m_GCLock = -1;
+VOLATILE(int32_t) g_no_gc_lock = -1;
```
Maoni0: >g_no_gc_lock [](start = 18, length = 12)

this shouldn't be in gccommon.cpp - it's only used internally by GC so should be in gc.cpp.

src/gc/gcimpl.h
```diff
@@ -91,14 +91,13 @@ class GCHeap : public IGCHeapInternal
...
     //flags can be GC_ALLOC_CONTAINS_REF GC_ALLOC_FINALIZE
-    Object*  Alloc (size_t size, uint32_t flags);
-    Object*  AllocAlign8 (size_t size, uint32_t flags);
     Object*  AllocAlign8 (gc_alloc_context* acontext, size_t size, uint32_t flags);
 private:
     Object*  AllocAlign8Common (void* hp, alloc_context* acontext, size_t size, uint32_t flags);
 public:
     Object*  AllocLHeap (size_t size, uint32_t flags);
     Object* Alloc (gc_alloc_context* acontext, size_t size, uint32_t flags);
+    bool UseThreadAllocationContexts();
```
Maoni0: this is no longer used.

src/gc/gccommon.cpp
swgillespie: ok!

src/gc/gcimpl.h
swgillespie: will delete.

src/gc/gc.cpp

swgillespie:
@Maoni0 done! thanks!

src/gc/gc.h
```diff
@@ -108,6 +108,7 @@ extern "C" uint8_t* g_gc_highest_address;
 extern "C" bool g_fFinalizerRunOnShutDown;
 extern "C" bool g_built_with_svr_gc;
 extern "C" uint8_t g_build_variant;
+extern "C" VOLATILE(int32_t) g_no_gc_lock;
```
Maoni0: >g_no_gc_lock [](start = 29, length = 12)

this should either just be in gc.cpp or in gcimpl.h. it's an pure internal implementation.

swgillespie: All of these variables are internal implementations, would you prefer is they all lived in one of those two files too?

Maoni0: yep. gc.h should only contain interface related stuff, not implementation only stuff.

---
In reply to: [105787160](https://github.com/dotnet/coreclr/pull/10065#discussion_r105787160) [](ancestors = 105787160)

swgillespie: Moving them to those two files will cause these variables to get duplicated for each of server GC and workstation GC - is that ok?

Maoni0: I am ok with that - I don't have a strong opinion on that though.

swgillespie: I don't either, the duplication just means I have to bludgeon the linker a little more to do the right thing 😆

swgillespie:
Doing a stress run tonight - please don't merge yet

swgillespie:
@Maoni0 (and @sergiy-k, who might also be interested)

As a status update: I did a stress run last night and it ran into trouble. I’m really perplexed because everything is completely hosed and nothing works – even the commits that I definitely tested and verified were working. I stepped through the allocation path with a debugger to verify it at the time and I completed several full test runs without incident, but now the first allocation immediately deadlocks in the allocation path.

The trouble is in `JIT_TrialAllocSFastSP`, where we attempt to acquire the alloc lock:

```asm
        inc     [g_global_alloc_lock + OFFSETOF__GlobalAllocLock__m_lock]
        jnz     JIT_NEW
```

For reasons I don’t understand fully, this is the sequence of events:
1.	`inc` increments the value of `g_global_alloc_lock.m_lock` from -1 to 0
2.	`jnz` jumps to `JIT_New` anyway (?????)
3.	`JIT_New` tries to acquire the lock again, but it deadlocks because it already has the lock and the alloc lock doesn't allow recursive acquires

After the INC is executed, EFLAGS has only the parity, adjust, and direction flags set (the value is 0x216 according to windbg). This is not what I'd expect at all, out of `inc`, it seems to me that the zero flag should be set. I doubt my processor broke overnight so it seems to be more likely that something is wrong with what the debugger is reporting to me as to the value of `g_global_alloc_lock`.

One major variable that has changed since I started this PR and now is that I upgraded my toolchain to VS2017. Today I will be troubleshooting this (probably also doing a build with VS2015 and diffing the results) and I hope to have it resolved by the end of the day today.

swgillespie:
It turns out that the deadlock is due to `g_global_alloc_lock.m_lock` being 32 bits, while using a `inc qword ptr` instruction to acquire it in assembly helpers. This may or may not work depending on whether or not the linker places something adjacent to `g_global_alloc_lock`, which explains why this worked with VS2015's linker and not with VS2017's.

I'll be restarting the stress run and reporting back with the results.

src/vm/gcheaputilities.h
```diff
@@ -192,11 +195,96 @@ class GCHeapUtilities {
...
+// can occur.
+//
+// This lock is acquired by all allocations when not using per-thread allocation contexts. 
+// It is acquired in two kinds of places:
+//   1) JIT_TrialAllocFastSP (and related assembly alloc helpers), which attempt to 
+//      acquire it but move into an alloc slow path if acquiring fails
+//      (but does not decrement the lock variable when doing so)
+//   2) Alloc and AllocAlign8 in gchelpers.cpp, which acquire the lock using
+//      the Acquire and Release methods below.
+class GlobalAllocLock {
```
Maoni0: >class GlobalAllocLock [](start = 0, length = 21)

This is only used on the EE side now. It shouldn't be in a gc header.

swgillespie: It's not, `gcheaputilities.h` is a EE-only header.

Maoni0: it's on the EE side, yes, and it does stuff with the GC. but GlobalAllocLock is only used by allocators which live in gchelpers.cpp. shouldn't this be then in gchelpers.h?

swgillespie: I tried to keep all of the GC utility functions and structs in one place with `gcheaputilities.h`, but I'd be fine moving it to gchelpers as well. It seems to me that gchelpers is more for allocation helpers - nobody including `gchelpers.h` (and only that file) would need to use `GlobalAllocLock`, while includers of `gcheaputilities.h` would (since they may be calling methods on IGCHeap directly).

swgillespie: Also, regarding this lock - I need to be in preemptive mode when I call `__SwitchToThread`, right? I think that might be wrong in this code here.

Maoni0: The point is this is not a GC utility :) It's only used by allocators which are separate from the GC. I just looked and gchelpers.h really just includes the higher level allocation APIs for EE to use (so it's included in common.h). Since this is purely internal to allocation helpers I would just have the code in gchelpers.cpp since no one uses it except gchelpers.cpp.

swgillespie: ok, will do!

Maoni0: re the lock implementation, yes __SwitchToThread should be taken in preemptive mode.

swgillespie:
Last night's stress run on x64 Windows was clean. I also did a stress run with x86 windows and it didn't print out the same "Reached time limit, exiting: ran X tests out of X" message that the x64 stress run did. I'm running it again under a debugger this morning to make sure there aren't any issues.

swgillespie:
The x86 stress job hits a fatal error due to a failure to expand the mark list: 

```
00 0b13caf0 779aa41e ntdll!NtTerminateProcess+0xc
01 0b13cbcc 74dcadc2 ntdll!RtlExitUserProcess+0x9e
02 0b13cbe0 0fe049bd KERNEL32!ExitProcessImplementation+0x12
03 0b13cbec 0fe07229 CoreCLR!EEPolicy::ExitProcessViaShim+0xd [e:\github\coreclr\staging\src\vm\eepolicy.cpp @ 545]
04 0b13cd2c 0fe053ac CoreCLR!SafeExitProcess+0x319 [e:\github\coreclr\staging\src\vm\eepolicy.cpp @ 522]
05 0b13d09c 104ec7ef CoreCLR!EEPolicy::HandleFatalError+0x16c [e:\github\coreclr\staging\src\vm\eepolicy.cpp @ 1424]
06 0b13d108 1051b96f CoreCLR!WKS::gc_heap::enque_pinned_plug+0x3f [e:\github\coreclr\staging\src\gc\gc.cpp @ 17489]
07 0b13d124 10508c22 CoreCLR!WKS::gc_heap::store_plug_gap_info+0x25f [e:\github\coreclr\staging\src\gc\gc.cpp @ 21446]
08 0b13d518 104f285f CoreCLR!WKS::gc_heap::plan_phase+0x18f2 [e:\github\coreclr\staging\src\gc\gc.cpp @ 21876]
09 0b13d598 104f268a CoreCLR!WKS::gc_heap::gc1+0x17f [e:\github\coreclr\staging\src\gc\gc.cpp @ 15382]
0a 0b13d5c8 104c573e CoreCLR!WKS::gc_heap::garbage_collect+0x84a [e:\github\coreclr\staging\src\gc\gc.cpp @ 16946]
0b 0b13d600 1051db86 CoreCLR!WKS::GCHeap::GarbageCollectGeneration+0x42e [e:\github\coreclr\staging\src\gc\gc.cpp @ 35247]
0c 0b13d630 104d1637 CoreCLR!WKS::gc_heap::try_allocate_more_space+0x306 [e:\github\coreclr\staging\src\gc\gc.cpp @ 13186]
0d 0b13d64c 104ccc93 CoreCLR!WKS::gc_heap::allocate_more_space+0x17 [e:\github\coreclr\staging\src\gc\gc.cpp @ 13473]
0e 0b13d66c 104c3a5c CoreCLR!WKS::gc_heap::allocate+0xa3 [e:\github\coreclr\staging\src\gc\gc.cpp @ 13504]
0f 0b13d718 10096d4f CoreCLR!WKS::GCHeap::Alloc+0x17c [e:\github\coreclr\staging\src\gc\gc.cpp @ 34567]
```

This is a known issue and isn't related to this PR. Other than that, the stress run looks good on x86.

swgillespie:
Are there any objections towards merging this? I think it's ready to go.

Maoni0:
I approved but somehow it's not showing up....

swgillespie:
Oh, that's weird - thanks!

swgillespie:
thanks!

PR #10067, opened 2017-03-09
Fix AssertionDsc::Equals OAK_NO_THROW's handling

### LINKED ISSUE #9936: JIT: AssertionDsc::Equals doesn't handle OAK_NO_THROW properly
Redundant assertions are comparing as not-equal, wasting assertion index slots.  See [#9773 (comment)](https://github.com/dotnet/coreclr/pull/9773#discussion_r103097827) and [#9773 (comment)](https://github.com/dotnet/coreclr/pull/9773#issuecomment-283876281).

mikedn:
`op2` of a `OAK_NO_THROW` assertions is `O2K_INVALID` and `HasSameOp2` always returns false for `O2K_INVALID`. As a result `OAK_NO_THROW` assertions always compare unequal and duplicate assertions are generated that waste space in the assertion table.

Also, `HasSameOp1` only checks `op1.vn` but that is bogus for `OAK_NO_THROW` assertions, most of the time is 0.

Fixes #9936

### DIFF
--- a/src/jit/compiler.h
+++ b/src/jit/compiler.h
@@ -5826,8 +5826,20 @@ public:
 
         bool HasSameOp1(AssertionDsc* that, bool vnBased)
         {
-            return (op1.kind == that->op1.kind) &&
-                   ((vnBased && (op1.vn == that->op1.vn)) || (!vnBased && (op1.lcl.lclNum == that->op1.lcl.lclNum)));
+            if (op1.kind != that->op1.kind)
+            {
+                return false;
+            }
+            else if (op1.kind == O1K_ARR_BND)
+            {
+                assert(vnBased);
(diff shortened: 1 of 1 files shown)


mikedn:
jit-diff summary:
```
Total bytes of diff: -63 (0.00 % of base)
    diff is an improvement.
         -63 : System.Private.CoreLib.dasm (0.00 % of base)
         -52 : System.Private.CoreLib.dasm - System.DefaultBinder:BindToMethod(int,ref,byref,ref,ref,ref,byref):ref:this
         -11 : System.Private.CoreLib.dasm - System.Text.UnicodeEncoding:GetChars(long,int,long,int,ref):int:this
2 total methods with size differences (2 improved, 0 regressed).
```

In those 2 affected methods a couple of range checks are now eliminated, previously they weren't because the assertion table filled up.

pgavlin:
LGTM.

@JosephTremoulet @briansull can one of you take a look as well?

JosephTremoulet:
Yep, looks good, thanks.

Issue #10022, opened 2017-03-08
JIT/Methodical/Boxing/misc/concurgc.il failing in CI with a timeout


adityamandaleeka:
Hit this on a PR I opened, and saw that it also happened on another recent job. Here are the two instances:

https://ci.dot.net/job/dotnet_coreclr/job/master/job/x86_checked_windows_nt_prtest/3373/
https://ci.dot.net/job/dotnet_coreclr/job/master/job/x86_checked_windows_nt_prtest/3369/

```
cmdLine:D:\j\workspace\x86_checked_w---b3a226f6\bin\tests\Windows_NT.x86.Checked\JIT\Methodical\Boxing\misc\_relconcurgc_il\_relconcurgc_il.cmd Timed Out
18:42:02         Test Harness Exitcode is : -100
```

gkhanna79:
CC @RussKeldorph

RussKeldorph:
I've seen this time out in several x86 jobs but I can't repro locally.  Assuming it is a test issue, but worth at least cursory investigation.

pgavlin:
Looking into it now.

pgavlin:
I can't repro this either, but I did notice that the test is not using volatile reads and writes where it should be. #10072 fixes those issues; once it's merged I plan to close this bug until it hits again.

mikedn:
Another one: https://ci.dot.net/job/dotnet_coreclr/job/master/job/x86_checked_windows_nt_prtest/3883/

pgavlin:
For posterity, the crash dump for that failed build is located here: https://dumpling.azurewebsites.net/api/dumplings/archived/1d2e8c8f93185b2c93a36929f499d2d307595470

Unfortunately the test job did not upload artifacts so we do not have symbols for the dump.

However, I have looked at another dump from a different instance of this failure, and here is what seems to be happening:
- Thread A instantiates a `Thread` object and calls `Thread.Start`, which kicks off thread B
- Thread B immediately invokes the GC using `GC.Collect`
- Thread A ends up [blocked on `GCHeap:: WaitForGCEvent` waiting for the GC to complete](https://github.com/dotnet/coreclr/blob/master/src/gc/gcee.cpp#L389-L428) with `Thread.Start` on the stack
- Thread B completes its GC and immediately begins another, and so on

[...]

PR #10073, opened 2017-03-09
Replace HNDTYPE defines with enum

adityamandaleeka:
As part of the work to move handle-related code to the GC interface and stop VM code from using things objecthandle.h directly, this change turns the #define'd `HNDTYPE_*` values into an enum class `HandleType` that's in gcinterface.h.

In the future, I'd like all the code that touches these types to use the new type, but there is so much code that uses integral types to deal with these values that it isn't practical to change everything right now. I've changed some functions (where practical) to use the new type, and for the rest I've added casts liberally to keep the existing code working.


jkotas:
This is just a nice to have cleanup, not strictly required for the standalone GC work, correct? If it is the case, I would like to see it done completely so that it is net improvement. I do not think it is a net improvement right now because of the casts that are everywhere.

adityamandaleeka:
@jkotas I'm actually removing all the casts and making it a plain enum instead of enum class. We can do the work to make it enum class and use that type everywhere later.

jkotas:
Also, I am not sure whether it is a good idea to make this `enum class HandleType`...

adityamandaleeka:
@jkotas I just pushed a commit undoing all the `enum class` stuff which necessitated all the casts. It's much simpler. Please let me know what you think.

jkotas:
Looks much better to me know.

adityamandaleeka:
The JIT test failure is a known issue: https://github.com/dotnet/coreclr/issues/10022

Issue #9730, opened 2017-02-23
Debugging is broken on macOS Sierra (10.12.4 Beta)


gregg-miskelly:
1. Install latest beta version of MacOS
2. Install .NET Core SDK, VS Code, and the C# extension, and try to debug

Result:
RegisterForRuntimeStartup will send failures back to the debugger call back with HRESULT E_INVALIDARG. The dbgshim logging isn't very helpful, it just has:

    {0x41e81-0x10b2f85f0} TRACE [PROCESS] at InvokeStartupCallback.1716: CreateProcessModules failed for pid 29295

gregg-miskelly:
It looks like the issue is that the output from vmmap changed, so the sscanf line fails to parse it. Example line that it fails to parse:

    __TEXT                 00000001022aa000-00000001025d1000 [ 3228K  2652K     0K     0K] r-x/rwx SM=COW          /usr/local/share/dotnet/shared/Microsoft.NETCore.App/1.1.0/libcoreclr.dylib

The vmmap command executed is `/usr/bin/vmmap -interleaved 30184 -wide` (where 30184 is the process id of the target process.

gregg-miskelly:
CC @lt72 @noahfalk @mikem8361

mikem8361:
Fixed PR #10078

PR #10078, opened 2017-03-10
Fix debugging on OS X Sierra

mikem8361:
Issue #9730 

The output of the vmmap changed between OS X versions.  Changed the sscanf formatting to deal with both old/new formats.


cxubrix:
After update to 10.12.4 and installation of 2.0.0-preview1-005670 the issue is still there and debuger fails to connect from vscode.

`Unable to start debugging. Unexpected CLRDBG output from command "-target-attach 2353"`

mikem8361:
Here is a work around that worked for Gregg Miskelly:
1.	Download https://dotnet.myget.org/F/dotnet-core/api/v2/package/runtime.osx.10.10-x64.Microsoft.NETCore.Runtime.CoreCLR/1.1.2-servicing-25123-01
2.	Open the resulting file as a zip, and copy out runtimes/osx.10.10-x64/native/libdbgshim.dylib
3.	Open a terminal
4.	cd ~/.vscode/extensions/ms-vscode.csharp-1.8.0/.debugger
5.	cp <path-to-extracted-libdbgshim> .
As I said above, but just to repeat here, I expect to have a real fix out soon. But hopefully this will unblock folks in the mean time.

PR #10083, opened 2017-03-10
[Ryujit/ARM32] Fix wrong comparion

hqueue:
Fix wrong comparision in previous commit.

(This is a part of #8496)


mikedn:
@BruceForstall PTAL

PR #10087, opened 2017-03-10
[x86/Linux][GDB-JIT] Fix crash after changing default calling convention

lucenticus:
This PR fixes crash of GDB JIT interface after changing default calling convention to cdecl in PR #9977
@janvorli @jkotas PTAL
\CC: @parjong


lucenticus:
@jkotas Thank you!

Issue #10091, opened 2017-03-10
runtest.cmd broken when VS2015 installed without VS2017


RussKeldorph:
Error on [this line]( https://github.com/dotnet/coreclr/blob/0ce065c5f28f840f02c9621a45bb9e3daed87e1e/tests/runtest.cmd#L123)

I believe the fix is to use delayed expansion.

swgillespie:
`runtest.cmd` is also broken when run from a VS2015 command prompt, even with VS2017 installed:

```
E:\GitHub\coreclr\staging>tests\runtest.cmd x86 Checked
Visual Studio 2015 or 2017 (Community is free) is a prerequisite to build this repository.
See: https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/developer-guide.md#prerequisites

E:\GitHub\coreclr\staging>tests\runtest.cmd x86 Checked vs2015
'_msbuildexe' is not recognized as an internal or external command,
operable program or batch file.
Visual Studio 2015 or 2017 (Community is free) is a prerequisite to build this repository.
See: https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/developer-guide.md#prerequisites
```

tannergooding:
Yes, that would be an appropriate fix here. I will have a fix out shortly.

RussKeldorph:
Correcting misleading title.  Thanks @swgillespie for knowing what I meant not what I wrote.

PR #10092, opened 2017-03-10
Updating tests/runtest.cmd to use delayed expansion for instances of `_msbuildexe`

tannergooding:
Resolves https://github.com/dotnet/coreclr/issues/10091

### DIFF
--- a/tests/runtest.cmd
+++ b/tests/runtest.cmd
@@ -120,7 +120,7 @@ if /i "%__VSVersion%" == "vs2017" (
   set "__VCToolsRoot=%VS140COMNTOOLS%\..\..\VC"
 
   set _msbuildexe="%ProgramFiles(x86)%\MSBuild\14.0\Bin\MSBuild.exe"
-  if not exist %_msbuildexe% set _msbuildexe="%ProgramFiles%\MSBuild\14.0\Bin\MSBuild.exe"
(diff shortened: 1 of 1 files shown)


tannergooding:
@RussKeldorph, I've validated that this fixes the issue on a VS2015 only machine and on a VS2015/VS2017 SxS machine.

tests/runtest.cmd
```diff
@@ -129,7 +129,7 @@ if not exist "%__VCToolsRoot%\vcvarsall.bat"          goto NoVS
 if not exist "%__VSToolsRoot%\VsDevCmd.bat"           goto NoVS
 
 :: Does MSBuild really exist?
-if not exist %_msbuildexe% echo Error: Could not find MSBuild.exe.  Please see https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/developer-guide.md for build instructions. && exit /b 1
+if not exist !_msbuildexe! echo Error: Could not find MSBuild.exe.  Please see https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/developer-guide.md for build instructions. && exit /b 1
```
RussKeldorph: I consider use of delayed expansion confusing except where absolutely required, so I probably wouldn't make any changes except the first, but that's personal preference.  Your call.

tannergooding: I'll remove them since they aren't strictly necessary (if statements are parsed in one go, rather than line-by-line, so it is needed there, but not here).

Issue #6809, opened 2016-08-19
Problems with TryStartNoGCRegion 


mattwarren:
As part of writing a blog post (["Preventing .NET Garbage Collections with the TryStartNoGCRegion API"](http://mattwarren.org/2016/08/16/Preventing-dotNET-Garbage-Collections-with-the-TryStartNoGCRegion-API/)), I came across some strange behaviour when using `TryStartNoGCRegion`.

With test code like this:

``` csharp
try
{
    Console.WriteLine("\nCalling TryStartNoGCRegion(..) with totalSize = {0:N0} MB ({1:N0} K or {2:N0} bytes)",
                        size / 1024.0 / 1024.0, size / 1024.0, size);
    var succeeded = GC.TryStartNoGCRegion(size);
    Console.WriteLine("{0:N0} MB ({1:N0} K or {2:N0} bytes) {3}",
                        size / 1024.0 / 1024.0, size / 1024.0, size, succeeded ? "SUCCEEDED" : "FAILED");
}
catch (Exception ex)
{
    Console.WriteLine("{0:N0} MB ({1:N0} K or {2:N0} bytes) - {3} {4}",
                      size / 1024.0 / 1024.0, size / 1024.0, size, ex.GetType().Name, ex.Message);
}
```

I get the following results:

| totalSize | Result |
| --: | :-- |
| 240 \* 1024 \* 1024 | Returns true |
| 242 \* 1024 \* 1024 | Hangs |
| 244 \* 1024 \* 1024 | Hangs |
| 246 \* 1024 \* 1024 | Hangs |
| 248 \* 1024 \* 1024 | Hangs |
| 250 \* 1024 \* 1024 | Hangs |
| 252 \* 1024 \* 1024 | Hangs |
| 254 \* 1024 \* 1024 | Hangs |
| 256 \* 1024 \* 1024 | Throws ArgumentOutOfRangeException("totalSize is too large") |

It seems that there are some values of `totalSize` that cause `TryStartNoGCRegion` to never return? On some runs, the values that hang will instead return `false`, but more often than not it just hangs?

**Note:** There's tests were all done on .NET 4.6,1, 32-bit, Server GC (`<gcServer enabled="true"/>`).

swgillespie:
Thanks for the report! cc @Maoni0 

I have a repro, here are the interesting threads:

```
   0  Id: 45d8.47b8 Suspend: 1 Teb: 006e7000 Unfrozen
 # ChildEBP RetAddr  
[...]

PR #10096, opened 2017-03-10
Porting desktop GC changes to coreclr. This fixes the following:

Maoni0:
1) https://github.com/dotnet/coreclr/issues/6809
2) when we do a minimal GC, we need to maintain the states correctly (sync up tables and clear bricks for the portion that we don't need in a normal GC but do need in minimal GC)


Maoni0:
@swgillespie

swgillespie:
👍

Issue #10097, opened 2017-03-10
We have duplicate IL for System.Private.Corelib.dll


gkhanna79:
@Petermarcu commented on [Tue Feb 07 2017](https://github.com/dotnet/core-setup/issues/1484)

## Steps to reproduce
Look in the tar.gz's

## Expected  behavior
No System.Private.Corelib.dll next to System.Private.Corelib.ni.dll

## Actual behavior
They are both there.

## Environment data

---

@ericstj commented on [Tue Feb 14 2017](https://github.com/dotnet/core-setup/issues/1484#issuecomment-279846310)

@gkhanna79 feel free to fix the duplicate entries in your CoreCLR packages.  EG: https://github.com/dotnet/coreclr/blob/89ba46aa6e113486428ac7528f662e3a0a5a55eb/src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/ubuntu/14.04/Microsoft.NETCore.Runtime.CoreCLR.pkgproj#L21-L22

---

@jkotas commented on [Sun Feb 26 2017](https://github.com/dotnet/core-setup/issues/1484#issuecomment-282580714)

FWIW, this duplication was introduced originally to make nuget guardrails happy: https://github.com/dotnet/coreclr/pull/3043#issuecomment-180515876 . Hopefully, this is not a problem anymore.

---

@ericstj commented on [Mon Feb 27 2017](https://github.com/dotnet/core-setup/issues/1484#issuecomment-282780562)

@jkotas that's never been a problem for System.Private.Corelib.  It was needed to match ref of mscorlib with lib of mscorlib, but never S.P.C since there is no ref exposed for that.  This is purely a bug in the coreclr package.

---

@gkhanna79 commented on [Tue Feb 28 2017](https://github.com/dotnet/core-setup/issues/1484#issuecomment-283148865)

@brianrob @noahfalk Do you rely on having the IL version of SPC.dll to be present next to CoreCLR in SharedFX?

---

@brianrob commented on [Tue Feb 28 2017](https://github.com/dotnet/core-setup/issues/1484#issuecomment-283149249)

@gkhanna79, I don't know of any reason why we need the IL version.  Confirming - if we did need IL for something, it's still in the NI, right?

---

@noahfalk commented on [Tue Feb 28 2017](https://github.com/dotnet/core-setup/issues/1484#issuecomment-283157617)

No dependency that I am aware of from the diagnostics side.

---

[...]

PR #10101, opened 2017-03-10
Remove S.P.C.dll from nuget package

gkhanna79:
@jkotas PTAL

Fixes https://github.com/dotnet/coreclr/issues/10097


gkhanna79:
Confirmed that Pri1 test build is clean.

jkotas:
tests\src\managed\Compilation\Compilation.cs may need updating - it assumes that there is System.Private.CoreLib.dll. Not sure how many other places like that are there.

It may be better to put the corelib native image into the package as System.Private.CoreLib.dll (consistent with what we are doing for R2R native images).

gkhanna79:
I am doing test run to identify issues from this change - hopefully, this will come through.

With that said, if the IL is required by some tests, I am just wondering if the fix is, instead, to exclude S.P.C.dll from sharedFX package, as opposed to excluding it from CoreCLR nuget package. What do you think @jkotas?

jkotas:
The native image contains the IL. For example, you can `/r:<native image>` on the C# compiler command line just fine. I would love to be in a situation where people just do not need to care that there is .ni.dll suffix sometimes. We are almost there. CoreLib is the only exception as far as I know.

Issue #10102, opened 2017-03-10
Port fix for macOS Sierra to 1.1 release


gregg-miskelly:
Install latest beta version of MacOS
Install .NET Core SDK, VS Code, and the C# extension, and try to debug

Result:
CoreCLR debugging fails

Notes:
This was originially reported on the GitHub for the C# VS Code extension (https://github.com/OmniSharp/omnisharp-vscode/issues/1220). It has been the most actively reported issue we have had thus far with ~15 folks adding a "me too", and I am sure many more folks who ran into the problem, but didn't bother to "me too" it.

PR #10103, opened 2017-03-10
Fix debugging on OS X Sierra

mikem8361:
Issue #10102

The output of the vmmap changed between OS X versions. Changed the sscanf formatting to deal with both old/new formats.


mikem8361:
@noahfalk, @janvorli review? It is a very simple change, but the ship room wants two reviewers.

noahfalk:
Looks fine to me. (Of course the entire thing still looks really fragile, but I recognize fixing that is a bigger issue)

PR #10104, opened 2017-03-11
Fix overflow when computing dwWeight in CPUGroupInfo::InitCPUGroupInfoArray

hanblee:
`dwWeight` is computed as nr_active<sup>m_nGroups</sup> which can overflow as the number of CPU groups increase.  This fix uses least common multiple of nr_active of different CPU groups to  compute `dwWeight` instead.
@Maoni0 PTAL - this should address `DivideByZeroException` issue on machines with large number of cores, but I don't have access to such machine to test at the moment.

### DIFF
--- a/src/utilcode/util.cpp
+++ b/src/utilcode/util.cpp
@@ -964,8 +964,10 @@ DWORD LCM(DWORD u, DWORD v)
         dwWeight = LCM(dwWeight, (DWORD)m_CPUGroupInfoArray[i].nr_active);
     }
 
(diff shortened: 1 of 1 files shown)


Maoni0:
thanks @hanblee. could you please also update the comment below this to reflect your change? it still says it can only address 4*64 CPUs.

hanblee:
Thanks for your feedback @Maoni0. I updated the comment to reflect the change.

hanblee:

hanblee:

PR #10105, opened 2017-03-11
Rename Type.IsSzArray to Type.IsSZArray and make public.

JonHanna:
For dotnet/corefx#16566

### DIFF
--- a/src/mscorlib/src/System/RuntimeHandles.cs
+++ b/src/mscorlib/src/System/RuntimeHandles.cs
@@ -153,7 +153,7 @@ namespace System
             return (corElemType == CorElementType.Array || corElemType == CorElementType.SzArray);
         }
 
-        internal static bool IsSzArray(RuntimeType type)
+        internal static bool IsSZArray(RuntimeType type)
         {
             CorElementType corElemType = GetCorElementType(type);
             return (corElemType == CorElementType.SzArray);
--- a/src/mscorlib/src/System/Type.cs
+++ b/src/mscorlib/src/System/Type.cs
@@ -1182,10 +1182,10 @@ namespace System
             get { return IsArrayImpl(); }
         }
 
-        internal virtual bool IsSzArray
+        public virtual bool IsSZArray
         {
             [Pure]
-            get { return false; }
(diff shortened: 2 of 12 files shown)


src/mscorlib/src/System/RuntimeHandles.cs
```diff
@@ -153,7 +153,7 @@ internal static bool IsArray(RuntimeType type)
             return (corElemType == CorElementType.Array || corElemType == CorElementType.SzArray);
         }
 
-        internal static bool IsSzArray(RuntimeType type)
+        internal static bool IsSZArray(RuntimeType type)
```
JonHanna: This rename isn't necessary, but the consistency seemed worthwhile.

src/mscorlib/src/System/Type.cs
```diff
@@ -1182,7 +1182,7 @@ public bool IsArray
             get { return IsArrayImpl(); }
         }
 
-        internal virtual bool IsSzArray
+        public virtual bool IsSZArray
         {
             [Pure]
             get { return false; }
```
AtsushiKan: @JonHanna  The default implemention now needs to be "throw new NotImplementedException();"

AtsushiKan:
Because the default implementation of IsSZArray is now to throw NYI, any corelib types that derive from Type may need to override this now to preserve the existing behavior. System.Reflection.Emit.TypeBuilder comes to mind.

src/mscorlib/src/System/Type.cs
JonHanna: Why can't we just keep the default false return?

AtsushiKan: The API was approved based on the "throw NotImplemented" behavior. If a third party type fails to override this, we want to get a clear exception to that effect, not an assumed answer that will be wrong 10% of the time.

JonHanna:
Test Windows_NT x64 Release Priority 1 Build and Test

JonHanna:
Test Windows_NT x64 Release Priority 1 Build and Test

Issue #9698, opened 2017-02-21
Investigate using JIT_Memset over Memset in Span.Clear


ahsonkhan:
From PR #9598 

https://github.com/dotnet/coreclr/pull/9598#discussion_r101274665

> Can `JIT_MemSet` be used and does it offer any improvement? (Not sure what path it gets used on) https://github.com/dotnet/coreclr/pull/7198

> Worth trying. I would call JIT_MemSet from the implementation of the PInvoke if it is available and faster. I would not call it directly from here. The condition for when it is available and faster potentially involves architectures, OS versions, and compiler/C-runtime versions. It is better for such condition to be on C++ side.

PR #10110, opened 2017-03-11
Optimize Span.Clear() on AMD64

helloguo:
This PR optimizes `Span.Clear()` on AMD64 by using `initblk `IL to call `JIT_MemSet`. Before 512 bytes, the baseline uses C# manged implementation. And after 512 Byte, it eventually calls CRT `memset`, which uses `rep stosb` for this copy bucket. The prototype calls `JIT_MemSet`, which has better implementation for short length up to 512 bytes and same `rep stosb` after 512 bytes. Performance measurements were made on a Skylake Desktop - Intel(R) Core(TM)i7-6700K CPU @ 4.00 GHz 32 GB RAM with Windows 10 Enterprise. 

One micro benchmark as shown below is used to test performance gain. The performance gain (from 22% to 58%) is shown in below table.

            begin = DateTime.Now.Ticks;
            for (uint i = 0; i < ITERATION; i++)
            {
                byteSpan.Clear();
            }
            time = DateTime.Now.Ticks - begin;

image

Fix https://github.com/dotnet/coreclr/issues/9698

### DIFF
--- a/src/mscorlib/src/System/Span.cs
+++ b/src/mscorlib/src/System/Span.cs
@@ -614,7 +614,7 @@ namespace System
                 return;
             
 #if AMD64
-            if (byteLength > uint.MaxValue) goto PInvoke;
+            if (byteLength > 4096) goto PInvoke;
             Unsafe.InitBlockUnaligned(ref b, 0, (uint)byteLength);
             return;
 #endif // AMD64
--- a/src/mscorlib/src/System/Span.cs
+++ b/src/mscorlib/src/System/Span.cs
@@ -612,7 +612,13 @@ namespace System
         {
             if (byteLength == 0)
                 return;
-
+            
+#if AMD64
+            if (byteLength > uint.MaxValue) goto PInvoke;
+            Unsafe.InitBlockUnaligned(ref b, 0, (uint)byteLength);
+            return;
+#endif // AMD64
+            // TODO: Optimize this method on X86 machine
             // Note: It's important that this switch handles lengths at least up to 22.
             // See notes below near the main loop for why.
 


helloguo:
@ahsonkhan PTAL

src/mscorlib/src/System/Span.cs
```diff
@@ -612,7 +612,11 @@ internal static unsafe void ClearWithoutReferences(ref byte b, nuint byteLength)
         {
             if (byteLength == 0)
                 return;
-
+            
+#if AMD64
+            Unsafe.InitBlockUnaligned(ref b, 0, (uint)byteLength);
```
jkotas: It is not ok to just cast byteLength to uint. It can be more than 2GB.

jkotas: One way to fix it would be to call: InitBlockUnaligned for smaller sizes (say, several kB) and PInvoke for sizes over that.

ahsonkhan: Why not just return after the call and remove the #else condition?

ahsonkhan:
What does the performance look like for Span size > 512 bytes on AMD64?

helloguo:
@jkotas Thanks for your comments. Add copy length checking before calling `InitBlockUnaligned`.

helloguo:
@ahsonkhan Because x86 does not have JIT_MemSet, we do x64 first. We may create an issue to follow up x86 optimization. When span size is larger than 512 bytes, for 1000 bytes length, the prototype/baseline is 0.72. For 2000 bytes length, the prototype/baseline is 0.75.

src/mscorlib/src/System/Span.cs
```diff
@@ -612,7 +612,13 @@ internal static unsafe void ClearWithoutReferences(ref byte b, nuint byteLength)
         {
             if (byteLength == 0)
                 return;
-
+            
+#if AMD64
+            if (byteLength > uint.MaxValue) goto PInvoke;
```
jkotas: Can this be something smaller than uint.MaxValue, like several kB - similar to what is done for memcpy?

The implementation of InitBlockUnaligned is not friendly to the GC. It can lead to long GC pauses for large sizes. The PInvoke path does not have this problem.

helloguo: Good point. So for 4000 bytes length, we see 17% improvement. For 8000 bytes length, we see 7% improvement. Considering GC, how about setting threshold as 4096 bytes?

jkotas: Sounds good to me.

PR #10114, opened 2017-03-11
Build Linux/x86 JIT as an altjit running on Windows

BruceForstall:
This allows generating the same code that the Linux/x86
JIT does without requiring setting up a Linux/x86 environment.

Ideally, we would also do this for Linux/AMD64, but that
requires implementing or enabling a Windows version of the
JIT-EE interface getSystemVAmd64PassStructInRegisterDescriptor()
API.


BruceForstall:
@parjong @pgavlin PTAL
cc @dotnet/jit-contrib

BruceForstall:
Subsumed by #10120

PR #10118, opened 2017-03-11
Fix delayed expansion scripting error in runtest.cmd

BruceForstall:
@dotnet/jit-contrib Can someone take a quick look at this trivial script fix?


pgavlin:
LGTM

PR #10120, opened 2017-03-11
Build Linux altjit for x86 and amd64

BruceForstall:
With this change, we build two new JITs: (1) a JIT that runs on
Windows amd64 and targets Linux amd64, and (2) a JIT that runs
on Windows x86 and targets Linux x86. Each is built as an altjit named
linuxnonjit.dll. This is useful for debugging, or generating asm code or diffs.
You can even easily create Windows/non-Windows asm diffs
(either to compare the asm, or compare the generated code size). In both
cases, you don't need to set up a Linux environment (especially the
currently immature Linux x86 environment) to do many aspects of JIT
development.

For this to work for amd64, the JIT-EE interface method
getSystemVAmd64PassStructInRegisterDescriptor() was changed
to always be built in, by defining `FEATURE_UNIX_AMD64_STRUCT_PASSING_ITF`
in all AMD64 builds. The `_ITF` suffix indicates that this is
functionality specific to implementing the JIT-EE interface
contract. There were many places in the VM that used this
interchangeably with `FEATURE_UNIX_AMD64_STRUCT_PASSING`. Now,
`FEATURE_UNIX_AMD64_STRUCT_PASSING` means code in the VM needed
to implement the amd64 struct passing feature, but not required to implement the
JIT-EE interface contract. In particular, MethodTables compute
and cache the "eightbyte" info of structs when loading a type.
This is not done when only `FEATURE_UNIX_AMD64_STRUCT_PASSING_ITF`
is set, to avoid altering MethodTable behavior on non-Unix
AMD64 builds. Instead, if `getSystemVAmd64PassStructInRegisterDescriptor()`
is called on a non-Unix build (by the altjit), the `ClassifyEightBytes()`
function is called, and nothing is cached. Hopefully (though it was
hard for me to guarantee by observation), calling `ClassifyEightBytes()`
does not have any side effects on MethodTables. It doesn't really matter,
since if called for altjit, we don't care too much about running.

The previously used `PLATFORM_UNIX` define is now insufficient.
I introduced the `#define` macros `_HOST_UNIX_` to indicate the
JIT being built will run on Unix, and `_TARGET_UNIX_` to indicate
the JIT is generating code targeting Unix. Some things were
converted to use the `UNIX_AMD64_ABI` define, which makes more
sense.


BruceForstall:
@briansull @janvorli PTAL
cc @dotnet/jit-contrib

BruceForstall:
@briansull @janvorli @dotnet/jit-contrib ping

src/jit/CMakeLists.txt
```diff
@@ -225,6 +222,10 @@ if ((CLR_CMAKE_PLATFORM_ARCH_I386 OR CLR_CMAKE_PLATFORM_ARCH_AMD64) AND WIN32)
     # On x86, build RyuJIT/ARM32 cross-compiling altjit.
     # On amd64, build RyuJIT/ARM64 cross-compiling altjit.
     add_subdirectory(protononjit)
+
+    # Build Linux/x86-running-on-Windows altjit or
+    # Linux/AMD64-running-on-Windows altjit..
+    add_subdirectory(linuxnonjit)
```
pgavlin: Would you mind adding flags to the build that we can set to skip the `linuxnonjit` and `protononjit` builds? Also, perhaps `linuxcrossjit` and `protocrossjit` are better names for these project.

src/jit/compiler.cpp
```diff
@@ -4990,11 +4990,11 @@ int Compiler::compCompile(CORINFO_METHOD_HANDLE methodHnd,
     // with an ARM-targeting "altjit").
     info.compMatchedVM = IMAGE_FILE_MACHINE_TARGET == info.compCompHnd->getExpectedTargetArchitecture();
 
-#if defined(ALT_JIT) && defined(UNIX_AMD64_ABI)
-    // ToDo: This code is to allow us to run UNIX codegen on Windows for now. Remove when appropriate.
-    // Make sure that the generated UNIX altjit code is skipped on Windows. The static jit codegen is used to run.
+#if (defined(_TARGET_UNIX_) && !defined(_HOST_UNIX_)) || (!defined(_TARGET_UNIX_) && defined(_HOST_UNIX_))
+    // The host and target platforms don't match. This info isn't handled by the existing
+    // getExpectedTargetArchitecture() JIT-EE interface method.
```
pgavlin: Can we use `CORINFO_EE_INFO::osType` to detect this at runtime?

pgavlin: (not that it's necessary to do this at runtime yet--I'm just curious)

src/jit/CMakeLists.txt
BruceForstall: What would the benefit of this be? Just so we build fewer things? Who would use it? Developers or CI?

Personally, I don't like building extra stuff. However, it's also gives you a build check that you probably aren't breaking other platform builds.

src/jit/compiler.cpp
BruceForstall: Looks like that might be possible. Noted for future possible improvement.

src/jit/CMakeLists.txt
pgavlin: > What would the benefit of this be? Just so we build fewer things? Who would use it? Developers or CI?

Me. :) I'd love to avoid building the cross-targeting JITs on Windows since I regularly work and build on most of our target platforms and can get the coverage I miss from CI. The Windows build already feels a mite sluggish compared to other platforms; adding another JIT that I'm not likely to use just makes this worse.

> Personally, I don't like building extra stuff. However, it's also gives you a build check that you probably aren't breaking other platform builds.

I don't mind depending on CI for this. It's definitely nice to have the abilities that you're adding, though, so they should be enabled by default.

mikedn: > Would you mind adding flags to the build that we can set to skip the linuxnonjit and protononjit builds?

Funny, I was thinking of asking the same thing. But then I changed my mind because it doesn't really affect me. Usually I just build the clrjit project in VS and copy the dll to the bin dir. Sure, I still need to do a complete build from time to time but for the most part it's just build clrjit/copy/run test.cmd and repeat.

BruceForstall: btw, I agree about the names not being great. We have a matrix of host arch, host platform, target arch, target platform, and JIT flavor (legacy, RyuJIT) to accommodate.

BruceForstall:
@jkotas care to look at the VM side changes?

src/jit/CMakeLists.txt
JosephTremoulet: I regularly build with `skipconfigure skipmscorlib skiptests skipbuildpackages`... surprised that `skipnonjit` would be a bridge too far

BruceForstall: @JosephTremoulet I don't believe there are any flags currently for building or not building portions of the native build (except JIT32, which is a special case).

PR #10124, opened 2017-03-11
Improve encoding performance

jkotas:
This is low-risk stop gap measure to improve encoding performance because of we may not have enough time to improve the test coverage (https://github.com/dotnet/corefx/issues/16334) and refactor to code to use corefxlab high-performance encoding routines for 2.0.

The change is to avoid passing the key loop control variables by ref to the invalid character fallback routines. Taking address of a variable prevents RuyJIT from enregistering it.

Results: UTF8 decoding 1k of ASCII characters is 1.65x faster, similar for other affected codepaths.

### DIFF
--- a/src/mscorlib/src/System/Text/UTF32Encoding.cs
+++ b/src/mscorlib/src/System/Text/UTF32Encoding.cs
@@ -193,6 +193,8 @@ namespace System.Text
 
             // For fallback we may need a fallback buffer
             EncoderFallbackBuffer fallbackBuffer = null;
+            char* charsForFallback;
+
             if (encoder != null)
             {
                 highSurrogate = encoder.charLeftOver;
@@ -250,7 +252,9 @@ namespace System.Text
                     chars--;
 
                     // Do the fallback
-                    fallbackBuffer.InternalFallback(highSurrogate, ref chars);
+                    charsForFallback = chars;
+                    fallbackBuffer.InternalFallback(highSurrogate, ref charsForFallback);
+                    chars = charsForFallback;
 
                     // We're going to fallback the old high surrogate.
                     highSurrogate = '\0';
@@ -271,7 +275,9 @@ namespace System.Text
                 if (Char.IsLowSurrogate(ch))
                 {
                     // We have a leading low surrogate, do the fallback
-                    fallbackBuffer.InternalFallback(ch, ref chars);
+                    charsForFallback = chars;
+                    fallbackBuffer.InternalFallback(ch, ref charsForFallback);
+                    chars = charsForFallback;
 
                     // Try again with fallback buffer
                     continue;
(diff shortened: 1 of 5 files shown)


src/mscorlib/src/System/Text/UTF32Encoding.cs
```diff
@@ -516,6 +534,7 @@ internal override unsafe int GetCharCount(byte* bytes, int count, DecoderNLS bas
 
             // For fallback we may need a fallback buffer
             DecoderFallbackBuffer fallbackBuffer = null;
+            char* charsForFallback;
```
danmosemsft: Is this used? 

In the log i only see src\System\Text\UTF32Encoding.cs(537,19): warning CS0168: The variable 'charsForFallback' is declared but never used [D:\j\workspace\x64_release_w---0575cb46\src\mscorlib\System.Private.CoreLib.csproj]

src/mscorlib/src/System/Text/UTF8Encoding.cs
```diff
@@ -369,7 +371,9 @@ internal override unsafe int GetByteCount(char* chars, int count, EncoderNLS bas
 
                     // Do our fallback.  Actually we already know its a mixed up surrogate,
                     // so the ref pSrc isn't gonna do anything.
-                    fallbackBuffer.InternalFallback(unchecked((char)ch), ref pSrc);
+                    pSrcForFalback = pSrc; // Avoid passing pSrc by reference to allow it to be enregistered
```
danmosemsft: typo 'falback'

danmosemsft: and elsewhere

src/mscorlib/src/System/Text/UnicodeEncoding.cs
```diff
@@ -1444,8 +1466,10 @@ internal override unsafe int GetCharCount(byte* bytes, int count, DecoderNLS bas
                             fallbackBuffer.InternalInitialize(byteStart, charEnd);
                         }
 
-                        if (!fallbackBuffer.InternalFallback(byteBuffer, bytes, ref chars))
+                        charsForFallback = chars; // Avoid passing chars by reference to allow it to be enregistered
+                        if (!fallbackBuffer.InternalFallback(byteBuffer, bytes, ref charsForFallback))
                         {
+                            chars = charsForFallback;
```
danmosemsft: It might be clearer instead to set chars in only one place ie
```c#
                        charsForFallback = chars; // Avoid passing chars by reference to allow it to be enregistered
                         bool result = fallbackBuffer.InternalFallback(byteBuffer, bytes, ref charsForFallback);
                             chars = charsForFallback;
                         if (result)
                          {		                          {

danmosemsft:
LGTM although there is a possibly related test failure.

Is there an issue to fix the JIT so it can do this by itself without this ugly workaround?

jkotas:
> Is there an issue to fix the JIT so it can do this by itself without this ugly workaround?

The analysis to prove that the address taken does not escape gets complex fast. The JIT does a little bit of it today, the C/C++ optimizing backends do some more of it; but you never want to take address of your perf critical loop control locals in any case and you need to structure your code accordingly.

The encoding fallback implementation is pretty poorly structured. I hope we will get a chance to refactor it as part of aligning the implementation with the corefxlab one. The fallback support were not in the code when I have fine tuned the encoding originally. They were added together with the ref on the perf critical locals later by somebody who did not know what he is doing.

src/mscorlib/src/System/Text/UTF32Encoding.cs
jkotas: Fixed.

jkotas:

PR #10128, opened 2017-03-12
Remove dead code

BruceForstall:
Remove code under `#ifdef PROTO_JIT`, which hasn't been defined
in a long time. I'm presuming that since this hasn't been defined
in a long time that the code isn't needed.


BruceForstall:
@dotnet/jit-contrib PTAL

PR #10132, opened 2017-03-13
Convert GenTree* to GenTreeCall* as much as possible

BruceForstall:
This simplifies reading the code.

### DIFF
--- a/src/jit/assertionprop.cpp
+++ b/src/jit/assertionprop.cpp
@@ -2099,7 +2099,7 @@ void Compiler::optAssertionGen(GenTreePtr tree)
             if ((tree->gtFlags & GTF_CALL_NULLCHECK) || ((tree->gtFlags & GTF_CALL_VIRT_KIND_MASK) != GTF_CALL_NONVIRT))
             {
                 //  Retrieve the 'this' arg
-                GenTreePtr thisArg = gtGetThisArg(tree);
+                GenTreePtr thisArg = gtGetThisArg(tree->AsCall());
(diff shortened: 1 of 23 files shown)


BruceForstall:
Interestingly, doing this reduces the number of asserts we run. In particular, for:
```
GenTree* tree;
tree->gtCall.gtCallObjp
```
The `gtCall` becomes AsCall(), which does an assert that `gtOper == GT_CALL`. When converting to the "real type" form:
```
GenTreeCall* call;
call->gtCallObjp
```
there is no such assert. I don't think we need to worry too much about this, since converting from `GenTree*` to `GenTreeCall*` uses `AsCall()` which does the assert.

The debugging experience is much better, as you can hover over `gtCall`, use it in QuickWatch, etc.

BruceForstall:
@dotnet/jit-contrib PTAL

mikedn:
There should be some kind of rule: thou shall not pass a GenTree* around if the object is known to be of a derived type :)

pgavlin:
> There should be some kind of rule: thou shall not pass a GenTree* around if the object is known to be of a derived type :)

Yes, definitely. Under-utilization of this knowledge is endemic to the compiler.

BruceForstall:
Random failure?
```
D:\j\workspace\x64_release_w---0575cb46\bin\tests\Windows_NT.x64.Release\TestWrappers\baseservices.threading\baseservices.threading.XUnitWrapper.cs(17537): error : baseservices_threading._mutex_misc_waitone2_waitone2_._mutex_misc_waitone2_waitone2_cmd [FAIL] [D:\j\workspace\x64_release_w---0575cb46\tests\runtest.proj]
```

PR #10134, opened 2017-03-13
Fix typo in Compiler::gtHasRef()

seanshpark:
Maybe from copy-paste typo?


seanshpark:
@BruceForstall PTAL

BruceForstall:
Nice catch. LGTM.
cc @dotnet/jit-contrib

Issue #10135, opened 2017-03-13
[Linux/ARM32] cross architecture build error


hseok-oh:
Build command:
./build.sh release cross arm crosscomponent

```
CMake Warning (dev) at src/coreclr/hosts/unixcorerun/CMakeLists.txt:35 (add_dependencies):
  Policy CMP0046 is not set: Error on non-existent dependency in
  add_dependencies.  Run "cmake --help-policy CMP0046" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.

  The dependency target "coreclr" of target "corerun" does not exist.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at src/coreclr/hosts/unixcoreconsole/CMakeLists.txt:34 (add_dependencies):
  Policy CMP0046 is not set: Error on non-existent dependency in
  add_dependencies.  Run "cmake --help-policy CMP0046" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.

  The dependency target "coreclr" of target "coreconsole" does not exist.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at src/debug/daccess/CMakeLists.txt:65 (add_dependencies):
  Policy CMP0046 is not set: Error on non-existent dependency in
  add_dependencies.  Run "cmake --help-policy CMP0046" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.

  The dependency target "coreclr" of target "daccess" does not exist.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Error at src/debug/daccess/CMakeLists.txt:47 (add_custom_command):
  Error evaluating generator expression:

    $<TARGET_FILE:coreclr>

  No target "coreclr"

.......
/usr/bin/ld: cannot find -lmscorrc_debug
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [src/ToolBox/superpmi/superpmi/superpmi] Error 1
make[1]: *** [src/ToolBox/superpmi/superpmi/CMakeFiles/superpmi.dir/all] Error 2

/usr/bin/ld: cannot find -lmscorrc_debug
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [src/ToolBox/superpmi/mcs/mcs] Error 1
make[1]: *** [src/ToolBox/superpmi/mcs/CMakeFiles/mcs.dir/all] Error 2

[...]

PR #10138, opened 2017-03-13
Fix dependency to coreclr and mscorrc in Linux

hseok-oh:
Remove dependency with coreclr in some projects for Linux
Recover build mscorrc for Linux cross architecture


hseok-oh:
related issue: #10135 
related PR: #10045

hseok-oh:
@dotnet-bot help

hseok-oh:
@dotnet-bot test Ubuntu x86 Debug please
@dotnet-bot test Ubuntu x86 Checked please
@dotnet-bot test Ubuntu x86 Release please

hseok-oh:
@dotnet-bot test Ubuntu x86 Debug please

src/coreclr/hosts/unixcoreconsole/CMakeLists.txt
```diff
@@ -31,8 +31,4 @@ if(NOT CLR_CMAKE_PLATFORM_ANDROID)
   )
 endif()
 
-add_dependencies(coreconsole
-    coreclr
```
janvorli: It is true that these dependencies were not really needed, since they are not build time, but rather runtime dependencies. But that makes me wonder - why do we build corerun and coreconsole during cross components build? I have tried to exclude the add_subdirectory(src/coreclr/hosts) for NOT DEFINED CLR_CROSS_COMPONENTS_BUILD and the build completed successfuly.

hseok-oh: I think so. Just I have one question, is it has any problem in windows environment if we remove this build dependency? It seems to build corerun and coreconsole during cross component build in Windows (and not installed), but I'm not good at Windows environment. If it also have no problem in Windows, I think it's better to merge add_subdirectory(src/coreclr/hosts) with Windows and Linux environment, by excluding add_subdirectory(src/coreclr/hosts) at root/CMakeLists.txt and use if statement using NOT DEFINED CLR_CROSS_COMPONENTS_BUILD instead of WIN32 in src/coreclr/CMakeLists.txt.

hseok-oh: I've tried to merge add_subdirectory(src/coreclr/hosts), but it looked like have some complicate build dependency problem. It's better to exclude the add_subdirectory(src/coreclr/hosts) for NOT DEFINED CLR_CROSS_COMPONENTS_BUILD as you suggested, and remain as another work item that is clean-up build dependencies.

rahku: >why do we build corerun and coreconsole during cross components build? 
I do not think this is true. cross-components build only builds dll/exe listed in file https://github.com/dotnet/coreclr/blob/master/crosscomponents.cmake

rahku: >why do we build corerun and coreconsole during cross components build?

I don't this is true. Cross-component build should only be building dll/exe's listed in file https://github.com/dotnet/coreclr/blob/master/crosscomponents.cmake

src/debug/daccess/CMakeLists.txt
```diff
@@ -63,4 +63,4 @@ if(CLR_CMAKE_PLATFORM_UNIX)
   )
 
   add_dependencies(daccess coreclr dactablerva_header)
-endif(CLR_CMAKE_PLATFORM_UNIX)
+endif(CLR_CMAKE_PLATFORM_UNIX AND NOT CLR_CROSS_COMPONENTS_BUILD)
```
rahku: Instead of these changes could you change this line from 
add_dependencies(daccess coreclr dactablerva_header)  
to
add_dependencies(daccess dactablerva_header)

and see if it works.

rahku: isn't these steps required to generate cross-arch mscordaccore?

src/coreclr/hosts/unixcoreconsole/CMakeLists.txt
rahku: But we process most of cmakelists.txt files unless it is excluded using CLR_CROSS_COMPONENTS_BUILD

src/debug/daccess/CMakeLists.txt
hseok-oh: We cannot build cross-arch mscordaccore in Linux yet. But someday we may try to build mscordaccore. If these dependencies need for mscordaccore, it's better to remain dependencies, and remove coreclr dependency only.

src/coreclr/hosts/unixcoreconsole/CMakeLists.txt
hseok-oh: @rahku Thanks for explanation. Then cmake traverses most of CMakeLists.txt, but finally corerun and coreconsole is not build, right?

rahku: This is correct

janvorli: @hseok-oh 
>  I think it's better to merge add_subdirectory(src/coreclr/hosts) with Windows and Linux environment, by excluding add_subdirectory(src/coreclr/hosts) at root/CMakeLists.txt

That is not possible. The reason this is included from the root CMakeLists.txt separate is that the Unix hosts are native Unix apps that are not built on top of PAL and so they need to include Unix platform header files. This is something that's disabled later in the root CMakeFiles.txt since none of the runtime code except the PAL is allowed to include these headers.

janvorli: @rahku 
>I don't this is true. Cross-component build should only be building dll/exe's listed in file https://github.com/dotnet/coreclr/blob/master/crosscomponents.cmake

Ah, right, I keep forgetting that we are still processing the CMakeLists.txt files but some targets are just not built since nothing needs them.

hseok-oh: @janvorli OK. I've fix to exclude `add_subdirectory(src/coreclr/hosts)` and `add_subdirectory(src/ildasm/unixcoreclrloader)` for NOT DEFINED CLR_CROSS_COMPONENTS_BUILD in the root CMakeFiles.txt.

janvorli: I think we actually don't need to exclude the hosts and unixcoreclrloader when you have removed the dependency of the hosts on coreclr. As @rahku said, although the cmake tree is parsed, the hosts are not being built. My original question stemmed from my not realizing that we are not building these. 
In the existing CMakeLists.txt files, we don't exclude targets from the build for CLR_CROSS_COMPONENTS_BUILD unless the exclusion is necessary. I can actually see that I was too eager to exclude the mscoree, mscorpe and mscorrc about two weeks ago. The issue I had could have been fixed by just a local change in the src/dlls/mscoree/coreclr/CMakeLists.txt to exclude the inject_debug_resources target for CLR_CROSS_COMPONENTS_BUILD.
I have just verified locally that it works too. Could you please augment your change to make the mscoree, mscorpe and mscorrc unconditional, revert the hosts and unixcoreclr change and instead of that wrap `add_custom_target(inject_debug_resources.....` in the src/dlls/mscoree/coreclr/CMakeLists.txt in `if(NOT CLR_CROSS_COMPONENTS_BUILD)` ?

hseok-oh:
@janvorli Fixed. PTAL

PR #10140, opened 2017-03-13
Improve LIR dumping

BruceForstall:
1. Use the LIR node dumper to display nodes to be generated by
codegen, since we're in LIR form at that point. Add a new
"prefix message" argument to allow "Generating: " to prefix all
such lines.
2. Fix off-by-one error in LIR dump due to `#ifdef` versus `#if`.
3. Remove extra trailing line for each LIR node. This interfered
with (1). But I always thought it was unnecessarily verbose; I don't
believe there is any ambiguity without that extra space.
4. Add dTreeLIR()/cTreeLIR() functions for use in the debugger.

### DIFF
--- a/src/jit/codegenarm.cpp
+++ b/src/jit/codegenarm.cpp
@@ -501,10 +501,12 @@ void CodeGen::genCodeForTreeNode(GenTreePtr treeNode)
 
 #ifdef DEBUG
     lastConsumedNode = nullptr;
+    if (compiler->verbose)
+    {
+        unsigned seqNum = treeNode->gtSeqNum; // Useful for setting a conditional break in Visual Studio
+        compiler->gtDispLIRNode(treeNode, "Generating: ");
+    }
 #endif
-
-    JITDUMP("Generating: ");
-    DISPNODE(treeNode);
 
     // contained nodes are part of their parents for codegen purposes
     // ex : immediates, most LEAs
--- a/src/jit/gentree.cpp
+++ b/src/jit/gentree.cpp
@@ -12002,16 +12002,22 @@ void Compiler::gtDispTreeRange(LIR::Range& containingRange, GenTree* tree)
 //
 // Arguments:
 //    node - the LIR node to dump.
+//    prefixMsg - an optional prefix for each line of output.
 //
-void Compiler::gtDispLIRNode(GenTree* node)
+void Compiler::gtDispLIRNode(GenTree* node, const char* prefixMsg /* = nullptr */)
 {
-    auto displayOperand = [](GenTree* operand, const char* message, IndentInfo operandArc, IndentStack& indentStack) {
+    auto displayOperand = [](GenTree* operand, const char* message, IndentInfo operandArc, IndentStack& indentStack, const char* prefixMsg = nullptr) {
         assert(operand != nullptr);
         assert(message != nullptr);
 
+        if (prefixMsg != nullptr)
+        {
+            printf("%s", prefixMsg);
+        }
+
         // 49 spaces for alignment
         printf("%-49s", "");
-#ifdef FEATURE_SET_FLAGS
+#if FEATURE_SET_FLAGS
         // additional flag enlarges the flag field by one character
         printf(" ");
 #endif
@@ -12022,7 +12028,6 @@ void Compiler::gtDispLIRNode(GenTree* node)
         operandArc = IIArc;
 
         printf("  t%-5d %-6s %s\n", operand->gtTreeID, varTypeName(operand->TypeGet()), message);
-
     };
 
     IndentStack indentStack(this);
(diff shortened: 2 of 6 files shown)


BruceForstall:
@dotnet/jit-contrib PTAL

src/jit/codegenarm.cpp
```diff
@@ -501,11 +501,13 @@ void CodeGen::genCodeForTreeNode(GenTreePtr treeNode)
 
 #ifdef DEBUG
     lastConsumedNode = nullptr;
+    if (compiler->verbose)
+    {
+        unsigned seqNum = treeNode->gtSeqNum; // Useful for setting a conditional break in Visual Studio
```
pgavlin: Can you not set a breakpoint on `treeNode->gtSeqNum` itself?

src/jit/gentree.cpp
```diff
@@ -12002,16 +12002,22 @@ void Compiler::gtDispTreeRange(LIR::Range& containingRange, GenTree* tree)
...
+void Compiler::gtDispLIRNode(GenTree* node, const char* prefixMsg /* = nullptr */)
 {
-    auto displayOperand = [](GenTree* operand, const char* message, IndentInfo operandArc, IndentStack& indentStack) {
+    auto displayOperand = [](GenTree* operand, const char* message, IndentInfo operandArc, IndentStack& indentStack, const char* prefixMsg = nullptr) {
         assert(operand != nullptr);
         assert(message != nullptr);
 
+        if (prefixMsg != nullptr)
+        {
+            printf("%s", prefixMsg);
```
pgavlin: Minor nit: you could capture the length of `prefixMsg` and use it to adjust the alignment below. I'd be interested to see how the codegen dumps looked in both cases.

src/jit/codegenarm.cpp
BruceForstall: Seems like it. I just copy-and-pasted that from the other codegen files. Since it doesn't hurt, it made sense to keep it the same.

src/jit/gentree.cpp
BruceForstall: The issue is that the prefixMsg is output for both the operand (here), and the operator (below), and the operator doesn't have any leading space. And, I want the operator and operands to be indented by the same amount, when a prefixMsg is given.

Here's an example from codegen:
```
Generating: N037 (  1,  4) [000034] ------------       t34 =    const(h)  int    0x8B6604 ftn REG NA
Generating:                                                  /--*  t34    int
Generating: N039 (  3,  6) [000035] ------------       t35 = *  indir     int    REG NA
Generating:                                                  /--*  t35    int    control expr
Generating: N041 ( 69, 33) [000008] --C-G-------             *  call      void   System.Console.WriteLine $VN.Void
```
A "full dump" looks like:
```
                                                 /--*  t32    int    control expr
N029 ( 45, 22) [000005] --C-G-------        t5 = *  call      float  Program.f2 $c3
                                                 /--*  t5     float
N031 ( 49, 25) [000021] DAC-G-----L-             *  st.lclVar float  V02 tmp1         d:2 REG NA
N033 (  3,  2) [000022] ------------       t22 =    lclVar    float  V02 tmp1         u:2 (last use) REG NA $c3
                                                 /--*  t22    float
N035 (???,???) [000033] ------------             *  putarg_stk [+0x00] float  REG NA
N037 (  1,  4) [000034] ------------       t34 =    const(h)  int    0x8B6604 ftn REG NA
                                                 /--*  t34    int
N039 (  3,  6) [000035] ------------       t35 = *  indir     int    REG NA
                                                 /--*  t35    int    control expr
N041 ( 69, 33) [000008] --C-G-------             *  call      void   System.Console.WriteLine $VN.Void
N043 (  2,  2) [000013] ------------                il_offset void   IL offset: 25 REG NA
N045 (  1,  1) [000011] ------------       t11 =    const     int    100 REG NA $42
                                                 /--*  t11    int
N047 (  2,  2) [000012] ------------             *  return    int    REG NA $1c0
```

BruceForstall:
@pgavlin @dotnet/jit-contrib any additional comments?

src/jit/gentree.cpp
pgavlin: This is hugely picky of me, but is there any chance we could avoid printing the message before the operands? No big deal if there isn't.

BruceForstall:
@pgavlin Done. Sample codegen output now:
```
Generating: N245 (  1,  1) [000115] ------------      t115 =    const     int    0 REG NA $40
                                                             /--*  t115   int
Generating: N247 (???,???) [000210] ------------             *  putarg_stk [+0x00] int    REG NA
IN0031:        push     0
Generating: N249 (  1,  1) [000113] ------------      t113 =    lclVar    ref    V00 this         u:2 esi (last use) REG esi RV $80
                                                             /--*  t113   ref
Generating: N251 (???,???) [000211] ------------      t211 = *  putarg_reg ref    REG ecx
```

pgavlin:
LGTM. Thanks for making the change!

PR #10143, opened 2017-03-13
Use VSD_STUB_CAN_THROW_AV instead of explicit TARGET_XXX

parjong:
This commit introduces VSD_STUB_CAN_THROW_AV, and annotates code related with VSD stub frame fixup using VSD_STUB_CAN_THROW_AV instead of explicit TARGET_XXX.


jkotas:
Thanks!

PR #10144, opened 2017-03-13
[x86/Linux] Fix IL_STUB_PInvoke with RetBuf

### LINKED ISSUE #10027: [x86/Linux] HFA tests failure
Failed Tests

 - JIT/jit64/hfa/main/testE/hfa_nd2E_r/hfa_nd2E_r.sh
 - JIT/jit64/hfa/main/testE/hfa_nf2E_d/hfa_nf2E_d.sh
 - JIT/jit64/hfa/main/testE/hfa_nf2E_r/hfa_nf2E_r.sh
 - JIT/jit64/hfa/main/testE/hfa_sd2E_d/hfa_sd2E_d.sh
 - JIT/jit64/hfa/main/testE/hfa_sd2E_r/hfa_sd2E_r.sh
 -  JIT/jit64/hfa/main/testE/hfa_sf2E_d/hfa_sf2E_d.sh
 - JIT/jit64/hfa/main/testE/hfa_sf2E_r/hfa_sf2E_r.sh
 - JIT/jit64/hfa/main/testE/hfa_sf2E_d/hfa_sf2E_d.sh
 - JIT/jit64/hfa/main/testE/hfa_sf2E_r/hfa_sf2E_r.sh

% All of these have a native function with RetBuf. Crashes when we P/Invoke it.

Error message

```
FAILED   - [ 5702 :    3 sec ] JIT/jit64/hfa/main/testE/hfa_nd2E_d/hfa_nd2E_d.sh
               BEGIN EXECUTION
               /root/dotnet-test/coreclr_unittest_20170228_083238_x86_x86-emu/coreclr-unittest-170227-233308/overlay/corerun hfa_nd2E_d.exe
               ./hfa_nd2E_d.sh: line 131:  7932 Segmentation fault      (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" hfa_nd2E_d.exe $CLRTestExecutionArguments
               Expected: 100
               Actual: 139
               END EXECUTION - FAILED
```

```
Program received signal SIGSEGV, Segmentation fault.
0xf4f17f0d in identity_HFA01 (hfa=...)
    at /var/jenkins/workspace/CLR/x86_Dbg_RT/repo/code/tests/src/JIT/jit64/hfa/main/dll/hfa_native.cpp:118
118     /var/jenkins/workspace/CLR/x86_Dbg_RT/repo/code/tests/src/JIT/jit64/hfa/main/dll/hfa_native.cpp: No such file or directory.
(gdb) i r
eax            0x3f800000       1065353216
ecx            0x3f800000       1065353216
edx            0xf4f64ac3       -185185597
ebx            0xffffd298       -11624
esp            0xffffc928       0xffffc928
ebp            0xffffc928       0xffffc928
esi            0x8087300        134771456
edi            0xffffc948       -14008
eip            0xf4f17f0d       0xf4f17f0d <identity_HFA01(HFA01)+13>
eflags         0x10286  [ PF SF IF RF ]
cs             0x23     35
ss             0x2b     43
ds             0x2b     43
es             0x2b     43
fs             0x0      0
gs             0x63     99
(gdb) disassemble identity_HFA01
Dump of assembler code for function identity_HFA01(HFA01):
   0xf4f17f00 <+0>:     push   %ebp
   0xf4f17f01 <+1>:     mov    %esp,%ebp
   0xf4f17f03 <+3>:     mov    0x8(%ebp),%eax
   0xf4f17f06 <+6>:     mov    %eax,%ecx
   0xf4f17f08 <+8>:     lea    0xc(%ebp),%edx
   0xf4f17f0b <+11>:    mov    (%edx),%edx
=> 0xf4f17f0d <+13>:    mov    %edx,(%eax)
   0xf4f17f0f <+15>:    mov    %ecx,%eax
   0xf4f17f11 <+17>:    pop    %ebp
   0xf4f17f12 <+18>:    ret    $0x4
End of assembler dump.
```

At the point it crashed, EAX must have been the address of return buffer. However EAX is 0x3f800000 which is 1.0 as float. Seems like calling convention problem.

cc: @janvorli @jkotas @parjong

wateret:
Fix calling convention and IL_STUB_PInvoke for native functions
which was problematic for native functions that has
a RetBuf argument(struct size <= 8) on x86/Linux.

Fix #10027

### DIFF
--- a/src/jit/compiler.cpp
+++ b/src/jit/compiler.cpp
@@ -1007,6 +1007,7 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
         // and also examine the clsHnd to see if it is an HFA of count one
         useType = getPrimitiveTypeForStruct(structSize, clsHnd);
     }
+
 #endif // FEATURE_UNIX_AMD64_STRUCT_PASSING
 
 #ifdef _TARGET_64BIT_
--- a/src/jit/compiler.cpp
+++ b/src/jit/compiler.cpp
@@ -997,6 +997,7 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
 
 #else // not UNIX_AMD64
 
+#ifndef UNIX_X86_ABI
     // The largest primitive type is 8 bytes (TYP_DOUBLE)
     // so we can skip calling getPrimitiveTypeForStruct when we
     // have a struct that is larger than that.
@@ -1007,6 +1008,7 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
         // and also examine the clsHnd to see if it is an HFA of count one
         useType = getPrimitiveTypeForStruct(structSize, clsHnd);
     }
+#endif
 
 #endif // FEATURE_UNIX_AMD64_STRUCT_PASSING
 
--- a/src/jit/compiler.cpp
+++ b/src/jit/compiler.cpp
@@ -997,7 +997,6 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
 
 #else // not UNIX_AMD64
 
-#ifndef UNIX_X86_ABI
     // The largest primitive type is 8 bytes (TYP_DOUBLE)
     // so we can skip calling getPrimitiveTypeForStruct when we
     // have a struct that is larger than that.
@@ -1008,8 +1007,6 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
         // and also examine the clsHnd to see if it is an HFA of count one
         useType = getPrimitiveTypeForStruct(structSize, clsHnd);
     }
-#endif
-
 #endif // FEATURE_UNIX_AMD64_STRUCT_PASSING
 
 #ifdef _TARGET_64BIT_
--- a/src/vm/callingconvention.h
+++ b/src/vm/callingconvention.h
@@ -1702,12 +1702,12 @@ inline BOOL HasRetBuffArg(MetaSig * pSig)
 }
 
 #ifdef UNIX_X86_ABI
+// For UNIX_X86_ABI and unmanaged function, we always need RetBuf if the return type is VALUETYPE
 inline BOOL HasRetBuffArgUnmanagedFixup(MetaSig * pSig)
 {
     WRAPPER_NO_CONTRACT;
-    CorElementType type = pSig->GetReturnTypeNormalized();
-
-    // For UNIX_X86_ABI and unmanaged function, we always need RetBuf if the return type is VALUETYPE
+    // We cannot just pSig->GetReturnType() here since it will return ELEMENT_TYPE_VALUETYPE for enums
+    CorElementType type = pSig->GetRetTypeHandleThrowing().GetVerifierCorElementType();
     return type == ELEMENT_TYPE_VALUETYPE;
 }
 #endif
--- a/src/vm/callingconvention.h
+++ b/src/vm/callingconvention.h
@@ -1415,8 +1415,11 @@ void ArgIteratorTemplate<ARGITERATOR_BASE>::ComputeReturnFlags()
             }
 #endif
 
+#ifndef UNIX_X86_ABI
             if  (size <= ENREGISTERED_RETURNTYPE_INTEGER_MAXSIZE)
                 break;
+#endif
+
 #endif // UNIX_AMD64_ABI && FEATURE_UNIX_AMD64_STRUCT_PASSING
         }
 #endif // ENREGISTERED_RETURNTYPE_INTEGER_MAXSIZE
--- a/src/vm/callingconvention.h
+++ b/src/vm/callingconvention.h
@@ -1415,11 +1415,8 @@ void ArgIteratorTemplate<ARGITERATOR_BASE>::ComputeReturnFlags()
             }
 #endif
 
-#ifndef UNIX_X86_ABI
             if  (size <= ENREGISTERED_RETURNTYPE_INTEGER_MAXSIZE)
                 break;
-#endif
-
 #endif // UNIX_AMD64_ABI && FEATURE_UNIX_AMD64_STRUCT_PASSING
         }
 #endif // ENREGISTERED_RETURNTYPE_INTEGER_MAXSIZE
@@ -1703,6 +1700,17 @@ inline BOOL HasRetBuffArg(MetaSig * pSig)
     ArgIterator argit(pSig);
     return argit.HasRetBuffArg();
 }
+
+#ifdef UNIX_X86_ABI
+inline BOOL HasRetBuffArgUnmanagedFixup(MetaSig * pSig)
+{
+    WRAPPER_NO_CONTRACT;
+    CorElementType type = pSig->GetReturnTypeNormalized();
+
+    // For UNIX_X86_ABI and unmanaged function, we always need RetBuf if the return type is VALUETYPE
+    return type == ELEMENT_TYPE_VALUETYPE;
+}
+#endif
 
 inline BOOL IsRetBuffPassedAsFirstArg()
 {
--- a/src/vm/dllimport.cpp
+++ b/src/vm/dllimport.cpp
@@ -3684,10 +3684,12 @@ static void CreateNDirectStubWorker(StubState*         pss,
         // return buffer in correct register.
         // The return structure secret arg comes first, however byvalue return is processed at
         // the end because it could be the HRESULT-swapped argument which always comes last.
-        fMarshalReturnValueFirst = HasRetBuffArg(&msig);
 
 #ifdef UNIX_X86_ABI
-        fMarshalReturnValueFirst |= HasRetBuffArgUnmanagedFixup(&msig);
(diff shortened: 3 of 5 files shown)


wateret:
cc: @janvorli @jkotas @parjong

wateret:
With the fix on `ilmarshalers.h`, IL_STUB_PInvoke generates code pushing RetBuf on stack as first argument.

**Disasm of IL_STUB_PInvoke for Identity_HFA02**

Master:
```
; Assembly listing for method DomainBoundILStubClass:IL_STUB_PInvoke(struct):struct
...
       83EC08       sub      esp, 8                     // Begin argument push
       F30F7E45CC   movq     xmm0, qword ptr [ebp-34H]
       83EC08       sub      esp, 8
       660FD60424   movq     dword ptr [esp], xmm0      // End argument push
       C745B008000000 mov      dword ptr [ebp-50H], 8
       8BC4         mov      eax, esp
       8945B4       mov      dword ptr [ebp-4CH], eax
       8D05843775F4 lea      eax, G_M23675_IG04
       8945B8       mov      dword ptr [ebp-48H], eax
       8B45C4       mov      eax, dword ptr [ebp-3CH]
       C6400800     mov      byte  ptr [eax+8], 0
       8B45C8       mov      eax, dword ptr [ebp-38H]

G_M23675_IG03:
       FFD0         call     eax.       // call native function
       83C408       add      esp, 8
```
this patch applied:
```
       83EC04       sub      esp, 4                     // Begin argument push
       F30F7E45D4   movq     xmm0, qword ptr [ebp-2CH]
       83EC08       sub      esp, 8
       660FD60424   movq     dword ptr [esp], xmm0
       8D45DC       lea      eax, bword ptr [ebp-24H]
       50           push     eax                         // End argument push
       C745B80C000000 mov      dword ptr [ebp-48H], 12
       8BC4         mov      eax, esp
       8945BC       mov      dword ptr [ebp-44H], eax
       8D058FE7DEF1 lea      eax, G_M23675_IG04
       8945C0       mov      dword ptr [ebp-40H], eax
       8B45CC       mov      eax, dword ptr [ebp-34H]
       C6400800     mov      byte  ptr [eax+8], 0
       8B45D0       mov      eax, dword ptr [ebp-30H]

G_M23675_IG03:
       FFD0         call     eax        // call native function
       83C404       add      esp, 4
```

src/jit/compiler.cpp
```diff
@@ -1007,6 +1008,7 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
         // and also examine the clsHnd to see if it is an HFA of count one
         useType = getPrimitiveTypeForStruct(structSize, clsHnd);
     }
+#endif
```
parjong: #10027 states that Identity_HFA01 return by EAX (without RetBuf). Does this change reflect that? It seems that we need to use ``structSize <= sizeof(float)`` instead of ``structSize <= sizeof(double)``.

src/vm/ilmarshalers.h
```diff
@@ -607,12 +607,15 @@ class ILMarshaler
...
             // and use byrefNativeReturn for all other structs.
+            // for UNIX_X86_ABI, we always need a return buffer argument for any size of structs.
             switch (nativeSize)
             {
+#ifndef UNIX_X86_ABI
                 case 1: typ = ELEMENT_TYPE_U1; break;
                 case 2: typ = ELEMENT_TYPE_U2; break;
                 case 4: typ = ELEMENT_TYPE_U4; break;
                 case 8: typ = ELEMENT_TYPE_U8; break;
+#endif
```
parjong: Similarly, does the following patch still resolve this issue?
``
                 case 1: typ = ELEMENT_TYPE_U1; break;
                  case 2: typ = ELEMENT_TYPE_U2; break;
                  case 4: typ = ELEMENT_TYPE_U4; break;
#ifndef UNIX_X86_ABI
                  case 8: typ = ELEMENT_TYPE_U8; break;
#endif
```

File: src/jit/compiler.cpp (hunk 1 above)
Comment (wateret) [inline src/jit/compiler.cpp]: As I mentioned here(https://github.com/dotnet/coreclr/issues/10027#issuecomment-285281093), 

P/Invoke is done like:
```
(Managed Code) --(1)--> (IL_STUB_PInvoke) --(2)--> (Native Code)
```

I think it(return by EAX) is the convention for (1). We actually need return buffer for (2).
Now I see this again, this change is probably unnecessary since it will change the calling convention for all managed code.

File: src/vm/ilmarshalers.h (hunk 2 above)
Comment (wateret) [inline src/vm/ilmarshalers.h]: Similarly to https://github.com/dotnet/coreclr/pull/10144#discussion_r105602129, for native code we always need a RetBuf regardless of the size of struct. I think this part is the one that fixes (2) part.

File: src/jit/compiler.cpp (hunk 1 above)
Comment (parjong) [inline src/jit/compiler.cpp]: Do you mean that native code always uses return buffer? If so, please ignore my previous comment.

Comment (wateret) [inline src/jit/compiler.cpp]: I think so. https://github.com/dotnet/coreclr/issues/10027#issuecomment-285309609 (`IsUnmanagedValueTypeReturnedByRef` always returns true) implies that. Anyway the change here may affect the managed code calling convention, even if it works fine. I should look at it again. Thank you!

File: src/vm/callingconvention.h [hunk 3]
```diff
@@ -1415,8 +1415,11 @@ void ArgIteratorTemplate<ARGITERATOR_BASE>::ComputeReturnFlags()
             }
 #endif
 
+#ifndef UNIX_X86_ABI
             if  (size <= ENREGISTERED_RETURNTYPE_INTEGER_MAXSIZE)
```
jkotas: It a common assumption in managed code that wrapping a primitive type in a struct has no overhead. This change breaks that assumption for Linux x86.

Can we avoid changing the managed calling convention, and change this for the unmanaged calling convention only?

wateret: Could I ask a good way to do that? I don't see a trivial way to find out if this is managed or unmanaged.

There is `m_fUnmanagedCallConv` flag which is used only with `FEATURE_INTERPRETER` on but it is not even defined in class members.

jkotas: If the comment is right, you should be able to tweak it here:

https://github.com/dotnet/coreclr/blob/e5994fa5507a5f08058193ff26dc3698cd2e6444/src/vm/dllimport.cpp#L3687

wateret:
@jkotas Thank you for your review. I updated the PR. PTAL.

src/vm/callingconvention.h
```diff
@@ -1701,6 +1701,17 @@ inline BOOL HasRetBuffArg(MetaSig * pSig)
...
 }
 
+#ifdef UNIX_X86_ABI
+inline BOOL HasRetBuffArgUnmanagedFixup(MetaSig * pSig)
+{
+    WRAPPER_NO_CONTRACT;
+    CorElementType type = pSig->GetReturnTypeNormalized();
+
+    // For UNIX_X86_ABI and unmanaged function, we always need RetBuf if the return type is VALUETYPE
+    return type == ELEMENT_TYPE_VALUETYPE;
```
jkotas: This will return false for a simple struct like `struct X { int x; }` because of it will get normalized to int, but you want it to return true.

I think you need something like:

`pSig->GetReturnTypeHandleThrowing().GetVerifierCorElementType() == ELEMENT_TYPE_VALUETYPE`

src/vm/dllimport.cpp
```diff
@@ -3685,7 +3685,12 @@ static void CreateNDirectStubWorker(StubState*         pss,
         // The return structure secret arg comes first, however byvalue return is processed at
         // the end because it could be the HRESULT-swapped argument which always comes last.
         fMarshalReturnValueFirst = HasRetBuffArg(&msig);
-#endif
+
+#ifdef UNIX_X86_ABI
+        fMarshalReturnValueFirst |= HasRetBuffArgUnmanagedFixup(&msig);
```
jkotas: I would make this 

```
#ifdef UNIX_X86_ABI
    fMarshalReturnValueFirst = ... maybe even inline HasRetBuffArgUnmanagedFixup here since it is trivial and nothing else needs it ...
#else
    fMarshalReturnValueFirst = HasRetBuffArg(&msig); 
#endif
```

src/vm/callingconvention.h
wateret: Wouldn't `GetReturnType()` work? instead of `GetRetTypeHandleThrowing()`

jkotas: `GetReturnType()` returns `ELEMENT_TYPE_VALUETYPE` for enums

wateret: Since `void ArgIteratorTemplate<ARGITERATOR_BASE>::ComputeReturnFlags()` uses `GetReturnType()` as well. 

https://github.com/dotnet/coreclr/blob/19e1cd299da301d1b2cb18ae8479595dbfc7224c/src/vm/callingconvention.h#L1329

jkotas: Yes, but then it does a ton of stuff in the `ELEMENT_TYPE_VALUETYPE` case.

Enums are encoded as ELEMENT_TYPE_VALUETYPE in signatures, but they are treated as primitive types for calling convention. `pSig->GetReturnTypeHandleThrowing().GetVerifierCorElementType()` will get the enum treated as primitive type.

wateret:
@jkotas Updated as your comment.

PR #10150, opened 2017-03-13
Re-insert Supplemental Payload into Helix payload

wtgodbe:
After our buildtools update last week, CoreCLR tests started failing to execute in Helix because the Supplemental Payload containing the Python test-runner scripts wasn't being found. This should fix that issue.


wtgodbe:
CC @MattGal @gkhanna79 - local testing showed jobs executing & passing in Helix with this change. This is a temporary fix until I integrate Matt's latest refactor of BuildTools Cloud Testing targets.

PR #10156, opened 2017-03-14
Cleanup most code in StringBuilder

jamesqo:
I tried looking into implementing https://github.com/dotnet/corefx/issues/14398 for `StringBuilder`, but the code is kinda messy right now. I cleaned it up a little bit by fixing typos in comments/removing unnecessary ones, adding XML docs, using expression-bodied members, removing obvious assert messages, etc. Not sure whether I still want to pursue the corefx issue because I think it could prove hard to refactor `StringBuilder`, but here is the cleanup at any rate.

@jkotas @AlexGhiondea PTAL


jkotas:
cc @alexperovich Maybe useful to unify StringBuilder with CoreRT first. This cleanup will be hard to port otherwise.

jamesqo:
@jkotas 

> Maybe useful to unify StringBuilder with CoreRT first. This cleanup will be hard to port otherwise.

What are some of the differences between the coreclr/corert StringBuilders? How much work would it be to unify them?

jkotas:
> What are some of the differences between the coreclr/corert StringBuilders

Diff the two files, understand and resolve each delta.

Number of similar refactoring changes went in recently, one example from many: https://github.com/dotnet/coreclr/pull/10270/

alexperovich:
I just submitted #10357 to move StringBuilder to shared. Once that is in you can modify the version in shared and everything should work.

danmosemsft:
@jamesqo do you want to resolve the merge now Alex's change went in?

jamesqo:
@danmosemsft I'm a little busy right now, but I'll resolve conflicts as soon as I have time.

jamesqo:
Fixed merge conflicts. @jkotas PTAL

jkotas:
Build breaks...

jamesqo:
@dotnet-bot test this please (404 errors)

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -1322,161 +1210,62 @@ public StringBuilder Insert(int index, char[] value, int startIndex, int charCou
...
-        // the buffer at index. Existing characters are shifted to make room for the new text.
-        // The capacity is adjusted as needed. 
-        // 
         [CLSCompliant(false)]
-        public StringBuilder Insert(int index, ulong value)
-        {
-            Contract.Ensures(Contract.Result<StringBuilder>() != null);
-            return Insert(index, value.ToString(), 1);
-        }
+        public StringBuilder Insert(int index, ulong value) => Insert(index, value.ToString(), 1);
```
danmosemsft: @jkotas what value is there in retaining [CLSCompilant(false)]? Does our C# compiler check these attributes?

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -1322,161 +1210,62 @@ public StringBuilder Insert(int index, char[] value, int startIndex, int charCou
...
+        public StringBuilder AppendFormat(String format, Object arg0, Object arg1, Object arg2) => AppendFormatHelper(null, format, new ParamsArray(arg0, arg1, arg2));
 
         public StringBuilder AppendFormat(String format, params Object[] args)
         {
             if (args == null)
             {
                 // To preserve the original exception behavior, throw an exception about format if both
                 // args and format are null. The actual null check for format is in AppendFormatHelper.
-                throw new ArgumentNullException((format == null) ? nameof(format) : nameof(args));
+                string paramName = format == null ? nameof(format) : nameof(args);
```
danmosemsft: I think this was fine before, but in any case, parens around `format == null` please

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -1768,7 +1559,7 @@ public bool Equals(StringBuilder sb)
                     thisChunkIndex = thisChunk.m_ChunkLength + thisChunkIndex;
                 }
 
-                // Decrement the pointer to the 'this' StringBuilder
+                // TODO: Doc the fields.
```
danmosemsft: What is this for?

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -274,27 +248,29 @@ void ISerializable.GetObjectData(SerializationInfo info, StreamingContext contex
...
-                Debug.Assert(currentBlock.m_ChunkOffset >= 0, "Negative offset");
+                Debug.Assert(currentBlock.m_ChunkLength <= currentBlock.m_ChunkChars.Length);
+                Debug.Assert(currentBlock.m_ChunkLength >= 0);
+                Debug.Assert(currentBlock.m_ChunkOffset >= 0);
 
                 StringBuilder prevBlock = currentBlock.m_ChunkPrevious;
                 if (prevBlock == null)
                 {
-                    Debug.Assert(currentBlock.m_ChunkOffset == 0, "First chunk's offset is not 0");
+                    Debug.Assert(currentBlock.m_ChunkOffset == 0);
```
danmosemsft: maybe just `Debug.Assert(currentBlock.m_ChunkPrevious != null || currentBlock.m_ChunkOffset == 0)`

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -509,29 +486,28 @@ public int Length
...
 
                 int delta = value - Length;
-                // if the specified length is greater than the current length
                 if (delta > 0)
                 {
-                    // the end of the string value of the current StringBuilder object is padded with the Unicode NULL character
-                    Append('\0', delta);        // We could improve on this, but who does this anyway?
+                    // Pad ourselves with null characters.
+                    // TODO: Improve on comment here?
+                    Append('\0', delta); // We could improve on this, but who does this anyway?
```
danmosemsft: I don't know what this comment means either. I would remove it.

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -938,150 +943,73 @@ public StringBuilder Remove(int startIndex, int length)
...
-            Contract.Ensures(Contract.Result<StringBuilder>() != null);
-
-            if (null == value)
-            {
-                //Appending null is now a no-op.
-                return this;
-            }
-            return Append(value.ToString());
-        }
+        public StringBuilder Append(object value) => value == null ? this : Append(value.ToString());
```
danmosemsft: parens around `value == null` please

src/mscorlib/shared/System/Text/StringBuilder.cs
jkotas: > Does our C# compiler check these attributes?

Yes.

It is not strictly needed in implementation assemblies, but it is convenient to have reference assemblies and implementation to be marked with same attributes.

FXDC cares about the CLSCompilant according to https://github.com/dotnet/corefx/issues/17067#issuecomment-294915792.

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -938,150 +943,73 @@ public StringBuilder Remove(int startIndex, int length)
...
 
-        // Appends all of the characters in value to the current instance.
         public StringBuilder Append(char[] value)
         {
             Contract.Ensures(Contract.Result<StringBuilder>() != null);
 
-            if (null != value && value.Length > 0)
+            if (value?.Length > 0)
```
danmosemsft: @stephentoub did you tell me at some point that it was more efficient if it's a hot path to use check null and length separately rather than use the null coalescing operator? I compared the IL of the two and it's not clera to me.

jamesqo: @danmosemsft Please see related Roslyn issue: https://github.com/dotnet/roslyn/issues/17395 TL;DR it's more efficient if the null-condition is on the right hand side, there's no difference if it's on the left hand side.

danmosemsft:
@jamesqo do you plan to complete this one?

jamesqo:
@danmosemsft Yup, I have forgot about this PR. Thanks for reminding me.

src/mscorlib/shared/System/Text/StringBuilder.cs
jamesqo: @danmosemsft, there is a `break` statement after the assert because this means we've reached a tail node. Without the `break` we would enter an infinite loop.

jamesqo:
@danmosemsft PTAL. I realized that I actually omitted a lot of work that I had on my laptop but didn't push to GitHub, so you may want to review in-depth again. I also responded to your PR feedback.

danmosemsft:
@jamesqo thanks I'll take a look in a bit (ping me if I don't)

Since this is basically cleanup, I suggest we wait to merge until after the 2.0 fork this week. Sound good?

jamesqo:
@danmosemsft, sounds good.

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -25,140 +25,146 @@ namespace System.Text
...
-            m_MaxCapacity = Int32.MaxValue;
+
+            m_MaxCapacity = int.MaxValue;
             if (capacity == 0)
             {
                 capacity = DefaultCapacity;
             }
-            if (capacity < length)
-                capacity = length;
+            capacity = Math.Max(capacity, length);
```
AlexGhiondea: Is this actually faster?

jamesqo: @AlexGhiondea This has nothing to do with speed, I thought this was cleaner.

AlexGhiondea: It is clearer, but it introduces a method call in this code path. 
Given that this is not on a 'hot' path, I think this change should be fine.

jamesqo: @AlexGhiondea I do not think it will introduce a method call since it will be optimized by the JIT.

AlexGhiondea: For this particular method (Math.Max) there is a high chance that the JIT will inline the call (the implementation of the method is very small). 

I am fine with the change either way (I think it is clear to read in both versions :) )

danmosemsft:
@AlexGhiondea did you CR, do you want to signoff/give other feeedback?

jamesqo:
@danmosemsft I do not think he CR'd. He made only one comment, and it was about performance.

AlexGhiondea:
I had one comment about using Math.Max but that is not blocking in this case.

I will take another look and respond back by lunch time today.

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -1204,55 +1158,15 @@ public StringBuilder Insert(int index, String value)
...
-        // StringBuilder is not changed.
-        // 
-        public StringBuilder Insert(int index, short value)
-        {
-            Contract.Ensures(Contract.Result<StringBuilder>() != null);
-            return Insert(index, value.ToString(), 1);
-        }
+        public StringBuilder Insert(int index, short value) => Insert(index, value.ToString(), 1);
 
-        // Returns a reference to the StringBuilder with ; value inserted into
```
AlexGhiondea: Any particular reason to remove these comments?

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -1334,161 +1237,62 @@ public StringBuilder Insert(int index, char[] value, int startIndex, int charCou
...
-        // 
         [CLSCompliant(false)]
-        public StringBuilder Insert(int index, ulong value)
-        {
-            Contract.Ensures(Contract.Result<StringBuilder>() != null);
-            return Insert(index, value.ToString(), 1);
-        }
+        public StringBuilder Insert(int index, ulong value) => Insert(index, value.ToString(), 1);
 
-        // Returns a reference to this string builder with value inserted into
```
AlexGhiondea: Any reason to remove this comment?

danmosemsft:
@jamesqo ..

src/mscorlib/shared/System/Text/StringBuilder.cs
jamesqo: @AlexGhiondea I believe this is fairly obvious from the naming of the method. The comment only takes up a lot of space.

src/mscorlib/shared/System/Text/StringBuilder.cs
jamesqo: @AlexGhiondea I believe this is fairly obvious from the naming of the method. The comment only takes up a lot of space.

jamesqo:
@AlexGhiondea @danmosemsft I think it was the right decision to remove the comments above `Insert` since most people know what that method does, and the comments were taking up a lot of space. Do you still disagree, and is there any other PR feedback? If not, I think it would be nice to merge this PR quickly to avoid the chance of further merge conflicts.

src/mscorlib/shared/System/Text/StringBuilder.cs
AlexGhiondea: @jamesqo I am not sure the name of the method conveys all the information in the comment. If the only reason to remove them is saving some space, I would rather we keep them around.

jamesqo: @AlexGhiondea I'm not going to push too much here, but I disagree. I'm sure developers know what it means to `Insert` something in a `List<T>`, so wouldn't most be able to guess what `Insert` does for `StringBuilder`? How else do you think they could interpret it?

If you insist on me keeping the comments, I will have to convert all of them to triple-slash XML doc comments since I consider double-slash ones very ad hoc/messy.

danmosemsft: What does `; value` mean anyway? seems corrupted. My 2c is it's fine to rmove these.

danmosemsft: But I don't care so long as this super old PR gets merged :)

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -25,140 +25,146 @@ namespace System.Text
...
             {
                 throw new ArgumentOutOfRangeException(nameof(startIndex), SR.ArgumentOutOfRange_StartIndex);
             }
             Contract.EndContractBlock();
 
-            if (value == null)
-            {
-                value = String.Empty;
-            }
+            value = value ?? string.Empty;
```
stephentoub: I would prefer we not make changes like this that add additional unnecessary writes.

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -232,15 +236,10 @@ private StringBuilder(SerializationInfo info, StreamingContext context)
...
-                if (persistedCapacity < persistedString.Length)
-                {
-                    persistedCapacity = persistedString.Length;
-                }
-                if (persistedCapacity > persistedMaxCapacity)
-                {
-                    persistedCapacity = persistedMaxCapacity;
-                }
+                persistedCapacity = Math.Max(persistedCapacity, persistedString.Length);
+                persistedCapacity = Math.Min(persistedCapacity, persistedMaxCapacity);
```
stephentoub: This could just be:
```C#
persistedCapacity = Math.Min(Math.Max(DefaultCapacity, persistedString.length), persistedMaxCapacity);
```

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -213,16 +220,13 @@ private StringBuilder(SerializationInfo info, StreamingContext context)
...
                         break;
                 }
             }
 
             // Check values and set defaults
-            if (persistedString == null)
-            {
-                persistedString = String.Empty;
-            }
+            persistedString = persistedString ?? string.Empty;
```
stephentoub: Same comment about unnecessary writes.  Yes, this is slightly shorter C#, but at the expense of more work (even if small).

src/mscorlib/shared/System/Text/StringBuilder.cs
jamesqo: @stephentoub This is not optimized by the JIT?

stephentoub: I've not checked. But even if it is, it's more IL.

jamesqo:
@stephentoub Your PR feedback has been addressed.

danmosemsft:
@jamesqo I'm fine with removing the comments. Can you please rebase? Sorry this has taken so long.

jamesqo:
@danmosemsft Done. (Sorry for the delay, I thought I had done it earlier.)

danmosemsft:
@stephentoub any other feedback? if not we can merge

stephentoub:

jamesqo:
@stephentoub Checks have passed :)

Issue #9488, opened 2017-02-10
[ARM32/Linux] Cross-architectrue crossgen: loading System.Private.CoreLib.ni.dll failure


hseok-oh:
Failure loading process of `System.Private.CoreLib.ni.dll` generated by cross-architecture crossgen (x86-host / arm-target)

gdb backtrace
```
Thread 1 "corerun" received signal SIGSEGV, Segmentation fault.
0x763d6d26 in Module::GetAssembly (this=0x0) at /nfs/rbp3/coreclr/src/vm/ceeload.inl:390
390         return m_pAssembly;
(gdb) bt
#0  0x763db396 in Module::GetAssembly (this=0x0) at /nfs/rbp3/coreclr/src/vm/ceeload.inl:390
#1  0x765df0ce in ModuleSecurityDescriptor::VerifyDataComputed (this=0x76cbb25c <g_TrapReturningThreads>) at /nfs/rbp3/coreclr/src/vm/securitymeta.cpp:1579
#2  0x766dc506 in AssemblySecurityDescriptor::ResolvePolicy (this=0x6b708, pSharedSecDesc=0x6b890, fShouldSkipPolicyResolution=0)
    at /nfs/rbp3/coreclr/src/vm/securitydescriptorassembly.cpp:535
#3  0x76651214 in Assembly::SetDomainAssembly (this=0x6ba88, pDomainAssembly=0x6b670) at /nfs/rbp3/coreclr/src/vm/assembly.cpp:1267
#4  0x766bb436 in DomainAssembly::SetAssembly (this=0x6b670, pAssembly=0x6ba88) at /nfs/rbp3/coreclr/src/vm/domainfile.cpp:1954
#5  0x766bc790 in DomainAssembly::Allocate (this=0x6b670) at /nfs/rbp3/coreclr/src/vm/domainfile.cpp:2880
#6  0x766b85e4 in DomainFile::DoIncrementalLoad (this=0x6b670, level=FILE_LOAD_ALLOCATE) at /nfs/rbp3/coreclr/src/vm/domainfile.cpp:585
#7  0x76633100 in AppDomain::TryIncrementalLoad (this=0x42fe0, pFile=0x6b670, workLevel=FILE_LOAD_ALLOCATE, lockHolder=...) at /nfs/rbp3/coreclr/src/vm/appdomain.cpp:7190
[...]

PR #10159, opened 2017-03-14
[Linux] Define critical section for cross components

hseok-oh:
It modifies size of critical section same in all architecture.
Finally it makes same size of Module in all architecture.
It enables cross-crossgen (x86-host/arm-target) in Linux.

related issue: #9488

### DIFF
--- a/src/pal/inc/pal.h
+++ b/src/pal/inc/pal.h
@@ -2577,11 +2577,7 @@ PALIMPORT BOOL PALAPI PAL_VirtualUnwindOutOfProc(CONTEXT *context,
 #warning 
 #error  PAL_CS_NATIVE_DATA_SIZE is not defined for this architecture
 #endif
-#define CS_NATIVE_DATA_STRUCT_SIZE 128
     
-// CRITICAL_SECTION should pair with PAL_CRITICAL_SECTION because it is casted explicitly.
-// CRITICAL_SECTION uses same size of CNativeDataStorage (CS_NATIVE_DATA_STRUCT_SIZE) in all architecture
-//  using extra space for compativility of critical section in cross-architecture component.
 typedef struct _CRITICAL_SECTION {
     PVOID DebugInfo;
     LONG LockCount;
@@ -2592,10 +2588,10 @@ typedef struct _CRITICAL_SECTION {
 
     BOOL bInternal;
     volatile DWORD dwInitState;
-    struct CSNativeDataStorage
+    union CSNativeDataStorage
     {
         BYTE rgNativeDataStorage[PAL_CS_NATIVE_DATA_SIZE]; 
-        BYTE rgExtraStorage[CS_NATIVE_DATA_STRUCT_SIZE - PAL_CS_NATIVE_DATA_SIZE]; // make sure the storage is machine-pointer-size aligned
+        void * pvAlign; // make sure the storage is machine-pointer-size aligned
     } csnds;    
 } CRITICAL_SECTION, *PCRITICAL_SECTION, *LPCRITICAL_SECTION;
 
--- a/src/pal/inc/pal.h
+++ b/src/pal/inc/pal.h
@@ -2577,8 +2577,11 @@ PALIMPORT BOOL PALAPI PAL_VirtualUnwindOutOfProc(CONTEXT *context,
 #warning 
 #error  PAL_CS_NATIVE_DATA_SIZE is not defined for this architecture
 #endif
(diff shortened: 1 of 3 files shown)


hseok-oh:

janvorli:
@hseok-oh I wonder - isn't the problem actually caused by the fact that the ifdefs to define the PAL_CS_NATIVE_DATA_SIZE use _ARM_ etc instead of the _TARGET_ARM_ etc?

hseok-oh:
@janvorli For crossgen, that's true. But I just have a concern that struct `CRITICAL_SECTION` can be used for host environment, not target in another cross-arch components. Is it used for target only?

janvorli:
@hseok-oh I think you are right. E.g. when initializing PAL, critical section is used by the code of e.g. crossgen for its synchronization purposes, but it is also used in the Module structure as you've mentioned, where it represents the target critical section. 
@jkotas it seems fragile that cross tool crossgen e.g. for arm running on x86 use the layout of the Module class of the host architecture for the target one. If we would like to enable cross tools that would have different address sizes for host and target, this would become a problem on Windows too. 
Do you see some reasonably easy way to make crossgen use the Module structure with the correct target layout instead of unifying the layout by making the CRITICAL_SECTION size the same for all targets on Unix like this PR does?

jkotas:
> it seems fragile that cross tool crossgen e.g. for arm running on x86 use the layout of the Module class of the host architecture for the target one

Yes, it is fragile. The fix should be to define the target layout for CRITICAL_SECTION in inc\crosscomp.h and than use that.

> cross tools that would have different address sizes for host and target, this would become a problem on Windows too

Right, the cross-targeting crossgen depends on the same bitness of host and target in many places. The debugger DAC has the same problem.

hseok-oh:
@janvorli @jkotas Fixed. PTAL

src/vm/crst.h
```diff
@@ -296,7 +296,11 @@ friend class Crst;
 #endif
 
     union {
+#ifdef CROSSGEN_COMPILE
+        T_CRITICAL_SECTION  m_criticalsection;
```
jkotas: > we will need to bifurcate the CrstBase and some other classes derived from it (CrstStatic, CrstExplicitInit)

I do not think this is needed. crossgen does not have any synchronization because of it is single threaded. Crst operations are implemented as no-opts in crossgen: https://github.com/dotnet/coreclr/blob/master/src/vm/crossgencompile.cpp#L111.

So it is safe to change the Crst data fields to anything we want.

Some comments about it would be useful though.

src/vm/crst.h
```diff
@@ -296,7 +296,11 @@ friend class Crst;
 #endif
 
     union {
+#ifdef CROSSGEN_COMPILE
```
jkotas: This may rather be `CROSS_COMPILE`  so that it kicks for any kind of cross-compilation (debugger DAC), not just for crossgen.

src/pal/inc/pal.h
```diff
@@ -2589,10 +2592,10 @@ typedef struct _CRITICAL_SECTION {
 
     BOOL bInternal;
     volatile DWORD dwInitState;
-    union CSNativeDataStorage
+    struct CSNativeDataStorage
     {
         BYTE rgNativeDataStorage[PAL_CS_NATIVE_DATA_SIZE]; 
-        VOID * pvAlign; // make sure the storage is machine-pointer-size aligned
+        BYTE rgExtraStorage[CS_NATIVE_DATA_STRUCT_SIZE - PAL_CS_NATIVE_DATA_SIZE]; // make sure the storage is machine-pointer-size aligned
```
hseok-oh: I've recovered `pal/inc/pal.h` so `CS_NATIVE_DATA_STRUCT_SIZE` is not used any more.

src/vm/crst.h
hseok-oh: Other cross-compilation (debugger, DAC) works correctly when we use `T_CRITICAL_SECTION` here? I cannot be convinced because I'm not familiar with these components.

jkotas: I think so

danmosemsft:
@hseok-oh what is the status of this PR?

hseok-oh:
@danmosemsft I'm not working this PR now because of other issues. But I'll resume this soon to enable cross-crossgen.

hseok-oh:
I confirmed that this issue is solved on recent master branch. So I close this PR.

PR #10162, opened 2017-03-14
need to account for the allocated bytes accurately for the following cases

Maoni0:
portion in ac unused during a GC
when a min object is absorbed into an ac

### DIFF
--- a/src/gc/gc.cpp
+++ b/src/gc/gc.cpp
@@ -5792,9 +5792,10 @@ void gc_heap::fix_allocation_context (alloc_context* acontext, BOOL for_gc_p,
         alloc_contexts_used ++;
     }
 
-
     if (for_gc_p)
     {
+        // We need to update the alloc_bytes to reflect the portion that we have not used  
+        acontext->alloc_bytes -= (acontext->alloc_limit - acontext->alloc_ptr);  
         acontext->alloc_ptr = 0;
         acontext->alloc_limit = acontext->alloc_ptr;
     }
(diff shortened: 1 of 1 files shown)


Maoni0:
@jorive please let me know if you are still hitting problems. thanks!

jorive:
@Maoni0 This is working correctly now. Thank you! =)

swgillespie:
Could this be related to this corefx test failure? https://github.com/dotnet/corefx/issues/16606

Maoni0:
the osx test seems to have been failing for a long time so I will disregard it.

Maoni0:
@swgillespie @adityamandaleeka PTAL (I think I will also need your CR when I port this to 1.1.0)

src/gc/gc.cpp
```diff
@@ -11373,6 +11374,13 @@ void gc_heap::adjust_limit_clr (uint8_t* start, size_t limit_size,
         }
         acontext->alloc_ptr = start;
     }
+    else  
+    {  
+        // If the next alloc context is right up against the current one it means we are absorbing the min  
+        // object, so need to account for that.  
+        acontext->alloc_bytes += (start - acontext->alloc_limit);
```
adityamandaleeka: This is more of a question than a comment on this change, but it looks like this function is also used when we are allocating in the LOH (it's called from `a_fit_free_list_large_p`). If we're working in the LOH, shouldn't this code be modifying `alloc_bytes_loh` instead of `alloc_bytes`?

The managed API for getting allocated bytes seems to add `alloc_bytes_loh` and `alloc_bytes` together, so this wouldn't have any effect on that, but are we doing the right thing internally?

danmosemsft:
@dotnet-bot test OSX x64 Checked Build and Test please  (Sierra update)

danmosemsft:
@dotnet-bot test OSX x64 Checked Build and Test (Sierra update)

src/gc/gc.cpp
Maoni0: LOH ac's are tracked differently - for each large object allocation we just have a local ac on the stack and the size of the ac will just be what you asked for. so all we need to do is adding that number of bytes to alloc_bytes_loh.

adityamandaleeka: Got it. Yes, I see this in `allocate_large_object`.

adityamandaleeka:
LGTM

swgillespie:
LGTM

PR #10169, opened 2017-03-14
Fix tz rules

tarekgh:
On Linux when the caller ask for the TZ adjustment rules we convert the internal stored rule data from UTC to local time. we used to TZI.Convert which is wrong because calculating the start the daylight start time should always not include the daylight delta because we didn't start the daylight saving yet. and calculating the end of daylight should always include the daylight delta because we were in the daylight saving period.

The fix here is manually calculating the start and end of the daylight saving inside the rule

Fixes https://github.com/dotnet/corefx/issues/17028

### DIFF
--- a/src/mscorlib/src/System/TimeZoneInfo.Unix.cs
+++ b/src/mscorlib/src/System/TimeZoneInfo.Unix.cs
@@ -128,7 +128,7 @@ namespace System
                 return Array.Empty<AdjustmentRule>();
             }
 
-            // The rules we use in Unix cares mostly about the start and end dates but doesn't fill the transition start and end info.
+            // The rules we use in Unix care mostly about the start and end dates but don't fill the transition start and end info.
             // as the rules now is public, we should fill it properly so the caller doesn't have to know how we use it internally
             // and can use it as it is used in Windows
 
@@ -138,9 +138,13 @@ namespace System
             {
                 var rule = _adjustmentRules[i];
                 var start = rule.DateStart.Kind == DateTimeKind.Utc ?
+                            // At the daylight start we didn't start the daylight saving yet then we convert to Local time
+                            // by adding the _baseUtcOffset to the UTC time
                             new DateTime(rule.DateStart.Ticks + _baseUtcOffset.Ticks, DateTimeKind.Unspecified) :
                             rule.DateStart;
                 var end = rule.DateEnd.Kind == DateTimeKind.Utc ?
+                            // At the daylight saving end, the UTC time is mapped to local time which is already shifted by the daylight delta
+                            // we calculate the local time by adding _baseUtcOffset + DaylightDelta to the UTC time
                             new DateTime(rule.DateEnd.Ticks + _baseUtcOffset.Ticks + rule.DaylightDelta.Ticks, DateTimeKind.Unspecified) :
                             rule.DateEnd;
 
--- a/src/mscorlib/src/System/TimeZoneInfo.Unix.cs
+++ b/src/mscorlib/src/System/TimeZoneInfo.Unix.cs
@@ -128,8 +128,8 @@ namespace System
                 return Array.Empty<AdjustmentRule>();
             }
 
-            // The rules we use in Unix cares mostly about the start and end dates but doesn�t fill the transition start and end info.
-            // as the rules now is public, we should fill it properly so the caller doesn�t have to know how we use it internally
+            // The rules we use in Unix cares mostly about the start and end dates but doesn't fill the transition start and end info.
+            // as the rules now is public, we should fill it properly so the caller doesn't have to know how we use it internally
             // and can use it as it is used in Windows
 
             AdjustmentRule[] rules = new AdjustmentRule[_adjustmentRules.Length];
--- a/src/mscorlib/src/System/TimeZoneInfo.Unix.cs
+++ b/src/mscorlib/src/System/TimeZoneInfo.Unix.cs
@@ -128,8 +128,8 @@ namespace System
                 return Array.Empty<AdjustmentRule>();
             }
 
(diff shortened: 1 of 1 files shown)


src/mscorlib/src/System/TimeZoneInfo.Unix.cs
```diff
@@ -128,8 +128,8 @@ public AdjustmentRule[] GetAdjustmentRules()
                 return Array.Empty<AdjustmentRule>();
             }
 
-            // The rules we use in Unix cares mostly about the start and end dates but doesn�t fill the transition start and end info.
-            // as the rules now is public, we should fill it properly so the caller doesn�t have to know how we use it internally
+            // The rules we use in Unix cares mostly about the start and end dates but doesn't fill the transition start and end info.
```
danmosemsft: nit, "don't" as "rules" is plural of course

danmosemsft: nit "cares"->"care"

danmosemsft:
does this fix https://github.com/dotnet/corefx/issues/17036 and https://github.com/dotnet/corefx/issues/17028?

Do either of those also need "IsAmbiguousTime" check for when we move clocks forward?

src/mscorlib/src/System/TimeZoneInfo.Unix.cs
```diff
@@ -138,10 +138,10 @@ public AdjustmentRule[] GetAdjustmentRules()
             {
                 var rule = _adjustmentRules[i];
                 var start = rule.DateStart.Kind == DateTimeKind.Utc ?
-                            new DateTime(TimeZoneInfo.ConvertTime(rule.DateStart, this).Ticks, DateTimeKind.Unspecified) :
+                            new DateTime(rule.DateStart.Ticks + _baseUtcOffset.Ticks, DateTimeKind.Unspecified) :
```
danmosemsft: Is it worth a comment explaining why one has the daylightdelta and not the other? After all, that confused whoever wrote the original code.

tarekgh:
@danmosemsft this change is fixing dotnet/corefx#17028 as I mentioned in the comment above. I am going to look at dotnet/corefx#17036  next.

dotnet/corefx#17028 already using IsAmbiguousTime and should still use it to be reliable during running the test at daylight saving ends period. When I look at dotnet/corefx#17036  will tell what we need to do there,

src/mscorlib/src/System/TimeZoneInfo.Unix.cs
```diff
@@ -138,10 +138,10 @@ public AdjustmentRule[] GetAdjustmentRules()
...
                             rule.DateStart;
                 var end = rule.DateEnd.Kind == DateTimeKind.Utc ?
-                            new DateTime(TimeZoneInfo.ConvertTime(rule.DateEnd, this).Ticks - 1, DateTimeKind.Unspecified) :
+                            new DateTime(rule.DateEnd.Ticks + _baseUtcOffset.Ticks + rule.DaylightDelta.Ticks, DateTimeKind.Unspecified) :
```
danmosemsft: So this code works equally well for clock forward as well as clock back?

danmosemsft:
@tarekgh thanks, I misread your comment. LGTM although I'm not very knowledgeable about this code.

src/mscorlib/src/System/TimeZoneInfo.Unix.cs
tarekgh: yes. rule.DateEnd is UTC time which is mapping to the local time (with the daylight delta adjustment). 

for instance Pacific Time zone has _baseUtcOffset -08:00 hours and the daylight saving ends at 09:00 AM (which is equivalent to 2:00 AM when we having the daylight saving is on which is one hour delta) 

then the end time would be 09:00 - 08:00 + 01:00 which will give us 02:00 AM which is exactly the time the daylight saving end.

src/mscorlib/src/System/TimeZoneInfo.Unix.cs
```diff
@@ -138,10 +138,14 @@ public AdjustmentRule[] GetAdjustmentRules()
...
-                            new DateTime(TimeZoneInfo.ConvertTime(rule.DateStart, this).Ticks, DateTimeKind.Unspecified) :
+                            // At the daylight start we didn't start the daylight saving yet then we convert to Local time
+                            // by adding the _baseUtcOffset to the UTC time
+                            new DateTime(rule.DateStart.Ticks + _baseUtcOffset.Ticks, DateTimeKind.Unspecified) :
                             rule.DateStart;
                 var end = rule.DateEnd.Kind == DateTimeKind.Utc ?
-                            new DateTime(TimeZoneInfo.ConvertTime(rule.DateEnd, this).Ticks - 1, DateTimeKind.Unspecified) :
+                            // At the daylight saving end, the UTC time is mapped to local time which is already shifted by the daylight delta
+                            // we calculate the local time by adding _baseUtcOffset + DaylightDelta to the UTC time
+                            new DateTime(rule.DateEnd.Ticks + _baseUtcOffset.Ticks + rule.DaylightDelta.Ticks, DateTimeKind.Unspecified) :
```
eerhardt: Do you still need the `-1` here?

tarekgh: Actually no :-) from my testing I am seeing rule.DateEnd already subtracted the 1

eerhardt: perfect.  that's what I was hoping.

tarekgh:
Test OSX x64 Checked Build and Test please

tarekgh:
test Windows_NT x86 Checked Build and Test please

Issue #10076, opened 2017-03-10
JIT\Methodical\tailcall_v4\hijacking failed in CI


adityamandaleeka:
Just hit this on one of my PRs. It looks like this isn't a JIT issue, but a test/infrastructure issue. If I'm reading this output correctly, the test passed, but perhaps took longer than expected, so the harness thinks it timed out. 

https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_windows_nt_prtest/3876/
```
 BEGIN EXECUTION
16:04:31          "D:\j\workspace\debug_windows---17180f1b\tests\..\bin\tests\Windows_NT.x64.Debug\Tests\Core_Root\corerun.exe" hijacking.exe 
16:04:31         Executed 485279 GCs
16:04:31         Executed 3284875 tail calls
16:04:31         PASS
16:04:31         Expected: 100
16:04:31         Actual: 100
16:04:31         END EXECUTION - PASSED
16:04:31         PASSED
16:04:31         cmdLine:D:\j\workspace\debug_windows---17180f1b\bin\tests\Windows_NT.x64.Debug\JIT\Methodical\tailcall_v4\hijacking\hijacking.cmd Timed Out
16:04:31         Test Harness Exitcode is : -100
```

gkhanna79:
@RussKeldorph PTAL at the test and see if it should indeed take a long time or was this instance an outlier.

RussKeldorph:
@gkhanna79 I suspect this was an outlier, though I think we've seen a few tests that are exhibiting this behavior.  Possibly something else in the test pass that is oversubscribing the machine's resources and starving certain tests.  @pgavlin is looking into it.

PR #10171, opened 2017-03-14
Do not wait on std{out,err} when a test times out.

pgavlin:
The test wrapper currently waits on each of the process, stdout copy,
and stderr copy in turn. This makes it difficult to determine whether
or not the test did in fact time out, as the time in which it may
write to its output streams is effectively twice or thrice the
time used to determine success or failure (i.e. the process wait
timeout). This can cause situations like what we've seen in #10076,
where the harness reports that the process timed out but the test's
output indicates a successful run.

This change only waits on stdout/stderr if the process completes within
the time allotted; otherwise it cancels the copies and promptly reports
the timeout.

### DIFF
--- a/tests/src/Common/Coreclr.TestWrapper/CoreclrTestWrapperLib.cs
+++ b/tests/src/Common/Coreclr.TestWrapper/CoreclrTestWrapperLib.cs
@@ -71,23 +71,27 @@ namespace CoreclrTestLib
 
                 process.Start();
 
-                Task copyOutput = process.StandardOutput.BaseStream.CopyToAsync(outputStream);
-                Task copyError = process.StandardError.BaseStream.CopyToAsync(errorStream);
+                var cts = new CancellationTokenSource();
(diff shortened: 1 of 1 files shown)


pgavlin:
@jashook @wtgodbe @RussKeldorph PTAL

tests/src/Common/Coreclr.TestWrapper/CoreclrTestWrapperLib.cs
```diff
@@ -71,23 +71,27 @@ public int RunTest(string executable, string outputFile, string errorFile)
 
                 process.Start();
 
-                Task copyOutput = process.StandardOutput.BaseStream.CopyToAsync(outputStream);
-                Task copyError = process.StandardError.BaseStream.CopyToAsync(errorStream);
+                var cts = new CancellationTokenSource();
+                Task copyOutput = process.StandardOutput.BaseStream.CopyToAsync(outputStream, 4096, cts.Token);
```
RussKeldorph: MSDN says the default size is 81920.  Assuming the current default isn't 4096, any reason not to use the default?

PR #10172, opened 2017-03-14
JIT: preliminaries to improve types

AndyAyersMS:
Refactor the logic in `impDevirtualizeCall` so that the part that determines
the type of a tree for ref types is now a new utility method that can be called
elsewhere.

Update the utility to examine calls more closely. For inline candidates that
return shared types, try and use the context to get to the unshared version of the
type. For calls that are not inline candidates, look at the type in the signature
available to the jit w/o context.

Call the utility when we've created a temp for an inlinee's argument and the
argument is not modified in the inlinee body. If we already thought we knew the
type of the temp exactly, ensure that this new information agrees.

Rework the logic in `impDevirtualizeCall` in anticipation of interface call
devirtualization. Update the diagnostic stream to indicate the kind of call
devirtualized nd the primary reason why devirtualization happened.
Avoid fetching class and method names unless they're going to be used. Likewise
try not to fetch attributes if we already have them on hand.

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -22337,6 +22337,39 @@ GenTreePtr Compiler::fgInlinePrependStatements(InlineInfo* inlineInfo)
                 }
                 else
                 {
+                    // We're going to assign the argument value to the
+                    // temp we use for it in the inline body.
+                    //
+                    // If we know the argument's value can't be
+                    // changed within the method body, try and improve
+                    // the type of the temp.
+                    if (!inlArgInfo[argNum].argHasLdargaOp && !inlArgInfo[argNum].argHasStargOp)
+                    {
+                        GenTree*             argNode        = inlArgInfo[argNum].argNode;
+                        bool                 isExact        = false;
+                        bool                 isNonNull      = false;
+                        CORINFO_CLASS_HANDLE refClassHandle = gtGetClassHandle(argNode, &isExact, &isNonNull);
+
+                        if (refClassHandle != nullptr)
+                        {
+                            const unsigned tmpNum = inlArgInfo[argNum].argTmpNum;
+
+                            // If we already had an exact type for
+                            // this temp, this new information had
+                            // better agree with what we knew before.
+                            if (lvaTable[tmpNum].lvClassIsExact)
+                            {
+                                assert(isExact);
+                                assert(refClassHandle == lvaTable[tmpNum].lvClassHnd);
+                            }
+                            else
+                            {
+                                lvaTable[tmpNum].lvClassHnd     = refClassHandle;
+                                lvaTable[tmpNum].lvClassIsExact = isExact;
+                            }
+                        }
+                    }
+
                     /* Create the temp assignment for this argument */
 
                     CORINFO_CLASS_HANDLE structHnd = DUMMY_INIT(0);
--- a/src/jit/gentree.cpp
+++ b/src/jit/gentree.cpp
@@ -16334,6 +16334,207 @@ CORINFO_CLASS_HANDLE Compiler::gtGetStructHandle(GenTree* tree)
     return structHnd;
 }
 
+//------------------------------------------------------------------------
+// gtGetClassHandle: find class handle for a ref type
+//
+// Arguments:
+//    tree -- tree to find handle for
+//    isExact   [out] -- whether handle is exact type
+//    isNonNull [out] -- whether tree value is known not to be null
+//
+// Return Value:
+//    nullptr if class handle is unknown,
+//        otherwise the class handle.
+//    isExact set true if tree type is known to be exactly the handle type,
+//        otherwise actual type may be a subtype.
+//    isNonNull set truee if tree value is known not to be null,
+//        otherwise a null value is possible.
+
+CORINFO_CLASS_HANDLE Compiler::gtGetClassHandle(GenTreePtr tree, bool* isExact, bool* isNonNull)
+{
+    // Set default values for our out params.
+    *isNonNull                    = false;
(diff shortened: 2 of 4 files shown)


AndyAyersMS:
@JosephTremoulet PTAL
cc @dotnet/jit-contrib

Impacts about 80 methods in the jit-diff set. Has more impact when jitting, since many interesting shared methods are not inline candidates in R2R mode.

For instance when jitting we can now devirtualize a call whose object is the result the List<T>.get_Item.

src/jit/flowgraph.cpp
```diff
@@ -22337,6 +22337,39 @@ GenTreePtr Compiler::fgInlinePrependStatements(InlineInfo* inlineInfo)
...
+                        {
+                            const unsigned tmpNum = inlArgInfo[argNum].argTmpNum;
+
+                            // If we already had an exact type for
+                            // this temp, this new information had
+                            // better agree with what we knew before.
+                            if (lvaTable[tmpNum].lvClassIsExact)
+                            {
+                                assert(isExact);
+                                assert(refClassHandle == lvaTable[tmpNum].lvClassHnd);
```
JosephTremoulet: How do we know the new info won't be compatible but less precise than the old info?

src/jit/gentree.cpp
```diff
@@ -16334,6 +16334,207 @@ CORINFO_CLASS_HANDLE Compiler::gtGetStructHandle(GenTree* tree)
...
+//    tree -- tree to find handle for
+//    isExact   [out] -- whether handle is exact type
+//    isNonNull [out] -- whether tree value is known not to be null
+//
+// Return Value:
+//    nullptr if class handle is unknown,
+//        otherwise the class handle.
+//    isExact set true if tree type is known to be exactly the handle type,
+//        otherwise actual type may be a subtype.
+//    isNonNull set truee if tree value is known not to be null,
```
JosephTremoulet: typo: truee -> true

src/jit/importer.cpp
```diff
@@ -18498,34 +18431,55 @@ void Compiler::impDevirtualizeCall(GenTreeCall*            call,
...
+    // Is the call an interface call?
+    const bool isInterface = (baseClassAttribs & CORINFO_FLG_INTERFACE) != 0;
+
+    // See what we know about the type of 'this' in the call.
+    bool                 isExact      = false;
+    bool                 objIsNonNull = false;
+    CORINFO_CLASS_HANDLE objClass     = gtGetClassHandle(thisObj, &isExact, &objIsNonNull);
+
+    // Bail if we know nothing.
+    if (objClass == nullptr)
```
JosephTremoulet: How common is this case?  Worth moving the check up to before fetching `baseClass`/`baseClassAttribs`?

src/jit/importer.cpp
```diff
@@ -18376,110 +18387,33 @@ void Compiler::impDevirtualizeCall(GenTreeCall*            call,
...
 
 #if defined(DEBUG)
-
     // Bail if devirt is disabled.
     if (JitConfig.JitEnableDevirtualization() == 0)
     {
         return;
     }
 
+    const bool doPrint = JitConfig.JitPrintDevirtualizedMethods() == 1;
```
JosephTremoulet: I'm confused what you're trying to achieve with `doPrint` in release builds here... looks like you only fill in the info strings under `#if DEBUG`, so release would just print the placeholders and the only info you really get is the total number of devirtualized calls per kind, right?  If we're going to want that information from release builds, would it be better to fire an ETW event or something?

src/jit/flowgraph.cpp
AndyAyersMS: This is a bit of future proofing.

Currently the argument temps -- created over in `impInlineFetchArg` -- don't get any class/exact info. If/when they do, we will type them based on the method signature.

Any correct program will pass an argument value with the signature type or some subtype. So the argument type should be as good or better a bound on the type than the signature type.

In the update case, it might be worth checking for proper subtype relationship here, though if we somehow "downgrade" the type to a superclass, we still have a safeguard in the VM side of devirtualization, which also does this check.

src/jit/importer.cpp
AndyAyersMS: Makes sense to reorder and check for objType first, since we only get a type maybe 50% of the time. Will update.

src/jit/importer.cpp
AndyAyersMS: `doPrint` only exists in `DEBUG`. There is no release mode diagnostic stream for devirtualization.

Not sure if devirtualization warrants an VM/ETW callback/event. Inlining needs it because other runtime subsystems are sensitive to inlining -- for instance rejitting needs to know what methods have been inlined.

If we do add a notification mechanism it will traffic only in handles and string literals to avoid any need for dynamic string allocation.

JosephTremoulet: Right, thanks, I was just misreading the code there, I like it how you have it.

AndyAyersMS:
Pushed an update that fixes the typo and reorders the checks -- also fixed a dump mode AV and tweaked dump mode formatting a bit.

AndyAyersMS:
Windows debug timed out in one of the [chaos tests](https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_windows_nt_prtest/4072/testReport/junit/JIT_Generics/_Coverage_chaos65204782cs_o_chaos65204782cs_o_/_Coverage_chaos65204782cs_o_chaos65204782cs_o_cmd/) -- can't repro locally, so am going to retry.

AndyAyersMS:
OSX failure is a known issue, so will ignore it.

PR #10174, opened 2017-03-14
Add support for testing via ILLINK

swaroop-sridhar:
This commit has the following changes:
1) Setup the testing infrastructure to run the CoreCLR tests after linking through ILLinker 
2) The linker is taken as an argument for runtest.cmd, typically the corebuild of 
    https://github.com/mono/linker
3) Tests are run by linking the test binaries and all libraries except System.Private.Corelib
4) For tests that use reflection, setup Reflection roots in source directories as _reflect.xml


swaroop-sridhar:
@sbomer please review
CC @erozenfeld

tests/src/CLRTest.Execute.Batch.targets
```diff
@@ -213,21 +217,83 @@ IF NOT "%CLRCustomTestLauncher%"=="" (
...
       ]]></BatchCLRTestLaunchCmds>
       <BatchCLRTestLaunchCmds Condition="'$(CLRTestKind)' == 'BuildAndRun'"><![CDATA[
+
+set ExePath=$(InputAssemblyName)
+set LinkBin=__Link
+set Assemblies=-a System.Private.CoreLib
+set ReflectionRoots=
+
+IF defined DoLink ( 
+    IF NOT EXIST !ILLINK! (
```
sbomer: Would it be possible to move the linker-specific logic to its own section, something like IllinkBatchScript, instead of putting it in BatchCLRTestLaunchCmds? You could have BatchCLRTestLaunchCmds check ExePath and set it to $(InputAssemblyName) if none of the previous steps have modified the ExePath.

swaroop-sridhar:
@sbomer: Yes, runtest.sh will be changed in a separate PR.
I'll do the refactoring you've suggested. Thanks.

tests/src/CLRTest.Execute.Batch.targets
swaroop-sridhar: I've moved the linker testing and cleanup code to their own sections.
But unlinke ilasm/ildasm these sections will always be invoked.

The decision of whether to run via linker or not is not static, but done at runtime, since we don't want to rebuild the tests to tests with/without illink.

tests/src/dir.targets
```diff
@@ -184,4 +184,9 @@
 
   </Target>
 
+  <Target Name="CopyReflectRoots">
```
erozenfeld: Nit: Can you change these to CopyReflectionRoots?

jashook:
@dotnet-bot test CI Please

swaroop-sridhar:
@dotnet-bot  test Windows_NT x64 Debug Build and Test

PR #10177, opened 2017-03-14
Support bracket enclosed names in TZ POSIX rules

eerhardt:
The POSIX time zone rules at the end of a tzfile have been updated to allow digits, plus signs and minus signs in the time zone names by enclosing the name in `<` and `>` characters. TimeZoneInfo's parsing logic was written using an older version of the spec, so it didn't have support for this format.

Fix https://github.com/dotnet/corefx/issues/16962

@tarekgh

### DIFF
--- a/src/mscorlib/src/System/TimeZoneInfo.Unix.cs
+++ b/src/mscorlib/src/System/TimeZoneInfo.Unix.cs
@@ -927,7 +927,7 @@ namespace System
         /// Creates an AdjustmentRule given the POSIX TZ environment variable string.
         /// </summary>
         /// <remarks>
-        /// See http://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html for the format and semantics of this POSX string.
+        /// See http://man7.org/linux/man-pages/man3/tzset.3.html for the format and semantics of this POSX string.
         /// </remarks>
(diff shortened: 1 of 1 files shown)


tarekgh:
LGTM. thanks @eerhardt

eerhardt:
test OSX x64 Checked Build and Test

src/mscorlib/src/System/TimeZoneInfo.Unix.cs
```diff
@@ -1189,8 +1189,29 @@ private static bool TZif_ParseMDateRule(string dateRule, out int month, out int
...
+            bool isBracketEnclosed = index < posixFormat.Length && posixFormat[index] == '<';
+            if (isBracketEnclosed)
+            {
+                // move past the opening bracket
+                index++;
+                
+                string result = TZif_ParsePosixString(posixFormat, ref index, c => c == '>');
+                
+                // move past the closing bracket
+                index++;
```
stephentoub: Do we need to verify that it is in fact a closing brace? Is it possible with malformed input we might get here and increment even if there isn't one?

eerhardt: The only case where TZif_ParsePosixString returns is if the "breakCondition" func returns true or it gets to the end of the string.  If it hits the end of the string, incrementing index one more past the length won't cause problems. But just to be safe I can add the check that the current char is the closing brace before incrementing.

tarekgh:
test Windows_NT x64 Debug Build and Test please

eerhardt:

PR #10173, opened 2017-03-14
Change CoreCLR to use Sierra

mmitche:

PR #10178, opened 2017-03-14
Move System.Convert code to shared partition

alexperovich:


src/mscorlib/shared/System/Convert.cs
```diff
@@ -258,79 +248,74 @@ public static Object ChangeType(Object value, TypeCode typeCode, IFormatProvider
...
+            Debug.Assert(value != null, "[Convert.DefaultToType]value!=null");
             if (targetType == null)
             {
                 throw new ArgumentNullException(nameof(targetType));
             }
             Contract.EndContractBlock();
 
-            RuntimeType rtTargetType = targetType as RuntimeType;
-
-            if (rtTargetType != null)
```
jkotas: This does not do the same thing as what it did before. The operator == is overloaded for Type, and so this will turn into more expensive call than what used to be here.

It probably does the same thing as before, except that it is quite a bit slower.

@AtsushiKan Any thoughts on this?

alexperovich: I needed to do this because corert doesn't have RuntimeType. What could I use that would not loose the perf and work in both corert and coreclr?

jkotas: Actually, it is using the slower overloaded operator== already, so this change is not making it worse.

alexperovich: Would using `ReferenceEquals` be faster and still correct?

jkotas: Yes, `ReferenceEquals` would be faster, generate smaller native code and work correctly. 

It is depending on internal implementation detail that there is just one instance of RuntimeType for given type. There is a ton of code out depending on this detail already and thus we are not ever going to be able to change it, so taking one more dependency on it should not be a problem.

alexperovich:

alexperovich:

alexperovich:
@jkotas who owns the CI for coreclr? This failure doesn't make any sense.

jkotas:
The CI failure is know - getting fixed in https://github.com/dotnet/coreclr/pull/10173

PR #10180, opened 2017-03-14
Add IBC support to managed build

dpodder:
Add a new build switch, `ibcinstrument`, that adds `/Tuning` on the `crossgen` command line when building System.Private.CoreLib and its peers. Automatically consume IBC optdata during builds when these conditions are met:
1. `ibcinstrument` is *not* passed to the build,
2. optdata is available
3. ibcmerge is available

Note that `optdata` will not yet be restored with this change; once packages for master are made available, a new package reference will still need to be added. This PR attempts to unblock manually testing IBC, as well as surrounding CI/infrastructure work.


dpodder:
@janvorli @adiaaida PTAL
/cc @dagood @briansull @dotnet/rap-team @weshaggard @joperezr @AlexGhiondea

adiaaida:
Two questions: where does this work assume ibcmerge will be installed (on the path? Some known directory?)? And do we have a version of ibcmerge that runs on linux to make the build.sh changes necessary? I understand putting in the support, but if we can't actually run ibc on linux, we should probably make ibcinstrument fail early/be ignored on linux.

adiaaida:
Also, can you include a sample project.json in the tests directory (maybe?) that can be copied into optdata/project.json so one doesn't have to be generated on the fly?

dpodder:
@adiaaida ibcmerge is installed through BuildTools, and is available from within the context of msbuild. For CI, if we need to expose it to non-MSBuild based tools, the easiest thing I can think of is adding a new msbuild target so that you could read the output of, say, `dotnet msbuild /t:FindBuildTool /p:tool=ibcmerge.exe` or something like that?

ibcmerge does not currently work on Linux; the work is planned, but for now it requires the desktop framework. We don't want `build.sh` to fail early because it actually works; `/Tuning` doesn't require ibcmerge to work just to produce instrumented images.

Re sample.json: great idea, I'll add one.

adiaaida:
Ok, so if we're just calling build.cmd, it knows where to find it, assuming it's been installed through BuildTools

weshaggard:
@dpodder @dagood How is this and https://github.com/dotnet/corefx/pull/17053 connected? I would expect for CoreCLR to consume IBCMerge it would need a similar set of changes to get all the necessary pieces restored. It doesn't automatically happen as part of consuming BuildTools unless it is correctly configured.

dpodder:
@adiaaida Yes, that's correct.

dpodder:
@weshaggard Good point, the three `OptionalToolSource*` parameters need to be fed to msbuild. In my testing I've been setting local environment variables; I wasn't sure of the best way to consume secret parameter values. I can add them as extra passthrough parameters to mirror CoreFX, if that's the preferred approach (environment vars would still work locally, since msbuild can read them).

weshaggard:
>environment vars would still work locally, since msbuild can read them

Depending on environment variables is almost always a bad idea as there are various reliability, discoverability, and security concerns with doing so. I would suggest either we pass them as "extra parameters" that just flow through to msbuild when we call build. I wouldn't plumb a ton of extra arguments into build.cmd/sh as there are already way too many IMO.

dagood:
@weshaggard @dpodder About relation to https://github.com/dotnet/corefx/pull/17053: in CoreFX, only the build defs needed changes (and a CoreFX-specific Tools-Override update). There aren't any similar changes in CoreCLR because the defs aren't checked in here, but the "live" build definitions will need to be changed in the same way. (cc @wtgodbe) 

The `OptionalToolSource*` props are only intended to be passed in to `sync`, so I wouldn't think you'd need to pass them through `build.cmd/sh`. (The VSTS builds call `sync` first.) In fact, I'd recommend calling the VSTS vars `PB_OptionalToolSource*` and passing them in manually to make sure it's working as intended rather than through the environment.

dpodder:
@adiaaida test script added, PTAL.
@weshaggard @dagood Sounds good, thanks. I'll make the change and push an update when ready.

tests/scripts/optdata/bootstrap.py
```diff
@@ -0,0 +1,82 @@
...
+    """Returns the Build_OS component used by the build system."""
+    if os.name == 'nt':
+        return 'Windows_NT'
+    else:
+        sysname = os.uname()[0]
+        return 'OSX' if sysname.casefold() == 'Darwin'.casefold() else sysname
+
+def get_optdata_version(tool):
+    """Returns the version string specified in project.json for the given tool."""
+    package_name = 'optimization.{0}.CoreCLR'.format(tool)
```
adiaaida: I don't really have a preference on this, but in other places, we've done this as

package_name = 'optimization.%s.CoreCLR' % (tool)

Likewise below

package_name = 'optimization.%s-%s.%s.CoreCLR' % (get_buildos(), arch, tool)

tests/scripts/optdata/bootstrap.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python
```
adiaaida: Does this script basically just copy the project.json file into the appropriate directory in src/.nuget?

tests/scripts/optdata/bootstrap.py
```diff
@@ -0,0 +1,82 @@
...
+ARCH_LIST = ['x64', 'x86']
+TOOL_LIST = ['IBC', 'PGO']
+
+def get_buildos():
+    """Returns the Build_OS component used by the build system."""
+    if os.name == 'nt':
+        return 'Windows_NT'
+    else:
+        sysname = os.uname()[0]
+        return 'OSX' if sysname.casefold() == 'Darwin'.casefold() else sysname
```
adiaaida: Can we just use sysname.lower() here? I know casefold is more aggressive than lower, but since all we are comparing to is a lowercase 'Darwin', I don't think we need that aggressive of a a lower, and casefold doesn't work in Python2 without importing extra libraries. If possible, I'd like this script to be as compatible with both Python2 and Python3 as possible since the main dependency in CoreCLR is on Python2.

tests/scripts/optdata/bootstrap.py
```diff
@@ -0,0 +1,82 @@
...
+
+    if not path.exists(path.dirname(TARGET_FILE)):
+        os.makedirs(path.dirname(TARGET_FILE))
+    shutil.copyfile(ORIGIN_FILE, TARGET_FILE)
+
+    print("Bootstrapping optdata is complete.")
+    for tool in TOOL_LIST:
+        for arch in ARCH_LIST:
+            optdata_dir = get_optdata_dir(tool, arch)
+            print("  * Copy {a} {t} files into: {d}".format(a=arch, t=tool, d=optdata_dir))
```
adiaaida: What is actually happening here? Is this just a suggestion to the user where to copy their data to?

tests/scripts/optdata/bootstrap.py
```diff
@@ -0,0 +1,82 @@
...
+from os import path
+import shutil
+import subprocess
+import sys
+
+# Display the docstring if the user passes -h|--help
+argparse.ArgumentParser(description=__doc__).parse_args()
+
+SCRIPT_ROOT = path.dirname(path.realpath(__file__))
+REPO_ROOT = path.realpath(path.join(SCRIPT_ROOT, '..', '..', '..'))
```
adiaaida: I'd be more comfortable with the user being required to pass the repo_root, on the off chance that this file gets moved, or something about the repo structure changes. I doubt it would, but it's something to guard against.

tests/scripts/optdata/bootstrap.py
dpodder: Sounds good, I like consistency. I'll change it.

tests/scripts/optdata/bootstrap.py
dpodder: That's one small part of it--if that was all I would have probably just written cmd/sh scripts. The bigger part is building up and printing the path(s) where optdata (PGO and/or IBC) needs to get copied for it to be consumed. This is harder than the project.json itself, and figuring it out was much easier in python than a shell script.

tests/scripts/optdata/bootstrap.py
dpodder: Yes, good catch; I didn't realize `casefold()` required extra packages on python2. I'll just use `lower()`.

tests/scripts/optdata/bootstrap.py
dpodder: Yes, and this is what is nontrivial--see my reply to the comment earlier.

tests/scripts/optdata/bootstrap.py
dpodder: Hmm, good point. How about if I added some assertions like make sure ROOT\src\.nuget exists where we think it should? I also would like to keep the script easy to use, and counting the number of `..`s needed is even more error-prone IMO.

config.json
```diff
@@ -186,6 +186,18 @@
...
+      "valueType": "property",
+      "values": [ true, false ],
+      "defaultValue": false
+    },
+    "OptimizationDataDir": {
+      "description": "Sets the path where the build should look for IBC profile data.",
+      "valueType": "property",
+      "values": [],
+      "defaultValue": ""
+    },
```
dagood: I'm not opposed to this if it makes things easier for you, but I haven't been updating/using config.jsons for new features so I don't accrue maintenance costs. The plan is to (somehow) make these common properties shared: https://github.com/dotnet/buildtools/issues/1085.

adiaaida:
This looks good to me.

config.json
dpodder: Thanks - I really like the idea of inheriting shared properties such as these from a common source. From the issue though it looks like this support isn't there yet? I'd like to move forward with the current implementation for now. We should be able to remove these entries from the local config.json later once they're implemented in a shared config, right?

dagood: That's correct: no support, and we can remove them later.

My point is just that I tend to avoid adding them in the first place (using the ` -- /p:Foo=bar` syntax instead) to avoid the extra maintenance cost. I'm fine leaving these in the PR, just a note.

dpodder: Ah, I see what you mean. I originally tried the `-- /p:Foo=bar` syntax too, but that turned out to be too fragile; there can only be one `--`, and it makes the arguments even more sensitive to ordering. One of the early CI failures was caused by this (it was trying to pass `-rebuild` to `run.exe` on the command line, after I had already fed it `--` followed by msbuild args). I'd much rather keep it in its current form until the shared configs are ready.

config.json
```diff
@@ -342,6 +354,24 @@
       "values": [],
       "defaultValue": ""
     },
+    "OptionalToolSource": {
```
weshaggard: Do you actually need these here? Do they work with if you pass them after "--"? I ask because I don't see them being passed through the scripts.

dpodder: These are the same as the `EnableProfileGuidedOptimization` and `OptimizationDataDir` properties; the only difference is they are secrets, so they will need to be passed in from CI/VSTS or from users manually. VSTS support requires changes to not-checked-in pipeline definitions. CI support is coming in later PR(s). I want to go ahead with this PR before the CI changes are ready so that I can unblock manual IBC builds, which other work is dependent upon.

They *can* be passed directly to msbuild after `--`, but as I also noted above, that is much more fragile. No other parameters currently used in CI rely on `--`, which I think is a good thing. I'd really much rather add these settings to config.json for now, and remove them together with the non-secret ones once they're available in a shared config.

weshaggard: Sounds good. 

As an assign I really think the amount of argument parsing that is happening in these build scripts is way out of control which is at least part of the trouble with the extra parameter fragility.

dpodder:
Thanks for the detailed feedback @adiaaida @dagood @weshaggard!

PR #10182, opened 2017-03-14
Remove potential overflow in SafeBSTRHandle

botaberg:
I was working on porting SafeBSTRHandle to CoreRT and while in code review, it was noted that SafeBSTRHandle.Allocate has an opportunity for integer overflow which would be left unhandled. I fixed it in CoreRT, and figured I'd offer up the fix here as well.

This PR changes the type of the local variable lenInBytes from uint to ulong. This avoids the integer overflow and works out well because bstr.Initialize, to which lenInBytes is given as an argument, takes an argument of type ulong.

### DIFF
--- a/src/mscorlib/src/System/Security/SafeBSTRHandle.cs
+++ b/src/mscorlib/src/System/Security/SafeBSTRHandle.cs
@@ -13,7 +13,7 @@ namespace System.Security
 
         internal static SafeBSTRHandle Allocate(uint lenInChars)
         {
-            ulong lenInBytes = lenInChars * sizeof(char);
+            ulong lenInBytes = (ulong)lenInChars * sizeof(char);
             SafeBSTRHandle bstr = Interop.OleAut32.SysAllocStringLen(IntPtr.Zero, lenInChars);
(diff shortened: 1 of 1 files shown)


danmosemsft:
@dotnet-bot test this (OSX was broken earlier)

danmosemsft:
@dotnet-bot test this please (politer and more functional)

src/mscorlib/src/System/Security/SafeBSTRHandle.cs
```diff
@@ -13,7 +13,7 @@ internal sealed class SafeBSTRHandle : SafeBuffer
 
         internal static SafeBSTRHandle Allocate(uint lenInChars)
         {
-            uint lenInBytes = lenInChars * sizeof(char);
+            ulong lenInBytes = lenInChars * sizeof(char);
```
jkotas: This is not fixing anything. The multiplication will be performed as uint, and only then the result gets cast to ulong. So if there is an integer overflow during the multiplication, you will get the exact same result as before. Instead, you would need to add the cast to at least one of the multiplicants.

PR #10186, opened 2017-03-15
[x86/Linux] Invoke gcResetForBB() in genFuncletProlog

wateret:
A trivial fix.

Looks like this code is missing. It is present in all architectures other than x86.

cc: @jkotas @janvorli @parjong


danmosemsft:
@dotnet-bot test OSX x64 Checked Build and Test (Sierra update)

PR #10188, opened 2017-03-15
[x86/Linux] Fix crash when GC triggered inside funclets

wateret:
cc: @jkotas @janvorli @parjong

Related Issue : #10187

### DIFF
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -4163,6 +4163,14 @@ bool EECodeManager::EnumGcRefs( PREGDISPLAY     pContext,
         GC_NOTRIGGER;
     } CONTRACTL_END;
 
+#ifdef WIN64EXCEPTIONS
+    if (flags & ParentOfFuncletStackFrame)
+    {
+        LOG((LF_GCROOTS, LL_INFO100000, "Not reporting this frame because it was already reported via another funclet.\n"));
+        return true;
+    }
+#endif // WIN64EXCEPTIONS
+
     GCInfoToken gcInfoToken = pCodeInfo->GetGCInfoToken();
     unsigned  curOffs = pCodeInfo->GetRelOffset();
 
@@ -4586,10 +4594,6 @@ bool EECodeManager::EnumGcRefs( PREGDISPLAY     pContext,
     unsigned ptrAddr;
     unsigned lowBits;
 
-#ifdef WIN64EXCEPTIONS
-    if (!pCodeInfo->IsFunclet())
-#endif
-    {
 
     /* Process the untracked frame variable table */
 
@@ -4647,8 +4651,6 @@ bool EECodeManager::EnumGcRefs( PREGDISPLAY     pContext,
                   DAC_ARG(DacSlotLocation(info.ebpFrame ? REGI_EBP : REGI_ESP,
                                           info.ebpFrame ? EBP - ptrAddr : ptrAddr - ESP,
                                           true)));
-    }
-
     }
 
 #if VERIFY_GC_TABLES
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -4249,6 +4249,18 @@ bool EECodeManager::EnumGcRefs( PREGDISPLAY     pContext,
 
 #endif // _DEBUG
 
+#ifdef WIN64EXCEPTIONS   // funclets
+    //
+    // If we're in a funclet, we do not want to report the incoming varargs.  This is
+    // taken care of by the parent method and the funclet should access those arguments
+    // by way of the parent method's stack frame.
+    //
+    if(pCodeInfo->IsFunclet())
+    {
+        return true;
+    }
+#endif // WIN64EXCEPTIONS
+
     /* What kind of a frame is this ? */
 
     FrameType   frameType = FR_NORMAL;
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -4249,18 +4249,6 @@ bool EECodeManager::EnumGcRefs( PREGDISPLAY     pContext,
 
 #endif // _DEBUG
 
-#ifdef WIN64EXCEPTIONS   // funclets
-    //
-    // If we're in a funclet, we do not want to report the incoming varargs.  This is
-    // taken care of by the parent method and the funclet should access those arguments
-    // by way of the parent method's stack frame.
-    //
-    if(pCodeInfo->IsFunclet())
-    {
-        return true;
-    }
-#endif // WIN64EXCEPTIONS
-
     /* What kind of a frame is this ? */
 
     FrameType   frameType = FR_NORMAL;
@@ -4598,6 +4586,10 @@ bool EECodeManager::EnumGcRefs( PREGDISPLAY     pContext,
     unsigned ptrAddr;
     unsigned lowBits;
 
+#ifdef WIN64EXCEPTIONS
+    if (!pCodeInfo->IsFunclet())
+#endif
+    {
 
     /* Process the untracked frame variable table */
 
@@ -4655,6 +4647,8 @@ bool EECodeManager::EnumGcRefs( PREGDISPLAY     pContext,
                   DAC_ARG(DacSlotLocation(info.ebpFrame ? REGI_EBP : REGI_ESP,
                                           info.ebpFrame ? EBP - ptrAddr : ptrAddr - ESP,
                                           true)));
+    }
+
     }
 
 #if VERIFY_GC_TABLES
@@ -4763,6 +4757,18 @@ bool EECodeManager::EnumGcRefs( PREGDISPLAY     pContext,
 #if VERIFY_GC_TABLES
     _ASSERTE(*castto(table, unsigned short *)++ == 0xBABE);
 #endif
+
+#ifdef WIN64EXCEPTIONS   // funclets
+    //
+    // If we're in a funclet, we do not want to report the incoming varargs.  This is
+    // taken care of by the parent method and the funclet should access those arguments
+    // by way of the parent method's stack frame.
+    //
+    if(pCodeInfo->IsFunclet())
+    {
+        return true;
+    }
+#endif // WIN64EXCEPTIONS
 
     /* Are we a varargs function, if so we have to report all args
        except 'this' (note that the GC tables created by the x86 jit


wateret:
Current patch makes those two TCs pass, I just copied the added code from same code of USE_GC_INFO_DECODER on.

Here are a list that I'm not sure about.
* Currently funclets has nothing on its stack so there is nothing that can be a GC root. If so, we still have REG_EXCEPTION_OBJECT(eax), but I'm not sure that it is handled.
* Not sure if the added code is at the right place in EnumGcRefs().
* Why does not X86/Linux use GcInfoDecoder?

src/vm/eetwain.cpp
```diff
@@ -4249,6 +4249,18 @@ bool EECodeManager::EnumGcRefs( PREGDISPLAY     pContext,
 
 #endif // _DEBUG
 
+#ifdef WIN64EXCEPTIONS   // funclets
+    //
+    // If we're in a funclet, we do not want to report the incoming varargs.  This is
```
jkotas: This is skipping way more than just the incoming varargs

jkotas:
> Why does not X86/Linux use GcInfoDecoder?

In the early days, CLR was written just for x86 and the GC info encoding was very x86 specific. When we have started porting to other platforms, portable GcInfoDecoder was introduced but x86 was never switched to it because of the GcInfoDecoder would not be as efficient on x86 and it is work to switch it.

src/vm/eetwain.cpp
wateret: It crashes while [Process the untracked frame variable table](https://github.com/dotnet/coreclr/blob/bcc0a1d528b78f5e05bf7973090f33ef2403e685/src/vm/eetwain.cpp#L4590-L4646). Would funclets need to do something with these untracked variables? I thought they are handled by main function frame.

By the way does untracked frame variable mean ref-type local variables?

wateret: cc: @jkotas @janvorli

wateret:
@jkotas I updated the commit to skip untracked frame variables for funclets. Still not sure if this is a right way though.

jkotas:
The delta looks good to me now. Are you doing any GC stress testing on Linux x86? It is the best way to verify whether the GC reporting works correctly.

parjong:
@jkotas I once tried, but there were too many stack alignment issues. I'm currently working on fixing them.

wateret:
@jkotas The current patch only skips `Process the untracked frame variable table` part, but it still do `Process the frame variable lifetime table` for funclets. I'm not sure we need this for funclet.
I also thought this is a bit strange because we need to (at least) move `table` by skipping untracked table so that we can correctly process variable lifetime table.

jkotas:
> I also thought this is a bit strange because we need to (at least) move table by skipping untracked table

Agree. This looks like a problem.

src/vm/eetwain.cpp
janvorli: @wateret I believe the only difference between untracked and tracked variables is that the untracked are assumed to be alive in the whole function while the tracked ones liveness is tracked precisely (by instruction address ranges). And only locals containing references or byrefs are reported. @BruceForstall please correct me if I am wrong.

BruceForstall: @janvorli that's correct

wateret:
@jkotas Fixed that and the indentation.

janvorli:
@wateret, @jkotas so is the current state of the PR correct? Looking at what we do for the target architectures where we use GCInfoDecoder, we only skip reporting untracked variables for filter funclets there. The comment in that code says:
```
        // Filters are the only funclet that run during the 1st pass, and must have
        // both the leaf and the parent frame reported.  In order to avoid double
        // reporting of the untracked variables, do not report them for the filter.
```
But this code skips reporting untracked variables for all funclets.
Thinking about it, it seems that the same rule that the comment mentions should apply to the non-exceptionally invoked finallys (both their frame and all the frames upto the containing function frame are alive on the stack), which would cover the issue this PR is fixing. 
@jkotas are the non-exceptionally called finallys always inlined on the other architectures so that they don't require the special treatment or am I missing something?

jkotas:
> @jkotas are the non-exceptionally called finallys always inlined on the other architectures

No, the non-exceptional finallys are called via a call on the other architectures as well.

The reporting here needs to be in sync with what the JIT does. The best way to verify whether it is in sync is to run GC stress.

parjong:
@jkotas @janvorli @wateret I found that this PR is necessary to run 'Hello, World' example under GCStress=3. Without this, CLR hits the following assert failure:
```
TID 7f35: STACKWALK    TransitionFrame::UpdateRegDisplay(ip:F25D141F, sp:FFFFC3B0)
TID 7f35: STACKWALK: [002] CONSIDER: FRAMELESS: PC= f25d141f  SP= ffffc3b0  method=IL_STUB_PInvoke [funclet]
TID 7f35: STACKWALK: Found Non-Filter funclet @ SP: FFFFC3B0, m_crawl.pFunc = F6338D34; FuncletParentCallerSP: FFFFC430
TID 7f35: STACKWALK: [002] CALLBACK: FRAMELESS: PC= f25d141f  SP= ffffc3b0  method=IL_STUB_PInvoke [funclet]
TID 7f35: Scanning Frameless method F6338D34M EIP = F25D141F &EIP = FFFFC3AC
TID 7f35: Scanning Frame for method DomainBoundILStubClass:IL_STUB_PInvoke

Assert failure(PID 32565 [0x00007f35], Thread: 32565 [0x7f35]): !CREATE_CHECK_STRING(pMT && pMT->Validate())
    File: /home/parjong/projects/dotnet/coreclr/src/vm/object.cpp Line: 1719
```

wateret:
> Looking at what we do for the target architectures where we use GCInfoDecoder, we only skip reporting untracked variables for filter funclets there.

@janvorli Oh I see. That means we shouldn't skip untracked variables. However wouldn't we double report 2nd pass funclets as well as filters? I thought as long as we have a funclet frame, we have main function's frame upper in stack.

wateret:
Last part of output with GC logs. Backtrace is in #10187

```
TID 0008:     IGCHeap::Promote: Promote GC Root *FFFFC9F0 = F2E1757C MT = F63B81BCT
TID 0008: STACKWALK: [007] CONSIDER: FRAMELESS: PC= f4f6c379  SP= ffffca10  method=WriteLine
TID 0008: STACKWALK: [007] CALLBACK: FRAMELESS: PC= f4f6c379  SP= ffffca10  method=WriteLine
TID 0008: Scanning Frameless method F255E434M EIP = F4F6C379 &EIP = FFFFC93C
TID 0008: Scanning Frame for method System.IO.SyncTextWriter:WriteLine
TID 0008:     IGCHeap::Promote: Promote GC Root *FFFFCA1C = F2E17654 MT = F255E590T
TID 0008: STACKWALK: [008] CONSIDER: FRAMELESS: PC= f4f6c338  SP= ffffca30  method=WriteLine
TID 0008: STACKWALK: [008] CALLBACK: FRAMELESS: PC= f4f6c338  SP= ffffca30  method=WriteLine
TID 0008: Scanning Frameless method F63B5ED8M EIP = F4F6C338 &EIP = FFFFC93C
TID 0008: Scanning Frame for method System.Console:WriteLine
TID 0008: STACKWALK: [009] CONSIDER: FRAMELESS: PC= f4f6ca1f  SP= ffffca40  method=foo                                                 [0/9812]
TID 0008: STACKWALK: [009] CALLBACK: FRAMELESS: PC= f4f6ca1f  SP= ffffca40  method=foo
TID 0008: Scanning Frameless method F63B25A8M EIP = F4F6CA1F &EIP = FFFFC93C
TID 0008: Scanning Frame for method Class1:foo
    Frame  local at [EBP-0CH]:     Untracked  local at [EBP-1CH]:
TID 0008:     IGCHeap::Promote: Promote GC Root *FFFFCA9C = F31991E0 MT = F6348384T
    Untracked  local at [EBP-20H]:
    Untracked  local at [EBP-24H]:
    Untracked  local at [EBP-28H]:
    Untracked  local at [EBP-2CH]:
    Untracked  local at [EBP-30H]:
    Untracked  local at [EBP-34H]:
    Untracked  local at [EBP-38H]:
    Untracked  local at [EBP-3CH]:
    Untracked  local at [EBP-40H]:
    Untracked  local at [EBP-44H]:
    Untracked  local at [EBP-48H]:
    Untracked  local at [EBP-4CH]:
    Untracked  local at [EBP-50H]:
    Untracked  local at [EBP-54H]:
    Untracked  local at [EBP-58H]:
    Untracked  local at [EBP-5CH]:
    Untracked  local at [EBP-60H]:
    Untracked  local at [EBP-64H]:
    Untracked  local at [EBP-68H]:
    Untracked  local at [EBP-6CH]:
    Untracked  local at [EBP-70H]:
TID 0008:     IGCHeap::Promote: Promote GC Root *FFFFCA48 = F2E17750 MT = F59E3238T
    Untracked  local at [EBP-74H]:
TID 0008:     IGCHeap::Promote: Promote GC Root *FFFFCA44 = F31991E0 MT = F6348384T
TID 0008: STACKWALK: [00a] CONSIDER: FRAMELESS: PC= f4f6ce1a  SP= ffffcac0  method=foo [funclet]
TID 0008: STACKWALK: Found Non-Filter funclet @ SP: FFFFCAC0, m_crawl.pFunc = F63B25A8; FuncletParentCallerSP: FFFFCB50
TID 0008: STACKWALK: [00a] CALLBACK: FRAMELESS: PC= f4f6ce1a  SP= ffffcac0  method=foo [funclet]
TID 0008: Scanning Frameless method F63B25A8M EIP = F4F6CE1A &EIP = FFFFC93C
TID 0008: Scanning Frame for method Class1:foo
    Untracked  local at [EBP-1CH]:

Program received signal SIGSEGV, Segmentation fault.
0xf6d5873a in MethodTable::GetFlag (this=0x61006c,
    flag=MethodTable::enum_flag_HasComponentSize)
    at /home/hanjoung/ws/dotnet/coreclr/src/vm/methodtable.h:3997
3997            return m_dwFlags & flag;
```

janvorli:
@wateret could you please additionally set trEnumGCRefs to true (and also keep the dspPtr set to true as you did now) and get me the full GC log both with and without your fix via gist (https://gist.github.com/)?
Also, in the failure case, could you please add stacktrace at the time of failure? I would like to see the parameter values in the callstack in relation to the GC log.

wateret:
@janvorli Here is the log and backtrace you mentioned. https://gist.github.com/wateret/43c18a5006ff52a9621b81c2c10caaf3

wateret:
@janvorli In the log I put up on Gist, is this double-relocation for one GC?

https://gist.github.com/wateret/43c18a5006ff52a9621b81c2c10caaf3#file-dump-L3853 (funclet)
https://gist.github.com/wateret/43c18a5006ff52a9621b81c2c10caaf3#file-dump-L3889 (main function)

```
3853 TID 0008:     GC Root FFFFCB1C RELOCATED F2E35E9C -> F2E177CC  MT = F6348384T
3889 TID 0008:     GC Root FFFFCB1C RELOCATED F2E177CC -> F2E094FC  MT = 006E0068T
```

It looks like relocation happened twice in GC count 1. Am I right? The crash occurs when promoting `*FFFFCB1C` on GC count 2.

wateret:
@janvorli It seems like there is a way not to double-report for those that use GCInfoDecoder.
https://github.com/dotnet/coreclr/blob/release/1.0.0-rc2/src/vm/gcinfodecoder.cpp#L574-L582

On arm32/Linux, I get the below message in main function. The funclet reports but main function does not.

```
TID 5798: Not reporting this frame because it was already reported via another funclet.
```

janvorli:
@wateret so it seems that an oposite approach to the one you have used in this PR is used there. 
If you look at the , you'll see this code at the beginning:
```C++
    // In order to make ARM more x86-like we only ever report the leaf frame
    // of any given function. We accomplish this by having the stackwalker
    // pass a flag whenever walking the frame of a method where it has
    // previously visited a child funclet
    if (WantsReportOnlyLeaf() && (inputFlags & ParentOfFuncletStackFrame))
    {
        LOG((LF_GCROOTS, LL_INFO100000, "Not reporting this frame because it was already reported via another funclet.\n"));
        return true;
    }
```

The WantsReportOnlyLeaf() is a bit set by the GC encoder on request from the JIT if `compiler->ehAnyFunclets()` returns true, which means that it is set for all functions that contain any funclets.

The ParentOfFuncletStackFrame in the input flags is passed by the stack walker to the EECodeManager::EnumGcRefs when it finds a frame that's a parent of a funclet.

So I think we should do the same thing for x86 Linux. I'm actually not sure why the WantsReportOnlyLeaf() check is necessary, since if the stack walker reports that the current frame is a parent of a funclet, then there must be a funclet in the function.

So I would suggest that instead of your fix, we would add the following to the beginning of EECodeManager::EnumGcRefs for x86 Unix:

```C++
    if (inputFlags & ParentOfFuncletStackFrame)
    {
        LOG((LF_GCROOTS, LL_INFO100000, "Not reporting this frame because it was already reported via another funclet.\n"));
        return true;
    }
```

wateret:
@janvorli Thank you, I update the code. For the sake of lifetime and arg reg table, GCInfoDecoder's approach(funclet does report, main function does not) looks more suitable.

janvorli:

janvorli:
@wateret I assume the change fixed the problem and it can be merged, right?

wateret:
@janvorli Yes, I found no regression on my local unit test(pri0).

wateret:

wateret:
@janvorli Could we merge this please?

janvorli:
@wateret I am sorry for the delay.

PR #10193, opened 2017-03-15
[RyuJIT/ARM32] Fix lvOnFrame for struct args

mskvortsov:
This change removes struct-related NYIs introduced earlier to hide assertion failure in `raMarkStkVars` at `regalloc.cpp:6747`:
```c++
        /* Some basic checks */

        // It must be in a register, on frame, or have zero references.

        noway_assert(varDsc->lvIsInReg() || varDsc->lvOnFrame || varDsc->lvRefCnt == 0);
```
NYI stats for `CodeGenBringUpTests` with this change applied (chk build):
```
    745 codegenarm.cpp:1882 - NYI: genRegCopy
    391 lsraarm.cpp:1334 - NYI: Unimplemented node type lclHeap
    256 morph.cpp:4659 - NYI_ARM: fgMorphMultiregStructArgs
    190 lsraarm.cpp:900 - NYI: lowering double
    151 lowerarm.cpp:149 - NYI_ARM: Lowering for GT_STORE_OBJ isn't implemented
      6 lsraarm.cpp:1334 - NYI: Unimplemented node type memoryBarrier
      1 lsraarm.cpp:1334 - NYI: Unimplemented node type obj
```
(Previous stats are available in the umbrella issue #8496.)

### DIFF
--- a/src/jit/lclvars.cpp
+++ b/src/jit/lclvars.cpp
@@ -676,11 +676,6 @@ void Compiler::lvaInitUserArgs(InitVarDscInfo* varDscInfo)
                 codeGen->regSet.rsMaskPreSpillRegArg |= regMask;
             }
         }
-        else
-        {
-            varDsc->lvOnFrame = true; // The final home for this incoming register might be our local stack frame
-        }
-
 #else // !_TARGET_ARM_
 #if defined(FEATURE_UNIX_AMD64_STRUCT_PASSING)
         SYSTEMV_AMD64_CORINFO_STRUCT_REG_PASSING_DESCRIPTOR structDesc;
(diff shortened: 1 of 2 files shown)


BruceForstall:
Can you explain the reasoning behind this change? Can you show a test case that requires it? Is this something specific to RyuJIT/ARM32, or does it apply also to the ARM32 legacy back-end (you've changed code common to both).

mskvortsov:
Full testsuite launch confirms no regressions for the legacy arm32 backend.
And that comment in `lclvars.cpp` seems to be on my side too:
```c++
         // The final home for this incoming register might be our local stack frame.
         // For System V platforms the final home will always be on the local stack frame.
         varDsc->lvOnFrame = true;
```

PR #10194, opened 2017-03-15
Prepare MethodBase.cs for migration to shared partition.

AtsushiKan:
The preparation steps are:

- Get rid of the roommating of unrelated types in the same file (INVOCATION_FLAGS).
- Split off internal methods and apis that are inherently unportable (usually the static methods) into MethodBase.Unsharable.cs.


AtsushiKan:
cc @jkotas 
cc @alexperovich

src/mscorlib/src/System/Reflection/MethodBase.Unsharable.cs
```diff
@@ -0,0 +1,161 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
jkotas: I would call it MethodBase.CoreCLR.cs - same naming convention is used for CoreCLR specific partial classes in corefx.

src/mscorlib/src/System/Reflection/MethodBase.Unsharable.cs
```diff
@@ -0,0 +1,161 @@
...
+            StackCrawlMark stackMark = StackCrawlMark.LookForMyCaller;
+            return RuntimeMethodInfo.InternalGetCurrentMethod(ref stackMark);
+        }
+        #endregion
+
+        #region Internal Members
+        // used by EE
+        private IntPtr GetMethodDesc() { return MethodHandle.Value; }
+
+#if FEATURE_APPX
```
jkotas: Nit: You can delete this empty ifdef while you are on it.

src/mscorlib/src/System/Reflection/MethodBase.Unsharable.cs
```diff
@@ -0,0 +1,161 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+//
```
jkotas: Nit: Delete these blank comment lines (all files)

jkotas:
LGTM otherwise

src/mscorlib/src/System/Reflection/MethodBase.Unsharable.cs
AtsushiKan: 'kay - I figured not hardcoding the platform was more flexible but since there's prior precedent...

PR #10195, opened 2017-03-15
Fix ToTitleCase Functionality for Dutch Cultures

rionmonster:
fixes https://github.com/dotnet/corefx/issues/16770

Added support for properly handling scenarios where the "IJ" characters at the beginning of a string were not being properly capitalized in Dutch cultures (i.e. "nl-BE" and "nl-NL") as [per Section 3.13 of the Unicode Standard](http://www.unicode.org/versions/Unicode9.0.0/ch03.pdf#page=81)

The fix basically consists of three parts:

- Determine if the current culture is Dutch (via an `IsDutchCulture()` method)
- If so, check if the current position within the string has an "IJ" at the current position (via an `IsIjAtCurrentPosition()` method).
- If both of these conditions are met, increment the `charLen` value, which will be passed to the `AddTitlecaseLetter()` method and will treat the "IJ" as a surrogate pair and capitalize both of them

I wasn't sure how granular to separate the methods or if it would be better to simply consolidate them into a single `ApplySpecialCaseForDutchCultures()` type method or similar. Another option would be to possibly store a flag for the current culture (and if it was Dutch) to avoid checking it repeated as it won't change within the context of a `TextInfo` class.

The following test cases were used to test this : 

```
[Fact]
public void ToTitleCaseDutchTest()
{
        string[] dutchCultureInfos = new string[] { "nl-BE", "nl-NL" };
        foreach (string cultureInfo in dutchCultureInfos)
        {
                TextInfo ti = CultureInfo.GetCultureInfo(cultureInfo).TextInfo;
                Assert.Equal("IJ IJ IJ IJ", ti.ToTitleCase("ij iJ Ij IJ"));
                Assert.Equal("IJzeren Eigenschappen", ti.ToTitleCase("ijzeren eigenschappen"));
                Assert.Equal("Lake IJssel", ti.ToTitleCase("lake iJssel"));
                Assert.Equal("Boba N' IJango Fett PEW PEW", ti.ToTitleCase("Boba n' Ijango fett PEW PEW"));
         }
}
```

It's worth noting that I am not a native Dutch speaker, so if any special-case scenarios need to be added, let me know.

@tarekgh

### DIFF
--- a/src/mscorlib/src/System/Globalization/TextInfo.cs
+++ b/src/mscorlib/src/System/Globalization/TextInfo.cs
@@ -474,7 +474,7 @@ namespace System.Globalization
             StringBuilder result = new StringBuilder();
             string lowercaseData = null;
             // Store if the current culture is Dutch (special case)
-            bool isDutchCulture = CultureName.StartsWith("nl-",StringComparison.OrdinalIgnoreCase);
+            bool isDutchCulture = CultureName.StartsWith("nl-", StringComparison.OrdinalIgnoreCase);
 
             for (int i = 0; i < str.Length; i++)
             {
--- a/src/mscorlib/src/System/Globalization/TextInfo.cs
+++ b/src/mscorlib/src/System/Globalization/TextInfo.cs
@@ -482,6 +482,15 @@ namespace System.Globalization
                 charType = CharUnicodeInfo.InternalGetUnicodeCategory(str, i, out charLen);
                 if (Char.CheckLetter(charType))
                 {
+                    // Special case to check for Dutch specific titlecasing with
+                    // "IJ" characters at the beginning of word
+                    if (IsDutchCulture(str) && IsIjAtCurrentPosition(str, i))
+                    {
+                        // Increment the charLen to treat the "IJ" as a surrogate
+                        // pair and capitalize both letters
+                        charLen++;
+                    }
+
                     // Do the titlecasing for the first character of the word.
                     i = AddTitlecaseLetter(ref result, ref str, i, charLen) + 1;
 
@@ -682,6 +691,40 @@ namespace System.Globalization
                  || uc == UnicodeCategory.TitlecaseLetter
                  || uc == UnicodeCategory.ModifierLetter
                  || uc == UnicodeCategory.OtherLetter);
+        }
+
+
+        /// 
+        ///  Checks if the current culture is Dutch (i.e. "nl-BE" or "nl-NL") to
+        ///  determine if special casing rules should be applied. (Dutch Titlecasing)
+        ///
+        private bool IsDutchCulture(String str)
+        {
+            // Validate inputs
+            if (str == null)
+            {
+                throw new ArgumentNullException("string");
+            }
+
+            // Ignore if not Dutch
+            return IndexOfStringOrdinalIgnoreCase("nl-BE,nl-NL", _cultureName, 0, 11) != -1;
+        }
+
+
+        /// 
+        ///  Performs a case-insensitive check for the characters "ij" at the current
+        ///  position within a string. (Dutch Titlecasing)
+        ///
+        private bool IsIjAtCurrentPosition(String str, int position)
+        {
+            // Validate inputs
+            if (str == null)
+            {
+                throw new ArgumentNullException("string");
+            }
+
+            // Determine if "ij" is at the current position in the string
+            return IndexOfStringOrdinalIgnoreCase(str, "ij", position, 2) != -1;
         }
 
         //
--- a/src/mscorlib/src/System/Globalization/TextInfo.cs
+++ b/src/mscorlib/src/System/Globalization/TextInfo.cs
@@ -473,6 +473,8 @@ namespace System.Globalization
 
             StringBuilder result = new StringBuilder();
             string lowercaseData = null;
+            // Store if the current culture is Dutch (special case)
+            bool isDutchCulture = IsDutchCulture(str);
 
             for (int i = 0; i < str.Length; i++)
             {
@@ -484,7 +486,7 @@ namespace System.Globalization
                 {
                     // Special case to check for Dutch specific titlecasing with
                     // "IJ" characters at the beginning of word
-                    if (IsDutchCulture(str) && IsIjAtCurrentPosition(str, i))
+                    if (isDutchCulture && IsIjAtCurrentPosition(str, i))
                     {
                         // Increment the charLen to treat the "IJ" as a surrogate
                         // pair and capitalize both letters
--- a/src/mscorlib/src/System/Globalization/TextInfo.cs
+++ b/src/mscorlib/src/System/Globalization/TextInfo.cs
@@ -474,7 +474,7 @@ namespace System.Globalization
             StringBuilder result = new StringBuilder();
             string lowercaseData = null;
             // Store if the current culture is Dutch (special case)
-            bool isDutchCulture = IsDutchCulture(str);
+            bool isDutchCulture = CultureName.StartsWith("nl-",StringComparison.OrdinalIgnoreCase);
 
             for (int i = 0; i < str.Length; i++)
             {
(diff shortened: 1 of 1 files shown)


src/mscorlib/src/System/Globalization/TextInfo.cs
```diff
@@ -482,6 +482,15 @@ public unsafe String ToTitleCase(String str)
                 charType = CharUnicodeInfo.InternalGetUnicodeCategory(str, i, out charLen);
                 if (Char.CheckLetter(charType))
                 {
+                    // Special case to check for Dutch specific titlecasing with
+                    // "IJ" characters at the beginning of word
+                    if (IsDutchCulture(str) && IsIjAtCurrentPosition(str, i))
```
tarekgh: >IsDutchCulture(str) [](start = 24, length = 19)

this can be cached in a local variable instead of calling the same method again inside the loop

rionmonster: Just updated this by adding the `isDutchCulture` variable outside of the loop, which should address that issue.

src/mscorlib/src/System/Globalization/TextInfo.cs
```diff
@@ -482,6 +482,15 @@ public unsafe String ToTitleCase(String str)
...
+                    {
+                        // Increment the charLen to treat the "IJ" as a surrogate
+                        // pair and capitalize both letters
+                        charLen++;
+                    }
+
```
tarekgh: It is better we optimize this case as AddTitlecaseLetter call allocate extra string. we can do the following:

if (isDutchCulture && i < str.Lenth - 1 && (str[i] == 'i" || str[i] == 'I") && (str[i+1] == 'j' || str[i+1] == 'J') )
{
    result.Append("IJ");
    i+= 2;
}

src/mscorlib/src/System/Globalization/TextInfo.cs
```diff
@@ -684,6 +693,40 @@ private static bool IsLetterCategory(UnicodeCategory uc)
...
+        ///  Checks if the current culture is Dutch (i.e. "nl-BE" or "nl-NL") to
+        ///  determine if special casing rules should be applied. (Dutch Titlecasing)
+        ///
+        private bool IsDutchCulture(String str)
+        {
+            // Validate inputs
+            if (str == null)
+            {
+                throw new ArgumentNullException("string");
+            }
```
tarekgh: you don't need this check

src/mscorlib/src/System/Globalization/TextInfo.cs
```diff
@@ -684,6 +693,40 @@ private static bool IsLetterCategory(UnicodeCategory uc)
                  || uc == UnicodeCategory.OtherLetter);
         }
 
+
+        /// 
+        ///  Checks if the current culture is Dutch (i.e. "nl-BE" or "nl-NL") to
+        ///  determine if special casing rules should be applied. (Dutch Titlecasing)
+        ///
+        private bool IsDutchCulture(String str)
```
tarekgh: >String str [](start = 36, length = 10)

you don't need this parameter either

src/mscorlib/src/System/Globalization/TextInfo.cs
```diff
@@ -684,6 +693,40 @@ private static bool IsLetterCategory(UnicodeCategory uc)
...
+        {
+            // Validate inputs
+            if (str == null)
+            {
+                throw new ArgumentNullException("string");
+            }
+
+            // Ignore if not Dutch
+            return IndexOfStringOrdinalIgnoreCase("nl-BE,nl-NL", _cultureName, 0, 11) != -1;
```
tarekgh: >return IndexOfStringOrdinalIgnoreCase("nl-BE,nl-NL", _cultureName, 0, 11) != -1; [](start = 12, length = 80)

this need to do something CultureName.StartsWith("nl-", StringComparison.OrdinalIgnoreCase)

by the way, does nl-BE & nl-NL are the only Dutch cultures? do consider de-DE too? how did you get the list?

src/mscorlib/src/System/Globalization/TextInfo.cs
```diff
@@ -684,6 +695,40 @@ private static bool IsLetterCategory(UnicodeCategory uc)
...
+        {
+            // Validate inputs
+            if (str == null)
+            {
+                throw new ArgumentNullException("string");
+            }
+
+            // Determine if "ij" is at the current position in the string
+            return IndexOfStringOrdinalIgnoreCase(str, "ij", position, 2) != -1;
+        }
```
tarekgh: you don't need this if you apply my previous suggestion

src/mscorlib/src/System/Globalization/TextInfo.cs
tarekgh: I double checked and the check CultureName.StartsWith("nl-", StringComparison.OrdinalIgnoreCase) should be enough for all Dutch cultures

src/mscorlib/src/System/Globalization/TextInfo.cs
```diff
@@ -684,6 +693,40 @@ private static bool IsLetterCategory(UnicodeCategory uc)
...
```
rionmonster: I'll add all of these changes shortly and let you know when they are all done.

src/mscorlib/src/System/Globalization/TextInfo.cs
tarekgh: use CultureName.StartsWith("nl-", StringComparison.OrdinalIgnoreCase) as we need to check all cultures starts with "nl-" 

using == will be case sensitive so if you have the name in the form "NL-NL"  for instance will not work

rionmonster:
All of these proposed fixes / changes should be in this latest commit.

src/mscorlib/src/System/Globalization/TextInfo.cs
```diff
@@ -473,6 +473,8 @@ public unsafe String ToTitleCase(String str)
 
             StringBuilder result = new StringBuilder();
             string lowercaseData = null;
+            // Store if the current culture is Dutch (special case)
+            bool isDutchCulture = CultureName.StartsWith("nl-",StringComparison.OrdinalIgnoreCase);
```
tarekgh: nit: please add space after the comma.

tarekgh:
I added a small comment but other than that LGTM.

did you test it? and will you be willing to add the test in corefx repo after we get this fix there? 

thanks for your effort with this fix.

rionmonster:
I've fixed the minor spacing issue (I honestly don't know how that managed to sneak into there). 

> did you test it? note that you need to test it with Dutch cultures and non Dutch cultures.

With regards to testing, I have been running it against the existing CoreFX branch and that suite of tests as well as a test that I wrote (and mentioned in the original PR message) to specifically test this fix. I have a build running at present with all of these recent changes, and when that is done, we should be good to go.

Would you recommend building another entire test case similar to the first one, but using non-Dutch cultures? I assumed that the existing suite of tests would indicate any breaking changes or issues with non-Dutch cultures, but I could easily build one to test the same functionality with different cultures (i.e. normal "Ij" titlecasing as opposed to "IJ" casing).

> and will you be willing to add the test in corefx repo after we get this fix there?

Sure, I have a branch with that test all ready to go once we get the green light.

tarekgh:
> Would you recommend building another entire test case similar to the first one, but using non-Dutch cultures? I assumed that the existing suite of tests would indicate any breaking changes or issues with non-Dutch cultures, but I could easily build one to test the same functionality with different cultures (i.e. normal "Ij" titlecasing as opposed to "IJ" casing).

I would suggest convert the test method you are using to be a Theory and then provide the data (like the culture name, the string to title case it and the expected string). in the data you can pass different cultures and different strings (with IJ's) and then the expected return string.

rionmonster:
I'll work on getting that together tonight. I'll let you know what it looks like so you can point out if anything looks out of the ordinary (or just plain wrong) :)

tarekgh:

rionmonster:
I've added a Theory-based approach which should cover both Dutch and non-Dutch test cases to my CoreFX branch: 

```
public static IEnumerable<object[]> DutchTitleCaseInfo_TestData()
{
        yield return new object[] { "nl-NL", "IJ IJ IJ IJ", "ij iJ Ij IJ" };
        yield return new object[] { "nl-be", "IJzeren Eigenschappen", "ijzeren eigenschappen" };
        yield return new object[] { "NL-NL", "Lake IJssel", "lake iJssel" };
        yield return new object[] { "NL-BE", "Boba N' IJango Fett PEW PEW", "Boba n' Ijango fett PEW PEW" };
        yield return new object[] { "en-us", "Ijill And Ijack", "ijill and ijack" };
        yield return new object[] { "de-DE", "Ij Ij IJ Ij", "ij ij IJ ij" };
        yield return new object[] { "he-il", "Ijon't Know What Will Happen.", "Ijon't know what Will happen." };
}

[Theory]
[MemberData(nameof(DutchTitleCaseInfo_TestData))]
public void ToTitleCaseDutchTest(string cultureName, string expected, string actual)
{
        TextInfo ti = CultureInfo.GetCultureInfo(cultureName).TextInfo;
        Assert.Equal(expected, ti.ToTitleCase(actual));
}
```
With regards to building and targeting OSX, does this have to be done from a Mac (or a machine running OSX)? I attempted to do this from my local Windows machine using the following:

```
build x64 Checked osxmscorlib
```

However, it doesn't look like the build worked as expected (as there were _tons_ of `Unreachable Code Detected` warnings that popped up. It did say that it successfully built the project, however when attempting to target this build using the CoreFX project, it didn't seem to work. I could be missing something trivial, but I don't seem to know exactly what it is :)

tarekgh:
> With regards to building and targeting OSX, does this have to be done from a Mac

yes.

rionmonster:
Hmmm.

This could be problematic as I don't have one. Any ideas / suggestions (aside from buying one)?

tarekgh:
why you want to test on MAC? you can test on Windows or Linux (by creating a VM)

rionmonster:
@tarekgh 

I've had a PR ready with a handful of tests for this issue for CoreFX. Would you recommend me sending that up or should I wait for a specific issue to be created?

tarekgh:
@rionmonster please go ahead and submit the PR in corefx and link it with this issue.

PR #10202, opened 2017-03-15
Migrate MethodBase.cs over to the shared partition.

AtsushiKan:
Migrate MethodBase.cs over to the shared partition.

As a side effect, this also lights up the new api IsConstructedGenericMethod on CoreClr.


AtsushiKan:

PR #10203, opened 2017-03-15
CoreFx #15622 Dictionary<TKey, TValue>.Remove(TKey, out TValue)

WinCPP:
New API to return the value associated with key being removed, if the key is present.

@karelz @danmosemsft Kindly review.

### DIFF
--- a/src/mscorlib/src/System/Collections/Generic/Dictionary.cs
+++ b/src/mscorlib/src/System/Collections/Generic/Dictionary.cs
@@ -590,6 +590,9 @@ namespace System.Collections.Generic
             entries = newEntries;
         }
 
+        // The overload Remove(TKey key, out TValue value) is a copy of this method with one additional
+        // statement to copy the value for entry being removed into the output parameter.
+        // Code has been intentionally duplicated for performance reasons.
         public bool Remove(TKey key)
         {
             if (key == null)
@@ -625,6 +628,51 @@ namespace System.Collections.Generic
                     }
                 }
             }
+            return false;
+        }
+
+        // This overload is a copy of the overload Remove(TKey key) with one additional
+        // statement to copy the value for entry being removed into the output parameter.
+        // Code has been intentionally duplicated for performance reasons.
+        public bool Remove(TKey key, out TValue value)
+        {
+            if (key == null)
+            {
+                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.key);
+            }
+
+            if (buckets != null)
+            {
+                int hashCode = comparer.GetHashCode(key) & 0x7FFFFFFF;
+                int bucket = hashCode % buckets.Length;
+                int last = -1;
+                for (int i = buckets[bucket]; i >= 0; last = i, i = entries[i].next)
+                {
+                    if (entries[i].hashCode == hashCode && comparer.Equals(entries[i].key, key))
+                    {
+                        if (last < 0)
+                        {
+                            buckets[bucket] = entries[i].next;
+                        }
+                        else
+                        {
+                            entries[last].next = entries[i].next;
+                        }
+
+                        value = entries[i].value;
+
+                        entries[i].hashCode = -1;
+                        entries[i].next = freeList;
+                        entries[i].key = default(TKey);
+                        entries[i].value = default(TValue);
+                        freeList = i;
+                        freeCount++;
+                        version++;
+                        return true;
+                    }
+                }
+            }
+            value = default(TValue);
             return false;
         }
 


src/mscorlib/src/System/Collections/Generic/Dictionary.cs
```diff
@@ -614,6 +622,9 @@ public bool Remove(TKey key)
                         {
                             entries[last].next = entries[i].next;
                         }
+
+                        value = entries[i].value;
```
danmosemsft: It might be slightly clearer if this line went directly after the if on line 615

src/mscorlib/src/System/Collections/Generic/Dictionary.cs
```diff
@@ -592,11 +592,19 @@ private void Resize(int newSize, bool forceNewHashCodes)
...
+        }
+
+        public bool Remove(TKey key, out TValue value)
+        {
             if (key == null)
             {
                 ThrowHelper.ThrowArgumentNullException(ExceptionArgument.key);
             }
 
+            value = default(TValue);
```
danmosemsft: this could move just before the return false.

src/mscorlib/src/System/Collections/Generic/Dictionary.cs
```diff
@@ -592,11 +592,19 @@ private void Resize(int newSize, bool forceNewHashCodes)
 
         public bool Remove(TKey key)
         {
+            TValue unused;
+            return Remove(key, out unused);
```
danmosemsft: That doesn't seem to compile in our compiler right now. If it did, I think it's fine.

src/mscorlib/src/System/Collections/Generic/Dictionary.cs
mikedn: This implementation of `public bool Remove(TKey key, out TValue value)` penalizes many users of `public bool Remove(TKey key)` by introducing an unnecessary copy (and initialization) of the value.

danmosemsft: @mikedn Any suggestions? A branch? Duplicate the body of the function?

mikedn: Duplicating the code might be the only option that doesn't impact performance. But it too has its cost since we're looking at a generic type. Using a branch to avoid the unnecessary copy might be a reasonable compromise.

But before doing any of that the performance impact should be measured.

danmosemsft:
@WinCPP could you please do some ad-hoc performance measurements? A simple console app is fine. 

This should help: https://github.com/dotnet/coreclr/blob/master/Documentation/workflow/UsingYourBuild.md
(Make sure to add some Console.WriteLine into the corelib somewhere to be certain it's your modified version getting used. )

Even easier, just [install 2.0](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/dogfooding.md) and paste in the Dictionary code into your own code -- both the old and new implementations. Make sure to build release.

Then I'd set up some dictionaries with different numbers of entries and confirm the timings with each implementation. Timing with Stopwatch is fine, and you'd probably do a bunch of iterations and drop outliers and get the standard deviation.

Perhaps @mikedn can point to a gist of code someone else has used for this purpose. I haven't got one handy.

danmosemsft:
This is a trivial one https://gist.github.com/jamesqo/0c98cf76934b51d0737143520a616074

Many folks have used [Benchmark.NET](https://github.com/dotnet/BenchmarkDotNet)
https://gist.github.com/benaadams/792d2734ef569d45be42
https://gist.github.com/anonymous/edca38b045d5a6d420766f79096cc6d8

safern:
Or even something simple like this one:

https://github.com/dotnet/coreclr/pull/9923#issuecomment-283880405

src/mscorlib/src/System/Collections/Generic/Dictionary.cs
WinCPP: @danmosemsft @mikedn Thanks for the review comments. I will work on getting the performance data. This review gives me a close and significant insight into working of C# and how to think differently when wearing a C# vs C++ hat.

WinCPP:
Thanks @danmosemsft @safern for the guidance. I will work on the performance part.

src/mscorlib/src/System/Collections/Generic/Dictionary.cs
WinCPP: @mikedn hmm. A few questions, mainly as a learning for me...

> Duplicating the code might be the only option that doesn't impact performance. But it too has its cost since we're looking at a generic type.
1. Are we referring to maintenance cost? But then what is special in context of generic type... Seems there is more here... Please share.

> Using a branch to avoid the unnecessary copy might be a reasonable compromise.
2. In non-duplication way, I'm not able to understand how we could avoid assignment at line 626, since we have to squirrel off the value before being overwritten. I think I am not able to understand what is meant by _branch_ here. Appreciate your inputs.

3. I was curious why the C# compiler would not inline entire `Remove(key, out value)` into the `Remove(key)` and distill off the unnecessary copies via intermediate code optimizations. For long I have always wondered why the chained constructors also won't inline (as seen in IL for even Release mode).

Performance data points I will collect
a. Old Remove implementation
b. Nested Remove implementation as in the PR with code re-org.
c. Branched version (but I need to be sure I understand what is meant by _branch_ in this context).

Thanks!

mikedn: > Are we referring to maintenance cost? But then what is special in context of generic type...

Maintenance cost too since we'll end up with 2 very similar pieces of code. But this is a generic type, this means that if you duplicate that code you're not adding just a single method, you're adding many more - one for each `Dictionary` instantiation where `TValue` is a value type (e.g. one for `Dictionary<int, int>`, another one for `Dictionary<int, bool>` and so on).

> In non-duplication way, I'm not able to understand how we could avoid assignment at line 626...

I don't know exactly what @danmosemsft had in mind when he said branch but one possible option is to pass a `bool` parameter to the method indicating whether to perform the copy or not. That would imply moving all this code to a new private method and having both `Remove` overloads call that method.

> I was curious why the C# compiler would not inline entire Remove(key, out value) into the...

The C# compiler doesn't do inlining, that's the job of the JIT compiler since inlining normally needs to take into account various "low level" aspects (e.g. native code size, calling convention). Anyway, there's too much code to inline.

> Performance data points I will collect

Make sure to measure the case of large value types. I don't expect this change to have a measurable impact on `Dictionary<int, int>` but the story may be different for something like `Dictionary<int, Matrix>`.

WinCPP:
For the performance testing, I want to use my local clr + corefx. I'm stuck in that. Please help. I want to use local CoreCLR because it has implementation and CoreFx because it has facade.

1. I built CoreCLR (build.cmd x64 Release) and configured CoreFx to use it by following the steps in "[Testing with private CoreCLR bits](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/developer-guide.md)". I had passed `-release` to .\build.cmd while building CoreFX. 
2. As mentioned in "[Using your local CoreFx build](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/dogfooding.md)", I modified the .csproj and Nuget.config to pick up Microsoft.Private.CoreFx.NETCoreApp from my release folder.

But the `dotnet restore` fails with following errors, which I am not able to solve. Please help. Sorry for dumping entire output, but I'm totally clueless :-(
```
D:\Progs\C#Progs\DictPerf>e:\dotnet2.0\dotnet.exe restore
  Restoring packages for D:\Progs\C#Progs\DictPerf\DictPerf.csproj...
e:\dotnet2.0\sdk\2.0.0-preview1-005440\NuGet.targets(97,5): error : Unable to resolve 'Microsoft.NETCore.Platforms (>= 2.0.0-beta-25211-0)' for '.NETCoreApp,Version=v2.0'. [D:\Progs\C#Progs\DictPerf\DictPerf.csproj]
e:\dotnet2.0\sdk\2.0.0-preview1-005440\NuGet.targets(97,5): error : Unable to resolve 'Microsoft.NETCore.Platforms (>= 2.0.0-beta-25211-0)' for '.NETCoreApp,Version=v2.0 (win7-x64)'. [D:\Progs\C#Progs\DictPerf\DictPerf.csproj]
  Lock file has not changed. Skipping lock file write. Path: D:\Progs\C#Progs\DictPerf\obj\project.assets.json
  Restore failed in 127.88 ms for D:\Progs\C#Progs\DictPerf\DictPerf.csproj.

  Errors in D:\Progs\C#Progs\DictPerf\DictPerf.csproj
      Unable to resolve 'Microsoft.NETCore.Platforms (>= 2.0.0-beta-25211-0)' for '.NETCoreApp,Version=v2.0'.
      Unable to resolve 'Microsoft.NETCore.Platforms (>= 2.0.0-beta-25211-0)' for '.NETCoreApp,Version=v2.0 (win7-x64)'.

  NuGet Config files used:
      D:\Progs\C#Progs\DictPerf\NuGet.Config
      C:\Users\WinCPP\AppData\Roaming\NuGet\NuGet.Config

  Feeds used:
      M:\corefx\bin\packages\Release
      https://dotnet.myget.org/F/dotnet-core/api/v3/index.json
      https://api.nuget.org/v3/index.json
```

safern:
@WinCPP it seems like the package name/version is wrong in your csproj. Would you mind sharing the ``
```
<ItemGroup>
   <PackageReference ... >

**Comment by WinCPP** [comment]
@safern please find below the csproj contents:

```XML
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.0</TargetFramework>
    <RuntimeFrameworkVersion>2.0.0-beta-001776-00</RuntimeFrameworkVersion>
	<RuntimeIdentifier>win7-x64</RuntimeIdentifier>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.Private.CoreFx.NETCoreApp" Version="4.4.0-beta-25211-0" />
  </ItemGroup>
</Project>
```
As mentioned in the step in the [doc](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/dogfooding.md), I made sure that the `Include` name matches and the version is same as in the `Release` folder in my build...

WinCPP:
Going by the error string that I'm getting,
```
      Unable to resolve 'Microsoft.NETCore.Platforms (>= 2.0.0-beta-25211-0)' for '.NETCoreApp,Version=v2.0'.
```
does it mean that I have to build by local CoreClr and / or CoreFx with .NETCoreApp version 2.0? And then only I can use my local bits...? Because following is the output for `dotnet --info` command in the git shell for my local corefx repo.
```
M:\corefx [issue-15622 ≡ +0 ~3 -0 !]> dotnet --info
.NET Command Line Tools (1.0.0-preview2-003131)

Product Information:
 Version:            1.0.0-preview2-003131
 Commit SHA-1 hash:  635cf40e58

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.10586
 OS Platform: Windows
 RID:         win10-x64
```

karelz:
Note that ".NET Command Line Tools" version != .NET Core version. The tools can target multiple .NET Core versions (like VS).

Here's example of .NET Core only API addition, if that helps: https://github.com/dotnet/corefx/pull/16996/files

WinCPP:
@karelz The link discusses about performance, but I could not get the fix for my issue...

Actually there are two local builds to be used CoreClr and CoreFx and now I notice that each of them has user your build / dogfooding documents [here](https://github.com/dotnet/coreclr/blob/master/Documentation/workflow/UsingYourBuild.md) and [here ](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/dogfooding.md) respectively, with former pointing to latter.

Steps in the former document for using local CoreClr bits worked perfectly. However this being a case of adding new API with implementation in CoreClr and facade in CoreFx, I thought I have to do the steps in CoreFx dogfooding document as well; and there it fails. So are the steps for using local CoreFx not to be done and, just do steps for local CoreClr? Then, how would I get the new API on CoreFx facade?

safern:
adding @joperezr as he might certainly know the answer to help you @WinCPP.

WinCPP:
I was still not able to solve the above problems so I abandoned that approach. I made copy of System.Collections.Tests and modified its contents and added performance tests for only two implementations of Remove:

```C#
API implementation
OLD - bool Remove(TKey key)
NEW - bool Remove(Tkey key) that internally uses Remove(TKey key, out value)
```
Following is the performance data that I got. Please let me know if it makes sense.

```
For a dictionary of int, int
Number of iterations: 20
Number of entries processed in each iteration: 2000000
Data points: timings in milliseconds
            OLD             NEW
           19.00           22.00
           19.00           21.00
           20.00           21.00
           19.00           21.00
           19.00           21.00
           19.00           21.00
           19.00           21.00
           19.00           21.00
           19.00           21.00
           19.00           24.00
           19.00           21.00
           19.00           21.00
           19.00           22.00
           19.00           21.00
           19.00           21.00
           19.00           21.00
           19.00           21.00
           19.00           21.00
           19.00           21.00
           19.00           21.00
Average:   19.05           21.25
Std Dev:    0.217944947     0.698212002


Dictionary of int, struct of 8 ints
Number of iterations: 20
Number of entries processed in each iteration: 2000000
Data points: timings in milliseconds
            OLD            NEW
           23.00          25.00
           22.00          24.00
           22.00          24.00
           22.00          24.00
           22.00          24.00
           23.00          24.00
           22.00          24.00
           22.00          24.00
           22.00          24.00
           23.00          24.00
           22.00          27.00
           22.00          24.00
           22.00          24.00
           22.00          24.00
           22.00          24.00
           22.00          24.00
           22.00          24.00
           22.00          24.00
           22.00          24.00
           22.00          25.00
Average:   22.15          24.25
Std Dev:    0.357071421    0.698212002
```

WinCPP:
@danmosemsft @mikedn @karelz @safern Please review the above data.

danmosemsft:
Thanks @WinCPP . This is a very hot method, so we don't want to take a 10% performance regression. I would try this approach now

> pass a bool parameter to the method indicating whether to perform the copy or not. That would imply moving all this code to a new private method and having both Remove overloads call that method.

Unless @jkotas has another idea.

jkotas:
> Unless @jkotas has another idea.

Well, the Dictionary is big and fat type already and each extra convenience APIs is making it fatter. The lookup loop is replicated in the implementation many times, so duplicating it one more time is hardly going to register on the radar. If we wanted to avoid this duplication, I think that the naïve implementation would be the best way to do it:

```
if(!TryGetValue(key, out value)) return false;
Remove(key);
return true;
```

Yes, it is not as fast as possible, but I do not see a problem with it. Dictionary is not the fastest collection for any given case. It is a collection that is fast enough for most common uses.

The performance effects of adding every convenience API imaginable over time to the common collections is something for fxdc to think about. I am not sure whether this particular API would made it if it was my call... .

karelz:
@jkotas what kind of performance impact do you see in adding more convenience APIs? Size of the type? Code locality?
If we accept that convenience APIs don't need to be super-performant, do you see any other problem with the simple solution you outlined above?

WinCPP:
>>pass a bool parameter to the method indicating whether to perform the copy or not. That would imply moving all this code to a new private method and having both Remove overloads call that method.

> Unless @jkotas has another idea.

@danmosemsft There is an out parameter that will have to be mandatorily assigned before leaving the private method. Essentially something like this,

```C#
if (FoundInDictionary)
{
    // Do the remove stuff
    //
    if (WantTheValueOut)
    {
        value = entries[i].value;
    }
    else
    {
        value = default(TValue); // out parameter hence all return paths should assign a value
    }
    // Do the actual remove stuff
    //
    return true;
}

value = default(TValue);
return false;
```
Therefore, a copy or a create is inevitable and will penalize the old `Remove(TKey key)` implementation invariably. So I am not sure if branching is going to help, unless it were a `ref` parameter instead of `out`. Appreciate your thoughts. Do we still want to collect stats for this?

I will capture stats for `Remove(TKey key, out value)` as a wrapper around `TryGetValue` and `Remove(TKey key)` as suggested by @jkotas ... The issue mentions this as a performance problem as double look ups happen on the dictionary for same key, once for fetching value an then for removal. Perhaps the data might help us...

Just a layman's wishful thinking... I always wish if inlining for compilers could be smart to detect plain vanilla function overloads in which one wraps the other (`Remove(TKey key)` and `Remove(TKey key, out TValue value)` pair or the chained constructors)... In such case the expression tree for the inner call could be duplicated into the outer wrapper during intermediate code generation and then aggressively optimized... regardless of whether it is concrete or generic type.

jkotas:
> @jkotas what kind of performance impact do you see in adding more convenience APIs? Size of the type?

Yes, size of the type - that gets multipled for generic types. Contributes to disk footprint, startup time, ... .

> If we accept that convenience APIs don't need to be super-performant

Then they can be extension methods against the interface... .

mikedn:
> Just a layman's wishful thinking... I always wish if inlining for compilers could be smart to detect plain vanilla function overloads in which one wraps the other (Remove(TKey key) and Remove(TKey key, out TValue value) pair or the chained constructors)...

What does that solve? You still end up with a bunch of additional native code being generated.

To add some numbers to this story: the existing `Remove` method has 300-500 bytes depending on the involved key and value types (e.g. `Dictionary<int, byte>.Remove(int)` has ~400 bytes). Every additional unshared instantiation adds another 300-500 bytes. Duplicating the `Remove` code in the new overload doubles that amount.

mikedn:
>```C#
>if(!TryGetValue(key, out value)) return false;
>Remove(key);
>return true;
>```

Maybe we can build both `Remove` overloads on top of `FindEntry`? It's been a while since I starred at the dictionary code but it seems that there's no reason to have multiple lookups except to save a range check or two.

WinCPP:
> Maybe we can build both Remove overloads on top of FindEntry?

Current find entry returns the index for the found entry by abstracting the direct bucket / entry list. Remove APIs work at the _found_ location to adjust the root / next links for the buckets / entries to _effect_ a removal; essentially this _removal_ code is again a common block which is in fact dependent on results of the _FindEntry_ functional block. It is between these two common functional blocks that the second `Remove` overload saves off the `value` for the entry that was located by _FindEntry_ block and is about to be cleaned up by the _removal_ block...

Hope I am making sense...

WinCPP:
Following is the new performance data. The columns represent the following functions / functionalities. Question: Looking at the data, duplication of code is appears to be the way to go? @danmosemsft I haven't checked performance of branching implementation ... for now. Kindly refer to [this](https://github.com/dotnet/coreclr/pull/10203#issuecomment-287524274) comment above. Appreciate your inputs.

```
TryGet       : Untouched TryGet API
OrgRemove    : Existing Remove API
NewRemoveGet : New Remove Overload, a copy of OrgRemove code with additional functionality
               to copy the `value` before effecting the _removal_ part. Currently in PR.
NewRemove    : Remove API (no value ouput) as wrapper around NewRemoveGet. Currently in PR.
RemoveWrap   : New Remove Overload implemented as wrapper around `TryGet` and 'OrgRemove`

Observations:
a. RemoveWrap timing is sum of timings for TryGet and Remove (original) APIs and has duplicated search effort.
b. OrgRemove is faster than NewRemove that internally NewRemoveGet


For a dictionary of int, int
Number of iterations: 20
Number of entries processed in each iteration: 2000000
Data points: timings in milliseconds

Run     TryGet  OrgRemove   NewRemoveGet    NewRemove   RemoveWrap
1        18       19            21             21          37
2        18       19            21             21          37
3        18       19            21             21          37
4        18       19            21             21          37
5        18       19            21             21          37
6        18       19            21             21          37
7        18       19            21             21          37
8        18       19            21             21          37
9        18       19            21             21          37
10       18       19            21             21          37
11       18       19            21             21          37
12       18       19            21             21          37
13       18       19            21             21          37
14       18       19            21             21          37
15       18       19            21             21          37
16       18       19            21             21          37
17       18       19            21             21          37
18       18       19            21             21          37
19       18       19            21             21          37
20       18       19            21             21          37
-------------------------------------------------------------
Avg      18       19            21             21          37
Std Dev   0        0             0              0           0



For a dictionary of int, struct of 8 ints
Number of iterations: 20
Number of entries processed in each iteration: 2000000
Data points: timings in milliseconds

Run     TryGet  OrgRemove   NewRemoveGet    NewRemove   RemoveWrap
1        20       23            24             26          42
2        20       22            24             25          42
3        20       22            24             25          42
4        20       22            26             25          42
5        20       22            25             25          42
6        20       22            24             25          42
7        20       22            24             25          42
8        20       22            24             25          42
9        20       22            24             25          42
10       20       22            24             25          42
11       20       22            24             25          42
12       20       22            24             26          42
13       20       22            24             25          42
14       20       22            24             25          42
15       20       22            24             25          42
16       20       22            24             25          42
17       20       22            24             25          42
18       20       22            24             25          42
19       20       22            24             25          42
20       20       22            24             25          42
-------------------------------------------------------------
Avg      20       22.05         24.15          25.1        42
Std Dev   0    0.217944947    0.476969601       0.3         0
```

karelz:
@jkotas @mikedn the impact on size/
I have created https://github.com/dotnet/corefx/issues/17275 to consider migrating Core-only 2.0 newly added APIs into extension methods.

@WinCPP you seem to be truly "lucky" to hit all interesting design/principle decisions around CoreFX I have seen in last few months :( (this is number 3 for you, while other contributors get typically 1 at most and even that is usually pretty rare). I am very sorry for that!
Anyway, I would recommend to wait until Tuesday if we come to decision on the topic (member vs. extension method). I hope that it is not too big inconvenience for you. If the discussion is inconclusive on Tuesday, we may decide to just add this API as a member and then potentially migrate it later to extension method together with all other APIs we added as Core-only in 2.0 so far.

WinCPP:
> you seem to be truly "lucky" to hit all interesting design/principle decisions around CoreFX I have seen in last few months :( (this is number 3 for you, while other contributors get typically 1 at most and even that is usually pretty rare). I am very sorry for that!

@karelz No! In fact I feel lucky (really lucky... without quotes) that I got to work on such issues. Even if these issues have now entered kernel mode, I am fortunate that I could participate into discussions and data gathering up to this point. These are valuable learning opportunities for me; let them keep coming. For now, I will sleep (pun not intended and also intended :-) ) and wait for the issues to come back into 'grabs' mode with the decision tag... Thank you for all the opportunities to contribute.

src/mscorlib/src/System/Collections/Generic/Dictionary.cs
svick: The actual C# 7.0 syntax is `out var _`. Using that seems to compile fine for me.

WinCPP:
Hi! Gentle reminder... :) Is this in a state that I can work on...? Thanks!

karelz:
We need to review the list of APIs, incl. `Remove` - @ianhays created a list here: https://github.com/dotnet/corefx/issues/17275#issuecomment-288841228

I think this one falls into [1] category (https://github.com/dotnet/corefx/issues/17275#issuecomment-288197971), so we will likely take it.
To be on safe side, I would suggest to wait until next API review meeting on Tuesday.

safern:
@karelz I think this falls into [1] category as well. Where you able to review this yesterday in the API review meeting? Any updates so that we can get this API exposed for 2.0.0?

karelz:
We should get to it next Tue. I marked the main issue as blocking-release to pop in our API review meeting.
I entirely forgot about the issue and forgot to mark it api-ready-for-review, sorry for that. To my defense, we were booked with 2 other large higher-priority reviews which took the whole 2h segment anyway :(.

safern:
Hey @WinCPP the API review group decided this method should stay as an instance method (see: https://github.com/dotnet/corefx/issues/17275#issuecomment-291636863). 

So now we should be able to move forward with this implementation and make a decision on which codepath to follow regarding the performance results you got.

karelz:
@WinCPP FYI: We also approved adding the same method as extension method -- see https://github.com/dotnet/corefx/issues/17918

WinCPP:
Thanks @safern @karelz . So to resume the discussion, from the data points in [this comment](https://github.com/dotnet/coreclr/pull/10203#issuecomment-287538615), I am of opinion that it would be better to have the new `Remove` overload as a code level copy of the original Remove, with additional statement to copy the value into output parameter, rather than trying to realize one in terms of other in any possible way. Kindly let me know if this sounds ok.

danmosemsft:
@WinCPP I agree.

WinCPP:
Please ignore above commit. I am trying to sync my sandbox branch with dotnet/master and I have messed up miserably. Just thought I'd drop a note here... Will let you know once I have code ready for review.

WinCPP:
@dotnet-bot Please test OSX10.12 x64 Checked Build and Test

WinCPP:
@danmosemsft @safern @karelz @jkotas Please review the latest commit, referring to [this](https://github.com/dotnet/coreclr/pull/10203#issuecomment-291930763) comment for more information.

Between, I am raising another PR in CoreFx. Also I believe same changes need to be reflected in CoreRt repository as well...?

danmosemsft:
@WinCPP your timing table (newremove and newremoveget) were taken with the exact code you have in PR right now?

danmosemsft:
@jkotas to see timings.

WinCPP:
@danmosemsft @jkotas I'm sorry. I haven't yet pushed the changes... they are in committed state in my local copy. Oops.

Between, I wanted to check. @danmosemsft as you suggested, I have moved the following line,
```C#
value = default(TValue);
```
to end of the new overload Remove implementation to avoid double assignment to value for the 'key exists' path. So do we want fresh perf data with this minor change...?

WinCPP:
@danmosemsft the new source is in... sorry.

safern:
> Between, I am raising another PR in CoreFx. Also I believe same changes need to be reflected in CoreRt repository as well...?

Yes this changes should be in CoreRT as well.

WinCPP:
@safern Yup I will do that. I just wanted to confirm as I saw similar code in CoreRt and had similar in past... but I just wanted to be sure :)

danmosemsft:
@wincpp yes it would do no harm to run your perf test program again.

WinCPP:
@danmosemsft Yup I'm trying to get that done. I updated and sync'd my branch to dotnet/master and somehow my entire corefx repo is now referring to version 4.0 giving weird compilation errors. Don't know what is happening. I've been seeing v4.0 issues and xunit issues after I installed v5.0 as per [link](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/performance-tests.md) and slowly things started falling apart... and today almost everything is almost corrupt :(

WinCPP:
@danmosemsft @karelz ... I created entirely new local CoreClr and CoreFx sandboxes and built them. CoreClr builds fine as x64 Debug as well as Release. CoreFx builds successfully as x64 Debug (using x64 CoreClr Release). However the release build is still giving same errors. So seems like it was nothing to do with my sandboxes (or with v5.0 update)... Are the following errors related to `EventCounter` already known?

I think for now I will proceed with CoreClr Release build and CoreFx Debug build for the `Remove(K, V)` performance testing...

```
System\Diagnostics\Tracing\EventCounter.cs(239,26): error CS0436: The type 'EventCounter' in 'D:\WinCPP\corefx\src\System.Diagnostics.Tracing\src\System\Diagnostics\Tracing\EventCounter.cs' conflicts with the imported type 'EventCounter' in 'System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e'. Using the type defined in 'D:\WinCPP\corefx\src\System.Diagnostics.Tracing\src\System\Diagnostics\Tracing\EventCounter.cs'. [D:\WinCPP\corefx\src\System.Diagnostics.Tracing\src\System.Diagnostics.Tracing.csproj]
System\Diagnostics\Tracing\EventCounter.cs(270,71): error CS0436: The type 'EventCounter' in 'D:\WinCPP\corefx\src\System.Diagnostics.Tracing\src\System\Diagnostics\Tracing\EventCounter.cs' conflicts with the imported type 'EventCounter' in 'System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e'. Using the type defined in 'D:\WinCPP\corefx\src\System.Diagnostics.Tracing\src\System\Diagnostics\Tracing\EventCounter.cs'. [D:\WinCPP\corefx\src\System.Diagnostics.Tracing\src\System.Diagnostics.Tracing.csproj]
System\Diagnostics\Tracing\EventCounter.cs(229,31): error CS0436: The type 'EventCounter' in 'D:\WinCPP\corefx\src\System.Diagnostics.Tracing\src\System\Diagnostics\Tracing\EventCounter.cs' conflicts with the imported type 'EventCounter' in 'System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e'. Using the type defined in 'D:\WinCPP\corefx\src\System.Diagnostics.Tracing\src\System\Diagnostics\Tracing\EventCounter.cs'. [D:\WinCPP\corefx\src\System.Diagnostics.Tracing\src\System.Diagnostics.Tracing.csproj]
```

WinCPP:
@danmosemsft @karelz Furthermore, I am getting follow error related 'PerfRunner.exe` not present... I'm doing performance testing as per the guidelines for first time. Last time, I had written a custom application that used stop watch... Kindly advise how I should get around the following error. Thanks!
```
The application to execute does not exist: 'D:\WinCPP\corefx\bin\AnyOS.AnyCPU.Release\System.Collections.Tests\netcoreapp\PerfRunner.exe'
```

danmosemsft:
@brianrob @mellinoe

jkotas:
> System\Diagnostics\Tracing\EventCounter.cs(239,26): 

This is transient mismatch between CoreLib and CoreFX. It will get cleared with the next time CoreCLR gets updated in CoreFX. To workaround it, you can remove EventCounter.cs from `src\System.Diagnostics.Tracing\src\System.Diagnostics.Tracing.csproj`.

WinCPP:
Thanks @jkotas . I will try the fix. But then I have the PerfRunner.exe issue too..

@danmosemsft @jkotas Using my previous hackish way, I have performance figures for the `Remove(Key)` and `Remove(Key, out Value)` API with latter being a code copy of former, as discussed above, with additional statement to set the output parameter on key found / not found paths. Note that `BigData` is a structure of 8 ints (ref. [previous](https://github.com/dotnet/coreclr/pull/10203#issuecomment-287538615) performance data)

```
-----------------------------------------------------------
Run    Dictionary<int, int>        Dictionary<int, BigData>
-----------------------------------------------------------
      Remove(K)  Remove(K,V)        Remove(K)  Remove(K,V)
 1       23          31                 25          25
 2       22          24                 24          25
 3       24          24                 24          28
 4       22          24                 24          25
 5       22          24                 24          25
 6       37          24                 24          25
 7       22          24                 24          24
 8       22          24                 24          25
 9       22          24                 24          25
10       22          24                 24          31
11       23          24                 24          32
12       22          39                 24          25
13       22          24                 24          25
14       22          24                 24          31
15       37          24                 24          32
16       22          24                 24          24
17       22          24                 24          32
18       22          27                 24          25
19       22          26                 58          25
20       22          24                 24          25
21       22          24                 24          25
22       22          24                 24          25
23       22          24                 31          25
24       22          24                 32          25
25       24          24                 25          32
26       22          55                 24          24
27       22          24                 24          25
28       22          27                 24          25
29       22          24                 24          25
30       22          24                 24          31
31       22          24                 24          32
32       22          24                 24          24
33       22          24                 24          25
34       22          24                 24          31
35       22          24                 24          32
36       22          24                 24          25
37       22          30                 25          25
38       22          24                 24          25
39       22          24                 24          25
40       22          24                 24          25
-----------------------------------------------------------
Avg     22.9       25.675              25.3       26.625
StdDev  3.27        5.43                5.49       2.94
```

WinCPP:
Closed the PR by wrong empty force push. I have added [this](https://github.com/WinCPP/coreclr/blob/2ebcd77ed09428ca4456e6cd3b48f783be0f5a13/src/mscorlib/src/System/Collections/Generic/Dictionary.cs#L635) line to assign default value before throwing null key exception.

src/mscorlib/src/System/Collections/Generic/Dictionary.cs
```diff
@@ -628,6 +628,49 @@ public bool Remove(TKey key)
...
 
+        public bool Remove(TKey key, out TValue value)
+        {
+            if (key == null)
+            {
+                value = default(TValue);
+                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.key);
+            }
+
+            if (buckets != null)
```
safern: We don't need to duplicate this for loop code, we have a private method called `FindEntry` which returns the index to that entry and if it is not found it returns -1 but it won't update the last value so we could probably add a `FindEntryRemove(key)` where it finds the entry and does the bucket/entries update and then return the value... something like:

```cs
private int FindEntryRemove(TKey key)
{
       if (buckets != null)
       {
           int hashCode = comparer.GetHashCode(key) & 0x7FFFFFFF;
           int bucket = hashCode % buckets.Length;
           int last = -1;
           for (int i = buckets[bucket]; i >= 0; last = i, i = entries[i].next)
           {
               if (entries[i].hashCode == hashCode && comparer.Equals(entries[i].key, key))
               {
                    if (last < 0)
                   {
                        buckets[bucket] = entries[i].next;
                   }
                   else
                   {
                        entries[last].next = entries[i].next;
                   }
                   return i;
               }
            }
       }
       return -1;
}
```

And then both removes could be implemented around that private API, something like:

```cs
int i = FindEntryRemove(key);
if (i > 0)
{
    value = entries[i].value;
    FreeEntry(i);
    return true;
}
value = default(TValue);
return false;
```

And we could add a private method called FreeEntry(int index), that basically does:

```cs
entries[i].hashCode = -1;
entries[i].next = freeList;
entries[i].key = default(TKey);
entries[i].value = default(TValue);
freeList = i;
freeCount++;
version++;
```

@jkotas @danmosemsft any thoughts?

jkotas: I am not sure whether it would be a win. You would need to collect throughput and size data to tell what is better. I am ok with the implementation in one big method as is.

I would rather look into how to make the code for this implementation smaller. For example, caching pointer to `entries[i]` in a local may help to reduce native code size speed and `ref Entry entry = ref entries[i];`

WinCPP: @safern I believe this is in the pursuit of reducing duplicated code. But then the duplication, although restricted to few lines, happens between `FindEntry` and `FindEntryAndRemove` for the _finding_ part. Unless we want to merge them and pass a `boolean` to indicate 'remove' that will default to `false`...? ... potentially introducing a branch on the paths where existing `FindEntry` is used. Either ways, the method `FindEntryAndRemove` still _feels_ incomplete from design point of view because we have removed but not free'd the entry and freeing part comes as post fix...

@jkotas 
1. Curiosity question, wouldn't the C# compiler perform local CSE on the four lines which fetch the entry ... i.e. `entry[i].*` lines... with effect being that the entry is computed only once? This is based on my experience with C / C++ compilers though...
1. How does the usage of `default(TValue)` work? Does it initialize and then do a memberwise copy to destination (due to operator =) or, does it directly construct the default into memory referenced by the destination? If it is latter, does it still do memberwise init or it does sort of memset to zeros using span APIs? Thought in my head: cycles spent in memberwise init to default for a value type will be proportional to its size.

jkotas: C# compiler does very few trivial optimizations only. Optimizations like CSE are done by the JIT. Whether or not the optimization kicks in depends on number of factors. The best way to know for sure is to look the disassembly of the generated code and measure.

One thing to keep in mind is that the array indexing in C# is bound-checked. It restricts the optimizations that the JIT can do transparently, for example it cannot do the bound-check speculatively because of it would result in spurious exceptions.

I have glanced over the native code generated for `Dictionary<int,int>.Remove`. The CSE for `entry[i]` kicks in some cases, but not others. If you cache `ref Entry entry = ref entries[i];` at the start of the loop and replace all 9 occurrences of `entries[i]` with it, the native code generated for this method should be approximately 20% smaller.

jkotas: Either way, I think this one is fine to merge as is, and the optimizations for both the new and existing methods should be done in separate PR.

jkotas: Opened https://github.com/dotnet/coreclr/issues/10883

src/mscorlib/src/System/Collections/Generic/Dictionary.cs
```diff
@@ -628,6 +628,49 @@ public bool Remove(TKey key)
             return false;
         }
 
+        public bool Remove(TKey key, out TValue value)
+        {
+            if (key == null)
+            {
+                value = default(TValue);
```
jkotas: What is the point of setting the value on exception thrown? It is not how similar method works. For example, `Dictionary.TryGetValue` won't set the value on exception.

src/mscorlib/src/System/Collections/Generic/Dictionary.cs
WinCPP: @jkotas thanks for the elaborate answer. I helps me understand more internals. I had not given a thought that the `operator []` would do bounds checking... took it for plain simple C / C++ array access.

src/mscorlib/src/System/Collections/Generic/Dictionary.cs
WinCPP: @jkotas hmmm as I mentioned earlier, in the initial code I did leave out assigning default to the `out` parameter referring to some method but I am not able to find out which. I did the latest change based on `Dictionary.TryGetValue` itself which actually assigned default value on false path (and in fact there is no exception since it is a `Try` method)... please refer [here](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Collections/Generic/Dictionary.cs#L633).

So looked up MSDN for what should be the behavior of an `out` parameter in C# ([link](https://msdn.microsoft.com/en-us/library/t3c3bfhx.aspx)) and it says this...

> Although variables passed as out arguments do not have to be initialized before being passed, the called method is required to assign a value before the method returns.

And then this aligned with the `TryGetValue` behavior mentioned previously and also with COM which mentions that callee is supposed to assign value to out parameter even on E_FAIL, in which case it may be a NULL pointer... hence...

stephentoub:
> Although variables passed as out arguments do not have to be initialized before being passed, the called method is required to assign a value before the method returns.

This is only applicable when the method returns to its caller; if it instead throws an exception, this does not apply.

src/mscorlib/src/System/Collections/Generic/Dictionary.cs
stephentoub: > Although variables passed as out arguments do not have to be initialized before being passed, the called method is required to assign a value before the method returns.

This is only applicable when the method returns to its caller; if it instead throws an exception, this does not apply.

WinCPP: @stephentoub Thanks for the clarification. Removing.

jkotas:
> did the latest change based on Dictionary.TryGetValue itself which actually assigned default value on false path

Assigning on the false path is fine. I am questioning assigning on the exception throwing path. Dictionary.TryGetValue does not assign on exception throwing path either.

WinCPP:
@jkotas ... yup got it. Removing. Now I recollect, the exception was inside `FindEntry` based on which I had initially actually not included default assignment on exception path... but then I forgot how I had reached to that conclusion...

WinCPP:
Hmm... then as for the behavior of the API, can we say that the value in the `out` parameter will remain unaltered after the `Remove(K, out V)` if the API throws a null key exception? I could not find test case for `Dictionary.TryGetValue` where null key is passed... so wondering what should be the expected behavior. Or do we want to say behavior not defined...? I have one test case for the new overload for null key exception and I will modify that accordingly.

Between... just curiosity.... the other `Try*` methods that I have used ... `Enum.TryParse`, `Int32.TryParse` do not throw exceptions which is what I thought is the reason for 'Try*` methods... i.e. eliminate the exception handling blocks... I do appreciate that a null key is different thing altogether, but then the user said 'try' and the API tried, but there is nothing in there...

jkotas:
`Try*` does not apply to argument validation. If you pass invalid argument to the API, it will throw. TryParse methods are not different - for example, if you pass bad NumberStyles to int.TryParse, you will get an exception there as well.

stephentoub:
> then as for the behavior of the API, can we say that the value in the out parameter will remain unaltered after the Remove(K, out V) if the API throws a null key exception?

The value is undefined in such situations, but ideally in most situations if the target previously had a value assigned it would remain unchanged. From a language perspective, `out` is only considered an assignment for definite assignment purposes if the call returns rather than throws.

WinCPP:
Thanks @jkotas @stephentoub . I'll update the test case to expect that the contents of the out parameter remain unaltered...

WinCPP:
@safern @jkotas @stephentoub Please review the latest commit. Thanks.

src/mscorlib/src/System/Collections/Generic/Dictionary.cs
```diff
@@ -628,6 +628,48 @@ public bool Remove(TKey key)
             return false;
         }
 
+        public bool Remove(TKey key, out TValue value)
```
danmosemsft: Maybe a comment explaining that we're duplicating the code of Remove(TKey) for performance reasons. Othrwise someone will want to refactor it later...

WinCPP: Hmmm... But then @jkotas has opened #10883 ... would that cover this as well...?

jkotas: #10883 is not about eliminating the code duplication between the different overloads.

WinCPP: Oops ok. @danmosemsft @jkotas I have added a comment above both overloads. Kindly review the same.

danmosemsft:
Looks good thanks for doing this @WinCPP !

Nit: if you'd made the comment addition in a 2nd commit, rather than squashing locally, we could easily see it was the only change you make. When we merge, there is a squash button so there's no need to squash everything yourself.

danmosemsft:
I'll merge when CI's green.

danmosemsft:
@dotnet-bot test tizen armel cross debug build (https://github.com/dotnet/coreclr/issues/10897)

danmosemsft:
@dotnet-bot test tizen armel cross release build (https://github.com/dotnet/coreclr/issues/10897)

WinCPP:
@danmosemsft  Oops ok. Will keep that in mind next time onwards.

danmosemsft:
Ignoring Tizen as per #10897

omariom:
@jkotas
> Yes, size of the type - that gets multipled for generic types. 

Even when the method doesn't use any of the generic params? Like `return 1;`

jkotas:
Yes, it is the case today.

PR #10205, opened 2017-03-15
Fix debugging on OS X Sierra on 1.0.x

mikem8361:
Issue #10204

The output of the vmmap changed between OS X versions. Changed the sscanf formatting to deal with both old/new formats.

### DIFF
--- a/src/pal/src/thread/process.cpp
+++ b/src/pal/src/thread/process.cpp
@@ -2666,6 +2666,12 @@ CreateProcessModules(
     // VM_ALLOCATE            0000000105bac000-0000000105bad000 [    4K] r--/rw- SM=SHM
     // MALLOC (admin)         0000000105bad000-0000000105bae000 [    4K] r--/rwx SM=ZER
     // MALLOC                 0000000105bae000-0000000105baf000 [    4K] rw-/rwx SM=ZER
+
+    // OS X Sierra (10.12.4 Beta)
+    // REGION TYPE                      START - END             [ VSIZE  RSDNT  DIRTY   SWAP] PRT/MAX SHRMOD PURGE    REGION DETAIL
+    // Stack                  00007fff5a930000-00007fff5b130000 [ 8192K    32K    32K     0K] rw-/rwx SM=PRV          thread 0
+    // __TEXT                 00007fffa4a0b000-00007fffa4a0d000 [    8K     8K     0K     0K] r-x/r-x SM=COW          /usr/lib/libSystem.B.dylib
+    // __TEXT                 00007fffa4bbe000-00007fffa4c15000 [  348K   348K     0K     0K] r-x/r-x SM=COW          /usr/lib/libc++.1.dylib
     char *line = NULL;
     size_t lineLen = 0;
     int count = 0;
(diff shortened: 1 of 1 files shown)


ckiszka:
@mikem8361 -- First, thank you for resolving this issue. I've been following it for a hot minute. Is this patch in the daily builds or do I have to clone and build?

mikem8361:
>>>>@mikem8361<https://github.com/mikem8361> -- First, thank you for resolving this issue. I've been following it for a hot minute. Is this patch in the daily builds or do I have to clone and build?

@leecow Are there daily builds of the 1.0/1.1 service branches?

leecow:
Looks like it. Here are the Keep in mind these should be used only for validation only and are not yet signed.

[1.0](https://dotnetcli.blob.core.windows.net/dotnet/preview/Installers/1.0.5-servicing-004818-00/dotnet-osx-x64.1.0.5-servicing-004818-00.pkg) / [1.1](https://dotnetcli.blob.core.windows.net/dotnet/release/1.1.0/Installers/1.1.2-servicing-001422-00/dotnet-osx-x64.1.1.2-servicing-001422-00.pkg)

ckiszka:
Thanks guys. I really appreciate your efforts. It looks like I must be suffering from a different issue. I'm going to try and rollback my OS X version. I keep getting (below) this error no matter what I do.

-------------------------------------------------------------------
You may only use the Microsoft .NET Core Debugger (vsdbg) with
Visual Studio Code, Visual Studio or Visual Studio for Mac software
to help you develop and test your applications.
-------------------------------------------------------------------
Unable to attach to CoreCLR.

mikem8361:
That looks like the symptom (“Unable to attach to CoreCLR”) of the issue I fixed. Did you copy libdbgshim.dylib over the one VSCode/C# extension provides? The fix is in this component that is part of VS (specifically the C# VSCode extension.

@greggmiskelly can tell you the exact path I think.

ckiszka:
I did not copy the libdbgshim.dylib. Let me research on where this file is located.

ckiszka:
I just noticed you're referencing VS Code. Im using Visual Studio for Mac IDE. I assume the fix is different?

mikem8361:
Assume it was VSCode because I didn’t release we have VS for Mac.  The fix is the same but I’m not sure of the location of libdbgshim.dylib.

ckiszka:
I found the location. Where do I pickup the latest file?

ckiszka:
Again, thanks for the direction man.

mikem8361:
The latest file is in the packages Lee sent the links to.  Install the 1.0 or 1.1 pkg (you may have already done that) and copy the libdbgshim.dylib out there.

ckiszka:
BOOM! You're the man! Thank you. My problem is resolved. However, for rookies like me, you need to "show package contents" on the Visual Studio for Mac IDE app and the latest Microsoft .NET Core in "/usr/local/share/dotnet/shared". Replace the "libdbgshim.dylib" from the latest .NET Core over the Visual Studio for Mac version. 

Thanks again @mikem8361!!!!

PR #10210, opened 2017-03-15
Make user specify RID for publishing test native bins into

wtgodbe:


gkhanna79:
Wouldnt we need to update the build definitions to consume this?

wtgodbe:
@gkhanna79 yes, I plan on doing so

PR #10214, opened 2017-03-16
Take crash dumps for timed-out tests.

pgavlin:
This change updates the test infrastructure to collect crash dumps for
tests that time out. We've been seeing a number of tests that only time
out in CI (e.g. #10076); hopefully this will help us root-cause the
timeouts.

### DIFF
--- a/tests/src/Common/Coreclr.TestWrapper/CoreclrTestWrapperLib.cs
+++ b/tests/src/Common/Coreclr.TestWrapper/CoreclrTestWrapperLib.cs
@@ -7,12 +7,89 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Win32.SafeHandles;
 
 namespace CoreclrTestLib
 {
+    static class DbgHelp
+    {
+        public enum MiniDumpType : int
+        {
+            MiniDumpNormal                          = 0x00000000,
+            MiniDumpWithDataSegs                    = 0x00000001,
+            MiniDumpWithFullMemory                  = 0x00000002,
+            MiniDumpWithHandleData                  = 0x00000004,
+            MiniDumpFilterMemory                    = 0x00000008,
+            MiniDumpScanMemory                      = 0x00000010,
+            MiniDumpWithUnloadedModules             = 0x00000020,
+            MiniDumpWithIndirectlyReferencedMemory  = 0x00000040,
+            MiniDumpFilterModulePaths               = 0x00000080,
+            MiniDumpWithProcessThreadData           = 0x00000100,
+            MiniDumpWithPrivateReadWriteMemory      = 0x00000200,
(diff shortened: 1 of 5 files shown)


pgavlin:
@RussKeldorph @wtgodbe @schaabs PTAL

pgavlin:
Ping.

tests/src/Common/Coreclr.TestWrapper/CoreclrTestWrapperLib.cs
```diff
@@ -22,6 +99,63 @@ public class CoreclrTestWrapperLib
...
+                var processEntry = new Kernel32.ProcessEntry32 { Size = sizeof(Kernel32.ProcessEntry32) };
+
+                bool success = Kernel32.Process32First(snapshot, ref processEntry);
+                while (success)
+                {
+                    if (processEntry.ParentProcessID == ppid)
+                    {
+                        try
+                        {
+                            Process c = Process.GetProcessById(processEntry.ProcessID);
```
RussKeldorph: Would `using` be appropriate here?

pgavlin: No: the value may be returned via the `child` out parameter; using would dispose it before the return.

tests/src/Common/Coreclr.TestWrapper/CoreclrTestWrapperLib.cs
```diff
@@ -22,6 +99,63 @@ public class CoreclrTestWrapperLib
...
+            using (var crashDump = File.OpenWrite(path))
+            {
+                var flags = DbgHelp.MiniDumpType.MiniDumpWithFullMemory | DbgHelp.MiniDumpType.MiniDumpIgnoreInaccessibleMemory;
+                return DbgHelp.MiniDumpWriteDump(process.Handle, process.Id, crashDump.SafeFileHandle, flags, IntPtr.Zero, IntPtr.Zero, IntPtr.Zero);
+            }
+        }
+
+        static unsafe bool TryFindChildProcessByName(Process process, string childName, out Process child)
+        {
+            IntPtr snapshot = Kernel32.CreateToolhelp32Snapshot(Kernel32.Toolhelp32Flags.TH32CS_SNAPPROCESS, 0);
```
RussKeldorph: Is SafeHandle appropriate here, or is that useless since no Thread.Abort?

pgavlin: We could use a `SafeHandle`; it just requires implementing a new derived class of the appropriate parent (in this case `SafeHandleMinusOneIsInvalid`).

tests/src/Common/Coreclr.TestWrapper/CoreclrTestWrapperLib.cs
RussKeldorph: Ah, right.  Sorry.  Are you missing a Dispose in the caller, then?  Not sure if you subscribe to the philosophy that resource management is stupid on the exit path of an app.

Issue #10049, opened 2017-03-09
JIT: improve aliasing model for implicit-by-ref struct passing


JosephTremoulet:
There are cases on both ARM64 and Windows x64 where the calling convention dictates that a struct which is passed by value at source be passed by reference (after making a copy in the caller's frame) in the generated code.  Today we do this rewrite in the importer, and the callee's view of such parameters is the same as normal byrefs.  This inhibits promoting these structs, removing redundant loads of their fields, etc.  We'd get better codegen if we either modeled these as by-value until after later in the phase list or introduced internal annotations in the jit so that we could model these byref pointers as both non-null and restrict.

russellhadley:
Keep in 2.0 for Span on Kestrel.

JosephTremoulet:
Seeing as how we don't do much optimization on structs that we don't promote, I think the task here should really be to enable promotion of implicit-by-ref struct parameters (which conceptually we could do by copying the parameter to a local struct on method entry, rewriting all uses of the parameter to be uses of the local, and promoting the local).

JosephTremoulet:
Turns out we make struct promotion decisions before we do address-taken analysis; a promoted struct that's address-taken just has its fields laid out together on the frame and the codegen basically looks like it wasn't promoted.  So inserting a copy at function entry would just add a useless mem-to-mem copy in those cases.

JosephTremoulet:
Looks like the address-taken analysis, implicit-by-ref marking, and struct promotion should be reorderable, so I'll try swapping them around and find out if I'm right about that.

ahsonkhan:
@JosephTremoulet, any updates on this issue?

JosephTremoulet:
Per my previous notes, I spent some time working on running things in this order:

  1. address-taken analysis
  2. implicit by-ref rewrite
  3. struct promotion

It ran into issues where we lost some optimization capability because we do field-sensitive address-taken analysis, which depends on running struct promotion first for the address-taken analysis' bookkeeping.  So I've switched and am working on running things in this order:

  1. struct promotion
  2. address-taken analysis
  3. implicit by-ref rewrite

I have that doing the right thing in some basic testcases, but am still working through broader correctness (currently hitting some issues with Vector types) and assessing CQ/TP impact.  (That work-in-progress is in #10453).

JosephTremoulet:
Resolved in #10453.

Issue #10031, opened 2017-03-08
Span<T> methods may not inline when called from shared methods


AndyAyersMS:
Several of `Span<T>`'s methods invoke methods on other generic classes, eg `ByReference<T>` or `Unsafe<T>`. If the span methods are in turn invoked from a method where `T` requires runtime lookup, this need for runtime lookup passes transitively to generic callees made by the span method. Unfortunately this transitive need blocks inlining of the span methods as CoreCLR is not able to support inlining methods that perform runtime lookups. Inlining is blocked even when the span methods are given the `AggressiveInlining` attribute.

We can get around this by implementing (where possible) the generic methods invoked by Span as jit intrinsics. For example the span indexer calls `Unsafe.Add<T>(ref T, int)`  and that method only depends on `sizeof(T)` which is is known to be 4 or 8 for shared methods (since sharing implies `T` is a ref type and so is pointer sized).

RussKeldorph:
@russellhadley @AndyAyersMS Can we get this assigned to someone or moved out of 2.0?

russellhadley:
Assigning you @AndyAyersMS.  You can triage it further.

AndyAyersMS:
Simple example: we'd like the span ctor and indexer to inline into `X<T>.C`, when `T` is a ref type.
```C#
using System;
using System.Runtime.CompilerServices;

class X<T>
{
    [MethodImpl(MethodImplOptions.NoInlining)]
    public static T C(T[] a)
    {
        var span = new Span<T>(a);
        T result = span[span.Length / 2];
        return result;
    }
}

class Y
{
    public static int Main(string[] args)
    {
        string[] vals = new string[] {"a", "b", "c", "d", "e"};
        string z = X<string>.C(vals);
        return z.Equals("c") ? 100 : -1;
    }
}
```
When `T` is a value type `X<T>` is not shared and we can inline the span methods. Eg for `X<int>`:
```
Inlines into 06000001 X`1[Int32][System.Int32]:C(ref):int
  [1 IL=0003 TR=000004 06000DB0] [aggressive inline attribute] System.Span`1[Int32][System.Int32]:.ctor(ref):this
    [0 IL=0004 TR=000086 060001F4] [FAILED: does not return] System.ThrowHelper:ThrowArgumentNullException(int)
    [2 IL=0055 TR=000057 06003F11] [below ALWAYS_INLINE size] System.Runtime.CompilerServices.JitHelpers:GetArrayData(ref):byref
[...]

PR #10216, opened 2017-03-16
Update SpanBench test

JosephTremoulet:
- Re-enable the tests that were disabled in e859c309.
 - Re-work how tests are invoked from the command line to require less
   boilerplate
     - Now each test has a single [Benchmark] entrypoint
     - That entrypoint invokes its test's single inner-loop by wrapping it
       in a lambda that it passes to a new helper Invoke method (shared
       across all tests) which handles the xunit vs. command-line
       differences..
     - Main finds the entrypoints by using reflection to search for the
       [Benchmark] attributes, so the explicit list of stringified test
       names is no longer needed and the command line will run the same
       set of tests that xunit-perf does.
 - The new SpanAPI tests now get invoked when this tests is run from the
   command line as well.
 - Add [NoInlining] to the SpanAPI tests' kernels.
 - Add some heap writes and conditional writes to prevent deadcode
   optimization from eliminating tests' kernels (assuming we don't do
   interprocedural constant propagation or deadcode, or store sinking and
   final value calculation, or PRE...).
 - Split the Index SpanAPI tests into one version that uses a
   loop-invariant index and thus should get hoisted out of the loop, and
   another version that uses variant indices.

### DIFF
--- a/tests/src/Common/test_dependencies/project.json
+++ b/tests/src/Common/test_dependencies/project.json
@@ -2,6 +2,7 @@
   "dependencies": {
     "Microsoft.NETCore.Platforms": "2.0.0-beta-25118-01",
     "Microsoft.Private.CoreFx.NETCoreApp": "4.4.0-beta-25118-01",
+    "System.Memory": "4.4.0-beta-25118-01",
     "System.Runtime.CompilerServices.Unsafe": "4.4.0-beta-25118-01"
   },
   "frameworks": {
--- a/tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
+++ b/tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
@@ -27,13 +27,15 @@ namespace Span
         const int FillAllIterations = 1;
         const int BaseIterations = 1;
 #else
+        // Appropriately-scaled iteration counts for the various benchmarks
         const int BubbleSortIterations = 100;
         const int QuickSortIterations = 1000;
         const int FillAllIterations = 100000;
         const int BaseIterations = 10000000;
 #endif
 
-        const int Size = 1024;
+        // Default length for arrays of mock input data
+        const int DefaultLength = 1024;
 
         // Helpers
         #region Helpers
@@ -44,47 +46,263 @@ namespace Span
             public T[] C0;
         }
 
-        /*[MethodImpl(MethodImplOptions.NoInlining)]
-        static void TestFillAllSpan(Span<byte> span)
+        // Copying the result of a computation to Sink<T>.Instance is a way
+        // to prevent the jit from considering the computation dead and removing it.
+        private sealed class Sink<T>
         {
-            for (int i = 0; i < span.Length; ++i) {
-                span[i] = unchecked((byte)i);
+            public T Data;
+            public static Sink<T> Instance = new Sink<T>();
+        }
+
+        // Use statics to smuggle some information from Main to Invoke when running tests
+        // from the command line.
+        static bool IsXunitInvocation = true; // xunit-perf leaves this true; command line Main sets to false
+        static int CommandLineInnerIterationCount = 0;   // used to communicate iteration count from BenchmarkAttribute
+                                                         // (xunit-perf exposes the same in static property Benchmark.InnerIterationCount)
+        static bool DoWarmUp; // Main sets this when calling a new benchmark routine
+
+
+        // Invoke routine to abstract away the difference between running under xunit-perf vs running from the
+        // command line.  Inner loop to be measured is taken as an Action<int>, and invoked passing the number
+        // of iterations that the inner loop should execute.
+        static void Invoke(Action<int> innerLoop, string nameFormat, params object[] nameArgs)
+        {
+            if (IsXunitInvocation)
+            {
+                foreach (var iteration in Benchmark.Iterations)
+                    using (iteration.StartMeasurement())
+                        innerLoop((int)Benchmark.InnerIterationCount);
+            }
+            else
+            {
+                if (DoWarmUp)
+                {
+                    // Run some warm-up iterations before measuring
+                    innerLoop(CommandLineInnerIterationCount);
+                    // Clear the flag since we're now warmed up (caller will
+                    // reset it before calling new code)
+                    DoWarmUp = false;
+                }
+
+                // Now do the timed run of the inner loop.
+                Stopwatch sw = Stopwatch.StartNew();
+                innerLoop(CommandLineInnerIterationCount);
+                sw.Stop();
+
+                // Print result.
+                string name = String.Format(nameFormat, nameArgs);
+                double timeInMs = sw.Elapsed.TotalMilliseconds;
+                Console.WriteLine("{0}: {1}ms", name, timeInMs);
             }
         }
 
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        static void TestFillAllArray(byte[] data)
+        // Helper for the sort tests to get some pseudo-random input
+        static int[] GetUnsortedData(int length)
         {
-            for (int i = 0; i < data.Length; ++i) {
-                data[i] = unchecked((byte)i);
-            }
-        }
-
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        static void TestFillAllReverseSpan(Span<byte> span)
-        {
-            for (int i = span.Length; --i >= 0;) {
-                span[i] = unchecked((byte)i);
-            }
-        }
-
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        static void TestFillAllReverseArray(byte[] data)
-        {
-            for (int i = data.Length; --i >= 0;) {
-                data[i] = unchecked((byte)i);
-            }
-        }
-
-        static int[] GetUnsortedData()
-        {
-            int[] unsortedData = new int[Size];
+            int[] unsortedData = new int[length];
             Random r = new Random(42);
             for (int i = 0; i < unsortedData.Length; ++i)
             {
                 unsortedData[i] = r.Next();
             }
             return unsortedData;
+        }
+        #endregion // helpers
+
+        // Tests that implement some vary basic algorithms (fill/sort) over spans and arrays
+        #region Algorithm tests
+
+        #region TestFillAllSpan
+        [Benchmark(InnerIterationCount = FillAllIterations)]
+        [InlineData(DefaultLength)]
+        public static void FillAllSpan(int length)
+        {
+            byte[] a = new byte[length];
+
+            Invoke((int innerIterationCount) =>
+            {
+                Span<byte> s = new Span<byte>(a);
+                for (int i = 0; i < innerIterationCount; ++i)
+                {
+                    TestFillAllSpan(s);
+                }
+            },
+            "TestFillAllSpan({0})", length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void TestFillAllSpan(Span<byte> span)
+        {
+            for (int i = 0; i < span.Length; ++i)
+            {
+                span[i] = unchecked((byte)i);
+            }
+        }
+        #endregion
+
+        #region TestFillAllArray
+        [Benchmark(InnerIterationCount = FillAllIterations)]
+        [InlineData(DefaultLength)]
+        public static void FillAllArray(int length)
+        {
+            byte[] a = new byte[length];
+
+            Invoke((int innerIterationCount) =>
+            {
+                for (int i = 0; i < innerIterationCount; ++i)
+                {
+                    TestFillAllArray(a);
+                }
+            },
+            "TestFillArray({0})", length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void TestFillAllArray(byte[] data)
+        {
+            for (int i = 0; i < data.Length; ++i)
+            {
+                data[i] = unchecked((byte)i);
+            }
+        }
+        #endregion
+
+        #region TestFillAllReverseSpan
+        [Benchmark(InnerIterationCount = FillAllIterations)]
+        [InlineData(DefaultLength)]
+        public static void FillAllReverseSpan(int length)
+        {
+            byte[] a = new byte[length];
+
+            Invoke((int innerIterationCount) =>
+            {
+                Span<byte> s = new Span<byte>(a);
+                for (int i = 0; i < innerIterationCount; ++i)
+                {
+                    TestFillAllReverseSpan(s);
+                }
+            },
+            "TestFillAllReverseSpan({0})", length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void TestFillAllReverseSpan(Span<byte> span)
+        {
+            for (int i = span.Length; --i >= 0;)
+            {
+                span[i] = unchecked((byte)i);
+            }
+        }
+        #endregion
+
+        #region TestFillAllReverseArray
+        [Benchmark(InnerIterationCount = FillAllIterations)]
+        [InlineData(DefaultLength)]
+        public static void FillAllReverseArray(int length)
+        {
+            byte[] a = new byte[length];
+
+            Invoke((int innerIterationCount) =>
+            {
+                for (int i = 0; i < innerIterationCount; ++i)
+                {
+                    TestFillAllReverseArray(a);
+                }
+            },
+            "TestFillAllReverseArray({0})", length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void TestFillAllReverseArray(byte[] data)
+        {
+            for (int i = data.Length; --i >= 0;)
+            {
+                data[i] = unchecked((byte)i);
+            }
+        }
+        #endregion
+
+        #region TestQuickSortSpan
+        [Benchmark(InnerIterationCount = QuickSortIterations)]
+        [InlineData(DefaultLength)]
+        public static void QuickSortSpan(int length)
+        {
+            int[] data = new int[length];
+            int[] unsortedData = GetUnsortedData(length);
+
+            Invoke((int innerIterationCount) =>
+            {
+                Span<int> span = new Span<int>(data);
+
+                for (int i = 0; i < innerIterationCount; ++i)
+                {
+                    Array.Copy(unsortedData, data, length);
+                    TestQuickSortSpan(span);
+                }
+            },
+            "TestQuickSortSpan({0})", length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void TestQuickSortSpan(Span<int> data)
+        {
+            if (data.Length <= 1)
+            {
+                return;
+            }
+
+            int lo = 0;
+            int hi = data.Length - 1;
+            int i, j;
+            int pivot, temp;
+            for (i = lo, j = hi, pivot = data[hi]; i < j;)
+            {
+                while (i < j && data[i] <= pivot)
+                {
+                    ++i;
+                }
+                while (j > i && data[j] >= pivot)
+                {
+                    --j;
+                }
+                if (i < j)
+                {
+                    temp = data[i];
+                    data[i] = data[j];
+                    data[j] = temp;
+                }
+            }
+            if (i != hi)
+            {
+                temp = data[i];
+                data[i] = pivot;
+                data[hi] = temp;
+            }
+
+            TestQuickSortSpan(data.Slice(0, i));
+            TestQuickSortSpan(data.Slice(i + 1));
+        }
+        #endregion
+
+        #region TestBubbleSortSpan
+        [Benchmark(InnerIterationCount = BubbleSortIterations)]
+        [InlineData(DefaultLength)]
+        public static void BubbleSortSpan(int length)
+        {
+            int[] data = new int[length];
+            int[] unsortedData = GetUnsortedData(length);
+
+            Invoke((int innerIterationCount) =>
+            {
+                Span<int> span = new Span<int>(data);
+
+                for (int i = 0; i < innerIterationCount; i++)
+                {
+                    Array.Copy(unsortedData, data, length);
+                    TestBubbleSortSpan(span);
+                }
+            },
+            "TestBubbleSortSpan({0})", length);
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
@@ -93,10 +311,13 @@ namespace Span
             bool swap;
             int temp;
             int n = span.Length - 1;
-            do {
+            do
+            {
                 swap = false;
-                for (int i = 0; i < n; i++) {
-                    if (span[i] > span[i + 1]) {
+                for (int i = 0; i < n; i++)
+                {
+                    if (span[i] > span[i + 1])
+                    {
                         temp = span[i];
                         span[i] = span[i + 1];
                         span[i + 1] = temp;
@@ -107,6 +328,84 @@ namespace Span
             }
             while (swap);
         }
+        #endregion
+
+        #region TestQuickSortArray
+        [Benchmark(InnerIterationCount = QuickSortIterations)]
+        [InlineData(DefaultLength)]
+        public static void QuickSortArray(int length)
+        {
+            int[] data = new int[length];
+            int[] unsortedData = GetUnsortedData(length);
+
+            Invoke((int innerIterationCount) =>
+            {
+                for (int i = 0; i < innerIterationCount; i++)
+                {
+                    Array.Copy(unsortedData, data, length);
+                    TestQuickSortArray(data, 0, data.Length - 1);
+                }
+            },
+            "TestQuickSortArray({0})", length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void TestQuickSortArray(int[] data, int lo, int hi)
+        {
+            if (lo >= hi)
+            {
+                return;
+            }
+
+            int i, j;
+            int pivot, temp;
+            for (i = lo, j = hi, pivot = data[hi]; i < j;)
+            {
+                while (i < j && data[i] <= pivot)
+                {
+                    ++i;
+                }
+                while (j > i && data[j] >= pivot)
+                {
+                    --j;
+                }
+                if (i < j)
+                {
+                    temp = data[i];
+                    data[i] = data[j];
+                    data[j] = temp;
+                }
+            }
+            if (i != hi)
+            {
+                temp = data[i];
+                data[i] = pivot;
+                data[hi] = temp;
+            }
+
+            TestQuickSortArray(data, lo, i - 1);
+            TestQuickSortArray(data, i + 1, hi);
+        }
+        #endregion
+
+        #region TestBubbleSortArray
+        [Benchmark(InnerIterationCount = BubbleSortIterations)]
+        [InlineData(DefaultLength)]
+        public static void BubbleSortArray(int length)
+        {
+            int[] data = new int[length];
+            int[] unsortedData = GetUnsortedData(length);
+
+            Invoke((int innerIterationCount) =>
+            {
+                for (int i = 0; i < innerIterationCount; i++)
+                {
+                    Array.Copy(unsortedData, data, length);
+                    TestBubbleSortArray(data);
+                }
+            },
+            "TestBubbleSortArray({0})", length);
+        }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
         static void TestBubbleSortArray(int[] data)
@@ -114,10 +413,13 @@ namespace Span
             bool swap;
             int temp;
             int n = data.Length - 1;
-            do {
+            do
+            {
                 swap = false;
-                for (int i = 0; i < n; i++) {
-                    if (data[i] > data[i + 1]) {
+                for (int i = 0; i < n; i++)
+                {
+                    if (data[i] > data[i + 1])
+                    {
                         temp = data[i];
                         data[i] = data[i + 1];
                         data[i + 1] = temp;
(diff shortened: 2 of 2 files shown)


JosephTremoulet:
@AndyAyersMS @ahsonkhan PTAL.  This seems to be exposing some bug (the checked jit asserts when compiling the tests that I'm un-commenting here), so not ready to merge until I track that down, but it's looking like a jit bug so I think this test change is complete enough for review.
/cc @dotnet/jit-contrib

JosephTremoulet:
These lab failures are not what I expected:

```
08:22:46   Unhandled Exception: System.IO.FileNotFoundException: Could not load file or assembly 'System.Memory, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
08:22:46            at Span.SpanBench.TestSpanFill[T](T[] array, Int32 iterationCount)
08:22:46            at Span.SpanBench.<>c__DisplayClass74_0`1.<InvokeTestSpanFill>b__0(Int32 innerIterationCount) in D:\j\workspace\x86_checked_w---b3a226f6\tests\src\JIT\Performance\CodeQuality\Span\SpanBench.cs:line 905
08:22:46            at Span.SpanBench.Invoke(Action`1 innerLoop, String nameFormat, Object[] nameArgs) in D:\j\workspace\x86_checked_w---b3a226f6\tests\src\JIT\Performance\CodeQuality\Span\SpanBench.cs:line 81
08:22:46            at Span.SpanBench.InvokeTestSpanFill[T](Int32 length) in D:\j\workspace\x86_checked_w---b3a226f6\tests\src\JIT\Performance\CodeQuality\Span\SpanBench.cs:line 905
08:22:46            at Span.SpanBench.TestSpanFillByte(Int32 length) in D:\j\workspace\x86_checked_w---b3a226f6\tests\src\JIT\Performance\CodeQuality\Span\SpanBench.cs:line 888
08:22:46            at Span.SpanBench.Main(String[] args) in D:\j\workspace\x86_checked_w---b3a226f6\tests\src\JIT\Performance\CodeQuality\Span\SpanBench.cs:line 1130
```

I'd say the references are messed up, but somehow we're building fine and then failing to run... since the SpanAPI tests weren't previously run when invoked from the command line, it's possible it's been like this for a while but the tests just haven't been running.  It runs when I build locally, though.  I'm investigating, but if anybody has suggestions I'm all ears.

AndyAyersMS:
I believe there is a separate place where test runtime dependencies are declared, though this area has been changing recently so my understanding may outdated.

You might see if tests/src/Common/test_dependencies/project.lock.json has been updated locally or needs to be updated (there maybe some explicit way to update this now).

 @wtgodbe any ideas?

JosephTremoulet:
> I believe there is a separate place where test runtime dependencies are declared

Ah, thanks for the tip.  I can indeed repro now with a fresh sync and a clean enlistment -- looks like GenerateLayout is no longer including a copy of System.Memory.dll.  Must have been a recent change, since this worked for me a day or two ago...

JosephTremoulet:
I tried adding `"System.Memory": "4.4.0-beta-25114-01"` (which is what we have in [benchmarks/project.json](https://github.com/dotnet/coreclr/blob/7d8e1382eea401386890bc38065fa9b8ef76141e/tests/src/JIT/config/benchmark/project.json#L13)) to [test_dependencies](https://github.com/dotnet/coreclr/blob/7d8e1382eea401386890bc38065fa9b8ef76141e/tests/src/Common/test_dependencies/project.json#L2).  I have no idea if that's a reasonable thing to do, but it did change my failure mode from a MissingFileException to `System.TypeLoadException: A value type containing a by-ref instance field, such as Span<T>, cannot be used as the type for a class instance field.`.  That's in fact a perfectly reasonable error with my change, so this feel like progress.  I'll rewrite to avoid trying to declare spans as instance members of ref classes...

JosephTremoulet:
> I'll rewrite to avoid trying to declare spans as instance members of ref classes

That turns out to be harder than I thought.  I got rid of the `Sink<Span<T>>` cases, but am still getting the same `TypeLoadException`.  I think it's complaining about the span members of the display classes that were generated for my lambdas.  This must be a known issue, right?  How do people usually deal with the situation that they wish their lambda could capture a Span?

ahsonkhan:
>  lambda could capture a Span?

Can a lambda capture a span (which is stack only)? cc @KrzysztofCwalina, @jkotas

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
```diff
@@ -44,315 +46,392 @@ private sealed class TestClass<T>
...
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        static void TestBubbleSortSpan(Span<int> span)
+        // Tests that implement some vary basic algorithms (fill/sort) over spans and arrays
+        #region Algorithm tests
+
+        #region TestFillAllSpan
+        [Benchmark(InnerIterationCount = FillAllIterations)]
+        [InlineData(DefaultLength)]
+        public static void FillAllSpan(int length)
         {
```
ahsonkhan: I think we can remove this test now since we have a Span.Fill method that does this, no? Or is this meant to test writing into a span as well using the indexer?

JosephTremoulet:
It turned out to be trivial to move all the lambda-captured spans into the lamdba methods themselves in this case.  I can now run the test locally again, and I've pushed the updates.

Still don't know if just adding `System.Memory` to `test_dependencies` like that is ok or not...

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
```diff
@@ -44,315 +46,392 @@ private sealed class TestClass<T>
...
+        // of iterations that the inner loop should execute.
+        static void Invoke(Action<int> innerLoop, string nameFormat, params object[] nameArgs)
         {
-            for (int i = span.Length; --i >= 0;) {
-                span[i] = unchecked((byte)i);
+            if (IsXunitInvocation)
+            {
+                foreach (var iteration in Benchmark.Iterations)
+                    using (iteration.StartMeasurement())
+                        innerLoop((int)Benchmark.InnerIterationCount);
```
ahsonkhan: This is interesting. How would we run the benchmarks now? None of the tests with "Benchmark" attributes have iteration.StartMeasurement(). I thought the methods with Benchmark attributes are the entry point.

jkotas:
> Can a lambda capture a span (which is stack only)?

No. Check https://github.com/dotnet/csharplang/pull/264 for list of other limitations.

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
```diff
@@ -447,82 +532,189 @@ public static void TestSpanDangerousGetPinnableReferenceByte(int length)
...
+
+            Invoke((int innerIterationCount) => TestSpanIndexHoistable<T>(array, length, innerIterationCount),
+                "TestSpanIndexHoistable<{0}>({1})", typeof(T).Name, length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void TestSpanIndexHoistable<T>(T[] array, int length, int iterationCount)
+        {
+            var sink = Sink<T>.Instance;
+            var span = new Span<T>(array);
```
ahsonkhan: Isn't this test also measuring the construction of Span from an array and not just indexer? When compared to the array indexer this would give worst results.

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
```diff
@@ -447,82 +532,189 @@ public static void TestSpanDangerousGetPinnableReferenceByte(int length)
...
+        {
+            var array = new T[length];
+            Invoke((int innerIterationCount) => TestArrayIndexHoistable<T>(array, length, innerIterationCount),
+                "TestArrayIndexHoistable<{0}>({1})", typeof(T).Name, length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void TestArrayIndexHoistable<T>(T[] array, int length, int iterationCount)
+        {
+            var sink = Sink<T>.Instance;
```
ahsonkhan: Is the execution time of `Sink<T>.Instance` and `Sink<Span<T>>.Instance` similar and negligible?

tests/src/Common/test_dependencies/project.json
```diff
@@ -2,6 +2,7 @@
   "dependencies": {
     "Microsoft.NETCore.Platforms": "2.0.0-beta-25114-01",
     "Microsoft.Private.CoreFx.NETCoreApp": "4.4.0-beta-25114-01",
+    "System.Memory": "4.4.0-beta-25114-01",
```
ahsonkhan: I am not sure exactly why this is needed. Isn't Span<T> also already defined in mscorlib?

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
```diff
@@ -412,13 +500,31 @@ public static void TestSpanDangerousCreateByte(int length)
...
-            foreach (var iteration in Benchmark.Iterations)
-                using (iteration.StartMeasurement())
-                    for (int i = 0; i < Benchmark.InnerIterationCount; i++)
-                        span = Span<string>.DangerousCreate(testClass, ref testClass.C0[0], testClass.C0.Length);
+            InvokeTestSpanDangerousCreate<string>(length);
+        }
+
+        static void InvokeTestSpanDangerousCreate<T>(int length)
+        {
+            TestClass<T> testClass = new TestClass<T>();
```
ahsonkhan: By doing this, how are we still testing the non-reference case of Span<byte>? Both Span<string> and Span<byte> go down this path.

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
```diff
@@ -447,82 +546,189 @@ public static void TestSpanDangerousGetPinnableReferenceByte(int length)
...
+            Invoke((int innerIterationCount) => TestSpanIndexVariant<T>(array, length, innerIterationCount),
+                "TestSpanIndexVariant<{0}>({1})", typeof(T).Name, length);
         }
 
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void TestSpanIndexVariant<T>(T[] array, int length, int iterationCount)
+        {
+            var sink = Sink<T>.Instance;
+            var span = new Span<T>(array);
+            int mask = (length < 2 ? 0 : (length < 8 ? 1 : 7));
```
ahsonkhan: How is the index variant for lengths >= 10 which is true for 3/4 test cases?

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
```diff
@@ -717,12 +953,23 @@ public static void TestSpanClearByte(int length)
...
+
+            Invoke((int innerIterationCount) => TestSpanClear<T>(array, innerIterationCount),
+                "TestSpanClear<{0}>({1})", typeof(T).Name, length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void TestSpanClear<T>(T[] array, int iterationCount)
+        {
+            var span = new Span<T>(array);
+            for (int i = 0; i < iterationCount; i++)
```
ahsonkhan: Measures construction + clear. Is the hopes/goal that since we clear "iterationCount" times that it would dwarf the constructor time? If not, when comparing to Array.Clear, span has a disadvantage.

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
```diff
@@ -781,14 +1028,33 @@ public static void TestSpanAsBytesByte(int length)
...
+            var sink = Sink<byte>.Instance;
+            var span = new Span<T>(array);
+
+            for (int i = 0; i < iterationCount; i++)
+            {
+                var byteSpan = span.AsBytes();
+                // Under a condition that we know is false but the jit doesn't,
+                // add a read from 'byteSpan' to make sure it's not dead, and an assignment
+                // to 'span' so the AsBytes call won't get hoisted.
+                if (untrue) { sink.Data = byteSpan[0]; span = new Span<T>(); }
```
ahsonkhan: This is interesting. I get the read from 'byteSpan', but why do we need to assign to 'span'? If we don't, will the call to AsBytes gets hoisted outside of the for loop and will only run once? Is that what "hoisted" means?

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
```diff
@@ -846,148 +1126,67 @@ public static void TestSpanSliceStringChar(int length)
...
-            for (int i = 0; i < QuickSortIterations; i++)
-            {
-                Array.Copy(unsortedData, data, Size);
-                TestQuickSortSpan(span);
-            }
-        }
-
-        static void BubbleSortSpanBase()
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void TestSpanSliceStringChar(string s, int iterationCount, bool untrue)
```
ahsonkhan: Propagate the function name change to methods that are called? i.e. Add 'Wrapper' to this method name?

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
```diff
@@ -846,148 +1126,67 @@ public static void TestSpanSliceStringChar(int length)
...
-            return args.Select(arg => new object[] { arg });
-        }
-
-        static IEnumerable<object[]> TestFuncs = MakeArgs(
-            /*"FillAllSpanBase", "FillAllArrayBase",
-            "FillAllReverseSpanBase", "FillAllReverseArrayBase",
-            "BubbleSortSpanBase", "BubbleSortArrayBase",
-            "QuickSortSpanBase", "QuickSortArrayBase"*/
-        );
+                BenchmarkAttribute benchAttr = m.GetCustomAttribute<BenchmarkAttribute>();
```
ahsonkhan: This is awesome :) Good to know.

ahsonkhan:
> Still don't know if just adding `System.Memory` to `test_dependencies` like that is ok or not...

For a standalone application, I had to include a PackageReference to System.Memory or else I would get compiler errors like this:
Program.cs(79,28): error CS0246: The type or namespace name 'Span<>' could not be found (are you missing a using directive or an assembly reference?) [C:\Users\ahkha\Documents\SpanDemo\FastSpan\spantest.csproj]

So, I would imagine this is necessary for the tests to run via command line. That would be my assumption. How come this was not necessary for when benchmarks run via xunit?

ahsonkhan:
> Split the Index SpanAPI tests into one version that uses a
loop-invariant index and thus should get hoisted out of the loop, and
another version that uses variant indices.

What is the motivation for this?

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
JosephTremoulet: xunit-performance calls the [Benchmark]-attributed methods with logic similar to what's in Main now.  Those methods now build a delegate and pass it to this method, which does iteration.StartMeasurement().  Note that the references to `Benchmark` in method bodies are referring to the `Xunit.Benchmark` class, not the `BenchmarkAttribute` that `[Benchmark]` instantiates; the static class fetches the data from the attributes (again, similar to what Main now does) and makes it available to the benchmark code via properties on the `Benchmark` class (much like what I'm doing with `InnerIterationCount` for the command-line case now).

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
JosephTremoulet: I think it's reasonable to expect people will keep writing open-coded loops like this and have them in the benchmark.  I'm not wedded to the idea, though, if you feel strongly.

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
JosephTremoulet: `TestSpanDangerousCreateByte` calls `InvokeTestSpanDangerousCreate<byte>` calls `Invoke` calls the lambda call `TestSpanDangerousCreate<byte>` calls `Span<byte>.DangerousCreate`.  Am I missing something?

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
JosephTremoulet: Right, when I wrapped the tests in lambdas to make them callable from the command line, I lost the ability to put setup code in the same method as the kernel but outside the measurement loop.  So then there was a decision to be made about whether the spans created like this should be created inside the kernel method or created outside the loop and passed in.  With the SpanAPI tests, I opted to create the spans as locals in the kernel methods; passing the spans in would make measuring on win-x64 skewed since it would see them as byrefs (pending fixing #10049), so I thought this would best mirror the codegen we want to be tracking.  My expectation is (as you surmised in a later comment) that the loop's run-time will dwarf out the span creation time because it iterates so many times.

All that said, for IndexHoistable in particular, we'd hope to get to a point that the kernel loop doesn't even include the store... at that point I'd hope the running-time is effectively zero and the difference between creating the span here and not doing it in the array case would be too small to measure.

For any of these tests that you feel it's important, we could split them back out and have one method that gets invoked from Xunit (which could have a span local that gets initialized outside the timing loop) and another that gets invoked from the command line (which would look like this and we'd hope the codegen was similar enough).  My inclination is not to do that.

Regarding the choice of passing the spans into the method or not, while I made the spans local to the kernel methods for the SpanAPI tests, I made them parameters in the other version.  I just thought it would be good to have both approaches measured, and now hopefully the other tests will see a bump when #10049 gets implemented.

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
JosephTremoulet: I believe so.  Any type initialization should get triggered during the warm-up runs, so then the measured code will just verify that the type is initialized and load the field.

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
JosephTremoulet: It follows the sequence 0,1,2,3,4,5,6,7,0,1,2,3,4,5,6,7,0,1,2,3,...  By "variant" here I just meant "not the same on each iteration of the loop" so that the JIT can't hoist the span access out of the loop.

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
JosephTremoulet: Yes, the loop will both iterate and zero out the whole span on each iteration; construction will just load the length out of the array.

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
JosephTremoulet: I haven't checked if we currently hoist AsBytes, but yes it could and yes that's what "hoisted" means and yes that's why I included the assignment to span.

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
JosephTremoulet: In the other tests, the [Benchmark] method is suffixed with a type and the kernel method is not.  Since the kernel method isn't generic in this case, there wasn't a type to suffix the [Benchmark] method with, so I added a "Wrapper" suffix to it.  It's a bit easier to inspect jitting of the kernel methods if they aren't overloads, is why I bothered.  That said, if you want them to have the same name, I'll change it, though wouldn't it be better to remove "Wrapper" from the other method?

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
JosephTremoulet: :)

ahsonkhan:
> Given the issues, I'm not wedded to keeping the Hoistable one if you think it would be better removed.

I am for keeping it. I just wanted to get an understanding of what insight we could get from that data point.

ahsonkhan:
> it's reasonable to expect people will keep writing open-coded loops

Agreed.

ahsonkhan:
> My inclination is not to do that.

I agree with this as well. As long as the benchmark results are reasonable (and comparisons valid), I see no reason to change the tests.

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
ahsonkhan: > it's reasonable to expect people will keep writing open-coded loops

Agreed.

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
ahsonkhan: > Am I missing something?

Nope, I was confused.

Question: In the call graph you specified, at what point do we start measuring?

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
ahsonkhan: > It's a bit easier to inspect jitting of the kernel methods if they aren't overloads, is why I bothered. 

Fair point. No change needed then.

ahsonkhan:
LGTM

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
```diff
@@ -364,12 +445,7 @@ public static void BubbleSortArray()
         [InlineData(1000)]
         public static void TestSpanConstructorByte(int length)
         {
-            var array = new byte[length];
-            Span<byte> span;
-            foreach (var iteration in Benchmark.Iterations)
-                using (iteration.StartMeasurement())
-                    for (int i = 0; i < Benchmark.InnerIterationCount; i++)
-                        span = new Span<byte>(array);
+            InvokeTestSpanConstructor<byte>(length);
```
ahsonkhan: I just learned (from @KrzysztofCwalina) how you could  run benchmark tests with Type passed in as InlineData and avoid having multiple type specific definitions (like TestSpanConstructorByte and TestSpanConstructorString). In essence, have generic benchmark methods.

So, instead of having:
```C#
...
public static void TestSpanConstructorByte(int length)
...
public static void TestSpanConstructorString(int length)
```
You could do this:
```C#
[Benchmark(InnerIterationCount = 100000)]
[InlineData(1, typeof(byte))]
[InlineData(10, typeof(byte))]
[InlineData(100, typeof(byte))]
[InlineData(1000, typeof(byte))]
[InlineData(1, typeof(string))]
[InlineData(10, typeof(string))]
[InlineData(100, typeof(string))]
[InlineData(1000, typeof(string))]
public static void TestSpanConstructor(int length, Type type)
{
    MethodInfo method = typeof(MemoryTests).GetMethod("InvokeTestSpanConstructor").MakeGenericMethod(type);
    object[] parameters = { length };
    method.Invoke(null, parameters);
    /*foreach (var iteration in Benchmark.Iterations)
    {
        using (iteration.StartMeasurement())
        {
            for (int i = 0; i < Benchmark.InnerIterationCount; i++)
                method.Invoke(null, parameters);
        }
    }*/
}
public static void InvokeTestSpanConstructor<T>(int length)
{
    var array = new T[length];
    Invoke((int innerIterationCount) => TestSpanConstructor<T>(array, innerIterationCount, false),
"TestSpanConstructor<{0}>({1})", typeof(T).Name, length);
}
```

@JosephTremoulet, do you refactoring the tests this way is a good idea?

ahsonkhan:
I tried to compare the results of the following:
**A:**
```C#
[Benchmark(InnerIterationCount = BaseIterations)]
[InlineData(1)]
public static void TestSpanConstructorByte1(int length)
{
    InvokeTestSpanConstructor<byte>(length);
}

[Benchmark(InnerIterationCount = BaseIterations)]
[InlineData(1)]
public static void TestSpanConstructorString1(int length)
{
    InvokeTestSpanConstructor<string>(length);
}

public static void InvokeTestSpanConstructor<T>(int length)
{
    var array = new T[length];
    Invoke((int innerIterationCount) => TestSpanConstructor<T>(array, innerIterationCount, false), "TestSpanConstructor<{0}>({1})", typeof(T).Name, length);
}

/*[MethodImpl(MethodImplOptions.NoInlining)]*/
static void TestSpanConstructor<T>(T[] array, int iterationCount, bool untrue)
{
    var sink = Sink<T>.Instance;

    for (int i = 0; i < iterationCount; i++)
    {
        var span = new Span<T>(array);
        // Under a condition that we know is false but the jit doesn't,
        // add a read from 'span' to make sure it's not dead, and an assignment
        // to 'array' so the constructor call won't get hoisted.
        if (untrue) { sink.Data = span[0]; array = new T[iterationCount]; }
    }
}
```
**B:**
```C#
[Benchmark(InnerIterationCount = BaseIterations)]
[InlineData(1, false)]
public static void TestSpanConstructorByte(int length, bool untrue)
{
    var array = new byte[length];
    var sink = Sink<byte>.Instance;
    foreach (var iteration in Benchmark.Iterations)
        using (iteration.StartMeasurement())
            for (int i = 0; i < InnerIterationCount; i++)
            {
                var span = new Span<byte>(array);
                if (untrue) { sink.Data = span[0]; array = new byte[InnerIterationCount]; }
            }
}

[Benchmark(InnerIterationCount = BaseIterations)]
[InlineData(1, false)]
public static void TestSpanConstructorString(int length, bool untrue)
{
    var array = new string[length];
    var sink = Sink<string>.Instance;
    foreach (var iteration in Benchmark.Iterations)
        using (iteration.StartMeasurement())
            for (int i = 0; i < InnerIterationCount; i++)
            {
                var span = new Span<string>(array);
                if (untrue) { sink.Data = span[0]; array = new string[InnerIterationCount]; }
            }
}
```
**C:**
```C#
[Benchmark(InnerIterationCount = BaseIterations)]
[InlineData(1, typeof(byte))]
[InlineData(1, typeof(string))]
public static void TestSpanConstructor(int length, Type type)
{
    MethodInfo method = typeof(MemoryTests).GetMethod("InvokeTestSpanConstructor").MakeGenericMethod(type);
    object[] parameters = { length };
    method.Invoke(null, parameters);
}
```

**Here are the results:**

     Test Name                                                                             | Metric                | Iterations |    AVERAGE |    STDEV.S |        MIN |        MAX
    :------------------------------------------------------------------------------------- |:--------------------- |:----------:| ----------:| ----------:| ----------:| ----------:
     System.Buffers.Tests.MemoryTests.TestSpanConstructor(length: 1, type: typeof(byte))   | Duration              |    101     |     48.790 |      7.526 |     44.325 |     78.395
     System.Buffers.Tests.MemoryTests.TestSpanConstructor(length: 1, type: typeof(string)) | Duration              |     20     |    540.724 |     63.090 |    480.255 |    684.930
     System.Buffers.Tests.MemoryTests.TestSpanConstructorByte1(length: 1)                  | Duration              |    101     |     49.833 |      9.072 |     44.213 |     88.616
     System.Buffers.Tests.MemoryTests.TestSpanConstructorString1(length: 1)                | Duration              |     21     |    528.542 |     53.536 |    477.277 |    692.126
     System.Buffers.Tests.MemoryTests.TestSpanConstructorByte(length: 1, untrue: False)    | Duration              |    101     |      0.000 |      0.000 |      0.000 |      0.002
     System.Buffers.Tests.MemoryTests.TestSpanConstructorString(length: 1, untrue: False)  | Duration              |    101     |      0.009 |      0.085 |      0.000 |      0.856

A and C yield very similar results (approximately 50ms for byte, and 540ms for string). However, I would image the "real" results come from B, i.e. essentially 0ms for byte and 0.009 ms for string (for the given iteration count). Given that, aren't we getting inaccurate data when refactor the tests to go through the delegates and Invoke call?

How will we reason about such data? Is it just to highlight the proportional difference between byte/string and compare one run to another as code changes and see improvements? Does the "real" execution time not matter so much?

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
JosephTremoulet: IMO the different overloads are easier to read since they don't need the reflection magic, and they also have the benefit you mentioned in your edit.  I don't mind changing it if you disagree, though.

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
JosephTremoulet: Inside `Invoke`, we [start measuring](https://github.com/JosephTremoulet/coreclr/blob/198fa375dde167a433167fd4575846768879ffce/tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs#L73) just before invoking the delegate that's bound to the [lambda body](https://github.com/JosephTremoulet/coreclr/blob/198fa375dde167a433167fd4575846768879ffce/tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs#L511) that calls `TestSpanDangerousCreate<byte>`, and stop measuring [once that delegate call returns](https://github.com/JosephTremoulet/coreclr/blob/198fa375dde167a433167fd4575846768879ffce/tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs#L75).

JosephTremoulet:
@ahsonkhan, I'm having trouble reproducing your results.  I verified that `TestSpanConstructor<string>` is running into #10031, but when I switched to a non-generic `TestSpanConstructorString`, even though we then get the inlining, it only made a small performance difference.  From comparing the [disassembly](https://gist.github.com/JosephTremoulet/e38a53eb5234b938dfc6d236150019c4) of the byte and nongeneric string versions, it would appear that [this check](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Span.cs#L45) (which the JIT eliminates in the byte case because the `default(T) == null` clause is false there) is probably causing the slowdown for the string version, but I can't figure out how you wouldn't be running into the same.  Can you share the disassembly you're getting for your `TestSpanConstructorString` from case B?

FYI, I also  tried moving the `Sink<T>.Instance` calls out of the kernel methods, and found the difference in runtime to be negligible (in fact, in the generic case with T=String, while getting the address of the field does indeed involve some dictionary lookup stuff, it actually made the running time slightly faster because the first part of that sequence is redundant with part of the type check sequence in the loop, so leaving `Span<T>.Instance` in the kernel method let us remove a load from the loop).  I don't have much of a preference either way, so let me know if you want me to move those out.

AndyAyersMS:
The jit code to optimize away the runtime type tests is fairly fragile, anything other than `typeof(T) == typeof(..sometype..)` is unlikely to be recognized today. You might open an issue and assign to me for this array type-check idiom; perhaps there is some way to clean this up.

JosephTremoulet:
> You might open an issue and assign to me for this array type-check idiom; perhaps there is some way to clean this up

I wouldn't expect us to be able to clean this one up; I think it is the moral equivalent of array store type checks, which is in the `Span<T>` constructor that takes a `T[]` because arrays are covariant (and have the checks at stelems) but Spans are not (and don't have the checks at stores).

Though as you get farther in propagating exact types, we could start applying that knowledge for this and array store checks when the array allocation is in scope.  Do you want an issue for that?

AndyAyersMS:
Maybe I am too pessimistic -- there are patterns in morph for the `typeof(T) == obj.GetType()` checks, we should at least see why they don't kick in here.

In a non-shared context we should be able to know the types well enough.

ahsonkhan:
> but I can't figure out how you wouldn't be running into the same. Can you share the disassembly you're getting for your TestSpanConstructorString from case B?

> A and C both include a delegate call (from Invoke to the lambda) and a direct call (from the lambda to TestSpanConstructor<T>. These are fixed costs (evidently < 50ms) that we should be able to drown out by making the iteration count high enough (e.g. target 1000ms).

I was actually testing/running against slow span which likely doesn't have a lot of the .net core 2.0 changes in the JIT/etc. I think you have addressed the main point of my comparison. I wasn't really showcasing why string or generic is so much larger but in general the test time is not measuring the real time of the API call.

ahsonkhan:
> FYI, I also  tried moving the `Sink<T>.Instance` calls out of the kernel methods, and found the difference in runtime to be negligible ...
... so let me know if you want me to move those out.

No change needed.

ahsonkhan:
> (and have the checks at stelems)

What is stelems?

JosephTremoulet:
> What is stelems?

Store array element operations

ahsonkhan:
Looking at scenario B, I must be doing something incorrect, because this line is wrong/won't compile:
```C#
for (int i = 0; i < InnerIterationCount; i++)
```
It should instead be:
```C#
for (int i = 0; i < Benchmark.InnerIterationCount; i++)
```
Fixing this, I cannot reproduce the weird numbers that I initially reported for B (and it makes sense, possibly I had a local variable InnerIterationCount=1 or something).

I tried to run tests again for two scenarios and here is what I observed:
1a) xunit.perf (No lambda)
System.Buffers.Tests.MemoryTests.TestSpanConstructorByte(length: 1, untrue: False) - 29.768
System.Buffers.Tests.MemoryTests.TestSpanConstructorString(length: 1, untrue: False) - 429.595

1b) xunit.perf (lambda)
System.Buffers.Tests.MemoryTests.TestSpanConstructorByte(length: 1, untrue: False) - 8.690
System.Buffers.Tests.MemoryTests.TestSpanConstructorString(length: 1, untrue: False) - 437.956

1a) shows Span<byte> constructor take ~30ms while 1b) shows approximately 8- 10ms. Span<string> constructor uniformly takes around 400-440ms. Why a 3x reduction in runtime when we use lambdas?

**Regardless, this is not blocking, and we should merge this PR!**

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
```diff
@@ -44,315 +46,394 @@ private sealed class TestClass<T>
...
-        static void TestFillAllArray(byte[] data)
-        {
-            for (int i = 0; i < data.Length; ++i) {
-                data[i] = unchecked((byte)i);
-            }
-        }
+        // Use statics to smuggle some information from Main to Invoke when running tests
+        // from the command line.
+        static bool IsXunitInvocation = true; // xunit-perf leaves this true; command line Main sets to false
+        static int InnerIterationCount = 0;   // used to communicate iteration count from BenchmarkAttribute
```
ahsonkhan: **Suggestion:** Rename this to something other than InnerIterationCount to avoid confusion between this and BenchmarkAttribute.InnerIterationCount. If code evolves, I don't want to accidentally run 0 iterations instead by using InnerIterationCount in the wrong places where BenchmarkAttribute.InnerIterationCount should have been used.

Maybe ExposedInnerIterationCount?

JosephTremoulet: I could call it `CommandLineIterationCount`?

JosephTremoulet:
> Why a 3x reduction in runtime when we use lambdas?

I really don't know.  Do you see that result repeated consistently?  8ms/20ms are really short intervals to try to measure accurately.

JosephTremoulet:
> Regardless, this is not blocking, and we should merge this PR!

Great, but I still don't know if it's ok to add System.Memory to `tests/src/Common/test_dependencies/project.json` like that.  @wtgodbe, care to weigh in?

I also notice that the System.Memory references over in jit/config/benchmark get updated by @dotnet-bot when it updates references, and I don't know where the logic is that controls that to know if it would automatically start updating this reference in tandem or not.  @stephentoub, maybe you know?

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
ahsonkhan: That works.

ahsonkhan:
>  Do you see that result repeated consistently?

I tried a few times. I could try again with larger iterations.

jkotas:
> if it's ok to add System.Memory to `tests/src/Common/test_dependencies/project.json `

Yes, it is ok to add it.

> I don't know where the logic is that controls

The auto update logic goes through all project.jsons and updates all version numbers it can find. It should kick in automatically for System.Memory, without any other special opt-in.

ahsonkhan:
Part of https://github.com/dotnet/corefxlab/issues/1314

Issue #10161, opened 2017-03-14
[x86/Linux] Vector3Interop test failures


parjong:
Reverse P/Invoke test in ``JIT.SIMD.Vector3Interop_r`` failed:
```
FAILED   - [ 5474 :    7 sec ] JIT/SIMD/Vector3Interop_r/Vector3Interop_r.sh
...
               callBack_RPInvoke_Vector3Arg:
                   iVal 123
                   Sum0,1,2 = 1400, 14, 140
                   str abcdefg
               RPInvoke Vector3Arg test failed
```

parjong:
``v1`` and ``v2`` seem to be reversed:
```
    // Argument pass test
    // Test if the managed side correctly reads 12-byte Vector3 argument from the native side
    // and meet the backend assumption 1 and 2.
    static void callBack_RPInvoke_Vector3Arg(
        int i,
        Vector3 v1,
        [MarshalAs(UnmanagedType.LPStr)] string s,
        Vector3 v2)
    {
        // sum = (1, 2, 3) dot (1, 2, 3) = 14
        float sum0 = Vector3.Dot(v1, v1);
        // sum = (10, 20, 30) dot (10, 20, 30) = 1400
        float sum1 = Vector3.Dot(v2, v2);
        // sum = (10, 20, 30) dot (1, 2, 3) = 140
        float sum2 = Vector3.Dot(v2, v1);

        Console.WriteLine("callBack_RPInvoke_Vector3Arg:");
        Console.WriteLine("    iVal {0}", i);
        Console.WriteLine("    Sum0,1,2 = {0}, {1}, {2}", sum0, sum1, sum2);
        Console.WriteLine("    str {0}", s);

        result = (sum0 == 14) && (sum1 == 1400) && (sum2 == 140) && (s == "abcdefg") && (i == 123);
    }
```

parjong:
@jkotas @BruceForstall Could you let me know why x86 pushes arguments from left to right? Is there any problem with R2L order?

jkotas:
> I guessed that windows native compiler uses L2R order. 

This is microoptimization from the early .NET days. The arguments in IL are pushed L2R order, and so this allowed the JIT to generate slightly better code with less work for push/pop architectures like x86. I suspect that the difference is neglible today. However, number of other implementation details picked up dependency on this order, and so changing it would not be exactly trivial task.

It has nothing to do with windows native compilers.

jkotas:
[...]

PR #10217, opened 2017-03-16
[x86/Linux] Reorder stack arguments on reverse P/Invoke

parjong:
x86 native stack layout is different from x86 managed stack layout, which results in #10161.

This commit revises UMThunkStub to reorder stack arguments.

### DIFF
--- a/src/vm/dllimportcallback.cpp
+++ b/src/vm/dllimportcallback.cpp
@@ -1385,8 +1385,11 @@ VOID UMThunkMarshInfo::RunTimeInit()
     MetaSig sig(pMD);
     ArgIterator argit(&sig);
     int numRegistersUsed = 0;
-    m_ecxArgOffset = -1;
-    m_edxArgOffset = -1;
+
+    //
+    // m_cbStackArgSize represents the number of arg bytes for the MANAGED signature
+    //
+    m_cbStackArgSize = 0;
 
     int offs = 0;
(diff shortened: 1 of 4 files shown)


parjong:

parjong:
@jkotas Please take a look.

src/vm/dllimportcallback.cpp
```diff
@@ -1430,6 +1430,76 @@ VOID UMThunkMarshInfo::RunTimeInit()
...
+    // | stkarg 1  |
+    // | stkarg 0  |
+    //
+    // This stub bridges the gap between them.
+    //
+    char *pCurSrc = pSrc;
+    char *pCurDst = pDst + m_cbStackArgSize;
+
+    MetaSig sig(pMD);
+
```
jkotas: This will be super slow to do for every reverse PInvoke.

parjong: @jkotas I tried to revise reverse P/Invoke IL stub as you suggested, but failed.. 

As I understand, we need to change the global argument ordering to avoid this kind of hack, but it requires non-trivial amount of work.

jkotas: Could you please open an issue to track creating a better performing implementation and link it from here?

parjong: #10225 is opened.

jkotas: Thanks!

PR #10224, opened 2017-03-16
Fix Convert.FromBase64CharArray with empty array

stephentoub:
A recent changed caused a CoreFx test failure, verifying that FromBase64CharArray could be used successfully with an empty array.  It started failing with an out of range exception.

cc: @alexperovich, @jkotas


jkotas:
Thanks!

Issue #10056, opened 2017-03-09
Most recent build tools update broke support for building on VS2017


tannergooding:
The following commit broke support for building on VS2017: 8f23fca0ff57c97707d4f6b5068e8044d2dcc13c. Note: The support for building on VS2017 was validated before the commit, but was actually merged just after the aforementioned was merged.

The failure resembles the following:
> D:\coreclr\Tools\Microsoft.CSharp.Core.targets(106,11): error MSB4064: The "OverrideToolHost" parameter is not suppo
rted by the "Csc" task. Verify the parameter exists on the task, and it is a settable public instance property. [D:\
coreclr\src\mscorlib\ref\mscorlib.csproj]

I have confirmed that reverting the commit resolves the issue (but I don't think we should revert it). Instead, we should find out what in the new toolset has broken the VS2017 support and update the toolset again.

tannergooding:
@chcosta, @gkhanna79

tannergooding:
Also FYI. @agocke, @jaredpar from the Roslyn Compiler Team, as they may be able to assist in the diagnosis.

chcosta:
I can't repro this at the moment.  I don't get very far when trying to build in a VS2017 developer command prompt.

```
BUILD: Commencing build of native components for Windows_NT.x64.Release
BUILD: Using environment: "D:\VisualStudio2017\Common7\Tools\\..\..\VC\Auxiliary\Build\vcvarsall.bat" x86_amd64
**********************************************************************
** Visual Studio 2017 Developer Command Prompt v15.0.26228.4
** Copyright (c) 2017 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x86_x64'
BUILD: Regenerating the Visual Studio solution
CMake Error: Could not create named generator Visual Studio 15 2017 Win64

Generators
  Visual Studio 14 2015 [arch] = Generates Visual Studio 2015 project files.
                                 Optional [arch] can be "Win64" or "ARM".
  Visual Studio 12 2013 [arch] = Generates Visual Studio 2013 project files.
                                 Optional [arch] can be "Win64" or "ARM".
  Visual Studio 11 2012 [arch] = Generates Visual Studio 2012 project files.
                                 Optional [arch] can be "Win64" or "ARM".
  Visual Studio 10 2010 [arch] = Generates Visual Studio 2010 project files.
                                 Optional [arch] can be "Win64" or "IA64".
[...]

PR #10229, opened 2017-03-16
Support devirtulization of Encoders

benaadams:
Seal and make private the internal classes `UTF8Encoder` and `UTF8Decoder` 

@AndyAyersMS @jkotas @stephentoub PTAL

### DIFF
--- a/src/mscorlib/src/System/Text/EncoderBestFitFallback.cs
+++ b/src/mscorlib/src/System/Text/EncoderBestFitFallback.cs
@@ -16,7 +16,7 @@ using System.Diagnostics.Contracts;
 namespace System.Text
 {
     [Serializable]
-    internal class InternalEncoderBestFitFallback : EncoderFallback
+    internal sealed class InternalEncoderBestFitFallback : EncoderFallback
     {
         // Our variables
         internal Encoding encoding = null;
--- a/src/mscorlib/src/System/Text/Encoding.cs
+++ b/src/mscorlib/src/System/Text/Encoding.cs
@@ -1430,7 +1430,7 @@ namespace System.Text
         }
 
         [Serializable]
-        internal class DefaultEncoder : Encoder, IObjectReference, ISerializable
+        internal sealed class DefaultEncoder : Encoder, IObjectReference, ISerializable
         {
             private Encoding m_encoding;
             [NonSerialized] private bool m_hasInitializedEncoding;
@@ -1552,7 +1552,7 @@ namespace System.Text
         }
 
         [Serializable]
-        internal class DefaultDecoder : Decoder, IObjectReference, ISerializable
+        internal sealed class DefaultDecoder : Decoder, IObjectReference, ISerializable
         {
             private Encoding m_encoding;
             [NonSerialized]
--- a/src/mscorlib/src/System/Text/UTF32Encoding.cs
+++ b/src/mscorlib/src/System/Text/UTF32Encoding.cs
@@ -998,7 +998,7 @@ namespace System.Text
         }
 
         [Serializable]
-        internal class UTF32Decoder : DecoderNLS
+        private sealed class UTF32Decoder : DecoderNLS
         {
             // Need a place to store any extra bytes we may have picked up
             internal int iChar = 0;
--- a/src/mscorlib/src/System/Text/UTF7Encoding.cs
+++ b/src/mscorlib/src/System/Text/UTF7Encoding.cs
@@ -628,7 +628,7 @@ namespace System.Text
         [Serializable]
         // Of all the amazing things... This MUST be Decoder so that our com name
         // for System.Text.Decoder doesn't change
-        private class Decoder : DecoderNLS, ISerializable
+        private sealed class Decoder : DecoderNLS, ISerializable
         {
             /*private*/
             internal int bits;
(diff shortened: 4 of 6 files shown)


src/mscorlib/src/System/Text/ASCIIEncoding.cs
```diff
@@ -763,13 +763,46 @@ public override bool IsSingleByte
...
-            return new DecoderNLS(this);
+            return new ASCIIDecoder(this);
         }
 
-
         public override Encoder GetEncoder()
         {
-            return new EncoderNLS(this);
+            return new ASCIIEncoder(this);
+        }
```
stephentoub: So now Encoding.ASCII.GetEncoder().AnyVirtual() ends up getting devirtualized all the way through?

benaadams:
Example [pattern in current Kestrel](https://github.com/aspnet/KestrelHttpServer/blob/feature/dev-si/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/MemoryPoolIterator.cs#L1254-L1316) that am targeting
```csharp
var decoder = Encoding.UTF8.GetDecoder();
// ...
decoder.Convert(...);
```

src/mscorlib/src/System/Text/ASCIIEncoding.cs
benaadams: Checking; I don't think so yet. The GetEncoder call will be devirtualized; but then I think it needs to inline to then devirtualize at the next step; but I don't think inilining post devirtualization is there yet?

benaadams:
Can't build coreclr after installing latest VS2017 updated to test https://github.com/dotnet/coreclr/issues/10056 😢

benaadams:
@stephentoub looks like not yet, following
```csharp
{
    var bytes = new byte[] { 0x20, 0x20, 0x20, 0x20 };
    var chars = new char[4];

    var decoder = System.Text.Encoding.ASCII.GetDecoder();

    var converted = decoder.GetChars(bytes, 0, bytes.Length, chars, 0);

    Console.WriteLine($"Converted {converted} bytes");
}

```
Devirtualizes as so
```
Devirtualized virtual call to Encoding:GetDecoder; now direct call to ASCIIEncoding:GetDecoder [final class]
Devirtualized virtual call to Encoding:get_CodePage; now direct call to Encoding:get_CodePage [final class]
Devirtualized virtual call to Encoding:get_CodePage; now direct call to Encoding:get_CodePage [final class]
Devirtualized virtual call to Encoding:SetDefaultFallbacks; now direct call to UTF8Encoding:SetDefaultFallbacks [exact]
Devirtualized virtual call to StreamWriter:set_AutoFlush; now direct call to StreamWriter:set_AutoFlush [exact]
Devirtualized virtual call to Encoder:Reset; now direct call to OSEncoder:Reset [exact]
Devirtualized virtual call to Encoder:GetBytes; now direct call to OSEncoder:GetBytes [final class]
Devirtualized virtual call to Stream:get_CanWrite; now direct call to ConsoleStream:get_CanWrite [final class]
```
With inlines
```
Successfully inlined Encoding:get_ASCII():ref (6 IL bytes) (depth 1) [below ALWAYS_INLINE size]
Devirtualized virtual call to Encoding:GetDecoder; now direct call to ASCIIEncoding:GetDecoder [final class]
Successfully inlined String:Format(ref,ref):ref (14 IL bytes) (depth 1) [below ALWAYS_INLINE size]
Successfully inlined ParamsArray:.ctor(ref):this (33 IL bytes) (depth 2) [profitable inline]
**************** Inline Tree
Inlines into 06000001 Program:Main(ref)
  [1 IL=0025 TR=000033 0600317B] [below ALWAYS_INLINE size] Encoding:get_ASCII():ref
  [0 IL=0042 TR=000044 060030A2] [FAILED: target not direct] Decoder:GetChars(ref,int,int,ref,int):int:this
  [0 IL=0030 TR=000036 06003198] [FAILED: target not direct] Encoding:GetDecoder():ref:this
  [2 IL=0059 TR=000069 060003F6] [below ALWAYS_INLINE size] String:Format(ref,ref):ref
    [3 IL=0003 TR=000099 06000AE6] [profitable inline] ParamsArray:.ctor(ref):this
  [0 IL=0064 TR=000074 06000055] [FAILED: noinline per IL/cached result] Console:WriteLine(ref)
Budget: initialTime=270, finalTime=334, initialBudget=2700, currentBudget=2700
Budget: initialSize=1727, finalSize=1996
```
Thought hopefully in future

stephentoub:
> looks like not yet

Thanks.  In that case does adding the new derived types add any value at this point?

src/mscorlib/src/System/Text/ASCIIEncoding.cs
AndyAyersMS: When if type information needs to propagate from callee to caller (eg a method returning an exact or sealed subtype of its declared return type) then it can lead to devirtualization but not inlining.

I would very much like to fix this, but it is lower priority than fixing cases where the jit loses types or doesn't get the best type information available.

danmosemsft:
@dotnet-bot test OSX10.12 x64 Checked Build and Test please (Sierra update)

benaadams:
> does adding the new derived types add any value at this point?

I could just seal the internal Utf8 coders? Doesn't introduce new types and then would be ready if/when devirtualization supports it.

Would also changing them from internal to private be an issue?

jkotas:
> I could just seal the internal Utf8 coders?

Just sealing Utf8 encoders sounds fine.

>Would also changing them from internal to private be an issue?

Sounds fine too.

benaadams:
Sealed and made private some other encoders and fallbacks; many were already both sealed and private

benaadams:
OSX Tests have started passing again 😄

PR #10231, opened 2017-03-16
Fixing more file-class correspondence violations.

AtsushiKan:
This is partly to pave the way for sharing more
*Info.cs files, and partly because this has
annoyed me for years.

This commit includes a couple of opportunistic
*.CoreCLR.cs splits but I stopped after the easy
ones as it was getting error prone to do them
all at once.


AtsushiKan:
cc @jkotas

jkotas:
Build error: '/mnt/j/workspace/dotnet_coreclr/master/arm_emulator_cross_debug_ubuntu_prtest/src/mscorlib/src/System/Reflection/FieldInfo.CoreClr.cs' could not be found. 

Casing?

jkotas:
LGTM otherwise.

PR #10236, opened 2017-03-16
Move the MemberInfo-derived Info classes to the shared partition.

AtsushiKan:
Note: The only wrinkle is the fact that the CoreClr "=="
overrides "know" that the Runtime implementations of these
types inherit their Equals() method from System.Object.
This is not something we can put in the shared space.

The operator will still work as before without this special
checking and it is unlikely we're making much of use it
as "==" has no reliable semantics on the CoreClr versions
of these types.


AtsushiKan:
cc @jkotas 
cc @alexperovich 

(I'll let the bot mirror them over to corert.)

AtsushiKan:

PR #10238, opened 2017-03-16
Reduce case iteration counts in SeekUnroll.

pgavlin:
This test was taking too long to run on x86. After investigation, it
does not appear that there is a fundamental problem with the code
generated for the test; rather, it seems that we are running an
excessive number of iterations when using this test as a correctness
test. This change reduces the number of times we run each case from 10
to 1 unless otherwise specified on the command line.

[Here is an example of a job where this test has hit the timeout.](https://ci.dot.net/job/dotnet_coreclr/job/master/job/x86_release_windows_nt/967/)

### DIFF
--- a/tests/src/JIT/Performance/CodeQuality/SIMD/SeekUnroll/SeekUnroll.cs
+++ b/tests/src/JIT/Performance/CodeQuality/SIMD/SeekUnroll/SeekUnroll.cs
@@ -151,7 +151,17 @@ public static class SeekUnroll
             InnerIterations = 100000;
         }
 
-        int manualLoopCount = (args != null && args.Length >= 1) ? int.Parse(args[0]) : 1;
+        int manualLoopCount = 1;
+        if (args == null || args.Length == 0)
+        {
+            Console.WriteLine("Warning: no iteration count specified; defaulting to 1 iteration per case");
+            Console.WriteLine("To use multiple iterations per case, pass the desired number of iterations as the first command-line argument to this test");
+        }
+        else
+        {
+            manualLoopCount = int.Parse(args[0]);
+        }
+
         foreach(int index in IndicesToTest)
         {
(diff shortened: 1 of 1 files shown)


pgavlin:
@JosephTremoulet @AndyAyersMS @dotnet/jit-contrib PTAL

pgavlin:
cc @RussKeldorph

pgavlin:
> Seems more sensible to me to reduce InnerIterations for x86 rather than do this...

What value do you see in running all 5 cases for 10 iterations each during a normal test run? Note that this change does not affect the iteration count when running as a perf test.

> Also if x64 does not time out it is probably worth opening an issue to better understand why x86 does.

I've already looked into this. The root cause appears to be due to the fact that we retire about 5-6x as many instructions on x86 due to the use of longs in `FindByte`. The primary differences are the decomposed index into a `Vector<long>` and the use of a helper call to multiply two longs. The former is especially impactful, as it accounts for 20 extra instructions per call to `FindByte` (or ~1 trillion extra instructions over the entire test).

AndyAyersMS:
In the past we've tried to keep these tests somewhat viable as benchmarks when run as exes. I suppose getting back just one timed result by default might be suggestive enough to remember that this test needs extra inputs to trigger benchmark mode, and one can go look at the sources and see how to make that happen. So unless @JosephTremoulet chimes in I'll leave it up to you to decide what to do.

JosephTremoulet:
Or the test could just print out something like "command-line argument IterationCount not supplied; defaulting to 1 iteration", that ought to be "suggestive enough to remember that this test needs extra inputs to trigger benchmark mode".

pgavlin:
> Or the test could just print out something like "command-line argument IterationCount not supplied; defaulting to 1 iteration"

Good suggestion. I will make that change.

PR #10239, opened 2017-03-16
Clean up GenTreePutArgStk

BruceForstall:
Consolidate constructors; go from 4 to 2.

Reduce `#ifdef` in callers.

### DIFF
--- a/src/jit/gentree.h
+++ b/src/jit/gentree.h
@@ -4592,34 +4592,34 @@ struct GenTreePutArgStk : public GenTreeUnOp
     // Don't let clang-format mess with the GenTreePutArgStk constructor.
     // clang-format off
 
-    GenTreePutArgStk(genTreeOps oper,
-                     var_types  type,
-                     GenTreePtr op1,
-                     unsigned   slotNum
-                     PUT_STRUCT_ARG_STK_ONLY_ARG(unsigned numSlots)
-                     , bool putInIncomingArgArea = false
-                     , GenTreeCall* callNode = nullptr)
+    GenTreePutArgStk(genTreeOps   oper,
+                     var_types    type,
+                     GenTreePtr   op1,
+                     unsigned     slotNum
+                     PUT_STRUCT_ARG_STK_ONLY_ARG(unsigned numSlots),
+                     bool         putInIncomingArgArea = false,
+                     GenTreeCall* callNode = nullptr)
         : GenTreeUnOp(oper, type, op1 DEBUGARG(/*largeNode*/ false))
         , gtSlotNum(slotNum)
 #if defined(UNIX_X86_ABI)
         , gtPadAlign(0)
 #endif
+#if FEATURE_FASTTAILCALL
+        , gtPutInIncomingArgArea(putInIncomingArgArea)
+#endif // FEATURE_FASTTAILCALL
 #ifdef FEATURE_PUT_STRUCT_ARG_STK
         , gtPutArgStkKind(Kind::Invalid)
         , gtNumSlots(numSlots)
         , gtNumberReferenceSlots(0)
         , gtGcPtrs(nullptr)
 #endif // FEATURE_PUT_STRUCT_ARG_STK
-    {
-#if FEATURE_FASTTAILCALL
-        gtPutInIncomingArgArea = putInIncomingArgArea;
-#endif // FEATURE_FASTTAILCALL
 #ifdef DEBUG
-        gtCall = callNode;
+        , gtCall(callNode)
 #endif
+    {
     }
 
-    // clang-format on
+// clang-format on
 
 #if FEATURE_FASTTAILCALL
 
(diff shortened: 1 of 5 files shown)


BruceForstall:
Wow. clang-format is complaining about where `// clang-format on` is put. Great.

BruceForstall:
@pgavlin PTAL
cc @dotnet/jit-contrib

src/jit/gentree.h
```diff
@@ -4589,105 +4589,58 @@ struct GenTreePutArgStk : public GenTreeUnOp
...
-                     unsigned slotNum PUT_STRUCT_ARG_STK_ONLY_ARG(unsigned numSlots)
-                         PUT_STRUCT_ARG_STK_ONLY_ARG(bool isStruct),
-                     bool _putInIncomingArgArea = false DEBUGARG(GenTreeCall* callNode = nullptr)
-                         DEBUGARG(bool largeNode = false))
-        : GenTreeUnOp(oper, type DEBUGARG(largeNode))
+                     GenTreePtr op1,
+                     unsigned   slotNum
+                     PUT_STRUCT_ARG_STK_ONLY_ARG(unsigned numSlots)
+                     , bool putInIncomingArgArea = false
+                     , GenTreeCall* callNode = nullptr)
```
JosephTremoulet: Why do these two lines start with commas instead of having the commas at the end of the preceding lines?  Looks weird this way.

BruceForstall: I can change it; doesn't matter. Originally, it was this way because there were `#ifdef` around the lines.

src/jit/gentree.h
```diff
@@ -4589,105 +4589,58 @@ struct GenTreePutArgStk : public GenTreeUnOp
...
-#if defined(UNIX_X86_ABI)
-        , gtPadAlign(0)
-#endif
-#ifdef FEATURE_PUT_STRUCT_ARG_STK
-        , gtPutArgStkKind(Kind::Invalid)
-        , gtNumSlots(numSlots)
-        , gtNumberReferenceSlots(0)
-        , gtGcPtrs(nullptr)
-#endif // FEATURE_PUT_STRUCT_ARG_STK
+    static bool putInIncomingArgArea()
```
pgavlin: I would make this a constexpr member function instead of a static:

```
constexpr bool putInIncomingArgArea()
{
    return false;
}
```

BruceForstall: @pgavlin We don't have any use of `constexpr` in our source base, possibly because of compiler issues. I don't want to be the first one with this minor change.

BruceForstall:
```
09:16:25   cmdLine:D:\j\workspace\debug_windows---17180f1b\bin\tests\Windows_NT.x64.Debug\GC\Regressions\v2.0-rtm\544701\544701\544701.cmd Timed Out
```

src/jit/gentree.h
pgavlin: We have had some compiler issues in the past when building for Desktop, but IIRC those have been worked around (see e.g. the comments in debugreturn.h). If we can't make this a `constexpr`, let's at least make it a `const` member function.

BruceForstall:
CentOS failure:
```
.............Expected thread to sleep for 2000 ms (and not be interrupted).
Average delta: 858993403 ms,  acceptable delta: 150


FAILED: threading/SleepEx/test2/paltest_sleepex_test2. Exit code: 1
```

src/jit/gentree.h
pgavlin: FWIW, we do have a use of `constexpr`, but only one--see gdbjit.h.

BruceForstall:
Another sleep failure?!? OSX:
```
The sleep function slept for 651 ms when it should have slept for 500 ms


FAILED: threading/SleepEx/test1/paltest_sleepex_test1. Exit code: 1
```

pgavlin:
@BruceForstall those sleep tests look like bad tests... AFAIK the only guarantee on any OS is that `sleep(x)` will sleep for at least `x` units of time (never fewer). I've filed #10261 to track fixing this.

src/jit/lower.cpp
```diff
@@ -940,18 +940,11 @@ GenTreePtr Lowering::NewPutArg(GenTreeCall* call, GenTreePtr arg, fgArgTabEntryP
...
-#if FEATURE_FASTTAILCALL
         putArg = new (comp, GT_PUTARG_STK)
             GenTreePutArgStk(GT_PUTARG_STK, type, arg, info->slotNum PUT_STRUCT_ARG_STK_ONLY_ARG(info->numSlots),
-                             call->IsFastTailCall() DEBUGARG(call));
-#else
-        putArg = new (comp, GT_PUTARG_STK)
-            GenTreePutArgStk(GT_PUTARG_STK, type, arg,
-                             info->slotNum PUT_STRUCT_ARG_STK_ONLY_ARG(info->numSlots) DEBUGARG(call));
-#endif
+                             call->IsFastTailCall(), call);
```
briansull: Thanks for getting rid of these ifdefs

briansull:
LGTM

PR #10240, opened 2017-03-16
Fix test failure detection when dumps are enabled.

pgavlin:


pgavlin:
@RussKeldorph PTAL

tests/runtest.cmd
```diff
@@ -233,12 +233,13 @@ del %CORE_ROOT%\mscorlib.ni.dll
...
 set __BuildLogRootName=TestRunResults
 call :msbuild "%__ProjectFilesDir%\runtest.proj" /p:Runtests=true /clp:showcommandline
+set __errorlevel=%errorlevel%
 
 if "%__CollectDumps%"=="true" (
     python "%__DumplingHelperPath%" collect_dump %errorlevel% "%__CrashDumpFolder%" %__StartTime% "CoreCLR_Tests"
 )
 
-if errorlevel 1 (
+if %__errorlevel%==1 (
```
RussKeldorph: Run "help if" at cmd prompt.  'if errorlevel' is a >= comparison.

RussKeldorph: (...assuming it's possible for msbuild to return something greater than 1)

pgavlin: fixed

PR #10242, opened 2017-03-16
Add reflection roots for tests calling CreateInstance()

swaroop-sridhar:
Add the XML files identifying the reflection roots to keep the
type (and constructors threof) which are used by tests as
Activator.CreateInstance(typeof(T))

This will help these tests pass while running under ILLINIK


swaroop-sridhar:
@erozenfeld please review

swaroop-sridhar:
@dotnet-bot help

swaroop-sridhar:
@dotnet-bot test illink Windows_NT Checked illink

erozenfeld:
Do we need to keep all types in the assemblies? Can't we just keep whatever are the arguments to Activator.CreateInstance(typeof(T)) ?

swaroop-sridhar:
Yes we only need the constructors of the types called by typeof() in this case.
But since there are many tests needing reflection roots, I've been adding assembly.* for tests failing because of artifacts necessary from the test itself. However, I've have added the precise members necessary in the case of libraries.

I believe this is sufficient, because:
1) It checks that the test passes if we keep the app intact and link the libraries (except corelib)
2) Eventual goal should be to remove the .xml file

swaroop-sridhar:
@erozenfeld, looking at the tests in question, writing out the types to include in the XML files would be  no different from writing "*". The tests define a few classes, and instantiate various template expansions of them.

Issue #10048, opened 2017-03-09
corefx test failure with jitstressregs: assert in CoreCLR from dotnet.exe


BruceForstall:
```
  Assert failure(PID 3280 [0x00000cd0], Thread: 5948 [0x173c]): (ptrAssemblyArray != NULL) && (cbAssemblyArrayLength > 0)
15:40:10   
15:40:10   CORECLR! AssemblyNative::LoadFromStream + 0x14E (0x00007ff9`56ef570e)
15:40:10   SYSTEM.PRIVATE.CORELIB.NI! <no symbol> + 0x0 (0x00007ff9`55ff3ff8)
15:40:10   SYSTEM.PRIVATE.CORELIB.NI! <no symbol> + 0x0 (0x00007ff9`55fab927)
15:40:10   <no module>! <no symbol> + 0x0 (0x00007ff8`f8879fb4)
15:40:10   <no module>! <no symbol> + 0x0 (0x0000004e`cb24a638)
15:40:10   <no module>! <no symbol> + 0x0 (0x0000004e`ca8a6ff8)
15:40:10   <no module>! <no symbol> + 0x0 (0x0000004e`e484bd40)
15:40:10   <no module>! <no symbol> + 0x0 (0x00000000`00000001)
15:40:10       File: d:\j\workspace\x64_checked_w---463bcea0\src\vm\assemblynative.cpp Line: 660
15:40:10       Image: D:\j\workspace\x64_checked_w---463bcea0\_\fx\bin\testhost\netcoreapp-Windows_NT-Release-x64\dotnet.exe
15:40:10   
15:40:10   Finished running tests.  End time=15:40:10.90, Exit code = 123456789
```

For instance:
```
@dotnet-bot test Windows_NT x64 corefx_jitstressregs0x1000
@dotnet-bot test Windows_NT x86 corefx_jitstressregs0x1000
```

BruceForstall:
@RussKeldorph : @sivarv said you'd seen this before. Did anyone verify it works without JitStressRegs?

gkhanna79:
@RussKeldorph I am labeling this Codegen for now. Please update it if you think this is not codegen related.

RussKeldorph:
[...]

Issue #8160, opened 2016-11-17
RyuJIT/x86: CoreFx System.Runtime.Tests Exit code = -2146233022


RussKeldorph:
coreclr *(assuming root=c:\git\coreclr)*
```
build x86 checked skiptests
```
corefx *(assuming root=c:\git\corefx)*
```
build -Release -SkipTests -TestNugetRuntimeId=win7-x86 -- /p:BUILDTOOLS_OVERRIDE_RUNTIME=c:\git\coreclr\bin\Product\Windows_NT.x86.Checked
cd bin\tests\AnyOS.AnyCPU.Release\System.Runtime.Tests\netcoreapp1.1.netcoreapp1.1
set COMPlus_AltJit=*
set COMPlus_AltJitName=protojit.dll
RunTests.cmd c:\git\corefx\packages
```
**Output**
```
Using c:\git\corefx\packages as folder for resolving package dependencies.
Executing in C:\git\corefx\bin\tests\AnyOS.AnyCPU.Release\System.Runtime.Tests\default.netcoreapp1.1\
Hard linking dependent files...
Finished linking needed files, moving to running tests.
Running tests... Start time: 16:50:22.53
Command(s):
call CoreRun.exe xunit.console.netcore.exe System.Runtime.Tests.dll  -xml testResults.xml -notrait Benchmark=true -notrait category=nonnetcoreapp1.1tests  -notrait category=OuterLoop -notrait category=failing -notrait category=nonwindowstests
xUnit.net console test runner (32-bit .NET Core)
Copyright (C) 2014 Outercurve Foundation.

Discovering: System.Runtime.Tests
Discovered:  System.Runtime.Tests
Starting:    System.Runtime.Tests
   System.Tests.IntPtrTests.ImplicitCast [SKIP]
      Condition(s) not met: \"Is64Bit\"
   System.Tests.IntPtrTests.GetHashCodeRespectAllBits [SKIP]
      Condition(s) not met: \"Is64Bit\"
   System.Tests.IntPtrTests.Ctor_VoidPointer_ToPointer [SKIP]
      Condition(s) not met: \"Is64Bit\"
   System.Tests.IntPtrTests.Subtract [SKIP]
      Condition(s) not met: \"Is64Bit\"
   System.Tests.IntPtrTests.Ctor_Long [SKIP]
      Condition(s) not met: \"Is64Bit\"
   System.Tests.IntPtrTests.Size [SKIP]
[...]

PR #10245, opened 2017-03-17
LoadFromStream should throw badimageformat early on

rahku:
Otherwise it violates the invariant at https://github.com/dotnet/coreclr/blob/master/src/vm/assemblynative.cpp#L660

Fixes issue #10048 and issue #8160

### DIFF
--- a/src/mscorlib/src/System.Private.CoreLib.txt
+++ b/src/mscorlib/src/System.Private.CoreLib.txt
@@ -854,6 +854,7 @@ BadImageFormat_StreamPositionInvalid = Corrupt .resources file.  The specified p
 BadImageFormat_ResourcesDataInvalidOffset = Corrupt .resources file. Invalid offset '{0}' into data section.
 BadImageFormat_NegativeStringLength = Corrupt .resources file. String length must be non-negative.
 BadImageFormat_ParameterSignatureMismatch = The parameters and the signature of the method don't match.
+BadImageFormat_BadILFormat = Bad IL format.
 
 ; Cryptography
 ; These strings still appear in bcl.small but should go away eventually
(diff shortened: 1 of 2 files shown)


rahku:
PTAL @gkhanna79

src/mscorlib/src/System.Private.CoreLib.txt
```diff
@@ -854,6 +854,7 @@ BadImageFormat_StreamPositionInvalid = Corrupt .resources file.  The specified p
 BadImageFormat_ResourcesDataInvalidOffset = Corrupt .resources file. Invalid offset '{0}' into data section.
 BadImageFormat_NegativeStringLength = Corrupt .resources file. String length must be non-negative.
 BadImageFormat_ParameterSignatureMismatch = The parameters and the signature of the method don't match.
+BadImageFormat_BadILFormat = Bad IL format.
```
gkhanna79: Why are we adding this instead of using https://github.com/dotnet/coreclr/pull/10245/files#diff-82c8be167dc0db14d2000fe3135b426dL171?

src/mscorlib/src/System/Runtime/Loader/AssemblyLoadContext.cs
```diff
@@ -137,6 +137,11 @@ public Assembly LoadFromStream(Stream assembly, Stream assemblySymbols)
                 throw new ArgumentNullException(nameof(assembly));
             }
 
+            if(assembly.Length <= 0)
```
justinvp: Nit: space

src/mscorlib/src/System.Private.CoreLib.txt
rahku: Are you referring to "Arg_BadImageFormatException = Format of the executable (.exe) or library (.dll) is invalid. " I was matching the message that we get for .net core 1.1. Also above message seems confusing that we are loading a dll but in this case it is not. This is a stream.

rahku:
All checks passed previously and I only did whitespace change

PR #10246, opened 2017-03-17
[Linux] 'root' as a special case in UTIL_IsExecuteBitsSet

parjong:
The current PAL implementation checks whether execution bit is set for the current executable, or not.

Unfortunately, it does not consider when the current user is root, which results in coreclr_initialize failure with status 0x8007001f.

This commit revises UTIL_IsExecuteBitsSet to consider root (euid == 0) as a special case.

### DIFF
--- a/src/pal/src/misc/utils.cpp
+++ b/src/pal/src/misc/utils.cpp
@@ -124,7 +124,12 @@ BOOL UTIL_IsExecuteBitsSet( struct stat * stat_data )
     }
     
     /* Check for read permissions. */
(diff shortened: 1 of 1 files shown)


parjong:
We may revise INIT_FindEXEPath not to check execution bit, but I'm not sure which approach is better.

parjong:
@jkotas  Could you please take a look?

jkotas:
Looks good. Though, I do not think that this whole method is needed in CoreCLR - it may be nice to look into deleting it and its callers at some point.

PR #10248, opened 2017-03-17
Fix ValueNumStore::IsVNArrLenUnsignedBound checks

mikedn:
`VNFuncApp::m_func` should be checked first to be sure that `m_args[0]` and `m_args[1]` contain valid VNs. It's not clear if `m_func` can ever be something other than a binary func in the case of a relop VN but the code doesn't look right as it is.


mikedn:
cc @JosephTremoulet

PR #10259, opened 2017-03-17
JIT: Fix noway_assert when inlining

### LINKED ISSUE #9891: JIT: figure out how a noway_assert in the importer is firing
There are some as of yet unknown cases where [this](https://github.com/dotnet/coreclr/blob/7b8f9d6c8f5b196b8cf7da6f4c49bf4149f3e15a/src/jit/importer.cpp#L9934) `noway_assert` can fire.

The assert is there to try and ensure that if the jit is inlining a method that uses `starg` on a particular argument the jit has also decided to use a new local to represent that argument in the inlined body (rather than propagating the caller's argument expression directly into the inlinee's body). This should be guaranteed by the analysis of the inlinee done during `fgFindJumpTargets` (which notes `starg` instructions in the inline arg info) and the logic in `impInlineFetchArg`.

So far I have been unable to create a repro case. Some hypotheses below.

- There is an asymmetry in `fgFindJumpTargets` between `starg` and `ldarga` for inlinees. Abstractly both should have the same effect on how the jit propagates the arg value. However the `starg` case calls into `compMapILargNum` for inlinees, which seemingly could cause `starg` impact to be misplaced for methods with hidden arguments. The invariant in the rest of the code is that inlinee arg numbers refer to IL arguments and do not take hidden arguments into account. So far it looks like this asymmetry is harmless because we bail out of `lvaInitTypeRef` early for inlinees so the arg number is not adjusted by the call to `compMapILargNum`.
- We are seeing IL with bad `starg` numbers. We need to tread cautiously with invalid IL when evaluating an inline, because we are not sure the particular method will actually ever be called. So we tend to defer failure and perhaps this `noway_assert` is the first indication of the actual problem. Note the code in `fgFindJumpTargets` will still record properties for nonexistent args since the local table size is at least 16 entries. So for small invalid `starg` values the arg info will still record the right `starg` state (but other aspects of the var will be bogus), while for large invalid `starg` values the `starg` observation is suppressed. IL cases created with bad `starg` numbers trigger a raft of asserts, but so far I have not been able to hit this particular `noway_assert`.

Inlining for `starg` inlinees was enabled in #6572.

AndyAyersMS:
The jit keeps track of which inline arguments can be modified via starg
during the inlinee IL prescan to ensure that a temp is used to represent
such arguments in the inlined body. There is a noway_assert the a starg
is procecssed during importation that double-checks that a temp is really
being used, and this assert was firing.

The prescan guraded the starg tracking with a check to avoid an out-of-bounds
write into the argument table when scanning bad IL with bogus argument numbers.
The predicate was checking the argument number against lvaTableCount.
Unfortunately for inlinees this value is not related to the number of arguments.
In particular it some rare cases it may be an underestimate so the prescan might
end up failing to note a starg. If the the jit then tries to inline and the
caller passes a constant, the assert will then fire.

Fix is to update the guard to use the number of callee arguments. This is
known to be less than the table size by earlier checks.

Added a test case.

Closes #9891.

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -4546,7 +4546,7 @@ void Compiler::fgFindJumpTargets(const BYTE* codeAddr, IL_OFFSET codeSize, BYTE*
                 else
                 {
                     // account for possible hidden param
-                    varNum = compMapILargNum(varNum); 
+                    varNum = compMapILargNum(varNum);
 
                     // This check is only intended to prevent an AV.  Bad varNum values will later
                     // be handled properly by the verifier.
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -4535,17 +4535,22 @@ void Compiler::fgFindJumpTargets(const BYTE* codeAddr, IL_OFFSET codeSize, BYTE*
                 }
 
                 varNum = (sz == sizeof(BYTE)) ? getU1LittleEndian(codeAddr) : getU2LittleEndian(codeAddr);
-                varNum = compMapILargNum(varNum); // account for possible hidden param
 
-                // This check is only intended to prevent an AV.  Bad varNum values will later
-                // be handled properly by the verifier.
-                if (varNum < lvaTableCnt)
+                if (isInlining)
                 {
-                    if (isInlining)
+                    if (varNum < impInlineInfo->argCnt)
                     {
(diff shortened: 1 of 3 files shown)


AndyAyersMS:
@JosephTremoulet PTAL
cc @dotnet/jit-contrib

briansull:
LGTM

AndyAyersMS:
No diffs (as expected) in the framework dasm set.

PR #10262, opened 2017-03-17
Updates for the handling of Profile data in the JIT

briansull:
Use the type weight_t instead of unsigned when dealing with BasicBlock weights.
    Added new method getCalledCount to return the number of times that the method was called.
    Added new method fgProfileRunsCount to return the number of scenario runs for the profile data.
    Made the method fgIsUsingProfileWeight public instead of protected
    Renamed fgCalledWeight to fgCalledCount
    Updated the logic in getBBWeight and added comments explaining how it works.

### DIFF
--- a/src/jit/compiler.h
+++ b/src/jit/compiler.h
@@ -3552,7 +3552,7 @@ public:
     bool                 fgSlopUsedInEdgeWeights;  // true if their was some slop used when computing the edge weights
     bool                 fgRangeUsedInEdgeWeights; // true if some of the edgeWeight are expressed in Min..Max form
     bool                 fgNeedsUpdateFlowGraph;   // true if we need to run fgUpdateFlowGraph
-    BasicBlock::weight_t fgCalledWeight;           // count of the number of times this method was called
+    BasicBlock::weight_t fgCalledCount;            // count of the number of times this method was called
                                                    // This is derived from the profile data
                                                    // or is BB_UNITY_WEIGHT when we don't have profile data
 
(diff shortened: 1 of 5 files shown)


briansull:
@dotnet/jit-contrib PTAL

briansull:
@dpodder /cc

src/jit/compiler.h
```diff
@@ -4528,12 +4528,22 @@ class Compiler
...
+    //                           or if we have some fake profile data for the stress mode
     bool fgIsUsingProfileWeights()
     {
         return (fgHaveProfileData() || fgStressBBProf());
     }
-    void fgInstrumentMethod();
+
+    // fgProfileRunsCount - returns total number of scenario runs for the profile data
+    //                      or BB_UNITY_WEIGHT when we aren't using profile data.
+    unsigned fgProfileRunsCount()
```
dpodder: Just making sure: this is being intentionally made public as well, right?

briansull: Yes the other new method that I added  is public as well:  fgProfileRunsCount

src/jit/lclvars.cpp
```diff
@@ -2394,31 +2394,92 @@ unsigned Compiler::lvaLclExactSize(unsigned varNum)
...
+        //
+        // 1. For methods that do not have IBC data the called weight will is always 100 (BB_UNITY_WEIGHT)
+        //     and the entry point bbWeight value is almost always 100 (BB_UNITY_WEIGHT)
+        // 2.  For methods that do have IBC data the called weight is the actual number of calls
+        //     from the IBC data and the entry point bbWeight value is almost always the actual
+        //     number of calls from the IBC data.
+        //
+        // "almost always" - except for the rare case where a loop backedge jumps to BB01
+        //
+        // We also perform a rouinding operation by adding half of the 'calledCount' before performing
```
BruceForstall: >rouinding [](start = 29, length = 9)

typo: rouinding

BruceForstall:
:shipit:

PR #10264, opened 2017-03-17
Invariant globalization

tarekgh:
This change is to introduce the config switch to control using the OS globalization or just run in the invariant mode.


tarekgh:
CC @jkotas @stephentoub @ellismg @danmosemsft @weshaggard @Petermarcu @janvorli

src/mscorlib/src/System/Text/Normalization.Windows.cs
```diff
@@ -49,6 +55,12 @@ internal static bool IsNormalized(String strInput, NormalizationForm normForm)
 
         internal static String Normalize(String strInput, NormalizationForm normForm)
         {
+            if (CultureData.InvariantMode)
+            {
+                // work ordinal, then all charcaters are normalized in this mode
```
danmosemsft: typo "charcaters " in a few places

tarekgh: thanks Dan :-)

src/mscorlib/src/System/CLRConfig.cs
```diff
@@ -9,6 +9,30 @@
...
 namespace System
 {
+    // CLRConfig is mainly reading the config switch values. this is used when we cannot use the AppContext class 
+    // one example, is using the context switch in the globalization code which require to read the switch very 
+    // early even before the appdomain get initialized.
+    // In general AppContext should be used instead of CLRConfig if there is no reason prevent that.
+    internal class CLRConfig
+    {
+
+#if CORECLR
```
jkotas: This should not be needed - CORECLR is always defined.

src/corefx/System.Globalization.Native/icushim.cpp
```diff
@@ -170,6 +170,13 @@ bool FindLibWithMajorMinorSubVersion(int* majorVer, int* minorVer, int* subVer)
     return false;
 }
 
+static int32_t g_icuPresent = 1;
```
stephentoub: It may not actually matter, but did you consider reversing this? i.e. starting it at 0 == not present, and then setting it to 1 only once ICU has been detected and all of the function pointers appropriately initialized?

tarekgh:
@dotnet-bot help

src/mscorlib/src/System/CLRConfig.cs
tarekgh: I have the #else case to avoid compiling the code for non CORECLR. I may be missing something here.

src/mscorlib/src/System/Text/Normalization.Windows.cs
jkotas: I do not understand what is `work ordinal` trying to say.

src/mscorlib/src/Interop/Unix/System.Globalization.Native/Interop.Collation.cs
```diff
@@ -60,6 +60,10 @@ public override bool IsInvalid
 
             protected override bool ReleaseHandle()
             {
+                if (CultureData.InvariantMode)
```
stephentoub: In what situation would ReleaseHandle be called but find InvariantMode == true?

src/mscorlib/src/System/Text/Normalization.Unix.cs
```diff
@@ -4,13 +4,20 @@
...
 using System.Text;
+using System.Globalization;
 
 namespace System.Text
 {
     static partial class Normalization
     {
         public static bool IsNormalized(this string strInput, NormalizationForm normalizationForm)
         {
+            if (CultureData.InvariantMode)
```
jkotas: This check should be after ValidateArguments

src/mscorlib/src/System/Text/Normalization.Unix.cs
```diff
@@ -25,6 +32,12 @@ public static bool IsNormalized(this string strInput, NormalizationForm normaliz
 
         public static string Normalize(this string strInput, NormalizationForm normalizationForm)
         {
+            if (CultureData.InvariantMode)
```
jkotas: Same here

src/mscorlib/src/System/Globalization/CompareInfo.Invariant.cs
```diff
@@ -0,0 +1,230 @@
...
+
+using System.Diagnostics;
+using System.Diagnostics.Contracts;
+
+namespace System.Globalization
+{
+    public partial class CompareInfo
+    {
+        internal static unsafe int InvariantIndexOfOrdinal(string source, string value, int startIndex, int count, bool ignoreCase)
+        {
```
stephentoub: Presumably these arguments have all already been validated?

src/mscorlib/src/System/Globalization/CompareInfo.Invariant.cs
```diff
@@ -0,0 +1,230 @@
...
+            fixed (char* pSource = source) fixed (char* pValue = value)
+            {
+                char* pSrc = &pSource[startIndex];
+                int index = InvariantFindStringOrdinal(pSrc, count, pValue, value.Length, true, ignoreCase);
```
stephentoub: Nit: can you use named arguments for things like the true here and false below?  From context I'm assuming it's about first vs last, but it'd be easier to read if it was clear from the call site.

src/mscorlib/src/Interop/Unix/System.Globalization.Native/Interop.Collation.cs
tarekgh: This is not really needed now. I had the initial implementation which was keeping the sort handle but now I don't initialize it at all so it is not needed now. I'll convert this to assert instead

src/mscorlib/src/System/Globalization/CompareInfo.Invariant.cs
tarekgh: yes. but I can add some asserts for it

src/mscorlib/src/System/Globalization/CompareInfo.Invariant.cs
tarekgh: sure. it is good idea anyway :-)

tarekgh:

src/mscorlib/src/System/Globalization/CompareInfo.Invariant.cs
```diff
@@ -0,0 +1,230 @@
...
+                        sourceChar = InvariantToUpper(source[ctrSource]);
+                        if (sourceChar != firstValueChar)
+                        {
+                            continue;
+                        }
+
+                        for (ctrValue = 1; ctrValue < valueCount; ctrValue++)
+                        {
+                            sourceChar = InvariantToUpper(source[ctrSource + ctrValue]);
+                            valueChar = InvariantToUpper(value[ctrValue]);
```
stephentoub: We already have routines for doing this kind of thing (though I don't know if we have it for this exact thing), used for ASCII-only performance paths.  Can we look at reusing where possible?  That would help to a) ensure the fallback is using well-tested code paths, b) any optimizations to one would apply to the other, and c) for places where we don't currently have ASCII-only fast paths, we might be able to leverage this work to add some.

stephentoub: Related to that, how are we thinking about the performance of this invariant mode?  Presumably one of the scenarios for it is apps that don't want to carry the ICU data around because they care about being lean, in which case throughput may also be of importance?

src/mscorlib/src/System/Globalization/CompareInfo.cs
```diff
@@ -170,12 +176,20 @@ public static CompareInfo GetCompareInfo(String name)
...
+            {
+                return true;
+            }
             char *pChar = &ch;
             return IsSortable(pChar, 1);
         }
 
         public static unsafe bool IsSortable(string text)
         {
+            if (CultureData.InvariantMode)
```
jkotas: This should be after the argument validation at least; or even better in the common worked method.

src/mscorlib/src/Interop/Unix/System.Globalization.Native/Interop.Collation.cs
jkotas: How can we get here for the invariant mode?

src/mscorlib/src/System/Globalization/CultureData.Unix.cs
```diff
@@ -14,6 +14,20 @@ namespace System.Globalization
 {
     internal partial class CultureData
     {
+        private static bool s_invariantGlobalizationMode = InitGlobalizationInvariantMode();
```
jkotas: This should be initialized explicitly during startup, not using implicit constructor. The implicit constructor as written will turn into two checks each time. 1. have we initialized already? 2. What the value is? So these checks are twice as expensive than what they need to be.

Or even better - we may want to cache the flag in a field in the objects with hot paths like TextInfo, and check the instance field there. Instance fields are faster to access than static fields.

src/mscorlib/src/System/Globalization/CompareInfo.Invariant.cs
```diff
@@ -0,0 +1,230 @@
...
+                int index = InvariantFindStringOrdinal(pSrc, count, pValue, value.Length, false, ignoreCase);
+                if (index >= 0)
+                {
+                    return index + startIndex - count + 1;
+                }
+                return -1;
+            }
+        }
+
+        private static unsafe int InvariantFindStringOrdinal(char* source, int sourceCount, char* value, int valueCount, bool start, bool ignoreCase)
```
stephentoub: What does it mean that we use both "Invariant" and "Ordinal" in the name?  Do they connote different things?

src/mscorlib/src/System/Globalization/CultureData.Unix.cs
stephentoub: Since corelib is crossgen'd, I assume the normal tricks around static readonly bools becoming JIT-time consts wouldn't apply, right?

src/mscorlib/src/System/Globalization/TextInfo.cs
```diff
@@ -311,20 +327,118 @@ public virtual String ListSeparator
...
+
+                string result = string.FastAllocateString(s.Length);
+                fixed (char* pResult = result)
+                {
+                    for (int j = 0; j < i; j++)
+                    {
+                        pResult[j] = pSource[j];
+                    }
+                    
+                    pResult[i] = (Char)(pSource[i] & ~0x20);
```
danmosemsft: lower case `char` 
and above also

src/mscorlib/src/System/Globalization/TextInfo.Windows.cs
```diff
@@ -49,44 +41,44 @@ private unsafe string ChangeCase(string s, bool toUpper)
...
+                                                    pSource,
+                                                    nLengthInput,
+                                                    pResult,
+                                                    nLengthInput,
+                                                    null,
+                                                    null,
+                                                    _sortHandle);
+            }
+
+            if (0 == ret)
```
danmosemsft: I don't see anything in the coding guidelines about Yoda conditionals but I think we avoid them.

src/mscorlib/src/System/CLRConfig.cs
jkotas: Do you mean CoreRT? I do not think we will want any of CLRConfig.cs in CoreRT.

src/mscorlib/src/System/Globalization/CultureData.Unix.cs
```diff
@@ -14,6 +14,20 @@ namespace System.Globalization
...
+        
+        private static bool InitGlobalizationInvariantMode()
+        {
+            bool invariantEnabled = CLRConfig.GetBoolValue("System.Globalization.Invariant");
+            if (!invariantEnabled && Interop.GlobalizationInterop.ICUPresent() == 0)
+            {
+                string message = "Couldn't find a valid ICU package installed on the system. " + 
+                                 "Set the configuration flag System.Globalization.Invariant to true if want to run with no globalization support.";
```
danmosemsft: ` if want to ` -> `if you want to`

src/mscorlib/src/System/Globalization/CompareInfo.cs
```diff
@@ -459,6 +491,34 @@ internal static unsafe int CompareOrdinalIgnoreCase(string strA, int indexA, int
                     length--;
                 }
 
+                if (CultureData.InvariantMode)
+                {
+                    while (length != 0)
```
stephentoub: Is the sole difference between this section and the one above it that this one doesn't limit itself to just ASCII chars, so that we don't call to the OS in invariant mode?  Consider a) adding a comment, and b) moving this out into a separate function.

src/mscorlib/src/System/Globalization/CompareInfo.cs
```diff
@@ -170,12 +176,20 @@ public static CompareInfo GetCompareInfo(String name)
 
         public static unsafe bool IsSortable(char ch)
         {
+            if (CultureData.InvariantMode)
+            {
+                return true;
+            }
```
stephentoub: Rather than having these guards in every function, would it be possible to, for example, implement a special CultureInfo for invariant all in managed code with functions like you have, and then if ICU isn't present / the config switch is thrown, just return that invariant CulutureInfo from anything that can give back a CultureInfo?  Maybe that's not feasible, but it seems like it could help to consolidate the logic better.

src/mscorlib/src/System/Globalization/CompareInfo.cs
```diff
@@ -459,6 +491,34 @@ internal static unsafe int CompareOrdinalIgnoreCase(string strA, int indexA, int
                     length--;
                 }
 
+                if (CultureData.InvariantMode)
```
danmosemsft: is it necessary to duplicate this block? can't the condition above be changed from
`while (length != 0 && (*a <= 0x80) && (*b <= 0x80))`
to
`while (length != 0 && ((*a <= 0x80) && (*b <= 0x80)) || CultureData.InvariantMode)`

stephentoub:
How are we testing this?

src/mscorlib/src/System/CLRConfig.cs
danmosemsft: Unless this code may one day be shared with CoreRT, it will never get compiled without CORECLR defined. It's defined in System.Private.CoreLib.csproj

danmosemsft:
What is your thinking about testing that we don't accidentally load ICU when we aren't supposed to? It would be easy to introduce a code path a month from now that does this.
Should we be able to periodically run certain tests in a situation where ICU isn't available? Or we rely on just checking periodically under the debugger?
Could we add code that gets called when ICU was loaded, and make it (when enabled by a flag that some of our tests set) cause a failfast?

src/corefx/System.Globalization.Native/icushim.cpp
tarekgh: I'll do that

src/mscorlib/src/System/Globalization/CompareInfo.Invariant.cs
tarekgh: I have already used the existing current optimized implementation in the functionality like string comparisons. but for IndexOf, looks we don't have a version for that. in short, I used whatever exist and provided the missing. Please point me to any implementation I should used if I missed it.
For the performance I agree in general. for the areas we can optimize it more we can look at it and do whatever needed there. I think the current implementation is reasonable for this check in and we can follow up for the optimization later. is this ok?

src/mscorlib/src/System/Globalization/CompareInfo.Invariant.cs
tarekgh: I'll remove the word Ordinal to reduce the confusion. it was meant in the invariant mode we do ordinal comparison.

src/mscorlib/src/System/Text/Normalization.Unix.cs
tarekgh: will do that

src/mscorlib/src/System/Text/Normalization.Unix.cs
tarekgh: will do that

src/mscorlib/src/System/Globalization/CompareInfo.cs
tarekgh: will move it after the validation. there is no common working method. we have forked implementation for different OS's

src/mscorlib/src/Interop/Unix/System.Globalization.Native/Interop.Collation.cs
tarekgh: Yes you are right. I have converted this check to assert instead.

src/mscorlib/src/System/Globalization/TextInfo.cs
tarekgh: will do that

src/mscorlib/src/System/Globalization/TextInfo.Windows.cs
tarekgh: I didn't change this code. also I am not sure why this bad to use?

src/mscorlib/src/System/Globalization/CultureData.Unix.cs
tarekgh: will fix that

src/mscorlib/src/System/Globalization/CompareInfo.cs
tarekgh: will do that

tarekgh:
For the testing question, I have already created a test which testing almost all the globalization functionality when having the config switch is on. The test expect different values that if we have the switch is off. I have been testing it on Windows and Linux. also I have tested on system having ICU installed and not having the ICU installed with and without the config switch. 
Also as you notice I have many asserts in the code in the paths that calls the OS. actually in my testing I was having these asserts as exception thrown. and I have successfully ran the test which exercise all the globalization functionality. also the test ran fine on system doesn't have ICU installed

danmosemsft:
> I have tested on system having ICU installed and not having the ICU installed with and without the config switch.

I guess what I'm wondering is how easy it would be for a dev 6 months from now to accidentally introduce a load of ICU in a non-ICU codepath and not notice. Unless we either have something in the product that notices it loads (if that's even possible) or we regularly run our tests on a box without ICU available, as you've done manually, we probably wouldn't notice.

src/mscorlib/src/System/Globalization/TextInfo.Windows.cs
danmosemsft: As you know in native code people often write `constant == variable` (aka Yoda conditionals) out of habit so that if they mistype `constant = variable` it will cause a compile error. In C# though there is no implicit conversion in most cases to boolean so eg for integers there is no danger of successfully compiling `constant = variable`. So in code bases I"ve worked on, typically it's written `variable == constant` for readability.
Just a nit and I don't see it in the style guidelines although it does seem largely followed here.

src/mscorlib/src/System/Globalization/CompareInfo.cs
tarekgh: I have spent some time before thinking about how we can implement the whole thing here. I have looked at the alternatives. for CultureInfo we already doing almost your idea mentioned here. but for other classes like CompareInfo/TextInfo I preferred the current implementation. to make it like what you suggested, I'll have to subclass CompareInfo/TextInfo and override all methods and then return such objects to the callers of such properties (like CultureInfo.CompareInfo, CultureInfo.TextInfo) . I dismissed this idea because a) we are not returning CompareInfo as we used to but we are returning a subclass of CompareInfo b) I was trying to avoid the unknown perf cost c) this will restrict us if we decided to expose a constructors for such classes in the future d) we'll have to use most of the existing code in CompareInfo anyway. we can refactor the code for this one though.

tarekgh:
@danmosemsft I understand your point. we need to be more robust to guard against mistakes in the future. you are talking about ICU here but we should think about Windows too which I am seeing them as same.  I think the best here is we create a static analysis tool that can scan the whole code and ensure all paths to the pinvokes are guarded.

src/mscorlib/src/System/Globalization/CompareInfo.cs
tarekgh: I didn't do that for the perf reason. this will ass extra check with every character.

src/mscorlib/src/System/Globalization/TextInfo.Windows.cs
tarekgh: I'll fix this one

tarekgh:
@jkotas @stephentoub @danmosemsft I think I have addressed all the feedback. please let me know if I missed anything or we are ok to go. Thanks.

src/mscorlib/src/System/Globalization/CompareInfo.cs
danmosemsft: What about 

``` C#
             byte comparator = CultureData.InvariantMode ? 0xFF : 0x80;

             fixed (char* ap = strA) fixed (char* bp = strB)
             {		             
                 char* a = ap + indexA;
                 char* b = bp + indexB;
 		 
                 while (length != 0 && (*a <= comparator) && (*b <= comparator))
                 {
```

that would avoid the duplication al,so.

danmosemsft: Or a separate function as @stephentoub suggests. It's just confusing two have two blocks like this IMO.

src/mscorlib/src/System/Globalization/CompareInfo.Invariant.cs
```diff
@@ -0,0 +1,238 @@
...
+            }
+
+            byte [] keyData;
+            if (source.Length == 0)
+            { 
+                keyData = EmptyArray<Byte>.Value;
+            }
+            else
+            {
+                // As we word as ordinal, then generate the sort key as the codepoint ordinal values
```
ellismg: I don't understand what this comment means.

src/mscorlib/src/System/Globalization/CompareInfo.cs
tarekgh: I think this is a good idea. I'll apply it. I'll use 0xFFFF (not 0xFF)

src/mscorlib/src/System/Globalization/CompareInfo.Invariant.cs
tarekgh: in the invariant mode, all string comparison are done as ordinal. so when generating the sort keys we generate it according to this fact. so comparing the sort keys will be same as comparing teh actual strings. I can change the comment to be clearer. I agree the comment is not clear as I meant to say "compare" instead of "word" :-(

src/mscorlib/src/System/Globalization/CompareInfo.cs
danmosemsft: Yes thanks I meant 0xFFFF :) ... this should get a good comment, though, as I had to think about it.

src/mscorlib/src/System/Globalization/TextInfo.Windows.cs
```diff
@@ -49,44 +41,44 @@ private unsafe string ChangeCase(string s, bool toUpper)
...
+            //
+            //  Create the result string.
+            //
+            string result = string.FastAllocateString(nLengthInput);
+
+            fixed (char* pSource = s)
+            fixed (char* pResult = result)
+            {
+                ret = Interop.Kernel32.LCMapStringEx(_sortHandle != IntPtr.Zero ? null : _textInfoName,
+                                                    toUpper ? LCMAP_UPPERCASE | linguisticCasing : LCMAP_LOWERCASE | linguisticCasing,
```
ellismg: NIT:  Would this be clearer as `(lingusticCasing | (toupper ? LCMAP_UPPERCASE : LCMAP_LOWERCASE))`.

Your call.

src/mscorlib/src/System/Globalization/CultureData.Windows.cs
```diff
@@ -21,6 +21,11 @@ namespace System.Globalization
 
     internal partial class CultureData
     {
+        internal static void InitGlobalizationInvariantMode()
+        {
+            s_invariantGlobalizationMode = CLRConfig.GetBoolValue(c_InvariantModeConfigSwitch);
```
danmosemsft: Why would I ever enable this on Windows? To be certain of behavioral compat with my Unix users?

src/mscorlib/src/System/Globalization/CompareInfo.Invariant.cs
danmosemsft: @benadams who may know of something.

src/mscorlib/src/System/Globalization/CultureData.Windows.cs
Petermarcu: So your app behaves the same no matter where you run it.

src/mscorlib/src/System/Globalization/CompareInfo.Invariant.cs
```diff
@@ -0,0 +1,238 @@
...
+                            return ctrSource;
+                        }
+                    }
+                }
+                else
+                {
+                    char firstValueChar = value[0];
+                    for (ctrSource = 0; ctrSource <= lastSourceStart; ctrSource++)
+                    {
+                        sourceChar = source[ctrSource];
```
danmosemsft: Why special case the first character instead of just 
```c#
                else
                {
                    for (ctrSource = 0; ctrSource <= lastSourceStart; ctrSource++)
                    {
                        for (ctrValue = 0; ctrValue < valueCount; ctrValue++)
                        {
                            sourceChar = source[ctrSource + ctrValue];
                            valueChar = value[ctrValue];

                            if (sourceChar != valueChar)
                            {
                                break;
                            }
                        }
...
```

In the two case insensitive blocks, you've special cased the first character of the source to avoid having to ToUpperCase it more than once, but in the case sensitive blocks, there's no value is there?

danmosemsft:
LGTM otherwise, if the others folk sign off.

src/classlibnative/config/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
```
jkotas: Creating a new directory, makefile and lib to make for just a single function looks like overkill. Can't this be added to existing one?

src/mscorlib/src/System/Globalization/CultureData.cs
```diff
@@ -308,6 +309,10 @@ private static StringStringDictionary RegionNames
         private static volatile StringCultureDataDictionary s_cachedRegions;
         private static volatile StringStringDictionary s_RegionNames;
 
+        private const string c_InvariantModeConfigSwitch = "System.Globalization.Invariant";
+        private static bool s_invariantGlobalizationMode;
+        internal static bool InvariantMode { get { return s_invariantGlobalizationMode; } }
```
jkotas: There are other beforefieldinit statics in this type (e.g. `s_lock`), so this will still require before field init check on every access...

src/mscorlib/src/System/Globalization/CultureData.Unix.cs
jkotas: Right, that's why it would be best to cache the bool in instance fields where possible - instance fields are the fastest to access.

src/corefx/System.Globalization.Native/icushim.cpp
```diff
@@ -170,6 +170,13 @@ bool FindLibWithMajorMinorSubVersion(int* majorVer, int* minorVer, int* subVer)
     return false;
 }
 
+static int32_t g_icuPresent = 0;
+
+extern "C" int32_t GlobalizationNative_ICUPresent()
+{
+    return g_icuPresent;
```
jkotas: So this means that we will *always* load ICU if there is one available, and then not use it. It is wasteful to load 20MB binary that we are going to not use, and makes it hard to verify the correctness by construction.

This should be changed so that the ICU is loaded only when the globalization is invariant globalization is not enabled.

src/mscorlib/src/System/Globalization/CompareInfo.Invariant.cs
```diff
@@ -0,0 +1,238 @@
...
+                    }
+                }
+            }
+
+            return -1;
+        }
+
+        private static char InvariantToUpper(char c)
+        {
+            return ('a' <= c && c <= 'z') ? (char)(c - 0x20) : c;
```
jkotas: Most other places use the uint trick to save one of the checks (e.g. ToLowerAsciiInvariant). Do the same here?

src/mscorlib/src/System/Globalization/CompareInfo.Invariant.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            if ((options & ValidSortkeyCtorMaskOffFlags) != 0)
+            {
+                throw new ArgumentException(SR.Argument_InvalidFlag, nameof(options));
+            }
+
+            byte [] keyData;
+            if (source.Length == 0)
+            { 
+                keyData = EmptyArray<Byte>.Value;
```
jkotas: `EmptyArray<byte>` is not in CoreRT. Use `Array.Empty<byte>()` instead.

jkotas:
> the best here is we create a static analysis tool that can scan the whole code and ensure all paths to the pinvokes are guarded.

Similar one-off tools never pay off, and they do not guarantee that the thing *actually* works.

Your primary quality plan has to be regular testing with this on. For example, make the globalization tests with this on to be part of regular outter loop run.

src/corefx/System.Globalization.Native/icushim.cpp
stephentoub: I agree.  Wouldn't that be easily addressed just by having InitializeICUShim first check the config flag?  Seems like the rest would fall out naturally.

tarekgh: there is no easy way to access the config flag without doing pinvoke and passing the flag to system.Globalization.Native. I'll change this to have the  managed side specifically pinvoke to initialize the ICU if the flag is not set

src/mscorlib/src/System/Globalization/CultureData.Unix.cs
tarekgh: sorry I am not clear about that. I am using this bool static in many static methods. if I cache it in instance field how I am going to use it in such static methods? could you please clarify more what you mean here?

src/mscorlib/src/System/Globalization/CompareInfo.cs
```diff
@@ -182,6 +192,11 @@ public static unsafe bool IsSortable(string text)
                 throw new ArgumentNullException(nameof(text));
             }
 
+            if (CultureData.InvariantMode)
```
jkotas: This should be after the `0 == text.Length` check.

src/mscorlib/src/System/Text/Normalization.Unix.cs
```diff
@@ -13,6 +14,12 @@ public static bool IsNormalized(this string strInput, NormalizationForm normaliz
         {
             ValidateArguments(strInput, normalizationForm);
 
+            if (CultureData.InvariantMode)
+            {
+                // work ordinal, then all characters are normalized in this mode
```
jkotas: I still do not understand what is `work ordinal` trying to say here.

src/mscorlib/src/System/Globalization/CultureData.Unix.cs
jkotas: Right, this would not work everywhere. But it would work for the string manipulation methods on TextInfo or CompareInfo that are probably the most impacted ones.

Feel free to ignore this suggestion if you do not think it is a good idea. It would be useful to pick a few of the leanest paths that the check is getting added to and measure perf impact of this change on them.

src/mscorlib/src/System/Globalization/CultureData.cs
jkotas: It may be best to move the flag into a separate class - it would also make it better shareable with CoreRT.

src/corefx/System.Globalization.Native/icushim.cpp
stephentoub: Sounds good.

src/mscorlib/src/System/Globalization/CultureData.cs
stephentoub: Subsequent to this change, should we consider exposing this publicly from somewhere?  I could see someone wanting to check it for verification purposes.

tarekgh: > Subsequent to this change, should we consider exposing this publicly from somewhere? I could see someone wanting to check it for verification purposes

This is can be achieved by reading the config flag using AppContext.

tarekgh:
I am still testing on Linux so hold on till I finish testing it. thanks.

tarekgh:
I have finished testing on Linux. please let me know if you have any other feedback or we are good to go.

tarekgh:
I forgot to mention, I have applied @jkotas suggestion for the static field which is created a separate class to have the static field and get it initialized there. also for the critical paths in CompareInfo and TextInfo, I have created instance fields to cache the static value so we'll not do 2 checks as Jan mentioned.

danmosemsft:
I have no more feedback

jkotas:
Any of perf measurements?

src/mscorlib/src/System/Globalization/CompareInfo.cs
```diff
@@ -741,9 +796,22 @@ public unsafe virtual int IndexOf(String source, String value, int startIndex, i
...
 
+            if (_invariantMode)
+                return IndexOfOrdinal(source, value, startIndex, count, ignoreCase: (options & (CompareOptions.IgnoreCase | CompareOptions.OrdinalIgnoreCase)) != 0);
+
             return IndexOfCore(source, value, startIndex, count, options, null);
         }
 
+        internal static unsafe int IndexOfOrdinal(string source, string value, int startIndex, int count, bool ignoreCase)
+        {
+            if (GlobalizationMode.Invariant)
```
jkotas: Move this to the caller in this file, so that it can use the faster instance check?

src/mscorlib/src/System/Globalization/CompareInfo.cs
```diff
@@ -927,9 +998,22 @@ public unsafe virtual int LastIndexOf(String source, String value, int startInde
...
 
+            if (_invariantMode)
+                return InvariantLastIndexOf(source, value, startIndex, count, (options & (CompareOptions.IgnoreCase | CompareOptions.OrdinalIgnoreCase)) != 0);
+
             return LastIndexOfCore(source, value, startIndex, count, options);
         }
 
+        internal static unsafe int LastIndexOfOrdinal(string source, string value, int startIndex, int count, bool ignoreCase)
+        {
+            if (GlobalizationMode.Invariant)
```
jkotas: Same here

src/mscorlib/src/System/Text/Normalization.Windows.cs
```diff
@@ -49,6 +56,13 @@ internal static bool IsNormalized(String strInput, NormalizationForm normForm)
 
         internal static String Normalize(String strInput, NormalizationForm normForm)
         {
+            if (GlobalizationMode.Invariant)
+            {
+                // In Invariant mode we assume all characters are normalized.
```
jkotas: This mean that there will be a bunch of URL parsing tests failing when this is turned on because of they depend on normalization. Have you tried running all corefx tests with this (not just the globalization ones) - what is failing?

src/vm/ceemain.cpp
```diff
@@ -162,6 +162,7 @@
 #ifdef FEATURE_USE_LCID
 #include "nlsinfo.h"
 #endif 
+#include "clrconfiguration.h"
```
jkotas: Why do you need to include it here?

src/classlibnative/inc/clrconfiguration.h
```diff
@@ -0,0 +1,17 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
```
jkotas: This file was moved to vm

src/vm/clrconfiguration.cpp
```diff
@@ -0,0 +1,21 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
jkotas: Nit: It may be nice to call it clrconfignative.* - the "native" suffix is used to differentiate the file with FCalls/QCalls from the regular impl file.

tarekgh:
> Any of perf measurements?

I didn't have chance to measure the perf yet. I'll do that later if this is ok as I already applied your suggest using the instance fields.

src/mscorlib/src/System/Text/Normalization.Windows.cs
tarekgh: I have ran the whole tests but didn't enable the flag for all tests. I just enabled the flag for the test I created specifically covering the globalization. so I am not expecting the URL test to fail as we don't enable the invariant mode on such test. this apply to the current globalization tests, without enabling the invariant mode we shouldn't change the behavior. 

I am going to do more tests on Linux as I expect some failures after we have changed the way we initialize ICU.

src/mscorlib/src/System/Globalization/CompareInfo.cs
tarekgh: This called from different places. I am going to convert this method to be instance method instead of static and use the fast check.

src/mscorlib/src/System/Text/Normalization.Windows.cs
jkotas: >  changed the way we initialize ICU

Why should it cause tests to fail?

tarekgh: if there is any library using System.Globalization.Native.so will need to initialize the ICU manually as we stopped initializing it automatically

jkotas: CoreLib should be the only one using System.Globalization.Native.so

src/vm/ceemain.cpp
tarekgh: I'll remove it. it was left from previous changes we already removed

src/mscorlib/src/System/Text/Normalization.Windows.cs
tarekgh: good. I was not sure of that. then we should be good here.

tarekgh:
@jkotas I pushed the changes you have requested. thanks

src/mscorlib/src/System/Globalization/CompareInfo.cs
```diff
@@ -741,9 +796,22 @@ public unsafe virtual int IndexOf(String source, String value, int startIndex, i
             if ((options & ValidIndexMaskOffFlags) != 0 && (options != CompareOptions.Ordinal))
                 throw new ArgumentException(SR.Argument_InvalidFlag, nameof(options));
 
+            if (_invariantMode)
+                return IndexOfOrdinal(source, value, startIndex, count, ignoreCase: (options & (CompareOptions.IgnoreCase | CompareOptions.OrdinalIgnoreCase)) != 0);
+
             return IndexOfCore(source, value, startIndex, count, options, null);
         }
 
+        internal unsafe int IndexOfOrdinal(string source, string value, int startIndex, int count, bool ignoreCase)
```
jkotas: Does not need to be unsafe.

src/mscorlib/src/System/Globalization/CompareInfo.cs
```diff
@@ -927,9 +998,22 @@ public unsafe virtual int LastIndexOf(String source, String value, int startInde
                 return LastIndexOfOrdinal(source, value, startIndex, count, ignoreCase: true);
             }
 
+            if (_invariantMode)
+                return InvariantLastIndexOf(source, value, startIndex, count, (options & (CompareOptions.IgnoreCase | CompareOptions.OrdinalIgnoreCase)) != 0);
+
             return LastIndexOfCore(source, value, startIndex, count, options);
         }
 
+        internal unsafe int LastIndexOfOrdinal(string source, string value, int startIndex, int count, bool ignoreCase)
```
jkotas: Same here

src/mscorlib/src/System/Globalization/GlobalizationMode.Unix.cs
```diff
@@ -0,0 +1,24 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Globalization
+{
+    internal sealed partial class GlobalizationMode
+    {
+        internal static bool GetGlobalizationInvariantMode()
```
jkotas: private?

src/mscorlib/src/System/Globalization/GlobalizationMode.Windows.cs
```diff
@@ -0,0 +1,14 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Globalization
+{
+    internal sealed partial class GlobalizationMode
+    {
+        internal static bool GetGlobalizationInvariantMode()
```
jkotas: Same here

jkotas:
LGTM modulo few nits.

src/mscorlib/src/System/Globalization/GlobalizationMode.cs
```diff
@@ -0,0 +1,14 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Globalization
+{
+    internal sealed partial class GlobalizationMode
+    {
+        private const string c_InvariantModeConfigSwitch = "System.Globalization.Invariant";
+        private static bool s_invariantMode = GetGlobalizationInvariantMode();
```
jkotas: This can be readonly

jkotas: Or even `internal static bool Invariant { get; } = GetGlobalizationInvariantMode();`

tarekgh:
@jkotas thanks for your review and the feedback. I have one question. one of your comments you asked to remove the #if CORECLR. my question is, when these changes get merged with the corert, I think this will break. right? GlobalizationMode.Invariant get initialized by the QCall which is not in coreRT. so I think we'll need to add the #if CORECLR in the initialization to avoid the QCall calling in corert.

tarekgh:

jkotas:
> when these changes get merged with the corert, I think this will break. right?

I expect that we are not going to share this file with CoreRT - this file will be CoreCLR specific. There will be CoreRT specific version of it that reads it directly from AppContext and has potentially other CoreRT specific tweaks.

jkotas:
Also, could you please  work @russellhadley and @erozenfeld to figure out the plan how to get this integrated with the linker that they are working on? This can make standalone apps that opt into invariant globalization smaller; and it can be also used to test that you got all paths to ICU covered with the condition - the linker should tree shake all calls to ICU.

tarekgh:
> I expect that we are not going to share this file with CoreRT - this file will be CoreCLR specific. There will be CoreRT specific version of it that reads it directly from AppContext and has potentially other CoreRT specific tweaks.

I was just wondering about the calls of GlobalizationMode.Invarian in many places inside the globalization code. sore corert should have the definition of GlobalizationMode or otherwise the code will not compile. I may be missing something but anyway I'll go ahead and merge the changes here and we can handle any break if we get any.

tarekgh:

tarekgh:
I'll work with @russellhadley and @erozenfeld as you suggested too.

russellhadley:
Yes, we should be able to deadcode the appropriate code paths if the invariant flags is know at publish time.  @erozenfeld can you make sure that we have a backlog item for this?

tarekgh:
@erozenfeld please let me know if you need anything from my side. I'll be happy to help.

Issue #10059, opened 2017-03-09
[x86/Linux] JIT.Methodical.Invoke.deep._il_reldeep1 test failure


parjong:
CLR hits the following assert while running ``JIT.Methodical.Invoke.deep._il_reldeep1`` test:
```
Assert failure(PID 9972 [0x000026f4], Thread: 9972 [0x26f4]): Assertion failed 'getEmitter()->emitMaxStackDepth <= maxAllowedStackDepth' in 'JitTest.Test:Main():int' (IL size 1014)

    File: /home/parjong/projects/dotnet/coreclr/src/jit/codegencommon.cpp Line: 3176
    Image: /home/parjong/projects/dotnet/dotnet-overlay/Linux.x86.Debug.Debug/corerun

Aborted (core dumped)
```

parjong:
Here is the stack trace and the value of interest when hitting the assert:
```
Assert failure(PID 3353 [0x00000d19], Thread: 3353 [0x0d19]): Assertion failed 'getEmitter()->emitMaxStackDepth <= maxAllowedStackDepth' in 'JitTest.Test:Main():int' (IL size 1014)

    File: /home/parjong/projects/dotnet/coreclr/src/jit/codegencommon.cpp Line: 3207
    Image: /home/parjong/projects/dotnet/dotnet-overlay/Linux.x86.Debug.Debug/corerun


Program received signal SIGABRT, Aborted.
0xf7fd9d80 in __kernel_vsyscall ()
(gdb) bt 15
#0  0xf7fd9d80 in __kernel_vsyscall ()
#1  0xf7ce2687 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#2  0xf7ce5ab3 in __GI_abort () at abort.c:89
#3  0xf790fde1 in PROCEndProcess (hProcess=0xffffff01, uExitCode=123456789, bTerminateUnconditionally=1) at /home/parjong/projects/dotnet/coreclr/src/pal/src/thread/process.cpp:1385
#4  0xf790ff51 in TerminateProcess (hProcess=0xffffff01, uExitCode=123456789) at /home/parjong/projects/dotnet/coreclr/src/pal/src/thread/process.cpp:1301
#5  0xf6cb1d32 in TerminateOnAssert () at /home/parjong/projects/dotnet/coreclr/src/utilcode/debug.cpp:183
[...]

PR #10266, opened 2017-03-17
Fix Linux/x86 call alignment calculation and insertion

### LINKED ISSUE #10122: [x86/Linux] Nested calls don't handle stack alignment properly
On Linux/x86, the stack must be aligned to 16 bytes when a call is made. We do this by computing the number of alignment bytes required, saving this on the first PUTARG_STK node, and generating an additional, preceding "sub esp, N" instruction before the first stack argument is generated. Then, after the call, we add back the alignment value, if any.

The alignment is calculated in fgArgInfo::ArgsAlignPadding().

The problem is, it assumes the stack is aligned when the arguments for a call start being pushed. This is not true for nested calls, for which we may push some args for an outer call, then push some args for an inner call, make the inner call, push some more args for the outer call, then make the outer call.

Consider the following program:
```
using System;

public class Program
{

    public static float f1(float a)
    {
        Console.WriteLine(a);
        return a + 1.0f;
    }

    public static float f2(float a, float b)
    {
        Console.WriteLine(a);
        return a + b + 1.0f;
    }

    {
        float b = f2(2.0f, f1(1.0f));
        Console.WriteLine(b);

        return 100;
    }
}
```
which generates this code:
```
movss    xmm0, dword ptr [@RWD00]
sub      esp, 8 // f2 takes 8 bytes of stack arguments, so generate an 8-byte alignment adjustment
sub      esp, 4 // "push" the float argument
movss    dword ptr [esp], xmm0
movss    xmm0, dword ptr [@RWD04]
sub      esp, 12 // f1 takes 4 bytes of stack arguments, so generate a 12-byte alignment adjustment
                        // NOTE: the stack has already been adjusted by 12 bytes! No alignment adjustment is needed!
sub      esp, 4 // "push" the inner float arg
movss    dword ptr [esp], xmm0
call     [Program:f1(float):float]
add      esp, 12 // restore adjustment
fstp     dword ptr [TEMP_01 ebp-04H]
movss    xmm0, dword ptr [TEMP_01 ebp-04H]
sub      esp, 4 // "push" second outer arg
movss    dword ptr [esp], xmm0
call     [Program:f2(float,float):float]
add      esp, 8 // "restore" alignment
```

Option 1 for fixing this would be to have ArgsAlignPadding() take into account all argument and alignment stack adjustments that have been made by outer calls at the point the inner call args start being pushed.

Option 2 for fixing this would be to hoist all nested calls, preventing arguments from multiple nested calls being on the stack simultaneously (one one call's arguments would be set up and pushed at a time). This is done (and required by ABI) for all non-x86 targets. It might lead to sub-optimal code due to the need for temps to store intermediate results.

Related issues: #10002, #10059.

BruceForstall:
The existing system of calculating per-call alignment adjustments
did not consider the possibility of nested calls. On x86, some
arguments for a call might be pushed on the stack before a
nested call's arguments start being pushed. Thus, a call can't
consider only its own arguments when determining the stack level.

Instead, use the existing genStackLevel variable, updated
dynamically during code generation, to determine the baseline
stack alignment when a call first needs to push something on
the stack, or if it has no stack arguments, whether it needs to
align the stack before a call.

One wrinkle here is that the fgAddCodeRef() function for adding
throw blocks for array bounds checks, and other similar helpers,
need to know the stack level on entry, and this stack level is
computed during argument morphing. There is a tough phase ordering
problem here. So, we bail out and force an EBP frame if there are
any such throw helpers in the function. When using an EBP frame,
the helper block doesn't need to know the stack level -- it is
only needed for ESP frames, needed for unwinding. Note that this
bail out already existed if the same helper needed multiple different
stack levels on entry. We could do slightly better without too much
work by not bailing out for top-level calls with no stack arguments.

Fixes #10122

### DIFF
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -107,7 +107,7 @@ CodeGen::CodeGen(Compiler* theCompiler) : CodeGenInterface(theCompiler)
     m_stkArgVarNum = BAD_VAR_NUM;
 #endif
 
-#if defined(DEBUG) && defined(UNIX_X86_ABI)
+#if defined(UNIX_X86_ABI)
     curNestedAlignment = 0;
     maxNestedAlignment = 0;
 #endif
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -107,6 +107,11 @@ CodeGen::CodeGen(Compiler* theCompiler) : CodeGenInterface(theCompiler)
     m_stkArgVarNum = BAD_VAR_NUM;
 #endif
 
+#if defined(DEBUG) && defined(UNIX_X86_ABI)
+    curNestedAlignment = 0;
+    maxNestedAlignment = 0;
+#endif
+
     regTracker.rsTrackInit(compiler, &regSet);
     gcInfo.regSet        = &regSet;
     m_cgEmitter          = new (compiler->getAllocator()) emitter();
@@ -3201,7 +3206,7 @@ void CodeGen::genGenerateCode(void** codePtr, ULONG* nativeSizeOfCode)
                                         genTypeStSz(TYP_LONG) +       // longs/doubles may be transferred via stack, etc
                                         (compiler->compTailCallUsed ? 4 : 0); // CORINFO_HELP_TAILCALL args
 #if defined(UNIX_X86_ABI)
-        maxAllowedStackDepth += genTypeStSz(TYP_INT) * 3; // stack align for x86 - allow up to 3 INT's for padding
+        maxAllowedStackDepth += maxNestedAlignment;
 #endif
         noway_assert(getEmitter()->emitMaxStackDepth <= maxAllowedStackDepth);
     }
--- a/src/jit/codegenlinear.h
+++ b/src/jit/codegenlinear.h
@@ -178,6 +178,40 @@ void genAlignStackBeforeCall(GenTreePutArgStk* putArgStk);
 void genAlignStackBeforeCall(GenTreeCall* call);
 void genRemoveAlignmentAfterCall(GenTreeCall* call);
 
+#if defined(DEBUG) && defined(UNIX_X86_ABI)
+
+    unsigned curNestedAlignment;    // Keep track of alignment adjustment required during codegen.
+    unsigned maxNestedAlignment;    // The maximum amount of alignment adjustment required.
+
+    void SubtractNestedAlignment(unsigned adjustment)
+    {
+        assert(curNestedAlignment >= adjustment);
+        unsigned newNestedAlignment = curNestedAlignment - adjustment;
+        if (curNestedAlignment != newNestedAlignment)
+        {
+            JITDUMP("Adjusting stack nested alignment from %d to %d\n", curNestedAlignment, newNestedAlignment);
+        }
+        curNestedAlignment = newNestedAlignment;
+    }
+
+    void AddNestedAlignment(unsigned adjustment)
+    {
+        unsigned newNestedAlignment = curNestedAlignment + adjustment;
+        if (curNestedAlignment != newNestedAlignment)
+        {
+            JITDUMP("Adjusting stack nested alignment from %d to %d\n", curNestedAlignment, newNestedAlignment);
+        }
+        curNestedAlignment = newNestedAlignment;
+
+        if (curNestedAlignment > maxNestedAlignment)
+        {
+            JITDUMP("Max stack nested alignment changed from %d to %d\n", maxNestedAlignment, curNestedAlignment);
+            maxNestedAlignment = curNestedAlignment;
+        }
+    }
+
+#endif
+
 #ifdef FEATURE_PUT_STRUCT_ARG_STK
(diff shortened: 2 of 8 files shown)


BruceForstall:
@seanshpark @parjong This change fixes to Linux/x86 alignment problem detailed in #10122 . The asm diffs look good to me. Can you test it on whatever Linux/x86 testing you do to see if it fixes the problem?

With this, I'm not sure we need to define `FEATURE_FIXED_OUT_ARGS` as described in #10012 .

BruceForstall:
@briansull PTAL
cc @dotnet/jit-contrib

briansull:
LGTM

parjong:
@BruceForstall I'm not sure whether the following result is valid or not as I currently have a problem with overlay creation (#10299), the current patch looks fine except that it introduces two regressions (which looks like #10059):
```
FAILED   - [ 1335 :    1 sec ] JIT/Directed/perffix/primitivevt/mixed2_cs_do/mixed2_cs_do.sh
               BEGIN EXECUTION
               /root/dotnet-test/coreclr_unittest_20170320_111107_x86_x86-emu/coreclr-unittest-170320-021145/Core_Root/corerun mixed2_cs_do.exe
               
               Assert failure(PID 2212 [0x000008a4], Thread: 2212 [0x08a4]): Assertion failed 'getEmitter()->emitMaxStackDepth <= maxAllowedStackDepth' in 'testout1:Func_0_1_1_1_1():long' (IL size 242)
               
                   File: /home/parjong/projects/dotnet/coreclr/src/jit/codegencommon.cpp Line: 3206
                   Image: /root/dotnet-test/coreclr_unittest_20170320_111107_x86_x86-emu/coreclr-unittest-170320-021145/Core_Root/corerun
               
               ./mixed2_cs_do.sh: line 131:  2212 Aborted                 (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" mixed2_cs_do.exe $CLRTestExecutionArguments
               Expected: 100
               Actual: 134
               END EXECUTION - FAILED
         - [ 1336 :    2 sec ] JIT/Directed/perffix/primitivevt/mixed2_cs_r/mixed2_cs_r.sh
FAILED   - [ 1337 :    2 sec ] JIT/Directed/perffix/primitivevt/mixed2_cs_ro/mixed2_cs_ro.sh
               BEGIN EXECUTION
               /root/dotnet-test/coreclr_unittest_20170320_111107_x86_x86-emu/coreclr-unittest-170320-021145/Core_Root/corerun mixed2_cs_ro.exe
               
               Assert failure(PID 2265 [0x000008d9], Thread: 2265 [0x08d9]): Assertion failed 'getEmitter()->emitMaxStackDepth <= maxAllowedStackDepth' in 'testout1:Func_0_1_1_1_1():long' (IL size 242)
               
                   File: /home/parjong/projects/dotnet/coreclr/src/jit/codegencommon.cpp Line: 3206
                   Image: /root/dotnet-test/coreclr_unittest_20170320_111107_x86_x86-emu/coreclr-unittest-170320-021145/Core_Root/corerun
               
               ./mixed2_cs_ro.sh: line 131:  2265 Aborted                 (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" mixed2_cs_ro.exe $CLRTestExecutionArguments
               Expected: 100
               Actual: 134
               END EXECUTION - FAILED
```

parjong:
This PR fixes #10001.

BruceForstall:
@parjong you wrote:

> The number of failures increases from 41 into 43 with this patch.

Is this from a full run of the dotnet/coreclr test tree? Or do you run a subset of the tests? If it is a subset, what are the tests that are run?

Can you tell me the list of the 41 failing tests?

parjong:
@BruceForstall The results come from a full run of coreclr unittest (priority 0).

Here are related XMLs:
 - [XML (41 failures)](https://github.com/dotnet/coreclr/files/856702/coreclrtests.xml.41.txt)
 - [XML (43 failures)](https://github.com/dotnet/coreclr/files/856703/coreclrtests.xml.43.txt)

BruceForstall:
@parjong check my latest commit; I think it should fix that assert.

parjong:
@BruceForstall Wow! Updated PR resolves all the assert failures when I checked manually. I'm currently running full test, but I could let you know its result tomorrow as I'll be absent in the afternoon.

I found that there is a compile error in Release build:
```
/home/parjong/projects/dotnet/coreclr/src/jit/codegencommon.cpp:3209:33: error: use of undeclared identifier 'maxNestedAlignment'
        maxAllowedStackDepth += maxNestedAlignment;
                                ^
```

BruceForstall:
@parjong I fixed the release build issue (I hadn't noticed the assert was noway_assert(), which is available in release).

BruceForstall:
@parjong This is ready to be merged, if you think it looks good. PTAL.

I know #10012 might mean we reimplement this again, but until that happens, if this unblocks your work, it's at least better that what is there right now.

parjong:
@BruceForstall LGTM, Thank you! 

This PR allows me to focus on stack misalignment issue while @seanshpark is working on #10012.

PR #10270, opened 2017-03-17
Move ParameterInfo.cs and Assembly.cs to shared partition.

AtsushiKan:
Module.cs is almost reconciled except for one line
that CoreRt will have to change so we'll leave this
unshared for now.

__Filter.cs was moved into a private member of Module
(as in CoreRt).

ReflectionOnlyLoad now throws PlatformNotSupportedException
(as in CoreRt).

Removed private fields from PropertyInfo (not used in
managed, not referenced from VM.)


AtsushiKan:
cc @jkotas 
cc @alexperovich

src/mscorlib/src/System/Reflection/ParameterInfo.CoreCLR.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using System.Runtime.Serialization;
+
+namespace System.Reflection
+{
+    public partial class ParameterInfo : ICustomAttributeProvider, IObjectReference
+    {
+        // this is an internal api for DynamicMethod. A better solution is to change the relationship
+        // between ParameterInfo and ParameterBuilder so that a ParameterBuilder can be seen as a writer
+        // api over a ParameterInfo. However that is a possible breaking change so it needs to go through some process first
```
jkotas: SetName/SetAttributes can be move to RuntimeParameterInfo if we wanted clean this up a bit. Change `ParameterInfo[] LoadParameters()` to return `RuntimeParameterInfo[]` instead, and the rest just should just fall out.

(If you would like to keep it the way you have it - that's fine with me to.)

jkotas:
LGTM

Issue #8398, opened 2016-11-30
AssemblyLoadContext throws System.InvalidOperationException: Resolved assembly's simple name should be the same as of the requested assembly.


christianhuening:
I am trying to load some plugin style DLLs from a folder with a custom AssemblyLoadContext implementation like so:
```csharp
 public class LIFEAssemblyLoader : AssemblyLoadContext
    {
        private readonly string _folderPath;


        /// <summary>
        ///   Create a new assembly loader.
        /// </summary>
        /// <param name="folderPath">Base path for the dependency resolving.</param>
        public LIFEAssemblyLoader(string folderPath) {
            _folderPath = folderPath;
            this.Resolving += OnResolving;
        }

        private Assembly OnResolving(AssemblyLoadContext assemblyLoadContext, AssemblyName assemblyName)
        {
            return Load(assemblyName);
        }

        /// <summary>
        ///   Load the requested assembly. Performs recursive calls to load required dependencies.
        /// </summary>
        /// <param name="assemblyName">Assembly description.</param>
        /// <returns>The loaded assembly.</returns>
        protected override Assembly Load(AssemblyName assemblyName) {
            var deps = DependencyContext.Default;
            var res = deps.CompileLibraries.Where(d => d.Name.Contains(assemblyName.Name)).ToList();
            if (res.Count > 0) {
                return Assembly.Load(new AssemblyName(res.First().Name));
            }
            var dllPath = _folderPath + Path.DirectorySeparatorChar + assemblyName.Name + ".dll";
            var fileinfo = new FileInfo(dllPath);
            if (File.Exists(fileinfo.FullName)) {
                return LoadFromAssemblyPath(fileinfo.FullName);
            }
            return Assembly.Load(assemblyName);
        }
    }
```

Now for every DLL that is attempted to be loaded an exception is thrown like so:
```
Caught type load error while Loading model code. Error was: System.IO.FileLoadException: Could not load file or assembly 'System.Runtime, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. An operation is not legal in the current state. (Exception from HRESULT: 0x80131509)
File name: 'System.Runtime, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' ---> System.InvalidOperationException: Resolved assembly's simple name should be the same as of the requested assembly.
[...]

Issue #6187, opened 2016-07-08
AssemblyLoadContext: allow customizing key used lookup cached assemblies


natemcmaster:
When AssemblyLoadContext triggers an assembly load, it first checks for cached assemblies already loaded. If I'm not mistaken, the key used for the cache is "Name+Version+Culture+PublicKey".

It would be useful to add additional assembly characteristics to the key (i.e. AssemblyInformationalVersion or AssemblyFileVersion).

Scenario: patch version upgrades to assemblies often carry the same name and AssemblyVersionAttribute, but patch upgrades may carry implementation changes (or security patches). If the default load context loaded an older assembly and assembly load context wants a newer, patched assembly, the load context will still get the older version because the patched version's assembly name+version didn't change.

Currently the only way to ensure the newer assembly version is loaded is to explicitly trigger an assembly load by calling `AssemblyLoadContext.LoadFromAssemblyName` before an automatic load has been triggered.

natemcmaster:
cc @gkhanna79

natemcmaster:
Btw, this could help us enable ASP.NET/EF Core tools on .NET Standard class library projects and improve how they work for .NET Core apps.

cc @NTaylorMullen @prafullbhosale

jkotas:
`AssemblyLoadContext` is used to resolve assembly references. The assembly reference (ie key for `AssemblyLoadContext`) is specified as exactly `Name+Version+Culture+PublicKey` in the [ECMA 335](https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/dotnet-standards.md) spec. The entire .NET ecosystem depends on this, it is not something that can be realistically changed. 

Could you please describe the problem/scenario you are trying to solve, without going into details on how to solve it?

gokarnm:
I'm hitting this same issue with my scenario, here is what I'm trying to do.
 
I'm creating a custom AssemblyLoadContext in my application to load a plugin from a directory. The plugin can have additional dependencies on CoreFX or third party libraries. These dependencies can be same as my application's dependencies but the plugin may use a different version. A newer version of a CoreFX package could mean a patch version update in which case only the AssemblyInformationalVersion would be changed but the AssemblyVersion may remain the same. The plugin directory contains the plugin DLL and all its dependencies.
 
[...]

Issue #5463, opened 2016-06-03
AssemblyLoadContext documentation


jkotas:
There are a lot of questions being asked on how this works - we need to have documentation for it.

jkotas:
cc @gkhanna79

jkotas:
https://blogs.msdn.microsoft.com/dotnet/2016/05/27/making-it-easier-to-port-to-net-core/#comment-77765 has some frequently asked questions that this documentation should provide answers to.

gkhanna79:
Sounds good @jkotas I will look into this next week.

Ninds:
On the blog at the link above you write :

"AssemblyLoadContext does not provide data isolation. Objects can be passed between them without"

.. but is this not then inconsistent with having separate static variables per context ?

Essentially I want parallelism with a very low dependency on having to write thread-aware code. Appdomains gave that to me.

jkotas:
I am not sure where you see the inconsistency. For both AppDomain and AssemblyLoadContext, the code that sets up your computation has to be thread-aware.

jkotas:
@Ninds Your example makes sense. It should be possible to build this with AssemblyLoadContext.

Ninds:
ah great. so isolated static variables in each context ?
But I guess we would have to build our own proxies of class which are to be called across context boundaries .. but since they should be very minimal and light that should not be a big problem.

So will we be getting some doc on this feature soon ?
Will AssemblyContexts be added to .NET Framework in the future too ?

Ninds:
.. one more thing will exceptions propagate across context boundaries ?

jkotas:
> isolated static variables in each context ?

Correct.

>  some doc on this feature soon ?

@gkhanna79 is going to start working on the docs soon.

> Will AssemblyContexts be added to .NET Framework in the future too ?

It is in our backlog - we do not have the date for it.

> one more thing will exceptions propagate across context boundaries ? 

Yes.

Ninds:
That's GREAT !
Thanks very much !
I'll come back later with any other questions.

OpinionatedGeek:
Hi

I'd like to chime in with a couple of other AssemblyLoadContext questions. (I'm told this is the place to ask them so they'll be addressed in the documentation - apologies if this isn't the right place.)

* How do I unload/discard assemblies?
* When I unload the assemblies, is the memory allocated to their objects freed (made available for GC)?

[...]

PR #10271, opened 2017-03-17
Initial document on AssemblyLoadContext Design

gkhanna79:
skip ci please

@jkotas PTAL

### DIFF
--- a/Documentation/design-docs/assemblyloadcontext.md
+++ b/Documentation/design-docs/assemblyloadcontext.md
@@ -1,4 +1,4 @@
-**LoadContext** can be viewed as a container for assemblies, their code and data. Whenever an assembly is loaded, it is loaded within a load context - independent of whether the load was triggered explicitly (e.g. via *Assembly.Load*), implicitly (e.g. resolving static assembly references from the manifest) or dynamically (by emitting code on the fly).
+**LoadContext** can be viewed as a container for assemblies, their code and data (e.g. statics). Whenever an assembly is loaded, it is loaded within a load context - independent of whether the load was triggered explicitly (e.g. via *Assembly.Load*), implicitly (e.g. resolving static assembly references from the manifest) or dynamically (by emitting code on the fly).
 
 This concept is not new to .NET Core but has existed since the days of .NET Framework (see [this](https://blogs.msdn.microsoft.com/suzcook/2003/05/29/choosing-a-binding-context/) for details) where it operated behind the scenes and not exposed for the developer to interact with, aside from loading your assembly in one based upon the API used to perform the load.
 
@@ -7,6 +7,7 @@ In .NET Core, we have exposed a [managed API surface](https://github.com/dotnet/
 * Ability to load multiple versions of the same assembly within a given process (e.g. for plugin frameworks)
 * Ability to load assemblies explicitly in a context isolated from that of the application.
 * Ability to override assemblies being resolved from application context.
+* Ability to have isolation of statics (as they are tied to the **LoadContext**)
 * Expose LoadContext as a first class concept for developers to interface with and not be a magic.
 
 ## Types of LoadContext
@@ -17,19 +18,21 @@ Every .NET Core app has a **LoadContext** instance created during .NET Core Runt
 ### Custom LoadContext
 For scenarios that wish to have isolation between loaded assemblies, applications can create their own **LoadContext** instance by deriving from **System.Runtime.Loader.AssemblyLoadContext** type and loading the assemblies within that instance.
 
+Multiple assemblies with the same simple name cannot be loaded into a single load context (*Default* or *Custom*). Also, .Net Core ignores strong name token for assembly binding process.
 
 ## How Load is attempted
 
 ### Basics
 If an assembly *A1* triggers the load of an assembly *C1*, the latter's load is attempted within the **LoadContext** instance of the former (which is also known as the *RequestingAssembly* or *ParentAssembly*). 
 
-Dynamically generates assemblies add a slight twist since they do not have a *ParentAssembly/RequestingAssembly* per-se. Thus, they are associated with the load context of their *Creator Assembly* and any subsequent loads (static or dynamic) will use that load context.
+Dynamically generated assemblies add a slight twist since they do not have a *ParentAssembly/RequestingAssembly* per-se. Thus, they are associated with the load context of their *Creator Assembly* and any subsequent loads (static or dynamic) will use that load context.
 
 ### Resolution Process
-If the assembly was already present in *A1's* context, either because we had successfully loaded it earlier or because we failed to load it for some reason, we return the corresponding status (and assembly reference for the success case).
+If the assembly was already present in *A1's* context, either because we had successfully loaded it earlier, or because we failed to load it for some reason, we return the corresponding status (and assembly reference for the success case).
+
 However, if *C1* was not found in *A1's* context, the *Load* method override in *A1's* context is invoked. 
 
-* For *Custom LoadContext*, this override is an oppurtunity to load an assembly **before** the fallback (see below) to *Default LoadContext* is attempted to resolve the load. 
+* For *Custom LoadContext*, this override is an opportunity to load an assembly **before** the fallback (see below) to *Default LoadContext* is attempted to resolve the load. 
 
 * For *Default LoadContext*, this override always returns *null* since *Default Context* cannot override itself.
 
@@ -45,7 +48,7 @@ If the *Default LoadContext* fallback also did not resolve the load (or was not 
 
 * **System.Private.CoreLib.dll** is only loaded once, and into the **Default LoadContext**, during the .NET Core Runtime startup as it is a logical extension of the same. It cannot be loaded into **Custom LoadContext**.
 * Currently, custom **LoadContext** cannot be unloaded once created. This is a feature we are looking into for a future release.
-* If an attempt is made to load a R2R image from the same location in multiple load context's, then precompiled code can only be used from the first image that got loaded. The subsequent images will have their code JITted. This happens because subsequent loading binaries from the same location results in OS mapping them to the same memory as the previous one was mapped to and thus, could corrupt internal state information required for use precompiled code.
+* If an attempt is made to load a [Ready-To-Run (R2R)](https://github.com/dotnet/coreclr/blob/master/Documentation/botr/readytorun-overview.md) image from the same location in multiple load context's, then precompiled code can only be used from the first image that got loaded. The subsequent images will have their code JITted. This happens because subsequent loading binaries from the same location results in OS mapping them to the same memory as the previous one was mapped to and thus, could corrupt internal state information required for use precompiled code.
 
 ## Tests
 
@@ -59,15 +62,25 @@ Most of the **AssemblyLoadContext** [API surface](https://github.com/dotnet/core
 
 This property will return a reference to the *Default LoadContext*.
 
+### Load
+
+This method should be overriden in a *Custom LoadContext* if the intent is to override the assembly resolution that would be done during fallback to *Defaut LoadContext*
+
 ### LoadFromAssemblyName
 
 This method can be used to load an assembly into a load context different from the load context of the currently executing assembly.
 
+### Resolving
+
+This event is raised to give the last oppurtunity to a *LoadContext* instance to attempt to resolve a requested assembly that has neither been resolved by **Load** method, nor by fallback to **Default LoadContext**.
+
 ## Assembly Load APIs and LoadContext
 
-As part of .NET Standard 2.0 effort, certain assembly load APIs off the **AppDomain** type, which were present in Desktop .NET Framework, have been brought back. The following maps the APIs to the load context in which they will load the assembly:
+As part of .NET Standard 2.0 effort, certain assembly load APIs off the **Assembly** type, which were present in Desktop .NET Framework, have been brought back. The following maps the APIs to the load context in which they will load the assembly:
 
-* Assembly.Load - loads the assembly into the *Default LoadContext*
-* Assemby.LoadFrom - loads the assembly into the *Default LoadContext*
+* Assembly.Load - loads the assembly into the context of the assembly that triggers the load.
+* Assembly.LoadFrom - loads the assembly into the *Default LoadContext*
 * Assembly.LoadFile - creates a new (anonymous) load context to load the assembly into.
 * Assembly.Load(byte[]) - creates a new (anonymous) load context to load the assembly into.
+
+If you need to influence the load process or the load context in which assemblies are loaded, please look at the various Load* APIs exposed by **AssemblyLoadContext** [API surface](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Loader/ref/System.Runtime.Loader.cs).
--- a/Documentation/design-docs/assemblyloadcontext.md
+++ b/Documentation/design-docs/assemblyloadcontext.md
@@ -1,0 +1,63 @@
+**LoadContext** can be viewed as a container for assemblies, their code and data. Whenever an assembly is loaded, it is loaded within a load context - independent of whether the load was triggered explicitly (e.g. via *Assembly.Load*), implicitly (e.g. resolving static assembly references from the manifest) or dynamically (by emitting code on the fly).
+
+This concept is not new to .NET Core but has existed since the days of .NET Framework (see [this](https://blogs.msdn.microsoft.com/suzcook/2003/05/29/choosing-a-binding-context/) for details) where it operated behind the scenes and not exposed for the developer to interact with, aside from loading your assembly in one based upon the API used to perform the load.
+
+In .NET Core, we have exposed a [managed API surface](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Loader/ref/System.Runtime.Loader.cs) that developers can use to interact with it - to inspect loaded assemblies or create their own **LoadContext** instance. Here are some of the scenarios that motivated this work:
+
+* Ability to load multiple versions of the same assembly within a given process (e.g. for plugin frameworks)
+* Ability to load assemblies explicitly in a context isolated from that of the application.
+* Ability to override assemblies being resolved from application context.
+* Expose LoadContext as a first class concept for developers to interface with and not be a magic.
+
+## Types of LoadContext
+### Default LoadContext
+
+Every .NET Core app has a **LoadContext** instance created during .NET Core Runtime startup that we will refer to as the *Default LoadContext*. All application assemblies (including their transitive closure) are loaded within this **LoadContext** instance.
+
+### Custom LoadContext
+For scenarios that wish to have isolation between loaded assemblies, applications can create their own **LoadContext** instance by deriving from **System.Runtime.Loader.AssemblyLoadContext** type and loading the assemblies within that instance.
+
+
+## How Load is attempted
+
+### Basics
+If an assembly *A1* triggers the load of an assembly *C1*, the latter's load is attempted within the **LoadContext** instance of the former (which is also known as the *RequestingAssembly* or *ParentAssembly*). 
+
+Dynamically generates assemblies add a slight twist since they do not have a *ParentAssembly/RequestingAssembly* per-se. Thus, they are associated with the load context of their *Creator Assembly* and any subsequent loads (static or dynamic) will use that load context.
+
+### Resolution Process
+If the assembly was already present in *A1's* context, either because we had successfully loaded it earlier or because we failed to load it for some reason, we return the corresponding status (and assembly reference for the success case).
+However, if *C1* was not found in *A1's* context, the *Load* method override in *A1's* context is invoked. 
+
+* For *Custom LoadContext*, this override is an oppurtunity to load an assembly **before** the fallback (see below) to *Default LoadContext* is attempted to resolve the load. 
+
+* For *Default LoadContext*, this override always returns *null* since *Default Context* cannot override itself.
+
+If the *Load* method override does not resolve the load, fallback to *Default LoadContext* is attempted to resolve the load incase the assembly was already loaded there. If the operating context is *Default LoadContext*, there is no fallback attempted since it has nothing to fallback to.
+
+If the *Default LoadContext* fallback also did not resolve the load (or was not applicable), the *Resolving* event is invoked against *A1's* load context. This is the last oppurtunity to attempt to resolve the assembly load. If there are no subscribers for this event, or neither resolved the load, a *FileNotFoundException* is thrown.
+
+## PInvoke Resolution
+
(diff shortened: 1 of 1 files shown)


jkotas:
@agocke @davidfowl @gokarnm Could you please take a look and tell whether it has answers to everything you ever wanted to know about AssemblyLoadContext? Thanks!

Documentation/design-docs/assemblyloadcontext.md
```diff
@@ -0,0 +1,63 @@
...
+
+Most of the **LoadContex** [API surface](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Loader/ref/System.Runtime.Loader.cs) is self-explanatory. Key APIs/Properties, though, are described below:
+
+### Default
+
+This property will return a reference to the *Default LoadContext*.
+
+### LoadFromAssemblyName
+
+This method can be used to load an assembly into a load context different from the current one.
```
jkotas: I would say:

This method can be used to load an assembly using a load context different from the load context of currently executing assembly.

Documentation/design-docs/assemblyloadcontext.md
```diff
@@ -0,0 +1,63 @@
...
+* Currently, custom **LoadContext** cannot be unloaded once created. This is a feature we are looking into for a future release.
+* If an attempt is made to load a R2R image from the same location in multiple load context's, then precompiled code can only be used from the first image that got loaded. The subsequent images will have their code JITted. This happens because subsequent loading binaries from the same location results in OS mapping them to the same memory as the previous one was mapped to and thus, could corrupt internal state information required for use precompiled code.
+
+## Tests
+
+Tests are present [here](https://github.com/dotnet/corefx/tree/master/src/System.Runtime.Loader).
+
+## API Surface
+
+Most of the **LoadContex** [API surface](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Loader/ref/System.Runtime.Loader.cs) is self-explanatory. Key APIs/Properties, though, are described below:
```
jkotas: It would be useful to explain the relationship between the classic APIs `Assembly.Load`, `Assembly.LoadFrom`, `Assembly.LoadFile`, `Assembly.Load(byte[])` and AssemblyLoadContexts. What is the load context used by each of these APIs.

Documentation/design-docs/assemblyloadcontext.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Default
+
+This property will return a reference to the *Default LoadContext*.
```
jkotas: Explain where the list of the assemblies for the default load context comes from?

jkotas:
LGTM otherwise. Thank you for writing this!

Documentation/design-docs/assemblyloadcontext.md
gkhanna79: I mentioned its composition at the start of the writeup.

Documentation/design-docs/assemblyloadcontext.md
gkhanna79: Fixed.

Documentation/design-docs/assemblyloadcontext.md
gkhanna79: @jkotas Is this question in context of the APIs we have exposed for NS 2.0?

jkotas: Yes. @rahku may be the best person to fill it in since he has done the work on it.

Documentation/design-docs/assemblyloadcontext.md
```diff
@@ -0,0 +1,63 @@
...
+## How Load is attempted
+
+### Basics
+If an assembly *A1* triggers the load of an assembly *C1*, the latter's load is attempted within the **LoadContext** instance of the former (which is also known as the *RequestingAssembly* or *ParentAssembly*). 
+
+Dynamically generates assemblies add a slight twist since they do not have a *ParentAssembly/RequestingAssembly* per-se. Thus, they are associated with the load context of their *Creator Assembly* and any subsequent loads (static or dynamic) will use that load context.
+
+### Resolution Process
+If the assembly was already present in *A1's* context, either because we had successfully loaded it earlier or because we failed to load it for some reason, we return the corresponding status (and assembly reference for the success case).
+However, if *C1* was not found in *A1's* context, the *Load* method override in *A1's* context is invoked.
```
agocke: Maybe it should be noted whether or not `mscorlib` is an exception to this? My understanding is that mscorlib can never be reloaded.

Documentation/design-docs/assemblyloadcontext.md
```diff
@@ -0,0 +1,63 @@
...
+
+## PInvoke Resolution
+
+*Custom LoadContext* can override the **AssemblyLoadContext.LoadUnmanagedDll** method to intercept PInvokes from within the **LoadContext** instance so that can be resolved from custom binaries. If not overridden, or if the resolution is not able to resolve the PInvoke, the default PInvoke mechanism will be used as fallback.
+
+## Constraints
+
+* Currently, custom **LoadContext** cannot be unloaded once created. This is a feature we are looking into for a future release.
+* If an attempt is made to load a R2R image from the same location in multiple load context's, then precompiled code can only be used from the first image that got loaded. The subsequent images will have their code JITted. This happens because subsequent loading binaries from the same location results in OS mapping them to the same memory as the previous one was mapped to and thus, could corrupt internal state information required for use precompiled code.
```
agocke: What about if the custom load context requests the load from another load context? e.g.,

```csharp
protected override Assembly Load(AssemblyName name) => Default.LoadFromAssemblyName(name);
```

Documentation/design-docs/assemblyloadcontext.md
agocke: Also, an explanation for why no `Assembly.Load(byte[])` but an `Assembly.Load(Stream)` would be useful.

I ended up wrapping an `ImmutableArray<byte>` in a `MemoryStream` but it seems a little weird to have to do that.

Documentation/design-docs/assemblyloadcontext.md
agocke: My observation is that doing this allows types (unification and static values) to persist if this is done, so that may be an interesting thing to call out.

gkhanna79: If LC_A attempts to resolve from LC_B, which already has the assembly loaded, then there is no second load attempted since the assembly is already found in LC_B (See the process section). This is really about reloading the assembly within a different context.

jkotas: > My observation is that doing this allows types (unification and static values)

Yes, you can build a graph (typically a tree) between different AssemblyLoadContext to control what gets unified.

Documentation/design-docs/assemblyloadcontext.md
jkotas: Stream is more flexible/efficient. `byte[]` can be wrapped in MemoryStream easily, but the other direction cannot be done efficiently.

`Load(byte[])` requires the assembly to be stored in the exact right sized byte array. If you have the assembly in something else, you have to pay the cost of allocating (potentially large) `byte[]`.

gkhanna79: > Yes. @rahku may be the best person to fill it in since he has done the work on it.

Had chatted with @rahku about this. Document is now updated to reflect this.

Documentation/design-docs/assemblyloadcontext.md
gkhanna79: Done.

Documentation/design-docs/assemblyloadcontext.md
```diff
@@ -0,0 +1,73 @@
...
+
+* For *Custom LoadContext*, this override is an oppurtunity to load an assembly **before** the fallback (see below) to *Default LoadContext* is attempted to resolve the load. 
+
+* For *Default LoadContext*, this override always returns *null* since *Default Context* cannot override itself.
+
+If the *Load* method override does not resolve the load, fallback to *Default LoadContext* is attempted to resolve the load incase the assembly was already loaded there. If the operating context is *Default LoadContext*, there is no fallback attempted since it has nothing to fallback to.
+
+If the *Default LoadContext* fallback also did not resolve the load (or was not applicable), the *Resolving* event is invoked against *A1's* load context. This is the last oppurtunity to attempt to resolve the assembly load. If there are no subscribers for this event, or neither resolved the load, a *FileNotFoundException* is thrown.
+
+## PInvoke Resolution
```
davidfowl: Should we add something about resource assemblies?

agocke:
One more request: could we get some info about the use of `SetProfileOptimizationRoot`?

Documentation/design-docs/assemblyloadcontext.md
```diff
@@ -0,0 +1,73 @@
...
+### Custom LoadContext
+For scenarios that wish to have isolation between loaded assemblies, applications can create their own **LoadContext** instance by deriving from **System.Runtime.Loader.AssemblyLoadContext** type and loading the assemblies within that instance.
+
+
+## How Load is attempted
+
+### Basics
+If an assembly *A1* triggers the load of an assembly *C1*, the latter's load is attempted within the **LoadContext** instance of the former (which is also known as the *RequestingAssembly* or *ParentAssembly*). 
+
+Dynamically generates assemblies add a slight twist since they do not have a *ParentAssembly/RequestingAssembly* per-se. Thus, they are associated with the load context of their *Creator Assembly* and any subsequent loads (static or dynamic) will use that load context.
```
gokarnm: Typo _Dynamically generates assemblies_ => _Dynamically generated assemblies_

Documentation/design-docs/assemblyloadcontext.md
```diff
@@ -0,0 +1,73 @@
...
+
+### LoadFromAssemblyName
+
+This method can be used to load an assembly into a load context different from the load context of the currently executing assembly.
+
+## Assembly Load APIs and LoadContext
+
+As part of .NET Standard 2.0 effort, certain assembly load APIs off the **AppDomain** type, which were present in Desktop .NET Framework, have been brought back. The following maps the APIs to the load context in which they will load the assembly:
+
+* Assembly.Load - loads the assembly into the *Default LoadContext*
```
gokarnm: I'm a bit concerned with how the `Assembly.Load` and `Assembly.LoadFrom` seem to be "custom ALC unfriendly", similar to `AssemblyLoadContext.Default`, as a developer can use it to write code that works fine under default ALC (the common case) but potentially break when the same code is run under a custom ALC.

As someone who is developing a plugin framework leveraging the ALC feature, I'm looking at how third party framework developers can be directed towards the "pit of success" where the code they write works seamlessly under a custom ALC.

Maybe having a "Best Practices" section about the writing ALC friendly code, or/and the API documentation for these three APIs mentioning the pitfall of using default context would be helpful. Something on the lines of _"This API always loads the assembly in the default Load Context. Consider using this [alternate API] if your code is not tied to the default load context and can run in any load context"_.

Some inputs for the proposed "Best Practices" section

1. When to override Load method vs. implement Resolve event handler for a custom ALC based on @gkhanna79 advice on this [thread](https://github.com/dotnet/coreclr/issues/8398#issuecomment-264235647).

1. `AssemblyLoadContext.Default` returns the reference to the default load context, use `AssemblyLoadContext.GetLoadContext(Assembly.GetExecutingAssembly())` if you want a reference to the current executing assembly's load context.

1. `Assembly.Load` and `Assembly.LoadFrom` always load the assembly into the default load context. If your code is not tied to the default load context and can be run in any load context, consider using `AssemblyLoadContext.Load` or `AssemblyLoadContext.LoadFromAssemblyPath` methods on the executing assembly's load context. You can get the executing assembly's load context using `AssemblyLoadContext.GetLoadContext(Assembly.GetExecutingAssembly())`.

davidfowl: I thought Assembly.Load used the "current" load context. If it doesn't how would any of this work for plugins?

Documentation/design-docs/assemblyloadcontext.md
```diff
@@ -0,0 +1,73 @@
...
+### Basics
+If an assembly *A1* triggers the load of an assembly *C1*, the latter's load is attempted within the **LoadContext** instance of the former (which is also known as the *RequestingAssembly* or *ParentAssembly*). 
+
+Dynamically generates assemblies add a slight twist since they do not have a *ParentAssembly/RequestingAssembly* per-se. Thus, they are associated with the load context of their *Creator Assembly* and any subsequent loads (static or dynamic) will use that load context.
+
+### Resolution Process
+If the assembly was already present in *A1's* context, either because we had successfully loaded it earlier or because we failed to load it for some reason, we return the corresponding status (and assembly reference for the success case).
+However, if *C1* was not found in *A1's* context, the *Load* method override in *A1's* context is invoked. 
+
+* For *Custom LoadContext*, this override is an oppurtunity to load an assembly **before** the fallback (see below) to *Default LoadContext* is attempted to resolve the load.
```
natemcmaster: Doesn't resolution first check the default load context? Observed behavior is that Load on *A1*'s context is not called if the default load context already contains an assembly identity matching *C1*. https://github.com/dotnet/coreclr/issues/6187

Also, spelling: oppurtunity => opportunity

Documentation/design-docs/assemblyloadcontext.md
```diff
@@ -0,0 +1,73 @@
...
+## How Load is attempted
+
+### Basics
+If an assembly *A1* triggers the load of an assembly *C1*, the latter's load is attempted within the **LoadContext** instance of the former (which is also known as the *RequestingAssembly* or *ParentAssembly*). 
+
+Dynamically generates assemblies add a slight twist since they do not have a *ParentAssembly/RequestingAssembly* per-se. Thus, they are associated with the load context of their *Creator Assembly* and any subsequent loads (static or dynamic) will use that load context.
+
+### Resolution Process
+If the assembly was already present in *A1's* context, either because we had successfully loaded it earlier or because we failed to load it for some reason, we return the corresponding status (and assembly reference for the success case).
+However, if *C1* was not found in *A1's* context, the *Load* method override in *A1's* context is invoked.
```
natemcmaster: Doesn't resolution first check the default load context? Observed behavior is that Load on *A1*'s context is not called if the default load context already contains an assembly identity matching *C1*. Doc bug or implementation bug? cref https://github.com/dotnet/coreclr/issues/6187

Documentation/design-docs/assemblyloadcontext.md
```diff
@@ -0,0 +1,73 @@
...
+
+## PInvoke Resolution
+
+*Custom LoadContext* can override the **AssemblyLoadContext.LoadUnmanagedDll** method to intercept PInvokes from within the **LoadContext** instance so that can be resolved from custom binaries. If not overridden, or if the resolution is not able to resolve the PInvoke, the default PInvoke mechanism will be used as fallback.
+
+## Constraints
+
+* **System.Private.CoreLib.dll** is only loaded once, and into the **Default LoadContext**, during the .NET Core Runtime startup as it is a logical extension of the same. It cannot be loaded into **Custom LoadContext**.
+* Currently, custom **LoadContext** cannot be unloaded once created. This is a feature we are looking into for a future release.
+* If an attempt is made to load a R2R image from the same location in multiple load context's, then precompiled code can only be used from the first image that got loaded. The subsequent images will have their code JITted. This happens because subsequent loading binaries from the same location results in OS mapping them to the same memory as the previous one was mapped to and thus, could corrupt internal state information required for use precompiled code.
```
natemcmaster: Can you define "R2R image" in this document?

Documentation/design-docs/assemblyloadcontext.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Tests
+
+Tests are present [here](https://github.com/dotnet/corefx/tree/master/src/System.Runtime.Loader).
+
+## API Surface
+
+Most of the **AssemblyLoadContext** [API surface](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Loader/ref/System.Runtime.Loader.cs) is self-explanatory. Key APIs/Properties, though, are described below:
```
natemcmaster: > Most of the AssemblyLoadContext API surface is self-explanatory.

This sentence is meaningless. I suggest removing it.

Documentation/design-docs/assemblyloadcontext.md
gkhanna79: What specific thing about resource assemblies are you looking for @davidfowl ?

Documentation/design-docs/assemblyloadcontext.md
gkhanna79: Mentioned at https://github.com/dotnet/coreclr/pull/10271/files#diff-5adec7752d04e4f6a6b40f455af2ee55R29

natemcmaster: > If the assembly was already present in *A1's* context, either because we had successfully loaded it earlier or because we failed to load it for some reason, we return the corresponding status (and assembly reference for the success case).

I read that sentence differently then. I understood "successfully loaded" to mean "successfully loaded _in A1's context._" 

So you're saying it means "successfully loaded _in A1's load context or the default context_", or is it "successfully loaded _in any load context_"?

Documentation/design-docs/assemblyloadcontext.md
gkhanna79: Done.

Documentation/design-docs/assemblyloadcontext.md
gkhanna79: This is present to explain why this document is not explaining functionality of all the APIs.

Documentation/design-docs/assemblyloadcontext.md
gkhanna79: This sentence is in the context of A1's LC.

gkhanna79:
Needless to say, when I write when I am half-asleep, I miss on few things. Now that I am wide-awake, I have pushed an update that I hope addresses most of the comments (for the ones not addressed, I have put up questions for the comment author).

Documentation/design-docs/assemblyloadcontext.md
```diff
@@ -0,0 +1,73 @@
...
+* For *Default LoadContext*, this override always returns *null* since *Default Context* cannot override itself.
+
+If the *Load* method override does not resolve the load, fallback to *Default LoadContext* is attempted to resolve the load incase the assembly was already loaded there. If the operating context is *Default LoadContext*, there is no fallback attempted since it has nothing to fallback to.
+
+If the *Default LoadContext* fallback also did not resolve the load (or was not applicable), the *Resolving* event is invoked against *A1's* load context. This is the last oppurtunity to attempt to resolve the assembly load. If there are no subscribers for this event, or neither resolved the load, a *FileNotFoundException* is thrown.
+
+## PInvoke Resolution
+
+*Custom LoadContext* can override the **AssemblyLoadContext.LoadUnmanagedDll** method to intercept PInvokes from within the **LoadContext** instance so that can be resolved from custom binaries. If not overridden, or if the resolution is not able to resolve the PInvoke, the default PInvoke mechanism will be used as fallback.
+
```
gkhanna79: Done.

gkhanna79:
> One more request: could we get some info about the use of SetProfileOptimizationRoot?

I will merge this PR and have @brianrob add context around it.

Documentation/design-docs/assemblyloadcontext.md
gokarnm: The latest version of the document says that `Assembly.Load` loads the assembly into the context of the assembly that triggers the load, which addresses my concern and seems to be the correct behavior.
What about `Assembly.Assembly.LoadFrom` though? Does it load into the default context or the "current" context?

gkhanna79:
LoadFrom loads into the default context.



Thanks,
Gaurav


-------- Original message --------

gokarnm:
Hi @gkhanna79, I'm curious about why `Assembly.Load` and `Assembly.LoadFrom' have different behavior in terms of the load context they load into, could you provide more insight? If `Assembly.LoadFrom' uses default context, how would you use it with plugins, it could cause potential issues when used from code running in a custom ALC?

gkhanna79:
@gokarnm Assembly.Load* APIs have been (re)introduced with respect to the Desktop .NET Framework.

- Assembly.Load, in .NET Core, works exactly how it works on Desktop.
- Assembly.LoadFrom, on Desktop, worked differently based upon certain heuristics that gave inconsistent/non-deterministic results. Sometimes, based upon rules, it would load in Default Context and at other times, it would load in its own _LoadFrom_ context. When we had enabled this for our internal infrastructure, when bringing up Windows Phone, we realized that simply loading in the default context didn't actually break anyone and worked as expected. Thus, for .NET Core, we chose to adopt the same semantic, making it more predictable.

Plugins written to use these APIs should expect to behave the same way as they did on the Desktop (afterall, these APIs are introduced from Desktop Compat standpoint as well). If someone wishes to load in a particular context, then they should use ALC APIs.

gokarnm:
@gkhanna79 , thanks for the details!

In Desktop .NET Framework, AppDomain was commonly used for implementing plugin frameworks and developers would have relatively less knowledge of load contexts. I would appreciate if the API documentation for `Assembly.LoadFrom` had a note, something like - 
_"This API always loads the assembly into the default load context. To load your assemblies in a particular context, use APIs on the AssemblyLoadContext type."_ 

That would partially address my concern that developers would use `Assembly.LoadFrom` as they are familiar with it from .NET Framework, but it may not be the right API in some scenarios, causing their code to break when run under a custom load context.

gkhanna79:
> I would appreciate if the API documentation for Assembly.LoadFrom had a note, something like

This is called out at the very end of the section of the Assembly Load APIs.

gokarnm:
>This is called out at the very end of the section of the Assembly Load APIs.

I noticed that, my thinking was - developers may not even know about ALC and read this document, or the ALC API reference. If the [API reference](https://docs.microsoft.com/en-us/dotnet/core/api/system.reflection.assembly) for `Assembly.LoadFrom` had a similar note, it would be more discoverable. Maybe some of this content is planned to flow into API reference? Sorry for my ignorance.

gkhanna79:
@danmosemsft Can the above be tracked as part of documentation update for APIs exposed out of NS 2.0?

danmosemsft:
@karelz is tracking documentation...

karelz:
Please file a bug - either here in CoreFX, or in [docs repo](https://github.com/dotnet/docs/issues) (BTW: @mairaw which one do you prefer?)

Ideally: Link the docs page. Suggest changes (mention text before and after).
Once we review it for correctness, we will ask docs team to make the changes.

mairaw:
No preference but https://github.com/dotnet/docs seems more appropriate for doc issues.

ro-jo:
When I search Bing or Google for AssemblyLoadContext documentation, it leads me to this page: #5463 

I read through the whole post looking for a link to some documentation, maybe I missed it.  Then I found an unqualified link to this discussion.

I haven't got time to read every word in this discussion or other discussions: where is the documentation for AssemblyLoadContext please?

Is this it, surely not?: https://docs.microsoft.com/en-gb/dotnet/api/system.runtime.loader.assemblyloadcontext?view=netcore-2.0

BrainCrumbz:
For future ref., this should be the "dedicated" doc page you were looking for: https://github.com/dotnet/coreclr/blob/master/Documentation/design-docs/assemblyloadcontext.md

PR #10273, opened 2017-03-17
Remove JIT-EE version ifdefs

BruceForstall:
Currently, we maintain two versions of the JIT-EE interface in the
source tree: the current version, and .NET 4.6 version. This is
managed by the `COR_JIT_EE_VERSION` define. Any new changes to the JIT-EE
interface must be put under `#if COR_JIT_EE_VERSION > 460`. As of now,
there are quite a few of these `#if` statements in the tree.

We started doing this so we could build and ship a CTP (Community
Technology Preview) version of the JIT from the current source tree that
runs against earlier versions of the runtime that are widely available
to customers, so those customers can try out the new JIT and give us
early feedback. This was done during the RyuJIT/x64 bring-up before
shipping .NET 4.6, to get as much feedback as possible before replacing
JIT64. After 4.6 shipped, we re-based the old version of the JIT-EE
interface to the 4.6 version, and started adding the above `#if`
conditions, with the thought that if we ever shipped another CTP, it
would be expected to run on top of 4.6. We have not shipped such a
thing, and we currently have no plans to build another CTP in the same
way we did before. (Early adopters can try out the "live" JIT in the
.NET Core open source project, although that JIT can't be used on the
desktop.)

So, to simplify the code, and reduce the cost of adding to or changing
the JIT-EE interface, we are removing the `COR_JIT_EE_VERSION` define.


BruceForstall:
@dotnet/jit-contrib PTAL
cc @jkotas

PR #10274, opened 2017-03-18
Fix yet another casing issue.

AndyAyersMS:
This might be what is causing the TFS mirror to have hiccups.


AndyAyersMS:
@ellismg PTAL
cc @dotnet/jit-contrib

ellismg:
LGTM.

PR #10275, opened 2017-03-18
Lower long compares that are materialized into a register

### LINKED ISSUE #7210: RyuJIT x86: lower long compares that are materialized into a register
This is an optimization to decrease register pressure. See #7189 for more details.

mikedn:
Fixes #7210 

Rather than attempting to emulate the current codegen implementation in lower this implementation basically does a 64 bit subtraction and converts its result to 0/1 using a single `SETcc` instruction and no conditional branches. A typical signed compare ends up generating 3 instructions (`CMP`, `SBB`, `SETcc`) rather than 6 that the current implementation generates. Compares involving constants may sometimes require loading the constants in registers but the current implementation doesn't handle this any better.

I added 2 new opers - `GT_CMP` and `GT_SETCC`. `GT_CMP` isn't strictly required but without it we generate `SUB`/`SBB` rather than `CMP`/`SBB` and the first `SUB` wastes a register. `GT_SETCC` is required at least for unsigned compares because we need to check the carry flag and not the actual subtraction result (that could be done with a normal relop). IMO it would best to lower all relops and `GT_JTRUE` to `GT_CMP`, `GT_SETCC` and `GT_JCC` but that's another story.


mikedn:
[Failure](https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_windows_nt_prtest/4287/) of `JIT_Generics._Coverage_chaos65204782cs_o_chaos65204782cs_o_._Coverage_chaos65204782cs_o_chaos65204782cs_o_cmd` seems unrelated since this doesn't affect x64.

mikedn:
jit-diff summary:
```
Total bytes of diff: -1752 (-0.02 % of base)
    diff is an improvement.
        -455 : System.Private.CoreLib.dasm (-0.02 % of base)
        -253 : System.Private.Uri.dasm (-0.32 % of base)
        -173 : System.Reflection.Metadata.dasm (-0.06 % of base)
        -162 : Microsoft.CodeAnalysis.VisualBasic.dasm (-0.01 % of base)
        -152 : Microsoft.CodeAnalysis.CSharp.dasm (-0.01 % of base)
18 total files with size differences (18 improved, 0 regressed).
          12 : System.Private.CoreLib.dasm - DateTimeFormat:FormatCustomizedTimeZone(struct,struct,ref,int,bool,ref)
           5 : System.Linq.Expressions.dasm - EqualInt64LiftedToNull:Run(ref):int:this
           5 : System.Linq.Expressions.dasm - EqualUInt64LiftedToNull:Run(ref):int:this
           5 : System.Linq.Expressions.dasm - NotEqualInt64LiftedToNull:Run(ref):int:this
           5 : System.Linq.Expressions.dasm - NotEqualUInt64LiftedToNull:Run(ref):int:this
        -156 : System.Reflection.Metadata.dasm - MetadataReader:InitializeTableReaders(struct,ubyte,ref,ref):this
         -70 : Microsoft.CSharp.dasm - ExpressionBinder:FoldConstI8Op(int,ref,ref,ref,ref,int):ref:this
         -63 : System.Private.Uri.dasm - Uri:CheckAuthorityHelper(int,char,char,byref,byref,ref,byref):char:this
         -60 : Microsoft.CodeAnalysis.CSharp.dasm - Binder:FoldNeverOverflowBinaryOperators(int,ref,ref):ref
         -44 : Microsoft.CodeAnalysis.VisualBasic.dasm - ExpressionEvaluator:PerformCompileTimeBinaryOperation(char,byte,ref,ref,ref):ref
144 total methods with size differences (138 improved, 6 regressed).
```

mikedn:
Sample generated code: https://gist.github.com/mikedn/75d4356aa5b320eea338234864e9de70

danmosemsft:
@dotnet/jit-contrib can someon eplease review?

BruceForstall:
@dotnet-bot test this please

mikedn:
This should wait for the reverse-ops removal if that PR is to be merged.

src/jit/gentree.h
```diff
@@ -4949,18 +4949,18 @@ struct GenTreeAllocObj final : public GenTreeUnOp
 #endif
 };
 
-struct GenTreeJumpCC final : public GenTree
+struct GenTreeCC final : public GenTree
 {
```
CarolEidt: It would be useful to have a comment here describing what opcodes this is for.

src/jit/lower.cpp
```diff
@@ -1998,11 +1999,144 @@ GenTree* Lowering::LowerTailCallViaHelper(GenTreeCall* call, GenTree* callTarget
 void Lowering::LowerCompare(GenTree* cmp)
 {
 #ifndef _TARGET_64BIT_
-    LIR::Use cmpUse;
-
-    if ((cmp->gtGetOp1()->TypeGet() == TYP_LONG) && BlockRange().TryGetUse(cmp, &cmpUse) &&
-        cmpUse.User()->OperIs(GT_JTRUE))
+    if (cmp->gtGetOp1()->TypeGet() == TYP_LONG)
     {
+// TODO-ARM: This code should be enabled for ARM32 once support for GT_CMP and GT_SETCC is added.
```
CarolEidt: This needs a `CLANG_FORMAT_COMMENT_ANCHOR`

src/jit/lower.cpp
```diff
@@ -1998,11 +1999,144 @@ GenTree* Lowering::LowerTailCallViaHelper(GenTreeCall* call, GenTree* callTarget
...
+#if _TARGET_X86_
+        LIR::Use cmpUse;
+        if (!BlockRange().TryGetUse(cmp, &cmpUse) || !cmpUse.User()->OperIs(GT_JTRUE))
+        {
```
CarolEidt: This could use a comment describing the condition you're testing for (e.g. "Is this a compare which produces a value or just sets the condition code?")

src/jit/gtlist.h
```diff
@@ -155,6 +155,7 @@ GTNODE(GT               , ">"            ,GenTreeOp          ,0,GTK_BINOP|GTK_RE
 // Because of this there is no need to also add GT_TEST_LT/LE/GE/GT opers.
 GTNODE(TEST_EQ          , "testEQ"       ,GenTreeOp          ,0,GTK_BINOP|GTK_RELOP)
 GTNODE(TEST_NE          , "testNE"       ,GenTreeOp          ,0,GTK_BINOP|GTK_RELOP)
+GTNODE(CMP              , "cmp"          ,GenTreeOp          ,0,GTK_BINOP|GTK_NOVALUE)
```
CarolEidt: Why is this not `GTK_RELOP`? I suspect there's some reason, but it's extremely confusing that `OperIsCompare()` returns false for `GT_CMP`.

CarolEidt:
:clock1:

src/jit/lower.cpp
mikedn: I'll add something referencing the `#if 0` below. Once that `#if` is removed this is no longer needed.

src/jit/lower.cpp
mikedn: Hmm, the comment applies of the #if below it, I don't see why should it be indented.

src/jit/gtlist.h
mikedn: My main concern with making `cmp` a relop is that some code may try to reverse it but that's not possible since there's no condition associated with it. There's also the fact that unlike other relops `cmp` does not return a value and thus it needs some special casing (e.g. `dstCount` setting in `TreeNodeInfoInitCmp`).

We do end up with some code like `else if (tree->OperIsCompare() || tree->OperIs(GT_CMP))` that perhaps looks a bit strange. As stated in the initial PR comment it would be better to lower all relops to avoid this kind of duplication. See #10861 for how that may look.

src/jit/lower.cpp
```diff
@@ -1998,11 +1999,144 @@ GenTree* Lowering::LowerTailCallViaHelper(GenTreeCall* call, GenTree* callTarget
...
+
+                //
+                // If the compare is signed then (x LT|GE y) can be transformed into ((x SUB y) LT|GE 0).
+                // If the compare is unsigned we can still use SUB but we need to check the Carry flag,
+                // not the actual result. In both cases we can simply check the appropiate condition flags
+                // and ignore the actual result:
+                //     SUB_LO loSrc1, loSrc2
+                //     SUB_HI hiSrc1, hiSrc2
+                //     SETCC|JCC (signed|unsigned LT|GE)
+                // If loSrc1 happens to be 0 then the first SUB can be eliminated and the second one can
```
mikedn: This comment is incorrect, it should say `If loSrc2 happens to be 0`.

src/jit/gtlist.h
mikedn: I've moved `GT_CMP` to a separate section, together with `GT_SETCC` and `GT_JCC` as they are related and LIR specific. I also added comment explaining their use of condition flags.

pgavlin:
@mikedn `GTF_REVERSE_OPS` has now been removed from LIR. Is this PR ready for review, or do you have pending updates?

mikedn:
@pgavlin It's ready, I already removed reverse ops stuff from it.

BruceForstall:
@dotnet-bot test this please

src/jit/lower.cpp
```diff
@@ -1990,6 +1990,7 @@ GenTree* Lowering::LowerTailCallViaHelper(GenTreeCall* call, GenTree* callTarget
 //
 // Notes:
 //    - Decomposes long comparisons that feed a GT_JTRUE (32 bit specific).
+//    - Decomposes long comparisons that produce a value (X86 specific).
```
pgavlin: Is there a particular reason that we decompose these operators here rather than in the decomposition proper?

mikedn: Decomposition takes care of nodes that produce a long value, comparisons only consume long values.

I think there's a more general question: do we still need decomposition or all it does can now be done during lowering? It's possible that the current approach is influenced by representation issues that precede LIR.

pgavlin: > Decomposition takes care of nodes that produce a long value, comparisons only consume long values

Yes, but those long values are necessarily decomposed. There is precedent for checking the user of a freshly-decomposed node and then decomposing the user as well; this is how we handle `GT_LONG` nodes that are part of field list arguments.

> I think there's a more general question: do we still need decomposition or all it does can now be done during lowering?

Yeah, this is a question that we need to think through. Off the top of my head it seems like we should at least be able to do
```
foreach block:
    foreach node:
        range newNodes = decomposeNode(node)
        lowerRange(newNodes)
```
rather than today's
```
foreach block:
    decompose_block()
    lower_block()
```

mikedn: > There is precedent for checking the user of a freshly-decomposed node and then decomposing the user as well; this is how we handle GT_LONG nodes that are part of field list arguments.

That sounds like a special case and not something that decomposition usually does.

I'd love to actually move all this code to decomposition but I don't see a natural way of doing it, short of simply extracting the code into a separate function and moving the function to `DecomposeLongs`.

> rather than today's

I think that we actually used the first version in the early days of decomposition. I switched to the second version to fix a bug. If memory serves me well the bug was related to embedded (shudder) trees so the current approach version may be obsolete.

src/jit/lower.cpp
```diff
@@ -1998,11 +1999,140 @@ GenTree* Lowering::LowerTailCallViaHelper(GenTreeCall* call, GenTree* callTarget
...
+                //     SUB_HI hiSrc1, hiSrc2
+                //     SETCC|JCC (signed|unsigned LT|GE)
+                // If loSrc2 happens to be 0 then the first SUB can be eliminated and the second one can
+                // be turned into a CMP because the first SUB would have set carry to 0. This effectively
+                // transforms a long compare against 0 into an int compare of the high part against 0.
+                //
+                // (x LE|GT y) can to be transformed into ((x SUB y) LE|GT 0) but checking that a long value
+                // is greater than 0 is not so easy. We need to turn this into a positive/negative check
+                // like the one we get for LT|GE compares, this can be achieved by swapping the compare:
+                //     (x LE|GT y) becomes (y GE|LT x)
```
pgavlin: Should this be `(x LE|GT y) becomes (y GT|LE x)`?

pgavlin: Ah, nvm. Misread it.

CarolEidt:
>         return;

Perhaps I'm missing something here, but is this correct? We have only compared `hiOp1` and `hiOp2`.

---
Refers to: src/jit/codegenxarch.cpp:5948 in 042d48f. [](commit_id = 042d48f56ab73dd6644963a53cf69e2f79ebb422, deletion_comment = True)

CarolEidt:
>     assert(varTypeIsLong(op2->TypeGet()));

It would be prudent to add asserts that `op1` and `op2` are `GT_LONG`.

---
Refers to: src/jit/codegenxarch.cpp:5926 in 042d48f. [](commit_id = 042d48f56ab73dd6644963a53cf69e2f79ebb422, deletion_comment = True)

mikedn:
@CarolEidt 

> Perhaps I'm missing something here, but is this correct? We have only compared hiOp1 and hiOp2.

Where is that? It's not clear what line of code your code refers to.

CarolEidt:
Ignore my last two comments - for some reason I found myself looking at an old commit.

BruceForstall:
@mikedn Can you fix the conflict?

src/jit/codegenxarch.cpp
```diff
@@ -6357,6 +6093,34 @@ void CodeGen::genSetRegToCond(regNumber dstReg, GenTreePtr tree)
     }
 }
 
+//------------------------------------------------------------------------
+// genCodeForShift: Generates a setcc instruction for a GT_SETCC node.
```
briansull: This method is named genCodeForSetcc, not genCodeForShift

mikedn:
Fixed conflicts and incorrect function comment

PR #10284, opened 2017-03-18
hoisted the accounting for LOH into allocate_large_objects to avoid…

Maoni0:
… missing counting in certain stages during bgc when we go through bgc_loh_alloc_clr instead of adjust_limit_clr


Maoni0:
@jkotas @rahku @swgillespie this should fix https://github.com/dotnet/corefx/pull/17269

jkotas:
Thanks!

Issue #6056, opened 2016-06-29
Enable Arm64 build in build pipeline


gkhanna79:
With https://github.com/dotnet/coreclr/issues/6000, we now generate Arm64 nuget packages for Windows. We should, now, build it in the pipeline.

@wtgodbe Can you PTAL? This should be similar to the x64 build definitions you have. For now please *do not* enable signing and publishing of the artifacts - just build.

gkhanna79:
CC @weshaggard

wtgodbe:
Do you expect us to ever want to sign binaries for an Arm64 build? If not, I can add logic to sign.builds to only sign binaries if the architecture is not Arm64, and we can reuse the same build definition that we use for x64. If not, I'll probably have to create another build definition (like the one we have for x86).

weshaggard:
We will likely want to sign them at some point but I think it is better to disable them in the sign.builds file for now as opposed to creating new build definitions.

wtgodbe:
@jhendrixMSFT what's the best way to apply this condition to sign.builds (only sign non-arm64 binaries)?

jhendrixMSFT:
You should condition the FilesToSign item groups and targets GetFilesToSignItems and CleanUpMarkerFiles to be executed when BuildArch != arm64.

wtgodbe:
I just queued an orchestrated build that should build arm64 Windows alongside x64. I'll update once the build completes.

wtgodbe:
@gkhanna79 @weshaggard @jhendrixMSFT it looks like arm64 Windows builds require a toolset_dir to be passed in via the command line - do any of you guys have an example of what this looks like? Without it, the arm64 build fails.

wtgodbe:
I see in netci.groovy we do "build.cmd /toolset_dir C:\\ats2' - can we assume the same thing will work on our build machines?

jhendrixMSFT:
@wtgodbe Seems unlikely as this directory doesn't exist on the build machines.  What is it supposed to contain?

wtgodbe:
I don't actually know, the documentation in build.cmd is very vague for it.

"arm64 builds currently use private toolset which has not been released yet"

"A toolset directory is required for the Arm64 Windows build. Use the toolset_dir argument."

@rahku do you know more about this?

wtgodbe:
@jashook do you know anything about the 'toolset_dir' arg for arm64 windows builds? What is the directory supposed to contain? Where does it come from?

[...]

PR #6000, opened 2016-06-25
Generate nuget packages for Arm64 build

gkhanna79:
@weshaggard @ericstj @pgavlin @rahku  PTAL

CC @haripulapaka

PR #10286, opened 2017-03-19
First steps for generating nuget packages for ARM64/Android.

cydhaselton:
Adds RIDS and related conditionals/variables to the appropriate files as part of Enabling Android/ARM64 for .NET Core ([issue here](https://github.com/dotnet/core-setup/issues/1651)).

Code appears to have been refactored since the original PR to which I was referred, cc-ing @janvorli and @qmfrederik for review in case of missed items


cydhaselton:
Also tagging @gkhanna79 for review because of #6056 and #6000

src/.nuget/Microsoft.NETCore.Native/Microsoft.NETCore.Native.props
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <SupportedPackageOSGroups>Linux;OSX</SupportedPackageOSGroups>
+    <SupportedPackageOSGroups>Linux;Android;OSX</SupportedPackageOSGroups>
```
gkhanna79: Is there a reason you are enabling this here?

CC @jkotas

src/.nuget/dir.props
```diff
@@ -23,7 +23,7 @@
 
     <RuntimeOS Condition="'$(RuntimeOS)' == ''">$(OSRid)</RuntimeOS>
 
-    <SupportedPackageOSGroups Condition="'$(SupportedPackageOSGroups)' == ''">Windows_NT;OSX;Linux</SupportedPackageOSGroups>
+    <SupportedPackageOSGroups Condition="'$(SupportedPackageOSGroups)' == ''">Windows_NT;OSX;Android;Linux</SupportedPackageOSGroups>
```
gkhanna79: What is the __DistroRid for Android build?

src/.nuget/Microsoft.NETCore.Native/Microsoft.NETCore.Native.props
cydhaselton: Because Linux and OSX were enabled here. If it's not applicable I can remove; I'm basically working from PRs and extrapolating based on refactored code so some of the edits may not be needed. Feel free to let me know and I'll make changes

src/.nuget/dir.props
cydhaselton: android.21-arm64 and/or android.21-arm, unless there's a difference between RID as registered in CoreFX and __DistroRid?

cydhaselton:
@jkotas and/or @gkhanna79:  Is there a definitive list of files to which a new RID should be added and/or detected for nuget package support?

src/.nuget/Microsoft.NETCore.Native/Microsoft.NETCore.Native.props
gkhanna79: I think you can skip this file from the change.

src/.nuget/dir.props
gkhanna79: Was this updated?

cydhaselton: Not yet; still working out where __DistroRid originates and/or is generated (also see question below)

cydhaselton: @gkhanna79 The test hack I added to build.sh generates __DistroRid for Android build and, with following commits, allows package build for Android.

I've added a TODO for figuring out a reliable way to generate __DistroRid for Android. Since I generated the commit changes from the `runtime.OS._filename_.props files and guesswork, a once-over to make sure I haven't missed anything would be appreciated

cydhaselton:
@gkhanna79 The test hack I added to build.sh generates __DistroRid for Android build and, with following commits, allows package build for Android.

I've added a TODO for figuring out a reliable way to generate __DistroRid for Android. Since I generated the commit changes from the `runtime.OS._filename_.props files and guesswork, a once-over to make sure I haven't missed anything would be appreciated

cydhaselton:
Copying @mellinoe for additional visibility into package generation

cydhaselton:
cc-ing @qmfrederik @mellinoe and @janvorli for additional visibility into package generation

cydhaselton:
Also copying @schellap for input

build.sh
```diff
@@ -74,7 +74,9 @@ initTargetDistroRid()
         if [ "$__BuildOS" == "Linux" ]; then
             if [ ! -e $ROOTFS_DIR/etc/os-release ]; then
                 echo "WARNING: Can not determine runtime id for current distro."
-                export __DistroRid=""
+                #Test hack to generate __DistroRid for Android
+                #TODO Find reliable way to generate __DistroRid for Android.
+                export __DistroRid="android.21-arm64"
```
gkhanna79: Just to be clear - we cannot commit with this change. Have you considered specifying this at command (e.g. __DistroRid=... ./build.sh) and then adding a check in this function if __DistroRid is already defined or not?

src/.nuget/Microsoft.NETCore.Native/Microsoft.NETCore.Native.props
```diff
@@ -3,4 +3,4 @@
   <PropertyGroup>
     <SupportedPackageOSGroups>Linux;OSX</SupportedPackageOSGroups>
   </PropertyGroup>
-</Project>
\ No newline at end of file
+</Project>
```
gkhanna79: Undo change to this file.

src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/runtime.Android.Microsoft.NETCore.Runtime.CoreCLR.props
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <_PlatformDoesNotSupportNiFiles Condition="'$(Platform)' == 'arm'">true</_PlatformDoesNotSupportNiFiles>
+    <_PlatformDoesNotSupportNiFiles Condition="'$(Platform)' == 'armel'">true</_PlatformDoesNotSupportNiFiles>
+    <_PlatformDoesNotSupportNiFiles Condition="'$(Platform)' == 'arm64'">true</_PlatformDoesNotSupportNiFiles>
+  </PropertyGroup>
+  <ItemGroup>
+    <NativeBinary Include="$(BinDir)libcoreclr.so" />
+    <NativeBinary Condition="'$(_PlatformDoesNotSupportNiFiles)' != 'true'" Include="$(BinDir)libcoreclrtraceptprovider.so" />
```
gkhanna79: This check is incorrect. __PlatformDoesNotSupportNiFiles_ should be used for platforms where we cannot generate NI images right now. They just happen to be same as where this binary does not build.

If you are able to build this for Android, please fix this check accordingly so that it can be included for that platform.

FYI - @hqueue @hseok-oh @jyoungyun

src/.nuget/dir.props
```diff
@@ -149,6 +162,14 @@
       <Platform>arm64</Platform>
     </OfficialBuildRID>
   </ItemGroup>
+  <ItemGroup Condition="$(SupportedPackageOSGroups.Contains(';Android;'))">
+    <OfficialBuildRID Include="android.21-arm64">
+      <Platform>arm64</Platform>
+    </OfficialBuildRID>
+    <OfficialBuildRID Include="android.21-arm">
```
gkhanna79: Have you tested the build for Arm32 for Android? If so, your change for __DsitroRid should account for it as well.

build.sh
cydhaselton: @gkhanna79:  I did not, but that is a good idea.

src/.nuget/dir.props
cydhaselton: I've only tested...and am only able to test...for arm64.  I can add arm into this change but won't be able to test it.

gkhanna79: In that case, please undo the Android for Arm change.

config.json
```diff
@@ -327,7 +327,7 @@
     "RuntimeId": {
       "description": "Specifies the OS to build Core_Root for",
       "valueType": "property",
-      "values": [ "debian.8-x64", "fedora.23-x64", "opensuse.42.1-x64", "osx.10.12-x64", "rhel.7-x64", "ubuntu.14.04-x64", "ubuntu.16.04-x64", "ubuntu.16.10-x64" ],
+      "values": [ "debian.8-x64", "fedora.23-x64", "opensuse.42.1-x64", "osx.10.10-x64", "rhel.7-x64", "ubuntu.14.04-x64", "ubuntu.16.04-x64", "ubuntu.16.10-x64", "android.21-arm64", "android.21-arm" ],
```
jyoungyun: I think you need to undo osx version change.

build.sh
jyoungyun: IMHO, you need to save `ANDROID_PLATFORM` variable as a file when you build rootfs and then you can use it in building coreclr. :)

cydhaselton: @jyoungyun This may be better; so far I've tried every iteration of detecting `__DistroRid` in this function and all have failed except for `! [[ -n "$__DistroRid" ]]`, which is the same as hardcoding that value.

cydhaselton: @gkhanna79 See recent commit

src/.nuget/Microsoft.NETCore.Native/Microsoft.NETCore.Native.props
cydhaselton: @gkhanna79 Done

config.json
cydhaselton: That's odd…don't remember making that change. Fixed

src/.nuget/dir.props
cydhaselton: Removed

build.sh
cydhaselton: FYI, Walking back all changes to build.sh and going route @jyoungyun  recommended. Too difficult getting logic check in `initTargetDistroRid()` to work without causing CI errors

cydhaselton:
Tagging @qmfrederik to review/ok addition to `build-android-rootfs.sh`

cydhaselton:
General question: I'm still getting up to speed with CI builds. Is there anything I need to do with the previous commits in this PR that have failed build checks if a later commit or commits are successful?

qmfrederik:
@cydhaselton The change to `build-android-rootfs` is fine with me.

Regarding to the CI builds, only the last commit matters. If the previous commits are no longer relevant, you can squash them, and they will disappear from the list of commits in this pull request.

Would it be easier to treat Android as a Linux distribution? That would avoid the need to add a new OS group.

cydhaselton:
Another (probably obvious) question: is there a way for me manually to re-run a CI build for a specific commit?

qmfrederik:
@cydhaselton You can force the CI build to re-run, but only for the last commit in your PR.

cydhaselton:
@gkhanna79 @jyoungyun and others…any other changes/additions I should make to this PR before merge?

build.sh
```diff
@@ -73,13 +73,18 @@ initTargetDistroRid()
...
+                else
+                    echo "WARNING: Can not determine runtime id for current distro."
+                    export __DistroRid=""
+                fi
             else
                 source $ROOTFS_DIR/etc/os-release
                 export __DistroRid="$ID.$VERSION_ID-$__BuildArch"
             fi
-        fi
+         fi
```
jyoungyun: Remove redundant space :)

cydhaselton: Good catch!  Fixed.

build.sh
```diff
@@ -73,8 +73,13 @@ initTargetDistroRid()
     if [ $__CrossBuild == 1 ]; then
         if [ "$__BuildOS" == "Linux" ]; then
             if [ ! -e $ROOTFS_DIR/etc/os-release ]; then
-                echo "WARNING: Can not determine runtime id for current distro."
-                export __DistroRid=""
+                if [ -e $ROOTFS_DIR/android_platform ]; then
+                    source $ROOTFS_DIR/android_platform
+                    export __DistroRid="$RID"
```
gkhanna79: I presume that $RID is being set by the source statement earlier,right? What is the value it gets?

config.json
```diff
@@ -327,7 +327,7 @@
     "RuntimeId": {
       "description": "Specifies the OS to build Core_Root for",
       "valueType": "property",
-      "values": [ "debian.8-x64", "fedora.23-x64", "opensuse.42.1-x64", "osx.10.12-x64", "rhel.7-x64", "ubuntu.14.04-x64", "ubuntu.16.04-x64", "ubuntu.16.10-x64" ],
+      "values": [ "debian.8-x64", "fedora.23-x64", "opensuse.42.1-x64", "osx.10.12-x64", "rhel.7-x64", "ubuntu.14.04-x64", "ubuntu.16.04-x64", "ubuntu.16.10-x64", "android.21-arm64", "android.21-arm" ],
```
gkhanna79: Were you not planning to remove the "arm" entry?

build.sh
gkhanna79: Nevermind - saw it further below.

src/.nuget/Microsoft.NETCore.ILAsm/runtime.Android.Microsoft.NETCore.ILAsm.props
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
```
gkhanna79: Thinking a bit more about this, why do you need to add these android specific files? Isn't the Linux version sufficient?

src/.nuget/dir.props
```diff
@@ -77,6 +78,13 @@
         <PackageRID Condition="'$(PortableBuild)' == '1'">osx-$(ArchGroup)</PackageRID>
       </PropertyGroup>
     </When>
+    <When Condition="'$(_runtimeOSFamily)' == 'android'">
+      <PropertyGroup>
+        <PackageRID>android.21-$(ArchGroup)</PackageRID>
```
gkhanna79: Should we have a check for arm64 since that is all we are supporting for now?

config.json
cydhaselton: I did, but reverted when working on fixing initTargetDistroRid logic in build.sh and forgot to re-remove.
Fixed

src/.nuget/Microsoft.NETCore.ILAsm/runtime.Android.Microsoft.NETCore.ILAsm.props
cydhaselton: I'm not sure; I'm still piecing together how nuget packages are generated. Since I don't know all of the places a) where it would be important to separate Android from Linux and b) where it wouldn't matter, I'm taking a "separate everything now, integrate with Linux after" approach.

I can remove the files and see if package build still works.

cydhaselton: @gkhanna79: So the build is successful, including packages, once `packageTargetOsGroup` for Android is removed from `dir.props` and conditionals added to `Microsoft.NETCore.Runtime.CoreCLR.pkgproj`

So my questions are:

1. What elements distinguish one package group from another? Or: What factors determine the creation of a new OS group?
2. Is it possible that a separate package group would be needed for Android in the future?

cydhaselton:
@gkhanna79 @qmfrederik  and others; re-posting this question in general discussion for greater visibility. Code discussion [here](https://github.com/dotnet/coreclr/pull/10286#pullrequestreview-30691338):

So my questions are:
1. What elements distinguish one package group from another? Or: What factors determine the creation of a new OS group?
2. Is it possible that a separate package group would be needed for Android in the future?

cydhaselton:
cc @hseok-oh and @ericstj for insight into OS Group package info

danmosemsft:

cydhaselton:
Also cc-ing @jkotas as per [documentation](https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/contributing-workflow.md); is there a person or persons that are responsible for the packaging subsystem that I should copy on this for feedback on the OS Group question?

jkotas:
You have the right people copied on this already  (@gkhanna79, @ericstj).

cydhaselton:
Thanks, @jkotas. Would it be appropriate for me to update the CODE_OWNERS.TXT with that info?

jkotas:
Yes, it sounds good to me.

cydhaselton:
Polite ping to @ericstj and @gkhanna79 again for input into above questions

src/.nuget/Microsoft.NETCore.ILAsm/runtime.Android.Microsoft.NETCore.ILAsm.props
gkhanna79: Given that Android is sufficiently distinct, but still Linux based, I am fine if a new OS group for it is used while leveraging the existing Linux definitions of the components.

CC @chcosta

src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/runtime.Android.Microsoft.NETCore.Runtime.CoreCLR.props
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <_PlatformDoesNotSupportNiFiles Condition="'$(Platform)' == 'arm'">true</_PlatformDoesNotSupportNiFiles>
+    <_PlatformDoesNotSupportNiFiles Condition="'$(Platform)' == 'armel'">true</_PlatformDoesNotSupportNiFiles>
+    <_PlatformDoesNotUseNiFiles Condition="'$(Platform)' == 'arm64'">true</_PlatformDoesNotUseNiFiles>
+  </PropertyGroup>
+  <ItemGroup>
+    <NativeBinary Include="$(BinDir)libcoreclr.so" />
+    <NativeBinary Condition="'$(_PlatformDoesNotSupportNiFiles)' != 'true'" Include="$(BinDir)libcoreclrtraceptprovider.so" />
```
gkhanna79: This just got recently changed to be supported. Thus, keeping it unified with the Linux definition will help streamline where it needs to be enabled vs not.

cydhaselton: Should I revert this change then or keep it as is with the separate `PlatformDoesNotUseNiFiles` condition?

src/.nuget/Microsoft.NETCore.ILAsm/runtime.Android.Microsoft.NETCore.ILAsm.props
cydhaselton: Unfortunately, unless I'm missing something, `packageTargetOSGroup` means adding Android-specific runtime-prefixed files (`runtime.Android...`).

src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/runtime.Android.Microsoft.NETCore.Runtime.CoreCLR.props
cydhaselton: cc @gkhanna79

src/.nuget/Microsoft.NETCore.ILAsm/runtime.Android.Microsoft.NETCore.ILAsm.props
cydhaselton: cc @gkhanna79

chcosta: I think it's reasonable for Android to use the Linux `_packageTargetOSGroup` to import the runtime package imports

src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/runtime.Android.Microsoft.NETCore.Runtime.CoreCLR.props
danmosemsft: @gkhanna79 for q above.

src/.nuget/Microsoft.NETCore.ILAsm/runtime.Android.Microsoft.NETCore.ILAsm.props
cydhaselton: @chcosta and @gkhanna79: Still not sure how to achieve what you are suggesting; if I create a `packageTargetOSGroup` for Android without adding the `runtime.Android` files package build fails.  Is there another OS that does this that I could reference?

src/.nuget/dir.props
```diff
@@ -34,6 +34,7 @@
   <!-- derive an OS Group based on the OS Family -->
   <PropertyGroup>
     <_derivedPackageTargetOSGroup Condition="'$(_derivedPackageTargetOSGroup)' == '' and '$(_runtimeOSFamily)' == 'osx'">OSX</_derivedPackageTargetOSGroup>
+    <_derivedPackageTargetOSGroup Condition="'$(_derivedPackageTargetOSGroup)' == '' and '$(_runtimeOSFamily)' == 'android'">Android</_derivedPackageTargetOSGroup>
```
chcosta: @cydhaselton, `_packageTargetOSGroup` is the property which handles the runtime package import.  If you deleted this line, then I'd expect it to default to importing "Linux" runtime package projects.  That's probably not quite what we want though because it would then be ignoring the "Android" piece of the `SupportedPackageOSGroups`.  Instead, I would leave this code as-is, but add a conditional property definition here https://github.com/dotnet/coreclr/blob/8e986affba47efc1ce85ab0bb9428a6c28dae279/src/.nuget/dir.props#L47, which is something like

```
<_packageTargetOSGroup Condition="'$(_derivedPackageTargetOSGroup)' == 'Android'">Linux</_packageTargetOSGroup>
```

and maybe add a comment explaining that Android is using the same runtime pkgproj's as Linux.
```

Comment (cydhaselton) [inline src/.nuget/dir.props]: @chcosta FYI, Implementing the above causes the package build for `Microsoft.NETCore.Runtime.CoreCLR` to fail. Specifically the build ignores the conditional I've copied from the `runtime.Android` file to the `runtime.Linux`file  to exclude `System.Private.Corelib.Ni.dll` from the package. If I use the existing condition for Ni  it does work…but since `Ni` files can be generated post-build on Android it shouldn't be used.

Comment (cydhaselton) [inline src/.nuget/dir.props]: Polite ping to @gkhanna79 and @ericstj  for input into above question.

Comment (gkhanna79) [inline src/.nuget/dir.props]: @cydhaselton SPC.ni.dll generation has changed in the last week - can you please rebase and see if you are still running into this issue?

Comment (cydhaselton) [inline src/.nuget/dir.props]: @gkhanna79 I am not; rebase resolved the issue. See updated code

**Comment by cydhaselton** [comment]
@gkhanna79 The above looks messy; did you mean "merge" instead of "rebase?

**Comment by gkhanna79** [comment]
What is looking messy?

**Comment by cydhaselton** [comment]
> What is looking messy?
@gkhanna79 Ignore…see edited message. Basically didn't do `git push -f` after rebase.

**Comment by cydhaselton** [comment]

**Comment by cydhaselton** [comment]
@gkhanna79, @ericstj and others…is there anything else that needs to be done for this PR?

File: config.json [hunk 15]
```diff
@@ -345,7 +345,7 @@
     "RuntimeId": {
       "description": "Specifies the OS to build Core_Root for",
       "valueType": "property",
-      "values": [ "debian.8-x64", "fedora.24-x64", "fedora.25-x64", "opensuse.42.1-x64", "osx.10.12-x64", "rhel.7-x64", "ubuntu.14.04-x64", "ubuntu.16.04-x64", "ubuntu.16.10-x64" ],
+      "values": [ "debian.8-x64", "fedora.23-x64", "fedora.25-x64", "opensuse.42.1-x64", "osx.10.12-x64", "rhel.7-x64", "ubuntu.14.04-x64", "ubuntu.16.04-x64", "ubuntu.16.10-x64", "android.21-arm64" ],
```
ericstj: This change to fedora appears to be a bad merge.

src/.nuget/dir.props
```diff
@@ -102,6 +111,13 @@
         <SymbolFileExtension>.dwarf</SymbolFileExtension>
       </PropertyGroup>
     </When>
+    <When Condition="'$(_runtimeOSFamily)' == 'android'">
+      <PropertyGroup>
+        <LibraryFileExtension>.so</LibraryFileExtension>
+<!--symbols included in .so, like Linux, but can be generated externally and if so, uses .debug ext-->
```
ericstj: nit: indentation

config.json
cydhaselton: Crap. Fixed

src/.nuget/dir.props
cydhaselton: Fixed

ericstj:
I've reviewed and commented on stuff that stuck out to me.  I don't see anything else concerning.  Most of these changes are around how coreclr decides to factor its package source which I don't have an opinion on so long as it produces the correct results.  I'd like for @gkhanna79 or @chcosta to give explicit sign off on that.

cydhaselton:
Thanks Eric.
@gkhanna79 and/or @chcosta let me know if there's anything you see that needs to be fixed.

cydhaselton:
Polite re-ping to @gkhanna79 and @chcosta … is there anything I need to change in this PR before merge?

cross/build-android-rootfs.sh
```diff
@@ -136,7 +137,11 @@ dpkg -x $__Android_Cross_Dir/deb/libunwind_1.2.20170304_$__AndroidArch.deb $__An
 
 cp -R $__Android_Cross_Dir/tmp/$__AndroidArch/data/data/com.termux/files/usr/* $__ToolchainDir/sysroot/usr/
 
+# Generate platform file for build.sh script to assign to __DistroRid
+# TODO: replace with logic that generates RID based on $__ApiLevel and $__BuildArch
+echo "Generating platform file..."
 
+echo "RID=android.21-arm64" > $__ToolchainDir/sysroot/android_platform
```
chcosta: Any reason not to address this TODO and use the ApiLevel / BuildArch here?

src/.nuget/dir.props
```diff
@@ -163,6 +179,11 @@
       <Platform>arm64</Platform>
     </OfficialBuildRID>
   </ItemGroup>
+  <ItemGroup Condition="$(SupportedPackageOSGroups.Contains(';Android;'))">
+    <OfficialBuildRID Include="android.21-arm64">
```
chcosta: Do you want to include the Portable rid?

chcosta:
Two additional comments, and then LGTM

cross/build-android-rootfs.sh
cydhaselton: @chcosta The TODO was addressed in previous commit by adding logic to `build-android-rootfs.sh`; I accidentally left the comment. Will remove comment and echo commands.

src/.nuget/dir.props
cydhaselton: @chcosta At this line? I don't think so…a quick check of master shows that the other supported OSs don;t include it at this point. Should it be?

@gkhanna79 Anything you see that needs changing?

gkhanna79: OfficialBuildRID is only used for platform targets that the official .NET Core build is going to support and we should add it only when we enable building for it. Otherwise, this is not required.

@cydhaselton Can you please undo this change for now?

cydhaselton: Out of town but will push change when I return.

cydhaselton: @gkhanna79 Removed

cydhaselton:
@dotnet-bot test Tizen armel Cross Debug Build and Windows_NT x64 Debug Build and Test

cydhaselton:

cydhaselton:
@gkhanna79 and @chcosta are there any other changes needed for this PR?

cydhaselton:
FYI will be keeping this branch around until I have time to write up documentation.

Issue #9806, opened 2017-02-26
Math.Atan2 does not follow IEEE 754 for all inputs


tannergooding:
The IEEE 754 spec indicates:
* atan2(±∞, −∞) is ±3π/4
* atan2(±∞, +∞) is ±π/4.

However, the current implementation returns NaN for the specified inputs.

danmosemsft:
cc @dcwuser also

dcwuser:
For reference, here is the spec: https://www.csee.umbc.edu/~tsimo1/CMSC455/IEEE-754-2008.pdf. The required special cases for atan2 are on page 45. I have written a test program for all of them, and the ones listed in this bug are the only ones not satisfied. (Caveat: while I was careful to distinguish +0 from -0 on input, I used a comparator that regards them as equal on output.)

Two points:

A. It wouldn't hurt for us to fulfill the spec.

B. Of all the special cases listed in the spec, these are the least important and most debatable. To derive these cases, you have to regard separate infinities as equal, and we certainly don't do that in other contexts, for example we say -inf/+inf = NaN, not -inf/+inf = -1.

So I would say: if this is easy to fix, and the fix doesn't impact performance, and we have have cycles, let's fix it. But if any of those conditions are violated, let's not.

gkhanna79:
CC @AlexGhiondea

tannergooding:
Self assigning since this is fixed by https://github.com/dotnet/coreclr/pull/10295

stephentoub:
Reopening to track test fixes

tannergooding:
Fixed with https://github.com/dotnet/corefx/pull/17217

Issue #10289, opened 2017-03-19
Math.Pow on non-windows is incorrect for X = -0.0, and Y < 0 where it is an odd integer


tannergooding:
We have special handling in the PAL layer to ensure that `pow` handles certain edge cases correctly. However, the handling for X = -0.0 and Y < 0 (see https://github.com/dotnet/coreclr/blob/master/src/pal/src/cruntime/math.cpp#L376) is invalid according to the IEEE spec.

The spec actually indicates:
For the pow function (integral exponents get special treatment):
* pow (x, ±0) is 1 for any x (even a zero, quiet NaN, or infinity)
* pow (±0, y) is ±∞ and signals the divideByZero exception for y an odd integer < 0
* pow (±0, −∞) is +∞ with no exception
* pow (±0, +∞) is +0 with no exception
* pow (±0, y) is +∞ and signals the divideByZero exception for finite y < 0 and not an odd integer
* pow (±0, y) is ±0 for finite y > 0 an odd integer
* pow (±0, y) is +0 for finite y > 0 and not an odd integer
* pow (−1, ±∞) is 1 with no exception
* pow (+1, y) is 1 for any y (even a quiet NaN)
* pow (x, y) signals the invalid operation exception for finite x < 0 and finite non-integer y.

This means, that for `Pow(+0.0, -3.0)` we should be returning `+∞` and for `Pow(-0.0, -3.0)` we should be returning `−∞`

tannergooding:
Self-assigning. This is resolved by https://github.com/dotnet/coreclr/pull/10295

tannergooding:
Fixed with https://github.com/dotnet/corefx/pull/17217

Issue #9807, opened 2017-02-26
Math.Pow does not follow IEEE 754 for all inputs


tannergooding:
The IEEE 754 spec defines the following inputs as special:

For the pown function (integral exponents only):
* pown (x, 0) is 1 for any x (even a zero, quiet NaN, or infinity)
* pown (±0, n) is ±∞ and signals the divideByZero exception for odd integral n < 0
* pown (±0, n) is +∞ and signals the divideByZero exception for even integral n < 0
* pown (±0, n) is +0 for even integral n > 0
* pown (±0, n) is ±0 for odd integral n > 0.

For the pow function (integral exponents get special treatment):
* pow (x, ±0) is 1 for any x (even a zero, quiet NaN, or infinity)
* pow (±0, y) is ±∞ and signals the divideByZero exception for y an odd integer < 0
* pow (±0, −∞) is +∞ with no exception
* pow (±0, +∞) is +0 with no exception
* pow (±0, y) is +∞ and signals the divideByZero exception for finite y < 0 and not an odd integer
* pow (±0, y) is ±0 for finite y > 0 an odd integer
* pow (±0, y) is +0 for finite y > 0 and not an odd integer
* pow (−1, ±∞) is 1 with no exception
* pow (+1, y) is 1 for any y (even a quiet NaN)
* pow (x, y) signals the invalid operation exception for finite x < 0 and finite non-integer y.

For the powr function (derived by considering only exp(y * log(x))):
* powr (x, ±0) is 1 for finite x > 0
* powr (±0, y) is +∞ and signals the divideByZero exception for finite y < 0
* powr (±0, −∞) is +∞
* powr (±0, y) is +0 for y > 0
* powr (+1, y) is 1 for finite y
* powr (x, y) signals the invalid operation exception for x < 0
* powr (±0, ±0) signals the invalid operation exception
* powr (+∞, ±0) signals the invalid operation exception
* powr (+1, ±∞) signals the invalid operation exception
* powr (x, qNaN) is qNaN for x ≥ 0
* powr (qNaN, y) is qNaN. 

However, the current implementation does not return the correct values for a number of these inputs (especially around the handling of NaN).

danmosemsft:
@dcwuser are you interested in getting cc'd on issues of such type? you did work in CoreFX to improve our handling of math edge cases.

dcwuser:
Thanks @danmosemsft, yes I am interested.

[...]

PR #10295, opened 2017-03-19
Removing the special handling in classlibnative for atan2(±∞, ±∞) and pow(-1.0, ±∞).

tannergooding:
These functions were incorrectly adjusting the handling for specific edge cases and making them differ from the result specified in IEEE 754.

This resolves https://github.com/dotnet/coreclr/issues/9806 and https://github.com/dotnet/coreclr/issues/9807

### DIFF
--- a/src/pal/tests/palsuite/c_runtime/pow/test1/test1.cpp
+++ b/src/pal/tests/palsuite/c_runtime/pow/test1/test1.cpp
@@ -106,12 +106,18 @@ int __cdecl main(int argc, char **argv)
         { -2.7182818284590452,     1,                     -2.7182818284590452,        PAL_EPSILON * 10 },     // x: -(e)                       expected: e
         { -2.7182818284590452,     PAL_POSINF,             PAL_POSINF,                0 },                    // x: -(e)
 
+        { -1.0,                    PAL_NEGINF,             1.0,                       PAL_EPSILON * 10 },
+        { -1.0,                    PAL_POSINF,             1.0,                       PAL_EPSILON * 10 },
+
         { -0.0,                    PAL_NEGINF,             PAL_POSINF,                0 },
         { -0.0,                   -1,                      PAL_NEGINF,                0 },
         { -0.0,                   -0.0,                    1,                         PAL_EPSILON * 10 },
         { -0.0,                    0,                      1,                         PAL_EPSILON * 10 },
         { -0.0,                    1,                     -0.0,                       PAL_EPSILON },
         { -0.0,                    PAL_POSINF,             0,                         PAL_EPSILON },
+
+        {  PAL_NAN,               -0.0,                    1.0,                       PAL_EPSILON * 10 },
+        {  PAL_NAN,                0,                      1.0,                       PAL_EPSILON * 10 },
 
         {  0.0,                    PAL_NEGINF,             PAL_POSINF,                0 },
         {  0.0,                   -1,                      PAL_POSINF,                0 },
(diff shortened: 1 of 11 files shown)


tannergooding:
FYI. @mellinoe

tannergooding:
Also fixed the logic for Pow in the PAL layer to match the current windows behavior and the IEEE spec. This should resolve https://github.com/dotnet/coreclr/issues/10289

tannergooding:
The `HAVE_COMPATIBLE_EXP` logic is currently broken as well. It is failing for Mac and Linux, when it shouldn't be. However, the initial fix I did, seems to block CMAKE on the `Linux ARM Emulator Cross Build`.

tannergooding:
The OSX10.12 job is failing, but in a manner unrelated to this PR.

tannergooding:
OSX10.12 jobs in general seem to have been failing for a while now.

tannergooding:
This should be good now, @mellinoe.

mellinoe:
Someone more familiar with the native code here should probably review this, but I'm supportive of the direction.

tannergooding:
tagging @janvorli, @AndyAyersMS, and @jkotas, who reviewed things last time I touched this code 😄

src/pal/tests/palsuite/c_runtime/pow/test1/test1.cpp
```diff
@@ -106,13 +106,19 @@ int __cdecl main(int argc, char **argv)
         { -2.7182818284590452,     1,                     -2.7182818284590452,        PAL_EPSILON * 10 },     // x: -(e)                       expected: e
         { -2.7182818284590452,     PAL_POSINF,             PAL_POSINF,                0 },                    // x: -(e)
 
+        { -1.0,                    PAL_NEGINF,             1.0,                       PAL_EPSILON * 10 },
```
tannergooding: I have a work item to clean up these tests in general (actually doing validation for exact values, and against negative zero, etc). I've already done this in the corefx pr, for the most part. I just haven't gotten to it here yet.

tannergooding: As for why PAL_EPSILON * 10, the comment on PAL_EPSILON explains.

Basically, we are allowed 15-17 digits of precision. That is 0.xxxxxxxxxxxxxxxxx (validate against PAL_EPSILON) or x.xxxxxxxxxxxxxxxx (PAL_EPSILON * 10) or xx.xxxxxxxxxxxxxxx (PAL_EPSILON * 100) or even 0.0xxxxxxxxxxxxxxxxx (PAL_EPSILON / 10).

If we just used PAL_EPSILON everywhere, we would get incorrect results for larger or smaller results.

dcwuser: I'm not talking about everywhere, I'm talking about the specific tests you added in this PR, where we can expect the result to be exact, e.g. (-1)^(-inf) = 1.0, not 1.0 +/ 1.0E-15.

tannergooding: Yes, its probably the case that we can expect an exact result here. I was just being consistent with the existing convention here.

There are plenty of things that could be improved. For example, the PAL layer (as a compatibility test) currently checks if `atan2(0.0, -0.1)` is within `0.0000001` digits of PI, (https://github.com/dotnet/coreclr/blob/master/src/pal/src/configure.cmake#L797), such logic is incredibly broken 😄 

I have been working on a separate PR that fixes all of this up, and it will be corrected when that goes out. This PR, at the very least, will fix the validate function to properly check:
* Exact results
* NaN
* -0.0
* +0.0
* Negative Infinity
* Positive Infinity
* Results that are within 15 digits of the expected value

tannergooding: That being said, if you feel strongly about this going in as inexact (especially considering the 1500 other scenarios we validate that are also inexact), I can fix it now, rather than as a whole in the separate PR.

dcwuser: No, this is very minor. I just wanted to understand the reason. It's fine to fix them all in one go later.

tannergooding:

Looks like jobs are passing now. Will rebase if it fails again.

tannergooding:
*Ping* Could I get a review of the PAL/native changes here.

danmosemsft:
ping of @janvorli, @AndyAyersMS, and @jkotas could someone look?

PR #10296, opened 2017-03-19
System.IO.Path multiple VolumeSeperators throwed NotSupportedExceptio…

ViktorHofer:
…n instead of ArgumentException on Windows

CoreFX issue: https://github.com/dotnet/corefx/issues/15098

### DIFF
--- a/src/mscorlib/shared/System/IO/Path.Windows.cs
+++ b/src/mscorlib/shared/System/IO/Path.Windows.cs
@@ -67,11 +67,14 @@ namespace System.IO
                 startIndex += 2;
 
                 if ((path.Length > 0 && path[0] == PathInternal.VolumeSeparatorChar)
-                    || (path.Length >= startIndex && path[startIndex - 1] == PathInternal.VolumeSeparatorChar && !PathInternal.IsValidDriveChar(path[startIndex - 2]))
-                    || (path.Length > startIndex && path.IndexOf(PathInternal.VolumeSeparatorChar, startIndex) != -1))
+                    || (path.Length >= startIndex && path[startIndex - 1] == PathInternal.VolumeSeparatorChar && !PathInternal.IsValidDriveChar(path[startIndex - 2])))
                 {
                     throw new NotSupportedException(SR.Argument_PathFormatNotSupported);
                 }
+
+                // Only one VolumeSeperator char is allowed in a path on Windows.
+                if (path.Length > startIndex && path.IndexOf(PathInternal.VolumeSeparatorChar, startIndex) != -1)
+                    throw new ArgumentException(SR.Arg_PathIllegal);
             }
 
             // Technically this doesn't matter but we used to throw for this case


src/mscorlib/shared/System/IO/Path.Windows.cs
```diff
@@ -67,11 +67,14 @@ public static string GetFullPath(string path)
...
 
                 if ((path.Length > 0 && path[0] == PathInternal.VolumeSeparatorChar)
-                    || (path.Length >= startIndex && path[startIndex - 1] == PathInternal.VolumeSeparatorChar && !PathInternal.IsValidDriveChar(path[startIndex - 2]))
-                    || (path.Length > startIndex && path.IndexOf(PathInternal.VolumeSeparatorChar, startIndex) != -1))
+                    || (path.Length >= startIndex && path[startIndex - 1] == PathInternal.VolumeSeparatorChar && !PathInternal.IsValidDriveChar(path[startIndex - 2])))
                 {
                     throw new NotSupportedException(SR.Argument_PathFormatNotSupported);
                 }
+
+                // Only one VolumeSeperator char is allowed in a path on Windows.
```
danmosemsft: typo, "Separator"

danmosemsft:
Is there a test for this (it would be in CoreFX)? I assume not as it would be failing right now.

Assuming @JeremyKuhne okays this change, could you please also prepare a PR against CoreFX that has a test that verifies it?

This explains how to run that test against your coreclr change here:
https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/developer-guide.md#testing-with-private-coreclr-bits

You will also want to run that test against desktop, to make sure it expects the same behavior of desktop, which I believe is done with `msbuild /t:buildandtest /p:targetgroup=netfx` in the correct corefx `tests` folder. You'll likely have to do `build -framework:netfx` from the root first.

src/mscorlib/shared/System/IO/Path.Windows.cs
```diff
@@ -67,11 +67,14 @@ public static string GetFullPath(string path)
...
+                if (path.Length > startIndex && path.IndexOf(PathInternal.VolumeSeparatorChar, startIndex) != -1)
+                    throw new ArgumentException(SR.Arg_PathIllegal);
```
ianhays: I assume the desktop Error doesn't include a paramname?

ianhays:
>Is there a test for this (it would be in CoreFX)? I assume not as it would be failing right now.

There is. They just fail on desktop last I checked. [This is the test](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Extensions/tests/System/IO/PathTests.cs#L478)

ViktorHofer:
Yeah there are test in corefx and I'm already preparing additional tests. As soon as it syncs and I'm ready I will submit a PR for that in corefx.

src/mscorlib/shared/System/IO/Path.Windows.cs
JeremyKuhne: No paramname on desktop

ViktorHofer:
As net461 runs on netstandard 2.0 we need to mimic that behavior and not net462s behavior. Then the work would still make sense. Right?

JeremyKuhne:
> As net461 runs on netstandard 2.0 we need to mimic that behavior and not net462s behavior. Then the work would still make sense. Right?

No, as Core is supposed to be in the lead- not chained behind NetFX improvements. In addition, if you ported your app from 4.6.2 to Core it would go "backwards" in behavior.  4.6.2 also adds long path support- we should, in theory, fail on long paths if we made 4.6.1 behavior the standard (which is very undesirable).

That said we definitely need to update the tests here to validate appropriately based on CoreFx/NetFx and the compat switch setting on NetFx when running there. So there is work here to fix the tests- just not the code. Note that it is super important to get these tests up to date to help prevent unintended regressions on both Core and NetFx.

I'm sorry that I messed up when opening the original issue. I totally spaced on the fact that I had updated 4.6.2 with regards to colon checks. :/

PR #10306, opened 2017-03-20
Reconciles Type.cs with CoreRT version for move to shared partition.

AtsushiKan:


AtsushiKan:
cc @jkotas 
cc @alexperovich

src/mscorlib/src/System/Type.CoreCLR.cs
```diff
@@ -0,0 +1,253 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+//
+//
+//
```
jkotas: Nit: I would be nice to delete this useless comment block

Issue #10033, opened 2017-03-08
gdbjit.cpp is inconsistent in using new and new (nothrow)


janvorli:
There are about 40 usages of new (nothrow) and about 14 throwing news without any particular reason. We should unify it.

janvorli:
cc: @lucenticus, @ayuckhulk, @kvochko, @Dmitri-Botcharnikov

ayuckhulk:
I'll look into this.

In some places there are also no error handling paths (for failed `new (nothrow)` case) so its also should be fixed.

@janvorli  Am I correct that it's better to use `(nothrow)` everywhere?

janvorli:
@ayuckhulk in fact, I would prefer the opposite, using throwing new and then wrap the entry point functions in try / catch. That would remove some code churn caused by checking for nullptr and make the code easier to read.
I think that the only entry points are NotifyGdb::MethodCompiled and NotifyGdb::MethodDropped, right?
Maybe some of the code will need to be hardened against the oom exceptions so that no memory leaks and no data structures are left in an inconsistent state. The code already uses NewArrayHolder and NewHolder at many if not all places, which takes automatically care of releasing the memory in case of an exception.

ayuckhulk:
@janvorli  I think we can close this issue.

PR #10307, opened 2017-03-20
[Linux][GDB-JIT] Add lexical scopes info for local variables

ayuckhulk:
This PR adds lexical scopes info for local variables in GDBJIT. It also includes related code refactoring and memory leak fix.

Summary:
* Pass addresses of lexical scopes for local variables from SymbolReader to GDBJIT
* Fix memory leak when passing structures from SymbolReader to GDBJIT
* Generate lexical scope DWARF entries
* Move global variables to local scope to avoid confusion about lifetimes of methods, lines and symbol names

@mikem8361, @janvorli PTAL

### DIFF
--- a/src/vm/gdbjit.cpp
+++ b/src/vm/gdbjit.cpp
@@ -530,7 +530,7 @@ GetDebugInfoFromPDB(MethodDesc* MethodDescPtr, SymbolsInfo** symInfo, unsigned i
     if (*symInfo == nullptr)
         return E_FAIL;
     locals.size = methodDebugInfo.localsSize;
-    locals.localsName = new (nothrow) char *[locals.size];
+    locals.localsName = new (nothrow) NewArrayHolder<char>[locals.size];
     if (locals.localsName == nullptr)
         return E_FAIL;
     locals.localsScope = new (nothrow) LocalsInfo::Scope [locals.size];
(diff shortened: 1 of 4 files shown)


ayuckhulk:

ayuckhulk:

src/vm/gdbjit.cpp
```diff
@@ -1692,13 +1802,16 @@ static int getNextPrologueIndex(int from, const SymbolsInfo *lines, int nlines)
...
-NotifyGdb::AddrSet codeAddrs;
+static NotifyGdb::AddrSet codeAddrs;
 
 /* Create ELF/DWARF debug info for jitted method */
 void NotifyGdb::MethodCompiled(MethodDesc* MethodDescPtr)
 {
+    FunctionMemberPtrArrayHolder method;
+
+    int SymbolCount = 0;
+    NewArrayHolder<Elf_Symbol> SymbolNames;
```
janvorli: Could you please be consistent with the naming convention used in this function and use camel casing for locals?

src/vm/gdbjit.cpp
```diff
@@ -554,6 +584,15 @@ GetDebugInfoFromPDB(MethodDesc* MethodDescPtr, SymbolsInfo** symInfo, unsigned i
         }
     }
 
+    if (methodDebugInfo.locals)
```
janvorli: It would be better to add a destructor to the MethodDebugInfo and let it destroy the locals and points. Then just letting the methodDebugInfo go out of scope would perform the proper cleanup. and it would enable proper cleanup even in case of exeption unwinding. I have suggested to convert all nothrow news in this file to throwing ones in #10033 and with that, automatic cleanup would be required.

ayuckhulk:
@janvorli Thank you!

PR #10319, opened 2017-03-20
Permit null object in `GenTreeUseEdgeIterator`

JosephTremoulet:
We use `GT_FIELD` for static field loads, which have no instance pointer.
Update the iterator constructor to recognize this as a valid state rather
than fail an assertion.

### DIFF
--- a/src/jit/gentree.cpp
+++ b/src/jit/gentree.cpp
@@ -9322,9 +9322,15 @@ GenTreeUseEdgeIterator::GenTreeUseEdgeIterator(GenTree* node)
             return;
 
         case GT_FIELD:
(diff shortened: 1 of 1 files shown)


JosephTremoulet:
@pgavlin PTAL
/cc @dotnet/jit-contrib

JosephTremoulet:
> How did you catch this?

by trying to dump a method with a static field load

pgavlin:
> by trying to dump a method with a static field load

Hah, got it. It is possible that there are some other issues like this: I assume that we did not catch this because under normal circumstances we primarily (only?) use the operand iterator in LIR (where these nodes do not exist).

PR #10321, opened 2017-03-20
Issue #7060 - Make StreamReader.ReadBuffer non-virtual

### LINKED ISSUE #7060: StreamReader.ReadBuffer should not be virtual
It's an [internal virtual method](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/IO/StreamReader.cs#L563) that is overriden nowhere except `NullStreamReader`. Since it's used in many places throughout the class, like `ReadToEnd` / `ReadLine` / etc, it may be worthwhile to make it non-virtual.

WinCPP:
Fixes #7060 

@ianhays @karelz Kindly review.

### DIFF
--- a/src/mscorlib/src/System/IO/StreamReader.cs
+++ b/src/mscorlib/src/System/IO/StreamReader.cs
@@ -540,7 +540,7 @@ namespace System.IO
             return sb.ToString();
         }
         
-        internal virtual int ReadBuffer() {
+        internal int ReadBuffer() {
             charLen = 0;
             charPos = 0;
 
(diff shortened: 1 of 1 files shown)


ianhays:
@jamesqo mentioned in #7060 that this is overridden in `NullStreamReader`. We'll need to do something there as well.

jamesqo:
@ianhays The red diff is the implementation that has been removed from `NullStreamReader`.

ianhays:
>The red diff is the implementation that has been removed from NullStreamReader.

Oops. Nothing to see here, move along.

Issue #10320, opened 2017-03-20
Memory leak when WaitAny on the same Task keeps timing out


svick:
As reported by @ayende in [a blog post](https://ayende.com/blog/177346/the-memory-leak-in-the-network-partition), when `Task.WaitAny` is being called on the same unfinished `Task` in a loop with a timeout, it causes a memory leak in the `m_continuationObject` list of the `Task`.

Minimal repro:

```c#
using System.Threading.Tasks;

public static class Program
{
    public static void Main()
    {
        var tasks = new[] { Task.Delay(-1) };
        while (true)
        {
            Task.WaitAny(tasks, millisecondsTimeout: 1); // memory leak!
        }
    }
}
```

The `m_continuationObject` of the `Task` in question is a `List<object>` containing thousands of `CompleteOnInvokePromise`, even though the `WaitAny()` calls corresponding to those objects already completed.

stephentoub:
Definitely a bug.  WaitAny is calling to TaskFactory.CommonCWAnyLogic(Task[]), which doesn't take a timeout.  It then waits with a timeout on the resulting task, but that continuation created has know knowledge of the timeout and thus sits around until any of the constituent tasks completes, which in the case mentioned, will never happen.

PR #10331, opened 2017-03-20
Fix leak in Task.WaitAny(..., int)

stephentoub:
WaitAny is effectively built on top of WhenAny, creating a continuation from the supplied tasks and then blocking on that continuation.  When a timeout is provided, it blocks with that timeout.  But if it doesn't complete within the timeout, it ends up leaking the continuations it created into the constituent tasks.  The fix is simply to force the returned continuation to complete, such that its continuation logic does all of the appropriate cleanup.

Fixes https://github.com/dotnet/coreclr/issues/10320
cc: @kouvel, @alexperovich


danmosemsft:
@dotnet-bot test osx10.12 x64 checked build and test (https://github.com/dotnet/coreclr/issues/8321)

kouvel:
(#8321)

PR #10332, opened 2017-03-20
[Local GC] Forbid inclusion of gcscan.h from VM and DAC directories

swgillespie:
This change removes the inclusion of `gcscan.h` from three places in the VM and once place in the DAC. For the VM uses, `IGCHeap::RuntimeStructuresValid` was added so that the VM can ask the GC whether structures are in valid state (previously done through `GCScan::GetGcRuntimeStructuresValid`). For the DAC use, the DAC variable that `GCScan::GetGcRuntimeStructuresValid` reads is referenced directly (`gc_structures_invalid_cnt`).

### DIFF
--- a/src/vm/object.cpp
+++ b/src/vm/object.cpp
@@ -19,7 +19,6 @@
 #include "eeconfig.h"
 #include "gcheaputilities.h"
 #include "field.h"
-#include "gcscan.h"
 #include "argdestination.h"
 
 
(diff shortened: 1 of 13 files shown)


swgillespie:
x86 job failed with https://github.com/dotnet/coreclr/issues/10022

swgillespie:
@Maoni0 @jkotas PTAL? Most of the work to enable this was done as a part of https://github.com/dotnet/coreclr/pull/9255 - so the functionality of `GCScan::GetGcRuntimeStructuresValid` can be implemented on the DAC side by reading one of the globals on the GC DAC struct.

src/vm/object.cpp
```diff
@@ -1776,9 +1775,10 @@ VOID Object::ValidateInner(BOOL bDeep, BOOL bVerifyNextHeader, BOOL bVerifySyncB
         // try to validate next object's header
         if (bDeep 
             && bVerifyNextHeader 
-            && GCScan::GetGcRuntimeStructuresValid ()
+            && GCHeapUtilities::IsGCHeapInitialized()
```
jkotas: IsGCHeapInitialized call does not look necessary - if we have an object to validate, it sure must have come from GCHeap and thus the GCHeap is initialized.

swgillespie: ok!

swgillespie:

swgillespie:
thanks!

PR #10335, opened 2017-03-21
[x86/Linux] Log UpdateRegDisplay

parjong:
This commit adds various UpdateRegDisplay logs for x86.


parjong:
@janvorli Please take a look.

PR #10342, opened 2017-03-21
[PAL] Do NOT check execute bit in INIT_FindEXEPath

parjong:
This commit revises INIT_FindEXEPath not to check execute bit if the path is fully specified (as discussed in #10246).

### DIFF
--- a/src/pal/src/init/pal.cpp
+++ b/src/pal/src/init/pal.cpp
@@ -105,7 +105,7 @@ static PCRITICAL_SECTION init_critsec = NULL;
 static int Initialize(int argc, const char *const argv[], DWORD flags);
 static BOOL INIT_IncreaseDescriptorLimit(void);
 static LPWSTR INIT_FormatCommandLine (int argc, const char * const *argv);
-static LPWSTR INIT_FindEXEPath(LPCSTR exe_name);
+static LPWSTR INIT_EXEPath(LPCSTR exe_name);
 
 #ifdef _DEBUG
 extern void PROCDumpThreadList(void);
@@ -406,7 +406,7 @@ Initialize(
         }
 
         /* find out the application's full path */
-        exe_path = INIT_FindEXEPath(argv[0]);
+        exe_path = INIT_EXEPath(argv[0]);
         if (NULL == exe_path)
         {
             ERROR("Unable to find exe path\n");
@@ -1130,14 +1130,10 @@ static LPWSTR INIT_FormatCommandLine (int argc, const char * const *argv)
 
 /*++
 Function:
-  INIT_FindEXEPath
-
-Abstract:
-    Determine the full, canonical path of the current executable by searching
-    $PATH.
+  INIT_EXEPath
 
 Parameters:
-    LPCSTR exe_name : file to search for
+    LPCSTR exe_name : the full path of the current executable
 
 Return:
     pointer to buffer containing the full path. This buffer must be released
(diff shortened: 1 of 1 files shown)


parjong:
It seems that path lookup case needs execution bit check.

parjong:

jkotas:
I think that the whole UTIL_IsExecuteBitsSet and INIT_FindEXEPath methods can be removed. They seem to be only used in by getPath to find path to process to launch all over the place. I do not think that CoreCLR is using this feature anywhere - and looking for a process to launch all over the place is not good for security.

janvorli:
I agree with @jkotas. CoreCLR hosts are already expected to pass in full absolute path. See coreclr_initialize function comment:
//  exePath                 - **Absolute** path of the executable that invoked the ExecuteAssembly

src/pal/src/init/pal.cpp
```diff
@@ -1150,269 +1146,66 @@ Notes 2:
...
+    if ( -1 == stat( exe_path, &theStats ) )
     {
-        WARN("$PATH isn't set.\n");
-        if (env_path != NULL)
-        {
-            free(env_path);
-        }
-
-        goto last_resort;
+      ERROR( "The file does not exist\n" );
```
janvorli: A nit - wrong indentation here and at several other places.

src/pal/src/init/pal.cpp
```diff
@@ -1150,269 +1146,66 @@ Notes 2:
...
     }
 
-    free(env_path);
-    TRACE("No %s found in $PATH (%s)\n", exe_name, EnvironGetenv("PATH", FALSE));
-
-last_resort:
-    /* last resort : see if the executable is in the current directory. This is
-       possible if it comes from a exec*() call. */
-    if (0 == stat(exe_name,&theStats))
+    return_size=MultiByteToWideChar(CP_ACP,0,real_path,-1,NULL,0);
```
janvorli: A nit - please add spaces after commas and around the `=` operator.

parjong:
@dotnet-bot test Windows_NT arm Cross Debug Build please.

src/pal/src/init/pal.cpp
```diff
@@ -1150,269 +1146,66 @@ Notes 2:
...
 
-            return return_value;
-        }
-        else
-        {
-            ERROR("found %s in current directory, but it isn't executable!\n",
-                  exe_name);
-        }                                                                   
+    return_value = reinterpret_cast<LPWSTR>(InternalMalloc((return_size*sizeof(WCHAR))));
+    if ( NULL == return_value )
```
janvorli: A nit - could you please unify spacing after the opening and before the closing parentheses in your change to no spaces?

src/pal/src/init/pal.cpp
```diff
@@ -1150,269 +1146,66 @@ Notes 2:
     This doesn't handle the case of directories with the desired name
     (and directories are usually executable...)
 --*/
-static LPWSTR INIT_FindEXEPath(LPCSTR exe_name)
+static LPWSTR INIT_EXEPath(LPCSTR exe_path)
```
janvorli: A nit - INIT_NormalizeEXEPath seems like a better name to me.

parjong: How about INIT_ConvertEXEPath? As I understand, the current implementation converts LPCSTR into LPWSTR.

janvorli: Ah, right, then your new name is better.

parjong:
@janvorli PR is updated. Please take a look.

src/pal/src/init/pal.cpp
```diff
@@ -1150,269 +1146,61 @@ Notes 2:
     This doesn't handle the case of directories with the desired name
```
janvorli: This part of the comment doesn't seem to make sense anymore.

src/pal/src/init/pal.cpp
```diff
@@ -1130,14 +1130,10 @@ static LPWSTR INIT_FormatCommandLine (int argc, const char * const *argv)
 
 /*++
 Function:
-  INIT_FindEXEPath
-
-Abstract:
-    Determine the full, canonical path of the current executable by searching
-    $PATH.
+  INIT_ConvertEXEPath
```
janvorli: Could you please add "Abstract:" with short description of what the function does? It was present for the original function.

src/pal/src/init/pal.cpp
```diff
@@ -1150,269 +1146,61 @@ Notes 2:
...
-
-        /* path_ptr is NULL if there's no ':' after this directory */
-        cur_dir=path_ptr;
+        ERROR("realpath() failed!\n");
+        return NULL;
     }
 
-    free(env_path);
-    TRACE("No %s found in $PATH (%s)\n", exe_name, EnvironGetenv("PATH", FALSE));
+    return_size = MultiByteToWideChar(CP_ACP,0,real_path,-1,NULL,0);
```
janvorli: A nit - could you please add spaces after the commas here too?

src/pal/src/init/pal.cpp
```diff
@@ -1150,269 +1146,61 @@ Notes 2:
...
-        TRACE("last resort failed : executable %s is not in the current "
-              "directory\n",exe_name);
+            TRACE("full path to executable is %s\n", real_path.GetString());
+        }
     }
 
-    ERROR("executable %s not found anywhere!\n", exe_name);
-    return NULL;
+    return return_value;
 #else // !__APPLE__
```
janvorli: I have just noticed that we also have this code for `__APPLE__` here. Could you please remove it and keep just the code you've updated for all platforms?

PR #10350, opened 2017-03-21
Add /silent option to crossgen

BruceForstall:
This sets the NGenOptions.fSilent flag, which prevents displaying
the final output message.

This is useful for JIT asm diff generation.


BruceForstall:
@jkotas PTAL

Issue #10351, opened 2017-03-21
Move CoreCLR OSX packages to 10.12


gkhanna79:
10.12 is the minimum platform supported for 2.0 for OSX. Corresponding CoreFX issue is https://github.com/dotnet/corefx/issues/17337

CC @weshaggard @ericstj @wtgodbe @chcosta

Issue #10353, opened 2017-03-21
Flip OSX SOS and TestBuild to 10.12 when CoreFX for 10.12 is published


gkhanna79:
See https://github.com/dotnet/coreclr/issues/10351

gkhanna79:
This is done.

PR #10354, opened 2017-03-21
Move OSX packages to use 10.12 as the RID

gkhanna79:
Fixes https://github.com/dotnet/coreclr/issues/10351

I will switch SOS and TestBuild to use 10.12 as the RID once CoreFX packages are published for the same. Tracking this with https://github.com/dotnet/coreclr/issues/10353.

@weshaggard 

CC @chcosta @wtgodbe @dagood


gkhanna79:
@weshaggard Are you fine with this change?

PR #10357, opened 2017-03-21
Move StringBuilder to shared partition

alexperovich:


alexperovich:
@dotnet-bot test Ubuntu x64 Checked Build and Test please (Java error in jenkins)

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -2128,11 +2252,11 @@ unsafe private void ReplaceInPlaceAtChunk(ref StringBuilder chunk, ref int index
         }
 
         /// <summary>
-        /// We have to prevent modification off the end of an array.
+        /// We have to prevent hackers from causing modification off the end of an array.
```
jkotas: Revert ... this comment change was likely done on purpose.

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -799,6 +798,141 @@ public StringBuilder AppendLine(string value)
...
+
+            return currentLength;
+        }
+
+        /// <summary>
+        /// Throw away the current contents in this StringBuffer and replace it with a new char[]
+        /// NOTE: The buffer in StringBuilder is *not* NULL-terminated.
+        /// This is only called from MCG code
+        /// </summary>
+        internal unsafe void ReplaceBuffer(char* newBuffer)
```
jkotas: This is redundant with:

```
MethodImplAttribute(MethodImplOptions.InternalCall)]
internal unsafe extern void ReplaceBufferInternal(char* newBuffer, int newLength);
```

FCalls, and I expect that the FCall reference will explode once it hits ProjectN.

It may be best to move these interop specific goo to CoreCLR/CoreRT specific files and deal with it separately.

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -2147,6 +2271,26 @@ unsafe private static void ThreadSafeCopy(char* sourcePtr, char[] destination, i
                 }
             }
         }
+
+        unsafe private static void ThreadSafeCopy(char[] source, int sourceIndex, char* destinationPtr, int destinationIndex, int count)
```
danmosemsft: I think this should be 'private static unsafe void'

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -1855,9 +1981,7 @@ public StringBuilder Replace(String oldValue, String newValue, int startIndex, i
                         replacements = new int[5];
                     else if (replacementsCount >= replacements.Length)
                     {
-                        int[] newArray = new int[replacements.Length * 3 / 2 + 4];     // grow by 1.5X but more in the begining
-                        Array.Copy(replacements, newArray, replacements.Length);
-                        replacements = newArray;
+                        Array.Resize(ref replacements, replacements.Length * 3 / 2 + 4); // grow by 1.5X but more in the begining
```
danmosemsft: nit typo 'begining'

src/mscorlib/shared/System/Text/StringBuilder.cs
danmosemsft: I am not sure what MCG is, mentioned in the comment, but it seems to be special to CoreRT or PN and not relevant to CoreCLR.

danmosemsft:
@dotnet-bot test linux arm emulator cross debug build (resx break earlier)
@dotnet-bot test linux arm emulator cross release build (resx break earlier)

src/mscorlib/System.Private.CoreLib.csproj
```diff
@@ -323,7 +323,7 @@
     <Compile Include="$(BclSourcesRoot)\System\String.Searching.cs" />
     <Compile Include="$(BclSourcesRoot)\System\StringComparer.cs" />
     <Compile Include="$(BclSourcesRoot)\System\StringComparison.cs" />
-    <Compile Include="$(BclSourcesRoot)\System\Text\StringBuilder.cs" />
+    <Compile Include="$(BclSourcesRoot)\System\Text\StringBuilder.CoreClr.cs" />
```
jkotas: Nit: It is `*.CoreCLR.cs` for all other similar files.

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -2147,6 +2121,26 @@ unsafe private static void ThreadSafeCopy(char* sourcePtr, char[] destination, i
                 }
             }
         }
+
+        private static unsafe void ThreadSafeCopy(char[] source, int sourceIndex, char* destinationPtr, int destinationIndex, int count)
```
jkotas: It is weird that this overloads takes destinationIndex. It should just take destinationPtr and the caller should add destinationIndex to destinationPtr - to be consistent with the other overload that takes the source as pointer.

jkotas:
LGTM otherwise.

src/mscorlib/shared/System/Text/StringBuilder.cs
alexperovich: It appears that this overload was only used in the CoreRt specific stuff, so I moved it there.

PR #10365, opened 2017-03-22
Enable Portable Windows RID

gkhanna79:
This change will produce packages with portable RID (e.g. win-x64).

@weshaggard PTAL

CC @RussKeldorph @wtgodbe @chcosta

### DIFF
--- a/config.json
+++ b/config.json
@@ -451,7 +451,7 @@
           }
         },
         "portable": {
-          "description": "Skips restoring packages during build.",
+          "description": "Triggers a portable build.",
           "settings": {
             "PortableBuild": "1"
           }
--- a/config.json
+++ b/config.json
@@ -210,6 +210,12 @@
       "values": [ "x86", "x64", "arm", "arm64" ],
       "defaultValue": "${CPUArch}"
     },
+    "PortableBuild": {
+      "description": "Indicates if this is a portable build.",
+      "valueType": "property",
+      "values": [ "0", "1"],
+      "defaultValue": "0"
+    },
     "Rebuild": {
       "description": "Specifies rebuild target.",
       "valueType": "target",
@@ -442,6 +448,12 @@
           "description": "Skips restoring packages during build.",
           "settings": {
             "RestoreDuringBuild": false
+          }
+        },
+        "portable": {
+          "description": "Skips restoring packages during build.",
+          "settings": {
+            "PortableBuild": "1"
           }
         },
         "BuildTarget": {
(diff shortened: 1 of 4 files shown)


weshaggard:
cc @ericstj

src/.nuget/dir.props
```diff
@@ -61,9 +61,13 @@
...
       <PropertyGroup>
-        <PackageRID Condition="'$(ArchGroup)' == 'x86' OR '$(ArchGroup)' == 'x64'">win7-$(ArchGroup)</PackageRID>
-        <PackageRID Condition="'$(ArchGroup)' == 'arm'">win8-$(ArchGroup)</PackageRID>
-        <PackageRID Condition="'$(ArchGroup)' == 'arm64'">win10-$(ArchGroup)</PackageRID>
+        <RIDPlatform>win7</RIDPlatform>
+        <RIDPlatform Condition="'$(ArchGroup)' == 'arm'">win8</RIDPlatform>
+        <RIDPlatform Condition="'$(ArchGroup)' == 'arm64'">win10</RIDPlatform>
+        
+        <!-- Set the platform part of the RID if we are doing a portable build -->
+        <RIDPlatform Condition="'$(PortableBuild)' == '1'">win</RIDPlatform>
```
weshaggard: So even though we aren't going to use the versionless rid in the host fallback we still plan to switch to versionless rids in our packages?

config.json
```diff
@@ -444,6 +450,12 @@
             "RestoreDuringBuild": false
           }
         },
+        "portable": {
+          "description": "Skips restoring packages during build.",
```
weshaggard: description isn't correct.

src/.nuget/dir.props
gkhanna79: Yes, the host needs to fallback from the latest RID since external developer authored packages will not be following this process (to start with). However, we should be transitioning to the base RID.

config.json
gkhanna79: Fixed.

weshaggard:
Is there a reason you are doing this conditionally? Is it just until we verify and push the packages through? Or do you intend to ship both sets?

gkhanna79:
This is being unconditionally until we decide we no longer are going to build RID specific packages (which, semantically, is equivalent to building distro specific packages).

gkhanna79:
BTW, I have changes (locally) to build portable OSX as well and need to test portable Linux. Will be updating the PR today for it.

weshaggard:
So I assume we need similar changes for corefx as well?

gkhanna79:
Yes, we need to @weshaggard.

PR #10371, opened 2017-03-22
Fix several devirtualization issues

### LINKED ISSUE #10311: Assertion failure: Consistency check failed: FAILED: !implSlot.IsNull() && "Valid method implementation was not found."
CoreFx tests with Checked runtime

**Sample CoreFx build/test Log**
https://ci.dot.net/job/dotnet_coreclr/job/master/job/jitstress/job/x64_checked_windows_nt_corefx_baseline/122/consoleFull

**Sample failure output**
```
2:26:46   CORECLR! CHECK::Trigger + 0x2C4 (0x00007ff8`0418ea6c)
22:26:46   call D:\j\workspace\x64_checked_w---a7bd363e\_\fx\bin\testhost\netcoreapp-Windows_NT-Release-x64\\dotnet.exe xunit.console.netcore.exe System.Xml.RW.WrappedReader.Tests.dll  -xml testResults.xml -notrait Benchmark=true -notrait category=nonnetcoreapptests  -notrait category=IgnoreForCI -notrait category=OuterLoop -notrait category=failing -notrait category=nonwindowstests
22:26:46   CORECLR! VirtualCallStubManager::Resolver + 0x1594 (0x00007ff8`045b37ec)
22:26:46   CORECLR! VirtualCallStubManager::GetTarget + 0x37E (0x00007ff8`045aaeee)
22:26:46   CORECLR! MethodTable::GetMethodDescForInterfaceMethod + 0x79A (0x00007ff8`046d7f32)
22:26:46   CORECLR! resolveVirtualMethodHelper + 0x2DC (0x00007ff8`0447b7a0)
22:26:46   CORECLR! CEEInfo::resolveVirtualMethod + 0x175 (0x00007ff8`0447b425)
22:26:46   CLRJIT! Compiler::impDevirtualizeCall + 0x414 (0x00007ff8`02d53640)
22:26:46   CLRJIT! Compiler::impImportCall + 0x1D4F (0x00007ff8`02d62eaf)
22:26:46   CLRJIT! Compiler::impImportBlockCode + 0x5B5D (0x00007ff8`02d5baf9)
22:26:46   CLRJIT! Compiler::impImportBlock + 0xED (0x00007ff8`02d55475)
22:26:46       File: d:\j\workspace\x64_checked_w---a7bd363e\src\vm\virtualcallstub.cpp Line: 2278
22:26:46       Image: D:\j\workspace\x64_checked_w---a7bd363e\_\fx\bin\testhost\netcoreapp-Windows_NT-Release-x64\dotnet.exe
```

AndyAyersMS:
When doing interface devirtualization, if the object type is canonical,
ensure that the owner type is too. The jit may present a mixed set
when inlining a shared method into a non-shared method. Ideally the
jit would also be able to present exact object types in such cases but
currently it cannot guarantee this. Closes #10311.

Adjust contracts to address some contract violations seen in desktop
testing. Make the helper non-static and fold in some of the info that
was passed from the caller to bring the desktop and CoreCLR implementations
closer.

Disallow interface devirt if the method is final but the class is not
exact or final, since derived classes can still override final methods
when implementing interfaces.

Don't try and devirtualize interface calls from com objects.

Add some related test cases.

### DIFF
--- a/src/vm/jitinterface.cpp
+++ b/src/vm/jitinterface.cpp
@@ -8714,12 +8714,16 @@ void CEEInfo::getMethodVTableOffset (CORINFO_METHOD_HANDLE methodHnd,
 }
 
 /*********************************************************************/
-static CORINFO_METHOD_HANDLE resolveVirtualMethodHelper(MethodDesc* callerMethod,
-                                                        CORINFO_METHOD_HANDLE baseMethod,
-                                                        CORINFO_CLASS_HANDLE derivedClass,
-                                                        CORINFO_CONTEXT_HANDLE ownerType)
+CORINFO_METHOD_HANDLE CEEInfo::resolveVirtualMethodHelper(CORINFO_METHOD_HANDLE baseMethod,
+                                                          CORINFO_CLASS_HANDLE derivedClass,
+                                                          CORINFO_CONTEXT_HANDLE ownerType)
 {
-    STANDARD_VM_CONTRACT;
+    CONTRACTL {
+        SO_TOLERANT;
+        THROWS;
+        GC_TRIGGERS;
+        MODE_PREEMPTIVE;
+    } CONTRACTL_END;
 
     MethodDesc* pBaseMD = GetMethod(baseMethod);
     MethodTable* pBaseMT = pBaseMD->GetMethodTable();
@@ -8747,6 +8751,15 @@ static CORINFO_METHOD_HANDLE resolveVirtualMethodHelper(MethodDesc* callerMethod
 
     if (pBaseMT->IsInterface())
     {
+
+#ifdef FEATURE_COMINTEROP
+        // Don't try and devirtualize com interface calls.
+        if (pDerivedMT->IsComObjectType())
+        {
+            return nullptr;
+        }
+#endif // FEATURE_COMINTEROP
+
         // Interface call devirtualization.
         //
         // We must ensure that pDerivedMT actually implements the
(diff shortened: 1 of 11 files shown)


AndyAyersMS:
@jkotas PTAL. This is preliminary but seems to fix issues where the jit doesn't have exact types.
cc @dotnet/jit-contrib 

Not real happy with having to drop the final method check from interface call devirt, but right now I can't tell whether a method is an explicit interface implementation or an inherited one that happens to be marked sealed (the latter is probably rare but possible), and so relying just on the final bit on the method can lead to the wrong devirtualization.

jkotas:
LGTM

src/vm/jitinterface.cpp
```diff
@@ -8760,7 +8773,17 @@ static CORINFO_METHOD_HANDLE resolveVirtualMethodHelper(MethodDesc* callerMethod
         // safely devirtualize.
         if (ownerType != nullptr)
         {
-            pDevirtMD = pDerivedMT->GetMethodDescForInterfaceMethod(GetTypeFromContext(ownerType), pBaseMD);
+            TypeHandle OwnerClsHnd = GetTypeFromContext(ownerType);
+            MethodTable* pOwnerMT = OwnerClsHnd.GetMethodTable();
+
+            // If the derived class is a shared class, make sure the
+            // owner class is too.
+            if (pDerivedMT->IsSharedByGenericInstantiations())
```
jkotas: Took me some time to convince myself that this should be ok ...

AndyAyersMS: Agree this is confusing. The jit isn't given or getting exact types in many cases. But when inlining a shared method into an unshared one, it ends up with the exact method type. So it can end up with an mixed view of types when devirtualizing a call in an inlinee.

For virtual calls this mixed view of types didn't matter since an exact type and its canonical type have the same vtable.

For invariant interface calls the mixed view likewise should not have mattered since the instance types do not figure in target resolution. I suspect one could add an equivalence check somewhere in the interface call resolution path that could fix the mixed case. However I was reluctant to make any changes there. Hence this fix.

By using shared types, we may lose the ability to devirtualize variant interface call since we can no longer reason about the relationships of the instance types. However it turns out we'd already lost those with the `CanCastToInterface` call above.

So the only diffs from this change are from no longer devirtualizing interface calls where the target is a final method; the changes on the VM side don't alter the jit's codegen.

Longer term I hope to selectively uplevel the types visible to the jit.

PR #10372, opened 2017-03-22
Switch coreclr corelib to use resx

alexperovich:


alexperovich:
@jkotas this change should fix the GenerateResource problem.

jkotas:
Thanks!

ViktorHofer:
Thanks to @tarekgh we probably found the culprit:

```
public virtual String EncodingName
{
get
{
return Environment.GetResourceString("Globalization.cp_" + m_codePage.ToString());
}
} 
so the failures related reading resources
```

EncodingName gets called e.g. here:

```
[Theory]
        [MemberData(nameof(Encodings_TestData))]
        public void EncodingName(ASCIIEncoding encoding)
        {
            Assert.NotEmpty(encoding.EncodingName); // US-ASCII in en-US
        }
```

Tools-Override/resources.targets
```diff
@@ -0,0 +1,76 @@
+<?xml version="1.0" encoding="utf-8"?>
```
weshaggard: Is the intent to always override this or is it to move this update into BuildTools?

alexperovich: I think this update should be moved into buildtools, but I believe there is a msbuild issue that needs to be fixed. I don't have a machine I can repro the task exec issue on though. MSBuild was trying to find a CLR v4 runtime to run the task on and I think this came from the defaulting logic here: https://github.com/Microsoft/msbuild/blob/6851538897f5d7b08024a6d8435bc44be5869e53/src/Build/Instance/TaskFactories/AssemblyTaskFactory.cs#L369

weshaggard: OK please file a tracking issue for moving to BuildTools then.

danmosemsft: @alexperovich I can't remember much about this, but <DisableOutOfProcTaskHost>true</> may be part of the solution.

alexperovich: dotnet/buildtools#1402 filed

Tools-Override/resources.targets
```diff
@@ -0,0 +1,76 @@
...
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <UsingTask TaskName="GenerateResourcesCode" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll" />
+
+  <PropertyGroup>
+    <GenerateResourceMSBuildRuntime>CurrentRuntime</GenerateResourceMSBuildRuntime>
```
weshaggard: @alexperovich I'm looking to merge this file into BuildTools can you tell me what these 2 properties are for? I don't seem the used anywhere.

alexperovich: These tell the msbuild targets here: https://github.com/Microsoft/msbuild/blob/master/src/Tasks/Microsoft.Common.CurrentVersion.targets#L2885 to use the current runtime and architecture for transforming resx -> resources. By default msbuild tries to spin up an out-of-process task host for the project's runtime and architecture so the resources are produced using the same runtime/architecture that the application is running as. For our resources this doesn't matter because they are all strings.

PR #10376, opened 2017-03-22
[x86/Linux] 16-byte aligned TheUMEntryPrestub

parjong:
This commit revises TheUMEntryPrestub to align its stack frame and check alignment before invoking TheUMEntryPrestubWorker.


parjong:
@janvorli Please take a look.

Issue #10139, opened 2017-03-13
[x86/Linux] ReadyToRun related features are not implemented, yet.


ragmani:
The testcase "readytorun/mainv1/mainv1.sh" is failed.
```
FAILED   - [ 7025 :    6 sec ] readytorun/mainv1/mainv1.sh
               BEGIN EXECUTION

               Assert failure(PID 22714 [0x000058ba], Thread: 22714 [0x58ba]): false && "ZapUnwindData::GetAlignment"
                   File: /home/parjong/projects/dotnet/coreclr/src/zap/zapcode.cpp Line: 1249
                   Image: /root/dotnet-test/coreclr_unittest_20170217_164016_x86_x86-emu/coreclr-unittest-170217-074046/overlay/crossgen

               Microsoft (R) CoreCLR Native Image Generator - Version 4.5.22220.0
               Copyright (c) Microsoft Corporation.  All rights reserved.

               ./mainv1.sh: line 98: 22714 Aborted                 (core dumped) $CORE_ROOT/crossgen -readytorun -platform_assemblies_paths $CORE_ROOT:`pwd` -out test.ni.dll test.dll

               Assert failure(PID 22782 [0x000058fe], Thread: 22782 [0x58fe]): false && "ZapUnwindData::GetAlignment"
                   File: /home/parjong/projects/dotnet/coreclr/src/zap/zapcode.cpp Line: 1249
                   Image: /root/dotnet-test/coreclr_unittest_20170217_164016_x86_x86-emu/coreclr-unittest-170217-074046/overlay/crossgen

               Microsoft (R) CoreCLR Native Image Generator - Version 4.5.22220.0
               Copyright (c) Microsoft Corporation.  All rights reserved.

               ./mainv1.sh: line 99: 22782 Aborted                 (core dumped) $CORE_ROOT/crossgen -readytorun -platform_assemblies_paths $CORE_ROOT:`pwd` -out fieldgetter.ni.dll fieldgetter.dll

               Assert failure(PID 22798 [0x0000590e], Thread: 22798 [0x590e]): false && "ZapUnwindData::GetAlignment"
                   File: /home/parjong/projects/dotnet/coreclr/src/zap/zapcode.cpp Line: 1249
                   Image: /root/dotnet-test/coreclr_unittest_20170217_164016_x86_x86-emu/coreclr-unittest-170217-074046/overlay/crossgen

               Microsoft (R) CoreCLR Native Image Generator - Version 4.5.22220.0
[...]

PR #10378, opened 2017-03-22
[x86/Linux] add three functions for _X86_TARGET_

ragmani:
There is no functions for _X86_TARGET_.
  - ZapUnwindData::GetAlignment()
  - DWORD ZapUnwindData::GetSize()
  - void ZapUnwindData::Save(ZapWriter * pZapWriter)

### DIFF
--- a/src/vm/codeman.cpp
+++ b/src/vm/codeman.cpp
@@ -988,7 +988,7 @@ PTR_VOID GetUnwindDataBlob(TADDR moduleBase, PTR_RUNTIME_FUNCTION pRuntimeFuncti
 #elif defined(_TARGET_X86_) && defined(FEATURE_PAL)
     PTR_UNWIND_INFO pUnwindInfo(dac_cast<PTR_UNWIND_INFO>(moduleBase + RUNTIME_FUNCTION__GetUnwindInfoAddress(pRuntimeFunction)));
 
-    *pSize = ALIGN_UP(sizeof(UNWIND_INFO), sizeof(BYTE));
+    *pSize = ALIGN_UP(sizeof(UNWIND_INFO), sizeof(DWORD));
 
     return pUnwindInfo;
 
--- a/src/vm/codeman.cpp
+++ b/src/vm/codeman.cpp
@@ -988,7 +988,7 @@ PTR_VOID GetUnwindDataBlob(TADDR moduleBase, PTR_RUNTIME_FUNCTION pRuntimeFuncti
 #elif defined(_TARGET_X86_) && defined(FEATURE_PAL)
     PTR_UNWIND_INFO pUnwindInfo(dac_cast<PTR_UNWIND_INFO>(moduleBase + RUNTIME_FUNCTION__GetUnwindInfoAddress(pRuntimeFunction)));
 
-    *pSize = ALIGN_UP(sizeof(ULONG), sizeof(DWORD));
+    *pSize = ALIGN_UP(sizeof(UNWIND_INFO), sizeof(BYTE));
 
     return pUnwindInfo;
 
--- a/src/vm/codeman.cpp
+++ b/src/vm/codeman.cpp
@@ -985,6 +985,13 @@ PTR_VOID GetUnwindDataBlob(TADDR moduleBase, PTR_RUNTIME_FUNCTION pRuntimeFuncti
 
     return pUnwindInfo;
 
+#elif defined(_TARGET_X86_) && defined(FEATURE_PAL)
+    PTR_UNWIND_INFO pUnwindInfo(dac_cast<PTR_UNWIND_INFO>(moduleBase + RUNTIME_FUNCTION__GetUnwindInfoAddress(pRuntimeFunction)));
+
+    *pSize = ALIGN_UP(sizeof(ULONG), sizeof(DWORD));
+
+    return pUnwindInfo;
+
 #elif defined(_TARGET_ARM_)
 
     // if this function uses packed unwind data then at least one of the two least significant bits
--- a/src/zap/zapcode.cpp
+++ b/src/zap/zapcode.cpp
@@ -1206,7 +1206,7 @@ void ZapUnwindData::Save(ZapWriter * pZapWriter)
 
 UINT ZapUnwindData::GetAlignment()
 {
-    return sizeof(ULONG);
+    return sizeof(BYTE);
 }
 
 DWORD ZapUnwindData::GetSize()
@@ -1222,8 +1222,6 @@ void ZapUnwindData::Save(ZapWriter * pZapWriter)
 
     PVOID pData = GetData();
     DWORD dwSize = GetBlobSize();
-
-    UNWIND_INFO * pUnwindInfo = (UNWIND_INFO *)pData;
 
     pZapWriter->Write(pData, dwSize);
 }
--- a/src/zap/zapcode.cpp
+++ b/src/zap/zapcode.cpp
@@ -1202,7 +1202,7 @@ void ZapUnwindData::Save(ZapWriter * pZapWriter)
 #endif //REDHAWK
 }
 
-#elif defined(_TARGET_X86_)
+#elif defined(_TARGET_X86_) && defined(FEATURE_PAL)
 
 UINT ZapUnwindData::GetAlignment()
 {
@@ -1212,14 +1212,6 @@ UINT ZapUnwindData::GetAlignment()
 DWORD ZapUnwindData::GetSize()
 {
     DWORD dwSize = ZapBlob::GetSize();
-
-#ifndef REDHAWK
-    // Add space for personality routine, it must be 4-byte aligned.
-    // Everything in the UNWIND_INFO has already had its size included in size
-    dwSize = AlignUp(dwSize, sizeof(ULONG));
-
-    dwSize += sizeof(ULONG);
-#endif //REDHAWK
 
     return dwSize;
 }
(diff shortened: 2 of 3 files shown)


ragmani:
@parjong This PR is from #10139

src/zap/zapcode.cpp
```diff
@@ -1202,6 +1202,49 @@ void ZapUnwindData::Save(ZapWriter * pZapWriter)
...
+{
+    DWORD dwSize = ZapBlob::GetSize();
+
+#ifndef REDHAWK
+    // Add space for personality routine, it must be 4-byte aligned.
+    // Everything in the UNWIND_INFO has already had its size included in size
+    dwSize = AlignUp(dwSize, sizeof(ULONG));
+
+    dwSize += sizeof(ULONG);
+#endif //REDHAWK
```
parjong: While working on #8889, I overhear that x86 has no personality routine space (please refer to ``reservePersonalityRoutineSpace``), but I'm not sure whether the line should be aligned with that, or not.

@jkotas @janvorli Could you correct me if there is something wrong?

jkotas: Yes, this should do the same thing as `reservePersonalityRoutineSpace` in VM

parjong:

ragmani:
I modified it as removing personality routine and succeeded in generating to ni with readytorun option.

ragmani:
I confirmed that simple ni assembly runs normally.
```
root@03c7d4352d95:/home/jang/DotNet/enable_x86/work# /home/jang/DotNet/enable_x86/work/ragmani/overlay.debug/corerun HelloDotNetCoreCS.dll 
Hello World
root@03c7d4352d95:/home/jang/DotNet/enable_x86/work# /home/jang/DotNet/enable_x86/work/ragmani/overlay.debug/corerun HelloDotNetCoreCS.ni.exe 
Hello World
```

src/zap/zapcode.cpp
```diff
@@ -1202,6 +1202,32 @@ void ZapUnwindData::Save(ZapWriter * pZapWriter)
...
+}
+
+void ZapUnwindData::Save(ZapWriter * pZapWriter)
+{
+    ZapImage * pImage = ZapImage::GetImage(pZapWriter);
+
+    PVOID pData = GetData();
+    DWORD dwSize = GetBlobSize();
+
+    UNWIND_INFO * pUnwindInfo = (UNWIND_INFO *)pData;
```
jkotas: This is unnecessary - pUnwindInfo is not used.

src/vm/codeman.cpp
```diff
@@ -985,6 +985,13 @@ PTR_VOID GetUnwindDataBlob(TADDR moduleBase, PTR_RUNTIME_FUNCTION pRuntimeFuncti
 
     return pUnwindInfo;
 
+#elif defined(_TARGET_X86_) && defined(FEATURE_PAL)
+    PTR_UNWIND_INFO pUnwindInfo(dac_cast<PTR_UNWIND_INFO>(moduleBase + RUNTIME_FUNCTION__GetUnwindInfoAddress(pRuntimeFunction)));
+
+    *pSize = ALIGN_UP(sizeof(ULONG), sizeof(DWORD));
```
jkotas: This is not correct (it will be always 4).

src/zap/zapcode.cpp
```diff
@@ -1202,6 +1202,32 @@ void ZapUnwindData::Save(ZapWriter * pZapWriter)
 #endif //REDHAWK
 }
 
+#elif defined(_TARGET_X86_) && defined(FEATURE_PAL)
+
+UINT ZapUnwindData::GetAlignment()
+{
+    return sizeof(ULONG);
```
jkotas: Can this be just sizeof(BYTE) ? The JIT case does not seem to have any extra alignment. It can be same here.

ragmani: chagned to BYTE in here and  _GetUnwindDataBlob_ function.

src/vm/codeman.cpp
ragmani: I modified to _"*pSize = ALIGN_UP(sizeof(UNWIND_INFO), sizeof(BYTE));"_

src/zap/zapcode.cpp
ragmani: I remoted it.

src/vm/codeman.cpp
janvorli: Aligning to sizeof(BYTE) doesn't make sense, it would always return the first parameter of the macro unchanged (align ensures that the value returned by the macro is a multiple of the second parameter and sizeof(BYTE) is 1 and all numbers are multiple of 1)

src/vm/codeman.cpp
```diff
@@ -985,6 +985,13 @@ PTR_VOID GetUnwindDataBlob(TADDR moduleBase, PTR_RUNTIME_FUNCTION pRuntimeFuncti
 
     return pUnwindInfo;
 
+#elif defined(_TARGET_X86_) && defined(FEATURE_PAL)
+    PTR_UNWIND_INFO pUnwindInfo(dac_cast<PTR_UNWIND_INFO>(moduleBase + RUNTIME_FUNCTION__GetUnwindInfoAddress(pRuntimeFunction)));
+
+    *pSize = ALIGN_UP(sizeof(UNWIND_INFO), sizeof(BYTE));
```
janvorli: The first argument to ALIGN_UP is correct now, but the second should be sizeof(DWORD)

janvorli:

janvorli:

parjong:
@jkotas @janvorli Could you please take a look? I overhear that this PR resolves 2 readytorun test failures.

src/zap/zapcode.cpp
```diff
@@ -1202,6 +1202,30 @@ void ZapUnwindData::Save(ZapWriter * pZapWriter)
 #endif //REDHAWK
 }
 
+#elif defined(_TARGET_X86_) && defined(FEATURE_PAL)
+
+UINT ZapUnwindData::GetAlignment()
+{
+    return sizeof(BYTE);
```
janvorli: This should be sizeof(ULONG) like on the other platforms.

src/zap/zapcode.cpp
```diff
@@ -1202,6 +1202,30 @@ void ZapUnwindData::Save(ZapWriter * pZapWriter)
...
+}
+
+DWORD ZapUnwindData::GetSize()
+{
+    DWORD dwSize = ZapBlob::GetSize();
```
janvorli: Please add `dwSize = AlignUp(dwSize, sizeof(ULONG));` here, like we use on other platforms.

src/zap/zapcode.cpp
```diff
@@ -1202,6 +1202,30 @@ void ZapUnwindData::Save(ZapWriter * pZapWriter)
...
+}
+
+void ZapUnwindData::Save(ZapWriter * pZapWriter)
+{
+    ZapImage * pImage = ZapImage::GetImage(pZapWriter);
+
+    PVOID pData = GetData();
+    DWORD dwSize = GetBlobSize();
+
+    pZapWriter->Write(pData, dwSize);
```
janvorli: And finally here, I believe we should add the following after the write (again, like we do on other platforms)
```C++
    DWORD dwPad = AlignmentPad(dwSize, sizeof(DWORD));
    if (dwPad != 0)
        pZapWriter->WritePad(dwPad);
```

src/zap/zapcode.cpp
parjong: This change was introduced according to the following comment by @jkotas:
>Can this be just sizeof(BYTE) ? The JIT case does not seem to have any extra alignment. It can be same here.

Could you let us know the semantics of ``GetAlignment``? x86/Linux currently emits no unwind data, and thus ``sizeof(BYTE)`` also looks fine.

janvorli: Ok, I was comparing it just to the other platforms. I've missed that comment from @jkotas and he knows better than me the semantics here. So let's ignore all of my comments.

src/vm/codeman.cpp
```diff
@@ -985,6 +985,13 @@ PTR_VOID GetUnwindDataBlob(TADDR moduleBase, PTR_RUNTIME_FUNCTION pRuntimeFuncti
 
     return pUnwindInfo;
 
+#elif defined(_TARGET_X86_) && defined(FEATURE_PAL)
+    PTR_UNWIND_INFO pUnwindInfo(dac_cast<PTR_UNWIND_INFO>(moduleBase + RUNTIME_FUNCTION__GetUnwindInfoAddress(pRuntimeFunction)));
+
+    *pSize = ALIGN_UP(sizeof(UNWIND_INFO), sizeof(DWORD));
```
jkotas: This does not look right. There is no alignment when the unwind blob is getting produced, and so there should be no alignment here where it is getting consumed either.

You should see a crashes during GC with R2R images because of this bug.

jkotas: Maybe this should be identical to the !WIN64EXCEPTIONS path:

```
PTR_VOID GetUnwindDataBlob(TADDR moduleBase, PTR_RUNTIME_FUNCTION pRuntimeFunction, /* out */ SIZE_T * pSize)
{
    *pSize = 0;
    return dac_cast<PTR_VOID>(pRuntimeFunction->UnwindData + moduleBase);
}
```

janvorli: Hmm, I am confused here. The added alignment was my request, since we do align in a similar way on AMD64 too (see few lines above).

jkotas: The structure of the unwind info is different on x86. There is no actual isolated unwind info blob like on x64. The unwind info is part of GCInfo on x86. So mirroring x64 is not the right strategy.

janvorli: Hmm, I see, the size is used to get location of the GCInfo from the address of the unwind data blob, so returning zero size like the Windows x86 is the right thing to do here. I can see now that that's what all the callers of GetUnwindDataBlob expect for x86.

janvorli: @ragmani, @parjong I am sorry for causing the confusion here.

janvorli: Btw, it looks like we will need to modify the following functions to take into account the fact that there is no personality routine for x86 Linux:
NativeImageJitManager::IsFilterFunclet
ReadyToRunJitManager::IsFilterFunclet

parjong: @jkotas Should we set ``*pSize`` as 0? x86/Linux produces a unwind info that consists of  ``FunctionLength`` unlike x86/Windows. (from ``inc/win64unwind.h``).

jkotas: I see - this may be just `sizeof(UNWIND_INFO)` for clarity then.

parjong: #10750 is submitted for cleanup.

PR #10391, opened 2017-03-22
[Local GC] Opt-out of GC Stress for FEATURE_STANDALONE_GC

swgillespie:
GCStress requires a level of coordination between the EE and the VM that isn't appropriate (yet) for a standalone GC, so this PR opts-out of it. Any developer validating their changes to a standalone GC and wishing to run GCStress can do a non-standalone build and validate their changes that way until we re-enable it for standalone GC.

### DIFF
--- a/src/gc/gc.cpp
+++ b/src/gc/gc.cpp
@@ -208,8 +208,10 @@ void GCStatistics::AddGCStats(const gc_mechanisms& settings, size_t timeInMSec)
 
     if (is_induced (settings.reason))
         cntReasons[(int)reason_induced]++;
+#ifdef STRESS_HEAP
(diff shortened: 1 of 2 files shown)


swgillespie:
@dotnet-bot  test Windows_NT x64 Debug Build and Test

swgillespie:
@Maoni0 PTAL? Is this reasonable?

Maoni0:
chatted with Sean - we do want to make stress work. Currently stress has its shortcomings and it's not executed very efficiently. we should have a workitem that makes that happen without exposing internal GC mechanism on the interface.

swgillespie:
thanks!

PR #10398, opened 2017-03-22
Jump stubs

BruceForstall:


BruceForstall:
Converted internal document on jump stubs to Markdown and add to CoreCLR document directory.

@jkotas @kouvel @RussKeldorph PTAL

PR #10400, opened 2017-03-22
Change setBBProfileWeight so that we no longer multiply a profile counts by 100

briansull:
Fix fgAddSyncMethodEnterExit to properly set the IBC profile weight when we have profile data
In fgComputeEdgeWeights properly set fgCalledCount when we inserted a new first block using fgFirstBBisScratch
Fix fgOptimizeUncondBranchToSimpleCond to properly set the new BasicBlock's weight and flags

### DIFF
--- a/src/jit/block.h
+++ b/src/jit/block.h
@@ -545,8 +545,7 @@ struct BasicBlock : private LIR::Range
     void setBBProfileWeight(unsigned weight)
     {
         this->bbFlags |= BBF_PROF_WEIGHT;
-        // Check if the multiplication by BB_UNITY_WEIGHT will overflow.
-        this->bbWeight = (weight <= BB_MAX_WEIGHT / BB_UNITY_WEIGHT) ? weight * BB_UNITY_WEIGHT : BB_MAX_WEIGHT;
+        this->bbWeight = weight;
     }
 
     // modifyBBWeight -- same as setBBWeight, but also make sure that if the block is rarely run, it stays that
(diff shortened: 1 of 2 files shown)


briansull:
@dotnet/jit-contrib PTAL
@dpodder PTAL

AndyAyersMS:
Presumably the old "scaled up" version was intended to allow for some fractional counts. Do we no longer want or need this? Seems like this should warrant some discussion.

I'd like to understand better how this fits into the larger picture -- do you have an overall plan you can point me at?

Are you going to make compensating changes in code that compares weights to BB_MAX_WEIGHT or BB_VERY_HOT_WEIGHT? Note what the inliner does in impMakeDiscretionaryInlineObservations doesn't make much sense so for that case it may not matter, but for other cases....

briansull:
When we have IBC Profile weights we use the exact profile counts.

The displayed block weight are still displayed as fractional weights based upon the entry block being set to 1.00  (see BasicBlock::getBBWeight )

The value of BB_VERY_HOT_WEIGHT is only used in one place as was essentially wrong before as having only 3 calls to a method would result in a value of 300.  (The only use is by the 32-bit x86 JIT to 16-byte align method entry points fir Ngen code)

briansull:
I have observed an issue with inlining where it apparently assigns all block weights for blocks that it imports to be the same values as the inlined location.  And it sets BBF_PROF_WEIGHT on all of these blocks.  We will revisit that at some point in the future.

briansull:

briansull:

briansull:

PR #10401, opened 2017-03-22
replace the linear time algorithm from lclvars with the constant one.

### LINKED ISSUE #10314: crossgen timeout JIT\Regression\JitBlue\DevDiv_255294
There are 120000 basic blocks in this tests. In the jit mode CoreCLR enables MinOPT 
```
compiler.cpp
if ((unsigned)JitConfig.JitMinOptsCodeSize() < info.compILCodeSize)
{
  JITLOG((LL_INFO10, "IL Code Size exceeded, using MinOpts for method %s\n", info.compFullName));
  theMinOptsValue = true;
}
```
it hides the quadratic algorithm:
```
lclvars.cpp:lvaMarkLclRefs:
  // Lazy init (If a block is not dominated by any other block, we'll redo this every time, but it'll be fast)
  BlockSetOps::AssignNoCopy(this, lvaMarkRefsCurBlock->bbDoms, fgGetDominatorSet(lvaMarkRefsCurBlock));
  BlockSetOps::RemoveElemD(this, lvaMarkRefsCurBlock->bbDoms, fgFirstBB->bbNum);
```
Crossgen doesn't drop down to MinOpts and spends much time in lvaMarkLclRefs.
I suggest to create a new block flag  isDominatedByNonNormalEntry and calculate it in fgComputeDoms. The flag will allow to get rid of the quadratic algorithm in this place.

sandreenko:
It allows to answer the question: "Is block dominated by non normal
entry" - in O(1). 
It is asked per block and time complexity was O(N^2), where N is the number of blocks in IR.
The third commit fixes #10314

### DIFF
--- a/src/jit/block.h
+++ b/src/jit/block.h
@@ -438,6 +438,8 @@ struct BasicBlock : private LIR::Range
 #define BBF_CLONED_FINALLY_BEGIN 0x100000000 // First block of a cloned finally region
 #define BBF_CLONED_FINALLY_END 0x200000000   // Last block of a cloned finally region
 
+#define BBF_DOMINATED_BY_NOT_NORMAL_ENTRY 0x400000000 // Full set of dominators contains not normal entry.
+
 // Flags that relate blocks to loop structure.
 
 #define BBF_LOOP_FLAGS (BBF_LOOP_PREHEADER | BBF_LOOP_HEAD | BBF_LOOP_CALL0 | BBF_LOOP_CALL1)
@@ -1171,6 +1173,16 @@ public:
 
     void MakeLIR(GenTree* firstNode, GenTree* lastNode);
     bool IsLIR();
+
+    void SetDominatedByNotNormalEntryFlag()
+    {
+        bbFlags |= BBF_DOMINATED_BY_NOT_NORMAL_ENTRY;
+    }
+
+    bool IsDominatedByNotNormalEntryFlag()
+    {
+        return (bbFlags & BBF_DOMINATED_BY_NOT_NORMAL_ENTRY) != 0;
+    }
 };
 
 template <>
--- a/src/jit/block.h
+++ b/src/jit/block.h
@@ -438,7 +438,7 @@ struct BasicBlock : private LIR::Range
 #define BBF_CLONED_FINALLY_BEGIN 0x100000000 // First block of a cloned finally region
 #define BBF_CLONED_FINALLY_END 0x200000000   // Last block of a cloned finally region
 
-#define BBF_DOMINATED_BY_NOT_NORMAL_ENTRY 0x400000000 // Full set of dominators contains not normal entry.
+#define BBF_DOMINATED_BY_EXCEPTIONAL_ENTRY 0x400000000 // Block is dominated by exceptional entry.
 
 // Flags that relate blocks to loop structure.
 
@@ -880,11 +880,6 @@ struct BasicBlock : private LIR::Range
     unsigned bbDfsNum;   // The index of this block in DFS reverse post order
                          // relative to the flow graph.
 
-#if ASSERTION_PROP
-    // A set of blocks which dominate this one *except* the normal entry block. This is lazily initialized
-    // and used only by Assertion Prop, intersected with fgEnterBlks!
-#endif
-
     IL_OFFSET bbCodeOffs;    // IL offset of the beginning of the block
     IL_OFFSET bbCodeOffsEnd; // IL offset past the end of the block. Thus, the [bbCodeOffs..bbCodeOffsEnd)
                              // range is not inclusive of the end offset. The count of IL bytes in the block
@@ -1079,9 +1074,7 @@ struct BasicBlock : private LIR::Range
     GenTree*     FirstNonPhiDefOrCatchArgAsg();
 
     BasicBlock()
-        :
-        VARSET_INIT_NOCOPY(bbLiveIn, VarSetOps::UninitVal())
-        , VARSET_INIT_NOCOPY(bbLiveOut, VarSetOps::UninitVal())
+        : VARSET_INIT_NOCOPY(bbLiveIn, VarSetOps::UninitVal()), VARSET_INIT_NOCOPY(bbLiveOut, VarSetOps::UninitVal())
     {
     }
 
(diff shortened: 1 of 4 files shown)


sandreenko:
@dotnet-bot test Windows_NT x86 Debug Build

sandreenko:
@dotnet-bot  test Tizen armel Cross Debug Build

sandreenko:
It also fix a bug when local variables that live only in EH blocks without predecessors were not marked with the volatile hint.

sandreenko:
PTAL @RussKeldorph  @BruceForstall  @dotnet/jit-contrib.

BruceForstall:
Have you done asm diffs (core and desktop: all plat including SuperPMI)? Any diffs?

Do you have any measurements showing improvement?

BruceForstall:
test Windows_NT x64 throughput
test Ubuntu14.04 throughput

src/jit/lclvars.cpp
```diff
@@ -3434,23 +3434,9 @@ void Compiler::lvaMarkLclRefs(GenTreePtr tree)
...
-            BlockSetOps::AssignNoCopy(this, lvaMarkRefsCurBlock->bbDoms, fgGetDominatorSet(lvaMarkRefsCurBlock));
-            BlockSetOps::RemoveElemD(this, lvaMarkRefsCurBlock->bbDoms, fgFirstBB->bbNum);
-        }
-        assert(fgEnterBlksSetValid);
-        if (!BlockSetOps::IsEmptyIntersection(this, lvaMarkRefsCurBlock->bbDoms, fgEnterBlks))
-        {
-            varDsc->lvVolatileHint = 1;
-        }
+        SetVolatileHint(varDsc);
     }
```
BruceForstall: Is bbDoms no longer used? Can you remove it?

src/jit/block.h
```diff
@@ -438,6 +438,8 @@ struct BasicBlock : private LIR::Range
 #define BBF_CLONED_FINALLY_BEGIN 0x100000000 // First block of a cloned finally region
 #define BBF_CLONED_FINALLY_END 0x200000000   // Last block of a cloned finally region
 
+#define BBF_DOMINATED_BY_NOT_NORMAL_ENTRY 0x400000000 // Full set of dominators contains not normal entry.
+
```
BruceForstall: Instead of "not normal" can we call it "Exceptional"? e.g., BBF_DOMINATED_BY_EXCEPTIONAL_ENTRY, fgCompDominatedByExceptionalEntryBlocks(), etc.?

src/jit/lclvars.cpp
sandreenko: Yes, thanks, I didn't not notice that.

src/jit/block.h
sandreenko: Fixed, it is a bit unclear because we include call_finally block for ARM and I don't think that we could call it exceptional. I do not know should we mark all finally blocks or not with this flag.
Do we expect better performance if we SetVolatileHint on local variables that live in finally blocks?

BruceForstall: >  it is a bit unclear because we include call_finally block for ARM and I don't think that we could call it exceptional.

Well, you could say "Special" and define that somewhere :-)

> Do we expect better performance if we SetVolatileHint on local variables that live in finally blocks?

Not sure. Is it perf, or correctness?

src/jit/block.h
```diff
@@ -881,7 +883,6 @@ struct BasicBlock : private LIR::Range
 #if ASSERTION_PROP
     // A set of blocks which dominate this one *except* the normal entry block. This is lazily initialized
     // and used only by Assertion Prop, intersected with fgEnterBlks!
-    BlockSet bbDoms;
```
BruceForstall: You should define the entire `#if` here.

src/jit/block.h
sandreenko: varDsc->lvVolatileHint affects only perf.

sandreenko: @briansull maybe you know, do we want lvVolatileHint flag to be set in finally blocks?

briansull: Setting lvVolatileHint forces any LclVars that are referenced in finally or catch blocks to get stack allocated

sandreenko: It means that we want to SetDominatedByExceptionalEntryFlag for fgEnterBlks (it includes normal entry and blocks from EH table), but also for finally and special code EH.
Does it sound right?

sandreenko:
@dotnet-bot test  
OSX10.12 x64 Checked Build and Test 

sandreenko:
@BruceForstall ptal

src/jit/flowgraph.cpp
```diff
@@ -24931,3 +24934,26 @@ unsigned Compiler::fgMeasureIR()
 
     return nodeCount;
 }
+
+//------------------------------------------------------------------------
+// fgCompDominatedByExceptionalEntryBlocks: compute blocks that are
+// dominated by not normal entry.
+//
+void Compiler::fgCompDominatedByExceptionalEntryBlocks()
+{
```
BruceForstall: You could add, for safety:
```
assert(fgEnterBlksSetValid);
```

src/jit/flowgraph.cpp
```diff
@@ -24931,3 +24934,26 @@ unsigned Compiler::fgMeasureIR()
...
+    for (unsigned i = 1; i <= fgBBNumMax; ++i)
+    {
```
BruceForstall: You don't need to run this loop at all if fgEnterBlksSet contains exactly one element, since that one element will be the normal entry block (you could assert this).

src/jit/block.h
```diff
@@ -1171,6 +1161,16 @@ struct BasicBlock : private LIR::Range
...
+
+    void SetDominatedByExceptionalEntryFlag()
+    {
+        bbFlags |= BBF_DOMINATED_BY_EXCEPTIONAL_ENTRY;
+    }
+
+    bool IsDominatedByExceptionalEntryFlag()
+    {
+        return (bbFlags & BBF_DOMINATED_BY_EXCEPTIONAL_ENTRY) != 0;
+    }
```
sandreenko: So, the existing algorithm set this bit only on blocks that are dominated by a block from fgEnterBlks and this block is not normal entry.
So if any block was added after computing fgComputeEnterBlocksSet it was not reflected in fgEnterBlks and lclVars in this block were not marked with the hint.

sandreenko: We add new EH only during fgFindBasicBlocks or inlining. It happens before computing dominators.

src/jit/flowgraph.cpp
sandreenko: Fixed

src/jit/flowgraph.cpp
sandreenko: Fixed

PR #10404, opened 2017-03-23
porting desktop changes to enable up to 1024 CPUs

Maoni0:
@swgillespie @adityamandaleeka this is part of what you've reviewed that doesn't already exist in core.

### DIFF
--- a/src/gc/gc.cpp
+++ b/src/gc/gc.cpp
@@ -53,9 +53,13 @@ BOOL bgc_heap_walk_for_etw_p = FALSE;
 #define LOH_PIN_QUEUE_LENGTH 100
 #define LOH_PIN_DECAY 10
 
+#ifdef BIT64
 // Right now we support maximum 1024 procs - meaning that we will create at most
 // that many GC threads and GC heaps. 
 #define MAX_SUPPORTED_CPUS 1024
+#else
+#define MAX_SUPPORTED_CPUS 64
+#endif // BIT64
 
 #ifdef GC_CONFIG_DRIVEN
 int compact_ratio = 0;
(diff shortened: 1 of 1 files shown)


src/gc/gc.cpp
```diff
@@ -5082,28 +5082,28 @@ class heap_select
...
 uint8_t* heap_select::sniff_buffer;
 unsigned heap_select::n_sniff_buffers;
 unsigned heap_select::cur_sniff_index;
-uint8_t heap_select::proc_no_to_heap_no[MAX_SUPPORTED_CPUS];
-uint8_t heap_select::heap_no_to_proc_no[MAX_SUPPORTED_CPUS];
-uint8_t heap_select::heap_no_to_numa_node[MAX_SUPPORTED_CPUS];
-uint8_t heap_select::heap_no_to_cpu_group[MAX_SUPPORTED_CPUS];
-uint8_t heap_select::heap_no_to_group_proc[MAX_SUPPORTED_CPUS];
-uint8_t heap_select::numa_node_to_heap_map[MAX_SUPPORTED_CPUS+4];
+uint16_t heap_select::proc_no_to_heap_no[MAX_SUPPORTED_CPUS];
```
jkotas: Would it make sense to allocate this dynamically? This is 12kB of statics - most of it is unused in typical config.

Maoni0: I did want these to not be dynamically allocated because I need it to be fast. however, this really only applies to 64-bit - 32-bit doesn't support 1024 CPUs anyway. Since 32-bit windows do not support processor groups, I can just make MAX_SUPPORTED_CPUS 64. I dunno if 32-bit Linux support more than 64 but our code doesn't support it anyway.

Maoni0: I did want these to not be dynamically allocated because I need it to be fast. however, this really only applies to 64-bit - 32-bit doesn't support 1024 CPUs anyway. Since 32-bit windows do not support processor groups, I can just make MAX_SUPPORTED_CPUS 64. I dunno if 32-bit Linux support more than 64 but our code doesn't support it anyway.

PR #10407, opened 2017-03-23
Fix fallout from switching corelib to use resx

jkotas:
Fixes #https://github.com/dotnet/corefx/issues/17413

### DIFF
--- a/src/mscorlib/src/System/Threading/Tasks/TPLETWProvider.cs
+++ b/src/mscorlib/src/System/Threading/Tasks/TPLETWProvider.cs
@@ -24,7 +24,7 @@ namespace System.Threading.Tasks
     [EventSource(
         Name = "System.Threading.Tasks.TplEventSource",
         Guid = "2e5dba47-a3d2-4d16-8ee0-6671ffdcd7b5",
-        LocalizationResources = System.CoreLib.Name)]
+        LocalizationResources = "FxResources.System.Private.CoreLib.SR")]
     internal sealed class TplEtwProvider : EventSource
     {
         /// Used to determine if tasks should generate Activity IDs for themselves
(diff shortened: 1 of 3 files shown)


src/mscorlib/src/System/Threading/Tasks/TPLETWProvider.cs
```diff
@@ -24,7 +24,7 @@ namespace System.Threading.Tasks
     [EventSource(
         Name = "System.Threading.Tasks.TplEventSource",
         Guid = "2e5dba47-a3d2-4d16-8ee0-6671ffdcd7b5",
-        LocalizationResources = System.CoreLib.Name)]
+        LocalizationResources = "FxResources.System.Private.CoreLib.SR")]
```
danmosemsft: This apparently didn't cause any corefx tests to fail. Should we have an issue for that?

danmosemsft:
@ViktorHofer

src/mscorlib/src/System/Threading/Tasks/TPLETWProvider.cs
jkotas: It caused System.Diagnostics.Tracing tests to fail on my machine. That's why I have fixed it.

jkotas: And the System.Diagnostics.Tracing tests are failing in https://github.com/dotnet/corefx/pull/17419 as well. The failure is not showing up in the Jenkins summary - presumably because of there is a lot other failures.

From the full log - the test is crashing before printing anything:

```
20:27:11   call D:\j\workspace\windows_nt_de---4526f5ff\bin\testhost\netcoreapp-Windows_NT-Debug-x86\\dotnet.exe xunit.console.netcore.exe System.Diagnostics.Tracing.Tests.dll  -xml testResults.xml -notrait Benchmark=true -notrait category=nonnetcoreapptests  -notrait category=IgnoreForCI -notrait category=OuterLoop -notrait category=failing -notrait category=nonwindowstests
20:27:11   python DumplingHelper.py collect_dump 0 C:\Users\DOTNET~1\AppData\Local\Temp\CoreRunCrashDumps 
... the crash dump got uploaded ...
```

ViktorHofer:
Thanks @jkotas! I was also working on it, but compilation took far too long on my machine.

PR #10413, opened 2017-03-23
Add heuristic to trigger GC to finalize dead threads and clean up han…

### LINKED ISSUE #6602: Thread object's handles accumulate over time to a high number
A thread that is created allocates some handles in Thread::AllocHandles(). After it terminates, the managed thread object needs to be collected by a GC for the handles to be released. Some applications that are mostly idle but have a timer ticking periodically will cause a thread pool thread to be created on each tick, and over time the number of handles accumulates to a high value. Thread creation adds some memory pressure, but that does not force a GC until a sufficiently large handle count, and for some mostly idle apps, that can be very long.

Add a time-based heuristic that takes into account the number of threads that have terminated but have not yet been released by the GC, to trigger a GC and keep the number of open handles to a relatively low value.

kouvel:
…dles and memory

A thread that is started allocates some handles in Thread::AllocHandles(). After it terminates, the managed thread object needs to be collected by a GC for the handles to be released. Some applications that are mostly idle but have a timer ticking periodically will cause a thread pool thread to be created on each tick, since the thread pool preserves threads for a maximum of 20 seconds currently. Over time the number of handles accumulates to a high value. Thread creation adds some memory pressure, but that does not force a GC until a sufficiently large handle count, and for some mostly idle apps, that can be very long. The same issue arises with directly starting threads as well.

Fixes #6602:
- Track a dead thread count separately from the current dead thread count. This is the count that will contribute to triggering a GC, once it reaches a threshold. The count is tracked separately so that it may be reset to zero when a max-generation GC occurs, preventing dead threads that survive a GC due to references from contributing to triggering a GC again in this fashion.
- If the count exceeds a threshold, enumerate dead threads to see which GC generation the corresponding managed thread objects are in. If the duration of time since the last GC of the desired generation also exceeds a threshold, trigger a preferably non-blocking GC on the finalizer thread.
- Removed a couple of handles and some code paths specific to user-requested thread suspension, which is not supported on CoreCLR

### DIFF
--- a/src/gc/gcinterface.h
+++ b/src/gc/gcinterface.h
@@ -163,6 +163,7 @@ struct segment_info
 // one for the object header, and one for the first field in the object.
 #define min_obj_size ((sizeof(uint8_t*) + sizeof(uintptr_t) + sizeof(size_t)))
 
+#define min_finalizing_generation 1
 #define max_generation 2
 
 // The bit shift used to convert a memory address into an index into the
@@ -526,6 +527,9 @@ public:
     // throughout the VM.
     virtual HRESULT GarbageCollect(int generation = -1, BOOL low_memory_p = FALSE, int mode = collection_blocking) = 0;
 
+    // Gets the smallest GC generation that runs finalizers.
+    virtual unsigned GetMinFinalizingGeneration() = 0;
+
     // Gets the largest GC generation. Also used extensively throughout the VM.
     virtual unsigned GetMaxGeneration() = 0;
 
@@ -727,6 +731,7 @@ public:
     SVAL_DECL(uint32_t, gcHeapType);
 #endif
 
+    static const uint32_t minFinalizingGeneration;
     SVAL_DECL(uint32_t, maxGeneration);
 };
 
--- a/src/gc/gcinterface.h
+++ b/src/gc/gcinterface.h
@@ -163,7 +163,6 @@ struct segment_info
 // one for the object header, and one for the first field in the object.
 #define min_obj_size ((sizeof(uint8_t*) + sizeof(uintptr_t) + sizeof(size_t)))
 
-#define min_finalizing_generation 1
 #define max_generation 2
 
 // The bit shift used to convert a memory address into an index into the
@@ -527,9 +526,6 @@ public:
     // throughout the VM.
     virtual HRESULT GarbageCollect(int generation = -1, BOOL low_memory_p = FALSE, int mode = collection_blocking) = 0;
 
-    // Gets the smallest GC generation that runs finalizers.
-    virtual unsigned GetMinFinalizingGeneration() = 0;
-
     // Gets the largest GC generation. Also used extensively throughout the VM.
     virtual unsigned GetMaxGeneration() = 0;
 
@@ -731,7 +727,6 @@ public:
     SVAL_DECL(uint32_t, gcHeapType);
 #endif
 
-    static const uint32_t minFinalizingGeneration;
     SVAL_DECL(uint32_t, maxGeneration);
 };
 
--- a/src/vm/threads.cpp
+++ b/src/vm/threads.cpp
@@ -6012,7 +6012,7 @@ void ThreadStore::IncrementDeadThreadCountForGCTrigger()
         return;
     }
 
-    SIZE_T gcLastMilliseconds = gcHeap->GetLastGCStartTime(gcHeap->GetMaxGeneration());
+    SIZE_T gcLastMilliseconds = gcHeap->GetLastGCStartTime(max_generation);
     SIZE_T gcNowMilliseconds = gcHeap->GetNow();
     if (gcNowMilliseconds - gcLastMilliseconds < s_DeadThreadGCTriggerPeriodMilliseconds)
     {
@@ -6092,8 +6092,7 @@ void ThreadStore::TriggerGCForDeadThreadsIfNecessary()
     IGCHeap *gcHeap = GCHeapUtilities::GetGCHeap();
     _ASSERTE(gcHeap != nullptr);
     SIZE_T generationCountThreshold = static_cast<SIZE_T>(s_DeadThreadCountThresholdForGCTrigger) / 2;
-    SIZE_T newDeadThreadGenerationCounts[3] = {0};
-    int countedMaxGeneration = _countof(newDeadThreadGenerationCounts) - 1;
+    SIZE_T newDeadThreadGenerationCounts[max_generation + 1] = {0};
     {
         ThreadStoreLockHolder threadStoreLockHolder;
         GCX_COOP();
@@ -6115,20 +6114,21 @@ void ThreadStore::TriggerGCForDeadThreadsIfNecessary()
                 continue;
             }
 
-            int exposedObjectGeneration = min(countedMaxGeneration, static_cast<int>(gcHeap->WhichGeneration(exposedObject)));
+            int exposedObjectGeneration = gcHeap->WhichGeneration(exposedObject);
             SIZE_T newDeadThreadGenerationCount = ++newDeadThreadGenerationCounts[exposedObjectGeneration];
             if (exposedObjectGeneration > gcGenerationToTrigger && newDeadThreadGenerationCount >= generationCountThreshold)
             {
                 gcGenerationToTrigger = exposedObjectGeneration;
-                if (gcGenerationToTrigger >= countedMaxGeneration)
+                if (gcGenerationToTrigger >= max_generation)
                 {
                     break;
                 }
             }
         }
 
-        // Make sure that enough time has elapsed since the last GC of the desired generation. Otherwise, the memory pressure
-        // would be sufficient to trigger GCs automatically.
+        // Make sure that enough time has elapsed since the last GC of the desired generation. We don't want to trigger GCs
+        // based on this heuristic too often. Give it some time to let the memory pressure trigger GCs automatically, and only
+        // if it doesn't in the given time, this heuristic may kick in to trigger a GC.
         SIZE_T gcLastMilliseconds = gcHeap->GetLastGCStartTime(gcGenerationToTrigger);
         SIZE_T gcNowMilliseconds = gcHeap->GetNow();
         if (gcNowMilliseconds - gcLastMilliseconds < s_DeadThreadGCTriggerPeriodMilliseconds)
@@ -6153,7 +6153,7 @@ void ThreadStore::TriggerGCForDeadThreadsIfNecessary()
                 continue;
             }
 
-            if (gcGenerationToTrigger < countedMaxGeneration &&
+            if (gcGenerationToTrigger < max_generation &&
                 static_cast<int>(gcHeap->WhichGeneration(exposedObject)) > gcGenerationToTrigger)
             {
                 continue;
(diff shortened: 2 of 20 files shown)


kouvel:
I investigated whether it would be possible to delete handles and some unmanaged memory associated with a thread upon the thread exiting. This does not seem to be feasible due to races that would exist without the use of a lock to read thread state, which would be necessary on almost every Thread API. The current mechanism eliminates the possibility of the native Thread object and its handles from being deleted concurrently with a Thread API by associating their lifetime with the managed Thread object's finalizer.

So, going back to the original solution of adding a heuristic to trigger GC based on the number of dead threads accumulated since the last finalizing GC and time elapsed since the last finalizing GC.

@jkotas @Maoni0

src/gc/gcinterface.h
```diff
@@ -526,6 +527,9 @@ class IGCHeap {
     // throughout the VM.
     virtual HRESULT GarbageCollect(int generation = -1, BOOL low_memory_p = FALSE, int mode = collection_blocking) = 0;
 
+    // Gets the smallest GC generation that runs finalizers.
+    virtual unsigned GetMinFinalizingGeneration() = 0;
```
jkotas: I do not understand what this means or why you need it. The finalizers run even after gen0 GC:

```
using System;

class Program
{
    ~Program()
    {
        Console.WriteLine("Here");
    }

    static void foo()
    {
        GC.KeepAlive(new Program());
    }

    {
        GC.Collect();
        foo();
        GC.Collect(0);
        for (;;);
    }
}
```

src/vm/threads.cpp
```diff
@@ -6009,6 +6042,80 @@ void ThreadStore::TransferStartedThread(Thread *thread, BOOL bRequiresTSL)
...
+void ThreadStore::TriggerFinalizingGCIfNecessary()
+{
+    if (!m_TriggerFinalizingGC)
+    {
+        return;
+    }
+    m_TriggerFinalizingGC = false;
+
+    IGCHeap *gcHeap = GCHeapUtilities::GetGCHeap();
+    gcHeap->GarbageCollect(gcHeap->GetMinFinalizingGeneration(), FALSE, collection_non_blocking);
```
jkotas: If the waking threads have done just enough work to get them promoted to Gen2, this won't cleanup anything. Should this rather be full GC?

src/vm/threads.h
```diff
@@ -3885,8 +3908,10 @@ class Thread: public IUnknown
 
 private:
     // For suspends:
+#ifndef FEATURE_CORECLR // CoreCLR does not support user-requested thread suspension
```
jkotas: You can delete this. No need to add `FEATURE_CORECLR` ifdefs.

src/gc/gcinterface.h
```diff
@@ -727,6 +731,7 @@ class IGCHeap {
     SVAL_DECL(uint32_t, gcHeapType);
 #endif
 
+    static const uint32_t minFinalizingGeneration;
```
swgillespie: Ideally there won't be any static fields on `IGCHeap` - the two existing ones (`gcHeapType` and `maxGeneration`) will be getting removed soon. Can we avoid putting this on the interface itself? (Provided it's necessary, per Jan's comment)

src/vm/threads.cpp
kouvel: I had discussed this with @Maoni0 before, the concern was that the mostly idle apps that are running into this issue may have a large GC heap and this would cause those apps to do a full GC every period. Also, in all of the cases where I saw this issue manifest, the scenario was such that neither gen 1 nor gen 2 GCs were happening frequently enough. So doing a gen 1 GC should cover most of the cases that are running into this issue. On the other hand, as you said this would not cover thread objects that are promoted to gen 2. Thoughts on the tradeoff? Maybe it could trigger a gen 2 GC between a number of gen 1 GCs. Or if there is a way to see which generation an object is in, it could walk the dead threads' exposed objects and trigger a GC for the highest generation containing a dead thread (but that would include reachable dead threads too).

src/gc/gcinterface.h
kouvel: Hmm somehow I was under the other impression, looks like gen 0 is fine. I'll fix to specify 1 or max based on the other discussion. Doing only gen 0 GC probably would not cover enough cases.

src/gc/gcinterface.h
kouvel: I'll remove this and will just specify a number relative to 0 or max

src/vm/threads.cpp
jkotas: > Or if there is a way to see which generation an object is in,

WhichGeneration

> that would include reachable dead threads too

If you wanted to handle these better, you can have a flag on the thread that you will set once you have considered it once.

I would think that this needs to do Gen2 GCs at least sometimes if we do not ever want to hear about it again.

Maoni0: I thought you wanted this for the idling scenario where nothing is happening aside from the timer creating threads? If there's work going on, you'd have gen1 GCs (or sometimes gen2 GCs) happening anyway so you don't need to worry about this. So if you are triggering GCs, I'd think a gen1 GC would suffice 'cause all the previously created threads would be dead.

src/gc/gcinterface.h
Maoni0: you can just use (max_generation - 1).

src/vm/threads.cpp
kouvel: The work that's happening is not sufficient to automatically trigger a gen 2 GC (and usually not enough to trigger a gen 1 GC either), but in cases where a gen 1 GC is triggered automatically it's possible that threads that are still referenced get promoted to gen 2, but a gen 2 GC is not triggered frequently enough to clean up those threads, which can continue to increase the handle count as before.

Maoni0: Ok, I think this is different from the scenario I understood when we talked. The scenario I understood was the threads that you created before were all dead (except the current one) - nothing holding onto them - if something is holding onto them it means they are useful in some way and if that's the case, some work is in progress and GCs will probably be triggered anyway. But it sounds like this is not the case because you worry about a little bit of work is going on that's enough to trigger gen1's. I am hesitant to have it trigger gen2 GCs because then we might be back to square one where it's triggering too many gen2 GCs so we have to be careful about triggering gen2's. You could call WhichGeneration and see if the percentage of gen2 objects is high enough, then trigger a gen2 but I would experiment with this carefully.

kouvel: As far as the observed scenarios are concerned, a gen 1 GC would suffice. Slightly different scenarios may trigger gen 1 GCs as a result of medium-level work, while referencing created threads for a short enough amount of time that they are still garbage but are alive long enough to get promoted to gen 2. I think it's beneficial to not leave any room for this being a continuing problem in the future. I plan on using WhichGeneration to trigger precisely the generation of GC that would benefit for this purpose, so this shouldn't cause excessive GCs in any scenario.

Maoni0: okie that sounds good

kouvel:
Made all changes from above

kouvel:
@dotnet-bot test this please

kouvel:
Updated commit description

src/vm/threads.cpp
```diff
@@ -6009,6 +5984,182 @@ void ThreadStore::TransferStartedThread(Thread *thread, BOOL bRequiresTSL)
...
+    if (!g_fEEStarted) // required for FinalizerThread::EnableFinalization() below
+    {
+        return;
+    }
+
+    int gcGenerationToTrigger = 0;
+    SIZE_T newDeadThreadGenerationCounts[3] = {0};
+    int countedMaxGeneration = _countof(newDeadThreadGenerationCounts) - 1;
+    {
+        GCX_COOP();
```
jkotas: It looks dangerous to be switching to cooperating mode here after the thread is marked as dead. Can we just do the simple checks of the time and dead thread count thresholds here, kick the finalizer thread if they are hit, and do the more fine grained checks on the finalizer thread?

kouvel: Done. Also fixed to clear the count even if a GC is not triggered after scanning threads due to not enough time having elapsed since the last GC at the same generation, so that it would not scan threads every time a thread becomes dead after the count threshold is reached and before the time threshold is reached.

jkotas:
It may be nice to add a test that sets DeadThreadGCTriggerPeriodMilliseconds to very low number and then creates a ton of threads - to verify that there is nothing completely broken.

src/vm/threads.cpp
```diff
@@ -6009,6 +5984,193 @@ void ThreadStore::TransferStartedThread(Thread *thread, BOOL bRequiresTSL)
...
+    {
+        // Not safe to touch CLR state
+        return;
+    }
+
+    int gcGenerationToTrigger = 0;
+    IGCHeap *gcHeap = GCHeapUtilities::GetGCHeap();
+    _ASSERTE(gcHeap != nullptr);
+    SIZE_T generationCountThreshold = static_cast<SIZE_T>(s_DeadThreadCountThresholdForGCTrigger) / 2;
+    SIZE_T newDeadThreadGenerationCounts[3] = {0};
```
Maoni0: >3 [](start = 41, length = 1)

this should really be max_generation + 1

src/vm/threads.cpp
```diff
@@ -6009,6 +5984,193 @@ void ThreadStore::TransferStartedThread(Thread *thread, BOOL bRequiresTSL)
...
+                continue;
+            }
+
+            Object *exposedObject = OBJECTREFToObject(thread->GetExposedObjectRaw());
+            if (exposedObject == nullptr)
+            {
+                continue;
+            }
+
+            int exposedObjectGeneration = min(countedMaxGeneration, static_cast<int>(gcHeap->WhichGeneration(exposedObject)));
```
Maoni0: >min [](start = 42, length = 3)

not sure why you needed to do this min - WhichGeneration is not going to return anything that's > max_generation

src/vm/threads.cpp
```diff
@@ -6009,6 +5984,193 @@ void ThreadStore::TransferStartedThread(Thread *thread, BOOL bRequiresTSL)
...
+                gcGenerationToTrigger = exposedObjectGeneration;
+                if (gcGenerationToTrigger >= countedMaxGeneration)
+                {
+                    break;
+                }
+            }
+        }
+
+        // Make sure that enough time has elapsed since the last GC of the desired generation. Otherwise, the memory pressure
+        // would be sufficient to trigger GCs automatically.
```
Maoni0: not sure what to make of this comment - if not enough time has elapsed, we wouldn't want to trigger a GC anyway just because we don't want to trigger GCs too often.

src/vm/threads.cpp
```diff
@@ -6009,6 +5984,193 @@ void ThreadStore::TransferStartedThread(Thread *thread, BOOL bRequiresTSL)
...
+        SIZE_T gcLastMilliseconds = gcHeap->GetLastGCStartTime(gcGenerationToTrigger);
+        SIZE_T gcNowMilliseconds = gcHeap->GetNow();
+        if (gcNowMilliseconds - gcLastMilliseconds < s_DeadThreadGCTriggerPeriodMilliseconds)
+        {
+            return;
+        }
+
+        // For threads whose exposed objects are in the generation of GC that will be triggered or in a lower GC generation,
```
Maoni0: > threads whose exposed objects are in the generation of GC that will be triggered or in a lower GC generation, [](start = 14, length = 110)

When are dead threads removed from the thread list?

kouvel:
Made changes from above

src/vm/threads.cpp
kouvel: They're removed in the Thread destructor, which is typically called from the finalizer thread (`CallFinalizerOnThreadObject, ..., Thread::DoExtraWorkForFinalizer => Thread::CleanupDetachedThreads => Thread::DecExternalCount`)

PR #10426, opened 2017-03-23
Enable FEATURE_METADATA_EMIT_ALL for non-crossgen compiles

### LINKED ISSUE #8418: ildasm roundtrip failing for JIT\opt\perf\doublealign\Locals\Locals
This test started failing after recent corefx update

BruceForstall:
This allows ilasm roundtrip test to work with NetStandard 2.0
changes where C# compiler emits a ".permissionset" attribute
into the assembly that ildasm emits. This define enables the
APIs that ilasm uses to process this attribute.

Fixes #8418


src/md/compiler/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ set(MDCOMPILER_SOURCES
   import.cpp
   importhelper.cpp
   mdutil.cpp
+  newmerger.cpp
```
jkotas: newmerger.cpp should be out. It is pretty sizable chunk and it is only used by C++ link.exe, nothing else is using it. It should be fine to delete it.

BruceForstall: You're fast; I hadn't even hit "enter" to tag you yet...

I had to add newmerger.cpp to fix a build break due to the following in regmeta.h in the RegMeta class:

```
#ifdef FEATURE_METADATA_EMIT_ALL
    NEWMERGER   m_newMerger;                // class for handling merge
#endif //FEATURE_METADATA_EMIT_ALL
```

jkotas: I would rather delete from here and everywhere else.

BruceForstall: Its use is pretty pervasive in regmeta.cpp, regmeta_compilersupport.cpp, and regmeta_emit.cpp, all under `#ifdef FEATURE_METADATA_EMIT_ALL`.

Should I go back, and just have ilasm ignore errors returned from m_pEmitter->DefinePermissionSet()? Seems like we wouldn't get a true round trip in that case.

jkotas: The new merged is pretty straightforward to delete: https://github.com/jkotas/coreclr/commit/cf82a1007283bd1923b095ea225f83931dd9835e

PR #10429, opened 2017-03-23
Migrate the Reflection serialization holders to the shared partition.

AtsushiKan:


src/mscorlib/src/System/Reflection/MemberSerializationStringGenerator.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    internal static class MemberSerializationStringGenerator
+    {
+        //
+        // Generate the "Signature2" binary serialization string for PropertyInfos
+        //
+        // Because the string is effectively a file format for serialized Reflection objects, it must be exactly correct. If missing
+        // metadata prevents generating the string, this method throws a MissingMetadata exception.
+        // 
+        public static string SerializationToString(this PropertyInfo property) => ((RuntimePropertyInfo)property).SerializationToString();
```
AtsushiKan: While I don't want to do it as part of _this_ PR, it is possible to share MemberSerializationStringGenerator.cs as well (CoreRT implements SerializationToString() as portable extension methods.)

The dead-code cleanup job on the CoreCLR side will be a bit involved as its SerializationToString() weaves in and out of FCalls and the normal ToString() path in a rather awkward dance.

jkotas:
Maybe be worth running the corefx test to verify that nothing fell through the cracks.

AtsushiKan:
I ran the serialization tests I wrote when I ported those two files over to N earlier this year. (And I buddy built the files against N again to make sure they still passed on that side.)

PR #10430, opened 2017-03-23
Use roslyn common props

chcosta:
Fixes https://github.com/dotnet/coreclr/issues/10056

Import Roslyn common props so that we don't get mismatched imports on VS2017

/cc @weshaggard @gkhanna79

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -1,3 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <!--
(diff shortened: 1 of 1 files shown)


dir.props
```diff
@@ -197,4 +198,7 @@
     <PackageIndex Include="$(PackageIndexFile)" />
   </ItemGroup>
 
+  <!-- Use Roslyn Compilers to build -->
+  <Import Project="$(RoslynPropsFile)" Condition="'$(RunningOnUnix)'!='true' and Exists('$(RoslynPropsFile)') and '$(UseRoslynCompilers)'!='false'" />
```
gkhanna79: Did we validate the VS 2017 build with this change?

chcosta: I did a local build in a VS2017 dev command prompt.  Is there additional validation you want?

gkhanna79:
Can you validate it with and without VS 2015 installed as well?

chcosta:
Validated a clean build both with and without VS2015 installed.

gkhanna79:
Thank you - looks good!

PR #10432, opened 2017-03-23
Jit: track ref types for inlinee locals and args

AndyAyersMS:
The jit may use new temporaries for inline locals and args. For temps
representing ref type locals and args, propagate the class handle from
the appropriate signature. This makes more types to be available for
devirtualization.

Also in addition to capturing arg ref types, cleanup `impInlineFetchArg`
by commoning out the table lookups and adding a header comment.


AndyAyersMS:
@JosephTremoulet PTAL
cc @dotnet/jit-contrib 

No diffs outside of about 20 new devirtualizations in the jit-diff framework set.

PR #10436, opened 2017-03-23
[Local GC] Redirect fatal errors to the EE

swgillespie:
Today, when the GC encounters a fatal error, it calls the macro `EEPOLICY_HANDLE_FATAL_ERROR`, which captures the current instruction pointer and calls into the EE to tear down the process. This PR does the same thing, except it moves the IP retrieval to `GCToOSInterface` (seemed to be an appropriate place for it, more appropriate than a free function in `gc.cpp`) and adds `HandleFatalError` to `GCToEEInterface`, which the GC will call when it can't proceed.

Thinking ahead, EEPolicy will need to be aware of fatal errors coming from the GC - the IP that we give it will be meaningless without the base address of `gc.dll` (which the EE will know - it needs to record this for the stress log to work). Then again, this may not be worth doing, because 1) it's hopefully extremely rare that the GC falls over, and 2) the GC today only has two places where errors are fatal, and each of the two cases uses a different exit code, so today the exit code is the only thing necessary to uniquely determine what went wrong.

### DIFF
--- a/src/gc/env/gcenv.os.h
+++ b/src/gc/env/gcenv.os.h
@@ -277,6 +277,12 @@ public:
     // Return:
     //  Time stamp in milliseconds
     static uint32_t GetLowPrecisionTimeStamp();
+
+    // Gets the current instruction pointer.
+    // Return:
+    //  The value of the instruction pointer immediately after
+    //  the call to this function.
+    static void* GetCurrentInstructionPointer();
 };
 
 #endif // __GCENV_OS_H__
--- a/src/gc/env/gcenv.os.h
+++ b/src/gc/env/gcenv.os.h
@@ -277,12 +277,6 @@ public:
     // Return:
     //  Time stamp in milliseconds
     static uint32_t GetLowPrecisionTimeStamp();
-
-    // Gets the current instruction pointer.
-    // Return:
-    //  The value of the instruction pointer immediately after
-    //  the call to this function.
-    static void* GetCurrentInstructionPointer();
 };
 
 #endif // __GCENV_OS_H__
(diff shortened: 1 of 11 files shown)


swgillespie:
@Maoni0 @jkotas PTAL?

src/gc/env/gcenv.os.h
```diff
@@ -277,6 +277,12 @@ class GCToOSInterface
     // Return:
     //  Time stamp in milliseconds
     static uint32_t GetLowPrecisionTimeStamp();
+
+    // Gets the current instruction pointer.
+    // Return:
+    //  The value of the instruction pointer immediately after
+    //  the call to this function.
+    static void* GetCurrentInstructionPointer();
```
jkotas: I do not think this should be on GCToOSInterface. Getting current IP is not OS service, it is more like compiler intrinsic service. If we really wanted to have it, it should be together with other compiler intrinsics, e.g. there is a cluster of them in gcenv.base.h - look for MemoryBarrier.

However, I am not sure whether this is needed at all: There are just two odd places (more like bugs than features) where this is called, and they are distinguished by the error code already, the IP is not necessary to distinguish them. We should just have the error code in the callback, and not worry about the IP. It should be fine to call `EEPOLICY_HANDLE_FATAL_ERROR` on the VM side whatever IP it happens to pick up.

swgillespie: Ok, will do! I noticed that we're only calling it twice and with two different error codes too, so that sounds reasonable to me.

Maoni0: on a related note when I look at this code, it'd be nice to remove all the code under #ifndef NO_CATCH_HANDLERS unless anyone objects...

jkotas: > NO_CATCH_HANDLERS 

This was removed in CoreCLR a while ago.

Maoni0: oh I was looking at desktop src :D NM then

swgillespie:
thanks!

PR #10437, opened 2017-03-23
Some BitVecOps improvements

mikedn:


mikedn:
@BruceForstall @JosephTremoulet Do you think this stuff is useful?

* Long bitvector allocation was going through a virtual function
* In some cases (especially dataflow related) multiple BitVecOps operations were needed (e.g. intersect+union or equals+intersect) and some also resulted in unnecessary allocations
* There was some unnecessary copying, especially around making empty bitvectors - it's simpler to just zero-out some memory than copying previously zeroed-out memory
* In some cases new bitvectors were allocated instead of reusing existing ones

Might sound nice but I haven't been able to measure any perf improvement when crossgening corelib. Not even on x86 where it looks like we use 64 bit bitvectors and thus always allocate. I wonder if it's worth the trouble.

JosephTremoulet:
Sure, I'm happy to review/commit changes that improve our bitvector APIs and discipline; it has some benefit in terms of not needing to wonder again whether these issues are costly, propagating good patterns when people copy code, etc.  And I've got nothing against making the compiler do less work, even if the throughput difference gets lost in the measurement noise.

@pgavlin might have some thoughts, too.

We could also see if the throughput lab tests can measure anything... @dotnet-bot help

BruceForstall:
This definitely seems useful. It can be difficult to measure changes like this, but they certainly add up across many changes. It might also be that corelib doesn't exercise these cases enough.

@pgavlin has had success measuring total retired instructions count using callgrind on Linux and has experimented with pin on Windows. Perhaps he could suggest a way to measure this effect (or just do the measurement). He's unavailable for a few days, though.

JosephTremoulet:
Hmm, not finding it.  @adiaaida, what are the magic words to run throughput tests?

BruceForstall:
Maybe:

test Windows_NT_x64 perf
test Windows_NT_x86 perf

BruceForstall:
@DrewScoggins Can you answer how we get a perf (throughput) run from a PR?

I think the results are only available internally, aren't they?

BruceForstall:
The changes LGTM. I assume there are no asm diffs?

DrewScoggins:
test Windows_NT x64 throughput

DrewScoggins:
Here are the results for the throughput run, https://benchview/compare?jobid=31375&comparejobids=[31362]&.  Looking at the results we are seeing marginal gains, about .5%, across the whole suite.  A few DLLs improved by 1%, but nothing too dramatic as far as I can tell.

DrewScoggins:
test Ubuntu14.04 x64 throughput

DrewScoggins:
test Ubuntu14.04 throughput

BruceForstall:
@DrewScoggins Thanks. .5% is probably better than I expected for this change. I don't have much experience looking at these results. Can I assume that's greater than the noise?

AndyAyersMS:
Looking at the past 200 runs or so for just System.Private.CoreLib I'd say the noise level is around +/- 1%, so I don't believe we can reliably measure below that level.

BruceForstall:
Unfortunately, we need much more precise throughput numbers for them to be useful for measuring small, incremental improvements.

AndyAyersMS:
Agreed -- too bad we don't have instructions retired for these runs, it would probably be useful here.

AndyAyersMS:
FWIW here's the x64 windows corelib data for the past 200 runs. Roughly 5550 milliseconds +/ 50. Maybe a slight improving trend.

image

Data from above was:
image

So maybe a small win? Hard to be sure.

DrewScoggins:
Yeah. We are planning to add instructions retired, we just have not gotten around to it yet.

DrewScoggins:
@BruceForstall You can trigger a run with the phrases that I used. Don't want to say them and trigger another run. You can find the results on the Benchview site which is internal only as of right now.

mikedn:
@BruceForstall Yes, there aren't any diffs. 

There may be more room for improvement but it starts to stray into the "here be dragons" kind of thing (e.g. it may be possible to share empty long bit vectors if we know for sure they won't be changed) so IMO this is good to go as is. You may want to squash it during merge.

pgavlin:
@mikedn: you should be able to get an idea of how these changes might affect throughput by running something like `crossgen System.Private.CoreLib.dll` using the `callgrind` tool in `valgrind` (if you have a spare Linux/OSX machine) or by using any of the various `icount` [Pin](https://software.intel.com/en-us/articles/pin-a-dynamic-binary-instrumentation-tool) tools (usable on any OS, but must be built from source using GNU make--yes, that means some slightly tricky setup on Windows).

PR #10438, opened 2017-03-23
GetContainingObject is inconsistent with the lowest address it chooses.

Maoni0:
For profiling purpose it only cares about condemned ranges; for byref
validation it cares about any object on the heap.


Maoni0:
@swgillespie PTAL

PR #10439, opened 2017-03-24
Android cross-build script update

cydhaselton:
Updates download URLs and folders for libandroid-support-dev and libandroid-support packages. Downloads were failing and consequently cross-build failed early due to missing headers.

Tagging @qmfrederik for review


cydhaselton:
Unwanted changes in PR…removing and re-submitting

qmfrederik:
@cydhaselton PS - if you're going to update the build script, Termux now also packages libunwind so we can just use the Termux package instead of compiling it ourselves.

cydhaselton:
@qmfrederik Sweet. Will add.

cydhaselton:
@qmfrederik At some point we should come up with logic to handle updated packages; I'll add it as a TODO in the script

PR #10440, opened 2017-03-24
Add PNSE messages

danmosemsft:
Tightening up on PNSE because it looks like should actually have implementations and got overlooked (eg: most in Marshal)

I'm doing this by making sure they all have messages, either a specific one or the `SR.Arg_PlatformNotSupported` that they'd get by default anyway. That will help me grep again in future.

For `throw new PlatformNotSupportedException(WHAT_TEXT_HERE);` in this change please let me know whether we should have an implementation, and if not why it's not possible. Then I'll fix the message / open an issue as needed.

I didn't look at `new NotSupportedException()` as I'm assuming those are all carried over from desktop, ie., we've consistently only used that when no other platform supports the codepath either.

@jkotas

### DIFF
--- a/src/mscorlib/Resources/Strings.resx
+++ b/src/mscorlib/Resources/Strings.resx
@@ -4335,6 +4335,9 @@
   <data name="PlatformNotSupported_SecureBinarySerialization" xml:space="preserve">
     <value>Secure binary serialization is not supported on this platform.</value>
   </data>
+  <data name="PlatformNotSupported_StrongNameSigning" xml:space="preserve">
+    <value>Strong name signing is not supported on this platform.</value>
+  </data>
   <data name="PlatformNotSupported_WinRT" xml:space="preserve">
     <value>Windows Runtime is not supported on this operating system.</value>
   </data>
--- a/src/mscorlib/Resources/Strings.resx
+++ b/src/mscorlib/Resources/Strings.resx
@@ -1,64 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!-- 
-    Microsoft ResX Schema 
-    
-    Version 2.0
-    
-    The primary goals of this format is to allow a simple XML format 
-    that is mostly human readable. The generation and parsing of the 
-    various data types are done through the TypeConverter classes 
-    associated with the data types.
-    
-    Example:
-    
-    ... ado.net/XML headers & schema ...
-    <resheader name="resmimetype">text/microsoft-resx</resheader>
-    <resheader name="version">2.0</resheader>
-    <resheader name="reader">System.Resources.ResXResourceReader, System.Windows.Forms, ...</resheader>
-    <resheader name="writer">System.Resources.ResXResourceWriter, System.Windows.Forms, ...</resheader>
-    <data name="Name1"><value>this is my long string</value><comment>this is a comment</comment></data>
-    <data name="Color1" type="System.Drawing.Color, System.Drawing">Blue</data>
-    <data name="Bitmap1" mimetype="application/x-microsoft.net.object.binary.base64">
-        <value>[base64 mime encoded serialized .NET Framework object]</value>
-    </data>
-    <data name="Icon1" type="System.Drawing.Icon, System.Drawing" mimetype="application/x-microsoft.net.object.bytearray.base64">
-        <value>[base64 mime encoded string representing a byte array form of the .NET Framework object]</value>
-        <comment>This is a comment</comment>
-    </data>
-                
-    There are any number of "resheader" rows that contain simple 
-    name/value pairs.
-    
-    Each data row contains a name, and value. The row also contains a 
-    type or mimetype. Type corresponds to a .NET class that support 
-    text/value conversion through the TypeConverter architecture. 
-    Classes that don't support this are serialized and stored with the 
-    mimetype set.
-    
-    The mimetype is used for serialized objects, and tells the 
-    ResXResourceReader how to depersist the object. This is currently not 
-    extensible. For a given mimetype the value must be set accordingly:
-    
-    Note - application/x-microsoft.net.object.binary.base64 is the format 
-    that the ResXResourceWriter will generate, however the reader can 
-    read any of the formats listed below.
-    
-    mimetype: application/x-microsoft.net.object.binary.base64
-    value   : The object must be serialized with 
-            : System.Runtime.Serialization.Formatters.Binary.BinaryFormatter
-            : and then encoded with base64 encoding.
-    
-    mimetype: application/x-microsoft.net.object.soap.base64
-    value   : The object must be serialized with 
-            : System.Runtime.Serialization.Formatters.Soap.SoapFormatter
-            : and then encoded with base64 encoding.
-
-    mimetype: application/x-microsoft.net.object.bytearray.base64
-    value   : The object must be serialized into a byte array 
-            : using a System.ComponentModel.TypeConverter
-            : and then encoded with base64 encoding.
-    -->
   <xsd:schema id="root" xmlns="" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:msdata="urn:schemas-microsoft-com:xml-msdata">
     <xsd:import namespace="http://www.w3.org/XML/1998/namespace" />
     <xsd:element name="root" msdata:IsDataSet="true">
@@ -4361,6 +4302,12 @@
   <data name="PartitionerStatic_CurrentCalledBeforeMoveNext" xml:space="preserve">
     <value>MoveNext must be called at least once before calling Current.</value>
   </data>
+  <data name="PlatformNotSupported_ArgIterator" xml:space="preserve">
+    <value>ArgIterator is not supported on this platform.</value>
+  </data>
+     <data name="PlatformNotSupported_ComInterop" xml:space="preserve">
+    <value>COM Interop is not supported on this platform.</value>
+  </data>
   <data name="PlatformNotSupported_NamedSynchronizationPrimitives" xml:space="preserve">
     <value>The named version of this synchronization primitive is not supported on this platform.</value>
   </data>
@@ -4373,6 +4320,9 @@
   <data name="PlatformNotSupported_ReflectionOnly" xml:space="preserve">
     <value>ReflectionOnly loading is not supported on this platform.</value>
   </data>
+  <data name="PlatformNotSupported_Remoting" xml:space="preserve">
+    <value>Remoting is not supported on this platform.</value>
+  </data>   
   <data name="PlatformNotSupported_RequiresLonghorn" xml:space="preserve">
     <value>This operation is only supported on Windows Vista and above.</value>
   </data>
--- a/src/mscorlib/Resources/Strings.resx
+++ b/src/mscorlib/Resources/Strings.resx
@@ -4305,7 +4305,7 @@
   <data name="PlatformNotSupported_ArgIterator" xml:space="preserve">
     <value>ArgIterator is not supported on this platform.</value>
   </data>
-     <data name="PlatformNotSupported_ComInterop" xml:space="preserve">
+  <data name="PlatformNotSupported_ComInterop" xml:space="preserve">
     <value>COM Interop is not supported on this platform.</value>
   </data>
   <data name="PlatformNotSupported_NamedSynchronizationPrimitives" xml:space="preserve">
@@ -4331,6 +4331,9 @@
   </data>
   <data name="PlatformNotSupported_RequiresW2kSP3" xml:space="preserve">
     <value>This operation is only supported on Windows 2000 SP3 or later operating systems.</value>
+  </data>
+  <data name="PlatformNotSupported_SecureBinarySerialization" xml:space="preserve">
+    <value>Secure binary serialization is not supported on this platform.</value>
   </data>
   <data name="PlatformNotSupported_WinRT" xml:space="preserve">
     <value>Windows Runtime is not supported on this operating system.</value>
--- a/src/mscorlib/shared/System/Reflection/StrongNameKeyPair.cs
+++ b/src/mscorlib/shared/System/Reflection/StrongNameKeyPair.cs
@@ -50,14 +50,14 @@ namespace System.Reflection
 
         public StrongNameKeyPair(string keyPairContainer)
         {
-            throw new PlatformNotSupportedException(WHAT_TEXT_HERE);
+            throw new PlatformNotSupportedException(SR.PlatformNotSupported_StrongNameSigning);
         }
 
         public byte[] PublicKey
         {
             get
             {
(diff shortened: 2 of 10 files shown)


src/mscorlib/src/System/Threading/WaitHandle.cs
```diff
@@ -448,7 +448,7 @@ public static int WaitAny(WaitHandle[] waitHandles, int millisecondsTimeout)
                                         WaitHandle toWaitOn)
         {
 #if PLATFORM_UNIX
-            throw new PlatformNotSupportedException();
+            throw new PlatformNotSupportedException(WHAT_TEXT_HERE);
```
danmosemsft: These are all the SignalAndWait and in a previous discussion with @jkotas and @kouvel we determined we could implement these for Unix, so I'll open an issue.

src/mscorlib/src/System/Runtime/InteropServices/NonPortable.cs
```diff
@@ -15,7 +15,7 @@ public static int GetHRForException(Exception e)
 
         public static int AddRef(System.IntPtr pUnk)
         {
-            throw new PlatformNotSupportedException();
+            throw new PlatformNotSupportedException(SR.Arg_PlatformNotSupported);
```
danmosemsft: I assume all of the ones in this file are "can/will never implement"?

src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs
```diff
@@ -1172,7 +1172,7 @@ public static Type GetTypeFromCLSID(Guid clsid)
         //====================================================================
         public static IntPtr /* IDispatch */ GetIDispatchForObject(Object o)
         {
-            throw new PlatformNotSupportedException();
+            throw new PlatformNotSupportedException(WHAT_TEXT_HERE);
```
danmosemsft: supportable or not?

src/mscorlib/src/System/Runtime/InteropServices/NonPortable.cs
jkotas: All the ones in this file should have PlatformNotSupported_ComInterop. Note that this file is only included for Unix builds.

src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs
```diff
@@ -1772,7 +1772,7 @@ public static IntPtr SecureStringToBSTR(SecureString s)
 #if FEATURE_COMINTEROP
             return s.MarshalToBSTR();
 #else
-            throw new PlatformNotSupportedException();
```
jkotas: This is bug/inconsistency. We have implemented BSTR on Unix, even though it is COM interopy. Mono took the same approach. We should remove the FEATURE_COMINTEROP ifdef around this one and make it work on Unix - to be consistent with other methods that take BSTRs.

src/mscorlib/src/System/Exception.cs
```diff
@@ -520,8 +520,8 @@ private MethodBase GetExceptionMethodFromString()
 
         protected event EventHandler<SafeSerializationEventArgs> SerializeObjectState
         {
-            add { throw new PlatformNotSupportedException(); }
-            remove { throw new PlatformNotSupportedException(); }
+            add { throw new PlatformNotSupportedException(WHAT_TEXT_HERE); }
+            remove { throw new PlatformNotSupportedException(WHAT_TEXT_HERE); }
```
jkotas: Secure binary serialization is not supported on this platform.

src/mscorlib/src/System/Environment.cs
```diff
@@ -821,13 +821,6 @@ private static void SetEnvironmentVariableCore(string variable, string value)
             if (string.IsNullOrEmpty(value) || value[0] == '\0')
                 value = null;
 
-            if (AppDomain.IsAppXModel() && !AppDomain.IsAppXDesignMode())
-            {
-                // Environment variable accessors are not approved modern API.
-                // so we throw PlatformNotSupportedException.
```
jkotas: Other environment APIs are stubbed out for UWP to return empty string or null (look for IsAppXModel in this file). It does not make sense to remove the check here, without doing something about other Environment APIs for AppX.

danmosemsft: Seems reasonable to just remove all the special casing in anticipation of getting the env API approved, unless you object

src/mscorlib/src/System/Environment.cs
```diff
@@ -241,7 +241,7 @@ public static String MachineName
                 if (AppDomain.IsAppXModel() && !AppDomain.IsAppXDesignMode())
                 {
                     // Getting Computer Name is not a supported scenario on Store apps.
```
jkotas: The MachineName property should be deleted in CoreLib. The actual implementation was moved to corefx (https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Extensions/src/System/Environment.Uap.cs#L27).

src/mscorlib/shared/System/Reflection/StrongNameKeyPair.cs
```diff
@@ -50,14 +50,14 @@ protected StrongNameKeyPair(SerializationInfo info, StreamingContext context)
 
         public StrongNameKeyPair(string keyPairContainer)
         {
-            throw new PlatformNotSupportedException();
+            throw new PlatformNotSupportedException(WHAT_TEXT_HERE);
```
jkotas: Strong name signing is not support on this platform.

src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs
danmosemsft: https://github.com/dotnet/coreclr/issues/10443

danmosemsft:
@dotnet-bot help please

danmosemsft:
@dotnet-bot test Windows_NT x64 corefx_baseline please

src/mscorlib/src/System/Environment.cs
```diff
@@ -233,27 +226,6 @@ public static String ExpandEnvironmentVariables(String name)
...
-                {
-                    // Getting Computer Name is not a supported scenario on Store apps.
-                    throw new PlatformNotSupportedException();
-                }
-
-                // In future release of operating systems, you might be able to rename a machine without
-                // rebooting.  Therefore, don't cache this machine name.
-                StringBuilder buf = new StringBuilder(MaxMachineNameLength);
-                int len = MaxMachineNameLength;
-                if (Win32Native.GetComputerName(buf, ref len) == 0)
```
jkotas: Nit: Can delete GetComputerName PInvoke as well.

jkotas:
LGTM

danmosemsft:
@dotnet-bot test Windows_NT x64 corefx_baseline please

danmosemsft:
@dotnet-bot test tizen armel cross debug build (./b41488.sh: line 132:  1710 Segmentation fault      (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" b41488.exe $CLRTestExecutionArguments)

danmosemsft:
`Finished running tests.  End time=22:46:28.49, Exit code = -1073741819`
on my box with CLR override I got 
` Finished running tests.  End time=23:42:32.62, Exit code = -1073740972` 
one time and pass the other time.

I think this is not due to my changes but perhaps due to mismatch of CoreFX and CoreCLR state.

src/mscorlib/Resources/Strings.resx
```diff
@@ -1,64 +1,5 @@
...
-    mimetype: application/x-microsoft.net.object.soap.base64
-    value   : The object must be serialized with 
-            : System.Runtime.Serialization.Formatters.Soap.SoapFormatter
-            : and then encoded with base64 encoding.
-
-    mimetype: application/x-microsoft.net.object.bytearray.base64
-    value   : The object must be serialized into a byte array 
-            : using a System.ComponentModel.TypeConverter
-            : and then encoded with base64 encoding.
-    -->
```
stephentoub: @danmosemsft, whenever I resave resx files in VS, this gets added back.  I'm not sure it's worth fighting against that.

src/mscorlib/Resources/Strings.resx
```diff
@@ -4382,6 +4332,12 @@
   <data name="PlatformNotSupported_RequiresW2kSP3" xml:space="preserve">
     <value>This operation is only supported on Windows 2000 SP3 or later operating systems.</value>
   </data>
+  <data name="PlatformNotSupported_SecureBinarySerialization" xml:space="preserve">
+    <value>Secure binary serialization is not supported on this platform.</value>
+  </data>
+  <data name="PlatformNotSupported_StrongNameSigning" xml:space="preserve">
+    <value>Strong name signing is not supported on this platform.</value>
```
stephentoub: Nit: "Strong name" => "Strong-name"

src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs
```diff
@@ -408,7 +408,7 @@ public static void Copy(IntPtr source, IntPtr[] destination, int startIndex, int
         //====================================================================
         public static byte ReadByte([MarshalAs(UnmanagedType.AsAny), In] Object ptr, int ofs)
         {
-            throw new PlatformNotSupportedException();
+            throw new PlatformNotSupportedException(); // https://github.com/dotnet/coreclr/issues/10442
```
stephentoub: Any reason not to include a specific message in these?  Can we take the opportunity of a PNSE being thrown to tell the dev getting the exception what they should use instead?

src/mscorlib/src/System/Runtime/InteropServices/NonPortable.cs
```diff
@@ -159,37 +159,37 @@ public static bool IsComObject(object o)
...
         public static int ReleaseComObject(object o)
         {
-            throw new PlatformNotSupportedException();
+            throw new PlatformNotSupportedException(SR.PlatformNotSupported_ComInterop);
         }
 
         public static void ZeroFreeBSTR(System.IntPtr s)
         {
-            throw new PlatformNotSupportedException();
+            throw new PlatformNotSupportedException(SR.PlatformNotSupported_ComInterop);
```
stephentoub: Why can't ZeroFreeBSTR be supported?  The implementation looks trivial.  Is it just that we don't expect it to be useful?

src/mscorlib/src/System/Threading/WaitHandle.cs
```diff
@@ -448,7 +448,7 @@ public static int WaitAny(WaitHandle[] waitHandles, int millisecondsTimeout)
                                         WaitHandle toWaitOn)
         {
 #if PLATFORM_UNIX
-            throw new PlatformNotSupportedException();
+            throw new PlatformNotSupportedException(SR.Arg_PlatformNotSupported); // https://github.com/dotnet/coreclr/issues/10441
```
stephentoub: How are we deciding when to throw a generic error message versus a more specific one?

src/mscorlib/Resources/Strings.resx
jkotas: strong-name, strongname, strong name ... I do not think we have any consistency.

stephentoub: We may be inconsistent.  But from a grammar perspective, when it's used as an adjective, there should be a dash.  Otherwise it can be parsed as strong (name signing) rather than (strong name) signing.  Like I said, a nit.

src/mscorlib/src/System/Runtime/InteropServices/NonPortable.cs
jkotas: Agree - can be included in #10443

src/mscorlib/Resources/Strings.resx
danmosemsft: I didn't realize that. Boo! I filed a bug about 10 years ago to fix the .resx template. I didn't realize the designer would need a fix as well. I removed these in corefx also. If they come back I"ll leave them.

src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs
danmosemsft: I didn't include a message because these are in the category of "maybe we should implement" (hence the issue link also). THere's an email thread about it.

src/mscorlib/src/System/Runtime/InteropServices/NonPortable.cs
danmosemsft: It is supported on Windows (marshal.cs). I'll note in #10443 for Unix.

src/mscorlib/src/System/Threading/WaitHandle.cs
danmosemsft: I put in a specific message whenever we will likely ship that way. I put in a generic one with an issue when we may fix it.

src/mscorlib/Resources/Strings.resx
danmosemsft: I"ll put in my other commit

src/mscorlib/src/System/Threading/WaitHandle.cs
stephentoub: > I put in a generic one with an issue when we may fix it. 

So is the plan then to audit these all again before we ship 2.0 and put in specific messages for anything remaining?  Is there an issue tracking that?

danmosemsft: The issue is the specific issue for each one, in this case in the comment. If we don't do the implementation, potentially we should go add a nice message. I don't see this as a particularly high priority vs. adding implementation elsewhere. (If I get a message "Frob is not supported on this platform" vs. "Operation is not supported on this platform." when I call Frob() I don't know I'm greatly better off)

I'm more concerned with the 500-1000 PNSE in CoreFX (note grep doesn't find them all -- there is also `<GeneratePlatformNotSupportedAssembly Condition="'$(TargetsUnix)' == 'true'">true</GeneratePlatformNotSupportedAssembly>`). We did go through them once at least.

stephentoub: > I'm more concerned with the 500-1000 PNSE in CoreFX 

Of course.  I'm just wondering why we'd go to the trouble of adding this specific error messages for some cases but not others.  To me, the most important place to have specific error messages is when a) it's not immediately obvious what caused it (e.g. the method is generally supported but not if a certain argument or certain combination of arguments is used) or b) there's an alternative we can recommend.  For everything else, I think the default message is fine, and that we shouldn't be spending time or space on resources adding in specific messages.  I also like consistency, though.

danmosemsft: I'm fine with that. I figured there was some value when it was a whole technology. Eg., CAS. It also codified my mental process of figuring out why we don't support it...

PR #10445, opened 2017-03-24
Remove dead strings

danmosemsft:
Remove the 524 of 1554 strings that aren't used in corelib

Used https://gist.github.com/danmosemsft/9da8ef840afac4edc6efd6db19638a76 which hopefully accounts for all the variations of loading strings that don't go through SR or use  generated ID's.

@alexperovich

### DIFF
--- a/src/mscorlib/Resources/Strings.resx
+++ b/src/mscorlib/Resources/Strings.resx
@@ -2870,6 +2870,9 @@
   <data name="NotSupported_ByRefLike" xml:space="preserve">
     <value>Cannot create boxed ByRef-like values.</value>
   </data>
+  <data name="NotSupported_ByRefLikeArray" xml:space="preserve">
+    <value>Cannot create arrays of ByRef-like values.</value>
+  </data>
   <data name="NotSupported_ByRefReturn" xml:space="preserve">
     <value>ByRef return value not supported in reflection invocation.</value>
   </data>
@@ -2944,6 +2947,9 @@
   </data>
   <data name="NotSupported_KeyCollectionSet" xml:space="preserve">
     <value>Mutating a key collection derived from a dictionary is not allowed.</value>
+  </data>
+  <data name="NotSupported_VoidArray" xml:space="preserve">
+    <value>Arrays of System.Void are not supported.</value>
   </data>
   <data name="NotSupported_ManagedActivation" xml:space="preserve">
     <value>Cannot create uninitialized instances of types requiring managed activation.</value>
--- a/src/mscorlib/Resources/Strings.resx
+++ b/src/mscorlib/Resources/Strings.resx
@@ -3174,7 +3174,7 @@
     <value>Secure binary serialization is not supported on this platform.</value>
   </data>
   <data name="PlatformNotSupported_StrongNameSigning" xml:space="preserve">
-    <value>Strong name signing is not supported on this platform.</value>
+    <value>Strong-name signing is not supported on this platform.</value>
   </data>
   <data name="PlatformNotSupported_WinRT" xml:space="preserve">
     <value>Windows Runtime is not supported on this operating system.</value>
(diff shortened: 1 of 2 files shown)


danmosemsft:
@dotnet-bot test Windows_NT x64 corefx_baseline please

danmosemsft:
@dotnet-bot test Windows_NT x64 corefx_baseline please

danmosemsft:
Also discovered `COMPlusThrow` and `FCThrow` ...

danmosemsft:
@jkotas did I miss any places to find resource strings?

stephentoub:
@danmosemsft, can you do a full CoreFx run against your changes as well?  There was recently a case where corefx tests were looking for specific error strings via reflection, and it'd be good to know which tests we need to fix up, if there are any relying on strings you're deleting.

danmosemsft:
@stephentoub I did (that's the "jit - corefx" run which uses `/p:CoreCLROverridePath` against corefx head) and it seems to have hit something I thought I fixed, which I'll take a look at.

stephentoub:
Ok, thanks.

danmosemsft:
Missed some more `COMPlusThrow..`.

@dotnet-bot test Windows_NT x64 corefx_baseline please

danmosemsft:
@alexperovich can you please eyeball whether you recall any way of loading resources that I might have missed? As far as I know I got them all.

danmosemsft:
Test failed because of missing "NotSupported_Void[]" but the resx only ever had "NotSupported_VoidArray". I don't know how one was being translated to the other, but added.

danmosemsft:
@dotnet-bot test Windows_NT x64 corefx_baseline please

danmosemsft:
(Incidentally I'm doing this mostly because it makes it more feasible to share with CoreRT)

alexperovich:
> Test failed because of missing "NotSupported_Void[]" but the resx only ever had "NotSupported_VoidArray". I don't know how one was being translated to the other, but added.

I had to rename `NotSupported_Void[]` to `NotSupported_VoidArray` because `NotSupported_Void[]` is not a valid identifier.

I can't think of anything you have missed.

danmosemsft:
The only failure is Linq Expressions again, with ` -1073741819` .. last time when I ran locally this happened sporadically so I believe it's a mismatch somehow and unrelated.

@stephentoub / @jkotas could one of you please review? Feel free to merge if it looks ok.

danmosemsft:
Something is odd then because the test that was failing before because `NotSupported_Void[]` was missing is no longer failing even though I hadn't renamed that to `NotSupported_VoidArray`. Still that's correct so I've done it now and I"ll test again.

danmosemsft:
Just Linq expressions again. I believe I just need a CR signoff.

AndyAyersMS:
Took a look at why Linq Expressions test fails. Was not sure if this is a known issue.

System.Linq.Expressions.Tests.CompilerTests.CompileDeepTree_NoStackOverflowFast intentionally runs code on a thread with a very small stack, presumably to ensure that expression compilation stack usage is within nominal expectations. On a checked runtime it hits a stack overflow while loading a class and trying to create the class display name.

Guessing in a release built runtime the stack overflow doesn't happen and the test passes. So either we should bump the stack size up a bit to allow it to also pass on checked runtimes, or somehow mark the test as one that should not be run with a checked runtime.

Also interesting in this test suite are a number of examples of bad IL that the jit rejects -- presumably those are intentional (?). I have not looked that closely at these.

cc @RussKeldorph @VSadov

RussKeldorph:
@AndyAyersMS Thanks.  I had started looking at it, but I had just gotten to the stack overflow, not root cause.  I'll follow up to see which of your options is appropriate.

JonHanna:
@AndyAyersMS there's some bad IL that is there on purpose, because there are some constructs that should be rejected by expressions. It would probably be okay to disable those in a context that couldn't cope with them, esp. since the same difficulty would be the condition that is meant to be guarded against won't come up.
There is quite likely some bad IL from tests that generate IL and which are just correct enough to test what they needed to test. If there are any like that, those should ideally be fixed, though disabling would be a reasonable starting point.
CC: @hughbe

AndyAyersMS:
From what I can see the bad IL in the tests doesn't cause any problems, the jit detects it and signals to the VM which raises the expected exceptions and everything rolls forward as it should.

All I was getting at was that it makes running the test under the debugger a bit more interesting as there are lots of "expected" exceptions to skip over on the way to an unexpected exception....

Issue #9879, opened 2017-03-01
Assembly.LoadFrom should try to fall back to Assembly.Load before failing


billwert:
On CoreCLR, LoadFrom is prohibited from loading things that are in the TPA. This breaks certain patterns developers use for plug-in models, such as this:
```csharp
            var exePath = Assembly.GetExecutingAssembly().Location;
            exePath = exePath.Replace(Path.GetFileName(exePath), "");
            foreach(var plugin in Directory.GetFiles(exePath, "*.dll"))
            {
                var a = Assembly.LoadFrom(plugin);
                // do plugin-y things
            }
```

Instead of failing we could fall back to the Load code path, since it is all the same context anyway.

cc: @rahku

jkotas:
Make sense to me

PR #10446, opened 2017-03-24
Allow assemblies in TPA to loaded using LoadFrom

rahku:


rahku:
Fixes #9879

rahku:
PTAL @jkotas

src/binder/clrprivbindercoreclr.cpp
```diff
@@ -159,8 +159,13 @@ HRESULT CLRPrivBinderCoreCLR::BindUsingPEImage( /* in */ PEImage *pPEImage,
                 {
                     if (pCoreCLRFoundAssembly->GetIsInGAC())
                     {
-                        // If we were able to bind to a TPA assembly, then fail the load
```
jkotas: We may want to delete or move the `IsMscorlib` case above - so that this kicks for it as well.

src/binder/clrprivbindercoreclr.cpp
```diff
@@ -159,8 +159,13 @@ HRESULT CLRPrivBinderCoreCLR::BindUsingPEImage( /* in */ PEImage *pPEImage,
...
-                        IF_FAIL_GO(HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND));
+                        SString tpaAssemblyPath;
+                        SString assemblyPath;
+                        BinderGetImagePath(pPEImage, assemblyPath);
+                        BinderGetImagePath(pCoreCLRFoundAssembly->GetPEImage(), tpaAssemblyPath);
+                        if(!tpaAssemblyPath.EqualsCaseInsensitive(assemblyPath))
```
jkotas: You should be able to just compare `PEImage` pointers. PEImages are in a hashtable with path as the key (check `PEImage::OpenImage`) so you should get the same one for same path.

jkotas:
Test be in corefx?

src/binder/clrprivbindercoreclr.cpp
rahku: Yes this is correct! Missed it.

src/binder/clrprivbindercoreclr.cpp
```diff
@@ -159,8 +159,13 @@ HRESULT CLRPrivBinderCoreCLR::BindUsingPEImage( /* in */ PEImage *pPEImage,
...
+                            // If we were able to bind to a TPA assembly, then fail the load
+                            IF_FAIL_GO(HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND));
```
rahku: also with the exception you get very generic message string. 
Unhandled Exception: System.IO.FileLoadException: Could not load file or assembly 'test, Version=0.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
I am planning to change it to following
Unhandled Exception: System.IO.FileLoadException: Could not load file or assembly 'test, Version=0.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. Assembly with same identity found at standard probing locations.

what do you think?

jkotas: Yes, better exception message would be nice.

rahku:
@gkhanna79 PTAL

src/binder/clrprivbindercoreclr.cpp
```diff
@@ -157,10 +149,12 @@ HRESULT CLRPrivBinderCoreCLR::BindUsingPEImage( /* in */ PEImage *pPEImage,
                 hr = BindAssemblyByNameWorker(pAssemblyName, &pCoreCLRFoundAssembly, true /* excludeAppPaths */);
                 if (SUCCEEDED(hr))
                 {
+                    // Is identity same as TPA assembly
                     if (pCoreCLRFoundAssembly->GetIsInGAC())
                     {
-                        // If we were able to bind to a TPA assembly, then fail the load
-                        IF_FAIL_GO(HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND));
+                        // Fail bind if requested assembly is at different location than TPA assembly
+                        if(pPEImage != pCoreCLRFoundAssembly->GetPEImage())
```
gkhanna79: What happens today if we have user assembly A1 1.0.0.0 loaded from C:\foo into a load context (Default or custom) and later get a request to load A1 1.0.0.0 from c:\foo2? Iwoudl expect that we will find a AssemblySpec match against what is already loaded and return that from the binder - is that not true? If so, why wont we follow the same pattern here?

rahku: Yes this is correct if you do loadfrom for both the assemblies. I believe the behavior that we have currently is fine for LoadFrom because there are explicit calls to do loadfrom with different path to same assembly. The user seems to be just playing around here. However, for TPA assemblies these can conflict can happen between Assembly.Load & Assembly.LoadFrom. User does not have total idea what files are in TPA and so if a conflict does happen it is better to throw an exception instead of returning an assembly from TPA.


I am trying to be explicit and throwing an exception instead of returning or loading assembly at different location.

gkhanna79: But your bind above will only succeed if AssemblySpec matches, independent of the paths involved. And that means, you can use the bindResult (which comes from the assembly specified in TPA) to resolve the incoming request as opposed to fail it.

If AssemblySpec was different, and any assembly on TPA does not match it, then your bind will fail automatically.

rahku: Yes. I am not very strongly against doing what you are suggesting ...only rationale I had was to be explicit and cause an error instead of loading dll at some other location. But we already do that for ALC.LoadFromAssemblyPath wherein we return the already loaded assembly instead of an error. I will do the change as you are suggesting. @jkotas do you have any objection?

jkotas: ok with me

rahku: To be clear this is what I am implementing:
LoadFrom(assemblypathWithSameIdentityasTPA)
- return TPA assembly if already loaded
- load TPA assembly if not already loaded and return that 
- above two bullets will be the behavior irrespective of whether the assembly path matches with TPA assembly

gkhanna79: > above two bullets will be the behavior irrespective of whether the assembly path matches with TPA assembly

Sounds good to me. Please ensure you test with mismatched specs to ensure the expected behaviors are in place for passing and failure scenarios. Also confirm that we do the same for other LoadFrom* loads as well.

rahku: Yes if spec does not match what is present in TPA then loadfrom will fail for default context. I have verified this.

rahku:
PTAL. I have modified the behavior. I have also verified that other LoadFrom i.e. LoadFromNativeImagePath & LoadFromStream exhibit the same behavior in default ALC.

rahku:

src/binder/clrprivbindercoreclr.cpp
gkhanna79: Agreed - we can make it consistent though attempt to load SPC.dll is basically equivalent to axe your own foot.

src/binder/clrprivbindercoreclr.cpp
```diff
@@ -159,8 +159,8 @@ HRESULT CLRPrivBinderCoreCLR::BindUsingPEImage( /* in */ PEImage *pPEImage,
                 {
                     if (pCoreCLRFoundAssembly->GetIsInGAC())
```
gkhanna79: Do we need this GetIsInGAC check any more @jkotas? I think we should just remove it.

src/binder/clrprivbindercoreclr.cpp
rahku: I would keep the error for corelib in place. Not much of value to change it.

src/binder/clrprivbindercoreclr.cpp
```diff
@@ -159,8 +159,8 @@ HRESULT CLRPrivBinderCoreCLR::BindUsingPEImage( /* in */ PEImage *pPEImage,
...
                     {
-                        // If we were able to bind to a TPA assembly, then fail the load
-                        IF_FAIL_GO(HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND));
+                        *ppAssembly = pCoreCLRFoundAssembly.Extract();
+                        goto Exit;
```
gkhanna79: If you add an "else" to the "if", you will not need the goto and the "else" can handle the binding of PEImage not found (which is what happens below).

This will also make the code cleaner to read.

src/binder/clrprivbindercoreclr.cpp
rahku: I think we do. In case the bind is successfull but not to a TPA assembly we do not want to set ppAssembly .

src/vm/assemblynative.cpp
```diff
@@ -507,9 +507,12 @@ Assembly* AssemblyNative::LoadFromPEImage(ICLRPrivBinder* pBinderContext, PEImag
         spec.GetFileOrDisplayName(0, name);
         COMPlusThrowHR(COR_E_FILELOAD, dwMessageID, name);
     }
+
+    BINDER_SPACE::Assembly* assem;
+    assem = BINDER_SPACE::GetAssemblyFromPrivAssemblyFast(pAssembly);
     
-    PEAssemblyHolder pPEAssembly(PEAssembly::Open(pParentAssembly, pILImage, pNIImage, pAssembly, FALSE));
-    
+    PEAssemblyHolder pPEAssembly(PEAssembly::Open(pParentAssembly, assem->GetPEImage(), assem->GetNativePEImage(), pAssembly, FALSE));
```
gkhanna79: Looks good otherwise. Did you run the ALC tests against this change?

rahku: yes

src/binder/clrprivbindercoreclr.cpp
rahku: there is too many if's in play here and for all the else I want to fall back to binding of PEImage not found. So will keep the way it is.

src/binder/clrprivbindercoreclr.cpp
gkhanna79: I think it is for consistency, but I am fine either ways.

PR #10455, opened 2017-03-24
Move '#pragma warning(disable:21000)' to correct place

mattwarren:
Whilst I was writing a blog post ['A Hitchhikers Guide to the CoreCLR Source Code'](http://mattwarren.org/2017/03/23/Hitchhikers-Guide-to-the-CoreCLR-Source-Code/#top-10-lists) I noticed that a `#pragma warning(disable:21000)` was mis-aligned.

I'm pretty sure that having it in the wrong place doesn't cause a problem, because the other methods are smaller (i.e. this doesn't actually *fix* anything). So I understand if you don't want to take the PR.

### DIFF
--- a/src/debug/di/process.cpp
+++ b/src/debug/di/process.cpp
@@ -4525,10 +4525,6 @@ void CordbProcess::QueueFakeConnectionEvents()
 // from the runtime controller. This represents the last amount of processing
 // the DI gets to do on an event before giving it to the user.
 //
-#ifdef _PREFAST_
-#pragma warning(push)
-#pragma warning(disable:21000) // Suppress PREFast warning about overly large function
-#endif
 void CordbProcess::DispatchRCEvent()
 {
     INTERNAL_API_ENTRY(this);
@@ -4729,6 +4725,10 @@ void CordbProcess::DbgAssertAppDomainDeleted(VMPTR_AppDomain vmAppDomainDeleted)
 //    A V2 shim can provide a proxy calllack that takes these events and queues them and 
 //    does the real dispatch to the user to emulate V2 semantics.
 //
+#ifdef _PREFAST_
+#pragma warning(push)
+#pragma warning(disable:21000) // Suppress PREFast warning about overly large function
+#endif
 void CordbProcess::RawDispatchEvent(
     DebuggerIPCEvent *          pEvent, 
     RSLockHolder *              pLockHolder,


mattwarren:
Are the build failures (['OSX10.12 x64 Checked Build and Test'](https://ci.dot.net/job/dotnet_coreclr/job/master/job/checked_osx10.12_flow_prtest/456/) and ['Tizen armel Cross Debug Build'](https://ci.dot.net/job/dotnet_coreclr/job/master/job/armel_cross_debug_tizen_prtest/121/)) caused by this PR? I can't really make sense of the build output?

janvorli:
@mattwarren they cannot be related. The _PREFAST_ is never defined out of windows.

danmosemsft:
@dotnet-bot test OSX10.12 x64 Checked Build and Test please (timeout)
@dotnet-bot test Tizen armel Cross Debug Build  please (infra)

danmosemsft:
> I'm pretty sure that having it in the wrong place doesn't cause a problem, because the other methods are smaller (i.e. this doesn't actually fix anything). 

You mean, it makes no difference in the build? I am not sure when we define _PREFAST_ but maybe it's not defined normally.

mattwarren:
> You mean, it makes no difference in the build? I am not sure when we define PREFAST but maybe it's not defined normally.

Sorry, I wasn't clear. What I mean is that when `_PREFAST_` is defined, I don't think my PR actually fixes anything, it's just a code tidy-up (so I understand if you don't want to accept it)

**Before the PR/fix** the `#pragma warning(disable:21000)` started on [line 4530](https://github.com/dotnet/coreclr/blob/master/src/debug/di/process.cpp#L4530) and it was 'popped' on [line 5886](https://github.com/dotnet/coreclr/blob/master/src/debug/di/process.cpp#L5886). This means that the following methods all have the warning disabled:

- `CordbProcess::DispatchRCEvent()`
- `CordbProcess::DbgAssertAppDomainDeletedCallback(..)`
- `CordbProcess::DbgAssertAppDomainDeleted(..)`
- `CordbProcess::RawDispatchEvent(..)`

**After the PR** only `CordbProcess::RawDispatchEvent(..)` is included in the `pragma` as I believe it's the only large method that really needs the warning disabled, as it's [1,152 lines long](https://github.com/dotnet/coreclr/blob/master/src/debug/di/process.cpp#L4732-L5884)

danmosemsft:
OK, thanks @mattwarren !

danmosemsft:
PS We passed around your great blog post...

mattwarren:
> PS We passed around your great blog post...

Thanks, you might also like ['The 68 things the CLR does before executing a single line of your code'](http://mattwarren.org/2017/02/07/The-68-things-the-CLR-does-before-executing-a-single-line-of-your-code/) and ['The .NET IL Interpreter'](http://mattwarren.org/2017/03/30/The-.NET-IL-Interpreter/) `</shameless plug>`

PR #10460, opened 2017-03-24
Move DateTime to shared partition

alexperovich:
Depends on dotnet/corefx#17484


src/mscorlib/shared/Interop/Windows/Kernel32/Interop.GetSystemTimeAsFileTime.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        [DllImport(Libraries.Kernel32)]
+        private extern static unsafe void GetSystemTimeAsFileTime(long* lpSystemTimeAsFileTime);
+
+        [DllImport(Libraries.Kernel32)]
+        private extern static unsafe void GetSystemTimePreciseAsFileTime(long* lpSystemTimeAsFileTime);
+
+        private unsafe delegate void GetSystemTimeAsFileTimeFn(long* lpSystemTimeAsFileTime);
+
+        private static Lazy<GetSystemTimeAsFileTimeFn> _getSystemTimeAsFileTime = new Lazy<GetSystemTimeAsFileTimeFn>(GetGetSystemTimeAsFileTimeFn);
```
jkotas: This method is performance sensitive. Going through delegate like this will likely make it several times slower. I do no think it is acceptable perf regression.

src/mscorlib/shared/Interop/Unix/System.Native/Interop.GetSystemTimeAsTicks.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal unsafe partial class Sys
+    {
+        [DllImport(Libraries.SystemNative, EntryPoint = "SystemNative_GetSystemTimeAsTicks")]
```
jkotas: It would be useful to measure how much overhead the PInvoke adds on Unix. I suspect we may need to use the PInvoke on Unix as well to avoid the perf regression there.

jkotas:
I see some good changes in CoreRT when I diff the CoreCLR and CoreRT:

- dateData renamed to _dateData, 
- Contract.Requires changed to Debug.Assert

Preserve these? 

LGTM otherwise.

jkotas:

PR #10462, opened 2017-03-24
CoreFx 14486 Remove argument validation in DangerousCreate

WinCPP:
Changes in CoreClr code corresponding to [dotnet/corefx #14486](https://github.com/dotnet/corefx/issues/14486).

@ahsonkhan @karelz Kindly review.


ahsonkhan:
@WinCPP, maybe explicitly add to the xml comments that we don't do null checks on the object or bound checks on the length (same with the change in corefx).

WinCPP:
> @WinCPP, maybe explicitly add to the xml comments that we don't do null checks on the object or bound checks on the length (same with the change in corefx).

Oops seeing this just now, I ended my day early yesterday... But the PR is merged...still want that to go in via another PR..?

ahsonkhan:
Go for it.

Issue #8267, opened 2016-11-23
Local GC: consolidate the data types on the GC interface


Maoni0:
this includes -

use portable types, eg, no BOOL
for opaque contexts like profiling/ETW contexts, use one data type (it uses both void* and size_t).

Issue #4093, opened 2016-04-05
Inconsistent finalizer test behavior on Ubuntu


swgillespie:
See http://dotnet-ci.cloudapp.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_tst_prtest/501/testReport/junit/GC.Scenarios.DoublinkList/dlcollect/dlcollect/ for a failure. I failed to get a repro after running this for about a week last week, but it happened again in CI.
- http://dotnet-ci.cloudapp.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_tst_prtest/545/testReport/GC.Scenarios.DoublinkList/dlcollect/dlcollect/history/
- http://dotnet-ci.cloudapp.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_tst_prtest/545/testReport/GC.Scenarios.LeakWheel/leakwheel/leakwheel/history/

swgillespie:
Spoke to @Maoni0 about this one - I have a repro but since it only repros on checked Ubuntu builds it's not very high priority. Moving to 1.1.0.

jkotas:
#7558 - GC.API.GCHandleCollector.Usage.Usage test failed with "Case 3 failed: threshold not increasing!" on Ubuntu

swgillespie:
@jkotas that's interesting, never seen that one before:

```
09:41:43 FAILED   - GC/API/GCHandleCollector/Usage/Usage.sh
09:41:43                BEGIN EXECUTION
09:41:43                /mnt/j/workspace/dotnet_coreclr/master/checked_ubuntu_tst_prtest/bin/tests/Windows_NT.x64.Checked/Tests/coreoverlay/corerun Usage.exe
09:41:43                Case 1 Passed!
09:41:43                100, 100
09:41:43                Case 2 Passed!
09:41:43                Case 3 failed: threshold not increasing!
09:41:43                
09:41:43                Test Failed!
09:41:43                Expected: 100
09:41:43                Actual: 1
09:41:43                END EXECUTION - FAILED
```

jkotas:
It was seen before in #4400 that was resolved as duplicate of this issue.

swgillespie:
ah, you're right - sorry for my lapse in memory. I suppose it's much less common than the other manifestations.

AndyAyersMS:
Here's another instance with a crash dump... note this is from an unmerged PR so it might be something I did. But it might not be since earlier versions of this that are very similar did not hit this problem.

[...]

PR #10463, opened 2017-03-24
[Local GC] Use standard C++ types (bool) and consistent types (void*) on the interface

swgillespie:
Fixes https://github.com/dotnet/coreclr/issues/8267. The idea is to use `bool` instead of `BOOL` on the interface itself, and to use `void*` instead of `size_t` as the profiling context in some diagnostic machines. The EE still uses `size_t` as its context, so it casts `void*` to `size_t` in the appropriate places.

### DIFF
--- a/src/gc/gc.cpp
+++ b/src/gc/gc.cpp
@@ -36172,7 +36172,7 @@ CFinalize::FinalizeSegForAppDomain (AppDomain *pDomain,
 bool
 CFinalize::FinalizeAppDomain (AppDomain *pDomain, bool fRunFinalizers)
 {
-    BOOL finalizedFound = FALSE;
+    bool finalizedFound = false;
 
     unsigned int startSeg = gen_segment (max_generation);
 
@@ -36182,7 +36182,7 @@ CFinalize::FinalizeAppDomain (AppDomain *pDomain, bool fRunFinalizers)
     {
         if (FinalizeSegForAppDomain (pDomain, fRunFinalizers, Seg))
         {
-            finalizedFound = TRUE;
+            finalizedFound = true;
         }
     }
 
--- a/src/gc/gc.cpp
+++ b/src/gc/gc.cpp
@@ -21332,7 +21332,7 @@ void gc_heap::relocate_in_loh_compact()
         generation_free_obj_space (gen)));
 }
 
-void gc_heap::walk_relocation_for_loh (size_t profiling_context, record_surv_fn fn)
+void gc_heap::walk_relocation_for_loh (void* profiling_context, record_surv_fn fn)
 {
     generation* gen        = large_object_generation;
     heap_segment* seg      = heap_segment_rw (generation_start_segment (gen));
@@ -24245,7 +24245,7 @@ void gc_heap::walk_relocation_in_brick (uint8_t* tree, walk_relocate_args* args)
     }
 }
 
-void gc_heap::walk_relocation (size_t profiling_context, record_surv_fn fn)
+void gc_heap::walk_relocation (void* profiling_context, record_surv_fn fn)
 {
     generation* condemned_gen = generation_of (settings.condemned_generation);
     uint8_t*  start_address = generation_allocation_start (condemned_gen);
@@ -24301,7 +24301,7 @@ void gc_heap::walk_relocation (size_t profiling_context, record_surv_fn fn)
     }
 }
 
-void gc_heap::walk_survivors (record_surv_fn fn, size_t context, walk_surv_type type)
+void gc_heap::walk_survivors (record_surv_fn fn, void* context, walk_surv_type type)
 {
     if (type == walk_for_gc)
         walk_survivors_relocation (context, fn);
@@ -24316,7 +24316,7 @@ void gc_heap::walk_survivors (record_surv_fn fn, size_t context, walk_surv_type 
 }
 
 #if defined(BACKGROUND_GC) && defined(FEATURE_EVENT_TRACE)
-void gc_heap::walk_survivors_for_bgc (size_t profiling_context, record_surv_fn fn)
+void gc_heap::walk_survivors_for_bgc (void* profiling_context, record_surv_fn fn)
 {
     // This should only be called for BGCs
     assert(settings.concurrent);
@@ -30863,7 +30863,7 @@ BOOL gc_heap::large_object_marked (uint8_t* o, BOOL clearp)
     return m;
 }
 
-void gc_heap::walk_survivors_relocation (size_t profiling_context, record_surv_fn fn)
+void gc_heap::walk_survivors_relocation (void* profiling_context, record_surv_fn fn)
 {
     // Now walk the portion of memory that is actually being relocated.
     walk_relocation (profiling_context, fn);
@@ -30876,7 +30876,7 @@ void gc_heap::walk_survivors_relocation (size_t profiling_context, record_surv_f
 #endif //FEATURE_LOH_COMPACTION
 }
 
-void gc_heap::walk_survivors_for_loh (size_t profiling_context, record_surv_fn fn)
+void gc_heap::walk_survivors_for_loh (void* profiling_context, record_surv_fn fn)
 {
     generation* gen        = large_object_generation;
     heap_segment* seg      = heap_segment_rw (generation_start_segment (gen));;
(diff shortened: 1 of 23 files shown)


src/gc/gc.cpp
```diff
@@ -35760,30 +35760,30 @@ size_t GCHeap::GetFinalizablePromotedCount()
...
         gc_heap* hp = gc_heap::g_heaps [hn];
         if (hp->finalize_queue->FinalizeAppDomain (pDomain, fRunFinalizers))
-            foundp = TRUE;
+            foundp = true;
     }
     return foundp;
 
 #else //MULTIPLE_HEAPS
-    return pGenGCHeap->finalize_queue->FinalizeAppDomain (pDomain, fRunFinalizers);
+    return !!pGenGCHeap->finalize_queue->FinalizeAppDomain (pDomain, fRunFinalizers);
```
jkotas: I would be nice to convert `CFinalize::FinalizeAppDomain` to use bool too to avoid the !!

jkotas:
LGTM

src/gc/gc.cpp

swgillespie:
@Maoni0 PTAL also?

src/gc/gcimpl.h
```diff
@@ -157,8 +157,8 @@ class GCHeap : public IGCHeapInternal
     //returns the generation number of an object (not valid during relocation)
     unsigned WhichGeneration (Object* object);
     // returns TRUE is the object is ephemeral 
-    BOOL    IsEphemeral (Object* object);
-    BOOL    IsHeapPointer (void* object, BOOL small_heap_only = FALSE);
+    bool IsEphemeral (Object* object);
+    bool IsHeapPointer (void* object, bool small_heap_only = FALSE);
```
Maoni0: >FALSE [](start = 61, length = 5)

any reason why you are not changing FALSE to false here? same with GetValidSegmentSize.

src/gc/gcinterface.h
```diff
@@ -546,16 +546,16 @@ class IGCHeap {
     virtual HRESULT Initialize() = 0;
 
     // Returns whether nor this GC was promoted by the last GC.
-    virtual BOOL IsPromoted(Object* object) = 0;
+    virtual bool IsPromoted(Object* object) = 0;
 
     // Returns true if this pointer points into a GC heap, false otherwise.
-    virtual BOOL IsHeapPointer(void* object, BOOL small_heap_only = FALSE) = 0;
+    virtual bool IsHeapPointer(void* object, bool small_heap_only = false) = 0;
```
Maoni0: >IsHeapPointer [](start = 17, length = 13)

what about callers of these methods? eg, we call IsHeapPointer with TRUE in object.cpp

src/gc/gcimpl.h
swgillespie: No reason, it looks like I missed those - will fix.

src/gc/gcinterface.h
swgillespie: I'm not sure why MSVC *isn't* issuing a warning - I'd expect it to since it did for a bunch of other similar conversions. I can see if I can track down and fix the callers.

src/vm/gcenv.ee.cpp
```diff
@@ -1176,13 +1176,13 @@ void GCToEEInterface::DiagWalkFReachableObjects(void* gcContext)
 // don't get confused.
 void WalkMovedReferences(uint8_t* begin, uint8_t* end, 
                          ptrdiff_t reloc,
-                         size_t context, 
-                         BOOL fCompacting,
-                         BOOL fBGC)
+                         void* context, 
+                         bool fCompacting,
+                         bool fBGC)
```
Maoni0: >bool fBGC [](start = 25, length = 9)

I see you've changed these BOOLs to bools, but how this is called in gc.cpp remains unchanged. this is called in methods that gc_heap::walk_survivors calls, eg, in gc_heap::walk_survivors_for_bgc we are calling this method:

            fn (plug_start, 
                plug_end,
                0,              // Reloc distance == 0 as this is non-compacting
                profiling_context,
                FALSE,          // Non-compacting
                TRUE);          // BGC

arguably you could leave this method's arg types unchanged as it's not strictly on the interface...

src/gc/gcinterface.h
Maoni0: yeah, I was a bit surprised too... perhaps we are disabling this warning somewhere (that would be surprising too).

jkotas: The warning is only issued if you are converting BOOL variable to bool. It is warning you about that there will be one or two extra machine instructions to do this conversion.

The warning is not issues you are converting FALSE/TRUE constant to bool. The compiler will optimize this conversion out, and so it does not bother warning you.

I agree that it would be nice to get the upper case FALSE/TRUE changed to the lower case one.

swgillespie: I can't see any pragmas that would disable the warning anywhere and I'm also having trouble getting MSVC to emit a warning, even when compiling one of their code snippets (from their docs) that's supposed to trigger the warning: https://godbolt.org/g/p36GyD. So that's strange and unfortunate.

src/vm/gcenv.ee.cpp
swgillespie: I figured that these function pointer types are exposed via the interface by virtue of being arguments to interface functions. I can change them back to `BOOL` or fix the call sites - both are fine with me, whatever you'd prefer.

jkotas: Lowercase bool is where we should be converging to.

src/gc/gcinterface.h
Maoni0: it's unfortunate that we don't see the warning when we call these methods 'cause we are converting BOOL to bool. if you turn the build output to verbose level, perhaps somewhere you can see C4800 is disabled (not necessarily in code, it could also just be in some build scripts).

src/vm/gcenv.ee.cpp
Maoni0: yeah, that's why I said not *strictly* :-) either we change this and change the callers too; or we leave this unchanged but I wouldn't want it to be in an inbetween state.

src/gc/gcinterface.h
jkotas: I am pretty sure that C4800 is enabled. As I have said, the warning is not issues for all BOOL to bool conversions. It is issued only if the conversion results in extra native code to be produced.  

```
void bar(bool f)
{
}

void foo(BOOL f)
{
    bar(TRUE); // No warning
    bar(FALSE); // No warning
    bar(f); // WARNING!
}
```

Maoni0: ahh sorry didn't read your last comment in its complete form; that's kind of unfortunate - I'd prefer it just give the warning even when converting the const to bool.... it's easy to change the code and useful to know where you are doing that.

swgillespie: With my latest commit I fixed all of the callsites that I could find for the methods I changed on the interface.

Maoni0:
> inline BOOL IsGCInProgress(bool bConsiderGCStart = FALSE)

false here.

---
Refers to: src/gc/gc.h:266 in f9c3e37. [](commit_id = f9c3e371e8f62b294ed685b35440e7f39f3da82b, deletion_comment = False)

Maoni0:
> inline BOOL IsServerHeap()

bool here.

---
Refers to: src/gc/gc.h:274 in f9c3e37. [](commit_id = f9c3e371e8f62b294ed685b35440e7f39f3da82b, deletion_comment = False)

swgillespie:
@Maoni0 Those aren't exposed on the external-facing interface, should I be converting `BOOL` -> `bool` on that interface as well?

jkotas:
@swgillespie The place we want to end up is that there is no upper case BOOL in the GC at all. Unfortunately, we were not able to just do find`n`replace because of there are some BOOL variables that are used with interlocked operations or other sensitive situations. These places needs to be converted to int32_t or something similar once we run into them.

But otherwise, converting everything else from the upper case BOOL to lower case bool is goodness.

swgillespie:
Ok, sounds good!

Maoni0:
I am mainly objecting to the inconsistency... I am fine with doing the general work of replacing BOOL with bool later but something inbetween (as in, you have a method that's supposed to return BOOL but you are returning false) seems less desirable.

swgillespie:
@dotnet-bot Ubuntu x64 Checked Build and Test

https://github.com/dotnet/coreclr/issues/4093

swgillespie:

swgillespie:
thanks!

PR #10466, opened 2017-03-24
Fix for JitStress test failures

### LINKED ISSUE #10461: Assertion failed 'firstILBlock->hasProfileWeight()'
**Repro (of many)**
x64 checked
set COMPlus_JitStress=2
run coreclr tests

**Example CI run**
https://ci.dot.net/job/dotnet_coreclr/job/master/job/jitstress/job/x64_checked_windows_nt_jitstress2/124/

**Sample output**
```
D:\j\workspace\x64_checked_w---8a0029a8\bin\tests\Windows_NT.x64.Checked\TestWrappers\JIT.Directed\JIT.Directed.XUnitWrapper.cs(21527): error : JIT_Directed._nullabletypes_castclassinterface_r_castclassinterface_r_._nullabletypes_castclassinterface_r_castclassinterface_r_cmd [FAIL] [D:\j\workspace\x64_checked_w---8a0029a8\tests\runtest.proj]
        
        Assert failure(PID 3148 [0x00000c4c], Thread: 3532 [0x0dcc]): Assertion failed 'firstILBlock->hasProfileWeight()' in 'NullableTest5:BoxUnboxToNQ(ref):bool' (IL size 46)
        
            File: d:\j\workspace\x64_checked_w---8a0029a8\src\jit\flowgraph.cpp Line: 12704
            Image: D:\j\workspace\x64_checked_w---8a0029a8\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\CoreRun.exe
```

briansull:
Fixes #10461 - Assertion failed 'firstILBlock->hasProfileWeight()'

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -12699,14 +12699,21 @@ void Compiler::fgComputeEdgeWeights()
         // Did we allocate a scratch block as the new first BB?
         if (fgFirstBBisScratch())
         {
-            firstILBlock = firstILBlock->bbNext;
-            // The second block is expected to have a profile-derived weight
+            // Skip past any/all BBF_INTERNAL blocks that may have been added before the first real IL block.
+            // These internal blocks all should be BBJ_NONE blocks.
+            //
+            while (firstILBlock->bbFlags & BBF_INTERNAL)
+            {
+                assert(firstILBlock->bbJumpKind == BBJ_NONE);
+                firstILBlock = firstILBlock->bbNext;
+            }
+            // The 'firstILBlock' is now expected to have a profile-derived weight
             assert(firstILBlock->hasProfileWeight());
         }
 
-        // If the first block only has one ref then we use it's weight
-        // for fgCalledCount. Otherwise we have backedge's into the first block,
-        // so instead we use the sum of the return block weights for fgCalledCount.
+        // If the first block only has one ref then we use it's weight for fgCalledCount.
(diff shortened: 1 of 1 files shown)


briansull:
@dotnet-bot test Windows_NT Jitstress2
@dotnet-bot test Ubuntu Jitstress2

briansull:
@dotnet/jit-contrib PTAL

AndyAyersMS:
Do we ever go back and add weights to all the skipped internal blocks? Looks like we only do it for the first one.

Also it seems odd to me that there is an odd mixture of predicates here -- seems like the initial `if` should check the internal flag and then, inside the if, you could assert that when this first block is internal it is also scratch.

briansull:
I believe that the weights for the first and other internal blocks have already been added and we don't really need the code here that sets the profile weight at all.  

I'm not trying to fix everything that might be incorrect with this checkin, just getting the JIT-CI unblocked.

Currently there are several cases where the weight don't get set properly 
and these cases cause us to fail when calculating the edge weights.

We still use the the profile weights that we have and there aren't any correctness issues that are exposed by having incorrect weights.

briansull:
I'm also going to have to investigate the remaining couple of JitStress=2 failures
Where the inserted blocks aren't BBJ_NONE.

briansull:
@dotnet-bot test Windows_NT Jitstress2
@dotnet-bot test Ubuntu Jitstress2

briansull:
Passes JitStress=2

PR #9920, opened 2017-03-02
Vector conversion support

helloguo:
#9318
 
This is preliminary. Do not merge.

This build is on top of PR#9318. It fixes failures and implements the remaining intrinsics.

@sivarv @russellhadley PTAL

Fixes #9317

PR #10467, opened 2017-03-24
Add vector conversion tests

helloguo:
This PR adds tests for vector Conversion, Narrow and Widen. It is built on #9318 and #9920.


helloguo:
@BruceForstall PTAL

PR #10469, opened 2017-03-24
Change the Jit Disasm message when we have profile data

briansull:
It will now include information about the edge weights and the value of fgCalledCount.
Example:

```
; Assembly listing for method System.Number:TrailingZeros(ref,int):bool
; optimized code
; rsp based frame
; fully interruptible
; with IBC profile data, edge weights are valid, and fgCalledCount is 145

```


briansull:
@dotnet/jit-contrib PTAL

BruceForstall:
LGTM

PR #10480, opened 2017-03-25
Issue #7061 Remove private _Prop from StreamReader, StreamWriter

### LINKED ISSUE #7061: Are the *_Prop properties in StreamReader and StreamWriter still needed?
There are a [bunch of properties,](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/IO/StreamReader.cs#L1114) such as `CharLen_Prop` in StreamReader, that seemingly just wrap an underlying field except with an uglier syntax. This is apparently for "performance for MarshalByRefObject-derived types," since accessing instance fields of such objects is slow. However, I fail to see how these properties make things any faster, since all they do is wrap the underlying field (and their accessors will probably be inlined). Maybe it would be OK to remove them?

cc @jkotas

WinCPP:
@jamesqo @jkotas @karelz Kindly review.

Fixes #7061

### DIFF
--- a/src/mscorlib/src/System/IO/StreamReader.cs
+++ b/src/mscorlib/src/System/IO/StreamReader.cs
@@ -777,16 +777,16 @@ namespace System.IO
 
         private async Task<String> ReadLineAsyncInternal()
         {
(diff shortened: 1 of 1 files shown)


jkotas:
Thanks!

karelz:
Thanks! I love deleting unnecessary code ;-)

PR #10483, opened 2017-03-25
Update for Android cross build script

cydhaselton:
Updated URLs for libandroid-support and libandroid-support-dev downloads.
Added URLs and folders for libunwind and libunwind-dev
Removed libunwind build.
Added TODO for package download logic

Tagging @qmfrederik as original script creator and @jkotas for visibility for merging (current script in master has broken URLs for package downloads; breaks Android cross-build)

### DIFF
--- a/cross/build-android-rootfs.sh
+++ b/cross/build-android-rootfs.sh
@@ -94,6 +94,7 @@ echo Generating the $__BuildArch toolchain
 $__NDK_Dir/build/tools/make_standalone_toolchain.py --arch $__BuildArch --api $__ApiLevel --install-dir $__ToolchainDir
 
 # Install the required packages into the toolchain
+# TODO: Add logic to get latest pkg version instead of specific version number
 rm -rf $__Android_Cross_Dir/deb/
 rm -rf $__Android_Cross_Dir/tmp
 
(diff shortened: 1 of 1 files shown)


jkotas:
@qmfrederik Does the change look good?

cross/build-android-rootfs.sh
```diff
@@ -124,38 +127,40 @@ dpkg -x $__Android_Cross_Dir/deb/libuuid-dev_1.0.3_$__AndroidArch.deb $__Android
...
 dpkg -x $__Android_Cross_Dir/deb/liblzma-dev_5.2.3_$__AndroidArch.deb $__Android_Cross_Dir/tmp/$__AndroidArch/
 dpkg -x $__Android_Cross_Dir/deb/liblzma_5.2.3_$__AndroidArch.deb $__Android_Cross_Dir/tmp/$__AndroidArch/
+dpkg -x $__Android_Cross_Dir/deb/libunwind-dev_1.2.20170304_$__AndroidArch.deb $__Android_Cross_Dir/tmp/$__AndroidArch/
+dpkg -x $__Android_Cross_Dir/deb/libunwind_1.2.20170304_$__AndroidArch.deb $__Android_Cross_Dir/tmp/$__AndroidArch/
 
 cp -R $__Android_Cross_Dir/tmp/$__AndroidArch/data/data/com.termux/files/usr/* $__ToolchainDir/sysroot/usr/
 
 # Prepare libunwind
-if [ ! -d $__libunwind_Dir ]; then
+#if [ ! -d $__libunwind_Dir ]; then
```
qmfrederik: I would suggest to remove the entire block of code instead of commenting it out; this code is obsolete now that Termux packages libunwind and Git will still keep the old code in its history.

qmfrederik:
@jkotas I left one comment about code which I think can be removed rather than just commented out; other than that, looks good to me.

It's unfortunate Termux removes the old packages from its repository when they publish new packages, this will cause the build script to break every time Termux updates a package.

cross/build-android-rootfs.sh
cydhaselton: @qmfrederik I left it in just in case the Termux package doesn't include your fix for libunwind; would make it easier to roll back.

Alternative: remove and copy from corefx script if needed. Thoughts?

cydhaselton:
@jkotas Are there any other changes I should make for this PR?

cross/build-android-rootfs.sh
qmfrederik: @cydhaselton I think git is the easiest way to roll back changes; so I would just remove the lines.

I think that in general, the idea is to use Git to retrieve any old/deleted code, and use issues in GitHub to keep track of to-do items.

cydhaselton:
@jkotas Good point. Latest commit removes commented code.

cross/build-android-rootfs.sh
cydhaselton: @qmfrederik Makes sense. Lines removed in latest commit

jkotas:
Thanks!

janvorli:
@cydhaselton, @qmfrederik - Regarding the changing version - what if we detected the version present in the termux folder dynamically? You can get the  http://termux.net/dists/stable/main/binary-aarch64/ and it returns a html file with a list of files in the folder. You can even pass it a format option to pick one of three formats, the `?F=0` being the simplest. The downside is that it is html formatted, but I guess it would be easy to parse with e.g. awk.

PR #10484, opened 2017-03-25
Correctly update bitvector in optAddVnAssertionMapping

mikedn:
When short bitvectors are used only the first assertion was stored in the map stored bitvector, all subsequent assertions were stored into copies of that bitvector.

### DIFF
--- a/src/jit/assertionprop.cpp
+++ b/src/jit/assertionprop.cpp
@@ -1483,13 +1483,15 @@ void Compiler::optPrintVnAssertionMapping()
  */
 void Compiler::optAddVnAssertionMapping(ValueNum vn, AssertionIndex index)
 {
-    ASSERT_TP cur;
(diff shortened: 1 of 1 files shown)


mikedn:
jit-diff summary:
```
Total bytes of diff: -5 (0.00 % of base)
    diff is an improvement.
          -3 : Microsoft.CodeAnalysis.CSharp.dasm (0.00 % of base)
          -2 : System.Reflection.Metadata.dasm (0.00 % of base)
2 total files with size differences (2 improved, 0 regressed).
          -3 : Microsoft.CodeAnalysis.CSharp.dasm - Lexer:AdvanceIfMatches(char):bool:this
          -2 : System.Reflection.Metadata.dasm - SignatureTypeEncoder:FunctionPointer(ubyte,int,int):struct:this
2 total methods with size differences (2 improved, 0 regressed).
```

mikedn:
cc @JosephTremoulet

PR #10486, opened 2017-03-26
Change typedef for BOOL in gcenv.base.h to int

adityamandaleeka:
I'm not sure why this was defined this way here before, but some GC changes I'm about to make were breaking the GCSample project because of this.

### DIFF
--- a/src/gc/env/gcenv.base.h
+++ b/src/gc/env/gcenv.base.h
@@ -37,7 +37,7 @@
 // Aliases for Win32 types
 //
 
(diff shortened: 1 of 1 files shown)


adityamandaleeka:

jkotas:
It was likely done to match https://github.com/dotnet/corert/blob/2b686a511ce22753b11a3d8152a9cf81b59350d8/src/Native/Runtime/inc/CommonTypes.h#L38

PR #10490, opened 2017-03-27
[WIP]Ensure corerun using the given executable path.

### LINKED ISSUE #5631: CoreRun tries to execute .exe from incorrect path.
This was tested and observed at both ARM/Linux and x64/Linux.

Reproduction
- Install coreclr & corefx at directory `/A/`
- Install `a.exe` at directory `/B/`
- Create another `a.exe` (different from `/B/a.exe`) at `/A/` (for a more dramatic failure, create a null file, `/A/a.exe`)
- At `/A/`, run `./corerun /B/a.exe`

Expected
- The file `/B/a.exe` is executed.

Result
- The file `/A/a.exe` is executed.

Note that if `/A/a.exe` does not exist, it properly executed `B/a.exe`.

mazong1123:
Fix the bug that corerun trying to load the executable from current directory even if the user has specified a full path of the executable.

Fix #5631

### DIFF
--- a/src/binder/assemblybinder.cpp
+++ b/src/binder/assemblybinder.cpp
@@ -554,7 +554,7 @@ namespace BINDER_SPACE
                                          /* in */  BOOL                 fNgenExplicitBind,
                                          /* in */  BOOL                 fExplicitBindToNativeImage,
                                          /* in */  BOOL                 fUseExplicitFilePath,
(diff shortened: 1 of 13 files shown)


mazong1123:
Some tests on Windows failed. I'm gonna fix them.

PR #10493, opened 2017-03-27
Path may contain spaces. Wrap in quotes.

### LINKED ISSUE #10492: Build instructions: registering DLL fails due to spaces.
The Windows [build instructions](https://github.com/dotnet/coreclr/blob/master/Documentation/building/windows-instructions.md) specify that msdia120.dll needs to be registered. However, the provided command will generally fail since %VSINSTALLDIR% expands to a path with spaces in it.

Error:
image

Need to wrap the path with quotes. Fix upcoming.

bendono:
Fixes #10492. Added quotes around path.


jkotas:
Thanks!

Issue #6757, opened 2016-08-16
Jitted Code Pitching Support 


sergign60:
We propose to support releasing memory that is allocated for methods that have been executed and will never be executed again or can be recompiled. That is needed for devices with limited memory capacity. 

Also when GDB JIT feature (#6278) is enabled there is additional memory consumption (~700 bytes on ARM32 per method with several lines) for ELF debugging information. Such memory can be released when jitted code for method itself released.

The existing execution model of coreclr without releasing memory for executed code and debugging info creates serious obstacles for applications running on devices with limited memory or even can make impossible their execution in certain cases.

@papaslavik @brucehoult @Dmitri-Botcharnikov @leemgs @lemmaa @wateret

jkotas:
The earlier versions of CLR had this mechanism. It was called code-pitching but it was removed because of it was very unreliable. The code for it was part of "Shared Source Common Language Infrastructure" that you can find around if you would like to take a look. The hard problem with code pitching is to guarantee that the method being discarder is not executing, and that nobody is holding pointer to it.

jkotas:
cc @noahfalk @dotnet/jit-contrib

sergign60:
Is it possible to define that the method is not executing by stopping the program execution and inspection of stack frames for all threads?

jkotas:
Yes, the original code pitching implementation have done something similar.

sergign60:
except for FCALLs? Because I read in comments 
// FCall is a high-performance alternative to ECall. Unlike ECall, FCall
// methods do not necessarily create a frame.   Jitted code calls directly
// to the FCall entry point.   It is possible to do operations that need
// to have a frame within an FCall, you need to manually set up the frame
// before you do such operations.

jkotas:
FCalls are a problem because they are implemented in C++ or hand-written assembly, and so they cannot be discarder.

jkotas:
> needed for devices with limited memory capacity

It may be a good idea to do analysis on how much memory can be saved by doing this for the typical applications. The JITed code does not typically dominate the working set.

papaslavik:
For data we have garbage collection... So it sort of has a control on the memory increase for data.

@jkotas I wonder if you mean the meta-data more than just data compared to the code here...

papaslavik:
Also it's interesting if it's possible to write a test using serialization an de-serialization or something like that which causes an unbounded increase in the memory for the code of dynamic methods.

papaslavik:
[...]

PR #6769, opened 2016-08-17
Jitted Code Dropping Feature implemented

sergign60:
The given pull request proposes a way for resolving issue "Jitted Code Dropping Support" #6757
Its distinctive features and algorithm are:
1. All its code is under #if defined(FEATURE_JIT_DROPPING) and doesn't mess up with other code
2. This feature is working only if the options INTERNAL_JitDropEnabled != 0 and  INTERNAL_JitDropMemThreshold > 0
3. Jitted code can be dropped only for methods that are not Dynamic or FCall
4. If the size of the generated native code exceeds the value of INTERNAL_JitDropMethodSizeThreshold this code is placed in the special heap code list.  Each heap block in this list stores the code for only one method and has the sufficient size for the code of a method aligned to 4K. The pointers to such methods are stored in the "CalledMethods" hash map.
5. If the entrypoint of a method is backpatched this method is excluded from the "CalledMethods" hash map.
6. When the total size of the generated native code exceeds the value of  INTERNAL_JitDropMemThreshold option, the execution of the program is stopped and stack frames for all the threads are inspected and pointers to methods being executed are stored in the "ExecutedMethods" hash map
7. The code for all the methods from the "CalledMethods" that are not in the "ExecutedMethods" is dropped. (All heap blocks for these methods are set in the initial state and can be reused for newly compiled methods,  pointers to the code for non-executed methods are set to NULL).
8. The coreclr code with this feature is built by the option

./build.sh cmakeargs -DFEATURE_JIT_DROPPING=true

PR #10496, opened 2017-03-27
Jitted Code Pitching Feature implemented

sergign60:
The given pull request proposes a way for resolving issue "Jitted Code Dropping Support" #6757
Its distinctive features and algorithm are:

    1. All its code is under #if defined(FEATURE_JIT_PITCHING) and doesn't mess up with other code
    2. This feature is working only if the options INTERNAL_JitPitchEnabled != 0 and INTERNAL_JitPitchMemThreshold > 0
    3. Jitted code can be pitched only for methods that are not Dynamic, FCall or Virtual
    4. If the size of the generated native code exceeds the value of INTERNAL_JitPitchMethodSizeThreshold this code is placed in the special heap code list. Each heap block in this list stores the code for only one method and has the sufficient size for the code of a method aligned to 4K. The pointers to such methods are stored in the "CalledMethods" hash map.
    5. If the entrypoint of a method is backpatched this method is excluded from the "CalledMethods" hash map and stored in "NotForPitchingMethods" hashmap.
    6. When the total size of the generated native code exceeds the value of INTERNAL_JitPitchMemThreshold option, the execution of the program is stopped and stack frames for all the threads are inspected and pointers to methods being executed are stored in the "ExecutedMethods" hash map
    7. The code for all the methods from the "CalledMethods" that are not in the "ExecutedMethods" is pitched. (All heap blocks for these methods are set in the initial state and can be reused for newly compiled methods, pointers to the code for non-executed methods are set to NULL).
    8. If the code for the given method is pitched once, this method is stored in the "NotForPitchingMethods" hashmap. Thus, if this method is compiled the second time, it is considered as called repeatedly, therefore, pitching for it is inexpedient, and the newly compiled code stored in the usual heap.
    9. The coreclr code with this feature is built by the option

./build.sh cmakeargs -DFEATURE_JIT_PITCHING=true

### DIFF
--- a/src/vm/codeman.cpp
+++ b/src/vm/codeman.cpp
@@ -2186,7 +2186,7 @@ HeapList* EEJitManager::NewCodeHeap(CodeHeapRequestInfo *pInfo, DomainCodeHeapLi
     } CONTRACT_END;
 
     size_t initialRequestSize = pInfo->getRequestSize();
-    size_t minReserveSize = VIRTUAL_ALLOC_RESERVE_GRANULARITY; //     ( 64 KB)           
+    size_t minReserveSize = VIRTUAL_ALLOC_RESERVE_GRANULARITY; //     ( 64 KB)
 
 #ifdef _WIN64
     if (pInfo->m_hiAddr == 0)
@@ -2390,12 +2390,12 @@ void* EEJitManager::allocCodeRaw(CodeHeapRequestInfo *pInfo,
     {
         // Let us create a new heap.
 
-        DomainCodeHeapList *pList = GetCodeHeapList(pInfo->m_pMD, pInfo->m_pAllocator);
+        DomainCodeHeapList *pList = GetCodeHeapList(pInfo, pInfo->m_pAllocator);
         if (pList == NULL)
         {
             // not found so need to create the first one
             pList = CreateCodeHeapList(pInfo);
-            _ASSERTE(pList == GetCodeHeapList(pInfo->m_pMD, pInfo->m_pAllocator));
+            _ASSERTE(pList == GetCodeHeapList(pInfo, pInfo->m_pAllocator));
         }
         _ASSERTE(pList);
 
@@ -2478,22 +2478,28 @@ CodeHeader* EEJitManager::allocCode(MethodDesc* pMD, size_t blockSize, CorJitAll
 
     SIZE_T totalSize = blockSize;
 
+    CodeHeader * pCodeHdr = NULL;
+
+    CodeHeapRequestInfo requestInfo(pMD);
+#if defined(FEATURE_JIT_PITCHING)
+    if (pMD && pMD->IsPitchable() && CLRConfig::GetConfigValue(CLRConfig::INTERNAL_JitPitchMethodSizeThreshold) < blockSize)
+    {
+        requestInfo.SetDynamicDomain();
+    }
+#endif
+
 #if defined(USE_INDIRECT_CODEHEADER)
     SIZE_T realHeaderSize = offsetof(RealCodeHeader, unwindInfos[0]) + (sizeof(T_RUNTIME_FUNCTION) * nUnwindInfos); 
 
     // if this is a LCG method then we will be allocating the RealCodeHeader
     // following the code so that the code block can be removed easily by 
     // the LCG code heap.
-    if (pMD->IsLCGMethod())
+    if (requestInfo.IsDynamicDomain())
     {
         totalSize = ALIGN_UP(totalSize, sizeof(void*)) + realHeaderSize;
         static_assert_no_msg(CODE_SIZE_ALIGN >= sizeof(void*));
     }
 #endif  // USE_INDIRECT_CODEHEADER
-
-    CodeHeader * pCodeHdr = NULL;
-
-    CodeHeapRequestInfo requestInfo(pMD);
 
     // Scope the lock
     {
@@ -2521,7 +2527,7 @@ CodeHeader* EEJitManager::allocCode(MethodDesc* pMD, size_t blockSize, CorJitAll
         pCodeHdr = ((CodeHeader *)pCode) - 1;
 
 #ifdef USE_INDIRECT_CODEHEADER
-        if (pMD->IsLCGMethod())
+        if (requestInfo.IsDynamicDomain())
         {
             pCodeHdr->SetRealCodeHeader((BYTE*)pCode + ALIGN_UP(blockSize, sizeof(void*)));
         }
@@ -2550,7 +2556,7 @@ CodeHeader* EEJitManager::allocCode(MethodDesc* pMD, size_t blockSize, CorJitAll
     RETURN(pCodeHdr);
 }
 
-EEJitManager::DomainCodeHeapList *EEJitManager::GetCodeHeapList(MethodDesc *pMD, LoaderAllocator *pAllocator, BOOL fDynamicOnly)
+EEJitManager::DomainCodeHeapList *EEJitManager::GetCodeHeapList(CodeHeapRequestInfo *pInfo, LoaderAllocator *pAllocator, BOOL fDynamicOnly)
 {
     CONTRACTL {
         NOTHROW;
@@ -2564,7 +2570,7 @@ EEJitManager::DomainCodeHeapList *EEJitManager::GetCodeHeapList(MethodDesc *pMD,
 
     // get the appropriate list of heaps
     // pMD is NULL for NGen modules during Module::LoadTokenTables
-    if (fDynamicOnly || (pMD != NULL && pMD->IsLCGMethod()))
+    if (fDynamicOnly || (pInfo != NULL && pInfo->IsDynamicDomain()))
     {
         ppList = m_DynamicDomainCodeHeaps.Table();
         count = m_DynamicDomainCodeHeaps.Count();
@@ -2605,7 +2611,7 @@ HeapList* EEJitManager::GetCodeHeap(CodeHeapRequestInfo *pInfo)
 
     // loop through the m_DomainCodeHeaps to find the AppDomain
     // if not found, then create it
-    DomainCodeHeapList *pList = GetCodeHeapList(pInfo->m_pMD, pInfo->m_pAllocator);
+    DomainCodeHeapList *pList = GetCodeHeapList(pInfo, pInfo->m_pAllocator);
     if (pList)
     {
         // Set pResult to the largest non-full HeapList
@@ -2726,7 +2732,7 @@ bool EEJitManager::CanUseCodeHeap(CodeHeapRequestInfo *pInfo, HeapList *pCodeHea
        }
    }
 
-   return retVal; 
+   return retVal;
 }
 
 EEJitManager::DomainCodeHeapList * EEJitManager::CreateCodeHeapList(CodeHeapRequestInfo *pInfo)
--- a/src/vm/codepitchingmanager.cpp
+++ b/src/vm/codepitchingmanager.cpp
@@ -1,0 +1,522 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+// ===========================================================================
+// File: CodePitchingManager.cpp
+//
+
+// ===========================================================================
+// This file contains the implementation for code pitching.
+// Its distinctive features and algorithm are:
+//
+// 1. All its code is under #if defined(FEATURE_JIT_PITCHING) and doesn't mess up with other code
+// 2. This feature is working only if the options INTERNAL_JitPitchEnabled != 0 and INTERNAL_JitPitchMemThreshold > 0
+// 3. Jitted code can be pitched only for methods that are not Dynamic, FCall or Virtual
+// 4. If the size of the generated native code exceeds the value of INTERNAL_JitDPitchMethodSizeThreshold this code is
+//    placed in the special heap code list. Each heap block in this list stores the code for only one method and has the
+//    sufficient size for the code of a method aligned to 4K. The pointers to such methods are stored in the
+//    "PitchingCandidateMethods" hash map.
+// 5. If the entrypoint of a method is backpatched this method is excluded from the "PitchingCandidateMethods" hash map
+//    and stored in "NotForPitchingMethods" hashmap.
+// 6. When the total size of the generated native code exceeds the value of INTERNAL_JitPitchMemThreshold option, the
+//    execution of the program is stopped and stack frames for all the threads are inspected and pointers to methods
+//    being executed are stored in the "ExecutedMethods" hash map
+// 7. The code for all the methods from the "PitchingCandidateMethods" that are not in the "ExecutedMethods" is pitched.
+//    (All heap blocks for these methods are set in the initial state and can be reused for newly compiled methods, pointers
+//     to the code for non-executed methods are set to nullptr).
+// 8. If the code for the given method is pitched once, this method is stored in the "NotForPitchingMethods" hashmap. Thus,
+//    if this method is compiled the second time, it is considered as called repeatedly, therefore, pitching for it is inexpedient,
+//    and the newly compiled code stored in the usual heap.
+// 9. The coreclr code with this feature is built by the option
+//     ./build.sh cmakeargs -DFEATURE_JIT_PITCHING=true
+// ===========================================================================
+
+#include "common.h"
+
+#ifndef DACCESS_COMPILE
+
+#if defined(FEATURE_JIT_PITCHING)
+
+#include "nibblemapmacros.h"
+#include "threadsuspend.h"
+
+static PtrHashMap* s_pPitchingCandidateMethods = nullptr;
+static PtrHashMap* s_pPitchingCandidateSizes = nullptr;
+static SimpleRWLock* s_pPitchingCandidateMethodsLock = nullptr;
+
+static PtrHashMap* s_pExecutedMethods = nullptr;
+static SimpleRWLock* s_pExecutedMethodsLock = nullptr;
+
+static PtrHashMap* s_pNotForPitchingMethods = nullptr;
+static SimpleRWLock* s_pNotForPitchingMethodsLock = nullptr;
+
+#ifdef _DEBUG
+static PtrHashMap* s_pPitchedMethods = nullptr;
+static SimpleRWLock* s_pPitchedMethodsLock = nullptr;
+#endif
+
+static ULONG s_totalNCSize = 0;
+static SimpleRWLock* s_totalNCSizeLock = nullptr;
+
+static ULONG s_jitPitchedBytes = 0;
+
+static INT64 s_JitPitchLastTick = 0;
+
+static bool s_JitPitchInitialized = false;
+
+
+static BOOL IsOwnerOfRWLock(LPVOID lock)
+{
+    // @TODO - SimpleRWLock does not have knowledge of which thread gets the writer
+    // lock, so no way to verify
+    return TRUE;
+}
+
+static void CreateRWLock(SimpleRWLock** lock)
+{
+    if (*lock == nullptr)
+    {
+        void *pLockSpace = SystemDomain::GetGlobalLoaderAllocator()->GetLowFrequencyHeap()->AllocMem(S_SIZE_T(sizeof(SimpleRWLock)));
+        SimpleRWLock *pLock = new (pLockSpace) SimpleRWLock(COOPERATIVE_OR_PREEMPTIVE, LOCK_TYPE_DEFAULT);
+
+        if (FastInterlockCompareExchangePointer(lock, pLock, NULL) != NULL)
+            SystemDomain::GetGlobalLoaderAllocator()->GetLowFrequencyHeap()->BackoutMem(pLockSpace, sizeof(SimpleRWLock));
+    }
+}
+
+static PtrHashMap* CreateHashMap(SimpleRWLock* rwLock)
+{
+    PtrHashMap *pMap = new (SystemDomain::GetGlobalLoaderAllocator()->GetLowFrequencyHeap()) PtrHashMap();
+    LockOwner lock = {rwLock, IsOwnerOfRWLock};
+    pMap->Init(32, nullptr, FALSE, &lock);
+    return pMap;
+}
+
+static void InitializeJitPitching()
+{
+    if (!s_JitPitchInitialized)
+    {
+        CreateRWLock(&s_pNotForPitchingMethodsLock);
+        CreateRWLock(&s_pPitchingCandidateMethodsLock);
+        CreateRWLock(&s_totalNCSizeLock);
+
+        {
+            SimpleReadLockHolder srlh(s_pNotForPitchingMethodsLock);
+            if (s_pNotForPitchingMethods == nullptr)
+            {
+                s_pNotForPitchingMethods = CreateHashMap(s_pNotForPitchingMethodsLock);
+            }
+        }
+
+        {
+            SimpleReadLockHolder srlh(s_pPitchingCandidateMethodsLock);
+            if (s_pPitchingCandidateMethods == nullptr)
+            {
+                s_pPitchingCandidateMethods = CreateHashMap(s_pPitchingCandidateMethodsLock);
+                s_pPitchingCandidateSizes = CreateHashMap(s_pPitchingCandidateMethodsLock);
+            }
+        }
+
+        s_JitPitchInitialized = true;
+    }
+}
+
+static COUNT_T GetFullHash(MethodDesc* pMD)
+{
+    const char *moduleName = pMD->GetModule()->GetSimpleName();
+
+    COUNT_T hash = HashStringA(moduleName);         // Start the hash with the Module name
+
+    SString className, methodName, methodSig;
+
+    pMD->GetMethodInfo(className, methodName, methodSig);
+
+    hash = HashCOUNT_T(hash, className.Hash());   // Hash in the name of the Class name
+    hash = HashCOUNT_T(hash, methodName.Hash());  // Hash in the name of the Method name
+    hash = HashCOUNT_T(hash, 0xffffffff & (ULONGLONG)pMD);
+
+    return hash;
+}
+
+bool MethodDesc::IsPitchable()
+{
+    if ((CLRConfig::GetConfigValue(CLRConfig::INTERNAL_JitPitchEnabled) == 0) ||
+        (CLRConfig::GetConfigValue(CLRConfig::INTERNAL_JitPitchMemThreshold) == 0))
+        return FALSE;
+
+    InitializeJitPitching();
+
+    if (IsLCGMethod() || IsVtableMethod() || IsInterface() || IsVirtual())
+        return FALSE;
+
+    _ASSERTE(s_pNotForPitchingMethodsLock != nullptr && s_pNotForPitchingMethods != nullptr);
+
+    {
+        SimpleReadLockHolder srlh(s_pNotForPitchingMethodsLock);
+        UPTR key = (UPTR)GetFullHash(this);
+        MethodDesc *pFound = (MethodDesc *)s_pNotForPitchingMethods->LookupValue(key, (LPVOID)this);
+        if (pFound != (MethodDesc *)INVALIDENTRY)
+        {
+            return FALSE;
+        }
+    }
+    return TRUE;
+}
+
+EXTERN_C bool LookupOrCreateInNotForPitching(MethodDesc* pMD)
+{
+    CONTRACTL
+    {
+        MODE_COOPERATIVE;
+        GC_TRIGGERS;
+        THROWS;
+    }
+    CONTRACTL_END;
+
+    if (pMD != nullptr && pMD->IsPitchable())
+    {
+        UPTR key = (UPTR)GetFullHash(pMD);
+
+        _ASSERTE(s_pNotForPitchingMethodsLock != nullptr && s_pNotForPitchingMethods != nullptr);
+
+        {
+             SimpleReadLockHolder srlh(s_pNotForPitchingMethodsLock);
+             MethodDesc *pFound = (MethodDesc *)s_pNotForPitchingMethods->LookupValue(key, (LPVOID)pMD);
+             if (pFound != (MethodDesc *)INVALIDENTRY)
+                 return TRUE;
+        }
+
+        {
+             SimpleWriteLockHolder swlh(s_pNotForPitchingMethodsLock);
+             s_pNotForPitchingMethods->InsertValue(key, (LPVOID)pMD);
+        }
+    }
+    return FALSE;
+}
+
+static void LookupOrCreateInPitchingCandidate(MethodDesc* pMD, ULONG sizeOfCode)
+{
+    CONTRACTL
+    {
+        MODE_COOPERATIVE;
+        GC_TRIGGERS;
+        THROWS;
+    }
+    CONTRACTL_END;
+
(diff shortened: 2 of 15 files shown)


sergign60:
https://github.com/dotnet/coreclr/pull/6769

sergign60:
https://github.com/dotnet/coreclr/issues/6757

sergign60:
Testing of Checked and Debug builds on the CoreCLR test  suite with the environment variables
```
COMPlus_JitDropMemThreshold=1
COMPlus_JitDropMethodSizeThreshold=1
COMPlus_JitDropEnabled=1
```
does not show any regression

sergign60:
@Dmitri-Botcharnikov @papaslavik @seanshpark

sergign60:
@dotnet-bot Windows_NT x86 Checked Build and Test

sergign60:

janvorli:
CC: @noahfalk

noahfalk:
I haven't had a chance to go through this in depth, I'll try to make time to give it more scrutiny soon (tomorrow?). For now a few thoughts on various areas:

1) Correctness - I haven't scrutinized carefully, but I suspect there are still race conditions which would cause this code to crash, even if the CoreCLR/CoreFx unit tests aren't stressing the runtime hard enough to hit them. I don't know what error-rate you'd consider acceptable for your usage, but if we wanted to get rigorous someone will need to do a full escape analysis on that code pointer to determine everywhere it might still be stored at the time it is being considered for deletion. For example consider any of the call paths which invoke GetMethodEntryPoint(), GetNativeCode(), GetSingleCallableAddrOfCode(), etc. The code pointer could be sitting in a register or stack location corresponding to that C++ code that will eventually be invoked when the caller returns. The JIT may have no explicit invariant that prevents it from caching the call target read via an indirection cell across a GC safe-point, in which case the code pointer could be stored in jitted code registers too.

2) Compat with diagnostic tools - I think I already commented on the earlier iteration of this proposal that debuggers, profiler, and ETW all appear likely to fail when the code is deleted. I don't consider it a bar you have to meet to check-in, or even to use the feature productively in some scenario of your own, but it would definately block us from enabling it by default.

3) Coordination with other jit updating technologies - I've got tiered compilation in progress over in 
#10478 . Initially there are some small areas of overlap that might be good to converge on, such as hooking into MethodDesc::IsPointingToStableNativeCode() and MethodDesc::IsNativeCodeStableAfterInit(). This should let you leverage other work that attempts to handle non-stable code pointers correctly and uniformly. However your requirements to be able to delete code are likely stricter than my requirements for tiered compilation. In your case it is an AV if you call the old code pointer whereas my PR only provides for eventual convergence on the most recent code pointer. As the tiered compilation project moves forward I anticipate there will be more opportunities for useful sharing.

sergign60:
@noahfalk 
many thanks for your comments. Just now I'm busy in another project, so I can look through them only at weekend.  So now I can say only that I fully agree with 2.

noahfalk:
Its cool stuff you are working on here @sergign60. I've been taking a closer look through this and had some follow up comments. IMO the minimum bar to check this in should be:
1) Current runtime behavior doesn't regress (using the off-by-default ifdef mostly solves this)
2) It doesn't introduce unnecessary complexity/confusion for other developers likely to work in or around this code in the future.
I do not think that correctness, completeness, or performance of the implementation needs to be a blocking issue as long as we've gotten it suitably decoupled, though I'm happy to continue pointing out potential issues if you find it useful.


The main thing that prevents this from being ready for checkin are a number of places where it adds some confusion/complexity - some easy to remedy but others might take a little more legwork:
1) Throughout our code base the term we usually use for deleting code is 'code pitching', including various places in public documented APIs. It would be good to use matching terminology in this feature.
2) How about renaming CalledMethods to PitchingCandidateMethods? I think that better reflects the intended usage of that set no?
3) We already have a CodeHeap implementation that supports releasing/reallocating methods -  HostCodeHeap in dynamicmethod.cpp. Rather than add a new 3rd kind of heap allocation it would be good to understand what it would take to make the existing m_isDynamicDomain in CodeHeapRequestInfo suitable for your needs. This could eliminate a lot of #ifdefs that are otherwise scattered throughout the allocator and code manager.
4) A lot of the logic in prestub.cpp should probably be moved into a new dedicated file/type, perhaps CodePitchingManager.cpp? This manager and the various sets being tracked should exist per-AppDomain, as AppDomain is the natural container for code in general. I haven't thought far enough ahead to figure out if this is a likely to be a long term stable location for this functionality, but it seems reasonably compartmentalized for now.
5) Can we avoid snaking an extra argument through MethodDesc::DoPrestub and MethodDesc::MakeJitWorker? It appears it roughly corresponds to the define FEATURE_JIT_DROPPING being true, so I am hoping extra runtime arguments aren't necessary.
6) In method.cpp we should try to avoid adding yet more complexity to TryGetMultiCallableAddrOfCode. In my PR I'm about to change IsPointingToNativeCode() -> IsPointingToStableNativeCode() and it would be a natural fit to mark pitchable methods as always returning FALSE from that method. In that second case it appears this PR created a condition in which the method claims to have a stable entry point but it doesn't actually have a stable entry point. Rather than redefining the meaning of HasStableEntryPoint() its probably better to modify the logic elsewhere so that we actually give those methods stable entrypoints (with a fixup precode) or we avoid ever marking the entrypoint as stable.
7) It would be good to add some comments describing the current state of the feature to the code, so that other devs encountering it are able to form the right expectations about it.

sergign60:
@noahfalk thanks for your comments. I'll look at them more closely at the nearest weekend.

sergign60:
@noahfalk thanks again for your comments. As you can see I did renaming work. 
3. & 5. & 6. Sure, I'll think through this variants

sergign60:
@noahfalk 
about 5. As I see another way is using CORJIT_FLAGS (+CORJIT_FLAG_JIT_PITCHING=41) by setting it in PreStubWorker. But this way requires additional argument too. What do you think?

sergign60:
@noahfalk 
I've fixed the item 5. Please look at the diffs.

sergign60:
@noahfalk 
Next week I'll be working on the item 3 in your list.

noahfalk:
@sergign60 - Sorry I've been out of touch for a while, I just got back from vacation in China. I'll take a look as soon as I can

sergign60:
@noahfalk  Please don't worry :)

src/vm/prestub.cpp
```diff
@@ -269,6 +275,11 @@ PCODE MethodDesc::MakeJitWorker(COR_ILMETHOD_DECODER* ILHeader, CORJIT_FLAGS fla
          GetMethodTable()->GetDebugClassName(),
          m_pszDebugMethodName));
 
+#if defined(FEATURE_JIT_PITCHING)
+    bool fDoJitPitching = flags.IsSet(CORJIT_FLAGS::CORJIT_FLAG_DO_PITCHING);
+    CheckStacksAndPitch(fDoJitPitching);
```
noahfalk: I was wondering if you needed any argument at all. For example imagine the code here was:

    CheckStacksAndPitch(TRUE);

Would that break the feature in some way? It wasn't clear to me what is being gained by identifying only the calls that originated in PreStubWorker rather than doing code pitching any time we JIT?

src/vm/prestub.cpp
```diff
@@ -487,6 +498,7 @@ PCODE MethodDesc::MakeJitWorker(COR_ILMETHOD_DECODER* ILHeader, CORJIT_FLAGS fla
 
             EX_TRY
             {
+                PTR_PCODE addrOfSlot = GetAddrOfSlot();
```
noahfalk: This line appears unneeded

src/vm/prestub.cpp
```diff
@@ -1321,7 +1344,15 @@ PCODE MethodDesc::DoPrestub(MethodTable *pDispatchingMT)
 #endif
         LOG((LF_CLASSLOADER, LL_INFO10000,
                 "    In PreStubWorker, method already jitted, backpatching call point\n"));
-
+#if defined(FEATURE_JIT_PITCHING)
+        if ((CLRConfig::GetConfigValue(CLRConfig::INTERNAL_JitPitchEnabled) != 0) &&
```
noahfalk: Lets further lower the footprint by wrapping all this logic inside a single call. Something like:

    #if defined(FEATURE_JIT_PITCHING)
        MarkMethodNotPitchable();
    #endif

src/vm/prestub.cpp
```diff
@@ -2108,6 +2139,17 @@ EXTERN_C PCODE STDCALL ExternalMethodFixupWorker(TransitionBlock * pTransitionBl
                 pCode = PatchNonVirtualExternalMethod(pMD, pCode, pImportSection, pIndirection);
             }
         }
+
+#if defined (FEATURE_JIT_PITCHING)
```
noahfalk: Similar to above, lets reduce this down to a single call out:

    #if defined (FEATURE_JIT_PITCHING)
        MarkMethodNotPitchingCandidate(pMD);
    #endif

Your call on naming and whether or not this method should be the same or different than the one above (I note one added to the NotForPitching set, the other didn't).

noahfalk:
Overall 1 and 2 look good, 4 looked most of the way there but I left a few comments in spots I hope that we can further reduce the footprint. Originally I was thinking for 4 you would put the static code inside a new class, but on further thought isolating it with the define and file seems perfectly sufficient to meet the bar of giving organization clarity to other developers.

5 I left you a comment on above.

3, 6 I know you said were coming so I haven't reviewed anything further in regards to them.

7 I wasn't sure whether you believed it was done or not? The kind of comment I was looking for would describe how well you believe the feature works and what its current limitations are. Something similar to the comment I added here:
https://github.com/dotnet/coreclr/blob/master/src/vm/tieredcompilation.cpp#L22
Apologies if I just missed it.

sergign60:
@noahfalk 
I guess that I've (almost) finished item 3. - switching to HostCodeHeap. Please review when you can.

sergign60:

sergign60:
@noahfalk 
Hi, I think that item 6 is ready too.

sergign60:
@dotnet-bot  test Tizen armel Cross Debug Build

sergign60:

sergign60:

src/jit/codegencommon.cpp
```diff
@@ -3240,7 +3240,11 @@ void CodeGen::genGenerateCode(void** codePtr, ULONG* nativeSizeOfCode)
     }
 #endif // EMIT_TRACK_STACK_DEPTH
 
-    *nativeSizeOfCode                 = codeSize;
+    if (nativeSizeOfCode != nullptr)
+    {
+        *nativeSizeOfCode = codeSize;
+    }
+
```
noahfalk: I'm a little surprised this would be needed. If you found a code path that calls here with nativeSizeOfCode = NULL we should probably fix the caller. But if you let me know what the callstack is that gets here I can help assess.

src/vm/codeman.cpp
```diff
@@ -2186,19 +2189,36 @@ HeapList* EEJitManager::NewCodeHeap(CodeHeapRequestInfo *pInfo, DomainCodeHeapLi
...
-    size_t minReserveSize = VIRTUAL_ALLOC_RESERVE_GRANULARITY; //     ( 64 KB)           
+#if defined(FEATURE_JIT_PITCHING)
+    size_t minReserveSize = 0;
+    if (CLRConfig::GetConfigValue(CLRConfig::INTERNAL_JitPitchEnabled) != 0)
+        minReserveSize = OS_PAGE_SIZE; //     ( 4 KB)
+    else
+        minReserveSize = VIRTUAL_ALLOC_RESERVE_GRANULARITY; //     ( 64 KB)
+#else
+    size_t minReserveSize = VIRTUAL_ALLOC_RESERVE_GRANULARITY; //     ( 64 KB)
+#endif
```
noahfalk: are these size adjustments necessary now that you are using HostCodeHeap? The HostCodeHeap internally uses a free list to re-use space that was previously allocated.

src/vm/codeman.cpp
```diff
@@ -2214,7 +2234,15 @@ HeapList* EEJitManager::NewCodeHeap(CodeHeapRequestInfo *pInfo, DomainCodeHeapLi
     size_t reserveSize = requestAndHeadersSize;
     if (reserveSize < minReserveSize)
         reserveSize = minReserveSize;
+
+#if defined(FEATURE_JIT_PITCHING)
+    if (CLRConfig::GetConfigValue(CLRConfig::INTERNAL_JitPitchEnabled) != 0)
+        reserveSize = ALIGN_UP(reserveSize, OS_PAGE_SIZE);
+    else
+        reserveSize = ALIGN_UP(reserveSize, VIRTUAL_ALLOC_RESERVE_GRANULARITY);
+#else
```
noahfalk: With HostCodeHeap you probably don't want this. The HostCodeHeap does use VirtualAlloc with MEM_RESERVE so it is going to expect that its heaps aren't smaller than the minimum reservation size.

src/vm/method.hpp
```diff
@@ -258,6 +258,17 @@ class MethodDesc
...
+    {
+        LIMITED_METHOD_DAC_CONTRACT;
+
+#if defined(FEATURE_JIT_PITCHING)
+        if (IsPitchable() && HasTemporaryEntryPoint())
+           return FALSE;
+#endif
+        return HasStableEntryPoint();
+    }
+
```
noahfalk: This still feels like it adds a lot of confusion about what it means for the code pointer to be stable so unfortunately I'm also not a big fan of this. Looking at the problem more I've only come up with one solution so far that seems decent: using a fixup precode. If after reading through all of this you object then I'm happy to brainstorm further with you, but there are a few reasons why I think its a good choice:
a) It will align with concepts already in the code making it much easier to understand, and more likely to 'just work' when someone else changes code here without explicitly retesting or reasoning about code pitching.
b) It reuses machinery we have used in other scenarios for a long time, so you get the benefit of that in the field testing. Its probably going to be more reliable that the scheme we've got here and I am not aware of any known issues.
c) It should be easy to convert from where you are now to the fixup prestub.

In terms of what that solution looks like, not sure if you have familiarized yourself with the concept already but if not a fixup precode is a little fragment of code that can be rewritten as needed. It can be used as the stable entrypoint of a method regardless whether that method has code jitted or not. Initially the fragment of code looks like a single instruction "call PrecodeFixupThunk". When that code is called it will run PrecodeFixupThunk, which then calls ThePreStub, PreStubWorker, etc. After the code is generated the fixup prestub will be backpatched so that it now reads "jmp <addr_of_jitted_code>". You can also call Reset (ie https://github.com/dotnet/coreclr/blob/master/src/vm/method.cpp#L2483) to revert back to the state where the method will call the prestub again.

In practical terms to use it I think you only need to do this:
a) Add IsPitchable() to the list of conditions that make IsNativeCodeStableStableAfterInit() return false. That will cause DoPrestub() to automatically allocate and use the precode for you: https://github.com/dotnet/coreclr/blob/master/src/vm/prestub.cpp#L1370
https://github.com/dotnet/coreclr/blob/master/src/vm/prestub.cpp#L1531

b) When using a precode, you will need to have to have allocated a NativeCodeSlot. This where we hold the pointer to the jitted code because the stable entrypoint pointer has now been used to point at the precode. Update NeedsNativeCodeSlot to return TRUE for pitchable or potentially pitchable methods:
https://github.com/dotnet/coreclr/blob/master/src/vm/methodtablebuilder.cpp#L7044

c) In your new PitchNativeCode function, call Precode::Reset() to point the precode back to the prestub. You would no longer clear the stable entrypoint or clear the HasStableEntrypoint bit. You can call GetNativeCode() to get the previous jitted code pointer, use the current FreeCodeMemory function to free it, and then use SetNativeCodeInterlocked(NULL) to clear the slot. Technically you probably still have narrow race conditions with some other code in prestub, but even if you don't want to do the careful analysis to find them all its probably a much smaller set of potential races than the current mechanism is exposed to.

src/vm/prestub.cpp
noahfalk: Just a ping that this is still an open issue. Feel free to resolve it at whatever pace works for you.

src/vm/method.hpp
sergign60: @noahfalk Thanks a lot! Unfortunately I'll go to vacation for next two weeks. By the way, my measurements show that about 70% of code are pitched now on the CoreCLR test suite.

src/jit/codegencommon.cpp
sergign60: @noahfalk 
I set this checking when I didn't know well coreclr just in case. No didn't see any code path with null nativeSizeOfCode

src/vm/codeman.cpp
sergign60: @noahfalk I'll remove this

src/vm/codeman.cpp
sergign60: @noahfalk I'll remove this

src/vm/method.hpp
noahfalk: No worries, this is your feature so you get to set the pace, whether that is having it ready tomorrow or next year ; ) Enjoy your vacation!

sergign60:
@noahfalk  I did switching to precode. 
If it is good for you I'll have two items
1 Deleting extra argument through MethodDesc::DoPrestub and MethodDesc::MakeJitWorker
2. Commentaries

sergign60:
@noahfalk 
the latest test results 
```
± env | grep Pitch
COMPlus_JitPitchMemThreshold=1
COMPlus_JitPitchMethodSizeThreshold=1
COMPlus_JitPitchEnabled=1
COMPlus_JitPitchPrintStat=1

=======================
     Test Results
=======================
# CoreCLR Bin Dir  : ./coreclr/bin/Product/Linux.x64.Debug
# Tests Discovered : 7102
# Passed           : 6392
# Failed           : 0
# Skipped          : 710
=======================
```

noahfalk:
@sergign60 - sorry I wasn't clear if you were waiting for me to review again or you planned to do more work before the review?

sergign60:
@noahfalk I planned to do more work before the next review.

sergign60:
@noahfalk Please do reviewing.

src/vm/codeman.cpp
```diff
@@ -2121,6 +2121,9 @@ void CodeHeapRequestInfo::Init()
         m_pAllocator = m_pMD->GetLoaderAllocatorForCode();
     m_isDynamicDomain = (m_pMD != NULL) ? m_pMD->IsLCGMethod() : false;
     m_isCollectible = m_pAllocator->IsCollectible() ? true : false;
+#if defined(FEATURE_JIT_PITCHING)
+    m_isJitPitchedDomain = false;
```
noahfalk: >m_isJitPitchedDomain = false; [](start = 4, length = 29)

It still seems like there is much more special casing here than is necessary. Does it cause issues to use m_DynamicDomainCodeHeaps without creating a 2nd set of heaps?
Similarly it wasn't clear that we needed a seperate m_isJitPitchedDomain field, IsJitPitchableDomain() accessor, or any of the checks in codemap.cpp that appeared to clone the behavior of the dynamic domain case?

Let me know if I missed something, but I'm imagining you could simply set m_isDynamicDomain = (the current logic) || (the condition for pitchable code). Then rely on m_isDynamicDomain to drive all the right behavior throughout the rest of codeman.cpp. You would probably need to duplicate the logic for detecting pitchable methods in EEJitManager::GetCodeHeapList, but two small edits still seems like a significant complexity improvement over the current set of edits.

src/vm/loaderallocator.cpp
```diff
@@ -66,6 +66,7 @@ LoaderAllocator::LoaderAllocator()
     m_pLastUsedCodeHeap = NULL;
     m_pLastUsedDynamicCodeHeap = NULL;
     m_pJumpStubCache = NULL;
+    m_pLastUsedJitPitchedCodeHeap = NULL;
```
noahfalk: If the cleanup I suggested in codeman works, this can be removed too.

src/vm/loaderallocator.hpp
```diff
@@ -115,6 +115,7 @@ class LoaderAllocator
     void * m_pLastUsedCodeHeap;
     void * m_pLastUsedDynamicCodeHeap;
     void * m_pJumpStubCache;
+    void * m_pLastUsedJitPitchedCodeHeap;
```
noahfalk: >void * m_pLastUsedJitPitchedCodeHeap; [](start = 4, length = 37)

If the cleanup I suggested in codeman works, this can be removed too.

noahfalk:
Alright, I think 3 in my list above can probably be taken further (see my comments above), but otherwise things are looking pretty good. Almost done 👍 

(Completely optional on your part, but now that you are using precode for the method switching mechanism you may find that a number of the special cases you avoided pitching before no longer cause problems when pitched)

sergign60:

sergign60:
@noahfalk You're right JitPitchedDomain Code Heap is redundant and deleted

 Test results are:
```
COMPlus_JitPitchMemThreshold=1
COMPlus_JitPitchMethodSizeThreshold=1
COMPlus_JitPitchEnabled=1
COMPlus_JitPitchPrintStat=1

=======================
     Test Results
=======================
# CoreCLR Bin Dir  : ./coreclr/bin/Product/Linux.x64.Debug
# Tests Discovered : 7104
# Passed           : 6394
# Failed           : 0
# Skipped          : 710
=======================
38 minutes and 24 seconds taken to run CoreCLR tests.
```

sergign60:
@noahfalk Added two features: time interval control & jit pitching notification

sergign60:

sergign60:
@noahfalk 
The test results:
```
=======================
     Test Results
=======================
# CoreCLR Bin Dir  : ./coreclr/bin/Product/Linux.x64.Debug
# Tests Discovered : 7104
# Passed           : 6394
# Failed           : 0
# Skipped          : 710
=======================
41 minutes and 4 seconds taken to run CoreCLR tests.
11607 ~/DROP git:(master) ✗ » env | grep COMPl
COMPlus_JitPitchMemThreshold=1
COMPlus_JitPitchMethodSizeThreshold=1
COMPlus_JitPitchEnabled=1
COMPlus_JitPitchPrintStat=1
```

src/pal/prebuilt/inc/xclrdata.h
```diff
@@ -3333,7 +3333,7 @@ enum __MIDL___MIDL_itf_xclrdata_0000_0008_0005
     {
         CLRDATA_METHNOTIFY_NONE	= 0,
         CLRDATA_METHNOTIFY_GENERATED	= 0x1,
-        CLRDATA_METHNOTIFY_DISCARDED	= 0x2
+        CLRDATA_METHNOTIFY_PITCHED	= 0x2
```
noahfalk: This file might be regenerated in the future from an idl we have on desktop. Its probably better to treat this as a public contract we don't change.

noahfalk:
Modulo that one comment I added on naming in the prebuilt header I think we are good to go. Do you think things are complete from your side @sergign60 or did you want to make more changes first?

@janvorli - Are you happy with where this has landed? I haven't been judging this on correctness because it is behind a disabled ifdef by default. I've been analyzing from the perspective of not regressing anything + not adding undue complexity to the rest of the runtime.

@sergign60 - One other heads up, I've got a large change for tiered compilation that will intersect with this change a bit. I suspect it won't be in for at least a week so if we merge your stuff before then I'll merge on top of your change. If your PR stays unchecked in for longer you might need to merge to mine instead. Assuming you go in first and I merge I'll do a best effort to update your work correctly, but part of the disadvantage about being behind an off-by-default define is that incoming changes might break this at anytime and there won't be any tests running to detect or prevent those regressions. We've gotten your feature's footprint small enough that intersecting changes probably won't occur that often though.

sergign60:
@noahfalk  Just statistics on the CoreCLR test suite:

pitched methods:  458052
jitted bytes:  232893742
pitched bytes:  141934600
percentage (pitched/jitted): 60.94%

sergign60:
@noahfalk @janvorli I've removed some special cases from avoided pitching methods. Also I"ve fixed statistics calculation, previous data was wrong :( Now the most of tests gives above 50% percentage.

sergign60:
@noahfalk @janvorli I've removed some special cases from avoided pitching methods. Also I"ve fixed statistics calculation, previous data was wrong :( Now the most of tests gives above 50% percentage.
=======================
     Test Results
=======================
# CoreCLR Bin Dir  : ./coreclr/bin/Product/Linux.x64.Debug
# Tests Discovered : 7104
# Passed           : 6394
# Failed           : 0
# Skipped          : 710
=======================

sergign60:
@noahfalk @janvorli  
 I've removed some special cases from avoided pitching methods. Also I"ve fixed statistics calculation, previous data was wrong :( Now the most of tests gives above 50% percentage.
=======================
     Test Results
=======================
# CoreCLR Bin Dir  : ./coreclr/bin/Product/Linux.x64.Debug
# Tests Discovered : 7104
# Passed           : 6394
# Failed           : 0
# Skipped          : 710
=======================

sergign60:
@noahfalk @janvorli

noahfalk:
I'm happy 👍 

@janvorli - I know you are pretty busy at the moment. Could you let us know if you intended to review this, if you'd like me to find someone else in your stead, or you want to go ahead and merge this without further review? Thanks!

janvorli:
@noahfalk I'd like to find some time and review it too. I hope I'll be able to do that tomorrow morning.

src/vm/codepitchingmanager.cpp
```diff
@@ -0,0 +1,526 @@
...
+                        MethodDesc *pMD = (MethodDesc *) i.GetValue();
+                        UPTR key = (UPTR)GetFullHash(pMD);
+                        MethodDesc *pFound = (MethodDesc *)s_pExecutedMethods->LookupValue(key, (LPVOID)pMD);
+                        ++i;
+                        if (pFound == (MethodDesc *)INVALIDENTRY)
+                        {
+                            pMD->PitchNativeCode();
+                        }
+                    }
+                    for (PtrHashMap::PtrIterator i = s_pExecutedMethods->begin(); !i.end();)
```
janvorli: A nit - can you please reuse the `i` from the above? It is confusing in the debugger when there are two variables of the same name.

src/vm/codepitchingmanager.cpp
```diff
@@ -0,0 +1,526 @@
...
+    CreateRWLock(&s_pPitchingCandidateMethodsLock);
+
+    // Now we have a lock we can use to synchronize the remainder of the init.
+
+    UPTR key = (UPTR)GetFullHash(pMD);
+
+    if (s_pPitchingCandidateMethods == NULL)
+    {
+        SimpleWriteLockHolder swlh(s_pPitchingCandidateMethodsLock);
+        if (s_pPitchingCandidateMethods == NULL)
```
janvorli: The lazy initialization of all the maps and locks make the code harder to read, obscuring the real functionality. It doesn't seem to save much. It seems it would be better to just allocate all of them in a single initialization function where you would know that the pitching is active. Did you have any particular reason for using the lazy allocations?

src/vm/codepitchingmanager.cpp
```diff
@@ -0,0 +1,526 @@
...
+                        {
+                            pMD->PitchNativeCode();
+                        }
+                    }
+                    for (PtrHashMap::PtrIterator i = s_pExecutedMethods->begin(); !i.end();)
+                    {
+                        MethodDesc *pMD = (MethodDesc *) i.GetValue();
+                        UPTR key = (UPTR)GetFullHash(pMD);
+                        ++i;
+                        s_pExecutedMethods->DeleteValue(key, (LPVOID)pMD);
```
janvorli: Is it really ok to delete values from the hash map during the iteration?

src/vm/codepitchingmanager.cpp
```diff
@@ -0,0 +1,526 @@
...
+            EX_END_CATCH(SwallowAllExceptions);
+        }
+    }
+}
+
+EXTERN_C void SavePitchingCandidate(MethodDesc* pMD, ULONG sizeOfCode)
+{
+    if (pMD->IsPitchable())
+    {
+        if (sizeOfCode > 0 && CLRConfig::GetConfigValue(CLRConfig::INTERNAL_JitPitchMethodSizeThreshold) < sizeOfCode)
```
janvorli: The first part of the condition is not necessary, the second would never pass for sizeOfCode being zero.

sergign60:
@janvorli 
CC: @noahfalk 
Thank you a lot for your for your comments. They are very useful. I think that I've satisfied all of them

src/vm/codepitchingmanager.cpp
```diff
@@ -0,0 +1,508 @@
...
+        pitchedBytes = (ULONGLONG)s_pPitchingCandidateSizes->LookupValue((UPTR)this, nullptr);
+        _ASSERTE(pitchedBytes != (ULONGLONG)INVALIDENTRY);
+        if (pitchedBytes == (ULONGLONG)INVALIDENTRY)
+            pitchedBytes = 0;
+        s_jitPitchedBytes += (pitchedBytes >> 1);
+    }
+    {
+        SimpleWriteLockHolder swlh(s_pPitchingCandidateMethodsLock);
+        s_pPitchingCandidateMethods->DeleteValue(key, (LPVOID)this);
+        s_pPitchingCandidateMethods->Compact();
```
janvorli: @sergign60 Last detail I wonder about - do we need to run the Compact after each DeleteValue? Does the s_pExecutedMethods->Clear() call happen that rarely that we care about the fact that the hash map is not compact?

sergign60: @janvorli  I think it's not needed. I'll remove it.

sergign60: @janvorli  I need some time for testing

sergign60:

sergign60:
@noahfalk @janvorli 
The latest CoreCLR testsuite result:

```
=======================
     Test Results
=======================
# CoreCLR Bin Dir  : ./coreclr/bin/Product/Linux.x64.Debug
# Tests Discovered : 7104
# Passed           : 6394
# Failed           : 0
# Skipped          : 710
=======================
$ env | grep COMPl
COMPlus_JitPitchMemThreshold=1
COMPlus_JitPitchMethodSizeThreshold=1
COMPlus_JitPitchEnabled=1
COMPlus_JitPitchPrintStat=1
```

sergign60:

sergign60:

sergign60:

sergign60:

sergign60:

sergign60:
@dotnet-bot test Windows_NT arm Cross Debug Build  please

sergign60:
@dotnet-bot  test FreeBSD x64 Checked Build please

sergign60:

sergign60:

sergign60:

sergign60:

sergign60:
@dotnet test OSX10.12 x64 Checked Build and Test please

sergign60:

sergign60:

sergign60:

sergign60:

noahfalk:
@sergign60 - sorry I had forgotten about this for a while, are you ready to have this merged now? I think earlier there were some CI issues but it appears you got those addressed and both JanV and I are happy with it.

I have another PR that is going to have merge conflicts so I'd like to get yours in ASAP and then I can merge on top of it. Otherwise if yours doesn't go in now you will need to do some merge work after my PR commits.

sergign60:
@noahfalk  yes, I'm ready to merge it

sergign60:
@noahfalk many thanks for your help!

noahfalk:
@sergign60 - very welcome and thanks for all your work : )

PR #10497, opened 2017-03-27
Incorrect parameter order found by PVS-Studio

DieselMachine84:
I used PVS-Studio static analyzer to check this project. I would like to suggest a variant of the way to fix the error, detected with the help of V3066 diagnostic https://www.viva64.com/en/w/V3066/

### DIFF
--- a/src/mscorlib/src/System/Reflection/RuntimeModule.cs
+++ b/src/mscorlib/src/System/Reflection/RuntimeModule.cs
@@ -464,7 +464,7 @@ namespace System.Reflection
 
             RuntimeType retType = null;
             Object keepAlive = null;
-            GetType(GetNativeHandle(), className, throwOnError, ignoreCase, JitHelpers.GetObjectHandleOnStack(ref retType), JitHelpers.GetObjectHandleOnStack(ref keepAlive));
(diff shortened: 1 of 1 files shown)


src/mscorlib/src/System/Reflection/RuntimeModule.cs
```diff
@@ -464,7 +464,7 @@ public override Type GetType(String className, bool throwOnError, bool ignoreCas
 
             RuntimeType retType = null;
             Object keepAlive = null;
-            GetType(GetNativeHandle(), className, throwOnError, ignoreCase, JitHelpers.GetObjectHandleOnStack(ref retType), JitHelpers.GetObjectHandleOnStack(ref keepAlive));
+            GetType(GetNativeHandle(), className, ignoreCase, throwOnError, JitHelpers.GetObjectHandleOnStack(ref retType), JitHelpers.GetObjectHandleOnStack(ref keepAlive));
```
jkotas: This is not the right fix. The right fix is to swap the argument names in the GetType declaration at the top of the file. 

There is no actual functionality bug here. It is just a cosmetic issue that the declaration has the argument names swapped.

danmosemsft:
@DieselMachine84 do you have time to fix this so we can merge the pR? thanks for flagging it.

DieselMachine84:
Yes, I'll swap the parameters in declaration

jkotas:
Thanks!

PR #10499, opened 2017-03-27
Move Pointer.cs to the shared partition.

AtsushiKan:
(The two internal helpers are not yet used
on CoreRT but that's only because CoreRT
doesn't yet implement Pointer support
for Reflection invoke. Exposing them
unconditionally now to save us the trouble
of removing the #ifdef's later.)


src/vm/mscorlib.h
```diff
@@ -752,10 +752,6 @@ DEFINE_CLASS(PARAMETER_MODIFIER,    Reflection,             ParameterModifier)
 
 DEFINE_CLASS(POINTER,               Reflection,             Pointer)
 
-DEFINE_CLASS_U(Reflection, Pointer, ReflectionPointer)
```
jkotas: Why are you removing this? It is here to verify the layout is in sync between managed and unmanaged sides.

src/mscorlib/shared/System/Reflection/Pointer.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            Debug.Assert(ptrType.IsRuntimeImplemented()); // CoreCLR: For CoreRT's sake, _ptrType has to be declared as "Type", but in fact, it is always a RuntimeType. Code on CoreCLR expects this.
+            _ptr = ptr;
+            _ptrType = ptrType;
+        }
+
+        private Pointer(SerializationInfo info, StreamingContext context)
+        {
+            _ptr = ((IntPtr)(info.GetValue("_ptr", typeof(IntPtr)))).ToPointer();
+            _ptrType = (Type)info.GetValue("_ptrType", typeof(Type));
```
jkotas: Throw if ptrType is not RuntimeImplemented?

src/vm/mscorlib.h
AtsushiKan: Ok, thought I was just removing dead code - didn't realize what the "_U" signified.

src/mscorlib/shared/System/Reflection/Pointer.cs
AtsushiKan: Fixed.

src/mscorlib/shared/System/Reflection/Pointer.cs
```diff
@@ -0,0 +1,61 @@
...
+            if (!_ptrType.IsRuntimeImplemented())
+                throw new SerializationException(SR.Arg_MustBeType);
+        }
+
+        public static object Box(void* ptr, Type type)
+        {
+            if (type == null)
+                throw new ArgumentNullException(nameof(type));
+            if (!type.IsPointer)
```
jkotas: Nit: Can be `(!type.IsPointer || type.IsRuntimeImplemented())` to make the code a bit smaller.

AtsushiKan: Arguably, CoreCLR should have used SR.Arg_MustBeType for the !IsRuntimeImplemented() case. The current message is confusing.

jkotas: Changing the error message to be less confusing is fine.

PR #10502, opened 2017-03-27
Don't ignore exceptions thrown from handlers of some events

kouvel:
Fixes dotnet/corefx#14747:
- Events include: AssemblyLoadContext.Unloading, AppDomain.ProcessExit
- Made the same change for AppDomain.DomainUnload for consistency, but it's not raised

### DIFF
--- a/src/vm/appdomain.cpp
+++ b/src/vm/appdomain.cpp
@@ -7475,7 +7475,7 @@ void AppDomain::ProcessUnloadDomainEventOnFinalizeThread()
 {
     CONTRACTL
     {
-        NOTHROW;
+        THROWS;
         GC_TRIGGERS;
         MODE_COOPERATIVE;
     }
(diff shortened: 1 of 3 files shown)


kouvel:
@dotnet-bot test this please

kouvel:
I was not planning on adding a test for this, as it would require crashing a process

stephentoub:
> I was not planning on adding a test for this, as it would require crashing a process

Ok, though if you decide to, we have the remote invoke support in the CoreFx repo for that, and we do crash processes in some cases.

kouvel:
Oh ok, I'll take a look at the crashing ones and see how it is done in a reliable manner

kouvel:
I don't think the remote invoke infra would work for this. That would catch exceptions thrown by the thread making the call into test code, but the relevant events here are raised on the finalizer thread and that will cause the program to crash in an uncatchable manner.

kouvel:
Also for existing tests when it rethrows the exception to crash the child process, I'm seeing my postmortem debugger kick in. We would probably need some COMPlus variable in a release build of CoreCLR to disable crashing the process and instead to just terminate it without raising an exception but with an exit code that would indicate a crash.

stephentoub:
> that will cause the program to crash in an uncatchable manner.

That's ok. The program will end with an exit code other than for success, and the main process will see that and judge the outcome accordingly.

Issue #10504, opened 2017-03-27
Default Interface Method support 


yizhang82:
[C# Language Spec Link](https://github.com/dotnet/csharplang/blob/master/proposals/default-interface-methods.md)
Runtime Spec Link (TBD)

List of work items: 

- [x] Fix up ILDAsm to support default interface methods (non-abstract virtual methods) on interfaces
- [x] Fix up ILDAsm to allow non-virtual instance methods on interface
- [x] Allow type loader to ignore not implemented interface methods on classes
- [x] Allow non-virtual interface instance methods in type loader
- [x] Support calling protected interface non-virtual instance method from derived class (TBD)
- [x] Hook up VSD interface dispatch
- [x] Throw in the case of diamond inheritance case
- [x] Support Shared generics and variance
- [x] Support looking up method impls
- [ ] Static constructors with initializers only (no static constructor by itself - needs more investigation)
- [ ] Support constrained interface call on value types
- [x] Handle interface precodes properly
- [x] Correct interface lookup algorithm (and validation) once design is finalized
- [x] Proper typeloader/ILDAsm error checking/validation for interface methods once design is finalized (private/protected/internal, virtual vs non-virtual)
- [ ] Reflection to tell the difference between default interface method and a regular
- [ ] NGEN support
- [x] R2R support
- [ ] Interaction with COM interop

gkhanna79:
@yizhang82 Is this for .NET Core 2.0?

yizhang82:
No. It's not part of .NET Core 2.0 work. I don't think we have a shipping date yet. Right now we are in early design/prototype phase.

yizhang82:
We also need engineering work to support producing packages for Roslyn CI. Tracked here: https://github.com/dotnet/core-eng/issues/833

mattwarren:
> Tracked here: [dotnet/core-eng#833](https://github.com/dotnet/core-eng/issues/833)

Just one thing, that 'core-eng' repo doesn't appear to be publicly accessible. 

I'm not saying it should be, but I just wanted to let you know that anyone who isn't a M/S employee can't see that link you posted.

MichalStrehovsky:
Broken out the remaining tasks into individual issues and now tracking within the [Default interfaces project](https://github.com/dotnet/coreclr/projects/6).

PR #10505, opened 2017-03-27
Default Interface Method Prototype

yizhang82:
CoreCLR Issue - https://github.com/dotnet/coreclr/issues/10504
C# Language Spec (WIP) - https://github.com/dotnet/csharplang/blob/master/proposals/default-interface-methods.md

As suggested by @jkotas, I'm moving my work to dev/defaultintf branch in the main CoreCLR repo. I'm going to keep submitting periodically submit new PRs to dev/defaultintf branch going forward - this is going to make collaboration between various runtime and language team a bit easier. 

What this PR has:
* Allow non-abstract non-public interface methods in ILDAsm and change our IL.targets to use ILDAsm we built for the time being, until we have a good set of toolsets that has the new ILDAsm. 
* Allow type loader to ignore not implemented interfaces (we may need to add more validation later)
* Hooked up VSD interface dispatch to look for default interface method in the absence of proper implementation from the class itself. 
* A naive implementation of default interface method look up based on InterfaceMap, from derived class to base class - just enough for prototyping on both CLR and C# sides. Once we finalize the design, it'll be replaced by the real thing. 
* Support for shared generics by properly instantiate MethodDesc which inserts instantiating stubs to pass the secret MethodTable argument for the particular generic instantiation. One particular interesting aspect of this is that we need to lie to the JIT about those interface methods at call site requires secret argument so that it emits the correct code at callsite. 
* Support for explicit override by looking up methodimpls (implicit method override is not supported)
* Support for interface variance
* Temporary workaround for interface precode. Today we always assume interface has precode
* Added tests
  * Note that the C# code is simply there to make writing IL a bit easier - the IL are disassembled from C# generated code and then modified by hand. Once things got more stable we can delete them, or see if it makes sense to switch over to a new C# compiler for positive cases that we care about.  

TODO - See https://github.com/dotnet/coreclr/issues/10504

@jkotas @davidwrighton PTAL

//cc @fadimounir


jkotas:
cc @migueldeicaza @kumpera

src/ilasm/assembler.cpp
```diff
@@ -698,14 +698,19 @@ void Assembler::StartMethod(__in __nullterminated char* name, BinStr* sig, CorMe
...
                     {
-                        if(OnErrGo) report->error("Non-virtual, non-abstract instance method in interface\n");
+                        // @DIM_TODO - What about non-virtuals?
+                        if(OnErrGo) report->error("Non-virtual instance method in interface\n");
                         else
                         {
-                            report->warn("Non-virtual, non-abstract instance method in interface, set to such\n");
-                            flags = (CorMethodAttr)(flags |mdVirtual | mdAbstract);
+                            report->warn("Non-virtual instance method in interface, set to such\n");
+                            flags = (CorMethodAttr)(flags | mdVirtual);
```
davidwrighton: I think that in ILASM we should permit non-virtuals. This would be to allow either the feature to be implemented in the runtime, or for tests that the feature is blocked by the typeloader to be put in place. In either case, we shouldn't mystically mark the method as virtual. (However, we should consider spewing a warning)

src/vm/method.cpp
```diff
@@ -1869,7 +1873,8 @@ BOOL MethodDesc::AcquiresInstMethodTableFromThis() {
         IsSharedByGenericInstantiations()  &&
         !HasMethodInstantiation() &&
         !IsStatic() &&
-        !GetMethodTable()->IsValueType();
+        !GetMethodTable()->IsValueType() &&
+        !IsDefaultInterfaceMethod();
```
davidwrighton: >IsDefaultInterfaceMethod [](start = 9, length = 24)

Could we rename this predicate to be IsInterfaceMethodWithImplementation()? We may support non-virtual interface methods, and calling them "default" interface methods would be confusing.

src/vm/methodtablebuilder.cpp
```diff
@@ -2881,17 +2886,6 @@ MethodTableBuilder::EnumerateClassMethods()
...
-        {
-            if (IsMdRTSpecialName(dwMemberAttrs) || IsMdVirtual(dwMemberAttrs))
-            {
-                CONSISTENCY_CHECK(CheckPointer(strMethodName));
-                if (strcmp(strMethodName, COR_CCTOR_METHOD_NAME))
-                {
-                    BuildMethodTableThrowException(BFA_NONAB_NONCCTOR_METHOD_ON_INT);
-                }
-            }
-        }
```
davidwrighton: Why are you removing the check against cctor's on interfaces? It seems to me that since we've not talked positively about adding static fields to interfaces, we should keep cctors disabled.

src/vm/methodtablebuilder.cpp
```diff
@@ -7667,6 +7704,23 @@ MethodTableBuilder::PlaceInterfaceMethods()
                 PlaceMethodFromParentEquivalentInterfaceIntoInterfaceSlot(itfSlotIt, pCurItfEntry, &rgInterfaceDispatchMapTypeIDs, dwCurInterface);
             }
 #endif
+
+            /*
```
davidwrighton: Please don't check in commented out code without a comment as to what's going on here.

src/vm/methodtablebuilder.cpp
```diff
@@ -11024,11 +11078,13 @@ void MethodTableBuilder::VerifyVirtualMethodsImplemented(MethodTable::MethodData
             MethodTable::MethodIterator it(hData);
             for (; it.IsValid() && it.IsVirtual(); it.Next())
             {
-                if (it.GetTarget().IsNull())
-                {
-                    MethodDesc *pMD = it.GetDeclMethodDesc();
-                    BuildMethodTableThrowException(IDS_CLASSLOAD_NOTIMPLEMENTED, pMD->GetNameOnNonArrayClass());
-                }
+                // @DESIGN - What is the right level of check if the interface itself does not have default implementation
```
davidwrighton: Please add a DIM_TODO to the comment so we can find these bits of commented out logic.

davidwrighton:
:shipit:

src/ilasm/assembler.cpp
yizhang82: Yep. I haven't got to supporting instance methods yet. I'm looking into it now and have removed these checks from ILAsm.

src/vm/methodtablebuilder.cpp
yizhang82: If I read the code correctly, we already allow static cctors in interfaces (since we allow static methods already), and the check is for non-cctor that are virtual or special name. I can keep the special name check for now.

src/vm/method.cpp
yizhang82: I'm a bit unsure about this one. Default interface method feature do have wider implications (such as non-virtual instance methods, etc). But the specific default interface method scenario in the context of runtime should be referring to non-abstract virtual instance method. You can also argue that IsInterfaceMethodWithImplementation can also refer to instance methods in interface, which is not what we want here.
I'll add a bit of comment in the method header to explain / clarify that it is non-abstract virtual instance method. I can change it if you feel strongly, though.

src/vm/methodtablebuilder.cpp
yizhang82: I've removed the code for now. It might be useful going forward when we go for a new implementation but I can easily add it back.

src/vm/methodtablebuilder.cpp
yizhang82: Thanks. Fixed/Replaced with DIM_TODO

yizhang82:
Thanks @davidwrighton for review!

PR #10506, opened 2017-03-27
Create an IGCHandleTable interface

adityamandaleeka:
This PR introduces the concept of IGCHandleTable, which the VM will use when dealing with handles and handle table-related things. Today, the VM code is directly including heads like objecthandle.h, handletable.h, and handletablepriv.h, which should only be used on the GC side of things. 

In order to keep this change simple, I've only included `Initialize` and `Shutdown` functions on the interface. This way, we can focus on the mechanics of where things are and how the VM gets an instance of GCHandleTable. Once this is in, I will follow up with changes that add all the other functionality (like handle creation, destruction, storing objects in handles, etc.) that the VM uses. In the end, the VM code will no longer be using the handle-related headers and only go through this interface for all its needs.

@Maoni0 @swgillespie @sergiy-k

### DIFF
--- a/src/gc/gccommon.cpp
+++ b/src/gc/gccommon.cpp
@@ -21,6 +21,7 @@ SVAL_IMPL_INIT(uint32_t,IGCHeap,gcHeapType,IGCHeap::GC_HEAP_INVALID);
 SVAL_IMPL_INIT(uint32_t,IGCHeap,maxGeneration,2);
 
 IGCHeapInternal* g_theGCHeap;
+IGCHandleTable* g_theGCHandleTable;
 
 #ifdef FEATURE_STANDALONE_GC
 IGCToCLR* g_theGCToCLR;
@@ -145,7 +146,7 @@ namespace SVR
     extern void PopulateDacVars(GcDacVars* dacVars);
 }
 
-bool InitializeGarbageCollector(IGCToCLR* clrToGC, IGCHeap** gcHeap, GcDacVars* gcDacVars)
+bool InitializeGarbageCollector(IGCToCLR* clrToGC, IGCHeap** gcHeap, IGCHandleTable** gcHandleTable, GcDacVars* gcDacVars)
 {
     LIMITED_METHOD_CONTRACT;
 
@@ -153,6 +154,14 @@ bool InitializeGarbageCollector(IGCToCLR* clrToGC, IGCHeap** gcHeap, GcDacVars* 
 
     assert(gcDacVars != nullptr);
     assert(gcHeap != nullptr);
+    assert(gcHandleTable != nullptr);
+
+    IGCHandleTable* handleTable = CreateGCHandleTable();
+    if (handleTable == nullptr)
+    {
+        return false;
+    }
+
 #ifdef FEATURE_SVR_GC
     assert(IGCHeap::gcHeapType != IGCHeap::GC_HEAP_INVALID);
 
@@ -169,7 +178,6 @@ bool InitializeGarbageCollector(IGCToCLR* clrToGC, IGCHeap** gcHeap, GcDacVars* 
 #else
     heap = WKS::CreateGCHeap();
     WKS::PopulateDacVars(gcDacVars);
-
 #endif
 
     if (heap == nullptr)
@@ -187,6 +195,7 @@ bool InitializeGarbageCollector(IGCToCLR* clrToGC, IGCHeap** gcHeap, GcDacVars* 
     assert(clrToGC == nullptr);
 #endif
 
+    *gcHandleTable = handleTable;
     *gcHeap = heap;
     return true;
 }
--- a/src/gc/gcinterface.h
+++ b/src/gc/gcinterface.h
@@ -171,11 +171,12 @@ struct segment_info
 
 class Object;
 class IGCHeap;
+class IGCHandleTable;
 
 // Initializes the garbage collector. Should only be called
 // once, during EE startup. Returns true if the initialization
 // was successful, false otherwise.
-bool InitializeGarbageCollector(IGCToCLR* clrToGC, IGCHeap **gcHeap, GcDacVars* gcDacVars);
+bool InitializeGarbageCollector(IGCToCLR* clrToGC, IGCHeap** gcHeap, IGCHandleTable** gcHandleTable, GcDacVars* gcDacVars);
 
 // The runtime needs to know whether we're using workstation or server GC 
 // long before the GCHeap is created. This function sets the type of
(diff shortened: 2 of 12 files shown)


Maoni0:
what would be the relationship between gchandletable.h and handletable.h?

adityamandaleeka:
@Maoni0 With the current design, gchandletable.h will just contain things related to the GCHandleTable class while handletable.h will continue to contain handletable.inl and the Hnd* functions it has today.

On a related note, I can see handletable.h and objecthandle.h being consolidated in the future, since after my changes are complete, there won't be much left in objecthandle.h

src/gc/gcinterface.h
```diff
@@ -384,6 +385,13 @@ typedef void (* record_surv_fn)(uint8_t* begin, uint8_t* end, ptrdiff_t reloc, s
 typedef void (* fq_walk_fn)(BOOL, void*);
 typedef void (* fq_scan_fn)(Object** ppObject, ScanContext *pSC, uint32_t dwFlags);
 typedef void (* handle_scan_fn)(Object** pRef, Object* pSec, uint32_t flags, ScanContext* context, BOOL isDependent);
+class IGCHandleTable {
```
adityamandaleeka: Sounds good.

adityamandaleeka:
@swgillespie Added a commit (that I'll squash) with those changes.

src/gc/gccommon.cpp
```diff
@@ -169,7 +174,6 @@ bool InitializeGarbageCollector(IGCToCLR* clrToGC, IGCHeap** gcHeap, GcDacVars*
 #else
     heap = WKS::CreateGCHeap();
     WKS::PopulateDacVars(gcDacVars);
-
 #endif
 
     if (heap == nullptr)
```
adityamandaleeka: Done

Maoni0:
@adityamandaleeka it seems a bit odd that there are 2 headers - one called gchandletable.h while the other one handletable.h which is also about gc handletable...it would be good to indicate if one is an interface or impl...

src/gc/sample/gchandletable_forsample.cpp
```diff
@@ -0,0 +1,12 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+// 
+
+// This file just includes the regular gchandletable.cpp file, but adds some
+// headers that are necessary for it to build in the GCSample project.
+
+#include "common.h"
+#include "gcenv.h"
+
```
jkotas: Why do we need this? Why can't this be done same way as the rest to keep the same simple?

jkotas: You should be able to `#include "gcenv.h"` in gchandletable.cpp, so that you do not need this separate files. gcenv.h is included in all other GC files.

adityamandaleeka: Thanks. I originally was not including gcenv.h because it was in the VM directory, but it looks like we will eventually have a similar file in the GC side too. I'll verify that including that works and update the PR. I'm happy to remove this ugly hack 👍

adityamandaleeka:
@dotnet-bot 

adityamandaleeka:
@Maoni0 I completely agree that the header situation is confusing. As I said above, I think we can combine handletable.h and objecthandle.h in the future to reduce some of the confusion (into something like handletable.gc.h to indicate it's only for GC-side use).

As for gchandletable.h vs handletable.h, I can rename gchandletable.h to gchandletableimpl.h (similar to gcimpl.h) to make it less confusing.

jkotas:
> handletable.gc.h

This name does not feel right - we do not have anything named `*.gc.h`. I do not see any problem with handletable.h.

adityamandaleeka:

This was the failure:
```
14:04:52 FAILED   - JIT/Performance/CodeQuality/Roslyn/CscBench/CscBench.sh
14:04:52                BEGIN EXECUTION
14:04:52                /mnt/j/workspace/dotnet_coreclr/master/checked_ubuntu_tst_prtest/bin/tests/Windows_NT.x64.Checked/Tests/coreoverlay/corerun CscBench.exe
14:04:52                
14:04:52                Unhandled Exception: System.MissingMethodException: Method not found: 'Void System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw(System.Exception)'.
14:04:52                   at System.Threading.Tasks.Parallel.ThrowSingleCancellationExceptionOrOtherException(ICollection exceptions, CancellationToken cancelToken, Exception otherException)
14:04:52                   at System.Threading.Tasks.Parallel.ForWorker[TLocal](Int32 fromInclusive, Int32 toExclusive, ParallelOptions parallelOptions, Action`1 body, Action`2 bodyWithState, Func`4 bodyWithLocal, Func`1 localInit, Action`1 localFinally) in /mnt/j/workspace/dotnet_corefx/master/ubuntu14.04_release/src/System.Threading.Tasks.Parallel/src/System/Threading/Tasks/Parallel.cs:line 1222
14:04:52                   at System.Threading.Tasks.Parallel.For(Int32 fromInclusive, Int32 toExclusive, ParallelOptions parallelOptions, Action`1 body) in /mnt/j/workspace/dotnet_corefx/master/ubuntu14.04_release/src/System.Threading.Tasks.Parallel/src/System/Threading/Tasks/Parallel.cs:line 474
14:04:52                   at Microsoft.CodeAnalysis.CSharp.CSharpCompilation.GetDiagnostics(CompilationStage stage, Boolean includeEarlierStages, CancellationToken cancellationToken)
14:04:52                   at Microsoft.CodeAnalysis.CSharp.CSharpCompilation.CompileImpl(CommonPEModuleBuilder moduleBuilder, Stream win32Resources, Stream xmlDocStream, Boolean emittingPdb, DiagnosticBag diagnostics, Predicate`1 filterOpt, CancellationToken cancellationToken)
14:04:52                   at Microsoft.CodeAnalysis.Compilation.Compile(CommonPEModuleBuilder moduleBuilder, Stream win32Resources, Stream xmlDocStream, Boolean emittingPdb, DiagnosticBag diagnostics, Predicate`1 filterOpt, CancellationToken cancellationToken)
14:04:52                   at Microsoft.CodeAnalysis.Compilation.Emit(EmitStreamProvider peStreamProvider, EmitStreamProvider pdbStreamProvider, EmitStreamProvider xmlDocumentationStreamProvider, EmitStreamProvider win32ResourcesStreamProvider, IEnumerable`1 manifestResources, EmitOptions options, IMethodSymbol debugEntryPoint, CompilationTestData testData, Func`1 getHostDiagnostics, CancellationToken cancellationToken)
14:04:52                   at Microsoft.CodeAnalysis.Compilation.Emit(Stream peStream, Stream pdbStream, Stream xmlDocumentationStream, Stream win32Resources, IEnumerable`1 manifestResources, EmitOptions options, IMethodSymbol debugEntryPoint, CompilationTestData testData, Func`1 getHostDiagnostics, CancellationToken cancellationToken)
14:04:52                   at Microsoft.CodeAnalysis.Compilation.Emit(Stream peStream, Stream pdbStream, Stream xmlDocumentationStream, Stream win32Resources, IEnumerable`1 manifestResources, EmitOptions options, IMethodSymbol debugEntryPoint, CancellationToken cancellationToken)
14:04:52                   at CscBench.CompileBench()
14:04:52                   at CscBench.Bench()
14:04:52                   at CscBench.Main()
14:04:52                ./CscBench.sh: line 139: 12611 Aborted                 (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" CscBench.exe $CLRTestExecutionArguments
14:04:52                Expected: 100
14:04:52                Actual: 134
14:04:52                END EXECUTION - FAILED
```

adityamandaleeka:
@jkotas @Maoni0 @swgillespie PTAL. I've removed the hack for the sample, renamed the new header file, and fixed Windows x86 (had to move GCHeap initialization to after write barrier setup).

I'll squash those changes prior to merge.

PR #10510, opened 2017-03-27
Implement TLS field handles for RyuJIT/x86

BruceForstall:
This isn't used by .NET Core, but helps us run desktop testing.


BruceForstall:
@briansull @dotnet/jit-contrib PTAL

briansull:
LGTM

PR #10503, opened 2017-03-27
small fgOptimizeDelegateConstructor code refactoring

sandreenko:
Delete unnecessary condition.
Use the default methods to work with lists.
ArgList construction reordering.

PR #10512, opened 2017-03-27
Ctor Jit-EE interface change

sandreenko:
CoreRT needs to be able to pass not only constant lookup there.
Please review only the second commit, the first will be merged in the separate PR #10503.

I want to do this change before the actual fix because it is a breaking change. I will not be able to verify spmi diffs for the fix if the interface change and the fix are in the same PR.


AndyAyersMS:
LGTM.

MichalStrehovsky:
You don't have to update the JitInterface GUID?

PR #10513, opened 2017-03-28
Crossgen support for ExcludeHotMethodCode and ExcludeColdMethodCode

briansull:
Added two new profile data flags: ExcludeHotMethodCode and ExcludeColdMethodCode

Refactored zapimage CompileProfileData into four new methods:
CompileHotRegion, CompileColdRegion, PlaceMethodIL and ProfileDisableInlining
Preserve the CompileStatus and methodProfilingDataFlags used when we compile methods during CompileHotRegion()
Then during CompileColdRegion retrieve these value and exclude the method from the AOT native image if the status is COMPILE_EXCLUDED.

### DIFF
--- a/src/inc/corbbtprof.h
+++ b/src/inc/corbbtprof.h
@@ -140,7 +140,7 @@ enum TypeProfilingDataFlags
     ReadMethodTable               = 0,  // 0x00001
     ReadEEClass                   = 1,  // 0x00002
     WriteEEClass                  = 2,  // 0x00004
-//  ReadStoredEnumData            = 3,  // 0x00008
+//  ReadStoredEnumData            = 3,  // 0x00008  // obsolete
     ReadFieldDescs                = 4,  // 0x00010
     ReadCCtorInfo                 = 5,  // 0x00020
     ReadClassHashTable            = 6,  // 0x00040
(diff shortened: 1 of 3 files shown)


briansull:
@dpodder PTAL
@AndyAyersMS PTAL

src/inc/corbbtprof.h
```diff
@@ -1,6 +1,7 @@
-// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
jkotas: Keep the license

briansull:
Sorry that was an inadvertent change.

src/inc/corbbtprof.h
briansull: Sorry that was an inadvertent change.

briansull:
@dpodder PTAL
@AndyAyersMS PTAL

src/zap/zapimage.cpp
```diff
@@ -1588,33 +1588,75 @@ ZapImage::CompileStatus ZapImage::CompileProfileDataWorker(mdToken token, unsign
...
+            unsigned methodProfilingDataFlags = pTokenInfo->flags;
+
+            // Hot methods can be marked to be excluded from the AOT native image.
+            // We also need to disable inlining of such methods.
+            //
+            if ((methodProfilingDataFlags & (1 << ExcludeHotMethodCode)) != 0)
+            {
+                // Disable the inlining of this method
+                //
+                // @ToDo: Figure out how to disable inlining for this method.
```
AndyAyersMS: Would prefer that this "method is excluded hot" information come through to the jit as a new method attribute, rather than baking in a policy here and (for instance) setting the noinline bit on the method.

src/zap/zapimage.cpp
```diff
@@ -1638,10 +1680,14 @@ void ZapImage::CompileProfileData()
...
                     }
                 }
             }
+
+            // Update the 'flags' saved in the profileDataHashTable hash table.
+            //
+            hashBBUpdateFlagsAndCompileResult(token, methodProfilingDataFlags, compileResult);
         }
         // record the start of hot Generics methods.
         m_iGenericsMethod = m_MethodCompilationOrder.GetCount();
```
AndyAyersMS: Not really part of your change, but am just curious if I understand the code here -- it looks like there are never any generics emitted into the hot section?

src/zap/zapimage.cpp
```diff
@@ -2022,20 +2088,64 @@ ZapImage::CompileStatus ZapImage::TryCompileMethodWorker(CORINFO_METHOD_HANDLE h
...
+
+        // Cold methods can be marked to be excluded from the AOT native image.
+        // We can reduced the size of the AOT native image by selectively
+        // excluding the code for some of the cold methods.
+        //
+        if ((methodProfilingDataFlags & (1 << ExcludeColdMethodCode)) != 0)
+        {
+            // returning COMPILE_EXCLUDED excludes this method from the AOT native image
+            return COMPILE_EXCLUDED;
+        }
```
AndyAyersMS: If you returned different status for hot and cold exclusions then the profileDataHashTable could potentially be mined after the fact to display the breakdown of method dispositions.

src/zap/zapimage.cpp
briansull: Yes, I agree. It also should only apply during AOT time compilation.
If JIT time request to inlining happens we should allow that.

I'm also considering making the **disable-inlining-hot** its own flag.
That would mean that a particular method could be Exclude-Hot without being Disable-Inlining (or vice a versa).  The initial implementation could tie them together but we could experiment easier.

src/zap/zapimage.cpp
briansull: I also took a look at that and it appear to me to be wrong and a left over from a time where we laid out all non-generic codes first followed by the generic code.  It feeds into a structure that records information about the structure of Ngen image. But I don't believe that the information has a consumer anymore, so it can be wrong without any impact. 
In any case we now lay out the code for generics in place along with the non-generic code, so that value isn't of any use to us anymore and could be removed.

src/zap/zapimage.cpp
briansull: OK I will make that change.

Issue #10141, opened 2017-03-13
Wrong comment about GC pinning bit in object.h


303248153:
I found this comment in https://github.com/dotnet/coreclr/blob/master/src/vm/object.h
```
// lose GC marking bit and the pinning bit
// A method table pointer should always be aligned.  During GC we set the least 
// significant bit for marked objects and we set the second to least significant
// bit for pinned objects.  So if we want the actual MT pointer during a GC
// we must zero out the lowest 2 bits.
```
I think it's wrong, GC pinning bit actually stored in object header

``` c++
void SetPinned()
{
    assert (!(gc_heap::settings.concurrent));
    GetHeader()->SetGCBit();
}

void SetGCBit() { m_uSyncBlockValue |= BIT_SBLK_GC_RESERVE; }
```
And I verified in lldb
![1](https://cloud.githubusercontent.com/assets/16222094/23843132/ab715616-07f4-11e7-84c5-8e618f14318a.jpg)

And no code is touching `m_pMethTab & 2`, am I right and should this comment be fixed?

swgillespie:
That would seem to be the case, yeah. I suppose that this was historically the case, though we do use the GC bit on the sync block now as you observed. Feel free to submit a PR with a fixed comment, we'd be happy to take it!

303248153:
Thanks for response.
Also the code in `GetGCSafeMethodTable` use a hard code `3` for clear the MARKED_BIT and the pinned bit not using now, should it be fixed together?
```
PTR_MethodTable GetGCSafeMethodTable() const
    {
        LIMITED_METHOD_CONTRACT;
        SUPPORTS_DAC;

        // lose GC marking bit and the pinning bit
        // A method table pointer should always be aligned.  During GC we set the least 
        // significant bit for marked objects and we set the second to least significant
        // bit for pinned objects.  So if we want the actual MT pointer during a GC
        // we must zero out the lowest 2 bits.
        return dac_cast<PTR_MethodTable>((dac_cast<TADDR>(m_pMethTab)) & ~((UINT_PTR)3));
}
```
I think it can be replaced with
```
return dac_cast<PTR_MethodTable>((dac_cast<TADDR>(m_pMethTab)) & (~MARKED_BIT));
```

swgillespie:
Sure, feel free to do that too!

PR #10515, opened 2017-03-28
Fix wrong comment about pinning bit and change GetGCSafeMethodTable

303248153:
Refer to issue #10141
Since pinning bit is moved to sync block,
I change the comment in GetGCSafeMethodTable, and replace the old hard code 3 with MARKED_BIT.

It's a dangerous change, I didn't see any code in gc use the last but one bit, if any of you known please say it.
I also run tests on windows after this change, only one test failed:

```
BEGIN EXECUTION
 "C:\coreclr-fork\coreclr\bin\tests\Windows_NT.x64.Debug\Tests\C
ore_Root\corerun.exe" SizeConstTest.exe

Unhandled Exception: System.Runtime.InteropServices.COMException: The data area passed to a system call is too small (Exception from HRESULT: 0x8007007A)
   at System.StubHelpers.ValueClassMarshaler.ConvertToNative(IntPtr dst, IntPtr
src, IntPtr pMT, CleanupWorkList& pCleanupWorkList)
   at Test.TakeByValTStr(S_CHARArray_ByValTStr s, Int32 size)
   at Test.SizeConstByValTStr()
   at Test.Main(String[] args)
Expected: 100
Actual: -532462766
END EXECUTION - FAILED
FAILED
```

This test also failed before I perform this modification, so I think it's irrelevant.

### DIFF
--- a/src/vm/object.h
+++ b/src/vm/object.h
@@ -638,7 +638,6 @@ class Object
         // A method table pointer should always be aligned.
         // During GC we set the least significant bit for marked objects.
         // So if we want the actual MT pointer during a GC we must zero out the lowest bit.
-        // In the past we had put pinning bit here but now it's in sync block.
         return dac_cast<PTR_MethodTable>((dac_cast<TADDR>(m_pMethTab)) & (~MARKED_BIT));
     }
 
--- a/src/vm/object.h
+++ b/src/vm/object.h
@@ -634,12 +634,12 @@ class Object
         LIMITED_METHOD_CONTRACT;
         SUPPORTS_DAC;
 
-        // lose GC marking bit and the pinning bit
-        // A method table pointer should always be aligned.  During GC we set the least 
-        // significant bit for marked objects and we set the second to least significant
-        // bit for pinned objects.  So if we want the actual MT pointer during a GC
-        // we must zero out the lowest 2 bits.
-        return dac_cast<PTR_MethodTable>((dac_cast<TADDR>(m_pMethTab)) & ~((UINT_PTR)3));
+        // lose GC marking bit
+        // A method table pointer should always be aligned.
+        // During GC we set the least significant bit for marked objects.
+        // So if we want the actual MT pointer during a GC we must zero out the lowest bit.
+        // In the past we had put pinning bit here but now it's in sync block.
+        return dac_cast<PTR_MethodTable>((dac_cast<TADDR>(m_pMethTab)) & (~MARKED_BIT));
     }
 
     // There are some cases where it is unsafe to get the type handle during a GC.
--- a/src/debug/daccess/dacimpl.h
+++ b/src/debug/daccess/dacimpl.h
@@ -1682,9 +1682,6 @@ public:
         // clear the GC flag bits off the MethodTable
         // equivalent to Object::GetGCSafeMethodTable()
         *mt &= ~1;
-        // Ensure nobody use the last but one bit
-        // Should be removed after CI
-        _ASSERTE((*mt & 3) == 0);
         return true;
     }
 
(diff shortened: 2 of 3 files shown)


src/vm/object.h
```diff
@@ -634,12 +634,12 @@ class Object
...
-        // A method table pointer should always be aligned.  During GC we set the least 
-        // significant bit for marked objects and we set the second to least significant
-        // bit for pinned objects.  So if we want the actual MT pointer during a GC
-        // we must zero out the lowest 2 bits.
-        return dac_cast<PTR_MethodTable>((dac_cast<TADDR>(m_pMethTab)) & ~((UINT_PTR)3));
+        // lose GC marking bit
+        // A method table pointer should always be aligned.
+        // During GC we set the least significant bit for marked objects.
+        // So if we want the actual MT pointer during a GC we must zero out the lowest bit.
+        // In the past we had put pinning bit here but now it's in sync block.
```
adityamandaleeka: > In the past we had put pinning bit here but now it's in sync block.
Remove this. It makes it sound like the pinning bit was the lowest bit in the past.

adityamandaleeka:
There's also code in the DAC that's using this technique and should probably be kept in sync.

https://github.com/dotnet/coreclr/blob/master/src/debug/daccess/dacimpl.h#L1682

adityamandaleeka:
@Maoni0 Do you know what the history of this was?

adityamandaleeka:
@303248153 My general advice for testing potentially dangerous changes like this, even if you're entirely sure about the change, is to add a temporary check that verifies (in all builds) that your assumption is true and then run it through all the test CI jobs you can. Once everything is passes, you can remove that check (or make it a normal assert, depending on whether the perf tradeoff is worth it).

In this particular case, you want to make sure that the last two bits are both 0 after you clear the mark bit (take a look at `_ASSERTE_ALL_BUILDS`).

Maoni0:
@adityamandaleeka originally both were on the methodtable*, for server GC the pinned bit was moved to syncblk (for perf); at some point it was decided that there was no need to have this distinction - pinning should be fairly infrequent anyway.

303248153:
@adityamandaleeka Thanks for your advice,
Looks like the assertion cause a test error, but I can't figure out the what's the reason.
I'll test on my environment and provide more information later.

303248153:
I run "GC\Regressions\v2.0-rtm\544701\544701\544701.cmd" on my windows 8.1 x64 5 times and they all passed, I saw "Timed Out" in [the error log](https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_windows_nt_prtest/4681/testReport/junit/GC_Regressions/_v2_0_rtm_544701_544701_544701_/_v2_0_rtm_544701_544701_544701_cmd/), maybe just because assertion cost more times? How can I solve this?

adityamandaleeka:
@303248153 That time-out has nothing to do with your change, and has been observed before.

303248153:
Looks like all test passed after I change 3 to 1 and added the assertion, now can I remove the assertion and wait for merge?

adityamandaleeka:
@303248153 Yes, please remove the assertions.

303248153:

adityamandaleeka:
@303248153 Please address my comment above, and then we can merge your PR.

303248153:
Ok, I didn't see that before :p

jkotas:
It would be better to just say that the low two bits are used by the GC. 

It would be nice to treat the fact that the GC is using just one of them right now and that it is for marking as GC implementation detail that the VM knows nothing about.

303248153:
@jkotas
That's reasonable to me, but the comment still should be fixed, I will open another PR and continue to see

PR #10516, opened 2017-03-28
Fix comment on mark::recover_plug_info

303248153:
I think the comment on recover_plug_info has a small mistake.
The original text is
```
We should think about whether it's really necessary to have to copy back the pre plug
info since it was already copied during compacting plugs. But if a plug doesn't move
by < 3 ptr size, it means we'd have to recover pre plug info.
```

It should be "doesn't move by >= 3 ptr" or "only move by < 3 ptr",
I replaced them with

```
We should think about whether it's really necessary to have to copy back the pre plug
info since it was already copied during compacting plugs. But if a plug doesn't move
by >= 3 ptr size (the size of gap_reloc_pair), it means we'd have to recover pre plug info.
```

### DIFF
--- a/src/gc/gc.cpp
+++ b/src/gc/gc.cpp
@@ -5589,7 +5589,7 @@ public:
 
     // We should think about whether it's really necessary to have to copy back the pre plug
     // info since it was already copied during compacting plugs. But if a plug doesn't move
-    // by < 3 ptr size, it means we'd have to recover pre plug info.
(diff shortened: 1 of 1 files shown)


danmosemsft:
@swgillespie

swgillespie:
Seems correct to me, thank you! We do have a fair amount of outdated or incorrect comments and it's great to get them cleaned up. Thanks!

PR #10519, opened 2017-03-28
[x86/Linux] Align frame before SetObjectAppDomain call

parjong:
This commit revises JIT_TrialAlloc::EmitSetAppDomain to align stack frame before SetObjectAppDomain call (discovered while GC stress test).


parjong:
\CC @wateret

parjong:
@janvorli CI is all green :) Please take a look.

PR #10525, opened 2017-03-28
Ensure using the given executable path.

### LINKED ISSUE #5631: CoreRun tries to execute .exe from incorrect path.

mazong1123:
Fix the bug that corerun trying to load the executable from current directory even if the user has specified a full path of the executable.

Fix #5631

### DIFF
--- a/src/coreclr/hosts/corerun/corerun.cpp
+++ b/src/coreclr/hosts/corerun/corerun.cpp
@@ -470,6 +470,18 @@ bool TryRun(const int argc, const wchar_t* argv[], Logger &log, const bool verbo
         return false;
     }
 
+    StackSString tpaList;
+    if (!managedAssemblyFullName.IsEmpty())
+    {
+        // Target assembly should be added to the tpa list. Otherwise corerun.exe
+        // may find wrong assembly to execute.
+        // Details can be found at https://github.com/dotnet/coreclr/issues/5631
+        tpaList = managedAssemblyFullName;
+        tpaList.Append(W(';'));
+    }
+
+    tpaList.Append(hostEnvironment.GetTpaList());
+
     //-------------------------------------------------------------
 
     // Create an AppDomain
@@ -499,7 +511,7 @@ bool TryRun(const int argc, const wchar_t* argv[], Logger &log, const bool verbo
     };
     const wchar_t *property_values[] = { 
         // TRUSTED_PLATFORM_ASSEMBLIES
-        hostEnvironment.GetTpaList(),
+        tpaList,
         // APP_PATHS
         appPath,
         // APP_NI_PATHS
--- a/src/binder/assemblybinder.cpp
+++ b/src/binder/assemblybinder.cpp
@@ -553,7 +553,6 @@ namespace BINDER_SPACE
                                          /* in */  PEAssembly          *pParentAssembly,
                                          /* in */  BOOL                 fNgenExplicitBind,
                                          /* in */  BOOL                 fExplicitBindToNativeImage,
-                                         /* in */  BOOL                 fUseExplicitFilePath,
                                          /* in */  bool                 excludeAppPaths,
                                          /* out */ Assembly           **ppAssembly)
     {
@@ -612,7 +611,6 @@ namespace BINDER_SPACE
                                         assemblyPath,
                                         fDoNgenExplicitBind,
                                         fExplicitBindToNativeImage,
-                                        fUseExplicitFilePath,
                                         excludeAppPaths,
                                         &bindResult));
             }
@@ -986,7 +984,6 @@ namespace BINDER_SPACE
                                          PathString         &assemblyPath,
                                          BOOL                fNgenExplicitBind,
                                          BOOL                fExplicitBindToNativeImage,
-                                         BOOL                fUseExplicitFilePath,
                                          bool                excludeAppPaths,
                                          BindResult         *pBindResult)
     {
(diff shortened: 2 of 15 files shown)


mazong1123:
@jkotas @gkhanna79 @danmosemsft  just fixed #5631. Could you please review this PR? Thanks.

gkhanna79:
I believe the behaviour you are seeing is because of the following (based upon the description in https://github.com/dotnet/coreclr/issues/5631):

1) CoreCLR is in /A
2) CoreRun is in /A
3) A.dll is in /A

When you attempt to execute `./corerun /B/A.dll`, TPAList had /A/A.dll in it since you copied it next to CoreCLR (and TPA is generated from the folder containing CoreCLR). As a result, when we try to find A, we find it in TPA and resolve it from there.

I think the right fix is to have the absolute path of the assembly being executed to be added to TPA list (i.e. add https://github.com/dotnet/coreclr/blob/master/src/coreclr/hosts/corerun/corerun.cpp#L411 to TPA list). 

I do not believe you need to change the binder.

mazong1123:
@gkhanna79 Unfortunately, the absolute path of the assembly in TPA list will never be used. In [AssemblyBinder::BindByTpaList](https://github.com/dotnet/coreclr/blob/master/src/binder/assemblybinder.cpp#L1341), we only check the simple name of the assembly. Since there's no flag indicates whether we're passing an absolute path, it looks like we have no chance to add special logic inside `AssemblyBinder::BindByTpaList`.

I have to compromise to the usage of BindByTpaList so I added a [special check](https://github.com/mazong1123/coreclr/blob/d237e6329f85132429176a0644cf6d93c9437ff4/src/binder/assemblybinder.cpp#L1023):

```cpp
        if (fUseExplicitFilePath)
        {
            // We're using explicit file path so we have already got the result.
            pBindResult->SetResult(pAssembly);
            GO_WITH_HRESULT(S_OK);
        }
        else
        {
```
Please let me know if there's any better idea :)

mazong1123:
I got a better idea today. Add following code into [AssemblySpec::LoadAssembly(LPCWSTR pFilePath)](https://github.com/dotnet/coreclr/blob/master/src/vm/assemblyspec.cpp#L975) so that we can correct the wrong IL file path loaded in application context. But I'm stuck by finding an existing method to extract file name without extension from pFilePath. I believe there must be an existing function/method to process this. 

```cpp
    // TODO: Extract file name without extension as simpleName from pFilePath.
    PathString simpleName(pFilePath);

    if (spec.GetAppDomain()->GetTPABinderContext()->GetAppContext()->IsTpaListProvided())
    {
        BINDER_SPACE::SimpleNameToFileNameMap* tpaMap = spec.GetAppDomain()->GetTPABinderContext()->GetAppContext()->GetTpaList();
        const BINDER_SPACE::SimpleNameToFileNameMapEntry* existingEntry = tpaMap->LookupPtr(simpleName.GetUnicode());
        if (existingEntry != nullptr && wcscmp(existingEntry->m_wszILFileName, pFilePath) != 0)
        {
            BINDER_SPACE::SimpleNameToFileNameMapEntry mapEntry;
            mapEntry.m_wszSimpleName = simpleName.GetUnicode();
            mapEntry.m_wszILFileName = const_cast<LPWSTR>(pFilePath);
            mapEntry.m_wszNIFileName = existingEntry->m_wszNIFileName;

            tpaMap->AddOrReplace(mapEntry);
        }
    }

```

mazong1123:
@gkhanna79 do you have time to take another look? Basically I correct the wrong IL file path at the begining of AssemblySpec::LoadAssembly so we don't have to change the binder.

mazong1123:
In case anyone is interested, there is another way to fix this issue which I gave up because it requires much more effort and should be implemented on Linux and Windows separately. The idea is to append the given assembly path to the tpa list directly and hanlde the duplicate file names(paths). For instance, on linux, we can append `managedAssemblyAbsolutePath` to tpa list after [unixcoreruncommon/coreruncommon.cpp:332](https://github.com/dotnet/coreclr/blob/master/src/coreclr/hosts/unixcoreruncommon/coreruncommon.cpp#L332).  This approach requires much more effort so I gave up (Is this the OFFICIAL way to fix this issue?).

danmosemsft:
@gkhanna79 could you take a look..

gkhanna79:
>  the absolute path of the assembly in TPA list will never be used. In AssemblyBinder::BindByTpaList

The absolute path is in the TPAEntry structure (see https://github.com/dotnet/coreclr/blob/master/src/binder/assemblybinder.cpp#L1347).

Did you try adding the path to the assembly to be executed on the TPA constructed by the host, which would alleviate the need to change the binder?

mazong1123:
> The absolute path is in the TPAEntry structure (see https://github.com/dotnet/coreclr/blob/master/src/binder/assemblybinder.cpp#L1347).

@gkhanna79 you are right. Sorry for the confusing. I just updated my previous comment.

> Did you try adding the path to the assembly to be executed on the TPA constructed by the host, which would alleviate the need to change the binder?

Actually I'm going to **update** the path on the TPA if it is incorrect. Recall we have following:

CoreCLR is in /A
CoreRun is in /A
A.dll is in /A
Another A.dll is in /B/

If we execute `/A/corerun /B/A.dll` , /A/A.dll will be loaded instead of /B/A.dll which is wrong. In `AssemblySpec::LoadAssembly` , I just check tpa list:

1. If there's an entry whose simpleName is `A`
2. And if the entry's `m_wszILFileName` does not equal to `/B/A.dll` 
See https://github.com/dotnet/coreclr/pull/10525/files#diff-66fe4bc960c4bc1ee9736ffa3711d3c4R1043

If that entry exists, I just change the `m_wszILFileName` to /B/A.dll and keep the rest:
https://github.com/dotnet/coreclr/pull/10525/files#diff-66fe4bc960c4bc1ee9736ffa3711d3c4R1047

gkhanna79:
Your approach does not sounds right to me. Did you try the suggestion I made in https://github.com/dotnet/coreclr/pull/10525#issuecomment-291398490? If you realpath the path passed to you at the command line for the DLL to be executed and add it as the first entry in TPA, I would expect your scenario would work without any changes to the binder.

danmosemsft:
@mazong1123 do you expect to get a chance to look? would love to get this closed out.

mazong1123:
Yes I'll work on this issue this weekend. Sorry for the late as I've got a bunch of daily jobs to deal with this week.

mazong1123:
I just added the path of target assembly to the tpa list and the bug on Linux fixed. I'll fix it on Windows if necessary.

mazong1123:
@gkhanna79 Fixed this bug on both Linux and Windows. Would you please take a look?

src/coreclr/hosts/corerun/corerun.cpp
```diff
@@ -470,6 +470,18 @@ bool TryRun(const int argc, const wchar_t* argv[], Logger &log, const bool verbo
         return false;
     }
 
+    StackSString tpaList;
+    if (!managedAssemblyFullName.IsEmpty())
+    {
+        // Target assembly should be added to the tpa list. Otherwise corerun.exe
+        // may find wrong assembly to execute.
+        // Details can be found at https://github.com/dotnet/coreclr/issues/5631
+        tpaList = managedAssemblyFullName;
```
gkhanna79: Shouldn't you need to resolve it to full path (via realpath or such API) to ensure you have the complete path in the TPA?

mazong1123: IMO the path has already been translate to the full path at by calling [WszGetFullPathName](https://github.com/mazong1123/coreclr/blob/75406dc96e96c38627e246d65712d7d3beaa2460/src/coreclr/hosts/corerun/corerun.cpp#L397) previously. 

For Unix version, we have the similar logic at [coreconsole.cpp](https://github.com/dotnet/coreclr/blob/551a7c5563105f7ec9a9d9273204d0a07a02c4fa/src/coreclr/hosts/unixcoreconsole/coreconsole.cpp#L98) and [osxbundlerun.cpp](https://github.com/dotnet/coreclr/blob/32f0f9721afb584b4a14d69135bea7ddc129f755/src/coreclr/hosts/osxbundlerun/osxbundlerun.cpp#L33)

gkhanna79: Thanks - the change looks good to me.

gkhanna79:
@mazong1123 Let me know if this can be merged.

mazong1123:
@gkhanna79 Please go ahead to merge this PR. Thanks :)

wtgodbe:
@gkhanna79 I just confirmed that I can successfully run Crossgen tests by selectively reverting this change.

PR #10533, opened 2017-03-28
Add PGO support for Clang/LLVM on Unix

dpodder:
Extend PGO support from VC++ on WIN32 to Clang/LLVM on UNIX as well.
* Just like on Windows: if profile data is missing, skip enabling PGO (allows non-PGO builds in branches where we don't publish PGO data).
* PGO with LTO requires additional dependencies (namely a discoverable `ld.gold` and `LLVMgold.so`). To protect against broken support and keep the build flexible across a wider array of distros, attempt to detect whether PGO compilation would work (using cmake's `try_compile()`), and fall back to a non-PGO/non-LTO build if the test fails.

### DIFF
--- a/pgosupport.cmake
+++ b/pgosupport.cmake
@@ -6,10 +6,19 @@ function(clr_pgo_unknown_arch)
     endif()
 endfunction(clr_pgo_unknown_arch)
 
+function(append_prop_strings TargetName PropertyName)
+    foreach(Flag IN LISTS ARGN)
+        set_property(TARGET ${TargetName} APPEND_STRING PROPERTY ${PropertyName} " ${Flag}")
+    endforeach(Flag)
+endfunction(append_prop_strings)
+
 # Adds Profile Guided Optimization (PGO) flags to the current target
 function(add_pgo TargetName)
     if(WIN32)
         set(ProfileFileName "${TargetName}.pgd")
+    elseif(UNIX)
+        # Clang/LLVM uses one profdata file for the entire repo
+        set(ProfileFileName "coreclr.profdata")
     endif(WIN32)
 
     set(CLR_CMAKE_OPTDATA_PACKAGEWITHRID "optimization.${CLR_CMAKE_TARGET_OS}-${CLR_CMAKE_TARGET_ARCH}.PGO.CoreCLR")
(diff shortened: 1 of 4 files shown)


pgosupport.cmake
```diff
@@ -19,25 +28,57 @@ function(add_pgo TargetName)
...
+            if(UNIX)
+                ## On Unix we need to pass PGO flags to the compiler as well as the linker
+                target_compile_options(${TargetName} PRIVATE $<${IsReleaseConfig}:-flto -fprofile-instr-use=${ProfilePath}>)
+            endif(UNIX)
+        endif(CLR_CMAKE_HAVE_PGO AND EXISTS ${ProfilePath})
+    endif(CLR_CMAKE_PGO_INSTRUMENT)
 endfunction(add_pgo)
 
+# Detect whether PGO is supported
+if(UNIX)
```
ellismg: Other places in coreclr, we have configure.cmake which we use to do configure like tests.  Can we do the same thing here?  Using check_cxx_source_compiles should let you do this check without having to check in a new file.

dpodder: Thanks! I wasn't aware of configure.cmake, but it definitely looks like the better approach. Will make the change.

pgosupport.cmake
```diff
@@ -19,25 +28,57 @@ function(add_pgo TargetName)
...
+                elseif(UNIX)
+                    append_prop_strings(${TargetName} LINK_FLAGS_${Config} -flto -fuse-ld=gold -fprofile-instr-use=${ProfilePath})
                 endif(WIN32)
-            endif(EXISTS ${ProfilePath})
-        endif(CLR_CMAKE_PGO_INSTRUMENT)
-    endforeach(ConfigType)
+            endforeach(Config)
+            if(UNIX)
+                ## On Unix we need to pass PGO flags to the compiler as well as the linker
+                target_compile_options(${TargetName} PRIVATE $<${IsReleaseConfig}:-flto -fprofile-instr-use=${ProfilePath}>)
```
janvorli: Since the makefile generator is not a multi-target one, you could just use `if (UPPERCASE_CMAKE_BUILD_TYPE STREQUAL RELEASE OR UPPERCASE_CMAKE_BUILD_TYPE STREQUAL RELWITHDEBINFO)`, which we do in other places. Unless something leaked in without my knowledge, we use the generator expressions for WIN32 only, which uses a multi-target generator.

pgosupport.cmake
janvorli: I agree with @ellismg

pgosupport.cmake
```diff
@@ -19,25 +28,57 @@ function(add_pgo TargetName)
...
-    set(ConfigTypeList RELEASE RELWITHDEBINFO)
+    set(ConfigList RELEASE RELWITHDEBINFO)
+    set(IsReleaseConfig "$<OR:$<CONFIG:RELEASE>,$<CONFIG:RELWITHDEBINFO>>")
 
-    foreach(ConfigType IN LISTS ConfigTypeList)
-        set(LinkFlagsProperty "LINK_FLAGS_${ConfigType}")
-        if(CLR_CMAKE_PGO_INSTRUMENT)
+    if(CLR_CMAKE_PGO_INSTRUMENT)
+        # Unfortunately LINK_FLAGS_* don't support generator expressions, so we need to use a loop
+        foreach(Config IN LISTS ConfigList)
```
dpodder: Sure, I'm happy to rewrite it without the loop and without the helper function to keep it easier to follow.

Problem with the suggestion for LINK_FLAGS is that even on Unix, we need the flags applied only when configuring for release or relwithdebinfo. If I add it to LINK_FLAGS, they would get applied even during a debug build. Generator expressions would be the go-to answer, but they're not supported for LINK_FLAGS or its variants.

I did some private testing and it seems like cmake does the right thing with LINK_FLAGS_{CONFIG}, even when targeting a single-config build system like `make`. The only other approach I can think of is to use target_link_libraries() to add linker flags instead--which *do* support generator expressions--but they end up putting the flags inside of the "additional libraries" msbuild property, which I didn't like the sight of (it works, but seems like more of a hack to me).

pgosupport.cmake
```diff
@@ -19,25 +28,57 @@ function(add_pgo TargetName)
...
             if(WIN32)
-                set_property(TARGET ${TargetName} APPEND_STRING PROPERTY ${LinkFlagsProperty} "/LTCG /GENPROFILE")
+                append_prop_strings(${TargetName} LINK_FLAGS_${Config} /LTCG /GENPROFILE)
```
dpodder: I can remove it, but I've already been setting it (on Windows) and it seems harmless to repeat it (PGO doesn't work without LTCG). Do we prefer for /LTCG to be centralized and a hard-fail if missed?

pgosupport.cmake
```diff
@@ -19,25 +28,57 @@ function(add_pgo TargetName)
...
+            elseif(UNIX)
```
dpodder: Sounds good, I'll change it.

pgosupport.cmake
dpodder: This seemed to work, at least on the version of cmake I was testing, but I'll double-check. Happy to change it if necessary, although if this does work down to the minimum version we officially support, I like the generator approach if only to keep it parallel with `WIN32`.

janvorli: My point was that this is the only place that uses generator expressions for non WIN32, so it would be inconsistent with the rest of the stuff. I'd prefer keeping the convention used at other places.

pgosupport.cmake
janvorli: I am sorry. I don't know why I have not taken into account the fact that it should be set for optimized builds only for Unix too. So I'd also use the `if (UPPERCASE_CMAKE_BUILD_TYPE STREQUAL RELEASE OR UPPERCASE_CMAKE_BUILD_TYPE STREQUAL RELWITHDEBINFO)` for Unix like in the other place I've suggested or use double set_property the same way as on Windows, whichever you prefer.

pgosupport.cmake
janvorli: Hmm, I guess it is safer to keep it at both places then.

pgosupport.cmake
dpodder: Ah, okay. I'll make the change.

dpodder:
@janvorli Thanks for the feedback, made the changes. PTAL

PR #10535, opened 2017-03-29
Add shared directory README

alexperovich:


src/mscorlib/shared/README.md
```diff
@@ -0,0 +1,15 @@
+# .NET Core Runtime Shared Directory
+
+This directory contains the shared sources for System.Private.CoreLib. These are shared between dotnet/corert and dotnet/coreclr.
+
+The sources are synchronized with a mirroring tool that watches for new commits on either side and creates new pull requests (as @dotnet-bot) in the other repository.
+
+## Getting clean CI and merging the mirror PRs
+
+Once the mirror PR is created there is a high chance that the new code will require changes to get a clean CI. Any changes can be added to the PR by checking out the PR branch and adding new commits. Please follow the following guidelines for modifying these PRs.
```
jkotas: I would just say `there is a chance`. There is high chance today because of we are moving a bunch of stuff around. But once the dust settles, it should be much smoother - especially for bug-fix level changes.

src/mscorlib/shared/README.md
```diff
@@ -0,0 +1,15 @@
+# .NET Core Runtime Shared Directory
```
jkotas: `.NET Core Runtime Shared Directory` sounds too general. Call it `System.Private.CoreLib Shared Sources` ?

jkotas:
Add something about conventions: that the shared directory should have minimum of CoreCLR or CoreRT specific code, ifdefs should be minimized but still can be used for one liners, the partial classes should be called *.CoreRT.cs / *.CoreCLR.cs, and otherwise it follows https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

jkotas:
FYI @marek-safar

src/mscorlib/shared/README.md
```diff
@@ -0,0 +1,15 @@
...
+The sources are synchronized with a mirroring tool that watches for new commits on either side and creates new pull requests (as @dotnet-bot) in the other repository.
+
+## Getting clean CI and merging the mirror PRs
+
+Once the mirror PR is created there is a high chance that the new code will require changes to get a clean CI. Any changes can be added to the PR by checking out the PR branch and adding new commits. Please follow the following guidelines for modifying these PRs.
+
+ - **DO NOT** modify the commits made by @dotnet-bot in any way.
+ - **TRY** to only make changes outside of shared.
+   - Changes made in the shared folder in additional commits will get mirrored properly if the mirror PR is merged with a **REBASE**
+ - **ALWAYS** Merge the mirror PR with the **REBASE** option.
```
danmosemsft: If I make a fix directly to the mirror PR, will it ever mirror back?  (Or not, because it does not mirror its own mirror commits)
If not, this should make clear I need to go edit the other repo myself

src/mscorlib/shared/README.md
```diff
@@ -0,0 +1,15 @@
+# .NET Core Runtime Shared Directory
+
+This directory contains the shared sources for System.Private.CoreLib. These are shared between dotnet/corert and dotnet/coreclr.
```
danmosemsft: you could add links to https://github.com/dotnet/coreclr/tree/master/src/mscorlib/shared and https://github.com/dotnet/corert/tree/master/src/System.Private.CoreLib/shared

src/mscorlib/shared/README.md
alexperovich: If you add additional commits to the Mirror PR that change the shared directory it will mirror those if the PR is merged with rebase. It marks its own commits so it can skip them. The added commits will be analyzed like normal. If the PR isn't merged with rebase then it won't even see those commits and problems will happen.

PR #10546, opened 2017-03-29
Fix duplicate fcall entries for GetAssemblyName

### LINKED ISSUE #10508: Duplicate fcall entries for AssemblyName.nGetFileInformation and AssemblyLoadContext.nGetFileInformation
Both of these  have the same name and point at AssemblyNameNative::GetFileInformation.
https://github.com/dotnet/coreclr/blob/76b7d89f6e32ff51ae4809827163864c40a6a76f/src/vm/ecalllist.h#L626
https://github.com/dotnet/coreclr/blob/76b7d89f6e32ff51ae4809827163864c40a6a76f/src/vm/ecalllist.h#L639

This is causing a runtime failure when both methods are called in the same app.

We're hitting this assert, and then the throw results in app failure:
https://github.com/dotnet/coreclr/blob/32f0f9721afb584b4a14d69135bea7ddc129f755/src/vm/ecall.cpp#L408

At runtime we're seeing `Could not load file or assembly '<assemblyName>'. Unspecified error (Exception from HRESULT: 0x80004005 (E_FAIL)`.  Thats because in the case we hit this the AssemblyLoadContext was the victim of that throw and it was in the middle of loading an assembly.

I believe this has been a latent issue for some time and was just recently visible since the AssemblyName.GetAssemblyName method was exposed as part of ns2.0 work.

@gkhanna79 helped me debug this.  I believe we need to do what's suggested here: https://github.com/dotnet/coreclr/blob/32f0f9721afb584b4a14d69135bea7ddc129f755/src/vm/ecall.cpp#L402-L406

/cc @gkhanna79 @sbomer

rahku:
fixes #10508


ericstj:
LGTM

Issue #9706, opened 2017-02-21
[x86/Linux] Unit test generics recursion instantiation failure


seanshpark:
Unit test Loader.classloader.generics.Instantiation.Recursion.genrecur fails with
```
      Assertion failed '!"Unknown node in codegen"' in
      'DomainBoundILStubClass:IL_STUB_UnboxingStub(struct):this' (IL size 31)
       src/jit/codegenxarch.cpp Line: 2215
```

Code block which dropped this is a one big switch-case and didn't handle `genTreeOps::GT_OBJ`

seanshpark:
Simplified test code:
```
using System;

#pragma warning disable 0414


class Test23
{
  interface I<T>
  {
    void m(T x);
  }

  struct S1<T> : I<S2<T>>
  {
    public void m(S2<T> x)
    {
    }
  }

  struct S2<T> : I<S1<T>>
  {
    public void m(S1<T> x)
    {
    }
  }

  public static void DoTheTest()
  {
    S2<object> d = new S2<object>();
    S1<object> s1 = new S1<object>();
    I< S2<object> > is2 = s1;

    /*
      this fails with
      Assertion failed '!"Unknown node in codegen"' in
      'DomainBoundILStubClass:IL_STUB_UnboxingStub(struct):this' (IL size 31)
       src/jit/codegenxarch.cpp Line: 2215
    */
    is2.m(d);

    //I< S1<object> > is1 = d;
    //is1.m(s1);
  }
}

public class Program
{
    public static int Main(string[] args)
    {
        Test23.DoTheTest();
        return 100;
    }
}
```

seanshpark:
https://github.com/dotnet/coreclr/blob/master/Documentation/design-docs/multi-reg-call-nodes.md

seanshpark:
This is done while code gen with `IL_STUB_UnboxingStub`.
@BruceForstall , could you help me with this? Where could this GT_OBJ should be transformed(?) or where in the code should I look at?

seanshpark:
callstack
```
assertAbort() at error.cpp:287 0xb1a3ccef	
CodeGen::genCodeForTreeNode() at codegenxarch.cpp:2,215 0xb1c633e7	
CodeGen::genCodeForBBlist() at codegenlinear.cpp:385 0xb19f9f39	
CodeGen::genGenerateCode() at codegencommon.cpp:3,083 0xb19e19a7	
[...]

PR #10550, opened 2017-03-29
[x86/Linux] Use Portable ArrayInitializeWorker

parjong:
x86 uses optimized ArrayInitializeWorker implementation. Unfortunately, C++ unwinder cannot unwind the stack frame by this optimized ArrayIniitliazeWorker, which leads to baseservices.exceptions.simple.ArrayInit test failure.

This commit revises it to use portable ArrayInitializeWorker instead of optimized one for x86/Linux.

### DIFF
--- a/src/classlibnative/bcltype/arraynative.cpp
+++ b/src/classlibnative/bcltype/arraynative.cpp
@@ -175,7 +175,7 @@ void ArrayInitializeWorker(ARRAYBASEREF * arrayRef,
 
     PCODE ctorFtn = pCanonMT->GetSlot(slot);
 
-#ifdef _X86_
+#if defined(_TARGET_X86_) && !defined(FEATURE_PAL)
     BEGIN_CALL_TO_MANAGED();
 
 
@@ -206,7 +206,7 @@ void ArrayInitializeWorker(ARRAYBASEREF * arrayRef,
     }
 
     END_CALL_TO_MANAGED();
-#else // _X86_
+#else // _TARGET_X86_ && !FEATURE_PAL
     //
     // This is quite a bit slower, but it is portable.
     //
@@ -230,7 +230,7 @@ void ArrayInitializeWorker(ARRAYBASEREF * arrayRef,
 
         offset += size;
     }
-#endif // _X86_
+#endif // !_TARGET_X86_ || FEATURE_PAL
 }
 
 


parjong:
\CC @seanshpark

seanshpark:
@parjong, patch works fine.

janvorli:
@parjong I am fine with the change, but I am curious what happened when the C++ unwinder was trying to unwind it. Did it get wrong eip / esp or something else happened? I don't see anything that should prevent the unwinder to work properly.

janvorli:
@parjong actually, would you be willing to do a quick experiment locally? To try to revert this change and then change the code to use the non DEBUG code for DEBUG too (the C version of making the call instead of the asm code) and see if it fixes the problem?

parjong:
@janvorli I'll let you know when I get the result.

parjong:
@janvorli I tested the following patch, and it also works for me:
```diff
diff --git a/src/classlibnative/bcltype/arraynative.cpp b/src/classlibnative/bcltype/arraynative.cpp
index c54d2f3..b834515 100644
--- a/src/classlibnative/bcltype/arraynative.cpp
+++ b/src/classlibnative/bcltype/arraynative.cpp
@@ -175,7 +175,7 @@ void ArrayInitializeWorker(ARRAYBASEREF * arrayRef,

     PCODE ctorFtn = pCanonMT->GetSlot(slot);

-#if defined(_TARGET_X86_) && !defined(FEATURE_PAL)
+#if defined(_TARGET_X86_)
     BEGIN_CALL_TO_MANAGED();


@@ -190,23 +190,14 @@ void ArrayInitializeWorker(ARRAYBASEREF * arrayRef,

         BYTE* thisPtr = (((BYTE*) OBJECTREFToObject (*arrayRef)) + offset);

-#ifdef _DEBUG
-        __asm {
-            mov ECX, thisPtr
-            mov EDX, pElemMT // Instantiation argument if the type is generic
-            call    [ctorFtn]
-            nop                // Mark the fact that we can call managed code
-        }
-#else // _DEBUG
         typedef void (__fastcall * CtorFtnType)(BYTE*, BYTE*);
         (*(CtorFtnType)ctorFtn)(thisPtr, (BYTE*)pElemMT);
-#endif // _DEBUG

         offset += size;
     }

     END_CALL_TO_MANAGED();
-#else // _TARGET_X86_ && !FEATURE_PAL
+#else // _TARGET_X86_
     //
     // This is quite a bit slower, but it is portable.
     //
@@ -230,7 +221,7 @@ void ArrayInitializeWorker(ARRAYBASEREF * arrayRef,

         offset += size;
     }
-#endif // !_TARGET_X86_ || FEATURE_PAL
+#endif // !_TARGET_X86_
 }
```

If you want, I'll revise this as you suggested.

seanshpark:
So then, I'll check  #9706 issue after this revision.

jkotas:
> `(*(CtorFtnType)ctorFtn)(thisPtr, (BYTE*)pElemMT);`

I do not see how this can work on Linux. I think you would also need to rearrange the arguments to match Linux fastcall.

This is not worth to spend time on optimizing. If you would like to actually optimize it, rewrite it in C#.

janvorli:
@jkotas Linux fastcall has the same order of arguments, see the code generated by clang:
```C
typedef void (__fastcall * CtorFtnType)(BYTE*, BYTE*);

__fastcall void ctorFtn(BYTE* thisPtr, BYTE* pElemMT)
{
}

void test()
{
  BYTE *thisPtr = (BYTE*)1;
  BYTE *pElemMT = (BYTE*)2;
  (*(CtorFtnType)ctorFtn)(thisPtr, (BYTE*)pElemMT);
}
```
```asm
test():                               # @test()
        mov     ecx, 1
        mov     edx, 2
        jmp     ctorFtn(unsigned char*, unsigned char*)          # TAILCALL
```

parjong:
@jkotas When I tested with clang, fastcall calling convention also passes the first argument via ecx, and the second argument via edx. The register argument reordering is required as CLR current simulates ``fastcall`` via ``__attribute__((stdcall, regparm(3)))``. This may explain why this optimization is working.

I guess that once #10410 landed. We may simply use ``__fastcall`` instead of ``__attribute__((stdcall, regparm(3)))``, which allows us not to use ``SWIZZLE_REGARG_ORDER`` for x86/Linux.

Anyway, I'm fine with the current form.

janvorli:
@jkotas I am fine with using the portable version too, I was just curious if the problem with unwind was caused by the inline asm code, which this test has confirmed.

janvorli:
Btw, the following tool is awesome for quickly looking at how various versions of clang, gcc and other compilers compile certain code (for various target architectures), what kind of unwind info it generates etc:
https://gcc.godbolt.org/#

PR #10566, opened 2017-03-29
[Local GC] Remove static fields from GC interface

swgillespie:
This PR eliminates the two static fields currently hanging off of `IGCHeap`: `maxGeneration` and `gcHeapType`. 

For `gcHeapType`, the current GC heap type is split into two variables on each side of the GC/EE interface; `g_heap_type` on the EE side and `g_gc_heap_type` on the GC side. Both components record the heap type upon initialization (for the EE, it's recorded after asking the GC to initialize in either SVR or WKS mode - for the GC, it's recorded after *being* asked to initialize in either SVR or WKS mode). The principal advantage of this (other than eliminating the static variable on the interface) is that `g_heap_type` can be made into a dacvar, which simplifies the DAC code a little bit.

For `maxGeneration`, the only reason it existed in the first place is so the DAC can read it (the EE can call `GetMaxGeneration` on IGCHeap - the DAC can't dispatch virtually on IGCHeap so it read the field instead). This isn't necessary anymore, because one of the dac vars that the GC exports is `max_gen`, which is exactly `maxGeneration`. Since the DAC already uses `max_gen`, we can just delete `maxGeneration` without any problems.

In this PR I also removed the `max_generation` define from `gcinterface.h`, since it does not make sense for this define to coexist with `IGCHeap::GetMaxGeneration`. All but one use of `max_generation` can be trivially replaced with `IGCHeap::GetMaxGeneration`, except one: some new code in `threads.cpp` allocates an array on the stack of size `max_generation`, which now does not work because `GetMaxGeneration` is not known at compile-time. I have changed this to an alloca so that it has the same effect, but I am unsure if this is a reasonable. (In particular I am wary of the security implications - I'm not sure of the threat model for dynamically-loaded components like the JIT or a standalone GC). If this is undesirable, I could change this to a heap allocation.


swgillespie:
@jkotas @Maoni0 @kouvel PTAL?

Maoni0:
I would just do a heap allocation instead of alloca for the usage in threads.cpp.

swgillespie:
@Maoni0 done, thanks!

swgillespie:
Are there any other thoughts/objections about this PR? I think it's ready to merge otherwise.

Maoni0:
sorry didn't get time yesterday...looking at it now

swgillespie:
no problem! thank you!

Maoni0:
I would allocate newDeadThreadGenerationCounts once, clear it in TriggerGCForDeadThreadsIfNecessary and use it that way, instead of calling new every time. you could add it as a new field in ThreadStore and initialize it in ThreadStore::InitThreadStore, unless Kount has objections.

kouvel:
^ That sounds good to me.

Other changes to VM stuff looks fine to me. Out of curiosity, why do you choose to have a virtual function for the max generation instead of just a global constant?

src/vm/threads.cpp
```diff
@@ -6153,7 +6160,7 @@ void ThreadStore::TriggerGCForDeadThreadsIfNecessary()
                 continue;
             }
 
-            if (gcGenerationToTrigger < max_generation &&
+            if (gcGenerationToTrigger < gcHeap->GetMaxGeneration() &&
                 static_cast<int>(gcHeap->WhichGeneration(exposedObject)) > gcGenerationToTrigger)
```
kouvel: Probably should remove the cast here

swgillespie:
> Out of curiosity, why do you choose to have a virtual function for the max generation instead of just a global constant?

For Local GC I was hoping to keep as many details about the GC out of the rest of the runtime, including the number of generations. I don't expect we'll be changing that particular detail of the GC, but we have the flexibility to do so if we'd like.

src/vm/threads.cpp
swgillespie: ok!

swgillespie:
@Maoni0 

> you could add it as a new field in ThreadStore and initialize it in ThreadStore::InitThreadStore

I tried this and ran into trouble - `ThreadStore::InitThreadStore` is called before the GC is initialized, and (at least the way it's set up now) we can't determine the generation count until the GC has been initialized. Would you be opposed to initializing it on first use, instead?

Maoni0:
that's fine.

kouvel:
LGTM

src/vm/threads.cpp
```diff
@@ -6088,11 +6090,21 @@ void ThreadStore::TriggerGCForDeadThreadsIfNecessary()
...
+    {
+        // initialize this field on first use with an entry for every table.
+        s_DeadThreadGenerationCounts = new (nothrow) SIZE_T[gcHeap->GetMaxGeneration() + 1];
+        if (!s_DeadThreadGenerationCounts)
+        {
+            return;
+        }
+    }
+
+    memset(s_DeadThreadGenerationCounts, 0, sizeof(SIZE_T) * (gcHeap->GetMaxGeneration() + 1));
```
Maoni0: >gcHeap->GetMaxGeneration() [](start = 62, length = 26)

nit...you could totally cache the return from this instead of making the virtual call a few times throughput this method...

swgillespie: you're right, will fix.

clang/gcc has `__attribute__((pure))` that enable the compiler to CSE calls like these that are known to be pure - perhaps we could put that attribute on some interface functions. I don't think msvc has any equivalent functionality, sadly...

swgillespie:
thanks!

PR #10567, opened 2017-03-29
Reflect PortableRIDs in Identity packages

gkhanna79:
Identity packages in CoreCLR were missing references to portable RID packages.

@ericstj PTAL


ericstj:

ericstj:
This looks reasonable.  Note that this still means you'll get the versioned RIDs if you restore with something like `win10-x64` or `osx.10.12-x64`.  All this change does is make sure that `win-x64` or `osx-x64` will give you something where before they gave nothing.

gkhanna79:
Correct.

Issue #10290, opened 2017-03-19
`Single.ToString` and `Double.ToString` do not differentiate between -0.0 and +0.0


tannergooding:
This can cause significant confusion when debugging tests or strings where this differentiation matters.

tarekgh:
@gkhanna79 why you think this is v2.0? we shipped like that before and even this is the behavior on the full framework. I am moving it to the future if you don't have any objection.

tannergooding:
This also impacts IEEE compliance for floating-point values.

Namely 5.12.1 External character sequences representing zeros, infinities, and NaNs:
> The conversions (described in 5.4.2) from supported formats to external character sequences and back that recover the original floating-point representation, shall recover zeros, infinities, and quiet NaNs, as well as non-zero finite numbers. In particular, signs of zeros and infinities are preserved

This, much like https://github.com/dotnet/coreclr/issues/13106, would be a "breaking change" for `-0.0` -- CC. @jkotas, @eerhardt, @danmosemsft

tannergooding:
Correct, today you cannot round trip `-0.0` (as it is always converted as just `0.0`, with no sign).

PR #10572, opened 2017-03-29
Fix Double.ToString performance on Linux and OSX

tarekgh:


tarekgh:
@jkotas could you please have a look? thanks.

CC @tannergooding

src/pal/src/cruntime/misctls.cpp
```diff
@@ -128,6 +128,20 @@ PAL_ctime( const PAL_time_t *clock )
     return retval;
 }
 
+unsigned int GetExponent(double d)
+{
+    return (*((unsigned int*)&d + 1) >> 20) & 0x000007ff;
+}
+
+unsigned long long GetMantissa(double d)
```
tannergooding: don't we tend to use `UINT64` elsewhere?

src/pal/src/cruntime/misctls.cpp
```diff
@@ -150,250 +164,155 @@ See MSDN for more information.
...
-        *sign = 1;
-    }
+    const int SCALE_NAN = (int)0x80000000;
+    const int SCALE_INF = 0x7FFFFFFF;
 
-    if ( count > ECVT_MAX_COUNT_SIZE )
+    if (GetExponent(value) == 0x7ff)
     {
-        count = ECVT_MAX_COUNT_SIZE;
+        *dec = GetMantissa(value) != 0 ? SCALE_NAN : SCALE_INF;
```
tannergooding: Doesn't Windows return 1 for the `dec` in both cases?

src/pal/src/cruntime/misctls.cpp
```diff
@@ -150,250 +164,155 @@ See MSDN for more information.
...
+        *sign = GetSign(value);
+        lpStartOfReturnBuffer[0] = '\0';
+        return lpStartOfReturnBuffer;
```
tannergooding: I think we want to keep the `goto done`, that way PERF_EXIT is called appropriately.

src/pal/src/cruntime/misctls.cpp
```diff
@@ -150,250 +164,155 @@ See MSDN for more information.
...
     }
 
-    /* Get the string to work with. */
-    sprintf_s( TempBuffer, sizeof(TempBuffer), FORMAT_STRING, value );
+#ifndef __APPLE__
+    // OSX doesn't support evct_r. it support evct but we cannot use it either as it is not thread safe
+    ecvt_r(value, count, dec, sign, lpStartOfReturnBuffer, ECVT_MAX_BUFFER_SIZE);
+    return lpStartOfReturnBuffer;
```
tannergooding: Same thing about PERF_EXIT

src/pal/src/cruntime/misctls.cpp
```diff
@@ -150,250 +164,155 @@ See MSDN for more information.
...
-
-        *dec = 1;
-        goto done;
+        *sign = 1;
     }
-
-    /* Check to see if it is infinite. */
-    if ( strcmp( "Inf", TempBuffer ) == 0 || strcmp( "-Inf", TempBuffer ) == 0 )
+    
+    if (value == 0.0)
```
tannergooding: This isn't correctly handling sign for negative zero.

src/pal/src/cruntime/misctls.cpp
tarekgh: why?

---
In reply to: [108801136](https://github.com/dotnet/coreclr/pull/10572#discussion_r108801136) [](ancestors = 108801136)

src/pal/src/cruntime/misctls.cpp
janvorli: I would also prefer using explicit size types.

tarekgh: I'll change that.

---
In reply to: [108800801](https://github.com/dotnet/coreclr/pull/10572#discussion_r108800801) [](ancestors = 108800801)

src/pal/src/cruntime/misctls.cpp
```diff
@@ -150,250 +164,155 @@ See MSDN for more information.
...
+    //
 
-        *dec = 1;
-        if ( *TempBuffer == '-' )
-        {
-            *sign = 1;
-        }
-        goto done;
+    int exponentIndex = tempBufferLength - 1;
+    while (TempBuffer[exponentIndex] != 'e' && exponentIndex > 0)
```
tannergooding: nit: It might not matter here, but isn't `strrchr` more efficient, since it might use vectorization in some cases?

src/pal/src/cruntime/misctls.cpp
```diff
@@ -150,250 +164,155 @@ See MSDN for more information.
...
-        if ( isdigit(*lpTempBuffer) )
-        {
-            *lpReturnBuffer = *lpTempBuffer;
-            LoopIndex++;
-            lpReturnBuffer++;
-        }
-        lpTempBuffer++;
-
-        if ( LoopIndex == count + 1 )
+        if (TempBuffer[mantissaIndex] >= '0' && TempBuffer[mantissaIndex] <= '9')
```
tannergooding: nit: It might not matter here, but would two memcpys be more efficient?

src/pal/src/cruntime/misctls.cpp
tarekgh: That is fine. the current behavior is if you are trying to print -0.0 it will produce 0.0 and not -0.0. you may try it your self. so I am not really changing the behavior

---
In reply to: [108801243](https://github.com/dotnet/coreclr/pull/10572#discussion_r108801243) [](ancestors = 108801243)

src/pal/src/cruntime/misctls.cpp
tannergooding: So that `PERFLogFunctionExit` is called in the correct circumstances. If we aren't actually using these ever, then I would rather see us remove them. If they are being used, then the entry and exit calls should be appropriately matched.

jkotas:
FreeBSD build break: use of undeclared identifier 'ecvt_r'

src/pal/src/cruntime/misctls.cpp
tannergooding: It's a trivial check to fix it (just use the `*sign = GetSign(value)` call instead of the `if (value < 0.0)` check) and I'd rather see it fixed so that we can actually differentiate -0.0 from +0.0 in the future (https://github.com/dotnet/coreclr/issues/10290).

tarekgh:
> FreeBSD build break: use of undeclared identifier 'ecvt_r'

I'll look at this one.

src/pal/src/cruntime/misctls.cpp
```diff
@@ -150,250 +164,155 @@ See MSDN for more information.
...
+        (*dec)++;
     }
     else
     {
-        *lpStartOfReturnBuffer = '\0';
-    }
-
-done:
-
-    LOGEXIT( "_ecvt returning %p (%s)\n", lpStartOfReturnBuffer , lpStartOfReturnBuffer );
```
tannergooding: As in a comment above, we should either keep these (because they are used in some scenarios) or we should remove the matching `PERF_ENTRY(_ecvt);` calls on L167.

tarekgh: ok, I'll do that

src/pal/src/cruntime/misctls.cpp
tarekgh: we need to avoid copying the sign and the decimal point. so I have to check the numbers

tannergooding: Yeah, I think the string is small enough that it probably won't matter, but I would imagine that two memcpy's (from first digit to decimal point and from decimal point to last digit) might be marginally faster, if we already know where the decimal point and first digit are.

src/pal/src/cruntime/misctls.cpp
tarekgh: returning 1 is wrong. dec here is really the exponent (or the scale). returning SCALE_NAN or SCALE_INF is the right thing here.

tannergooding: Are we sure that differing from Windows here won't cause a breakage or difference in behavior downstream? If so, then fine. Otherwise, we should match Windows and return 1.

tarekgh: look at 

https://github.com/dotnet/coreclr/blob/52a816d3011f4d03b80b5940dc036b40d701f52d/src/classlibnative/bcltype/number.cpp#L147

we do that before calling _ecvt. actually we can remove this check from here but I just kept it in case we port the code to the managed later.

tannergooding: Alright. Although, It looks like we should never be hitting this conditional anyways (because of that check) and we might just be able to replace the statement with an assert that `value` is neither INFINITY or NAN.

src/pal/src/cruntime/misctls.cpp
tarekgh: I am not going to change this now as there is tracking issue for that. the reason is I am trying to avoid inconsistency between platforms for now. when we address the tracking issue we should go and validate all platforms

tannergooding: Windows does return a sign of 1 for -0.0 though, so not making the change is a difference in behavior.

I think we currently (downstream) ignore the value of sign when value is 0 though, so it doesn't really matter other than being one more place to track down when the issue is resolved.

I'm fine with leaving it, but would appreciate if you could call out the handling difference here in the tracking issue.

tarekgh:
@dotnet-bot help

tarekgh:

src/pal/src/cruntime/misctls.cpp
```diff
@@ -150,250 +155,150 @@ See MSDN for more information.
...
+    _ASSERTE(GetExponent(value) != 0x7ff);
 
-    if ( count > ECVT_MAX_COUNT_SIZE )
-    {
-        count = ECVT_MAX_COUNT_SIZE;
-    }
+#if !defined(__APPLE__) && !defined(__FreeBSD__)
+    // OSX and FreeBSD doesn't support evct_r. it support evct but we cannot use it either as it is not thread safe
+    ecvt_r(value, count, dec, sign, lpStartOfReturnBuffer, ECVT_MAX_BUFFER_SIZE);
+    goto done;
```
janvorli: A nit - the `goto done` is not needed here, the `done` label is right after the #endif.

tarekgh: I would keep it for explicit jumping just in case in the future we may add more code either before the done or before goto

tarekgh:
#10390

tarekgh:

tarekgh:
The failure in System.Threading.Tests.WaitHandleTests is unrelated to my change so I am merging the changes

Issue #10543, opened 2017-03-29
Debug command !dso does not work in server GC mode for .net core app


imcarolwang:
In .net core app, add following runtimeconfig, !dso doesn't work; without the config !dso just work.

> {
>   "runtimeOptions":{ 
>   "configProperties":{
> 			"System.GC.Server": true
> 		     }
> 		   }
> }

swgillespie:
@imcarolwang Can you provide some more details? What version of the CLI are you using? What operating system? What sort of "don't work" - is the output wrong or does it crash?

imcarolwang:
I have repro app built on Win10 with the latest version of .net core SDK installed.
Operating system: tried on Windows10, Windows2012 R2, issue repro on both OS.

**The Details:**

A helloworld app would repro, it's created via "dotnet new console --framework netcoreapp2.0", add   RuntimeIdentifiers node to the **Helloworld.csproj**, like below:
```
<Project Sdk="Microsoft.NET.Sdk"> 
	<PropertyGroup>
	    <OutputType>Exe</OutputType>
 	    <TargetFramework>netcoreapp2.0</TargetFramework>
            <RuntimeFrameworkVersion>2.0.0-beta-001834-00</RuntimeFrameworkVersion>
            <RuntimeIdentifiers>win7-x64</RuntimeIdentifiers>
	</PropertyGroup>
</Project>
```
**program.cs**:
```
using System;
using System.Diagnostics;

namespace helloworld
{
    class Program
    {
        static void Main(string[] args)
        {
            string str = "helloworld";
            Console.WriteLine("calling the debugger.break, test string: {0}", str);
            Debugger.Break();
            Console.WriteLine("press anykey to exit ...");
            Console.Read();
        }
    }
}
```
dotnet build and run, then **dotnet publish HelloWorld.csproj -c Release -r win7x64 -o ReleaseFolder**

update file **helloworld.runtimeconfig.json** under ReleaseFolder with the aforementioned config of **"System.GC.Server": true**

run **windbg -g helloworld.exe**, check command !dso, the output from windbg(**doesn't work**):
```
0:000> !dso
OS Thread Id: 0x438c (0)
RSP/REG          Object           Name
```
without that config, type !dso the output is like the following(**works**):
```
0:000> !dso
OS Thread Id: 0xc24 (0)
RSP/REG          Object           Name
000000D56A97DE68 0000029b8b3c86f0 System.Byte[]
[...]

PR #10576, opened 2017-03-30
Fix an issue where the DAC and GC versions of the heap_segment data s…

swgillespie:
…tructure did not agree on the location of the `heap` field.

Unfortunately it looks like I missed a `static_assert`, which would have revealed that `dac_heap_segment.heap` and `heap_segment.heap` did not have the same offsets. This broke the SOS `DumpStackObject` command for Server GC. This PR ensures that their offsets are correct (by rearranging fields on `heap_segment`) and adds a static assert to ensure that their offsets are the same.

Fixes https://github.com/dotnet/coreclr/issues/10543

cc @Maoni0 @mikem8361 @adityamandaleeka PTAL?

### DIFF
--- a/src/gc/gcpriv.h
+++ b/src/gc/gcpriv.h
@@ -4171,12 +4171,11 @@ public:
     size_t          flags;
     PTR_heap_segment next;
     uint8_t*        background_allocated;
-    uint8_t*        plan_allocated;
-    uint8_t*        saved_bg_allocated;
-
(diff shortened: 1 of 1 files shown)


Maoni0:
:shipit:

adityamandaleeka:

adityamandaleeka:
That test failure is unrelated (and should be fixed now).

swgillespie:
thanks!

PR #10578, opened 2017-03-30
[Linux] Clean up Windows-specific EH code

parjong:
This commit cleans up EH code not to invoke the following Windows-specific methods in Linux:
 - InitSavedExceptionInfo
 - CLRAddVectoredHandlers
 - CLRRemoveVectoredHandlers

### DIFF
--- a/src/vm/ceemain.cpp
+++ b/src/vm/ceemain.cpp
@@ -1893,6 +1893,7 @@ part2:
 
                 TerminateStackProbes();
 
+#ifndef FEATURE_PAL
                 // Unregister our vectored exception and continue handlers from the OS.
                 // This will ensure that if any other DLL unload (after ours) has an exception,
                 // we wont attempt to process that exception (which could lead to various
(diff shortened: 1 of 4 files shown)


janvorli:
@parjong I am not sure if this change wins us anything. Instead of 3 ifdefs, we now have 7 of them. The goal was always to minimize the number of ifdefs when possible, usually by ifdefing out bodies of functions that are not needed on Unix instead of ifdefing out the whole method definitions, declarations and callers.

parjong:
I proposed this PR as I could see that some features (such as vectored handler) are not supported in non-Windows only after I analyzed excep.cpp while reading exceptionhandling.cpp. This PR makes it explicit from the caller side (e.g. exceptionhandling.cpp), which may save analyze time.

As you pointed out, I'm also not sure whether this approach is better or not (it is just my preference).  Feel free to close this PR if you feel that there is no gain.

danmosemsft:
@janvorli please close as suggested if you think so.

Issue #10560, opened 2017-03-29
Update optimization flags/levels in jit interface


JosephTremoulet:
With AOT scenarios now including ngen and crossgen and ready-to-run, desktop and coreclr and corert, and with work in flight on IBC and tiered jitting, it's a good time to take another look at the flags used to control which optimizations the jit performs, and try to get a set in place that will give us the right interface going forward to evolve our codegen/policies in the different scenarios.

JosephTremoulet:
The current setup works like so:

 - When the EE asks the JIT to compile a method, it passes `CorJitFlag`s, which include independently-settable:
   - `CORJIT_FLAG_SPEED_OPT`
   - `CORJIT_FLAG_SIZE_OPT`
   - `CORJIT_FLAG_MIN_OPT` "disable optimization"
   - `CORJIT_FLAG_DEBUG_CODE` "no code mangling"
 - The JIT translates these to `JitFlag`s; the mapping is 1:1
 - The JIT then translates these into three independent things:
   - "debuggable code" vs not
   - Which optimizations to run.  We set this to `CFLG_MINOPT`, which is defined
     as `CFLG_TREETRANS`, if MIN_OPT was requested or debuggable code was
     requested or we're jitting a cctor and the cctor isn't an inlinee;
     otherwise, we set this to `CFLG_MAXOPT`, which is defined as all of them:
     - CLFLG_REGVAR
     - CLFLG_RNGCHKOPT
     - CLFLG_DEADASGN
     - CLFLG_CODEMOTION
     - CLFLG_QMARK
     - CLFLG_TREETRANS
     - CLFLG_INLINING
     - CLFLG_CONSTANTFOLD
     - CLFLG_STRUCTPROMOTE
   - Whether to favor size, speed, or "blended" (setting both flags would just
     result in optimizing for size, and cctors are forced to always optimize
     for side)

JosephTremoulet:
Pulling from the list of flags today and some of the [discussion](https://github.com/dotnet/coreclr/pull/10478#issuecomment-290131924) around tiered jitting, we know that the following are some of the goals that we may have for code to be compiled:

 - Make sure stepping through code in the debugger will work well (we have this today in the "debuggable code" flag)
 - Make sure the jit is using the most straightforward/bulletproof lowering (we have this today in minopts)
 - Make sure the jit is going to spend as little time as possible jitting (we don't have this today, and will want this for fjit)
[...]

PR #10580, opened 2017-03-30
Add Tier0 jit flag

noahfalk:
This flag provides a hook to change the JIT policy in the future and diverge tier0 compilation from the min_opt policy.

This is a follow up change discussed in #10478 

We can replace this flag as needed with even more flexible settings mechanisms, but this is a simple place to start.

### DIFF
--- a/src/inc/corinfo.h
+++ b/src/inc/corinfo.h
@@ -214,11 +214,11 @@ TODO: Talk about initializing strutures before use
 #endif
 
 // Update this one
-SELECTANY const GUID JITEEVersionIdentifier = { /* 61783541-8fc0-44ce-80f7-7789b93a3309 */
-    0x61783541,
-    0x8fc0,
-    0x44ce,
-    { 0x80, 0xf7, 0x77, 0x89, 0xb9, 0x3a, 0x33, 0x09 }
+SELECTANY const GUID JITEEVersionIdentifier = { /* f00b3f49-ddd2-49be-ba43-6e49ffa66959 */
+    0xf00b3f49,
+    0xddd2,
+    0x49be,
+    { 0xba, 0x43, 0x6e, 0x49, 0xff, 0xa6, 0x69, 0x59 }
 };
 
 //////////////////////////////////////////////////////////////////////////////////////////////////////////
--- a/src/inc/corjit.h
+++ b/src/inc/corjit.h
@@ -146,6 +146,8 @@ public:
         CORJIT_FLAG_USE_PINVOKE_HELPERS     = 36, // The JIT should use the PINVOKE_{BEGIN,END} helpers instead of emitting inline transitions
         CORJIT_FLAG_REVERSE_PINVOKE         = 37, // The JIT should insert REVERSE_PINVOKE_{ENTER,EXIT} helpers into method prolog/epilog
         CORJIT_FLAG_DESKTOP_QUIRKS          = 38, // The JIT should generate desktop-quirk-compatible code
+        CORJIT_FLAG_TIER0                   = 39, // This is the initial tier for tiered compilation which should generate code as quickly as possible
+        CORJIT_FLAG_TIER1                   = 40, // This is the final tier (for now) for tiered compilation which should generate high quality code
     };
 
     CORJIT_FLAGS()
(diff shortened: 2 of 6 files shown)


noahfalk:
@AndyAyersMS @JosephTremoulet @BruceForstall @cmckinsey

AndyAyersMS:
I'd rather see this a distinct flag value in corjit.h with the "mapping" onto minopts done on the jit side over in `compInitOptions`. That way it is simpler to experiment with new behaviors.

JosephTremoulet:
> I'd rather see this a distinct flag value in corjit.h with the "mapping" onto minopts done on the jit side over in compInitOptions. That way it is simpler to experiment with new behaviors.

I think it could be useful to have three different views and two mappings:
 - The code to re-jit can just say "this is Tier0", like this change
 - The cor jit flags communicate goals from the VM to the JIT (so Tier0 would map to a FastJit cor jit flag which we don't have yet, the intent of #10560 was to figure out what's the right set there)
 - The jit side would then map the interface flags to the set of optimizations and heuristic knob settings that it wants to run

(and extrapolating from that, I think it would make sense e.g. to move the "use minopts for cctors" rule to the VM side...)

Or am I over-thinking it and the interface should just be the current flags plus a bit to set on tier 0?

AndyAyersMS:
I'd like to see a discretionary policy decisions happen in one place if possible, which likely means they should happen in the jit. If the VM knows something about a method that entails certain codegen, then it can ask for it specifically. But if it is up to the jit what kind of code is appropriate, then let's try and put all that logic in one place, and have the VM pass in whatever extra information it has to the jit, so the jit can make reasonable choices.

noahfalk:
I'm happy to do whatever works for you guys. @JosephTremoulet if you agree with @AndyAyersMS let me know and I'll make the update. If you want to continue discussing I'll wait to make further changes until an agreement has been reached. No rush at my end.

JosephTremoulet:
> I'd like to see a discretionary policy decisions happen in one place if possible

That sounds reasonable.  Extrapolating from that, then, it seems like
 - When the EE passes `CORJIT_FLAG_MIN_OPT` to the jit, this really simply means "the NoOptimization flag was set in the method's metadata (or COMPLUS var)" (this seems to be how the VM sets it today)
 - We could get rid of `CORJIT_FLAG_SIZE_OPT` and `CORJIT_FLAG_SIZE_OPT` (these seem to only be set for COMPLUS vars?)
 - The JIT should be able to distinguish single-shot jit vs tier 0 vs tier 1 -- this could be two new flags, or (maybe?) some combination of one flag and `#ifdef FEATURE_TIERED_COMPILATION`
 - The JIT should really/eventually be informing the `TieredCompilationManager` what conditions should prompt re-jitting at the next tier (presumably by compiling checks and conditional calls to a "rejit this" helper into the jitted code), though that raises some interesting questions about what conditions to bake into a method being compiled AOT...

Does that all sound right?

noahfalk:
Thanks Joe!

> •The JIT should really/eventually be informing the  TieredCompilationManager  what conditions should prompt re-jitting at the next tier (presumably by compiling checks and conditional calls to a "rejit this" helper into the jitted code), though that raises some interesting questions about what conditions to bake into a method being compiled AOT...

I think it is very reasonable for arbitrary jit instrumentation to be a trigger for tiered compilation, but it is not something I am trying to build in the short term. I think it will be something we want longer term and I'll avoid doing anything that would make it harder to achieve when the time comes.

> •The JIT should be able to distinguish single-shot jit vs tier 0 vs tier 1 -- this could be two new flags, or (maybe?) some combination of one flag and  #ifdef FEATURE_TIERED_COMPILATION 

If you want single-shot to potentially be different than tier 1 we should have a distinct flag. Even when FEATURE_TIERED_COMPILATION is on, there are still going to be methods that we determine can't be tiered for various reasons.

> •We could get rid of  CORJIT_FLAG_SIZE_OPT  and  CORJIT_FLAG_SIZE_OPT  (these seem to only be set for COMPLUS vars?)

I'm not opposed at all to the principle, but just for logistics I'm going to leave my change scoped to work that is immediately relevant for tiered jitting.

JosephTremoulet:
LGTM (assuming formatting fix), except I'd expect we should update the JIE/EE interface guid

AndyAyersMS:
LGTM too.

noahfalk:
> assuming formatting fix

Did I miss a comment somewhere? I'm not sure what you are refering to.

AndyAyersMS:
There are formatting legs failing in CI. Details should be there in the logs.

You can:
* fix by hand, or 
* apply the patches from the CI, or
* clone the dotnet/jitutils repo and run the jit-format tool in fix mode

noahfalk:
Ah thanks! I checked the logs and all makes sense once again.

noahfalk:

PR #10583, opened 2017-03-30
[ARM32] Let R2R set header flag of NI file for ARM32

hqueue:
Set header flag of NI file for ARM32 generated from ReadyToRun

### DIFF
--- a/src/zap/zapheaders.cpp
+++ b/src/zap/zapheaders.cpp
@@ -34,11 +34,14 @@ void ZapImage::SaveCorHeader()
     corHeader.MinorRuntimeVersion = VAL16(COR_VERSION_MINOR);
     corHeader.Flags = VAL32(COMIMAGE_FLAGS_IL_LIBRARY);
 
-#ifdef _TARGET_X86_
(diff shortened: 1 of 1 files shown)


hqueue:
Thank you for the feedback.

I should have discussed this before making PR after reading `corflags` in inc/corhdr.h as below.
```
//  32BITREQUIRED  32BITPREFERRED
//        0               0         :   no special meaning, MachineType and ILONLY flag determine image requirements
//        0               1         :   illegal, reserved for future use
//        1               0         :   image is x86-specific
//        1               1         :   image is platform neutral and prefers to be loaded 32-bit when possible
```
I found that R2R sets MachineType as ARM/Linux for ARM/Linux.

Is it enough for ARM/Linux by setting` MachineType` correctly without updating `corflags` in NI file?
I'd like to make sure whether my assumption is right or not.

jkotas:
> Is it enough for ARM/Linux by settingMachineType correctly

Yes, that should be enough.

hqueue:
Thanks! Closing this PR.

PR #10584, opened 2017-03-30
Move ObjectFromHandle to the GCHandleTable interface

adityamandaleeka:
This change touches a lot of files, but it's essentially the same idea everywhere: Rather than using the `ObjectFromHandle` macro that was in objecthandle.h, put `ObjectFromHandle` on the `GCHandleTable` interface and modify the callers to call that. 

The DAC will use HndFetchHandle directly since for now we don't want to deal with getting it to call virtual functions on the GC side.

### DIFF
--- a/src/gc/gchandletable.cpp
+++ b/src/gc/gchandletable.cpp
@@ -22,3 +22,13 @@ void GCHandleTable::Shutdown()
 {
     Ref_Shutdown();
 }
+
+void* GCHandleTable::GetHandleTableContext(HHANDLETABLE hTable)
+{
+    return (void*)((uintptr_t)::HndGetHandleTableADIndex(hTable).m_dwIndex);
+}
+
+HHANDLETABLE GCHandleTable::GetHandleTableForHandle(OBJECTHANDLE handle)
+{
+    return ::HndGetHandleTable(handle);
+}
--- a/src/gc/gchandletable.cpp
+++ b/src/gc/gchandletable.cpp
@@ -23,12 +23,12 @@ void GCHandleTable::Shutdown()
     Ref_Shutdown();
 }
 
-void* GCHandleTable::GetHandleTableContext(HHANDLETABLE hTable)
+void* GCHandleTable::GetHandleTableContext(void* handleTable)
 {
-    return (void*)((uintptr_t)::HndGetHandleTableADIndex(hTable).m_dwIndex);
+    return (void*)((uintptr_t)::HndGetHandleTableADIndex((HHANDLETABLE)handleTable).m_dwIndex);
 }
 
-HHANDLETABLE GCHandleTable::GetHandleTableForHandle(OBJECTHANDLE handle)
+void* GCHandleTable::GetHandleTableForHandle(OBJECTHANDLE handle)
 {
-    return ::HndGetHandleTable(handle);
+    return (void*)::HndGetHandleTable(handle);
 }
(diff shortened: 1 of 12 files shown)


adityamandaleeka:
@Maoni0 @swgillespie @sergiy-k PTAL

jkotas:
It is ok to do the debug-only validation via the interface, but it has to stay as simple pointer dereference in the release builds.

adityamandaleeka:
You're right. I'll make an `ObjectFromHandle` on the VM side that calls the interface method in debug and does a simple dereference in release.

adityamandaleeka:
@jkotas @Maoni0 PTAL

src/gc/gchandletable.cpp
```diff
@@ -22,3 +22,8 @@ void GCHandleTable::Shutdown()
 {
     Ref_Shutdown();
 }
+
+Object* GCHandleTable::ObjectFromHandle(OBJECTHANDLE handle)
+{
+    return OBJECTREFToObject(::HndFetchHandle(handle));
```
jkotas: When you look at what HndFetchHandle actually does, it gets AppDomain index, and then validates a bunch of things on the AppDomain. All these are bad dependencies for local GC. Have you thought about what we want to about those?

I am thinking that the whole validation should stay on the VM side, and the only method on the GC/EE interface should be a method to fetch AppDomain index from handle table. (It would be nice to not call it AppDomain index, but something more general - like `void* context`.)

Maoni0: yep, @swgillespie and I talked about what should be done wrt AD. aside from the alloc/survived bytes info that GC tracks for appdomain resource manager purpose, ideally everything else should goto the EE side.

adityamandaleeka: OK. I was already planning to expose the AppDomain index for a handle table via the interface, so I'll do what @jkotas suggested above.

src/gc/gchandletable.cpp
```diff
@@ -22,3 +22,13 @@ void GCHandleTable::Shutdown()
 {
     Ref_Shutdown();
 }
+
+uint32_t GCHandleTable::GetHandleTableContext(HHANDLETABLE hTable)
```
jkotas: I would make this `void*` and cast it to int32_t on the VM side.

src/gc/gcinterface.h
```diff
@@ -149,10 +149,18 @@ struct segment_info
...
 };
 
+
+/*
+ * handle to handle table
+ */
+typedef DPTR(struct HandleTable) PTR_HandleTable;
+typedef DPTR(PTR_HandleTable) PTR_PTR_HandleTable;
+typedef PTR_HandleTable HHANDLETABLE;
+typedef PTR_PTR_HandleTable PTR_HHANDLETABLE;
```
jkotas: This leaks the impl details to the interface. I think HHANDLETABLE should be void* on the interface, and it should be case internally to PTR_HandleTable, etc.

src/vm/exceptionhandling.h
```diff
@@ -253,7 +253,11 @@ class ExceptionTracker
 
         if (NULL != m_hThrowable)
         {
+#ifndef DACCESS_COMPILE
             return ObjectFromHandle(m_hThrowable);
```
jkotas: Is this change still needed?

jkotas:
LGTM otherwise

adityamandaleeka:
@jkotas Thanks, I've addressed your feedback.

src/gc/gchandletable.cpp
```diff
@@ -22,3 +22,13 @@ void GCHandleTable::Shutdown()
 {
     Ref_Shutdown();
 }
+
+void* GCHandleTable::GetHandleTableContext(void* handleTable)
+{
+    return (void*)((uintptr_t)::HndGetHandleTableADIndex((HHANDLETABLE)handleTable).m_dwIndex);
```
jkotas: In future PR, we should also change `m_dwIndex` to `void* m_pContext`.

PR #10587, opened 2017-03-30
Fix Issue #10022.

pgavlin:
This issue occurred because the JIT currently inserts the P/Invoke
method prolog into the first block of a function without ensuring that
this block will only execute once. This is not safe, as executing the
P/Invoke prolog multiple times can create cycles in the frame list and
casue the stack unwinder to hang.

This change ensures that lowering inserts a new, single-entry basic
block at the beginning of the function if the existing entry block has
more than one predecessor.

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -420,6 +420,9 @@ BasicBlock* Compiler::fgNewBasicBlock(BBjumpKinds jumpKind)
 
 void Compiler::fgEnsureFirstBBisScratch()
 {
+    // This method does not update predecessor lists and so must only be called before they are computed.
+    assert(!fgComputePredsDone);
+
     // Have we already allocated a scratch block?
 
     if (fgFirstBBisScratch())
@@ -439,11 +442,6 @@ void Compiler::fgEnsureFirstBBisScratch()
             block->inheritWeight(fgFirstBB);
         }
 
-        // The first basic block will have a ref count of at least one. This corresponds to the virtual edge
-        // from the prolog to the entry block. We need to remove this virtual edge before adding the scratch
-        // block as a predecessor.
-        fgFirstBB->bbRefs--;
-        fgAddRefPred(fgFirstBB, block);
         fgInsertBBbefore(fgFirstBB, block);
     }
     else
@@ -8069,6 +8067,16 @@ bool Compiler::fgMoreThanOneReturnBlock()
 void Compiler::fgAddInternal()
 {
     noway_assert(!compIsForInlining());
+
+#ifndef LEGACY_BACKEND
+    // The RyuJIT backend requires a scratch BB into which it can safely insert a P/Invoke method prolog if one is
+    // required. Create it here.
+    if (info.compCallUnmanaged != 0)
+    {
+        fgEnsureFirstBBisScratch();
+        fgFirstBB->bbFlags |= BBF_DONT_REMOVE;
+    }
+#endif // !LEGACY_BACKEND
 
     /*
         <BUGNUM> VSW441487 </BUGNUM>
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -438,6 +438,12 @@ void Compiler::fgEnsureFirstBBisScratch()
         {
             block->inheritWeight(fgFirstBB);
         }
+
+        // The first basic block will have a ref count of at least one. This corresponds to the virtual edge
+        // from the prolog to the entry block. We need to remove this virtual edge before adding the scratch
+        // block as a predecessor.
+        fgFirstBB->bbRefs--;
+        fgAddRefPred(fgFirstBB, block);
         fgInsertBBbefore(fgFirstBB, block);
     }
     else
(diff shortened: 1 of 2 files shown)


pgavlin:

src/jit/lower.cpp
```diff
@@ -2835,6 +2835,13 @@ void Lowering::InsertPInvokeMethodProlog()
 
     JITDUMP("======= Inserting PInvoke method prolog\n");
 
+    // If the entry block has more than one incoming edge, insert a preceding scratch BB into which we can safely
+    // insert the P/Invoke method prolog.
+    if (comp->fgFirstBB->countOfInEdges() > 1)
```
BruceForstall: Seems like an unnecessary optimization. Why not just unconditionally call fgEnsureFirstBBisScratch()?

pgavlin: The check is cheap and avoids unnecessarily allocating a new block.

src/jit/flowgraph.cpp
```diff
@@ -438,6 +438,12 @@ void Compiler::fgEnsureFirstBBisScratch()
         {
             block->inheritWeight(fgFirstBB);
         }
+
+        // The first basic block will have a ref count of at least one. This corresponds to the virtual edge
+        // from the prolog to the entry block. We need to remove this virtual edge before adding the scratch
+        // block as a predecessor.
+        fgFirstBB->bbRefs--;
+        fgAddRefPred(fgFirstBB, block);
```
BruceForstall: Why wasn't this a problem before?

We don't always have predecessor lists built when this function is called. Seems like this would be bad in that case.

pgavlin: > Why wasn't this a problem before?

I would assume that this was not a problem before because this function was always called before predecessor lists were created.

> We don't always have predecessor lists built when this function is called. Seems like this would be bad in that case.

It looks like `fgAddRefPred` is designed to handle that case: if we don't have predecessor lists build when `fgAddRefPred` is called, it just bumps `bbRefs`.

src/jit/lower.cpp
BruceForstall: I'd prefer the simplicity of always doing it.

Actually, I think you want to always do it and then change the code to always append to the first block (which will be guaranteed to be a scratch block). Otherwise, in the case a scratch block already exists (which might be the case already, today), the code prepends to other code that's already there. So, we're prepending before debug variable initializations, IBC counts, reverse pinvoke enter calls, just-my-code callbacks, etc.

AndyAyersMS: Other callers don't bother with this optimization so I would agree with Bruce, just call it unconditionally.

AndyAyersMS: In fact we should probably call it much earlier, as soon as we realize we're going to have inline pinvokes, and just assert that was called here.

src/jit/flowgraph.cpp
BruceForstall: Maybe you should assert that bbRefs > 0?

You could also avoid making this change by calling fgEnsureFirstBBisScratch() during fgMorphCall if we see such a call, in preparation for having it available later.

src/jit/lower.cpp
pgavlin: > Actually, I think you want to always do it and then change the code to always append to the first block

Yeah, that's better.

BruceForstall: E.g.,
```
assert(fgFirstBBisScratch());
```

src/jit/flowgraph.cpp
```diff
@@ -420,6 +420,9 @@ BasicBlock* Compiler::fgNewBasicBlock(BBjumpKinds jumpKind)
 
 void Compiler::fgEnsureFirstBBisScratch()
 {
+    // This method does not update predecessor lists and so must only be called before they are computed.
+    assert(!fgComputePredsDone);
+
```
briansull: Thanks for adding this pre-condition

src/jit/flowgraph.cpp
```diff
@@ -8064,6 +8068,16 @@ void Compiler::fgAddInternal()
 {
     noway_assert(!compIsForInlining());
 
+#ifndef LEGACY_BACKEND
+    // The RyuJIT backend requires a scratch BB into which it can safely insert a P/Invoke method prolog if one is
+    // required. Create it here.
+    if (info.compCallUnmanaged != 0)
+    {
+        fgEnsureFirstBBisScratch();
+        fgFirstBB->bbFlags |= BBF_DONT_REMOVE;
```
briansull: Was there a case where this new block was getting removed/merged before you had populated it with a statement node?

pgavlin: Yes. In the case of P/Invoke prologs this block remains empty until lowering.

briansull: OK and I guess we removed  the empty block, unless the BBF_DONT_REMOVE is set.

Maybe we should look into having fgEnsureFirstBBisScratch always set that flag, or maybe that is doing to much as we still want to merge the blocks for the other cases.

briansull:
LGTM

PR #10592, opened 2017-03-30
Add test coverage for ldtoken of open generic methods

MichalStrehovsky:
Regression tests for dotnet/corert#3152.


MichalStrehovsky:
@fadimounir PTAL

PR #10593, opened 2017-03-30
Enable StandAloneGC and Server GC

jeffschwMSFT:
Change necessary to enable standalonegc on server gc.  Now CreateThread in gcenv.windows.cpp matches gcenv.os.cpp.

There was an oversight in GCToOSInterface::CreateThread where the newly created suspended thread was not resumed.  The result is a hang if you build standalone and run with servergc.

Repo:
build.cmd buildstandalonegc skiptests
set CORECLR_SERVER_GC
corerun.exe helloworld.exe

### DIFF
--- a/src/gc/windows/gcenv.windows.cpp
+++ b/src/gc/windows/gcenv.windows.cpp
@@ -597,8 +597,8 @@ bool GCToOSInterface::CreateThread(GCThreadFunction function, void* param, GCThr
         ::SetThreadAffinityMask(gc_thread, (DWORD_PTR)1 << affinity->Processor);
     }
 
-	ResumeThread(gc_thread);
-	CloseHandle(gc_thread);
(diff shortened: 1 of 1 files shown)


jeffschwMSFT:
@Maoni0 @swgillespie

Maoni0:
Cool! thanks Jeff!!

src/gc/windows/gcenv.windows.cpp
```diff
@@ -597,6 +597,9 @@ bool GCToOSInterface::CreateThread(GCThreadFunction function, void* param, GCThr
         ::SetThreadAffinityMask(gc_thread, (DWORD_PTR)1 << affinity->Processor);
     }
 
+	ResumeThread(gc_thread);
```
janvorli: A nit - could you please convert tabs to spaces in these two lines?

PR #10596, opened 2017-03-30
Lookup System.Runtime and mscorlib references by Assembly name

### LINKED ISSUE #10595: ILAsm: Incorrectly uses assembly alias rather than assembly name to determine BaseAsmRef
# Issue

In determining the base library assembly ref used for fundamental types such as `System.Object`, `System.String`, etc; the assembler incorrectly tries to determine if `System.Runtime` is referenced by looking at Assembly aliases, rather than Assembly names.

# Repo

Consider the following assembly definition:

```cil
.assembly TestAssembly { }

.assembly extern System.Runtime as SomethingElse
{
}

.method static void Main()
{
    .entrypoint
    newobj instance void [SomethingElse]System.Object::.ctor()
    pop
    ret
}
```

When processed by ilasm it will not detect the fact that `System.Runtime` is referenced (since there is an alias for `System.Runtime` and it is not equal to `System.Runtime`) and will instead insert a reference to `mscorlib` and use that to resolve `System.Object`:

```cil
.assembly TestAssembly { }

.assembly extern System.Runtime
{
}

.assembly extern mscorlib
{
}

.method static void Main()
{
    .entrypoint
    newobj instance void [mscorlib]System.Object::.ctor()
    pop
    ret
}

```

# Desired Result

ilasm should look for `System.Runtime` by assembly simple name, rather than by alias. If a reference to `System.Runtime` is found, then it should be considered the BaseAsmRef regardless of its alias. Otherwise, the fallback to mscorlib is performed as always.

mletterle:
Rather than by alias.

Fixes #10595

### DIFF
--- a/src/ilasm/asmman.cpp
+++ b/src/ilasm/asmman.cpp
@@ -131,6 +131,17 @@ mdToken             AsmMan::GetAsmRefTokByName(__in __nullterminated const char*
     AsmManAssembly* tmp = GetAsmRefByName(szAsmRefName);
     return(tmp ? tmp->tkTok : mdAssemblyRefNil);
 }
+AsmManAssembly*     AsmMan::GetAsmRefByAsmName(__in __nullterminated const char* szAsmName)
+{
+    AsmManAssembly* ret = NULL;
+    if(szAsmName)
+    {
+        for(int i=0; (ret = m_AsmRefLst.PEEK(i))&&
+            (strcmp(ret->szName,szAsmName)); i++);
+    }
+    return ret;
+}
+
 //==============================================================================================================
 void    AsmMan::SetModuleName(__inout_opt __nullterminated char* szName)
 {
--- a/src/ilasm/asmman.hpp
+++ b/src/ilasm/asmman.hpp
@@ -272,6 +272,8 @@ public:
     void    SetManifestResFile(__in __nullterminated char* szFileName, ULONG ulOffset);
     void    SetManifestResAsmRef(__in __nullterminated char* szAsmRefName);
 
+    AsmManAssembly*     GetAsmRefByAsmName(__in __nullterminated const char* szAsmName);
+    
     mdToken             GetFileTokByName(__in __nullterminated char* szFileName);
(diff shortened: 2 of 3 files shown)


danmosemsft:
@dotnet/jit-contrib do you guys own ilasm?

danmosemsft:

danmosemsft:
@dotnet-bot test Ubuntu x64 Checked Build and Test (log gone)

danmosemsft:
@dotnet/jit-contrib  could you please evaluate this.

RussKeldorph:
@mletterle Apologies (again) for missing this PR.  If you're happy with where #10878 is going, could you close this one?  Thank you very much for helping make ilasm better.  In the future, feel free to mention me directly on any ilasm/ildasm/jit-related issues or PRs so hopefully I don't miss them.

mletterle:
@RussKeldorph While the two PRs do solve the same problem, the code they contain is pretty orthogonal, my PR changes the way BaseAsmRef is determined which will affect how it is determined wherever that call is used. Which may fix other unreferenced issues. Though there is probably a discussion to be had on the whole "BaseAsmRef" idea in general.

Thanks for helping address these long standing issues! Exciting times in cil land...

rartemev:

rartemev:
@mletterle As I can see you changes solve the original issue but still cannot handle chained aliases like test from my PR does.

mletterle:
This PR is specifically to avoid aliases all together and look at the simple name of the Assembly to determine if it is `System.Runtime` for example.

Also I don't believe it's possible to define an alias that refers to another alias. 

Partition II § 6.3 doesn't appear to provide a mechanism for that since it says:

> The dotted name used in .assembly extern shall exactly match the name of the assembly **as declared
with an .assembly directive**, in a case-sensitive manner

emphasis mine.

rartemev:

rartemev:
Since all CI tests have passed it looks good to me. Besides that I am still not sure about alias chains so I'd rather ask @RussKeldorph to clarify this question.

RussKeldorph:
@mletterle Thanks!

PR #10598, opened 2017-03-31
[Local GC] Move some EE-specific finalize-on-unload logic out of the GC

swgillespie:
`IGCHeap::FinalizeAppDomain` ultimately ends up calling `CFinalize::FinalizeSegForAppDomain`, which finalizes every eligible object on the segment. Today, a little bit of EE introspection is done: if the object we are looking at is "agile and finalizable", it might not get finalized. In particular, if the object being finalized is a thread that is started or some other "agile" non-thread object, it'll get migrated to another app domain.

The GC does not care about these details so this PR moves this logic to the EE as a part of `GCToEEInterface`, since it's an EE-specific detail (`FEATURE_REDHAWK` does not do this check). With this change the GC no longer depends on the Thread class's method table.

### DIFF
--- a/src/gc/gc.cpp
+++ b/src/gc/gc.cpp
@@ -36101,43 +36101,21 @@ CFinalize::FinalizeSegForAppDomain (AppDomain *pDomain,
         // if it has the index we are looking for. If the methodtable is null, it can't be from the
         // unloading domain, so skip it.
         if (method_table(obj) == NULL)
-            continue;
-
-        // eagerly finalize all objects except those that may be agile.
-        if (obj->GetAppDomainIndex() != pDomain->GetIndex())
-            continue;
-
-#ifndef FEATURE_REDHAWK
-        if (method_table(obj)->IsAgileAndFinalizable())
         {
-            // If an object is both agile & finalizable, we leave it in the
-            // finalization queue during unload.  This is OK, since it's agile.
-            // Right now only threads can be this way, so if that ever changes, change
-            // the assert to just continue if not a thread.
-            _ASSERTE(method_table(obj) == g_pThreadClass);
-
-            if (method_table(obj) == g_pThreadClass)
-            {
-                // However, an unstarted thread should be finalized. It could be holding a delegate
-                // in the domain we want to unload. Once the thread has been started, its
-                // delegate is cleared so only unstarted threads are a problem.
-                Thread *pThread = ((THREADBASEREF)ObjectToOBJECTREF(obj))->GetInternal();
-                if (! pThread || ! pThread->IsUnstarted())
-                {
-                    // This appdomain is going to be gone soon so let us assign
-                    // it the appdomain that's guaranteed to exist
-                    // The object is agile and the delegate should be null so we can do it
-                    obj->GetHeader()->ResetAppDomainIndexNoFailure(SystemDomain::System()->DefaultDomain()->GetIndex());
-                    continue;
-                }
-            }
-            else
-            {
-                obj->GetHeader()->ResetAppDomainIndexNoFailure(SystemDomain::System()->DefaultDomain()->GetIndex());
-                continue;
-            }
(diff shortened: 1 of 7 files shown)


src/gc/gc.cpp
```diff
@@ -36101,43 +36101,21 @@ CFinalize::FinalizeSegForAppDomain (AppDomain *pDomain,
         // if it has the index we are looking for. If the methodtable is null, it can't be from the
         // unloading domain, so skip it.
         if (method_table(obj) == NULL)
+        {
             continue;
+        }
 
-        // eagerly finalize all objects except those that may be agile.
+        // skip objects not in the unloading appdomain
         if (obj->GetAppDomainIndex() != pDomain->GetIndex())
```
jkotas: This should be inside the ShouldFinalizeObjectForUnload callback as well. `obj->GetAppDomainIndex()` and `pDomain->GetIndex()` are bad dependencies.

swgillespie: will do!

src/vm/gcenv.ee.cpp
```diff
@@ -1349,3 +1349,39 @@ void GCToEEInterface::HandleFatalError(unsigned int exitCode)
 {
     EEPOLICY_HANDLE_FATAL_ERROR(exitCode);
 }
+
+bool GCToEEInterface::ShouldFinalizeObjectForUnload(AppDomain* pDomain, Object* obj)
+{
+    MethodTable *pMT = obj->GetMethodTable();
```
jkotas: This can just `return true` in CoreCLR - it should be never actually called. We do not have AppDomains in CoreCLR.

swgillespie: Ok! Are we still using `#ifndef FEATURE_CORECLR` for this, or should I replace the body of this method with `return true`?

jkotas: You can just replace it with `return true`.

swgillespie:
@Maoni0 PTAL?

src/gc/gc.cpp
```diff
@@ -36101,43 +36101,15 @@ CFinalize::FinalizeSegForAppDomain (AppDomain *pDomain,
         // if it has the index we are looking for. If the methodtable is null, it can't be from the
         // unloading domain, so skip it.
         if (method_table(obj) == NULL)
+        {
             continue;
+        }
 
-        // eagerly finalize all objects except those that may be agile.
-        if (obj->GetAppDomainIndex() != pDomain->GetIndex())
+        // does the EE actually want us to finalize this object? (is it agile?)
```
Maoni0: >(is it agile?) [](start = 65, length = 14)

I wouldn't even say (is it agile?) here - it's just decision that EE makes for whatever reason it chooses.

src/vm/gcenv.ee.cpp
```diff
@@ -1349,3 +1349,10 @@ void GCToEEInterface::HandleFatalError(unsigned int exitCode)
 {
     EEPOLICY_HANDLE_FATAL_ERROR(exitCode);
 }
+
+bool GCToEEInterface::ShouldFinalizeObjectForUnload(AppDomain* pDomain, Object* obj)
+{
+    // CoreCLR does not have appdomains, so this code path is dead. Other runtimes may
+    // choose to inspect the object being finalized here.
+    return true;
```
Maoni0: >  return true; [](start = 2, length = 14)

For changes like this that are very specific to core, you might want to have a uniformed way to annotate them so it'll be easy for you to know to take care of them when we need to port these back to desktop. Perhaps mark with [desktop todo] or something like that.

swgillespie: will do!

src/gc/gc.cpp

swgillespie:

swgillespie:
thanks!

Issue #5825, opened 2016-06-16
Fix C4996 instances


gkhanna79:

rahku:
disable of C4996 error is at following three places
1. /wd4996 from utilcode\CMakeLists.txt 
2.  in function GetOSVersion() in file vm\ceemain.cpp
3.  in file inc\legacyactivationshim.h:#pragma warning(disable:4996)

PR #10601, opened 2017-03-31
1. Remove disable of C4996

rahku:
2. delete some dead code
3. Remove call to deprecated OS apis GetVersion & GetVersionEx.  GetOSVersion api in VM sets OS version to be 0.0.0. Do not use GetOSVersion() anymore.
4. Add implementation of IsWindows8OrAbove & IsWinrtSupported (donno why this was deleted earlier)


rahku:
Fixes issue #10601

rahku:
@vancem could you take a look at changes in eventtrace.cpp. They were ifdef'd out as FEATURE_CORESYSTEM is defined in coreclr repo. So I am removing the changes.

rahku:
@BruceForstall FYI EE no longer sends OS version info to jit. Not sure if this would impact superpmi. I am also removing an assert in legacy jit

src/vm/ceemain.cpp
```diff
@@ -3401,19 +3401,13 @@ void ContractRegressionCheck()
...
-// warning C4996: 'GetVersionExW': was declared deprecated
-// externalapis\windows\winblue\sdk\inc\sysinfoapi.h(442)
-// Deprecated. Use VerifyVersionInfo* or IsWindows* macros from VersionHelpers.
-#pragma warning( disable : 4996 )
-    return WszGetVersionEx(lposVer);
-#pragma warning( default : 4996 )
+    if (lposVer == nullptr)
+        return FALSE;
+    memset(lposVer, 0, sizeof(OSVERSIONINFO));
+    lposVer->dwPlatformId = VER_PLATFORM_WIN32_NT;
```
janvorli: How about Unix? We have GetVersionEx implementation in PAL that sets dwPlatformId to VER_PLATFORM_UNIX.
And if we don't want to call the PAL, could you please remove the GetVersionEx implementation from PAL and also remove the PAL tests that excercise it?
Also, the `#define WszGetVersionEx GetVersionExW` should be removed from winwrap.h in this case.

src/vm/ceemain.cpp
```diff
@@ -3401,19 +3401,13 @@ void ContractRegressionCheck()
...
 
 //
-// GetOSVersion - Gets the real OS version bypassing the OS compatibility shim
-// Mscoree.dll resides in System32 dir and is always excluded from compat shim.
-// This function calls mscoree!shim function via mscoreei ICLRRuntimeHostInternal interface
-// to get the OS version. We do not do this PAL or coreclr..we direclty call the OS
-// in that case.
+// GetOSVersion - OS does not provide apis to get the OSVersion
 //
 BOOL GetOSVersion(LPOSVERSIONINFO lposVer)
```
jkotas: We should delete this, and instead fix the places where it is called.

vancem:
> @vancem could you take a look at changes in eventtrace.cpp. They were ifdef'd out as FEATURE_CORESYSTEM is defined in coreclr repo. So I am removing the changes.

I am OK with these changes.

rahku:
PTAL @gkhanna79 @janvorli

src/mscorlib/src/System/Environment.cs
```diff
@@ -361,69 +361,38 @@ public static Version Version
...
-        internal static bool IsWindows8OrAbove
-        {
-            get
-            {
-                return true;
-            }
-        }
-
+#if !FEATURE_PAL
+        private static Lazy<bool> s_IsWindows8OrAbove = new Lazy<bool>(() =>
```
gkhanna79: Is there a reason you used VerifyVersionInfo as opposed to GetVersion? For Win8+ platforms, it will return 6.2 for non-manifest applications.

Also, did you confirm VerifyVersionInfoW is present in OneCore?

rahku: GetVersion* apis have been deprecated and should not be used. VerifyVersionInfo  api is suggested method to compare OS version now
>For Win8+ platforms, it will return 6.2 for non-manifest applications.

and that is OK since I only need to find if the os is win8 & above.

>Also, did you confirm VerifyVersionInfoW is present in OneCore?

Yes it is present in onecore

rahku:
Fixes issue #5825

PR #10606, opened 2017-03-31
Prepare TimeZoneInfo.cs for move to shared partition.

AtsushiKan:
TimeZoneInfo.cs needs no changes but it brings in
another file CurrentTimeZone.cs.

Which drags in yet another file (Hashtable.cs)
which is widely used inside CoreCLR but aside
from this one file, it seems to be stuff that'll
will never be ported over or already has been
ported over sans Hashtable.

So we'll refactor the memoization logic of
CurrentTimeZone.cs into its own partial file
and share the rest. CoreCLR will use continue to use Hashtable.
CoreRT will use ConcurrentUnifier.

And while we're at it, will rename it to
CurrentSystemTimeZone.cs to match the actual class name.

Once we make a corresponding change on the CoreRT side,
both TimeZone.cs and CurrentSystemTimeZone.cs will
be fully synced and can move to the shared partition.

### DIFF
--- a/src/mscorlib/src/System/CurrentSystemTimeZone.cs
+++ b/src/mscorlib/src/System/CurrentSystemTimeZone.cs
@@ -29,12 +29,8 @@ namespace System
 {
     [Obsolete("System.CurrentSystemTimeZone has been deprecated.  Please investigate the use of System.TimeZoneInfo.Local instead.")]
     [Serializable]
-    internal class CurrentSystemTimeZone : TimeZone
+    internal partial class CurrentSystemTimeZone : TimeZone
     {
-        // The per-year information is cached in in this instance value. As a result it can
-        // be cleaned up by CultureInfo.ClearCachedData, which will clear the instance of this object
-        private readonly Hashtable m_CachedDaylightChanges = new Hashtable();
-
         // Standard offset in ticks to the Universal time if
         // no daylight saving is in used.
         // E.g. the offset for PST (Pacific Standard time) should be -8 * 60 * 60 * 1000 * 10000.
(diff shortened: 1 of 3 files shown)


AtsushiKan:
(This diff's much better in CodeFlow...)

src/mscorlib/src/System/CurrentSystemTimeZone.cs
```diff
@@ -29,12 +29,8 @@ namespace System
 {
     [Obsolete("System.CurrentSystemTimeZone has been deprecated.  Please investigate the use of System.TimeZoneInfo.Local instead.")]
     [Serializable]
-    internal class CurrentSystemTimeZone : TimeZone
+    internal partial class CurrentSystemTimeZone : TimeZone
     {
-        // The per-year information is cached in in this instance value. As a result it can
-        // be cleaned up by CultureInfo.ClearCachedData, which will clear the instance of this object
```
jkotas: There is a bug: CultureInfo.ClearCachedData() does not clear this table. The code was under `FEATURE_CORECLR` ifdef originally, but it was brought back together with CurrentSystemTimeZone. Could you please fix it while you are on it?

AtsushiKan: Wait, it looks like it is clearing it -  CultureInfo.ClearCachedData() calls TimeZone.ResetTimeZone which does what it always has - clear out the cached "currentTimeZone." which eventually orphans the table along with the current timezone.

jkotas: I see. Sorry about the confusion.

AtsushiKan:
An identical commit passed all CI's except Tizen a couple of hours ago - Tizen is complete on this one now. Merging.

PR #10609, opened 2017-03-31
[RyuJIT/ARM32] Localloc implementation in codegen

alpencolt:
Algorithm is almost the same as for other platforms but code
is little bit refactored to improve readability.

Added test to cover almost all possible branches of codegen.
Since arrays allocation of int type is needed multiplication and overflow
checks new tests cover only byte arrays.


alpencolt:

alpencolt:
Part of #8496 
+@Dmitri-Botcharnikov

alpencolt:

alpencolt:
@BruceForstall could you take a look at this PR?

BruceForstall:
Looks like span tests are failing
```
13:10:21 FAILED   - JIT/Performance/CodeQuality/Span/SpanBench/SpanBench.sh
13:10:21                BEGIN EXECUTION
13:10:21                /mnt/j/workspace/dotnet_coreclr/master/checked_ubuntu_tst_prtest/bin/tests/Windows_NT.x64.Checked/Tests/coreoverlay/corerun SpanBench.exe
13:10:21                TestSpanClear<Byte>(1): 0.5819ms
13:10:21                TestSpanClear<Byte>(10): 0.0027ms
13:10:21                TestSpanClear<Byte>(100): 0.0009ms
13:10:21                TestSpanClear<Byte>(1000): 0.0009ms
13:10:21                TestSpanClear<String>(1): 0.02ms
13:10:21                TestSpanClear<String>(10): 0.0045ms
13:10:21                TestSpanClear<String>(100): 0.0042ms
13:10:21                TestSpanClear<String>(1000): 0.0049ms
13:10:21                TestArrayClear<Byte>(1): 0.01ms
13:10:21                TestArrayClear<Byte>(10): 0.0025ms
13:10:21                TestArrayClear<Byte>(100): 0.0022ms
13:10:21                TestArrayClear<Byte>(1000): 0.0022ms
13:10:21                TestArrayClear<String>(1): 0.01ms
13:10:21                TestArrayClear<String>(10): 0.0024ms
13:10:21                TestArrayClear<String>(100): 0.0021ms
13:10:21                TestArrayClear<String>(1000): 0.0024ms
13:10:21                
13:10:21                Unhandled Exception: System.TypeLoadException: Could not load type 'System.Span' from assembly 'System.Memory, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
13:10:21                   at Span.SpanBench.TestSpanAsBytes[T](T[] array, Int32 iterationCount, Boolean untrue)
13:10:21                   at Span.SpanBench.Invoke(Action`1 innerLoop, String nameFormat, Object[] nameArgs)
13:10:21                   at Span.SpanBench.Main(String[] args)
13:10:21                ./SpanBench.sh: line 200: 37312 Aborted                 (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" $ExePath $CLRTestExecutionArguments
13:10:21                Expected: 100
13:10:21                Actual: 134
13:10:21                END EXECUTION - FAILED
```

src/jit/codegencommon.cpp
```diff
@@ -5601,6 +5601,7 @@ void CodeGen::genPushCalleeSavedRegisters(regNumber initReg, bool* pInitRegZeroe
 void          CodeGen::genPushCalleeSavedRegisters()
 #endif
 {
+
```
BruceForstall: nit: remove this gratuitous whitespace addition

src/jit/codegencommon.cpp
```diff
@@ -5653,6 +5654,15 @@ void          CodeGen::genPushCalleeSavedRegisters()
...
     rsPushRegs |= RBM_LR; // We must save the return address (in the LR register)
 
+#if defined(_TARGET_ARM_)
+    if (compiler->compLocallocUsed)
+    {
+        // Don't forget store REG_SAVED_LOCALLOC_SP previous value
+        // since it will be trashed with SP
+        rsPushRegs |= RBM_SAVED_LOCALLOC_SP;
+    }
+#endif // _TARGET_ARM_
```
BruceForstall: This is the wrong place for this. Instead, add the following to CodeGen::genFinalizeFrame() 

```
#if defined(_TARGET_ARM_)
        // If there are any reserved registers, add them to the 
        if (regSet.rsMaskResvd != RBM_NONE)
        {
            regSet.rsSetRegsModified(regSet.rsMaskResvd);
        }
#endif // _TARGET_ARM_
```
(just before:
```
#ifdef DEBUG
    if (verbose)
    {
        printf("Modified regs: ");
```

The idea is that REG_SAVED_LOCALLOC_SP is already correctly in the rsMaskResvd mask.

The legacy JIT does this during code generation, but RyuJIT needs to do this here instead. (It appears to be ok, and cause zero diffs, to do this for legacy JIT as well.)

src/jit/codegencommon.cpp
```diff
@@ -8200,6 +8210,13 @@ void CodeGen::genFinalizeFrame()
             maskCalleeRegsPushed |= maskExtraRegs;
         }
     }
+
+    if (compiler->compLocallocUsed)
+    {
+        // SP is stored in temporary register if localloc is used
+        // This register is REG_SAVED_LOCALLOC_SP
+        maskCalleeRegsPushed |= RBM_SAVED_LOCALLOC_SP;
+    }
```
BruceForstall: remove this

tests/src/JIT/CodeGenBringUpTests/LocallocB_N.cs
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public static int Main()
+    {
+        int ret;
+
+        ret = LocallocB_N(1);
+        if (ret != -1) {
+            Console.WriteLine("Failed on index: " + ret);
+            return Fail;
```
BruceForstall: When a test fails, can you please make the output unique to indicate which case failed? E.g., you have 4 failure modes here, and all output looks the same. I'd like something like:
```
Test 1: Failed on index: ...
...
Test 2: Failed on index: ...
```
etc., just so it's easier to see which test failed without debugging it.

BruceForstall: This comment applies to all your test files.

---
In reply to: [109272951](https://github.com/dotnet/coreclr/pull/10609#discussion_r109272951) [](ancestors = 109272951)

tests/src/JIT/CodeGenBringUpTests/LocallocCnstB1.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        for (int i = 0; i < 1; i++)
+        {
+            if (!CHECK(a[i], (byte) i)) return i;
+        }
+
+        return -1;
+    }
+
```
BruceForstall: Please verify that in a C# release compilation (e.g., /o+), that the JIT actually does generate a localloc. I'm worried that a really smart compiler could figure out what is going on here and optimize the entire function away.

BruceForstall:
The GS cookie check work is unrelated to the localloc work, as far as I can tell. Can you submit it as a separate PR?

src/jit/codegenarm.cpp
```diff
@@ -44,7 +44,39 @@ void CodeGen::genSetRegToIcon(regNumber reg, ssize_t val, var_types type, insFla
 //
 void CodeGen::genEmitGSCookieCheck(bool pushReg)
 {
-    NYI("ARM genEmitGSCookieCheck");
+    noway_assert(compiler->gsGlobalSecurityCookieAddr || compiler->gsGlobalSecurityCookieVal);
+
```
BruceForstall: as far as I can tell, this GS cookie check function is almost exactly the same as the one for arm64. Can we share them, and avoid having this code duplication?

src/jit/codegenarm.cpp
```diff
@@ -1315,6 +1351,328 @@ void CodeGen::genMultiRegCallStoreToLocal(GenTreePtr treeNode)
...
+void CodeGen::genLclHeap(GenTreePtr tree)
+{
+    assert(tree->OperGet() == GT_LCLHEAP);
+
+    GenTreePtr size = tree->gtOp.gtOp1;
+    noway_assert((genActualType(size->gtType) == TYP_INT) || (genActualType(size->gtType) == TYP_I_IMPL));
+
+    // Result of localloc will be returned in regCnt.
+    // Also it used as temporary register in code generation
+    // for soring allocation size
```
BruceForstall: >soring [](start = 11, length = 6)

typo: soring

alpencolt:
It looks there is bug in master. I've checked last commit and Span test is failing too even the legacy backend is used.
Look to #10658

src/jit/codegenarm.cpp
alpencolt: You right, code is mostly the same, I've moved it to the codegencommon. But I had to guard with few `#ifdef` statements which looks in my opinion little bit ugly.

Furthermore a lot of code for both ARM platforms is identical. It looks there is need to make some abstraction level (e.g. codegenarmcommon) for this platforms. Do you have ideas how to do it?

tests/src/JIT/CodeGenBringUpTests/LocallocCnstB1.cs
alpencolt: It generates for `LocallocCnstB1.cs` test with `\o+` following code:
```ASM
000026  2300           movs    r3, 0
000028  B408           push    {r3}
00002A  B408           push    {r3}
00002C  466B           mov     r3, sp
00002E  F8C9 3004      str     r3, [r9+4]
```
Allocated array (with size 8, which is stack alignment) is located in `[r9+4]`, legacy backend does the same.
But `CHECK()` method was inlined, I've added `NoInline` attribute to fix it.

alpencolt:

PR #10611, opened 2017-03-31
Do not throw an assertion on no mscorlib when determining to use legacy debug options

### LINKED ISSUE #10608: ILAsm: Unable to create debugging information when only referencing System.Runtime
# Issue

When attempting to insert `DebuggableAttributes`, ilasm only considers `mscorlib` it does not consider `System.Runtime`. If the assembly only contains `System.Runtime` ilasm crashes with a assertion error.

# Repo

Consider the following il:

```cil
.assembly TestAssembly { }

.assembly extern System.Runtime 
{
}

.method static void Main()
{
    .entrypoint
    newobj instance void [System.Runtime]System.Object::.ctor()
    pop
    ret
}
```

Attempting to ilasm this and using, for example, the "debug=impl" option results in an Assert failure.

# Desired Result

ilasm should properly get the `System.Diagnostic.DebuggableAttribute` reference from whatever has been determined to be the "BaseAsmRef"

# Note

This is actually a duplicate of an [old connect bug of mine](https://connect.microsoft.com/VisualStudio/feedback/details/754111/ilasm-crashes-when-generating-pdb-files-for-an-assembly-that-does-not-reference-mscorlib).

mletterle:
It's okay to use debuggable attributes from other BaseAsmRef's

Fixes #10608

### DIFF
--- a/src/ilasm/asmman.cpp
+++ b/src/ilasm/asmman.cpp
@@ -297,9 +297,10 @@ void AsmMan::EmitDebuggableAttribute(mdToken tkOwner)
     else
     {
         AsmManAssembly *pAssembly = GetAsmRefByName("mscorlib");
-        _ASSERTE(pAssembly != NULL);
-        PREFIX_ASSUME(pAssembly != NULL);
-        fOldStyle = (pAssembly->usVerMajor == 1);
+        if(pAssembly != NULL)
+        {
+            fOldStyle = (pAssembly->usVerMajor == 1);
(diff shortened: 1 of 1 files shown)


danmosemsft:

danmosemsft:
@dotnet/jit-contrib do you own ilasm?

RussKeldorph:
@danmosemsft Yes, ilasm goes in area-Codegen.

PR #10613, opened 2017-03-31
NS2.0 Inject more missing types into CoreRT by sharing them.

AtsushiKan:
- ArgIterator was cleaned up for sharing but actually
  sharing it today is too messy with TypedReference
  being in the wrong namespace on half of CoreRT.

  Some preparatory fixes discovered during the
  attempt will be going in on CoreRT...


jkotas:
I think it may be better for the RuntimeArgumentHandle and ArgIterator to stay unshared since they depend on runtime impl details.

LGTM otherwise

AtsushiKan:

AtsushiKan:

PR #10615, opened 2017-03-31
Cross-Platform Sample-Based Profiler

brianrob:
Implement a sample-based profiler primarily targeted at non-Windows platforms.  It has been tested on Windows, Linux and OSX.  This PR contains:

1. The profiler itself.
2. The event pipe, which is responsible for sending profiler events to the configured source.  Right now, there is only a single source.
3. Ability to enable the profiler using a COMPLUS variable for testing (actual configuration APIs will be implemented in a future PR).
4. Ability to dump profile data to a JSON file for testing (production file format will be implemented in a future PR).
5. Implementation of PAL_nanosleep for use by the profiler and event pipe.

Testing performed:
1. Functional testing of the PR changes.
2. CoreCLR functional test pass.
3. CoreCLR functional test pass with the profiler enabled.
4. Some testing with the profiler and GCStress 1 enabled in an attempt to catch deadlocks in thread suspension.


brianrob:
cc: @valenis

brianrob:

brianrob:

src/inc/crsttypes.h
```diff
@@ -236,6 +237,7 @@ int g_rgCrstLevelMap[] =
     3,			// CrstDynamicMT
     3,			// CrstDynLinkZapItems
     7,			// CrstEtwTypeLogHash
+    11,			// CrstEventPipe
```
brianrob: This is actually tool-generated code.  All of the lines in this array have tabs.  If we want to have spaces here, we probably need to do that as part of porting the tool to run in the coreclr repo.

src/vm/eventpipe.cpp
```diff
@@ -0,0 +1,234 @@
...
+    }
+    CONTRACTL_END;
+
+    if(!s_tracingInitialized)
+    {
+        return;
+    }
+
+    // Take the lock and enable tracing.
+    CrstHolder _crst(&s_initCrst);
```
brianrob: Yes, Enable will be callable in-proc.  I just haven't exposed it yet.

src/vm/eventpipe.cpp
```diff
@@ -0,0 +1,234 @@
...
+        MODE_PREEMPTIVE;
+        PRECONDITION(pThread != NULL);
+    }
+    CONTRACTL_END;
+
+    EX_TRY
+    {
+        if(s_pJsonFile != NULL)
+        {
+            static CommonEventFields eventFields;
```
brianrob: I was going for fewer allocations, but this actually would have turned into a bug as soon as multiple callers exist.  I've fixed this.

src/vm/eventpipejsonfile.h
```diff
@@ -0,0 +1,40 @@
...
+        void FormatCallStack(StackContents &stackContents, SString &resultStr);
+
+        // The output file stream.
+        CFileStream *m_pFileStream;
+
+        // Keep track of if an error has been encountered while writing.
+        bool m_writeErrorEncountered;
+
+        // File-open timestamp for use when calculating event timestamps.
+        LARGE_INTEGER m_FileOpenTimeStamp;
```
brianrob: Fixed.  Thanks for pointing this out.

src/vm/eventpipe.h
```diff
@@ -0,0 +1,117 @@
...
+    DWORD ThreadID;
+};
+
+struct StackContents
+{
+    const static unsigned int MAX_STACK_DEPTH = 100;
+
+    // Array of IP values from a stack crawl.
+    // Top of stack is at index 0.
+    UINT_PTR StackFrames[MAX_STACK_DEPTH];
```
brianrob: Yes, you're right.  Done.

src/pal/src/misc/time.cpp
```diff
@@ -394,3 +394,40 @@ GetTickCount64()
...
+
+    if(requestedTime == NULL)
+    {
+        return -1;
+    }
+
+    req.tv_sec = requestedTime->Seconds;
+    req.tv_nsec = requestedTime->Nanoseconds;
+
+    result = nanosleep(&req, &rem);
```
brianrob: I went ahead and fixed the implementation of PAL_nanosleep to handle interruptions.  I've also removed the remainingTime parameter.

I didn't do anything just yet with requestedTime.  I can convert the parameter to a long, but was trying to match the signature of nanosleep so that it is general use (vs later users having to make changes if they want to sleep for seconds vs. nanoseconds).  Ideally, no one sleeps for very long, of course.

Would you rather block callers from specifying multisecond sleeps, and if this is ever a requirement, the code will have to be fixed, or do you think we should leave it as is.  I don't have a strong opinion either way, but wanted to bring up my thoughts during implementation.

src/vm/sampleprofiler.cpp
```diff
@@ -0,0 +1,145 @@
...
+        MODE_ANY;
+    }
+    CONTRACTL_END;
+
+    // The sampling thread will watch this value and exit
+    // when profiling is disabled.
+    s_profilingEnabled = false;
+
+    // Wait for the sampling thread to clean itself up.
+    while(s_pSamplingThread != NULL)
```
brianrob: Thanks for pointing this out.  Fixed.

src/vm/sampleprofiler.cpp
```diff
@@ -0,0 +1,145 @@
...
+    }
+    CONTRACTL_END;
+
+    // Complete thread initialization and start the profiling loop.
+    if(s_pSamplingThread->HasStarted())
+    {
+        // Switch to pre-emptive mode so that this thread doesn't starve the GC.
+        GCX_PREEMP();
+
+        while(s_profilingEnabled && !g_fEEShutDown)
```
brianrob: Fixed.  Yes, you're right.  The g_fEEShutDown was in there from some testing I was doing, and I don't believe it's needed.

src/inc/crsttypes.h
janvorli: Ah, ok, I didn't know that it is autogenerated.

src/pal/src/misc/time.cpp
janvorli: I also don't have a strong opinion here. My feeling was that it is not worth adding a new type just to be able to pass the argument that can specify long times when I don't see a usage for a non-interruptible sleep longer than 1s.

src/vm/eventpipe.h
```diff
@@ -0,0 +1,154 @@
...
+struct CommonEventFields
+{
+    // Timestamp generated by QueryPerformanceCounter.
+    LARGE_INTEGER TimeStamp;
+
+    // Thread ID.
+    DWORD ThreadID;
+};
+
+struct StackContents
```
janvorli: A nit - now that it is fully encapsulated, you could change it to `class`

brianrob: Done.

src/pal/src/misc/time.cpp
brianrob: Fair enough.  I've made this change as well.

src/vm/eventpipe.cpp
```diff
@@ -0,0 +1,234 @@
...
+        GC_NOTRIGGER;
+        MODE_ANY;
+    }
+    CONTRACTL_END;
+
+    // TODO: Implement filtering.
+    return false;
+}
+
+void EventPipe::WriteEvent(GUID& providerID, INT64 eventID, BYTE *pData, size_t length, bool sampleStack)
```
vancem: I am thinking that we may wish to have abstractions for both providers and events (thus you pass a Provider&  or frankly just a Event& (which points at the provider).    I really don't like GUIDs and I would like to hide them in most code.

src/vm/eventpipe.h
```diff
@@ -0,0 +1,154 @@
...
+
+    const static unsigned int MAX_STACK_DEPTH = 100;
+
+    // Array of IP values from a stack crawl.
+    // Top of stack is at index 0.
+    UINT_PTR m_stackFrames[MAX_STACK_DEPTH];
+
+    // Parallel array of MethodDesc pointers.
+    // Used for debug-only stack printing.
+    MethodDesc* m_methods[MAX_STACK_DEPTH];
```
vancem: If this is debug-only should it be #if DEBUG?

src/vm/eventpipejsonfile.cpp
```diff
@@ -0,0 +1,131 @@
...
+    if(FAILED(m_pFileStream->OpenForWrite(outFilePath)))
+    {
+        delete(m_pFileStream);
+        m_pFileStream = NULL;
+        return;
+    }
+
+    QueryPerformanceCounter(&m_fileOpenTimeStamp);
+
+    SString fileStart(W("{\n\"StackSource\" : {\n\"Samples\" : [\n"));
```
vancem: I realize this is all throw away, but it would be good to have the trace start time (as a DateTime) in the output as well as things like the process ID process name, machine name.    You don't have to do it in this code (since it is throw away), but ultimately we want such things.

src/vm/sampleprofiler.cpp
```diff
@@ -0,0 +1,155 @@
...
+        while(s_profilingEnabled)
+        {
+            // Check to see if we can suspend managed execution.
+            if(ThreadSuspend::SysIsSuspendInProgress() || (ThreadSuspend::GetSuspensionThread() != 0))
+            {
+                // Skip the current sample.
+#ifdef FEATURE_PAL
+                PAL_nanosleep(s_samplingRateInNs);
+#else
+                ClrSleepEx(1, FALSE);
```
vancem: What is the granularity of ClrSleepEx?

src/vm/sampleprofiler.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+#include "common.h"
+#include "sampleprofiler.h"
+#include "hosting.h"
+#include "threadsuspend.h"
+
+Volatile<BOOL> SampleProfiler::s_profilingEnabled = false;
+Thread* SampleProfiler::s_pSamplingThread = NULL;
+CLREventStatic SampleProfiler::s_threadShutdownEvent;
+#ifdef FEATURE_PAL
```
vancem: Can we hide this 'fast wait' behind a common API?  I would like to have as few #if FEATURE_PAL ifdefs as possible.

PR #10544, opened 2017-03-29
Removing Slice on string overloads that return ROS and adding AsSpan

ahsonkhan:
Related to: https://github.com/dotnet/corefx/pull/17407

PR #10617, opened 2017-03-31
Temporary removing string slice span bench test.

ahsonkhan:
Work around while we wait for:
https://github.com/dotnet/coreclr/pull/10544

### DIFF
--- a/tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
+++ b/tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
@@ -1009,7 +1009,7 @@ namespace Span
                 Array.Clear(array, 0, length);
         }
         #endregion
-
+#if false
         #region TestSpanAsBytes<T>
         [Benchmark(InnerIterationCount = BaseIterations)]
         [InlineData(1)]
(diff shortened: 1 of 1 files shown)


ahsonkhan:
cc @AndyAyersMS, @jkotas.

I have to break the circular dependency here.

AndyAyersMS:
LGTM.

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
```diff
@@ -1108,43 +1108,7 @@ public static void TestSpanNonPortableCastFromIntToByte(int length)
         }
         #endregion
 
-        #region TestSpanSliceStringChar<T>
```
jkotas: It would be better to just to put this under `#if false`

jkotas:
LGTM otherwise

tests/src/JIT/Performance/CodeQuality/Span/SpanBench.cs
ahsonkhan: I agree, but I plan to add this back fairly quickly.

ahsonkhan:

Issue #10620, opened 2017-03-31
coreclr doesn't enforce the delegate transparency rule for some IL sequences.


sandreenko:
importer.cpp : impImportBlockCode has code to enforce  the delegate transparency rule:

```
// Insert delegate callout here.
if (opcode == CEE_NEWOBJ && (mflags & CORINFO_FLG_CONSTRUCTOR) && (clsFlags & CORINFO_FLG_DELEGATE))
{
  // In coreclr the delegate transparency rule needs to be enforced even if verification is disabled
  typeInfo              tiActualFtn          = impStackTop(0).seTypeInfo;
  CORINFO_METHOD_HANDLE delegateMethodHandle = tiActualFtn.GetMethod2();
  impInsertCalloutForDelegate(info.compMethodHnd, delegateMethodHandle, resolvedToken.hClass);
}
```

and GetMethod2() was added in PR #4999 as workaround for tests JIT/Methodical/delegate/_simpleoddpower_il_d/_simpleoddpower_il_d.sh
JIT/Methodical/delegate/_simpleoddpower_il_d/_simpleoddpower_il_r.sh
where such IL sequence exists:
```
    IL_01d0:  ldloc.0
    IL_01d1:  dup
    IL_01d2:  
    	call instance method int32 * () OddPower::GetMthdPtr1()
    IL_01d8:  newobj     instance void OddPower/MyDeleg2::.ctor(object,
                                                                native int)
```

call returns TI_ERROR type for pointers, because
`verMakeTypeInfo: case CORINFO_TYPE_PTR: // for now, pointers are treated as an error`
and it was done before the initial GitHub commit (2010).

So in this situation on the EE stack we have TI_ERROR type with nullptr in m_method.
GetMethod2 returns null, delegateMethodHandle == nullptr and impInsertCalloutForDelegate does nothing because isDelegateCreationAllowed returns false for nullptr.

jkotas:
This should be deleted. We should not be enforcing any security transparency rules in CoreCLR anymore.

sandreenko:
isDelegateCreationAllowed  was deleted.

PR #10626, opened 2017-03-31
delete GetMethod2

sandreenko:
The description is in #10620 .
I need to delete this methods because I want to add another class inside the union in typeInfo for resolver token. It is necessary for dotnet/corert#2102. All methods that work with typeInfo should check type.

### DIFF
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -12814,10 +12814,17 @@ void Compiler::impImportBlockCode(BasicBlock* block)
 
 #ifdef FEATURE_CORECLR
                     // In coreclr the delegate transparency rule needs to be enforced even if verification is disabled
-                    typeInfo              tiActualFtn          = impStackTop(0).seTypeInfo;
-                    CORINFO_METHOD_HANDLE delegateMethodHandle = tiActualFtn.GetMethod2();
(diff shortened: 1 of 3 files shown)


sandreenko:
PTAL @BruceForstall @pgavlin @dotnet/jit-contrib.

src/jit/importer.cpp
```diff
@@ -12814,10 +12814,17 @@ void Compiler::impImportBlockCode(BasicBlock* block)
 
 #ifdef FEATURE_CORECLR
                     // In coreclr the delegate transparency rule needs to be enforced even if verification is disabled
```
jkotas: You can delete this whole block instead and the impInsertCalloutForDelegate method. We are not doing any delegate transparency check in CoreCLR. It is left-over from Silverlight in a browser.

sandreenko: I am not familiar with that enough, but it sounds great. Thanks.

sandreenko: Do we want to delete isDelegateCreationAllowed from JitEE interface?

jkotas: > Do we want to delete isDelegateCreationAllowed from JitEE interface?

It would be nice. It should be fine to do it together with the next JITEE interface change.

PR #10632, opened 2017-03-31
Remove unused IsFastSort/SpecialSort string logic

bbowyersmyth:
Every new string that goes through IndexOf, LastIndexOf, Compare, and Equals with a StringComparison.OrdinalIgnoreCase first gets checked to see if it's all ASCII. As part of this check, if it is ASCII, the string is also scanned to see if it has a hyphen, single quote, or control characters and is marked as a STRING_STATE_SPECIAL_SORT instead. However STRING_STATE_SPECIAL_SORT is never independently used so there is no need to do this additional work.

IsFastSort/IS_FAST_CASING are aliases for IsAscii so they have been removed.
IsAscii/IS_ASCII is defined as STRING_STATE_SPECIAL_SORT || STRING_STATE_FAST_OPS. Special sort is a subset of fast ops (ascii) so it is safe to remove it from there if STRING_STATE_SPECIAL_SORT is no longer being set.
The other related macros are unused and have been removed.

Nothing else is calling GetHighCharState and I have scanned the codebase for 0xC0000000 usage. Any other systems to be aware of that might be reading that bit?


bbowyersmyth:
HighCharHelper::IsHighChar and the related HighCharTable array are still referenced in RegMeta::DefineUserString where a flag is written to the "user string". I think that ends up in EEStringData where the flag is not used at all except for cloning. I might send another PR for that if I can be sure.

bbowyersmyth:
Looks like 'IsFastSort' is still being referenced somewhere. I can't find it in any of the source files and the build completes locally. Any clues?

tarekgh:
IsFastSort is needed for the string comparison operation optimization. Fast sort is different than Ascii because you can have some ascii characters which can affect the sorting operation (at least on Linux). so we need to keep IsFastSort.

https://github.com/dotnet/coreclr/blob/cdff8b0babe5d82737058ccdae8b14d8ae90160d/src/mscorlib/src/System/Globalization/CompareInfo.Unix.cs#L200

tarekgh:
> Looks like 'IsFastSort' is still being referenced somewhere. I can't find it in any of the source files and the build completes locally. Any clues?

looks you are building Windows code and not Linux

bbowyersmyth:
Ah yes I see it in the Unix log. Thanks, will investigate.

bbowyersmyth:
@tarekgh It looks like they could all use IsAscii no? Sort order shouldn't have an impact on IndexOf, LastIndexOf, StartsWith, and EndsWith with OrdinalIgnoreCase for an ASCII string.

tarekgh:
No we cannot use IsAscii, there is some characters in the ascii range which has special handling in Linux. like \r\n.

you may look at the issue https://github.com/dotnet/corefx/issues/4587 which we introduced IsFastSort to fix it. at the end of the issue I have some summary which can be helpful describing the behavior.

Issue #4853, opened 2016-05-08
Tests failing due to expected constructor not running


RussKeldorph:
I'm not certain, but these tests look like they may be failing because they are expecting a constructor to run that isn't.   I may also be incorrect about these tests failing for the same issue.  This grouping is just based on an initial impression.

Tests
```
JIT\Methodical\cctor\misc\tail_il_d\tail_il_d.cmd
JIT\Methodical\cctor\misc\tail_il_r\tail_il_r.cmd
```
Output
```
Testing .cctor() invocation by calling virtual method

Before running virtual method
After running virtual method
in Main() after f(ref b), measure.a is 0
FAILED
Expected: 100
Actual: -1
```
Tests
```
JIT\Methodical\cctor\misc\threads3_il_d\threads3_il_d.cmd
JIT\Methodical\cctor\misc\threads3_il_r\threads3_il_r.cmd
```
Output
```
Before calling static field
Thread #0 in .cctor(), measure.a is 0
Thread #1 in .cctor(), measure.a is 0
Thread #3 in .cctor(), measure.a is 0
Thread #2 in .cctor(), measure.a is 0
Thread #3 in .cctor(), after +=8, measure.a is 3
Thread #1 in .cctor(), after +=8, measure.a is 2
Thread #2 in .cctor(), after +=8, measure.a is 4
Thread #4 in .cctor(), measure.a is 0
Thread #4 in .cctor(), after +=8, measure.a is 5
Thread #0 in .cctor(), after +=8, measure.a is 1

After calling static field
in Main(), measure.a is 5
FAILED
Expected: 100
Actual: -1
```
Tests
```
JIT\Methodical\cctor\simple\precise3_il_d\precise3_il_d.cmd
JIT\Methodical\cctor\simple\precise3_il_r\precise3_il_r.cmd
```
Output
```
Testing .cctor() invocation by calling virtual method

Before running virtual method
After running virtual method
in Main() after f(ref b), measure.a is 0
FAILED
Expected: 100
Actual: -1
```

RussKeldorph:
I'm hoping this is just a test issue, but I can't say much for sure yet.

LLITCHEV:
For the tail_il_d and tail_il_r this is a test issue.
[...]

PR #10633, opened 2017-03-31
Jit: fix a few issues with single def local tracking

AndyAyersMS:
Fixes for three issues that came up in desktop testing.

1. Don't track class types for locals when in import only mode,
since the jit may be looking at an uninstantiated generic method.
In these cases the jit's use of TYP_REF to represent CORINFO_TYPE_VAR
confuses the tracking code. In import only mode the jit is not going
to use the information, so there is no need to track it at all.

    Import only mode is tied to verification and CoreCLR runs in full
trust mode, so no test was added.

2. Allow `lvaUpdateClass` to be called more than once but assert
that the second and subsequent calls provide the exact same update
as the first call.

    This can happen for single def ref class locals whose definition
block is reimported. Reimportation is triggered by some tolerable
non-ref type mismatches at block joins and so ref type information
should be consistent for each importation attempt.

    Add a couple of IL test cases translated from destkop MC++ tests
that will trigger this kind of reimportation.

3. Bail out of devirtualization (for desktop only) if the base class
has precise initialization semantics, since desktop seems to trigger
class initialization for ref type callvirts (contrary to ECMA355 I.8.9.5).
See #4853 for some discussion. Since this is desktop only behavior
no test was added.

### DIFF
--- a/src/jit/gentree.cpp
+++ b/src/jit/gentree.cpp
@@ -16379,6 +16379,14 @@ CORINFO_CLASS_HANDLE Compiler::gtGetClassHandle(GenTreePtr tree, bool* isExact, 
     *isExact                      = false;
     CORINFO_CLASS_HANDLE objClass = nullptr;
 
+    // Bail out if we're just importing and not generating code, since
+    // the jit uses TYP_REF for CORINFO_TYPE_VAR locals and args, but
+    // these may not be ref types.
+    if (compIsForImportOnly())
+    {
+        return objClass;
+    }
+
     // Bail out if the tree is not a ref type.
     var_types treeType = tree->TypeGet();
     if (treeType != TYP_REF)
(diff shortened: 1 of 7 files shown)


AndyAyersMS:
@briansull PTAL
cc @dotnet/jit-contrib 

No diffs in CoreCLR on jit-diff over frameworks.

briansull:
Looks Good

AndyAyersMS:
Looks like the new tests fail on x86. Not too surprising since I ported them from x64 MC++. 

Let me see if I can fix them to be less arch-dependent.

briansull:
One additional comment on this

_Reimportation is triggered by some tolerable non-ref type mismatches at block joins and so ref type information should be consistent for each importation attempt._

I think that re-importation is also triggered when a null is push on the IL stack on one path and then on the alternate path a non-null ref type is pushed.

AndyAyersMS:
The cases I see in the importer that trigger reimportation do not involve ref types:
* int32 and native int --> native int
* double and float --> double
* & and native int --> &
* & and int32 --> &

PR #10637, opened 2017-04-01
Remove file system/security/util/URLString.cs

rahku:


rahku:
Fixes issue #8732

rahku:
PTAL @jkotas I have removed function AppDomainSetup.NormalizePath and replaced its usage with Path.GetFullPath.

rahku:
PTAL @JeremyKuhne I have removed the changes that you did for this comment. As it is not applicable for coreclr as there is no quirking config files available in coreclr.
// Long path checks can be quirked, and as loading default quirks too early in the setup of an AppDomain is risky.

src/mscorlib/src/System/AppDomainSetup.cs
```diff
@@ -213,230 +208,29 @@ public String ApplicationBase
...
-                    path = AppDomain.NormalizePath(path, fullCheck: true);
+                Value[(int)LoaderInformation.ApplicationBaseValue] = (value == null || value.Length == 0)?null:Path.GetFullPath(value);
             }
-
-            return path;
         }
-
+        
         public String ConfigurationFile
         {
```
jkotas: This method is not used anywhere. It can be deleted.

jkotas: Bonus points for deleting all other properties in AppDomainSetup that are not used anywhere :-)

jkotas:
LGTM modulo comment.

PR #10638, opened 2017-04-01
Jit: decrement finally nesting level when removing empty trys

### LINKED ISSUE #10621: RyuJIT/x86: assert failure when precompiling System.Security.AccessControl.dll
To repro:
1. Download the [System.Security.AccessControl.dll package](https://dotnet.myget.org/feed/dotnet-core/package/nuget/System.Security.AccessControl/4.4.0-preview1-25131-02)
2. `crossgen runtimes/win/lib/netcoreapp2.0/System.Security.AccessControl.dll`

You should see the following assert:
```
Assert failure(PID 31136 [0x000079a0], Thread: 35532 [0x8acc]): Assertion failed 'treeNode->gtVal.gtVal1 < compiler->compHndBBtabCount' in 'System.Security.AccessControl.Privilege:ToggleState(bool):this' (IL size 407)

    File: c:\dev\coreclr\src\jit\codegenxarch.cpp Line: 2142
    Image: c:\dev\coreclr\bin\Product\Windows_NT.x86.Checked\crossgen.exe
```

This assert does not fire if `COMPlus_JitMinOpts` is set to `1`.

@AndyAyersMS can you take a look?

AndyAyersMS:
For non-funclet EH models, the GT_END_LFIN statement tracks the
nesting level of the associated finally. When removing a try-finally
with an empty try we need to decrement this level for any try-finallys
nested within the associated finally.

Added a test case with several levels of empty try and nesting.

Closes #10621.

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -23375,6 +23375,7 @@ void Compiler::fgRemoveEmptyTry()
         // Handler index of any nested blocks will update when we
         // remove the EH table entry.  Change handler exits to jump to
         // the continuation.  Clear catch type on handler entry.
+        // Decrement nesting level of enclosed GT_END_LFINs.
         for (BasicBlock* block = firstHandlerBlock; block != endHandlerBlock; block = block->bbNext)
         {
             if (block == firstHandlerBlock)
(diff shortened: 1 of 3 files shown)


AndyAyersMS:
@brucefo @pgavlin PTAL
cc @dotnet/jit-contrib 

Let me know if you think it is viable/preferable to use `fgGetFinallyNestingLevel` when processing the `GT_END_LFIN` instead of tracking the level in the tree like we do now.

BruceForstall:
> Let me know if you think it is viable/preferable to use fgGetFinallyNestingLevel when processing the GT_END_LFIN instead of tracking the level in the tree like we do now.

Not sure. `fgGetNestingLevel` is pretty expensive. I would leave it as you have it now.

PR #10645, opened 2017-04-02
Move easy files to shared partition

jkotas:


jkotas:
cc @alexperovich @AtsushiKan

Issue #10647, opened 2017-04-02
Debug.Assert failure should terminate the program unconditionally


jkotas:
Repro:

```
using System;
using System.Diagnostics;

class Program
{
    static void Main(string[] args)
    {
        try
        {
            Debug.Assert(false);
        }
        catch
        {
        }

        Console.WriteLine("Hello World!");
    }
}
```

dotnet restore
dotnet run

Actual result: "Hello World!"
Expected result: Program is terminate with assert message

jkotas:
The problem is that the `Debug.Assert` throws a regular catchable exception today. Instead, it needs to fail fast. There is old [TODO](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Diagnostics/Debug.Windows.cs#L19) pointing to https://github.com/dotnet/corefx/issues/3708 that was never fixed and postponed to future instead.

jkotas:
This is import to fix for v2.0. 

We may be hitting assertion failures in our debug build tests runs today, but we never notice them because of they get caught as exception.

cc @danmosemsft @gkhanna79

danmosemsft:
We want literally `Environment.FailFast` ? Will that produce equally good debug experience?

jkotas:
We should keep the `Debugger.Break()` when debugger is attached: https://github.com/dotnet/coreclr/blob/f355c5789d5d7cd6ea10bee326c66f04641cfc90/src/mscorlib/src/System/Diagnostics/Debug.Windows.cs#L15

bartonjs:
I have to say, the messages going away and the process just crashing has been very confusing to me.  Is this the same treatment we have on netfx and I was just spoiled when asserts were printing to the console during corefx development?

Issue #7679, opened 2016-10-17
Test JIT/Regression/VS-ia64-JIT/V2.0-RTM/b539509/b539509/b539509.sh fails intermittently


alexperovich:
The test JIT/Regression/VS-ia64-JIT/V2.0-RTM/b539509/b539509/b539509.sh fails occasionally on CI on the Linux ARM emulator.
Run here: https://ci.dot.net/job/dotnet_coreclr/job/master/job/arm_emulator_cross_debug_ubuntu_prtest/1033/

Console:

> 00:24:06.507 FAILED   - JIT/Regression/VS-ia64-JIT/V2.0-RTM/b539509/b539509/b539509.sh
> 00:24:06.507                BEGIN EXECUTION
> 00:24:06.508                /bindings/tmp/arm32_ci_temp/coreclr/Windows_NT.x64.Debug/Tests/coreoverlay/corerun b539509.exe
> 00:24:06.508                qemu: uncaught target signal 6 (Aborted) - core dumped
> 00:24:06.508                ./b539509.sh: line 117: 51783 Aborted                 (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" b539509.exe $CLRTestExecutionArguments
> 00:24:06.509                Expected: 100
> 00:24:06.509                Actual: 134
> 00:24:06.510                END EXECUTION - FAILED

sjsinju:
I think the same issue as #6298

AndyAyersMS:
Seeing this happen more often lately, eg here: https://ci.dot.net/job/dotnet_coreclr/job/master/job/armel_cross_checked_tizen_prtest/86/console

sandreenko:
Can we disable this test before a fix?

RussKeldorph:
I also believe this is probably a Duplicate of #6298.

PR #10652, opened 2017-04-03
Debug.Assert FF

danmosemsft:
Fix https://github.com/dotnet/coreclr/issues/10647
Fix https://github.com/dotnet/corefx/issues/3708

@jkotas

### DIFF
--- a/src/mscorlib/src/System/Diagnostics/Debug.Unix.cs
+++ b/src/mscorlib/src/System/Diagnostics/Debug.Unix.cs
@@ -8,11 +8,7 @@ namespace System.Diagnostics
 {
     public static partial class Debug
     {
-        private static string NewLine => "\n"; 
-
-        private const string EnvVar_DebugWriteToStdErr = "COMPlus_DebugWriteToStdErr";
-        private static readonly bool s_shouldWriteToStdErr = 
-            Internal.Runtime.Augments.EnvironmentAugments.GetEnvironmentVariable(EnvVar_DebugWriteToStdErr) == "1";
+        private static readonly bool s_shouldWriteToStdErr = Environment.GetEnvironmentVariable("COMPlus_DebugWriteToStdErr") == "1";
 
         private static void ShowAssertDialog(string stackTrace, string message, string detailMessage)
         {
@@ -22,16 +18,11 @@ namespace System.Diagnostics
             }
             else
             {
-                // TODO: #3708 Determine if/how to put up a dialog instead.
-                var exc = new DebugAssertException(message, detailMessage, stackTrace);
-                if (!s_shouldWriteToStdErr) 
-                {
-                    // We always want to print out Debug.Assert failures to stderr, even if
-                    // !s_shouldWriteToStdErr, so if it wouldn't have been printed in
-                    // WriteCore (only when s_shouldWriteToStdErr), print it here.
-                    WriteToStderr(exc.Message);
-                }
-                throw exc;
+                // In Core, we do not show a dialog.
+                // Fail in order to avoid anyone catching an exception and masking
+                // an assert failure.
+                var ex = new DebugAssertException(message, detailMessage, stackTrace);
+                Environment.FailFast(ex.Message, ex);
             }
         }
 
--- a/src/mscorlib/src/System/Diagnostics/Debug.Windows.cs
+++ b/src/mscorlib/src/System/Diagnostics/Debug.Windows.cs
@@ -6,8 +6,6 @@ namespace System.Diagnostics
 {
     public static partial class Debug
     {
-        private static string NewLine => "\r\n";
-
         private static void ShowAssertDialog(string stackTrace, string message, string detailMessage)
         {
             if (Debugger.IsAttached)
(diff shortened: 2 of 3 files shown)


danmosemsft:
I don't understand this, do we expect these asserts to work in retail?
```
// Do not remove this, it is needed to retain calls to these conditional methods in release builds
#define DEBUG
```

Shouldn't all of `Debug` either have `[System.Diagnostics.Conditional("DEBUG")]` (if public) or be in `#if DEBUG` (if not)

jkotas:
> I don't understand this, do we expect these asserts to work in retail?

This is necessary to make the asserts work in Debug build of your program, running against Release build of the runtime. For example - if you remove this line, `Assert(bool condition)` implementation compiled into the shipping runtime bits would be no-op:

```
        [System.Diagnostics.Conditional("DEBUG")]
        public static void Assert(bool condition)
        {
             // This call is compiled in only if DEBUG is defined becase of Assert(string,string,string) has annotated with ConditionalAttribute
             // Assert(condition, string.Empty, string.Empty);
        }
```

src/mscorlib/src/System/Diagnostics/Debug.Unix.cs
```diff
@@ -22,16 +18,11 @@ private static void ShowAssertDialog(string stackTrace, string message, string d
...
-                    // We always want to print out Debug.Assert failures to stderr, even if
-                    // !s_shouldWriteToStdErr, so if it wouldn't have been printed in
-                    // WriteCore (only when s_shouldWriteToStdErr), print it here.
-                    WriteToStderr(exc.Message);
-                }
-                throw exc;
+                // In Core, we do not show a dialog.
+                // Fail in order to avoid anyone catching an exception and masking
+                // an assert failure.
+                var ex = new DebugAssertException(message, detailMessage, stackTrace);
```
jkotas: What are we getting out of the DebugAssertException exception? Why not just format the message?

danmosemsft: I wasn't sure whether tihs gets us better bucketing? per
```
        // This overload of FailFast will allow you to specify the exception object
        // whose bucket details *could* be used when undergoing the failfast process.
```

jkotas: @leculver Do you think a dummy DebugAssertException helps bucketing of failfasts triggered by assertion failures?

leculver: Generally speaking, yes that should make it obvious that it's a System.Diagnostics.Debug.Assert which ended up triggering the failfast (as opposed to just a different kind of failfast).  Without the exception, I'm not sure the message is surfaced in a meaningful way in watson (outside of actually debugging the dump itself).

The big difference is that a lot of frames are 'ignored' when processing stack frames for watson, so a callstack like: "user1 calls user2 calls user3 calls framework1 which asserts", you might get a bucket that looks like 'FailFast_User3" because we skipped 'framework1'.  Instead we'd look like 'DebugAsssertException_User3' which makes it a bit more clear it was an assert instead of raw failfast.

In any case, if anyone wants to see the real bucketing behavior, simply generate a full crash dump with the new code and a crash dump using only a string (no exception), then run !analyze -v' in windbg.  The failure buckets should be pretty different.

(Though to be clear, I'm fine with either behavior.  I do think the dummy DebugAssertException will add clarity though.)

danmosemsft: @jkotas LMK whether you signoff

danmosemsft:
@dotnet-bot test tizen armel cross debug build (https://github.com/dotnet/coreclr/issues/7679)

src/mscorlib/src/System/Diagnostics/Debug.Unix.cs
```diff
@@ -22,16 +18,11 @@ private static void ShowAssertDialog(string stackTrace, string message, string d
...
             else
             {
-                // TODO: #3708 Determine if/how to put up a dialog instead.
-                var exc = new DebugAssertException(message, detailMessage, stackTrace);
-                if (!s_shouldWriteToStdErr) 
-                {
-                    // We always want to print out Debug.Assert failures to stderr, even if
-                    // !s_shouldWriteToStdErr, so if it wouldn't have been printed in
-                    // WriteCore (only when s_shouldWriteToStdErr), print it here.
-                    WriteToStderr(exc.Message);
```
bartonjs: Why'd we stop printing the message? If Environment.FailFast is supposed to, it doesn't seem to be...

danmosemsft: As you see if COMPlus_DebugWriteToStdErr=1 it would go to stderr. I don't know why we were always writing it here, and I don't know why I removed it. We could put it back. Does it go to eg CI logs, and the FF message doesn't?

bartonjs: https://github.com/dotnet/corefx/issues/18603#issuecomment-295345117 shows what was presumed to be an assert that was tripping, and nothing made it into the CI logs.

jkotas: I think Environment.FailFast should print the message to the console. I cannot think about any good reason for not printing it.

PR #10653, opened 2017-04-03
Cut down FormatterServices

danmosemsft:
Contributes to https://github.com/dotnet/corefx/issues/16229

@jkotas


jkotas:
cc @russellhadley @erozenfeld If you are ever going to run ILLink on the shipping CoreLib, we may want to keep this method rooted.

danmosemsft:
Is the a way (like an rd.xml) we can mark it to root it?

PR #10662, opened 2017-04-03
Add JIT intrinsics support for vector conversion on AMD64 and x86

helloguo:
This PR adds JIT intrinsic support for vector conversion/narrow/widen on AMD64 SSE2, SSE34 and AVX. It is built on https://github.com/dotnet/coreclr/pull/9920 and https://github.com/dotnet/coreclr/pull/9318. The intrinsics will be tested by https://github.com/dotnet/coreclr/pull/10467.

The following APIs are accelerated by JIT intrinsics provided in this PR:

```
public static partial class Vector
{
    public static void Widen(Vector<byte> source, out Vector<ushort> dest1, out Vector<ushort> dest2);
    public static void Widen(Vector<ushort> source, out Vector<uint> dest1, out Vector<uint> dest2);
    public static void Widen(Vector<uint> source, out Vector<ulong> dest1, out Vector<ulong> dest2);
    public static void Widen(Vector<sbyte> source, out Vector<short> dest1, out Vector<short> dest2);
    public static void Widen(Vector<short> source, out Vector<int> dest1, out Vector<int> dest2);
    public static void Widen(Vector<int> source, out Vector<long> dest1, out Vector<long> dest2);
    public static void Widen(Vector<float> source, out Vector<double> dest1, out Vector<double> dest2);

    public static Vector<byte> Narrow(Vector<ushort> source1, Vector<ushort> source2);
    public static Vector<ushort> Narrow(Vector<uint> source1, Vector<uint> source2);
    public static Vector<uint> Narrow(Vector<ulong> source1, Vector<ulong> source2);
    public static Vector<sbyte> Narrow(Vector<short> source1, Vector<short> source2);
    public static Vector<short> Narrow(Vector<int> source1, Vector<int> source2);
    public static Vector<int> Narrow(Vector<long> source1, Vector<long> source2);
    public static Vector<float> Narrow(Vector<double> source1, Vector<double> source2);

    public static Vector<float> ConvertToSingle(Vector<int> value);
    public static Vector<float> ConvertToSingle(Vector<uint> value);
    public static Vector<double> ConvertToDouble(Vector<long> value);
    public static Vector<double> ConvertToDouble(Vector<ulong> value);
    public static Vector<int> ConvertToInt32(Vector<float> value);
    public static Vector<uint> ConvertToUInt32(Vector<float> value);
    public static Vector<long> ConvertToInt64(Vector<double> value);
    public static Vector<ulong> ConvertToUInt64(Vector<double> value);
}
```

The semantics of above APIs can be found at https://github.com/dotnet/corefx/issues/15957. The C# implementation of above APIs can be found at https://github.com/dotnet/corefx/pull/16276.

Fix https://github.com/dotnet/coreclr/issues/9317.

### DIFF
--- a/src/jit/emitxarch.cpp
+++ b/src/jit/emitxarch.cpp
@@ -94,7 +94,10 @@ bool emitter::IsThreeOperandBinaryAVXInstruction(instruction ins)
             ins == INS_vinsertf128 || ins == INS_punpckldq || ins == INS_phaddd || ins == INS_pminub ||
             ins == INS_pminsw || ins == INS_pminsb || ins == INS_pminsd || ins == INS_pminuw || ins == INS_pminud ||
             ins == INS_pmaxub || ins == INS_pmaxsw || ins == INS_pmaxsb || ins == INS_pmaxsd || ins == INS_pmaxuw ||
-            ins == INS_pmaxud);
+            ins == INS_pmaxud || ins == INS_vinserti128 || ins == INS_punpckhbw || ins == INS_punpcklbw ||
+            ins == INS_punpckhqdq || ins == INS_punpcklqdq || ins == INS_punpckhwd || ins == INS_punpcklwd ||
+            ins == INS_punpckhdq || ins == INS_packssdw || ins == INS_packsswb || ins == INS_packuswb ||
+            ins == INS_packusdw || ins == INS_vperm2i128);
 }
 
 // Returns true if the AVX instruction is a move operator that requires 3 operands.
@@ -105,8 +108,8 @@ bool emitter::IsThreeOperandBinaryAVXInstruction(instruction ins)
 // to indicate whether a 3-operand instruction.
 bool emitter::IsThreeOperandMoveAVXInstruction(instruction ins)
 {
-    return IsAVXInstruction(ins) &&
-           (ins == INS_movlpd || ins == INS_movlps || ins == INS_movhpd || ins == INS_movhps || ins == INS_movss);
+    return IsAVXInstruction(ins) && (ins == INS_movlpd || ins == INS_movlps || ins == INS_movhpd || ins == INS_movhps ||
+                                     ins == INS_movss || ins == INS_movlhps);
 }
 
 // ------------------------------------------------------------------------------
@@ -206,6 +209,14 @@ emitter::code_t emitter::AddVexPrefix(instruction ins, code_t code, emitAttr att
 // Returns true if this instruction, for the given EA_SIZE(attr), will require a REX.W prefix
 bool TakesRexWPrefix(instruction ins, emitAttr attr)
 {
+    // Because the current implementation of AVX does not have a way to distinguish between the register
+    // size specification (128 vs. 256 bits) and the operand size specification (32 vs. 64 bits), where both are
+    // required, the instruction must be created with the register size attribute (EA_16BYTE or EA_32BYTE),
+    // and here we must special case these by the opcode.
+    if (ins == INS_vpermq)
+    {
+        return true;
+    }
 #ifdef _TARGET_AMD64_
     // movsx should always sign extend out to 8 bytes just because we don't track
     // whether the dest should be 4 bytes or 8 bytes (attr indicates the size
@@ -342,7 +353,6 @@ unsigned RegEncoding(regNumber reg)
 // AVX:  specific bits within VEX prefix need to be set in bit-inverted form.
 emitter::code_t emitter::AddRexWPrefix(instruction ins, code_t code)
 {
-#ifdef _TARGET_AMD64_
     if (UseAVX() && IsAVXInstruction(ins))
     {
         // W-bit is available only in 3-byte VEX prefix that starts with byte C4.
@@ -351,7 +361,7 @@ emitter::code_t emitter::AddRexWPrefix(instruction ins, code_t code)
         // W-bit is the only bit that is added in non bit-inverted form.
         return code | 0x00008000000000ULL;
     }
-
+#ifdef _TARGET_AMD64_
     return code | 0x4800000000ULL;
 #else
     assert(!"UNREACHED");
@@ -3810,6 +3820,40 @@ void emitter::emitIns_R_R_R(instruction ins, emitAttr attr, regNumber targetReg,
     emitCurIGsize += sz;
 }
 
+/**********************************************************************************
+* emitIns_R_R_R_I: Add an instruction with three register operands and an immediate.
+*
+* Arguments:
+*    ins       - the instruction to add
+*    attr      - the emitter attribute for instruction
+*    targetReg - the target (destination) register
+*    reg1      - the first source register
+*    reg2      - the second source register
+*    ival      - the immediate value
+*/
+
+void emitter::emitIns_R_R_R_I(
+    instruction ins, emitAttr attr, regNumber targetReg, regNumber reg1, regNumber reg2, int ival)
+{
+    assert(IsSSEOrAVXInstruction(ins));
+    assert(IsThreeOperandAVXInstruction(ins));
+    // Currently vex prefix only use three bytes mode.
+    // size = vex + opcode + ModR/M + 1-byte-cns = 3 + 1 + 1 + 1 = 6
+    // TODO-XArch-CQ: We should create function which can calculate all kinds of AVX instructions size in future
+    UNATIVE_OFFSET sz = 6;
+
+    instrDesc* id = emitNewInstrCns(attr, ival);
+    id->idIns(ins);
+    id->idInsFmt(IF_RWR_RRD_RRD_CNS);
+    id->idReg1(targetReg);
+    id->idReg2(reg1);
+    id->idReg3(reg2);
+
+    id->idCodeSize(sz);
+    dispIns(id);
+    emitCurIGsize += sz;
+}
+
 #endif
 /*****************************************************************************
  *
@@ -6995,6 +7039,15 @@ void emitter::emitDispIns(
             printf("%s, ", emitRegName(id->idReg2(), attr));
             printf("%s", emitRegName(id->idReg3(), attr));
             break;
+        case IF_RWR_RRD_RRD_CNS:
+            assert(IsAVXInstruction(ins));
+            assert(IsThreeOperandAVXInstruction(ins));
+            printf("%s, ", emitRegName(id->idReg1(), attr));
+            printf("%s, ", emitRegName(id->idReg2(), attr));
+            printf("%s, ", emitRegName(id->idReg3(), attr));
+            val = emitGetInsSC(id);
+            goto PRINT_CONSTANT;
+            break;
 #endif
         case IF_RRW_RRW_CNS:
             printf("%s,", emitRegName(id->idReg1(), attr));
@@ -9514,7 +9567,34 @@ BYTE* emitter::emitOutputRI(BYTE* dst, instrDesc* id)
 
         assert(id->idGCref() == GCT_NONE);
         assert(valInByte);
-        assert(ins == INS_psrldq || ins == INS_pslldq);
+        // The left and right shifts use the same encoding, and are distinguished by the Reg/Opcode field.
+        regNumber regOpcode;
+        switch (ins)
+        {
+            case INS_psrldq:
+                regOpcode = (regNumber)3;
+                break;
+            case INS_pslldq:
+                regOpcode = (regNumber)7;
+                break;
+            case INS_psrld:
+            case INS_psrlw:
+            case INS_psrlq:
+                regOpcode = (regNumber)2;
+                break;
+            case INS_pslld:
+            case INS_psllw:
+            case INS_psllq:
+                regOpcode = (regNumber)6;
+                break;
+            case INS_psrad:
+                regOpcode = (regNumber)4;
+                break;
+            default:
+                assert(!"Invalid instruction for SSE2 instruction of the form: opcode reg, immed8");
+                regOpcode = REG_NA;
+                break;
+        }
 
         // Get the 'base' opcode.
         code = insCodeMI(ins);
@@ -9528,14 +9608,6 @@ BYTE* emitter::emitOutputRI(BYTE* dst, instrDesc* id)
             code = insEncodeReg3456(ins, reg, size, code);
         }
 
-        // In case of psrldq
-        // Reg/Opcode = 3
-        // R/M = reg1
-        //
-        // In case of pslldq
-        // Reg/Opcode = 7
-        // R/M = reg1
-        regNumber regOpcode = (regNumber)((ins == INS_psrldq) ? 3 : 7);
         unsigned regcode = (insEncodeReg345(ins, regOpcode, size, &code) | insEncodeReg012(ins, reg, size, &code)) << 8;
 
         // Output the REX prefix
@@ -10659,6 +10731,11 @@ size_t emitter::emitOutputInstr(insGroup* ig, instrDesc* id, BYTE** dp)
             dst = emitOutputRRR(dst, id);
             sz  = emitSizeOfInsDsc(id);
             break;
+        case IF_RWR_RRD_RRD_CNS:
+            dst = emitOutputRRR(dst, id);
+            sz  = emitSizeOfInsDsc(id);
+            dst += emitOutputByte(dst, emitGetInsSC(id));
+            break;
 #endif
 
         case IF_RRW_RRW_CNS:
@@ -10690,6 +10767,11 @@ size_t emitter::emitOutputInstr(insGroup* ig, instrDesc* id, BYTE** dp)
             assert(code & 0x00FF0000);
 
 #ifdef FEATURE_AVX_SUPPORT
+            if (TakesRexWPrefix(ins, size))
+            {
+                code = AddRexWPrefix(ins, code);
+            }
+
             if (TakesVexPrefix(ins))
             {
                 if (IsThreeOperandBinaryAVXInstruction(ins))
@@ -10718,11 +10800,16 @@ size_t emitter::emitOutputInstr(insGroup* ig, instrDesc* id, BYTE** dp)
 
             if (Is4ByteAVXInstruction(ins))
             {
-                // We just need to output the last byte of the opcode.
                 assert((code & 0xFF) == 0);
-                assert((code & 0xFF00) != 0xC000);
-                dst += emitOutputByte(dst, (code >> 8) & 0xFF);
-                code = 0;
+                if ((code & 0xFF00) == 0xC000)
+                {
+                    dst += emitOutputWord(dst, code | regcode);
+                }
+                else
+                {
+                    dst += emitOutputByte(dst, (code >> 8) & 0xFF);
+                    dst += emitOutputByte(dst, 0xC0 | (regcode >> 8));
+                }
             }
             else if (code & 0xFF000000)
             {
@@ -10732,27 +10819,25 @@ size_t emitter::emitOutputInstr(insGroup* ig, instrDesc* id, BYTE** dp)
                 if (Is4ByteSSE4Instruction(ins))
                 {
                     dst += emitOutputWord(dst, code);
-                    code = 0;
+                    dst += emitOutputByte(dst, 0xC0 | (regcode >> 8));
+                }
+                else
+                {
+                    assert((code & 0xFF00) == 0xC000);
+                    dst += emitOutputWord(dst, code | regcode);
                 }
             }
             else if (code & 0x00FF0000)
             {
                 dst += emitOutputByte(dst, code >> 16);
                 code &= 0x0000FFFF;
-            }
-
-            // Note that regcode is shifted by 8-bits above to align with RM byte.
-            if (code != 0)
-            {
                 assert((code & 0xFF00) == 0xC000);
                 dst += emitOutputWord(dst, code | regcode);
             }
             else
             {
-                // This case occurs for SSE4/AVX instructions.
-                // Note that regcode is left shifted by 8-bits.
-                assert(Is4ByteAVXInstruction(ins) || Is4ByteSSE4Instruction(ins));
-                dst += emitOutputByte(dst, 0xC0 | (regcode >> 8));
+                assert((code & 0xFF00) == 0xC000);
+                dst += emitOutputWord(dst, code | regcode);
             }
 
             dst += emitOutputByte(dst, emitGetInsSC(id));
--- a/src/jit/lsraxarch.cpp
+++ b/src/jit/lsraxarch.cpp
@@ -2676,6 +2676,90 @@ void Lowering::TreeNodeInfoInitSIMD(GenTree* tree)
             info->srcCount = 1;
             break;
 
+        case SIMDIntrinsicConvertToSingle:
+            info->srcCount = 1;
+            if (simdTree->gtSIMDBaseType == TYP_UINT)
+            {
+                // We need an internal register different from targetReg.
+                info->isInternalRegDelayFree = true;
+                info->internalIntCount       = 1;
+                info->internalFloatCount     = 2;
+                info->setInternalCandidates(lsra, lsra->allSIMDRegs() | lsra->allRegs(TYP_INT));
+            }
+            break;
+
+        case SIMDIntrinsicConvertToUInt32:
+        case SIMDIntrinsicConvertToInt32:
+            info->srcCount = 1;
+            break;
+
+        case SIMDIntrinsicWidenLo:
+        case SIMDIntrinsicWidenHi:
+            info->srcCount = 1;
+            if (varTypeIsIntegral(simdTree->gtSIMDBaseType))
+            {
+                // We need an internal register different from targetReg.
+                info->isInternalRegDelayFree = true;
+                info->internalFloatCount     = 1;
+                info->setInternalCandidates(lsra, lsra->allSIMDRegs());
+            }
+            break;
+
+        case SIMDIntrinsicConvertToInt64:
+        case SIMDIntrinsicConvertToUInt64:
+            // We need an internal register different from targetReg.
+            info->isInternalRegDelayFree = true;
+            info->srcCount               = 1;
+            info->internalIntCount       = 1;
+            if (comp->getSIMDInstructionSet() == InstructionSet_AVX)
+            {
+                info->internalFloatCount = 2;
+            }
+            else
+            {
+                info->internalFloatCount = 1;
+            }
+            info->setInternalCandidates(lsra, lsra->allSIMDRegs() | lsra->allRegs(TYP_INT));
+            break;
+
+        case SIMDIntrinsicConvertToDouble:
+            // We need an internal register different from targetReg.
+            info->isInternalRegDelayFree = true;
+            info->srcCount               = 1;
+            info->internalIntCount       = 1;
+#ifdef _TARGET_X86_
+            if (simdTree->gtSIMDBaseType == TYP_LONG)
+            {
+                info->internalFloatCount = 3;
+            }
+            else
+#endif
+                if ((comp->getSIMDInstructionSet() == InstructionSet_AVX) || (simdTree->gtSIMDBaseType == TYP_ULONG))
+            {
+                info->internalFloatCount = 2;
+            }
+            else
+            {
+                info->internalFloatCount = 1;
+            }
+            info->setInternalCandidates(lsra, lsra->allSIMDRegs() | lsra->allRegs(TYP_INT));
+            break;
+
+        case SIMDIntrinsicNarrow:
+            // We need an internal register different from targetReg.
+            info->isInternalRegDelayFree = true;
+            info->srcCount               = 2;
+            if ((comp->getSIMDInstructionSet() == InstructionSet_AVX) && (simdTree->gtSIMDBaseType != TYP_DOUBLE))
+            {
+                info->internalFloatCount = 2;
+            }
+            else
+            {
+                info->internalFloatCount = 1;
+            }
+            info->setInternalCandidates(lsra, lsra->allSIMDRegs());
+            break;
+
         case SIMDIntrinsicShuffleSSE2:
             info->srcCount = 2;
             // Second operand is an integer constant and marked as contained.
--- a/src/jit/simd.cpp
+++ b/src/jit/simd.cpp
@@ -2609,11 +2609,70 @@ GenTreePtr Compiler::impSIMDIntrinsic(OPCODE                opcode,
 
         // Unary operators that take and return a Vector.
         case SIMDIntrinsicCast:
+        case SIMDIntrinsicConvertToSingle:
+        case SIMDIntrinsicConvertToDouble:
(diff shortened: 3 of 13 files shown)


helloguo:
@BruceForstall @russellhadley @CarolEidt PTAL

helloguo:

src/jit/emitxarch.cpp
```diff
@@ -214,7 +217,8 @@ bool TakesRexWPrefix(instruction ins, emitAttr attr)
     // cased here.
     //
     // Rex_jmp = jmp with rex prefix always requires rex.w prefix.
-    if (ins == INS_movsx || ins == INS_rex_jmp)
+    // vpermq requires w bit to set to 1
+    if (ins == INS_movsx || ins == INS_rex_jmp || ins == INS_vpermq)
     {
```
BruceForstall: I don't think this is correct. vpermq is encoded with a VEX prefix (with w set to 1), not with a REX.W prefix. And this function is only asking about REX.W.

src/jit/emitxarch.cpp
```diff
@@ -3786,7 +3790,7 @@ void emitter::emitIns_R_R_I(instruction ins, emitAttr attr, regNumber reg1, regN
     instrDesc* id = emitNewInstrSC(attr, ival);
 
     // REX prefix
-    if (IsExtendedReg(reg1, attr) || IsExtendedReg(reg2, attr))
+    if (IsExtendedReg(reg1, attr) || IsExtendedReg(reg2, attr) || TakesRexWPrefix(ins, attr))
```
BruceForstall: >|| TakesRexWPrefix(ins, attr) [](start = 63, length = 29)

Did you add this because of vpermq? Why is that necessary? It wouldn't work on x86.

src/jit/emitxarch.cpp
```diff
@@ -3827,6 +3831,39 @@ void emitter::emitIns_R_R_R(instruction ins, emitAttr attr, regNumber targetReg,
     emitCurIGsize += sz;
 }
 
+/**********************************************************************************
+* emitIns_R_R_R_I: Add an instruction with three register operands and an immediate.
+*
+* Arguments:
+*    ins - the instruction to add
+*    targetReg - the target (destination) register
+*    reg1      - the first source register
```
BruceForstall: Please include all arguments, including `attr`

src/jit/emitxarch.cpp
```diff
@@ -10793,6 +10863,11 @@ size_t emitter::emitOutputInstr(insGroup* ig, instrDesc* id, BYTE** dp)
             assert(code & 0x00FF0000);
 
 #ifdef FEATURE_AVX_SUPPORT
+            if (TakesRexWPrefix(ins, size))
+            {
+                code = AddRexWPrefix(ins, code);
+            }
+
             if (TakesVexPrefix(ins))
```
BruceForstall: Why was this change necessary? It doesn't look like you added support for any instructions that take a REX prefix.

src/jit/simd.cpp
```diff
@@ -2617,6 +2617,75 @@ GenTreePtr Compiler::impSIMDIntrinsic(OPCODE                opcode,
         }
         break;
 
+        case SIMDIntrinsicConvertToSingle:
+        case SIMDIntrinsicConvertToDouble:
+        case SIMDIntrinsicConvertToInt32:
+        case SIMDIntrinsicConvertToUInt32:
+        case SIMDIntrinsicConvertToInt64:
+        case SIMDIntrinsicConvertToUInt64:
+        {
```
BruceForstall: These look the same as SIMDIntrinsicCast. Why not share that code, above?

src/jit/simd.cpp
```diff
@@ -2617,6 +2617,75 @@ GenTreePtr Compiler::impSIMDIntrinsic(OPCODE                opcode,
...
+            JITDUMP("SIMD Conversion is not supported on this platform\n");
+            return nullptr;
+#endif
+        }
+        break;
+
+        case SIMDIntrinsicNarrow:
+        {
+#ifdef _TARGET_AMD64_
+            assert(!instMethod);
```
BruceForstall: Why is this (and the other ones you've added) under `_TARGET_AMD64_`? I would remove the `#ifdef` and make it all platform (namely, x64 and x86, which are the only ones defining `FEATURE_SIMD` today).

src/jit/simd.cpp
```diff
@@ -2617,6 +2617,75 @@ GenTreePtr Compiler::impSIMDIntrinsic(OPCODE                opcode,
...
+        case SIMDIntrinsicWiden:
+        {
+#ifdef _TARGET_AMD64_
+            GenTree* dstAddrHi = impSIMDPopStack(TYP_BYREF);
+            GenTree* dstAddrLo = impSIMDPopStack(TYP_BYREF);
+            op1                = impSIMDPopStack(simdType);
+            GenTree* dupOp1    = fgInsertCommaFormTemp(&op1, gtGetStructHandleForSIMD(simdType, baseType));
+
+            // Widen the lower half and assign it to dstAddrLo.
+            simdTree = gtNewSIMDNode(simdType, op1, nullptr, SIMDIntrinsicWidenLo, baseType, size);
```
BruceForstall: >op1 [](start = 47, length = 3)

Should this refer to dupOp1 instead of op1?

src/jit/simd.cpp
```diff
@@ -2617,6 +2617,75 @@ GenTreePtr Compiler::impSIMDIntrinsic(OPCODE                opcode,
...
+            simdTree = gtNewSIMDNode(simdType, dupOp1, nullptr, SIMDIntrinsicWidenHi, baseType, size);
+            GenTree* hiDest =
+                new (this, GT_BLK) GenTreeBlk(GT_BLK, simdType, dstAddrHi, getSIMDTypeSizeInBytes(clsHnd));
+            GenTree* hiAsg = gtNewBlkOpNode(hiDest, simdTree, getSIMDTypeSizeInBytes(clsHnd),
+                                            false, // not volatile
+                                            true); // copyBlock
+            hiAsg->gtFlags |= ((simdTree->gtFlags | dstAddrHi->gtFlags) & GTF_ALL_EFFECT);
+
+            retVal = gtNewOperNode(GT_COMMA, simdType, loAsg, hiAsg);
+#else
```
BruceForstall: So it looks like we never create a SIMDIntrinsicWiden tree node, but instead "decompose" it into a high and low part that we put into a `GT_COMMA` like:
```
  GT_COMMA( *dstHi = SIMDIntrinsicWidenHi(op1) , *dstLo = SIMDIntrinsicWidenLow(op1) )
```
This seems a bit weird to me, as we normally use commas to create temps in the tree, e.g.:
```
  GT_COMMA( tmp = ... , tmp )
```
where the "result" of the comma is the `tmp` on the right-hand-side (second operand) of the `GT_COMMA`. That would imply the "value" represented by your tree is the `*dstLo` assignment, and the left side is "just" side-effect.

I wonder if you should instead create some kind of "merge" node instead, e.g.:
```
  SIMDIntrinsicWiden( *dstHi = SIMDIntrinsicWidenHi(op1) , *dstLo = SIMDIntrinsicWidenLow(op1) )
```
where the SIMDIntrinsicWiden codegen would be a nop, since the operands would do the work for it.

Comments?

src/jit/lsraxarch.cpp
```diff
@@ -2676,6 +2676,68 @@ void Lowering::TreeNodeInfoInitSIMD(GenTree* tree)
...
+
+        case SIMDIntrinsicConvertToInt64:
+        case SIMDIntrinsicConvertToUInt64:
+        case SIMDIntrinsicConvertToDouble:
+            // We need an internal register different from targetReg.
+            info->isInternalRegDelayFree = true;
+            info->srcCount               = 1;
+            info->internalIntCount       = 1;
+            if (comp->getSIMDInstructionSet() == InstructionSet_AVX || simdTree->gtSIMDBaseType == TYP_ULONG)
+            {
```
BruceForstall: coding style: please parenthesize sub-expressions (here and below)

src/jit/simdcodegenxarch.cpp
```diff
@@ -951,6 +1088,570 @@ void CodeGen::genSIMDIntrinsicUnOp(GenTreeSIMD* simdNode)
...
+    {
+        inst_RV_RV(ins, targetReg, op1Reg, targetType, emitActualTypeSize(targetType));
+    }
+    genProduceReg(simdNode);
+}
+
+//----------------------------------------------------------------------------------
+// genSIMDLo64BitConvert: "Generate code to convert lower-most 64-bit item (long <--> double)
+//
+// Arguments:
```
BruceForstall: Please fill this in.

nit: extra double quote in title line

BruceForstall:
Why are these not enabled for x86?

src/jit/simdcodegenxarch.cpp
```diff
@@ -951,6 +1088,570 @@ void CodeGen::genSIMDIntrinsicUnOp(GenTreeSIMD* simdNode)
...
+        assert(tmpRegsMask != RBM_NONE);
+        assert(genCountBits(tmpRegsMask) == 2);
+
+        regMaskTP tmpReg1Mask = genFindLowestBit(tmpRegsMask);
+        tmpRegsMask &= ~tmpReg1Mask;
+        regNumber tmpReg  = genRegNumFromMask(tmpReg1Mask);
+        regNumber tmpReg2 = genRegNumFromMask(tmpRegsMask);
+        assert(tmpReg != op1Reg && tmpReg2 != op1Reg);
+
+        // tmpReg:      mask, which contains multiple 0x53000000
```
BruceForstall: >which contains multiple 0x53000000 [](start = 30, length = 34)

What does "which contains multiple 0x53000000" mean?

BruceForstall: Maybe it would be helpful to have a comment showing a sample code sequence that you intend to generate.

---
In reply to: [109797093](https://github.com/dotnet/coreclr/pull/10662#discussion_r109797093) [](ancestors = 109797093)

src/jit/simdcodegenxarch.cpp
```diff
@@ -951,6 +1088,570 @@ void CodeGen::genSIMDIntrinsicUnOp(GenTreeSIMD* simdNode)
...
+
+//----------------------------------------------------------------------------------
+// genSIMDIntrinsic64BitConvert: Generate code for 64-bit SIMD Convert (long/ulong <-> double)
+//
+// Arguments:
+//    simdNode - The GT_SIMD node
+//
+// Notes:
+//    There are no instructions for converting to/from 64-bit integers, so for these we
+//    do the convertion an element at a time.
```
BruceForstall: >convertion [](start = 13, length = 10)

typo: convertion

src/jit/emitxarch.cpp
helloguo: Thanks for pointing it out. Yes, vpermq needs VEX w1 bit. How about creating two functions named TakesVexWPrefix() and AddVexWPrefix() to handle VEX?

src/jit/emitxarch.cpp
helloguo: Yes, it's because vpermq. Will find a better way to handle vpermq.

src/jit/emitxarch.cpp
helloguo: This is also because of vpermq. Will address vpermq.

src/jit/simd.cpp
helloguo: Sounds good. Will make the change.

src/jit/simd.cpp
helloguo: This PR does x64 first. It's a good point to make it all platform. We may create an issue to follow it up.

src/jit/simdcodegenxarch.cpp
helloguo: Sure. Will add comment.

helloguo:
@BruceForstall Thanks for your feedback. Will make the change.

BruceForstall:
@helloguo You should rewrite the VectorConvert.cs tests to use the JitLog test infrastructure that ensures that the given functions are accelerated, and make sure every intrinsic you've created is accelerated. Check VectorAdd.cs for example. You might also be able to use CheckValue<> instead of your own value comparisons.

src/jit/simd.cpp
BruceForstall: I think you need to make it all platform before this change is merged. We don't want to advance the two platforms independently; we want them to support the same functionality. Plus, you need to change the VectorConvert tests to check for actually generating accelerated code, and we don't want per-architecture tests.

src/jit/emitxarch.cpp
BruceForstall: We already have `TakesVexPrefix()` and `AddVexPrefix()`. Maybe what you want is a function that returns true if the VEX prefix W bit needs to be set (I assume nobody required this until now). Perhaps that should be called `RequiresVexPrefixWBit(instruction ins)`? Then, `AddVexPrefix()` could have code like:
```
if (RequiresVexPrefixWBit(ins))
{
   code |= ... <magic W bit mask> ...
}
```

src/jit/simd.cpp
helloguo: Tried to do something like:
```
SIMDIntrinsicWiden( *dstHi = SIMDIntrinsicWidenHi(op1) , *dstLo = SIMDIntrinsicWidenLow(op1) )
```

But it's not that straightforward.  I define a new node named `SIMDIntrinsicWidenRet`, which takes `loAsg `and `hiAsg ` as inputs. However, before it gets lowered and emitted, assertion happens at lir.cpp line 1503 https://github.com/dotnet/coreclr/blob/master/src/jit/lir.cpp#L1503.  I guess because the root node is `SIMDIntrinsicWidenRet`, it checks the IR from root to leaf. Can you give me any suggestions?

Since both `loAsg `and `hiAsg` are assignment already, it may be reasonable to put comma there?

BruceForstall: ok, I withdraw my suggestion. Go ahead and stick with your original `GT_COMMA` implementation.

russellhadley:
@BruceForstall, @helloguo Just looping back to this.  Are we in shape to get the AMD64 changes merged?  Has the follow up item for the X86 work been opened?

BruceForstall:
@russellhadley I haven't re-reviewed. I don't want it merged until x86 is fully supported equal to amd64.

russellhadley:
Why do we need to wait?  Doing it in two steps doesn't seem problematic to me.

BruceForstall:
@russellhadley Why do we need to rush it and not do it right at the first merge?

russellhadley:
@BruceForstall I don't see it as a rush, just more bake time and checking in early and often.  I'm not getting how the big bang is more "right" than in stages.

BruceForstall:
@russellhadley As a matter of principle, x86 and x64 have feature parity, including w.r.t. SIMD intrinsics. We don't want to regress that, even temporarily. (I can't be assured that any resources devoted to adding intrinsics don't disappear after merging 1/2 of the work.) As a practical matter, the VectorConvert test here needs to be rewritten to use JitLog to verify that intrinsics were used. We don't want to disable this test for x86 -- we want all tests running on all architectures.

russellhadley:
@BruceForstall There will be feature parity, just one will be slower than the other (i.e. X86 not use the intrinsics initially).  In terms of resource allocation I don't think PRs are the appropriate place for that discussion.  From my side I'm willing to approve the merge if the amd64 changes if they are correct and meet the design direction.

src/jit/simd.cpp
CarolEidt: @brucefo - I don't think so; I believe that `fgInsertCommaFormTemp` will update `op1` to point to the comma expression, and return a new node to reference the temp.

---
In reply to: [109793048](https://github.com/dotnet/coreclr/pull/10662#discussion_r109793048) [](ancestors = 109793048)

src/jit/simdcodegenxarch.cpp
```diff
@@ -951,6 +1088,600 @@ void CodeGen::genSIMDIntrinsicUnOp(GenTreeSIMD* simdNode)
...
+        }
+        inst_RV_RV(widenIns, targetReg, srcReg, simdType);
+    }
+    else
+    {
+        // targetReg = op1Reg (permute op1Reg when AVX)
+        // tmpReg = 0
+        // tmpReg = (0 > targetReg)           // (If signed) Get the sign bits
+        // punpck[l|h]dq  targetReg, tmpReg   // Interleave the sign bits
+        regNumber tmpReg = genRegNumFromMask(simdNode->gtRsvdRegs & RBM_ALLFLOAT);
```
CarolEidt: This comment is not correct - it needs to reflect the code being generated below.

src/jit/emitxarch.cpp
CarolEidt: This doesn't seem right to me. The VEX.W prefix, I believe, is generally the AVX encoding of the corresponding REX.W prefix. We do have encodings that use the VEX.W prefix, and it is set based on the opcode size. I think it may be that the vpermq encoding needs to have the correct operand size, in which case I believe that the W bit would be set appropriately.

---
In reply to: [109974462](https://github.com/dotnet/coreclr/pull/10662#discussion_r109974462) [](ancestors = 109974462)

src/jit/simdcodegenxarch.cpp
```diff
@@ -951,6 +1088,600 @@ void CodeGen::genSIMDIntrinsicUnOp(GenTreeSIMD* simdNode)
...
+
+        if (iset == InstructionSet_AVX)
+        {
+            // permute op1Reg and put it into targetReg
+            unsigned ival = 0xd4;
+            if (simdNode->gtSIMDIntrinsicID == SIMDIntrinsicWidenHi)
+            {
+                ival = 0xe8;
+            }
+            getEmitter()->emitIns_R_R_I(INS_vpermq, emitSize, targetReg, op1Reg, ival);
```
CarolEidt: I think that this is where the emitSize needs to be set correctly (EA_8BYTE?) to cause the W bit to be set correctly.

CarolEidt:
:clock1:

RussKeldorph:
@dotnet/jit-contrib

src/jit/emitxarch.cpp
helloguo: > The VEX.W prefix, I believe, is generally the AVX encoding of the corresponding REX.W prefix.

I agree. From wikipedia, "The W bit is equivalent to the REX prefix's W bit, and specifies a 64-bit operand; for non-integer instructions, it is a general opcode extension bit." But it's for 64-bit register only.

The function `AddRexWPrefix(instruction ins, code_t code)` https://github.com/dotnet/coreclr/blob/master/src/jit/emitxarch.cpp#L350, takes care of both VEX.W bit and REX.W bit. From this aspect, it's OK to let `TakesRexWPrefix` return `true ` for `vpermq` and use `AddRexWPrefix` to set VEX.W bit.

> We do have encodings that use the VEX.W prefix, and it is set based on the opcode size.

For ymm register, VEX.W bit is just a general opcode extension bit. So I think we need to set the VEX W bit explicitly for `vpermq` because `vpermq `uses ymm or xmm. 

I think we have two options here: 
        (1) reuse `TakesRexWPrefix `and `AddRexWPrefix `by letting `TakesRexWPrefix  `return `true `for `vpermq`.
        (2) define two new functions (e.g. `TakesVexWPrefix `and `AddVexWPrefix`) to handle VEX.W bit explicitly. Meanwhile, clean the code in `AddRexWPrefix ` so that `AddRexWPrefix ` only handles REX.W bit.

Any suggestions?

helloguo:
@CarolEidt @BruceForstall Just added x86 support except three conversions (double to long, long to double, double to ulong) because there are no corresponding hardware instructions. It's not straightforward to add intrinsics for these three conversions on 32 bits machines.

CarolEidt:
@helloguo - I think it is appropriate to leave the three conversions you mention as non-intrinsics; as you mention, the code will be complex. Since they will not be accelerated, it may be worth considering whether the C# implementation can utilize some of the other (supported) intrinsics. (Note that I haven't explored this possibility, so it may be that it would not be a productive approach). In any case, I think it's find to leave it as a future improvement.
However, I am still not comfortable with the addition of a special method for the VEX.W prefix, as I believe it should be handled in a more natural fashion by supplying the correct type size. Did you try that?

helloguo:
> However, I am still not comfortable with the addition of a special method for the VEX.W prefix, as I believe it should be handled in a more natural fashion by supplying the correct type size. Did you try that?

@CarolEidt If I understand correctly, you are referring that VEX.W bit provides same functionality as REX.W bit in certain scenarios. So that when we set correct type size, we can reuse `TakesRexWPrefix` and `AddRexWPrefix` to set VEX.W bit instead of having a special method. Because `TakesRexWPrefix` returns false when `attr `is not `EA_8BYTE` (https://github.com/dotnet/coreclr/blob/master/src/jit/emitxarch.cpp#L222), we need to set `attr `to `EA_8BYTE `for `vpermq ` in order to set VEX.W bit. 

However, if I just set `attr `to `EA_8BYTE `for `vpermq ` ( for example: `getEmitter()->emitIns_R_R_I(INS_vpermq, EA_8BYTE, targetReg, op1Reg, ival);`), an exception will be thrown. I'm thinking we can reuse `TakesRexWPrefix` and `AddRexWPrefix` by letting `TakesRexWPrefix` return `true `for `vpermq ` like following.  Any suggestions? 
```
bool TakesRexWPrefix(instruction ins, emitAttr attr)
{
#ifdef _TARGET_AMD64_
    // movsx should always sign extend out to 8 bytes just because we don't track
    // whether the dest should be 4 bytes or 8 bytes (attr indicates the size
    // of the source, not the dest).
    // A 4-byte movzx is equivalent to an 8 byte movzx, so it is not special
    // cased here.
    //
    // Rex_jmp = jmp with rex prefix always requires rex.w prefix.
    if (ins == INS_movsx || ins == INS_rex_jmp || ins == INS_vpermq)
    {
        return true;
    }

    if (EA_SIZE(attr) != EA_8BYTE)
    {
        return false;
    }
```

helloguo:

BruceForstall:
@helloguo Are you planning to address my previous request:

> You should rewrite the VectorConvert.cs tests to use the JitLog test infrastructure that ensures that the given functions are accelerated, and make sure every intrinsic you've created is accelerated. Check VectorAdd.cs for example. You might also be able to use CheckValue<> instead of your own value comparisons.

?

helloguo:
@BruceForstall Yes, modifying tests is in progress.

src/jit/lsraxarch.cpp
```diff
@@ -2676,6 +2676,91 @@ void Lowering::TreeNodeInfoInitSIMD(GenTree* tree)
...
+#ifdef _TARGET_AMD64_
+            if ((comp->getSIMDInstructionSet() == InstructionSet_AVX) || (simdTree->gtSIMDBaseType == TYP_ULONG))
+#else
+            if (simdTree->gtSIMDBaseType == TYP_LONG)
+            {
+                info->internalFloatCount = 3;
+            }
+            else if ((comp->getSIMDInstructionSet() == InstructionSet_AVX) || (simdTree->gtSIMDBaseType == TYP_ULONG))
+#endif
+            {
```
BruceForstall: Couldn't this be:
```
#ifdef _TARGET_X86_
            if (simdTree->gtSIMDBaseType == TYP_LONG)
            {
                info->internalFloatCount = 3;
            }
            else
#endif
                if ((comp->getSIMDInstructionSet() == InstructionSet_AVX) || (simdTree->gtSIMDBaseType == TYP_ULONG))
```
?

src/jit/simd.cpp
```diff
@@ -2617,6 +2621,61 @@ GenTreePtr Compiler::impSIMDIntrinsic(OPCODE                opcode,
...
+        case SIMDIntrinsicConvertToUInt64:
+        {
+#ifdef _TARGET_AMD64_
+            op1 = impSIMDPopStack(simdType, instMethod);
+
+            simdTree = gtNewSIMDNode(simdType, op1, nullptr, simdIntrinsicID, baseType, size);
+            retVal   = simdTree;
+#else
+            JITDUMP("SIMD Conversion is not supported on this platform\n");
+            return nullptr;
```
BruceForstall: Can you be more specific here, since ConvertToSingle/Double/Int32/UInt32 is supported. E.g.,
```
JITDUMP("SIMD Conversion to Int64/UInt64 is not supported on this platform\n");
```

src/jit/simdcodegenxarch.cpp
```diff
@@ -951,6 +1088,716 @@ void CodeGen::genSIMDIntrinsicUnOp(GenTreeSIMD* simdNode)
...
+    regNumber targetReg = simdNode->gtRegNum;
+    assert(targetReg != REG_NA);
+    var_types targetType = simdNode->TypeGet();
+
+    regNumber   op1Reg = genConsumeReg(op1);
+    instruction ins    = getOpForSIMDIntrinsic(simdNode->gtSIMDIntrinsicID, baseType);
+    if (intrinsicID == SIMDIntrinsicConvertToSingle && baseType == TYP_UINT)
+    {
+        regNumber tmpIntReg   = genRegNumFromMask(simdNode->gtRsvdRegs & RBM_ALLINT);
+        regMaskTP tmpRegsMask = (simdNode->gtRsvdRegs & RBM_ALLFLOAT);
```
BruceForstall: Please change all your code that finds/extracts registers from the gtRsvdRegs mask to instead call the `ExtractTempReg()` or `GetSingleTempReg()` APIs. You can use `AvailableTempRegCount()` for asserts, if desired, although generally just using the new APIs will give you the asserts you need.

src/jit/simdcodegenxarch.cpp
```diff
@@ -951,6 +1088,716 @@ void CodeGen::genSIMDIntrinsicUnOp(GenTreeSIMD* simdNode)
...
+        regMaskTP tmpReg3Mask = genFindLowestBit(tmpRegsMask);
+
+        tmpReg  = genRegNumFromMask(tmpReg1Mask);
+        tmpReg2 = genRegNumFromMask(tmpReg2Mask);
+        tmpReg3 = genRegNumFromMask(tmpReg3Mask);
+        assert(tmpReg != op1Reg && tmpReg2 != op1Reg && tmpReg3 != op1Reg);
+    }
+    else if (iset == InstructionSet_AVX || (baseType == TYP_ULONG))
+#endif
+    {
```
BruceForstall: This could be:
```
#ifdef _TARGET_X86_
    if (baseType == TYP_LONG)
    {
    ...
    }
#else
    else if (iset == InstructionSet_AVX || (baseType == TYP_ULONG))
```

CarolEidt:
> Because TakesRexWPrefix returns false when attris not EA_8BYTE (https://github.com/dotnet/coreclr/blob/master/src/jit/emitxarch.cpp#L222), we need to set attr to EA_8BYTEfor vpermq in order to set VEX.W bit.

Yes, I believe that's what I suggested. Note that `vpermq` is specifically for quadwords, which in xarch-speak is 8 bytes. So it makes sense to provide `EA_8BYTE` as the `emitSize` on the call to `emitIns_R_R_I`. (See my comment in simdcodegenxarch.cpp. I don't think we need (or want) a special case in `TakesRexWPrefix()`.

helloguo:
> Yes, I believe that's what I suggested. Note that vpermq is specifically for quadwords, which in xarch-speak is 8 bytes. So it makes sense to provide EA_8BYTE as the emitSize on the call to emitIns_R_R_I. (See my comment in simdcodegenxarch.cpp. I don't think we need (or want) a special case in TakesRexWPrefix().

@CarolEidt if we just set `attr `to `EA_8BYTE`, the L bit will not be set to 1 (https://github.com/dotnet/coreclr/blob/master/src/jit/emitxarch.cpp#L197-L200), which means we need a special case for `vpermq `in function `AddVexPrefix`.

CarolEidt:
@helloguo - Sorry for the delay. I spent some time looking into this, and you are right; it doesn't seem that there is a great way to cleanly support both register size specification and operand size specification in the current instruction formats and descriptors. I looked at what I did for the vgather variants (experimental work for https://github.com/dotnet/corefx/issues/1608), which have a similar encoding issue - the 'W' bit encodes single vs. double, while the 'L' bit (which in this case may take on both 0/1) encodes 128 vs. 256. Since the opcode implicitly encodes single vs. double, I had taken an approach similar to what you suggest above in `TakesRexWPrefix()`.

I suggest adding a very explicit comment along these lines:
```
// Because the current implementation of AVX does not have a way to distinguish between the register
// size specification (128 vs. 256 bits) and the operand size specification (32 vs. 64 bits), where both are
// required, the instruction must be created with the register size attribute (EA_16BYTE or EA_32BYTE),
// and here we must special case these by the opcode.
```
Does that seem reasonable?

russellhadley:
@pgavlin Can you add a note on the complexity of keeping the register and operand size specification separate?

BruceForstall:
The test changes look good to me.

src/jit/emitxarch.cpp
CarolEidt: @BruceForstall - for AVX instructions that encode REX-equivalent bits (e.g. the W bit), we handle them as for the REX encoding. In fact, for instructions that have both AVX and SSE encodings, we simply generate the AVX encoding from the same encoding bits when AVX encodings are enabled. So, this is the universal way to set 'W' bits, whether they wind up in an AVX or VEX prefix.
I find this approach preferable to adding a separate method. It is what I did for the vgather variants when I was experimenting with that, which also require both the W and L bits to be set appropriately.

CarolEidt: @helloguo - thanks for making this change; I think it is a better approach.

helloguo:

src/jit/simdcodegenxarch.cpp
helloguo: Thank you for your suggestion. Because these APIs (`ExtractTempReg()` or `GetSingleTempReg()`) were not there when this PR was first made, it's not very convenient to fix it in this PR. I will submit a fix once this PR is merged.

BruceForstall: Couldn't you just rebase to upstream/master, make the change, then push your branch again?

helloguo: Sure. Done.

BruceForstall: Looks great. Thanks.

russellhadley:
Given the dates for 2.0 (we're going to fork imminently), let's hold off on this until master opens for 2.1.

BruceForstall:
@helloguo I expect we will be able to merge this soon. In the meantime, can you update the change so it doesn't have any merge conflicts?

BruceForstall:
@helloguo Looks like GitHub still thinks there is a merge conflict in codegenlinear.h? Maybe you need to rebase to upstream/master?

helloguo:

helloguo:
@BruceForstall Done.

BruceForstall:
@dotnet-bot test this please

helloguo:

BruceForstall:
@helloguo Thanks for all the work!

PR #10663, opened 2017-04-03
CoreRT: make CORINFO_HELP_READYTORUN_DELEGATE_CTOR optimization for verifiable sequences.

sandreenko:
Save the token from CEE_LDFTN or CEE_LDVIRTFTN and restore it in the importer.
Use the token to always perform delegate ctor optimization.
It is jit part of dotnet/corert#2102.


sandreenko:
PTAL @pgavlin @dotnet/jit-contrib.

src/jit/_typeinfo.h
```diff
@@ -365,6 +367,13 @@ class typeInfo
         m_method = method;
     }
 
+    typeInfo(CORINFO_RESOLVED_TOKEN* token)
+    {
+        assert(token != nullptr && token->hMethod != nullptr && !isInvalidHandle(token->hMethod));
```
pgavlin: This assert should be broken into three successive asserts for better diagnosability:
```
assert(token != nullptr);
assert(token->hMethod != nullptr);
assert(!IsInvalidHandle(token->hMethod));
```

src/jit/importer.cpp
```diff
@@ -7289,6 +7291,13 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
         exactContextHnd = nullptr;
     }
 
+    if (opcode == CEE_NEWOBJ && clsFlags & CORINFO_FLG_DELEGATE)
```
pgavlin: This condition should be `(opcode == CEE_NEWOBJ) && ((clsFlags & CORINFO_FLG_DELEGATE) != 0)`

src/jit/importer.cpp
```diff
@@ -7369,7 +7378,8 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
             {
                 // New inliner morph it in impImportCall.
                 // This will allow us to inline the call to the delegate constructor.
-                call = fgOptimizeDelegateConstructor(call->AsCall(), &exactContextHnd);
+                assert(ldftnToken);
```
pgavlin: `assert(ldftnToken != nullptr)`

src/jit/titypes.h
```diff
@@ -6,6 +6,7 @@ DEF_TI(TI_ERROR, "<ERROR>")
 DEF_TI(TI_REF, "Ref")
 DEF_TI(TI_STRUCT, "Struct")
 DEF_TI(TI_METHOD, "Method")
+DEF_TI(TI_TOKEN, "Token")
```
pgavlin: I would add a flag to `typeInfo::m_bits` rather than adding a new `TI_type`:
a) the `TI_TOKEN` case is logically a specialization of `TI_METHOD`, and so doesn't require a new type
b) adding a new `TI_type` impacts more code than just that which uses `typeInfo` values

src/inc/corinfo.h
```diff
@@ -1568,6 +1568,9 @@ enum CorInfoTokenKind
 
     // token comes from CEE_NEWOBJ
     CORINFO_TOKENKIND_NewObj    = 0x200 | CORINFO_TOKENKIND_Method,
+
+    // token comes from CEE_LDVIRTFTN
+    CORINFO_TOKENKIND_LdvirtualToken = 0x400 | CORINFO_TOKENKIND_Method,
```
MichalStrehovsky: Nit: the existing values are all named after the IL instruction referring to the token. The consistent naming would be `CORINFO_TOKENKIND_Ldvirtftn`.

src/jit/_typeinfo.h
```diff
@@ -302,6 +306,8 @@ class typeInfo
         CORINFO_CLASS_HANDLE m_cls;
         // Valid only for type TI_METHOD
         CORINFO_METHOD_HANDLE m_method;
+        // Valid only for type TI_TOKEN
```
sandreenko: should be Valid only for token.

src/jit/importer.cpp
```diff
@@ -7289,6 +7297,21 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
         exactContextHnd = nullptr;
     }
 
+    if ((opcode == CEE_NEWOBJ) && ((clsFlags & CORINFO_FLG_DELEGATE) != 0))
+    {
+        // Only verifiable cases are supported.
+        // dup; ldvirtftn; newobj; or ldftn; newobj.
+        // IL test could contain unverifiable sequence, in this case optimization should not be done.
```
pgavlin: Can you provide an example of such a sequence?

sandreenko: Yes, it is from tests\src\JIT\Methodical\delegate\oddpower.il

```
    IL_01d1:  dup
    IL_01d2:  call instance method int32 * () OddPower::GetMthdPtr1()
    IL_01d8:  newobj     instance void OddPower/MyDeleg2::.ctor(object, native int)
```

src/jit/_typeinfo.h
```diff
@@ -300,8 +304,10 @@ class typeInfo
 
     union {
         CORINFO_CLASS_HANDLE m_cls;
-        // Valid only for type TI_METHOD
+        // Valid only for type TI_METHOD without IsToken
         CORINFO_METHOD_HANDLE m_method;
+        // Valid only for TI_TOKEN with IsToken
+        CORINFO_RESOLVED_TOKEN* m_token;
```
AndyAyersMS: A `CORINFO_RESOLVED_TOKEN` is fairly large. Seems like you only ever look at the `hMethod` field. So why not just reuse the `m_method` field for the method handle and use the new token bit to safeguard access?

AndyAyersMS: Never mind, I see it gets passed back into the runtime.

Anyways the cases below where you create heap tokens should go into jit-allocated memory. I don't think this happens with `new`. Can you double-check.

src/jit/importer.cpp
```diff
@@ -12272,7 +12295,8 @@ void Compiler::impImportBlockCode(BasicBlock* block)
                     return;
                 }
 
-                impPushOnStack(op1, typeInfo(resolvedToken.hMethod));
+                CORINFO_RESOLVED_TOKEN* heapToken = new CORINFO_RESOLVED_TOKEN(resolvedToken);
```
AndyAyersMS: Make sure this is going into jit-allocated memory so it doesn't leak.

src/jit/importer.cpp
```diff
@@ -12377,7 +12401,10 @@ void Compiler::impImportBlockCode(BasicBlock* block)
                     return;
                 }
 
-                impPushOnStack(fptr, typeInfo(resolvedToken.hMethod));
+                CORINFO_RESOLVED_TOKEN* heapToken = new CORINFO_RESOLVED_TOKEN(resolvedToken);
```
AndyAyersMS: Ditto here.

src/jit/_typeinfo.h
sandreenko: You are right. Fixed.
I thought that the global new is overloaded to allocate memory in the memory pool.

sandreenko:
@AndyAyersMS Does it look good now?

PR #10664, opened 2017-04-03
Fix InternalsVisibleTo when it references an assembly with some Debug…

### LINKED ISSUE #3541: JITTrackingEnabled on coreclr making internals visible to not work correctly
When the F# compiler emits a DebuggableAttribute with JitEnabled false, the built binary can no longer access internals in another assembly when internals visible has been correctly specified.

The attached zip file has two complete fsharp compilers and its coreclr dependencies.

[jit-tracking.zip](https://onedrive.live.com/redir?resid=24D4E4E2DE3A85A1!109903&authkey=!ALZ7NJzV9yqvDFc&ithint=file%2czip)

fsc.portable has jit tracking on.
fsc-pdbonly has jit tracking off.

Not that corerun fsc.exe, fails for fsc.portable and runs correctly for fsc-pdbonly.

Expected both to work.

Kevin Ransom

kouvel:
…gableAttribute flags

Fixes #3541
- Mask out the DebuggableAttribute bits from when comparing assembly spec flags for matching an InternalsVisibleTo reference to an assembly

### DIFF
--- a/src/inc/corhdr.h
+++ b/src/inc/corhdr.h
@@ -751,6 +751,7 @@ typedef enum CorAssemblyFlags
 
     afEnableJITcompileTracking   =  0x8000, // From "DebuggableAttribute".
     afDisableJITcompileOptimizer =  0x4000, // From "DebuggableAttribute".
+    afDebuggableAttributeMask    =  0xc000,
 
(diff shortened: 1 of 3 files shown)


kouvel:
Filed separate issue to figure out how to test this, a quick test didn't work

kouvel:
@jkotas @gkhanna79

PR #10668, opened 2017-04-03
[Linux/GDB-JIT] Fix 'this' pointer in struct methods

ayuckhulk:
This PR fixes `this` method argument for value types - `this` should always be a pointer to underlying structure.

@mikem8361, @janvorli PTAL


ayuckhulk:
@janvorli Thank you for the review!

PR #10670, opened 2017-04-03
[Local GC] Pull some logic using SystemDomain::System out of the GC

swgillespie:
In the same vein of my last few PRs, this PR moves some EE introspection out of the GC and onto `GCToEEInterface`:

In `generation_to_condemn`, the GC may choose to elevate a Gen 2 GC to be blocking if the system domain requires app domain cleanup. Instead of calling `RequiresAppDomainCleanup` on the system domain from the GC, this PR moves all of that logic to the EE side and instead has the GC invoke a method on `GCToEEInterface` to determine whether or not this GC should be elevated to a blocking collection on behalf of the system domain.

### DIFF
--- a/src/gc/env/gcenv.ee.h
+++ b/src/gc/env/gcenv.ee.h
@@ -71,6 +71,7 @@ public:
 
     static void HandleFatalError(unsigned int exitCode);
     static bool ShouldFinalizeObjectForUnload(AppDomain* pDomain, Object* obj);
+    static bool ForceFullGCToBeBlocking();
     static bool ShouldElevateForAppDomainCleanup();
     static bool EagerFinalized(Object* obj);
 };
--- a/src/gc/env/gcenv.ee.h
+++ b/src/gc/env/gcenv.ee.h
@@ -71,6 +71,7 @@ public:
 
     static void HandleFatalError(unsigned int exitCode);
     static bool ShouldFinalizeObjectForUnload(AppDomain* pDomain, Object* obj);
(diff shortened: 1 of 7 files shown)


swgillespie:
@Maoni0 @jkotas PTAL?

jkotas:
I have never understood why we have this check. There is nothing fundamental that requires GC for AppDomain unload to be blocking. As far as I can tell, it was just a quick workaround for a stress bug long time ago that somebody forgot to remove. I wanted to remove this check some time ago, but folks were too scared about exposing pre-existing stress bugs, so we are stuck with it.

Also, ShouldElevateForAppDomainCleanup is not quite the right name for it because of it returns true for unloading of collectible assemblies as well.

I would call the GC callback something like `ForceFullGCToBeBlocking()`, and maybe also include comment about what I have just said in the EE side of the implementation.

Maoni0:
not "forgot to remove" - rather it was done so we wouldn't AV and someone never found out why we AVed in the first place without it (I think it looked at it some and the AV happened on an object GC had not processed so someone else was going to look at it but never got around to). We can do some intense stress runs without it and see if we could repro the bug but I wouldn't remove it as part of this PR; or if you do, make sure to put a [DESKTOP TODO] so this is taken care of on desktop.

swgillespie:
Will do re: the renaming. I'll also make a note of this discussion in a comment.

swgillespie:
Thanks!

Issue #7994, opened 2016-11-04
Remove the VC Redist pre-req


Petermarcu:
We need to come up with a solution that doesn't require users to first install the VC Redist in order to install .NET Core.

gkhanna79:
Fixing this can be done by including the required files when building the RID split version of M.N.App. @weshaggard Do you still want uCRT binaries to go in their own package, like the apiset one, and be a dependency of CoreCLR package?

weshaggard:
I don't think I ever wanted them to be in their own package. We should see if the C++ team already provides them as a package and if so then we can use that but if not then I would be hesitant for us to package them by themselves and instead would prefer we package them directly in the CoreCLR package or the M.N.App package.

Petermarcu:
Why not just have them delivered with coreclr? That sounds simplest.

gkhanna79:
@wtgodbe As discussed, I am assigned this to you. Fixing this involves two stages:

1) Package the UCRT binaries into the CoreCLR nuget package. During the build, use `UniversalCRTSDKDir` environment variable to get the base location to find the binaries for the architecture in question. We will **not** be doing this for Arm64 since the problem is not applicable there. Also, we should package the binaries only during `Release` build.

2) Once the CoreCLR package with this change is published and Maestro update has been merged in Core-Setup, please update the Core-Setup Windows platform installers to remove the VCRedist installer dependency. The installers live under https://github.com/dotnet/core-setup/tree/master/packaging/windows.

PR #10671, opened 2017-04-03
Publish VC Redist DLLs as part of CoreCLR Package

wtgodbe:
CC @gkhanna79 @Petermarcu 

First part of addressing https://github.com/dotnet/coreclr/issues/7994

I tested locally, and this does indeed include VC Redist binaries in the runtime package, in `runtimes\win7-{arch}\native`, next to coreclr.dll & other native binaries.

@chcosta these VC Redist .dll's don't have symbol packages - will that be a problem for publishing? How can I get around that?

### DIFF
--- a/src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/runtime.Windows_NT.Microsoft.NETCore.Runtime.CoreCLR.props
+++ b/src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/runtime.Windows_NT.Microsoft.NETCore.Runtime.CoreCLR.props
@@ -20,6 +20,7 @@
     <NativeBinary Include="$(BinDir)mscorrc.dll" />
     <NativeBinary Include="$(BinDir)sos.dll" />
     <NativeBinary Include="$(BinDir)System.Private.CoreLib.ni.dll" />
+    <NativeBinary Include="$(UniversalCRTSDKDir)Redist\ucrt\DLLs\$(BuildArch)\*.dll" Condition="'$(BuildType)'=='Release' AND '$(BuildArch)' != 'arm64'" />
     <ArchitectureSpecificLibFile Include="$(BinDir)System.Private.CoreLib.dll" />
     <ArchitectureSpecificLibFile Include="$(BinDir)SOS.NETCore.dll" />
     <ArchitectureSpecificToolFile Include="$(BinDir)crossgen.exe" />
--- a/src/.nuget/dir.targets
+++ b/src/.nuget/dir.targets
@@ -1,6 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+
+  <Target Name="VerifyVCRedist" BeforeTargets="GetSymbolPackageFiles" Condition="'$(_runtimeOSFamily)' == 'win'">
+    <Error Condition="'$(UniversalCRTSDKDir)' == ''" Text="Unable to find VC Redist binaries - check that UniversalCRTSDKDir environment variable is set" />
+  </Target>
   <!--
     Finds symbol files and injects them into the package build.
   -->


src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/runtime.Windows_NT.Microsoft.NETCore.Runtime.CoreCLR.props
```diff
@@ -20,6 +20,7 @@
     <NativeBinary Include="$(BinDir)mscorrc.dll" />
     <NativeBinary Include="$(BinDir)sos.dll" />
     <NativeBinary Include="$(BinDir)System.Private.CoreLib.ni.dll" />
+    <NativeBinary Include="$(UniversalCRTSDKDir)Redist\ucrt\DLLs\$(BuildArch)\*.dll" Condition="'$(BuildType)'=='Release' AND '$(BuildArch)' != 'arm64'" />
```
gkhanna79: Can you please add a check to fail the build if UniversalCRTSDKDir is not defined?

gkhanna79: Did you get failure when generating symbol packages?

wtgodbe: @gkhanna79 done

wtgodbe: The symbol packages contain the .dll's, but no corresponding .pdb's.

gkhanna79: @chcosta What are your thoughts on this?

CC @weshaggard

chcosta: This shouldn't have an effect on symbol packages

src/.nuget/dir.targets
```diff
@@ -1,6 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+
+  <Target Name="VerifyVCRedist" BeforeTargets="GetSymbolPackageFiles" Condition="'$(_runtimeOSFamily)' == 'win'">
```
chcosta: `GetSymbolPackageFiles` -> `GetPackageFiles`

It's kind of weird (to me), to rely on the machine state for these vs picking them up from somewhere else, like how the apiset packages were handled.

wtgodbe: How were the apiset packages handled?

chcosta: There are apiset packages on myget for use in operating systems that don't include them.  I'd imagine you just package these vc redist files similarly and consume them via those packages or repackage them with coreclr.

chcosta: https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.Windows.ApiSets

wtgodbe: Who publishes that package/where do they get the binaries from? If we did the same here, wouldn't that have the same issue, since whoever publishes the VC Redist bits would have to get them from program files anyways?

chcosta: Yes, it's just changing the model slightly.  They still need to be picked up from somewhere for packaging/publishing.  You don't have to produce that package with every build though and you could just consume the package in coreclr rather than packaging the build machine state.  Somebody has to have a machine requirement on the binaries, but it doesn't have to be CoreClr.  The ApiSet packages used to be produced out of ProjectK, I believe.  CoreFx has moved away from using the ApiSet packages, but I think that's more an evolution of the product than an issue with the packages.  I don't know enough about the history to really fight for it being a better model, I'm just suggesting it as a possible alternative.

gkhanna79: We are intentionally choosing to not author another package that we need to maintain. Also, these files are part of the toolset that we are using to build.

wtgodbe: @chcosta I think having the target before GetSymbolPackageFiles is the right thing, as that Target uses the NativeBinary Item, which depends on VCRedist files existing on the machine - we should fail before doing any operations on that data if VCRedist isn't present, as that work would've been unnecessary

src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/runtime.Windows_NT.Microsoft.NETCore.Runtime.CoreCLR.props
```diff
@@ -20,6 +20,7 @@
     <NativeBinary Include="$(BinDir)mscorrc.dll" />
     <NativeBinary Include="$(BinDir)sos.dll" />
     <NativeBinary Include="$(BinDir)System.Private.CoreLib.ni.dll" />
```
weshaggard: Any reason this is scoped to Release?

gkhanna79: Since there are not debug binaries (which introduce additional dependencies that Release CRT build cannot support) for CRT that can be distributed, we statically link it for Debug build.

PR #10672, opened 2017-04-03
Fix a reference to System.Exception in a JIT test.

pgavlin:
This bad reference was casuing a failure in the x64 pri1 R2R jobs.


pgavlin:
@BruceForstall @AndyAyersMS @dotnet/jit-contrib PTAL

PR #10676, opened 2017-04-03
[Local GC] Move Weak Reference finalization out of the GC

swgillespie:
When finalizing objects on a segment, the GC checks whether or not an object's method table is that of `System.WeakReference` or `System.WeakReference<T>`, in which case it calls `FinalizeWeakReference` on the EE side to perform the actual work of finalizing a weak reference.

This PR moves all of this logic to `GCToEEInterface`, so that the GC does not know or care whether or not an object is a weak reference: it will call `GCToEEInterface::FinalizeIfWeakReference`, which returns `true` if the object was a weak reference (and was finalized) and `false` if the object wasn't a weak reference. The implementation of `GCToEEInterface::FinalizeIfWeakReference` does the method table check and, if the object is a weak reference, calls `FinalizeWeakReference` on it.

### DIFF
--- a/src/gc/gc.cpp
+++ b/src/gc/gc.cpp
@@ -36269,7 +36269,7 @@ CFinalize::ScanForFinalization (promote_func* pfn, int gen, BOOL mark_only_p,
 
                     assert (method_table(obj)->HasFinalizer());
 
-                    if (GCToEEInterface::FinalizeIfWeakReference(obj))
+                    if (GCToEEInterface::EagerFinalized(obj))
                     {
                         MoveItem (i, Seg, FreeList);
                     }
--- a/src/gc/gc.cpp
+++ b/src/gc/gc.cpp
@@ -36269,16 +36269,11 @@ CFinalize::ScanForFinalization (promote_func* pfn, int gen, BOOL mark_only_p,
 
                     assert (method_table(obj)->HasFinalizer());
 
-#ifndef FEATURE_REDHAWK
-                    if (method_table(obj) == pWeakReferenceMT || method_table(obj)->GetCanonicalMethodTable() == pWeakReferenceOfTCanonMT)
+                    if (GCToEEInterface::FinalizeIfWeakReference(obj))
                     {
-                        //destruct the handle right there.
-                        FinalizeWeakReference (obj);
                         MoveItem (i, Seg, FreeList);
                     }
-                    else
-#endif //!FEATURE_REDHAWK
-                    if ((obj->GetHeader()->GetBits()) & BIT_SBLK_FINALIZER_RUN)
+                    else if ((obj->GetHeader()->GetBits()) & BIT_SBLK_FINALIZER_RUN)
                     {
                         //remove the object because we don't want to
                         //run the finalizer
--- a/src/vm/gcenv.ee.cpp
+++ b/src/vm/gcenv.ee.cpp
@@ -1369,10 +1369,11 @@ bool GCToEEInterface::ShouldFinalizeObjectForUnload(AppDomain* pDomain, Object* 
     return true;
 }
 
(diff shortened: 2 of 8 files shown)


src/gc/gc.cpp
```diff
@@ -36269,16 +36269,11 @@ CFinalize::ScanForFinalization (promote_func* pfn, int gen, BOOL mark_only_p,
 
                     assert (method_table(obj)->HasFinalizer());
 
-#ifndef FEATURE_REDHAWK
-                    if (method_table(obj) == pWeakReferenceMT || method_table(obj)->GetCanonicalMethodTable() == pWeakReferenceOfTCanonMT)
+                    if (GCToEEInterface::FinalizeIfWeakReference(obj))
```
jkotas: This is a general purpose callback that can be used to eagerly finalize any type during GC instead of waiting for it to be finalized later by the finalizer thread. It just happens that the current implementation is using it for finalizing weak references.

I would call it `bool EagerFinalization(Object* obj)` or something similar.

swgillespie: Will do!

swgillespie:
The assert is due to the GC's `method_table` macro masking off the mark bit, while `Object`'s `GetMethodTable` does not (and asserts that the mark bit isn't set). I'll address this.

@Maoni0 PTAL as well?

src/gc/gc.cpp
Maoni0: Right. I would call this EagerFinalized just to make it obvious that it returns a boolean value whether it was eagerly finalized; and if it was GC will do the management on the finalize queue.

---
In reply to: [109552827](https://github.com/dotnet/coreclr/pull/10676#discussion_r109552827) [](ancestors = 109552827)

src/vm/gcenv.ee.cpp
```diff
@@ -1359,3 +1368,15 @@ bool GCToEEInterface::ShouldFinalizeObjectForUnload(AppDomain* pDomain, Object*
     // to move them to a new app domain instead of finalizing them here.
     return true;
 }
+
+bool GCToEEInterface::FinalizeIfWeakReference(Object* obj)
+{
+    MethodTable* pMT = obj->GetMethodTable();
+    if (pMT == pWeakReferenceMT || pMT == pWeakReferenceOfTCanonMT)
```
Maoni0: >| pMT ==  [](start = 33, length = 9)

why is that the call to GetCanonicalMethodTable is gone?

swgillespie: I missed that - thanks!

src/gc/gc.cpp
swgillespie: Ok!

Maoni0:
otherwise LGTM!

janvorli:
The leakwheel test has failed in the CI on Ubuntu x64 with:
```
Assert failure(PID 32074 [0x00007d4a], Thread: 32101 [0x7d65]): (dac_cast<TADDR>(m_pMethTab) & MARKED_BIT) == 0
File: /mnt/j/workspace/dotnet_coreclr/master/checked_ubuntu_prtest/src/vm/object.h Line: 242
```

swgillespie:
@janvorli Yeah, that's the assert I was talking about above - I need to use `GetGCSafeMethodTable` instead of `GetMethodTable` so that we mask off the mark bit instead of asserting that it's not set. I have a fix that I'm about to push.

swgillespie:
@dotnet-bot test OSX Release gcsimulator
@dotnet-bot test OSX Checked longgc
@dotnet-bot test Ubuntu Checked longgc	
@dotnet-bot test Windows_NT Checked longgc

swgillespie:
thanks!

PR #10678, opened 2017-04-03
Remove the SO_TOLERANT contact from CEEInfo::resolveVirtualMethodHelper

briansull:
that contract should only apply to CEEInfo::resolveVirtualMethod which uses the JIT_TO_EE_TRANSITION() transition macros.


briansull:
PTAL @AndyAyersMS @BruceForstall

PR #10682, opened 2017-04-04
Remove unused code under typo define

BruceForstall:
The code is under COR_JIT_EE_VER which never existed (it used to
be COR_JIT_EE_VERSION), thus the code has never been used.


BruceForstall:
@pgavlin PTAL
cc @dotnet/jit-contrib

PR #10685, opened 2017-04-04
[x86/Linux] Correctly unwind esp frames

parjong:
UnwindEspFrame already considers pushed argument size when unwinding esp frame, and thus SP should be current SP (not SP before argument pushes) when invoking UnwindEspFrame.

Unfortunately, OOPStackUnwinderX86::VirtualUnwind provides SP before argument pushes, which results in  incorrect unwinding inside baseservices.threading.monitor.enter.monitorenter.

This commit revises x86/Linux unwider to correctly unwind esp frames.


parjong:

PR #10406, opened 2017-03-23
[ARM32/Linux] Build libcoreclrtraceptprovider.so

hseok-oh:
- Build libcoreclrtraceptprovider.so for ARM32
- Include libcoreclrtraceptprovider.so in nuget package
- Except Tizen(armel): lttng is not supported in Tizen

related issue: #8684

PR #10686, opened 2017-04-04
Pass crossbuild and linux-id to CMakeLists.txt using CMake arguments

hseok-oh:
Remove in CMakeLists.txt
1) read cross-build from environment variable
2) read linux-id from file system

Instead, pass cross-build and linux-id from gen-buildsys-clang.sh using cmake arguments.


hseok-oh:
This is part of #10406 
cc/ @janvorli

PR #10692, opened 2017-04-04
Update module not found error message

janvorli:
This message can be misleading, since it says "The specified module
could not be found", but it would also report case when the module
was found, but one of the module's dependencies were not found.
Multiple people in the past were hit by that when their module was
not loaded, they were thinking that there is some problem with
the path coreclr is loading modules from or something like that and
spent non-trivial amount of time trying to figure that out.
The fact that the module could have missing dependencies is really
not obvious.

So I am updating the message to reflect that fact.

### DIFF
--- a/src/pal/src/misc/errorstrings.cpp
+++ b/src/pal/src/misc/errorstrings.cpp
@@ -76,7 +76,7 @@ ErrorString palErrorStrings[] =
     { ERROR_SEM_TIMEOUT, W("The semaphore timeout period has expired.\n") },
     { ERROR_INSUFFICIENT_BUFFER, W("The data area passed to a system call is too small.\n") },
     { ERROR_INVALID_NAME, W("The filename, directory name, or volume label syntax is incorrect.\n") },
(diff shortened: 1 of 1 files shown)


janvorli:
@adityamandaleeka  can you take a look please?

janvorli:
FYI: @yizhang82, @alexanderuv

bendono:
The message is an improvement, but it would be much more helpful if the message would specific which module could not be loaded. In Desktop I have always relied on Fusion logs, but Core doesn't have this functionality.

janvorli:
@bendono - the exception message already contains the module name. This is the message before my change:
```
System.DllNotFoundException : Unable to load DLL 'libLLVM.dylib': The specified module could not be found.
 (Exception from HRESULT: 0x8007007E)
```

bendono:
@janvorli Thank you. That looks good.

janvorli:

janvorli:
@dotnet-bot  test Tizen armel Cross Debug Build  please

PR #10695, opened 2017-04-04
Update VS2017 'Individual Components' install info

JosephTremoulet:
A few small changes to make it easier to compare selections in the
installer UI against the list in the doc:

 - Re-order the list to match the order presented in the installer UI
 - Place the individual items under groupings matching the installer
 - Change casing to match installer UI


danmosemsft:
LGTM

Issue #10694, opened 2017-04-04
Seemingly Arbitrary Random Objects in Test APIs


rionmonster:
When looking into another existing issue, I came across the following seemingly unused instances of a `Random` class similar to the the following instance:

```
public void TryGetValueValidations(K[] keys, V[] values, K key, V value)
{
        // Other code omitted for brevity
       
        // This is an example of a random Random instance
        Random r = new Random();
        Test.Eval(!tbl.TryGetValue(key, out val), "Err_011 Expected TryGetValue to return false");
        Test.Eval(val == null, "Err_012 Expected val to be null");
}
```

After a cursory look, it appeared a few times specifically within the Test APIs (`tests/src/baseservices/compilerservices/dynamicobjectproperties/testapis.cs`) section, but may be found in other places: 

  - [C:\dev\coreclr\tests\src\baseservices\compilerservices\dynamicobjectproperties\testapis.cs(94)](https://github.com/dotnet/coreclr/blob/master/tests/src/baseservices/compilerservices/dynamicobjectproperties/testapis.cs#L89-L119)
  - [C:\dev\coreclr\tests\src\baseservices\compilerservices\dynamicobjectproperties\testapis.cs(150)](https://github.com/dotnet/coreclr/blob/master/tests/src/baseservices/compilerservices/dynamicobjectproperties/testapis.cs#L121-L153)
     
Is there are reason for this or do these need to be removed?

stephentoub:
If the local is completely unused, it could be deleted.

adityamandaleeka:
@rionmonster Can this issue be closed now?

rionmonster:
@adityam Done :)

PR #10696, opened 2017-04-04
Removed Unused Local Random Instances

rionmonster:
Related to Issue #10694 

Discovered two cases of a `Random` instance that weren't being used within their respective methods and removed them.


adityamandaleeka:
LGTM, thanks.

PR #10698, opened 2017-04-04
Remove `GTF_REVERSE_OPS` from LIR.

pgavlin:
In HIR, this flag indicates that the second operand to a binary node will execute before
the node's first operand. LIR, however, no longer determines ordering via use edges, so
this flag only affects the order in which operands to a node are considered. The sole
constraint on this use ordering is that for a given node, the same ordering must be used
in liveness, LSRA, and the code generator; this is due to the correspondence between
use ordering and spill/reload/last-use ordering. As a result, the reverse ops flag is
unnecessary and rather unhelpful in LIR, causing little more than a bit of extra complexity
throughout the backend.

This change removes `GTF_REVERSE_OPS` from LIR by clearing this flag during rationalize
and verifying that it remains clear in `LIR::CheckLIR`. We could reuse this bit for an
additional backend-specific purpose in the future with a bit more work in the checker.

### DIFF
--- a/src/jit/codegenlinear.cpp
+++ b/src/jit/codegenlinear.cpp
@@ -1282,12 +1282,7 @@ void CodeGen::genConsumeOperands(GenTreeOp* tree)
 {
     GenTree* firstOp  = tree->gtOp1;
     GenTree* secondOp = tree->gtOp2;
-    if ((tree->gtFlags & GTF_REVERSE_OPS) != 0)
-    {
-        assert(secondOp != nullptr);
-        firstOp  = secondOp;
-        secondOp = tree->gtOp1;
-    }
+
     if (firstOp != nullptr)
     {
         genConsumeRegs(firstOp);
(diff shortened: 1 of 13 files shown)


pgavlin:
@BruceForstall @CarolEidt @dotnet/jit-contrib PTAL. As mentioned in the description, this PR is primarily discussion-oriented for now.

src/jit/codegenlinear.cpp
```diff
@@ -1507,71 +1502,28 @@ void CodeGen::genSetBlockSrc(GenTreeBlk* blkNode, regNumber srcReg)
 
 void CodeGen::genConsumeBlockOp(GenTreeBlk* blkNode, regNumber dstReg, regNumber srcReg, regNumber sizeReg)
 {
-    // We have to consume the registers, and perform any copies, in the actual execution order.
-    // The nominal order is: dst, src, size.  However this may have been changed
-    // with reverse flags on the blkNode and the setting of gtEvalSizeFirst in the case of a dynamic
-    // block size.
+    // We have to consume the registers, and perform any copies, in the actual execution order: dst, src, size.
```
CarolEidt: This is not strictly correct. I think it should be something like: "We have to consume the registers, and perform any copies, in the canonical operand order (dst, src, size), as that is the order used also by the register allocator."
Then, the comment below should also be changed to use "canonical order" (or whatever term is used) instead of "execution order".

CarolEidt:
I love the simplification - and it looks good to me aside from a suggested comment change.

---
In reply to: [291571962](https://github.com/dotnet/coreclr/pull/10698#issuecomment-291571962) [](ancestors = 291571962)

CarolEidt:
:shipit:

src/jit/codegenlinear.cpp
pgavlin: That wording is definitely better. I will make the change.

mikedn:
Is it safe to assume that this will be merged sooner or later? I have 2 PRs that would benefit from this simplification.

Also note that it looks like decomposition doesn't handle reverse ops properly anyway, at least in `DecomposeArith`. It should copy the reverse ops flag from the original node (that becomes "lo") to the "hi" node that it creates. And if we ever produce a `GT_LONG` with reverse ops then things become really hairy as we probably need to propagate that too.

BruceForstall:
@mikedn Yes, after the final .NET Core 2.0 integration (next week?)

PR #10699, opened 2017-04-04
Fix handling of empty environment variables in PAL

adityamandaleeka:
Fix https://github.com/dotnet/corefx/issues/17614

@stephentoub 

BTW, there are still codepaths (such as malloc failure) where 0 can be returned from this function and the error code isn't ENV_VAR_NOT_FOUND. We may want to make GetEnvironmentVariableCore treat anything other than ERROR_SUCCESS as a failure if the return value is 0.

### DIFF
--- a/src/pal/src/misc/environ.cpp
+++ b/src/pal/src/misc/environ.cpp
@@ -218,7 +218,13 @@ GetEnvironmentVariableW(
     }
     else if (size == 0)
     {
-        // handle error in GetEnvironmentVariableA
+        // If size is 0, it either means GetEnvironmentVariableA failed, or that
+        // it succeeded and the value of the variable is empty. Check GetLastError
+        // to determine which. If the call failed, we won't touch the buffer.
+        if (GetLastError() == ERROR_SUCCESS)
+        {
+            *lpBuffer = '\0';
+        }
     }
     else
     {


src/pal/src/misc/environ.cpp
```diff
@@ -218,7 +218,13 @@ GetEnvironmentVariableW(
     }
     else if (size == 0)
     {
-        // handle error in GetEnvironmentVariableA
+        // If size is 0, it either means GetEnvironmentVariableA failed, or that
+        // it suceeded and the value of the variable is empty. Check GetLastError
```
stephentoub: Nit: suceeded => succeeded

stephentoub:
> We may want to make GetEnvironmentVariableCore treat anything other than ERROR_SUCCESS as a failure if the return value is 0

Seems reasonable.

src/pal/src/misc/environ.cpp
adityamandaleeka: oops, fixed.

adityamandaleeka:
> Can you make sure we have appropriate tests for this in corefx, too?

@stephentoub What's the best way to add this to the CoreFX tests? I guess we'll have to set an empty variable before the process starts because (a) changes via setenv after process start are not going to be picked up by the PAL environment code and (b) Environment.SetEnvironmentVariable treats an empty string value as null.

stephentoub:
> I guess we'll have to set an empty variable before the process starts because

We have a test utility in corefx called RemoteInvoke that lets you invoke a delegate in another process, and you can set a ProcessStartInfo for that process, which includes what environment variables you want to set.  So you can essentially easily add a test that invokes code in another process where that process was created with a specific set of environment variables.  The only question would be whether the process start code that takes that collection of environment variables does special handling of empty values... I don't remember if it does.

Issue #7807, opened 2016-10-25
Query real swap file size and potential VA space on the target System


ramarag:
This functionality will improve the perf for the MemoryFailPoint feature.
Follow up on https://github.com/dotnet/coreclr/pull/7751

This feature is needed on xplat

ramarag:
enable https://github.com/dotnet/corefx/pull/13399/commits/755ce2f72caee5a29b29cd150cb30b08514482b6#diff-ab7a4b1b3e267c2adfbe272048b78932R30 on Xplat, once this issue is fixed

janvorli:
Fixed by #10700. @ramarag, is seems you can enable the checks that were disabled in #7945 for xplat now.

PR #10700, opened 2017-04-04
Add reporting swap file size to GlobalMemoryStatusEx on Unix

janvorli:
This change adds swap file size reporting to GlobalMemoryStatusEx on Linux, OSX and FreeBSD. Also a new PAL test was added to exercise the function.


janvorli:
CC: @ramarag - this fixes the #7807

src/pal/src/misc/sysinfo.cpp
```diff
@@ -244,11 +253,65 @@ GlobalMemoryStatusEx(
...
-    // TODO: implement getting memory details via sysinfo. On Linux, it provides swap file details that
-    // we can use to fill in the xxxPageFile members.
 
-#endif // HAVE_SYSCONF
+#endif // HAVE_SYSCTL
+
+    // Get swap file size, consider the ability to get the values optional
+    // (don't return FALSE from the GlobalMemoryStatusEx)
+#if HAVE_XSW_USAGE
+    // This is available on OSX
```
janvorli: We use OSX everywhere in our sources / package names / OS reporting, so I'd prefer sticking with it here for now.

PR #10711, opened 2017-04-04
Fix registry name enumeration

JeremyKuhne:
Same change as https://github.com/dotnet/corefx/pull/17826, also pulls dead permission checks. I'll be cutting out the rest of the unused SPCL reg code after this goes in. From the original change:

Key data (values and subkeys) can change while we're in the midst of enumerating. This is causing intermittent failures in tests as ERROR_NO_MORE_ITEMS surfaces as an exception as we try and iterate past removed data.

In addition, allocating 32K to get value names causes 32K * number of items to get natively allocated on the heap. While these allocs are returned on each iteration it could potentially cause significant grief for large keys and concurrent heap allocations. Above and beyond this, allocating 64K for what are typically < 20 character names is a bit crazy, and as such we'll optimize for the common case.

### DIFF
--- a/src/mscorlib/src/Microsoft/Win32/RegistryKey.cs
+++ b/src/mscorlib/src/Microsoft/Win32/RegistryKey.cs
@@ -48,21 +48,13 @@
 
 */
 
-
-using System;
-using System.Collections;
-using System.Collections.Generic;
-using System.Security;
-using System.Text;
-using System.Threading;
-using System.IO;
-using System.Runtime.Remoting;
-using System.Runtime.InteropServices;
 using Microsoft.Win32.SafeHandles;
-using System.Runtime.Versioning;
-using System.Globalization;
+using System;
+using System.Buffers;
+using System.Collections.Generic;
 using System.Diagnostics.Contracts;
-using System.Diagnostics.CodeAnalysis;
+using System.IO;
+using System.Text;
 
 namespace Microsoft.Win32
 {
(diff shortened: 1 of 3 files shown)


src/mscorlib/src/Microsoft/Win32/RegistryKey.cs
```diff
@@ -363,144 +353,131 @@ public RegistryKey OpenSubKey(String name)
...
-        internal int InternalValueCount()
+        /// <summary>
+        /// Retrieves an array of strings containing all the value names.
+        /// </summary>
+        public unsafe string[] GetValueNames()
         {
             EnsureNotDisposed();
-            int values = 0;
-            int junk = 0;
-            int ret = Win32Native.RegQueryInfoKey(hkey,
```
danmosemsft: You got rid of RegQueryInfo key and just loop for ERROR_NO_MORE_ITEMS, but you didn't do that in the corefx change.

danmosemsft:
I guess we have no tests for any of this code? Only the CoreFX version. How will we we know your code works?
Why isn't this code moved down into corelib?
It's a shame it has to be here as it's used in only 2 places!

src/mscorlib/src/Microsoft/Win32/RegistryKey.cs
JeremyKuhne: Because the code in CoreFX is public. I'm going to remove the count code as we don't use it in SPCL- and it doesn't buy much, and in fact may be worse in some cases.

JeremyKuhne:
> I guess we have no tests for any of this code?

Only indirect via the stuff that is publicly exposed.

> Why isn't this code moved down into corelib?

It is useful to have the library be truly portable- we really don't use much of the reg code in corelib. I'll probably open an issue to put the needed functions in a shared code file.

> It's a shame it has to be here as it's used in only 2 places!

I'm going to chop out everything but the actually used code- so it will be tighter here.

JeremyKuhne:
Fixes https://github.com/dotnet/corefx/issues/17037

PR #10724, opened 2017-04-05
Use the correct type for interlocked instructions

### LINKED ISSUE #10714: Incorrect and unexpected optimizer behavior for Interlocked.Exchange with short value
**NOTE:** _Most likely this does not just affect Interlocked.Exchange, but it's easy to reproduce this way._

When turning on the optimization for the following code you get different (and also **wrong**) output values than from building the application without turning on the optimizer.

```c#
using System;
using System.Threading;

class Program
{
    private static int _intToExchange = -1;
    private static short _innerShort = 2;

    static void Main( string[] args )
    {
        var oldValue = Interlocked.Exchange(ref _intToExchange, _innerShort);
        Console.WriteLine( "It was:   {0}", oldValue );
        Console.WriteLine( "It is:    {0}", _intToExchange );
        Console.WriteLine( "Expected: {0}", _innerShort );
    }
}
```

Without using optimization the output is as expected:
```
It was:   -1
It is:    2
Expected: 2
```

But when you turn on optimization (e.g. with `<Optimize>true</Optimize>` in .csproj) this is the output:
```
It was:   65535
It is:    -65534
Expected: 2
```

More about this bug can also be found here on [stackoverflow](http://stackoverflow.com/q/43206775/503059).

mikedn:
The interlocked operation type can, at least in theory, be `TYP_VOID` in the case of `GT_LOCKADD`. The first operand type is not relevant because it's the address of the memory location and not an indir.

The only option is to use the type of the second operand but that may be a small int type so `genActualType` needs to be used to get the correct type.

Fixes #10714

### DIFF
--- a/src/jit/codegenxarch.cpp
+++ b/src/jit/codegenxarch.cpp
@@ -3826,6 +3826,7 @@ void CodeGen::genLockedInstructions(GenTreeOp* treeNode)
     regNumber   targetReg = treeNode->gtRegNum;
     regNumber   dataReg   = data->gtRegNum;
     regNumber   addrReg   = addr->gtRegNum;
+    var_types   type      = genActualType(data->TypeGet());
     instruction ins;
 
     // The register allocator should have extended the lifetime of the address
@@ -3840,7 +3841,7 @@ void CodeGen::genLockedInstructions(GenTreeOp* treeNode)
     genConsumeOperands(treeNode);
     if (targetReg != REG_NA && dataReg != REG_NA && dataReg != targetReg)
     {
-        inst_RV_RV(ins_Copy(data->TypeGet()), targetReg, dataReg);
+        inst_RV_RV(ins_Copy(type), targetReg, dataReg);
         data->gtRegNum = targetReg;
 
         // TODO-XArch-Cleanup: Consider whether it is worth it, for debugging purposes, to restore the
@@ -3866,8 +3867,8 @@ void CodeGen::genLockedInstructions(GenTreeOp* treeNode)
 
     // all of these nodes implicitly do an indirection on op1
     // so create a temporary node to feed into the pattern matching
-    GenTreeIndir i = indirForm(data->TypeGet(), addr);
-    getEmitter()->emitInsBinary(ins, emitTypeSize(data), &i, data);
+    GenTreeIndir i = indirForm(type, addr);
+    getEmitter()->emitInsBinary(ins, emitTypeSize(type), &i, data);
 
     if (treeNode->gtRegNum != REG_NA)
     {
--- a/tests/src/JIT/Regression/JitBlue/GitHub_10714/GitHub_10714.cs
+++ b/tests/src/JIT/Regression/JitBlue/GitHub_10714/GitHub_10714.cs
@@ -1,0 +1,25 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+class GitHub_10714
+{
+    const int Passed = 100;
+    const int Failed = 0;
+    
+    static int intToExchange = -1;
+    static short innerShort = 2;
+    
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    static int Test() => Interlocked.Exchange(ref intToExchange, innerShort);
+
+    static int Main()
+    {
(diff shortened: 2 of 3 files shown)


mikedn:
cc @dotnet/jit-contrib

pgavlin:
LGTM. Any diffs outside of the repro case you added?

mikedn:
Nope, no diffs.

janvorli:
@mikedn your newly added test has failed in the CI: https://ci.dot.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_flow_prtest/5128/

mikedn:
So the test passed during my PR test [run](https://ci.dot.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_tst_prtest/4443/testReport/) and failed in another PR test [run](https://ci.dot.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_tst_prtest/4480/testReport/)

Hrm, looks like random failure, I'll have to try to reproduce it.

mikedn:
The code generated on Linux is slightly different but it's correct an the difference is not related to this fix:
```asm
G_M52595_IG02:
       48BF805E1171FA7F0000 mov      rdi, 0x7FFA71115E80
       BE01000000           mov      esi, 1
       E800000000           call     CORINFO_HELP_GETSHARED_NONGCSTATIC_BASE
       48B8B45E1171FA7F0000 mov      rax, 0x7FFA71115EB4
; this is different on Windows, presumably due to differences in relocation
       48BFB85E1171FA7F0000 mov      rdi, 0x7FFA71115EB8
       480FBF3F             movsx    rdi, word  ptr [rdi]
       8738                 xchg     dword ptr [rax], edi
       8BC7                 mov      eax, edi
```
vs. Windows generated code:
```asm
G_M52595_IG02:
       48B9805E1271FA7F0000 mov      rcx, 0x7FFA71125E80
       BA01000000           mov      edx, 1
       E81813755F           call     CORINFO_HELP_GETSHARED_NONGCSTATIC_BASE
       48B8B45E1271FA7F0000 mov      rax, 0x7FFA71125EB4
       480FBF15BE59E9FF     movsx    rdx, word  ptr [reloc classVar[0x71126080]]
       8710                 xchg     dword ptr [rax], edx
       8BC2                 mov      eax, edx
```
Could it be that the tests in the other PR ran using a JIT without this fix? This PR was merge at 2:28 and in the failing PR build log I see:
```
02:25:24     Build dotnet_coreclr » master » checked_ubuntu_prtest #5197 started
02:31:25     Build dotnet_coreclr » master » checked_windows_nt_bld_prtest #7756 started
```
So the runtime build started before this was merged and the test build started after this was merged.

janvorli:
@mikedn I guess you are right and that test run used the sources before your fix. I think that the reason is that the tests are built on Windows, so the new test was incorporated but the change in the JIT was not. 
Anyways, looking at the recent test results in the CI, all the tests are passing, so everything is ok now.

briansull:
Actually looking at the code I am a bit worried that the x64 Linux code could be wrong:

It looks like it  may just be using the lower 32-bits of a full 64-bit address:

 480FBF15BE59E9FF     movsx    rdx, word  ptr [reloc classVar[0x71126080]]

If this is the case then I would expect the code to A/V on this instruction.

On Windows we get an indirection of the full 64-bit address and the address use looks like it is in the same module as the other 64-bit address:
       48B8B45E1171FA7F0000 mov      rax, 0x7FFA71115EB4
       48BFB85E1171FA7F0000 mov      rdi, 0x7FFA71115EB8

janvorli:
@briansull the rdx is not the address, but the value to be stored.

briansull:
@janvorli yes, but that value is loaded from an absolute address that is adjacent to the one  used in the xchg instruction. (At least on the Windows version)

mikedn:
> It looks like it may just be using the lower 32-bits of a full 64-bit address:

Isn't that a RIP-relative displacement?

briansull:
@mikedn  I suppose that that is possible, if the test executes correctly then it is OK.

mikedn:
I checked, it's RIP-relative. The ModRM byte is 0x15 = 00010101 so mod = 00 and r/m = 101 as specified in the Intel manual.

Also, `movsx rdx, word ptr [reloc classVar[0x71126080]]` is generated on Windows, not on Linux. Linux builds do not generate this kind of code because [`IsPreferredExecutableRange`](https://github.com/dotnet/coreclr/blob/d8e91d3febea933dce2bb2e0825ffcea1c76480d/src/utilcode/util.cpp#L463) always return false.

PR #10740, opened 2017-04-05
Revert EncodingForwarder for the sake of the Perf

### LINKED ISSUE #10705: Eliminate EncodingForwarder performance impact
The cost introduced by the ```EncodingForwarder``` in cases when you dont even need it is way too punishing for applications that have to deal with lots of strings and encoding. In our measurements the cost of Encoding has a non-negligible cost (to the point that if not solved we will eventually will roll our own). 

This is an example of the cost:
utf8

The funny fact is that on construction we do know it is a ```UTF8Encoding```. 
```csharp
private static readonly UTF8Encoding Encoding = new UTF8Encoding();
```

The virtual call itself will probably will be tackled by the devirtualization work @AndyAyersMS is doing, but the ```EncodingForwarder``` will continue be there and it is a sizeable cost by itself.

tarekgh:
Fixes #10705


tarekgh:
@dotnet-bot help

tarekgh:

tarekgh:
CC @jkotas @jamesqo

jamesqo:
It seems unfortunate that 900 LOC has to be added back, but if `EncodingForwarder` was slowing perf meaningfully then :+1:

jkotas:
@tarekgh Could you please submit a matching PR to CoreRT as well?

src/mscorlib/src/System/Text/UTF32Encoding.cs
```diff
@@ -84,43 +84,108 @@ internal override void SetDefaultFallbacks()
...
+            Contract.EndContractBlock();
+
+            int byteCount = bytes.Length - byteIndex;
+
+            // Fix our input array if 0 length because fixed doesn't like 0 length arrays
+            if (bytes.Length == 0)
+                bytes = new byte[1];
+
+            fixed (char* pChars = s)
+                fixed ( byte* pBytes = bytes)
```
jkotas: There was optimization done in the EncodingForwarder version to use `bytes[0]` here, and other similar places. It would be nice to preserve it.

tarekgh:

tarekgh:
@jkotas I returned back the bytes[0] optimization and I am going to submit another PR for corert first thing in the morning tomorrow

tarekgh:

tarekgh:
@jkotas are you OK to merge it?

tarekgh:
OSX10.12 is know issue for timing out and slowness is not related to the changes here

PR #10741, opened 2017-04-05
Strip out unused reg code.

JeremyKuhne:
Removing the unused code to better be able to audit our CoreLib usage and to prepare for targeted code sharing with CoreFX. Wraps key usages in using statements.


tarekgh:
CC @Priya91

src/mscorlib/src/Microsoft/Win32/RegistryKey.cs
```diff
@@ -232,77 +137,44 @@ public void DeleteValue(String name, bool throwOnMissingValue)
...
-         */
+        /// <summary>
+        /// Retrieves a subkey or null if the operation failed.
+        /// </summary>
+        /// <param name="writable">True to open writable, otherwise opens the key read-only.</param>
         public RegistryKey OpenSubKey(string name, bool writable)
         {
             ValidateKeyName(name);
             EnsureNotDisposed();
-            name = FixupName(name); // Fixup multiple slashes to a single slash
```
tarekgh: > FixupName(name);  [](start = 18, length = 18)

looks we were supporting names with slashes. don't we need to keep this behavior? I guess we'll have different behavior than the desktop. right?

JeremyKuhne:
Note that another goal here is to focus any additional usages we might introduce (hopefully none).

src/mscorlib/src/Microsoft/Win32/RegistryKey.cs
JeremyKuhne: As it is internal only for this copy of the code I felt it was better to fail hard here and cut out a ton of code. This just allows you to specify keys with double separators and ending separators.

src/mscorlib/src/Microsoft/Win32/RegistryKey.cs
```diff
@@ -481,22 +318,6 @@ public unsafe string[] GetValueNames()
...
-         * doesn't exist.
-         *
-         * Note that <var>name</var> can be null or "", at which point the
-         * unnamed or default value of this Registry key is returned, if any.
-         *
-         * @param name Name of value to retrieve.
-         *
-         * @return the data associated with the value.
-         */
-        public Object GetValue(String name)
```
tarekgh: >public Object GetValue(String name) [](start = 8, length = 35)

we don't expose such APIs?

src/mscorlib/src/Microsoft/Win32/RegistryKey.cs
JeremyKuhne: Again, ideally we'd factor the CoreFX code to allow sharing the subset we need for CoreLib.

src/mscorlib/src/Microsoft/Win32/RegistryKey.cs
JeremyKuhne: Not from here. CoreFX has a completely independent and rewritten Registry implementation.

tarekgh: I see you have the version with the default parameters so looks this will be covered there but do you think this will work for reflection too?

---
In reply to: [110039278](https://github.com/dotnet/coreclr/pull/10741#discussion_r110039278) [](ancestors = 110039278)

tarekgh: got it. thanks

---
In reply to: [110039485](https://github.com/dotnet/coreclr/pull/10741#discussion_r110039485) [](ancestors = 110039485)

JeremyKuhne: This registry code isn't publicly exposed from CoreLib.

tarekgh:

JeremyKuhne:
The other thing I was trying to accomplish here is (pending code-sharing) being able to better deal with any additional fixes that come in for the "real" implementation in CoreFX.

JeremyKuhne:
@tarekgh Thanks for kicking off the other runs

src/mscorlib/src/System/Globalization/HijriCalendar.Win32.cs
```diff
@@ -42,54 +42,41 @@ private static int GetAdvanceHijriDate()
...
-            try
+            using (key = Registry.CurrentUser.OpenSubKey(InternationalRegKey, writable: false))
             {
-                // Open in read-only mode.
-                // Use InternalOpenSubKey so that we avoid the security check.
-                key = RegistryKey.GetBaseKey(RegistryKey.HKEY_CURRENT_USER).OpenSubKey(InternationalRegKey, false);
-            }
-            //If this fails for any reason, we'll just return 0.
-            catch (ObjectDisposedException) { return 0; }
-            catch (ArgumentException) { return 0; }
```
tarekgh: this could be a behavior change as we used to continue if we get ObjectDisposedException exceptions but now we'll throw.

tarekgh:
LGTM

src/mscorlib/src/System/Globalization/HijriCalendar.Win32.cs
JeremyKuhne: Yeah, but that should never happen.

jkotas:
> prepare for targeted code sharing with CoreFX

There is another copy of the registry in CoreRT. You may want to move it to the shared CoreLib partition to get sharing between CoreRT and CoreCLR for cheap.

JeremyKuhne:
> There is another copy of the registry in CoreRT. You may want to move it to the shared CoreLib partition to get sharing between CoreRT and CoreCLR for cheap.

@jkotas Good point- I'm going to create an issue about sharing the code and will call that pivot out.

JeremyKuhne:

JeremyKuhne:

JeremyKuhne:

PR #10743, opened 2017-04-05
Remove stress jobs that take CI cycles without adding value

RussKeldorph:
THIS IS JUST TO GET CI VALIDATION FOR A DEV BRANCH.  THIS IS NOT GOING INTO MASTER.

We should get enough coverage in this branch without these jobs running.


RussKeldorph:
@dotnet-bot test ci please

PR #10746, opened 2017-04-05
Remove stress jobs that take CI cycles without adding value

RussKeldorph:
We should get enough coverage in this branch without these jobs running.


RussKeldorph:
Note that this is not a pull request against master.

PR #10747, opened 2017-04-05
Make Encoding.Default return UTF8 without BOM emission

### LINKED ISSUE #10643: Change Encoding.Default to UTF-8 *without BOM* for Unix interoperability
This is a follow-up from [here](https://github.com/dotnet/standard/issues/260#issuecomment-290149417).

* `Encoding.Default` in the full framework on Windows reports the legacy "ANSI" code page (encoding).

* Core doesn't support these legacy pages (by default) so its `Encoding.Default` must return an encoding that makes sense in a cross-platform world.

* Currently, `Encoding.UTF8` is returned, which is a UTF-8 encoding _with BOM_.
  * The reason for returning `Encoding.UTF8` as `Encoding.Default` is a historic one, it seems, as the [source-code comment](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Text/Encoding.cs#L1235) mentions _Silverlight_(!).

* **UTF-8 with BOM is problematic on all Unix platforms, whose utilities do not expect a BOM and instead treat it as _data_, leading to unexpected results.**

* **Having `Encoding.Default` return a UTF-8 encoding _with BOM_ is pointless and confusing**, because it neither represents any platform's nor the framework's own true default, and may cause disruption in practice.

   * Note that both .NET Framework's and .NET Core's _true default_ is _BOM-less_ UTF-8, because that's what's used _in the absence of specifying an encoding_ with the [methods of the `System.IO` class](https://msdn.microsoft.com/en-us/library/system.io.file_methods%28v=vs.110%29.aspx).

tarekgh:
Fixes #10643

Encoding.Default today is returning UTF8 but with BOM emission enabled. This is causing many problem when this encoding used. for example in Console we ensure turning of BOM emission. In System IO we always use the default with no BOM. we got the feedback from Powershell team too.
Also most of the tools (especially on Linux) doesn't handle BOM and treat it as part of the text. 

https://github.com/dotnet/standard/issues/260

This is kind of breaking change for netcore but I am expecting this will be very limiting breaking as I am seeing most of the time people trying to do turn off the BOM.

### DIFF
--- a/src/mscorlib/src/System/Text/Encoding.cs
+++ b/src/mscorlib/src/System/Text/Encoding.cs
@@ -1230,7 +1230,7 @@ namespace System.Text
 
             Encoding enc;
 
-            // For silverlight and netcore we use UTF8 since ANSI isn't available
+            // For netcore we use UTF8 since ANSI isn't available
             enc = UTF8Encoding.s_defaultNoBOM;
 
             // This method should only ever return one Encoding instance
--- a/src/mscorlib/src/System/Text/Encoding.cs
+++ b/src/mscorlib/src/System/Text/Encoding.cs
@@ -1231,7 +1231,7 @@ namespace System.Text
             Encoding enc;
 
             // For netcore we use UTF8 since ANSI isn't available
-            enc = UTF8Encoding.s_defaultNoBOM;
+            enc = new UTF8Encoding.UTF8EncodingSealed(encoderShouldEmitUTF8Identifier: false);
 
             // This method should only ever return one Encoding instance
             return Interlocked.CompareExchange(ref defaultEncoding, enc, null) ?? enc;
--- a/src/mscorlib/src/System/Text/Encoding.cs
+++ b/src/mscorlib/src/System/Text/Encoding.cs
@@ -1230,10 +1230,8 @@ namespace System.Text
 
             Encoding enc;
 
-
-            // For silverlight we use UTF8 since ANSI isn't available
-            enc = UTF8;
-
+            // For silverlight and netcore we use UTF8 since ANSI isn't available
+            enc = UTF8Encoding.s_defaultNoBOM;
 
             // This method should only ever return one Encoding instance
             return Interlocked.CompareExchange(ref defaultEncoding, enc, null) ?? enc;
--- a/src/mscorlib/src/System/Text/UTF8Encoding.cs
+++ b/src/mscorlib/src/System/Text/UTF8Encoding.cs
@@ -58,11 +58,13 @@ namespace System.Text
         internal sealed class UTF8EncodingSealed : UTF8Encoding
         {
             public UTF8EncodingSealed() : base(encoderShouldEmitUTF8Identifier: true) { }
+            public UTF8EncodingSealed(bool encoderShouldEmitUTF8Identifier) : base(encoderShouldEmitUTF8Identifier) { }
         }
 
         // Used by Encoding.UTF8 for lazy initialization
         // The initialization code will not be run until a static member of the class is referenced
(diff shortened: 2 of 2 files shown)


tarekgh:
CC @danmosemsft @weshaggard @stephentoub  @jkotas

tarekgh:

src/mscorlib/src/System/Text/Encoding.cs
```diff
@@ -1230,10 +1230,8 @@ private static Encoding CreateDefaultEncoding()
 
             Encoding enc;
 
-
-            // For silverlight we use UTF8 since ANSI isn't available
-            enc = UTF8;
-
+            // For silverlight and netcore we use UTF8 since ANSI isn't available
```
jkotas: You can delete the reference to Silverlight while you are on it.

tarekgh:

tarekgh:
test Windows_NT x64 Debug Build and Test please

src/mscorlib/src/System/Text/UTF8Encoding.cs
```diff
@@ -58,11 +58,13 @@ public class UTF8Encoding : Encoding
         internal sealed class UTF8EncodingSealed : UTF8Encoding
         {
             public UTF8EncodingSealed() : base(encoderShouldEmitUTF8Identifier: true) { }
```
stephentoub: Do we still need this ctor, or could the s_default initialization just be `new UTF8EncodingSealed(encoderShouldEmitUTF8Identifier: true)?

src/mscorlib/src/System/Text/UTF8Encoding.cs
```diff
@@ -58,11 +58,13 @@ public class UTF8Encoding : Encoding
...
+            public UTF8EncodingSealed(bool encoderShouldEmitUTF8Identifier) : base(encoderShouldEmitUTF8Identifier) { }
         }
 
         // Used by Encoding.UTF8 for lazy initialization
         // The initialization code will not be run until a static member of the class is referenced
         internal static readonly UTF8EncodingSealed s_default = new UTF8EncodingSealed();
+        internal static readonly UTF8EncodingSealed s_defaultNoBOM = new UTF8EncodingSealed(encoderShouldEmitUTF8Identifier: false);
```
jkotas: This static cache should not be needed. The static cache in Encoding.cs is sufficient.

tarekgh:
Thanks @stephentoub and @jkotas, I have removed the static as suggested.

src/mscorlib/src/System/Text/Encoding.cs
```diff
@@ -1230,10 +1230,8 @@ private static Encoding CreateDefaultEncoding()
 
             Encoding enc;
 
-
-            // For silverlight we use UTF8 since ANSI isn't available
-            enc = UTF8;
-
+            // For netcore we use UTF8 since ANSI isn't available
+            enc = new UTF8Encoding.UTF8EncodingSealed(encoderShouldEmitUTF8Identifier: false);
```
benaadams: Rather than initializing via an interlocked; would it be easier to make the type fixed and more visible for devirtualization; as per UTF8Encoding?

```csharp
private static readonly UTF8EncodingSealed s_defaultEncoding
  = new UTF8EncodingSealed(encoderShouldEmitUTF8Identifier: false);
public static Encoding Default => s_defaultEncoding;
```

Since its the only static var in the class it shouldn't be initialized when its not used? (e.g. some other static var is accessed which causes its initialization)

jkotas: Yes, this would make sense. @benaadams Could you please submit PR?

PR #10748, opened 2017-04-05
[Local GC] Fix a bad merge

swgillespie:
Unfortunately in my last PR I made an error during a rebase that resulted in some unused code that was deleted in a now-squashed commit leaking back in.


swgillespie:
thanks!

PR #10749, opened 2017-04-05
[x86/Linux] Fix ResumeEsp before resume

parjong:
This commit fixes RseumeEsp just before resuming, which resolves 4 test failures:
 - JIT/IL_Conformance/Old/Conformance_Base/conv_ovf_r8_i
 - JIT/IL_Conformance/Old/Conformance_Base/conv_ovf_r8_i4
 - JIT/Methodical/Arrays/misc/_il_relinitializearray
 - JIT/Regression/VS-ia64-JIT/V1.2-M01/b12390

### DIFF
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -4033,6 +4033,50 @@ bool UnwindStackFrame(PREGDISPLAY     pContext,
 
 #endif // _TARGET_X86_
 
+#ifdef WIN64EXCEPTIONS
+#ifdef _TARGET_X86_
+size_t EECodeManager::GetResumeSp( PREGDISPLAY  pRD )
+{
+    EECodeInfo codeInfo(pRD->ControlPC);
+
+    PTR_CBYTE methodStart = PTR_CBYTE(codeInfo.GetSavedMethodCode());
+
+    GCInfoToken gcInfoToken = codeInfo.GetGCInfoToken();
+    PTR_VOID    methodInfoPtr = gcInfoToken.Info;
+    DWORD       curOffs = codeInfo.GetRelOffset();
+
+    CodeManStateBuf stateBuf;
+
+    stateBuf.hdrInfoSize = (DWORD)DecodeGCHdrInfo(gcInfoToken,
+                                                  curOffs,
+                                                  &stateBuf.hdrInfoBody);
+
+    PTR_CBYTE table = dac_cast<PTR_CBYTE>(methodInfoPtr) + stateBuf.hdrInfoSize;
+
+    hdrInfo *info = &stateBuf.hdrInfoBody;
+
+    _ASSERTE(info->epilogOffs == hdrInfo::NOT_IN_EPILOG && info->prologOffs == hdrInfo::NOT_IN_PROLOG);
+
+    bool isESPFrame = !info->ebpFrame && !info->doubleAlign;
+
+    if (codeInfo.IsFunclet())
+    {
+        // Treat funclet's frame as ESP frame
+        isESPFrame = true;
+    }
+
+    if (isESPFrame)
+    {
+        const size_t curESP = pRD->SP;
+        return curESP + GetPushedArgSize(info, table, curOffs);
+    }
+
+    const size_t curEBP = *pRD->GetEbpLocation();
+    return GetOutermostBaseFP(curEBP, info);
+}
+#endif // _TARGET_X86_
+#endif // WIN64EXCEPTIONS
+
 #ifndef CROSSGEN_COMPILE
 #ifndef WIN64EXCEPTIONS
 
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -4035,9 +4035,13 @@ bool UnwindStackFrame(PREGDISPLAY     pContext,
 
 #ifdef WIN64EXCEPTIONS
 #ifdef _TARGET_X86_
-size_t EECodeManager::GetResumeSp( PREGDISPLAY  pRD )
+size_t EECodeManager::GetResumeSp( PCONTEXT  pContext )
 {
-    EECodeInfo codeInfo(pRD->ControlPC);
+    PCODE currentPc = PCODE(pContext->Eip);
+
+    _ASSERTE(ExecutionManager::IsManagedCode(currentPc));
+
+    EECodeInfo codeInfo(currentPc);
 
     PTR_CBYTE methodStart = PTR_CBYTE(codeInfo.GetSavedMethodCode());
 
@@ -4067,11 +4071,11 @@ size_t EECodeManager::GetResumeSp( PREGDISPLAY  pRD )
 
     if (isESPFrame)
     {
-        const size_t curESP = pRD->SP;
+        const size_t curESP = (size_t)(pContext->Esp);
         return curESP + GetPushedArgSize(info, table, curOffs);
     }
 
-    const size_t curEBP = *pRD->GetEbpLocation();
+    const size_t curEBP = (size_t)(pContext->Ebp);
     return GetOutermostBaseFP(curEBP, info);
 }
 #endif // _TARGET_X86_
--- a/src/vm/exceptionhandling.cpp
+++ b/src/vm/exceptionhandling.cpp
@@ -1200,10 +1200,10 @@ ProcessCLRException(IN     PEXCEPTION_RECORD   pExceptionRecord
 
                 pThread->SetFrame(pLimitFrame);
 
+                FixContext(pContextRecord);
+
                 SetIP(pContextRecord, (PCODE)uResumePC);
             }
-
-            FixContext(pContextRecord);
 
 #ifdef STACK_GUARDS_DEBUG
             // We are transitioning back to managed code, so ensure that we are in
--- a/src/vm/exceptionhandling.cpp
+++ b/src/vm/exceptionhandling.cpp
@@ -123,6 +123,24 @@ bool FixNonvolatileRegisters(UINT_PTR  uOriginalSP,
                              bool      fAborting
                              );
 
+void FixContext(PCONTEXT pContextRecord)
+{
+#define FIXUPREG(reg, value)                                                                \
+    do {                                                                                    \
+        STRESS_LOG2(LF_GCROOTS, LL_INFO100, "Updating " #reg " %p to %p\n",                 \
+                pContextRecord->reg,                                                        \
+                (value));                                                                   \
+        pContextRecord->reg = (value);                                                      \
+    } while (0)
+
+#ifdef _TARGET_X86_
+    size_t resumeSp = EECodeManager::GetResumeSp(pContextRecord);
+    FIXUPREG(ResumeEsp, resumeSp);
+#endif // _TARGET_X86_
+
+#undef FIXUPREG
+}
+
 MethodDesc * GetUserMethodForILStub(Thread * pThread, UINT_PTR uStubSP, MethodDesc * pILStubMD, Frame ** ppFrameOut);
 
 #ifdef FEATURE_PAL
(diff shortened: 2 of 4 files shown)


parjong:
\CC @YongseopKim

parjong:
@jkotas @janvorli Is there any issue related with OSX CI?

janvorli:
@parjong yes, there is some issue with network connection speed to the OSX machines. I believe people are looking into it.

parjong:
Thanks you for reply :+1:

src/vm/eetwain.cpp
```diff
@@ -4033,6 +4033,50 @@ bool UnwindStackFrame(PREGDISPLAY     pContext,
...
+    }
+
+    if (isESPFrame)
+    {
+        const size_t curESP = pRD->SP;
+        return curESP + GetPushedArgSize(info, table, curOffs);
+    }
+
+    const size_t curEBP = *pRD->GetEbpLocation();
+    return GetOutermostBaseFP(curEBP, info);
```
janvorli: Does this work correctly for the cases where we have added stack alignment before pushing the arguments?

parjong: As I understand jitter records all push/pop for ESP frame, and thus this approach works for ESP frames.
For EBP frame, jitter reports frame size, and thus this approach seems to work as we could derive resume SP from ebp and frame size.

I am currently not sure about funclets in ebp framed function.

src/vm/i386/cgenx86.cpp
```diff
@@ -384,7 +384,7 @@ void HelperMethodFrame::UpdateRegDisplay(const PREGDISPLAY pRD)
 #endif // DACCESS_COMPILE
 
     pRD->pCurrentContext->Eip = pRD->ControlPC = m_MachState.GetRetAddr();
-    pRD->pCurrentContext->Esp = pRD->pCurrentContext->ResumeEsp = pRD->SP = (DWORD) m_MachState.esp();
+    pRD->pCurrentContext->Esp = pRD->SP = (DWORD) m_MachState.esp();
```
janvorli: Should we make the same change in TailCallFrame::UpdateRegDisplay and HijackFrame::UpdateRegDisplay? They set the Esp and ResumeEsp the same way too.

src/vm/exceptionhandling.cpp
```diff
@@ -441,13 +441,24 @@ void ExceptionTracker::UpdateNonvolatileRegisters(CONTEXT *pContextRecord, REGDI
...
+
+
 #if defined(_TARGET_X86_)
 
     UPDATEREG(Ebx);
     UPDATEREG(Esi);
     UPDATEREG(Edi);
     UPDATEREG(Ebp);
 
+    UPDATEREG_VAL(ResumeEsp);
```
janvorli: I am not excited about the updating of ResumeEsp at this place since it is a not a non-volatile register and the function is called UpdateNonvolatileRegisters. Unfortunately the only other place where we have the necessary data to update the ResumeESP is the caller and that's called FixNonvolatileRegisters, so it would not be any better. And I don't really want to rename the functions.

src/vm/i386/cgenx86.cpp
parjong: I think so, but I would like to isolate those changes from this change.

src/vm/exceptionhandling.cpp
parjong: We may revise ``GetResumeSp`` to take ``CONTEXT`` instead ``REGDISPLAY``.

Then, we could fix ``ResumeEsp`` inside ProcessCLRException just before calling ``ExceptionTracker::ResumeExecution``.

src/vm/eetwain.cpp
janvorli: Can you verify that it works for the funclets too? I have looked for a good candidate of a fcall that throws and can be easily forced to throw and has at least one stack argument and found ArrayNative::ArrayClear. It throws if the passed iIndex is below the lower bound of the array or if the iIndex + iLength is over the end of the array.
It looks like this FCall is invoked when you call Array.Clear static method.
So if you write a little test app and placed the following piece of code into a funclet, you should be able to put a breakpoint inside the `if (codeInfo.IsFunclet())` in the EECodeManager::GetResumeSp and then step through your code to see if it works as expected.
```C#
byte[] b = new byte[4];
Array.Clear(b, 0, 8);
```

src/vm/exceptionhandling.cpp
janvorli: Great, that sounds much better to me! And maybe we can just change the method to set the ResumeEsp internally in the context instead of returning it back. And maybe rename the method to something like FixResumeEsp then.

src/vm/eetwain.cpp
parjong: I'm sorry for unclear comment. 

I already found a case that ``GetPushedArgSize`` becomes incorrect while working on #10685. I tried to set ``baseSP`` as ``curESP + GetPushedArgSize(...)`` instead of ``pContext->pCurrentContext->ResumeEsp``, but ``JIT.Methodical.flowgraph.dev10_bug679008.fgloop`` fails with that approach.

It seems that we need to emit full arg info for all managed methods in order to fix this case. I'm not sure about the amount of work for this task.

src/vm/i386/cgenx86.cpp
```diff
@@ -402,6 +402,11 @@ void HelperMethodFrame::UpdateRegDisplay(const PREGDISPLAY pRD)
 
     ClearRegDisplayArgumentAndScratchRegisters(pRD);
 
+    //
+    // Fix up ResumeSp
+    //
+    pRD->pCurrentContext->ResumeEsp = EECodeManager::GetResumeSp(pRD);
```
jkotas: This assumes that the FCall is always called from JITed code. It is not always the case. For example, the FCall can be called by CallDescrWorker.

parjong: It is possible, but does not matter as we use ResumeEsp only when resuming from managed code.

janvorli: I think it will be more obvious after you make the change that you've suggested.

src/vm/exceptionhandling.cpp
parjong: It think that we first need to guarantee the correctness of ``GetPushedArgSize`` before implementing this change. I'll take a look.

src/vm/eetwain.cpp
janvorli: I have a quick idea. Maybe we can solve it in an easy way, based on the fact that we know that a funclet is called from native code and we know what the prolog of a funclet looks like. Since funclet cannot allocate any locals, we know that the ESP we are looking for is the ESP after the prolog of the funclet, which is ESP at which the asm helper called the funclet minus the return address size and that aligned to 16 bytes.

parjong: I'm not sure whether we could get ``ESP at which the asm helper called`` correctly.

parjong: I submitted #10782 to address this issue. Could you please take a look?

janvorli: I have quickly looked at the #10782, but my first impression is that it should be our last resort if there is no other way. I feel uneasy about adding the full argument info even for EBP frames. But I'll think about it more.
But as for my idea how to solve the funclet issue differently, the CallEHFunclet and CallEHFilterFunclet records the funclet caller SP in ExceptionTracker::m_EHClauseInfo.GetCallerStackFrameForEHClauseReference(). So in ExceptionTracker::ResumeExecution, we could read that value from the right exception tracker and pass it to the EECodeManager::GetResumeSp. If there are no issues that I don't see at this late time at night, I think that it would work.

parjong: I guess that the suggested approach may work for resuming, but I'm not sure whether this approach could address incorrect unwinding during GC and 1st pass EH.

Could you let me know your opinion?

parjong: One approach to optimize #10782 is to emit full arg info only for functions with funclet. Is it possible to add some flag on GCInfo header (to indicate arg info type)?

parjong:
@janvorli Revised PR as I mentioned, but it introduces baseservices.compilerservices.RuntimeWrappedException regression. I'll take a look into this issue.

parjong:
955fbc2 resolves baseservices.compilerservices.RuntimeWrappedException regression.

janvorli:
> I guess that the suggested approach works for resuming, but I'm not sure whether this approach could address incorrect unwinding during GC and 1st pass EH.
@parjong it should work in all of these cases. Whenever there is a funclet on the stack, there must be an exception tracker corresponding to the exception that caused the funclet to be called. And that tracker contains the caller SP of the funclet. 
The tricky part though is how to find the right exception tracker. It seems the unwinder would need to keep track of a "current" tracker and whenever we hit a funclet frame during unwinding, we would move to the next tracker.

But I was thinking about it and it seems we have two more options to solve the issue:
1. Do it by adding the full args info even for EBP frames, but only explicitly for calls to FCALLs if that's possible. That would seem to be the ideal way with minimal growth of the GC info.
2. Use the idea that was discussed a long time ago and create a lookup table for stack args size for each FCALL. As was pointed out, that size is already passed to the macros that are used to create FCALL table, but currently ignored. So we could start using it and whenever unwind goes from native to managed frame, we would locate the appropriate FCALL's entry in the lookup and get the size from there.
3. Maybe we should focus on completing the FEATURE_FIXED_OUT_ARGS work instead. That would remove the need to solve this problem.

@parjong can you please tell me your opinions on these options?

parjong:
> 1. Do it by adding the full args info even for EBP frames, but only explicitly for calls to FCALLs if that's possible. That would seem to be the ideal way with minimal growth of the GC info.

Full arg info are required only for functions with funclets. FCALL itself does not matter. All the function with funclet is ebp-framed, but we need to treat its funclets as if it is esp-framed (as we can't depend on ebp when unwinding them). Actually, #10782 already introduced ``emitFullArgInfo`` to control when we emit full arg info. Currently, it is always true when ``fullPtrMap`` which means that function itself is fully interruptible (changes in ``morph.cpp`` enforces all functions with funclet to be fully interruptible), or esp-framed. It is easy to change this in JIT. Unfortunately, there is no way to distinguish ebp-frame with full  full arg info and ebp-frame with partial arg info from VM, which enforces #10782 to emit full arg info for all interruptible frames. We could optimize this if we could introduce an additional flag in GCInfo header that allows VM to distinguish two types of ebp-frame.

> 1. Use the idea that was discussed a long time ago and create a lookup table for stack args size for each FCALL. As was pointed out, that size is already passed to the macros that are used to create FCALL table, but currently ignored. So we could start using it and whenever unwind goes from native to managed frame, we would locate the appropriate FCALL's entry in the lookup and get the size from there.

I once considered this approach, but concluded that it will not work. Recently, it turns out that x86 JIT emits nested calls. The presence of nested calls make it impossible to estimate stack alignment padding without additional meta data such as GCInfo.

> 1. Maybe we should focus on completing the FEATURE_FIXED_OUT_ARGS work instead. That would remove the need to solve this problem.

I totally agree that enabling ``FEATURE_FIXED_OUT_ARGS`` will be the best solution for this issue, but I overhear that it will require non-trivial amount of work: JIT itself, GC Info encoder/decoder, and unwinder should be revised all together.

janvorli:
@parjong it seems I got confused by the title of this PR. So the problem you are solving is not just FCALLs in a funclet, but all calls made from a funclet? That when we unwind from any such call back to the funclet, we cannot get the correct resume ESP without a fix?

danmosemsft:

parjong:
@janvorli Sorry for the confusion. Yes, this PR attempts to address general incorrect resume esp issue due to stack alignment padding (via using GCInfo). This issue currently appears only for FCALL as managed unwinder uses a heuristic to adjust resume esp (which may be incorrect in the presence of nested calls). 

I guess that we will suffer from incorrect resume esp issue when we unwind from any nested call even for managed functions without a fix like this.

This PR is based on the assumption that ``GetPushedArgSize`` is correct for ESP-framed functions and funclets, but unfortunately it turns out that ``GetPushedArgSize`` is incorrect for funclets.

#10782 attempts to fix this issue (makes ``GetPushedArgSize`` correct for funclets).

PR #10755, opened 2017-04-06
[Local GC] Move knowledge of the free object method table outside of the GC

swgillespie:
The EE provides a method table for a specific type of object, the "free object", which is used by the GC to keep the heap crawlable. A free object is conceptually similar to a managed array in that it consists of an object header and a field indicating how many array elements the object has (i.e. how large the free region is), followed by a region of space known to be free to the GC.

This PR ensures that the GC's knowledge of the free method table is acquired through `GCToEEInterface` upon startup. This PR also has the GC acquire two additional details on startup: 1) the size of `ArrayBase`, and the 2) offset of the `NumComponents` field of `ArrayBase`. These are used to locate the offset of the field into which the size of the free object is stored and to locate the first byte of the free object's free region.

### DIFF
--- a/src/gc/env/gcenv.ee.h
+++ b/src/gc/env/gcenv.ee.h
@@ -73,6 +73,7 @@ public:
     static bool ShouldFinalizeObjectForUnload(AppDomain* pDomain, Object* obj);
     static bool ForceFullGCToBeBlocking();
     static bool EagerFinalized(Object* obj);
+    static MethodTable* GetFreeObjectMethodTable();
 };
 
 #endif // __GCENV_EE_H__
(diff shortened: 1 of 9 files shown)


src/gc/env/gcenv.ee.h
```diff
@@ -71,6 +71,9 @@ class GCToEEInterface
 
     static void HandleFatalError(unsigned int exitCode);
     static bool ShouldFinalizeObjectForUnload(AppDomain* pDomain, Object* obj);
+    static MethodTable* GetFreeObjectMethodTable();
+    static size_t GetSizeOfArrayBase();
```
jkotas: These constants are used on very perf sensitive paths. I do not think it is ok to virtualize them. They have to be part of the fixed binary contract between VM and GC.

swgillespie: These methods are called once on GC startup and cached on the EE side - is that still a problem?

jkotas: Yes.

swgillespie: Ok, will fix!

swgillespie:
cc @adityamandaleeka @Maoni0 @sergiy-k

jkotas:
Need to resolve conflicts ... LGTM otherwise.

swgillespie:
thanks!

PR #10756, opened 2017-04-06
don't switch to COOP GC when logging CCW ref count

yizhang82:
When logging CCW ref count messages in UWP applications, XAML may call AddRef inside a GC and switching to GC cooperative mode inside a GC would guarantee a deadlock. Fortunately in CoreCLR we can ignore the domain unloading scenario and avoid the domain holder. Since we are not in coop mode, we also can't use ObjectFromhandle - I had to dereference the handle pointer directly. This should be fine since we only use the object* as an ID loosely and it doesn't matter if it is super accurate, and we don't touch the Object* anyway. 

This is useful when diagnosing WinRT related leaks using PerfView. 
@jkotas @vancem  PTAL

### DIFF
--- a/src/vm/comcallablewrapper.cpp
+++ b/src/vm/comcallablewrapper.cpp
@@ -945,43 +945,43 @@ void SimpleComCallWrapper::BuildRefCountLogMessage(LPCWSTR wszOperation, StackSS
     }
     CONTRACTL_END;
 
-    GCX_COOP();
-
-    // There's no way how to get the class name if the AD is unloaded. We will just skip it if it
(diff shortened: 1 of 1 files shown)


vancem:
I think it is fine to fetch the object ref while in preemptive mode because we don't mind if it is perfectly accurate or not.    

Can you try running 

    PerfView /Providers=ClrPrivate collect 

On a scenario to confirm that it does not deadlock and that you get a 'CCW ref count' view?

Thanks for doing this.

yizhang82:
@dotnet-bot Test Windows_NT x64 Release Priority 1 Build and Test please

yizhang82:
@dotnet-bot Test OSX10.12 x64 Checked Build and Test  please

PR #10757, opened 2017-04-06
[Local GC] Move handle creation to IGCHandleTable

adityamandaleeka:
This change adds handle creation functions to the IGCHandleTable interface, and modifies the VM to use the interface for handle creation rather than calling functions in objecthandle.h directly.

Rather than adding a separate function for each `Create__<type>__Handle` function (of which there are many) to the interface, I consolidated them to a few major categories and added the `Create__<type>__Handle` functions to a header on the VM side that will call the appropriate thing on the interface. This minimizes changes to VM source while also not adding clutter to the interface.

I've also moved the handle-related code that was in gcheaputilities to its own file, since there is now a lot more code there.

Most of the handle creation functions are pretty straightforward, but there are a couple of interesting cases:
- I removed `CreateDuplicateHandle` entirely since it's only used in one place in the VM and instead modified the caller to achieve the same result with the other functionality we already have on the interface.
- Rather than adding an interface method specific to WinRT weak handles, I made `CreateHandleWithExtraInfo` which takes a `void*` that can be anything the user wants. `CreateWinRTWeakHandle` is now a convenience function that lives in the VM side and calls `CreateHandleWithExtraInfo` with the `IWeakReference` as the extra info.

@Maoni0 @jkotas @swgillespie @sergiy-k

### DIFF
--- a/src/gc/gchandletableimpl.h
+++ b/src/gc/gchandletableimpl.h
@@ -19,6 +19,12 @@ public:
     virtual void* GetHandleTableForHandle(OBJECTHANDLE handle);
 
     virtual OBJECTHANDLE CreateHandleOfType(void* table, Object* object, int type);
+
+    virtual OBJECTHANDLE CreateHandleWithExtraInfo(void* table, Object* object, int type, void* pExtraInfo);
+
+    virtual OBJECTHANDLE CreateDependentHandle(void* table, Object* primary, Object* secondary);
+
+    virtual OBJECTHANDLE CreateGlobalHandleOfType(Object* object, int type);
 };
 
 #endif  // GCHANDLETABLE_H_
--- a/src/gc/gchandletableimpl.h
+++ b/src/gc/gchandletableimpl.h
@@ -17,6 +17,8 @@ public:
     virtual void* GetHandleTableContext(void* handleTable);
 
     virtual void* GetHandleTableForHandle(OBJECTHANDLE handle);
+
+    virtual OBJECTHANDLE CreateHandleOfType(void* table, Object* object, int type);
 };
 
 #endif  // GCHANDLETABLE_H_
--- a/src/vm/appdomain.hpp
+++ b/src/vm/appdomain.hpp
@@ -1325,8 +1325,16 @@ public:
 
     OBJECTHANDLE CreateDependentHandle(OBJECTREF primary, OBJECTREF secondary)
     {
-        WRAPPER_NO_CONTRACT;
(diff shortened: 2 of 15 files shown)


src/gc/gchandletableimpl.h
```diff
@@ -17,6 +17,14 @@ class GCHandleTable : public IGCHandleTable
     virtual void* GetHandleTableContext(void* handleTable);
 
     virtual void* GetHandleTableForHandle(OBJECTHANDLE handle);
+
+    virtual OBJECTHANDLE CreateHandleOfType(void* table, Object* object, int type);
```
swgillespie: Should the definition of `OBJECTHANDLE` be on the interface? I see it's defined in `vars.hpp` and `gcenv.base.h` today.

src/vm/appdomain.hpp
```diff
@@ -1323,7 +1326,9 @@ class BaseDomain
     OBJECTHANDLE CreateDependentHandle(OBJECTREF primary, OBJECTREF secondary)
     {
         WRAPPER_NO_CONTRACT;
```
swgillespie: This isn't a wrapper anymore - can we use `::CreateDepehdentHandle`'s contract here?

adityamandaleeka:
(the failure is obviously not related to these changes)

src/vm/appdomain.hpp
```diff
@@ -1238,7 +1239,9 @@ class BaseDomain
     OBJECTHANDLE CreateTypedHandle(OBJECTREF object, int type)
     {
         WRAPPER_NO_CONTRACT;
-        return ::CreateTypedHandle(m_hHandleTableBucket->pTable[GetCurrentThreadHomeHeapNumber()], object, type);
+
+        IGCHandleTable *pHandleTable = GCHandleTableUtilities::GetGCHandleTable();
+        return pHandleTable->CreateHandleOfType(m_hHandleTableBucket->pTable[GetCurrentThreadHomeHeapNumber()], OBJECTREFToObject(object), type);
```
jkotas: HandleTableBucket seems to be implementation detail of the HandleTable. Shouldn't it move to the GC side as well?

Also, it does not look right that I have to call to GC two times to create the handle: first to get thread home number, and then to actually create the handle.

adityamandaleeka: Yes, that will be addressed in a future PR. The home heap number will not be exposed anymore.

jkotas:
LGTM - as an intermediate point.

adityamandaleeka:
Thanks!

PR #10764, opened 2017-04-06
Clean up and prepare FileLoadException.cs for move to Shared Partition.

AtsushiKan:


AtsushiKan:
cc @jkotas

src/mscorlib/src/System/IO/FileLoadException.cs
```diff
@@ -2,182 +2,113 @@
...
-            info.AddValue("FileLoad_FileName", _fileName, typeof(String));
-
-            try
-            {
-                info.AddValue("FileLoad_FusionLog", FusionLog, typeof(String));
-            }
-            catch (SecurityException)
+            info.AddValue("FileLoad_FileName", FileName, typeof(string));
+            string fusionLog = FusionLog;
+            if (fusionLog != null)
```
jkotas: Why to add only if it is non-null now when it was added unconditionally before? Adding it only if non-null means that there will exception thrown & caught during deserialization.

src/mscorlib/src/System/IO/FileLoadException.cs
```diff
@@ -2,182 +2,113 @@
...
-        protected FileLoadException(SerializationInfo info, StreamingContext context) : base(info, context)
+        protected FileLoadException(SerializationInfo info, StreamingContext context)
+            : base(info, context)
         {
             // Base class constructor will check info != null.
 
-            _fileName = info.GetString("FileLoad_FileName");
+            FileName = info.GetString("FileLoad_FileName");
 
             try
```
jkotas: You can remove the this try/catch if you add it unconditionally.

src/mscorlib/src/System/IO/FileLoadException.cs
AtsushiKan: Hmm... I think I unified the "missing because of security" to "missing because of non-null or security" in the original code. Fixed.

PR #10771, opened 2017-04-06
Forbid LEA without index AND base.

sandreenko:
GT_LEA could have null op1 and null op2. There is no check that at least one operator is presented during a creation process.

This case was not reflected in GenTreeUseEdgeIterator.
It caused an assert  failure on a test with a static pinned array.

```
[FixedAddressValueType]
static int arr[]={1,2,3};

int e = arr[1]; // constant base + constant index * constant size.
```
It required minopts to prevent constant folding during Value numbering.

Fix DevDiv 400702. I spent some time but was not able to create a small repro for it.

### DIFF
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -1869,26 +1869,26 @@ bool CodeGen::genCreateAddrMode(GenTreePtr  addr,
         The following indirections are valid address modes on x86/x64:
 
             [                  icon]      * not handled here
-            [reg                   ]      * not handled here
+            [reg                   ]
             [reg             + icon]
-            [reg2 +     reg1       ]
-            [reg2 +     reg1 + icon]
-            [reg2 + 2 * reg1       ]
-            [reg2 + 4 * reg1       ]
-            [reg2 + 8 * reg1       ]
-            [       2 * reg1 + icon]
-            [       4 * reg1 + icon]
-            [       8 * reg1 + icon]
-            [reg2 + 2 * reg1 + icon]
-            [reg2 + 4 * reg1 + icon]
-            [reg2 + 8 * reg1 + icon]
+            [reg1 +     reg2       ]
+            [reg1 +     reg2 + icon]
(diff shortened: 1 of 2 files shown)


sandreenko:
PTAL @dotnet/jit-contrib

sandreenko:
@dotnet-bot test tizen armel Cross Debug Build

pgavlin:
@sandreenko it seems odd that we are seeing an address tree of the form that you've described, especially if this occurs only on CoreRT. Are we sure that the base should be a compile-time constant that isn't a relocatable handle (which would not be constant-foldable)?

sandreenko:
It is not connected with CoreRT, it is RyuJit x86 problem.
@BruceForstall we need your opinion about this PR. 
Possible solutions are:
1. forbid addr modes without registers, return false in doAddrMode() if it happens and:
1.1. do constant folding in morph( it will positively affect more code rather then addrModes) ;
1.2 do not fold constant trees in minopts;
1.3 do local constant folding from value numbering in minOpts (the best variant in my opinion, but it requires the separate issue).
2. allow addr modes without registers (it is the current PR), but change EdgeIterator;
3. return doAddrMode == true for this case, but check that it is actual constant and create constant instead of LEA. (I do not like this variant, it makes code less clear).

pgavlin:
(3) is my preferred solution, as it is minimally invasive and does not change the semantics of `GT_LEA` (unlike option 2).

BruceForstall:
I'm a bit confused about your suggestions. For 1, are you suggesting having `genCreateAddrMode()` (not `doAddrMode()`, which isn't a function that I see) return false if the address mode it computes is purely a constant? That would be my preference. I don't see the need to change any constant folding, in minopts. The comment for `genCreateAddrMode()` already says it doesn't handle pure `[icon]` addressing modes, but it didn't expect to see an add that creates such a thing. Changing it to return false in this case seems to strengthen its existing contract.

danmosemsft:

sandreenko:
PR was updated. Please look again.

PR #10775, opened 2017-04-06
NS2.0 Prepare FileNotFoundException for move to shared partition.

AtsushiKan:
This is exactly what we did for FileLoadException - except
this one's also in the service of finishing off the IO
namespace in CoreRT.


AtsushiKan:
cc @jkotas

PR #10778, opened 2017-04-06
Call custom parameterless constructor on structs through Activator

### LINKED ISSUE #6843: Activator.CreateInstance fails to call constructor for value types
See https://github.com/dotnet/roslyn/issues/1029 and http://roslyn.codeplex.com/workitem/465 for some context.

> The issue is basically an optimization introduced in Activator.CreateInstance around CLR 4.0 and present ever since. The optimization assumes that parameterless instantiation of generic T type does not cause sideeffects if T is found to be a struct and therefore instances can be cached. Parameterless struct constructors (which are permitted by spec) violate such assumptions and make optimization observable.

kouvel:
Fixes #6843
- Disabled caching struct types that have custom parameterless constructors in `ActivatorCache`
- Removed some things relevant to security, which don't apply to CoreCLR

### DIFF
--- a/src/mscorlib/src/System/RtType.cs
+++ b/src/mscorlib/src/System/RtType.cs
@@ -4701,15 +4701,12 @@ namespace System
             internal volatile CtorDelegate m_ctor;
             internal readonly RuntimeMethodHandleInternal m_hCtorMethodHandle;
             internal readonly MethodAttributes m_ctorAttributes;
-            // Is a security check needed before this constructor is invoked?
-            internal readonly bool m_bNeedSecurityCheck;
             // Lazy initialization was performed
             internal volatile bool m_bFullyInitialized;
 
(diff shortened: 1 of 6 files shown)


gkhanna79:
What about https://github.com/dotnet/coreclr/issues/6843#issuecomment-241566379?

kouvel:
> What about #6843 (comment)?

Regarding:
> [This code](https://github.com/dotnet/coreclr/blob/00c85b302a078d5d13cfe9ff8cb453cd8296d6aa/src/mscorlib/src/System/RtType.cs#L5143) makes incorrect assumption that value types have no default constructor.

That would be incorrect when value types with custom parameterless constructors are cached. I didn't find a decent way to cache it and call the constructor properly through reflection, so opted to not cache it.

PR #10782, opened 2017-04-06
[x86/Linux] Use GCInfo for funclet unwinding

parjong:
This commit revises x86/Linux unwinder to use GCInfo(GetPushedArgSize) for funclet unwinding.

This commit includes the following additional changes:
 - Enforce fully interruptible codegen for functions with EH funclet
 - Enforce full arg info emission even for EBP frames

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -16204,11 +16204,6 @@ void Compiler::fgSetOptions()
     if (info.compXcptnsCount > 0)
     {
         codeGen->setFramePointerRequiredEH(true);
-#ifdef UNIX_X86_ABI
-        assert(!codeGen->isGCTypeFixed());
-        // Enforce fully interruptible codegen for funclet unwinding
-        genInterruptible = true;
-#endif // UNIX_X86_ABI
     }
 
 #else // !_TARGET_X86_
@@ -16219,6 +16214,15 @@ void Compiler::fgSetOptions()
     }
 
 #endif // _TARGET_X86_
+
+#ifdef UNIX_X86_ABI
+    if (info.compXcptnsCount > 0)
+    {
+        assert(!codeGen->isGCTypeFixed());
+        // Enforce fully interruptible codegen for funclet unwinding
+        genInterruptible = true;
+    }
+#endif // UNIX_X86_ABI
 
     fgCheckArgCnt();
 
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -16202,7 +16202,14 @@ void Compiler::fgSetOptions()
     // to use a frame pointer because of EH. But until all the code uses
     // the same test, leave info.compXcptnsCount here.
     if (info.compXcptnsCount > 0)
+    {
         codeGen->setFramePointerRequiredEH(true);
+#ifdef UNIX_X86_ABI
+        assert(!codeGen->isGCTypeFixed());
+        // Enforce fully interruptible codegen for funclet unwinding
+        genInterruptible = true;
+#endif // UNIX_X86_ABI
+    }
 
 #else // !_TARGET_X86_
 
--- a/src/vm/eetwain.cpp
+++ b/src/vm/eetwain.cpp
@@ -2413,16 +2413,16 @@ unsigned scanArgRegTableI(PTR_CBYTE     table,
     _ASSERTE(*castto(table, unsigned short *)++ == 0xBABE);
 #endif
 
-    bool      isPartialArgInfo;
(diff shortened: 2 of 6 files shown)


parjong:
@dotnet-bot test Ubuntu x64 Formatting please

parjong:

danmosemsft:

src/vm/eetwain.cpp
```diff
@@ -2413,6 +2413,18 @@ unsigned scanArgRegTableI(PTR_CBYTE     table,
     _ASSERTE(*castto(table, unsigned short *)++ == 0xBABE);
 #endif
 
+    bool      isPartialArgInfo;
```
janvorli: A nit, could you please name it hasPartialArgInfo instead (like in "the current frame **has** partial arg info")?

janvorli:
@jkotas, @BruceForstall can you please share your opinion on making the funclets fully interruptible and generating full arg info for them? I would like to make sure there are no downsides I don't see.

jkotas:
The downsides are bigger GC info and lower JIT throughput. Each of these hacks make `FEATURE_FIXED_OUT_ARGS` more appealing.

src/jit/morph.cpp
```diff
@@ -16202,7 +16202,14 @@ void Compiler::fgSetOptions()
...
     // the same test, leave info.compXcptnsCount here.
     if (info.compXcptnsCount > 0)
+    {
         codeGen->setFramePointerRequiredEH(true);
+#ifdef UNIX_X86_ABI
+        assert(!codeGen->isGCTypeFixed());
+        // Enforce fully interruptible codegen for funclet unwinding
+        genInterruptible = true;
+#endif // UNIX_X86_ABI
+    }
```
BruceForstall: Please don't insert the code here; I don't want to dilute or confuse the comment here. Instead, add code below the `#endif // _TARGET_X86_` below, e.g.:

#ifdef UNIX_X86_ABI
    if (info.compXcptnsCount > 0)
    {
        assert(!codeGen->isGCTypeFixed());
        // Enforce fully interruptible codegen for funclet unwinding
        genInterruptible = true;
    }
#endif // UNIX_X86_ABI

BruceForstall:
As a data point, can you compute the increase in GC info size on System.Private.CoreLib.ni.dll? If you can use jit-diff (from jitutils), the `jit-diff diff` command takes a `--gcinfo` argument that might help, although I'm not sure if it automatically adds up the size of all GC info and compares it base vs. diff.

parjong:
@BruceForstall jit-diff is still unavailable on x86/Linux. It seems that jit-diff has a strict dependency on dotnet host which is unavailable, yet:
```
runtime/corerun jitutils/bin/jit-diff.dll diff
Environment variable JIT_UTILS_ROOT not found - no configuration loaded.

Unhandled Exception: System.InvalidOperationException: Unable to locate dotnet multiplexer
   at Microsoft.DotNet.Cli.Utils.Muxer.get_MuxerPath()
   at Microsoft.DotNet.Cli.Utils.MuxerCommandResolver.Resolve(CommandResolverArguments commandResolverArguments)
   at Microsoft.DotNet.Cli.Utils.CompositeCommandResolver.Resolve(CommandResolverArguments commandResolverArguments)
   at Microsoft.DotNet.Cli.Utils.CommandResolver.TryResolveCommandSpec(ICommandResolverPolicy commandResolverPolicy, String commandName, IEnumerable`1 args, NuGetFramework framework, String configuration, String outputPath, String applicationName)
   at Microsoft.DotNet.Cli.Utils.Command.Create(ICommandResolverPolicy commandResolverPolicy, String commandName, IEnumerable`1 args, NuGetFramework framework, String configuration, String outputPath, String applicationName)
   at ManagedCodeGen.jitdiff.TryCommand(String name, IEnumerable`1 commandArgs, Boolean capture)
   at ManagedCodeGen.jitdiff.Config.SetRID()
   at ManagedCodeGen.jitdiff.Config..ctor(String[] args)
   at ManagedCodeGen.jitdiff.Main(String[] args)
Aborted (core dumped)
```

BruceForstall:
@parjong Have you tried jit-diff on linux/x64? You could do Linux/x86 asm diffs there using the altjit.

parjong:
@BruceForstall I tried, but x64/Linux failed to load x86/Linux jit. x86/Linux JIT is linked to x86/Linux libunwind, but it is not available on x64/Linux environment.

BruceForstall:
@parjong ok. you could do it from x86 Windows.

PR #10783, opened 2017-04-07
Devirtualizable default encoding

benaadams:
Follow up on https://github.com/dotnet/coreclr/pull/10747

/cc @tarekgh @jkotas @stephentoub

### DIFF
--- a/src/mscorlib/src/System/Text/Encoding.cs
+++ b/src/mscorlib/src/System/Text/Encoding.cs
@@ -85,7 +85,11 @@ namespace System.Text
     [Serializable]
     public abstract class Encoding : ICloneable
     {
-        private static Encoding defaultEncoding;
+        // For netcore we use UTF8 as default encoding since ANSI isn't available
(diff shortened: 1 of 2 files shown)


tarekgh:
LGTM. thanks @benaadams

src/mscorlib/src/System/Text/Encoding.cs
```diff
@@ -85,7 +85,11 @@ namespace System.Text
     [Serializable]
     public abstract class Encoding : ICloneable
     {
-        private static Encoding defaultEncoding;
+        // For netcore we use UTF8 as default encoding since ANSI isn't available
+        private static readonly UTF8Encoding.UTF8EncodingSealed s_defaultEncoding  = new UTF8Encoding.UTF8EncodingSealed(encoderShouldEmitUTF8Identifier: false);
+
+        // Returns an encoding for the system's current ANSI code page.
```
iSazonov: ANSI? It seems this is misleading.

benaadams: Just moved comment

Comment above qualifies
`For netcore we use UTF8 as default encoding since ANSI isn't available`

PR #10801, opened 2017-04-07
Move public ThreadAbortException back to CoreLib

jkotas:


jkotas:
Contributes to https://github.com/dotnet/corefx/issues/16229

PR #10806, opened 2017-04-07
Publish portable OSX binaries into separate container in pipeline

wtgodbe:
Previously, both portable & non-portable OSX builds were using the same RID in the pipeline (osx.10.12). This meant they were both publishing into the same container, so whoever published last, "won". This change will allow portable binaries to be published into a container named "xxx-osx", much like portable linux binaries are published into a container named "xxx-linux". Merging this & getting packages published to MyGet is a prereq for running tests against portable binaries, so we may not be able to do that until the mac machines get fixed.

CC @gkhanna79

### DIFF
--- a/buildpipeline/pipelines.json
+++ b/buildpipeline/pipelines.json
@@ -109,7 +109,7 @@
         {
           "Name": "DotNet-CoreClr-Trusted-Mac",
           "Parameters": {
-            "Rid": "osx.10.12",
+            "Rid": "osx",
             "portableBuild": "-portable"
           }
         },


janvorli:
@gkhanna79, @wtgodbe this is the first time I hear about OSX portable binaries. It doesn't make sense to  me, the portable option is used for Linux only.

wtgodbe:
My understanding is that OSX Portable binaries are meant to be used on any version of OSX (10.10, 10.11, 10.12, etc). We also have the same concept for Windows. @gkhanna79 can elaborate

janvorli:
There is nothing we build differently, the portable option of the build.sh is ignored for OSX both for corefx and coreclr. Or is the goal to produce a so called portable OSX build that's equivalent to the "non-portable" in expectation that at some point in the future we might need it and we want to have our infrastructure prepared for it now?

gkhanna79:
@janvorli The portable build for Windows and OSX, unlike Linux, is really about packaging the binaries in a portable RID (e.g. OSX-x64 or win-x64). There is nothing special, aside from that, for them.

janvorli:
@gkhanna79 ah, ok, that makes sense then.

wtgodbe:
@gkhanna79 any objections, or should I go ahead and merge?

PR #10807, opened 2017-04-07
Temporarily disable symweb indexing

dagood:
This step fails builds when the share is full, and the coreclr/master build takes up a lot of space. This share is also used by servicing branches and corefx. Disabling this step for now will take this pressure off the infra until we find a different mitigation.

Indexing on symweb isn't absolutely necessary. It is one symbol acquisition dev flow out of a few. For example, MyGet will still work.

/cc @MattGal

### DIFF
--- a/buildpipeline/DotNet-CoreClr-Trusted-Windows-x86.json
+++ b/buildpipeline/DotNet-CoreClr-Trusted-Windows-x86.json
@@ -227,7 +227,7 @@
       }
     },
     {
-      "enabled": true,
+      "enabled": false,
       "continueOnError": false,
       "alwaysRun": false,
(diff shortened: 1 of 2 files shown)


buildpipeline/DotNet-CoreClr-Trusted-Windows-x86.json
```diff
@@ -227,7 +227,7 @@
       }
     },
     {
-      "enabled": true,
+      "enabled": false,
```
gkhanna79: Can we please have a tracking issue to re-enable this when everything is back to normal?

dagood: Good idea, created https://github.com/dotnet/coreclr/issues/10809 to track.

dagood:
Merging, checked-in defs aren't used by CI.

PR #10811, opened 2017-04-07
Remove RELOC_SUPPORT define

BruceForstall:
It's always defined, is always expected to be defined, and the build
doesn't work without it.

Also remove unused `SECURITY_CHECK` and `VERIFY_IMPORTER` defines.


BruceForstall:
@briansull PTAL
cc @dotnet/jit-contrib

briansull:
Looks Good

PR #10813, opened 2017-04-07
[Local GC] Move operations on CLREventStatic to the EE interface

swgillespie:
The problem with `CLREventStatic` is that some of them (namely, ones created with `CreateAutoEvent` or `CreateManualEvent` are host-aware: waiting on an event of this type may eventually call a host's implementation of `IHostAutoEvent` (https://msdn.microsoft.com/en-us/library/ms164466(v=vs.110).aspx) if the CLR is hosted. This does not have any bearing on CoreCLR, since CoreCLR (as of today) does not support the traditional hosting APIs. This does have a bearing on the full framework CLR, though, which does need to support those APIs.

This PR addresses this particular hosting problem by delegating all operations related to `CLREventStatic` to the EE: creation, destruction, and all operations. This means a few things:
  1. The GC operates on `CLREventStatic`s through opaque references - it does not know anything about the implementation given to it from the EE. This further implies that the GC must always operate on `CLREventStatic`s behind a pointer, so many fields in the GC have been changed to contain a pointer to a `CLREventStatic` instead of embedding one by-value.
  2. The GC uses the `IsValid` method to determine whether or not an event is valid. Instead of this, this PR adds an additional invariant: if a GC observes a non-null pointer to a `CLREventStatic`, it /must/ be valid. All codepaths in the GC that previously were calling `IsValid` can therefore be replaced with null checks, and the pointer can be nulled out after a call to `CloseEvent`.
 3. In the future, the GC itself should be free of knowledge of the hosting API. The "GC PAL", or the GC-specific implementations of `GCToOSInterface`, will still need to know of the hosting API in case it needs to invoke `IHostMemoryManager` to satisfy virtual memory operations (https://msdn.microsoft.com/en-us/library/ms164513(v=vs.110).aspx) or otherwise inform the host that VM operations are occurring.

Apologies for the large diff - a lot of this diff is in `gc.cpp` replacing uses of `.Set`, `.Reset`, and `.Wait` (and friends) with their respective APIs on `GCToEEInterface`.

cc @Maoni0 @adityamandaleeka @sergiy-k @jkotas


swgillespie:
I will be doing a stress run overnight tonight to ensure this didn't introduce any regressions.

jkotas:
Shouldn't this be on GC-OS interface instead?

jkotas:
> specific implementations of GCToOSInterface, will still need to know of the hosting API in case it needs to invoke IHostMemoryManager to satisfy virtual memory operations

Right, you would need *desktop-specific* implementation of GCToOSInterface if you wanted it to support desktop hosting.

swgillespie:
@jkotas I picked `GCToEEInterface` after talking to Maoni to avoid having to duplicate the logic of calling into the host before waiting on an event. I'm happy to move it GCToOSInterface if that's better.

swgillespie:
Regarding a separate implementation of GCToOSInterface - at least how it is right now, GCToOSInterface is dispatched statically, so different implementations have to be chosen at compile-time. This works fine for now since we are distinguishing implementations by OS, but if we distinguish implementations by target runtime, won't we have to have some sort of e.g. `#ifdef DESKTOP_CLR` stuff to select the appropriate desktop implementation? Though if we also made GCToOSInterface dispatch dynamically like GCToEEInterface, then we would't have that problem. (Though even then, we'd need to know to what runtime we are attached to know which implementation to select)

jkotas:
I think that the desktop hosting hooks - if we need them - should be its own interface that is separate and hidden from everything else.

swgillespie:
So in that case, all hosting hooks (the VM ones and all of the event ones for this PR) would go through that interface?

jkotas:
I think that the GC should keep having its own self-contained memory and synchronization OS adaption layer.

The desktop hosting hooks should be built into Windows version of this OS adaption layer, under ifdef so that you do not have to carry them if you do not need to. Something like:

```
bool GCToOSInterface::VirtualCommit(void* address, size_t size)
{
#ifdef DESKTOP_HOSTING
    if (g_pDesktopMemoryHosting != null)
    {
        return g_pDesktopMemoryHosting->EEVirtualAlloc(address, size, MEM_COMMIT, PAGE_READWRITE) != nullptr;
    }
#endif

    return ::VirtualAlloc(address, size, MEM_COMMIT, PAGE_READWRITE) != nullptr;
}
```

Note that the desktop hosting hooks are ad-hoc. Only a subset of methods on the GCToOSInterface will have them.

I think that adding support for desktop hosting hooks in the local GC is very low priority. I would be spending cycles on it as the very last thing; once everything else works. It may be even fine to not have it ever.

swgillespie:
> I think that adding support for desktop hosting hooks in the local GC is very low priority. I would be spending cycles on it as the very last thing; once everything else works. It may be even fine to not have it ever.

Ok! Sounds good.

A `CLREventStatic*` is returned by `IGCHeap::GetWaitForGCEvent` today, and the EE calls both `Set` and `Reset` on it. I imagine we could remove that problem by instead exposing new methods `SetWaitForGCEvent` and `ResetWaitForGCEvent` on the interface, freeing us from having to expose `CLREventStatic` at all. Does that sound reasonable if we move all of this synchronization code to `GCToOSInterface`

jkotas:
> I imagine we could remove that problem by instead exposing new methods `SetWaitForGCEvent` and `ResetWaitForGCEvent` on the interface, freeing us from having to expose `CLREventStatic` at all.

Yes, this sounds good.

swgillespie:
@jkotas Just to be make sure that I understand this correctly before I embark on these things:

1. GetWaitForGCEvent can be split into Set/ResetWaitForGCEvent, so `CLREventStatic` doesn't need to be exposed on the interface and the EE doesn't have knowledge of the GC's synchronization layer
2. Implement the API surface of `CLREventStatic` in the GC PAL. I was thinking of basing the Unix implementation off of [the CoreRT implementation](https://github.com/dotnet/corert/blob/master/src/Native/Runtime/unix/PalRedhawkUnix.cpp#L207) since that does pretty much everything that the GC needs. The Windows one just has to forward to Win32 API calls.
3. Change all uses of `CLREventStatic` in the GC to use the GC PAL implementation.

Does this sound reasonable?

jkotas:
I think you just need 1. to solve your immediate problem.

2. and 3. look like a low-priority cleanup to me. Do you think there is a good reason to do it now?

swgillespie:
@jkotas Today the GC can't link as a standalone library because the method symbols for `CLREventStatic` are undefined in `gc.dll`, since they all live in `coreclr.dll`. The GC can either provide our own implementation (GCToOSInterface) or use the EE's (GCToEEInterface), but as far as I can tell we do need to do one or the other. 3) goes hand-in-hand with 2 (and is trivial thanks to this define: https://github.com/dotnet/coreclr/blob/master/src/gc/gcpriv.h#L200). We will definitely need to do more validation for 2), though.

jkotas:
I see - make sense.

src/gc/env/gcenv.os.h
```diff
@@ -47,6 +47,42 @@ struct GCThreadAffinity
...
 };
 
+// An event is a synchronization object whose state can be set and reset
+// indicating that an event has occured. It is used pervasively throughout
+// the GC.
+class GCEvent {
+public:
+    // Closes the event. Attempting to use the event past calling CloseEvent
+    // is a logic error.
+    virtual void CloseEvent() = 0;
```
jkotas: This should not be virtual. The rest of the GCToOSInterface is not virtual either.

swgillespie: I was thinking that virtual would make for a cleaner design - will fix.

swgillespie: The main reason why I was hoping to keep this class virtual was to allow for implementations to differ in size from `GCEvent`.

swgillespie:
I've validated these changes on Windows and OSX and am in the process of validating them on Linux. The event implementation in `gc/unix/events.cpp` is adapted from CoreRT with `PthreadCondAttrHolder` commented-out for now, though I'm happy to fix that (gc/unix doesn't have access to holders but we could roll our own RAII wrapper) before merging if desired.

src/gc/env/gcenv.os.h
jkotas: The GC-OS interface is statically linked, not virtualized. I do not think there is value in getting it virtualized. It is hard to tell where to stop once you start doing that, e.g. should new/delete operators be virtualized too? We have been there in desktop framework.

The different implementations of GCEvent won't co-exist. We should always need just one - so we should just include the right one to the build.

swgillespie: Ok!

swgillespie: @jkotas Would you prefer that these methods be on `GCToOSInterface` itself or are they okay here on the `GCEvent` class?

jkotas: I think they are okay on the GCEvent class.

src/gc/gcimpl.h
```diff
@@ -6,7 +6,7 @@
 #ifndef GCIMPL_H_
 #define GCIMPL_H_
 
-#define CLREvent CLREventStatic
+#define CLREvent GCEvent
```
jkotas: Rename CLREvent -> GCEvent and get rid of this define?

swgillespie: I saw a comment in `gcimpl.h` next to this define so I left it in place: `Please leave these definitions intact.` If it's not being used anywhere I'm happy to remove it but I'm not sure if it serves another purpose.

src/gc/gcpriv.h
```diff
@@ -2833,9 +2831,9 @@ class gc_heap
 
     PER_HEAP
 #ifndef MULTIPLE_HEAPS
-    CLREvent gc_done_event;
+    CLREvent* gc_done_event;
```
jkotas: This is `CLREvent*` pointer to `Impl*` pointer to actual stuff now. Would it make sense to get rid of at least one of these indirections at least - keep the Create method on GCEvent directly instead of having it on the GCToOS interface?

swgillespie: Yeah, that makes sense now that `GCEvent` has a fixed size. Will do.

src/gc/gcimpl.h
jkotas: Comments that say "do not ever change this" without explaining the reason are meant to be ignored.

swgillespie:
To avoid making the diff any larger than it already is, my most recent changes don't rename `Create*EventNoThrow` to `Create*Event` like I did in previous iterations. I'm happy to make that change if desired.

src/gc/unix/events.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+uint32_t GCEvent::Wait(uint32_t timeout, bool alertable)
+{
+    assert(m_impl != nullptr);
+    return m_impl->Wait(timeout, alertable);
+}
+
+bool GCEvent::CreateAutoEventNoThrow(bool initialState)
+{
+    // [DESKTOP TODO] The difference between events and OS events is
```
jkotas: This comment is not applicable for Unix

jkotas:
LGTM

src/gc/unix/events.cpp
swgillespie: will fix.

swgillespie:
Thanks!

I'll be doing some additional validation on these changes today.

sergiy-k:
LGTM

swgillespie:

swgillespie:
thanks!

PR #10816, opened 2017-04-07
[Local GC] Move handle destruction code to IGCHandleTable

adityamandaleeka:
This is similar to the handle creation PR, but for destruction.

As before, I've consolidated the handle destruction functions for the interface and added the functions that the VM code calls to gchandletableutilities.h on the VM side. In the future, we can remove some/all of these by modifying the callers, but that's out of the scope of the Local GC work.

Interesting notes:
- WinRT handle destruction has some extra logic, but I moved all of that stuff to the VM side so the interface doesn't know anything about WinRT.
- SizedRef handle destruction also had some special logic, but it looks like nobody is using that in CoreCLR, so I just deleted it.

@Maoni0 @swgillespie @jkotas @sergiy-k

### DIFF
--- a/src/gc/gchandletable.cpp
+++ b/src/gc/gchandletable.cpp
@@ -55,3 +55,18 @@ OBJECTHANDLE GCHandleTable::CreateDependentHandle(void* table, Object* primary, 
 
     return handle;
 }
+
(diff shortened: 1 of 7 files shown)


danmosemsft:

adityamandaleeka:
I'll fix the merge conflict I just introduced in my other PR.

PR #10817, opened 2017-04-08
[Local GC] Fix some interface violations

swgillespie:
This PR fixes two linker errors that arise when building the GC standalone:
 1. A use of `_DebugBreak` in `GCHeap::WaitUntilGCComplete`, replaced with `GCToOSInterface::DebugBreak`
2. A use of `EEPOLICY_HANDLE_FATAL_ERROR` in `FATAL_GC_ERROR`, replaced with `GCToEEInterface::HandleFatalError`.

cc @Maoni0 @adityamandaleeka @sergiy-k @jkotas

### DIFF
--- a/src/gc/gcee.cpp
+++ b/src/gc/gcee.cpp
@@ -408,12 +408,8 @@ BlockAgain:
         dwWaitResult = WaitForGCEvent->Wait(DETECT_DEADLOCK_TIMEOUT, FALSE );
 
         if (dwWaitResult == WAIT_TIMEOUT) {
-            //  Even in retail, stop in the debugger if available.  Ideally, the
-            //  following would use DebugBreak, but debspew.h makes this a null
-            //  macro in retail.  Note that in debug, we don't use the debspew.h
-            //  macros because these take a critical section that may have been
-            //  taken by a suspended thread.
-            FreeBuildDebugBreak();
+            //  Even in retail, stop in the debugger if available.
+            GCToOSInterface::DebugBreak();
             goto BlockAgain;
         }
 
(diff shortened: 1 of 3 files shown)


src/gc/gcee.cpp
```diff
@@ -408,12 +408,8 @@ uint32_t GCHeap::WaitUntilGCComplete(bool bConsiderGCStart)
...
 
         if (dwWaitResult == WAIT_TIMEOUT) {
-            //  Even in retail, stop in the debugger if available.  Ideally, the
-            //  following would use DebugBreak, but debspew.h makes this a null
-            //  macro in retail.  Note that in debug, we don't use the debspew.h
-            //  macros because these take a critical section that may have been
-            //  taken by a suspended thread.
-            FreeBuildDebugBreak();
+            //  Even in retail, stop in the debugger if available.
+            GCToOSInterface::DebugBreak();
```
adityamandaleeka: It looks like `__FreeBuildDebugBreak` (used in release builds) checks whether `CLRConfig::INTERNAL_BreakOnRetailAssert` is set, and does nothing if it isn't. Is this change in behavior intentional?

swgillespie: Yeah, at least for now. In a future PR I'll be adding a mechanism by which we can query config values from the GC while build standalone, so we could choose do to that if we'd like. This is the only place in the GC that I found that honors `BreakOnRetailAssert`, though - all other places just call `DebugBreak` unconditionally and the process will die if there's no debugger attached.

jkotas: This code is under `DETECT_DEADLOCK` ifdef. This ifdef does not seem to be turned on, so it should not matter much what we do here.

adityamandaleeka: Sounds good.

PR #10819, opened 2017-04-08
Delete unused CreateVariableHandle function

adityamandaleeka:
As the title says, it doesn't look like this is used anywhere.

@jkotas PTAL


jkotas:
This is used by CoreRT. It should stay.

adityamandaleeka:
OK, in that case I'll add it to IGCHandleTable. Thanks!

adityamandaleeka:
Actually, I don't think I need to add anything to the interface, since it's just using the extra info field for the type.

PR #10823, opened 2017-04-08
Refactor variable strength handle creation

adityamandaleeka:
Code on the VM side currently calls a `CreateVariableHandle` function in objecthandle.h directly, which is mostly like any other handle creation function, except it also does some extra validation on the variable handle type.

This change refactors this a bit so that the VM side creates these handles via the interface, and the validation of the type happens on the GC side in `HndCreateHandle`.

@jkotas PTAL

### DIFF
--- a/src/gc/handletable.cpp
+++ b/src/gc/handletable.cpp
@@ -313,6 +313,10 @@ OBJECTHANDLE HndCreateHandle(HHANDLETABLE hTable, uint32_t uType, OBJECTREF obje
     }
 #endif // _DEBUG && !FEATURE_REDHAWK
 
+    // If we are creating a variable-strength handle, verify that the
+    // requested variable handle type is valid.
+    _ASSERTE(uType != HNDTYPE_VARIABLE || IS_VALID_VHT_VALUE(lExtraInfo));
+
     VALIDATEOBJECTREF(object);
 
     // fetch the handle table pointer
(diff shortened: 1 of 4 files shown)


src/gc/handletable.cpp
```diff
@@ -313,6 +313,14 @@ OBJECTHANDLE HndCreateHandle(HHANDLETABLE hTable, uint32_t uType, OBJECTREF obje
     }
 #endif // _DEBUG && !FEATURE_REDHAWK
 
+    // If we are creating a variable-strength handle, verify that the
+    // requested variable handle type is valid.
+    if (uType == HNDTYPE_VARIABLE && !IS_VALID_VHT_VALUE(lExtraInfo))
```
jkotas: I think this validation should be just assert (since we assert when it fails anyway).

adityamandaleeka: @jkotas Not sure I understood your comment correctly. Do we want this function to succeed in release builds if `lExtraInfo` is not a valid variable handle type?

jkotas: The handle table methods are expected to be called with valid arguments. If you call them with invalid arguments, they can succeed, crash, or whether in release builds. It is fine to assert on invalid arguments in debug builds. The error handling in this case should be consistent with the overall design. For example, bad uType passed to HndCreateHandle is not checked in release builds either.

adityamandaleeka: Got it, thanks! I misunderstood "since we assert when it fails anyway". I'll send an update.

adityamandaleeka:
@jkotas PTAL

PR #10827, opened 2017-04-09
Delete reference to model.xml since it no longer exists

jkotas:


jkotas:
cc @fujiy 

Have you noticed anything else in this doc that is outdated?

fujiy:
Just that
Thanks

PR #10830, opened 2017-04-09
Move files to shared partition

jkotas:
Moved a batch of files to shared CoreLib partition, applying formatting and other changes from CoreRT as appropriate.


src/mscorlib/shared/System/ArgumentException.cs
```diff
@@ -21,7 +21,7 @@ namespace System
     // message describing what was wrong and which parameter is incorrect.
     // 
     [Serializable]
-    public class ArgumentException : SystemException, ISerializable
```
danmosemsft: removed because it's on the baseclass alreayd?

src/mscorlib/shared/System/MissingFieldException.cs
```diff
@@ -9,64 +9,50 @@
...
-            get
-            {
-                if (ClassName == null)
-                {
-                    return base.Message;
-                }
-                else
-                {
-                    // do any desired fixups to classname here.
-                    return SR.Format(SR.MissingField_Name, (Signature != null ? FormatSignature(Signature) + " " : "") + ClassName + "." + MemberName);
```
danmosemsft: Looks like you fixed a bug in the ordering

src/mscorlib/shared/System/Runtime/Versioning/TargetFrameworkAttribute.cs
```diff
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-//
```
danmosemsft: nit, you didn't delete the purpose comment in this one.

src/mscorlib/shared/System/ArgumentException.cs
jkotas: Yes - it is not necessary. It was removed in CoreRT, and it is consistent with most other CoreLib exception types.

src/mscorlib/shared/System/Runtime/Versioning/TargetFrameworkAttribute.cs
jkotas: I have been reconciling diffs between CoreCLR and CoreRT, not much beyond that. The purpose comment was on both sides so I have kept it.

We have the purpose comment in number of files in the shared partition. I have deleted it only if it was deleted in CoreRT and it had zero information value.

src/mscorlib/shared/System/MissingFieldException.cs
jkotas: Hmm, this is implemented differently between CoreCLR and CoreRT. Not sure which side is right. I will exclude this file from this batch.

glenn-slayden:
@jkotas I got lost in GitHub  :-(  Unfortunately, it's a common occurrence for me.

Did I put the following comment in the right place? (wait for the in-page jump `#` after clicking...)

<a href=https://github.com/dotnet/coreclr/commit/1a9781b68b8ec66c223bb96f44ae1d2704ce6fdf#r33421958>1a9781b68b8ec66c223bb96f44ae1d2704ce6fdf#r33421958</a>

It doesn't seem worth entering an issue on.

PR #10835, opened 2017-04-09
add test with deep execution tree without stores.

sandreenko:
It will not fail if the compiler spills trees from eeStack from time to time or all recursive walkers are replaced by non-recursive algorithms.


pgavlin:
Is there a reason the existing `https://github.com/dotnet/coreclr/blob/master/tests/src/JIT/jit64/opt/cse/hugeexpr1.cs` is insufficient for this scenario?

sandreenko:
I didn't see any failures in ci when I disabled stack spilling and send PR.
I think hugeexpt1.cs doesn't have very deep tree, the biggest one has deep near 600 as I see.

pgavlin:
Hm. I am pretty sure that we're already covering this somewhere... LGTM if we aren't, in any case.

sandreenko:
I want to double check.

sandreenko:
It is pretty interesting that we see this fail only in Windows_NT x64 Debug Build and Test. It is what I tested locally.

And as you see there are no failures on existing tests. 
I think I should try to make it fails in Release and Checked. Maybe just to increase the depth of the tree.

sandreenko:
Ok, now it looks fine. It breaks in all modes (x64/x86 Debug/Release/Checked).
So I want to merge it as soon as testing will be finished.

PR #10837, opened 2017-04-10
Move files to shared CoreLib partition

jkotas:
- Reconcile diffs with CoreRT as appropriate
- Deleted unused internal SuppressMergeCheckAttribute and XmlIgnoreMemberAttribute


clr.coreclr.props
```diff
@@ -37,7 +37,7 @@
     <FeatureUseAsmGCWriteBarriers>true</FeatureUseAsmGCWriteBarriers>
     <!-- Setting this to "false" works only for workstation GC, not server. -->
     <FeatureSymDiff>true</FeatureSymDiff>
-    <FeatureReadyToRun Condition="'$(TargetArch)'!='arm64'">true</FeatureReadyToRun>
+    <FeatureReadyToRun>true</FeatureReadyToRun>
```
jkotas: Mismatch between C++ makefiles and C# project files that I run into

src/mscorlib/shared/Interop/Unix/System.Native/Interop.GetRandomBytes.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+internal partial class Interop
+{
+    internal unsafe partial class Sys
+    {
+        [DllImport(Interop.Libraries.SystemNative, EntryPoint = "SystemNative_GetNonCryptographicallySecureRandomBytes")]
```
danmosemsft: I can't find this entry point.. mabye it's a generated name?

clr.coreclr.props
danmosemsft: Does this change behavior -- is it worth a separate PR in case it needs revert?

src/mscorlib/shared/Interop/Unix/System.Native/Interop.GetRandomBytes.cs
jkotas: https://github.com/dotnet/corefx/blob/274d3c3be1412454db8aa2b31c50d23467c21e84/src/Native/Unix/System.Native/pal_random.cpp#L19

clr.coreclr.props
jkotas: will do

src/mscorlib/shared/Interop/Unix/System.Native/Interop.GetRandomBytes.cs
danmosemsft: Oh it's in corefx. I don't understand how that kind of call works yet.

PR #10840, opened 2017-04-10
Delete FEATURE_READYTORUN define for CoreLib build

jkotas:
This define does not have much value for the CoreLib build, and it was missing on arm64.
Also deleted dead props and targets files left over from the legacy build


jkotas:

PR #10842, opened 2017-04-10
Ensure Environment.ExitCode works correctly after app domain unloaded.

### LINKED ISSUE #6206: Environment.set_ExitCode doesn't change process' exit code
set_ExitCode is supposed to enable changing the exit code of an app with a void-returning Main method, but on CoreCLR setting it doesn't propagate accordingly, e.g. set it to 42, and the process' exit code is still 0.

mazong1123:
This PR addresses 2 problems of Environment.ExitCode:

1. Can't get correct exit code of main function.
2. Can't set %errorlevel%.

Details can be found on #6206

Fix #6206

### DIFF
--- a/src/coreclr/hosts/inc/coreclrhost.h
+++ b/src/coreclr/hosts/inc/coreclrhost.h
@@ -27,6 +27,10 @@ CORECLR_HOSTING_API(coreclr_initialize,
 
 CORECLR_HOSTING_API(coreclr_shutdown,
             void* hostHandle,
+            unsigned int domainId);
+
+CORECLR_HOSTING_API(coreclr_shutdown_2,
+            void* hostHandle,
             unsigned int domainId,
             int* latchedExitCode);
 
--- a/src/coreclr/hosts/inc/coreclrhost.h
+++ b/src/coreclr/hosts/inc/coreclrhost.h
@@ -27,7 +27,8 @@ CORECLR_HOSTING_API(coreclr_initialize,
 
 CORECLR_HOSTING_API(coreclr_shutdown,
             void* hostHandle,
-            unsigned int domainId);
+            unsigned int domainId,
+            int* latchedExitCode);
 
 CORECLR_HOSTING_API(coreclr_create_delegate,
             void* hostHandle,
--- a/src/dlls/mscoree/unixinterface.cpp
+++ b/src/dlls/mscoree/unixinterface.cpp
@@ -183,9 +183,9 @@ int coreclr_initialize(
     }
 #endif
 
-    ReleaseHolder<ICLRRuntimeHost2> host;
+    ReleaseHolder<ICLRRuntimeHost4> host;
 
-    hr = CorHost2::CreateObject(IID_ICLRRuntimeHost2, (void**)&host);
+    hr = CorHost2::CreateObject(IID_ICLRRuntimeHost4, (void**)&host);
     IfFailRet(hr);
 
     ConstWStringHolder appDomainFriendlyNameW = StringToUnicode(appDomainFriendlyName);
@@ -284,7 +284,7 @@ int coreclr_shutdown(
             void* hostHandle,
             unsigned int domainId)
 {
-    ReleaseHolder<ICLRRuntimeHost2> host(reinterpret_cast<ICLRRuntimeHost2*>(hostHandle));
+    ReleaseHolder<ICLRRuntimeHost4> host(reinterpret_cast<ICLRRuntimeHost4*>(hostHandle));
 
     HRESULT hr = host->UnloadAppDomain(domainId, true); // Wait until done
     IfFailRet(hr);
@@ -315,9 +315,7 @@ int coreclr_shutdown_2(
             unsigned int domainId,
             int* latchedExitCode)
 {
-    ICLRRuntimeHost4* hostHandleV4 = nullptr;
-    reinterpret_cast<ICLRRuntimeHost2*>(hostHandle)->QueryInterface(IID_ICLRRuntimeHost4, reinterpret_cast<void**>(&hostHandleV4));
-    ReleaseHolder<ICLRRuntimeHost4> host(reinterpret_cast<ICLRRuntimeHost4*>(hostHandleV4));
+    ReleaseHolder<ICLRRuntimeHost4> host(reinterpret_cast<ICLRRuntimeHost4*>(hostHandle));
 
     HRESULT hr = host->UnloadAppDomain2(domainId, true, latchedExitCode); // Wait until done
     IfFailRet(hr);
@@ -354,7 +352,7 @@ int coreclr_create_delegate(
             const char* entryPointMethodName,
             void** delegate)
 {
-    ICLRRuntimeHost2* host = reinterpret_cast<ICLRRuntimeHost2*>(hostHandle);
+    ICLRRuntimeHost4* host = reinterpret_cast<ICLRRuntimeHost4*>(hostHandle);
 
     ConstWStringHolder entryPointAssemblyNameW = StringToUnicode(entryPointAssemblyName);
     ConstWStringHolder entryPointTypeNameW = StringToUnicode(entryPointTypeName);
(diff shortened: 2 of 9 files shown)


mazong1123:
@jkotas Could you please review this PR? Thanks.

mazong1123:

mazong1123:

src/coreclr/hosts/inc/coreclrhost.h
```diff
@@ -27,7 +27,8 @@ CORECLR_HOSTING_API(coreclr_initialize,
 
 CORECLR_HOSTING_API(coreclr_shutdown,
             void* hostHandle,
-            unsigned int domainId);
+            unsigned int domainId,
```
janvorli: I don't think we can change the signature of the existing hosting API. There are other Microsoft projects like dotnet cli and possibly also community projects using the current APIs and this would break them. It seems that we need to create a new version of coreclr_shutdown with the added latchedExitCode and keep the old one unchanged. Named e.g. coreclr_shutdown_2.

mazong1123: @janvorli  Should I do the same change on `ICLRRuntimeHost::UnloadAppDomain` (keep the old one, add a new `ICLRRuntimeHost::UnloadAppDomain2`)?

BTW, can we just set the default value of latchedExitCode to NULL? That's what I did on  `ICLRRuntimeHost::UnloadAppDomain`

janvorli: Right, you cannot modify the ICLRRuntimeHost methods either. This is a COM interface and by COM rules, you need to add a new version of the whole interface, called ICLRRuntimeHost4 (there already is ICLRRuntimeHost3) and put the new method on it.

mazong1123: @janvorli I went through the code from coreclr initialization to shutdown. It seems like we need more efforts to address this issue. Therefore I just wrote down the design here before diving to the code. Please correct me if anything is incorrect. 

Current process of executing a managed assembly:
1. In `ExecuteManagedAssembly`, `coreclr_initialize` called to initialize the `hostHandle`.
2. in `coreclr_initialize` , a `CorHost2` object created, who implemented interface `ICLRRuntimeHost2`, and a pointer to `ICLRRuntimeHost2` is assigned to `hostHandle`.
3. In `ExecuteManagedAssembly`, `coreclr_execute_assembly` called with `hostHandle` passed in.
4. In `coreclr_execute_assembly`, `hostHandle` is cast to `ICLRRuntimeHost2`.
5. In `ExecuteManagedAssembly`, `coreclr_shutdown` called with `hostHandle` passed in.
6. In `coreclr_shutdown`, `hostHandle` is cast to `ICLRRuntimeHost2`.

To fix the `Environment.ExitCode` issue while keeping `ICLRRuntimeHost2` not being changed, I'm going to:
1. Create a new COM interface `ICLRRuntimeHost4` which introduces a new `UnloadAppDomain` method with `latchedExitCode` as its last parameter. 
2. Create a `CorHost3` object which implements `ICLRRuntimeHost4`.
3. Create `coreclr_initialize_2`, `coreclr_execute_assembly_2` and `coreclr_shutdown_2` functions. They have same logic to `coreclr_initialize`, `coreclr_execute_assembly` and `coreclr_shutdown` but use `CorHost3` and `ICLRRuntimeHost4` inside.

jkotas: You should not need to create CorHost4 . The new interface can be implemented by CorHost2 just fine. Also, you should not need to create `coreclr_initialize_2` or `coreclr_execute_assembly_2`.  `coreclr_shutdown_2` is the only one needed.

src/dlls/mscoree/unixinterface.cpp
```diff
@@ -299,6 +299,39 @@ int coreclr_shutdown(
...
+//  HRESULT indicating status of the operation. S_OK if the assembly was successfully executed
+//
+extern "C"
+int coreclr_shutdown_2(
+            void* hostHandle,
+            unsigned int domainId,
+            int* latchedExitCode)
+{
+    ICLRRuntimeHost4* hostHandleV4 = nullptr;
+    reinterpret_cast<ICLRRuntimeHost2*>(hostHandle)->QueryInterface(IID_ICLRRuntimeHost4, reinterpret_cast<void**>(&hostHandleV4));
```
janvorli: Instead of this query interface here, could you please change all references in this file from ICLRRuntimeHost2 to ICLRRuntimeHost4?

mazong1123: Sure, I've changed it.

mazong1123:
@dotnet-bot  test Tizen armel Cross Release Build please

src/vm/assembly.cpp
```diff
@@ -1965,6 +1965,10 @@ INT32 Assembly::ExecuteMainMethod(PTRARRAYREF *stringArgs, BOOL waitForOtherThre
     IfFailThrow(hr);
     
     END_ENTRYPOINT_THROWS;
+
+    // Sync the latched exit code with the return value of main function.
+    SetLatchedExitCode(iRetVal);
```
jkotas: Is this really needed? 

It will incorrectly overwrite latched exit code for non-void returning functions.

https://github.com/mazong1123/coreclr/blob/0c46640307dab1d8359f4e3b856a4390d9a6389d/src/vm/assembly.cpp#L1854 should take care of it for non-void returning main.

jkotas:
BTW: Once this goes through, we will also need to update the nuget aware dotnet host (https://github.com/dotnet/core-setup/blob/master/src/corehost/cli/coreclr.cpp) to use the new method when it is available.

src/vm/assembly.cpp
mazong1123: > Is this really needed?

The condition in this line will be false and let the latched exit code keep 0 at least for the example [here](https://github.com/dotnet/coreclr/issues/6206#issuecomment-231935917):
https://github.com/mazong1123/coreclr/blob/0c46640307dab1d8359f4e3b856a4390d9a6389d/src/vm/assembly.cpp#L1852

I don't know why we're checking the stringArgs here. Seems like we should remove the check. Do you have any idea about this line of code?

mazong1123:
@jkotas Do we need a shutdown2 method instead of updating the method [shutdown] (https://github.com/dotnet/core-setup/blob/master/src/corehost/cli/coreclr.cpp#L86)? I guess we need a [coreclr_shutdown_2_fn](https://github.com/dotnet/core-setup/blob/master/src/corehost/cli/coreclr.cpp#L22) as well?

src/vm/assembly.cpp
jkotas: > Seems like we should remove the check

Agree.

Could you please also verify that the following program will have exit code 123?

```
using System;

class My {
    static void Main() {
        Environment.ExitCode = 123;
    }
}
```

jkotas:
> Do we need a shutdown2 method instead of updating the method shutdown? 

Updating shutdown should do the trick.

> I guess we need a coreclr_shutdown_2_fn as well?

Agree.

src/vm/assembly.cpp
mazong1123: I just ran the sample program and `echo $?` gave me 123. Note I've removed the check and my sync code before testing. So it works.

mazong1123:
@jkotas I've fired a new issue here: https://github.com/dotnet/core-setup/issues/2050

I'll work on that once this PR merged.

jkotas:
I have just noticed that we have not unified the Windows and Unix corerun yet, and the Windows one still goes through the interface. Could you please also update these to use ICLRRuntimeHost4 ?

https://github.com/dotnet/coreclr/blob/master/src/coreclr/hosts/corerun/corerun.cpp
https://github.com/dotnet/coreclr/blob/master/src/coreclr/hosts/coreconsole/coreconsole.cpp

mazong1123:
@jkotas I created a new issue #10960  to track this change. I'll submit a new PR accordingly. Sorry for missing the Windows change.

Issue #10722, opened 2017-04-05
[RyuJIT/ARM32] Implement struct argument passing


hseok-oh:
In the comment, (#8496 (comment))
```
    268 morph.cpp:4610 - NYI_ARM: fgMorphMultiregStructArgs
```

And when we use ryujit for BringUpTest only, we can find 2 NYIs related with fgMorphMultiregStructArgs

- JIT/CodeGenBringUpTests/struct16args:Main
- JIT/CodeGenBringUpTests/StructFldAddr:Main

mskvortsov:
`fgMorphMultiregStructArgs` gets called by `fgMorphArgs` at [morph.cpp:4373](https://github.com/dotnet/coreclr/blob/a5fef2172ad19d4c58e67c9e5bee10d247cd8929/src/jit/morph.cpp#L4366-L4375):
```c++
#ifndef LEGACY_BACKEND
    // In the future we can migrate UNIX_AMD64 to use this
    // method instead of fgMorphSystemVStructArgs

    // We only build GT_FIELD_LISTs for MultiReg structs for the RyuJIT backend
    if (hasMultiregStructArgs)
    {
        fgMorphMultiregStructArgs(call);
    }
#endif // LEGACY_BACKEND
```
For the ARM case, `hasMultiregStructArgs` gets set on at [morph.cpp:3767](https://github.com/dotnet/coreclr/blob/a5fef2172ad19d4c58e67c9e5bee10d247cd8929/src/jit/morph.cpp#L3762-L3769):
```c++
#ifndef _TARGET_X86_
                // TODO-Arm: Does this apply for _TARGET_ARM_, where structs passed by value can be split between
                // registers and stack?
                if (size > 1)
                {
                    hasMultiregStructArgs = true;
                }
#endif // !_TARGET_X86_
```
Does an answer for the question posed in that comment exist? In general, considering poor readability of the `fgMorphArgs` function, are there any shortcuts for understanding the logic behind it? What are the steps required to implement struct-passing ABI for the RyuJIT/ARM32?

Some existing documentation on the subject: [jit-call-morphing.md](https://github.com/dotnet/coreclr/blob/fe98261c0fef181197380d0b42c482eb6eddcf94/Documentation/design-docs/jit-call-morphing.md), [first-class-structs.md](https://github.com/dotnet/coreclr/blob/fe98261c0fef181197380d0b42c482eb6eddcf94/Documentation/design-docs/first-class-structs.md)
[...]

PR #10846, opened 2017-04-10
[ARM32/RyuJIT] Argument passing for multireg struct

hseok-oh:
Enable fgMorphMultiregStructArgs by removing assert NYI
Fix fgMorphArgs to pass fgMorphMultiregStructArgs
Can pass morphing phase for argument passing of struct

related issue: #10722 #8496


hseok-oh:
cc/ @dotnet/arm32-contrib

hseok-oh:
### JIT/CodeGenBringUpTests/StructFldAddr
Before
`morph.cpp:4610 - NYI_ARM: fgMorphMultiregStructArgs`
After
`lsraarm.cpp:1438 - NYI: Unimplemented node type obj`

### JIT/CodeGenBringUpTests/struct16args
Before
`morph.cpp:4610 - NYI_ARM: fgMorphMultiregStructArgs`
After
`lsraarm.cpp:982 - NYI: lowering double`

BruceForstall:
cc @dotnet/jit-contrib

src/jit/morph.cpp
```diff
@@ -3185,7 +3185,16 @@ GenTreeCall* Compiler::fgMorphArgs(GenTreeCall* call)
         }
 
 #ifdef _TARGET_ARM_
-
+#ifndef LEGACY_BACKEND
+        if (reMorphing)
```
BruceForstall: Instead of adding any code, it looks like you can just adjust the `#ifdefs` to share arm64 code. That is, change the above to:
```
#if defined(_TARGET_ARM_) && defined(LEGACY_BACKEND)
```
and below from:
```
#elif defined(_TARGET_ARM64_)
```
to:
```
#elif defined(_TARGET_ARMARCH_)
```

src/jit/morph.cpp
```diff
@@ -3185,7 +3185,16 @@ GenTreeCall* Compiler::fgMorphArgs(GenTreeCall* call)
...
+        {
+            passUsingFloatRegs = isValidFloatArgReg(argEntry->regNum);
+        }
+        else
+        {
+            passUsingFloatRegs = !callIsVararg && (isHfaArg || varTypeIsFloating(argx));
+        }
+#else  // LEGACY_BACKEND
         bool passUsingIntRegs;
```
BruceForstall: Why don't you need to execute the following code (now under `LEGACY_BACKEND`) for RyuJIT/arm32? E.g., don't you need to check `!opts.compUseSoftFP`? And what about the rules implemented below about argument register alignment for long and double? It seems like you are altering the calling ABI.

src/jit/morph.cpp
```diff
@@ -3464,7 +3473,7 @@ GenTreeCall* Compiler::fgMorphArgs(GenTreeCall* call)
                     structPassingKind howToPassStruct;
                     structBaseType = getArgTypeForStruct(objClass, &howToPassStruct, originalSize);
 
-#ifdef _TARGET_ARM64_
+#if defined(_TARGET_ARM64_) || (defined(_TARGET_ARM_) && !defined(LEGACY_BACKEND))
                     if ((howToPassStruct == SPK_PrimitiveType) && // Passed in a single register
```
BruceForstall: Doesn't this change the ABI for ARM32?

(btw, please update the `#endif` comment when updating the `#if` condition.)

src/jit/morph.cpp
```diff
@@ -4606,7 +4608,7 @@ void Compiler::fgMorphMultiregStructArgs(GenTreeCall* call)
     //
     CLANG_FORMAT_COMMENT_ANCHOR;
 
-#ifdef _TARGET_ARM_
+#ifdef _TARGET_ARM
```
hqueue: Typo here.

hseok-oh:
Fixed
- match ryujit ABI with legacy
- morph struct argument to GT_FILED_LIST when size <= 4
- cannot handle split struct that using register and stack, and hfa yet

Example
```
public struct Rational
{
    public int num;
    public int den;
}
public struct RationalPolynomial
{
    public Rational a;
    public Rational b;
}
public class BringUpTest
{
    const int Pass = 100;
    const int Fail = -1;

    public static int StructFldAddr(RationalPolynomial rp)
    {   
        return rp.a.num + rp.b.num;
    }

    public static int Main()
    {   
        Rational a = new Rational();
        Rational b = new Rational();
        a.num = 3;
        a.den = 4;
        b.num = 2;
        b.den = 3;
        RationalPolynomial rp = new RationalPolynomial();
        rp.a = a;
        rp.b = b;
        int y = StructFldAddr(rp);
        if (y == 5) return Pass;
        else return Fail;
    }
}
```

Generated code: call `StructFldAddr(rp)`
```
IN0011: 000034  9800           ldr     r0, [sp] // [V02 loc2]
IN0012: 000036  9901           ldr     r1, [sp+0x04]    // [V02 loc2+0x04]
IN0013: 000038  9A02           ldr     r2, [sp+0x08]    // [V02 loc2+0x08]
IN0014: 00003A  9B03           ldr     r3, [sp+0x0c]    // [V02 loc2+0x0c]
IN0015: 00003C  F240 0400      movw    r4, LOW RELOC 0x74579a08
IN0016: 000040  F2C0 0400      movt    r4, HIGH RELOC 0x74579a08
IN0017: 000044  F8D4 C000      ldr     r12, [r4]
[0205AE1C] ptr arg pop  0
IN0018: 000048  47E0           blx     r12              // BringUpTest:StructFldAddr(struct):int
```

src/jit/jit.h
```diff
@@ -276,7 +276,7 @@
 #define FEATURE_UNIX_AMD64_STRUCT_PASSING_ONLY(x)
 #endif // defined(FEATURE_UNIX_AMD64_STRUCT_PASSING)
 
-#if defined(FEATURE_UNIX_AMD64_STRUCT_PASSING) || (defined(_TARGET_X86_) && !defined(LEGACY_BACKEND))
+#if defined(FEATURE_UNIX_AMD64_STRUCT_PASSING) || (!defined(_TARGET_64BIT_) && !defined(LEGACY_BACKEND))
```
CarolEidt: The associated `#else` clauses also need to be changed.

src/jit/morph.cpp
```diff
@@ -3759,14 +3769,19 @@ GenTreeCall* Compiler::fgMorphArgs(GenTreeCall* call)
                     }
                 }
 
-#ifndef _TARGET_X86_
+#ifdef _TARGET_64BIT_
                 // TODO-Arm: Does this apply for _TARGET_ARM_, where structs passed by value can be split between
```
CarolEidt: Presumably this comment should be removed.

CarolEidt:
> #else  // !_TARGET_X86_

This and the `#endif` also need to be changed.

---
Refers to: src/jit/morph.cpp:4174 in 3c27e6a. [](commit_id = 3c27e6a8fd279af2bcd7139728dd753bb637d7f1, deletion_comment = False)

src/jit/morph.cpp
```diff
@@ -4186,7 +4201,7 @@ GenTreeCall* Compiler::fgMorphArgs(GenTreeCall* call)
         }
 #endif // !LEGACY_BACKEND
 
-#if defined(_TARGET_X86_) && !defined(LEGACY_BACKEND)
+#if !defined(_TARGET_64BIT_) && !defined(LEGACY_BACKEND)
         if (isStructArg)
```
CarolEidt: Also here the `#endif` needs to be changed.

CarolEidt:
> //    Currently the implementation only handles ARM64 and will NYI for other architectures.

Needs to be updated

---
Refers to: src/jit/morph.cpp:4719 in 3c27e6a. [](commit_id = 3c27e6a8fd279af2bcd7139728dd753bb637d7f1, deletion_comment = False)

src/jit/morph.cpp
```diff
@@ -4766,43 +4778,63 @@ GenTreePtr Compiler::fgMorphMultiregStructArg(GenTreePtr arg, fgArgTabEntryPtr f
...
             assert(argValue->OperGet() == GT_OBJ);
 
             // We need to load the struct from an arbitrary address
             // and we can't read past the end of the structSize
-            // We adjust the second load type here
+            // We adjust the last load type here
             //
-            if (structSize < 2 * TARGET_POINTER_SIZE)
+            unsigned remains  = structSize % TARGET_POINTER_SIZE;
+            unsigned lastElem = elemCount - 1;
```
CarolEidt: I would prefer a more meaningful name such as `remainingRegisters` or `remainingSlots`.

src/jit/morph.cpp
```diff
@@ -4824,10 +4856,10 @@ GenTreePtr Compiler::fgMorphMultiregStructArg(GenTreePtr arg, fgArgTabEntryPtr f
         assert(varNum < lvaCount);
         LclVarDsc* varDsc = &lvaTable[varNum];
 
-        // At this point any TYP_STRUCT LclVar must be a 16-byte struct
+        // At this point any TYP_STRUCT LclVar must be aligned struct
```
CarolEidt: >be aligned [](start = 52, length = 10)

typo: "must be an aligned struct"

hseok-oh:
PTAL @BruceForstall @CarolEidt

src/jit/morph.cpp
```diff
@@ -3759,14 +3769,19 @@ GenTreeCall* Compiler::fgMorphArgs(GenTreeCall* call)
...
-                // registers and stack?
+#ifdef _TARGET_64BIT_
                 if (size > 1)
                 {
                     hasMultiregStructArgs = true;
                 }
-#endif // !_TARGET_X86_
+#elif defined(_TARGET_ARM_)
+                // TODO-Arm: Where structs passed by value can be split between
+                // registers and stack?
```
CarolEidt: Please phrase this either as a complete question or statment, e.g. "TODO-Arm: Handle the case where structs passed by value can be split between registers and stack."

CarolEidt:
:shipit:

PR #10851, opened 2017-04-10
Get rid of "#if FEATURE_RANDOMIZED_STRING_HASHING" in shared partition.

AtsushiKan:
Since CoreCLR defines it (for now) and CoreRT doesn't, let's not have
"#if's" on this in shared code. It just creates confusion.

The only ones are in StringComparer.cs and fortunately, they are quite
expendable as...

1. It's only purpose is serialize Hashtables and Dictionaries in
   a way that CLR 1.1 can deserialize them. We don't care about that in
   .NET Core.

2. In CoreCLR as of today, this infrastructure is not changing the serialization
   blob one iota since all the GetEqualityComparerForSerialization() methods
   return "this."

So this removes the appearance of downgrading the comparers to non-randomized
versions for serialization purposes (and yes, we could go on and remove
all the other "let's do it for Everett and RTM" stuff in Hashtable.cs, but
I want to keep this PR on topic. Hashtable.cs is not going to be imported
into CoreRT so it's low priority for me.)

### DIFF
--- a/src/mscorlib/shared/System/StringComparer.cs
+++ b/src/mscorlib/shared/System/StringComparer.cs
@@ -171,9 +171,6 @@ namespace System
 
     [Serializable]
     internal sealed class CultureAwareComparer : StringComparer
-#if FEATURE_RANDOMIZED_STRING_HASHING
-        , IWellKnownStringEqualityComparer
-#endif
     {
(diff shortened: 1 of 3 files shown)


AtsushiKan:

AtsushiKan:

PR #10852, opened 2017-04-10
[RyuJIT/ARM32] Encodable offsets for ld/st

### LINKED ISSUE #10793: [RyuJIT/ARM32] Assert failure on missing reservation of tmp reg
```
Assertion failed '(regSet.rsMaskResvd & RBM_OPT_RSVD) != 0'
```
```
#10 0xb16b044c in CodeGen::rsGetRsvdReg (this=<optimized out>) at /home/mskvortsov/git/coreclr/src/jit/codegen.h:142
#11 0xb16a8a24 in emitter::emitIns_R_R_I (this=0x8f8ec, ins=instruction::INS_orn, attr=emitAttr::EA_1BYTE, reg1=_regNumber_enum::REG_R0, reg2=<optimized out>, imm=<optimized out>, flags=insFlags::INS_FLAGS_DONT_CARE, opt=<optimized out>) at /home/mskvortsov/git/coreclr/src/jit/emitarm.cpp:2787
#12 0xb16b003a in emitter::emitInsLoadStoreOp (this=0x8f8ec, ins=<optimized out>, attr=<optimized out>, dataReg=<optimized out>, indir=<optimized out>) at /home/mskvortsov/git/coreclr/src/jit/emitarm.cpp:7596
```
Test cases:
```
GC/Scenarios/LeakGen/leakgen
JIT/Performance/CodeQuality/BenchI/BubbleSort2
JIT/Performance/CodeQuality/BenchI/Midpoint
JIT/Performance/CodeQuality/BenchI/Pi
JIT/Performance/CodeQuality/BenchI/TreeInsert
JIT/Regression/CLR-x86-JIT/V1-M09.5-PDC/b12624
JIT/Regression/CLR-x86-JIT/V1-M11-Beta1/b33922
JIT/Regression/CLR-x86-JIT/V1-M11-Beta1/b34945
JIT/Regression/CLR-x86-JIT/V1-M11-Beta1/b41495
JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b49335
JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b70992
JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b72828
JIT/Regression/CLR-x86-JIT/V1-M13-RTM/b92289
JIT/Regression/CLR-x86-JIT/V2.0-Beta2/b091942
```
cc @dotnet/arm32-contrib

mskvortsov:
Fixes #10793.
cc @dotnet/arm32-contrib

### DIFF
--- a/src/jit/emitarm.cpp
+++ b/src/jit/emitarm.cpp
@@ -1422,6 +1422,20 @@ DONE:
 
 /*****************************************************************************
  *
+ *  emitIns_valid_imm_for_ldst_offset() returns true when the immediate 'imm'
+ *   can be encoded as the offset in a ldr/str instruction.
+ */
+/*static*/ bool emitter::emitIns_valid_imm_for_ldst_offset(int imm, emitAttr size)
+{
+    if ((imm & 0x0fff) == imm)
+        return true; // encodable using IF_T2_K1
+    if (unsigned_abs(imm) <= 0x0ff)
+        return true; // encodable using IF_T2_H0
+    return false;
+}
+
+/*****************************************************************************
+ *
  *  Add an instruction with no operands.
  */
 
@@ -7608,10 +7622,26 @@ void emitter::emitInsLoadStoreOp(instruction ins, emitAttr attr, regNumber dataR
                 }
             }
         }
-        else
+        else // no Index
         {
-            // TODO check offset is valid for encoding
-            emitIns_R_R_I(ins, attr, dataReg, memBase->gtRegNum, offset);
+            if (emitIns_valid_imm_for_ldst_offset(offset, attr))
+            {
+                // Then load/store dataReg from/to [memBase + offset]
+                emitIns_R_R_I(ins, attr, dataReg, memBase->gtRegNum, offset);
+            }
+            else
+            {
+                // We require a tmpReg to hold the offset
+                regMaskTP tmpRegMask = indir->gtRsvdRegs;
+                regNumber tmpReg     = genRegNumFromMask(tmpRegMask);
+                noway_assert(tmpReg != REG_NA);
+
+                // First load/store tmpReg with the large offset constant
+                codeGen->instGen_Set_Reg_To_Imm(EA_PTRSIZE, tmpReg, offset);
+
+                // Then load/store dataReg from/to [memBase + tmpReg]
+                emitIns_R_R_R(ins, attr, dataReg, memBase->gtRegNum, tmpReg);
+            }
         }
     }
     else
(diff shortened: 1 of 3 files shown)


BruceForstall:
cc @dotnet/jit-contrib

mskvortsov:
@BruceForstall I'm not sure I understand you.

Here is the context of the assertion failure this PR fixes:
```
Generating: N105 ( 11, 19) [000061] -------------                il_offset void   IL offset: 35 REG NA
Generating: N107 (  1,  1) [000101] -------------      t101 =    lclVar    ref    V02 loc0         u:3 r0 (last use) REG r0 RV $2c1
                                                              ┌──▌  t101   ref    
Generating: N109 (???,???) [000159] -------------      t159 = ▌  lea(b+1024004) byref  REG NA
Generating: N111 (  1,  1) [000058] -------------       t58 =    const     int    1 reuse reg val REG r3 RV $41
IN0024:             movs    r3, 1
                                                              ┌──▌  t159   byref  
                                                              ├──▌  t58    int    
Generating: N113 (???,???) [000142] -A-XGO-------             ▌  storeIndir int    REG NA
							V02 in reg r0 is becoming dead  [000101]
							Live regs: 0071 {r0 r4 r5 r6} => 0070 {r4 r5 r6}
							Live vars: {V01 V02 V03 V04} => {V01 V03 V04}
							GC regs: 0021 {r0 r5} => 0020 {r5}

Assert failure(PID 4555 [0x000011cb], Thread: 4555 [0x11cb]): Assertion failed '(regSet.rsMaskResvd & RBM_OPT_RSVD) != 0' in 'LGen.LeakGen:MakeLeak(int):this' (IL size 52)
```
`GC/Scenarios/LeakGen/leakgen`, method `LGen.LeakGen:MakeLeak`

That large constant is used not to address a local variable on the stack frame, but to index inside a heap allocated array. How can the `compRsvdRegCheck` be relevant here?

briansull:
If a large non-encodable constant is needed in an address computation then the constant should be marked as non-contained and should be assigned its own register during the register allocation phase.
Then at codegen the path should be to load that register with the constant. and the lea instruction should be [register+register]

BruceForstall:
@mskvortsov I see now that the proposed change here simply copies what ARM64 does for ARM32 (that's the problem with too much duplicated code). If you update it to merge in my recent changes, I'll go ahead and merge it.

@briansull It seems like we should change address mode building to not generate [base + scale*index + offset] for ARM32/ARM64, esp. if the offset doesn't fit in the instruction (and currently requires a temp). It seems like we should just generate [reg + reg] and make the constant non-contained.

btw, there is one slightly bizarre thing in `CodeGen::genCreateAddrMode()`: we have the `SCALED_ADDR_MODES` `#define`, but it's always enabled, and in some places there is *also* an `#ifndef _TARGET_ARM64_`. Presumably ARM actually *will* see scaled index addressing modes.

PR #10859, opened 2017-04-10
Add back UnmanagedType.LPUTF8Str

jkotas:
Dropped it by accident when moving the file to shared CoreLib partition


alexperovich:
I don't enough logging context in the mirroring tool to see why this was skipped. I will add more so if it happens again I can diagnose.

PR #10861, opened 2017-04-10
[Experimental] Lower relops

mikedn:


mikedn:
@russellhadley You may find this experiment interesting, I was actually planning to generate BT and these changes should enable that.

russellhadley:
@mikedn are you of a mind to push this forward?  It seams like a good step.

src/jit/codegenxarch.cpp
```diff
@@ -8808,4 +8183,116 @@ void CodeGen::genAmd64EmitterUnitTests()
 
 #endif // _TARGET_AMD64_
 
+const CodeGen::GenConditionDesc& CodeGen::GetConditionDesc(GenCondition condition)
+{
+    // clang-format off
+    static constexpr GenConditionDesc map[]
```
mikedn: It's possible to flip compares (via `GenCondition::Reverse`) but that's not related to this table. This is based on the existing codegen approach to floating point compares - unfortunately some cannot be implemented using a single SETcc/Jcc. For example `bool b = float1 == float2;` is
```asm
vucomiss    xmm0,xmm1
setnp       al
jp          00007FFA1F3804DD
sete        al
```
Expanding these in lower is costly (we need to split the basic block) and unlikely to buy as anything so such expansion is still deferred to codegen.

I need to add comments to this table to show how entries map to `GenCondition` codes.

src/jit/lower.cpp
```diff
@@ -628,20 +620,14 @@ GenTree* Lowering::LowerSwitch(GenTree* node)
...
-                //   |__ GT_COND
-                //          |____GT_EQ
-                //                 |____ (switchIndex) (The temp variable)
-                //                 |____ (ICon)        (The actual case constant)
-                GenTreePtr gtCaseCond =
-                    comp->gtNewOperNode(GT_EQ, TYP_INT, comp->gtNewLclvNode(tempLclNum, tempLclType),
-                                        comp->gtNewIconNode(i, TYP_INT));
-                /* Increment the lvRefCnt and lvRefCntWtd for temp */
+                // GT_CMP switchValue, caseValue
+                // GT_JCC(EQ)
```
mikedn: Hmm, I'm not sure what you mean, this should generate identical code.

russellhadley: I was thinking more that it mapped better to how the ISA deals with it.  One compare that does all the flags, and branches that encode what bits to branch on.  This allows for single compare multiple consumer kind of things rather than keeping a single compare with the particular boolean relation in it.  (i.e. LT)

mikedn:
> are you of a mind to push this forward? It seams like a good step.

Yes, I'm going to continue with this. I also have the if-conversion experiment built on top of this because it avoids some nasty issues the existing code has (like [this](https://github.com/dotnet/coreclr/blob/master/src/jit/lsraxarch.cpp#L293) small SIMD/JTRUE optimizations that's in the wrong place).

There's an issue that I still need to solve somehow. We have dead code that reaches lowering and it is only removed after lowering, during liveness computation. After this change the dead code is removed only partially, a `GT_JCC` gets removed but its associated `GT_CMP` is not because there's no connection between them.

src/jit/lower.cpp
mikedn: Ah, that trick. One thing that I'm planning to try is to convert a switch like
```C#
switch (x) {
case 2:
case 3:
case 8:
    // do something
    break;
}
```
to generate BT/JCC.

mikedn:

src/jit/lower.cpp
mikedn: @russellhadley What use case for BT do you have in mind for MachineDepOpts?
I got it working for:
* bit tests of 64 bit "flags" (smaller code because it avoids a 64 bit immediate)
* `(x >> y) & 1` (saves an instruction)
* switches like the one mentioned above

Any other examples worth experimenting with?

russellhadley: Yes.  I'd be interested turning operations of 'or', 'xor' and 'and' of power of 2 (or 'not' pow 2) into their variants of bt{x}.  Some of these cases will feed a compare and just boil down to just the bt feeding the branch, but some won't.  (I'm just cribbing off of the native compiler peep file here)  Question: when you say you have the bit test of 64 bit "flags" working, does that mean you're recognizing trees that collapse into a test of a single bit and turning them into bt?  I was assuming that more of the cases than 64 bits would turn into a size/instruction count win.  

Cases where we should see it to start are like this:

 Replace:
```
       and   eax, X               ; X can be register, memory, or immediate.
       cmp   eax, #0              ; Must be zero.
       jeq   L1                   ; eq or ne, must be last and only use of cc, since the 
                                  ; replacement 'test' sets different flags than the 'cmp'.
```
 With:
```
       test  eax, X
       jeq   L1
```
If X is a power of 2 we can change it to 'bt' 

There's a lot of variants here just looking at the native compiler.

What kind of diffs are you getting out of our test code?

mikedn: > Question: when you say you have the bit test of 64 bit "flags" working, does that mean you're recognizing trees that collapse into a test of a single bit and turning them into bt?

The JIT already does an AND-CMP-to-TEST transform like the one you show in your example in `LowerCompare`. Where `BT` comes in handy is when you have an uncontainable immediate:
```asm
; if ((x & (1L << 42)) != 0) currently generates
       48B80000000000040000 mov      rax, 0x40000000000
       4885D0               test     rdx, rax
       740D                 je       SHORT G_M55888_IG03
; with BT this becomes
       0FBAE22A             bt       edx, 42
       770D                 ja       SHORT G_M55888_IG03
; and save 8-9 bytes of code
```
This is a kind of transform that cannot be done properly during codegen. Since the immediate value is not contained a register is allocated to it and the `mov` is generated before we get to see the `test` node and do something about it.

mikedn: > What kind of diffs are you getting out of our test code?

I only ran jitdiff on fx, I've yet to run it on tests:
* the `TEST` to `BT` transform described above saves 332 bytes (most in `System.Private.Uri` which happens to heavily use a 64-bit "flags" enum). No regressions.
* `(x >> y) & 1` appears to be a dud, 0 differences. I need to try the other variant of this pattern - `x & (1 << y)`.
* `BT` based switch table saves 3781 bytes. Diffs look like (example from https://github.com/dotnet/corefx/blob/master/src/System.Reflection.Metadata/src/System/Reflection/Internal/Utilities/BlobUtilities.cs#L337):
```diff
        41FFC9               dec      r9d
        4183F92C             cmp      r9d, 44
        7722                 ja       SHORT G_M55886_IG08
-       418BC1               mov      eax, r9d
-       4C8D0D22000000       lea      r9, [reloc @RWD00]
-       458B0C81             mov      r9d, dword ptr [r9+4*rax]
-       4C8D15BFFFFFFF       lea      r10, G_M55886_IG02
-       4D03CA               add      r9, r10
-       41FFE1               jmp      r9
+       B8FFF0FF7F           mov      eax, 0x7FFFF0FF
+       440FA3C8             bt       eax, r9d
+       7306                 jae      SHORT G_M55886_IG08
```
This seems by far the most useful use of BT and there are probably more opportunities for it:
* Roslyn sometimes use binary search instead of `switch` and my current implementation doesn't recognize that. It should be able to recognize code like `if (x == 2 || x == 3 || x == 8)` as well.
* This may be combined with switch table to binary search conversion (which the JIT currently doesn't do). That's quite a bit of work and outside the scope of this experiment.

mikedn: > I'd be interested turning operations of 'or', 'xor' and 'and' of power of 2 (or 'not' pow 2) into their variants of bt{x}

I see couple of potential uses for BTC/R/S:
* Avoid uncontained immediates like I've done for `test`. Should be relatively easy to implement but as seen in the `TEST` to `BT` case there aren't many opportunities for this, 64 bit "flags" aren't that common.
* `x | (1 << y)` (and the similar `&` and `^` variants). I suspect these are rare but this is an interesting experiment. All `BT mem, reg` instructions have poor performance and should be avoided. But how do we know in lowering if the first operand will be in memory or a register? The register allocator may or may not allocate a register to a lclvar.
* Actually use `BT{x}` for what it does best - a bit test and a bit change in the same instruction:
```C#
if ((x & 4) == 0) {
    x |= 4;
    // more code
}
```
Not sure if this can be reasonably implemented without SSA. And interestingly enough out of all popular native compilers ICC appears to be only one that handles this pattern.

russellhadley: > This is a kind of transform that cannot be done properly during codegen. Since the immediate value is not contained a register is allocated to it and the mov is generated before we get to see the test node and do something about it.

Admittedly this would need to happen pre-RA.  We could probably still do it as long as we can tell that the def of the mov is anticipated (use in the test is last use) but that still leaves the RA in the lurch with extra register pressure.  Ideally this would be done in lower so the RA doesn't have to allocated the extra kill.

mikedn:
@russellhadley I've added BTC/R/S generation for patterns mentioned above in https://github.com/dotnet/coreclr/pull/10861/commits/aea700d3903105163a0b3201e0fa0b60c6884282. Not very interesting, just some pattern matching after all. The biggest trouble turns out to be adding BT support in the emitter, it requires a lot of special casing and I'm still not sure it's all right.

Cases like `(x AND NOT(1 LSH y))` & co. are rare in fx diff, 1 in corelib, 1 in roslyn and 1 in corefx (BitArray).

I also added my old if-conversion experiment. Despite various limitation it still catches lots of cases in fx, hundreds probably.

Next experiment - another pet-peeve of mine - attempting to generate a single compare instruction for code like:
```C#
if (x < y)
    ...
else if (x > y)
    ...
else
    ...
```

karelz:
What is status of this PR? No update for last 5 months :(

mikedn:
Closing, it served its experimental purpose. I'll keep my branch for future reference and try to extract useful stuff from it.

PR #10866, opened 2017-04-10
Fix VSO bug 405852.

pgavlin:
This bug was due to a mismatch between the semantics of the FP modulus
operator as specified by ECMA-335 and as implemented by value numbering.

ECMA-335 requires the following behavior for FP modulus:
- If the divisor is 0 or the dividend is not finite, the result is NaN
- If the divisor is not finite and is not NaN, the result is the
  dividend.
Value numbering, however, simply used `fmod`, which does not have these
semantics. This change implements the required semantics in VN.

### DIFF
--- a/src/jit/valuenum.cpp
+++ b/src/jit/valuenum.cpp
@@ -206,6 +206,52 @@ T ValueNumStore::EvalOp(VNFunc vnf, T v0, T v1, ValueNum* pExcSet)
     }
 }
 
+struct FloatTraits
+{
+    static float NaN()
+    {
+        unsigned bits = 0xFFC00000u;
+        float result;
+        static_assert(sizeof(bits) == sizeof(result), "sizeof(unsigned) must equal sizeof(float)");
+        memcpy(&result, &bits, sizeof(result));
+        return result;
+    }
+};
+
+struct DoubleTraits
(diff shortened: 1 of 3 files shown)


pgavlin:
@briansull @JosephTremoulet @dotnet/jit-contrib PTAL

src/jit/valuenum.cpp
```diff
@@ -206,6 +206,52 @@ T ValueNumStore::EvalOp(VNFunc vnf, T v0, T v1, ValueNum* pExcSet)
...
+    {
+        unsigned long long bits = 0xFFF8000000000000ull;
+        double             result;
+        static_assert(sizeof(bits) == sizeof(result), "sizeof(unsigned long long) must equal sizeof(double)");
+        memcpy(&result, &bits, sizeof(result));
+        return result;
+    }
+};
+
+template <typename TFp, typename TFpTraits>
```
RussKeldorph: This seems hardly worth it even if it works, but could you make TFp be a public typedef of TFpTraits so you don't have to pass it separately?

pgavlin: I tried that, actually. The compiler didn't like me very much for it :)

PR #10867, opened 2017-04-10
Jit: fix issue with single-def type propagation

### LINKED ISSUE #10858: JIT: single-def local type propagation is too aggressive
If multiple values with different (but compatible) types can reach a the definition point of a single-def local, the jit might incorrectly update the type to the type of first reaching value.

This can lead to incorrect devirtualizations.

It's not easy to create this kind of IL with C#, but it is possible:
```C#
using System;

class B 
{
    public virtual string F() { return "B"; }
}

sealed class D : B
{
    public override string F() { return "D"; }
}

sealed class E : B
{
    public override string F() { return "E"; }
}

class X
{
    {
        // When optimizing IL, CSC will leave the newobj's on the stack
        // across the branches to the common def point.
        B b1 = (args.Length > 0) ? (B)new E() : (B)new D();
        B b2 = null;

        // Conditional flow here to forces b1 to a local instead of
        // remaining on the stack. So we have a single def point with
        // two reaching values.
        if (args.Length > 0)
        {
            b2 = new D();
        }
        else
        {
            b2 = new E();
        }

        // We should not be able to devirtualize either call.
        return b2.F()[0] + b1.F()[0] - 37;
    }
}
```

AndyAyersMS:
To avoid overly aggressive type propagation when there are multiple
reaching definitions, only propagate types to single-IL-def locals
when the defining value comes from the same basic block as the store.
We check this conservatively by insisting that the block's entry stack
be empty.

Added a test case where the jit will improperly devirtualize without
such a check.

Closes #10858.

### DIFF
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -10094,7 +10094,11 @@ void Compiler::impImportBlockCode(BasicBlock* block)
                     const bool isSingleILStoreLocal =
                         !lvaTable[lclNum].lvHasMultipleILStoreOp && !lvaTable[lclNum].lvHasLdAddrOp;
 
-                    if (isSingleILStoreLocal)
+                    // Conservative check that there is just one
+                    // definition that reaches this store.
+                    const bool hasSingleReachingDef = (block->bbStackDepthOnEntry() == 0);
+
+                    if (isSingleILStoreLocal && hasSingleReachingDef)
                     {
                         lvaUpdateClass(lclNum, op1, clsHnd);
                     }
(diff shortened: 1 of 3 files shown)


AndyAyersMS:
Should also fix various x86 desktop assertion failures; still verifying this.

7 diffs in the framework jit-diff set. They are all cases where we're now overly conservative, eg the stack is non-empty going into the block but the value that reaches the single-IL-def comes from within the block.

AndyAyersMS:
Desktop tests also look good.

PR #10869, opened 2017-04-10
Fix some static methods on ThreadPool to wait for initialization

### LINKED ISSUE #10521: Several static methods on ThreadPool don't have consistent thread-safe behavior
For example, `SetMinThreads` fails if called concurrently while another call is initializing the thread pool. It should be calling `EnsureInitialized` instead.

kouvel:
Fix for part of #10521


kouvel:
@stephentoub @gkhanna79

PR #10871, opened 2017-04-10
Package compatjit.dll in pipeline builds.

pgavlin:
Just what it says on the tin.

### DIFF
--- a/buildpipeline/DotNet-CoreClr-Trusted-Windows-x86.json
+++ b/buildpipeline/DotNet-CoreClr-Trusted-Windows-x86.json
@@ -203,7 +203,7 @@
       },
       "inputs": {
         "filename": "build-packages.cmd",
-        "arguments": "-BuildArch=$(Architecture) -BuildType=$(PB_BuildType) $(portableBuild)",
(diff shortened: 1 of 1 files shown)


pgavlin:
@dagood @wtgodbe @adiaaida PTAL

adiaaida:
LGTM.

buildpipeline/DotNet-CoreClr-Trusted-Windows-x86.json
```diff
@@ -203,7 +203,7 @@
       },
       "inputs": {
         "filename": "build-packages.cmd",
-        "arguments": "-BuildArch=$(Architecture) -BuildType=$(PB_BuildType) $(portableBuild)",
+        "arguments": "-BuildArch=$(Architecture) -BuildType=$(PB_BuildType) $(portableBuild) -ExtraParameters=/p:PackageCompatJit=1",
```
dagood: Any reason why `-ExtraParameters` instead of `--`? Looks like either way would do the same thing after looking at config.json.

pgavlin: `-ExtraParameters` is documented in the help message; AFAICT the ability to use `--` is not.

PR #10875, opened 2017-04-10
modify lastAvAddress when adjusting context for Stub

### LINKED ISSUE #5505: [JIT32/x86][GCStress=C] Loader\classloader\methodoverriding\regressions\549411\exploit fails
**Environment**
Windows x86 (JIT32)
set COMPlus_GCStress=C

**Output**

```
BEGIN EXECUTION
 "O:\tfs\pk\src\NDP\clr\bin\tests\Windows_NT.x86.Checked\Tests\Core_Root\corerun.exe" exploit.exe
FAIL
Expected: 100
Actual: 101
END EXECUTION - FAILED
FAILED
```

rahku:
Fixes #5505

### DIFF
--- a/src/vm/i386/excepx86.cpp
+++ b/src/vm/i386/excepx86.cpp
@@ -3703,6 +3703,13 @@ AdjustContextForVirtualStub(
     pExceptionRecord->ExceptionAddress = (PVOID)callsite;
     SetIP(pContext, callsite);
 
+#ifdef HAVE_GCCOVER
+    // Modify LastAVAddress saved in thread to distinguish between fake & real AV
+    // See comments in IsGcMarker in file excep.cpp for more details
+    pThread->SetLastAVAddress((LPVOID)GetIP(pContext));
+#endif    
+
+
     // put ESP back to what it was before the call.
     SetSP(pContext, dac_cast<PCODE>(dac_cast<PTR_BYTE>(GetSP(pContext)) + sizeof(void*)));
 
--- a/tests/src/Loader/classloader/methodoverriding/regressions/549411/exploit.csproj
+++ b/tests/src/Loader/classloader/methodoverriding/regressions/549411/exploit.csproj
@@ -12,7 +12,6 @@
     <OutputType>Exe</OutputType>
     <CLRTestKind>BuildAndRun</CLRTestKind>
     <CLRTestPriority>0</CLRTestPriority>
-    <GCStressIncompatible Condition="'$(Platform)' == 'x86'">true</GCStressIncompatible>
   </PropertyGroup>
 
   <ItemGroup>
--- a/tests/x86_jit32_issues.targets
+++ b/tests/x86_jit32_issues.targets
@@ -130,9 +130,6 @@
         <ExcludeList Include="$(XunitTestBinBase)\JIT\Generics\Coverage\chaos65204782cs_o\chaos65204782cs_o.cmd">
             <Issue>6714</Issue>
         </ExcludeList>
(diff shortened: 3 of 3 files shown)


rahku:

rahku:

rahku:
From comments in function [IsGCMarker](https://github.com/dotnet/coreclr/blob/master/src/vm/excep.cpp#L6928) in excep.cpp
        // on x86 we can't suspend EE to update the GC marker instruction so
        // we update it directly without suspending.  this can sometimes yield
        // a STATUS_ACCESS_VIOLATION instead of STATUS_CLR_GCCOVER_CODE.  in
        // this case we let the AV through and retry the instruction.  we'll
        // track the IP of the instruction that generated an AV so we don't
        // mix up a real AV with a "fake" AV.
        // see comments in function DoGcStress for more details on this race.
        // also make sure that the thread is actually in managed code since AVs
        // outside of of JIT code will never be potential GC markers

Essentially on x86 if there is an AV, clr exception processing mechanism needs to determine if that AV could potentially be due to race in update of GC marker instruction or a real AV. For that clr's vectored exception handler saves the location of AV address and causes resume of execution. If AV again happens at the same location then it is a real AV otherwise it is not. 

However this fails when AV happens in VSD resolve stub ( interface dispatch when target object is NULL). In this case when we hit the first AV clr's vectored exception handler saves the location of AV address to be VSD Resolve Stub and continues execution. This would av again at the same address and this time clr's [vectored exception handler](https://github.com/dotnet/coreclr/blob/master/src/vm/excep.cpp#L8273) would deem it as real AV. But further down the [execution](https://github.com/dotnet/coreclr/blob/master/src/vm/excep.cpp#L7875) we adjust the exception information to report that AV happened in managed code instead of in VSD resolve stub. Then again in ComplusFrameHandler in order to check if the current AV is a real AV or fake AV we call IsGCMarker. Now IsGCMarker starts to return true that this is a fake AV. This happens bcoz the  below condition now starts to be true
 pThread->GetLastAVAddress() != (LPVOID)GetIP(pContext) &&
This happens because GetLastAVAddress still return address in VSD resolve stub but from exception information we get AV address to the adjusted adress of managed code.

The fix is to also adjust last AV adress when adjusting the context for stubs.

rahku:
PTAL @gkhanna79

rahku:
@RussKeldorph are the changes in file tests/x86_jit32_issues.targets correct way to re-enable the failing tests

my run of Windows_NT x86 Checked Build and Test (Jit - GCStress=0xC) has three failures but I see the same results for dummy PR with no changes  https://github.com/dotnet/coreclr/pull/10887

RussKeldorph:
@rahku No, now that RyuJIT/x86 is the default, Win/x86 tests are disabled in `tests/issues.targets`.  There is a $(BuildArch) == x86 section if a failure is x86-specific.

I believe the GCStress failures are intermittent and known.  @pgavlin is investigating.

rahku:
never mind i see that it is disabled directly in csproj here https://github.com/dotnet/coreclr/blob/master/tests/src/Loader/classloader/methodoverriding/regressions/549411/exploit.csproj#L15 . I will make the changes to enable it.

rahku:

danmosemsft:
@dotnet-bot test Tizen armel Cross Debug Build  (biuld break)
@dotnet-bot test Tizen armel Cross Release Build  (biuld break)

rahku:
@gkhanna79 ping

rahku:
I also see this failure in my baseline run https://github.com/dotnet/coreclr/pull/10887

PR #10876, opened 2017-04-10
Fix a few errors in cast decomposition.

pgavlin:
The cursor was not guaranteed to be in the correct location after cast
decomposition due to the location in which the high result was typically
placed. This mispositioning can cause re-decomposition, which can lead
to asserts e.g. when encountering a GT_LONG node that was left behind by
a previously-decomposed node. This change ensures that the high result
immediately follows the cast node so that the cursor will be in the
correct location following decomposition of the cast.

### DIFF
--- a/src/jit/decomposelongs.cpp
+++ b/src/jit/decomposelongs.cpp
@@ -265,7 +265,7 @@ GenTree* DecomposeLongs::DecomposeNode(GenTree* tree)
         default:
         {
             JITDUMP("Illegal TYP_LONG node %s in Decomposition.", GenTree::NodeName(tree->OperGet()));
-            noway_assert(!"Illegal TYP_LONG node in Decomposition.");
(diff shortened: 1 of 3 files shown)


pgavlin:
@BruceForstall @mikedn @dotnet/jit-contrib PTAL

mikedn:
Looks like this also fixes a potential issue with an overflow check being done too early. Thanks!

pgavlin:
OS X and Ubuntu jobs appear to be infrastructural issues.

pgavlin:
@dotnet-bot 
test Ubuntu x64 Checked Build and Test

pgavlin:
https://github.com/dotnet/coreclr/pull/10886 fixes the *nix issues.

pgavlin:
@dotnet-bot
test Ubuntu x64 Checked Build and Test

PR #10877, opened 2017-04-10
[Local GC] Move handle null checking code to VM side

adityamandaleeka:
With this change, `ObjectHandleIsNull` and `IsHandleNullUnchecked`, which are used by the VM, are no longer in objecthandle.h.

`IsHandleNullUnchecked` is now a function in vm/gchandletableutilities.h, and the few places that were calling `ObjectHandleIsNull` now just check for NULL directly.

### DIFF
--- a/src/gc/handletable.h
+++ b/src/gc/handletable.h
@@ -216,18 +216,6 @@ FORCEINLINE BOOL HndIsNull(OBJECTHANDLE handle)
 }
 
 
-
-/*
- * inline handle checking
- */
-FORCEINLINE BOOL HndCheckForNullUnchecked(OBJECTHANDLE handle)
-{
-    LIMITED_METHOD_CONTRACT;
-
-    return (handle == NULL || (*(_UNCHECKED_OBJECTREF *)handle) == NULL);
-}
-
-
 /*
  *
  * Checks handle value for null or special value used for free handles in cache.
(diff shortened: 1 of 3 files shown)


src/vm/loaderallocator.cpp
```diff
@@ -546,7 +546,8 @@ BOOL QCALLTYPE LoaderAllocator::Destroy(QCall::LoaderAllocatorHandle pLoaderAllo
 
     BEGIN_QCALL;
 
-    if (ObjectHandleIsNull(pLoaderAllocator->GetLoaderAllocatorObjectHandle()))
+    OBJECTHANDLE objHandle = pLoaderAllocator->GetLoaderAllocatorObjectHandle();
+    if (*(Object**)objHandle == NULL)
```
jkotas: Why not to keep the `ObjectHandleIsNull`?

adityamandaleeka: No reason... I just didn't think it was valuable since the `== NULL` suffices. It was also used in just two places and in two asserts so it wasn't a big deal to modify the callers.

I don't have a particularly strong opinion on this, so if you feel strongly I can add `ObjectHandleIsNull` to gchandletableutilities.

jkotas: I think it is better to have it in gchandleutilities than to have casts that violate the abstraction in random places, even if it is just two places.

adityamandaleeka: @jkotas OK, I added it back.

jkotas:
@dotnet-bot test this please

adityamandaleeka:
@dotnet-bot test Ubuntu arm Cross Release Build (the failure is in b26496 and is obviously not related).

PR #10882, opened 2017-04-11
Remove double initialization of variables in app domain creation

adityamandaleeka:
This is minor cleanup, but I noticed we were initializing these variables to NULL multiple times here.


adityamandaleeka:
@jkotas PTAL

PR #10886, opened 2017-04-11
Fix CI test error: permission error

hseok-oh:
Fix CoreCLR/CoreFX copy order: permission problem

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -1637,6 +1637,7 @@ def static calculateBuildCommands(def newJob, def scenario, def branch, def isPR
                     // Unpack the corefx binaries
                     buildCommands += "mkdir ./bin/CoreFxBinDir"
                     buildCommands += "tar -xf ./bin/build.tar.gz -C ./bin/CoreFxBinDir"
+                    buildCommands += "sudo chmod a+x ./bin/CoreFxBinDir/corerun"
 
                     // Call the ARM CI script to cross build and test using docker
                     buildCommands += """./tests/scripts/arm32_ci_script.sh \\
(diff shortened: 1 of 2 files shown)


hseok-oh:
@dotnet-bot test ci

hseok-oh:
@gkhanna79 PTAL

pgavlin:
LGTM. Any idea why this is suddenly necessary? What changed?

cc @adityamandaleeka @janvorli

janvorli:
@pgavlin my guess is that the issue is rather at the other side where we tar the files.

pgavlin:
> my guess is that the issue is rather at the other side where we tar the files.

Yeah, that was my hypothesis as well. Is that scripting in netci.groovy?

janvorli:
I am not sure. There is some tar-ing in the corefx netci.groovy, but those lines haven't changed recently.
@mmitche do you know where the problematic package is being tar-ed?

BruceForstall:
@hseok-oh It looks like this change causes tizen jobs to fail. See

https://ci.dot.net/job/dotnet_coreclr/job/master/job/armel_cross_debug_tizen_prtest/939/console

from log:
```
10:51:32 + chmod a+x ./bin/CoreFxBinDir/corerun
10:51:32 chmod: cannot access ‘./bin/CoreFxBinDir/corerun’: No such file or directory
```

pgavlin:
@BruceForstall do you think we should roll this back for now (i.e. is having the tizen jobs clean a higher priority than having the OSX/Ubuntu jobs clean)?

BruceForstall:
@pgavlin Not sure, because the Ubuntu ARM jobs now seem to be working, where they weren't before, and they do have this code.

Issue #9597, opened 2017-02-15
Closed static delegates serialization incorrectly throws InvalidCastExceptions


fadimounir:
This code should compile and run correctly on CoreCLR, just like it does on the Desktop CLR:

`using System;
using System.IO;
using System.Reflection;
using System.Runtime.Serialization;

public static class B279085Extensions
{
    public static string StaticExt(this Program thisPtr, string a, string b)
    {
        Console.WriteLine(a + b);
        return thisPtr._id + " = RES = " + a + b;
    }
    public static string StaticExt2(this Program thisPtr, string a, string b)
    {
        Console.WriteLine(b + a);
        return thisPtr._id + " = RES = " + b + a;
    }
}

[Serializable]
public class Program
{
    internal sealed class DummyFormatterConverter : IFormatterConverter
    {
        public object Convert(object value, TypeCode typeCode)
        {
            throw new NotImplementedException();
        }

        public object Convert(object value, Type type)
        {
            throw new NotImplementedException();
        }

        public bool ToBoolean(object value)
        {
            throw new NotImplementedException();
        }

        public byte ToByte(object value)
        {
            throw new NotImplementedException();
        }

        public char ToChar(object value)
        {
            throw new NotImplementedException();
        }

        public DateTime ToDateTime(object value)
        {
            throw new NotImplementedException();
        }

        public decimal ToDecimal(object value)
        {
            throw new NotImplementedException();
        }

        public double ToDouble(object value)
        {
            throw new NotImplementedException();
        }

        public short ToInt16(object value)
        {
            throw new NotImplementedException();
        }

        public int ToInt32(object value)
        {
            throw new NotImplementedException();
        }

        public long ToInt64(object value)
        {
            throw new NotImplementedException();
        }

        public sbyte ToSByte(object value)
        {
            throw new NotImplementedException();
        }

        public float ToSingle(object value)
        {
            throw new NotImplementedException();
        }

        public string ToString(object value)
        {
            throw new NotImplementedException();
        }

        public ushort ToUInt16(object value)
        {
            throw new NotImplementedException();
        }

        public uint ToUInt32(object value)
        {
            throw new NotImplementedException();
        }

        public ulong ToUInt64(object value)
        {
            throw new NotImplementedException();
        }
    }

    public string _id;

    public static string StaticDel(string a, string b)
    {
        Console.WriteLine(a + b);
        return "RES = " + a + b;
    }
    public static string StaticDel2(string a, string b)
    {
[...]

PR #10901, opened 2017-04-11
Fix closed delegate deserialization to account for extension methods

kouvel:
Functional fix for #9597:
- The code was incorrectly assuming that the serialized target type info refers to the target object's type; it actually refers to the method's reflected type, which may not be the same type when it's an extension method


kouvel:

kouvel:

PR #10902, opened 2017-04-11
Create ELF core dump on coreclr exceptions and signals.

mikem8361:
When coreclr aborts because of a unhandled managed exception or hardware signal, the
"createdump" utility is launched if the "COMPlus_DbgEnableMiniDump" environment
variable is set to "1".

This utility uses the DAC enumerate memory interface to determine what memory regions
should be written for the managed state plus all the normal Linux memory regions for
the native state like threads, stacks, DSO, etc.

There still needs some to configure some of the "policies" like the name/location of
the core dump and the type of minidump (MiniDumpNormal, MiniDumpWithFullAuxiliaryState,
MiniDumpWithPrivateReadWriteMemory, etc.).

Currently only supported on Ubuntu Linux.


src/pal/src/thread/process.cpp
```diff
@@ -2870,10 +2932,53 @@ void PROCNotifyProcessShutdown()
...
 PAL_NORETURN
-void
+VOID
 PROCAbort()
 {
+    // Do any shutdown cleanup before aborting or creating a core dump
     PROCNotifyProcessShutdown();
+
+    // If enabled, launch the create minidump utility and wait until it completes
+    char* enabled = getenv("COMPlus_DbgEnableMiniDump");
```
janvorli: The getenv is not an async signal safe function and this code can be called from signal handlers. Can we get the setting from the env var e.g. during PAL initialization and store it to a global variable? The rest of the functions here look ok from this perspective.

src/pal/src/thread/process.cpp
```diff
@@ -2861,6 +2869,60 @@ void PROCNotifyProcessShutdown()
...
+  
+--*/
+BOOL
+PROCAbortInitialize()
+{
+    if (g_szCoreCLRPath == nullptr)
+    {
+        return FALSE;
+    }
+    char program[MAX_LONGPATH];
```
janvorli: I would prefer not to use fixed length limits on paths. We have tried to erradicate all such limits in the past. We could allocate this buffer dynamically here since we know the size we need. Then we can pass the dynamically allocated  buffer to the g_argvCreateDump[0].

src/debug/createdump/memoryregion.h
```diff
@@ -0,0 +1,98 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+struct MemoryRegion 
+{
+private:
+    // All data members are deep copy safe
```
janvorli: Even the m_fileName?

mikem8361: Removed comment. Copy/paste error.

src/pal/src/thread/process.cpp
mikem8361: Will fix.

src/pal/src/thread/process.cpp
mikem8361: Will fix.

src/debug/createdump/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
+project(createdump)
+
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+# Set the RPATH of sos so that it can find dependencies without needing to set LD_LIBRARY_PATH
```
janvorli: A nit: should be Set the RPATH of createdump ...
But actually, why does the createdump need RPATH?

src/debug/createdump/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+    dumpwriter.cpp
+)
+
+_add_executable(createdump
+    ${CREATEDUMP_SOURCES}
+)
+
+target_link_libraries(createdump
+    corguids
+    coreclrpal
```
janvorli: Looking at the sources, I can be missing something, but I don't see any other dependency on PAL than the WideCharToMultiByte at one place, some windows types that come from the PAL headers and the guids. Since it is a app that's meant purely for Unix, I wonder what was the reason to not to make it native without any PAL dependency.

mikem8361:
Yes. I’m pretty sure createdump shouldn’t have a rpath. I’ll remove it.

The place I really need the path is loading the DAC with LoadLibraryA/GetProcAddress in crashinfo.cpp line 352.

mikem8361:

mikem8361:

mikem8361:

src/debug/createdump/CMakeLists.txt
mikem8361: @janvorli Are you ok with my changes and responses to your feedback so far? Is there anything else?

mikem8361:
@jkotas would you mind taking a quick look at this changes?  Thanks.

src/debug/createdump/CMakeLists.txt
jkotas: What's the size of the binary? How much does the PAL contribute to it?

mikem8361:
The release size of createdump is 2meg.  I’m not sure how to determine the size without the PAL other than doing a lot of work not to use it.  The biggest reason for using the PAL is LoadLibraryA of the DAC module.  The DAC needs it’s DLLMain(DLL_PROCESS_ATTACH) handler called so just using dlopen/dladdr to load it duplicates a lot of code.

jkotas:
> The release size of createdump is 2meg.

This sounds like a lot given how small the tool source code is. My guess is that the PAL is large portion of it. @Petermarcu has been looking for low hanging items to reduce the disk footprint. It looks like this would be one.

> just using dlopen/dladdr to load it duplicates a lot of code.

Is it really that much code to call DLLMain in the DAC? Shouldn't it be just about dladdr("DLLMain") and call the result?

mikem8361:
Ok, your right it looks like replacing LoadLibrary isn’t that big of a deal, but what about WideCharToMultiByte?

janvorli:
As for WideCharToMultiByte, you essentially need just UnicodeToUTF8 (CP_ACP means UTF-8 on Unix). The port was done from the managed code version and there are just a few dependencies on windows specific things - WCHAR, USHORT, LPCSTR, LPWSTR, DWORD and BYTE types, wcscpy_s and wcscat_s., so you could easily reuse the code just by defining these things and putting some ifdef around the included pal/utf8.h header.

mikem8361:
@jkotas, @janvorli what about this compromise: have createdump linked to the PAL in the DAC (mscordaccore) like DBI and SOS does.  This reduced the size to 460K or so.

janvorli:
I would be ok with that, since DAC is required by this tool anyways.

jkotas:
Sounds good to me.

PR #10905, opened 2017-04-11
Fix a number of reordering issues in shift decomposition.

pgavlin:
Shift decompostition contained a fair number of IR rewrites that were
unintentionally reordering nodes. This change fixes these instances s.t.
they no longer reorder code.

### DIFF
--- a/src/jit/decomposelongs.cpp
+++ b/src/jit/decomposelongs.cpp
@@ -1013,14 +1013,24 @@ GenTree* DecomposeLongs::DecomposeShift(LIR::Use& use)
 {
     assert(use.IsInitialized());
 
-    GenTree* tree      = use.Def();
-    GenTree* gtLong    = tree->gtGetOp1();
+    GenTree* shift     = use.Def();
+    GenTree* gtLong    = shift->gtGetOp1();
     GenTree* loOp1     = gtLong->gtGetOp1();
     GenTree* hiOp1     = gtLong->gtGetOp2();
-    GenTree* shiftByOp = tree->gtGetOp2();
+    GenTree* shiftByOp = shift->gtGetOp2();
 
-    genTreeOps oper        = tree->OperGet();
+    genTreeOps oper        = shift->OperGet();
     genTreeOps shiftByOper = shiftByOp->OperGet();
+
+    // tLo = ...
+    // ...
+    // tHi = ...
+    // ...
+    // tLong = long tLo, tHi
+    // ...
+    // tShiftAmount = ...
+    // ...
+    // tShift = shift tLong, tShiftAmount
 
     assert((oper == GT_LSH) || (oper == GT_RSH) || (oper == GT_RSZ));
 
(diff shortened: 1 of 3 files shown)


pgavlin:
@BruceForstall @mikedn @dotnet/jit-contrib PTAL

src/jit/decomposelongs.cpp
```diff
@@ -1048,15 +1058,27 @@ GenTree* DecomposeLongs::DecomposeShift(LIR::Use& use)
...
                 if (count < 32)
                 {
-                    // Hi is a GT_LSH_HI, lo is a GT_LSH. Will produce:
-                    // reg1 = lo
-                    // shl lo, shift
-                    // shld hi, reg1, shift
+                    // For shifts of < 32 bits, we transform the code to:
+                    //
+                    //     tLo = ...
+                    //           st.lclVar vLo, tLo
```
mikedn: Nit: IMO using the actual `genTreeOps` member names (STORE_LCL_VAR) would be preferable.

src/jit/decomposelongs.cpp
```diff
@@ -1013,15 +1013,25 @@ GenTree* DecomposeLongs::DecomposeShift(LIR::Use& use)
 {
     assert(use.IsInitialized());
 
-    GenTree* tree      = use.Def();
-    GenTree* gtLong    = tree->gtGetOp1();
+    GenTree* shift     = use.Def();
```
mikedn: Hurray for using descriptive variable names! But this should have been done separately, a lot of changes in this diff are caused by this and not by the actual bug fix.

mikedn:
SGTM, I'm not familiar with this code. It looks like the main issue was that the original source nodes were moved right before the shift which is both incorrect and unnecessary.

pgavlin:
>  It looks like the main issue was that the original source nodes were moved right before the shift which is both incorrect and unnecessary.

Yes, that was the problem in most cases. Oddly enough, some cases were also moving nodes earlier in the function.

src/jit/decomposelongs.cpp
pgavlin: > But this should have been done separately, a lot of changes in this diff are caused by this and not by the actual bug fix.

Yes, you're right. In this case I elected to fold the changes together b/c it helped me reason about the change, but I should have undone the renames before submitting the PR.

pgavlin:
@dotnet-bot help

pgavlin:
@dotnet-bot
test Windows_NT x86 corefx_baseline
test Windows_NT x86 corefx_jitstress1
test Windows_NT x86 corefx_jitstress2

src/jit/decomposelongs.cpp
```diff
@@ -1013,15 +1013,25 @@ GenTree* DecomposeLongs::DecomposeShift(LIR::Use& use)
...
+    // tLo = ...
+    // ...
+    // tHi = ...
+    // ...
+    // tLong = long tLo, tHi
+    // ...
+    // tShiftAmount = ...
+    // ...
+    // tShift = shift tLong, tShiftAmount
+
```
briansull: This comment is confusing to me. Why do you use names like tLo, tHi  and tLong here.
These names should match your chosen local variables.
Also the "..." could be filled in, But I'm not sure that we even need this comment here, 
unless it is explaining something better than the actual code itself.

PR #10907, opened 2017-04-12
Pass null instead of empty for PB_BuildType

wtgodbe:


wtgodbe:
@gkhanna79 @chcosta @MattGal we now have build type & architecture showing up in Mission Control - https://mc.dot.net/#/product/netcore/master/source/official~2Fcoreclr~2Fmaster~2F/type/build~2Fproduct~2F/build/20170412.04/workItem/Orchestration/analysis/external/Visual%20Studio%20Build%20Information

gkhanna79:
Looks great! Do you know why we do not see OSX Portable build (see https://github.com/dotnet/coreclr/blob/master/buildpipeline/pipelines.json#L143)?

Also, it is a bit non-intuitive to locate Portable builds for Linux. From a reporting perspective, could we have a "Linux" OS and list the Portable x64 and arm32 builds under it @wtgodbe

wtgodbe:
@gkhanna79 for the OSX portable build, looks like the OperatingSystem param got set to "RedHat" (https://github.com/dotnet/coreclr/blob/master/buildpipeline/pipelines.json#L149) in https://github.com/dotnet/coreclr/pull/10894, probably a copy-paste error. As for the Linux portable builds, I'll put up a fix for that as well.

Issue #10050, opened 2017-03-09
JIT: recognize/optimize Span<T> bounds checks more like array bounds checks


JosephTremoulet:
Array bounds checks are inserted by the jit at ldelem operations, and internally we use the `GenTreeBoundsChk` operator to represent them.  This operator may raise an exception and so has exceptional flow, but not normal flow (doesn't split blocks).  Span<T> bounds checks, by contrast, are user-written compare/branch/throw sequences.  It would be handy to shape the span bound check IR more like the array bound check IR (by pattern-matching it, or by making the Span<T> indexer an intrinsic whose bounds checks the JIT is responsible for materializing, or adding attributes like [Bound] (which would apply to the length field) and [BoundsChecked] (which would apply to methods taking an index) to make the JIT more generally responsible for materializing span bounds checks).  This would both simplify modifying our bound-check-targeting optimizations like range check elim at spans and help anywhere that split blocks make our optimizations conservative (e.g. loop hoisting).

mikedn:
> or adding attributes like [Bound] (which would apply to the length field) and [BoundsChecked] (which would apply to methods taking an index)

How would the index method parameter be identified? Assumed to be the first/single `int` parameter? This attribute based solution sounds interesting (though perhaps over complicated) as it may apply to `Slice(int start)` as well. But Slice has an `(int start, int length)` overload, I wonder what, if anything, should be done about that.

JosephTremoulet:
> How would the index method parameter be identified?  Assumed to be the first/single int parameter?

Yes, that's what I was imagining, or `BoundsCheckedAttribute` could be parameterized by an `int parameterIndex`.

russellhadley:
Keep in 2.0 for Span on Kestrel.

ahsonkhan:
@JosephTremoulet, any updates on this issue?

JosephTremoulet:
> @JosephTremoulet, any updates on this issue?

Just that the idea came up in offline discussion that if we have the IndexOutOfBoundsException be raised by something that can be recognized by the JIT as an intrinsic (perhaps a direct call to the associated CORINFO helper func?), then recognizing the pattern of such a call in a block that is conditionally called based on an unsigned compare would be something the JIT could do with reasonable complexity/speed, which makes it my favorite choice for how to implement this since it also leaves flexibility in how Span and friends are implemented in runtime/library sources.

AndyAyersMS:
[...]

PR #10910, opened 2017-04-12
Jit intrinsics for Span<T>.get_Item and ReadOnlySpan<T>.get_Item.

### LINKED ISSUE #10785: Implement Span indexers as intrinsics
Create jit intrinsics for `Span<T>.get_Item` and `ReadOnlySpan<T>.get_Item`. 

Verify that indexers used in loops can have bounds checks hoisted, or identify what else needs to be done to make this possible.

See #10031 and #10050 for some discussion.

Preliminary and somewhat rough version [here](https://github.com/AndyAyersMS/coreclr/tree/SpanIndexerIntrinsic).

AndyAyersMS:
Implement these two methods as optional-expand jit intrinsics.
Uses `GT_ARR_BOUNDS_CHECK` for the bounds check so in some cases
downstream code is able to eliminate redundant checks. Fully general
support (on par with arrays in most cases) is still work in progress.

Update one bit of code in the optimizer that assumed it knew the
tree types that appeared in a `GT_ARR_BOUNDS_CHECK`.

Add benchmark tests for Span and ReadOnlySpan indexers

Tests ability of jit to reason about indexer properties with respect
to loop bounds and related indexer uses. Some cases inspired by span
indexer usage in Kestrel.

Closes #10785.

Also addresses lack of indexer inlining noted in #10031. Span indexers
should now always be inlined, even when invoked from shared methods.

### DIFF
--- a/src/inc/corinfo.h
+++ b/src/inc/corinfo.h
@@ -213,12 +213,11 @@ TODO: Talk about initializing strutures before use
     #define SELECTANY extern __declspec(selectany)
 #endif
 
-// Update this one
-SELECTANY const GUID JITEEVersionIdentifier = { /* f00b3f49-ddd2-49be-ba43-6e49ffa66959 */
-    0xf00b3f49,
-    0xddd2,
-    0x49be,
-    { 0xba, 0x43, 0x6e, 0x49, 0xff, 0xa6, 0x69, 0x59 }
+SELECTANY const GUID JITEEVersionIdentifier = { /* 2d784407-bfb1-40d0-95ea-5ad77e46f645 */
+    0x2d784407,
+    0xbfb1,
+    0x40d0,
+    {0x95, 0xea, 0x5a, 0xd7, 0x7e, 0x46, 0xf6, 0x45}
 };
 
 //////////////////////////////////////////////////////////////////////////////////////////////////////////
@@ -959,6 +958,8 @@ enum CorInfoIntrinsics
     CORINFO_INTRINSIC_GetManagedThreadId,
     CORINFO_INTRINSIC_ByReference_Ctor,
     CORINFO_INTRINSIC_ByReference_Value,
+    CORINFO_INTRINSIC_Span_GetItem,
+    CORINFO_INTRINSIC_ReadOnlySpan_GetItem,
 
     CORINFO_INTRINSIC_Count,
     CORINFO_INTRINSIC_Illegal = -1,         // Not a true intrinsic,
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -3652,6 +3652,80 @@ GenTreePtr Compiler::impIntrinsic(GenTreePtr            newobjThis,
             retNode                     = field;
             break;
         }
+        case CORINFO_INTRINSIC_Span_GetItem:
+        case CORINFO_INTRINSIC_ReadOnlySpan_GetItem:
+        {
+            // Have index, stack pointer-to Span<T> s on the stack.
+            //
+            // Implement
+            //     if (index >= s->_length) throw new IndexOutOfRangeException();
+            //     return s->_pointer + index * sizeof(T) (for Span)
+            //     return *(s->_pointer + index * sizeof(T)) (for ReadOnlySpan)
+            //
+            // Signature should show one class type parameter
+            assert(sig->sigInst.classInstCount == 1);
+            CORINFO_CLASS_HANDLE spanElemHnd = sig->sigInst.classInst[0];
+            const unsigned       elemSize    = info.compCompHnd->getClassSize(spanElemHnd);
+            assert(elemSize > 0);
+
+            const bool isReadOnly = (intrinsicID == CORINFO_INTRINSIC_ReadOnlySpan_GetItem);
+
+            JITDUMP("\nimpIntrinsic: Expanding %sSpan<T>.get_Item, T=%s, sizeof(T)=%u\n", isReadOnly ? "ReadOnly" : "",
+                    info.compCompHnd->getClassName(spanElemHnd), elemSize);
+
+            GenTreePtr index          = impPopStack().val;
+            GenTreePtr ptrToSpan      = impPopStack().val;
+            GenTreePtr indexClone     = nullptr;
+            GenTreePtr ptrToSpanClone = nullptr;
+
+            if (verbose)
+            {
+                printf("with ptr-to-span\n");
+                gtDispTree(ptrToSpan);
+                printf("and index\n");
+                gtDispTree(index);
+            }
+
+            // We need to use both index and ptr-to-span twice, so clone or spill.
+            index = impCloneExpr(index, &indexClone, NO_CLASS_HANDLE, (unsigned)CHECK_SPILL_ALL,
+                                 nullptr DEBUGARG("Span.get_Item index"));
+            ptrToSpan = impCloneExpr(ptrToSpan, &ptrToSpanClone, NO_CLASS_HANDLE, (unsigned)CHECK_SPILL_ALL,
+                                     nullptr DEBUGARG("Span.get_Item ptrToSpan"));
+
+            // Bounds check
+            CORINFO_FIELD_HANDLE lengthHnd    = info.compCompHnd->getFieldInClass(clsHnd, 1);
+            const unsigned       lengthOffset = info.compCompHnd->getFieldOffset(lengthHnd);
+            GenTreePtr           length       = gtNewFieldRef(TYP_INT, lengthHnd, ptrToSpan, lengthOffset, false);
+            GenTreePtr           boundsCheck  = new (this, GT_ARR_BOUNDS_CHECK)
+                GenTreeBoundsChk(GT_ARR_BOUNDS_CHECK, TYP_VOID, index, length, SCK_RNGCHK_FAIL);
+
+            // Element access
+            GenTreePtr           index2long = gtNewCastNode(TYP_I_IMPL, indexClone, TYP_I_IMPL);
+            GenTreePtr           sizeofNode = gtNewIconNode(elemSize);
+            GenTreePtr           mulNode    = gtNewOperNode(GT_MUL, TYP_I_IMPL, index2long, sizeofNode);
+            CORINFO_FIELD_HANDLE ptrHnd     = info.compCompHnd->getFieldInClass(clsHnd, 0);
+            const unsigned       ptrOffset  = info.compCompHnd->getFieldOffset(ptrHnd);
+            GenTreePtr           ptrToSpan2 = gtClone(ptrToSpan, true);
+            GenTreePtr           data       = gtNewFieldRef(TYP_BYREF, ptrHnd, ptrToSpanClone, ptrOffset, false);
+            GenTreePtr           result     = gtNewOperNode(GT_ADD, TYP_BYREF, data, mulNode);
+
+            // Prepare result
+            var_types resultType = JITtype2varType(sig->retType);
+
+            if (isReadOnly)
+            {
+                result = gtNewOperNode(GT_IND, resultType, result);
+            }
+            else
+            {
+                assert(resultType == result->TypeGet());
+            }
+
+            retNode = gtNewOperNode(GT_COMMA, resultType, boundsCheck, result);
+
+            break;
+        }
+
         default:
             /* Unknown intrinsic */
             break;
--- a/src/vm/jitinterface.cpp
+++ b/src/vm/jitinterface.cpp
@@ -8634,9 +8634,9 @@ CorInfoIntrinsics CEEInfo::getIntrinsicID(CORINFO_METHOD_HANDLE methodHnd,
     else
     {
         MethodTable * pMT = method->GetMethodTable();
-        if (pMT->IsByRefLike() && pMT->GetModule()->IsSystem())
+        if (pMT->GetModule()->IsSystem())
         {
-            if (pMT->HasSameTypeDefAs(g_pByReferenceClass))
+            if (pMT->IsByRefLike() && pMT->HasSameTypeDefAs(g_pByReferenceClass))
             {
                 // ByReference<T> has just two methods: constructor and Value property
                 if (method->IsCtor())
@@ -8648,10 +8648,25 @@ CorInfoIntrinsics CEEInfo::getIntrinsicID(CORINFO_METHOD_HANDLE methodHnd,
                     _ASSERTE(strcmp(method->GetName(), "get_Value") == 0);
                     result = CORINFO_INTRINSIC_ByReference_Value;
                 }
-                *pMustExpand = true;
+                if (pMustExpand != nullptr)
+                {
+                    *pMustExpand = true;
+                }
             }
-
-            // TODO-SPAN: Span<T> intrinsics for optimizations
+            else if (pMT->HasSameTypeDefAs(MscorlibBinder::GetClass(CLASS__SPAN)))
+            {
+                if (method->HasSameMethodDefAs(MscorlibBinder::GetMethod(METHOD__SPAN__GET_ITEM)))
+                {
+                    result = CORINFO_INTRINSIC_Span_GetItem;
+                }
+            }
+            else if (pMT->HasSameTypeDefAs(MscorlibBinder::GetClass(CLASS__READONLY_SPAN)))
+            {
+                if (method->HasSameMethodDefAs(MscorlibBinder::GetMethod(METHOD__READONLY_SPAN__GET_ITEM)))
+                {
+                    result = CORINFO_INTRINSIC_ReadOnlySpan_GetItem;
+                }
+            }
         }
     }
 
--- a/src/vm/method.cpp
+++ b/src/vm/method.cpp
@@ -2404,7 +2404,7 @@ BOOL MethodDesc::IsFCallOrIntrinsic()
     if (IsFCall() || IsArray())
         return TRUE;
 
-    // Intrinsic methods on ByReference<T> or Span<T>
+    // Intrinsic methods on ByReference<T>, Span<T>, or ReadOnlySpan<T>
     MethodTable * pMT = GetMethodTable();
     if (pMT->IsByRefLike() && pMT->GetModule()->IsSystem())
         return TRUE;
--- a/tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
+++ b/tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
@@ -1,0 +1,925 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Text;
+using Xunit;
+using Microsoft.Xunit.Performance;
+
+[assembly: OptimizeForBenchmarks]
+[assembly: MeasureInstructionsRetired]
+
+namespace Span
+{
+    class Sink
+    {
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        public static Sink NewSink() { return new Sink(); }
+
+        public byte b;
+        public int i;
+    }
+
+    [AttributeUsage(AttributeTargets.Method, Inherited = false)]
+    class CategoryAttribute : Attribute
+    {
+        public CategoryAttribute(string name)
+        {
+            _name = name;
+        }
+        string _name;
+        public string Name => _name;
+    }
+
+    class IndexerBench
+    {
+        const int Iterations = 1000000;
+        const int DefaultLength = 1024;
+        const byte Expected = 70;
+        static bool HasFailure = false;
+
+        [Benchmark(InnerIterationCount = Iterations)]
+        [InlineData(DefaultLength)]
+        [Category("Indexer in-loop bounds check elimination")]
+        public static void Ref(int length)
+        {
+            byte[] a = GetData(length);
+
+            Invoke((int innerIterationCount) =>
+            {
+                Span<byte> s = new Span<byte>(a);
+                byte result = 0;
+                for (int i = 0; i < innerIterationCount; ++i)
+                {
+                    result = TestRef(s);
+                }
+                return result;
+            },
+            "Ref({0})", length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static byte TestRef(Span<byte> data)
+        {
+            ref byte p = ref data.DangerousGetPinnableReference();
+            int length = data.Length;
+            byte x = 0;
+
+            for (var idx = 0; idx < length; idx++)
+            {
+                x ^= Unsafe.Add(ref p, idx);
+            }
+
+            return x;
+        }
+
+        [Benchmark(InnerIterationCount = Iterations)]
+        [InlineData(DefaultLength)]
+        [Category("Indexer in-loop bounds check elimination")]
+        public static void Fixed1(int length)
+        {
+            byte[] a = GetData(length);
+
+            Invoke((int innerIterationCount) =>
+            {
+                Span<byte> s = new Span<byte>(a);
+                byte result = 0;
+                for (int i = 0; i < innerIterationCount; ++i)
+                {
+                    result = TestFixed1(s);
+                }
+                return result;
+            },
+            "Fixed1({0})", length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static unsafe byte TestFixed1(Span<byte> data)
+        {
+            fixed (byte* pData = &data.DangerousGetPinnableReference())
+            {
+                int length = data.Length;
+                byte x = 0;
+                byte* p = pData;
+
+                for (var idx = 0; idx < length; idx++)
+                {
+                    x ^= *(p + idx);
+                }
+
+                return x;
+            }
+        }
+
+        [Benchmark(InnerIterationCount = Iterations)]
+        [InlineData(DefaultLength)]
+        [Category("Indexer in-loop bounds check elimination")]
+        public static void Fixed2(int length)
+        {
+            byte[] a = GetData(length);
+
+            Invoke((int innerIterationCount) =>
+            {
+                Span<byte> s = new Span<byte>(a);
+                byte result = 0;
+                for (int i = 0; i < innerIterationCount; ++i)
+                {
+                    result = TestFixed2(s);
+                }
+                return result;
+            },
+            "Fixed2({0})", length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static unsafe byte TestFixed2(Span<byte> data)
+        {
+            fixed (byte* pData = &data.DangerousGetPinnableReference())
+            {
+                int length = data.Length;
+                byte x = 0;
+
+                for (var idx = 0; idx < length; idx++)
+                {
+                    x ^= pData[idx];
+                }
+
+                return x;
+            }
+        }
+
+        [Benchmark(InnerIterationCount = Iterations)]
+        [InlineData(DefaultLength)]
+        [Category("Indexer in-loop bounds check elimination")]
+        public static void Indexer1(int length)
+        {
+            byte[] a = GetData(length);
+
+            Invoke((int innerIterationCount) =>
+            {
+                Span<byte> s = new Span<byte>(a);
+                byte result = 0;
+                for (int i = 0; i < innerIterationCount; ++i)
+                {
+                    result = TestIndexer1(s);
+                }
+                return result;
+            },
+            "Indexer1({0})", length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static byte TestIndexer1(Span<byte> data)
+        {
+            int length = data.Length;
+            byte x = 0;
+
+            for (var idx = 0; idx < length; idx++)
+            {
+                x ^= data[idx];
+            }
+
+            return x;
+        }
+
+        [Benchmark(InnerIterationCount = Iterations)]
+        [InlineData(DefaultLength)]
+        [Category("Indexer in-loop bounds check elimination")]
+        public static void Indexer2(int length)
+        {
+            byte[] a = GetData(length);
+
+            Invoke((int innerIterationCount) =>
+            {
+                Span<byte> s = new Span<byte>(a);
+                byte result = 0;
+                for (int i = 0; i < innerIterationCount; ++i)
+                {
+                    result = TestIndexer2(s);
+                }
+                return result;
+            },
+            "Indexer2({0})", length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static byte TestIndexer2(Span<byte> data)
+        {
+            byte x = 0;
+
+            for (var idx = 0; idx < data.Length; idx++)
+            {
+                x ^= data[idx];
+            }
+
+            return x;
+        }
+
+        [Benchmark(InnerIterationCount = Iterations)]
+        [InlineData(DefaultLength)]
+        [Category("Indexer in-loop bounds check elimination")]
+        public static void Indexer3(int length)
+        {
+            byte[] a = GetData(length);
+
+            Invoke((int innerIterationCount) =>
+            {
+                Span<byte> s = new Span<byte>(a);
+                byte result = 0;
+                for (int i = 0; i < innerIterationCount; ++i)
+                {
+                    result = TestIndexer3(s);
+                }
+                return result;
+            },
+            "Indexer3({0})", length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static byte TestIndexer3(Span<byte> data)
+        {
+            Span<byte> data2 = data;
+
+            byte x = 0;
+
+            for (var idx = 0; idx < data2.Length; idx++)
+            {
+                x ^= data2[idx];
+            }
+
+            return x;
+        }
+
+        [Benchmark(InnerIterationCount=Iterations / 10)]
+        [InlineData(DefaultLength)]
+        [Category("Indexer in-loop bounds check elimination")]
+        public static void Indexer4(int length)
+        {
+            byte[] a = GetData(length);
+
+            Invoke((int innerIterationCount) =>
+            {
+                Span<byte> s = new Span<byte>(a);
+                byte result = 0;
+                int inner = Math.Max(1, (innerIterationCount / 10));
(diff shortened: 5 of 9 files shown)


AndyAyersMS:
@JosephTremoulet PTAL
cc @dotnet/jit-contrib @jkotas 

This lays groundwork and improves indexer perf, but does not address all the issues.

Will post some benchmark data and code snippets later.

src/inc/corinfo.h
```diff
@@ -213,12 +213,11 @@ TODO: Talk about initializing strutures before use
...
-SELECTANY const GUID JITEEVersionIdentifier = { /* f00b3f49-ddd2-49be-ba43-6e49ffa66959 */
-    0xf00b3f49,
-    0xddd2,
-    0x49be,
-    { 0xba, 0x43, 0x6e, 0x49, 0xff, 0xa6, 0x69, 0x59 }
+SELECTANY const GUID JITEEVersionIdentifier = { /* 2d784407-bfb1-40d0-95ea-5ad77e46f645 */
+    0x2d784407,
+    0xbfb1,
+    0x40d0,
+    {0x95, 0xea, 0x5a, 0xd7, 0x7e, 0x46, 0xf6, 0x45}
```
briansull: There have been a lot of GUID changes recently.
Is this required since we are only extending an enum?
 @sandreenko

AndyAyersMS: We could indeed defer the GUID update -- this is not a "must expand" intrinsic, so an older jit could just use the IL versions.

sandreenko: Yes, spmi should work fine without GUID update. It will show didn't find a key for this methods, but the other part will be valid.

AndyAyersMS:
x86 is using `GT_LCL_FLD` for the length access in the bounds check, so need to tweak the code in `optExtractArrIndex` a bit to compensate.

AndyAyersMS:
Ok, GUID left as is....

AndyAyersMS:
Results on the new Indexer benchmarks (Windows, x64). Some of these tests are
unfortunately susceptible to loop head alignment effects.

The intrinsic version is generally better with a couple regressions. Will add some analyses
of the regression cases subsequently.

Known length spans are now on par with the ref/fixed indexing (which
avoids using the indexer at all to bypass the bounds checks).


| Test                     |    Old  |     New | Change | Notes |
| :----------------------- | ------: | ------: | :----- | ----- |
|                Ref(1024) |  509.14 |  509.46 |        |       |
|             Fixed1(1024) |  505.08 |  511.06 |        |       |
|             Fixed2(1024) |  580.58 |  587.17 |        | Should match Fixed1. See dotnet/roslyn#18615 |
|           Indexer1(1024) |  665.05 |  615.26 | Better |       |
|           Indexer2(1024) |  661.12 |  620.36 | Better |       |
|           Indexer3(1024) |  616.77 |  616.24 |        |       |
|           Indexer4(1024) |  683.95 |  691.43 |        |       |
|           Indexer5(1024) |  783.21 |  912.77 | Worse  |       |
|           Indexer6(1024) |  899.16 | 1015.29 | Worse  |       |
|   ReadOnlyIndexer1(1024) |  666.35 |  620.83 | Better |       |
|   ReadOnlyIndexer2(1024) |  667.29 |  621.27 | Better |       |
|   WriteViaIndexer1(1024) | 2342.12 | 2343.55 |        |       |
|   WriteViaIndexer2(1024) | 1164.56 | 1120.31 |        |       |
|     KnownSizeArray(1024) |  613.33 |  499.22 | Better | No bounds checks left |
|      KnownSizeCtor(1024) |  616.80 |  506.78 | Better | No bounds checks left |
|     KnownSizeCtor2(1024) |  623.13 |  525.48 | Better | No bounds checks left |
|         SameIndex1(1024) |  997.03 |  833.67 | Better |       |
|         SameIndex2(1024) | 1309.56 | 1514.99 | Worse  |       |
|      CoveredIndex1(1024) | 1363.13 | 1191.16 | Better |       |

src/jit/importer.cpp
```diff
@@ -3652,6 +3652,82 @@ GenTreePtr Compiler::impIntrinsic(GenTreePtr            newobjThis,
             retNode                     = field;
             break;
         }
+        case CORINFO_INTRINSIC_Span_GetItem:
+        case CORINFO_INTRINSIC_ReadOnlySpan_GetItem:
+        {
+            // Have index, stack pointer-to Span<T> s on the stack.
+            //
+            // Implement
+            //     if (index >= s->_length) throw new IndexOutOfRangeException();
```
JosephTremoulet: This comment makes it look like we're not checking for negative indices.  Maybe it should read `if (index < 0 || index >= s->_length)`, or `if ((unsigned) index >= (unsigned) s->_length)`, or maybe this line should just be `BoundsCheck(index, s->_length)`

src/jit/importer.cpp
```diff
@@ -3652,6 +3652,82 @@ GenTreePtr Compiler::impIntrinsic(GenTreePtr            newobjThis,
...
+
+            // Bounds check
+            CORINFO_FIELD_HANDLE lengthHnd    = info.compCompHnd->getFieldInClass(clsHnd, 1);
+            const unsigned       lengthOffset = info.compCompHnd->getFieldOffset(lengthHnd);
+            GenTreePtr           length       = gtNewFieldRef(TYP_INT, lengthHnd, ptrToSpan, lengthOffset, false);
+            GenTreePtr           boundsCheck  = new (this, GT_ARR_BOUNDS_CHECK)
+                GenTreeBoundsChk(GT_ARR_BOUNDS_CHECK, TYP_VOID, index, length, SCK_RNGCHK_FAIL);
+
+            // Element access
+            GenTreePtr           index2long = gtNewCastNode(TYP_I_IMPL, indexClone, TYP_I_IMPL);
```
JosephTremoulet: Avoid the cast on 32-bit?  Maybe call this index2intptr?

src/jit/importer.cpp
```diff
@@ -3652,6 +3652,82 @@ GenTreePtr Compiler::impIntrinsic(GenTreePtr            newobjThis,
...
+            GenTreePtr           sizeofNode = gtNewIconNode(elemSize);
+            GenTreePtr           mulNode    = gtNewOperNode(GT_MUL, TYP_I_IMPL, index2long, sizeofNode);
+            CORINFO_FIELD_HANDLE ptrHnd     = info.compCompHnd->getFieldInClass(clsHnd, 0);
+            const unsigned       ptrOffset  = info.compCompHnd->getFieldOffset(ptrHnd);
+            GenTreePtr           ptrToSpan2 = gtClone(ptrToSpan, true);
```
JosephTremoulet: Looks like `ptrToSpan2` is unused now (in favor of `ptrToSpanClone`)

src/vm/jitinterface.cpp
```diff
@@ -8634,9 +8634,9 @@ CorInfoIntrinsics CEEInfo::getIntrinsicID(CORINFO_METHOD_HANDLE methodHnd,
     else
     {
         MethodTable * pMT = method->GetMethodTable();
-        if (pMT->IsByRefLike() && pMT->GetModule()->IsSystem())
+        if (pMT->GetModule()->IsSystem())
         {
-            if (pMT->HasSameTypeDefAs(g_pByReferenceClass))
+            if (pMT->IsByRefLike() && pMT->HasSameTypeDefAs(g_pByReferenceClass))
```
JosephTremoulet: Curious why you changed this (and if we need to worry that `GetModule()` or `IsSystem()` is significantly more expensive that `IsByRefLike()`).  Also, Span and ReadOnlySpan are both "byref-like" in the sense I'm used to that term being used (transitively contains a managed pointer, can't be put in the heap, etc.), are they not `IsByRefLike()` per this check?

src/vm/method.cpp
```diff
@@ -2404,7 +2404,7 @@ BOOL MethodDesc::IsFCallOrIntrinsic()
     if (IsFCall() || IsArray())
         return TRUE;
 
-    // Intrinsic methods on ByReference<T> or Span<T>
+    // Intrinsic methods on ByReference<T>, Span<T>, or ReadOnlySpan<T>
     MethodTable * pMT = GetMethodTable();
     if (pMT->IsByRefLike() && pMT->GetModule()->IsSystem())
```
JosephTremoulet: Looks like you left the `IsByRefLike()` guard here; is there a reason it's different than above?

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
```diff
@@ -0,0 +1,925 @@
...
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static byte TestKnownSizeArray(int innerIterationCount)
+        {
+            byte[] a = new byte[1024];
+            SetData(a);
+            Span<byte> data = new Span<byte>(a);
+            byte x = 0;
+
+            for (int i = 0; i < innerIterationCount; i++)
+            {
```
JosephTremoulet: Was there a particular reason to put both loop nests in the inner method here and in the next test?  Just to drown out the cost of allocating the array?

src/jit/importer.cpp
AndyAyersMS: Will update to show expansions in pseudo IR.

src/jit/importer.cpp
AndyAyersMS: Sure, we have `impImplicitIorI4Cast` we can use here.

src/jit/importer.cpp
AndyAyersMS: Will remove...

src/vm/jitinterface.cpp
AndyAyersMS: `IsByRefLike` is cheap. I'll move it back up to check along with the module.

src/vm/method.cpp
AndyAyersMS: Not any good reason. Fixing the earlier check to be consistent.

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
AndyAyersMS: In `Indexer4` the indexer-loop CQ is worse when tis nested inside some other loop; these tests highlights that issue.

Later tests like `KnownSizeArray` need to see the array creation site to propagate the length so everything has to be in one method.

JosephTremoulet: I may have misplaced the comment; I was curious in `KnownSizeArray` and friends why not generate a new array per inner iteration.  (not that it's particularly concerning, just curious since I noticed it departing from the pattern of inner iteration counting calls to the kernel method)

AndyAyersMS:
Still issues to sort out with the assumptions made in `optExtractdArrIndex`. Also CI system seems to be having various odd issues....

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
AndyAyersMS: I wanted to try and keep these more or less time-equivalent to the main indexer tests and though an array creation or array copy per inner iteration would skew the results quite a bit.

AndyAyersMS:
The regressions on Indexer5, Indexer6, and SameIndex2 all have the same root cause: extra null checks on the span byref. Here's Indexer5's inner loop:

<table>
<tr>
  <th>Before</th>
  <th>After</th>
</tr>
<tr>
<td>
<pre>
G_M36436_IG03:
   cmp      r8d, dword ptr [rcx+8]
   jae      SHORT G_M36436_IG06
   ...
   mov      r9, bword ptr [rcx]
   movsxd   r10, r8d
   movzx    r9, byte  ptr [r9+r10]
   xor      r9d, eax
   movzx    rax, r9b
   mov      dword ptr [rdx], r8d
   inc      r8d
   cmp      r8d, dword ptr [rcx+8]
   jl       SHORT G_M36436_IG03
</pre>
</td>
<td>
<pre>
G_M36436_IG03:
   cmp      r8d, dword ptr [rcx+8]
   jae      SHORT G_M36433_IG05
   cmp      dword ptr [rcx], ecx
   mov      r9, bword ptr [rcx]
   movsxd   r10, r8d
   movzx    r9, byte  ptr [r9+r10]
   xor      r9d, eax
   movzx    rax, r9b
   mov      dword ptr [rdx], r8d
   inc      r8d
   cmp      r8d, dword ptr [rcx+8]
   jl       SHORT G_M36433_IG03
</pre>
</td>
</tr>
</table>

AndyAyersMS:
Drilling in a bit on the regression cases: in the before case we copy the `_pointer` field (type `ByReference<T>`) to a local as part of inlining. So we know the local's address is not null. In the after case we're still looking at the original byref span argument, which we worry may be a null pointer. 

In the after case the tree in question is:
```
[000040] *--XG-------             |  |  /--*  indir     ubyte 
[000035] ------------             |  |  |  |        /--*  const     int    1
[000036] ------------             |  |  |  |     /--*  *         long  
[000034] ------------             |  |  |  |     |  \--*  cast      long <- int
[000029] ------------             |  |  |  |     |     \--*  lclVar    int    V03 loc1         
[000038] ----G-------             |  |  |  |  /--*  +         byref 
[000037] ----G-------             |  |  |  |  |  \--*  field     byref  _pointer
[000030] ------------             |  |  |  |  |     \--*  lclVar    byref  V00 arg0         
[000039] ---XG-------             |  |  |  \--*  comma     byref 
[000033] ---XG-------             |  |  |     \--*  arrBndsChk_Rng void  
[000028] ------------             |  |  |        +--*  lclVar    int    V03 loc1         
[000032] ----G-------             |  |  |        \--*  field     int    _length
[000026] ------------             |  |  |           \--*  lclVar    byref  V00 arg0         
       
```
Seems like we should be able to mitigate this by spilling `_pointer` to a local during the intrinsic expansion.

On the other hand we shouldn't have to do this -- we clearly recognize that the `_length` access in the first part of the expansion doesn't need an explicit check. The difference is that morph seemingly gets confused by the address arithmetic sitting above the `_pointer` access and doesn't realize there is another dereference to get to `_pointer`, and somehow thinks that indefinite-offset math should apply to the `_pointer` access too.

AndyAyersMS:
Spilling seems doable but a bit awkward -- either we need another level of comma or the `_pointer` load happens before the bounds check. The latter is probably ok.

May just defer on this since the perf issues only happen when there are writes in the loop and we have other issues with writes....

JosephTremoulet:
>  morph seemingly gets confused by the address arithmetic sitting above the _pointer access and doesn't realize there is another dereference to get to _pointer, and somehow thinks that indefinite-offset math should apply to the _pointer access too.

That seems bizarre, the code is trying to be careful.  I'd recommend at least filing an issue for that so we can take a closer look and fix whatever bug may be there.

AndyAyersMS:
Looking at SpanBench, there are 3 notable improvements and 3 regressions:
```
Improved:
TestSpanIndexHoistable<String> -- indexer inline
TestSpanIndexVariant<String> -- indexer inline
TestBubbleSortSpan -- fewer bounds checks

Regressed:
TestFillAllSpan -- extra null check
TestFillAllReverseSpan -- extra null check
TestQuickSortSpan -- extra null checks
```
So far all the regressions are caused by the extra null checks.

AndyAyersMS:
Here's a simple example that doesn't involve span. `P0`'s explicit null check is unnecessary.

I'll open a new issue for this, though for some reason I'm unable to just now. Should we hold off on merging the above until this is fixed?

```C#
unsafe struct F
{
    public F(int* x) { p = x; }
    public int* p;
}

static unsafe int P1(ref F f)
{
    return *(f.p);
}

static unsafe int P0(ref F f, int v)
{
    return *(f.p + v * 8);
}
```
```asm
;; P1
mov      rax, qword ptr [rcx]
mov      eax, dword ptr [rax]
ret

;; P0
cmp      dword ptr [rcx], ecx
mov      rax, qword ptr [rcx]
shl      edx, 3
movsxd   rdx, edx
ret
```

JosephTremoulet:
> Should we hold off on merging the above until this is fixed?

I'm inclined to say no, but I would like to understand why the sequence from inlining the explicit indexer method isn't hitting the same issue -- what does the IR from the same point in the phase list look like in that case?

AndyAyersMS:
The inliner introduces temps that cache the field values, so the code in morph doesn't hit this path. We could do the same in the intrinsic expansion, but getting the temp assignment in the right place requires another level of comma.

Initial fix for #10942 looks encouraging and takes care of the Indexer regressions. Checking SpanBench now.

AndyAyersMS:
SpanBench regressions called out earlier also taken care of by #10942.

I still see two sets of slowdowns in SpanBench that I can't easily explain:  TestSpanDangerousCreate<String> slows down by about 15% (47 -> 53) and TestSpanDangerousGetPinnableReference<String> by maybe 20% (16 -> 19). 

In DangerousCreate, the codegen diff in the test loop is that the span indexer gets inlined. However this is in dynamically unreachable part of the loop guarded by `if (untrue)` and the parts of the loop that are reachable have identical code, so it is hard to see how the indexer inlining matters. Both loops call the prejitted version of Span.DangerousCreate but the code for that is identical too.

In DangerousGetPinnableReference the innermost loop is identical code.

Can't get xunit-perf running just now or I'd try and verify that instructions retired is similar.

mikedn:
> However this is in dynamically unreachable part of the loop guarded by if (untrue) and the parts of the loop that are reachable have identical code, so it is hard to see how the indexer inlining matters. 

Does the loop body increases in size?

AndyAyersMS:
The new version is shorter because of the indexer intrinsic. Calling the indexer requires a runtime lookup which is a lot of code.
<details>
<table align="left">
<tr><th>Before</th><th>After</th></tr>
<tr valign="top">
<td><pre><code>
G_M28281_IG04:
  mov      r13, gword ptr [rsi+8]
  mov      rcx, r13
  mov      r8, qword ptr [r14]
  test     r8d, 1
  jne      SHORT G_M28281_IG05
  jmp      SHORT G_M28281_IG06
&nbsp;
G_M28281_IG05:
  mov      r8, qword ptr [r8+FFFFFFFFH]
&nbsp;
G_M28281_IG06:
  xor      edx, edx
  call     CORINFO_HELP_LDELEMA_REF
  mov      bword ptr [rsp+28H], rax
  mov      r13d, dword ptr [r13+8]
  mov      rdx, qword ptr [r14+16]
  test     rdx, rdx
  jne      G_M28281_IG12
  mov      rcx, rbx
  mov      rdx, 0x7FFF48587220
  call     CORINFO_HELP_RUNTIMEHANDLE_METHOD
  mov      rdx, rax
  mov      rax, bword ptr [rsp+28H]
  mov      dword ptr [rsp+20H], r13d
  lea      rcx, bword ptr [rsp+30H]
  mov      r8, rsi
  mov      r9, rax
  call     Span:DangerousCreate(...)
  test     bpl, bpl
; *** this branch always taken
  je       SHORT G_M28281_IG10 
  mov      rdx, qword ptr [r14+16]
  test     rdx, rdx
  jne      SHORT G_M28281_IG08
  mov      rcx, rbx
  mov      rdx, 0x7FFF48587220
  call     CORINFO_HELP_RUNTIMEHANDLE_METHOD
  mov      rdx, rax
&nbsp;
G_M28281_IG08:
  lea      rcx, bword ptr [rsp+30H]
  xor      r8d, r8d
  call     Span:get_Item(int):byref:this
  mov      rdx, gword ptr [rax]
  lea      rcx, bword ptr [r15+8]
  call     CORINFO_HELP_ASSIGN_REF
  mov      rcx, qword ptr [r14+24]
  test     rcx, rcx
  jne      SHORT G_M28281_IG09
  mov      rcx, rbx
  mov      rdx, 0x7FFF48587230
  call     CORINFO_HELP_RUNTIMEHANDLE_METHOD
  mov      rcx, rax
&nbsp;
G_M28281_IG09:
  call     CORINFO_HELP_NEWSFAST
  mov      rsi, rax
&nbsp;
G_M28281_IG10:
  inc      r12d
  cmp      r12d, edi
  jl       G_M28281_IG04
</td></pre></code>
<td><pre><code>
G_M28280_IG04:
  mov      r13, gword ptr [rsi+8]
  mov      rcx, r13
  mov      r8, qword ptr [r14]
  test     r8d, 1
  jne      SHORT G_M28280_IG05
  jmp      SHORT G_M28280_IG06
&nbsp;
G_M28280_IG05:
  mov      r8, qword ptr [r8+FFFFFFFFH]
&nbsp;
G_M28280_IG06:
  xor      edx, edx
  call     CORINFO_HELP_LDELEMA_REF
  mov      bword ptr [rsp+28H], rax
  mov      r13d, dword ptr [r13+8]
  mov      rdx, qword ptr [r14+16]
  test     rdx, rdx
  jne      G_M28280_IG11
  mov      rcx, rbx
  mov      rdx, 0x7FFF485871E0
  call     CORINFO_HELP_RUNTIMEHANDLE_METHOD
  mov      rdx, rax
  mov      rax, bword ptr [rsp+28H]
  mov      dword ptr [rsp+20H], r13d
  lea      rcx, bword ptr [rsp+30H]
  mov      r8, rsi
  mov      r9, rax
  call     Span:DangerousCreate(...)
  test     bpl, bpl
; *** this branch always taken
  je       SHORT G_M28280_IG09
  mov      edx, dword ptr [rsp+38H]
  cmp      edx, 0
  jbe      SHORT G_M28280_IG12
  mov      rdx, bword ptr [rsp+30H]
&nbsp;
&nbsp;
&nbsp;
&nbsp;
&nbsp;
&nbsp;
&nbsp;
&nbsp;
  mov      rdx, gword ptr [rdx]
  lea      rcx, bword ptr [r15+8]
  call     CORINFO_HELP_ASSIGN_REF
  mov      rcx, qword ptr [r14+24]
  test     rcx, rcx
  jne      SHORT G_M28280_IG08
  mov      rcx, rbx
  mov      rdx, 0x7FFF485871F0
  call     CORINFO_HELP_RUNTIMEHANDLE_METHOD
  mov      rcx, rax
&nbsp;
G_M28280_IG08:
  call     CORINFO_HELP_NEWSFAST
  mov      rsi, rax
&nbsp;
G_M28280_IG09:
  inc      r12d
  cmp      r12d, edi
  jl       G_M28280_IG04                                
</td></pre></code></tr></table>
</details>

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
```diff
@@ -0,0 +1,925 @@
...
+                {
+                    result = WriteViaIndexer1(s);
+                }
+                return result;
+            },
+            "WriteViaIndexer1({0})", length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static byte WriteViaIndexer1(Span<byte> data)
```
JosephTremoulet: Calling this `TestWriteViaIndexer1` would be more consistent with the rest of the naming.

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
```diff
@@ -0,0 +1,925 @@
...
+                {
+                    result = WriteViaIndexer2(s);
+                }
+                return result;
+            },
+            "WriteViaIndexer2({0})", length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static byte WriteViaIndexer2(Span<byte> data)
```
JosephTremoulet: Calling this `TestWriteViaIndexer2` would be more consistent with the rest of the naming.  (the others besides WriteViaIndexer1/WriteViaIndexer2 look fine)

mikedn:
> The new version is shorter because of the indexer intrinsic.

I see, nice. Maybe the slowdown is caused by lack of jump target alignment, I assume that the code before the loop is identical so both loops are either aligned or not aligned.

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
AndyAyersMS: Yeah, I'll update this and the one below.

AndyAyersMS:
Slowdowns persist even with  `COMPlus_JitAlignLoops=1` -- though I will admit I haven't drilled into this to verify we get 16 byte method head alignment, that this option pads loop heads to 16 byte boundaries, and it uses the recommended padding sequences. Some comments say padding only goes to 8 bytes.

Given the branchiness at the top of the loop (where are you, cmov?) it could also be the branch predictor has some unfortunate interaction with branches in some other method -- many of those have different code in the after version.

russellhadley:
cmov #7447?  I don't know if we should just bring back qmark, or do the late recognition.  We should probably revisit this again.

mikedn:
> Where are you, cmov?

Somewhere [here](https://github.com/mikedn/coreclr/commit/f5857f8c45c0db45b0e3343c564e5ba89319d61a) :). But I don't think it will help with that particular code because it's an indir. My experimental implementation ignores them due to exceptions.

AndyAyersMS:
It comes from `impRuntimeLookup` so the defererence should be marked as non-excepting. But the dereference is conditional so probably beyond what we can hope for from cmov.

Even if we can't get rid of the branch it would be nice to get rid of that 8 byte immediate -1 somehow and branch over the indirection case to avoid back to back branches. We have
```
   y = (x & 1) == 0 ? x : *(x-1)
```
we might try instead something like:
```
   tmp = x & 1;
   y = (tmp != 0) ? *(x - tmp) : x;
```
or some other trick to get a cheap -1.

mikedn:
>  8 byte immediate -1 

8 byte? Where is that?

I'd rather fix the flow control, it's weird:
```asm
  test     r8d, 1
  jne      SHORT G_M28281_IG05
  jmp      SHORT G_M28281_IG06
G_M28281_IG05:
  mov      r8, qword ptr [r8+FFFFFFFFH] 
G_M28281_IG06:
```
that is
```asm
  test     r8d, 1
  je       SHORT G_M28281_IG06
  mov      r8, qword ptr [r8+FFFFFFFFH] 
G_M28281_IG06:
```

AndyAyersMS:
I edited out the encoded instruction bytes to save room. Here it is in all its glory:
```
G_M28280_IG05:
       4D8B80FFFFFFFF       mov      r8, qword ptr [r8+FFFFFFFFH]
```
The native compilers know lots of tricks to generate compact small immediates.

We should definitely fix the control flow....

mikedn:
> Here it is in all its glory:

Well, it's 4 bytes, not 8. And it probably should be 1 byte only since the displacement can be 8 bit in size and it is sign extended to 64 bit.

mikedn:
Yeah, `GenTreeAddrMode::gtOffset` is `unsigned` so the emitter sees `0xFFFFFFFF` as `0x00000000FFFFFFFF` instead of `0xFFFFFFFFFFFFFFFF`...

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
```diff
@@ -0,0 +1,925 @@
...
+        static byte TestSameIndex1(Span<byte> data)
+        {
+            byte x = 0;
+            byte y = 0;
+
+            for (var idx = 0; idx < data.Length; idx++)
+            {
+                x ^= data[idx];
+                y ^= data[idx];
+            }
```
JosephTremoulet: In `SameIndex1`, `SameIndex2`, and `CoveredIndex1`, I think you're writing the tests to ensure that we can remove a second bounds check based on a prior bounds check succeeding, in straight-line code.  But since you put the straight-line code in a loop and the indexer is the IV ranging from 0 to data.Length, once we handle the IV cases we'll be removing both range checks from the loop and no longer testing what you intended (if I'm following correctly).  Might want to make the loop bound an argument to the inner method or something.

It looks like maybe the same thing is happening in `WriteViaIndexer2`, but there you used 1 for the initial value of the IV, not sure if you were counting on that ruling out the IV-based removal or not.

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
```diff
@@ -0,0 +1,925 @@
...
+                {
+                    result = TestCoveredIndex1(s);
+                }
+                return result;
+            },
+            "CoveredIndex1({0})", length);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static byte TestCoveredIndex1(Span<byte> data)
```
JosephTremoulet: Our assertion prop currently isn't clever enough to get the cases here, even for arrays.  The covered-index cases we currently handle are:

  - a[K1] followed by a[K2], with K2 >= 0 and K1 >= K2  (with both K1 and K2 constant)
  - a[*] followed by a[0]  (where the "*" means "anything")

Might want to add testing for those (or I can with #10050).

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
AndyAyersMS: Good point, will update the loop bounds on those tests to be argument values.

On `WriteViaIndexer2` I could go back to 0 lower bounds. It started off as a clone of `WriteViaIndexer1`...

AndyAyersMS:
Umm, yeah, 8 nibbles, not 8 bytes. Those `F`s look bigger than they are.

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
AndyAyersMS: Will do.

AndyAyersMS:
Ok, tests updated. Hopefully this covers all the things that were brought up.

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
```diff
@@ -500,9 +500,9 @@ public static void WriteViaIndexer2(int length)
         [MethodImpl(MethodImplOptions.NoInlining)]
         static byte TestWriteViaIndexer2(Span<byte> data)
```
JosephTremoulet: This now definitely has the `idx = 0; idx < data.length; idx++` pattern that ought to let us remove the bounds checks from the loop regardless of their relation to one another; I'd suggest making start/length parameters to this like you did with the others.

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
```diff
@@ -754,6 +759,97 @@ static byte TestCoveredIndex1(Span<byte> data)
...
+        static byte TestCoveredIndex2(Span<byte> data, int start, int end)
+        {
+            byte x = 0;
+            byte y = 0;
+
+            for (var idx = start; idx < end; idx++)
+            {
+                x ^= data[idx];
+
+                if (idx == 100)
```
JosephTremoulet: May want to invert this condition so that the case whose optimization we're hoping to check is the one that usually executes.

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
```diff
@@ -754,6 +759,97 @@ static byte TestCoveredIndex1(Span<byte> data)
...
+        {
+            byte x = 0;
+            byte y = 0;
+            byte z = 0;
+
+            for (var idx = start; idx < end; idx++)
+            {
+                x ^= data[idx];
+
+                if (idx == 100)
```
JosephTremoulet: Same as `TestCoveredIndex2`, but additionally `x ^= data[idx]` isn't relevant to this pattern so you may want to remove it or make the `if` an `if/else` and sink `x ^= data[idx]` to the rare side.

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
AndyAyersMS: Will do.

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
AndyAyersMS: Good idea.

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
AndyAyersMS: Agree on the inversion.

Right now all the tests compute the same expected result, which we check.

The `x ^= ...` feeds into the return value so it has to be done every iteration, or else I need to fix my checking scheme to be more flexible.

AndyAyersMS:
The [Ubuntu arm cross leg](https://ci.dot.net/job/dotnet_coreclr/job/master/job/arm_cross_release_ubuntu_prtest/1155/) failed in ldelemnullarr2.exe. Looks like this has failed in other recent CI runs and may be flaky.

I have another update to push so will just let that kick off a retry.

AndyAyersMS:
Need one more test update. The test class must be public for xunit-perf to find the benchmarks.

Benchmark discovery is now working with some private xunit-perf bits, but perf data is still not showing up.

AndyAyersMS:
Same [ubuntu flakiness](https://ci.dot.net/job/dotnet_coreclr/job/master/job/arm_cross_release_ubuntu_prtest/1158/) in ldelemnullarr2.exe

@dotnet-bot retest Ubuntu arm Cross Release Build

AndyAyersMS:
This latest [Ubuntu arm cross leg](https://ci.dot.net/job/dotnet_coreclr/job/master/job/arm_cross_release_ubuntu_prtest/1177/) timed out.

Will give it one more shot...

@dotnet-bot retest Ubuntu arm Cross Release Build

AndyAyersMS:
Last 100 or more runs of the Ubuntu cross arm leg for any PR have now timed out or failed. 

@hseok-oh is this something you can look into?

tests/src/JIT/Performance/CodeQuality/Span/Indexer.cs
```diff
@@ -0,0 +1,1021 @@
...
+                    var invokeMethod = m.CreateDelegate(typeof(Action<int>)) as Action<int>;
+
+                    // All the benchmarks methods in this test use [InlineData] to specify how many times and with
+                    // what arguments they should be run.
+                    foreach (InlineDataAttribute dataAttr in m.GetCustomAttributes<InlineDataAttribute>())
+                    {
+                        foreach (object[] data in dataAttr.GetData(m))
+                        {
+                            // All the benchmark methods in this test take a single int parameter
+                            invokeMethod((int)data[0]);
```
JosephTremoulet: This calls a [Benchmark] method which in turn calls Invoke, and one of the arguments to Invoke is a string with the method name and argument. We have the MethodInfo and argument handy here, so we could compute the string here and communicate it to Invoke via a static field, to avoid the need to have that boilerplate string creation in each [Benchmark] method. I wouldn't stop to do that in this change, just wanted to mention the idea so that if we repeat this pattern again we could consider doing it then.

AndyAyersMS: I also considered adding a command line filtering setup so you could just run the part or parts of interest.

Once we've settled on a pattern I'd like to go back to some of the older benchmarks and update them to use it too.

hseok-oh:

PR #10911, opened 2017-04-12
[Local GC] Hide handle table internals from VM side

adityamandaleeka:
This change moves us closer to a clean separation between GC and VM when it comes to handles.

The VM side will no longer need to know about `HandleTableBucket`s and home heap numbers and all the details about how we choose to manage handles in the GC, and instead will just know about one type of handle store (which, internally, will be a `HandleTableBucket`*). It will be able to create these stores, add to them, destroy them, and so on. AppDomains will create their own handle stores and there will also be a global one.

I removed the parts of the interface that were using the actual HHANDLETABLE concept (like `GetHandleTableForHandle`) so the interface only exposes one type of "collection of handles".

@Maoni0 @jkotas @swgillespie @sergiy-k

### DIFF
--- a/src/gc/gchandletable.cpp
+++ b/src/gc/gchandletable.cpp
@@ -23,6 +23,16 @@ void GCHandleTable::Shutdown()
     Ref_Shutdown();
 }
 
+void* GCHandleTable::GetGlobalHandleTable()
+{
+    return (void*)g_HandleTableMap.pBuckets[0];
+}
+
+void* GCHandleTable::GetNewHandleTable(void* context)
+{
+    return (void*)::Ref_CreateHandleTableBucket(ADIndex((uintptr_t)context));
+}
+
 void* GCHandleTable::GetHandleContext(OBJECTHANDLE handle)
 {
     return (void*)((uintptr_t)::HndGetHandleTableADIndex(::HndGetHandleTable(handle)).m_dwIndex);
--- a/src/gc/gchandletable.cpp
+++ b/src/gc/gchandletable.cpp
@@ -38,9 +38,31 @@ void* GCHandleTable::GetHandleContext(OBJECTHANDLE handle)
     return (void*)((uintptr_t)::HndGetHandleTableADIndex(::HndGetHandleTable(handle)).m_dwIndex);
 }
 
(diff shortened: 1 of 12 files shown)


src/gc/gchandletable.cpp
```diff
@@ -23,39 +23,73 @@ void GCHandleTable::Shutdown()
...
 
-void* GCHandleTable::GetHandleTableContext(void* handleTable)
+void* GCHandleTable::GetGlobalHandleStore()
 {
-    return (void*)((uintptr_t)::HndGetHandleTableADIndex((HHANDLETABLE)handleTable).m_dwIndex);
+    return (void*)g_HandleTableMap.pBuckets[0];
 }
 
-void* GCHandleTable::GetHandleTableForHandle(OBJECTHANDLE handle)
+void* GCHandleTable::GetNewHandleStore(void* context)
```
jkotas: CreateHandleStore ?

adityamandaleeka: Sure, will change it.

jkotas:
There are way too many `void*` now ... it would be really nice to create a strong typed HANDLESTORE or something.

Looks good to me otherwise. Thanks!

danmosemsft:
@dotnet-bot test Tizen armel Cross Debug Build  (biuld break)
@dotnet-bot test Tizen armel Cross Release Build  (biuld break)

adityamandaleeka:
@jkotas Yea, I think I will make an `IGCHandleStore` which will have `Destroy`, `Uproot`, `ContainsHandle`, and all the handle creation functions that take `void* store`. The implementation of that interface will just hold on to the `HandleTableBucket*`. This will get rid of a bunch of void*s and also cut down on the casts we're doing. 

We should also rename `IGCHandleTable` to something else since it really doesn't correspond to a handle table.

I'll do these things in a later PR once I have a good design.

sergiy-k:
I agree with @jkotas regarding ```void*``` and extra casting caused by it. I'm OK with addressing it in a separate PR.

adityamandaleeka:
Thanks @jkotas and @sergiy-k. I'll make those changes in another PR soon (maybe today).

PR #10916, opened 2017-04-12
[Linux][GDB-JIT] Fix class size calculation

Buyduck:
We should use "m_BaseSize" filed of "MethodTable" to get correct class size.

@janvorli PTAL

cc @Dmitri-Botcharnikov @lucenticus @ayuckhulk


lucenticus:
LGTM

Buyduck:
@janvorli Thank you!

Issue #9968, opened 2017-03-05
None of the build badges are working


stephentoub:
https://github.com/dotnet/coreclr/blob/master/Documentation/workflow/OfficalAndDailyBuilds.md

They all point to an old server, hence they all show up as "inaccessible".

danmosemsft:
@stephentoub seems fixed..

stephentoub:
@danmosemsft, the problems I called it here https://github.com/dotnet/coreclr/pull/10021#issuecomment-285029233 are still problems.

gkhanna79:
Assigning to @stephentoub who has a PR out for this.

PR #10917, opened 2017-04-12
Fix build badges in OfficialAndDailyBuilds.md

stephentoub:
- None of the badge links were correct, resulting in every badge always showing as "Building" regardless of actual status
- The OSX badge links would 404.
- We were still using shields.io badges, which have been unreliable.

Fixes https://github.com/dotnet/coreclr/issues/9968
cc: @danmosemsft


stephentoub:
@mmitche, this fix highlights that these jobs have never been run:
http://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_centos7.1
http://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_debian8.4
http://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_rhel7.2
but I assume they should be run as part of merges to master or daily or something?

gkhanna79:
The change looks good to me though CentOS 7.1, Debian 8.4 and RH 7.2 are still shown as unknown. Is that expected @stephentoub ?

stephentoub:
>  though CentOS 7.1, Debian 8.4 and RH 7.2 are still shown as unknown

See my comment to @mmitche just above yours.  They show as unknown because they've never been run; that's not a problem with the badges but, assuming we actually care about them, which netci.groovy or whatever we use to trigger builds.

PR #10921, opened 2017-04-12
Target Portability Fix in WaitHandle

Rew:
Reorganize conditional compiler directives to have the "else" condition
be the fallback case (non Windows targets).  Originally it was coded
that the "if" was the fallback case (only unix) and the else was an
implied specific case (that only worked on windows targets).


Rew:
After thinking about this a bit, I am unsure if this change makes sense.  Without knowing the challenge to this being implemented in Unix, it is unclear if this should be a default feature that is not yet supported by Unix targets, or an extended feature that is only available on windows targets.  The InternalCall on SignalAndWaitOne seems to imply that it is intended to be implemented by all runtimes going forward, and should be included by default.  I'm leaning towards closing this.

PR #10924, opened 2017-04-12
Comment explaining use of FlushProcessWriteBuffers in updating card table

acl33:
Hope this is right, and correctly reflects the various discussions we've had...

### DIFF
--- a/src/vm/gcenv.ee.cpp
+++ b/src/vm/gcenv.ee.cpp
@@ -1275,9 +1275,17 @@ void GCToEEInterface::StompWriteBarrier(WriteBarrierParameters* args)
         // We need to make sure that other threads executing checked write barriers
         // will see the g_card_table update before g_lowest/highest_address updates.
         // Otherwise, the checked write barrier may AV accessing the old card table
-        // with address that it does not cover. Write barriers access card table
-        // without memory barriers for performance reasons, so we need to flush
-        // the store buffers here.
+        // with address that it does not cover. 
+        //
+        // Even x86's total store ordering is insufficient here because threads reading
+        // g_card_table do so via the instruction cache, whereas g_lowest/highest_address
+        // are read via the data cache.
+        //
+        // The g_card_table update is covered by section 8.1.3 of the Intel Software
+        // Development Manual, Volume 3A (System Programming Guide, Part 1), about
+        // "cross-modifying code": We need all _executing_ threads to invalidate
+        // their instruction cache, which FlushProcessWriteBuffers achieves by sending
+        // an IPI (inter-process interrupt).
         FlushProcessWriteBuffers();
 
         g_lowest_address = args->lowest_address;


acl33:
Note I've not mentioned Section 11.6 on self-modifying code (where an instruction is modified by a different linear address to which it is loaded from - i.e. we *might* need to flush even were the update not cross-thread, because the address written to (2 bytes after the start of the instruction) might not be in the same cache line as the start of the instruction....should I add that as well?

@mjp41 is this about what you were thinking?

src/gc/gc.cpp
```diff
@@ -7389,6 +7389,13 @@ int gc_heap::grow_brick_card_tables (uint8_t* start,
 
         seg_mapping_table = new_seg_mapping_table;
 
+        // The above write of the card table base address, is read by executing threads via the instruction cache.
+        // The update counts as cross-modifying code (section 8.1.3 of the Intel Software Development Manual),
+        // which requires all _executing_ threads to invaliadate their icache. FlushProcessWriteBuffers achieves
```
adityamandaleeka: Typo: 'invalidate'

src/gc/gc.cpp
```diff
@@ -7389,6 +7389,13 @@ int gc_heap::grow_brick_card_tables (uint8_t* start,
 
         seg_mapping_table = new_seg_mapping_table;
 
+        // The above write of the card table base address, is read by executing threads via the instruction cache.
```
adityamandaleeka: Nit: remove the comma after address

adityamandaleeka:
LGTM modulo those minor comments.

mjp41:
@acl33 should you add a Volume number?  It took as a while to find the right one?

Otherwise LGTM.

mjp41:
Actually, I think the following comments should also be updated:
https://github.com/dotnet/coreclr/blob/2dfc74df1f9fe238eb2ec25c4764b8b65432b010/src/vm/gcenv.ee.cpp#L1281

src/gc/gc.cpp
jkotas: I do not think that there is instruction cache involved here. `g_gc_card_table` is regular static, it won't be read via the instruction cache.

The comment about instruction cache should be next to https://github.com/dotnet/coreclr/blob/2dfc74df1f9fe238eb2ec25c4764b8b65432b010/src/vm/gcenv.ee.cpp#L1281 as @mjp41 pointed out. It is where we are patching the code inline.

acl33:
Thanks, all - you are quite right, I got the wrong one, it's StompWriteBarrier in gcenv.ee.cpp that I meant - it even has the comment I meant to update. So, have moved there.

This leaves me unable to explain why we need the FPWB I originally commented in gc.cpp - at least on x86. If there's anything I can write then please shout ;-)...

acl33:
@jkotas, @mjp41
Do you think the comment was also correct in its previous location, should I retain that? (This has changed totally since @adityamandaleeka 's first LGTM!!)

jkotas:
> Do you think the comment was also correct in its previous location, should I retain that?

I do not think that this comment is not applicable in the original location.

Issue #10900, opened 2017-04-11
Change SOS publish to use portable RIDs rather than distro specific RIDs.


lt72:
We have recently enabled building the product (nuget packages, binaries as applicable and installers) for portable RIDs (e.g. win-x64, osx-64) as opposed to distro specific RIDs (e.g. win7-x64). For this, the packages are also produced with the same RID. However, I notice (https://github.com/dotnet/coreclr/blob/master/src/ToolBox/SOS/NETCore/project.json#L9) that SOS.NETCore is published for netcoreapp1.0 and for specific RIDs. 
 
For SOS.NEtCore to be correctly included in portable build, it should target portable RIDs which will also require it to move to netcoreapp2.0. 

The required change is to publish SOS for netcoreapp2.0 usign portable RIDs.

PR #10925, opened 2017-04-12
Remove distro specific RIDs from SOS.NETCore's project.json.

mikem8361:
Removed the "runtimes" section from the project.json.

Issue: https://github.com/dotnet/coreclr/issues/10900

### DIFF
--- a/src/ToolBox/SOS/NETCore/project.json
+++ b/src/ToolBox/SOS/NETCore/project.json
@@ -11,14 +11,5 @@
        "portable-net45+win8" 
       ] 
     } 
-  },
-  "runtimes": {
-    "win7-x86": {},
-    "win7-x64": {},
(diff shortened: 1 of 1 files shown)


src/ToolBox/SOS/NETCore/project.json
```diff
@@ -11,14 +11,5 @@
        "portable-net45+win8" 
       ] 
     } 
-  },
```
gkhanna79: Does this produce the expected artifacts for both portable and non-portable builds (and do they work)?

Do "build.cmd -portable" for triggering a portable build.

mikem8361:
The “-portable” builds and works on Windows and Linux.

gkhanna79:
Please feel free to merge if you think the change is correct. Sorry, I dont have enough context on the build of this component to review it.

PR #10927, opened 2017-04-12
Enable building Core_Root for any RID against packages

wtgodbe:
This change does the following things:

- Requires an RID to be passed when building tests against packages
- Builds a Core_Root only for that RID
- Removes the guarantee that you'll always be able to run tests locally after building tests against packages (i.e. if you build tests for Ubuntu, you won't also get a Core_Root useable on Windows).
- Enables building tests/Core_Root with portable windows & OSX binaries
- Changes pipeline build to run Windows & OSX tests using portable binaries

CC @gkhanna79 this will get all Helix tests running against portable binaries

@MattGal do you know when Helix jobs will be working again? I can't fully validate this until I can run Helix jobs with bits from this change, which I haven't been able to do yet.


wtgodbe:
I have now been able to validate that tests still pass in Helix with these changes - @gkhanna79 @MattGal can you guys PTAL before I merge?

build-test.cmd
```diff
@@ -305,14 +299,34 @@ if exist "%CORE_ROOT%" rd /s /q "%CORE_ROOT%"
...
 
 set RuntimeIdArg=
+set TargetsWindowsArg=
 
 if defined __RuntimeId (
     set RuntimeIdArg=-RuntimeID="%__RuntimeId%"
 )
 
+if "%__TargetsWindows%"=="1" (
+    set TargetsWindowsArg=-TargetsWindows=true
```
gkhanna79: Why don't you add these to "unprocessedbuildargs" list?

gkhanna79: Never mind - I see you need to use it at other places too.

PR #10929, opened 2017-04-12
[Local GC] Handle table low-hanging interface violations

swgillespie:
Unfortunately this was the only low-hanging fruit I could find in the handle table. The other interface violations look to be more complex to deal with, in particular:
  1. Async pinned handles migrating on app domain unload (and in general, the handle table knowing about `System.Threading.OverlappedData`)
  2. Handle creation will throw on failure instead of returning null.
  3. Some handle table code uses PAL exception handling
  4. Validation routines that reference both EE and GC details (e.g. https://github.com/dotnet/coreclr/blob/master/src/gc/handletablescan.cpp#L964)

cc @adityamandaleeka @sergiy-k @Maoni0 @jkotas

### DIFF
--- a/src/gc/handletablescan.cpp
+++ b/src/gc/handletablescan.cpp
@@ -949,7 +949,7 @@ static void VerifyObjectAndAge(_UNCHECKED_OBJECTREF *pValue, _UNCHECKED_OBJECTRE
     if (minAge >= GEN_MAX_AGE || (minAge > thisAge && thisAge < static_cast<int>(g_theGCHeap->GetMaxGeneration())))
     {
         _ASSERTE(!"Fatal Error in HandleTable.");
-        EEPOLICY_HANDLE_FATAL_ERROR(COR_E_EXECUTIONENGINE);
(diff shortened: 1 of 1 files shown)


jkotas:
LGTM 

> Async pinned handles migrating on app domain unload

This will be a set of not very pretty callbacks, but it should be manageable.

> Handle creation will throw on failure instead of returning null.

Yes, these should return NULL instead.

> Some handle table code uses PAL exception handling

This is part of the async pinned handle problem.

> Validation routines that reference both EE and GC details

This is not really specific to handle table. The object validation callback is needed by gc.cpp as well.

adityamandaleeka:
LGTM

swgillespie:
Thanks both!

@adityamandaleeka Do you think I'd hit any merge conflicts with your work if I tried to address any of the above issues?

adityamandaleeka:
@swgillespie Don't worry about it. I will hold off on touching the async pinned handle code if you are going to work on it.

swgillespie:
@adityamandaleeka Sounds good then, thanks!

PR #10930, opened 2017-04-12
HelperMethodFrame missing at certain places in method MarshalNative::PtrToStructureHelper

rahku:
AV can occur at https://github.com/dotnet/coreclr/blob/master/src/vm/marshalnative.cpp#L202 if invalid address is obtained from native side. When that happens since no HMF is erected x86 stackwalker will miss walking certain portions of managed code. Fix is to create HMF at start of method.


rahku:
PTAL @jkotas @gkhanna79

jkotas:
Shouldn't we be rather fail fasting in this case?

LGTM otherwise (it is consistent with the other methods near by).

rahku:
Yes unless someone specifies LegacyCorruptedStateExceptionsPolicy

PR #10383, opened 2017-03-22
[x86/Linux] fix broken stack in DelayLoad_Helper when converting to AT&T syntax assembly.

ragmani:
```
Dump of assembler code for function DelayLoad_Helper:
   0xf713123b <+0>:	push   %esi
   0xf713123c <+1>:	push   %edi
   0xf713123d <+2>:	push   %ecx
   0xf713123e <+3>:	push   %edx
   0xf713123f <+4>:	mov    0x10(%esp),%ecx
   0xf7131243 <+8>:	mov    0x14(%esp),%edx
   0xf7131247 <+12>:	mov    %ebx,0x10(%esp)
   0xf713124b <+16>:	mov    %ebp,0x14(%esp)
   0xf713124f <+20>:	lea    0x14(%esp),%ebp
   0xf7131253 <+24>:	mov    %esp,%esi
   0xf7131255 <+26>:	pushw  $0x0
   0xf7131258 <+29>:	push   %ecx
   0xf7131259 <+30>:	push   %edx
   0xf713125a <+31>:	push   %eax
   0xf713125b <+32>:	push   %esi
   0xf713125c <+33>:	call   0xf729dcf0 <DynamicHelperWorker(TransitionBlock*, TADDR*, DWORD, Module*, INT)>
   0xf7131261 <+38>:	test   %eax,%eax
   0xf7131263 <+40>:	jne    0xf713126f <DelayLoad_Helper+52>
   0xf7131265 <+42>:	mov    (%esi),%eax
   0xf7131267 <+44>:	add    $0x8,%esp
   0xf713126a <+47>:	pop    %edi
   0xf713126b <+48>:	pop    %esi
   0xf713126c <+49>:	pop    %ebx
   0xf713126d <+50>:	pop    %ebp
=> 0xf713126e <+51>:	ret    
   0xf713126f <+52>:	pop    %edx
   0xf7131270 <+53>:	pop    %ecx
   0xf7131271 <+54>:	pop    %edi
   0xf7131272 <+55>:	pop    %esi
   0xf7131273 <+56>:	pop    %ebx
   0xf7131274 <+57>:	pop    %ebp
   0xf7131275 <+58>:	jmp    *%eax
```

src/vm/i386/asmhelpers.S
```
.macro DYNAMICHELPER frameFlags, suffix

NESTED_ENTRY DelayLoad_Helper\suffix, _TEXT, NoHandler
    STUB_PROLOG_2_HIDDEN_ARGS

    mov         esi, esp

[...]

PR #10932, opened 2017-04-13
[x86/Linux] changed to the AT&T syntax to solve a problem related to broken stack.

ragmani:
This is a problem similar to #10383.


parjong:
@janvorli Please take a look.

PR #10933, opened 2017-04-13
Fix clang build error when enabling NODEBASH_STATS

hqueue:
After enabling NODEBASH_STATS in Linux, following error occurs when building CoreCLR with clang.

```
/opt/code/github/hqueue/coreclr/src/jit/gentree.cpp:453:8: error: C++ requires a type specifier for all declarations
inline hashme(genTreeOps op1, genTreeOps op2)
~~~~~~ ^
```


hqueue:

PR #10935, opened 2017-04-13
[x86/Linux] Use appropriate FCALL macro for SafeBuffer methods

parjong:
SafeBuffer::PtrToStructure and SafeBuffer::StructureToPtr takes byref arguments, but are not annotated by appropriate macros(such as FCXXXX3_IVI / FCXXXX3_VII). 

This results in argument order mismatch called from JIT, which results in assert failures while running the following FX unittests:
 - System.Security.Cryptography.X509Certificates.Tests
 - System.IO.UnmanagedMemoryStream.Tests
 - System.Security.SecureString.Tests
 - System.Runtime.InteropServices.Tests
 - System.Net.Primitives.Functional.Tests

This commit uses appropriate macros for these methods in order to fix assert failures.


jkotas:
Thanks!

PR #10936, opened 2017-04-13
Introduce API for codegen getting temp registers from gtRsvdRegs

BruceForstall:
LSRA puts a set of temporary registers needed by a node in the
gtRsvdRegs register mask. Currently, evey codegen function that
needs to use a temporary register manually manipulates this mask.

Introduce a few functions to make this simpler and more regular:

1. GetSingleTempReg() // Gets a temp; asserts there is exactly one temp reg.
2. ExtractTempReg() // Gets the lowest temp, removes it from gtRsvdRegs so subsequent calls don't see it.
3. AvailableTempRegs() // Returns a count of available temp registers.

All take an optional register mask, so you can extract a register from
just the set specified by the mask (typically RBM_ALLINT or RBM_ALLFLOAT).

### DIFF
--- a/src/jit/gentree.cpp
+++ b/src/jit/gentree.cpp
@@ -17403,3 +17403,68 @@ regMaskTP ReturnTypeDesc::GetABIReturnRegs()
 
     return resultMask;
 }
+
+#ifndef LEGACY_BACKEND
+
+//------------------------------------------------------------------------
+// The following functions manage the gtRsvdRegs set of temporary registers
+// created by LSRA during code generation.
+
+//------------------------------------------------------------------------
+// AvailableTempRegCount: return the number of available temporary registers in the (optional) given set
+// (typically, RBM_ALLINT or RBM_ALLFLOAT).
+//
+// Arguments:
+//    mask - (optional) Check for available temporary registers only in this set.
+//
+// Return Value:
+//    Count of available temporary registers in given set.
+//
+unsigned GenTree::AvailableTempRegCount(regMaskTP mask /* = (regMaskTP)-1 */) const
+{
+    return genCountBits(gtRsvdRegs & mask);
+}
+
+//------------------------------------------------------------------------
+// GetSingleTempReg: There is expected to be exactly one available temporary register
+// in the given mask in the gtRsvdRegs set. Get that register. No future calls to get
+// a temporary register are expected. Removes the register from the set, but only in
+// DEBUG to avoid doing unnecessary work in non-DEBUG builds.
+//
+// Arguments:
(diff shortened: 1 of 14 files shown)


BruceForstall:

hqueue:
Thanks! It was quite inconvenient to manipulate `gtRsvdRegs`.

mikedn:
It's not clear why these new functions are in `GenTree` instead of `CodeGen`, considering that they're primarily used by `CodeGen`.

BruceForstall:
@mikedn I suppose putting them in CodeGen works. Since they are solely to manipulate GenTree::gtRsvdRegs, I thought putting them on GenTree more tightly associated them with their data.

src/jit/gentree.cpp
```diff
@@ -17403,3 +17403,68 @@ regMaskTP ReturnTypeDesc::GetABIReturnRegs()
...
+// AvailableTempRegs: return the number of available temporary registers in the (optional) given set
+// (typically, RBM_ALLINT or RBM_ALLFLOAT).
+//
+// Arguments:
+//    mask - (optional) Check for available temporary registers only in this set.
+//
+// Return Value:
+//    Count of available temporary registers in given set.
+//
+unsigned GenTree::AvailableTempRegs(regMaskTP mask /* = (regMaskTP)-1 */)
```
pgavlin: Nit: I would name this `AvailableTempRegCount`.

src/jit/gentree.cpp
```diff
@@ -17403,3 +17403,68 @@ regMaskTP ReturnTypeDesc::GetABIReturnRegs()
...
+// and return it. Remove this register from the temporary register set, so it won't
+// be returned again.
+//
+// Arguments:
+//    mask - (optional) Extract an available temporary register only in this set.
+//
+// Return Value:
+//    Available temporary register in given mask.
+//
+regNumber GenTree::ExtractTempReg(regMaskTP mask /* = (regMaskTP)-1 */)
```
pgavlin: I don't love the fact that this API is stateful, but I'm having trouble articulating exactly why.

I would be more comfortable encapsulating the temp registers in a type that provides the APIs you've added. Perhaps something like:
```
class TempRegSet
{
public:
    unsigned Count();

    template<bool expectSingleRegister = false>
    regNumber ExtractNextRegister();
};
```

The `GenTree` API would then become something like:
```
TempRegSet GenTree::GetReservedRegisters(regMaskTP mask = -1);
```

BruceForstall:

BruceForstall:

BruceForstall:
@pgavlin ping

Issue #1986, opened 2015-11-07
Set affinity for server GC threads


jkotas:
Body of `set_thread_affinity_mask_for_heap` is ifdefed out for CoreCLR. @Maoni0 thinks that it is important to fix this.

benaadams:
From botr [Garbage Collection Design](https://github.com/dotnet/coreclr/blob/master/Documentation/botr/garbage-collection.md)

> **Efficient locking:** The thread affinity of allocation contexts and quantums guarantee that there is only ever a single thread writing to a given allocation quantum. As a result, there is no need to lock for object allocations, as long as the current allocation context is not exhausted.

Could be a bug; rather than just perf?

jkotas:
This issue is about thread affinity of server GC threads that affects performance only.

The botr paragraph you have quoted talks about thread affinity of allocation quantums that is unrelated concept.

benaadams:
Ah, makes sense then

benaadams:
Resolved by https://github.com/dotnet/coreclr/pull/4770 ?

janvorli:
@benaadams it was fixed for Windows only, we still need to fix it for Unix.

gkhanna79:
@janvorli Is this still planned for fixing in 2.0?

janvorli:
Yes, I am currently finishing it. I have most of the stuff done.

janvorli:
Implemented by #10938

PR #10938, opened 2017-04-13
Add NUMA and thread affinity support for Unix

janvorli:
This change adds new PAL functions for NUMA and thread affinity support
for Unix and also enables related code in GC and VM for FEATURE_PAL.

It doesn't reflect the limits imposed by CGROUPS on systems with
CGROUPS enables yet.

### DIFF
--- a/src/pal/src/numa/numa.cpp
+++ b/src/pal/src/numa/numa.cpp
@@ -77,7 +77,7 @@ static const WORD NO_GROUP = 0xffff;
 
 /*++
 Function:
-  FreeLookupArrays
+  AllocateLookupArrays
 
 Allocate CPU and group lookup arrays
 --*/
@@ -151,6 +151,9 @@ NUMASupportInitialize()
         for (int i = 0; i < numaNodesCount; i++)
         {
             int st = numa_node_to_cpus(i, mask);
+            // The only failure that can happen is that the mask is not large enough
+            // but that cannot happen since the mask was allocated by numa_allocate_cpumask
+            _ASSERTE(st == 0);
             unsigned int nodeCpuCount = numa_bitmask_weight(mask);
             unsigned int nodeGroupCount = (nodeCpuCount + MaxCpusPerGroup - 1) / MaxCpusPerGroup;
             g_groupCount += nodeGroupCount;
@@ -164,6 +167,9 @@ NUMASupportInitialize()
         for (int i = 0; i < numaNodesCount; i++)
         {
             int st = numa_node_to_cpus(i, mask);
+            // The only failure that can happen is that the mask is not large enough
+            // but that cannot happen since the mask was allocated by numa_allocate_cpumask
+            _ASSERTE(st == 0);
             unsigned int nodeCpuCount = numa_bitmask_weight(mask);
             unsigned int nodeGroupCount = (nodeCpuCount + MaxCpusPerGroup - 1) / MaxCpusPerGroup;
             for (int j = 0; j < g_cpuCount; j++)
@@ -172,6 +178,8 @@ NUMASupportInitialize()
                 {
                     if (currentGroupCpus == MaxCpusPerGroup)
                     {
+                        g_groupToCpuCount[currentGroup] = MaxCpusPerGroup;
+                        g_groupToCpuMask[currentGroup] = GetFullAffinityMask(MaxCpusPerGroup);
                         currentGroupCpus = 0;
                         currentGroup++;
                     }
@@ -579,6 +587,7 @@ GetProcessAffinityMask(
                         // The process has affinity in more than one group, in such case
                         // the function needs to return zero in both masks.
                         processMask = 0;
+                        systemMask = 0;
                         group = NO_GROUP;
                         break;
                     }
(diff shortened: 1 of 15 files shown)


janvorli:
Fixes #1986

jkotas:
How are we going to test this?

src/pal/src/numa/numa.cpp
```diff
@@ -0,0 +1,683 @@
...
+)
+{
+    PERF_ENTRY(GetNumaHighestNodeNumber);
+    ENTRY("GetNumaHighestNodeNumber(HighestNodeNumber=%p)\n", HighestNodeNumber);
+    *HighestNodeNumber = (ULONG)g_highestNumaNode;
+
+    BOOL success = TRUE;
+
+    LOGEXIT("GetNumaProcessorNodeEx returns BOOL %d\n", success);
+    PERF_EXIT(GetNumaProcessorNodeEx);
```
swgillespie: nit: is this supposed to be GetNumaHighestNodeNumber?

janvorli:
@dotnet-bot test tizen armel cross debug build

janvorli:
@jkotas, as for the individual functions, I have tested them locally on my native Linux box that has 8 cores in two NUMA nodes. Regarding different NUMA configurations testing, some can be done on a VM. As for performance testing, I don't have a clear plan yet, ideally, we could use some real world benchmark like the techempower running on a physical machine with more processors / NUMA nodes and measure performance with the NUMA stuff enabled and disabled.

src/pal/src/numa/numa.cpp
janvorli: Thank you for spotting that, I have fixed it.

janvorli:

janvorli:

janvorli:

janvorli:

hseok-oh:

janvorli:
@swgillespie, @adityamandaleeka could you please review?

src/pal/src/numa/numa.cpp
```diff
@@ -0,0 +1,683 @@
...
+int g_groupCount = 0;
+// The highest NUMA node available
+int g_highestNumaNode = 0;
+
+static const int MaxCpusPerGroup = 8 * sizeof(KAFFINITY);
+static const WORD NO_GROUP = 0xffff;
+
+/*++
+Function:
+  FreeLookupArrays
```
adityamandaleeka: AllocateLookupArrays

src/pal/src/numa/numa.cpp
```diff
@@ -0,0 +1,683 @@
...
+    {
+        struct bitmask *mask = numa_allocate_cpumask();
+        int numaNodesCount = numa_max_node() + 1;
+
+        g_cpuCount = numa_num_possible_cpus();
+        g_groupCount = 0;
+
+        for (int i = 0; i < numaNodesCount; i++)
+        {
+            int st = numa_node_to_cpus(i, mask);
```
adityamandaleeka: Looks like we're not doing anything with this return value. Assert that it's 0?

adityamandaleeka: Same thing below

src/pal/src/numa/numa.cpp
```diff
@@ -0,0 +1,683 @@
...
+                    g_cpuToAffinity[j].Group = currentGroup;
+                    g_cpuToAffinity[j].Number = currentGroupCpus;
+                    g_groupAndIndexToCpu[currentGroup * MaxCpusPerGroup + currentGroupCpus] = j;
+                    currentGroupCpus++;
+                }
+            }
+
+            if (currentGroupCpus != 0)
+            {
+                g_groupToCpuCount[currentGroup] = currentGroupCpus;
```
adityamandaleeka: `g_groupToCpuCount[currentGroup]` is set here for whatever `currentGroup` is when we get here. What if we filled a group and moved on to the next one in the for loop above (near line 175)?

src/pal/src/numa/numa.cpp
```diff
@@ -0,0 +1,683 @@
...
+                    WORD g = g_cpuToAffinity[i].Group;
+                    if (group == NO_GROUP || g == group)
+                    {
+                        group = g;
+                        processMask |= ((DWORD_PTR)1) << g_cpuToAffinity[i].Number;
+                    }
+                    else
+                    {
+                        // The process has affinity in more than one group, in such case
+                        // the function needs to return zero in both masks.
```
adityamandaleeka: The comment here says we need to return 0 in both masks, but unless I'm missing something we're not setting `systemMask` to 0, right?

src/pal/src/numa/numa.cpp
janvorli: Sigh. Thank you for spotting that.

src/pal/src/numa/numa.cpp
janvorli: Adding assert, the only failure of this function could be that the mask is not large enough, but since the mask is allocated by the numa_allocate_cpumask function, that cannot happen.

src/pal/src/numa/numa.cpp
janvorli: Yes, we do. It is missing. Thank you for spotting it.

src/pal/src/numa/numa.cpp
janvorli: Right. Thank you!

janvorli:
@adityamandaleeka I've reflected your feedback.

PR #10941, opened 2017-04-13
Fix FieldDesc::GetExactFieldType when FieldDesc doesn't exactly represent the owner type

ayuckhulk:
This PR fix FieldDesc::GetExactFieldType when FieldDesc doesn't exactly represent the owner type (like in generics).

To get the exact type from FieldDesc, field signature should point past the calling convention.

@janvorli, @jkotas PTAL

### DIFF
--- a/src/vm/field.cpp
+++ b/src/vm/field.cpp
@@ -888,6 +888,10 @@ TypeHandle FieldDesc::GetExactFieldType(TypeHandle owner)
         GetSig(&pSig, &cSig);
         SigPointer sig(pSig, cSig);
 
+        ULONG callConv;
+        IfFailThrow(sig.GetCallingConv(&callConv));
+        _ASSERTE(callConv == IMAGE_CEE_CS_CALLCONV_FIELD);
+
         // Get the generics information
         SigTypeContext sigTypeContext(GetExactClassInstantiation(owner), Instantiation());
 
--- a/src/vm/field.cpp
+++ b/src/vm/field.cpp
@@ -895,9 +895,8 @@ TypeHandle FieldDesc::GetExactFieldType(TypeHandle owner)
         // Get the generics information
         SigTypeContext sigTypeContext(GetExactClassInstantiation(owner), Instantiation());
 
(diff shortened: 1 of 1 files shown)


src/vm/field.cpp
```diff
@@ -886,6 +886,13 @@ TypeHandle FieldDesc::GetExactFieldType(TypeHandle owner)
         PCCOR_SIGNATURE pSig;
         DWORD cSig;
         GetSig(&pSig, &cSig);
+
+        // Point past the calling convention
+        BYTE * pOldPtr = (BYTE*) pSig;
+        unsigned callConv = CorSigUncompressData(pSig);
+        _ASSERTE(callConv == IMAGE_CEE_CS_CALLCONV_FIELD);
```
jkotas: It would be better to call GetCallingConv on the SigPointer, e.g. like what is done here: https://github.com/dotnet/coreclr/blob/7ced6b540e5c2fd7ed49661dc0b182d65a413949/src/vm/siginfo.cpp#L2925

ayuckhulk: Thank you for the suggestion!

ayuckhulk:
@jkotas Is it possible that a field signature does not have calling conv info?

jkotas:
I think it should have it always.

I was kind of surprised when I saw this PR - I would not have expected to have a bug here. There may be something more going on.

ayuckhulk:

hseok-oh:

ayuckhulk:
@jkotas, so there were two issues with FieldDesc::GetExactFieldType:
* It always reported `System.Int16` (cortype index `0x6` or `IMAGE_CEE_CS_CALLCONV_FIELD`) when FieldDesc doesn't exactly represent the owner type
* The wrong module was used to load the type, but it was ok for `System.Int16`

ayuckhulk:
@jkotas Is this PR ready to be merged?

jkotas:
Thanks!

ayuckhulk:
@jkotas Thank you!

chunseoklee:
@ayuckhulk could you make PR into relase/2.0.0 branch ?

PR #10948, opened 2017-04-13
Fix perf of DateTime.Now on Unix

eerhardt:
Profiling DateTime.Now on Unix shows that a lot of time is spent in GetPreviousAdjustmentRule and AdjustmentRule::Equals. Instead, it is safe to use ReferenceEquals, which increases the performance of DateTime.Now and other time zone operations on Unix.

Partial fix for https://github.com/dotnet/corefx/issues/3571

See investigation https://github.com/dotnet/corefx/issues/3571#issuecomment-293762131

@tarekgh @stephentoub @mazong1123

### DIFF
--- a/src/mscorlib/src/System/TimeZoneInfo.cs
+++ b/src/mscorlib/src/System/TimeZoneInfo.cs
@@ -172,10 +172,11 @@ namespace System
             DateTime adjustedTime = ConvertTime(dateTimeOffset, this).DateTime;
 
             bool isAmbiguous = false;
-            AdjustmentRule rule = GetAdjustmentRuleForAmbiguousOffsets(adjustedTime);
+            int? ruleIndex;
+            AdjustmentRule rule = GetAdjustmentRuleForAmbiguousOffsets(adjustedTime, out ruleIndex);
             if (rule != null && rule.HasDaylightSaving)
             {
-                DaylightTimeStruct daylightTime = GetDaylightTime(adjustedTime.Year, rule);
+                DaylightTimeStruct daylightTime = GetDaylightTime(adjustedTime.Year, rule, ruleIndex);
                 isAmbiguous = GetIsAmbiguousTime(adjustedTime, rule, daylightTime);
             }
 
@@ -232,10 +233,11 @@ namespace System
             }
 
             bool isAmbiguous = false;
-            AdjustmentRule rule = GetAdjustmentRuleForAmbiguousOffsets(adjustedTime);
+            int? ruleIndex;
+            AdjustmentRule rule = GetAdjustmentRuleForAmbiguousOffsets(adjustedTime, out ruleIndex);
             if (rule != null && rule.HasDaylightSaving)
             {
-                DaylightTimeStruct daylightTime = GetDaylightTime(adjustedTime.Year, rule);
+                DaylightTimeStruct daylightTime = GetDaylightTime(adjustedTime.Year, rule, ruleIndex);
                 isAmbiguous = GetIsAmbiguousTime(adjustedTime, rule, daylightTime);
             }
 
@@ -263,15 +265,15 @@ namespace System
         }
 
         // note the time is already adjusted
-        private AdjustmentRule GetAdjustmentRuleForAmbiguousOffsets(DateTime adjustedTime)
+        private AdjustmentRule GetAdjustmentRuleForAmbiguousOffsets(DateTime adjustedTime, out int? ruleIndex)
         {
-            AdjustmentRule rule = GetAdjustmentRuleForTime(adjustedTime);
+            AdjustmentRule rule = GetAdjustmentRuleForTime(adjustedTime, out ruleIndex);
             if (rule != null && rule.NoDaylightTransitions && !rule.HasDaylightSaving)
             {
                 // When using NoDaylightTransitions rules, each rule is only for one offset.
                 // When looking for the Daylight savings rules, and we found the non-DST rule,
                 // then we get the rule right before this rule.
-                return GetPreviousAdjustmentRule(rule);
+                return GetPreviousAdjustmentRule(rule, ruleIndex);
             }
 
             return rule;
@@ -282,9 +284,14 @@ namespace System
         /// If the specified rule is the first AdjustmentRule, or it isn't in _adjustmentRules,
         /// then the specified rule is returned.
         /// </summary>
-        private AdjustmentRule GetPreviousAdjustmentRule(AdjustmentRule rule)
+        private AdjustmentRule GetPreviousAdjustmentRule(AdjustmentRule rule, int? ruleIndex)
         {
             Debug.Assert(rule.NoDaylightTransitions, "GetPreviousAdjustmentRule should only be used with NoDaylightTransitions rules.");
+
+            if (ruleIndex.HasValue && 0 < ruleIndex.Value && ruleIndex.Value < _adjustmentRules.Length)
+            {
+                return _adjustmentRules[ruleIndex.Value - 1];
+            }
 
             AdjustmentRule result = rule;
             for (int i = 1; i < _adjustmentRules.Length; i++)
@@ -413,10 +420,11 @@ namespace System
                 dateTime.Kind == DateTimeKind.Utc ? ConvertTime(dateTime, s_utcTimeZone, this, flags, cachedData) :
                 dateTime;
 
-            AdjustmentRule rule = GetAdjustmentRuleForTime(adjustedTime);
+            int? ruleIndex;
+            AdjustmentRule rule = GetAdjustmentRuleForTime(adjustedTime, out ruleIndex);
             if (rule != null && rule.HasDaylightSaving)
             {
-                DaylightTimeStruct daylightTime = GetDaylightTime(adjustedTime.Year, rule);
+                DaylightTimeStruct daylightTime = GetDaylightTime(adjustedTime.Year, rule, ruleIndex);
                 return GetIsAmbiguousTime(adjustedTime, rule, daylightTime);
             }
             return false;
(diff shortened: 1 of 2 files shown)


eerhardt:

tarekgh:
LGTM

tarekgh:
@eerhardt how much perf gain we are getting with this fix? the original issue was claiming it is 3x slower than Windows. so I am wondering with this fix what would be the perf.

eerhardt:
Using the `America/Chicago` time zone

```C#
            DateTime dt;
            for (int i = 0; i < 20; i++)
            {
                Stopwatch watch = new Stopwatch();
                watch.Start();
                for (int j = 0; j < 10000; j++)
                {
                    dt = DateTime.Now; dt = DateTime.Now; dt = DateTime.Now;
                    dt = DateTime.Now; dt = DateTime.Now; dt = DateTime.Now;
                    dt = DateTime.Now; dt = DateTime.Now; dt = DateTime.Now;
                }

                Console.Write(watch.ElapsedMilliseconds + "<br>");
            }
```

Produces

 **MacBook Pro Without Changes** |  **MacBook Pro With Changes**
| -------- | :-------------------------------------: | 
144<br>150<br>138<br>145<br>138<br>138<br>137<br>140<br>139<br>136<br>151<br>138<br>139<br>144<br>143<br>138<br>142<br>144<br>142<br>144<br> | 52<br>52<br>52<br>48<br>54<br>50<br>50<br>52<br>59<br>49<br>49<br>49<br>52<br>53<br>50<br>49<br>50<br>49<br>53<br>53<br> |

For comparison, my Windows machine runs that same code in about 20ms, so we are now below x3 slower than Windows.

Note that the issue title was written incorrectly to start.  After getting the benchmark environment fixed and before doing any perf investigation, `Linux DateTime.Now shows as being 35x slower than Windows DateTime.Now.`.  See https://github.com/dotnet/corefx/issues/3571#issuecomment-145930830

eerhardt:
I've decided to make one more change for this scenario.  Since scanning all the rules in `GetPreviousAdjustmentRule` was taking most of the time, I changed the code to remember the index of the AdjustmentRule into the `_adjustmentRules` array.  This allows us to get the previous adjustment rule without scanning.

With this change, here are the results of the DateTime.Now benchmark:

| **Windows x64** | **MacBook Pro before changes** | **MacBook Pro after changes** |
| :-------------------------------------: | :-------------------------------------: | :-------------------------------------: |
22<br>21<br>20<br>19<br>20<br>19<br>20<br>20<br>20<br>22<br>20<br>20<br>20<br>22<br>19<br>20<br>21<br>20<br>20<br>19<br> | 144<br>150<br>138<br>145<br>138<br>138<br>137<br>140<br>139<br>136<br>151<br>138<br>139<br>144<br>143<br>138<br>142<br>144<br>142<br>144<br> | 31<br>28<br>27<br>32<br>28<br>28<br>28<br>28<br>28<br>30<br>29<br>29<br>27<br>26<br>27<br>28<br>27<br>27<br>27<br>27<br>

@tarekgh @stephentoub @mazong1123 - please take another look

eerhardt:

src/mscorlib/src/System/TimeZoneInfo.cs
```diff
@@ -282,12 +284,23 @@ private AdjustmentRule GetAdjustmentRuleForAmbiguousOffsets(DateTime adjustedTim
...
         /// then the specified rule is returned.
         /// </summary>
-        private AdjustmentRule GetPreviousAdjustmentRule(AdjustmentRule rule)
+        private AdjustmentRule GetPreviousAdjustmentRule(AdjustmentRule rule, int? ruleIndex)
         {
+            Debug.Assert(rule.NoDaylightTransitions, "GetPreviousAdjustmentRule should only be used with NoDaylightTransitions rules.");
+
+            if (ruleIndex.HasValue && 0 < ruleIndex.Value && ruleIndex.Value < _adjustmentRules.Length)
+            {
+                return _adjustmentRules[ruleIndex.Value - 1];
```
stephentoub: I'm not familiar with what the rule indices mean, but is it correct that we compare `ruleIndex < _adjustmentRules.Length` but then access `_adjustmentRules[ruleIndex.Value - 1]`?  Naively in such situations I'd have expected it to be `<=` instead of `<`, or to be `ruleIndex.Value` rather than `ruleINdex.Value-1`.

src/mscorlib/src/System/TimeZoneInfo.cs
```diff
@@ -282,12 +284,23 @@ private AdjustmentRule GetAdjustmentRuleForAmbiguousOffsets(DateTime adjustedTim
...
+            }
+
             AdjustmentRule result = rule;
             for (int i = 1; i < _adjustmentRules.Length; i++)
             {
-                if (rule.Equals(_adjustmentRules[i]))
+                // use ReferenceEquals here instead of AdjustmentRule.Equals because 
+                // ReferenceEquals is much faster. This is safe because all the callers
+                // of GetPreviousAdjustmentRule pass in a rule that was retrieved from
+                // _adjustmentRules.  A different approach will be needed if this ever changes.
```
stephentoub: I assume there's nothing easy we could assert here to help catch if something did ever change?

src/mscorlib/src/System/TimeZoneInfo.cs
```diff
@@ -661,7 +677,8 @@ private static DateTime ConvertTime(DateTime dateTime, TimeZoneInfo sourceTimeZo
             // performance for the normal case at the expense of the 'ArgumentException'
             // case and Loss-less Local special cases.
             //
-            AdjustmentRule sourceRule = sourceTimeZone.GetAdjustmentRuleForTime(dateTime);
+            int? sourceRuleIndex;
+            AdjustmentRule sourceRule = sourceTimeZone.GetAdjustmentRuleForTime(dateTime, out sourceRuleIndex);
```
stephentoub: Is there anything we can/should be validating here about the results, e.g.
`Debug.Assert(sourceRule == null || sourceRuleIndex.HasValue);`?

eerhardt:

src/mscorlib/src/System/TimeZoneInfo.cs
```diff
@@ -282,12 +284,23 @@ private AdjustmentRule GetAdjustmentRuleForAmbiguousOffsets(DateTime adjustedTim
         /// If the specified rule is the first AdjustmentRule, or it isn't in _adjustmentRules,
         /// then the specified rule is returned.
         /// </summary>
-        private AdjustmentRule GetPreviousAdjustmentRule(AdjustmentRule rule)
+        private AdjustmentRule GetPreviousAdjustmentRule(AdjustmentRule rule, int? ruleIndex)
```
tarekgh: >GetPreviousAdjustmentRule(AdjustmentRule rule, int? ruleIndex) [](start = 31, length = 62)

Is it worth to have overload of this method which takes initialized index? this can eliminate the checks of the index initialization

src/mscorlib/src/System/TimeZoneInfo.cs
```diff
@@ -1076,6 +1099,7 @@ private AdjustmentRule GetAdjustmentRuleForTime(DateTime dateTime, bool dateTime
                 int compareResult = CompareAdjustmentRuleToDateTime(rule, previousRule, dateTime, date, dateTimeisUtc);
                 if (compareResult == 0)
                 {
+                    ruleIndex = median;
```
tarekgh: > ruleIndex = median; [](start = 19, length = 20)

what you think if we added the index inside the rule itself? and we don't have to pass it around with the rule.

tarekgh:
I added a couple of suggestion if you  want to consider it. but in general LGTM.

src/mscorlib/src/System/TimeZoneInfo.cs
eerhardt: This code is trying to get the rule immediately preceding the one being specified.  The values coming into the method are the `rule`, and optionally a pre-computed `ruleIndex`, which is the index of the specified `rule`.

The `ruleIndex` is the index into the array of the specified `rule`.  So it can never be `==` to `_adjustmentRules.Length`.

src/mscorlib/src/System/TimeZoneInfo.cs
eerhardt: My best attempt at an assert was the one I added:

`Debug.Assert(rule.NoDaylightTransitions, "GetPreviousAdjustmentRule should only be used with NoDaylightTransitions rules.");`

`NoDaylightTransitions` rules are only ever created on Unix when reading from the tzdata files, and outside callers can't create them since there is no API for `NoDaylightTransitions`.

I also verified that there are no public APIs that take AdjustmentRules on TimeZoneInfo.

src/mscorlib/src/System/TimeZoneInfo.cs
eerhardt: I thought of it, but that would involve changing serialization of these rules, which I wasn't prepared to do with this change.  Or we would have to set the index in all the constructors (both normal and during serialization).  I just felt this was a more targeted change.

src/mscorlib/src/System/TimeZoneInfo.cs
eerhardt: Do you mean just taking `int ruleIndex`?  I don't think there would be any callers of it, since the method that returns the rule can return `null` for the rule, and it returns `null` for the index as well.

src/mscorlib/src/System/TimeZoneInfo.cs
eerhardt: I can add this.  Do you mean for every call site of  `GetAdjustmentRuleForTime`?

eerhardt: I've added assertions for this.

eerhardt:

PR #10951, opened 2017-04-13
[Local GC] Scaffolding for loading a standalone GC

swgillespie:
This is a very early PR that tries to imagine what a standalone-GC-capable runtime might look like. I'm hoping to gather comments about this approach early and perhaps start a discussion about what we should be doing, if this is not a good approach. So I welcome any and all comments!

The objective of this PR is to allow for `FEATURE_STANDALONE_GC` to produce a runtime that is capable of running with or without a standalone GC, in the hopes of paving the way to a future in which `FEATURE_STANDALONE_GC` is enabled by default. This PR makes several changes to the CoreCLR build to reach towards this goal:

1. `FEATURE_STANDALONE_GC` becomes a EE-only feature define - it has no meaning to the GC. The meaning of `FEATURE_STANDALONE_GC` is defined to be "a runtime capable of loading a standalone GC, if one exists".
2. The EE's implementation of `GCToEEInterface` is built twice: once as it is today (now in `gcenv.ee.static.cpp`, as the implementation of the static `GCToEEInterface` class) and once as the implementation of `IGCToCLR`, dispatched virtually. To avoid symbol collisions, this PR places the standalone implementation in a separate namespace.
3. The GC itself is built twice: once as it is today and once with the `BUILD_AS_STANDALONE` macro defined. This macro does what `FEATURE_STANDALONE_GC` previously did - it sets up the GC to be able to be loaded dynamically. In practice, that means that it will use its GC-local implementation of `GCToOSInterface` and will dispatch on `GCToEEInterface` dynamically. Unlikely the EE, though, the two separate builds of the GC will not be linked together. The first, non-`BUILD_AS_STANDALONE` build gets linked into `coreclr.dll` as it does today, while the second `BUILD_AS_STANDALONE` build produces a static archive (that, in the future, will be a shared library).

This PR is an implementation of these ideas. First, `gcenv.ee.cpp` is included by two separate files, `gcenv.ee.standalone.cpp` and `gcenv.ee.static.cpp`, which are both build and linked into `coreclr.dll`. `gcenv.ee.common.cpp` contains code that can be shared between the two implementations. Second, the `gc/standalone` directory contains a CMakeLists.txt file that builds the GC in standalone mode. Finally, this PR contains some scaffold code that would load a standalone GC and initialize it. Since the GC currently doesn't link when building as a shared library, this code path is entirely untested and is only there for illustration purposes and as a rough skeleton for what it might look at when the time comes.

On Windows (release build), building `GCToEEInterface` twice results in a binary size increase of `5632` bytes, an increase of 0.1% over the existing binary size. Until a standalone GC is functional, we'd also probably want to link the `BUILD_AS_STANDALONE` GC into the runtime as well so that we can continue to test it, which would also increase the binary size (although the linker would be able to eliminate a lot of the code shared between the two components).
 
cc people who may be interested: @adityamandaleeka @sergiy-k @Maoni0 @jkotas @janvorli


src/CMakeLists.txt
```diff
@@ -149,6 +149,9 @@ if(IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/jit32")
 endif()
 
 add_subdirectory(gc)
```
jkotas: Do you still need the build the gc directory? The makefile in it is ugly because of it is referencing VM - I was not able to mirror it to corert because of it is causing build breaks there. It would be nice to get rid of it.

src/vm/gcenv.ee.common.cpp
```diff
@@ -0,0 +1,394 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
```
jkotas: Why did this needed to be factored into its own file?

jkotas:
> On Windows (release build), building GCToEEInterface twice results in a binary size increase of 5632 bytes, an increase of 0.1% over the existing binary size. 

I think this sounds fine. You can enable it by default.

> We'd also probably want to link the BUILD_AS_STANDALONE GC into the runtime as well so that we can continue to test it, which would also increase the binary size

I do not think this should be enabled by default. It should stay as you special build.

src/CMakeLists.txt
swgillespie: Yep, we can certainly do that! I can't think of any reason why that couldn't be done today.

sergiy-k:
I agree with @jkotas that we would not link the BUILD_AS_STANDALONE GC in the runtime by default. Instead, I would suggest to change to modify the jobs you have in the CI for standalone GC to compile GC and the runtime with BUILD_AS_STANDALONE and FEATURE_STANDALONE_GC enabled and then load the actual gc.dll.

swgillespie:
The last two commits introduce a `FEATURE_STANDALONE_GC_ONLY` feature define, which (when defined) results in *only* a standalone GC being linked into `coreclr.dll` for use with our CI jobs for testing. When defined, the EE still "loads" the GC dynamically in that it activates it by loading `InitializeGarbageCollector` using `GetProcAddress` and calling it, but the GC is still linked statically into the runtime. This is the "special build" that `build.cmd/sh buildstandalonegc` does.

A couple other fixes are bundled with these two commits to make this possible:
  1. In order to load `InitializeGarbageCollector` through `GetProcAddress`, the symbol must be exported from `libcoreclr.so`. In order to do this with ld, it needs to be in the export symbol list. To avoid polluting the existing exports file, I created a second exports file to use only when building with `FEATURE_STANDALONE_GC_ONLY` that is the same as the existing exports file, but with `InitializeGarbageCollector` on it. On Windows, this isn't necessary because `__declspec(dllexport)` is sufficient. (`__attribute__((visibility("default")))` is apparently not sufficient with clang/ld)

 2. `WRITE_BARRIER_CHECK` is disabled when using `FEATURE_STANDALONE_GC_ONLY`, otherwise the runtime doesn't link. This can be fixed later - it involves globals known to write barriers and the GC, similar to the other globals we already decoupled.

Other thoughts:
1. We will need to change the signature of `InitializeGarbageCollector` at least once in the coming week or so. The reason for this is that the EE first calls `InitializeHeapType` to set the heap type (server or workstation) before calling `InitializeGarbageCollector`. Ideally, we should only be exposing one symbol from `gc.dll` and have it do all of the initialization. This implies that we need some way to pass the heap type to `InitializeGarbageCollector` so that the GC knows what flavor of heap to use. More generally, we need some way to communicate the information from startup configs (concurrent GC, hoard VM, whether or not to affinitize server GC threads) to the GC when initializing. (Even more general, we need a mechanism for the GC to query the EE for its more verbose/debug configurations - what `EEConfig` provides today.)

2. I didn't turn `FEATURE_STANDALONE_GC` on by default as a part of this PR - I can either do that here or do it in a follow-up PR, either one is fine with me.

janvorli:
> In order to load InitializeGarbageCollector through GetProcAddress, the symbol must be exported from libcoreclr.so. In order to do this with ld, it needs to be in the export symbol list. To avoid polluting the existing exports file, I created a second exports file to use only when building with FEATURE_STANDALONE_GC_ONLY that is the same as the existing exports file, but with InitializeGarbageCollector on it. On Windows, this isn't necessary because __declspec(dllexport) is sufficient. (__attribute__((visibility("default"))) is apparently not sufficient with clang/ld)

I would not create a separate exports file, but rather add the InitializeGarbageCollector into the existing one. Having two such files could be a maintenance problem - it would be easy to forget to add a new symbol that needs to be exported to both.

swgillespie:
Ok! I was wary about adding new symbols to the export file because I didn't think we'd want to ship a version of `libcoreclr.so` with `InitializeGarbageCollector` exposed like that. Though, I suppose we *are* already doing that on windows.

swgillespie:
@dotnet-bot help

swgillespie:
@dotnet-bot test Ubuntu x64 Checked standalone_gc

swgillespie:
For some reason the CI has forgotten about the OSX and Ubuntu PR-triggered Standalone GC jobs, despite them being run weekly... https://ci.dot.net/job/dotnet_coreclr/job/master/job/x64_checked_ubuntu_standalone_gc/ . At any rate, I have tested it locally and verified that it works on Linux.

swgillespie:
Would there be any objections to going ahead and merging this? There should not be any implications for non-`FEATURE_STANDALONE_GC` builds.

swgillespie:
Going to close this PR and re-open it against a feature branch.

PR #10955, opened 2017-04-13
Make a GCHandleStore class and interface for use by the VM

adityamandaleeka:
This change formalizes the handle store concept that we just introduced, and replaces all the functions on `IGCHandleTable` that were taking `void* store` as the first argument with class methods on `IGCHandleStore`.

The diff isn't that bad, but it's thrown off by the reordering of things, so I recommend using the "Split" view and focusing on the right side :smile:.

@jkotas @sergiy-k @swgillespie @Maoni0

### DIFF
--- a/src/gc/gchandletable.cpp
+++ b/src/gc/gchandletable.cpp
@@ -17,35 +17,35 @@ IGCHandleManager* CreateGCHandleManager()
 
 void GCHandleStore::Uproot()
 {
-    Ref_RemoveHandleTableBucket(_underlyingBucket);
+    Ref_RemoveHandleTableBucket(&_underlyingBucket);
 }
 
 bool GCHandleStore::ContainsHandle(OBJECTHANDLE handle)
 {
-    return _underlyingBucket->Contains(handle);
+    return _underlyingBucket.Contains(handle);
 }
 
 OBJECTHANDLE GCHandleStore::CreateHandleOfType(Object* object, int type)
 {
-    HHANDLETABLE handletable = _underlyingBucket->pTable[GetCurrentThreadHomeHeapNumber()];
+    HHANDLETABLE handletable = _underlyingBucket.pTable[GetCurrentThreadHomeHeapNumber()];
     return ::HndCreateHandle(handletable, type, ObjectToOBJECTREF(object));
 }
 
 OBJECTHANDLE GCHandleStore::CreateHandleOfType(Object* object, int type, int heapToAffinitizeTo)
 {
-    HHANDLETABLE handletable = _underlyingBucket->pTable[heapToAffinitizeTo];
+    HHANDLETABLE handletable = _underlyingBucket.pTable[heapToAffinitizeTo];
     return ::HndCreateHandle(handletable, type, ObjectToOBJECTREF(object));
 }
 
 OBJECTHANDLE GCHandleStore::CreateHandleWithExtraInfo(Object* object, int type, void* pExtraInfo)
 {
-    HHANDLETABLE handletable = _underlyingBucket->pTable[GetCurrentThreadHomeHeapNumber()];
+    HHANDLETABLE handletable = _underlyingBucket.pTable[GetCurrentThreadHomeHeapNumber()];
     return ::HndCreateHandle(handletable, type, ObjectToOBJECTREF(object), reinterpret_cast<uintptr_t>(pExtraInfo));
 }
 
 OBJECTHANDLE GCHandleStore::CreateDependentHandle(Object* primary, Object* secondary)
 {
-    HHANDLETABLE handletable = _underlyingBucket->pTable[GetCurrentThreadHomeHeapNumber()];
+    HHANDLETABLE handletable = _underlyingBucket.pTable[GetCurrentThreadHomeHeapNumber()];
     OBJECTHANDLE handle = ::HndCreateHandle(handletable, HNDTYPE_DEPENDENT, ObjectToOBJECTREF(primary));
     ::SetDependentHandleSecondary(handle, ObjectToOBJECTREF(secondary));
 
@@ -54,7 +54,7 @@ OBJECTHANDLE GCHandleStore::CreateDependentHandle(Object* primary, Object* secon
 
 GCHandleStore::~GCHandleStore()
 {
-    Ref_DestroyHandleTableBucket(_underlyingBucket);
+    ::Ref_DestroyHandleTableBucket(&_underlyingBucket);
 }
 
 bool GCHandleManager::Initialize()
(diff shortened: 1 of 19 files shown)


adityamandaleeka:
Note: I really would like to rename `IGCHandleTable` before we ship since it has no relation to an actual handle table.

src/gc/gchandletableimpl.h
```diff
@@ -6,33 +6,49 @@
...
+    virtual void* GetHandleContext(OBJECTHANDLE handle);
 
-    virtual OBJECTHANDLE CreateHandleOfType(void* store, Object* object, int type, int heapToAffinitizeTo);
+    virtual IGCHandleStore* GetGlobalHandleStore();
 
-    virtual OBJECTHANDLE CreateHandleWithExtraInfo(void* store, Object* object, int type, void* pExtraInfo);
+    virtual IGCHandleStore* CreateHandleStore(void* context);
 
-    virtual OBJECTHANDLE CreateDependentHandle(void* store, Object* primary, Object* secondary);
+    virtual void DestroyHandleStore(IGCHandleStore* store);
```
jkotas: Shouldn't this rather be a method on IGCHandleStore?

src/gc/gchandletableimpl.h
```diff
@@ -6,33 +6,49 @@
...
+};
+
+extern GCHandleStore* g_gcGlobalHandleStore;
+
+class GCHandleTable : public IGCHandleTable
+{
+public:
+    virtual bool Initialize();
+
+    virtual void Shutdown();
```
jkotas: Why do we have the interface at all? Shouldn't it rather be methods on IGCHeap?

src/gc/gchandletableimpl.h
adityamandaleeka: I had it there at first but moved it here for symmetry with `CreateHandleStore`

src/gc/gchandletableimpl.h
adityamandaleeka: This interface (which will get a few more things added to it) is only about handle-related things, so @Maoni0 and I decided at the start to make it its own interface and keep it separate from `IGCHeap`.

We could move all these functions to a new 'section' in the `IGCHeap` interface, but that interface is already getting pretty broad in terms of what it does, so I'd rather not unless there's a compelling reason to.

jkotas: > We could move all these functions to a new 'section' in the IGCHeap interface, but that interface is already getting pretty broad in terms of what it does.

I do not see a problem with this. I think it is fine to keep adding the methods to the same interface - it is consistent at least.

jkotas: BTW: There is lame inefficiency in the CreateHandle method today that it calls the HomeHeapNumber virtually. It would be nice to fix to avoid this virtual call to get the home heap. Something to keep in mind as you are plumbing it through.

src/gc/gchandletableimpl.h
```diff
@@ -6,33 +6,49 @@
...
 
-    virtual void UprootHandleStore(void* store);
+    virtual OBJECTHANDLE CreateDependentHandle(Object* primary, Object* secondary);
 
-    virtual bool ContainsHandle(void* store, OBJECTHANDLE handle);
+    virtual ~GCHandleStore();
 
-    virtual OBJECTHANDLE CreateHandleOfType(void* store, Object* object, int type);
+private:
+    HandleTableBucket* _underlyingBucket;
```
jkotas: It would be better to have this inline: `HandleTableBucket _underlyingBucket` so that we do not have these unnecessary pointers to pointers to pointers... .

src/gc/gchandletable.cpp
```diff
@@ -8,86 +8,94 @@
...
 
-OBJECTHANDLE GCHandleTable::CreateGlobalHandleOfType(Object* object, int type)
+IGCHandleStore* GCHandleTable::CreateHandleStore(void* context)
 {
-    return ::HndCreateHandle(g_HandleTableMap.pBuckets[0]->pTable[GetCurrentThreadHomeHeapNumber()], type, ObjectToOBJECTREF(object)); 
+#ifndef FEATURE_REDHAWK
+    HandleTableBucket* newBucket = ::Ref_CreateHandleTableBucket(ADIndex((DWORD)(uintptr_t)context));
+    return new (nothrow) GCHandleStore(newBucket);
+#else
+    assert("CreateHandleStore is not implemented when FEATURE_REDHAWK is defined!");
```
jkotas: > CreateHandleStore is not implemented when FEATURE_REDHAWK is defined

This looks odd, but I see that you are not changing it.

src/gc/gchandletableimpl.h
adityamandaleeka: I looked into that and it will involve a few more changes. Some of the `Ref_` functions rely on pointer equivalence when comparing buckets (as in, they check whether the address of the bucket they're looking for matches a pointer value in `g_HandleTableMap.pBuckets`) and `HandleTableBucket`s are currently just heap allocated with `new`.

One way to do this properly would be to create `HandleTableStore`s in the function where we currently create the underlying buckets and store the bucket in the store (and of course in that case `g_HandleTableMap.pBuckets` will contain pointers to buckets that are within `HandleTableStores` so the existing pointer checks will continue to work).

Does that sound good? This is purely GC-side implementation work, so it can be changed independently of the interface work.

src/gc/gchandletable.cpp
adityamandaleeka: Yup, I caught it because of a linker error 😛

src/gc/gchandletableimpl.h
jkotas: Yes, something like this. I was thinking that the most straightforward way to do it would be to factor the initialization from `Ref_CreateHandleTableBucket` into separate function, like:

```
HandleTableBucket * Ref_CreateHandleTableBucket(void* context)
{
    HandleTableBucket * result = new (nothrow) HandleTableBucket();
    if (result == nullptr)
        return nullptr;

    if (!Ref_InitializeHandleTableBucket(result, context))
    {
        delete result;
        return nullptr;
    }

    return result;
}

bool Ref_InitializeHandleTableBucket(HandleTableBucket * bucket, void* context)
{
     // Contains the rest of the code from Ref_CreateHandleTableBucket
     // BTW: I have noticed that Ref_CreateHandleTableBucket will need work to switch over to non-throwing new operators...
}
```

jkotas: > This is purely GC-side implementation work, so it can be changed independently of the interface work.

That's true, but since you are working this work in the live tree, it would be a good practice to do the refactorings together with the perf mitigations to reduce their perf overhead.

adityamandaleeka: Sure, I'll update this PR with the changes we discussed. Thanks!

adityamandaleeka:
@jkotas PTAL

adityamandaleeka:
Thanks!

PR #10956, opened 2017-04-13
Jit: fix unnecessary null checks in some field accesses

### LINKED ISSUE #10942: JIT: unnecessary null checks in address trees that involve field access on byrefs
`P0`'s explicit null check is unnecessary. A similar check appears in the codegen for the proposed span indexer intrinsic expansion #10910, in cases where the span's pointer field cannot be promoted, and the check causes some performance regressions when using the intrinsic expansion instead of using the IL.

The explicit null check is inserted by morph in `fgMorphField`. Because interpretation of `GT_FIELD` is parent node dependent, morph needs to look for child field nodes and pass down the right context. There is some manipulation of address context in `fgMorphSmpOp` but it doesn't get this case right, and so the indefinite offset from `v * 8` is taken to apply to the dereference of `f.p` too.

```C#
unsafe struct F
{
    public F(int* x) { p = x; }
    public int* p;
}

static unsafe int P1(ref F f)
{
    return *(f.p);
}

static unsafe int P0(ref F f, int v)
{
    return *(f.p + v * 8);
}
```
```asm
;; P1
mov      rax, qword ptr [rcx]
mov      eax, dword ptr [rax]
ret

;; P0
cmp      dword ptr [rcx], ecx
mov      rax, qword ptr [rcx]
shl      edx, 3
movsxd   rdx, edx
ret
```

AndyAyersMS:
Morph was sometimes passing the existing MorphAddressContext down
to fgMorphField even when the field access was for a field value.
If that context contained indefinite offsets, morph would then
insert an explicit null check on the object pointer for the field
access.

Typically the field offset is small enough that this explicit check
is not needed. The implicit check done when fetching the field's value
is sufficient protection.

The fix is to have `fgMorphSmpOp` clear out the context for child
`GT_FIELD` nodes, unless the field parent is a `GT_ADDR`. Note
if there is an `op2` node the parent cannot be `GT_ADDR` so these
field children always get an empty context.

No tests added since this kicks in reasonably frequently in corelib and
elsewhere in frameworks.

Closes #10942.

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -11416,6 +11416,20 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
                 }
             }
 
+            // If gtOp1 is a GT_FIELD, we need to pass down the mac if
+            // its parent is GT_ADDR, since the address of the field
+            // is part of an ongoing address computation. Otherwise
+            // op1 represents the value of the field and so any address
+            // calculations it does are in a new context.
+            if ((op1->gtOper == GT_FIELD) && (tree->gtOper != GT_ADDR))
+            {
+                subMac1 = nullptr;
+
+                // The impact of this field's value to any ongoing
+                // address computation is handled below when looking
+                // at op2.
+            }
+
             tree->gtOp.gtOp1 = op1 = fgMorphTree(op1, subMac1);
 
 #if LOCAL_ASSERTION_PROP
@@ -11496,7 +11510,6 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
             // (These are used to convey parent context about how addresses being calculated
             // will be used; see the specification comment for MorphAddrContext for full details.)
             // Assume it's an Ind context to start.
-            MorphAddrContext subIndMac2(MACK_Ind);
             switch (tree->gtOper)
             {
                 case GT_ADD:
@@ -11517,6 +11530,17 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
                 default:
                     break;
             }
+
+            // If gtOp2 is a GT_FIELD, we must be taking its value,
+            // so it should evaluate its address in a new context.
+            if (op2->gtOper == GT_FIELD)
+            {
+                // The impact of this field's value to any ongoing
+                // address computation is handled above when looking
+                // at op1.
+                mac = nullptr;
+            }
+
             tree->gtOp.gtOp2 = op2 = fgMorphTree(op2, mac);
 
             /* Propagate the side effect flags from op2 */


AndyAyersMS:
@briansull PTAL
cc @dotnet/jit-contrib 

Jit-diff result
```
Total bytes of diff: -3019 (-0.02 % of base)
    diff is an improvement.
        -590 : System.Runtime.Extensions.dasm (-0.39 % of base)
        -565 : Microsoft.CodeAnalysis.dasm (-0.07 % of base)
        -517 : System.Security.Cryptography.Primitives.dasm (-2.95 % of base)
        -424 : System.Linq.Parallel.dasm (-0.07 % of base)
        -220 : System.Net.Http.dasm (-0.11 % of base)
19 total files with size differences (19 improved, 0 regressed).
           2 : Microsoft.CodeAnalysis.VisualBasic.dasm - AnonymousTypeTemplateSymbol:SynthesizeDebuggerDisplayAttribute():ref:this
        -373 : System.Security.Cryptography.Primitives.dasm - <ReadAsyncCore>d__42:MoveNext():this
        -190 : System.Runtime.Extensions.dasm - <ReadAsyncInternal>d__63:MoveNext():this
        -144 : System.Security.Cryptography.Primitives.dasm - <WriteAsyncCore>d__49:MoveNext():this
        -133 : System.Runtime.Extensions.dasm - <ReadBufferAsync>d__65:MoveNext():this
         -94 : Microsoft.CodeAnalysis.dasm - <ProcessCompilationEventsAsync>d__58:MoveNext():this
329 total methods with size differences (328 improved, 1 regressed).
```
Intereststingly this seems to kick in frequently in async code (which is featured in System.Runtime.Extensions). Surveyed many of the diffs, and they all showed elimination of directly anticipated explicit null checks. A few examples:
```
;;; System.Runtime.Extensions <ReadLineAsyncInternal>d__59:MoveNext():this
;;; Before
       mov      rcx, bword ptr [rbp+10H]
       cmp      dword ptr [rcx], ecx
       mov      rcx, bword ptr [rbp+10H]
       mov      rcx, gword ptr [rcx]
;;; After
       mov      rcx, bword ptr [rbp+10H]
       mov      rcx, gword ptr [rcx]

;;; System.Private.CoreLib System.Number:ParseNumber(byref,int,byref,ref,ref,bool):bool
;;; Before
       cmp      dword ptr [rbx], ebx
       mov      rcx, qword ptr [rbx+8]
;;; After
       mov      rcx, qword ptr [rbx+8]
```

AndyAyersMS:
Some of the diffs are just crazy. For instance there are 6 explicit null checks more or less in a row here
in the code for System.Security.Cryptography.Primitives `<ReadAsyncCore>d__42:MoveNext():this`. Only the first one is needed.
<table>
<tr><th>Before</th><th>After</th></tr>
<tr valign="top"><td><pre><code>
       mov      rcx, bword ptr [rbp+10H]
       mov      rcx, gword ptr [rcx]
       ....
       mov      rcx, bword ptr [rbp+10H]
       cmp      dword ptr [rcx], ecx
       mov      rcx, bword ptr [rbp+10H]
       mov      rcx, gword ptr [rcx]
       mov      r8, qword ptr [(reloc)]
       mov      rcx, gword ptr [rcx+r8+24]
       mov      r8, bword ptr [rbp+10H]
       cmp      dword ptr [r8], r8d
       mov      r8, bword ptr [rbp+10H]
       mov      r8, gword ptr [r8]
       mov      rdx, qword ptr [(reloc)]
       mov      r8, gword ptr [r8+rdx+24]
       mov      rcx, bword ptr [rbp+10H]
       mov      rcx, gword ptr [rcx]
</code></pre></td>
<td><pre><code>
       mov      rcx, bword ptr [rbp+10H]
       mov      rcx, gword ptr [rcx]
       ....
       mov      r8, gword ptr [rcx+r8+24]
       mov      rcx, r8
</td></code></pre></tr></table>

AndyAyersMS:
[The Tizen failure](https://ci.dot.net/job/dotnet_coreclr/job/master/job/armel_cross_release_tizen_prtest/1081/) is for b37131 which throws a NRE, but it comes from an array bounds check. On x64 there are no before/after diffs for this test. Also the Tizen debug leg passed and the jit should be generating the same code there as in the release leg. So suspect it is unrelated.

AndyAyersMS:
Let's see if the failure repros
@dotnet-bot retest Tizen armel Cross Release Build

briansull:
LGTM, Although this area could still be reworked further.

AndyAyersMS:
Do you have specific suggestions? Should we open an issue?

src/jit/morph.cpp
```diff
@@ -11416,6 +11416,20 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
                 }
             }
 
+            // If gtOp1 is a GT_FIELD, we need to pass down the mac if
+            // its parent is GT_ADDR, since the address of the field
+            // is part of an ongoing address computation. Otherwise
+            // op1 represents the value of the field and so any address
+            // calculations it does are in a new context.
+            if ((op1->gtOper == GT_FIELD) && (tree->gtOper != GT_ADDR))
```
JosephTremoulet: If you move this up above the `if (tree->gtOper == GT_ADD && subMac1 != nullptr)` check on line 11393, we could avoid doing that work when a `GT_FIELD` feeds a `GT_ADD`.

AndyAyersMS: I'll probably leave it as is because it is kind of subtle (maybe this is what Brian was alluding to).

We can change this check as you suggest, but if we also do likewise for the similar `op2` check, which seems like an entirely plausible thing to do, we could miss the need to null check an (ADD, FIELD, FIELD) case.

Probably more straightforward to account for the impact of `op1` on any ongoing address computation when we're looking at `op1` (and similarly for `op2`) rather than the swapped checking we do here.

Also not clear to me why other binary operators aren't considered here.

JosephTremoulet: Why do we need to null check (ADD, FIELD, FIELD)?

erozenfeld:
We can also look into extending optFoldNullCheck in EarlyProp to clean up cases like this.

src/jit/morph.cpp
AndyAyersMS: Well, I could be confused. 

It seems to me we need to inform whatever is going on upstream that when handling the ADD where at least one of the child op's is a FIELD we have discovered an indefinite offset, so upstream code may need an explicit null check.

JosephTremoulet: I'd think that (somewhat guessing here):

  1. If the pointer-valued child FIELD is loading a managed pointer, then we're not required to insert a null check (otherwise we'd be adding an additional null check at virtually every Span indexer, no?), and instead can rely on the null check have been performed before storing the value into the field.
  2. If the pointer-valued child FIELD is loading an object pointer, then us seeing an ADD here instead of an ADDR (FIELD obj, f) here indicates that the input was some sort of unsafe code that doesn't require a null check.

In any case, passing the address context on down to the recursive call that processes the 2nd FIELD could only lead to null checking the address that the FIELD loads from, which of course wouldn't serve the purpose of null checking the loaded value to which the indefinite offset is added.

AndyAyersMS:
I think many of these cases are ones where the fields are address-taken so don't end up in SSA.

AndyAyersMS:
Will defer any future cleanup/rework since this is mainly to unblock the span indexer changes.

src/jit/morph.cpp
briansull: Yeah what Joe mentioned what I was seeing.  Someone probably Dave Detlefs added this code.
The whole thing feels a bit off to me.  You can see that right before we call the MorphTree on the op1 and op2 we setup this special Context arg and pass it to just the three versions of the Morph function that can use it. 

I guess it does eliminate the extra null check for GT_FIELD expansions but I rather we did this a different way.

briansull: The null checks that are being added are for the case where we add some offset to a base pointer (typically a GC pointer) and the amount that we add might push us past the set of unmapped pages that start at address 0.  So if for instance the base pointer was a nullptr and we added 10,000 to pushing up into a valid mapped page we would access some other valid memory and no generate an access violation.  It is also be a security issue if we allow this pattern to occur.

briansull:
LGTM

src/jit/morph.cpp
JosephTremoulet: > the case where we add some offset to a base pointer (typically a GC pointer) and the amount that we add might push us past the set of unmapped pages

Right, but we're not saying that we insert null checks for every instance of pointer arithmetic we can find without a fold-onto-able dereference, are we?  I presume this code is here to implement the null check for `ldflda` (but at the same time do the optimization to omit its explicit null check if we find that it's feeding a dereference that would be on the guard page if the object were null).

>The whole thing feels a bit off to me

I agree.  I think it would feel cleaner to insert null checks where the language requires them (ldflda) in the importer, and fold them as an optimization, or perhaps have the importer simply flag them (`GTF_ADDR_NEEDS_NULLCHECK` or something) and then this code could clearly be generating or omitting those null checks, rather than leaving us guessing at what contract it's implementing.

AndyAyersMS: I'll open a new issue to track rethinking how we do this.

AndyAyersMS: #10984

PR #10957, opened 2017-04-13
Do not honor attribute HandleProcessCorruptedStateExceptions

### LINKED ISSUE #9045: Strip corrupted state exceptions handling

rahku:
Fixes #9045

### DIFF
--- a/src/vm/exceptionhandling.cpp
+++ b/src/vm/exceptionhandling.cpp
@@ -1011,6 +1011,10 @@ ProcessCLRException(IN     PEXCEPTION_RECORD   pExceptionRecord
                 CEHelper::SetupCorruptionSeverityForActiveException((STState == ExceptionTracker::STS_FirstRethrowFrame), (pTracker->GetPreviousExceptionTracker() != NULL),
                                                                     CEHelper::ShouldTreatActiveExceptionAsNonCorrupting());
             }
+
+            // Failfast if exception indicates corrupted process state            
+            if (pTracker->GetCorruptionSeverity() == ProcessCorrupting)
+                EEPOLICY_HANDLE_FATAL_ERROR(pExceptionRecord->ExceptionCode);
         }
 #endif // FEATURE_CORRUPTING_EXCEPTIONS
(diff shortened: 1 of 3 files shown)


rahku:
It does not do full cleanup of all the code required for handling HandleProcessCorruptedStateExceptions Attribute. But removes its functionality.

rahku:
PTAL @jkotas @gkhanna79

jkotas:
Can we fail fast immediately when the corrupting exception hits the managed code? It sounds like we will still run a lot of code (looking for handlers, etc.) with this fix when this happens.

src/vm/excep.cpp
```diff
@@ -11536,28 +11418,7 @@ BOOL CEHelper::CanMethodHandleException(CorruptionSeverity severity, PTR_MethodD
...
-
-            // Check if the method can handle the severity specified in the exception object.
-            fLookForExceptionHandlersInMethod = CEHelper::CanMethodHandleCE(pMethodDesc, severity, fCalculateSecurityInfo);
-        }
-        else
-        {
-            // Method is in a Pre-V4 assembly - allow it to be examined for processing the CE
-            fLookForExceptionHandlersInMethod = TRUE;
-        }
+        fLookForExceptionHandlersInMethod = FALSE;
```
gkhanna79: This is not an appropriate change - not only we are incompletely removing the functionality but if we need to add additional policies for this feature, this will not be available.

The right fix would be to check for CorruptionSeverity against the ExceptionTracker/ExInfo, when it is setup early in the personality routine, and failfast if it is corrupting.

rahku:
can you PTAL this now terminates the process in first pass of EH. I also modified a jit calli tests which was depending on HandleProcessCorruptedState attribute. I will add tests for this changed functionality in corefx repo.

PR #10958, opened 2017-04-13
Preserve VNs in fgMorphIntoHelperCall.

pgavlin:
This method is used to transform a node into a semantically-equivalent
helper call, and should preserve any value numbers that are present on
the the node.


pgavlin:
@briansull @JosephTremoulet @dotnet/jit-contrib PTAL

pgavlin:
> Wouldn't hurt to put some comments in the testcase indicating what the issue was. Does it fail to compile without this fix, or just produce slower code?

Can do. The original repro was yet another remorphing bug: remorphing during CSE was changing a long-typed `GT_UMOD` to a helper call without preserving the value number, which was causing some code that attempted to extract exceptions from VNs to fail.

briansull:
Looks fine

pgavlin:
The two failing builds timed out.

@dotnet-bot 

PR #10963, opened 2017-04-14
[x86/Linux] Use appropriate FCALL macro for DoToDecimal

parjong:
COMCurrency::DoToDecial takes CY as the 2nd argument (which is essentially same as LONGLONG). Unfortunately, this method does not use FCXXXX2_IV macro, which results in assert failure during FX System.Runtime.Tests run.

This commit replace FCXXXX2 with FCXXXX2_IV to fix the assertion failure during FX System.Runtime.Tests run.

### DIFF
--- a/src/classlibnative/bcltype/currency.cpp
+++ b/src/classlibnative/bcltype/currency.cpp
@@ -16,7 +16,7 @@
 #include "string.h"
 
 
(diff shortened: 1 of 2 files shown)


parjong:

parjong:

PR #10967, opened 2017-04-14
Change ICLRRuntimeHost2 to ICLRRuntimeHost4 for Windows corerun.

### LINKED ISSUE #10960: Use ICLRRuntimeHost4 in Windows corerun
Based on the discuss in [this PR](https://github.com/dotnet/coreclr/pull/10842#issuecomment-293937355), we need to unify the Windows and Unix corerun. Related issue #6206

mazong1123:
Since we've introduced ICLRRuntimeHost4 in PR #10842 for Unix, we apply it for Windows corerun as well.

Fix #10960

### DIFF
--- a/src/coreclr/hosts/coreconsole/coreconsole.cpp
+++ b/src/coreclr/hosts/coreconsole/coreconsole.cpp
@@ -541,15 +541,17 @@ bool TryRun(const int argc, const wchar_t* argv[], Logger &log, const bool verbo
 
     log << W("Unloading the AppDomain") << Logger::endl;
 
-    hr = host->UnloadAppDomain(
+    hr = host->UnloadAppDomain2(
         domainId, 
-        true);                          // Wait until done
+        true,
+        (int *)&exitCode);                          // Wait until done
 
     if (FAILED(hr)) {
         log << W("Failed to unload the AppDomain. ERRORCODE: ") << hr << Logger::endl;
         return false;
     }
 
+    log << W("App domain unloaded exit value = ") << exitCode << Logger::endl;
 
     //-------------------------------------------------------------
 
--- a/src/coreclr/hosts/coreconsole/coreconsole.cpp
+++ b/src/coreclr/hosts/coreconsole/coreconsole.cpp
@@ -71,7 +71,7 @@ class HostEnvironment
     // The list of paths to the assemblies that will be trusted by CoreCLR
     StringBuffer m_tpaList;
 
-    ICLRRuntimeHost2* m_CLRRuntimeHost;
+    ICLRRuntimeHost4* m_CLRRuntimeHost;
 
     HMODULE m_coreCLRModule;
 
@@ -326,8 +326,8 @@ public:
         return m_hostExeName;
     }
 
-    // Returns the ICLRRuntimeHost2 instance, loading it from CoreCLR.dll if necessary, or nullptr on failure.
-    ICLRRuntimeHost2* GetCLRRuntimeHost() {
+    // Returns the ICLRRuntimeHost4 instance, loading it from CoreCLR.dll if necessary, or nullptr on failure.
+    ICLRRuntimeHost4* GetCLRRuntimeHost() {
         if (!m_CLRRuntimeHost) {
 
             if (!m_coreCLRModule) {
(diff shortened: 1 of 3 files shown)


mazong1123:
@jkotas Could you please take a look? The arm build CI failure should not be caused by my changes.

jkotas:
The change looks good so far, but I do not see the new method getting called anywhere. Does it actually work?

mazong1123:
@jkotas I don't have a Windows box by hand this weekend. I'll confirm it on Monday.

BTW, I saw @stephentoub add a test in corefx repo (https://github.com/dotnet/corefx/pull/9939/commits/7608f2462a10f278f7407b76514b0fd3fda1ea9d), do we need to add a test in coreclr repo either?

jkotas:
We need to re-enable the test in the corefx repo, and potentially add more variants of it to exercise the distinct cases. Adding a test to CoreCLR repo for this one is not necessary.

mazong1123:
All right. I'll first make sure the Environment.ExitCode bug fixed on Windows so that this PR could be merged. Then I'll work on corefx repo for the test stuffs.

hseok-oh:

mazong1123:
@jkotas I just tried following code on Windows 7 x64:
```cpp
using System;
class Program
{
    static int Main(string[] args)
    {
        AppDomain.CurrentDomain.ProcessExit += (sender, e) => {
            Console.WriteLine("Original exit code:" + Environment.ExitCode);
            Environment.ExitCode = 123;
            Console.WriteLine("New exit code:" + Environment.ExitCode);
        };
        return 12;
    }
}
```

The output is

```
Original exit code:12
New exit code:123

C:\>echo %errorlevel%
12
```

So on Windows,  %errorlevel% won't changed after Environment.ExitCode set. I'll investigate this issue later.

jkotas:
The problem is that you are not calling UnloadAppDomain2 in this change anywhere. Changing the UnloadAppDomain to UnloadAppDomain2 in the files you are touching in this change should fix the problem.

mazong1123:
Thanks @jkotas . 

One thing to confirm: I just noticed that the `exitCode` variable on Windows is declared as  `DWORD` which is `unsigned long`. However, `wmain` returns `exitCode` as `int` 

Quote from [MSDN](https://msdn.microsoft.com/en-us/library/system.environment.exitcode(v=vs.110).aspx) also said the `exitCode` is an `int` instead of `unsigned long`:

>The ExitCode property is a signed 32-bit integer. To prevent the property from returning a negative exit code, you should not use values greater than or equal to 0x80000000.

So I guess I can safely cast `exitCode` to int as UnloadAppDomain2 needs an `int` instead of `DWORD`. Please let me know if there's any risk for doing this. Thanks.

mazong1123:
With the new commit, %errorlevel% can be correctly set via Environment.ExitCode:

Example 1:
```csharp
using System;
class Program
{
    static int Main(string[] args)
    {
        AppDomain.CurrentDomain.ProcessExit += (sender, e) => {
            Console.WriteLine("Original exit code:" + Environment.ExitCode);
            Environment.ExitCode = 123;
            Console.WriteLine("New exit code:" + Environment.ExitCode);
        };
        return 12;
    }
}
```

The output is

```
Original exit code:12
New exit code:123

C:\>echo %errorlevel%
123
```

Example 2:

```csharp
using System;

namespace hl
{
class Program
{
    {
        Environment.ExitCode = 123;
    }
}
}
```
The output is
```
C:\>echo %errorlevel%
123
```

jkotas:
> So I guess I can safely cast exitCode to int as UnloadAppDomain2 needs an int instead of DWORD.

Correct.

mazong1123:
@jkotas I added a new commit so the bug on Windows should be fixed (see https://github.com/dotnet/coreclr/pull/10967#issuecomment-295127865). Would you please take a look? Thanks :)

jkotas:
LGTM. Thanks!

jkotas:
The corefx test is still disabled against this bug. https://github.com/dotnet/corefx/blob/0ebc867d6bad9d777dd61afb78d0445c86143b34/src/System.Runtime.Extensions/tests/System/Environment.Exit.cs#L52

@mazong1123 Could you please submit PR to enable the test? Removing ActiveIssue attribute should do the trick.

PR #10972, opened 2017-04-14
[Ryujit/ARM32] LSRA support for arm32 floating-point register allocation

### LINKED ISSUE #9214: [Ryujit/ARM32] LSRA support for arm32 floating-point register allocation
Current LSRA can not handle variable with double type which requires a pair of register correctly for ARM32.  

About 34 tests are filed in `JIT/CodeGenBringUpTests/` due to assertion failure related to TYP_DOUBLE.

I made a draft PR (#9131) and got a feedback that this might require a bunch of implementation.
From now on, this issue will address LSRA issues related to double type variable in arm32.
I will update this issue after further investigation.

(This is a part of #8496)

hqueue:
Fix  #9214

Initial implementation of LSRA support for ARM32 floating point register allocation.
This fixes all LSRA failure of TYP_DOUBLE in `JIT/CodeGenBringUpTests/` and addresses more than 16000 NYI assertion related to TYP_DOUBLE in CoreCLR unit tests observed at https://github.com/dotnet/coreclr/issues/8496#issuecomment-294250881.
```
16155, c:\gh\coreclr\src\jit\lsraarm.cpp:230 - NYI: lowering double
```

For TYP_DOUBLE, we should handle overlapping FP registers.   
* Enable lowering and register allocation for TYP_DOUBLE
* Update `previousInterval` together in `tryAllocateFreeReg()` and `unassignPhysReg()`
* Consider TYP_DOUBLE for free registers in `freeRegister()` and `allocateRegisters()`
* Consider TYP_DOUBLE when updating reigsters in `resolveLocalRef()` and  `processBlockStartLocations()`

(This is a part of #8496)

## Example

```
FAILED   - [  14][   8s]JIT/CodeGenBringUpTests/DblAdd/DblAdd.sh
               BEGIN EXECUTION
               /mnt/albireo/mywork/device/rpi3/unittests/coreoverlays/coreoverlay.test20170414.lsra_fp/corerun DblAdd.exe

               Assert failure(PID 8766 [0x0000223e], Thread: 8766 [0x223e]): Assertion failed 'assignedInterval != nullptr' in 'BringUpTest:Main():int' (IL size 62)

                   File: /opt/code/github/hqueue/coreclr/src/jit/lsra.cpp Line: 6221
                   Image: /mnt/albireo/mywork/device/rpi3/unittests/coreoverlays/coreoverlay.test20170414.lsra_fp/corerun

               ./DblAdd.sh: line 132:  8766 Aborted                 (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" DblAdd.exe $CLRTestExecutionArguments
               Expected: 100
               Actual: 134
               END EXECUTION - FAILED
```

# Test
When running with   `COMPlus_AltJit=*   COMPlus_AltJitAssertOnNYI=0` for  `JIT/CodeGenBringUpTests/`

## Before
```
=======================
     Test Results
=======================
# CoreCLR Bin Dir  : 
# Tests Discovered : 139
# Passed           : 104
# Failed           : 35
# Skipped          : 0
=======================
```
Among 35 failures, 34 failurs are due to above LSRA assertion failure.

## After

```
=======================
     Test Results
=======================
# CoreCLR Bin Dir  : 
# Tests Discovered : 139
# Passed           : 138
# Failed           : 1
# Skipped          : 0
=======================
```
All 34 failures due to LSRA assertions are gone.

### DIFF
--- a/src/jit/lsra.cpp
+++ b/src/jit/lsra.cpp
@@ -6563,42 +6563,38 @@ regNumber LinearScan::rotateBlockStartLocation(Interval* interval, regNumber tar
 #endif // DEBUG
 
 #ifdef _TARGET_ARM_
-//-------------------------------------------------------------------------------
-// isOverlappedRegRecord : Is interval assigned to overlapped register?
+//--------------------------------------------------------------------------------------
+// isSecondHalfReg: Test if recRec is second half of double reigster
+//                  which is assigned to an interval.
 //
 // Arguments:
-//    assignedInterval - an interval assigned to some register
-//    physRegRecord - a register which be tested
+//    regRec - a register to be tested
+//    interval - an interval which is assigned to some register
+//
+// Assumptions:
+//    The interval must have been assigned to some register
 //
 // Return Value:
-//    True if an interval is assigned to a register which overlaps physRegRecord
+//    True only if regRec is second half of assignedReg in interval
 //
-bool LinearScan::isOverlappedRegRecord(Interval* assignedInterval, RegRecord* physRegRecord)
+bool LinearScan::isSecondHalfReg(RegRecord* regRec, Interval* interval)
 {
-    RegRecord* assignedReg = assignedInterval->assignedReg;
+    RegRecord* assignedReg = interval->assignedReg;
 
-    if (assignedReg == physRegRecord)
+    if (assignedReg != nullptr && interval->registerType == TYP_DOUBLE)
     {
-        return true;
-    }
-    else if (assignedInterval->registerType == TYP_DOUBLE)
-    {
-        // Check whether physRegRecord is overlapped by assignedReg
-        RegRecord* altPhysRegRecord;
-        regNumber  regNum = physRegRecord->regNum;
+        // interval should have been allocated to a valid double register
+        assert(genIsValidDoubleReg(assignedReg->regNum));
 
-        if (genIsValidFloatReg(regNum))
-        {
-            // Should not be a valid double reg
-            assert(!genIsValidDoubleReg(regNum));
+        // Find a second half RegRecord of double register
+        regNumber firstRegNum  = assignedReg->regNum;
+        regNumber secondRegNum = REG_NEXT(firstRegNum);
 
-            // Make regNum of overlapping double register
-            regNum = REG_PREV(regNum);
-            assert(genIsValidDoubleReg(regNum));
+        assert(genIsValidFloatReg(secondRegNum) && !genIsValidDoubleReg(secondRegNum));
 
-            altPhysRegRecord = getRegisterRecord(regNum);
-            return assignedReg == altPhysRegRecord;
-        }
+        RegRecord* secondRegRec = getRegisterRecord(secondRegNum);
+
+        return secondRegRec == regRec;
     }
 
     return false;
@@ -6815,7 +6811,7 @@ void LinearScan::processBlockStartLocations(BasicBlock* currentBlock, bool alloc
                 }
 #ifdef _TARGET_ARM_
                 // Consider overlapping floating point register for TYP_DOUBLE
-                else if (!assignedInterval->isConstant && isOverlappedRegRecord(assignedInterval, physRegRecord))
+                else if (!assignedInterval->isConstant && isSecondHalfReg(physRegRecord, assignedInterval))
                 {
                     assert(assignedInterval->assignedReg->regNum == REG_PREV(physRegRecord->regNum));
                 }
@@ -6954,7 +6950,7 @@ void LinearScan::freeRegister(RegRecord* physRegRecord)
 
                     // Don't call unassignPhysReg() again, because two FP registers of a double reister are already
                     // unassigned together when this function is called with valid double register number.
-                    regNumber  prevRegNum   = REG_PREV(physRegRecord->regNum);
+                    regNumber  prevRegNum    = REG_PREV(physRegRecord->regNum);
                     RegRecord* prevRegRecord = getRegisterRecord(prevRegNum);
 
                     // Both FP registers of a double reigster should not be allocated to the interval anymore.
--- a/src/jit/lsra.cpp
+++ b/src/jit/lsra.cpp
@@ -6292,6 +6292,21 @@ void LinearScan::unassignPhysReg(RegRecord* regRec, RefPosition* spillRefPositio
     checkAndClearInterval(regRec, spillRefPosition);
     regNumber thisRegNum = regRec->regNum;
 
+#ifdef _TARGET_ARM_
+    regNumber  nextRegNum = REG_NA;
+    RegRecord* nextRegRec = nullptr;
+
+    // Prepare second half RegRecord of a double register for TYP_DOUBLE
+    if (assignedInterval->registerType == TYP_DOUBLE)
+    {
+        assert(isFloatRegType(regRec->registerType));
+        assert(genIsValidDoubleReg(regRec->regNum));
+
+        nextRegNum = REG_NEXT(regRec->regNum);
+        nextRegRec = getRegisterRecord(nextRegNum);
+    }
+#endif // _TARGET_ARM_
+
 #ifdef DEBUG
     if (VERBOSE && !dumpTerse)
     {
@@ -6395,18 +6410,17 @@ void LinearScan::unassignPhysReg(RegRecord* regRec, RefPosition* spillRefPositio
         regRec->assignedInterval = regRec->previousInterval;
         regRec->previousInterval = nullptr;
 #ifdef _TARGET_ARM_
-        // Update overlapping floating point register for TYP_DOUBLE
+        // Update second half RegRecord of a double register for TYP_DOUBLE
         if (regRec->assignedInterval->registerType == TYP_DOUBLE)
         {
             assert(isFloatRegType(regRec->registerType));
-
-            regNumber  nextRegNum = REG_NEXT(regRec->regNum);
-            RegRecord* nextRegRec = getRegisterRecord(nextRegNum);
+            assert(genIsValidDoubleReg(regRec->regNum));
 
             nextRegRec->assignedInterval = nextRegRec->previousInterval;
             nextRegRec->previousInterval = nullptr;
         }
-#endif
+#endif // _TARGET_ARM_
+
 #ifdef DEBUG
         if (spill)
         {
@@ -6423,6 +6437,18 @@ void LinearScan::unassignPhysReg(RegRecord* regRec, RefPosition* spillRefPositio
     {
         regRec->assignedInterval = nullptr;
         regRec->previousInterval = nullptr;
+
+#ifdef _TARGET_ARM_
+        // Update second half RegRecord of a double register for TYP_DOUBLE
+        if (assignedInterval->registerType == TYP_DOUBLE)
+        {
+            assert(isFloatRegType(regRec->registerType));
+            assert(genIsValidDoubleReg(regRec->regNum));
+
+            nextRegRec->assignedInterval = nullptr;
+            nextRegRec->previousInterval = nullptr;
+        }
+#endif // _TARGET_ARM_
     }
 }
 
--- a/src/jit/lsra.cpp
+++ b/src/jit/lsra.cpp
@@ -5605,9 +5605,11 @@ regNumber LinearScan::tryAllocateFreeReg(Interval* currentInterval, RefPosition*
             {
                 availablePhysRegInterval->previousInterval = intervalToUnassign;
 #ifdef _TARGET_ARM_
-                // TODO-ARM-Throughput: For ARM,  we don't have to keep a same previous interval in two RegRecords,
-                // if we can use a interval from one RegRecord. But I'm not sure yet.
-                // Please take a look at unassignPhysReg().
+                // TODO-ARM-Throughput: For ARM, this should not be necessary, i.e. keeping a same
+                // previous interval in two RegRecords, because we will always manage the register
+                // assignment of TYP_DOUBLE intervals together.
+                // Later we should be able to remove this and update unassignPhysReg() where
+                // previousInterval is used. Please also take a look at unassignPhysReg().
 
                 // Update overlapping floating point register for TYP_DOUBLE
                 if (intervalToUnassign->registerType == TYP_DOUBLE)
@@ -6289,7 +6291,7 @@ void LinearScan::unassignPhysReg(RegRecord* regRec, RefPosition* spillRefPositio
 {
     Interval* assignedInterval = regRec->assignedInterval;
     assert(assignedInterval != nullptr);
-    checkAndClearInterval(regRec, spillRefPosition);
+
     regNumber thisRegNum = regRec->regNum;
 
 #ifdef _TARGET_ARM_
@@ -6304,6 +6306,20 @@ void LinearScan::unassignPhysReg(RegRecord* regRec, RefPosition* spillRefPositio
 
         nextRegNum = REG_NEXT(regRec->regNum);
         nextRegRec = getRegisterRecord(nextRegNum);
+
+        // Both two RegRecords should have been assigned to the same interval.
+        assert(assignedInterval == nextRegRec->assignedInterval);
+    }
+#endif // _TARGET_ARM_
+
+    checkAndClearInterval(regRec, spillRefPosition);
+
+#ifdef _TARGET_ARM_
+    if (assignedInterval->registerType == TYP_DOUBLE)
+    {
+        // Both two RegRecords should have been unassigned together.
+        assert(regRec->assignedInterval == nullptr);
+        assert(nextRegRec->assignedInterval == nullptr);
     }
 #endif // _TARGET_ARM_
 
@@ -6572,7 +6588,7 @@ regNumber LinearScan::rotateBlockStartLocation(Interval* interval, regNumber tar
 //    interval - an interval which is assigned to some register
 //
 // Assumptions:
-//    The interval must have been assigned to some register
+//    None
 //
 // Return Value:
 //    True only if regRec is second half of assignedReg in interval
@@ -6811,9 +6827,9 @@ void LinearScan::processBlockStartLocations(BasicBlock* currentBlock, bool alloc
                 }
 #ifdef _TARGET_ARM_
                 // Consider overlapping floating point register for TYP_DOUBLE
-                else if (!assignedInterval->isConstant && isSecondHalfReg(physRegRecord, assignedInterval))
+                else if (!assignedInterval->isConstant && assignedInterval->registerType == TYP_DOUBLE)
                 {
-                    assert(assignedInterval->assignedReg->regNum == REG_PREV(physRegRecord->regNum));
+                    assert(!assignedInterval->isActive || isSecondHalfReg(physRegRecord, assignedInterval));
                 }
 #endif // _TARGET_ARM_
                 else
@@ -6944,22 +6960,9 @@ void LinearScan::freeRegister(RegRecord* physRegRecord)
             if (nextRefPosition == nullptr || RefTypeIsDef(nextRefPosition->refType))
             {
 #ifdef _TARGET_ARM_
-                if (assignedInterval->registerType == TYP_DOUBLE && !genIsValidDoubleReg(physRegRecord->regNum))
-                {
-                    // physRegRecord is a second half of a double register.
-
-                    // Don't call unassignPhysReg() again, because two FP registers of a double reister are already
-                    // unassigned together when this function is called with valid double register number.
-                    regNumber  prevRegNum    = REG_PREV(physRegRecord->regNum);
-                    RegRecord* prevRegRecord = getRegisterRecord(prevRegNum);
-
-                    // Both FP registers of a double reigster should not be allocated to the interval anymore.
-                    assert(assignedInterval->assignedReg != prevRegRecord);
-                    assert(assignedInterval->assignedReg != physRegRecord);
-                }
-                else
+                assert((assignedInterval->registerType != TYP_DOUBLE) || genIsValidDoubleReg(physRegRecord->regNum));
 #endif // _TARGET_ARM_
-                    unassignPhysReg(physRegRecord, nullptr);
+                unassignPhysReg(physRegRecord, nullptr);
             }
         }
     }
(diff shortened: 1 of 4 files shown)


hqueue:
@dotnet-bot please test Windows_NT x64 Debug Build and Test

hqueue:

hqueue:
@dotnet/jit-contrib PTAL

/cc: @dotnet/arm32-contrib

hseok-oh:

src/jit/lsra.cpp
```diff
@@ -6373,6 +6387,24 @@ void LinearScan::unassignPhysReg(RegRecord* regRec, RefPosition* spillRefPositio
...
+            regNumber  nextRegNum = REG_NEXT(regRec->regNum);
+            RegRecord* nextRegRec = getRegisterRecord(nextRegNum);
+
+            nextRegRec->assignedInterval = nextRegRec->previousInterval;
+            nextRegRec->previousInterval = nullptr;
+
+            // TODO-ARM-Throughput: For ARM, we can reuse regRec->previousInterval,
+            // if following assertion is always true.
+            // Please take a look at tryAllocateFreeReg().
+            assert(nextRegRec->assignedInterval == regRec->assignedInterval);
```
CarolEidt: I don't believe that this is a correct assert. One or both of the two registers may have been allocated to a `TYP_FLOAT` interval.

src/jit/lsra.cpp
```diff
@@ -6709,6 +6785,13 @@ void LinearScan::processBlockStartLocations(BasicBlock* currentBlock, bool alloc
                     }
                     inVarToRegMap[assignedInterval->getVarIndex(compiler)] = REG_STK;
                 }
+#ifdef _TARGET_ARM_
+                // Consider overlapping floating point register for TYP_DOUBLE
+                else if (!assignedInterval->isConstant && isOverlappedRegRecord(assignedInterval, physRegRecord))
```
CarolEidt: >isOverlappedRegRecord(assignedInterval, physRegRecord) [](start = 58, length = 54)

It appears that this will return true for the case where `assignedInterval` is `TYP_FLOAT` but its `assignedReg` is not `physRegRecord` (i.e. the case below). In this case this assert will fire incorrectly.

src/jit/lsra.cpp
```diff
@@ -6836,7 +6919,20 @@ void LinearScan::freeRegister(RegRecord* physRegRecord)
             // we wouldn't unnecessarily link separate live ranges to the same register.
             if (nextRefPosition == nullptr || RefTypeIsDef(nextRefPosition->refType))
             {
+#ifdef _TARGET_ARM_
+                if (assignedInterval->registerType == TYP_DOUBLE && !genIsValidDoubleReg(physRegRecord->regNum))
+                {
+                    // Don't call unassignPhysReg() again, because two FP registers of a double reister are already
+                    // unassigned together when this function is called with valid double register number.
+                    assert(assignedInterval->assignedReg != physRegRecord);
```
CarolEidt: >assert(assignedInterval->assignedReg != physRegRecord); [](start = 20, length = 55)

I believe that this assert should be stronger. In fact, I think that it is problematic if this method is ever called with the other half of a `TYP_DOUBLE` register when the interval being ejected is itself `TYP_DOUBLE`. That is, such registers should always be freed together (and only the half that is the valid double register should be added to the set of registers to free).

src/jit/lsra.cpp
```diff
@@ -6836,7 +6919,20 @@ void LinearScan::freeRegister(RegRecord* physRegRecord)
...
+                }
+                else
+                {
+                    unassignPhysReg(physRegRecord, nullptr);
+                }
+#else
```
CarolEidt: In my view, it would be better to just put the "then" clause under the `#ifdef _TARGET_ARM_` (through the `else' line). This makes it clearer that the non-double case is common with the handling for other targets.

src/jit/lsra.cpp
```diff
@@ -7551,6 +7660,15 @@ void LinearScan::allocateRegisters()
...
+#ifdef _TARGET_ARM_
+            // Update overlapping floating point register for TYP_DOUBLE
+            if (currentInterval->registerType == TYP_DOUBLE)
+            {
+                assert(isFloatRegType(currentInterval->assignedReg->registerType));
+                regsToFree &= ~genRegMask(REG_NEXT(assignedRegister));
+            }
+#endif
+
             // If this interval is dead, free the register.
```
CarolEidt: If you kept only the valid half of a `TYP_DOUBLE` register in the sets of registers to free, then this would not be required.

src/jit/lsra.cpp
```diff
@@ -7564,11 +7682,29 @@ void LinearScan::allocateRegisters()
                     if (currentRefPosition->delayRegFree)
                     {
                         delayRegsToFree |= assignedRegBit;
+
+#ifdef _TARGET_ARM_
+                        // Update overlapping floating point register for TYP_DOUBLE
+                        if (currentInterval->registerType == TYP_DOUBLE)
+                        {
+                            assert(isFloatRegType(currentInterval->assignedReg->registerType));
+                            delayRegsToFree |= genRegMask(REG_NEXT(assignedRegister));
```
CarolEidt: Same comment as above.

src/jit/lsra.cpp
```diff
@@ -7564,11 +7682,29 @@ void LinearScan::allocateRegisters()
...
                     else
                     {
                         regsToFree |= assignedRegBit;
+
+#ifdef _TARGET_ARM_
+                        // Update overlapping floating point register for TYP_DOUBLE
+                        if (currentInterval->registerType == TYP_DOUBLE)
+                        {
+                            assert(isFloatRegType(currentInterval->assignedReg->registerType));
+                            regsToFree |= genRegMask(REG_NEXT(assignedRegister));
```
CarolEidt: And same here

CarolEidt:
:clock1:

src/jit/lsra.cpp
hqueue: Thank you for feedback :)

The assumption in this PR is as follows.
For TYP_DOUBLE interval, we keep following mappings between a TYP_DOUBLE interval and two RegRecords which 
consist a double register.

For example, if we have assigned `d0` register (which is actually `f0` and `f1` in this LSRA ) to a interval `Interval_1` which is TYP_DOUBLE`, we keep following mappings.
```
When Interval_1->registerType == TYP_DOUBLE

// When a TYP_DOUBLE register is assigned to a TYP_DOUBLE interval
Interval_1->assginedReg = RegRecord_f0       // from interval to regrecord
RegRecord_f0->assignedInterval = Interval_1  // from regrecord to interval
RegRecord_f1->assignedInterval = Interval_1  // from regrecord to interval

// When double register is freed later
Interval_1->assignedReg = NULL
RegRecord_f0->previousInterval = Interval_1
RegRecord_f1->previousInterval = Interval_1
```

> I don't believe that this is a correct assert. One or both of the two registers may have been allocated to a TYP_FLOAT interval.

With above assumption, I think we can assume that two (single float) registers have been already assigned to a same TYP_DOUBLE interval, because  this assertion is inside `if (regRec->assignedInterval->registerType == TYP_DOUBLE)`.

And for the case where one or both register have been allocated to `TYP_FLOAT` interval after being freed for `TYP_DOUBLE` interval, both `regRec` and `nextRegRec` should have null `previousInterval` when the code reaches here. I think I've find and update the code where `previousInterval` is partially set. (Can you suggest where?)
 
What do you think of it?

src/jit/lsra.cpp
hqueue: >  where `assignedInterval` is `TYP_FLOAT`  but its `assignedReg` is not `physRegRecord`

If `assignedInterval` is `TYP_FLOAT`, then `isOverlappedRegRecord()` returns true only when `assignedReg` is `physRegRecord`, because there is a comparion at line 6556 which checks `assignedInterva` is `TYP_DOUBLE`.
```
6556 +    else if (assignedInterval->registerType == TYP_DOUBLE)
```

If you have meant "but its`assignedReg` is ~~not~~ `physRegRecord`", then you are definitely right.
However I think the first comparison in the if-statement at line 6779 already met and such case may not reach here.
```
6779    if (!assignedInterval->isConstant && assignedInterval->assignedReg == physRegRecord)
```
Are you suggesting narrowing down cases where `isOverlappedRegRecord()` returns `true` to avoid ambiguity ?
If so, I will rename and update the function, because the function is only used here right now.

src/jit/lsra.cpp
hqueue: You are right and I will add stronger assertion here to check whether the first half of TYP_DOUBLE register.

src/jit/lsra.cpp
hqueue: Thanks!

src/jit/lsra.cpp
hqueue: I've also came to the similar conclusion after implementing `freeRegister()` to handle TYP_DOUBLE correctly, but was not sure.

Thank you for the confirmation :)

src/jit/lsra.cpp
hqueue: Thanks!

src/jit/lsra.cpp
hqueue: Thanks!

src/jit/lsra.cpp
hqueue: Or do I misunderstand your feedback?

src/jit/lsra.cpp
```diff
@@ -6502,6 +6534,49 @@ regNumber LinearScan::rotateBlockStartLocation(Interval* interval, regNumber tar
...
+    {
+        return true;
+    }
+    else if (assignedInterval->registerType == TYP_DOUBLE)
+    {
+        // Check whether physRegRecord is overlapped by assignedReg
+        RegRecord* altPhysRegRecord;
+        regNumber  regNum = physRegRecord->regNum;
+
+        if (genIsValidFloatReg(regNum))
```
CarolEidt: In what case would this NOT be a valid float reg?

src/jit/lsra.cpp
CarolEidt: You are right - I meant "its `assignedReg` IS `physRegRecord`". I think it would be good to rename and narrow the focus of the method.

---
In reply to: [112925890](https://github.com/dotnet/coreclr/pull/10972#discussion_r112925890) [](ancestors = 112925890)

src/jit/lsra.cpp
CarolEidt: What about the case where interval0 in d0(f0/f1) becomes inactive, and f0 is assigned to a `TYP_FLOAT` (say interval1). At this point both `previousInterval` fields reference interval0. Then interval1 becomes inactive and is reallocated to yet another `TYP_FLOAT`. At that point, its `previousInterval` will be set to interval1, but the `previousInterval` of f1 is still interval0. Have I missed something that will handle this case?

---
In reply to: [112926942](https://github.com/dotnet/coreclr/pull/10972#discussion_r112926942) [](ancestors = 112926942)

pgavlin:
I'm a bit nervous about the implementation approach taken by this PR.

If I'm reading it correctly, the implementation assigns double-typed intervals to even-numbered single-precision registers  (e.g. F0, F2, etc.) and then kills/frees the other half of the corresponding double register (e.g. F1 for F0, F3 for F2, etc.) appropriately. This seems like a poor-man's version of a more general solution for overlapping registers.

I can imagine that in the general case we might create a reg record per assignable register (e.g. a reg record for each of `F0`, `F1`, and `D0` on ARM32 or a reg record for each of `rax`, `eax`, `ax`, `ah`, and `al` on AMD64) with some way to discover which other reg records interfere (perhaps a bitmask). We would then use the register type information to choose an appropriate register to allocate to an interval and the interference information to mark registers as free or busy. It is possible that this approach might also ease some of the complications observed in this PR (e.g. assignedInterval/previousInterval bookkeeping).

CarolEidt:
@pgavlin - this is the implementation approach that was taken in the original arm32 register allocator (i.e. not strictly an approach taken by this PR).
I don't think it is useful to implement a complex solution for overlapping registers when:
* we have no plans (as far as I know) to use the high regs (e.g. `ah`) on xarch, and
* such a solution is likely to add more complexity than the simple aligned aliasing required for arm32 (i.e. double registers always have an even/odd pair).

Note that the bookkeeping you allude to would still be required, and would be unlikely (IMO) to be much different from what we see here - though one might argue that it could be extracted/abstracted somewhat.

pgavlin:
> this is the implementation approach that was taken in the original arm32 register allocator (i.e. not strictly an approach taken by this PR).

Sure; I did not mean to imply that this approach was unique.

> such a solution is likely to add more complexity than the simple aligned aliasing required for arm32 (i.e. double registers always have an even/odd pair).

I actually believe that it would do the opposite: though the data structures might become slightly more complex, I believe that the code would be more regular and easier to understand due to the formalization of the reasoning about overlap.

> Note that the bookkeeping you allude to would still be required, and would be unlikely (IMO) to be much different from what we see here

I think that a solution that uses independent register records to represent different register sizes that may alias other registers would at least prevent the issues with `previousInterval`.

hqueue:
> If I'm reading it correctly, the implementation assigns double-typed intervals to even-numbered single-precision registers (e.g. F0, F2, etc.) and then kills/frees the other half of the corresponding double register (e.g. F1 for F0, F3 for F2, etc.) appropriately. This seems like a poor-man's version of a more general solution for overlapping registers.

@pgavlin I also agree that it's not a neat approach. I was just trying to minimize changes by extending existing approach at  https://github.com/dotnet/coreclr/blob/master/src/jit/lsra.cpp#L6039-L6047 and https://github.com/dotnet/coreclr/blob/master/src/jit/lsra.cpp#L6266-L6273 .
It seems that it was a original arm32 implementation as described by @CarolEidt.

> I actually believe that it would do the opposite: though the data structures might become slightly more complex, I believe that the code would be more regular and easier to understand due to the formalization of the reasoning about overlap.

In both approach, i.e. using separate RegRecord for double reigster and using current apporach,  we always have to maintain and check interference between two kind of registers (single-float reg and double-float reg). Therefore I prefer @CarolEidt's opinion when considering the amount of code changes required.
Ideally it would be better to refactor or introduce new data structures eventually to LSRA. But for now, I'm afraid of changing a lot of code in LSRA this time.

hqueue:
@CarolEidt If you don't mind, I will update this PR with your feedback and with additional implementation, because I found there are some assertion failures remaining in `JIT/Methodical` tests.

src/jit/lsra.cpp
hqueue: You are right :)
How about replacing this `if` with `assert` ?

CarolEidt: Yes, I think that's the right answer.

CarolEidt:
> If you don't mind, I will update this PR with your feedback and with additional implementation, because I found there are some assertion failures remaining in JIT/Methodical tests.

Sounds good. Thanks.

pgavlin:
>Ideally it would be better to refactor or introduce new data structures eventually to LSRA to handle overlapping registers. But for now, I'm afraid of changing a lot of code in LSRA this time.

Okay. Can you file a work item to track cleaning this up in the future?

hqueue:
@CarolEidt I've update the PR with your feedback and bugfix for remaining assertion failure.
While implementing, I could understand your feedback which I was not sure.
There is no more question remaining regarding your comment by now.
I tried to follow your suggestions but it was almost rewritten with refactoring.

Before below test was failing with previous PR but it's also fixed by now. And no regression found in `JIT/CodeGenBringUpTests/`.

```
FAILED   - [   5][  38s]JIT/Methodical/fp/exgen/10w250d_cs_do/10w250d_cs_do.sh
               BEGIN EXECUTION
               /mnt/albireo/mywork/device/rpi3/unittests/coreoverlays/coreoverlay.test20170414.lsra_update1/corerun 10w250d_cs_do.exe

               Assert failure(PID 8720 [0x00002210], Thread: 8720 [0x2210]): Assertion failed 'assignedInterval != nullptr' in 'testout1:Func_0_1_1_1_1_1_1_1_1_1_1_1_1_1_1_1():double' (IL size 282)

                   File: /opt/code/github/hqueue/coreclr/src/jit/lsra.cpp Line: 6235
                   Image: /mnt/albireo/mywork/device/rpi3/unittests/coreoverlays/coreoverlay.test20170414.lsra_update1/corerun

               ./10w250d_cs_do.sh: line 132:  8720 Aborted                 (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" 10w250d_cs_do.exe $CLRTestExecutionArguments
               Expected: 100
               Actual: 134
               END EXECUTION - FAILED
```

> Okay. Can you file a work item to track cleaning this up in the future?
@pgavlin I've create issue at https://github.com/dotnet/coreclr/issues/11286

hqueue:
> Okay. Can you file a work item to track cleaning this up in the future?

@pgavlin I've create issue at https://github.com/dotnet/coreclr/issues/11286

hqueue:
@CarolEidt To help review, one of big change is commit https://github.com/dotnet/coreclr/pull/10972/commits/a4566a5800c4f141d4a0a00b311b66e5872e68c6.

* Make checkAndClearInterval() and checkAndAssignInterval() to be TYP_DOUBLE safe for ARM32
* Update assignPhysReg() and unassignPhysReg() to be remained TYP_DOUBLE safe for ARM32

Before there was a problematic situation when checkAndAssignInterval()(which is TYP_DOUBLE unsafe)
calls unassignPhysReg()(which is TYP_DOUBLE safe). Therefore checkAndAssignInterval() is invoked for f1 and f2 accidently.

src/jit/lsra.cpp
```diff
@@ -6709,6 +6809,13 @@ void LinearScan::processBlockStartLocations(BasicBlock* currentBlock, bool alloc
                     }
                     inVarToRegMap[assignedInterval->getVarIndex(compiler)] = REG_STK;
                 }
+#ifdef _TARGET_ARM_
+                // Consider overlapping floating point register for TYP_DOUBLE
+                else if (!assignedInterval->isConstant && isSecondHalfReg(physRegRecord, assignedInterval))
+                {
+                    assert(assignedInterval->assignedReg->regNum == REG_PREV(physRegRecord->regNum));
```
CarolEidt: This assert basically duplicates what is already checked by `isSecondHalfReg`, but doesn't check for the case where, somehow, `assignedInterval` is a `TYP_DOUBLE` for which the second half has not been appropriately updated. I think a better check would be:
```
else if (!assignedInterval->isConstant && assignedInterval->registerType == TYP_DOUBLE)
{
    assert(!assignedInterval->isActive || isSecondHalfReg(physRegRecord, assignedInterval));
}
```
Does that make sense?

src/jit/lsra.cpp
```diff
@@ -6502,6 +6562,45 @@ regNumber LinearScan::rotateBlockStartLocation(Interval* interval, regNumber tar
...
+//--------------------------------------------------------------------------------------
+// isSecondHalfReg: Test if recRec is second half of double reigster
+//                  which is assigned to an interval.
+//
+// Arguments:
+//    regRec - a register to be tested
+//    interval - an interval which is assigned to some register
+//
+// Assumptions:
+//    The interval must have been assigned to some register
```
CarolEidt: This assumption does not appear to be correct. If the interval is not assigned to a register, it will simply return false.

src/jit/lsra.cpp
```diff
@@ -6836,7 +6943,23 @@ void LinearScan::freeRegister(RegRecord* physRegRecord)
             // we wouldn't unnecessarily link separate live ranges to the same register.
             if (nextRefPosition == nullptr || RefTypeIsDef(nextRefPosition->refType))
             {
-                unassignPhysReg(physRegRecord, nullptr);
+#ifdef _TARGET_ARM_
+                if (assignedInterval->registerType == TYP_DOUBLE && !genIsValidDoubleReg(physRegRecord->regNum))
```
CarolEidt: I think that, when a register that is assigned to a `TYP_DOUBLE` `Interval` is freed, only the valid (lower) register should be added to the free set. Then this can be an assert, i.e.:
```
assert((assignedInterval->registerType != TYP_DOUBLE) || genIsValidDoubleReg(physRegRecord->regNum));
```

src/jit/lsra.cpp
```diff
@@ -5604,6 +5604,20 @@ regNumber LinearScan::tryAllocateFreeReg(Interval* currentInterval, RefPosition*
             else if ((bestScore & UNASSIGNED) != 0 && intervalToUnassign != nullptr)
             {
                 availablePhysRegInterval->previousInterval = intervalToUnassign;
+#ifdef _TARGET_ARM_
+                // TODO-ARM-Throughput: For ARM,  we don't have to keep a same previous interval in two RegRecords,
+                // if we can use a interval from one RegRecord. But I'm not sure yet.
+                // Please take a look at unassignPhysReg().
```
CarolEidt: One more suggestion: You might update this comment to add that it should not be necessary because we will always manage the register assignment of `TYP_DOUBLE` intervals together.

src/jit/lsra.cpp
```diff
@@ -6263,11 +6293,17 @@ void LinearScan::unassignPhysReg(RegRecord* regRec, RefPosition* spillRefPositio
...
+    regNumber  nextRegNum = REG_NA;
+    RegRecord* nextRegRec = nullptr;
+
+    // Prepare second half RegRecord of a double register for TYP_DOUBLE
+    if (assignedInterval->registerType == TYP_DOUBLE)
     {
-        regNumber  nextRegNum = REG_NEXT(regRec->regNum);
-        RegRecord* nextRegRec = getRegisterRecord(nextRegNum);
-        checkAndClearInterval(nextRegRec, spillRefPosition);
+        assert(isFloatRegType(regRec->registerType));
```
hqueue: I will add assertion here.

src/jit/lsra.cpp
hqueue: You are right :)

src/jit/lsra.cpp
hqueue: Thanks! I was also having hard time with this condition.
Your suggestion is much clear and make sense :)

src/jit/lsra.cpp
hqueue: Right! With the invariant we can just assert this one.

src/jit/lsra.cpp
hqueue: I will update this too.

hqueue:
@CarolEidt I've applied your feedback in a commit https://github.com/dotnet/coreclr/pull/10972/commits/849dd6e2a701f1d46738e95a6fc1ba77741151ea. Please take a look.

CarolEidt:
@hqueue - thanks for making the changes, and sorry for the delay in reviewing. This looks good.

PR #10975, opened 2017-04-14
Bring back typelib related attributes

yizhang82:
Not strictly required for net standard 2.0 but nice to have for tlbimp created dll to not have dangling references. 

@jkotas PTAL

### DIFF
--- a/src/mscorlib/shared/System/Runtime/InteropServices/TypeLibAttributes.cs
+++ b/src/mscorlib/shared/System/Runtime/InteropServices/TypeLibAttributes.cs
@@ -6,6 +6,17 @@ using System;
 
 namespace System.Runtime.InteropServices
 {
+    [AttributeUsage(AttributeTargets.Interface, Inherited = false)]
+    public sealed class TypeLibImportClassAttribute : Attribute
+    {
+        internal String _importClassName;
+        public TypeLibImportClassAttribute(Type importClass)
+        {
+            _importClassName = importClass.ToString();
+        }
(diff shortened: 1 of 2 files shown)


jkotas:
Why not to add them to corefx instead?

yizhang82:
@jkotas I didn't consider it but that's a good point. It's unlikely we'll ever poke at them in the runtime. let me look into adding them into S.P.InteropServices. Thanks!

hseok-oh:

src/mscorlib/shared/System/Runtime/InteropServices/TypeLibAttributes.cs
```diff
@@ -0,0 +1,166 @@
...
+        internal TypeLibFuncFlags _val;
+        public TypeLibFuncAttribute(TypeLibFuncFlags flags)
+        {
+            _val = flags;
+        }
+        public TypeLibFuncAttribute(short flags)
+        {
+            _val = (TypeLibFuncFlags)flags;
+        }
+        public TypeLibFuncFlags Value { get {return _val;} }
```
danmosemsft: tiny nit -- space before `return`

yizhang82: Hmm.. desktop code was like this (this is a simply a copy/paste). Fixed.

yizhang82:
I'll open a PR on corefx side.

PR #10983, opened 2017-04-14
Revert "Strip out unused reg code. (#10741)"

danmosemsft:
This reverts commit ed4f594abf41a71b126152bb8755051d0831e12d (https://github.com/dotnet/coreclr/pull/10741) which @adiaaida has determined via a process of elimination to be the cause of the broken performance runs.

It's not immediately obvious to us why it broke them because we don't know what registry keys it's expecting to read. @adiaaida can provide a repro case to @JeremyKuhne so he can break down his change and figure out the problem. That's probably quicker than debugging all the perf infrastructure.

meantime we want the runs to work.

### DIFF
--- a/src/mscorlib/System.Private.CoreLib.csproj
+++ b/src/mscorlib/System.Private.CoreLib.csproj
@@ -86,9 +86,11 @@
   <PropertyGroup Condition="'$(OsEnvironment)' == 'Unix'">
     <DebugType>portable</DebugType>
   </PropertyGroup>
+
   <PropertyGroup Condition="'$(TargetsOSX)' == 'true'">
     <DefineConstants>PLATFORM_OSX;$(DefineConstants)</DefineConstants>
   </PropertyGroup>
+
   <!-- Assembly attributes -->
   <PropertyGroup>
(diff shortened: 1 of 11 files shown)


danmosemsft:
I undersatnd from @valenis they are looking into adding a simple perf test into the CI system so a total break like this could not get through.

adiaaida:
@dotnet-bot test ci please

adiaaida:
@dotnet-bot test windows_nt x64 perf please

danmosemsft:

@adiaaida @valenis @brianrob  is the perf tests failure unrelated? Or is it the symptom we're trying to fix?

adiaaida:
@danmosemsft Yes. The perf failure is because there are quotes in the title of your PR. I'll look into a fix to harden against that.

I've started this job manually to confirm the changes fix our issue: https://ci2.dot.net/job/dotnet_coreclr/job/perf/job/master/job/perf_perflab_Windows_NT_x64_prtest/35/

danmosemsft:
@adiaaida thanks. If the kind good please hit merge if you get a chance before I do. The arm failure is unrelated and it makes sense to ignite it given this is fixing a break.

PR #10988, opened 2017-04-15
Move Span/ReadOnlySpan to shared CoreLib partition

jkotas:
Fix a few method names to better names used in CoreRT

Contributes to https://github.com/dotnet/corert/issues/2966


src/mscorlib/src/System/Runtime/CompilerServices/jithelpers.cs
```diff
@@ -216,14 +216,7 @@ static internal PinningHelper GetPinningHelper(Object o)
         private extern static bool IsAddressInStack(IntPtr ptr);
 #endif
 
-        static internal bool ByRefLessThan<T>(ref T refA, ref T refB)
```
ahsonkhan: What is the rationale for removing this?

src/vm/mscorlib.h
```diff
@@ -828,9 +827,6 @@ DEFINE_METHOD(INTERLOCKED,          COMPARE_EXCHANGE_OBJECT,CompareExchange, SM_
 DEFINE_CLASS(PINNING_HELPER,        CompilerServices,       PinningHelper)
 DEFINE_FIELD(PINNING_HELPER,        M_DATA,                 m_data)
 
-DEFINE_CLASS(ARRAY_PINNING_HELPER,  CompilerServices,       ArrayPinningHelper)
-DEFINE_FIELD(ARRAY_PINNING_HELPER,  M_ARRAY_DATA,           m_arrayData)
```
ahsonkhan: Removing this is causing CI errors. Also it is used [here](https://github.com/dotnet/coreclr/pull/10988/files#diff-bbced44286fd2ad6bcf70fbd3c276d57R6926).

d:\j\workspace\debug_windows---17180f1b\src\vm\jitinterface.cpp(6926): error C2065: 'FIELD__ARRAY_PINNING_HELPER__M_ARRAY_DATA': undeclared identifier [D:\j\workspace\debug_windows---17180f1b\bin\obj\Windows_NT.x64.Debug\src\vm\crossgen\cee_crossgen.vcxproj]

src/mscorlib/shared/System/Span.cs
```diff
@@ -508,12 +507,9 @@ public static ReadOnlySpan<char> AsSpan(this string text)
             if (text == null)
                 ThrowHelper.ThrowArgumentNullException(ExceptionArgument.text);
 
-            return new ReadOnlySpan<char>(ref text.GetFirstCharRef(), text.Length);
+            return new ReadOnlySpan<char>(ref text.GetRawStringData(), text.Length);
         }
-    }
 
-    internal static class SpanHelper
```
ahsonkhan: Should we separate the Span/SpanHelper class into a separate file, esp since it has internal statics used by both Span and ReadOnlySpan?

src/mscorlib/src/System/Runtime/CompilerServices/jithelpers.cs
```diff
@@ -216,14 +216,7 @@ static internal PinningHelper GetPinningHelper(Object o)
...
-        {
-            // The body of this function will be replaced by the EE with unsafe code!!!
-            // See getILIntrinsicImplementation for how this happens.
-            throw new InvalidOperationException();
-        }
-
-        static internal ref T GetArrayData<T>(T[] array)
+        static internal ref byte GetRawSzArrayData(this Array array)
```
ahsonkhan: METHOD__JIT_HELPERS__GET_ARRAY_DATA and 
METHOD__JIT_HELPERS__GET_RAW_SZ_ARRAY_DATA case both point to the same [code](https://github.com/dotnet/coreclr/blob/master/src/vm/jitinterface.cpp#L6935-L6954) and yet one returns a ref T while the other returns a ref byte? How does that work?

src/mscorlib/shared/System/ReadOnlySpan.cs
```diff
@@ -57,7 +57,7 @@ public ReadOnlySpan(T[] array, int start)
             if ((uint)start > (uint)array.Length)
                 ThrowHelper.ThrowArgumentOutOfRangeException();
 
-            _pointer = new ByReference<T>(ref Unsafe.Add(ref JitHelpers.GetArrayData(array), start));
+            _pointer = new ByReference<T>(ref Unsafe.Add(ref Unsafe.As<byte, T>(ref array.GetRawSzArrayData()), start));
```
ahsonkhan: Does having to pass `ref array.GetRawSzArrayData()` to `Unsafe.As<byte, T>` incur significant performance cost?

src/mscorlib/src/System/Runtime/CompilerServices/jithelpers.cs
jkotas: Not used anywhere.

src/mscorlib/src/System/Runtime/CompilerServices/jithelpers.cs
jkotas: The IL is same for both cases because of it depends on unsafe casts.

src/mscorlib/shared/System/Span.cs
jkotas: Agree.

src/mscorlib/shared/System/ReadOnlySpan.cs
jkotas: The JITed code is the same in both cases.

src/vm/mscorlib.h
jkotas: Accidental change ... fixed.

ahsonkhan:

ahsonkhan:

ahsonkhan:
LGTM

PR #10989, opened 2017-04-15
small clean-up and bug fix in genIntToIntCast

sandreenko:
Some commits to make code clearer and make DevDiv issue fix easier.

No x64 asm diff expected.
Fix DevDiv 406160.

### DIFF
--- a/src/jit/codegenxarch.cpp
+++ b/src/jit/codegenxarch.cpp
@@ -6482,10 +6482,12 @@ void CodeGen::genIntToIntCast(GenTreePtr treeNode)
     assert(genIsValidIntReg(targetReg));
     assert(genIsValidIntReg(sourceReg));
 
-    instruction ins  = INS_invalid;
-    emitAttr    size = EA_UNKNOWN;
+    instruction ins     = INS_invalid;
+    emitAttr    srcSize = EA_ATTR(genTypeSize(srcType));
+    emitAttr    dstSize = EA_ATTR(genTypeSize(dstType));
+    emitAttr    size    = EA_UNKNOWN;
 
-    if (genTypeSize(srcType) < genTypeSize(dstType))
+    if (srcSize < dstSize)
     {
         // Widening cast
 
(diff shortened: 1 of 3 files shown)


src/jit/codegenxarch.cpp
```diff
@@ -6674,8 +6670,6 @@ void CodeGen::genIntToIntCast(GenTreePtr treeNode)
     }
     else // non-overflow checking cast
     {
-        noway_assert(size < EA_PTRSIZE || srcType == dstType);
```
sandreenko: There were two cases how we could come here:
1.	if we came here from widening case, then size was srcSize, for amd64 it could not be greater than 4 for widening. x86 widening to long is the special case that was proceeded in the genLongToIntCast. So, for x86 it was unreachable and didn't check anything useful for amd64.	
2.	If we came here from narrowing, it means that size was dstSize.
a.	If it was cast to smaller type, then dstSize was less srcSize, when srcSize was less or equal EA_PTRSIZE. So, it was true and didn’t check anything useful;
b.	if it was cast with srcSize == dstSize, then it was true for amd64 because we do not perform ulong<->long casts without check. **For x86 there was an error** **because bool -> int cast was treated as narrowing cast (srcSize is size of actual type, that was uint) without check.**

hseok-oh:

sandreenko:
No spmi diffs now.

sandreenko:
So the error happened when after importer we had tree like:

```
 [000087] ---X-----U--             |     |     |  /--*  cast_ovfl long <- ulong <- uint
 [000086] ------------             |     |     |  |  \--*  lclVar    short  V00 arg0
```
that was transformed into 
`t87 = *  cast      int <- uint REG edx`
before the code generation phase,
However, I was not able to reproduce this behavior in CoreCLR.

sandreenko:

sandreenko:
PTAL @briansull  @dotnet/jit-contrib.

src/jit/codegenxarch.cpp
briansull: Are we changing the logic for both x64 and x86 with this PR?

sandreenko: We delete this assert for both platforms. The logic was not changed, I replaced size variable with exact using srcSize or dstSize, because it was hard to prove that this assert could be deleted.
So there are no any superpmi asm diffs for x64.

briansull:
LGTM

Issue #10980, opened 2017-04-14
Jit: improve codegen for impRuntimeLookup


AndyAyersMS:
`impRuntimeLookup` generates a conditional-deref tree:
```
[000077] ------------             *  stmtExpr  void  (IL   ???...  ???)
[000069] ------------ then        |        /--*  lclVar    long   V09 tmp2         
[000073] ------------             |     /--*  colon     long  
[000072] ------------ else        |     |  \--*  indir     long  
[000070] ------------             |     |     |  /--*  const     long   -1
[000071] ------------             |     |     \--*  +         long  
[000062] ------------             |     |        \--*  lclVar    long   V09 tmp2         
[000074] ------------             |  /--*  qmark     long  
[000067] ------------             |  |  |  /--*  const     int    0
[000068] Q-----------    if       |  |  \--*  ==        int   
[000065] ------------             |  |     |  /--*  const     int    1
[000066] ------------             |  |     \--*  &         int   
[000064] ------------             |  |        \--*  cast      int <- long
[000063] ------------             |  |           \--*  lclVar    long   V09 tmp2         
[000076] -A----------             \--*  =         long  
[000075] D------N----                \--*  lclVar    long   V10 tmp3         
```
for x64 this turns into the following native code:
```asm
       41F7C001000000       test     r8d, 1
       7502                 jne      SHORT G_M28280_IG05
       EB07                 jmp      SHORT G_M28280_IG06

G_M28280_IG05:
       4D8B80FFFFFFFF       mov      r8, qword ptr [r8+FFFFFFFFH]

G_M28280_IG06:
```
we ought to be able to produce something more compact like
```asm
       41F7C001000000       test     r8d, 1
       7502                 je       SHORT G_M28280_IG06
       4D8B80FFFFFFFF       mov      r8, qword ptr [r8+FFFFFFFFH]

G_M28280_IG06:
```
and maybe find smaller encodings for the immediate values.

mikedn:
Let's have a self contained repro to stare at:
```C#
private static Type Test<T>() => typeof(T);
```
generates
```asm
G_M57965_IG01:
       4883EC28             sub      rsp, 40
       48894C2420           mov      qword ptr [rsp+20H], rcx
G_M57965_IG02:
       488B4938             mov      rcx, qword ptr [rcx+56]
       488B09               mov      rcx, qword ptr [rcx]
       F7C101000000         test     ecx, 1
       7502                 jne      SHORT G_M57965_IG03
[...]

PR #10990, opened 2017-04-15
Narrow TEST instructions when possible

mikedn:
Contributes to #10980, #7641

### DIFF
--- a/src/jit/codegenxarch.cpp
+++ b/src/jit/codegenxarch.cpp
@@ -6222,10 +6222,27 @@ void CodeGen::genCompareInt(GenTreePtr treeNode)
     assert(!varTypeIsFloating(op2Type));
 
     instruction ins;
+    var_types   type = TYP_UNKNOWN;
 
     if (tree->OperIs(GT_TEST_EQ, GT_TEST_NE))
     {
         ins = INS_test;
+
+        // Unlike many xarch instructions TEST doesn't have a form with a 16/32/64 bit first operand and
+        // an 8 bit immediate second operand. But if the immediate value fits in 8 bits then we can simply
+        // emit a 8 bit TEST instruction, unless we're targeting x86 and the first operand is a non-byteable
+        // register.
+        // Note that lowering does something similar but its main purpose is to allow memory operands to be
+        // contained so it doesn't handle other kind of operands. It could do more but on x86 that results
+        // in additional register constrains and that may be worse than wasting 3 bytes on an immediate.
+        if (
+#ifdef _TARGET_X86_
+            (!op1->isUsedFromReg() || isByteReg(op1->gtRegNum)) &&
+#endif
+            (op2->IsCnsIntOrI() && genTypeCanRepresentValue(TYP_UBYTE, op2->AsIntCon()->IconValue())))
+        {
+            type = TYP_UBYTE;
+        }
     }
     else if (op1->isUsedFromReg() && op2->IsIntegralConst(0))
     {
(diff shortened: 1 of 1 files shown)


mikedn:
All diffs are like:
* `test ecx, 8` -> `test cl, 8`
* `test dword ptr [ebp+14H], 6` -> `test byte ptr [ebp+14H], 6`
* `jne G_M18532_IG10` -> `jne SHORT G_M18532_IG10`

Diff summary:
```
X64:

Total bytes of diff: -6342 (-0.05 % of base)
    diff is an improvement.
       -3497 : System.Private.CoreLib.dasm (-0.10 % of base)
        -801 : Microsoft.CodeAnalysis.VisualBasic.dasm (-0.06 % of base)
        -400 : Microsoft.CodeAnalysis.CSharp.dasm (-0.03 % of base)
        -347 : System.Linq.Parallel.dasm (-0.06 % of base)
        -216 : System.Private.Uri.dasm (-0.25 % of base)
32 total files with size differences (32 improved, 0 regressed).
        -148 : System.Private.CoreLib.dasm - Dictionary`2:System.Collections.IDictionary.set_Item(ref,ref):this (29 methods)
        -148 : System.Private.CoreLib.dasm - Dictionary`2:System.Collections.IDictionary.Add(ref,ref):this (29 methods)
         -96 : System.Private.CoreLib.dasm - GenericEqualityComparer`1:LastIndexOf(ref,struct,int,int):int:this (19 methods)
         -96 : System.Private.CoreLib.dasm - GenericEqualityComparer`1:IndexOf(ref,struct,int,int):int:this (19 methods)
         -90 : System.Private.CoreLib.dasm - CLRIPropertyValueImpl:CoerceScalarValue(int,ref):ref
862 total methods with size differences (862 improved, 0 regressed).

X86:

Total bytes of diff: -5786 (-0.06 % of base)
    diff is an improvement.
       -3078 : System.Private.CoreLib.dasm (-0.11 % of base)
        -679 : Microsoft.CodeAnalysis.VisualBasic.dasm (-0.06 % of base)
        -377 : System.Linq.Parallel.dasm (-0.08 % of base)
        -367 : Microsoft.CodeAnalysis.CSharp.dasm (-0.03 % of base)
        -324 : System.Private.Uri.dasm (-0.41 % of base)
32 total files with size differences (32 improved, 0 regressed).
        -132 : System.Private.CoreLib.dasm - Dictionary`2:System.Collections.IDictionary.set_Item(ref,ref):this (29 methods)
        -132 : System.Private.CoreLib.dasm - Dictionary`2:System.Collections.IDictionary.Add(ref,ref):this (29 methods)
         -94 : System.Private.CoreLib.dasm - ValueTuple`8:System.Runtime.CompilerServices.ITuple.get_Item(int):ref:this (15 methods)
         -88 : System.Private.CoreLib.dasm - ValueTuple`8:System.Collections.IStructuralEquatable.Equals(ref,ref):bool:this (15 methods)
         -87 : System.Private.CoreLib.dasm - CLRIPropertyValueImpl:CoerceScalarValue(int,ref):ref
792 total methods with size differences (792 improved, 0 regressed).
```

hseok-oh:

BruceForstall:
@dotnet/jit-contrib

omariom:
@mikedn 

`test dword ptr [ebp+14H], 6 -> test byte ptr [ebp+14H], 6`

Can it accidentally lead to [store-to-load forwarding stall](https://blogs.msdn.microsoft.com/oldnewthing/20170428-00/?p=96065).

mikedn:
> Can it accidentally lead to store-to-load forwarding stall.

AFAIK no, at least on current Intel processors. Note that the address doesn't change like it does in the example you link to. Only the load size is smaller so this should be fine.

omariom:
As I understand  it, stalls happen when the read is wider than a previous write to the same location.
As you are narrowing reads it should not happen.

russellhadley:
@erozenfeld Given that this change for just test is seeing .5% win on code size, we should see what the opportunity is for pushing converts up the expression trees.  This transform is a special case where we can retype the immediate which makes narrowing the type on the memory operand legal.  I'm just wondering if applying this in morph to the tree context could give us a bunch of code size and static instruction count reduction wins.

mikedn:
@russellhadley Narrowing earlier might have adverse side effects on x86 because it limits register choices. For example I tried to do this in lowering where we do something similar already and on x86 we sometimes end up with:
```asm
mov eax, esi
test al, 8
```

But more importantly, the JIT isn't well equipped to handle small int trees, it tends to assume that trees are either int or long. Only lowering sometimes generate byte trees.

russellhadley:
@mikedn Agreed, those are issues.   
- The byteable register issue on x86 is a problem but usually manageable by going after memory operands.  Additionally because of the byteable constraints the RA should be aggressively spilling byte register references and folding the memory accesses into the instructions. 
- The issue of not being able to handle trees that are typed as something other than long/int I think will need a hard look.  We end up generating a large number of extensions in the instruction stream so cleaning this up could have a significant impact.   One of my asks of @erozenfeld as he looks at morph is to come up with some guesstimate as to what this change would take engineering wise.

BruceForstall:
ping @dotnet/jit-contrib PTAL

src/jit/codegenxarch.cpp
```diff
@@ -6222,10 +6222,27 @@ void CodeGen::genCompareInt(GenTreePtr treeNode)
...
+        // contained so it doesn't handle other kind of operands. It could do more but on x86 that results
+        // in additional register constrains and that may be worse than wasting 3 bytes on an immediate.
+        if (
+#ifdef _TARGET_X86_
+            (!op1->isUsedFromReg() || isByteReg(op1->gtRegNum)) &&
+#endif
+            (op2->IsCnsIntOrI() && genTypeCanRepresentValue(TYP_UBYTE, op2->AsIntCon()->IconValue())))
+        {
+            type = TYP_UBYTE;
+        }
```
CarolEidt: I feel that there must be something I'm missing here, but it appears that you are comparing only the lower 8-bits of `op1` without any analysis to determine whether that is safe. Can you enlighten me?

mikedn: Note that this applies only to `TEST_EQ/NE` so it's really `(op1 AND op2) EQ/NE 0`. Of course, the result is determined by the bits set in the "mask" (`op2`) and if the mask fits in a byte then we can ignore bits 8-63 of `op1`.

In other words the analysis is there, it's `genTypeCanRepresentValue(TYP_UBYTE...`.

CarolEidt: But since this is actually testing equality it doesn't seem that it will be correct if higher bits of op1 are set

russellhadley: Think of it as testing of a mask.  If the result of the AND with the mask, in this case 0b00000110,  is zero (ZF flag set) then JE (which is an alias of JZ) branches, the NE case is just ZF == 0.    It always takes me a bit to get my head straight on this one.  Whether any bits were set in the upper part of op1 doesn't impact the result since they were going to be AND'd with constant zeros.   (I always have to remember that CMP is SUB and that's why not all CMPs can be made into TESTs)

CarolEidt:
Ah, of course. What I was missing is that we already have the test opcodes, not `GT_EQ` and `GT_NE`. Sorry for the confusion.

src/jit/codegenxarch.cpp
mikedn: It seems that getting rid of `GT_TEST_EQ`/`GT_TEST_NE` and having `GT_TEST`/`GT_JCC`/`GT_SETCC` instead could help clarify things :)

pgavlin: Do we ever issue `test r/m64, imm32` today? That particular encoding sign-extends the immediate and thus could not be replaced with `test r/m8, imm8`.

mikedn: > Do we ever issue test r/m64, imm32 today?

Yes.

> That particular encoding sign-extends the immediate and thus could not be replaced with test r/m8, imm8.

I don't quite follow. This transform is done only if the immediate value is in range 0..255. So if any sign extension is/would be done anywhere it doesn't matter since the value is positive.

pgavlin: > This transform is done only if the immediate value is in range 0..255

Ah, of course--you even pointed that out above. Thanks!

PR #10996, opened 2017-04-15
Do unsigned magic division

### LINKED ISSUE #10970: Missing div => mul by reciprocal case
This case was found in the Kestrel parser code.

For signed div by 10 we are successfully turning it into a imul by reciprocal, but for unsigned we're failing to do the transformation.

Repro: https://gist.github.com/russellhadley/3130b9bfb30567befbc03ae4830f5bfb

mikedn:
Fixes #10970

### DIFF
--- a/src/jit/lower.cpp
+++ b/src/jit/lower.cpp
@@ -4175,91 +4175,6 @@ GenTree* Lowering::LowerUnsignedDivOrMod(GenTreeOp* divMod)
 }
 
 //------------------------------------------------------------------------
-// GetSignedMagicNumberForDivide: Generates a magic number and shift amount for
-// the magic number division optimization.
-//
-// Arguments:
-//    denom - The denominator
-//    shift - Pointer to the shift value to be returned
-//
-// Returns:
-//    The magic number.
-//
-// Notes:
-//    This code is previously from UTC where it notes it was taken from
-//   _The_PowerPC_Compiler_Writer's_Guide_, pages 57-58. The paper is is based on
-//   is "Division by invariant integers using multiplication" by Torbjorn Granlund
-//   and Peter L. Montgomery in PLDI 94
-
-template <typename T>
-T GetSignedMagicNumberForDivide(T denom, int* shift /*out*/)
-{
-    // static SMAG smag;
-    const int bits         = sizeof(T) * 8;
-    const int bits_minus_1 = bits - 1;
-
-    typedef typename jitstd::make_unsigned<T>::type UT;
-
-    const UT two_nminus1 = UT(1) << bits_minus_1;
-
-    int p;
-    UT  absDenom;
-    UT  absNc;
-    UT  delta;
-    UT  q1;
-    UT  r1;
-    UT  r2;
-    UT  q2;
-    UT  t;
-    T   result_magic;
-    int result_shift;
-    int iters = 0;
-
-    absDenom = abs(denom);
-    t        = two_nminus1 + ((unsigned int)denom >> 31);
-    absNc    = t - 1 - (t % absDenom);        // absolute value of nc
-    p        = bits_minus_1;                  // initialize p
-    q1       = two_nminus1 / absNc;           // initialize q1 = 2^p / abs(nc)
-    r1       = two_nminus1 - (q1 * absNc);    // initialize r1 = rem(2^p, abs(nc))
-    q2       = two_nminus1 / absDenom;        // initialize q1 = 2^p / abs(denom)
-    r2       = two_nminus1 - (q2 * absDenom); // initialize r1 = rem(2^p, abs(denom))
-
-    do
-    {
-        iters++;
-        p++;
-        q1 *= 2; // update q1 = 2^p / abs(nc)
-        r1 *= 2; // update r1 = rem(2^p / abs(nc))
-
-        if (r1 >= absNc)
-        { // must be unsigned comparison
-            q1++;
-            r1 -= absNc;
-        }
-
-        q2 *= 2; // update q2 = 2^p / abs(denom)
-        r2 *= 2; // update r2 = rem(2^p / abs(denom))
-
-        if (r2 >= absDenom)
-        { // must be unsigned comparison
-            q2++;
-            r2 -= absDenom;
-        }
-
-        delta = absDenom - r2;
-    } while (q1 < delta || (q1 == delta && r1 == 0));
-
-    result_magic = q2 + 1; // resulting magic number
-    if (denom < 0)
-    {
-        result_magic = -result_magic;
-    }
-    *shift = p - bits; // resulting shift
-
-    return result_magic;
-}
-
-//------------------------------------------------------------------------
 // LowerSignedDivOrMod: transform integer GT_DIV/GT_MOD nodes with a power of 2
 // const divisor into equivalent but faster sequences.
 //
@@ -4335,12 +4250,12 @@ GenTree* Lowering::LowerSignedDivOrMod(GenTreePtr node)
 
         if (type == TYP_INT)
         {
-            magic = GetSignedMagicNumberForDivide<int32_t>(static_cast<int32_t>(divisorValue), &shift);
+            magic = MagicDivide::GetSigned32Magic(static_cast<int32_t>(divisorValue), &shift);
         }
         else
         {
 #ifdef _TARGET_64BIT_
-            magic = GetSignedMagicNumberForDivide<int64_t>(static_cast<int64_t>(divisorValue), &shift);
+            magic = MagicDivide::GetSigned64Magic(static_cast<int64_t>(divisorValue), &shift);
 #else
             unreached();
 #endif
--- a/src/jit/lower.cpp
+++ b/src/jit/lower.cpp
@@ -4244,6 +4244,11 @@ GenTree* Lowering::LowerSignedDivOrMod(GenTreePtr node)
 
     if (!isPow2(absDivisorValue))
     {
+        if (comp->opts.MinOpts())
+        {
+            return next;
+        }
+
 #ifdef _TARGET_XARCH_
         ssize_t magic;
         int     shift;
--- a/src/jit/lower.cpp
+++ b/src/jit/lower.cpp
@@ -144,7 +144,7 @@ GenTree* Lowering::LowerNode(GenTree* node)
 
         case GT_UDIV:
         case GT_UMOD:
-            LowerUnsignedDivOrMod(node);
+            return LowerUnsignedDivOrMod(node->AsOp());
             break;
 
         case GT_DIV:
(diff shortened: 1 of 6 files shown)


src/jit/lower.cpp
```diff
@@ -3841,46 +3841,266 @@ GenTree* Lowering::LowerAdd(GenTree* node)
...
+//    The magic number.
+//
+// Notes:
+//    This code is previously from UTC where it notes it was taken from
+//   _The_PowerPC_Compiler_Writer's_Guide_, pages 57-58. The paper is is based on
+//   is "Division by invariant integers using multiplication" by Torbjorn Granlund
+//   and Peter L. Montgomery in PLDI 94
+
+template <typename U>
+U GetUnsignedMagicNumberForDivide(U denom, int* shift /*out*/, bool* add /*out*/)
```
russellhadley: I'm interested to see if this shows up in throughput.   Just reading through, I was wondering if it would be worth pre-computing some set of common cases and leaving them in a table?

mikedn: There's only one occurrence in corelib and a couple of others in System.Runtime.Numerics and System.Xml.ReaderWriter so there's not much to measure. 

FWIW I measured how much time this function takes for `denom` = 10 and it's ~12ns on my machine. The time seems to reach ~70ns for larger denominators.

The book from which this code was adapted actually suggests the possibility of using a precomputed table but then the book was written in 1996, computers where kind of slow back then :)

mikedn: Interesting, the VC++ compiler has a table for 3-12.

Tornhoof: This is the classic algorithm, isn't it?
Did you consider Mölller and Granlund's updated one from 2011: Improved Division by Invariant Integers?

mikedn: @Tornhoof Yes, it's the original algorithm. I've seen "Improved Division by Invariant Integers" but it seems that it skewed towards multi word division so I didn't pay too much attention to it. Also, one of its basic premises is that that MULHI is slightly slower than a normal MUL but that doesn't seem to be the case anymore on current CPUs. Granlund's own instruction timing tables show that the latency dropped from 10 cycles (on Nehalem) to 3 cycles (on Haswell).

I was actually looking at another potential improvement: http://ridiculousfish.com/files/faster_unsigned_division_by_constants.pdf. It simplifies a bit the adjustment code required by stubborn divisors like 7. But I haven't seen any compiler using that and I have no idea if it's actually correct.

Tornhoof: Thanks for the explanation

danmosemsft:
@dotnet-bot test Tizen armel Cross Release Build  please  (break)
@dotnet-bot test Ubuntu arm Cross Release Build  please (break)

mikedn:
Well, this is pretty much done but there are a few questions:

1. This implements the "classic" algorithm that everyone uses. There is another algorithm for unsigned division that generates slightly shorter code for certain divisors (e.g. 7) and, interesting in the context of JIT, avoids the need for creating a temporary lclvar. But no native compiler I looked at (cl, clang, gcc, icc) uses this algorithm and I have no idea why. Perhaps it's actually incorrect, perhaps there are legal issues (though the reference code is in public domain). Should we looking into this more?
2. Should we use a precomputed table? This applies to the signed version as well.
3. Should this optimization be done in minopts/debug mode?  This applies to the signed version as well.

russellhadley:
1 - This seems like a good place to start - I think we should get this in before taking the next step.
2 - I think the pre-computed table is a good idea - for the "we're a JIT" reason.
3 - My thought is that we should only do this under opts.  (not minopt)  It doesn't reduce IR node count so it won't likely make a speedup.

mikedn:
I've added precomputed tables and disabled magic division optimization under minopts (for both signed and unsigned division). The tables cover the range 3-12, they're identical to the ones found in "The PowerPC Compiler Writer's Guide" so they're easy to double check.

russellhadley:
Remove the WIP tag on the PR?  @pgavlin do you have any feedback here?  LGTM.

src/jit/lower.cpp
```diff
@@ -3987,131 +3987,199 @@ GenTree* Lowering::LowerAdd(GenTree* node)
...
+        if (requiresDividendMultiuse)
+        {
+            LIR::Use dividendUse(BlockRange(), &divMod->gtOp1, divMod);
+            dividendLclNum = dividendUse.ReplaceWithLclVar(comp, curBBWeight);
+            dividend       = divMod->gtGetOp1();
         }
 
-        delta = absDenom - r2;
-    } while (q1 < delta || (q1 == delta && r1 == 0));
+        // Insert a new GT_MULHI node in front of the existing GT_UDIV/GT_UMOD node.
```
pgavlin: Nit: s/in front of/before/

src/jit/lower.cpp
```diff
@@ -3987,131 +3987,199 @@ GenTree* Lowering::LowerAdd(GenTree* node)
...
+            // divisor UMOD dividend = dividend SUB (div MUL divisor)
+            GenTree* divisor  = comp->gtNewIconNode(divisorValue, type);
+            GenTree* mul      = comp->gtNewOperNode(GT_MUL, type, div, divisor);
+            GenTree* dividend = comp->gtNewLclvNode(dividendLclNum, type);
+
+            divMod->SetOper(GT_SUB);
+            divMod->gtOp1 = dividend;
+            divMod->gtOp2 = mul;
+
+            BlockRange().InsertBefore(divMod, dividend, div, divisor, mul);
```
pgavlin: Nit: I think this is probably better ordered as `BlockRange().InsertBefore(divMod, div, divisor, mul, dividend)` in order to keep the lifetime of `dividend` shorter in the case its lclVar does not end up tracked.

src/jit/utils.cpp
```diff
@@ -1808,3 +1808,349 @@ float FloatingPointUtils::round(float x)
...
+template <typename T>
+const UnsignedMagic<T>* TryGetUnsignedMagic(T divisor)
+{
+    return nullptr;
+}
+
+template <>
+const UnsignedMagic<uint32_t>* TryGetUnsignedMagic(uint32_t divisor)
+{
+    // clang-format off
```
pgavlin: What sort of results were you getting from clang-format here?

src/jit/utils.cpp
```diff
@@ -1808,3 +1808,349 @@ float FloatingPointUtils::round(float x)
...
+    {
+        *shift = magic->shift;
+        *add   = magic->add;
+        return magic->magic;
+    }
+
+    typedef typename jitstd::make_signed<T>::type ST;
+
+    const unsigned bits         = sizeof(T) * 8;
+    const unsigned bits_minus_1 = bits - 1;
```
pgavlin: Nit: `bitsMinus1`, `twoNMinus1`, etc.

mikedn: Yeah, I should change those. I tried to be consistent with the signed version but there's no point in doing that.

src/jit/utils.cpp
mikedn:
```C++
    static const UnsignedMagic<uint32_t> table[]{
        {0xaaaaaaab, false, 1}, // 3
        {},
        {0xcccccccd, false, 2}, // 5
        {0xaaaaaaab, false, 2}, // 6
        {0x24924925, true, 3},  // 7
        {},
        {0x38e38e39, false, 1}, // 9
        {0xcccccccd, false, 3}, // 10
        {0xba2e8ba3, false, 3}, // 11
        {0xaaaaaaab, false, 3}, // 12
    };
```
I suppose it's reasonable, though I hate the misplaced opening brace :)

src/jit/lower.cpp
mikedn: >  in the case its lclVar does not end up tracked

Heh, that would be quite unfortunate. I'll update.

src/jit/utils.cpp
mikedn: Got rid of custom formatting, it's not worth it.

mikedn:
jit-diff summary:
```
Total bytes of diff: 143 (0.00 % of base)
    diff is a regression.
         133 : System.Runtime.Numerics.dasm (0.23 % of base)
          10 : System.Private.CoreLib.dasm (0.00 % of base)
2 total files with size differences (0 improved, 2 regressed), 77 unchanged.
         122 : System.Runtime.Numerics.dasm - BigNumber:FormatBigInteger(struct,ref,ref):ref
          11 : System.Runtime.Numerics.dasm - Number:Int32ToDecChars(long,byref,int,int)
          10 : System.Private.CoreLib.dasm - TimerQueue:get_TickCount():int
3 total methods with size differences (0 improved, 3 regressed), 65548 unchanged.
```

All diffs come from the newly added unsigned magic division optimization (that is, the signed magic division changes don't generate any diffs as expected).

1 occurrence in corelib and a few more in System.Runtime.Numerics, some `BigInteger` formatting code.

pgavlin:
>All diffs come from the newly added unsigned magic division optimization (that is, the signed magic division changes don't generate any diffs as expected).

Can you provide an example of one of the diffs?

mikedn:
A `x % 10` from `BigInteger` code:

```asm
; before
       mov      r15d, 10
       mov      eax, ecx
       xor      rdx, rdx
       div      edx:eax, r15d
       add      edx, 48
; after
       mov      eax, ecx
       mov      edx, 0xD1FFAB1E
       mov      dword ptr [rbp-64H], eax
       mul      edx:eax, edx
       shr      edx, 3
       imul     eax, edx, 10
       mov      edx, dword ptr [rbp-64H]
       sub      edx, eax
       add      edx, 48
```
The temporary lclvar makes an appearance...

pgavlin:
> The temporary lclvar makes an appearance...

Yeah, it would be nice if that reload was just a copy from `ecx`. Seems okay to me, though.

cc @CarolEidt @BruceForstall @dotnet/jit-contrib

pgavlin:
OS X failure looks infrastructural in nature.

src/jit/lower.cpp
```diff
@@ -3979,131 +3979,199 @@ GenTree* Lowering::LowerAdd(GenTree* node)
...
+            magic = MagicDivide::GetUnsigned64Magic(static_cast<uint64_t>(divisorValue), &add, &shift);
+#else
+            unreached();
+#endif
         }
 
-        q2 *= 2; // update q2 = 2^p / abs(denom)
-        r2 *= 2; // update r2 = rem(2^p / abs(denom))
+        // Depending on the "add" flag returned by GetUnsignedMagicNumberForDivide we need to generate:
+        // add == false (when divisor == 3 for example):
```
CarolEidt: Minor: I would probably have named the flag `needsAdd` or `useAdd` or something that implies that it is a boolean.

mikedn: "add" is the name used in "The PowerPC Compiler Writer's Guide". I don't like it but I don't think any variations around "add" are any better, IMO it should somehow indicate why that addition needs to be performed. Maybe I'll change it to "overflow" or something similar, it really indicates that the magic number is too large to fit in 32/64 bit.

PR #11001, opened 2017-04-16
Fix OOM exception thrown in case insensitive replace for an empty string

hughbe:
Found whilst exposing this API (https://github.com/dotnet/corefx/issues/1271)

The following code fails with an OOM, and takes a bunch of time:
```cs
"abc".Replace("", "def", StringComparison.CurrentCulture)
```


stephentoub:
Nice find. Thanks for the fix.

tarekgh:
@hughbe thanks for this catch. this is great!

PR #11004, opened 2017-04-17
Fix comment about pinning bit in object.h

303248153:
Refer to issue #10141 and PR #10515
Since pinning bit is moved to sync block, I fix the comment in GetGCSafeMethodTable.
In the previous PR @jkotas said it's better to keep the code unchanged so this PR only modify the comment.


jkotas:
Thanks!

PR #11010, opened 2017-04-17
In function Assembly.Load ignore AssemblyName.CodeBase even if it is set

### LINKED ISSUE #10561: System.NotSupportedException : Assembly.Load with a Codebase is not supported.
Either Assembly.GetName() should not set Codabase property or LoadFromAssemblyName should ignore CodeBase property in AssemblyName

rahku:
Fixes #10561

### DIFF
--- a/src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
+++ b/src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
@@ -101,14 +101,20 @@ namespace System.Reflection
         {
             Contract.Ensures(Contract.Result<Assembly>() != null);
             Contract.Ensures(!Contract.Result<Assembly>().ReflectionOnly);
-
+            
+            AssemblyName modifiedAssemblyRef = null;
             if (assemblyRef != null && assemblyRef.CodeBase != null)
             {
-                throw new NotSupportedException(SR.NotSupported_AssemblyLoadCodeBase);
+                modifiedAssemblyRef = (AssemblyName)assemblyRef.Clone();
+                modifiedAssemblyRef.CodeBase = null;
             }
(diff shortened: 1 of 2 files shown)


rahku:
PTAL @gkhanna79 @jkotas

src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
```diff
@@ -101,14 +101,16 @@ public static Assembly Load(AssemblyName assemblyRef)
         {
             Contract.Ensures(Contract.Result<Assembly>() != null);
             Contract.Ensures(!Contract.Result<Assembly>().ReflectionOnly);
-
-            if (assemblyRef != null && assemblyRef.CodeBase != null)
+            
+            AssemblyName modifiedAssemblyRef = null;
+            if (assemblyRef != null)
             {
-                throw new NotSupportedException(SR.NotSupported_AssemblyLoadCodeBase);
```
jkotas: Delete unused resource string?

src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
```diff
@@ -101,14 +101,16 @@ public static Assembly Load(AssemblyName assemblyRef)
...
+                modifiedAssemblyRef = (AssemblyName)assemblyRef.Clone();
```
jkotas: Clone is not a particularly cheap operation. It would be nice to do this only if `assemblyRef.CodeBase` non-null.

jkotas:
Is there going to be a matching corefx test update as well?

jkotas:
LGTM otherwise.

Issue #9040, opened 2017-01-21
Delegate's hash code is poor


stephentoub:
Delegate's GetHashCode just returns the hash code of the delegate type.  For a scenario where delegates are used as keys in a dictionary, this leads to obvious performance problems.  We should look at coming up with a better GetHashCode that properly factors in the target object and method, in all of the various forms a delegate can take.

mburbea:
When does it make sense to use delegates as keys to a dictionary? I guess it'd be best to just remove the `GetHashCode` impelementation and rely on the built-in object one. That's probably a breaking change.

stephentoub:
> When does it make sense to use delegates as keys to a dictionary

This came from a reported performance problem from a developer. I don't have details on their exact scenario.

> guess it'd be best to just remove the GetHashCode impelementation and rely on the built-in object one.

That would mean that delegates to the same method could have different hash codes. As you say, it'd be a big breaking change.

mburbea:
Well, it's still an improvement over delegates of the same type have the same key. (Unless your dictionary is literally keyed by `Delegate`. A key like `Func<int,int>` means every key hashes the same.)

I'd have to assume that delegates used for this purpose are cached, as creating new delegates isn't particularly cheap.

jkotas:
A significant breaking change is never an improvement.

mburbea:
@jkotas, I don't follow. Changing anything about it would be a "significant" breaking change. Equality would still work, and the usual case (cached delegates being used as keys) would see a performance improvement.

stephentoub:
> I don't follow

Code like:
```C#
static Dictionary<Action, State> s_data = ...;
static void SomeMethod() { ... }
...
State s;
if (s_data.TryGetValue(SomeMethod, out s))
{
    ...
}
```
would stop working, due to the delegate instances created for SomeMethod not being the same object and thus very likely having different hash codes.

While such a code example may not have great performance, such code exists.  The fact that at least one real-world app was experiencing slow downs as a result of this issue highlights that people do rely on delegates being used as keys in dictionaries, for whatever reason.

mburbea:
Thanks for the example, you're correct such code would fail if we changed to the default method. 

Building a proper hashcode for a delegate would be tricky as you'd have to consider the invocation lists as well.

jkotas:
[...]

PR #11019, opened 2017-04-17
Fixing Delegate's hash code's distribution

mountgellert:
Delegate's GetHashCode just returns the hash code of the delegate type. For a scenario where delegates are used as keys in a dictionary, this leads to obvious performance problems. We should look at coming up with a better GetHashCode that properly factors in the target object and method, in all of the various forms a delegate can take.

Fixes https://github.com/dotnet/coreclr/issues/9040

### DIFF
--- a/src/mscorlib/src/System/Delegate.cs
+++ b/src/mscorlib/src/System/Delegate.cs
@@ -179,7 +179,10 @@ namespace System
             else
                 return unchecked((int)((long)this._methodPtrAux));
             */
-            return GetType().GetHashCode();
+            if (_methodPtrAux.IsNull())
+                return ( _target != null ? RuntimeHelpers.GetHashCode(_target) * 33 : 0) + GetType().GetHashCode();
+            else
+                return GetType().GetHashCode();
         }
 
         public static Delegate Combine(Delegate a, Delegate b)
--- a/tests/src/CoreMangLib/cti/system/delegate/delegategethashcode1.cs
+++ b/tests/src/CoreMangLib/cti/system/delegate/delegategethashcode1.cs
@@ -218,7 +218,7 @@ namespace DelegateTest
         {
             bool retVal = true;
 
-            TestLibrary.TestFramework.BeginScenario("PosTest7:  Use the different instance's same instance method to create two delegate which delegate object is the same,their hashcode is equal");
+            TestLibrary.TestFramework.BeginScenario("PosTest7:  Use the different instance's same instance method to create two delegate which delegate object is the same, their hashcode is different");
 
             try
             {
@@ -226,7 +226,7 @@ namespace DelegateTest
                 booldelegate workDelegate = new booldelegate(new TestClass(1).StartWork_Bool);
                 booldelegate workDelegate1 = new booldelegate(new TestClass1(2).StartWork_Bool );
 
-                if (workDelegate.GetHashCode()!=workDelegate1.GetHashCode())
+                if (workDelegate.GetHashCode()==workDelegate1.GetHashCode())
                 {
                     TestLibrary.TestFramework.LogError("013", "HashCode is not excepted ");
                     retVal = false;


src/mscorlib/src/System/Delegate.cs
```diff
@@ -179,7 +179,7 @@ public override int GetHashCode()
             else
                 return unchecked((int)((long)this._methodPtrAux));
             */
-            return GetType().GetHashCode();
+            return ( _target != null ? _target.GetHashCode() : 0) * 33 + GetType.GetHashCode();
```
jkotas: `_target` should be included in the hash code only if  `_methodPtrAux` is null.

`_target` is overloaded field. It points other helper delegate objects when _methodPtrAux is non-null. The table at https://github.com/dotnet/coreclr/blob/5e03cb09e3754e4a3430ba424a1352b9c01ff098/src/vm/comdelegate.cpp#L3471 has the summary of the different options.

jkotas: Also, this needs to use `RuntimeHelpers.GetHashCode` so that it is not exposed to how custom types implement GetHashCode (e.g. they can implement it as `throw new NotSupportedException()`).

mountgellert: So something like
```
            if (_methodPtrAux.IsNull())
                return ( _target != null ? _target.GetHashCode() : 0) * 33 + GetType.GetHashCode();
            else
                return GetType.GetHashCode();
```
would be more suitable?

mountgellert: I don't really understand this comment; you mean instead of GetType().GetHashCode()?

mountgellert:
Now it seems to build.

mountgellert:
Windows_NT x64 Release Priority 1 Build and Test fails due to
```
                booldelegate workDelegate = new booldelegate(new TestClass(1).StartWork_Bool);
                booldelegate workDelegate1 = new booldelegate(new TestClass1(2).StartWork_Bool );
```

Having a different hash code with my changes. They are clearly different delegates, does it makes sense to have the same return code?

@jkotas if you agree on that, I'd modify the testcase as well so it would accept this (by basically flipping the comparism)

src/mscorlib/src/System/Delegate.cs
jkotas: I meant [`RuntimeHelpers.GetHashCode` ](https://msdn.microsoft.com/en-us/library/11tbk3h9(v=vs.110).aspx).

jkotas:
> Having a different hash code with my changes. They are clearly different delegates, does it makes sense to have the same return code?

Agree - the test needs updating.

> That one also does go through the invocationlist and not the target.

The invocationList will contain other Delegate object for MulticastDelegates, so it will end up calling your hashing function recursively.

> Does this mean that there is no easy way to solve this problem, and generally delegates shouldn't be used as keys in dictionaries?

I think your implementation works fine. It needs to be tested well though.

danmosemsft:
Are automated tests possible? (Ie., can we write a test where it's guaranteed that X and Y will have different hashcodes?). If so it would be great to have a CoreFX PR up.

mountgellert:
@danmosemsft Actually there is a test already that fails now, DelegateGetHashCode.PosTest7

src/mscorlib/src/System/Delegate.cs
mountgellert: Ah, to get the nonvirtual hashcode. Sure, updated.

mountgellert:
@jkotas I'm trying to figure out whether I need to update the Equals() method as well, can you advise.

jkotas:
No change needed in `Delegate.Equals`

mountgellert:
Than this is up for merge.

jkotas:
- delegategethashcode1.cs had a Unicode encoding, so it was showing up as binary in the diff. I have fixed the encoding of the file. Could you please resolve the conflict with it?

- We should add a regression test to corefx repo that fails without this change, and that will pass with this change. E.g. this test can create two delegates pointing to the same method on different instances, and verify that they have different hashcode. It can retry say 100 times to deal with the rare case of collisions. (There is a very theoretical change that it may fail incorrectly even after 100 times ... but it is many orders of magnitude smaller than overall system unreliability, so this is ok.) Are you interested in taking care of that?

jkotas:
https://github.com/dotnet/corefx/blob/master/src/System.Runtime/tests/System/DelegateTests.cs would be a good place to add this test.

danmosemsft:
To try your tests:
Here's how to run tests in corefx against your private corelib: https://github.com/dotnet/corefx/blob/c2588415e91ca1d44885ac0bbe9bf8268adc4b48/Documentation/project-docs/developer-guide.md#testing-with-private-coreclr-bits

mountgellert:
@jkotas sure, will do

mountgellert:
@jkotas I rebased and squashed the commits

Regards the other question - the test DelegateGetHashCode.PosTest7 does exactly this

tests/src/CoreMangLib/cti/system/delegate/delegategethashcode1.cs
```diff
@@ -218,15 +218,15 @@ public bool PosTest7()
...
+            TestLibrary.TestFramework.BeginScenario("PosTest7:  Use the different instance's same instance method to create two delegate which delegate object is the same, their hashcode is different");
 
             try
             {
                 DelegateGetHashCode delctor = new DelegateGetHashCode();
                 booldelegate workDelegate = new booldelegate(new TestClass(1).StartWork_Bool);
                 booldelegate workDelegate1 = new booldelegate(new TestClass1(2).StartWork_Bool );
 
-                if (workDelegate.GetHashCode()!=workDelegate1.GetHashCode())
+                if (workDelegate.GetHashCode()==workDelegate1.GetHashCode())
```
jkotas: This has a very small probability of failing intermittently, but it is probably ok.

jkotas:
Thanks!

PR #11022, opened 2017-04-17
Remove Alpine for pipe builds

gkhanna79:
skip ci please

This PR removes Alpine from CoreCLR official builds since CoreFX and Core-Setup already do not support it.

@weshaggard @wtgodbe PTAL

FYI - @Petermarcu


gkhanna79:
Merging to unblock official builds.

PR #10943, opened 2017-04-13
Reenable incremental linking

gkhanna79:
Fixes https://github.com/dotnet/coreclr/issues/3069

@jkotas PTAL

PR #11023, opened 2017-04-17
Remove /RELEASE option required for device drivers

gkhanna79:
Removes the /RELEASE Linker option that conflicted with /INCREMENTAL option and is required for device drivers (per https://msdn.microsoft.com/en-us/library/h8ksa72a.aspx).

@jkotas PTAL.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -328,7 +328,6 @@ if (WIN32)
   endif ()
     
   set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /LARGEADDRESSAWARE") # can handle addresses larger than 2 gigabytes
-  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /RELEASE") #sets the checksum in the header
   set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /NXCOMPAT") #Compatible with Data Execution Prevention
   set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /DYNAMICBASE") #Use address space layout randomization
   set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /PDBCOMPRESS") #shrink pdb size
--- a/src/dlls/mscoree/coreclr/CMakeLists.txt
+++ b/src/dlls/mscoree/coreclr/CMakeLists.txt
@@ -5,6 +5,10 @@ if (WIN32)
 
     set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /ENTRY:CoreDllMain")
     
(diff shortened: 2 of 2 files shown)


gkhanna79:

gkhanna79:

gkhanna79:
@RussKeldorph Are you aware of the following error that is in the WinX64 CI leg above:

```
D:\j\workspace\debug_windows---17180f1b\Tools\VersionTools.targets(61,5): error : Dependencies invalid: In 'D:\j\workspace\debug_windows---17180f1b\tests\src\JIT\config\benchmark\project.json', 'System.Runtime.CompilerServices.Unsafe 4.4.0-preview1-25210-01' must be '4.4.0-preview1-25214-03' (CoreFx) [D:\j\workspace\debug_windows---17180f1b\tests\build.proj]
15:28:41 
```

gkhanna79:
@wtgodbe Are you aware of this by any chance? This is hitting more than just my PR.

CC @danmosemsft

gkhanna79:
@dotnet-bot test this please

danmosemsft:
Dumpling didn't appear in the log, should have uploaded a dump

AndyAyersMS:
Build issue was my fault, too much of a time lag between when I tested and committed.

Should have been fixed by #11038.

gkhanna79:

gkhanna79:

gkhanna79:
Windows Debug tests fail for x64 with the following assert in InitThreadManager when RELEASE switch is not specified:

`Assert failure(PID 6628 [0x000019e4], Thread: 3164 [0x0c5c]): ((BYTE*)JIT_PatchedCodeLast - (BYTE*)JIT_PatchedCodeStart < 0x1000)
`
The assert fires because the gap is more than a page - specifically 0x50900 bytes. The file size of Coreclr.dll is also 27522KB.

When RELEASE switch is specified, the assert is not hit since the gap is 0x120 bytes. Also, CoreCLR.dll is 21589KB in size.

Thus, it looks like skipping the switch results in some additional native compiler debug build specific code to be generated that triggers off the assert.

@jkotas Any thoughts on this? Have you seen this before? I am going to follow up with the C++ team on this as well.

gkhanna79:
Fixing warnings introduced by https://github.com/dotnet/coreclr/pull/10943

jkotas:
We have hit it before in CoreRT actually: https://github.com/dotnet/corert/pull/2376 . The problem is that INCREMENAL linking adds a thunk around every function in .dll, even the ones written in assembly. It breaks all places where the low-level runtime code assumes that the address of a function is actually the address of a function.

It was fixed by unconditionally disabling incremental linking for CoreRT runtime. I did not find any reasonable way how to get around it. I think we should do the same for CoreCLR as well.

gkhanna79:
I have reached the same conclusion after getting a similar response from Linker folks. This means that I will close this PR and revert https://github.com/dotnet/coreclr/pull/10943 (which already had incremental linking disabled unconditionally).

jkotas:
We really only need to disable incremental linking for CoreCLR.dll. Other .dlls can have it enabled.

Also, it would be useful to add a comment on why the incremental building is disabled.

gkhanna79:
Yes, that is the plan I was going to pursue.

gkhanna79:
@jkotas PTAL at the latest changes.

PR #11027, opened 2017-04-17
Format spmi sources with clang-format

sandreenko:
The jit style was used.
My GitHub application was not able to pack all changes into one commit. It is why I separated them by folders.


sandreenko:
PTAL @BruceForstall, cc @dotnet/jit-contrib

jashook:
Are we going to be auto formatting the toolbox directory going forward?

sandreenko:
> Are we going to be auto formatting the toolbox directory going forward?

As I understand it is possible to add spmi or toolbox folders into jit-format list, but I am not familiar enough with jit-format source code.

BruceForstall:
> Are we going to be auto formatting the toolbox directory going forward?

I see no need to do that.

PR #11028, opened 2017-04-17
Add SecurityZone stub

danmosemsft:
Relates to https://github.com/dotnet/corefx/issues/18078 (still needs CoreFX exposure)

Stub just like the other CAS related members.
@ViktorHofer

### DIFF
--- a/src/mscorlib/shared/System/Security/SecurityException.cs
+++ b/src/mscorlib/shared/System/Security/SecurityException.cs
@@ -62,5 +62,16 @@ namespace System.Security
         public object PermitOnlySetInstance { get; set; }
         public string RefusedSet { get; set; }
         public string Url { get; set; }
+        public SecurityZone Zone { get; set; }
+    }
+
+    public enum SecurityZone
+    {
+        MyComputer   = 0,
+        Intranet     = 1,
+        Trusted      = 2,
+        Internet     = 3,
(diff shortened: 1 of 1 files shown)


ViktorHofer:
@danmosemsft Thanks for adding! Will you also update the corefx refs after coreclr sync or should I?

danmosemsft:
If you're offering -- sure, grab it. 
Signed off?

ViktorHofer:
Sure.

src/mscorlib/shared/System/Security/SecurityException.cs
```diff
@@ -62,5 +62,16 @@ protected SecurityException(SerializationInfo info, StreamingContext context)
         public object PermitOnlySetInstance { get; set; }
         public string RefusedSet { get; set; }
         public string Url { get; set; }
+        public SecurityZone Zone { get; set; }
```
ViktorHofer: Not sure how much we care about alphabetical order

ViktorHofer:
Just noticed that moving the enum into a separate file would have been better. (like other enums in that dir)

weshaggard:
@danmosemsft where do you plan to put SecurityZone in the contracts? I don't think this type is worth exposing at all as it is essentially dead and will never work.

Issue #4536, opened 2016-04-23
GC Stress Test does not run on Linux/Ubuntu


ramarag:
GC stress tests are failing with : 
http://dotnet-ci.cloudapp.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_tst_prtest/1055/testReport/junit/GC.Stress.Framework/ReliabilityFramework/ReliabilityFramework/

**[4/22/16 7:50:21 PM 1] Started
[4/22/16 7:50:21 PM 1] Argument: testmix_gc.config
[4/22/16 7:50:21 PM 1] Argument: /maximumExecutionTime:-1
[4/22/16 7:50:21 PM 1] Running tests...
[4/22/16 7:50:21 PM 1] Getting configuration...
[4/22/16 7:50:21 PM 1] Exception while running tests: System.IO.FileLoadException: Could not load file or assembly 'System.Xml.ReaderWriter, Version=4.0.11.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040 (FUSION_E_REF_DEF_MISMATCH))
File name: 'System.Xml.ReaderWriter, Version=4.0.11.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
   at ReliabilityConfig.GetTestsToRun(String testConfig)
   at ReliabilityConfig..ctor(String testConfig)
   at ReliabilityFramework.RunReliabilityTests(String testConfig, Boolean doReplay)
   at ReliabilityFramework.Main(String[] args)
[4/22/16 7:50:21 PM 1] Reliability framework is shutting down...
[4/22/16 7:50:21 PM 1] Shutdown w/ ret val of  -1**

mjsabby:
I've seen FUSION_E_REF_DEF_MISMATCH with R2R image being tried to load. If someone investigates this and it does turn out to be that could we explain it here?

swgillespie:
The original issue does not reproduce anymore, but the reliability framework still doesn't run on non-Windows platforms due to p/invokes into `kernel32.dll`. 

```
Unhandled Exception: System.DllNotFoundException: Unable to load DLL 'kernel32.dll': The specified module could not be found.
 (Exception from HRESULT: 0x8007007E)
   at ReliabilityFramework.IsDebuggerPresent()
   at ReliabilityFramework.MyDebugBreak(String extraData)
[...]

Issue #4534, opened 2016-04-23
GC Stress Test Sanity Run


ramarag:
Currently the smallest execution time for each GC stress is ~1 min, please consider dropping it down to seconds so that CI can run it or schedule a weekly full run whichever sounds good.

I am ensuring that all tests are loadable by sepcifying -1 to be the exceution time in https://github.com/dotnet/coreclr/pull/4533. Change it to seconds if possible

swgillespie:
Changing the units of time to seconds is kind of a pain, so I think we'll opt for the weekly CI run instead.

PR #11029, opened 2017-04-17
Some fixes for the GC Reliability Framework

swgillespie:
Addresses https://github.com/dotnet/coreclr/issues/4534 and https://github.com/dotnet/coreclr/issues/4536.

This PR is a collection of small fixes to improve the usage of the GC Reliability Framework. The RF was designed to run on the Desktop CLR and as such has a few desktop-isms that are fixed in this PR:
  1. The RF defines its own custom assembly loader, under the assumption that since it only depends on mscorlib, all incoming assembly load requests are tests. In .NET Core, this is not the case - CoreFX assemblies also get loaded using this loader, which is a problem for this custom assembly loader since it slaps a `.exe` suffix onto the assembly being loaded. To fix this, the assembly loader now crudely attempts to determine if it's loading a framework library by checking for a "System" prefix and probes `CORE_ROOT` (or the current directory) for framework libraries.
  2. The RF p/invoked into kernel32 in a few places  - these have been replaced with .NET Standard equivalents.
  3. The RF assumed that the environment variable `NUMBER_OF_PROCESSORS` is always defined. This PR makes it fall back to `Environment.ProcessorCount` if this environment variable is not defined.

Previously, users of the RF had to do some manual file copying to ensure that all dependencies are in their correct places and directory structures are correct. This PR makes three scenarios work:
  1. Running RF from `Core_Root` itself,
  2. Running RF from any directory, but with `CORE_ROOT` set to the Core_Root directory (CI scenario)
  3. Running RF from a directory where all dependencies have been copied in (the traditional scenario).

A nice benefit of this is that the GC sanity run now works fine on all platforms that I tested it on. We could turn the RF into a CI job pretty easily going forward, which IMO we should definitely do.

cc @sergiy-k @Maoni0 @adityamandaleeka

### DIFF
--- a/tests/src/GC/Stress/Framework/ReliabilityFramework.cs
+++ b/tests/src/GC/Stress/Framework/ReliabilityFramework.cs
@@ -560,6 +560,7 @@ public class ReliabilityFramework
         return (99);
     }
 
+#if !PROJECTK_BUILD
     [DllImport("kernel32.dll")]
     private extern static void DebugBreak();
 
@@ -568,6 +569,7 @@ public class ReliabilityFramework
 
     [DllImport("kernel32.dll")]
     private extern static void OutputDebugString(string debugStr);
+#endif
 
     /// <summary>
     /// Checks to see if we should block all execution due to a fatal error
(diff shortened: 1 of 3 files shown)


tests/src/GC/Stress/Framework/ReliabilityFramework.cs
```diff
@@ -1688,28 +1745,38 @@ private void TestPreLoader_Process(ReliabilityTest test, string[] paths)
     {
         Console.WriteLine("Preloading for process mode");
 
+        Console.WriteLine("basepath: {0}, asm: {1}", test.BasePath, test.Assembly);
```
swgillespie: I left all of this logging in because I usually end up adding it back whenever I use the RF - I'm happy to delete it if it's too ugly.

danmosemsft:
retest

adityamandaleeka:
> We could turn the RF into a CI job pretty easily going forward, which IMO we should definitely do.

👍

adityamandaleeka:
LGTM

swgillespie:
thanks!

Issue #10968, opened 2017-04-14
CoreCLR failed to build on Windows due to MSBuild error MSB1001: unknown switch --TargetsWindows=true


KarenHuang2016:
I tried to build CoreCLR on Windows with VS2017 RTM. It failed to build when execute msbuild.cmd. This issue is caused by master branch revision [42f1781](https://github.com/dotnet/coreclr/commit/42f1781463e973a7e764675715191b4f137d4e05#diff-f35ce63d6e75aecbc109ddc24e0d77f0).
This revision added a parameter _-TargetsWindow=true_ to msbuild.cmd. It seems the correct parameter should be _/p:TargetsWindows=true_.

You can repro the issue as the steps below:
1. git clone https://github.com/dotnet/coreclr D:\CoreCLR\src
2. Open a clean x86 prompt (C:\windows\syswow64\cmd.exe) and browse to D:\CoreCLR\src
3. build x86 release -Rebuild -- /clp:ShowCommandLine

Error info:
BUILDTEST: Creating test wrappers...
Tools are already initialized.
Running: D:\CoreCLR\src\Tools\dotnetcli\dotnet.exe D:\CoreCLR\src\Tools\run.exe D:\CoreCLR\src\config.json build -Project=D:\CoreCLR\src\tests\runtest.proj -BuildWrappers -MsBuildEventLogging=" " -MsBuildLog=/flp:Verbosity=normal;LogFile="D:\CoreCLR\src\bin\Logs\Tests_XunitWrapper_Windows_NT__x86__Release.log" -MsBuildWrn=/flp1:WarningsOnly;LogFile="D:\CoreCLR\src\bin\Logs\Tests_XunitWrapper_Windows_NT__x86__Release.wrn" -MsBuildErr=/flp2:ErrorsOnly;LogFile="D:\CoreCLR\src\bin\Logs\Tests_XunitWrapper_Windows_NT__x86__Release.err" -BuildOS=Windows_NT -BuildType=Release -BuildArch=x86    -Rebuild -- /clp:ShowCommandLine -TargetsWindows=true
[...]

PR #11031, opened 2017-04-17
Fix error caused by passing -TargetsWindows at end of Command Line

wtgodbe:
Resolves https://github.com/dotnet/coreclr/issues/10968. The problem was that I was passing "-TargetsWindows=true/false" after any unprocessed Build Arguments - but since the run tool interprets anything after the string "--" as arguments to pass directly to Msbuild, if any such arg was passed to build.cmd, then "-TargetsWindows=true" wouldn't get interpreted as an arg to the run tool, but rather as an arg to msbuild. The solution is to move "UnprocessedBuildArgs" to the very end of every invocation of msbuild-ing run.exe.

CC @KarenHuang2016 @danmosemsft


wtgodbe:
@dotnet-bot test this please

danmosemsft:
retest

PR #11032, opened 2017-04-17
Implement the createdump policy options and env variables

mikem8361:
Added these command line options:

-f, --name - dump path and file name. The pid can be placed in the name with %d. The default is "/tmp/coredump.%d"
-n, --normal - create minidump (default).
-h, --withheap - create minidump with heap.
-m, --micro - create triage minidump.
-d, --diag - enable diagnostic messages.

Added these environment variables:

COMPlus_DbgMiniDumpType - if set to "1" generate MiniDumpNormal, "2" MiniDumpWithPrivateReadWriteMemory, "3" MiniDumpFilterTriage. Default is MiniDumpNormal.
COMPlus_DbgMiniDumpName - if set, use as the template to create the dump path and file name. The pid can be placed in the name with %d. The default is "/tmp/coredump.%d".


danmosemsft:
retest

PR #11033, opened 2017-04-17
Move handle object manipulation functions to interface

adityamandaleeka:
This change moves functions related to modifying objects in handles to the IGCHandleManager interface.

I thought about having `StoreObjectInHandle` purely on the VM side, but there is some bookkeeping the GC has to do apart from just setting the pointer: It has to call the handle write barrier and also fire some ETW events. I believe our plan is to eventually have the logic and manifest for GC-related events purely on the GC side anyway.

Edit: this is done with my latest commits. ~~We should of course move the handle and app domain validation stuff that's being done over to the VM side. I already moved some of it when I moved `ObjectFromHandle` to the VM. We can refactor that to cover all the validation done by these functions and call it from the VM side stubs before they call the interface methods.~~

@jkotas @Maoni0 @swgillespie @sergiy-k

### DIFF
--- a/src/gc/gcinterface.h
+++ b/src/gc/gcinterface.h
@@ -449,7 +449,7 @@ public:
 
     virtual bool StoreObjectInHandleIfNull(OBJECTHANDLE handle, Object* object) = 0;
 
-    virtual Object* CompareAndSwapObjectInHandle(OBJECTHANDLE handle, Object* object, Object* comparandObject) = 0;
+    virtual Object* InterlockedCompareExchangeObjectInHandle(OBJECTHANDLE handle, Object* object, Object* comparandObject) = 0;
 };
 
 // IGCHeap is the interface that the VM will use when interacting with the GC.
--- a/src/gc/gcinterface.h
+++ b/src/gc/gcinterface.h
@@ -444,6 +444,12 @@ public:
     virtual void DestroyHandleOfUnknownType(OBJECTHANDLE handle) = 0;
 
     virtual void* GetExtraInfoFromHandle(OBJECTHANDLE handle) = 0;
+
+    virtual void StoreObjectInHandle(OBJECTHANDLE handle, Object* object) = 0;
+
+    virtual bool StoreObjectInHandleIfNull(OBJECTHANDLE handle, Object* object) = 0;
+
+    virtual Object* CompareAndSwapObjectInHandle(OBJECTHANDLE handle, Object* object, Object* comparandObject) = 0;
 };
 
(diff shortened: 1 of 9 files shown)


adityamandaleeka:
Added commits to move the app domain/object validation to the VM side. The logic is the same as it was before.

danmosemsft:
retest

adityamandaleeka:

Looks like an infrastructure issue: `Agent went offline during the build`

adityamandaleeka:
@swgillespie @jkotas PTAL

src/gc/gcinterface.h
```diff
@@ -444,6 +444,12 @@ class IGCHandleManager {
     virtual void DestroyHandleOfUnknownType(OBJECTHANDLE handle) = 0;
 
     virtual void* GetExtraInfoFromHandle(OBJECTHANDLE handle) = 0;
+
+    virtual void StoreObjectInHandle(OBJECTHANDLE handle, Object* object) = 0;
+
+    virtual bool StoreObjectInHandleIfNull(OBJECTHANDLE handle, Object* object) = 0;
+
+    virtual Object* CompareAndSwapObjectInHandle(OBJECTHANDLE handle, Object* object, Object* comparandObject) = 0;
```
jkotas: What was wrong with InterlockedCompareExchangeObjectInHandle ?

jkotas: InterlockedCompareExchangeObjectInHandle looked more self-describing to me.

jkotas:
LGTM

src/gc/gcinterface.h
adityamandaleeka: I felt that `InterlockedCompareExchangeObjectInHandle` was too Windows-specific (since that's the name of the actual Windows API that we use), while compare and swap is the generic term for the concept (https://en.wikipedia.org/wiki/Compare-and-swap).

If you feel strongly about this, I am happy to change it back.

jkotas: Interlocked CompareExchange is also name of the cross-plat .NET API, and we have `Interlocked ::CompareExchange` in gcenv.interlocked.inl.

The GCHeap interface has a Windows feel in it because of it is using CamelCase naming, so I do not think that name with Windows heritage looks unnatural in it.

BTW: C++ standard is calling it `atomic::compare_exchange` (http://en.cppreference.com/w/cpp/atomic/atomic).

adityamandaleeka: OK. I'll change the name back before merging this PR.

PR #11036, opened 2017-04-17
Remove EventCounter from System.Private.CoreLib

brianrob:
EventCounter should live in System.Diagnostics.Tracing (partial façade).  I'll add it back to the façade in a subsequent change in CoreFX.


danmosemsft:
restart tests

PR #11038, opened 2017-04-17
Fix test dependency on System.Runtime.CompilerServices.Unsafe

brianrob:
The test build complains that its dependencies aren't up-to-date.  Running the fix-up tool pull this updated version of the dependency.


brianrob:
This appears to have come from #10910.

danmosemsft:
Since it was a build break I already did this :)

brianrob:
Good deal.  Thanks.

AndyAyersMS:
Thanks.

Looks like when you introduce new dependencies you shouldn't let much time pass between when you test and when you commit.

PR #11039, opened 2017-04-18
Fix String.Replace implementation

tarekgh:
String.Replace was callin ga low level CompareInfo method IndexOfCore which cannot be called in some conditions like when using ordinal ignore case option or when running in invariant mode.
The fix is to call a higher level method in CompareInfo which can handle such cases becfore calling the lower level method

### DIFF
--- a/src/mscorlib/src/System/Globalization/CompareInfo.cs
+++ b/src/mscorlib/src/System/Globalization/CompareInfo.cs
@@ -801,7 +801,7 @@ namespace System.Globalization
             return IndexOfCore(source, value, startIndex, count, options, null);
         }
 
-        // The following IndexOf overload is mainly used by String.Replace. This overload assume the parameters are already validated
+        // The following IndexOf overload is mainly used by String.Replace. This overload assumes the parameters are already validated
         // and the caller is passing a valid matchLengthPtr pointer.
         internal unsafe int IndexOf(string source, string value, int startIndex, int count, CompareOptions options, int* matchLengthPtr)
         {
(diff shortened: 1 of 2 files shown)


tarekgh:
CC @stephentoub @danmosemsft

tarekgh:
https://github.com/dotnet/corefx/issues/18459

tarekgh:
CC @hughbe

src/mscorlib/src/System/Globalization/CompareInfo.cs
```diff
@@ -802,6 +801,53 @@ public unsafe virtual int IndexOf(string source, string value, int startIndex, i
             return IndexOfCore(source, value, startIndex, count, options, null);
         }
 
+        // The following IndexOf overload is mainly used by String.Replace. This overload assume the parameters are already validated
+        // and the caller is passing a valid matchLengthPtr pointer.
+        internal unsafe int IndexOf(string source, string value, int startIndex, int count, CompareOptions options, int* matchLengthPtr)
```
danmosemsft: It seems that matchLengthPtr is just a way of passing the length of the match (which is always either null or value.Length, unless you end up calling FindNLSStringEx). It's only a pointer because that's what FindNLSStringEx requires. So why not replace it everywhere in the managed code with just a ref int ?

src/mscorlib/src/System/Globalization/CompareInfo.cs
```diff
@@ -802,6 +801,53 @@ public unsafe virtual int IndexOf(string source, string value, int startIndex, i
             return IndexOfCore(source, value, startIndex, count, options, null);
         }
 
+        // The following IndexOf overload is mainly used by String.Replace. This overload assume the parameters are already validated
```
stephentoub: Nit: assume => assumes

PR #11042, opened 2017-04-18
Added reference to UsingCoreRun.md

mmayr-at:
I added a reference to UsingCoreRun.md on the debug instructions page. It was not obvious for me, that i might need some further set up for `corerun.exe` to debug the CoreCLR.

### DIFF
--- a/Documentation/building/debugging-instructions.md
+++ b/Documentation/building/debugging-instructions.md
@@ -1,7 +1,9 @@
 Debugging CoreCLR
 =================
 
-These instructions will lead you through debugging CoreCLR on Windows and Linux. They will be expanded to support OS X when we have good instructions for that.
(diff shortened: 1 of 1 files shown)


adityamandaleeka:
@mmayr-at Thanks for trying to clarify those instructions, but this isn't the right change to make. Debugging CoreCLR does not require corerun; it just happens to be the case that the example used in those instructions runs a HelloWorld program with corerun.

mmayr-at:
I'm aware of that debugging CoreCLR can be done in several ways. I was just looking for some quick instructions on how to start debugging and found this page. I had some problems running `corrun.exe`, so  I thought a reference to using `corrung.exe` would be helpful here.

danmosemsft:
@mmayr-at thanks for this contribution, but I'm going to close it based on @adityamandaleeka 's comment. We'd weclome if you take another issue either here or in corefx repo.

Issue #10729, opened 2017-04-05
[RyuJIT/ARM32] Abort on FailFast


mskvortsov:
PR #10609 uncovered (or introduced?) several failures across the full testsuite run:
```
Assert failure(PID 2250 [0x000008ca], Thread: 2250 [0x08ca]): !"About to FailFast. set ComPlus_AssertOnFailFast=0 if this is expected"
    File: /home/mskvortsov/git/coreclr/src/vm/jithelpers.cpp Line: 5555
    Image: /home/mskvortsov/clr/corerun

Aborted
```
The test cases are:
```
GC/API/GC/GetGeneration_fail/GetGeneration_fail.exe
GC/Features/LOHCompaction/lohcompactapi2/lohcompactapi2.exe
GC/Features/Pinning/PinningOther/PinnedCollect/PinnedCollect.exe
GC/Features/Pinning/PinningOther/PinnedHandle/PinnedHandle.exe
GC/Features/Pinning/PinningOther/PinnedObject/PinnedObject.exe
Interop/MarshalAPI/Miscellaneous/MarshalClassTests/MarshalClassTests.exe
Interop/MarshalAPI/OffsetOf/OffsetOf/OffsetOf.exe
Interop/NativeCallable/NativeCallableTest/NativeCallableTest.exe
...
```

mskvortsov:
@alpencolt Could you take a look?

alpencolt:
I will check. It can reproduce only on release build.

alpencolt:
Cannot reproduce on latest versions `coreclr` and `corefx`. Only `OffsetOf.exe` and `NativeCallableTest.exe` are failed. First uses structures, which aren't implemented in ARM32. The second are failed on legacy backend but with another exception:
```C#
Unhandled Exception: System.DllNotFoundException: Unable to load DLL 'NativeCallableDll': The specified module or one of its dependencies could not be found.
 (Exception from HRESULT: 0x8007007E)
   at Program.NativeMethods.CallManagedAdd(IntPtr callbackProc, Int32 n)
   at TestNativeCallable()
   at Program.TestNativeCallableValid()
   at Program.Main()
Aborted

```
Task can be postponed I think.

mskvortsov:
@alpencolt That exception means you missed `libNativeCallableDll.so` and others (see `bin/Product/Linux.arm.Debug/bin` of coreclr build output).

mskvortsov:
Another FailFast on current tip (`cb371832`):
```
[...]

PR #11044, opened 2017-04-18
[RyuJIT/ARM32] Use 'subs' instead of 'sub' on localloc

alpencolt:
Previous implementation used `sub` instruction for counting iterations while should use `subs` on
conditional execution.


alpencolt:
Fix for #10729
cc @dotnet/arm32-contrib

BruceForstall:
cc @dotnet/jit-contrib

PR #11049, opened 2017-04-18
Fix incorrect test of structSize during struct promotion

briansull:
Fixes the desktop test Crossgen NI and Ngen NI validation

For the Desktop build of crossgen FEATURE_SIMD is not defined so under desktop crossgen we would not struct promote a struct of size 32 (4 pointer sized fields).

For other builds FEATURE_SIMD is defined so we would allow struct promote of struct up to 63 bytes.
With this change we can struct promote a 64 byte struct with 4 SIMD fields.


BruceForstall:

BruceForstall:
@dotnet/jit-contrib

PR #11056, opened 2017-04-19
Honor disabled GC on Transitions during GC stress on Unix

parjong:
GC.Coverage.LargeObjectAlloc currently gets stuck in the presence of GC stress (level 3). 

It seems that  GC_ON_TRANSITIONS has logic to avoid this stuck, but this logic is enabled only for non-UNIX platforms (i.e. Windows).

This commit ports this logic in order to prevent hang during GC stress test.

### DIFF
--- a/src/vm/threads.h
+++ b/src/vm/threads.h
@@ -6735,29 +6735,28 @@ public:
                             Thread::TriggersGC(GetThread());                \
                         } while(0)
 
-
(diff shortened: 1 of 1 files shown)


parjong:
\CC @wateret

parjong:
@janvorli Please take a look.

janvorli:
@parjong I took the liberty to modify the title of this PR to better reflect what it fixes.

PR #11059, opened 2017-04-19
Update StringBuilder.AppendJoin to approved API 

jnm2:
See https://github.com/dotnet/corefx/issues/3419

- [x] Q: Is it worth specializing the `string[]` overloads to not call `.ToString()` on strings? The array implementation is generic, so I wonder if that optimization is already done.
- [x] Q: Is it worth specializing concat-only loops when the separator is null or empty so that `Append(&string.Empty, 0)` isn't called in the loop?
- [x] Q: Is it better to access each array element twice rather than putting every array element in a local to check for null?
- [x] Q: Do any tests need to be added to https://github.com/dotnet/coreclr/blob/master/tests/src/CoreMangLib/cti/system/text/stringbuilder or just https://github.com/dotnet/corefx/blob/master/src/System.Runtime/tests/System/Text/StringBuilderTests.cs?
<br>

- [x] TODO: corefx PR with tests on this PR

/cc @karelz @AlexGhiondea @joperezr @danmosemsft

### DIFF
--- a/src/mscorlib/shared/System/Text/StringBuilder.cs
+++ b/src/mscorlib/shared/System/Text/StringBuilder.cs
@@ -1081,25 +1081,23 @@ namespace System.Text
             return this;
         }
 
-        // Append joined values with a separator between each value.
-        public unsafe StringBuilder AppendJoin(char separator, params object[] values)
-        {
-            // Defer argument validation to the internal function
-            return AppendJoinCore(&separator, 1, values);
-        }
-
-        public unsafe StringBuilder AppendJoin(char separator, params string[] values)
-        {
-            // Defer argument validation to the internal function
-            return AppendJoinCore(&separator, 1, values);
-        }
+        
+        #region AppendJoin
 
         public unsafe StringBuilder AppendJoin(string separator, params object[] values)
         {
             separator = separator ?? string.Empty;
             fixed (char* pSeparator = separator)
             {
-                // Defer argument validation to the internal function
+                return AppendJoinCore(pSeparator, separator.Length, values);
+            }
+        }
+
+        public unsafe StringBuilder AppendJoin<T>(string separator, IEnumerable<T> values)
+        {
+            separator = separator ?? string.Empty;
+            fixed (char* pSeparator = separator)
+            {
                 return AppendJoinCore(pSeparator, separator.Length, values);
             }
         }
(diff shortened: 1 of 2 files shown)


danmosemsft:
For tests -- they should go in CoreFX, and yes, please have a PR up in CoreFX for them before we merge this one. You'll want to have the tests written anyway so you can run them against this before we merge it.
Here's how to run tests in corefx against your private corelib: https://github.com/dotnet/corefx/blob/c2588415e91ca1d44885ac0bbe9bf8268adc4b48/Documentation/project-docs/developer-guide.md#testing-with-private-coreclr-bits

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -1163,6 +1153,32 @@ private unsafe StringBuilder AppendJoinCore<T>(char* separator, int separatorLen
             return this;
         }
 
+        private unsafe StringBuilder AppendJoinCore<T>(char* separator, int separatorLength, T[] values)
+        {
+            if (values == null)
+                throw new ArgumentNullException(nameof(values));
```
danmosemsft: Much of corelib would do `ThrowHelper.ThrowArgumentNullException` to reduce IL. but I see we don't do it anywhere in this class so I guess this is fine.

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -1163,6 +1153,32 @@ private unsafe StringBuilder AppendJoinCore<T>(char* separator, int separatorLen
...
+            Contract.Ensures(Contract.Result<StringBuilder>() != null);
```
danmosemsft: Also optional, but we can delete Contract.Ensures. We have given up on enforcing contracts, and this one cannot easily become a Debug.Assert.
https://github.com/dotnet/corefx/issues/15443

src/mscorlib/shared/System/Text/StringBuilder.cs
jnm2: Happy to. Is there any reason for me not to?

src/mscorlib/shared/System/Text/StringBuilder.cs
```diff
@@ -1081,62 +1081,52 @@ public StringBuilder Append(char[] value)
...
-            return AppendJoinCore(&separator, 1, values);
+            separator = separator ?? string.Empty;
+            fixed (char* pSeparator = separator)
+            {
+                return AppendJoinCore(pSeparator, separator.Length, values);
+            }
         }
 
-        public unsafe StringBuilder AppendJoin<T>(string separator, IEnumerable<T> values)
+        public unsafe StringBuilder AppendJoin(string separator, params string[] values)
```
danmosemsft: If we are to match `String.Join` it is (by accident most likely) `value` not `values`. unlike for `object[]`. @terrajobst do we care? can we be values in both cases?

src/mscorlib/shared/System/Text/StringBuilder.cs
jnm2: @danmosemsft I'd have to add a `values` field to the ExceptionArgument enum, right? https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/ThrowHelper.cs#L342

danmosemsft:
>specializing the string[] overloads to not call .ToString() on strings

string.ToString() is of course `return this` which hopefully inlines very well so I don't think this is an issue

>Is it worth specializing concat-only loops when the separator is null or empty so that Append(&string.Empty, 0) isn't called in the loop?

I think only if this shows up in traces. there is a code complexity/code size cost especially for generics.

>s it better to access each array element twice rather than putting every array element in a local to check for null

I often see this done in "hot" paths (storing array element in a local) but I cannot say whether it ctually helps code gen.. @jkotas ?

src/mscorlib/shared/System/Text/StringBuilder.cs
danmosemsft: yep

jkotas:
> I often see this done in "hot" paths (storing array element in a local) but I cannot say whether it actually helps code gen.. @jkotas ?

It does help for complex hot paths where the JIT optimizations are not able to CSE the value. 

In this particular case, the JIT optimizations are likely to CSE the value because of it is very simple code; and even if they did not, it is unlikely to make a measurable difference because of the whole method is relatively expensive. Caching the value in local variable here does not help anything; but it does not hurt much either.

Note that there is a small observable difference due to the caching in local variable: The user-overridden ToString method can observe whether it is being called on the cached copy; or on the array element directly.

jkotas:
> `string.ToString()` is of course return this which hopefully inlines very well 

I do not think it will inline here because of the method is generic, string is reference type and thus the method will be shared generic code. As always, you need to measure whether the extra virtual call matters here... .

danmosemsft:
@jkotas is it reasonable to submit this without investigating these possible optimizations? This is a new API so there's no regression potential.

jkotas:
I do not have a problem with it.

danmosemsft:
@jnm2 please put up aCoreFX PR with tests before we merge ths..

src/mscorlib/shared/System/Text/StringBuilder.cs
terrajobst: We care, but I think `values` makes much more sense. I think we should follow `String.Join<T>` which used `values` as well:

https://apisof.net/catalog/System.String.Join%3CT%3E(String,IEnumerable%3CT%3E)

danmosemsft: @jnm2 please use values in all cases.

jnm2:
I have a slight preference for removing the local copy for the array, any objections?

@danmosemsft Yes, underway.

danmosemsft:
@jnm fine to not cache it, based on Jan's comment.

jnm2:
CoreFX PR is ready. https://github.com/dotnet/corefx/issues/18699

danmosemsft:
ARM failure is random. https://github.com/dotnet/coreclr/issues/11123

danmosemsft:
Thanks! Should be in CoreFX in 24 hours.

PR #11062, opened 2017-04-19
[x86/Linux] Relax inst_JMP assert condition

### LINKED ISSUE #11061: [x86/Linux] Assertion failed while running FX System.Dynamic.Runtime.Tests
System.Dynamic.Runtime.Tests hits the following assertion failure:
```

Discovering: System.Dynamic.Runtime.Tests
Discovered:  System.Dynamic.Runtime.Tests
Starting:    System.Dynamic.Runtime.Tests

Assert failure(PID 20537 [0x00005039], Thread: 20537 [0x5039]): Assertion failed 'tgtBlock->bbTgtStkDepth * sizeof(int) == genStackLevel || compiler->rpFrameType != FT_ESP_FRAME' in 'Microsoft.CSharp.RuntimeBinder.CSharpSetMemberBinder:PopulateSymbolTableWithName(ref,ref,ref):this' (IL size 26)

    File: /var/jenkins/workspace/CLR/x86_Dbg_RT/repo/code/src/jit/instr.cpp Line: 247
    Image: /var/jenkins/workspace/FX/FX_x86_Debug_Test/working/overlay/corerun
```

parjong:
In x86/Linux codegen, jump to another basic block is allowed only when two blocks have the same stack level, or the method is non-ESP framed.

Stack alignment padding is not included for the stack level of target basic block, but included for that of current basic block.

This difference results in the assertion failure discussed in #11061.

This commit relaxes assert condition to fix #11061.

### DIFF
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -2692,12 +2692,22 @@ void CodeGen::genExitCode(BasicBlock* block)
 
 void CodeGen::genJumpToThrowHlpBlk(emitJumpKind jumpKind, SpecialCodeKind codeKind, GenTreePtr failBlk)
 {
-    if (!compiler->opts.compDbgCode)
-    {
-        /* For non-debuggable code, find and use the helper block for
-           raising the exception. The block may be shared by other trees too. */
+    /* For non-debuggable code, find and use the helper block for
+       raising the exception. The block may be shared by other trees too.
 
-        BasicBlock* tgtBlk;
+       Otherwise, the code to throw the exception will be generated inline,
+       and we will jump around it in the normal non-exception case */
+    bool useExnThrowingBlock = !compiler->opts.compDbgCode;
+
+#if !FEATURE_FIXED_OUT_ARGS
+    /* Allow the use of exception throwing blocks only for EBP frame.
+       (two basic blocks may have different stack level due to push/pop) */
+    useExnThrowingBlock = useExnThrowingBlock && (compiler->rpFrameType != FT_ESP_FRAME);
+#endif
+
+    if (useExnThrowingBlock)
+    {
+       BasicBlock* tgtBlk;
 
         if (failBlk)
         {
@@ -2727,9 +2737,6 @@ void CodeGen::genJumpToThrowHlpBlk(emitJumpKind jumpKind, SpecialCodeKind codeKi
     }
     else
     {
-        /* The code to throw the exception will be generated inline, and
-           we will jump around it in the normal non-exception case */
-
         BasicBlock*  tgtBlk          = nullptr;
         emitJumpKind reverseJumpKind = emitter::emitReverseJumpKind(jumpKind);
         if (reverseJumpKind != jumpKind)
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -2692,15 +2692,11 @@ void CodeGen::genExitCode(BasicBlock* block)
 
 void CodeGen::genJumpToThrowHlpBlk(emitJumpKind jumpKind, SpecialCodeKind codeKind, GenTreePtr failBlk)
 {
-    // For non-debuggable code, find and use the helper block for
-    // raising the exception. The block may be shared by other trees too.
-    //
-    // Otherwise, the code to throw the exception will be generated inline,
-    // and we will jump around it in the normal non-exception case
-    bool useExnThrowingBlock = !compiler->opts.compDbgCode;
-
-    if (useExnThrowingBlock)
+    if (!compiler->opts.compDbgCode)
     {
+        /* For non-debuggable code, find and use the helper block for
+           raising the exception. The block may be shared by other trees too. */
+
         BasicBlock* tgtBlk;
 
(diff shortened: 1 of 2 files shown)


parjong:

parjong:
@BruceForstall @dotnet/jit-contrib Please take a look.

src/jit/codegencommon.cpp
```diff
@@ -2692,11 +2692,21 @@ void CodeGen::genExitCode(BasicBlock* block)
...
+
+       Otherwise, the code to throw the exception will be generated inline,
+       and we will jump around it in the normal non-exception case */
+    bool useExnThrowingBlock = !compiler->opts.compDbgCode;
+
+#if !FEATURE_FIXED_OUT_ARGS
+    /* Allow the use of exception throwing blocks only for EBP frame.
+       (two basic blocks may have different stack level due to push/pop) */
+    useExnThrowingBlock = useExnThrowingBlock && (compiler->rpFrameType != FT_ESP_FRAME);
+#endif
```
BruceForstall: I think you should be consistent with other code and use:
```
    useExnThrowingBlock = useExnThrowingBlock && isFramePointerUsed();
```

Also, please change the comment style to `//` instead of `/*` for the comments you change/add.

parjong:
@BruceForstall As you pointed out, the method of interest has only one jump (to exception throwing block). 

When this assertion failed, ``tgtBlock->bbTgtStkDepth`` is 1 and ``genStackLevel`` is 12. I guess that stack alignment padding is not considered when ``bbTgtStkDepth`` is computed.

This means that we need to update ``bbTgtStkDepth`` properly in order to fix this issue.

src/jit/codegencommon.cpp
```diff
@@ -2692,11 +2692,15 @@ void CodeGen::genExitCode(BasicBlock* block)
...
-    if (!compiler->opts.compDbgCode)
-    {
-        /* For non-debuggable code, find and use the helper block for
-           raising the exception. The block may be shared by other trees too. */
+    // For non-debuggable code, find and use the helper block for
+    // raising the exception. The block may be shared by other trees too.
+    //
+    // Otherwise, the code to throw the exception will be generated inline,
+    // and we will jump around it in the normal non-exception case
+    bool useExnThrowingBlock = !compiler->opts.compDbgCode;
```
BruceForstall: I'm ok with your changes to codegencommon.cpp as "cleanup" changes. But you didn't make any functional changes, so it might be better to just not change the file at all, for simplicity.

src/jit/instr.cpp
```diff
@@ -239,13 +239,18 @@ void CodeGen::instNop(unsigned size)
...
+// On the x86 we are pushing (and changing the stack level), but on x64 and other archs we have
+// a fixed outgoing args area that we store into and we never change the stack level when calling methods.
+//
+// Thus only on x86 do we need to assert that the stack level at the target block matches the current stack level.
+//
+#ifndef UNIX_X86_ABI
+    assert(tgtBlock->bbTgtStkDepth * sizeof(int) == genStackLevel || isFramePointerUsed());
+#else
+    // bbTgtStkDepth is a (pure) argument count (stack alignment padding should be excluded).
+    assert(tgtBlock->bbTgtStkDepth * sizeof(int) == genStackLevel - curNestedAlignment || isFramePointerUsed());
```
BruceForstall: Three "nits": (1) use CLANG_FORMAT_COMMENT_ANCHOR to defeat the clang-format "bug" pushing comments to the left if an `#if` follows, and (2) parenthesize expressions, (3) prefer `#ifdef` to `#ifndef`. So, I suggest:

```
    // On the x86 we are pushing (and changing the stack level), but on x64 and other archs we have
    // a fixed outgoing args area that we store into and we never change the stack level when calling methods.
    //
    // Thus only on x86 do we need to assert that the stack level at the target block matches the current stack level.
    //
    CLANG_FORMAT_COMMENT_ANCHOR;

#ifdef UNIX_X86_ABI
    // bbTgtStkDepth is a (pure) argument count (stack alignment padding should be excluded).
    unsigned argStackLevel = genStackLevel - curNestedAlignment;
#else // !UNIX_X86_ABI
    unsigned argStackLevel = genStackLevel;
#endif // !UNIX_X86_ABI
    assert((tgtBlock->bbTgtStkDepth * sizeof(int) == argStackLevel) || isFramePointerUsed());
```

src/jit/codegencommon.cpp
parjong: Okay, I'll revert the changes in this file.

src/jit/instr.cpp
parjong: @BruceForstall I'll add CLANG_FORMAT_COMMENT_ANCHOR; to make comment formatting better, but it seems that argStackLevel  will be unused for Release build (Please correct me if there is something wrong).

BruceForstall: You are correct. You can leave the asserts as-is if you wish, but please add more parentheses, as suggested.

parjong:

parjong:
@BruceForstall Updated PR per feedback. Please take a look.

Issue #10931, opened 2017-04-13
[x86/Linux] ReadyToRun TCs fail, again.


ragmani:
```
Program received signal SIGSEGV, Segmentation fault.
0xf6dbfb94 in Module::IsReadyToRun (this=0x3cb800) at /home/jang/DotNet/enable_x86/work/ragmani/coreclr/src/vm/ceeload.h:3053
3053	        return m_pReadyToRunInfo != NULL;
(gdb) bt
#0  0xf6dbfb94 in Module::IsReadyToRun (this=0x3cb800) at /home/jang/DotNet/enable_x86/work/ragmani/coreclr/src/vm/ceeload.h:3053
#1  0xf72040fc in Module::GetNativeOrReadyToRunImage (this=0x3cb800) at /home/jang/DotNet/enable_x86/work/ragmani/coreclr/src/vm/ceeload.cpp:4338
#2  0xf6eef56c in StubDispatchFixupWorker (pTransitionBlock=0xffffc284, siteAddrForRegisterIndirect=4074884992, sectionIndex=0, pModule=0x3cb800)
    at /home/jang/DotNet/enable_x86/work/ragmani/coreclr/src/vm/virtualcallstub.cpp:1234
#3  0xf7189fd4 in StubDispatchFixupStub () at /home/jang/DotNet/enable_x86/work/ragmani/coreclr/src/vm/i386/asmhelpers.S:743
#4  0x10baea55 in ?? ()
#5  0x10bb0afd in ?? ()
#6  0xf7189e39 in CallDescrWorkerInternal () at /home/jang/DotNet/enable_x86/work/ragmani/coreclr/src/vm/i386/asmhelpers.S:304
#7  0xf6f2a18d in CallDescrWorker (pCallDescrData=0xffffcb18) at /home/jang/DotNet/enable_x86/work/ragmani/coreclr/src/vm/callhelpers.cpp:135
#8  0xf6f29f3d in CallDescrWorkerWithHandler (pCallDescrData=0xffffcb18, fCriticalCall=0) at /home/jang/DotNet/enable_x86/work/ragmani/coreclr/src/vm/callhelpers.cpp:78
#9  0xf6f2be13 in MethodDescCallSite::CallTargetWorker (this=0xffffce10, pArguments=0xffffce80, pReturnValue=0xffffcb98, cbReturnValue=8)
    at /home/jang/DotNet/enable_x86/work/ragmani/coreclr/src/vm/callhelpers.cpp:645
[...]

PR #11063, opened 2017-04-19
[x86/Linux] fix a problem that stack was broken by changing stdcall to cdecl in case of readytorun.

ragmani:
add DynamicHelperArgsStub and change jmp to call.

### DIFF
--- a/src/vm/i386/asmhelpers.S
+++ b/src/vm/i386/asmhelpers.S
@@ -815,6 +815,13 @@ PATCH_LABEL ExternalMethodFixupPatchLabel
 NESTED_END ExternalMethodFixupStub, _TEXT
 
 #ifdef FEATURE_READYTORUN
+NESTED_ENTRY DynamicHelperArgsStub, _TEXT, NoHandler
+    CHECK_STACK_ALIGNMENT
+    call  eax
+    add   esp, 12
+    ret
+NESTED_END DynamicHelperArgsStub, _TEXT
+
 // ==========================================================================
 NESTED_ENTRY DelayLoad_MethodCall, _TEXT, NoHandler
     STUB_PROLOG_2_HIDDEN_ARGS
@@ -971,6 +978,7 @@ NESTED_ENTRY DelayLoad_Helper\suffix, _TEXT, NoHandler
     push        eax             // indirection cell address.
     push        esi             // pTransitionBlock
 
+    CHECK_STACK_ALIGNMENT
     call        C_FUNC(DynamicHelperWorker)
     test        eax,eax
     jnz         LOCAL_LABEL(TailCallDelayLoad_Helper\suffix)
--- a/src/vm/i386/asmhelpers.S
+++ b/src/vm/i386/asmhelpers.S
@@ -816,6 +816,7 @@ NESTED_END ExternalMethodFixupStub, _TEXT
 
 #ifdef FEATURE_READYTORUN
 NESTED_ENTRY DynamicHelperArgsStub, _TEXT, NoHandler
+    .cfi_def_cfa_offset 16
     CHECK_STACK_ALIGNMENT
     call  eax
     add   esp, 12
--- a/src/vm/i386/cgenx86.cpp
+++ b/src/vm/i386/cgenx86.cpp
@@ -1878,23 +1878,47 @@ PCODE DynamicHelpers::CreateReturnIndirConst(LoaderAllocator * pAllocator, TADDR
     END_DYNAMIC_HELPER_EMIT();
 }
 
+EXTERN_C VOID DynamicHelperArgsStub();
+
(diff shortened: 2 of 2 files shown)


ragmani:
cc @parjong @janvorli 
This PR from #10931 .

parjong:
coreclr_print_RVA.patch should be removed. Could you remove it f and squish commits?

src/vm/i386/asmhelpers.S
```diff
@@ -815,6 +815,13 @@ PATCH_LABEL ExternalMethodFixupPatchLabel
 NESTED_END ExternalMethodFixupStub, _TEXT
 
 #ifdef FEATURE_READYTORUN
+NESTED_ENTRY DynamicHelperArgsStub, _TEXT, NoHandler
+    CHECK_STACK_ALIGNMENT
+    call  eax
+    add   esp, 12
+    ret
+NESTED_END DynamicHelperArgsStub, _TEXT
```
parjong: As I understand, unwind may failed to unwind this function. The following change will make this function unwindable:
```
mov  [esp + 8], ebp
lea ebp, [esp +8]
CHECK_STACK_ALIGNMENT
call eax
mov esp, ebp
pop ebp
ret
```
@janvorli Could you please take a look?

janvorli: @parjong I agree that to make it unwindable, it requires a special treatment. But it is not necessary to create an EBP frame there and the way you suggested it, it would not work anyways (there is no CFI annotation, so the unwinder would have no idea where to get the EBP). When there are no annotations, the default is the CFA to be ESP + 4, which means the unwinder gets the EIP of the caller frame from [CFA-4], which is [ESP] and then assigns CFA to ESP.
So I would suggest modifying the DynamicHelperArgsStub as follows. The `.cfi_def_cfa_offset 16` tells the unwinder that the CFA at that point in the code is at ESP+16. 
```asm
NESTED_ENTRY DynamicHelperArgsStub, _TEXT, NoHandler
    .cfi_def_cfa_offset 16
    CHECK_STACK_ALIGNMENT
    call  eax
    add   esp, 12
    ret
NESTED_END DynamicHelperArgsStub, _TEXT

```

janvorli: But I wonder - could we instead modify the JIT to generate the call to the dynamic helper as cdecl and cleanup the extra arguments itself? That would remove the need for this stub and it would perform better due to less code and one less call level.

parjong: As I understand, JIT already treats the dynamic helpers as CDECL. The problem comes from something like argument shuffling stub created by ``DynamicHelpers::CreateHelperWithTwoArgs``.

janvorli: @parjong ok, I wasn't sure if we already treat them as cdecl. But what I really meant was that the JIT would know that the stub would add one or two arguments and cleanup the stack after the call to the stub with that knowledge in mind. Since the only purpose of the DynamicHelperArgsStub is to cleanup the extra arguments after the call, the stub would not be needed this way.
However, I don't know how complicated change that would be on the JIT side and if the JIT has a knowledge about these arguments at all.
We could create a github issue for future investigation of this possibility, I would not do that now.

janvorli:

src/vm/i386/asmhelpers.S
parjong: I'm not sure whether it is possible to guarantee 16-byte stack alignment with the suggested approach, but it is worthwhile to consider.

janvorli: @parjong I think that it could be done by pushing placeholders for the additional one or two arguments right in the jitted code and replacing them with the real values in the stub. So the call site e.g. for the case with two additional args would look like this:
```asm
push 0 // placeholder for the arg1
push 0 // placeholder for the arg2
call stub
```
and the stub would then look something like:
```asm
mov dword ptr [rsp+8], arg1
mov dword ptr [rsp+4], arg2
jmp target
```

parjong: As I remember, ``CORINFO_HELP_READYTORUN_DELEGATE_CTOR`` in the following code invokes the stub of interest:
```
G_M65422_IG01:
       55           push     ebp
       8BEC         mov      ebp, esp
       57           push     edi
       56           push     esi

G_M65422_IG02:
       FF1500000000 call     [CORINFO_HELP_READYTORUN_NEW]
       8BF0         mov      esi, eax
       8BCE         mov      ecx, esi
       FF1500000000 call     [CORINFO_HELP_READYTORUN_DELEGATE_CTOR]
```
Is there any detailed description on how readytorun works? We need to change some readytorun ABI in order to implement the suggested approach (``DynamicHelperFixup``  should update caller's frame directly.  Am I right?)

PR #11065, opened 2017-04-19
Stop treating test files as executable and fix encoding

yizhang82:
Turns out we accidentally made some test source code executable and git end up thinking they are executable files. Chmod -x fixed it. Also changing line ending and encoding to UTF-8 (only for one string marshaling test code that has unicode in it) to play nicer with git.


danmosemsft:
@dotnet-bot test tizen armel cross debug build (https://github.com/dotnet/coreclr/issues/11069)

danmosemsft:
LGTM

Issue #11057, opened 2017-04-19
[RyuJIT/ARM32] Implement setting the model kill for CORINFO_HELP_ASSIGN_BYREF helper


sjsinju:
It was not implemented to setting the model kill of CORINFO_HELP_ASSIGN_BYREF helper for ARM arch now on Compiler::compHelperCallKillSet function.

sjsinju:
Merged

PR #10721, opened 2017-04-05
[RyuJIT/ARM32] Implement for GT_STORE_OBJ

sjsinju:
In #10657, I commented that the messages for NYI were printed about GT_STORE_OBJ on running the CodeGenBringUpTests.

'Lowering::LowerBlockStore(GenTreeBlk* blkNode)' method implementation is just copied.
but after lowering phase, in code generation, codegenarm.cpp, below would be run.

- genCodeForCpBlk
- genCodeForInitBlk
- genCodeForCpBlkUnroll
- genCodeForInitBlkUnroll

'genCodeForCpBlk' and 'genCodeForInitBlk' are implemented in ARM/ARM64 by MEMCPY/MEMSET
but 'genCodeForCpBlkUnroll' and 'genCodeForInitBlkUnroll' are not implemented both ARM and ARM64.

Therefore those need to implement.

Related main issue : #8496

PR #11067, opened 2017-04-19
[WIP][RyuJIT/ARM32] Implement the model kill for CORINFO_HELP_ASSGN_BYREF helper

sjsinju:
The CORINFO_HELP_ASSIGN_BYREF helper is used in #10721 PR.
So the assertion of setting the model kill for CORINFO_HELP_ASSIGN_BYREF HELPER would be occurred on running some tests.

Related Main Issue : #8496, #11057

### DIFF
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -631,6 +631,8 @@ regMaskTP Compiler::compHelperCallKillSet(CorInfoHelpFunc helper)
             return RBM_WRITE_BARRIER_SRC_BYREF | RBM_WRITE_BARRIER_DST_BYREF | RBM_CALLEE_TRASH_NOGC;
 #elif defined(_TARGET_X86_)
             return RBM_ESI | RBM_EDI | RBM_ECX;
+#elif defined(_TARGET_ARM_)
+            return RBM_ARG_1 | RBM_ARG_0 | RBM_CALLEE_TRASH_NOGC;
 #else
             NYI("Model kill set for CORINFO_HELP_ASSIGN_BYREF on target arch");
             return RBM_CALLEE_TRASH;


BruceForstall:
In target.h, I see the following for ARM:
```
#ifdef LEGACY_BACKEND
  #define RBM_CALLEE_TRASH_NOGC   (RBM_R2|RBM_R3|RBM_LR)
#else
  #define RBM_CALLEE_TRASH_NOGC   RBM_CALLEE_TRASH
#endif
```
This doesn't make sense to me: the write barrier code in the VM used by the `LEGACY_BACKEND` version of the ARM JIT is exactly the same as the code used by RyuJIT. It seems like RyuJIT should also use the `LEGACY_BACKEND` define.

Also, I see in the VM, in arm\asmhelpers.asm:
```
; The ByRef write barriers have a slightly different interface:
;
; On entry:
;   r0  : the destination address (object reference written here)
;   r1  : the source address (points to object reference to write)
;
; On exit:
;   r0  : incremented by 4
;   r1  : incremented by 4
;   r2  : trashed
;   $__wbscratch : trashed
```
(__wbscratch is r3)

This seems to correspond to the `LEGACY_BACKEND` definition of `RBM_CALLEE_TRASH_NOGC` except I'm not sure why r0/r1 aren't included in that mask of trashed registers.

sjsinju:
@BruceForstall OK. I got it. I'll review it detail again. Thank you!

sjsinju:
@BruceForstall I found similar case in #9681. 
I'm sorry but I'm not good at english. so it seems to confuse to understand. 
I am wondering which is the correct code.
In #9681, below codes are added
```cpp
+#ifdef LEGACY_BACKEND
    #define RBM_CALLEE_TRASH_NOGC   (RBM_R2|RBM_R3|RBM_LR)
+#else
+   #define RBM_CALLEE_TRASH_NOGC   RBM_CALLEE_TRASH
+#endif
```
Do you mean above codes should be reverted? if not, should be my code changed to below likes #9681 code?
```cpp
#elif defined(_TARGET_ARM_)
            return RBM_ARG_1 | RBM_ARG_0 | RBM_CALLEE_TRASH_NOGC;
```
-->
```cpp
#elif defined(_TARGET_ARM_)
            return RBM_CALLEE_TRASH_NOGC;
#else
```
It would be appreciated if you give proper guidances.

sjsinju:
@BruceForstall Can I get your opinion?

BruceForstall:
@sjsinju sorry for the delay; I hope to investigate this soon.

BruceForstall:
After further investigation, I believe the ARM `LEGACY_BACKEND` case is slightly different, due to how the register allocator works. My recommendation in #9681 was too conservative. I think the code in target.h needs to be:
```
#ifdef LEGACY_BACKEND
    #define RBM_CALLEE_TRASH_NOGC   (RBM_R2|RBM_R3|RBM_LR)
#else
   #define RBM_CALLEE_TRASH_NOGC   (RBM_R0|RBM_R1|RBM_R2|RBM_R3|RBM_LR)
#endif
```
That is, R0 and R1 also need to be marked as trashed (since the helper does trash them). I think the `LEGACY_BACKEND` code generation / register allocator must cause that to happen "manually".

There is definitely some confusion in the naming used by the WriteBarrier code for ARM. We do not set NOGC_WRITE_BARRIERS, but we use some "NOGC" names. For ARM, the normal argument registers are used to pass the WriteBarrier arguments (compared to x86, which uses a set of custom calling convention helpers). However, even on ARM, the helper is special, since it does not trash all the normal callee-trash registers.

sjsinju:
@BruceForstall Thank you. 
I got your comment by email but it is not exist here. Did you delete that?
I would wait for your another opinion.

BruceForstall:
Yes, I wrote it, then deleted it as I learned more. There's a lot of confusing code related to this topic that I'm trying to fully understand.

PR #11077, opened 2017-04-19
Spill side-effects in impAssignMultiRegTypeToVar.

### LINKED ISSUE #10940: [osx.10.12-x64/release] Parameters are not evaluated from left to right on OSX in release mode.
- The following code fails on `osx.10.12-x64` runtime in `release` mode but succeeds in `debug` mode.
- It succeeds on `win81-x64` in `release` and `debug` mode
- I didn't check any other runtimes.

```
public static Foo Deserialize(Stream stream)
{
    using (var r = new BinaryReader (stream))
    {
        return new Foo (r.ReadString (), r.ReadDecimal ());
    }
}
```
The method `ReadDecimal` is called before `ReadString`. Surprisingly this happens only with runtime methods. If we use custom methods instead (for example extensions methods which trace the order of execution) , the code works. 

Find attached the [zip file](https://github.com/dotnet/coreclr/files/920145/ParametersEvaluationOrderTest.zip) which contains the `netcoreapp1.1` project that reproduce the issue: I have used the following commands to publish the project:

```
dotnet restore -r osx.10.12-x64 ParametersEvaluationOrderTest.csproj
dotnet publish -r osx.10.12-x64 -c release ParametersEvaluationOrderTest.csproj
```

pgavlin:
This importer function converts IR from the form
```
    tree
```
to
```
            /- tree
    STMT - =
            \- lclVar
```

in order to conform to the JIT's representation of multi-register values.
Today, it does not check whether or not the tree being spilled has any
side-effects that may interfere with the side-effects of trees that are
already on the evaluation stack, which can lead to bad codegen. This
change fixes this function to spill any side-effects on the stack
before generating the temp assign.

Fixes #10940.

### DIFF
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -15051,7 +15051,7 @@ void Compiler::impMarkLclDstNotPromotable(unsigned tmpNum, GenTreePtr src, CORIN
 GenTreePtr Compiler::impAssignMultiRegTypeToVar(GenTreePtr op, CORINFO_CLASS_HANDLE hClass)
 {
     unsigned tmpNum = lvaGrabTemp(true DEBUGARG("Return value temp for multireg return."));
-    impAssignTempGen(tmpNum, op, hClass, (unsigned)CHECK_SPILL_NONE);
+    impAssignTempGen(tmpNum, op, hClass, (unsigned)CHECK_SPILL_ALL);
     GenTreePtr ret = gtNewLclvNode(tmpNum, op->gtType);
 
     // TODO-1stClassStructs: Handle constant propagation and CSE-ing of multireg returns.
--- a/tests/src/JIT/Regression/JitBlue/GitHub_10940/GitHub_10940.cs
+++ b/tests/src/JIT/Regression/JitBlue/GitHub_10940/GitHub_10940.cs
@@ -1,0 +1,58 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+
+// This test checks whether or not the JIT properly spills side effects in the importer when dumping multi-reg values
+// to temps. If the JIT does not do so correctly, the calls to GetString() and GetDecimal() will be reordered and the
+// test will fail with exit code 0; if it does do so correctly, the calls will not be reordered and the test will
+// pass.
+
+class Test
+{
+    abstract class ValueSourceBase
+    {
+        public abstract string GetString();
+        public abstract decimal GetDecimal();
+        public abstract int GetReturnValue();
+    }
+
+    class ValueSource : ValueSourceBase
+    {
+        int rv;
+
+        public override string GetString()
+        {
+            rv = 0;
+            return "";
+        }
+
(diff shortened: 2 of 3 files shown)


pgavlin:
@briansull @CarolEidt @dotnet/jit-contrib PTAL

AndyAyersMS:
Would be a good idea to check that we're not breaking trees in such a way as to reintroduce cases of #9562. 

You should be able to spot the broken opts, if any, from jit-diffs on corelib.

Also, please add a test case.

pgavlin:
> Also, please add a test case.

Yes, I'm working on one.

briansull:
LGTM

pgavlin:
@AndyAyersMS: there is one diff in S.P.CoreLib on Linux and OS X. It is in EventSource:CreateManifestAndDescriptors:

```diff
 G_M43523_IG10:
        488BFB               mov      rdi, rbx
-       E800000000           call     EventSource:GetGuid(ref):struct
-       48898550FFFFFF       mov      qword ptr [rbp-B0H], rax
-       48899558FFFFFF       mov      qword ptr [rbp-A8H], rdx
-       488BFB               mov      rdi, rbx
        8B75D4               mov      esi, dword ptr [rbp-2CH]
        E800000000           call     EventSource:GetName(ref,int):ref
        488985B0FEFFFF       mov      gword ptr [rbp-150H], rax
+       488BFB               mov      rdi, rbx
+       E800000000           call     EventSource:GetGuid(ref):struct
+       48898550FFFFFF       mov      qword ptr [rbp-B0H], rax
+       48899558FFFFFF       mov      qword ptr [rbp-A8H], rdx
        488D3D00000000       lea      rdi, [(reloc)]
        E800000000           call     CORINFO_HELP_NEWSFAST
        488B95B0FEFFFF       mov      rdx, gword ptr [rbp-150H]
        F30F6F8550FFFFFF     movdqu   xmm0, qword ptr [rbp-B0H]
        F30F7F8530FFFFFF     movdqu   qword ptr [rbp-D0H], xmm0
        8B7DD4               mov      edi, dword ptr [rbp-2CH]
        893C24               mov      dword ptr [rsp], edi
        488985A8FEFFFF       mov      gword ptr [rbp-158H], rax
        488BF8               mov      rdi, rax
        488BF2               mov      rsi, rdx
        488B9530FFFFFF       mov      rdx, qword ptr [rbp-D0H]
        488B8D38FFFFFF       mov      rcx, qword ptr [rbp-C8H]
        4D8BC7               mov      r8, r15
        4C8B8D08FFFFFF       mov      r9, gword ptr [rbp-F8H]
        E800000000           call     ManifestBuilder:.ctor(ref,struct,ref,ref,int):this
```

It corresponds to [this source code](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs#L3336-L3337):
```
                manifest = new ManifestBuilder(GetName(eventSourceType, flags), GetGuid(eventSourceType), eventSourceDllName,
                                               resources, flags);
```

So this diff is expected: the baseline JIT was incorrectly reordering the calls to `GetName` and `GetGuid`.

AndyAyersMS:
Thanks for checking.

Changes LGTM too.

Issue #10421, opened 2017-03-23
JIT throughput: noway_assert


BruceForstall:
The JIT has about 3300 noway_assert. These are executed in non-DEBUG (aka, RELEASE) builds. Some might be frequently executed and thus costly. Instead of auditing all of them for relevance (i.e., in an optimization phase that can be backed out of), or apparently importance, we could change the noway_assert macro (conditionally) to collect a count of which ones are frequently executed, using a hash table from preprocessor __FILE__ and __LINE__ to execution count, dumped at the end of compilation. Then, we could convert the worst ones to simple asserts.

BruceForstall:
I implemented this. I ran "a lot" of the test tree, processed the data, and saw these as the top 75 dynamic occurrences of noway_asserts (some might actually get optimized away in release builds, as there are a few static constant asserts here). The columns are: total count, filename, line number, assertion text.
```
181470679, e:\gh\coreclr2\src\jit\morph.cpp, 14648, tree
180436108, e:\gh\coreclr2\src\jit\morph.cpp, 14649, tree->gtOper != GT_STMT
83096677, e:\gh\coreclr2\src\jit\assertionprop.cpp, 808, assertIndex <= optAssertionCount
83096677, e:\gh\coreclr2\src\jit\assertionprop.cpp, 807, assertIndex != NO_ASSERTION_INDEX
72024047, e:\gh\coreclr2\src\jit\liveness.cpp, 1632, lclNum < lvaCount
43294958, e:\gh\coreclr2\src\jit\lsra.cpp, 3003, count < MaxInternalRegisters
40851620, e:\gh\coreclr2\src\jit\lclvars.cpp, 2510, varNum < lvaCount
35418942, e:\gh\coreclr2\src\jit\morph.cpp, 10632, tree->OperKind() & GTK_SMPOP
33982229, e:\gh\coreclr2\src\jit\flowgraph.cpp, 5228, opcode < CEE_COUNT
32541990, e:\gh\coreclr2\src\jit\morph.cpp, 11463, tree->gtOper != GT_CALL
32533197, e:\gh\coreclr2\src\jit\morph.cpp, 13347, oper == tree->gtOper
30050847, e:\gh\coreclr2\src\jit\flowgraph.cpp, 16328, !(block->bbFlags & BBF_REMOVED)
27085069, e:\gh\coreclr2\src\jit\morph.cpp, 15542, stmt->gtOper == GT_STMT
[...]

PR #11080, opened 2017-04-19
Add code to measure dynamic noway_assert usage

BruceForstall:
Code is under `#if MEASURE_NOWAY` which is disabled by default.

Related to #10421

### DIFF
--- a/src/jit/compiler.cpp
+++ b/src/jit/compiler.cpp
@@ -1134,6 +1134,170 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
     return useType;
 }
 
+///////////////////////////////////////////////////////////////////////////////
+//
+// MEASURE_NOWAY: code to measure and rank dynamic occurences of noway_assert.
+// (Just the appearances of noway_assert, whether the assert is true or false.)
+// This might help characterize the cost of noway_assert in non-DEBUG builds,
+// or determine which noway_assert should be simple DEBUG-only asserts.
+//
+///////////////////////////////////////////////////////////////////////////////
+
+#if MEASURE_NOWAY
+
+struct FileLine
+{
+    char*    m_file;
+    unsigned m_line;
+    char*    m_condStr;
+
+    FileLine() : m_file(nullptr), m_line(0), m_condStr(nullptr)
+    {
+    }
+
+    FileLine(const char* file, unsigned line, const char* condStr) : m_line(line)
+    {
+        size_t newSize = (strlen(file) + 1) * sizeof(char);
+        m_file         = (char*)HostAllocator::getHostAllocator()->Alloc(newSize);
+        strcpy_s(m_file, newSize, file);
+
+        newSize   = (strlen(condStr) + 1) * sizeof(char);
+        m_condStr = (char*)HostAllocator::getHostAllocator()->Alloc(newSize);
+        strcpy_s(m_condStr, newSize, condStr);
+    }
+
+    FileLine(const FileLine& other)
+    {
+        m_file    = other.m_file;
+        m_line    = other.m_line;
+        m_condStr = other.m_condStr;
+    }
+
+    // GetHashCode() and Equals() are needed by SimplerHashTable
+
+    static unsigned GetHashCode(FileLine fl)
+    {
+        assert(fl.m_file != nullptr);
+        unsigned code = fl.m_line;
+        for (const char* p = fl.m_file; *p != '\0'; p++)
+        {
+            code += *p;
+        }
+        // Could also add condStr.
+        return code;
+    }
+
+    static bool Equals(FileLine fl1, FileLine fl2)
+    {
+        return (fl1.m_line == fl2.m_line) && (0 == strcmp(fl1.m_file, fl2.m_file));
+    }
+};
+
+typedef SimplerHashTable<FileLine, FileLine, size_t, JitSimplerHashBehavior> FileLineToCountMap;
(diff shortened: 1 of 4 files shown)


BruceForstall:
@pgavlin PTAL
cc @dotnet/jit-contrib

src/jit/error.h
```diff
@@ -80,7 +80,21 @@ extern void noWayAssertBodyConditional(
     );
 extern void noWayAssertBodyConditional(const char* cond, const char* file, unsigned line);
 
+// Define MEASURE_NOWAY to 1 to enable code to count and rank individual noway_assert calls by occurrence.
+// These asserts would be dynamically executed, but not necessarily fail. The provides some insight into
```
CarolEidt: This key point (that you're measuring how many are executed, not how many fail) was one I missed at first. It might be useful to mention that in the `MEASURE_NOWAY` section in compiler.cpp as well.

briansull:
LGTM

src/jit/compiler.cpp
```diff
@@ -1134,6 +1134,169 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
...
+            nacp[i].count = iter.GetValue();
+            nacp[i].fl    = iter.Get();
+            ++i;
+        }
+
+        qsort(nacp, count, sizeof(nacp[0]), NowayAssertCountMap::compare);
+
+        FILE* fout = jitstdout;
+
+        LPCWSTR strJitNowayAssertFile = JitConfig.JitNowayAssertFile();
```
pgavlin: Nit: how about `jitNowayAssertFilePath`?

src/jit/compiler.cpp
```diff
@@ -1134,6 +1134,169 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
...
+        if (strJitNowayAssertFile != nullptr)
+        {
+            FILE* file_out = _wfopen(strJitNowayAssertFile, W("a"));
+            if (file_out == nullptr)
+            {
+                fprintf(fout, "Failed to open JitFuncInfoLogFile");
```
pgavlin: s/`JitFuncInfoLogFile`/`JitNowayAssertFile`/

pgavlin: Also, it is a bit weird that we just carry on and dump to `jitstdout` if we can't open the log file.

src/jit/compiler.cpp
```diff
@@ -1134,6 +1134,169 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
...
+{
+    if (NowayAssertMap != nullptr)
+    {
+        // Iterate noway assert map, create sorted table by occurrence, dump it.
+        unsigned             count = NowayAssertMap->GetCount();
+        NowayAssertCountMap* nacp  = new NowayAssertCountMap[count];
+        unsigned             i     = 0;
+
+        for (FileLineToCountMap::KeyIterator iter = NowayAssertMap->Begin(), end = NowayAssertMap->End();
+             !iter.Equal(end); iter++)
```
pgavlin: `++iter`?

src/jit/compiler.cpp
```diff
@@ -1134,6 +1134,169 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
...
+
+    // GetHashCode() and Equals() are needed by SimplerHashTable
+
+    static unsigned GetHashCode(FileLine fl)
+    {
+        assert(fl.m_file != nullptr);
+        unsigned code = fl.m_line;
+        for (const char* p = fl.m_file; *p != '\0'; p++)
+        {
+            code += *p;
```
pgavlin: Nit: this doesn't really matter too much, but we could use a slightly better hash function here (derived from djb):
```
unsigned code = 5381 + fl.m_line;
for (unsigned i = 0; fl.m_file[i] != '\0'; i++)
{
    code = ((code << 5) + code) ^ fl.m_file[i];
}
```

src/jit/compiler.cpp
```diff
@@ -1134,6 +1134,169 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
...
+        }
+        // Could also add condStr.
+        return code;
+    }
+
+    static bool Equals(FileLine fl1, FileLine fl2)
+    {
+        return (0 == strcmp(fl1.m_file, fl2.m_file)) && (fl1.m_line == fl2.m_line);
```
pgavlin: Nit: might as well flip the order of operations here and save a little time.

src/jit/compiler.cpp
BruceForstall: fixed

src/jit/compiler.cpp
BruceForstall: that's a bit awkward, plus we already have JitFuncInfoLogFile as a model.

src/jit/compiler.cpp
BruceForstall: done

BruceForstall:
> Can we define MEASURE_NOWAY to 1 in debug/checked builds and add a config value to enable/disable its output? It'd be nice to be able to access this functionality in such builds without rebuilding.

Done.

> It would be nice to be able to get a per-method count of noway_asserts executed during compilation that is reported as part of the profile CSV.

Left as an exercise for the reader. (Would be easy to add.)

PR #11083, opened 2017-04-19
Update CheckProjects to not fail for non-dev CORE_ROOT

### LINKED ISSUE #10635: CheckProjects is fragile to running with cross targets.
CheckProjects assumes the tests were built in a CoreCLR repo. For cross targets such as Arm32, Arm64 this may not be correct causing the test to fail.

Note this bug is mostly just to add an issue tag to disable the test on Arm64.

### LINKED ISSUE #11079: JIT/CheckProjects test requires that Core_Root be relative to repo root
https://github.com/dotnet/coreclr/blob/master/tests/src/JIT/CheckProjects/CheckProjects.cs#L51-L62

This condition isn't met in Helix, causing this test to fail there - I don't think we should require that Core_Root be in a location relative to the repo root for a test to execute.

CC @AndyAyersMS @RussKeldorph

AndyAyersMS:
CheckProjects expects to be able to find the jit test project files
relative to the value of the `CORE_ROOT` environment variable. This
fails when we have test setups like cross-tests or helix where the
test binary tree has been copied away from the repo after building.

Tolerate these cases by returning a success code if the `CORE_ROOT`
path is not pointing at the expected place.

Fixes #11079.
Fixes #10635.

### DIFF
--- a/tests/src/JIT/CheckProjects/CheckProjects.cs
+++ b/tests/src/JIT/CheckProjects/CheckProjects.cs
@@ -52,9 +52,9 @@ internal class ScanProjectFiles
 
             if (binIndex < 0)
             {
-                Console.WriteLine("CORE_ROOT must be set to full path to repo test dir; was '{0}'.",
-                    coreRoot);
-                return -1;
+                Console.WriteLine("No bin directory found in CORE_ROOT path `{0}`," + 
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@jashook PTAL
cc @dotnet/jit-contrib

PR #11085, opened 2017-04-19
Skip failing tests in Helix

wtgodbe:


tests/issues.targets
```diff
@@ -409,7 +409,7 @@
 
     <!-- The following are tests that fail on non-Windows, which we must not run when building against packages -->
 
-    <ItemGroup Condition="'$(XunitTestBinBase)' != '' and '$(BuildTestsAgainstPackages)' == 'true' and '$(RuntimeID)' != ''">
+    <ItemGroup Condition="'$(XunitTestBinBase)' != '' and '$(BuildTestsAgainstPackages)' == 'true' and '$(TargetsWindows)' != 'true'">
```
gkhanna79: So, we never specified RuntimeID?

wtgodbe: We do specify RuntimeID, it's just not available here anymore. Previously, if a runtimeID was passed, that meant we were building tests for non-windows; now we always pass a runtimeID when building against packages, and have an explicit variable (TargetsWindows) that determines if we're building tests for Windows or non-Windows - I just missed updating this file when I made that change.

PR #11091, opened 2017-04-19
Refactoring: move some BasicBlock functions to block.cpp

BruceForstall:
(where they belong)

Delete unused containsStatement().


BruceForstall:
@briansull @dotnet/jit-contrib PTAL

PR #11092, opened 2017-04-19
[Local GC] Ensure that handle creation returns null on failure instead of throwing

swgillespie:
`HndCreateHandle` today throws an `OutOfMemoryException` when it fails to allocate a new handle. It's inappropriate for anything on the GC side of the interface to throw exceptions, and particularly OOM exceptions, since those are both EE-specific details. This PR addresses this problem to do what Redhawk does and return null when a handle can't be created.

The change to `HndCreateHandle` itself is simple, since CoreRT/ProjectN already return null on allocation failure. The rest of this PR addresses places that expect `HndCreateHandle` to throw instead of returning null:

1. `TableRelocateAsyncPinHandles` and `SegmentRelocateAsyncPinHandles` - `TableRelocateAsyncPinHandles` tries to catch the OOM exception that `HndCreateHandle` throws (called by `SegmentRelocateAsyncPinHandles`. To remedy this, `SegmentRelocateAsyncPinHandles` checks `HndCreateHandle`'s return and, if null, returns `false` to `TableRelocateAsyncPinHandles`, which it interprets as an OOM in the same way that it did before (by catching the exception).
2. All calls of `IGCHandleManager::CreateHandleOfType` from the EE must check whether its return value is null and throw an OOM if so. Luckily, all calls to this API are routed through helpers in `gchandleutilities.h`, so this check only has to be done in a limited number of places. I did find one handle creation scenario (in `appdomain.hpp` that didn't use the common helpers, so I changed it to use the helpers for the sake of uniformity.

This PR also modifies `Ref_InitializeHandleTableBucket` to not throw. `Ref_InitializeHandleTableBucket` already returns false on failure.

cc @adityamandaleeka @Maoni0 @sergiy-k @jkotas

### DIFF
--- a/src/gc/objecthandle.cpp
+++ b/src/gc/objecthandle.cpp
@@ -740,9 +740,9 @@ bool Ref_InitializeHandleTableBucket(HandleTableBucket* bucket, void* context)
 {
     CONTRACTL
     {
-        THROWS;
+        NOTHROW;
         WRAPPER(GC_TRIGGERS);
-        INJECT_FAULT(COMPlusThrowOM());
+        INJECT_FAULT(return false);
     }
     CONTRACTL_END;
 
@@ -759,13 +759,18 @@ bool Ref_InitializeHandleTableBucket(HandleTableBucket* bucket, void* context)
 
     HandleTableBucketHolder bucketHolder(result, n_slots);
 
-    result->pTable = new HHANDLETABLE[n_slots];
+    result->pTable = new (nothrow) HHANDLETABLE[n_slots];
+    if (!result->pTable)
+    {
+        return false;
+    }
+
     ZeroMemory(result->pTable, n_slots * sizeof(HHANDLETABLE));
 
     for (int uCPUindex=0; uCPUindex < n_slots; uCPUindex++) {
         result->pTable[uCPUindex] = HndCreateHandleTable(s_rgTypeFlags, _countof(s_rgTypeFlags), ADIndex((DWORD)(uintptr_t)context));
         if (!result->pTable[uCPUindex])
-            COMPlusThrowOM();
+            return false;
     }
 
     for (;;) {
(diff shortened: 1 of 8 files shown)


src/vm/gchandleutilities.h
```diff
@@ -63,95 +63,132 @@ inline BOOL ObjectHandleIsNull(OBJECTHANDLE handle)
 #ifndef DACCESS_COMPILE
 
 // Handle creation convenience functions
+inline OBJECTHANDLE CreateHandleCommon(IGCHandleStore* store, Object* object, int type)
```
jkotas: Nit: I think it would look a bit better to have the OOM check duplicated instead of having yet another layer.

swgillespie: Ok!

src/gc/objecthandle.cpp
```diff
@@ -792,9 +797,18 @@ bool Ref_InitializeHandleTableBucket(HandleTableBucket* bucket, void* context)
...
+        newMap = new (nothrow) HandleTableMap;
+        if (!newMap)
+        {
+            return false;
+        }
+
+        newMap->pBuckets = new (nothrow) HandleTableBucket * [ INITIAL_HANDLE_TABLE_ARRAY_SIZE ];
+        if (!newMap->pBuckets)
+        {
+            return false;
```
Maoni0: >   return false; [](start = 9, length = 16)

This is not exactly a new problem but if it failed to allocate memory you are not cleaning up the memory that was allocated above. It had the same problem exactly when it throws OOM often the program just terminates.

swgillespie: I thought this is what `HandleTableBucketHolder` does [here](https://github.com/dotnet/coreclr/blob/master/src/gc/objecthandle.cpp#L609)? The only control flow I see that suppresses the release of that holder is the one that returns true.

Maoni0: ahh, you are right of course - I did not notice the holder code above. As a side note - while looking at the usage of this, I noticed that CreateHandleStore is the one that calls this and returns nullptr if this returns false; and CreateHandleStore is called in AppDomain::Init which seems to expect things to throw OOM instead of handling nullptr returns but since CreateHandleStore doesn't throw and just returns nullptr, we should throw OOM there instead.

swgillespie: ah, you're right - will do!

swgillespie:
thanks!

Issue #6838, opened 2016-08-22
Unittest `Interop/StringMarshalling/LPSTR`


wateret:
This test always returns PASS even if it actually failed.
I was going to fix it by myself but the file `src/Interop/StringMarshalling/LPSTR/LPSTRTest.cs` is UTF-16 encoded and git thinks it is a binary file. So the changes are not shown on github. (Here is the changes. https://github.com/wateret/coreclr/commit/c4ff2e26a40e05b46b0f924d973ad11a18e815b3)
Is this file supposed to be in UTF-16?

/cc @tijoytom - the author of this file.

gkhanna79:
CC @yizhang82

danmosemsft:
@yizhang82 @tijoytom what is the importance for 2.0 & will it happen for 2.0? If not can you please move to Future.

yizhang82:
Thanks @wateret for looking into it! I just created a PR out to make git treat those files properly (and fix the encoding) https://github.com/dotnet/coreclr/pull/11065. Once it's merged, @wateret please feel free to submit a new PR (if you are still interested to make the fix). Thanks!
@danmosemsft Moving to future.

yizhang82:
Addressed by https://github.com/dotnet/coreclr/pull/11096

PR #11096, opened 2017-04-20
Fix unittest `LPSTRTest`

wateret:
The test case Interop.StringMarshalling.LPSTR.LPSTRTest
always returned exit code 100(pass) even if it actually failed.
This implements `ExitTest()` in the same manner with `LPTSTRTest`.

### DIFF
--- a/tests/src/Interop/StringMarshalling/LPSTR/LPSTRTestNative.cpp
+++ b/tests/src/Interop/StringMarshalling/LPSTR/LPSTRTestNative.cpp
@@ -123,16 +123,16 @@ extern "C" DLL_EXPORT int __cdecl Writeline(char * pFormat, int i, char c, doubl
 }
 
 
-typedef LPCTSTR (__stdcall * Test_DelMarshal_InOut)(/*[in]*/ LPCSTR s);
+typedef LPCWSTR (__stdcall * Test_DelMarshal_InOut)(/*[in]*/ LPCSTR s);
 extern "C" DLL_EXPORT BOOL __cdecl RPinvoke_DelMarshal_InOut(Test_DelMarshal_InOut d, /*[in]*/ LPCSTR s)
 {
-    LPCTSTR str = d(s);
-    const char *ret = "Return";    
+    LPCWSTR str = d(s);
+    LPCWSTR ret = W("Return");    
 
-    size_t lenstr = _tcslen(str);
-    size_t lenret = _tcslen(ret);
+    size_t lenstr = wcslen(str);
+    size_t lenret = wcslen(ret);
 
-    if((lenret != lenstr)||(_tcsncmp(str,ret,lenstr)!=0))
+    if((lenret != lenstr)||(wcsncmp(str,ret,lenstr)!=0))
     {
         printf("Error in RPinvoke_DelMarshal_InOut, Returned value didn't match\n");
         return FALSE;
--- a/tests/src/Interop/StringMarshalling/LPSTR/LPSTRTest.cs
+++ b/tests/src/Interop/StringMarshalling/LPSTR/LPSTRTest.cs
@@ -31,6 +31,25 @@ class Test
     }
     #endregion
 
+    #region "Helper"
+    // ************************************************************
+    // Returns the appropriate exit code
+    // *************************************************************
+    static int ExitTest()
+    {
+        if (fails == 0)
+        {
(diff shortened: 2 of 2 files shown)


wateret:
Related Issue : #6838

yizhang82:
@wateret Thanks. Did you re-base your change with latest master? I've already merged my PR - with that git should be able to treat those files as text and show diff properly. I still couldn't see your diff, though. 

Also, why is your change converting it back to UTF-16? My PR makes it UTF-8 (among other things, such as fixing the CR/LF and removing executable attributes) which Git should understands properly. Maybe that's why now git couldn't show the diff anymore?

wateret:
@yizhang82 Yes I rebased to current master. First I tried my changes but the CI fails for this test case(https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_windows_nt_prtest/5880/consoleFull#5992872172d31e50d-1517-49fc-92b3-2ca637122019).

Then I thought there must have been a reason that the source file is written in UTF16 so I was checking whether it works if converting it back to UTF16.

wateret:
However now it seems like that it fails TC build due to a file encoding problem.

wateret:
There is a string literal `"ň"` in this test case. Would the file encoding of source code have something to do with?

wateret:
@yizhang82 You may take a look at my changes if you look at my first commit(https://github.com/dotnet/coreclr/pull/11096/commits/d9b923fd805925c4cafc6e64bf7c4a625e9fc1e7).

yizhang82:
@wateret I think the test has a bug that got exposed by your change. It incorrectly uses char for LPTSTR which it should be using W("return"), wcslen/wcscmp instead. See https://github.com/dotnet/coreclr/blob/master/tests/src/Interop/StringMarshalling/LPSTR/LPSTRTestNative.cpp#L127-L144. Let me know if you want to take a stab at it. Otherwise I can fix it.

wateret:
@yizhang82 Thank you. I updated the commit and looks like it works but I'm not sure if it is what you meant, specifically this part: 

```
LPCWSTR str = (LPCWSTR) d(s);
```

which looks unnatural to me.

tests/src/Interop/StringMarshalling/LPSTR/LPSTRTestNative.cpp
```diff
@@ -126,13 +126,13 @@ extern "C" DLL_EXPORT int __cdecl Writeline(char * pFormat, int i, char c, doubl
 typedef LPCTSTR (__stdcall * Test_DelMarshal_InOut)(/*[in]*/ LPCSTR s);
 extern "C" DLL_EXPORT BOOL __cdecl RPinvoke_DelMarshal_InOut(Test_DelMarshal_InOut d, /*[in]*/ LPCSTR s)
 {
-    LPCTSTR str = d(s);
-    const char *ret = "Return";    
+    LPCWSTR str = (LPCWSTR) d(s);
```
yizhang82: Don't do the cast. Instead, change `Test_DelMarshal_InOut` to return LPCWSTR instead.

yizhang82:
@wateret Please see my comments. Just change the function pointer return type instead and you can avoid the cast. Looks good otherwise. Thanks for fixing this!

wateret:
@yizhang82 Updated as your comment and test done on my local machine. However I still wonder why this works.

When I look at `PinvokeDef.cs`,  the function `DelMarshal_InOut()` returns a `LPTStr`.

```cs
    [UnmanagedFunctionPointer(CallingConvention.StdCall, CharSet = CharSet.Unicode)]
    [return: MarshalAs(UnmanagedType.LPTStr)]
    public delegate string DelMarshal_InOut([MarshalAs(UnmanagedType.LPTStr)][In, Out]string s);
```

Could you please explain what guarantees `LPTStr` maps to `LPCWSTR`(not `LPCSTR`)? My guess is `CharSet = CharSet.Unicode` implies that.

yizhang82:
@wateret See https://github.com/dotnet/coreclr/blob/master/src/vm/mlinfo.cpp#L2136-L2149. These days, we don't support Win9X, and LPTSTR = UTF-16 regardless of CharSet setting.

wateret:
@yizhang82 Thank you for your answer :)

janvorli:
@yizhang82 can this be merged?

PR #11097, opened 2017-04-20
Enable corerun to select Globalization Invariant mode

seanshpark:
This will enable to choose Globalization Invariant mode from corerun.

For ports that don't have dotnet program, x86/Linux for example, we can only run with corerun.
As CoreFX `Invariant.Tests` requires to run in Globalization Invariant mode,  this can be possible with this change.
To turn it on, use environment variable as `CORECLR_GLOBAL_INVARIANT=1`
For example,
```
CORECLR_GLOBAL_INVARIANT=1 ./corerun xunit.console.netcore.exe Invariant.Tests.dll 
```

### DIFF
--- a/src/coreclr/hosts/unixcoreruncommon/coreruncommon.cpp
+++ b/src/coreclr/hosts/unixcoreruncommon/coreruncommon.cpp
@@ -272,7 +272,7 @@ const char* GetEnvValueBoolean(const char* envVariable)
         envValue = "0";
     }
     // CoreCLR expects strings "true" and "false" instead of "1" and "0".
-    return std::strcmp(envValue, "1") == 0 ? "true" : "false";
+    return (std::strcmp(envValue, "1") == 0 || strcasecmp(envValue, "true") == 0) ? "true" : "false";
 }
 
 int ExecuteManagedAssembly(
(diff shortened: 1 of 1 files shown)


seanshpark:
CC @parjong

src/coreclr/hosts/unixcoreruncommon/coreruncommon.cpp
```diff
@@ -260,6 +264,17 @@ void AddFilesFromDirectoryToTpaList(const char* directory, std::string& tpaList)
...
 
+const char* GetEnvValueBoolean(const char* envVariable)
+{
+    const char* envValue = std::getenv(envVariable);
+    if (envValue == nullptr)
+    {
+        envValue = "0";
+    }
+    // CoreCLR expects strings "true" and "false" instead of "1" and "0".
+    return std::strcmp(envValue, "1") == 0 ? "true" : "false";
```
tarekgh: can the value be set to "true" too? I mean should we test for "true"?

tarekgh:
other than my comment, LGTM

src/coreclr/hosts/unixcoreruncommon/coreruncommon.cpp
seanshpark: Yes, that would be better.

PR #8227, opened 2016-11-21
Fast path ExecutionContext.Default for AsyncLocal

benaadams:
Move back to Default ExecutionContext when AsyncLocals are cleared

Early exit if setting value to null on DefaultContext

Return `null` for get if context is default without checking.

/cc @stephentoub @jkotas

PR #11100, opened 2017-04-20
Faster path default EC

benaadams:
Trying again, follow up to https://github.com/dotnet/coreclr/pull/11095

Puts the Threadpool threads on the default context.

Less churn when moving between Default contexts (`ExecutionContext.RunDefaultContext`); with lighter context switching.

Also allows the opportunity to benefit from [Finally Cloning](https://github.com/dotnet/coreclr/blob/master/Documentation/design-docs/finally-optimizations.md#finally-cloning) by the Jit in `RunDefaultContext`.

Resolves https://github.com/dotnet/coreclr/issues/11126

### DIFF
--- a/src/mscorlib/shared/System/Threading/ExecutionContext.cs
+++ b/src/mscorlib/shared/System/Threading/ExecutionContext.cs
@@ -70,6 +70,26 @@ namespace System.Threading
                 ExecutionContext.RestoreDefault(currentThread, defaultContext);
             }
         }
+
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static void SetDefaults()
+        {
+            var currentThread = Thread.CurrentThread;
+            var defaultContext = ExecutionContext.Default;
+
+            // The common case is that these have not changed, 
+            // so avoid the GC memory barrier cost of a write if not needed.
+            if (currentThread.SynchronizationContext != null)
+            {
+                currentThread.SynchronizationContext = null;
+            }
+
+            if (currentThread.ExecutionContext != defaultContext)
+            {
+                currentThread.ExecutionContext = defaultContext;
+            }
+        }
     }
 
     [Serializable]
--- a/src/mscorlib/shared/System/Threading/ExecutionContext.cs
+++ b/src/mscorlib/shared/System/Threading/ExecutionContext.cs
@@ -71,7 +71,6 @@ namespace System.Threading
             }
         }
 
-
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         internal static void SetDefaults()
         {
@@ -130,6 +129,7 @@ namespace System.Threading
         {
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static ExecutionContext Capture()
         {
             ExecutionContext executionContext = Thread.CurrentThread.ExecutionContext;
--- a/src/mscorlib/shared/System/Threading/ExecutionContext.cs
+++ b/src/mscorlib/shared/System/Threading/ExecutionContext.cs
@@ -13,6 +13,7 @@
 
 using System.Diagnostics;
 using System.Diagnostics.Contracts;
+using System.Runtime.CompilerServices;
 using System.Runtime.ExceptionServices;
 using System.Runtime.Serialization;
 
@@ -40,6 +41,33 @@ namespace System.Threading
             if (currentThread.ExecutionContext != m_ec)
             {
                 ExecutionContext.Restore(currentThread, m_ec);
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static bool CurrentIsDefault(Thread currentThread, ExecutionContext defaultContext)
+        {
+            Debug.Assert(currentThread == Thread.CurrentThread);
+            Debug.Assert(defaultContext == ExecutionContext.Default);
+
+            return (currentThread.ExecutionContext == defaultContext && currentThread.SynchronizationContext == null);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static void UndoToDefault(Thread currentThread, ExecutionContext defaultContext)
+        {
+            Debug.Assert(currentThread == Thread.CurrentThread);
+            Debug.Assert(defaultContext == ExecutionContext.Default);
+
+            // The common case is that these have not changed, so avoid the cost of a write if not needed.
+            if (currentThread.SynchronizationContext != null)
+            {
+                currentThread.SynchronizationContext = null;
+            }
+
+            if (currentThread.ExecutionContext != defaultContext)
+            {
+                ExecutionContext.RestoreDefault(currentThread, defaultContext);
             }
         }
     }
@@ -165,16 +193,54 @@ namespace System.Threading
             ecsw.Undo(currentThread);
         }
 
+        internal static void RunDefaultContext(ContextCallback callback, Object state)
+        {
+            ExecutionContext defaultContext = Default;
+            Thread currentThread = Thread.CurrentThread;
+            if (ExecutionContextSwitcher.CurrentIsDefault(currentThread, defaultContext))
+            {
+                try
+                {
+                    callback(state);
+                }
+                finally
+                {
+                    ExecutionContextSwitcher.UndoToDefault(currentThread, defaultContext);
+                }
+            }
+            else
+            {
+                Run(defaultContext, callback, state);
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        internal static void RestoreDefault(Thread currentThread, ExecutionContext defaultContext)
+        {
+            Debug.Assert(currentThread == Thread.CurrentThread);
+            Debug.Assert(defaultContext == ExecutionContext.Default);
+
+            ExecutionContext previous = currentThread.ExecutionContext;
+            currentThread.ExecutionContext = defaultContext;
+
+            // For the purposes of dealing with context change, null counts as the default EC
+            if (previous != null)
+            {
+                OnContextChanged(previous, defaultContext);
+            }
+        }
+
         internal static void Restore(Thread currentThread, ExecutionContext executionContext)
         {
             Debug.Assert(currentThread == Thread.CurrentThread);
 
-            ExecutionContext previous = currentThread.ExecutionContext ?? Default;
+            ExecutionContext defaultContext = Default;
+            ExecutionContext previous = currentThread.ExecutionContext ?? defaultContext;
             currentThread.ExecutionContext = executionContext;
 
             // New EC could be null if that's what ECS.Undo saved off.
             // For the purposes of dealing with context change, treat this as the default EC
-            executionContext = executionContext ?? Default;
+            executionContext = executionContext ?? defaultContext;
 
             if (previous != executionContext)
             {
--- a/src/mscorlib/shared/System/Threading/ExecutionContext.cs
+++ b/src/mscorlib/shared/System/Threading/ExecutionContext.cs
@@ -30,9 +30,11 @@ namespace System.Threading
 
         internal void Undo(Thread currentThread)
         {
+            // Current thread passed as parameter to avoid extern native thread lookup call
             Debug.Assert(currentThread == Thread.CurrentThread);
 
-            // The common case is that these have not changed, so avoid the cost of a write if not needed.
+            // The common case is that these have not changed, so avoid the GC memory barrier 
+            // cost of an reference write if not needed.
             if (currentThread.SynchronizationContext != m_sc)
             {
                 currentThread.SynchronizationContext = m_sc;
@@ -47,7 +49,9 @@ namespace System.Threading
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         internal static bool CurrentIsDefault(Thread currentThread, ExecutionContext defaultContext)
         {
+            // Current thread passed as parameter to avoid extern current native thread lookup call
             Debug.Assert(currentThread == Thread.CurrentThread);
+            // Default context passed as parameter to avoid static inializer checks in NGen'd code
             Debug.Assert(defaultContext == ExecutionContext.Default);
 
             return (currentThread.ExecutionContext == defaultContext && currentThread.SynchronizationContext == null);
@@ -56,10 +60,13 @@ namespace System.Threading
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         internal static void UndoToDefault(Thread currentThread, ExecutionContext defaultContext)
         {
+            // Current thread passed as parameter to avoid extern current native thread lookup call
             Debug.Assert(currentThread == Thread.CurrentThread);
+            // Default context passed as parameter to avoid static inializer checks in NGen'd code
             Debug.Assert(defaultContext == ExecutionContext.Default);
 
-            // The common case is that these have not changed, so avoid the cost of a write if not needed.
+            // The common case is that these have not changed, so avoid the GC memory barrier 
+            // cost of an reference write if not needed.
             if (currentThread.SynchronizationContext != null)
             {
                 currentThread.SynchronizationContext = null;
@@ -67,6 +74,7 @@ namespace System.Threading
 
             if (currentThread.ExecutionContext != defaultContext)
             {
+                // Restoring from a non-default context, need to check for AsyncLocal changes
                 ExecutionContext.RestoreDefault(currentThread, defaultContext);
             }
         }
@@ -74,11 +82,12 @@ namespace System.Threading
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         internal static void SetDefaults()
         {
+            // Current thread take to local to minimize extern current native thread lookup call
             var currentThread = Thread.CurrentThread;
             var defaultContext = ExecutionContext.Default;
 
-            // The common case is that these have not changed, 
-            // so avoid the GC memory barrier cost of a write if not needed.
+            // The common case is that these have not changed, so avoid the GC memory barrier 
+            // cost of an reference write if not needed.
             if (currentThread.SynchronizationContext != null)
             {
                 currentThread.SynchronizationContext = null;
@@ -154,6 +163,7 @@ namespace System.Threading
 
         public static AsyncFlowControl SuppressFlow()
         {
+            // Current thread take to local to minimize extern current native thread lookup call
             Thread currentThread = Thread.CurrentThread;
             ExecutionContext executionContext = currentThread.ExecutionContext ?? Default;
             if (executionContext.m_isFlowSuppressed)
@@ -171,6 +181,7 @@ namespace System.Threading
 
         public static void RestoreFlow()
         {
+            // Current thread take to local to minimize extern current native thread lookup call
             Thread currentThread = Thread.CurrentThread;
             ExecutionContext executionContext = currentThread.ExecutionContext;
             if (executionContext == null || !executionContext.m_isFlowSuppressed)
@@ -193,11 +204,13 @@ namespace System.Threading
             if (executionContext == null)
                 throw new InvalidOperationException(SR.InvalidOperation_NullContext);
 
+            // Current thread take to local to minimize extern current native thread lookup call
             Thread currentThread = Thread.CurrentThread;
             ExecutionContextSwitcher ecsw = default(ExecutionContextSwitcher);
             try
             {
                 EstablishCopyOnWriteScope(currentThread, ref ecsw);
+                // Restore may throw, so need try+catch rather than try+finally, explanation below
                 ExecutionContext.Restore(currentThread, executionContext);
                 callback(state);
             }
@@ -215,8 +228,11 @@ namespace System.Threading
 
         internal static void RunDefaultContext(ContextCallback callback, Object state)
         {
+            // Local copies used and passed to minimise static inializer checks in NGen'd code
+            // and extern current native thread lookup calls
             ExecutionContext defaultContext = Default;
             Thread currentThread = Thread.CurrentThread;
+            // Fastest path for moving from default context to default context
             if (ExecutionContextSwitcher.CurrentIsDefault(currentThread, defaultContext))
             {
                 try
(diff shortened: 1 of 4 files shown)


benaadams:
Most of the Kestrel on libuv take this path

Before (call count underlined)


After (call count underlined)


The 18 of the other 22 are timer callbacks; which work off timer thread or do an Unsafe queue to threadpool and other 4 are thread starts. (didn't want to mess in this area yet)

The Socket version and IO completion doesn't seem to be on the default context (or it would take a different path and doesn't call Run) so there are a higher percentage that are in the regular `Run` path.

/cc  @jkotas @AndyAyersMS @stephentoub @davidfowl

benaadams:
Socket version



Upstream for the `Run`

stephentoub:
What impact does this have on throughout?

If code calls TP.UQUWI and then changes the EC (e.g. changes an AsyncLocal), that'll poison the TP thread and knock subsequent work items off the fast path? And the slow path is now slower than the fast path today?

benaadams:
> And the slow path is now slower than the fast path today?

Slow path is slightly faster than current as using the local Default in Restore means one less call to `CORINFO_HELP_GETSHARED_GCSTATIC_BASE`; but otherwise unchanged.

> If code calls TP.UQUWI and then changes the EC (e.g. changes an AsyncLocal), that'll poison the TP thread and knock subsequent work items off the fast path?

Off the faster-path and only for that Dispatch quantum; will still use the previous default context fast path; albeit with one extra call. Slow path is unchanged.

However have fixed.

One speed bump is as there are more explicit default context threads `ExecutionContext.Capture` will check `m_isFlowSuppressed` before returning the context; but then it won't call `CORINFO_HELP_GETSHARED_GCSTATIC_BASE` to get the default, which it would do if it was `null`; so not sure how much that factors; but probably less than it never inlining.

I'd like to move more threads onto the default context, like the IO and explicitly call `RunDefaultContext` rather than `Run`; but I think that needs more caution.

Preliminary local testing is faster; my threadpool/task tests are slower but with CPU time in kernel greatly increased - which paradoxically means its faster shown through higher SpinWait contention - need to improve how they measure 😆 

Will do further testing.

benaadams:
Feel a bit foolish, is easy to test by setting 
```
set COMPlus_ThreadPool_ForceMinWorkerThreads=1
set COMPlus_ThreadPool_ForceMaxWorkerThreads=1
```
Then all contention measurement drops out

benaadams:
Full tests using single threadpool thread https://gist.github.com/benaadams/bf56dc4424ca6888aba178074cdae2b6

Highlights
```
                                                                             Parallelism
```
Pre
```
QUWI No Queues (TP)              8.982 M     9.500 M     9.467 M     9.318 M     9.234 M
- Depth    2                     9.367 M     9.362 M     9.486 M     9.283 M     9.310 M
- Depth   16                     9.699 M     9.767 M     9.698 M     9.664 M     9.682 M
- Depth   64                     9.756 M     9.761 M     9.764 M     9.862 M     9.630 M
- Depth  512                     9.923 M     9.948 M     9.949 M     9.876 M     9.761 M

QUWI No Queues                  13.695 M    10.798 M     3.565 M     5.252 M     4.560 M
- Depth    2                     6.792 M     6.078 M     3.798 M     7.401 M     4.040 M
- Depth   16                     6.443 M     5.297 M     9.365 M     8.035 M     3.546 M
- Depth   64                     7.934 M     9.117 M     9.239 M     8.092 M     3.825 M
- Depth  512                     9.737 M     9.781 M     9.140 M     8.050 M     4.313 M
```
Post
```
QUWI No Queues (TP)             10.049 M    10.716 M    10.668 M    10.619 M    10.461 M
- Depth    2                    10.850 M    10.877 M    10.716 M    10.744 M    10.527 M
- Depth   16                    11.165 M    11.151 M    11.261 M    11.181 M    11.063 M
- Depth   64                    11.329 M    11.335 M    11.206 M    11.222 M    11.066 M
- Depth  512                    11.379 M    11.340 M    11.353 M    11.271 M    11.233 M

QUWI No Queues                  15.985 M    12.116 M     3.238 M     6.168 M     4.727 M
- Depth    2                     7.814 M     6.308 M     4.384 M     6.899 M     3.971 M
- Depth   16                     9.124 M     9.009 M     9.242 M     8.667 M     5.019 M
- Depth   64                     9.836 M    10.090 M    10.260 M     8.762 M     4.546 M
- Depth  512                    11.014 M    11.030 M    10.287 M     9.285 M     5.285 M
```

Pre
```
SubTask Chain Awaited (TP)       1.070 M     1.067 M     1.071 M     1.072 M     1.066 M
- Depth    2                     1.097 M     1.104 M     1.102 M     1.101 M     1.099 M
- Depth   16                     1.227 M     1.222 M     1.222 M     1.219 M     1.217 M
- Depth   64                     1.227 M     1.227 M     1.227 M     1.223 M     1.221 M
- Depth  512                     1.210 M     1.218 M     1.221 M     1.220 M     1.218 M

SubTask Chain Awaited            1.061 M     1.067 M     1.049 M     1.025 M   926.972 k
- Depth    2                     1.097 M     1.090 M     1.085 M     1.082 M   972.882 k
- Depth   16                     1.226 M     1.220 M     1.215 M     1.194 M     1.052 M
- Depth   64                     1.224 M     1.226 M     1.219 M     1.195 M     1.104 M
- Depth  512                     1.214 M     1.217 M     1.220 M     1.190 M     1.102 M
```
Post
```
SubTask Chain Awaited (TP)       1.112 M     1.120 M     1.117 M     1.115 M     1.122 M
- Depth    2                     1.139 M     1.136 M     1.135 M     1.138 M     1.136 M
- Depth   16                     1.242 M     1.243 M     1.245 M     1.246 M     1.245 M
- Depth   64                     1.243 M     1.241 M     1.245 M     1.240 M     1.237 M
- Depth  512                     1.232 M     1.232 M     1.235 M     1.235 M     1.238 M

SubTask Chain Awaited            1.113 M     1.082 M     1.103 M     1.087 M     1.023 M
- Depth    2                     1.138 M     1.138 M     1.130 M     1.110 M   955.189 k
- Depth   16                     1.247 M     1.245 M     1.239 M     1.224 M     1.119 M
- Depth   64                     1.246 M     1.236 M     1.235 M     1.220 M     1.083 M
- Depth  512                     1.236 M     1.233 M     1.231 M     1.213 M     1.080 M
```

Pre
```
Continuation Chain (TP)        753.935 k   747.613 k   746.931 k   744.462 k   745.137 k
- Depth    2                     1.226 M     1.223 M     1.215 M     1.214 M     1.225 M
- Depth   16                     2.867 M     2.876 M     2.836 M     2.829 M     2.800 M
- Depth   64                     3.321 M     3.297 M     3.303 M     3.281 M     3.176 M
- Depth  512                     3.456 M     3.473 M     3.480 M     3.436 M     3.439 M

Continuation Chain             749.259 k   748.858 k   737.790 k   732.280 k   683.638 k
- Depth    2                     1.222 M     1.225 M     1.199 M     1.181 M     1.047 M
- Depth   16                     2.821 M     2.842 M     2.708 M     2.631 M     2.121 M
- Depth   64                     3.294 M     3.295 M     3.171 M     3.097 M     2.237 M
- Depth  512                     3.488 M     3.521 M     3.400 M     3.306 M     2.891 M
```
Post
```
Continuation Chain (TP)        790.326 k   790.035 k   789.038 k   785.879 k   787.792 k
- Depth    2                     1.299 M     1.291 M     1.287 M     1.282 M     1.292 M
- Depth   16                     2.992 M     2.986 M     2.965 M     2.949 M     2.958 M
- Depth   64                     3.449 M     3.419 M     3.428 M     3.408 M     3.311 M
- Depth  512                     3.605 M     3.587 M     3.578 M     3.573 M     3.553 M

Continuation Chain             787.984 k   787.776 k   780.919 k   770.841 k   677.396 k
- Depth    2                     1.288 M     1.283 M     1.273 M     1.250 M     1.131 M
- Depth   16                     2.954 M     2.953 M     2.830 M     2.768 M     2.127 M
- Depth   64                     3.450 M     3.431 M     3.310 M     3.238 M     2.418 M
- Depth  512                     3.634 M     3.623 M     3.494 M     3.431 M     3.492 M
```

Pre
```
Yield Chain Awaited (TP)         1.953 M     1.919 M     1.919 M     1.901 M     1.885 M
- Depth    2                     2.348 M     2.349 M     2.328 M     2.330 M     2.328 M
- Depth   16                     3.027 M     2.999 M     2.990 M     2.982 M     3.006 M
- Depth   64                     3.107 M     3.099 M     3.092 M     3.172 M     3.044 M
- Depth  512                     3.175 M     3.111 M     3.191 M     3.133 M     3.115 M

Yield Chain Awaited              1.984 M     1.975 M     1.877 M     1.827 M     1.494 M
- Depth    2                     2.332 M     2.327 M     2.271 M     2.228 M     1.671 M
- Depth   16                     3.000 M     2.986 M     2.893 M     2.778 M     2.116 M
- Depth   64                     3.200 M     3.083 M     3.023 M     2.957 M     2.335 M
- Depth  512                     3.222 M     3.051 M     3.079 M     3.084 M     2.534 M
```
Post
```
Yield Chain Awaited (TP)         2.186 M     2.145 M     2.129 M     2.134 M     2.081 M
- Depth    2                     2.487 M     2.483 M     2.488 M     2.460 M     2.499 M
- Depth   16                     3.049 M     3.053 M     3.025 M     2.986 M     3.031 M
- Depth   64                     3.114 M     3.128 M     3.086 M     3.171 M     3.086 M
- Depth  512                     3.203 M     3.113 M     3.194 M     3.107 M     3.093 M

Yield Chain Awaited              2.209 M     2.203 M     2.099 M     2.023 M     1.665 M
- Depth    2                     2.476 M     2.466 M     2.398 M     2.369 M     1.855 M
- Depth   16                     3.022 M     3.024 M     2.937 M     2.810 M     2.175 M
- Depth   64                     3.156 M     3.093 M     3.030 M     3.023 M     2.555 M
- Depth  512                     3.148 M     3.055 M     3.104 M     3.156 M     2.464 M
```

Generally looks better

benaadams:

src/mscorlib/shared/System/Threading/ExecutionContext.cs
```diff
@@ -42,6 +45,59 @@ internal void Undo(Thread currentThread)
                 ExecutionContext.Restore(currentThread, m_ec);
             }
         }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static bool CurrentIsDefault(Thread currentThread, ExecutionContext defaultContext)
+        {
+            // Current thread passed as parameter to avoid extern current native thread lookup call
+            Debug.Assert(currentThread == Thread.CurrentThread);
+            // Default context passed as parameter to avoid static inializer checks in NGen'd code
```
kouvel: typo: inializer

src/mscorlib/shared/System/Threading/ExecutionContext.cs
```diff
@@ -42,6 +45,59 @@ internal void Undo(Thread currentThread)
...
+            // The common case is that these have not changed, so avoid the GC memory barrier 
+            // cost of an reference write if not needed.
+            if (currentThread.SynchronizationContext != null)
+            {
+                currentThread.SynchronizationContext = null;
+            }
+
+            if (currentThread.ExecutionContext != defaultContext)
+            {
+                currentThread.ExecutionContext = defaultContext;
```
kouvel: Should this call `ExecutionContext.RestoreDefault` as in `UndoToDefault` to check for AsyncLocal changes and raise events?

If so, then perhaps `UndoToDefault` could be renamed to `SetDefaults` and this overload with no parameters can call the one with parameters.

src/mscorlib/shared/System/Threading/ExecutionContext.cs
```diff
@@ -165,16 +226,65 @@ public static void Run(ExecutionContext executionContext, ContextCallback callba
...
+            // Passed to minimise static inializer checks in NGen'd code
+            // and extern current native thread lookup calls
+            Debug.Assert(currentThread == Thread.CurrentThread);
+            Debug.Assert(defaultContext == ExecutionContext.Default);
+
+            ExecutionContext previous = currentThread.ExecutionContext;
+            currentThread.ExecutionContext = defaultContext;
+
+            // For the purposes of dealing with context change, null counts as the default EC
+            if (previous != null)
```
kouvel: It looks like this would call `OnContextChanged` when `previous == defaultContext`. Maybe this should be `previous != defaultContext && previous != null`.

src/mscorlib/src/System/Threading/ThreadPool.cs
```diff
@@ -491,6 +491,9 @@ public IThreadPoolWorkItem Dequeue(ThreadPoolWorkQueueThreadLocals tl, ref bool
 
         internal static bool Dispatch()
         {
+            // Put threadpool thread on default context
+            ExecutionContextSwitcher.SetDefaults();
```
kouvel: Why is this necessary? Shouldn't each work item restore the context especially after these changes, or is that still missing somewhere?

benaadams: Threadpool threads don't start on Default context as the VM creates them rather than the managed code

benaadams: Though it should only need ever to be called once; so maybe there is a better way of doing it?

kouvel: Shouldn't the context be null to start with? Maybe `ExecutionContextSwitcher.CurrentIsDefault` should also check for null.

benaadams: If its `null` then it needs to do a memory barrier assign to `Default`; so its once per dispatch or once per loop.

src/mscorlib/shared/System/Threading/ExecutionContext.cs
benaadams: No, is only used for after running an `Unsafe` workitem and `Dispatch` when no user code is running

src/mscorlib/shared/System/Threading/ExecutionContext.cs
benaadams: Is called by `UndoToDefault` when previous != default so its already implied that previous isn't default. Probably should add an additional asset at top.

kouvel: Oh right, makes sense

src/mscorlib/src/System/Threading/ThreadPool.cs
kouvel: Why not leave it as null (and allow restore to null) and also check for null (maybe first) when checking for default? Is it significantly better to initialize to `Default` and check for only that?

benaadams: Does the VM call any init function when creating a threadpool thread? Or does it go straight to Dispatch?

kouvel: It seems like it may be better to keep the field set to null as far as possible rather than to keep it set to `Default`

kouvel: It doesn't look like the VM calls anything on thread creation

src/mscorlib/shared/System/Threading/ExecutionContext.cs
kouvel: Not sure about this, the unsafe work item could change async locals and it seems this is the only place where we don't notify when restoring the context. For the `Dispatch` case where no user code is running, if the EC needs to be initialized to `Default`, could probably do that directly and even avoid checking the `SynchronizationContext` which should be null anyway at that point.

src/mscorlib/shared/System/Threading/ExecutionContext.cs
kouvel: Yea an assert would be good, as it is exposed to future calls from elsewhere

src/mscorlib/src/System/Threading/ThreadPool.cs
benaadams: > Why not leave it as null (and allow restore to null) and also check for null (maybe first) when checking for default? Is it significantly better to initialize to Default and check for only that?

`SetDefaults` does check if its already default before assigning it. `null` is a bit of a weird one; if you `Capture` it `?? Deafult`; however if you are inspecting it directly then it counts a flow suppressed.

But mainly it doubles the number of branches and is whether the `JIT_XxxWriteBarrier` is called once per item executed or once per threadpool thread lifetime. 

As running on the default context is currently baked into `QueueUserWorkItemCallbackDefaultContext` and ATMB.`MoveNextRunner.RunWithDefaultContext` and with this change into `Task.ExecuteWithThreadLocal` it makes sense to just leave the threadpool thread on the default context; rather than moving to null and always restoring the Default context; especially since non-default contexts will always restore to previous; whether that is null or Default.

So with this change the 3 definitely run on default context paths call the new method `RunDefaultContext` and don't bother passing context as an argument.

Other flows which may be on the default context call `Run` as they were doing before, still taking the previous fast path, but receiving the advantage that the Threadpool is already on the default context rather than null so the context doesn't need to be assigned to Default or restored back to null.

Its just introducing an even faster path when the Run context is known to be Default upfront.

benaadams: i.e. running the default will never call `CORINFO_HELP_ASSIGN_REF`

src/mscorlib/shared/System/Threading/ExecutionContext.cs
benaadams: Currently if `Unsafe` changed async locals you'd wouldn't get notification at the end of the `Unsafe` call but you'd get them at the start of every regular threadpool item execution and the values would live in a kinda limbo land.

Also you could never move back onto the DefaultContext as even if you clear the values it won't move you back onto Default https://github.com/dotnet/coreclr/pull/8227#issue-190780413

benaadams: Its the poisoning @stephentoub referred to in https://github.com/dotnet/coreclr/pull/11100#issuecomment-296010509 that I resolved with https://github.com/dotnet/coreclr/pull/11100/commits/f358e024626f0d7ef5d721eb27509abbe8720c78

Can revert it to allow poisoning by `Unsafe` which would permanently move it onto slower path but maintain current behaviour?

src/mscorlib/src/System/Threading/ThreadPool.cs
kouvel: With this:
> It seems like it may be better to keep the field set to null as far as possible rather than to keep it set to Default

I was thinking the EC field could be forced to null always instead of default, that is, it would never be set to `Default`, such that:
- `Capture` checks for null and returns `Default`
- `Run` would check for `Default` and set the field to null if not already null
- `Restore` would only check for null
- `ExecutionContextSwitcher.CurrentIsDefault` would only check for null

It seems like this would restore the number of branches and satisfy the same requirement that the write barrier is not called when running with the default context, maybe with some fewer loads of `Default`.

`Capture` has an odd contract, but its behavior is separate and can remain unchanged even if the EC field on `Thread` is kept as null instead of `Default`.

But this approach seems fine to me as well.

Regarding a one-time initialization, it would be possible to have the VM notify the managed side in `Thread.CurrentThread` if the `Thread` object was created, but since CurrentThread is called a lot, it may well be worse.

src/mscorlib/shared/System/Threading/ExecutionContext.cs
kouvel: Yea that is broken. @stephentoub, thoughts on whether we should restore the context, and whether we should notify of async local changes upon restoring here? Not sure how important the `Unsafe` version is, but the same path applies to suppressed-flow contexts as well.

src/mscorlib/shared/System/Threading/ExecutionContext.cs
benaadams: Resolved

src/mscorlib/shared/System/Threading/ExecutionContext.cs
benaadams: Added

src/mscorlib/shared/System/Threading/ExecutionContext.cs
kouvel: > Yea that is broken

I was referring to your previous comment:

> Currently if Unsafe changed async locals you'd wouldn't get notification at the end of the Unsafe call but you'd get them at the start of every regular threadpool item execution and the values would live in a kinda limbo land.

src/mscorlib/src/System/Threading/ThreadPool.cs
benaadams: I kinda agree; but the EC mix with default and flow suppressed is more complicated so it would need a greater set of changes.

e.g. For `Capture` for `CurrentThread.ExecutionContext`

`EC == null` => `Default`
`EC != null && FlowSupressed` => `null`
`EC != null` =>`EC`

So `null` means flow suppressed, unless you recapture when it means Default

benaadams: i.e. could have a `FlowSupressed` context with `null` being `Default` instead

benaadams: i.e. could have a `FlowSupressed` context with `null` being `Default` instead; but that's probably a more dangerous change.

kouvel: > So null means flow suppressed, unless you recapture when it means Default

I didn't follow that. I think `Capture` could stay as is, if it returns null it means EC != null and FlowSuppressed == true. If `Capture` returns `Default` it means EC == null. A current flow-suppressed context means EC != null.

> i.e. could have a FlowSupressed context with null being Default instead; but that's probably a more dangerous change?

I didn't follow that either, could you please elaborate?

benaadams: Thinking...

benaadams: Will try using null for default internally

danmosemsft:
Hi @benaadams do you expect to take it further?

benaadams:
@danmosemsft yes, I need to invert what its doing as per @kouvel's suggestion (i.e. what `null` represents)

danmosemsft:
@benaadams interested in continuing?...

karelz:
@benaadams if you don't have time to work on it now, let's close the PR, until you have the time again ;)

benaadams:
Very interested; but swamped. Will close and revisit.

PR #10946, opened 2017-04-13
Clear GTF_ASG from non-stores in rationalize

JosephTremoulet:
Since rationalization hoists comma antecedents out into separate
statements, any side-effects that the antecedent had might no longer apply
to the comma's ancestor nodes.  Moreover, in LIR, side-effect flags need
only describe the immediate node; summary flags are computed when Ranges
are constructed.  So, update rationalize to clear the `GTF_ASG` flag from
any node that is not a store.  Failure to do so gives an
overly-conservative view of side-effects, which can inhibit dead store
removal.

Fixes #9633.

Issue #10731, opened 2017-04-05
JIT: consider doing some type opts earlier


AndyAyersMS:
The jit should now be able to optimize many of the common type specialization checks during importation.  Since the jit can also do some rudimentary branch opts during importation, early type specialization check optimization would allow the jit to skip importing chunks of IL guarded by type checks that are known to fail.

For example consider any non-intrinsic vector method, eg [`Vector<T>:op_Multiply`](https://github.com/dotnet/corefx/blob/b6d4e5fa4bddd130fa2f95e34f5487064ff6bd1b/src/System.Numerics.Vectors/src/System/Numerics/Vector.cs#L1969). Since`T` is constrained to non-ref types this method's instantiation parameter is always a known class at jit time. So roughly 90% of the IL in this method is unreachable by any particular instantiation and doesn't need to be imported.

This should provide a throughput boost. It might also make it easier for the jit to consider inlining such methods, because the jit can more accurately assess the "true" size of the method.

redknightlois:
We abuse inlining + constrained non-ref types a lots. Would give us a huge boost!!

AndyAyersMS:
Did some initial prototyping to refactor the type equality optimizations so they can be run during importation (right now they're embedded in morph). There are two main optimizations:
* Modify `Type.op_Equality(a, b)` to be simply `a == b`, when safe to do so (and likewise for inequality). This one is pretty easy to move forward since we have an intrinsic for `Type.op_Equality` and some facilities for identifing when `a` or `b` is known to be a `RuntimeType`.
* Modify `a == b` when either `a' or `b` is a `RuntimeType` constructed from a compile-time handle. Here we can forego fetching the type object and just compare the handle.

See [master...AndyAyersMS:EarlyTypeOpt](https://github.com/dotnet/coreclr/compare/master...AndyAyersMS:EarlyTypeOpt?expand=1)

However during importation we're only opportunistically folding a small subset of trees as we build them, and the folding is done via `gtFoldExpr` which does not take a bottom-up top-down approach. Not sure why, perhaps it is too costly, or redundant/overlaps with aspects of morph.

At any rate, unless we arrange to aggressively or opportunistically fold subtrees, higher level folding may be blocked because subtrees haven't been simplified. In particular if we push both the folding optimizations above forward, we still miss early folding in cases like:
```C#
    static bool IsInt<T>()
    {
        if (typeof(T) == typeof(int))
        {
            return true;
        }
        else
        {
            return false;
        }
    }
[...]

PR #11104, opened 2017-04-20
Inline List.Clear

benaadams:
Even though the two paths are small post jit (non ref and ref) the combination pre-jit in the same function makes it look to big too inline.

Contributes to https://github.com/dotnet/corefx/issues/18390

/cc @jkotas @AndyAyersMS

### DIFF
--- a/src/mscorlib/src/System/Collections/Generic/List.cs
+++ b/src/mscorlib/src/System/Collections/Generic/List.cs
@@ -350,6 +350,7 @@ namespace System.Collections.Generic
 
 
         // Clears the contents of List.
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public void Clear()
         {
             if (RuntimeHelpers.IsReferenceOrContainsReferences<T>())


benaadams:
Asm to inline

Non reference asm
```asm
G_M4562_IG01:
       0F1F440000           nop      

G_M4562_IG02:
       894118               mov      dword ptr [rcx+24], eax

G_M4562_IG03:
       C3                   ret    
```
Reference asm
```asm
G_M58440_IG01:
       0F1F440000           nop      

G_M58440_IG02:
       4C8B01               mov      r8, qword ptr [rcx]
       448B4118             mov      r8d, dword ptr [rcx+24]
       33D2                 xor      edx, edx
       895118               mov      dword ptr [rcx+24], edx
       4585C0               test     r8d, r8d
       7E10                 jle      SHORT G_M58440_IG04
       33D2                 xor      edx, edx

G_M58440_IG03:

G_M58440_IG04:
       C3                   ret     
```

benaadams:
I think the `mov      r8, qword ptr [rcx]` is extra in the ref version?

omariom:
@AndyAyersMS
Are there plans to make jit take into account the final size of methods?

benaadams:
@omariom this its covered by this issue? https://github.com/dotnet/coreclr/issues/10731

omariom:
@benaadams 
I even liked it but then completely forgot  it :smile:
After that change JIT will be inlining List.Clear only for non RoCR types.

Great to see that List.Clear is now just 2 machine instructions :+1:

AndyAyersMS:
@omariom no, not exactly -- to inline this w/o the aggressive inline attribute the jit would have to anticipate that a type optimization was going to happen in the inlinee and then estimate the impact this optimization would have on the method size. When and where the type optimization happens is not as crucial for the inliner.

When we cost an inlinee today, we do so by just looking at its IL stream. There may be things we can spot in our fairly shallow evaluation of that stream to suggest that type based optimizations are likely to have a size impact. Estimating the magnitude of the impact may be a bit trickier.

An alternative that we could pursue in crossgen is to generate code for methods "bottom-up" -- callees before callers. Then we could flag callees where final size is considerably less than would be expected from their IL size and use this as a part of the size estimates for inlines when compiling callers.

In a tiered jit world we might also be able to profit from such smaller-than-expected observations and reconsider inlining methods that we previously rejected.

AndyAyersMS:
Would have to drill in to see why that extra mov hangs around. It could be a vestige of the runtime lookup we create initially to call `IsReference...` and so perhaps something that #10946 cleans up.

mikedn:
@AndyAyersMS Isn't the extra mov there because it has an exception side effect? I suppose it could be removed because the following mov happens to have the same side effect but that's not something that the JIT does today.

AndyAyersMS:
Yeah, it could be. I haven't looked at the jit dumps to see exactly what is going on.

mikedn:
Dump looks like this:
```
[000006] ------------             *  stmtExpr  void  (IL 0x000...  ???)
[000003] #--X--------             |  /--*  indir     long
[000002] ------------             |  |  \--*  lclVar    ref    V00 this
[000005] -A-X--------             \--*  =         long
[000004] D------N----                \--*  lclVar    long   V02 tmp0
```
Once introduced it looks like this indir can't easily go away.

Come to think of it this is seemingly strange side effect of an implementation detail. You'd think that a method like
```C#
void Clear() {
    if (RuntimeHelpers.IsReferenceOrContainsReferences<T>()) {
        // do nothing
    }
}
```
should not throw null ref exception if `this` is `null` but it does that because the type context is obtained via `this`.

AndyAyersMS:
Yeah, that's what I was getting at, but I didn't articulate it very well. Let me try again...

The runtime helper is static, so we should be able to call it with a null `this`. But our runtime lookup expansion introduces a dereference. So evidently there's some implicit (or perhaps explicit) requirement when calling a method that needs to do a runtime lookup off of `this` that callers must do a null check, so `this` is known to be not-null on entry, and hence the deref in the lookup path will never fail and and is not semantically motivated, so can be removed, despite it looking like something that can cause an exception.

AndyAyersMS:
Ecma I.8.4.2 says that instance methods on generic types can't be called with null `this` pointers. But I don't know how this is enforced or whether trying to leverage this puts the jit into the temptingly dangerous "undefined behavior" bucket where it can just assume what it expects to be true.

Most times these methods are invoked via `callvirt` so presumably the primary enforcement is on the caller side. But that's not required. Hacking up a simple example like the above to pass a null this and use `call` opcode still triggers NRE because of the residual internal check. But presumably per I.8.4.2 we could optimize that away.

But unless we modify the jit to always remove the null check (even in minopts/debuggable mode), this would have the annoying effect of possibly changing behavior because of optimization.

We might also consider enforcing this more directly by having the jit interface demand a null check for any generic instance method call, regardless of opcode.

@jkotas any thoughts?

jkotas:
I think it is fine to have undefined behavior wrt NREs when a method on generic type is called with null thisptr.

It is implementation dependent which exact operations do dictionary lookups that would trigger the NRE. I do not think it makes sense to try to set the exact set of the operations in stone. Also, we do not have any extensive tests to verify that the NREs are maintained in the exact same spots, so it is likely it have changed many times without us noticing.

PR #11107, opened 2017-04-20
Num succ improvement

BruceForstall:


mikedn:
Maybe these overloads should have different names? It seems too easy to make mistakes by forgetting to pass a `Compiler*`.

BruceForstall:
Maybe? Any name suggestions?

The usage hasn't changed, just the implementation. Maybe we move to an iterator model later.

mikedn:
> Any name suggestions?

Sorry, no. This is EH related stuff and I'm not enough familiar with that to come up with a good name.

BruceForstall:
Test Windows_NT x64 throughput
Test Windows_NT x86 throughput
Test Ubuntu14.04 throughput

BruceForstall:

BruceForstall:
@pgavlin PTAL
cc @dotnet/jit-contrib

JosephTremoulet:
> It seems too easy to make mistakes by forgetting to pass a `Compiler*`.

+1

> Any name suggestions?

Maybe the ones that don't take a `Compiler*` could be `NumSuccNaive`/`GetSuccNaive`?

BruceForstall:
> It seems too easy to make mistakes by forgetting to pass a Compiler*.

Note that the usage model in this change is exactly the same as it is today.

> Any name suggestions?

NumSuccSimple/GetSuccSimple?

I think this could/should be done as a separate commit.

JosephTremoulet:
> NumSuccSimple/GetSuccSimple?

The thing I don't like about "Simple" is that it doesn't convey "Unsound"...

pgavlin:
When we last discussed this, I suggested creating templated versions of these functions rather than separating them into multiple overloads. Something like this: https://gist.github.com/pgavlin/7916e6aacae392498f624310312f96a4

I think that this option is worth considering, as it both avoids code duplication and forces the caller to be conscientious about their choice of algorithm. For better readability, a type like
```
enum class SuccessorsKind
{
    Naive,
    All
};
```
could be used as the template parameter.

BruceForstall:
btw, our throughput tests measured 0.2% improvement for x64, 0.5% improvement for x86. Though there's lots of noise, so I don't know how much I trust them.

CarolEidt:
I like the templated approach that @pgavlin suggests, though instead of `Naive` perhaps we want to be very explicit and use `Unsound`, as indirectly suggested by @JosephTremoulet

BruceForstall:
"Unsound" doesn't makes sense to me. I don't have a clear definition in my head of "soundness" (plus I hate that terminology). E.g., when passed Compiler*, GetSucc() only returns unique switch successors. That doesn't seem related to soundness. (Note that GetSucc() for BBJ_COND *always* returns only unique successors, which seems slightly weird.)

Also, neither returns all EH edges, as the AllSuccessorIter does. Using that is the only "real" "sound" option.

Personally, I find the template approach unnecessarily complicating to what is otherwise quite simple.

CarolEidt:
@BruceForstall - I'd forgotten all the subtleties of this. I agree that the renaming is probably best done as a separate commit. And given the inconsistencies it doesn't even sound like names such as `GetNumUniqueSucc` and `GetNumSuccEdges` would be correct.

PR #11108, opened 2017-04-20
Use C++ ranged 'for' for AllSuccessorIter use

BruceForstall:


mikedn:
Side-question: What do you think about using range for to iterate over basic blocks and statements?

BruceForstall:
> Side-question: What do you think about using range for to iterate over basic blocks and statements?

As long as it's simple, easy-to-understand, and as performant, then great!

We have foreach_block and foreach_treenode_execution_order macros, but it would be nicer to use built-in language constructs. (Note that it was only recently that we didn't have to worry about older compilers that didn't support ranged-for, which is why we have those macros in the first place.)

BruceForstall:
@swgillespie 2 GC tests failed in "Windows_NT x64 Debug Build and Test":
```
GC_LargeMemory._API_gc_gettotalmemory_gettotalmemory_._API_gc_gettotalmemory_gettotalmemory_cmd
GC_Regressions._v2_0_beta1_149926_149926_149926_._v2_0_beta1_149926_149926_149926_cmd
```
with:
```
Large Memory Machine required
and:
OOM while allocating a byte array of size 16774256
```

why? Is there some kind of infrastructure or test flakiness here?

mikedn:
> As long as it's simple, easy-to-understand, and as performant, then great!

Yes, with one exception: one needs to be careful with iterating and editing the list at the same time. Oh well, I'll try it and see how it looks.

swgillespie:
@BruceForstall I've never seen those two fail in that way before, I'll take a look.

BruceForstall:

BruceForstall:

BruceForstall:
@pgavlin PTAL
cc @dotnet/jit-contrib

Issue #11076, opened 2017-04-19
[Memory consumption] Compact entry points for ARM


gbalykov:
Hello.

We propose to add compact entry points for ARM, but not exactly the way they are done for x86. For MethodDescChunk with 5 MethodDescs compact entry points will look like this:

```
entry1:
  mov r12, pc
  b CentralJump
entry2:
  mov r12, pc
  b CentralJump
entry3:
  mov r12, pc
  b CentralJump
entry4:
  mov r12, pc
  b CentralJump
entry5:
  mov r12, pc
  b CentralJump
CentraJump:
  ldr pc, pThePreStubCompactARM
  nop
  dw pChunk
  dw pThePreStubCompactARM
```

Compact entry points are encoded in Thumb, each compact entry has 4 byte size, CentralJump is 14 bytes.

ThePreStubCompactARM finds CentralJump by looking into branch offset of compact entry. Then it calculates offset of the compact entry from CentralJump by program counter stored in r12 and, thus, calculates index of MethodDesc in MethodDescChunk. Then,ThePreStubCompactARM loads pointer to MethodDescChunk from CentralJump and obtains MethodDesc, corresponding to the calculated index (this is done by traversing all MethodDescs in MethodDescChunk starting from the first until the MethodDesc with required index is met). After that, ThePreStubCompactARM calls PreStubWorker with obtained MethodDesc, like original ThePreStub.

Estimation on Tizen GUI Xamarin application (https://developer.tizen.org/sites/default/files/documentation/puzzle2.zip) shows, for ReadyToRun mode, that reduction of code heap size is 76 KB, which is 20% of code heap and, approximately, 2.2% of overall per-application CoreCLR memory consumption. For No-Precompilation mode, reduction of code heap size is 216 KB, which is 28% of code heap and, approximately, 2.3% of overall per-application CoreCLR memory consumption.

ReadyToRun mode means the Tizen-default set of precompiled assemblies is in ReadyToRun format, No-Precompilation mode means that there are no precompiled assemblies.

However, we expect slightly lower performance. Could you, please, suggest a way to estimate performance changes? We want to consider enabling compact entry points for arm by default. What do you think about this?

We plan to open pull request for this change in two days.

@lemmaa @egavrin @Dmitri-Botcharnikov @ruben-ayrapetyan

jkotas:
> However, we expect slightly lower performance

The compact entrypoint is only used once to trigger JITing. There should be no performance impact.

[...]

PR #11109, opened 2017-04-20
Add compact entry points for ARM

gbalykov:
This pull request adds compact entry points for ARM. 

Related issue: #11076 

For MethodDescChunk with 5 MethodDescs compact entry points will look like this:

```
entry1:
  mov r12, pc
  b CentralJump
entry2:
  mov r12, pc
  b CentralJump
entry3:
  mov r12, pc
  b CentralJump
entry4:
  mov r12, pc
  b CentralJump
entry5:
  mov r12, pc
  b CentralJump
CentraJump:
  ldr pc, pThePreStubCompactARM
  nop
  dw pChunk
  dw pThePreStubCompactARM
```

Compact entry points are encoded in Thumb, each compact entry has 4 byte size, CentralJump is 14 bytes.

ThePreStubCompactARM finds CentralJump by looking into branch offset of compact entry. Then it calculates offset of the compact entry from CentralJump by program counter stored in r12 and, thus, calculates index of MethodDesc in MethodDescChunk. Then,ThePreStubCompactARM loads pointer to MethodDescChunk from CentralJump and obtains MethodDesc, corresponding to the calculated index (this is done by traversing all MethodDescs in MethodDescChunk starting from the first until the MethodDesc with required index is met). After that, ThePreStubCompactARM calls PreStubWorker with obtained MethodDesc, like original ThePreStub.

Estimation on Tizen GUI Xamarin application (https://developer.tizen.org/sites/default/files/documentation/puzzle2.zip) shows, for ReadyToRun mode, that reduction of code heap size is 76 KB, which is 20% of code heap and, approximately, 2.2% of overall per-application CoreCLR memory consumption. For No-Precompilation mode, reduction of code heap size is 216 KB, which is 28% of code heap and, approximately, 2.3% of overall per-application CoreCLR memory consumption.

ReadyToRun mode means the Tizen-default set of precompiled assemblies is in ReadyToRun format, No-Precompilation mode means that there are no precompiled assemblies.

Testing:

- CI: default
- CoreCLR test suite on ARM on Tizen: no regression happened 

Micro-benchmark, provided by @jkotas in https://github.com/dotnet/coreclr/issues/11076#issuecomment-296748946, showed next results:
- Cortex-A7 softfp:
      before: 12.81s (user time)
      after: 10.488s (user time)
- Cortex-A15 softfp:
      before: 5.3s (user time)
      after: 4.68s (user time)

### DIFF
--- a/src/vm/arm/asmhelpers.S
+++ b/src/vm/arm/asmhelpers.S
@@ -509,6 +509,24 @@ LOCAL_LABEL(UM2MThunk_WrapperHelper_ArgumentsSetup):
         NESTED_END ThePreStub, _TEXT
 
 // ------------------------------------------------------------------
+        NESTED_ENTRY ThePreStubCompactARM, _TEXT, NoHandler
+
+        // r12 - address of compact entry point + PC_REG_RELATIVE_OFFSET
+
+        PROLOG_WITH_TRANSITION_BLOCK
+
+        mov         r0, r12
+
+        bl          C_FUNC(PreStubGetMethodDescForCompactEntryPoint)
+
+        mov         r12, r0                                          // pMethodDesc
+
+        EPILOG_WITH_TRANSITION_BLOCK_TAILCALL
+
+        b           C_FUNC(ThePreStub)
+
+        NESTED_END ThePreStubCompactARM, _TEXT
+// ------------------------------------------------------------------
 // This method does nothing. It's just a fixed function for the debugger to put a breakpoint on.
         LEAF_ENTRY ThePreStubPatch, _TEXT
         nop
--- a/src/vm/method.cpp
+++ b/src/vm/method.cpp
@@ -4571,6 +4571,35 @@ c_CentralJumpCode = {
 };
 #include <poppack.h>
 
+#elif defined(_TARGET_ARM_)
+
+#include <pshpack1.h>
+struct CentralJumpCode {
+    BYTE m_ldrPC[4];
+    BYTE m_short[2];
+    MethodDescChunk *m_pChunk;
+    PCODE m_target;
+
+    inline void Setup(PCODE target, MethodDescChunk *pChunk) {
+        WRAPPER_NO_CONTRACT;
+
+        m_target = target;
+        m_pChunk = pChunk;
+    }
+
+    inline BOOL CheckTarget(TADDR target) {
+        WRAPPER_NO_CONTRACT;
+        return ((TADDR)m_target == target);
+    }
+}
+c_CentralJumpCode = {
+    { 0xDF, 0xF8, 0x08, 0xF0 },                         //   ldr pc, =pTarget
+    { 0x00, 0x00 },                                     //   short offset for alignment
+    0,                                                  //   pChunk
+    0                                                   //   pTarget
+};
+#include <poppack.h>
+
 #else
 #error Unsupported platform
 #endif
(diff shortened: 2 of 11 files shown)


gbalykov:
For ReadyToRun mode memory reduction is lower than for No-Precompilation mode. Could you, please, suggest, why this is happening?

src/vm/arm/asmhelpers.S
```diff
@@ -509,6 +509,29 @@ LOCAL_LABEL(UM2MThunk_WrapperHelper_ArgumentsSetup):
...
+
+        mov         r0, r12
+
+        bl          C_FUNC(PreStubGetMethodDescForCompactEntryPoint)
+
+        mov         r1, r0                                          // pMethodDesc
+
+        add         r0, sp, #__PWTB_TransitionBlock                 // pTransitionBlock
+
+        bl          C_FUNC(PreStubWorker)
```
jkotas: It think it would be better to tailcall to ThePrestub instead - it saves you from the extra plumbing required for debugger here. It is what we do in other similar places, e.g. check PrecodeFixupThunk.

src/vm/method.cpp
```diff
@@ -4597,23 +4708,56 @@ static_assert_no_msg((TEP_CENTRAL_JUMP_SIZE & 1) == 0);
...
     }
 
+#ifdef _TARGET_ARM_
+
+    // On ARM compact entry points are thumb
+    _ASSERTE ((addr & THUMB_CODE) != 0);
+    addr = addr - THUMB_CODE;
+
+    // Get offset for CentralJumpCode from current compact entry point
+    uint16_t *pBranchInstr = ((uint16_t *)addr) + 1;
```
jkotas: This needs to be DACized properly so that it does not break debugger: cast to `PTR_XXX` and dereference that, similar to how x86 is done.

More info about DAC: https://github.com/dotnet/coreclr/blob/master/Documentation/botr/dac-notes.md

src/vm/method.cpp
```diff
@@ -4597,23 +4708,56 @@ static_assert_no_msg((TEP_CENTRAL_JUMP_SIZE & 1) == 0);
...
+#endif // _TARGET_ARM_
+
     CentralJumpCode* pCentralJumpCode = PTR_CentralJumpCode(centralJump);
 
     // Always do consistency check in debug
     if (fSpeculative INDEBUG(|| TRUE))
     {
+#ifdef _TARGET_ARM_
+
+        if (memcmp (pCentralJumpCode, &c_CentralJumpCode, offsetof (CentralJumpCode, m_pChunk)) != 0)
```
jkotas: Needs to be DACize, e.g. read one byte at a time like x86.

gbalykov:
@dotnet-bot help

gbalykov:

gbalykov:
Test JIT/Regression/CLR-x86-JIT/V1-M09.5-PDC/b26863/b26863/b26863.sh in CI for Tizen failed and succeeded in two consecutive launches without any additional fixes

fail -> https://ci.dot.net/job/dotnet_coreclr/job/master/job/armel_cross_debug_tizen_prtest/1757/console
success -> https://ci.dot.net/job/dotnet_coreclr/job/master/job/armel_cross_debug_tizen_prtest/1758/console

What do you think about this? Could this be related to this pull request or this is a known issue in CI?

danmosemsft:
@jkotas for question above? although CI is green, should this merge?

PR #11115, opened 2017-04-20
Move readytorun tests into directory one level deeper

wtgodbe:
Some of these tests were failing in Helix because they depend on being able to find test binaries relative to the folder above them - this assumption wasn't maintained in Helix, since we consider each 2nd-level test directory to be its own independent test tree there. By creating a 2nd level directory inside the readytorun directory, we can circumvent this problem.


wtgodbe:
CC @gkhanna79 is this an alright change?

gkhanna79:
If the test passes with this change, I am fine with it.

PR #11117, opened 2017-04-20
ReadyToRun/Crossgen: Disable cross module generic instantiations from profile data

briansull:
When we have profile data we can have cross module references in our ZapSig's when building ReadyToRun images. Instead of creating the cross module generic type we instead return
a null TypeHandle to indicate failure so that we don't attempt to add generic types/methods
that have cross module references.  (i.e. MyList<System.Guid> )


briansull:
@dotnet-bot test Windows_NT Release r2r
@dotnet-bot test Windows_NT Checked r2r

briansull:
@dotnet-bot test Windows_NT Release r2r

PR #11125, opened 2017-04-21
Add Marshal.GetExceptionPointers

yizhang82:
Debugger folks need this to capture exceptions in dumps. We might need to throw PNSE in .NET native (or make up something there). 
Any objections? @jkotas @janvorli @AntonLapounov


jkotas:
LGTM otherwise

sunitaw:
I am using .net core 2.0 and still getting - cannot resolve symbol for GetExceptionPointers

luqunl:
@sunitaw, We are investigating this.

pharring:
We (Snapshot Debugger) use reflection to call this method in .NET Core / .NET Standard 2.0

PR #11132, opened 2017-04-21
Remove destructor from GCEvent and instead rely on the OS to clean up

swgillespie:
`GCEvent` is occasionally used in static contexts and there are a lot of platform-specific subtleties involved with the running of static destructors on process exit. It's safer if `GCEvent` just doesn't have a destructor at all and instead lets the OS clean up - there won't be any uses of `GCEvent` whose lifetime is shorter than the lifetime of the process.

Fixes https://github.com/dotnet/coreclr/issues/11111.

cc @Maoni0 @jkotas

### DIFF
--- a/src/gc/env/gcenv.os.h
+++ b/src/gc/env/gcenv.os.h
@@ -59,9 +59,6 @@ public:
     // Constructs a new uninitialized event.
     GCEvent();
 
-    // Destructs an event.
-    ~GCEvent();
-
     // Closes the event. Attempting to use the event past calling CloseEvent
     // is a logic error.
     void CloseEvent();
--- a/src/gc/env/gcenv.os.h
+++ b/src/gc/env/gcenv.os.h
@@ -50,6 +50,14 @@ struct GCThreadAffinity
 // An event is a synchronization object whose state can be set and reset
 // indicating that an event has occured. It is used pervasively throughout
 // the GC.
+//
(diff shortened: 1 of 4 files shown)


src/gc/env/gcenv.os.h
```diff
@@ -59,9 +59,6 @@ class GCEvent {
     // Constructs a new uninitialized event.
     GCEvent();
 
-    // Destructs an event.
```
jkotas: Comment may be useful.

swgillespie: done!

src/gc/env/gcenv.os.h
```diff
@@ -50,6 +50,14 @@ struct GCThreadAffinity
 // An event is a synchronization object whose state can be set and reset
 // indicating that an event has occured. It is used pervasively throughout
 // the GC.
+//
+// Note that GCEvent deliberately leaks its contents by not having a non-trivial destructor.
+// This is by design; since many uses of GCEvent have static lifetime, their destructors
```
jkotas: It needs to be all uses, otherwise we will have a leak.

swgillespie: Right - today GCEvent is embedded in classes that may not necessarily have static lifetime, but do in practice because their object graph is rooted by something with static lifetime (this is the case for `gc_heap` with Server GC). I'll be more clear about that.

PR #11134, opened 2017-04-21
Fix fgNewBBinRegion when inserting into filters.

pgavlin:
In order to ensure that the GC operates properly when handling
exceptions, the CLR ABI requires that control exits a filter region
throguh its terminal instruction. `fgNewBBinRegion` was violating this
invariant, however, which lead to GC stress failures when the GC was
invoked immediately after a filter finished executing but before control
entered any handlers. This change fixes the behavior of
`fgNewBBinRegion` when inserting new blocks into filter regions.

There are two cases to consider when inserting a new BB into a filter
region:
1. The filter region consists of multiple blocks
2. The filter region consists of a single block

The first case is straightforward: instead of inserting a new BB after
the last block of the filter, insert it before the last block of the
filter. Because the filter contains multiple blocks, the predecessor of
the filter must also be in the filter region.

The latter case requires splitting the single block, as inserting a new
block before the single block would otherwise change control flow (the
filter would begin with the new block rather than the single block). In
order to acheive this with minimal complexity, this change first inserts
a `BBJ_ALWAYS` block that transfers control to the existing single
block, then inserts a second new block immediately before the existing
single block. The second new block is then returned. To put it
visually, the transformation is from this:

```
    filter start: BBN (BBJ_EHFILTERRET)
```

to this:

```
    filter start: BBN + 1 (BBJ_ALWAYS -> BBN)
                  BBN + 2 (jumpKind)
                  BBN     (BBJ_EHFILTERRET)
```

and the function returns `BBN + 2`.

Fixes VSO 406163.


src/jit/flowgraph.cpp
```diff
@@ -17258,7 +17264,21 @@ BasicBlock* Compiler::fgFindInsertPoint(unsigned    regionIndex,
         bestBlk = goodBlk;
     }
 
-DONE:;
+DONE:
+
+    // If we are inserting into a filter and the best block is the end of the filter region, we need to
+    // insert after its predecessor instead: the CLR ABI that the terminal block of a filter region is
```
CarolEidt: Should be "the CLR ABI states that"?

src/jit/flowgraph.cpp
```diff
@@ -3189,9 +3189,15 @@ void Compiler::fgComputePreds()
         if (ehDsc->HasFilter())
         {
             ehDsc->ebdFilter->bbFlags |= BBF_JMP_TARGET | BBF_HAS_LABEL;
+
+            // The first block of a filter has an artifical extra refcount.
+            ehDsc->ebdFilter->bbRefs++;
```
CarolEidt: Why are these extra refcounts needed?

src/jit/flowgraph.cpp
pgavlin: Yes, good catch.

src/jit/flowgraph.cpp
pgavlin: They are expected by `fgExtendEHRegionBefore`.

pgavlin: To elaborate a bit, these blocks start out with an extra ref count. `fgComputePreds` only increments the ref count of a BB for formal incoming edges, which these blocks do not have. The relevant code in `fgExtendEHRegionBefore` is [here](https://github.com/dotnet/coreclr/blob/master/src/jit/flowgraph.cpp#L16588-L16590) and [here](https://github.com/dotnet/coreclr/blob/master/src/jit/flowgraph.cpp#L16627-L16629).

CarolEidt: Thanks - so I gather that this is not something that is new, but was previously omitted. I was trying to figure out how that was a new requirement.

pgavlin: > I gather that this is not something that is new, but was previously omitted
Yes, that's right.

src/jit/flowgraph.cpp
```diff
@@ -16569,10 +16575,18 @@ void Compiler::fgExtendEHRegionBefore(BasicBlock* block)
 #endif // DEBUG
             HBtab->ebdTryBeg = bPrev;
             bPrev->bbFlags |= BBF_TRY_BEG | BBF_DONT_REMOVE | BBF_HAS_LABEL;
+#if FEATURE_EH_FUNCLETS
+            bPrev->bbFlags |= BBF_FUNCLET_BEG;
+#endif // FEATURE_EH_FUNCLETS
+
```
BruceForstall: This doesn't look right. 'try' is not necessarily the beginning of a funclet

BruceForstall:
Why are you addressing the case "2. The filter region consists of a single block"? Does anyone use that? I thought we discussed that wasn't necessary. In particular, if there is a single `BBJ_EHFILTERRET` and a new block is being
created, then the caller must have split the block already, since they need to generate a BBJ_COND or BBJ_ALWAYS to branch to the new block.

src/jit/flowgraph.cpp
pgavlin: Yeah, I have a fix for this locally.

pgavlin:
> Why are you addressing the case "2. The filter region consists of a single block"? Does anyone use that?

Yes: `fgAddCodeRef` is called during `fgSimpleLowering` and may need to insert throw blocks into single-block filter regions.

BruceForstall:
Please run desktop SPMI diffs

src/jit/flowgraph.cpp
```diff
@@ -16569,6 +16569,15 @@ void Compiler::fgExtendEHRegionBefore(BasicBlock* block)
...
             bPrev->bbFlags |= BBF_TRY_BEG | BBF_DONT_REMOVE | BBF_HAS_LABEL;
+#if FEATURE_EH_FUNCLETS
+            if (fgFuncletsCreated && ((block->bbFlags & BBF_FUNCLET_BEG) != 0))
+            {
+                bPrev->bbFlags |= BBF_FUNCLET_BEG;
+                block->bbFlags &= ~BBF_FUNCLET_BEG;
+            }
+#endif // FEATURE_EH_FUNCLETS
+
+
```
BruceForstall: Seems like this isn't necessary, because if `block` is the beginning of a funclet, then it will also be the first block of a handler, and thus the handler check below will do the right thing. Plus, if you do this, I would think the handler check assert below would fail because you would have already cleared the bit.

src/jit/flowgraph.cpp
```diff
@@ -17258,7 +17288,21 @@ BasicBlock* Compiler::fgFindInsertPoint(unsigned    regionIndex,
...
     }
 
-DONE:;
+DONE:
+
+    // If we are inserting into a filter and the best block is the end of the filter region, we need to
+    // insert after its predecessor instead: the CLR ABI that the terminal block of a filter region is
+    // its exit block. If the filter region consists of a single block, a new block cannot be inserted
+    // without either splitting the single block before inserting a new block or inserting the new block
+    // before the single block and updating the filter description s.t. the inserted block is marked as
```
BruceForstall: >s.t. [](start = 67, length = 4)

Can you please spell out "s.t."?

src/jit/flowgraph.cpp
pgavlin: Done.

src/jit/flowgraph.cpp
pgavlin: Done.

src/jit/flowgraph.cpp
pgavlin: Done.

src/jit/flowgraph.cpp
```diff
@@ -17069,6 +17098,9 @@ BasicBlock* Compiler::fgFindInsertPoint(unsigned    regionIndex,
             regionIndex, dspBool(putInTryRegion), startBlk->bbNum, (endBlk == nullptr) ? 0 : endBlk->bbNum,
             (nearBlk == nullptr) ? 0 : nearBlk->bbNum, (jumpBlk == nullptr) ? 0 : jumpBlk->bbNum, dspBool(runRarely));
 
+    const bool insertingIntoFilter = !putInTryRegion && (startBlk == ehGetDsc(regionIndex - 1)->ebdFilter) &&
+                                     (endBlk == ehGetDsc(regionIndex - 1)->ebdHndBeg);
+
```
BruceForstall: I assume you hoisted this because it looked loop-insensitive. However, you've introduced a bug where we're accessing ebdFilter even if the EH region does not represent a filter, which is illegal (since it is a union; we really should assert if we could). Below, we can do this because we know at that point `blk == startBlk` and `blk` is the first block of a filter, and the function requires the caller to pass exactly a filter region as startBlk/endBlk in this case. You could also use `!putInTryRegion && ehGetDsc(regionIndex - 1)->HasFilter() && ...` (but now the number of duplicated `ehGetDsc()` is getting ridiculous)

src/jit/flowgraph.cpp
```diff
@@ -17437,6 +17481,21 @@ BasicBlock* Compiler::fgNewBBinRegion(BBjumpKinds jumpKind,
     // Now find the insertion point.
     afterBlk = fgFindInsertPoint(regionIndex, putInTryRegion, startBlk, endBlk, nearBlk, nullptr, runRarely);
 
+    // If afterBlk is nullptr, we must be inserting into a single-block filter region. Because the CLR ABI requires
+    // that control exists a filter via the last instruction in the filter range, this situation requires logically
```
BruceForstall: >exists [](start = 20, length = 6)

exists => exits

src/jit/flowgraph.cpp
pgavlin: Could this be rewritten as `const bool insertingIntoFilter = !putInTryRegion && (startBlk->bbCatchType == BBCT_FILTER) && (endBlk == ehGetDsc(regionIndex - 1)->ebdHndBeg);`?

src/jit/flowgraph.cpp
```diff
@@ -20246,6 +20305,19 @@ void Compiler::fgDebugCheckBBlist(bool checkBBNum /* = false */, bool checkBBRef
...
+            {
+                blockRefs++;
+            }
+            if (HBtab->HasFilter() && (HBtab->ebdFilter == block))
+            {
+                blockRefs++;
+            }
+        }
+
         if (checkBBNum)
```
BruceForstall: This looks expensive. Can you instead change the assert below from:

```
        /* Check the bbRefs */
        noway_assert(!checkBBRefs || block->bbRefs == blockRefs);
```
to:
```
if (checkBBRefs)
{
   if (block->bbRefs != blockRefs)
   {
        // Check to see if this block is the beginning of a filter or a handler and adjust the ref count appropriately.
        for (EHblkDsc *HBtab = compHndBBtab, *HBtabEnd = &compHndBBtab[compHndBBtabCount]; HBtab != HBtabEnd; HBtab++)
        {
            if (HBtab->ebdHndBeg == block)
            {
                blockRefs++;
            }
            if (HBtab->HasFilter() && (HBtab->ebdFilter == block))
            {
                blockRefs++;
            }
        }

        assert(block->bbRefs == blockRefs);
    }
}
```

Thus, it is only computed if required, on demand.

src/jit/flowgraph.cpp
BruceForstall: It seems like we should be checking to ensure the startBlk and endBlk are in the same EH table entry (even though the caller should presumably guarantee it)

---
In reply to: [113073328](https://github.com/dotnet/coreclr/pull/11134#discussion_r113073328) [](ancestors = 113073328)

pgavlin: Okay. How about
```
    bool insertingIntoFilter = false;
    if (!putInTryRegion)
    {
        EHblkDsc* const dsc = ehGetDsc(regionIndex - 1);
        insertingIntoFilter = dsc->HasFilter() && (startBlk == dsc->ebdFilter) && (endBlk == dsc->ebdHndBeg);
    }
```

BruceForstall:
Thanks for cleaning up the bbRefs inconsistencies!

src/jit/flowgraph.cpp
BruceForstall: That works.

src/jit/flowgraph.cpp
pgavlin: Done

src/jit/flowgraph.cpp
pgavlin: Done

pgavlin:
@BruceForstall desktop SPMI reported no asm diffs. I have verified that there are diffs on x86 in the expected cases, however.

pgavlin:

PR #11145, opened 2017-04-21
plumb runtime events into EventPipe

helloguo:
This PR plumbs runtime events into EventPipe.

For example, for event `GCStart`, the `FireEtwGCStart` will be modified as folowing:

```
inline BOOL EventEnabledGCStart() {return EventPipeEventEnabledGCStart() || (XplatEventLogger::IsEventLoggingEnabled() && EventXplatEnabledGCStart());}

inline ULONG FireEtwGCStart(
    const unsigned int  Count,
    const unsigned int  Reason
)
{
    ULONG status = EventPipeWriteEventGCStart(Count,Reason);
    if(XplatEventLogger::IsEventLoggingEnabled())
    {
        status &= FireEtXplatGCStart(Count,Reason);
    }
    return status;
}
```

And `EventPipeEventEnabledGCStart()` and `EventPipeWriteEventGCStart()` will be generated by script as shown here.
 
```

bool EventPipeEventEnabledGCStart()
{
    return EventPipeEventGCStart->IsEnabled();
}

extern "C" ULONG EventPipeWriteEventGCStart(
                  const unsigned int Count,
                  const unsigned int Reason)
{
    if (!EventPipeEventEnabledGCStart())
        return ERROR_SUCCESS;

    char stackBuffer[32];
    char *buffer = stackBuffer;
    unsigned int offset = 0;
    unsigned int size = 32;
    bool fixedBuffer = true;

    bool success = true;
    success &= WriteToBuffer(Count, buffer, offset, size, fixedBuffer);
    success &= WriteToBuffer(Reason, buffer, offset, size, fixedBuffer);

    if (!success)
    {
        if (!fixedBuffer)
            delete[] buffer;
        return ERROR_WRITE_FAULT;
    }

    EventPipe::WriteEvent(*EventPipeEventGCStart, (BYTE *)buffer, size);

    if (!fixedBuffer)
        delete[] buffer;

    return ERROR_SUCCESS;
}
```


helloguo:
@brianrob PTAL

src/scripts/genXplatLttng.py
```diff
@@ -593,7 +593,7 @@ def generateLttngFiles(etwmanifest,eventprovider_directory):
 
 bool WriteToBuffer(const BYTE *src, int len, char *&buffer, int& offset, int& size, bool &fixedBuffer)
 {
-	if (offset + len)
+	if (offset + len > size)
```
brianrob: Was this just a bug that was present in the implementation?

src/vm/eventpipe.cpp
```diff
@@ -29,6 +32,7 @@ void EventPipe::Initialize()
         (CrstFlags)(CRST_TAKEN_DURING_SHUTDOWN));
 
     s_pConfig = new EventPipeConfiguration();
+    InitProvidersAndEvents();
```
brianrob: Can you put a comment here that says that this calls into auto-generated code to initialize the runtime providers and events so that the EventPipe configuration lock isn't taken at runtime.

src/scripts/genXplatEventing.py
```diff
@@ -339,6 +338,12 @@ def generateClrallEvents(eventNodes,allTemplates):
         fnptype.extend(fnptypeline)
         fnptype.append("\n)\n{\n")
         fnbody.append(lindent)
+        fnbody.append("if (!EventEnabled")
```
brianrob: I don't think you need this call to EventEnabled<EventName>.  Here is some example generated code:

inline ULONG FireEtwGCStart(
                  const unsigned int  Count,
                  const unsigned int  Reason
)
{
                  if (!EventEnabledGCStart()) {return ERROR_SUCCESS;}
                  EventPipeWriteEventGCStart(Count,Reason);
                  return FireEtXplatGCStart(Count,Reason);
}

The call to EventEnabled is duplicate logic because the EventPipe write and the LTTng write will both do an IsEnabled check for the specific logging mechanism.  You should be able to safely remove this.  We still want the EventEnabled<Event> functions generated so that they can be called by VM code, they just don't need to be called here.

src/scripts/genXplatEventing.py
```diff
@@ -282,18 +282,17 @@ def generateClrallEvents(eventNodes,allTemplates):
         #generate EventEnabled
         clrallEvents.append("inline BOOL EventEnabled")
         clrallEvents.append(eventName)
-        clrallEvents.append("() {return XplatEventLogger::IsEventLoggingEnabled() && EventXplatEnabled")
-        clrallEvents.append(eventName+"();}\n\n")
+        clrallEvents.append("() {return ")
+        clrallEvents.append("EventPipeEventEnabled" + eventName + "() || ")
+        clrallEvents.append("(XplatEventLogger::IsEventLoggingEnabled() && EventXplatEnabled")
```
brianrob: The LTTng generated code right now looks like this:

extern "C" BOOL  EventXplatEnabledGCStart(){ return tracepoint_enabled(DotNETRuntime, GCStart); }

Because this change refactors the EventEnabled calls, the LTTng EventXplatEnabled<Event> functions need to check XplatEventLogger::IsEventLoggingEnabled() first.  I think that this may need to happen in the clretwallmain.h generated functions because XplatEventLogger is in the VM and this code is in the PAL.

build.sh
```diff
@@ -146,15 +146,27 @@ generate_event_logging_sources()
 # Event Logging Infrastructure
    __GeneratedIntermediate="$__IntermediatesDir/Generated"
    __GeneratedIntermediateEventProvider="$__GeneratedIntermediate/eventprovider_new"
+   __GeneratedIntermediateEventPipe="$__GeneratedIntermediate/eventpipe_new"
```
brianrob: I ran this change through the CI and got a build failure.  Can you double check this?  Log is at https://ci.dot.net/job/dotnet_coreclr/job/master/job/checked_windows_nt_bld_prtest/9133/consoleText.

brianrob:
FYI @valenis

src/scripts/genXplatLttng.py
helloguo: I think so.

src/vm/eventpipe.cpp
helloguo: Sure. Will do.

src/scripts/genXplatEventing.py
helloguo: Good point. How about something like this:

```
inline BOOL EventEnabledGCStart() {return EventPipeEventEnabledGCStart() || (XplatEventLogger::IsEventLoggingEnabled() && EventXplatEnabledGCStart());}

inline ULONG FireEtwGCStart(
                  const unsigned int  Count,
                  const unsigned int  Reason
)
{
                  ULONG status = EventPipeWriteEventGCStart(Count,Reason);
                  if(XplatEventLogger::IsEventLoggingEnabled())
                  {
                       status |= FireEtXplatGCStart(Count,Reason);   // just an example here
                  }
                  return status;
}
```

```
bool EventPipeEventEnabledGCStart()
{
    return EventPipeEventGCStart->IsEnabled();
}

extern "C" ULONG EventPipeWriteEventGCStart(
                  const unsigned int Count,
                  const unsigned int Reason)
{
    if (!EventPipeEventEnabledGCStart())
        return ERROR_SUCCESS;

    char stackBuffer[32];
    char *buffer = stackBuffer;
    unsigned int offset = 0;
    unsigned int size = 32;
    bool fixedBuffer = true;

    bool success = true;
    success &= WriteToBuffer(Count, buffer, offset, size, fixedBuffer);
    success &= WriteToBuffer(Reason, buffer, offset, size, fixedBuffer);

    if (!success)
    {
        if (!fixedBuffer)
            delete[] buffer;
        return ERROR_WRITE_FAULT;
    }

    EventPipe::WriteEvent(*EventPipeEventGCStart, (BYTE *)buffer, size);

    if (!fixedBuffer)
        delete[] buffer;

    return ERROR_SUCCESS;
}

```
```
extern "C" BOOL  EventXplatEnabledGCStart(){ return tracepoint_enabled(DotNETRuntime, GCStart); }

extern "C" ULONG  FireEtXplatGCStart(
                  const unsigned int Count,
                  const unsigned int Reason)
{
    if (!EventXplatEnabledGCStart())
        return ERROR_SUCCESS;

    // do trace point here

    return ERROR_SUCCESS;
}
```

brianrob: Yes, this is good.

build.sh
helloguo: Yes, it seems this error only happens on Windows platforms. I guess the reason is that in file eventpipeprovider.h line 16, the enumerate variable `Error ` may be defined somewhere else. If we use `enum class EventPipeEventLevel` instead of `enum EventPipeEventLevel`, this error will disappear. Of course, we need to use` EventPipeEventLevel::Critical` instead of` Critical` if we make this change. Make the change?

src/vm/eventpipe.cpp
helloguo: And put the guard `#ifdef PLATFORM_UNIX`?

brianrob: The guard should be #ifdef FEATURE_PAL.

build.sh
brianrob: Yes, please.

brianrob:
LGTM.  Can you please take a look at https://ci.dot.net/job/dotnet_coreclr/job/master/job/checked_osx10.12_prtest/2124/consoleText.  It looks like there is an OSX failure - a missing file etmdummy.h.  This does not appear to repro in the dev/perf_profiler branch.

brianrob:
Spoke with @helloguo and he is not able to reproduce this locally.  Given that, we'll merge this and investigate post-merge so that work can continue on top of this change.

Issue #7693, opened 2016-10-18
RyuJIT/x86: GC/API/GC/GetGenerationWR2 fails with GCStress=0xC


pgavlin:
e.g. https://ci.dot.net/job/dotnet_coreclr/job/master/job/jitstress/job/x86_ryujit_checked_windows_nt_gcstress0x3/8/

```
20:53:07 D:\j\workspace\x86_ryujit_ch---8fdd26c8\bin\tests\Windows_NT.x86.Checked\TestWrappers\GC\GC.XUnitWrapper.cs(3003): error : GC._API_GC_GetGenerationWR2_GetGenerationWR2_._API_GC_GetGenerationWR2_GetGenerationWR2_cmd [FAIL] [D:\j\workspace\x86_ryujit_ch---8fdd26c8\tests\runtest.proj]
20:53:07         
20:53:07   
20:53:07   Return code:      1
20:53:07   Raw output file:      D:\j\workspace\x86_ryujit_ch---8fdd26c8\bin\tests\Windows_NT.x86.Checked\Reports\GC\API\GC\GetGenerationWR2\GetGenerationWR2.output.txt
20:53:07   Raw output:
20:53:07   BEGIN EXECUTION
20:53:07          "D:\j\workspace\x86_ryujit_ch---8fdd26c8\tests\..\bin\tests\Windows_NT.x86.Checked\Tests\Core_Root\corerun.exe" GetGenerationWR2.exe 
20:53:07         Test should return with ExitCode 100 ...
20:53:07         Test Failed
20:53:07         Expected: 100
20:53:07         Actual: 1
20:53:07         END EXECUTION - FAILED
20:53:07         FAILED
20:53:07         Test Harness Exitcode is : 1
20:53:07         
20:53:07   To run the test:
20:53:07   > set CORE_ROOT=D:\j\workspace\x86_ryujit_ch---8fdd26c8\bin\tests\Windows_NT.x86.Checked\Tests\Core_Root
20:53:07   > D:\j\workspace\x86_ryujit_ch---8fdd26c8\bin\tests\Windows_NT.x86.Checked\GC\API\GC\GetGenerationWR2\GetGenerationWR2.cmd
20:53:07   
20:53:07         Expected: True
20:53:07         Actual:   False
[...]

PR #11151, opened 2017-04-21
Mark GetGenerationWR2 as GCStressIncompatible across all platforms

swgillespie:
Upon review, this test can behave poorly in the presence of particularly aggressive GCStress settings. The code that the JIT emits for this function is fully interruptible and, as such, it's possible for a GC to occur between the left and right hand sides of this expression:

```csharp
result = ( GC.GetGeneration( o ) == GC.GetGeneration( wf ));
```

which will cause the test to fail.

cc @sergiy-k @adiaaida , closes https://github.com/dotnet/coreclr/issues/7693


adiaaida:
LGTM

PR #11155, opened 2017-04-22
[NoMerge] Inline AsyncMethodBuilder Create+Start

benaadams:
Generally always called (and together) but not always inlined into callee

Contributes to https://github.com/dotnet/corefx/issues/18481


benaadams:
Start's Exception handling means it won't inline 😢

benaadams:
Lot of change for not huge gains

PR #11157, opened 2017-04-22
Simplify emitter's RegEncoding

mikedn:
VC++ does't recognize that `(c ? x - 16 : x) & 7` is just `x & 7` and we end up with 5 instructions instead of 1.


mikedn:
@BruceForstall Just a small perf glitch I noticed when I profiled with inlining disabled.

Before we had:
```asm
lea     eax, [r8-16]
cmp     eax, 15
lea     eax, [r8-16]
cmovbe  r8d, eax
and     r8d, 7
```
and now we have:
```asm
and     r8d, 7
```

BruceForstall:
cc @dotnet/jit-contrib

BruceForstall:
LGTM

PR #11161, opened 2017-04-23
Add "createdump" command to SOS.

mikem8361:
Uses the createdump utility code as a library.  TODO: the float point registers for the threads are not saved in a dump created with the SOS command.

Changed the "-m, --micro" option to "-t, --triage".

Lots of cleanup.

### DIFF
--- a/src/debug/createdump/main.cpp
+++ b/src/debug/createdump/main.cpp
@@ -1,0 +1,99 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#include "createdump.h"
+
+const char* g_help = "createdump [options] pid\n" 
+"-f, --name - dump path and file name. The pid can be placed in the name with %d. The default is '/tmp/coredump.%d'\n"
+"-n, --normal - create minidump.\n"
+"-h, --withheap - create minidump with heap (default).\n" 
+"-t, --triage - create triage minidump.\n" 
+"-d, --diag - enable diagnostic messages.\n";
+
+bool CreateDumpCommon(const char* programPath, const char* dumpPathTemplate, MINIDUMP_TYPE minidumpType, CrashInfo* crashInfo);
+
+//
+// Main entry point
+//
+int __cdecl main(const int argc, const char* argv[])
+{
+    MINIDUMP_TYPE minidumpType = MiniDumpWithPrivateReadWriteMemory;
+#ifdef ANDROID
+    const char* dumpPathTemplate = "/data/local/tmp/coredump.%d";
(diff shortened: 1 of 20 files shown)


src/inc/torelease.h
```diff
@@ -0,0 +1,77 @@
...
+// you place in it when the ToRelease class falls out of scope.  You may use it
+// just like you would a standard pointer to a COM object (including if (foo),
+// if (!foo), if (foo == 0), etc) except for two caveats:
+//     1. This class never calls AddRef and it always calls Release when it
+//        goes out of scope.
+//     2. You should never use & to try to get a pointer to a pointer unless
+//        you call Release first, or you will leak whatever this object contains
+//        prior to updating its internal pointer.
+template<class T>
+class ToRelease
```
janvorli: It would be nice to rename it to ReleaseHolder to match the same thing that we have in the holder.h for usage in the runtime.

src/debug/createdump/main.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+bool CreateDumpCommon(const char* programPath, const char* dumpPathTemplate, MINIDUMP_TYPE minidumpType, CrashInfo* crashInfo);
+
+//
+// Main entry point
+//
+int __cdecl main(const int argc, const char* argv[])
+{
+    MINIDUMP_TYPE minidumpType = MiniDumpWithPrivateReadWriteMemory;
+    const char* dumpPathTemplate = "/tmp/coredump.%d";
```
janvorli: It would be nice to conditionalize the tmp path for Android as we do in PAL to be future proof. Android doesn't have /tmp folder, it has /data/local/tmp instead.
This is a related piece of code from palinternal.h:
```C++
#ifndef __ANDROID__
#define TEMP_DIRECTORY_PATH "/tmp/"
#else
// On Android, "/tmp/" doesn't exist; temporary files should go to
// /data/local/tmp/
#define TEMP_DIRECTORY_PATH "/data/local/tmp/"
#endif
```

src/inc/torelease.h
mikem8361: I don't like the name either. I moved this template from SOS; in SOS it has to be a different name not to conflict with the runtime one in inc\holder.h.  SOS and createdump can't use the runtime holder because it pulls in and depends on other runtime macros and templates.

I renamed the file/macro to releaseholder.h/ReleaseHolder<> but backed out my change to SOS's util.h to include its ToRelease<> macro.

src/debug/createdump/main.cpp
mikem8361: I made the change even though I don't think we will support this on Android in the future. I don't think we would create a dump file on your phone (but you never know).

mikem8361:

PR #11164, opened 2017-04-23
Support FEATURE_DBGIPC override

parjong:
This commit supports FEATURE_DBGIPC override via cmakearg option. This commit does not change the default behavior.


parjong:
@dotnet-bot test Windows_NT arm Cross Debug Build  please

parjong:
There are two motivations.
 - Tizen has a bit strict policy on /tmp. The pipe access (created under /tmp) emits many kernel-level warnings, which makes system debugging difficult.
 - There are too many remaining debugging pipes under my /tmp (Especially when I debug CLR using gdb).

janvorli:
@parjong thank you for the details.

PR #11165, opened 2017-04-24
[WIP][x86/Linux] Improve R/PInvoke Performance

### LINKED ISSUE #10225: [x86/Linux] Reverse P/Invoke Performance
x86 native and managed code uses different stack layout (which results in #10161) 

#10217 attempts to fix #10161, but degrades reverse P/Invoke performance. 

We need to improve reverse P/Invoke performance.

parjong:
This commit revises reverse P/Invoke callbacks to simply loop over shuffle descriptions (created during during run-time initialization) instead of method signature in order to fix #10225.

### DIFF
--- a/src/vm/dllimportcallback.h
+++ b/src/vm/dllimportcallback.h
@@ -212,7 +212,7 @@ public:
         UINT32 Edx;
     };
 
-    VOID SetupArguments(char *pSrc, ArgumentRegisters *pArgRegs, char *pDst);
+    VOID ShuffleArguments(char *pSrc, ArgumentRegisters *pArgRegs, char *pDst);
 #endif // _TARGET_X86_ && FEATURE_STUBS_AS_IL
 
 private:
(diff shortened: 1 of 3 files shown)


parjong:

src/vm/dllimportcallback.h
```diff
@@ -225,6 +225,24 @@ class UMThunkMarshInfo
     UINT16            m_cbRetPop;           // stack bytes popped by callee (for UpdateRegDisplay)
 #if defined(FEATURE_STUBS_AS_IL)
     UINT32            m_cbStackArgSize;     // stack bytes pushed for managed code
+
+    struct ShuffleDescription
```
jkotas: This should be under x86 ifdef because of it is x86 specific hack.

jkotas:
This is ok incremental improvement, but Linux x86 is still going to be quite a bit slower compared to other platforms. Have you looked into what it would take to get rid of this shuffling completely, and make x86 work like other platforms?

src/vm/dllimportcallback.h
parjong: This is already under x86 ifdef.

parjong:
@jkotas I once tried to change the argument passing order, but it turns out that the whole JIT should be revised to support that. I'm not sure about the amount of work, but it seems to be huge.

parjong:
It seems that this commit will introduce several assert failure on FX unittest. I'll first analyze these assertion failures, and will be back on this PR.

jkotas:
> it turns out that the whole JIT should be revised to support that.

Maybe you can leave the JIT as is, and instead adjust the IL for the stub: add dummy arguments for things passed in registers; and reverse the order of the remaining arguments by changing `m_paramidx + argOffset` at 
https://github.com/dotnet/coreclr/blob/52a816d3011f4d03b80b5940dc036b40d701f52d/src/vm/mlinfo.cpp#L3287 to be something like `m_totalArguments - (m_paramidx + argOffset)`.

parjong:
Thank you for comment. I'll try that first.

Issue #10792, opened 2017-04-07
AssemblyLoadContext.LoadFromAssemblyPath does not set the file name when throwing FileNotFoundException 


MissaouiChedy:
The ` Assembly AssemblyLoadContext.LoadFromAssemblyPath(string path)` instance method throws a `FileNotFoundException` when it can not find the supplied `path`.

However, the `Filename` property in the thrown exception is `null` and it would be better if it was populated with the path of the file not found.

PR #11166, opened 2017-04-24
Set FileName property when raising FileNotFoundException in ALC

gkhanna79:
Fixes https://github.com/dotnet/coreclr/issues/10792

@jkotas PTAL


jkotas:
This adds the file name to the file not found exceptions, but not others. For example, if you load invalid binary (e.g. pass a path to the text file), you won't get the file path in the BadImageFormatException with this change. It is worse experience compared to desktop - you will get the path in BadImageFormatException message on desktop.

Should we attach the file name where the file is actually opened instead, so that we get the filename for all cases?

It would also avoid hitting the file system multiple times that this change does. It is not the end of the world, but it would be nice to avoid it.

gkhanna79:
Closed in favor https://github.com/dotnet/coreclr/pull/11238

PR #11167, opened 2017-04-24
[x86/Linux] Remove unnecessary ResumeEsp field

parjong:
#10749 (which revises CLR to fix up SP just before resuming) eliminates the need for tracking resume SP for each stack unwind step.

This commit eliminates unnecessary resume SP tracking code.


parjong:
@janvorli Please take a look.

Issue #11128, opened 2017-04-21
ARM32 AltJit is broken


mskvortsov:
`JIT/CodeGenBringUpTests` under `AltJit=* AltJitAssertOnNYI=0` after #10846 got merged:
```
=======================
     Test Results
=======================
# CoreCLR Bin Dir  : 
# Tests Discovered : 158
# Passed           : 94
# Failed           : 64
# Skipped          : 0
=======================
```
```
Assert failure(PID 11964 [0x00002ebc], Thread: 11964 [0x2ebc]): Assertion failed 'type[inx] == getJitGCType((BYTE)currentGcLayoutTy

    File: /home/mskvortsov/git/coreclr/src/jit/morph.cpp Line: 4912
```
The full testsuite run is entirely broken, almost every test fails with
```
Assert failure(PID 13525 [0x000034d5], Thread: 13525 [0x34d5]): Assertion failed 'foundStructArg' in 'System.String:Format(ref,ref,ref,ref,ref):ref' (IL size 17)
               
    File: /home/mskvortsov/git/coreclr/src/jit/morph.cpp Line: 4699
```
Before it was like
```
=======================
     Test Results
=======================
# CoreCLR Bin Dir  :
# Tests Discovered : 7086
# Passed           : 6170
# Failed           : 185
# Skipped          : 731
=======================
71 minutes and 37 seconds taken to run CoreCLR tests.
```
cc @dotnet/arm32-contrib

BruceForstall:
@hseok-oh can you please investigate?

PR #11169, opened 2017-04-24
[ARM32/RyuJIT] Fix failure related with struct argument passing

hseok-oh:
Block using NYI_ARM in morphing phase when struct value split between registers and stack
Fix type assign error for struct slot

related issue: #11128 
cc/ @dotnet/arm32-contrib


hseok-oh:

BruceForstall:
cc @dotnet/jit-contrib

PR #11185, opened 2017-04-24
Fix few clang analyzer warnings

janvorli:
This change fixes two issues reported by the clang analyzer. In both cases, uninitialized
variable value was used in an error code path.

### DIFF
--- a/src/pal/src/file/disk.cpp
+++ b/src/pal/src/file/disk.cpp
@@ -68,6 +68,7 @@ GetDiskFreeSpaceW(
     PathCharString dirNameBufferPathString;
     size_t length;
     char * dirNameBuffer;
+    const char * dirName;
     int size;
 
     PERF_ENTRY(GetDiskFreeSpaceW);
@@ -125,7 +126,7 @@ GetDiskFreeSpaceW(
         if ( size != 0 )
         {
             FILEDosToUnixPathA( dirNameBuffer );
-            statfsRetVal = statfs( dirNameBuffer, &fsInfoBuffer );
+            dirName = dirNameBuffer;
         }
         else
         {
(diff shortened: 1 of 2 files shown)


janvorli:

src/pal/src/file/disk.cpp
```diff
@@ -136,9 +137,11 @@ GetDiskFreeSpaceW(
     }
     else
     {
-        statfsRetVal = statfs( "/", &fsInfoBuffer );
+        dirName = "/";
```
gkhanna79: So the warning was about using a local to do the work?

src/pal/src/file/file.cpp
```diff
@@ -2999,8 +2999,11 @@ OUT  PLARGE_INTEGER lpFileSize)
             &dwFileSizeHigh
             );
 
-        lpFileSize->u.LowPart = dwFileSizeLow;
-        lpFileSize->u.HighPart = dwFileSizeHigh;
+        if (NO_ERROR == palError)
```
gkhanna79: What was this warning about?

src/pal/src/file/disk.cpp
janvorli: The warning was that in case the statfs here failed, we were using uninitialized `dirNameBuffer` variable in `FILEGetProperNotFoundError( dirNameBuffer, &dwLastError );`. Since `dirNameBuffer` is not a `const char *`, I have added the `dirName` that is const so that I can assign a pointer to the literal "/" to it without casting off the constness.

src/pal/src/file/file.cpp
janvorli: This one was caused by assigning uninitialized `dwFileSizeLow` to `lpFileSize->u.LowPart` in case the call to `InternalGetFileSize` failed. In that case, the `InternalGetFileSize` doesn't set its output parameter and so the `dwFileSizeLow` is a garbage.

janvorli:

Issue #3710, opened 2016-03-15
Enable Ready2Run test runs in Helix


gkhanna79:

gkhanna79:
Assigning to @wtgodbe since you are driving Helix integration.

PR #11187, opened 2017-04-24
Enable running R2R tests in Helix on non-Windows

wtgodbe:
This PR enables R2R tests for non-Windows in Helix by running crossgen on all CoreFX assemblies in Core_Root, in Helix, right before we begin executing. We touch a file named 'crossgenCompleted.mrk' inside of Core_Root once this process is done, then check for the existence of that file before trying to do it again (this way we don't unnecessarily CrossGen anything more than once).

Also, this will attempt to Crossgen every Core_Root .dll without the string '-ms-win-' in the name, but will fail harmlessly against non-FX .dll's. Since the time cost of crossgen-ing a .dll is pretty low, this is an easy way to solve the problem of only wanting to Crossgen CoreFX binaries (i.e. CrossGen will fail for other .dll's, but the failure will be swallowed).

Pending this going in, I'll open an issue against myself to revert this change once we can build tests on Linux, as there will be a better solution available at that time. 

Local testing showed tests passing, and stuff only getting crossgen'd once.

CC @MattGal @gkhanna79 

Resolves https://github.com/dotnet/coreclr/issues/3710

### DIFF
--- a/tests/helixprep.proj
+++ b/tests/helixprep.proj
@@ -110,7 +110,7 @@ EXIT /B %ERRORLEVEL%
     <PropertyGroup>
       <WrapperShContents>#!/bin/sh%0a</WrapperShContents>
       <WrapperShContents>$(WrapperShContents)%0a</WrapperShContents> 
-      <WrapperShContents>$(WrapperShContents)export CORE_ROOT="$HELIX_CORRELATION_PAYLOAD"%0a</WrapperShContents>
+      <WrapperShContents>$(WrapperShContents)export CORE_ROOT="%24HELIX_CORRELATION_PAYLOAD"%0a</WrapperShContents>
       <WrapperShContents>$(WrapperShContents)%0a</WrapperShContents>
       <WrapperShContents>$(WrapperShContents)for scriptFilePath in %24(find . -type f -iname '%2A.sh' ! -iname "runtests.sh" | sort)%0a</WrapperShContents>
       <WrapperShContents>$(WrapperShContents)do%0a</WrapperShContents>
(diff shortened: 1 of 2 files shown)


tests/helixprep.proj
```diff
@@ -110,14 +110,40 @@ EXIT /B %ERRORLEVEL%
...
+      <WrapperShContents>$(WrapperShContents)export CORE_ROOT="%24HELIX_CORRELATION_PAYLOAD"%0a</WrapperShContents>
       <WrapperShContents>$(WrapperShContents)%0a</WrapperShContents>
       <WrapperShContents>$(WrapperShContents)for scriptFilePath in %24(find . -type f -iname '%2A.sh' ! -iname "runtests.sh" | sort)%0a</WrapperShContents>
       <WrapperShContents>$(WrapperShContents)do%0a</WrapperShContents>
       <WrapperShContents>$(WrapperShContents) perl -pi -e 's/\r\n|\n|\r/\n/g' "%24scriptFilePath"%0a</WrapperShContents>
       <WrapperShContents>$(WrapperShContents) chmod +x "%24scriptFilePath"%0a</WrapperShContents>
       <WrapperShContents>$(WrapperShContents)done%0a</WrapperShContents>
       <WrapperShContents>$(WrapperShContents)%0a</WrapperShContents>
+
+      <!-- Crossgen FX assemblies, if running R2R tests.
```
gkhanna79: Have you tested running this in helix?

wtgodbe: @gkhanna79 yes, the tests I looked at were all passing.

wtgodbe:
Just kicked off https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_Build/explorer?_a=summary&buildId=700711, I'll keep an eye on the tests.

PR #11188, opened 2017-04-24
Allow FEATURE_INTERPRETER to be enabled by command-line option

parjong:


BruceForstall:
Who is going to define the cmake `FEATURE_INTERPRETER` variable? What platforms/configurations do you expect to enable the interpreter?

BruceForstall:
cc @dotnet/jit-contrib

parjong:
@BruceForstall I would like to take a look it just for fun.

parjong:
To be more specific, I recently read [this article](http://mattwarren.org/2017/03/30/The-.NET-IL-Interpreter/ ) and am interested in the interpreter internal.

BruceForstall:
Presumably you could just add the
```
add_definitions(-DFEATURE_INTERPRETER)
```
in a private branch. Why do you need to merge this change? I'm no cmake expert: how does merging this change make it easier?

Bottom-line: I don't want there to be any chance the interpreter is accidentally compiled in by default.

parjong:
This PR allows users to enable FEATURE_INTERPERTER using the following command-line:
```
$ ./build.sh cmakeargs "-DFEATURE_INTERPRETER=1"
```
I want to build and test this feature without additional patch, but I also do not want to turn it default. I updated PR to prevent accidental enable that you are currently worry. 

Now, FEATURE_INTERPRETER will be turned off by default if the additional command-line option mentioned above is not provided.

mattwarren:
@parjong glad your found the blog post useful, it was fun to write!]

BTW if you want to get all the statistics printed out at the end of a program run, you also need to make the following changes in `interpreter.h`, I didn't make that clear in the post

``` diff
// If this is set, we keep track of extra information about IL instructions executed per-method.
-#define INTERP_PROFILE 0
+#define INTERP_PROFILE 1

// If this is set, we track the distribution of IL instructions.
-#define INTERP_ILINSTR_PROFILE 0
+#define INTERP_ILINSTR_PROFILE 1
```

parjong:
@mattwarren Thanks you for comment :+1: I'll try.

PR #11192, opened 2017-04-25
[WIP] Move return blocks out of loops

AndyAyersMS:
Low-tech approach to moving return blocks out of loops. PR'ing this just so I can run perf via CI and see if it causes any correctness ripples. No merge.

Doesn't move the returns far enough sometimes, and leaves back to back branches in places, but layout in the various collection `Find` methods looks reasonable.

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -14634,12 +14634,18 @@ void Compiler::fgReorderBlocks()
             continue;
         }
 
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@dotnet-bot test windows_nt x64 perf

AndyAyersMS:
Perf results from above show some promise; left notes over in #9692.

PR #11193, opened 2017-04-25
Move DynamicHelperFrame::UpdateRegDisplay into cgenx86.cpp

parjong:
It is currently implemented in frames.h, but cgenx86.cpp seems to be an appropriate place.


parjong:
@janvorli Please take a look

Issue #11101, opened 2017-04-20
[x86/Linux] JIT.Methodical.Boxing.seh._dbgfilter fail (GCStress=3)


wateret:
With GCStress=3 JIT.Methodical.Boxing.seh._dbgfilter fails on x86/Linux.

```
OK
Current totals 0 and 0 and 0
OK
Current totals 0 and 0.314159265358979 and 0.309016994374947
OK
Current totals 0 and 0.942477796076938 and 0.896802246667421
OK
Current totals 0 and 1.88495559215388 and 1.70581924104237
Current totals 0.0201 and 3.14159265358979 and 1.70581924104237

Assert failure(PID 1 [0x00000001], Thread: 1 [0x0001]): !CREATE_CHECK_STRING(pMT && pMT->Validate())
    File: /home/hanjoung/ws/dotnet/coreclr/src/vm/object.cpp Line: 1719
    Image: /work/nfs/dotnet/to/corerun
```

wateret:
Part of Bactrace

```
#9  0xf6d12095 in DbgAssertDialog (szFile=0xf79ed430 "/home/hanjoung/ws/dotnet/coreclr/src/vm/object.cpp", iLine=1719, szExpr=0xf79ed79c "!CREA
TE_CHECK_STRING(pMT && pMT->Validate())") at /home/hanjoung/ws/dotnet/coreclr/src/utilcode/debug.cpp:735
#10 0xf6e6c36a in Object::ValidateInner (this=0xf249476c, bDeep=1, bVerifyNextHeader=1, bVerifySyncBlock=1) at /home/hanjoung/ws/dotnet/coreclr
/src/vm/object.cpp:1719
#11 0xf6e6ad29 in Object::Validate (this=0xf249476c, bDeep=1, bVerifyNextHeader=1, bVerifySyncBlock=1) at /home/hanjoung/ws/dotnet/coreclr/src/
vm/object.cpp:1695
#12 0xf6e6aaaa in Object::ValidatePromote (this=0xf249476c, sc=0xffffa668, flags=4) at /home/hanjoung/ws/dotnet/coreclr/src/vm/object.cpp:1314
#13 0xf7584fe2 in WKS::GCHeap::Promote (ppObject=0xffffcb58, sc=0xffffa668, flags=4) at /home/hanjoung/ws/dotnet/coreclr/src/gc/gc.cpp:33956
[...]

PR #11195, opened 2017-04-25
[x86/Linux] Do not report untracked from filters

wateret:
Related Issue : #11101


wateret:
@janvorli Please take a look.

This is corresponding code with https://github.com/dotnet/coreclr/blob/03d0d244bca38053d54eef6f4b6db51ae87f7700/src/vm/eetwain.cpp#L5008-L5016 (USE_GCINFO_ENCODER version)

PR #11064, opened 2017-04-19
[ARM32] clang 3.9 as a default for ARM cross build

hqueue:
As discussed in https://github.com/dotnet/core-setup/issues/1411, change default compiler for ARM cross build to clang 3.9.

Use clang 3.9 as a default compiler for ARM cross build

* Both arm and armel will be built using clang 3.9 as a default
* ARM CI will use clang 3.9 as a default
* ARM pipieline build will use clang 3.9 as a default

PR #11197, opened 2017-04-25
[release/2.0.0][ARM32] clang 3.9 as a default for ARM cross build

hqueue:
*This is a duplicate of https://github.com/dotnet/coreclr/pull/11064 in master branch*

As discussed in https://github.com/dotnet/core-setup/issues/1411, change default compiler for ARM cross build to clang 3.9.

Use clang 3.9 as a default compiler for ARM cross build

* Both arm and armel will be built using clang 3.9 as a default
* ARM CI will use clang 3.9 as a default
* ARM pipieline build will use clang 3.9 as a default

### DIFF
--- a/Documentation/building/linux-instructions.md
+++ b/Documentation/building/linux-instructions.md
@@ -18,8 +18,8 @@ Toolchain Setup
 Install the following packages for the toolchain: 
 
 - cmake 
-- llvm-3.5 
-- clang-3.5 
+- llvm-3.5 (llvm-3.9 for ARM cross build)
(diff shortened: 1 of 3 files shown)


hqueue:
@qmfrederik It is also being discussed at https://github.com/dotnet/core-setup/issues/1411 about using clang 3.9 for other architectures, including x64.

qmfrederik:
@hqueue Thanks, I updated dotnet/core-setup#1411. My question is about arm64 (aarch64); I think it's a big strange to make the change for arm and armel but not arm64 (which may need it the most).

gkhanna79:

Issue #10780, opened 2017-04-06
NullReferenceException during 'dotnet build' on ubuntu arm32


eerhardt:
It appears there is a JIT bug on ubuntu arm machines. While trying to get a .NET Core SDK that runs on my Raspberry Pi, I encountered an error:

```
Microsoft (R) Build Engine version 15.2.93.5465
Copyright (C) Microsoft Corporation. All rights reserved.

/home/ubuntu/dotnet/sdk/2.0.0-preview1-005720/Sdks/Microsoft.NET.Sdk/build/Microsoft.NET.Sdk.targets(93,5): error MSB4018: The "GenerateDepsFile" task failed unexpectedly. [/home/ubuntu/HelloWorld/HelloWorld.csproj]
/home/ubuntu/dotnet/sdk/2.0.0-preview1-005720/Sdks/Microsoft.NET.Sdk/build/Microsoft.NET.Sdk.targets(93,5): error MSB4018: System.NullReferenceException: Object reference not set to an instance of an object. [/home/ubuntu/HelloWorld/HelloWorld.csproj]
/home/ubuntu/dotnet/sdk/2.0.0-preview1-005720/Sdks/Microsoft.NET.Sdk/build/Microsoft.NET.Sdk.targets(93,5): error MSB4018:    at Microsoft.NET.Build.Tasks.LockFileLookup..ctor(LockFile lockFile) [/home/ubuntu/HelloWorld/HelloWorld.csproj]
/home/ubuntu/dotnet/sdk/2.0.0-preview1-005720/Sdks/Microsoft.NET.Sdk/build/Microsoft.NET.Sdk.targets(93,5): error MSB4018:    at Microsoft.NET.Build.Tasks.DependencyContextBuilder.Build() [/home/ubuntu/HelloWorld/HelloWorld.csproj]
```

Looking into that code, there was no way a NullReferenceException was coming from that method.

Using Console.WriteLines, I traced it down to [this exact line that was failing](https://github.com/dotnet/sdk/blob/09c2a6bd8e2c0ec1524361d3a560a1ef961e4367/src/Tasks/Microsoft.NET.Build.Tasks/LockFileLookup.cs#L26).  Note that `LibraryType` is a struct.

I have 2 repos of this bug, one is the long way and one is the short way.

### Short Repo
- Clone https://github.com/eerhardt/ArmJITBug
[...]

Issue #7147, opened 2016-09-12
JIT: Loop hoisting re-ordering exceptions


JosephTremoulet:
Loop-hoisting optimization mentions in a [comment](https://github.com/dotnet/coreclr/blob/cbd1f37b7b74647644686a9143d05fbdd15162c8/src/jit/optimizer.cpp#L6225-L6226) that "We assume that expressions will be hoisted so that they are evaluated in the same order as they would have been in the loop," but that assumption doesn't hold with the current implementation.  Consider the following code:

```c#
using System;
using System.Runtime.CompilerServices;

namespace N
{
    public class C
    {
        int f = 2;

        [MethodImpl(MethodImplOptions.NoInlining)]
        static bool cross(int k, int y, C c)
        {
            bool b = false;

            for (int i = 0; i < k; ++i)
            {
                // Here "c.f" is invariant, but is evaluated after "i / y"
                // which may throw a different kind of exception, so can't
                // be hoisted without potentially changing the exception kind
                // thrown.
                b = (i / y < i + c.f);
            }

            return b;
        }

        [MethodImpl(MethodImplOptions.NoInlining)]
        static bool swap(int k, int x, int y, C c)
        {
            bool b = false;

            for (int i = 0; i < k; ++i)
            {
                // Sub-expressions "x / y" and "c.f" are both invariant
                // w.r.t. this loop, and can be hoisted.  Since each can
                // raise an exception, and the exceptions have different
                // types, their relative order must be preserved -- the
                // hoisted copy of "x / y" must be evaluated before the
                // hoisted copy of "c.f"
                b = (x / y < i + c.f);
            }

            return b;
        }

        public static int Main(string[] args)
        {
            int errors = 0;

            try
            {
                cross(10, 0, null);
                // DivByZero should be raised from 'swap'; normal return
                // is an error.
                ++errors;
            }
            catch (DivideByZeroException)
            {
                // This is the correct result -- i / y should be evaluated and
                // raise this exception (before c.f raises nulllref).
            }
            catch
            {
                // Any exception other than DivideByZero is a failure
                ++errors;
            }

            try
            {
                swap(10, 11, 0, null);
                // DivByZero should be raised from 'swap'; normal return
                // is an error.
                ++errors;
            }
            catch (DivideByZeroException)
            {
                // This is the correct result -- x / y should be evaluated and
                // raise this exception (before c.f raises nulllref).
            }
            catch
            {
                // Any exception other than DivideByZero is a failure
[...]

PR #11198, opened 2017-04-25
Disable loop invariant code hoisting for referece GT_CLS_VAR

hseok-oh:
Disable optimizing loop invariant code hoisting for GT_CLS_VAR completely 

related issue: #10780

### DIFF
--- a/src/jit/optimizer.cpp
+++ b/src/jit/optimizer.cpp
@@ -6204,9 +6204,10 @@ bool Compiler::optHoistLoopExprsForTree(
             treeIsHoistable = false;
         }
 
-        // Avoid to hoist reference loading
-        if (tree->OperGet() == GT_CLS_VAR && tree->gtType == TYP_REF)
+        if (tree->OperGet() == GT_CLS_VAR)
         {
+            // It is unsafe to mark this as invariant,
+            // as we currently don't prevent it from being hoisted above the call to the class initializer.
             treeIsInvariant = false;
         }
     }
--- a/src/jit/optimizer.cpp
+++ b/src/jit/optimizer.cpp
@@ -6184,14 +6184,6 @@ bool Compiler::optHoistLoopExprsForTree(
                     treeIsHoistable = false;
                 }
             }
-            // Currently we must give up on reads from static variables (even if we are in the first block).
-            //
-            if (tree->OperGet() == GT_CLS_VAR)
-            {
-                // TODO-CQ: test that fails if we hoist GT_CLS_VAR: JIT\Directed\Languages\ComponentPascal\pi_r.exe
-                // method Main
-                treeIsHoistable = false;
-            }
         }
 
         // Is the value of the whole tree loop invariant?
@@ -6202,6 +6194,13 @@ bool Compiler::optHoistLoopExprsForTree(
         if (!treeIsInvariant)
         {
             treeIsHoistable = false;
+        }
+
+        // Currently we must give up on reads from static variables (even if we are in the first block).
+        //
+        if (tree->OperGet() == GT_CLS_VAR)
+        {
+            treeIsHoistable = treeIsInvariant = false;
         }
     }
 
--- a/src/jit/optimizer.cpp
+++ b/src/jit/optimizer.cpp
@@ -6193,22 +6193,24 @@ bool Compiler::optHoistLoopExprsForTree(
                 treeIsHoistable = false;
             }
         }
-
-        // Is the value of the whole tree loop invariant?
-        treeIsInvariant =
-            optVNIsLoopInvariant(tree->gtVNPair.GetLiberal(), lnum, &hoistCtxt->m_curLoopVnInvariantCache);
-
-        // Is the value of the whole tree loop invariant?
-        if (!treeIsInvariant)
-        {
-            treeIsHoistable = false;
-        }
-
+        
         if (tree->OperGet() == GT_CLS_VAR)
         {
             // It is unsafe to mark this as invariant,
             // as we currently don't prevent it from being hoisted above the call to the class initializer.
             treeIsInvariant = false;
+        }
+        else
+        {
+            // Is the value of the whole tree loop invariant?
+            treeIsInvariant =
+                optVNIsLoopInvariant(tree->gtVNPair.GetLiberal(), lnum, &hoistCtxt->m_curLoopVnInvariantCache);
+        }
+
+        // Is the value of the whole tree loop invariant?
+        if (!treeIsInvariant)
+        {
+            treeIsHoistable = false;
         }
     }
 
--- a/src/jit/optimizer.cpp
+++ b/src/jit/optimizer.cpp
@@ -6193,7 +6193,7 @@ bool Compiler::optHoistLoopExprsForTree(
                 treeIsHoistable = false;
             }
         }
-        
+
         if (tree->OperGet() == GT_CLS_VAR)
         {
             // It is unsafe to mark this as invariant,
--- a/src/jit/optimizer.cpp
+++ b/src/jit/optimizer.cpp
@@ -6205,9 +6205,9 @@ bool Compiler::optHoistLoopExprsForTree(
         }
 
         // Avoid to hoist reference loading
-        if (tree->OperGet() == GT_CLS_VAR && tree->gtType == TYP_BYREF)
+        if (tree->OperGet() == GT_CLS_VAR && tree->gtType == TYP_REF)
         {
-            treeIsHoistable = treeIsInvariant = false;
+            treeIsInvariant = false;
         }
     }
 
--- a/src/jit/optimizer.cpp
+++ b/src/jit/optimizer.cpp
@@ -6184,6 +6184,14 @@ bool Compiler::optHoistLoopExprsForTree(
                     treeIsHoistable = false;
                 }
             }
+            // Currently we must give up on reads from static variables (even if we are in the first block).
+            //
+            if (tree->OperGet() == GT_CLS_VAR)
+            {
+                // TODO-CQ: test that fails if we hoist GT_CLS_VAR: JIT\Directed\Languages\ComponentPascal\pi_r.exe
+                // method Main
+                treeIsHoistable =  false;
+            }
         }
 
         // Is the value of the whole tree loop invariant?
@@ -6196,9 +6204,8 @@ bool Compiler::optHoistLoopExprsForTree(
             treeIsHoistable = false;
         }
 
-        // Currently we must give up on reads from static variables (even if we are in the first block).
-        //
-        if (tree->OperGet() == GT_CLS_VAR)
+        // Avoid to hoist reference loading
+        if (tree->OperGet() == GT_CLS_VAR && tree->gtType == TYP_BYREF)
         {
             treeIsHoistable = treeIsInvariant = false;
         }
--- a/src/jit/optimizer.cpp
+++ b/src/jit/optimizer.cpp
@@ -6190,7 +6190,7 @@ bool Compiler::optHoistLoopExprsForTree(
             {
                 // TODO-CQ: test that fails if we hoist GT_CLS_VAR: JIT\Directed\Languages\ComponentPascal\pi_r.exe
                 // method Main
-                treeIsHoistable =  false;
+                treeIsHoistable = false;
             }
         }
 
--- a/tests/src/JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b79250/b79250.il
+++ b/tests/src/JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b79250/b79250.il
@@ -89,7 +89,7 @@ Start_Orphan_8:
          stind.i4
 End_Orphan_8:
            ldloc LOCAL_0x0
-          conv.ovf.i1.un
+          conv.i1
 Start_Orphan_9:
           nop
 End_Orphan_9:


hseok-oh:
### Before
```
IN0029: 000092      movw    r3, 0xf48
IN002a: 000096      movt    r3, 0x7160
IN002b: 00009A      ldr     r3, [r3]
IN002c: 00009C      adds    r5, r3, 4
 
......

IN0037: 0000C0      blx     r3          // CORINFO_HELP_GETSHARED_NONGCSTATIC_BASE
IN0038: 0000C2      mov     r3, r5
IN0039: 0000C4      ldr     r2, [r3]
```
Null reference exception at IN0039

### After
```
IN0033: 0000B4      blx     r3          // CORINFO_HELP_GETSHARED_NONGCSTATIC_BASE
IN0034: 0000B6      movw    r3, 0xf44
IN0035: 0000BA      movt    r3, 0x7160
IN0036: 0000BE      ldr     r3, [r3]
IN0037: 0000C0      adds    r3, 4
IN0038: 0000C2      ldr     r2, [r3]
```

BruceForstall:
Should this be an arm32-only change?

@JosephTremoulet PTAL
cc @dotnet/jit-contrib

JosephTremoulet:
I was about to suggest combining the new code with the code just above that's doing a partial workaround, but then I saw that that's what your first commit looked like.  Why did you need to change it?  The commit message "Bug fix" doesn't give me much to go on...

>Should this be an arm32-only change?

I don't think so.  The issue is that we aren't modeling data dependence between helper calls that trigger static initialization and loads of static fields; there's already a workaround that blocks hoisting loads of statics, which this change extends to more generally block trees that contain loads of statics.  So, assuming we can get a tree that contains a load of a static (which only gets initialized in that loop) on any platform, then this shouldn't be arm32-only, and should add a regression test.

BruceForstall:
Perhaps one of us should run desktop asm diffs and/or non-arm asm diffs.

briansull:
This looks like a correctness issue that we need to port into CoreCLR 2.0 Release and the 4.7.1 Desktop Release.

briansull:
Adding @erozenfeld

src/jit/optimizer.cpp
```diff
@@ -6203,6 +6203,12 @@ bool Compiler::optHoistLoopExprsForTree(
         {
             treeIsHoistable = false;
         }
+
+        // Avoid to hoist reference loading
+        if (tree->OperGet() == GT_CLS_VAR && tree->gtType == TYP_BYREF)
+        {
+            treeIsHoistable = treeIsInvariant = false;
+        }
```
briansull: Just now taking a look this code:

Due to the code block above (line 6189-6194) it would be impossible for treeIsHoistable top be true when `(tree->OperGet() == GT_CLS_VAR)`

What did you observe here?   Is this possibly a native compiler bug?

briansull: Also our coding convention do not permit multiple assignments in one statement:

Do NOT:  ` treeIsHoistable = treeIsInvariant = false;`

Instead:
```
 treeIsHoistable = false;
 treeIsInvariant = false;
```

hseok-oh:
@briansull @JosephTremoulet We could see GenTree for static field load after morphing
in x64:
```
 |     /--*  const     long   8 Fseq[#FirstElem] $142
 |  /--*  +         byref  <l:$500, c:$501>
 |  |  \--*  indir     ref    <l:$2c1, c:$385>
 |  |     \--*  const(h)  long   0x7f3398000e88 static Fseq[Package] $202
 \--*  =         byref  <l:$500, c:$501>
     \--*  lclVar    byref  V08 tmp6         d:3 <l:$500, c:$501>
```

in ARM:
```
 |     /--*  const     int    4 Fseq[#FirstElem] $45
 |  /--*  +         byref  $401
 |  |  \--*  clsVar    ref    Hnd=0x74d823e4 Fseq[Package] $400
 \--*  =         byref  $401
    \--*  lclVar    byref  V09 tmp7
```

In 64bit architecture, it can use const and indirect to get static field : https://github.com/dotnet/coreclr/blob/master/src/jit/morph.cpp#L6607-L6627
In 32bit architecture, it uses reference type clsVar to get static field: https://github.com/dotnet/coreclr/blob/master/src/jit/morph.cpp#L6645-L6658

CLS_VAR itself is not hoisted. But we can see hoisting when JIT tries to hoist byref type OP_ADD with children in ARM architecture. In hoisting operation, we can see log like this,
```
Hoisting a copy of [000110] into PreHeader for loop L00 <BB03..BB03>:
N002 (  1,  1) [000112] -------------             /--*  const     int    4 Fseq[#FirstElem] $45
N003 (  8, 12) [000110] ----G--------             *  +         byref  $401
N001 (  6, 10) [000111] ----G--------             \--*  clsVar    ref    Hnd=0x74d823e4 Fseq[Package] $400

This hoisted copy placed in PreHeader (BB10):
               [000155] -------------             /--*  nop       void
               [000156] ----G--------             *  comma     void
     (  1,  1) [000154] -------H-----             |  /--*  const     int    4 Fseq[#FirstElem] $45
     (  8, 12) [000152] ----G--H-----             \--*  +         byref  $401
     (  6, 10) [000153] ----G--H-----                \--*  clsVar    ref    Hnd=0x74d823e4 Fseq[Package] $400
```
To avoid hoisting optimization for parents of CLS_VAR, I assigned `treeIsInvariant = false`. And I used type check because we need to avoid hoisting optimization for loading operation from memory, not assigning constant value.

src/jit/optimizer.cpp
```diff
@@ -6203,6 +6203,12 @@ bool Compiler::optHoistLoopExprsForTree(
         {
             treeIsHoistable = false;
         }
+
+        // Avoid to hoist reference loading
+        if (tree->OperGet() == GT_CLS_VAR && tree->gtType == TYP_REF)
+        {
+            treeIsInvariant = false;
```
briansull: I would add a comment 

// It is unsafe to make this as invariant here as we currently don't prevent it from being hoisted above the call to the class initializer.

JosephTremoulet:
@hseok-oh, can you elaborate on "I used type check because we need to avoid hoisting optimization for loading operation from memory, not assigning constant value"?  If we e.g. had a load of an int-typed static field which we added some loop-invariant value to, wouldn't we be at risk of hoisting the add?  The CLS_VAR node would have type TYP_INT in that case, no?

briansull:
I think that Joseph is correct we should not allow invariant code motion regardless of the type.

There is an additional check that we could consider doing here as well, but it requires asking the VM
if the type use in the GT_CLS_VAR is a 'beforeFieldInit' type, so that we know if a call to the class constructor is necessary.

hseok-oh:
I agree with @JosephTremoulet. Then it returns to the first commit, and remains failure at `JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b79250/b79250/b79250.sh`. In fact, if we try to run this test in ARM32/Linux, we can get same failure message.

hseok-oh:
`JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b79250/b79250/b79250.sh` failed with overflow exception. 
This exception occur in 1st loop (label between Loop_0x0 and Loop_0x1 https://github.com/dotnet/coreclr/blob/master/tests/src/JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b79250/b79250.il#L55-L120), exactly, `conv.ovf.i1.un` instruction at https://github.com/dotnet/coreclr/blob/master/tests/src/JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b79250/b79250.il#L92

This instruction convert the value in top of stack to int8, and this value comes from the first local variable `LOCAL_0x0`. This local variable is set once by the 1st and 2nd instruction of Method_0x5a7bd7a1 (`ldc.i4 0x75811f64` and `stloc LOCAL_0x0`). And this variable is never modified until loop exit. 

It seems that failure with overflow exception is a correct result, because the value of LOCAL_0x0 (0x75811f64) is larger than int8.

src/jit/optimizer.cpp
```diff
@@ -6203,6 +6203,13 @@ bool Compiler::optHoistLoopExprsForTree(
         {
             treeIsHoistable = false;
         }
+
+        if (tree->OperGet() == GT_CLS_VAR)
```
JosephTremoulet: I'd suggest moving this up a few lines and sinking the call to `optVNIsLoopInvariant` into an `else` path, to avoid that call when we're just going to discard the result.  Then you should be able to simply remove the partial workaround above, since the check on 6204 will set `treeIsHoistable` to false if you're setting `treeIsInvariant` to false before it.

tests/src/JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b79250/b79250.il
```diff
@@ -89,7 +89,7 @@ Start_Orphan_8:
          stind.i4
 End_Orphan_8:
            ldloc LOCAL_0x0
-          conv.ovf.i1.un
+          conv.i1
```
JosephTremoulet: I followed your reasoning on why it's correct for this to throw an `OverflowException`, but I don't understand why your change exposed this.  Can you explain what was the diff in the generated code and why we weren't hitting the overflow exception before?

src/jit/optimizer.cpp
hseok-oh: Fixed it. Thanks.

tests/src/JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b79250/b79250.il
hseok-oh: `conv.i1` does not check overflow. By fixing this, we can avoid this arithmetic overflow exception and check expected exception: DivideByZero (https://github.com/dotnet/coreclr/blob/master/tests/src/JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b79250/b79250.il#L1552-L1556). 

I'll investigate why we weren't hitting the arithmetic overflow exception before.

hseok-oh:
I'm not using `jit-diff`. I'm testing in ARM32/Linux device and x64/Linux PC. The test `b79250` is used in Windows only. You can see this in `tests/testsUnsupportedOutsideWindows.txt`. I don't know it was run correctly in ARM32/Windows.

JosephTremoulet:
@hseok-oh Ok I'll look into why `b79250` regresses on windows with this change.  Also, I'm not sure I understand about `jit-diff` -- I meant to suggest running it to get a sense of how frequently this change leads to codegen diffs, so running it on an x64/Linux PC would be fine.  Is that not feasible?

hseok-oh:
@JosephTremoulet Sure. No problem to use Linux PC. 
BTW, I'll leave a vacation and I'll be back on May 9.

JosephTremoulet:
The failure in `b79250` is due to the fix here exposing #7147 in that test -- some of the first operations in the loop body are to divide a static by zero, and then to perform an overflowing convert of a constant.  The main method of the test makes sure that the divide by zero exception is what gets thrown.  When we hoist the divide and the convert, the hoisted divide happens first and we get the expected result.  When we block division of statics and hoist just the convert, the conversion happens first and we get the wrong exception type.

JosephTremoulet:
On the one hand, incorrectly hoisting static loads before their initialization seems "worse" than throwing the wrong kind of exception, so it's tempting to say we could disable `b79250` pending fixing #7147.  On the other hand, `b79250` was added to `test/JIT/regressions` for a reason, I would hate to re-regress it, especially while trying to stabilize 2.0.

JosephTremoulet:
jit-diff (win x64) reported no diffs, but IIRC that's because `CLS_VAR` doesn't get used in crossgen; I'll port to desktop and see what SPMI says...

JosephTremoulet:
Only three SPMI diffs.

JosephTremoulet:
I'm finding it difficult to write a regression test -- we import a static field load as a comma of the class init helper call and the CLS_VAR for the static.  If the class has beforefieldinit semantics, we mark the call hoistable, its only arguments are two constants, and (typically) we hoist it along with the static field access if we're going to hoist the field access.  If the class does not have beforefieldinit semantics, then we mark the loop as having "heap havoc" and the static field isn't considered invariant.  I'm trying to figure out what was happening in #10780 that broke that pattern and led us to hoist the field access but not the helper call...

JosephTremoulet:
What happens in #10780 is that the class has beforefieldinit semantics, so we mark the init helper call as hoistable, but then we don't hoist it because it can throw an exception and there were prior global side effects in the loop -- the fear is that hoisting it would suppress those global side-effects that should be visible in the handler.  The static field load can't throw an exception, so it still gets hoisted.  This doesn't reproduce on x86 because the nodes have different costs, and so on x86 we end up considering the static field load (or more precisely the add that it feeds into) unprofitable to hoist.

JosephTremoulet:
Note that the importer [generates a comma](https://github.com/dotnet/coreclr/blob/980c1204d68f54be77eb840cc3f2e4fe2df42a26/src/jit/importer.cpp#L13352) wrapping the init call and field reference.  If the entire comma is hoisted, the codegen is correct, since the init call gets hoisted before the field reference.  If the comma is not hoisted, and the field reference is a direct child of the comma, then the [existing workaround](https://github.com/dotnet/coreclr/blob/980c1204d68f54be77eb840cc3f2e4fe2df42a26/src/jit/optimizer.cpp#L6192-L6199) is sufficient to avoid bad codegen.  So the bad codegen only manifests if there is a hoistable intermediate expression that contains the field reference and feeds the comma.  In the case of #10780, that expression comes from [this bit of code](https://github.com/dotnet/coreclr/blob/980c1204d68f54be77eb840cc3f2e4fe2df42a26/src/jit/importer.cpp#L6062-L6067), which is triggered by `CORINFO_FLG_FIELD_STATIC_IN_HEAP` being set on the static field (which apparently means it is allocated as a boxed struct in the GC heap).  The comments indicate that the triggering conditions for that are simply that the static is a struct, but if I try using a struct when trying to generate a repro I suddenly see `CORINFO_ACCESS_ADDRESS` set for the field, which causes us to take the other path in [this code](https://github.com/dotnet/coreclr/blob/980c1204d68f54be77eb840cc3f2e4fe2df42a26/src/jit/importer.cpp#L6055) and represent the field in our IR by indir-of-immediate rather than `CLS_VAR`, which in turn avoids the bad codegen because of how conservatively we treat indir-of-immediate aliasing.

So... as far as determining how broad the impact of the bug is, I can say that it needs to be a static to which `CORINFO_FLG_FIELD_STATIC_IN_HEAP` applies but `CORINFO_ACCESS_ADDRESS` does not, but I don't know how rare that combination is.

JosephTremoulet:
Given everything above, I think we could fix #10780 without regressing `b79250` (or more generally exacerbating #7147) by targeting the change so that rather than blocking hoisting of entire `CLS_VAR` trees, we block just hoisting trees which contain a `CLS_VAR` whose `gtClsVarHnd`'s fieldInfo has `CORINFO_FLG_FIELD_INITCLASS` set, and which do not contain the entire comma with the init helper.  i.e., in loop hoisting,

  1. Communicate to callers whether a node is a function of a need-init cls_var
  2. At comma nodes, if op2 is a function of a need-init cls_var but op1 is a class init helper call, don't continue propagating "is a function of need-init cls_var" upward.
  3. Don't hoist any node that is a function of a need-init cls_var (but do still consider them invariant if applicable)
  4. At (init helper, cls_var) comma nodes, don't hoist the cls_var if not hoisting the init helper

JosephTremoulet:
Regarding how widely the bug applies:

I did manage to create a repro that hits on Win/x86:

```c#

namespace N
{
    public struct Constant
    {
        public int Value;

        public static Constant Twenty = new Constant() { Value = 20 };
    }

    static class C
    {
        static int Sum;
        static int Zero;

        static int Add(Constant constant)
        {
            return Sum += constant.Value;
        }


        static void Twentify(int n)
        {
            for(int i = 0; i < n; ++i)
            {
                Zero = 0;
                Add(Constant.Twenty);  // hoisting Constant.Twenty w/o hoisting initializer call => bug
            }
        }

        {
            Sum = 0;
            Twentify(5);
            return Sum + Zero;
        }
    }
}
```

I'm not sure I can get one on x64 since we prefer imgrel32 relocs for statics, which the jit models differently internally and never hoists (pending #6900).
Even for x86, triggering conditions are somewhat demanding:
 - Needs to be a struct-typed static to get `CORINFO_FLG_FIELD_STATIC_IN_HEAP` and the addition under the comma
 - Needs to be a use of the struct value, and not its address, to avoid taking the `CORINFO_ACCESS_ADDRESS` path (which leads to conservative hoisting, like imgrel32).  This pretty much means it needs to be a by-value argument to a call (in #10780, it was an argument to op_Equality), and in turn the call needs to be inlined so we can still recognize the load as invariant
 - The static needs to be one that requires an initializer and has `BeforeFieldInit` semantics (which, in C#, I believe translates to having nontrivial static fields initialized inline at their declaration)
 - We need to consider the static access profitable to hoist.  That comparison involves the number of registers on the target machine and the number of `lclVar`s live across the loop, which is why the repro from #10780 hit on arm32 but not x86 (which has fewer registers)
 - We need to *not* hoist the initializer itself, which may mean that the hoistable reference (which must be in the loop's first block to be hoistable) must come after some statement which may mutate the heap (but in a way that we're able to disambiguate against the field itself)
 - Dynamically, the statics for the given class must not be initialized yet on entry to the loop

JosephTremoulet:
Since we're interested in fixing this for 2.0 and that window is about to close, and @hseok-oh is away today, I'm going to take a stab at putting together a fix along the lines outlined [above](https://github.com/dotnet/coreclr/pull/11198#issuecomment-299870010).

briansull:
I would be good to get this fixed for 2.0
Can you include me on the CR for the fix.

JosephTremoulet:
I've put up the PR with the proposed fix at #11465.  The only tricky bit was that the importer was discarding the knowledge about which fields had `CORINFO_FLG_FIELD_INITCLASS` set, so in the PR I've added GTF_ flags to propagate that information.

hseok-oh:
Fixed by #11465

PR #11203, opened 2017-04-25
[RyuJIT/ARM32] Update lclFld codegen

mskvortsov:
Fixes runtime failures of
```
JIT/jit64/valuetypes/nullable/box-unbox/box-unbox/box-unbox011
JIT/jit64/valuetypes/nullable/box-unbox/generics/box-unbox-generics011
JIT/jit64/valuetypes/nullable/box-unbox/interface/box-unbox-interface016
JIT/jit64/valuetypes/nullable/box-unbox/value/box-unbox-value011
...
```
cc @dotnet/arm32-contrib

### DIFF
--- a/src/jit/codegenarm.cpp
+++ b/src/jit/codegenarm.cpp
@@ -491,7 +491,21 @@ void CodeGen::genCodeForTreeNode(GenTreePtr treeNode)
             unsigned varNum = treeNode->gtLclVarCommon.gtLclNum;
             assert(varNum < compiler->lvaCount);
 
-            emit->emitIns_R_S(ins_Move_Extend(targetType, treeNode->InReg()), size, targetReg, varNum, offs);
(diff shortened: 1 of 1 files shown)


mskvortsov:
@BruceForstall The fix copies Arm64 codegen, as usual. I'm not sure that introducing something like `genCodeForLclFld()` would be hugely profitable here; am I wrong?

BruceForstall:
>  I'm not sure that introducing something like genCodeForLclFld() would be hugely profitable here; am I wrong?

For now, this is fine. I'd like to do a more general cleanup / refactoring of genCodeForTreeNode() (and x86/x64 as well). If I get to it, that might allow more arm32/arm64 sharing.

BruceForstall:
cc @dotnet/jit-contrib

PR #11204, opened 2017-04-25
Fix GCInfo library used by ARM/ARM64 protononjit builds

BruceForstall:
Otherwise, using the normal GCInfo library, the altjits will
assert creating GC info.

### DIFF
--- a/src/gcinfo/CMakeLists.txt
+++ b/src/gcinfo/CMakeLists.txt
@@ -23,5 +23,15 @@ endif(CLR_CMAKE_PLATFORM_UNIX)
 add_subdirectory(lib)
 add_subdirectory(crossgen)
 
+if (CLR_CMAKE_PLATFORM_ARCH_I386)
+    # On x86, build RyuJIT/ARM32 cross-compiling altjit.
+    add_subdirectory(gcinfo_arm)
+endif ()
+
+if (CLR_CMAKE_PLATFORM_ARCH_AMD64)
+    # On amd64, build RyuJIT/ARM64 cross-compiling altjit.
+    add_subdirectory(gcinfo_arm64)
+endif ()
+
 _install (FILES gcinfoencoder.cpp
                DESTINATION gcinfo)
--- a/src/gcinfo/gcinfo_arm/CMakeLists.txt
+++ b/src/gcinfo/gcinfo_arm/CMakeLists.txt
@@ -1,0 +1,7 @@
+remove_definitions(-D_TARGET_X86_=1)
+add_definitions(-D_TARGET_ARM_)
+
+add_library_clr(gcinfo_arm
+    STATIC
+    ${GCINFO_SOURCES}
+)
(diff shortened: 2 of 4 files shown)


BruceForstall:
@briansull @CarolEidt PTAL
cc @dotnet/jit-contrib

src/gcinfo/CMakeLists.txt
```diff
@@ -23,5 +23,15 @@ endif(CLR_CMAKE_PLATFORM_UNIX)
...
+if (CLR_CMAKE_PLATFORM_ARCH_I386)
+    # On x86, build RyuJIT/ARM32 cross-compiling altjit.
+    add_subdirectory(arm_altjit)
+endif ()
+
+if (CLR_CMAKE_PLATFORM_ARCH_AMD64)
+    # On amd64, build RyuJIT/ARM64 cross-compiling altjit.
+    add_subdirectory(arm64_altjit)
+endif ()
+
```
briansull: Does this work for Desktop builds, or is it just for CoreCLR builds?

BruceForstall: We already have this logic in the desktop build system; this change is to add it to the cmake-based CoreCLR build system.

src/gcinfo/CMakeLists.txt
```diff
@@ -23,5 +23,15 @@ endif(CLR_CMAKE_PLATFORM_UNIX)
 add_subdirectory(lib)
 add_subdirectory(crossgen)
 
+if (CLR_CMAKE_PLATFORM_ARCH_I386)
+    # On x86, build RyuJIT/ARM32 cross-compiling altjit.
+    add_subdirectory(arm_altjit)
```
pgavlin: I don't love that we have `altjit` in the paths here, but I don't have a great alternative. `arm{,64}_cross_gcinfo`, maybe?

pgavlin: Or just match the directories to the names of the libraries (e.g. `add_subdirectory(gcinfo_arm64)`)

BruceForstall: Why don't you like "altjit" in the path?

pgavlin: Because these are builds of the `gcinfo` library, and the `gcinfo` library is logically distinct from the JIT.

BruceForstall: But these builds are only used by altjit.

BruceForstall: (and gcinfo is only used by JIT, for that matter)

pgavlin: Regardless, this is a separate component and should be named as such. It is confusing to navigate to a directory named "gcinfo" and find subdirectories referencing the JIT. For example, when I first looked at the PR, I thought that you had actually placed the directories underneath the JIT's directory (e.g. `src/jit/arm{,64}_altjit`).

BruceForstall:
Updated

PR #11205, opened 2017-04-25
[RyuJIT/ARM32] Workaround vstr/vldr addressing modes limitations

mskvortsov:
Here is a sample context from `JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b73786`, `ILGEN_0x4a792db4:Method_0x2a166130`:
```
N067 (  1,  1) [000062] -------------       t62 =    lclVar    ref    V10 loc3         u:2 r4 (last use) REG r4 RV $380
                                                  ┌──▌  t62    ref    
N069 (???,???) [000118] -------------      t118 = ▌  lea(b+224824) byref  REG NA
...
                                                  ┌──▌  t31    int    
N075 ( 13, 18) [000032] ---X---------       t32 = ▌  cast      float <- int REG f8 $4c1
                                                  ┌──▌  t118   byref  
                                                  ├──▌  t32    float  
N077 (???,???) [000086] -A-XGO----S--             ▌  storeIndir float  REG NA
```
The constant is too large to fit into `imm8` of the `vstr` being generated for the `storeIndir`, and ARMv7 floating point store doesn't support a form similar to integer `str Rdst, [Rbase + Roffset]`, as described in the reference manual. The same it true for a floating point loads.

cc @dotnet/arm32-contrib


BruceForstall:
cc @dotnet/jit-contrib

mskvortsov:
Or should those `lea` nodes be simply made not *contained*?

BruceForstall:
> Or should those lea nodes be simply made not contained?

That sounds like the right solution.

pgavlin:
> Or should those `lea` nodes be simply made not contained?

What code do we generate for a non-contained `lea` on ARM32?

mskvortsov:
The arm32 `genLeaInstruction()` will produce something like
```
movw    Rimm, #lo16
movt    Rimm, #hi16
add     Raddr, Rbase, Rimm
```
(Currently, however, it will not. `emitIns_R_AR(INS_lea, ...)` used by it will assert on the same inability to encode a large constant.)

And it turns out, that generating those three instructions for a single `GT_LEA` is considered as not a perfectly right way to do things according to the words from the arm64 version of `genLeaInstruction()`:
```c++
    // TODO-ARM64-CQ: The purpose of the GT_LEA node is to directly reflect a single target architecture
    //             addressing mode instruction.  Currently we're 'cheating' by producing one or more
    //             instructions to generate the addressing mode so we need to modify lowering to
    //             produce LEAs that are a 1:1 relationship to the ARM64 architecture.
```
So, ideally, all these decisions whether a constant fits into some instruction form, and questions of particular addressing mode availability should be resolved at the lowering phase.

I'm not sure which way to choose:

1. update `genLeaInstruction()` in the vein of arm64 implementation, and make `lea` with a large offset non-contained for FP loads/stores,
2. continue modifying `emitInsLoadStoreOp()` which already makes some decisions about `imm` fitting,
3. postpone this work till I have enough test cases, as currently I have only one.

mikedn:
Shouldn't `CodeGen::genCreateAddrMode` be copied to codegenxarch.cpp and codegenarm.cpp and modified according to each architecture's address mode capabilities? Address mode is inherently arch specific and having a common version doesn't make a lot of sense.

CarolEidt:
> Shouldn't CodeGen::genCreateAddrMode be copied to codegenxarch.cpp and codegenarm.cpp and modified according to each architecture's address mode capabilities? Address mode is inherently arch specific and having a common version doesn't make a lot of sense.

Absolutely. Also, this method has evolved (as indicated by its name) from a method that actually generated code. I would love to see this cleaned up, and it should definitely be architecture-specific.  We have also discussed moving the creation of addressing modes earlier in the phase order, so that this method doesn't have to be called multiple times to determine whether a node is likely to participate in an addressing mode. However, that has the potential to do both harm and good.

BruceForstall:
> Shouldn't CodeGen::genCreateAddrMode be copied to codegenxarch.cpp and codegenarm.cpp and modified according to each architecture's address mode capabilities? Address mode is inherently arch specific and having a common version doesn't make a lot of sense.

There is a lot of commonality between architectures, although the details vary. It is already `#ifdef'ed` for arm64, Having `LEGACY_BACKEND` still around complicates it more. One things that's quite confusing is that `SCALED_ADDR_MODES` and `LEA_AVAILABLE` are defined for all platforms. The code is relatively complex, and we wouldn't want to duplicate it, IMO. However, it could use some cleaning up, probably, to make it work more generically for all platforms.

BruceForstall:
> update genLeaInstruction() in the vein of arm64 implementation, and make lea with a large offset non-contained for FP loads/stores,

I think you should follow the arm64 model. Eventually, I'd like to see us get rid of INS_lea entirely from the ARM/ARM64 code path, but that's hard as long as `LEGACY_BACKEND` still exists, which shares the ARM encoder.

BruceForstall:
@mskvortsov Are you planning to get back to this?

mskvortsov:
@BruceForstall The plan is to update this after most currently pending PRs related to altjit/arm32 get landed.

mskvortsov:
@dotnet/arm64-contrib This PR needs an arm64 test.

BruceForstall:
@dotnet-bot  test @dotnet-bot test Windows_NT arm64 Checked

mskvortsov:

hqueue:
This also fixes #11958.

src/jit/emitarm.cpp
```diff
@@ -1436,6 +1436,18 @@ emitter::insSize emitter::emitInsSize(insFormat insFmt)
 
 /*****************************************************************************
  *
+ *  emitIns_valid_imm_for_vldst_offset() returns true when the immediate 'imm'
+ *   can be encoded as the offset in a vldr/vstr instruction.
+ */
```
briansull: You could also add 
`// Returns true when 'imm' is a non-negative multiple of 4 that is less than 1024`

src/jit/codegencommon.cpp
```diff
@@ -2170,7 +2170,7 @@ bool CodeGen::genCreateAddrMode(GenTreePtr  addr,
         noway_assert(op2);
         switch (op2->gtOper)
         {
-#ifndef _TARGET_ARM64_
+#if !defined(_TARGET_ARM64_) && !(defined(_TARGET_ARM_) && !defined(LEGACY_BACKEND))
             // TODO-ARM64-CQ: For now we don't try to create a scaled index on ARM64.
```
briansull: or for ARM32

src/jit/codegencommon.cpp
```diff
@@ -2079,7 +2079,7 @@ bool CodeGen::genCreateAddrMode(GenTreePtr  addr,
 
         switch (op1->gtOper)
         {
-#ifndef _TARGET_ARM64_
+#if !defined(_TARGET_ARM64_) && !(defined(_TARGET_ARM_) && !defined(LEGACY_BACKEND))
             // TODO-ARM64-CQ: For now we don't try to create a scaled index on ARM64.
```
briansull: or for ARM32

src/jit/codegencommon.cpp
```diff
@@ -2017,7 +2017,7 @@ bool CodeGen::genCreateAddrMode(GenTreePtr  addr,
 
                     goto AGAIN;
 
-#if SCALED_ADDR_MODES && !defined(_TARGET_ARM64_)
+#if SCALED_ADDR_MODES && !defined(_TARGET_ARM64_) && !(defined(_TARGET_ARM_) && !defined(LEGACY_BACKEND))
                 // TODO-ARM64-CQ: For now we don't try to create a scaled index on ARM64.
```
briansull: or for ARM32

src/jit/lsraarm.cpp
```diff
@@ -567,49 +567,54 @@ void Lowering::TreeNodeInfoInit(GenTree* tree)
...
-            if ((index != nullptr) && (cns != 0))
+            // An internal register may be needed too; the logic here should be in sync with the
+            // genLeaInstruction()'s requirements for a such register.
+            if (lea->HasBase() && lea->HasIndex())
             {
-                // ARM does not support both Index and offset so we need an internal register
-                info->internalIntCount = 1;
+                if (offset != 0)
+                {
+                    info->internalIntCount = 1;
```
briansull: // We need a register when we have all three: base reg, index reg and a non-zero offset

src/jit/lsraarm.cpp
```diff
@@ -567,49 +567,54 @@ void Lowering::TreeNodeInfoInit(GenTree* tree)
...
+                }
             }
-            else if (!emitter::emitIns_valid_imm_for_add(cns, INS_FLAGS_DONT_CARE))
+            else if (lea->HasBase())
             {
-                // This offset can't be contained in the add instruction, so we need an internal register
-                info->internalIntCount = 1;
+                if (!emitter::emitIns_valid_imm_for_add(offset, INS_FLAGS_DONT_CARE))
+                {
+                    info->internalIntCount = 1;
```
briansull: `// We need a register when we have an offset that is too large to encode in the add instruction`

src/jit/lsraarmarch.cpp
```diff
@@ -190,10 +191,36 @@ void Lowering::TreeNodeInfoInitIndir(GenTreePtr indirTree)
...
         cns                  = lea->gtOffset;
 
-        m_lsra->clearOperandCounts(addr);
-        // The srcCount is decremented because addr is now "contained",
-        // then we account for the base and index below, if they are non-null.
-        info->srcCount--;
+#ifdef _TARGET_ARM_
+        // ARM floating-point load/store doesn't support a form similar to integer
+        // ldr Rdst, [Rbase + Roffset] with offset in a register. The only supported
+        // form is vldr Rdst, [Rbase + imm] with a certain constraint on the imm.
```
briansull: `with a more limited constraint on the imm.`

src/jit/codegenarmarch.cpp
```diff
@@ -2637,6 +2637,162 @@ void CodeGen::genCodeForStoreBlk(GenTreeBlk* blkOp)
...
+        BitScanForward(&lsl, lea->gtScale);
+
+        assert(lsl <= 4);
+
+        if (offset != 0)
+        {
+            regNumber tmpReg = lea->GetSingleTempReg();
+
+#if defined(_TARGET_ARM_)
+            if (emitter::emitIns_valid_imm_for_add(offset, INS_FLAGS_DONT_CARE))
```
CarolEidt: It would be nice to avoid the repeated `#ifdef` for this. Could you add a default value for the second parameter (`INS_FLAGS_DONT_CARE` for arm and `EA_8BYTE` for arm64)? Those seem to be the most common values, here and elsewhere, and then you could omit the `#ifdef`s.

src/jit/codegenarmarch.cpp
```diff
@@ -2637,6 +2637,162 @@ void CodeGen::genCodeForStoreBlk(GenTreeBlk* blkOp)
...
+#elif defined(_TARGET_ARM64_)
+            if (emitter::emitIns_valid_imm_for_add(offset, EA_8BYTE))
+#endif // _TARGET_*
+            {
+                if (lsl > 0)
+                {
+#if defined(_TARGET_ARM_)
+                    // Generate code to set tmpReg = base + index*scale
+                    emit->emitIns_R_R_R_I(INS_add, size, tmpReg, memBase->gtRegNum, index->gtRegNum, lsl,
+                                          INS_FLAGS_DONT_CARE, INS_OPTS_LSL);
```
CarolEidt: Similarly, this is repeated 3 times with the same final two arguments. For this case, it might make sense to add a method such as `generateScaledAdd(size, targetReg, baseReg, indexReg, scale)`.

mskvortsov: @CarolEidt Thank you for the review. I've done the changes you and @briansull requested. Can this be merged now?

CarolEidt:
@briansull - could you have a look and merge if you're happy with the changes?

PR #11208, opened 2017-04-25
[Preview] Undo VM portion of #9789

### LINKED ISSUE #11088: [x86jit32] CscBench benchmark is broken.
We have not got results since April 5th, 2017. A private PR performance run to make running benchmarks more robust indicates that `corerun.exe` is abnormally terminating when running the benchmark.

Benchmark: tests\src\JIT\Performance\CodeQuality\Roslyn\CscBench.cs

RussKeldorph:
Maintain compat with JIT32 for 2.0 Preview.  #9789 introduced a performance optimization involving a change to the JIT and JIT interface.  Since this optimization was not added to JIT32, we should disable it for Preview, returning behavior to that prior to #9789 for Preview.  Since JIT32 is being removed post-Preview, this change does not need to go into master.

Fixes #11088.

### DIFF
--- a/src/vm/jitinterface.cpp
+++ b/src/vm/jitinterface.cpp
@@ -1439,11 +1439,6 @@ static CORINFO_FIELD_ACCESSOR getFieldIntrinsic(FieldDesc * field)
     {
         return CORINFO_FIELD_INTRINSIC_ZERO;
     }
-    else
-    if (MscorlibBinder::GetField(FIELD__BITCONVERTER__ISLITTLEENDIAN) == field)
-    {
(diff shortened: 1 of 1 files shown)


RussKeldorph:
Note that JIT32 is being removed post-preview, so this change is not being introduced in master.

RussKeldorph:
@dotnet-bot help

RussKeldorph:
Another option is to port #9789 to JIT32, but I prefer this option for Preview unless we know of a reason it wouldn't work.

/cc @dotnet/jit-contrib

pgavlin:
Seems okay to me.

PR #11210, opened 2017-04-25
Unsafe.As workaround (roundtrip via a local to avoid type mismatch)

ahsonkhan:
From https://github.com/dotnet/corefxlab/issues/1479#issuecomment-296325631

> The CoreLib internal version of Unsafe.As is missing the workaround for this that was done in the public OOB version of Unsafe.As.

Keeping it in sync with:
https://github.com/dotnet/corefx/blob/master/src/System.Runtime.CompilerServices.Unsafe/src/System.Runtime.CompilerServices.Unsafe.il#L268

cc @jkotas, @russellhadley

### DIFF
--- a/src/vm/jitinterface.cpp
+++ b/src/vm/jitinterface.cpp
@@ -6991,7 +6991,7 @@ bool getILIntrinsicImplementationForUnsafe(MethodDesc * ftn,
     else if (tk == MscorlibBinder::GetMethod(METHOD__UNSAFE__BYREF_AS)->GetMemberDef())
     {
         // Return the argument that was passed in.
-        static const BYTE ilcode[] = { CEE_LDARG_0, CEE_RET };
+        static const BYTE ilcode[] = { CEE_LDARG_0, CEE_STLOC_0, CEE_LDLOC_0, CEE_RET };
         methInfo->ILCode = const_cast<BYTE*>(ilcode);
         methInfo->ILCodeSize = sizeof(ilcode);
         methInfo->maxStack = 1;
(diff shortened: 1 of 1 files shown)


src/vm/jitinterface.cpp
```diff
@@ -6991,7 +6991,7 @@ bool getILIntrinsicImplementationForUnsafe(MethodDesc * ftn,
     else if (tk == MscorlibBinder::GetMethod(METHOD__UNSAFE__BYREF_AS)->GetMemberDef())
     {
         // Return the argument that was passed in.
-        static const BYTE ilcode[] = { CEE_LDARG_0, CEE_RET };
+        static const BYTE ilcode[] = { CEE_LDARG_0, CEE_STLOC_0, CEE_LDLOC_0, CEE_RET };
```
jkotas: Have you verified that it actually works?

ahsonkhan: I ran build at root of coreclr repo. How else should I verify that the change is correct?

jkotas: Right, this is what I would expect because of you are using undefined local.

ahsonkhan: Missing .locals (int32&). Yes.

ahsonkhan: How do I push a local variable on to the stack here?
I was looking for some useful instruction here: https://github.com/dotnet/coreclr/blob/master/src/jit/smopcodemap.def

jkotas: You would need to manufacture a signature for the local here. It can be done, but it is not a one liner. I think it is better to workaround it in C# in the Span implementation instead, like: https://github.com/dotnet/coreclr/pull/11212

ahsonkhan: > You would need to manufacture a signature for the local here. It can be done, but it is not a one liner. 

I want to learn how to do that (outside of this PR).

> think it is better to workaround it in C# in the Span implementation instead

I will close this PR then. Thanks

Issue #10918, opened 2017-04-12
Thread.Sleep behaves incorrectly on Linux if system time changes while thread sleeps


stager13:
Steps to reproduce:

Compile the following program:

```c#
using System;
using System.Threading;

namespace SleepTest
{
    class Program
    {
        static void Main(string[] args)
        {
            const int sleepTimeInS = 5;

            for (int i = 0; i < 100000; i++)
            {
                 Console.WriteLine($"Sleeping for {sleepTimeInS} seconds {i}..."

                 Thread.Sleep(sleepTimeInS * 1000);

                 Console.WriteLine("Ok");
            }
        }
    }
}
```

Run it and set system time to some time back (for example 10 minutes)
Thread.Sleep hangs instead of waking up each 5 seconds. I couldn't reproduce this problem on Windows, this code runs fine there. So it is platform-dependent. I also observed similar problem with AutoResetEvent.WaitOne(timeout) call under Linux.

 

OS version: Ubuntu 16.10 x86-64 (also reproducible on Ubuntu MATE on arm (Raspberry PI))
.NET core version: 1.0.1 (also reproducible on 2.0.0-preview1-005448)

janvorli:
I suspect we are using waits with absolute timeouts for synchronization primitives and use realtime clock source (don't set monotonic source explicitly). I remember I had to handle that when I was designing and implementing the event object for CoreRT PAL.

gkhanna79:
CC @kouvel 

@janvorli Should this be for 2.0?

janvorli:
@gkhanna79 it would be good to fix it for 2.0.0

gkhanna79:
Do you want @kouvel to help take a look?

janvorli:
@gkhanna79 I am working on a fix.

kouvel:
Thanks @janvorli, I'm afraid I won't be able to take a look at this any time soon.

janvorli:
Fixed by #11213

PR #11213, opened 2017-04-25
Fix thread wait to be resilient to real time clock changes on Linux

janvorli:
This change fixes a problem that happens when a thread uses a timed wait
and the real time clock change in the system (e.g. due to the user action,
time server sync, leap second, etc). In such case, the wait may end up waiting
for shorter or longer period of time.
It fixes the issue for Thread.Sleep and timed wait on a process local synchronization
primitives by using CLOCK_MONOTONIC instead of CLOCK_REALTIME for waiting on
cond vars.
Unfortunately, the waits on cross process synchronication primitives like
named mutexes always use CLOCK_REALTIME, so this fix works on process local primitives
only.


janvorli:
Fixes #10918

janvorli:

janvorli:

janvorli:

janvorli:

joepeer:
Shouldn't this be merged for at least the LTS version 1.0.5?
Without the fix we will get weird behavior when the clock gets adjusted for day light saving time.

janvorli:
@gkhanna79 I think that @joepeer is right that this change should be made to the LTS version too. Do you agree?

gkhanna79:
@janvorli Sounds reasonable. Please feel free to get PRs for 1.0.0 and 1.1.0 branches and email shiproom for them.

janvorli:
@joepeer I've ported the change to both 1.0.x and 1.1.x

joepeer:
Thanks.
Any idea when the new LTS that should include the fix will get released?

janvorli:
@gkhanna79 would probably be able to give you at least an approximate answer.

gkhanna79:
Actually, @leecow can give the timeframe of the next release for the 1.x servicing builds.

leecow:
We have been releasing 1.x updates roughly every 2 months. That said, a possible July release will be dependent upon the high impact issues submitted as candidates for the release.

PR #11215, opened 2017-04-25
Set RunCrossGen=true for non-Windows R2R runs

wtgodbe:
CC @gkhanna79 this is the change to get test Crossgen-ing on Non-Windows

### DIFF
--- a/tests/helixprep.proj
+++ b/tests/helixprep.proj
@@ -74,7 +74,7 @@
     Condition="'$(TargetsWindows)' == 'true' ">
 
     <PropertyGroup>
-      <CrossgenVar Condition="'$(Crossgen)' == 'true'">set RunCrossGen=true</CrossgenVar>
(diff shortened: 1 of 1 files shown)


tests/helixprep.proj
```diff
@@ -74,15 +74,15 @@
     Condition="'$(TargetsWindows)' == 'true' ">
 
     <PropertyGroup>
-      <CrossgenVar Condition="'$(Crossgen)' == 'true'">set RunCrossGen=true</CrossgenVar>
+      <CmdCrossgenVar Condition="'$(Crossgen)' == 'true'">set RunCrossGen=true</CrossgenVar>
```
gkhanna79: Why do we need to change the name of the variable?

wtgodbe: @gkhanna79 we don't necessarily, I just thought it'd be better to specify that it was only going to be used in the CMD file.

wtgodbe:
@dotnet-bot test this please

PR #11220, opened 2017-04-26
Fix timer callback time drift

### LINKED ISSUE #6408: Possible timer drift
Because `timerInfo->FiringTime` is set to `currentTime + timerInfo->Period` instead of `timerInfo->FiringTime + timerInfo->Period`, drift could happen when `currentTime` and `timerInfo->FiringTime` are different. What could cause them to be different? The [thread sleep](https://github.com/dotnet/coreclr/blob/726d1a3244b80bf963fd0d51e57d4bb90af1e426/src/vm/win32threadpool.cpp#L5011) might not get back exactly at the requested `nextFiringInterval`.

This means that `FireTimers` runs when one timer is expired, but potentially a little later. It runs at time `timerInfo->FiringTime + sleepDelta` where `sleepDelta` is the amount of time later than planned that it is scheduled and runs due to the previous thread sleep. At time `timerInfo->FiringTime + sleepDelta`, `currentTime = GetTickCount()` is taken. That means `currentTime = timerInfo->FiringTime + sleepDelta`. Then **each** time the timer does fire, `timerInfo->FiringTime = currentTime+timerInfo->Period`; substituting currentTime, `timerInfo->FiringTime = timerInfo->FiringTime + sleepDelta + timerInfo->Period`.

It would be nice if the timer itself was rock-solid. I'm curious why the team didn't go with `timerInfo->FiringTime += timerInfo->Period`.

Here it seems that `SleepEx` is expected to be within 150 ms or 10% for longer intervals: [test1.c](https://github.com/dotnet/coreclr/blob/726d1a3244b80bf963fd0d51e57d4bb90af1e426/src/pal/tests/palsuite/threading/SleepEx/test1/test1.c#L70) That doesn't make me feel better about the delay between the timer's planned expiry time and executing `currentTime = GetTickCount()`. I don't want my timers to be expected to drift up to 150 ms or 10% of the interval on every tick.

Thanks!

kouvel:
Fixes #6408:
- The change to S.T.Timer is the actual fix, but fixed the timer firing on the thread pool timer thread similarly as well

### DIFF
--- a/src/mscorlib/src/System/Threading/Timer.cs
+++ b/src/mscorlib/src/System/Threading/Timer.cs
@@ -265,7 +265,19 @@ namespace System.Threading
                             if (timer.m_period != Timeout.UnsignedInfinite)
                             {
                                 timer.m_startTicks = nowTicks;
-                                timer.m_dueTime = timer.m_period;
+                                uint elapsedForNextDueTime = elapsed - timer.m_dueTime;
+                                if (elapsedForNextDueTime < timer.m_period)
+                                {
+                                    // Discount the extra amount of time that has elapsed since the previous firing time to
+                                    // prevent timer ticks from drifting
+                                    timer.m_dueTime = timer.m_period - elapsedForNextDueTime;
+                                }
+                                else
+                                {
+                                    // Enough time has elapsed to fire the timer yet again. The timer is not able to keep up
+                                    // with the short period, have it fire 1 ms from now to avoid spinning without a delay.
+                                    timer.m_dueTime = 1;
+                                }
 
                                 //
                                 // This is a repeating timer; schedule it to run again.
--- a/src/vm/win32threadpool.cpp
+++ b/src/vm/win32threadpool.cpp
@@ -4755,15 +4755,30 @@ DWORD ThreadpoolMgr::FireTimers()
                                   timerInfo,
                                   QUEUE_ONLY /* TimerInfo take care of deleting*/);
 
-                timerInfo->FiringTime = currentTime+timerInfo->Period;
+                if (timerInfo->Period != 0 && timerInfo->Period != (ULONG)-1)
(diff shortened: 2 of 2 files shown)


stephentoub:
Hmm.  The fix looks good if we want to make this change, I'm just not entirely convinced it's a change we want to make.  There are some cases where you create a periodic timer with the goal of having it fire as close to set periods of wall clock time as possible, and for that, I agree, this is a desired change.  But there are other cases where you don't care about actual wall clock time, and instead want to do something repeatedly some period of time after the last time you did it, in which case you don't want this: if one firing is delayed, then you still want to do the next firing the original period of time after that one occurred, not after it was supposed to occur.

kouvel:
Possible behaviors:
1. Timer ticks at predefined intervals (without drift)
2. Timer's next tick occurs at a predefined interval after the start of previous tick
3. Timer's next tick occurs at a predefined interval after the end of the previous tick

Of the above cases, I feel that (1) and (3) are the most useful. If a timer didn't say anything, I would expect that it behaves like (1) or (3). Since our timer says it allows multiple callbacks to be raised concurrently, I would expect that it behaves like (1). It's not very clear that our timer behaves like (2) and I don't think that is very useful.

Any of these behaviors can be coded for regardless of how the timer behaves, by making the timer non-periodic and rescheduling the timer in a callback at the desired point and with the desired due time. Considering that (2) is pretty close to (1) and since I feel that (1) is the more useful / expected behavior for a timer that does not wait for a callback to complete before scheduling the next callback, it does not feel like a breaking change to me, it feels more like a bug fix.

CC @gkhanna79 

Thoughts?

stephentoub:
Ok, sounds reasonable.

kouvel:
Thanks!

kouvel:

PR #11221, opened 2017-04-26
[x86/Linux] Stack alignment check in Hijack Helpers

parjong:
These helpers are already aligned, and thus this commit simply adds a check.


jkotas:
Thanks!

PR #11222, opened 2017-04-26
[ARM32/Linux] Copy tests.zip only for CI test and use x86 unittest

hseok-oh:
- Use x86 unittest instead of x64: 32bit architecture
- Copy tests.zip only from x86 Windows build for CI test


hseok-oh:
@dotnet-bot test ci please

RussKeldorph:
@hseok-oh Is this causing all ARM CI jobs to fail?  For example: https://github.com/dotnet/coreclr/pull/11350

Issue #11159, opened 2017-04-23
Doesn't build with clang/llvm 4.0


juergenpf:
Using the clang/llvm 4.0 toolchain I'm getting build errors from pal.h, complaining about defining builtin functions _rotl, _rotr etc.

gkhanna79:
CC @janvorli

janvorli:
@juergenpf there are actually more issues than those two builtins. Clang 4.0 also complains about some clobber annotations in inline asm code, SSIZE_T_MIN constant, _mul128 argument types and few comparisons of pointers as > 0.

janvorli:
I almost have these issues fixed.

janvorli:
Fixed by #11226

PR #11226, opened 2017-04-26
Enable build with clang 4.0

janvorli:
This change enables build with clang 4.0 and fixes a bunch of new errors that
the stricter compiler was reporting.
The new clang has `_rotl` and `_rotr` as intrinsics, complains about > 0 comparison of pointers, reports as error invalid constant literal `-9223372036854775808LL` (the literal type suffix applies to the value without the sign and so it overflows the long long type) and also had issues with _umul128 parameter types.
Since UnsignedMultiply128 that is defined as _umul128 and none of its callers are used anywhere, I have just deleted it and all of its callers.


janvorli:
Fixes #11159
CC: @juergenpf

PR #11227, opened 2017-04-26
Enable property to skip build number in packaging.

gkhanna79:
@weshaggard PTAL

This will be ported to release/2.0.0 for stabilization.

CC @RussKeldorph @leecow


weshaggard:
You don't actually need to add this here, you just need to set it to false when we want to turn off the build numbers. Unless you want it hear only for documentation.

PR #11233, opened 2017-04-26
Fix lsra memory consumption

### LINKED ISSUE #11142: Test failure: JIT_Regression._JitBlue_DevDiv_255294_DevDiv_255294_DevDiv_255294_/_JitBlue_DevDiv_255294_DevDiv_255294_DevDiv_255294_cmd
Opened on behalf of @gkhanna79

The test `JIT_Regression._JitBlue_DevDiv_255294_DevDiv_255294_DevDiv_255294_/_JitBlue_DevDiv_255294_DevDiv_255294_DevDiv_255294_cmd` has failed.

    Unhandled Exception: OutOfMemoryException.
    /home/helixbot/dotnetbuild/work/f4cb4c15-243a-4fd1-9f7c-c609e2f8f545/Work/52a9157d-ed0e-498e-91fc-3551ee7dc58a/Unzip/JitBlue/DevDiv_255294/DevDiv_255294/DevDiv_255294.sh: line 200:  4775 Aborted                 (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" $ExePath $CLRTestExecutionArguments
    
    
    Return code:      1
    Raw output file:      /home/helixbot/dotnetbuild/work/f4cb4c15-243a-4fd1-9f7c-c609e2f8f545/Work/52a9157d-ed0e-498e-91fc-3551ee7dc58a/Unzip/Reports\JIT.Regression\JitBlue\DevDiv_255294\DevDiv_255294\DevDiv_255294.output.txt
    Raw output:
    BEGIN EXECUTION
    /home/helixbot/dotnetbuild/work/f4cb4c15-243a-4fd1-9f7c-c609e2f8f545/Payload/corerun DevDiv_255294.exe
    Expected: 100
    Actual: 134
    END EXECUTION - FAILED
    Test Harness Exitcode is : 1
    
    To run the test:
    > set CORE_ROOT=/home/helixbot/dotnetbuild/work/f4cb4c15-243a-4fd1-9f7c-c609e2f8f545/Payload
    > /home/helixbot/dotnetbuild/work/f4cb4c15-243a-4fd1-9f7c-c609e2f8f545/Work/52a9157d-ed0e-498e-91fc-3551ee7dc58a/Unzip/JitBlue/DevDiv_255294/DevDiv_255294/DevDiv_255294.sh
    
    Expected: True
    Actual:   False

        Stack Trace:

        
Build : Master - 20170421.04 (Core Tests)
Failing configurations:
- ubuntu.1604.amd64
  - x64
- debian.82.amd64
  - x64

sandreenko:
Do not allocate new buffer each time when we call Clear for bitset.
It removes quadratic memory consumption for predecessors in lsra (N * N/ 64 size* were allocated, where N is the number of blocks).

Use the same preallocated buffer for lsra preds.

There are no spmi asm diffs.
Fix #11142, it allocated 3GB of memory, now it takes 300 MB,

PTAL @dotnet/jit-contrib , as I understand I have to merge it into the release branch too.


russellhadley:
Pretty cool.  Does this move the needle on memstats or TP?

src/jit/lsra.h
```diff
@@ -1131,7 +1131,7 @@ class LinearScan : public LinearScanInterface
     int compareBlocksForSequencing(BasicBlock* block1, BasicBlock* block2, bool useBlockWeights);
     BasicBlockList* blockSequenceWorkList;
     bool            blockSequencingDone;
-    void addToBlockSequenceWorkList(BlockSet sequencedBlockSet, BasicBlock* block);
+    void addToBlockSequenceWorkList(BlockSet sequencedBlockSet, BasicBlock* block, BlockSet predSet);
     void removeFromBlockSequenceWorkList(BasicBlockList* listNode, BasicBlockList* prevNode);
```
BruceForstall: These should really use BlockSet_ValArg_T instead of BlockSet

src/jit/bitsetasuint64inclass.h
```diff
@@ -180,11 +180,10 @@ class BitSetUint64
 
     inline void ClearD(Env env)
     {
-        // Recall that ClearD does *not* require "*this" to be of the current epoch.
+        // Recall that ClearD requires "*this" to be of the current epoch.
         Uint64BitSetOps::ClearD(env, m_bits);
```
CarolEidt: Given that this is changing the requirements on ClearD, have you examined all the callsites to ensure that they are consistent with this? I would suggest that this undergo additional testing (e.g. desktop SPMI) to ensure that we don't encounter this assert.

briansull:
@sandreenko
 I would like to take this fix for the Desktop 4.7.1 release of the JIT as well.
@CarolEidt 
Can you confirm that it is low risk.

src/jit/lsra.cpp
```diff
@@ -1540,6 +1540,7 @@ int LinearScan::compareBlocksForSequencing(BasicBlock* block1, BasicBlock* block
 // Arguments:
 //    sequencedBlockSet - the set of blocks that are already sequenced
 //    block             - the new block to be added
+//    predSet           - the buffer to save predessesors set
```
BruceForstall: typo: predessesors

I think you should elaborate, e.g.: "a block set allocated by the caller used here as a temporary block set for constructing a predecessor set. Allocated by the caller to avoid reallocating a new block set with every call to this function"

src/jit/lsra.h
BruceForstall: Although I guess you don't want to pass by value, do you? Should it be `BlockSet&`?

---
In reply to: [113554764](https://github.com/dotnet/coreclr/pull/11233#discussion_r113554764) [](ancestors = 113554764)

src/jit/bitsetasshortlong.h
```diff
@@ -664,9 +664,11 @@ void BitSetOps</*BitSetType*/ BitSetShortLongRep,
                /*BitSetTraits*/ BitSetTraits>::ClearDLong(Env env, BitSetShortLongRep& bs)
 {
     assert(!IsShort(env));
-    // Recall that ClearD does *not* require "bs" to be of the current epoch.
-    // Therefore, we must allocate a new representation.
-    bs = MakeEmptyArrayBits(env);
```
pgavlin: This change makes me *very* nervous, as we do not seem to have any way to verify that the size of the bitset has not changed since we created it (i.e. there appears to be no way to check that the epoch at creation is the same as the epoch at the clear). I think you should introduce a new method (perhaps `ClearDUnsafe`? I don't have a great name for it) with the semantics you've introduced here.

CarolEidt:
@briansull - I think that this is low risk for the register allocator. I think that, with a full SPMI run on desktop, I would consider it low risk overall (i.e. ensuring that there aren't other places where ClearD is called with a bitset of an old epoch.)

src/jit/lsra.h
BruceForstall: Maybe:
```
void addToBlockSequenceWorkList(const BlockSet& sequencedBlockSet, BasicBlock* block, BlockSet& predSet);
```
since sequencedBlockSet is const/unchanged, and predSet is changed. If you don't use `&` then for short form bitsets you'll pass by value. It turns out that doesn't (currently) matter, since they don't allocate memory and you only care about reducing memory, but it would be weird for short bitsets to be passed by value and long bitsets to be passed by reference, accidentally.

pgavlin: A value arg should be fine here, since the bitset is cleared upon each call to `addBlockSequenceWorklist`. This should probably use `BlockSet_ValArg_T`, though.

src/jit/bitsetasshortlong.h
BruceForstall: @sandreenko We had talked about you converting existing ClearD() callers to call MakeEmpty() directly, for existing callers that could not easily be proven to work with the new semantics. I presume the existing semantics, and huge comment you deleted for ClearD(), were there for a reason.

Is there some reason you decided not to do that? Could you somehow convince yourself the new semantics work for every call site?

src/jit/bitsetasuint64inclass.h
sandreenko: I checked that there are no spmi failures or spmi diffs.

src/jit/bitsetasshortlong.h
sandreenko: I manually checked all calls to ClearD, the initial idea was to replace them with MakeEmpty when it is necessary. But It appeared that all current users of  ClearD do not change epoch.

CarolEidt: One can `assert(CheckEpoch(env))` which is being done below in bitsetasuint64inclass.h. I believe that assert could be made here as well.

pgavlin: > But It appeared that all current users of ClearD do not change epoch.

I see twenty uses of `ClearD` in the compiler today:
```
src/jit/assertionprop.cpp:4559:    BitVecOps::ClearD(apTraits, fgFirstBB->bbAssertionIn);
src/jit/codegenlegacy.cpp:12585:        VarSetOps::ClearD(compiler, compiler->compCurLife);
src/jit/copyprop.cpp:299:        VarSetOps::ClearD(this, optCopyPropKillSet);
src/jit/emit.cpp:1475:    VarSetOps::ClearD(emitComp, emitInitGCrefVars);
src/jit/emit.cpp:1476:    VarSetOps::ClearD(emitComp, emitPrevGCrefVars);
src/jit/emit.cpp:4567:    VarSetOps::ClearD(emitComp, emitThisGCrefVars); // This is initialized to Empty at the start of codegen.
src/jit/flowgraph.cpp:1818:        // Initialize the per-block bbReach sets. (Note that we can't just call BlockSetOps::ClearD()
src/jit/flowgraph.cpp:10014:        BlockSetOps::ClearD(this, bNext->bbReach);
src/jit/liveness.cpp:489:        VarSetOps::ClearD(this, fgCurUseSet);
src/jit/liveness.cpp:490:        VarSetOps::ClearD(this, fgCurDefSet);
src/jit/liveness.cpp:725:            VarSetOps::ClearD(this, inScope);
src/jit/liveness.cpp:915:        VarSetOps::ClearD(this, initVars);
src/jit/liveness.cpp:1151:        VarSetOps::ClearD(m_compiler, m_liveOut);
src/jit/liveness.cpp:1265:            VarSetOps::ClearD(m_compiler, m_liveIn);
src/jit/liveness.cpp:1266:            VarSetOps::ClearD(m_compiler, m_liveOut);
src/jit/lsra.cpp:1741:    BlockSetOps::ClearD(compiler, bbVisitedSet);
src/jit/lsra.cpp:9317:            VarSetOps::ClearD(compiler, sameResolutionSet);
src/jit/lsra.h:1109:        BlockSetOps::ClearD(compiler, bbVisitedSet);
src/jit/regalloc.cpp:1343:            VarSetOps::ClearD(this, varAsSet);
src/jit/regalloc.cpp:6351:            VarSetOps::ClearD(this, raLclRegIntf[i]);
```

How did you verify that the epoch of each bitset does not change between the creation of each bitset and the point at which they are cleared? If the epoch has changed, the failure mode could be pretty bad, as if the number of bits has increased we could end up scribbling over arbitrary memory.

pgavlin: > We had talked about you converting existing ClearD() callers to call MakeEmpty() directly, for existing callers that could not easily be proven to work with the new semantics.

This sounds like a much safer change to me.

pgavlin: (really, anything that avoids changing the semantics of existing calls is fine)

pgavlin: > One can `assert(CheckEpoch(env))` which is being done below in bitsetasuint64inclass.h. I believe that assert could be made here as well.

I don't believe that it can, as we don't seem to store the epoch for this sort of bitset anywhere.

src/jit/lsra.h
BruceForstall: Conceptually, we don't want to be passing these bitsets "by value".

pgavlin: That's fair. Your suggestion to use a reference parameter should suffice, then.

src/jit/bitsetasshortlong.h
CarolEidt: I may be confused here, but don't we use `Compiler` for the `Env` of all bit set implementations?

pgavlin: Possibly, but `CheckEpoch` is an instance method on class `BitSetUint64` that asserts that the `m_epoch` field of that class matches the provided epoch (i.e. the epoch retrieved from `Env`). The short/long bitset implementation does not have an equivalent of `m_epoch` that it can use to perform this check (at least not that I can find, but perhaps I'm missing something).

BruceForstall: Yes, except for a few minor cases where we don't want to participate in the "epoch" system.

CheckEpoch only exists for bitsetasuint64inclass, which we don't use, not for shortlong bitsets, which we do.

Ideally, we would implement a DEBUG ShortLong representation bitset that also carried epoch along with every bitset, and asserted that operations between two bitsets were in the same epoch.

src/jit/lsra.cpp
sandreenko: Fixed, thanks.

src/jit/lsra.h
sandreenko: Fixed

src/jit/bitsetasshortlong.h
CarolEidt: Ah, of course. That said, given that most of the methods do not work across Epochs, another option would be to run SPMI with a JIT that always uses the class implementation of BitSet.

pgavlin: >Ideally, we would implement a DEBUG ShortLong representation bitset that also carried epoch along with every bitset, and asserted that operations between two bitsets were in the same epoch.

https://github.com/dotnet/coreclr/issues/11236

sandreenko: > This  sounds like a much safer change to me.

I think that in this case if we replace all ClearD with MakeEmpty we will lose a valuable information.

pgavlin: > I think that in this case if we replace all ClearD with MakeEmpty we will lose a valuable information.

Yes, this is a good point. You should instead create a separate `ClearD` method with the desired semantics.

FWIW, I can attest that liveness would also be able to use the new clear method in all cases that it currently uses `ClearD`.

BruceForstall: Agreed. We could replace it with a new Clear() function that is non-destructive.

Now, it's possible that all the ClearD functions today can use the new semantics, meaning we don't need a new Clear function. It's just very suspicious that is true, given the huge comments about why ClearD is (currently) special.

pgavlin: > It's just very suspicious that is true, given the huge comments about why ClearD is (currently) special.

This is my concern as well, especially given that the potential impacts of a mismatch are rather severe.

src/jit/bitsetasuint64inclass.h
```diff
@@ -180,11 +180,10 @@ class BitSetUint64
...
 #ifdef DEBUG
-        // But it updates it to of the current epoch.
-        m_epoch = BitSetTraits::GetEpoch(env);
+        assert(CheckEpoch(env));
 #endif
```
BruceForstall: No need for `#ifdef DEBUG`. Just make it:
```
    inline void ClearD(Env env)
    {
        assert(CheckEpoch(env));
        Uint64BitSetOps::ClearD(env, m_bits);
    }
```

src/jit/bitsetasshortlong.h
sandreenko: When I first looked at ClearD I thought it just clear bits without this epoch magic.
It was unique function, but nobody used it in such way.
For example:
flowgraph.cpp:        
```
// Initialize the per-block bbReach sets. (Note that we can't just call BlockSetOps::ClearD()
// when re-running this computation, because if the epoch changes, the size and representation of the
// sets might change).
```
So I prefer not to create the second function, that will confuse us in the future, but to replace the first.

However: liveness, lsra, regalloc, assertionprop, copyprop and emit  do not change the epoch.
flowgraph does, but it call MakeEmpty before ClearD.

sandreenko:
@pgavlin @BruceForstall I updated PR according to our discussion.

src/jit/emit.cpp
```diff
@@ -1472,8 +1472,8 @@ void emitter::emitBegProlog()
     /* Nothing is live on entry to the prolog */
 
     // These were initialized to Empty at the start of compilation.
-    VarSetOps::ClearD(emitComp, emitInitGCrefVars);
-    VarSetOps::ClearD(emitComp, emitPrevGCrefVars);
+    VarSetOps::OldStyleClearD(emitComp, emitInitGCrefVars);
+    VarSetOps::OldStyleClearD(emitComp, emitPrevGCrefVars);
```
CarolEidt: This begs the question of whether these should not be initialized until CodeGen.

src/jit/flowgraph.cpp
```diff
@@ -1815,7 +1815,7 @@ void Compiler::fgComputeReachabilitySets()
 
     for (block = fgFirstBB; block != nullptr; block = block->bbNext)
     {
-        // Initialize the per-block bbReach sets. (Note that we can't just call BlockSetOps::ClearD()
```
CarolEidt: I think this comment is incorrect. I'm fine with leaving the code as-is, but at least the comment should be fixed.

src/jit/emit.cpp
pgavlin: I agree, but we can make that change (if necessary) with a post-2.0 PR.

CarolEidt:
I would like to see an issue to either remove all uses of `OldStyleClearD` or rename it, and to generally review the usage of the two flavors of `ClearD` and `MakeEmpty`

pgavlin:
> I would like to see an issue to either remove all uses of OldStyleClearD or rename it, and to generally review the usage of the two flavors of ClearD and MakeEmpty

@sandreenko filed https://github.com/dotnet/coreclr/issues/11263. Does that suffice?

src/jit/emit.cpp
sandreenko: Fixed

CarolEidt:
> @sandreenko filed #11263. Does that suffice?

Yes, thanks.

Issue #10443, opened 2017-03-24
Marshal.SecureStringToBSTR and ZeroFreeBSTR not implemented for Unix


danmosemsft:
We have not implemented `Marshal.SecureStringToBSTR` for Unix.

https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs#L1775

This is bug/inconsistency. We have implemented BSTR on Unix, even though it is COM interopy. Mono took the same approach. We should remove the FEATURE_COMINTEROP ifdef around this one and make it work on Unix - to be consistent with other methods that take BSTRs.

Also `Marshal.ZeroFreeBSTR` can be implemented for Unix.

danmosemsft:
`SecureStringToBSTR` seems to have high usage on Xamarin. `ZeroFreeBSTR` has some usage, but small.

danmosemsft:
This is a nice to have for 2.0

jkotas:
cc @yizhang82

yizhang82:
@botaberg Can you take a look?

botaberg:
I can take a look. Thanks!

botaberg:
@yizhang82 Is this issue still high-pri for 2.0? I know we talked about this for a bit offline, but I haven't gotten to implementing it yet.

danmosemsft:
@botaberg there is some usage, it's low but we've learned it's papercuts that break compat. Is there an idea of how much work this is? We should certainly do it if it's just a few lines. IMO.

yizhang82:
It's low pri but I agree we should simply go do it. It's definitely a bit more than a few lines - we would need to:
* Enable SecureStringToBSTR for CoreCLR (it should just work).
* Add corresponding p/invokes for S.P.CoreLib as InteropExtension / ExternalInterop that points to Windows API on Windows and stub them out for non-windows
* (Nice to have) Add corresponding C# implementation for those p/invokes for non-Windows (basically porting coreclr/src/palrt/bstr.cpp to managed code). 
* (Nice to have) Hook up those APIs to various Marshal.BSTR implementation for non-windows as well

jkotas:
This bug is specifically about `Marshal.SecureStringToBSTR` for CoreCLR. (1. in your list - it should be easy.)

Enabling BSTRs for CoreRT on Unix (2.-4. in your list) should be tracked in CoreRT repo with the rest of the interop porting work (https://github.com/dotnet/corert/issues/669).

[...]

PR #11234, opened 2017-04-26
Enable Marshal.SecureStringToBSTR and Marshal.ZeroFreeBSTR on Unix

botaberg:
We have already created a native implementation of BSTR in CoreCLR which is available in Unix: https://github.com/dotnet/coreclr/blob/master/src/palrt/bstr.cpp

This change fixes issue #10443 by ensuring that Marshal.SecureStringToBSTR and ZeroFreeBSTR are available for CoreCLR on Unix. Previously, SecureStringToBSTR would throw PlatformNotSupportedException and ZeroFreeBSTR was not available. These methods still will not be available on CoreRT on Unix, despite this code being shared between CoreCLR and CoreRT, because CoreRT does not have access to the native CoreCLR assembly which contains the native BSTR implementation.


src/mscorlib/shared/Interop/Windows/Interop.Libraries.cs
```diff
@@ -8,8 +8,15 @@ internal static partial class Libraries
     {
         internal const string BCrypt = "BCrypt.dll";
         internal const string Crypt32 = "crypt32.dll";
+
+#if !FEATURE_PAL
```
jkotas: We do not want to have `FEATURE_PAL` ifdefs in the shared partition.

jkotas:
This looks incomplete...

botaberg:
@jkotas You are right - I also had a change in Marshal.cs to remove the #ifdef. Is there anything else you are looking for that may be missing?

jkotas:
Tests?

botaberg:
Closing this for now. Will update with a full set of changes verified on Unix.

src/mscorlib/shared/Interop/Unix/Interop.Libraries.cs
```diff
@@ -8,5 +8,6 @@ internal static partial class Libraries
     {
         internal const string GlobalizationInterop = "System.Globalization.Native";
         internal const string SystemNative = "System.Native";
+        internal const string LibCoreCLR = "libcoreclr";
```
jkotas: mscorlib/shared is code shared between CoreCLR and CoreRT. We do not want to have LibCoreCLR defined here because of libcoreclr does not exist in CoreRT.

jkotas: BTW: It will cause build breaks once this code gets mirrored to CoreRT

botaberg: We will have to define the PInvoke to libcoreclr somewhere; what is the correct location for this. If we put it in Shared, can we wrap it in #if CORECLR tags?

The fundamental problem is that the coreclr dll exists and can be accessed on CoreCLR, but not on CoreRT. We are punting this problem until later, but we still need CoreRT to build.

yizhang82: Here is what you need to do:
* Your windows p/invoke can be shared between CoreCLR and CoreRT
* For Unix, put your coreclr p/invoke under #ifdef CORECLR, and #else throw PNSE

jkotas: @yizhang82 Should the BSTR alloc/free be abstracted behind PInvokeMarshal - similar to what is done for regular memory allocations?

botaberg:
@yizhang82 Can you take a look at these revisions please?

yizhang82:
@botaberg Looks good to me but please add tests into SecureString test to cover these scenarios

tests/src/Interop/MarshalAPI/String/StringMarshalingTest.cs
```diff
@@ -52,6 +52,28 @@ private void StringToBStrToString()
...
+
+            IntPtr BStr = Marshal.SecureStringToBSTR(secureString);
+            String str = Marshal.PtrToStringBSTR(BStr);
+
+            if (!str.Equals(ts))
+            {
+                throw new Exception();
+            }
+
+            Marshal.ZeroFreeBSTR(BStr);
```
yizhang82: Add this in finally?

yizhang82:
@jkotas I agree this looks better. Thanks! I think you might missed corresponding PInvokeMarshal change in CoreRT to match CoreCLR. Otherwise LGTM.

jkotas:
> I think you might missed corresponding PInvokeMarshal change 

Right - I can take care of it as the change is integrated to CoreRT by the CoreLib mirror.

tests/src/Interop/MarshalAPI/String/StringMarshalingTest.cs
```diff
@@ -52,6 +52,28 @@ private void StringToBStrToString()
...
+            SecureString secureString = new SecureString();
+            foreach (char character in ts)
+            {
+                secureString.AppendChar(character);
+            }
+
+            IntPtr BStr = Marshal.SecureStringToBSTR(secureString);
+            String str = Marshal.PtrToStringBSTR(BStr);
+
+            if (!str.Equals(ts))
```
danmosemsft: just Assert.Equals(str, ts) ? that dumps the strings in the test log if they differ.

botaberg:
@shrah want to take a look as well?

botaberg:
@jkotas @yizhang82 I'm investigating the test failure here - looks like stack overflow occurs in the test for SecureStringToBSTR. Have any pointers?

src/mscorlib/src/System/Runtime/RuntimeImports.cs
```diff
@@ -26,8 +26,13 @@ internal unsafe static void RhZeroMemory(ref byte b, nuint byteLength)
             }
         }
 
+        internal unsafe static void RhZeroMemory(IntPtr p, UIntPtr byteLength)
+        {
+            RhZeroMemory(p, (UIntPtr)byteLength);
```
jkotas: This needs to be `RhZeroMemory((void*)p, (uint)byteLength);` instead.

jkotas:
> looks like stack overflow occurs in the test for SecureStringToBSTR

@botaberg I have flagged the place where I believe the bug is.

src/mscorlib/src/System/Runtime/RuntimeImports.cs
```diff
@@ -26,8 +26,13 @@ internal unsafe static void RhZeroMemory(ref byte b, nuint byteLength)
             }
         }
 
+        internal unsafe static void RhZeroMemory(IntPtr p, UIntPtr byteLength)
+        {
+            RhZeroMemory((void*)p, (uint)byteLength);
```
jkotas: This should be casting to `nuint`

botaberg:
@jkotas does this look good?

PR #11235, opened 2017-04-26
Require a frame pointer in methods with throw blocks for sources that…

pgavlin:
… have mismatched stack depths.

This requirement effectively disables double-aligned frames (which are
effectively a special case of ESP frames). Without this change it is
possible for the JIT to generate a throw block with sources that have
mismatched offsets, which can lead to GC holes.

Fixes VSO 424019.


pgavlin:
@BruceForstall @briansull @dotnet/jit-contrib PTAL.

PR #2082, opened 2015-11-17
Fix RyuJIT assert 'emitCurIG != emitPrologIG'

BruceForstall:
This occurs when the call-to-finally block exists as the last code in the function
(including any funclets). After we generate all "normal" code, we start over and
generate code for all the prologs and epilogs. In this case, we didn't reset the
emitForceNewIG global, which determines if we should force a new IG to be created
the next time code is generated. It is used for specifying no-GC regions. This fix
initializes it (along with emitNOGCIG) at the beginning of generating and prolog
or epilog.

PR #11237, opened 2017-04-26
Sign SOS.NETCore assembly

mikem8361:
Github issue #2082

VS Bug #409061 https://devdiv.visualstudio.com/DevDiv/_workitems?id=409061&_a=edit

### DIFF
--- a/src/ToolBox/SOS/NETCore/SOS.NETCore.csproj
+++ b/src/ToolBox/SOS/NETCore/SOS.NETCore.csproj
@@ -12,6 +12,8 @@
     <NoStdLib>true</NoStdLib>
     <NoCompilerStandardLib>true</NoCompilerStandardLib>
     <UseOpenKey Condition="'$(UseOpenKey)'==''">true</UseOpenKey>
+    <SignAssembly>true</SignAssembly>
+    <DelaySign>true</DelaySign>
   </PropertyGroup>
 
   <!-- Default configurations to help VS understand the options -->


src/ToolBox/SOS/NETCore/SOS.NETCore.csproj
```diff
@@ -12,6 +12,8 @@
     <NoStdLib>true</NoStdLib>
     <NoCompilerStandardLib>true</NoCompilerStandardLib>
     <UseOpenKey Condition="'$(UseOpenKey)'==''">true</UseOpenKey>
+    <SignAssembly>true</SignAssembly>
```
weshaggard: These shouldn't be needed at all if sign.targets is being imported.

src/ToolBox/SOS/NETCore/SOS.NETCore.csproj
```diff
@@ -12,6 +12,8 @@
...
+    <DelaySign>true</DelaySign>
```
weshaggard: For anything that is being signed with the Open key you cannot delay sign it because it is a full key and the signing lab will reject it.

weshaggard: For anything that is being signed with the Open key you cannot delay sign it because it is a full key and the signing lab will reject it.

weshaggard:
When you build it locally is it signed?

Is the signing issue talking about strong name or authenticode signing? It isn't clear to me from the issue.

mikem8361:
From your previous comments I shouldn't need or have either of the properties I've added that my changes are not needed. The csproj for System.Private.CoreLib has both keys. 

After my changes I did "sn -vf SOS.NETCore.dll" which said it was a valid assembly. I didn't tried this before the changes.

I assumed it was just strong name signing that is required.  I know even less about authenticode sigining. Neither VS bug nor the GitHub issue give any indication which it is.

mikem8361:

mikem8361:
@weshaggard Can you approve this? Before my changes SOS.NETCore wasn't strong name signed and after it is. 

@gkhanna79 Do I have to go through ship room to get this into 1.1.x?

weshaggard:
>@weshaggard Can you approve this? Before my changes SOS.NETCore wasn't strong name signed and after it is.

I don't know how it is strong name signed after this change because you have it set to DelaySign=true which would mean that is delay signed at best. Also this doesn't say anything about authenticode signing to see whether or not it is correctly setup for signing. @wtgodbe might know more about signing in the coreclr repo.

weshaggard:
I did a quick verification of what comes out of coreclr and I think this assembly is already correctly strong name and authenticode signed.

I verified:
".nuget\packages\runtime.win7-x86.Microsoft.NETCore.Runtime.CoreCLR\1.1.0\runtimes\win7-x86\lib\netstandard1.0\SOS.NETCore.dll"

So I think this PR should be closed I think the only signing that is needed is part of the core-setup shared framework build, where we crossgen this assembly which invalidates the signatures and we then need to re-sign.

PR #11244, opened 2017-04-26
Fix a double-free encountered when failing to initialize a handle table bucket

swgillespie:
Before https://github.com/dotnet/coreclr/pull/11092, `Ref_InitializeHandleTableBucket` would return one of two ways: it would either return `true` or it would throw an `OutOfMemoryException`. As a result, the affected code path was effectively dead. After #11092, though, `Ref_InitializeHandleTableBucket` returns false on allocation failure, which caused us to take this path even after `Ref_InitializeHandleTableBucket` has already freed `result`, resulting in a double-free.

This issue was found with clang static analysis. cc @Maoni0 @sergiy-k @adityamandaleeka PTAL?

### DIFF
--- a/src/gc/objecthandle.cpp
+++ b/src/gc/objecthandle.cpp
@@ -608,7 +608,8 @@ HandleTableBucketHolder::~HandleTableBucketHolder()
         }
         delete [] m_bucket->pTable;
     }
-    delete m_bucket;
+
+    // we do not own m_bucket, so we shouldn't delete it here.
 }
 
 bool Ref_Initialize()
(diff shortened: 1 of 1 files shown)


swgillespie:
ah, you're right - will do.

I also noticed that `HandleTableBucketHolder` will attempt to delete `&g_gcGlobalHandleStore->_underlyingBucket` [here](https://github.com/dotnet/coreclr/blob/master/src/gc/objecthandle.cpp#L642), which is particularly evil since it is located at offset 0 within `GCHandleStore`, so `g_gcGlobalHandleStore` will also get deleted twice.

swgillespie:
thanks!

PR #11245, opened 2017-04-26
Fix to respect readonly prefix for multidimensional array Address method

### LINKED ISSUE #9414: Readonly prefix not respected for multidimensional array address method
The below code works on desktop CLR, but throws an `ArrayTypeMismatchException` on CoreCLR:

```csharp
using System;

class Program
{
    interface IFrobber
    {
        void Frob();
    }

    class Frobber : IFrobber
    {
        public void Frob()
        {
        }
    }

    class Foo<T> where T : IFrobber
    {
        public static void FrobAll(T[,] arr)
        {
            for (int i = 0; i < arr.Length; i++)
                arr[0, i].Frob();
        }
    }

    {
        Foo<IFrobber>.FrobAll(new Frobber[,] { { new Frobber() } });
    }
}
```

The `FrobAll` method has the following IL sequence in it:

```
readonly.
call instance !0& !T[0..., 0...]::Address(int32, int32)
constrained. !T
callvirt instance void Program/IFrobber::Frob()
```

As per ECMA-335 III.2.3 the `readonly.` prefix shall suppress the exact type check done by the `Address` method and cause it to return a controlled-mutability managed pointer to the element of the array. The VM is not suppressing the type check in the IL stub.

Note this works when optimizations are enabled.

kouvel:
Fixes #9414
- Added missing 'else' back from regression in https://github.com/dotnet/coreclr/commit/97b4ff0b438261ba11b357008630076054a6f25d#diff-edc46b80f57431489a82311948a8234dL6424, which has the effect of passing null to the Address method's instParam to indicate the readonly prefix in order to bypass the type check

### DIFF
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -7325,8 +7325,7 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
                 // instParam.
                 instParam = gtNewIconNode(0, TYP_REF);
             }
-
-            if (!exactContextNeedsRuntimeLookup)
+            else if (!exactContextNeedsRuntimeLookup)
             {
 #ifdef FEATURE_READYTORUN_COMPILER
                 if (opts.IsReadyToRun())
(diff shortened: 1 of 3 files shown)


jkotas:
cc @dotnet/jit-contrib @briansull You may want to cherry pick this to desktop.

briansull:
Thanks for the heads up @jkotas I will pick up this change

pgavlin:
cc @sandreenko

kouvel:
Thanks all

PR #11251, opened 2017-04-27
Bring back Marshal.Read/Write overloads that takes object

yizhang82:
The original desktop implementation is implemented on top of mscoree.ND_R* and ND_W * helpers with AsAny marshalling,which is super odd but I can understand the convenience. I've re-implemented the API on top of AsAny marshaler so that it could work cross-plat without having to introduce new helpers into PAL. I've also introduced a bit of error checking for null case (desktop simply AVs).
I've also added a simple test as well.
@jkotas @shrah PTAL
/cc @danmosemsft @Petermarcu

### DIFF
--- a/src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs
+++ b/src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs
@@ -31,6 +31,7 @@ namespace System.Runtime.InteropServices
     using System.Diagnostics;
     using System.Diagnostics.Contracts;
     using System.Runtime.InteropServices.ComTypes;
+    using System.StubHelpers;
 
     [Serializable]
     public enum CustomQueryInterfaceMode
@@ -408,7 +409,7 @@ namespace System.Runtime.InteropServices
         //====================================================================
         public static byte ReadByte([MarshalAs(UnmanagedType.AsAny), In] Object ptr, int ofs)
         {
-            throw new PlatformNotSupportedException(); // https://github.com/dotnet/coreclr/issues/10442
+            return ReadValueSlow(ptr, ofs, (IntPtr nativeHome, int offset) => { return Marshal.ReadByte(nativeHome, offset); });
         }
 
         public static unsafe byte ReadByte(IntPtr ptr, int ofs)
@@ -432,7 +433,7 @@ namespace System.Runtime.InteropServices
 
         public static short ReadInt16([MarshalAs(UnmanagedType.AsAny), In] Object ptr, int ofs)
         {
-            throw new PlatformNotSupportedException(); // https://github.com/dotnet/coreclr/issues/10442
+            return ReadValueSlow(ptr, ofs, (IntPtr nativeHome, int offset) => { return Marshal.ReadInt16(nativeHome, offset); });
         }
 
         public static unsafe short ReadInt16(IntPtr ptr, int ofs)
@@ -467,9 +468,9 @@ namespace System.Runtime.InteropServices
             return ReadInt16(ptr, 0);
         }
 
-        public static int ReadInt32([MarshalAs(UnmanagedType.AsAny), In] Object ptr, int ofs)
+        public static int ReadInt32(object ptr, int ofs)
         {
-            throw new PlatformNotSupportedException(); // https://github.com/dotnet/coreclr/issues/10442
+            return ReadValueSlow(ptr, ofs, (IntPtr nativeHome, int offset) => { return Marshal.ReadInt32(nativeHome, offset); });
         }
 
         public static unsafe int ReadInt32(IntPtr ptr, int ofs)
@@ -535,7 +536,7 @@ namespace System.Runtime.InteropServices
 
         public static long ReadInt64([MarshalAs(UnmanagedType.AsAny), In] Object ptr, int ofs)
         {
-            throw new PlatformNotSupportedException(); // https://github.com/dotnet/coreclr/issues/10442
+            return ReadValueSlow(ptr, ofs, (IntPtr nativeHome, int offset) => { return Marshal.ReadInt64(nativeHome, offset); });
         }
 
         public static unsafe long ReadInt64(IntPtr ptr, int ofs)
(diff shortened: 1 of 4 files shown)


Petermarcu:
Great. Thanks @yizhang82 . I agree on the perf and also that throwing an exceptions instead of AVing is a good thing :)

tests/src/Interop/MarshalAPI/ReadWrite/ReadWriteObject.cs
```diff
@@ -0,0 +1,119 @@
+﻿using System;
```
danmosemsft: Do we have any rules about when we test corelib via coreclr/tests vs. corefx tests? The latter makes it easier to test compat with other platforms and also keeps everything in one place. Of course coreclr is convenient..

src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs
```diff
@@ -749,16 +785,51 @@ public static unsafe void WriteInt64(IntPtr ptr, int ofs, long val)
...
+        // propagate the value back
+        // It's quite slow and is only there for backcompact
+        // I don't think we should spend time optimizing it
+        // People should really call the IntPtr overload instead
+        //====================================================================
+        private static unsafe void WriteValueSlow<T>(object ptr, int ofs, T val, Action<IntPtr, int, T> writeValueHelper)
+        {
+            // We AV on desktop if passing NULL. So this is technically a breaking change but is an improvement
+            if (ptr == null)
+                throw new ArgumentNullException("ptr");
```
danmosemsft: nit: `throw new ArgumentNullException(nameof(ptr));`

src/mscorlib/src/System/StubHelpers.cs
```diff
@@ -700,6 +700,17 @@ static internal IntPtr CreateNativeNCCEventArgsInstance(int action, object newIt
 
     internal static class MngdNativeArrayMarshaler
     {
+        // Needs to match exactly with MngdNativeArrayMarshaler in ilmarshalers.h
+        internal struct MarshalerState
+        {
+            IntPtr  m_pElementMT;
```
danmosemsft: nit:extra space

src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs
```diff
@@ -576,6 +577,41 @@ public static long ReadInt64(IntPtr ptr)
...
+        // It's quite slow and can return back dangling pointers
+        // It's only there for backcompact
+        // I don't think we should spend time optimizing it
+        // People should really call the IntPtr overload instead
+        //====================================================================
+        static unsafe T ReadValueSlow<T>(object ptr, int ofs, Func<IntPtr, int, T> readValueHelper)
+        {
+            // We AV on desktop if passing NULL. So this is technically a breaking change but is an improvement
+            if (ptr == null)
+                throw new ArgumentNullException("ptr");
```
danmosemsft: nit: `throw new ArgumentNullException(nameof(ptr));`

src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs
```diff
@@ -408,7 +409,7 @@ public static void Copy(IntPtr source, IntPtr[] destination, int startIndex, int
         //====================================================================
         public static byte ReadByte([MarshalAs(UnmanagedType.AsAny), In] Object ptr, int ofs)
         {
-            throw new PlatformNotSupportedException(); // https://github.com/dotnet/coreclr/issues/10442
+            return ReadValueSlow(ptr, ofs, (IntPtr nativeHome, int offset) => { return Marshal.ReadByte(nativeHome, offset); });
```
justinvp: Nit: Could be simplified as:

```c#
return ReadValueSlow(ptr, ofs, (nativeHome, offset) => Marshal.ReadByte(nativeHome, offset));
```

src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs
```diff
@@ -576,6 +577,41 @@ public static long ReadInt64(IntPtr ptr)
...
         }
 
+        //====================================================================
+        // Read value from marshaled object (marshaled using AsAny)
+        // It's quite slow and can return back dangling pointers
+        // It's only there for backcompact
+        // I don't think we should spend time optimizing it
+        // People should really call the IntPtr overload instead
+        //====================================================================
+        static unsafe T ReadValueSlow<T>(object ptr, int ofs, Func<IntPtr, int, T> readValueHelper)
```
justinvp: Nit: Specify visibility, e.g. `private static unsafe`

tests/src/Interop/MarshalAPI/ReadWrite/ReadWriteObject.cs
yizhang82: I don't think we have a policy in place.  In general for the more 'runtime'-y features such as interop it makes sense to put the tests in CoreCLR. Marshal APIs are sort like a gray area that can land either way. My preference is to put them in CoreCLR tests unless it's super clear they belong in CoreFX (such as in System.Runtime.InteropServices.dll, etc). Moving forward, I think we'd want CoreCLR tests to run on multiple runtime configuration as well (they are runtime tests that applies everywhere, after all)

src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs
yizhang82: That's better. Thanks

src/mscorlib/src/System/StubHelpers.cs
yizhang82: Fixed.

src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs
yizhang82: Thanks. I agree this looks simpler but given that Marshal.ReadByte has different overloads that takes object/IntPtr I'd like to make this a bit more explicit.

src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs
yizhang82: fixed.

src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs
justinvp: You can keep the parameter types and still remove the unnecessary return statement, e.g.

```c#
return ReadValueSlow(ptr, ofs, (IntPtr nativeHome, int offset) => Marshal.ReadByte(nativeHome, offset));
```

yizhang82: OK. That makes sense. changed

PR #11252, opened 2017-04-27
Enable interpreter build

parjong:
#11188 allows users to enable interpreter feature via command-line, but build fails when this feature is enabled due to the following two reasons:
 - FEATURE_INTERPRETER should NOT be defined for crossgen build, but #11188 enables it even for crossgen build
 - FEATURE_TIERED_COMPILATION (recently added) has an implementation incompatible with FEATURE_INTERPRETER

This commit fixes interpreter build error to properly enable interpreter feature with cmakeargs option.


src/inc/switches.h
```diff
@@ -233,5 +233,8 @@
 #define FEATURE_STACK_SAMPLING
 #endif // defined (ALLOW_SXS_JIT)
 
+#if defined(USE_INTERPRETER)
+#define FEATURE_INTERPRETER
+#endif // USE_INTERPRETER
```
BruceForstall: I wonder if, instead of adding another name (USE_INTERPRETER), it would be better to add, here:
```
#if defined(CROSSGEN_COMPILE)
#if defined(FEATURE_INTERPRETER)
#undef FEATURE_INTERPRETER // No interpreter in crossgen builds
#endif
#endif
```
?

parjong: Thank you for comment! It works well. I updated PR as you suggested.

PR #11253, opened 2017-04-27
Use relevant define for resumable frame SP handling (and update ifdef condition)

parjong:
m_pvResumableFrameTargetSP-related implementations in the stack walker
are explicitly ifdefed by _TARGET_XXX_.

This commit introduces relevant define (RECORD_RESUMABLE_FRAME_SP) and
uses it to ifdef relevant implementations.

This commit does not introduce any behavior changes.

### DIFF
--- a/src/vm/stackwalk.h
+++ b/src/vm/stackwalk.h
@@ -45,7 +45,7 @@ class AppDomain;
  #endif 
 #endif // _TARGET_X86_ && !FEATURE_PAL
 
-#if !defined(_TARGET_X86_)
+#if defined(WIN64EXCEPTIONS)
 #define RECORD_RESUMABLE_FRAME_SP
 #endif
 
--- a/src/vm/stackwalk.h
+++ b/src/vm/stackwalk.h
@@ -45,6 +45,10 @@ class AppDomain;
  #endif 
 #endif // _TARGET_X86_ && !FEATURE_PAL
 
+#if !defined(_TARGET_X86_)
+#define RECORD_RESUMABLE_FRAME_SP
+#endif
+
 //************************************************************************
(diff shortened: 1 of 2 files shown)


parjong:
@jkotas @janvorli Could you please take a look? I'm not sure whether x86/Linux needs these implementations, or not.

src/vm/stackwalk.h
```diff
@@ -45,6 +45,10 @@ class AppDomain;
  #endif 
 #endif // _TARGET_X86_ && !FEATURE_PAL
 
+#if !defined(_TARGET_X86_)
+#define RECORD_RESUMABLE_FRAME_SP
```
janvorli: @parjong have you tried if enabling it for x86 Linux too would work? It is used just as a safeguard and it seems to me that the reason why it was excluded for x86 is the Windows x86 EH and that it can be applied to x86 Linux.

parjong: I tried, and there is no notable difference. If this code is related with Funclet-based EH, I will revise this line as as another PR  (I would like not to make behavior change here).

janvorli: I believe it is meant for WIN64EXCEPTIONS.

parjong: Yes, I plan to apply the following patch:
```cpp
#if defined(WIN64EXCEPTIONS)
#define RECORD_RESUMABLE_FRAME_SP
#endif
```

parjong:
@janvorli Could you please take a look? Should I  update PR to include the changes that you suggested?

janvorli:
@parjong it would be nice to update it with the changes I've suggested. Otherwise LGTM.

PR #11254, opened 2017-04-27
[ARM/Linux] Enable build with FEATURE_DBGIPC=0

parjong:
#11164 allows users to disable DBGIPC support, but it introduces several build errors on ARM. Android has a similar issue, and thus #9961 introduces a workaround for Android.

This commit ports this workaround to Unix platforms.

### DIFF
--- a/src/debug/di/cordb.cpp
+++ b/src/debug/di/cordb.cpp
@@ -20,6 +20,13 @@
 #include "dbgtransportmanager.h"
 #endif // FEATURE_DBGIPC_TRANSPORT_DI
 
+#if defined(PLATFORM_UNIX) || defined(__ANDROID__)
+// _aligned_malloc and _aligned_free are unavailable on POSIX and Android
+#define HAVE_ALIGNED_MALLOC 0
+#else
+#define HAVE_ALIGNED_MALLOC 1
+#endif
+
 //********** Globals. *********************************************************
 #ifndef FEATURE_PAL
 HINSTANCE       g_hInst;                // Instance handle to this piece of code.
@@ -499,7 +506,7 @@ DbiGetThreadContext(HANDLE hThread,
     DT_CONTEXT *lpContext)
 {
     // if we aren't local debugging this isn't going to work
-#if !defined(_ARM_) || defined(FEATURE_DBGIPC_TRANSPORT_DI) || defined(__ANDROID__)
+#if !defined(_ARM_) || defined(FEATURE_DBGIPC_TRANSPORT_DI) || !HAVE_ALIGNED_MALLOC
     _ASSERTE(!"Can't use local GetThreadContext remotely, this needed to go to datatarget");
     return FALSE;
 #else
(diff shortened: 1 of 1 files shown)


parjong:
\CC @jyoungyun @chunseoklee @lemmaa

parjong:
@janvorli @jkotas Please take a look

src/debug/di/cordb.cpp
```diff
@@ -499,7 +506,7 @@ DbiGetThreadContext(HANDLE hThread,
     DT_CONTEXT *lpContext)
 {
     // if we aren't local debugging this isn't going to work
-#if !defined(_ARM_) || defined(FEATURE_DBGIPC_TRANSPORT_DI) || defined(__ANDROID__)
+#if !defined(_ARM_) || defined(FEATURE_DBGIPC_TRANSPORT_DI) || !HAVE_ALIGNED_MALLOC
```
janvorli: @mikem8361 do you think that we would ever want to enable build for Unix with FEATURE_DBGIPC_TRANSPORT_DI not defined as something that would functionally make sense and not just as a mean to disable debugging on platforms that cannot support the IPC transport?

mikem8361: Not sure. All I know is that when the unix transport support was added, it was really only not enabled for Windows.  For the platforms that can't support it or it just doesn't make sense (like on a phone), it is ok.  As a way to disable this transport for platforms that already support it, I don't have an opinion.

parjong: @janvorli Could you please share your opinion on this PR? We couldn't find any regressions during internal evaluation. Is this related with SOS?

janvorli: @parjong I guess I might have confused you. The reason why I was asking about it is that if we thought that we could have a scenario where the debugging would use a different transport, then I would rather fix the code below to not to use `_aligned_malloc` - the alignment requirement is small, so we can use regular malloc, allocate buffer larger by 15 bytes and place the context to an aligned address in this buffer. 
But if we think we will not need it, then ifdefining it out for all non-windows platforms is better. 
My thinking stemmed from your change that defined the HAVE_ALIGNED_MALLOC and my feeling was that the missing aligned malloc is not the actual reason why we ifdef out the code, since that could be easily fixed. The real reason is that there is no point in having the code compiled on Unix or Android. 
So based on these thoughts, I would prefer adding `|| defined(PLATFORM_UNIX)` to the condition instead of introducing the HAVE_ALIGNED_MALLOC.

parjong: I'm not sure, but it seems that else branch in these functions is for remote debugging. Am I right? If so, how about using ``SUPPORT_REMOTE_DEBUGGING`` instead of ``HAVE_ALIGNED_MALLOC``?

janvorli: @parjong On the contrary, the code in the else branch is for local debugging when the debugger code is running in the debugged process. So you could use a define named SUPPORT_LOCAL_DEBUGGING instead.

parjong:

PR #11257, opened 2017-04-27
Turn on unwind information at informational level rather than Info 

vancem:
This is basically a bug fix.   Quite a while ago we wanted to be able to collect stacks 
(which needed JIT names, and unwind information) without collecting all verbose CLR events 
(in particular the GC allocation Tick events).   We did this in most places but missed this one.

This change makes it so that setting event the Informational (rather than verbose) level will 
cause unwind information to be generated (needed for stack crawling on 64 bit machines).


vancem:
@brianrob @mjsabby

PR #11266, opened 2017-04-27
Array.Copy fixes

### LINKED ISSUE #10646: Array.Copy doesn't handle pointer-objref conversions
This should probably throw a `ArrayTypeMismatchException` like e.g. the `IntPtr[]` and `int*[]` case does, but it does something horrible instead:

```csharp
using System;

unsafe class Program
{
    static int Main()
    {
        int*[] inarr = new int*[] { (int*)1, (int*)2 };
        object[] oarr = new object[inarr.Length];
        Array.Copy(inarr, oarr, inarr.Length);
        Array.Copy(oarr, inarr, inarr.Length);
        return 100;
    }
}
```

Neither of these `Array.Copy` should succeed.

jkotas:
Fix #10646 and https://github.com/dotnet/corefx/issues/13816

### DIFF
--- a/src/classlibnative/bcltype/arraynative.cpp
+++ b/src/classlibnative/bcltype/arraynative.cpp
@@ -319,14 +319,16 @@ ArrayNative::AssignArrayEnum ArrayNative::CanAssignArrayTypeNoGC(const BASEARRAY
             return AssignDontKnow;
     }
     
-    const CorElementType srcElType = srcTH.GetSignatureCorElementType();
-    const CorElementType destElType = destTH.GetSignatureCorElementType();
+    const CorElementType srcElType = srcTH.GetVerifierCorElementType();
+    const CorElementType destElType = destTH.GetVerifierCorElementType();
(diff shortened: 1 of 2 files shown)


jkotas:
corefx test updates: https://github.com/dotnet/corefx/pull/19063

src/classlibnative/bcltype/arraynative.cpp
```diff
@@ -435,10 +435,6 @@ ArrayNative::AssignArrayEnum ArrayNative::CanAssignArrayType(const BASEARRAYREF
     if (srcTH.IsInterface() && destElType != ELEMENT_TYPE_VALUETYPE)
         return AssignMustCast;
 
-    // Enum is stored as a primitive of type dest.
```
gkhanna79: Should we call these out for breaking change?

jkotas: I do not think that these are breaking changes.

tarekgh:
LGTM. thanks for fixing it.

Issue #11223, opened 2017-04-26
[x86/Linux] JIT.Methodical.Boxing.seh._relfilter fail (GCStress=3)


wateret:
With GCStress=3 JIT.Methodical.Boxing.seh._relfilter fails on x86/Linux.

```
OK
Current totals 0 and 0 and 0
OK
Current totals 0 and 0.314159265358979 and 0.309016994374947
OK
Current totals 0 and 0.942477796076938 and 0.896802246667421
OK
Current totals 0 and 1.88495559215388 and 1.70581924104237
Current totals 0.0201 and 3.14159265358979 and 1.70581924104237
Filtering out the exception...

Assert failure(PID 1 [0x00000001], Thread: 1 [0x0001]): !CREATE_CHECK_STRING(pMT && pMT->Validate())
    File: /home/hanjoung/ws/dotnet/coreclr/src/vm/object.cpp Line: 1727
    Image: /work/nfs/dotnet/ttt/corerun
```

This is almost same issue with #11101(_dbgfilter) which was resolved  by #11195 (not reporting untracked from filter). However this time the most variables are encoded as tracked so we need to handle untracked.

As we discussed in #11101, We may need something like `GCInfo::gcMarkFilterVarsPinned()`.

wateret:
@janvorli Here's my first attempt to do `gcMarkFilterVarsPinned()` on x86/Linux.
https://github.com/wateret/coreclr/commit/dc2ec633820e5d992c44e2ea5a2c597245049bab

It basically uses the same `gcMarkFilterVarsPinned()` of non-JIT32_GCENCODER. All I had to do more was to reorder the generated varDscPtrs by lifetime splitting(in vpdBegOfs order). However the problem was pinned_OFFSET_FLAG is only for untracked. For tracked, the same bit is used as this_OFFSET_FLAG.

janvorli:
@wateret it is great that we can reuse the code almost as-is. 
> However the problem was pinned_OFFSET_FLAG is only for untracked. For tracked, the same bit is used as this_OFFSET_FLAG.

I am not sure I understand this. You say the problem "was", but I don't see any changes in the code that would do anything about it. 

And could you also explain why you had to reorder the  generated varDscPtrs by lifetime splitting? It is because for the x86 gc info they have to be ordered by the vpdBegOfs? 
[...]

PR #11281, opened 2017-04-28
[x86/Linux] Enable gcMarkFilterVarsPinned (WIN64EXCEPTIONS)

wateret:
cc: @janvorli 

Updated as the comments from #11223, but we still have an issue https://github.com/dotnet/coreclr/issues/11223#issuecomment-297604033 .


wateret:
For https://github.com/dotnet/coreclr/issues/11223#issuecomment-297604033, could we use lowBits=0x3(both "this" and "byref" flag on) as pinned objects? If there is no such case(a reference is "this" object and it is "byref" at the same time), maybe we can use 0x3 as pinned. Of course, in we should make a little change to EnumGcRef, setting flags for GCEnumCallback.

wateret:

janvorli:
@wateret `this` can be byref - for structs, it is the case. But I wonder how it works for e.g. amd64. The "this" and "pinned" bits are also shared there, aren't they?

wateret:
@janvorli It seems like non-x86 GCInfo(using GCInfoEncoder) has completely different structure from x86 GCInfo and it can handle this situation. Looks like non-x86 GCInfo has a flag `GcSlotFlags`.

```c
enum GcSlotFlags
{
    GC_SLOT_BASE      = 0x0,
    GC_SLOT_INTERIOR  = 0x1,
    GC_SLOT_PINNED    = 0x2,
    GC_SLOT_UNTRACKED = 0x4,

    // For internal use by the encoder/decoder
    GC_SLOT_IS_REGISTER = 0x8,
    GC_SLOT_IS_DELETED  = 0x10,
};
```

Sorry for the late response by the way. I was on vacation.

wateret:
@janvorli non-x86 GCInfo uses second bit as pinned(not this). It looks like it has some other way to verify an object is `this` or not.

janvorli:
@wateret it rather looks like the GCInfoEncoder doesn't handle "this" pointer in any way. And looking at how the this_OFFSET_FLAG is used, I can see only a single place where this flag is used for something else than debug logging:
https://github.com/dotnet/coreclr/blob/master/src/vm/eetwain.cpp#L5338.
This is used to get "this" pointer given a stack frame. Basically, it just scans all tracked slots until it finds one that contains "this" pointer based on the `this_OFFSET_FLAG` flag. The comments say that "this" is never untracked.
The GCInfoEncoder uses a different way. Based on what I can see in the code, JIT on these platforms always stores "this" in the current frame as an untracked slot. So the gcinfo contains an offset of this slot and so the "this" can be extracted from it without any search. See `EECodeManager::GetExactGenericsToken`. It actually contains a generics token or "this", which is used as generic token for member methods. That's why it doesn't need to mark slots as containing "this".
So I wonder if we could switch to this model for x86 Linux too. That would mean adding such offset to the gcinfo and dropping the this_OFFSET_FLAG usage. That would solve our problem, since we would then be able to mark both tracked or untracked slots as pinned.
That would mean updating JIT to always home "this", make its slot untracked and store its offset to the gc info. It actually looks like we might be able to reuse the EECodeManager::GetParamTypeArg with possibly some modifications for getting the slot. As you can see from the code of the method, it is equivalent to EECodeManager::GetExactGenericsToken on archs with the GCInfoEncoder.

wateret:
@janvorli Please take a look. I added a commit that forces this pointer to untracked and the `GetInstance()` now assumes that the first entry of untracked variables. I am not sure about this commit. But I ran unit test and there is no regression(pri0, gcstress=3) and this makes two more tests below pass.

-  JIT.Methodical.Boxing.seh._relfilter._relfilter
-  JIT.Methodical.eh.basics.trythrowexcept_r.trythrowexcept_r

janvorli:
@BruceForstall can you take a look at the JIT changes, please?

BruceForstall:
This conversation probably happened a while ago, but have we considered having Linux/x86 use GCInfoEncoder/Decoder? That is, use the `!JIT32_GCENCODER` path in the JIT? I'm not sure what the impact would be (how hard it would be to get it to work, or what the delta in GC info size would be), but perhaps that would be an option. Certainly, all platforms that use GCInfoEncoder also use funclet model. GCInfoEncoder has been used for arm32, arm64, amd64, and ia-64, so it works with different "bitness". It's easier to work with, from an API perspective. If we never had to touch the JIT32_GCENCODER path, that would make me more comfortable, too.

parjong:
@BruceForstall The current x86/Linux unwinder is based on ``JIT32_GCENCODER``. We need to revise all the EH-related implementation to enable GCInfoEncode/Decoder (including JIT changes to emit proper unwind info). I guess that the required amount of work would be almost similar to that of enabling funclet-based EH (``WIN64EXCEPTION`` in EE, and ``FEATURE_EH_FUNCLET`` in JIT).

src/vm/eetwain.cpp
```diff
@@ -5356,6 +5371,18 @@ OBJECTREF EECodeManager::GetInstance( PREGDISPLAY    pContext,
     _ASSERTE(*castto(table, unsigned short *) == 0xBABE);
 #endif
 
+#else // WIN64EXCEPTIONS
+
+    // Untracked table must have at least one entry - this pointer
+    _ASSERTE(info.untrackedCnt > 0);
```
wateret: @janvorli I checked all the call sites to `GetInstance()` - from `StackFrameIterator::NextRaw(void)` and `CrawlFrame::GetThisPointer()`. These call sites check if the function is static so it looks OK to assume that we always have "this" in the function. And I think the changes in lclvars.cpp ensures "this" pointer. "this" seems to be always lclNum=0 so it goes to the first entry.

wateret:
@BruceForstall I agree with @parjong, IMHO It would not be small amount of work.

BruceForstall:
@parjong Ok, I had forgotten that the x86 unwinder uses the x86 GC info to provide unwinding information, since there are no platform-defined unwind codes.

BruceForstall:
> 

Couldn't you just put the entire clause under `#ifdef`, e.g.:
```
#ifndef WIN64EXCEPTIONS
            // For WIN64EXCEPTIONS, "this" must always be in untracked variables
            // so we cannot have "this" in variable lifetimes.

            if (compiler->lvaIsOriginalThisArg(varNum) && compiler->lvaKeepAliveAndReportThis())
            {
                // Encoding of untracked variables does not support reporting
                // "this". So report it as a tracked variable with a liveness
                // extending over the entire method.

                thisKeptAliveIsInUntracked = true;
                continue;
            }
#endif // !WIN64EXCEPTIONS
```

---
Refers to: src/jit/gcinfo.cpp:494 in af203f3. [](commit_id = af203f3c0698bb84c8c8616a244a3474dea6ee3b, deletion_comment = False)

src/jit/gcencode.cpp
```diff
@@ -106,6 +106,293 @@ ReturnKind GCInfo::getReturnKind()
...
+#endif // DEBUG
+                    }
+                }
+            }
+        } // HasFilter
+    }     // Foreach EH
+}
+
+void GCInfo::gcInsertVarPtrDscSplit(varPtrDsc* desc, varPtrDsc* begin)
+{
```
BruceForstall: Please add a function header comment describing what this does and why. In particular, why do we need to sort the insertions for `JIT32_GCENCODER` but not for the other case? Could this function be made to work for both (e.g., `#ifdef` to sort for `JIT32_GCENCODER` and not worry about sorting for the opposite?) That would reduce `#ifdefs` in `GCInfo::gcMarkFilterVarsPinned()`.

src/jit/gcencode.cpp
```diff
@@ -1972,7 +2259,14 @@ size_t GCInfo::gcMakeRegPtrTable(BYTE* dest, int mask, const InfoHdr& header, un
...
                 }
 
+#ifndef WIN64EXCEPTIONS
                 if (compiler->lvaIsOriginalThisArg(varNum) && compiler->lvaKeepAliveAndReportThis())
+#else
+                // For WIN64EXCEPTIONS, "this" must always be in untracked variables
+                // so we cannot have "this" in variable lifetimes
+                if (false)
+#endif
+
```
BruceForstall: Same comment applies here that I wrote for GCInfo.cpp.

src/jit/lclvars.cpp
```diff
@@ -3292,6 +3292,13 @@ void Compiler::lvaSortByRefCount()
         {
             varDsc->lvTracked = 0;
         }
+#if defined(JIT32_GCENCODER) && defined(WIN64EXCEPTIONS)
+        else if (lvaIsOriginalThisArg(lclNum))
+        {
+            // For x86/Linux, we cannot have this pointer in tracked variables, so we set this pointer untracked
+            varDsc->lvTracked = 0;
```
BruceForstall: Does it need to be untracked in all cases? What if there is no EH and/or no generics context? In that case, the VM doesn't need to look up `this` in the GC info, so we shouldn't need to treat is specially.

src/vm/eetwain.cpp
```diff
@@ -4816,8 +4816,22 @@ bool EECodeManager::EnumGcRefs( PREGDISPLAY     pContext,
                         printf("%cP+%02XH]: ", frameType, +dspOffs);
                 }
 #endif
+
+                unsigned flags = CHECK_APP_DOMAIN;
+#ifndef WIN64EXCEPTIONS
```
BruceForstall: Do you need to change the code above that references this_OFFSET_FLAG? (since that is no longer used?)

BruceForstall:
Is there a succinct way to summarize what the contract between the JIT and VM is with this change? It seems, for Linux/x86:
1. `this` pointer must be untracked.
2. `this` must be the first untracked variable.

It appears you have changed the meaning of the bits as defined by gcinfo.h:
```
//  Note for untracked locals the flags allowed are "pinned" and "byref"
//   and for tracked locals the flags allowed are "this" and "byref"
//  Note that these definitions should also match the definitions of
//   GC_CALL_INTERIOR and GC_CALL_PINNED in VM/gc.h
//
const unsigned  byref_OFFSET_FLAG  = 0x1;  // the offset is an interior ptr
const unsigned pinned_OFFSET_FLAG  = 0x2;  // the offset is a pinned ptr
const unsigned   this_OFFSET_FLAG  = 0x2;  // the offset is "this"
```
In particular, `this_OFFSET_FLAG` is no longer used. Should you put it under `#ifdef`? e.g.,
```
#if !defined(WIN64EXCEPTIONS)
const unsigned   this_OFFSET_FLAG  = 0x2;  // the offset is "this"
#endif
```
Do the conditions above apply to all methods? Or just methods with EH filters or methods with `this` as generics context?

Hopefully we are capturing all the Linux/x86 ABI changes in `Documentation\botr\clr-abi.md`.

wateret:
> Do the conditions above apply to all methods? Or just methods with EH filters or methods with this as generics context?

@BruceForstall I think "just methods with EH filters or methods with this as generics context" is the case. I'll update the PR. Thank you!

src/jit/lclvars.cpp
wateret: @BruceForstall I agree. As I mentioned in https://github.com/dotnet/coreclr/pull/11281#issuecomment-302983448 I'll update it.

src/vm/eetwain.cpp
wateret: @BruceForstall I think we should keep this change here for the sake of !WIN64EXCEPTION(i.e. x86/Windows). Since we want to change x86/Linux only.

src/jit/gcencode.cpp
wateret: @BruceForstall 

> why do we need to sort the insertions for JIT32_GCENCODER but not for the other case?

To summarize,
- x86/Windows : We don't need `gcMarkFilterVarsPinned` and `gcInsertVarPtrDscSplit`.
- x86/Linux : We do need `gcMarkFilterVarsPinned` but we have **x86-GCInfo**. It requires varPtrList to be sorted by vpdBegOfs ascending. That's why we need `gcInsertVarPtrDscSplit` for only this case.
- the others : We do need `gcMarkFilterVarsPinned` and **GCInfoEncoder** does not require varPtrList is sorted.

> Could this function be made to work for both (e.g., #ifdef to sort for JIT32_GCENCODER and not worry about sorting for the opposite?) That would reduce #ifdefs in GCInfo::gcMarkFilterVarsPinned().

Thank you for pointing that out. When I first wrote this patch, I just did not want to change the current behavior just in case. I'll try what you said then apply it to the patch.

src/jit/lclvars.cpp
wateret: @BruceForstall I was going to change the condition to:

```cpp
        else if (lvaIsOriginalThisArg(lclNum) && lvaKeepAliveAndReportThis())
```

However we need something like `lvaKeepAliveAndReportThis()` from VM. Could I ask how to get a method info from vm(`GetInstance():eetwain.cpp`)? 
Otherwise we may need to add a flag in GCInfo header.

BruceForstall: Isn't this just part of the contract between JIT and VM? That is, JIT only reports `this` in the GC information in exactly the cases that VM requires? If `lvaKeepAliveAndReportThis()` captures that condition, then it seems like that would be sufficient.

Can you explain why you think the JIT needs to ask a VM a question, or why we need a GCInfo bit? (E.g., what would the bit say?)

---
In reply to: [117710335](https://github.com/dotnet/coreclr/pull/11281#discussion_r117710335) [](ancestors = 117710335)

wateret:

wateret:
@BruceForstall I updated the commit as your comment. Please take a look.

wateret:
@BruceForstall Could you take a look please?

BruceForstall:
@janvorli Do you want to take another look?

wateret:
@BruceForstall Yes it is ready to merge. Thank you!

PR #11294, opened 2017-04-28
Correctly enumerate loaded assemblies in AppDomain

gkhanna79:
Use the correct function to enumerate loaded assemblies in AppDomain.

Fixes https://github.com/dotnet/corefx/issues/18877

@jkotas PTAL


src/mscorlib/src/System/AppDomain.cs
```diff
@@ -646,6 +646,15 @@ public override String ToString()
             return StringBuilderCache.GetStringAndRelease(sb);
         }
 
+        [MethodImpl(MethodImplOptions.InternalCall)]
+        [SuppressUnmanagedCodeSecurity]
```
jkotas: SuppressUnmanagedCodeSecurity is not necessary for InternalCall.

gkhanna79: Fixed.

jkotas:
LGTM otherwise

PR #11304, opened 2017-04-29
Fix #11273 by adjusting the fix for VSO 406163.

### LINKED ISSUE #11273: Assert failure(PID 5480 [0x00001568], Thread: 10988 [0x2aec]): Assertion failed 'handlerGetsXcptnObj(compiler->compCurBB->bbCatchTyp)'
https://ci.dot.net/job/dotnet_coreclr/job/master/job/jitstress/job/x64_checked_windows_nt_minopts/156/

```
Assert failure(PID 5480 [0x00001568], Thread: 10988 [0x2aec]): Assertion failed 'handlerGetsXcptnObj(compiler->compCurBB->bbCatchTyp)' in 'C:Test3b(ref,int):int' (IL size 45)
20:52:49         
20:52:49             File: d:\j\workspace\x64_checked_w---a16971a7\src\jit\codegenxarch.cpp Line: 2135
20:52:49             Image: D:\j\workspace\x64_checked_w---a16971a7\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\CoreRun.exe
20:52:49         
20
```

pgavlin:
Issue #11273 revealed an issue with the fix for VSO 406163: as of today,
the compiler models the incoming exception object in a handler by
looking for `GT_CATCH_ARG` in the first block of a handler and if found
marking the register that contains the exception object as holding a GC
pointer. Unfortunately, this breaks down under the transform that the
original fix (#11134) was performing for single-block filters. That
transformation was from this:

```
    filter start: BBN (BBJ_EHFILTERRET)
```

to this:

```
    filter start: BBN + 1 (BBJ_ALWAYS -> BBN)
                  BBN + 2 (jumpKind)
                  BBN     (BBJ_EHFILTERRET)
```

After this transform, `BBN` still contains the `GT_CATCH_ARG`, but the
register containing the exception object must be live through `BBN + 1`
as well as live in to `BBN`.

There are a number of possible solutions to this problem (e.g. modeling
the liveness of the exception object in some fashion or splitting the
single-block filter either after the catch arg or before the filter
ret), but many of them are complicated by the fact that this
transformation needs to operate on both HIR and LIR and must be be able
to run in a variety of locations, including after registers have been
allocated. Instead, this change takes the approach of ensuring that no
single-block filter regions exist on platforms that target the JIT32 GC
encoder by inserting a spill of the exception object to a local var in
its own block at the start of a filter.

### DIFF
--- a/src/jit/compiler.h
+++ b/src/jit/compiler.h
@@ -3187,7 +3187,7 @@ private:
     static fgWalkPreFn impFindValueClasses;
     void impSpillLclRefs(ssize_t lclNum);
 
-    BasicBlock* impPushCatchArgOnStack(BasicBlock* hndBlk, CORINFO_CLASS_HANDLE clsHnd);
+    BasicBlock* impPushCatchArgOnStack(BasicBlock* hndBlk, CORINFO_CLASS_HANDLE clsHnd, bool isSingleBlockFilter);
 
     void impImportBlockCode(BasicBlock* block);
 
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -17055,8 +17055,8 @@ bool Compiler::fgCheckEHCanInsertAfterBlock(BasicBlock* blk, unsigned regionInde
 //
 // Return Value:
 //    A block with the desired characteristics, so the new block will be inserted after this one.
-//    If there is no suitable location, return nullptr. This should basically never happen except in the case of
-//    single-block filters.
+//    If there is no suitable location, return nullptr. This should basically never happen.
+//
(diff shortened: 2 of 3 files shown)


pgavlin:

BruceForstall:
Any diffs?

BruceForstall:
You need to update:
```
// Return Value:
//    A block with the desired characteristics, so the new block will be inserted after this one.
//    If there is no suitable location, return nullptr. This should basically never happen except in the case of
//    single-block filters.
```

pgavlin:
No SPMI diffs. I will run diffs with jit-diff on Monday.

PR #11317, opened 2017-05-01
Fix current publish build warnings

dagood:
Cleans up publish build warnings.

---

CoreCLR publishes two Drops, one named by OfficialBuildId, another by BuildNumber. Having both makes it so you don't need to translate from one to another to find a CoreCLR build's Drop.

When the `dropMetadataContainerName` is the same, the later step fails. (This was only a warning because the step is "continue on error".) The metadata name is used to create a VSTS Artifact. I changed the metadata name based on what the Drop is named after.

---

Remove "Copy Publish Artifact: PublishLogs" because there are no *.log files being created during the publish build. There is only a `init-tools.cmd` call and direct `msbuild` invocations.

### DIFF
--- a/buildpipeline/DotNet-Trusted-Publish.json
+++ b/buildpipeline/DotNet-Trusted-Publish.json
@@ -305,25 +305,6 @@
     },
     {
       "enabled": true,
-      "continueOnError": false,
-      "alwaysRun": true,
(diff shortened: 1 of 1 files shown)


dagood:
Thanks! Merging now without CI complete because CI doesn't use the pipebuild files.

PR #11325, opened 2017-05-01
Adding missing icu headers for Android

cydhaselton:
urename.h and ustring.h contain the u_stncpy and its definition; if they are not in the list of icushim.h headers, Android build fails after moving to icu 59.1 from 58.2.

cc @jkotas, @richlander  to locate icu and/or System.Globalization.Native code owners

### DIFF
--- a/src/corefx/System.Globalization.Native/icushim.h
+++ b/src/corefx/System.Globalization.Native/icushim.h
@@ -32,6 +32,8 @@
 #include <unicode/usearch.h>
 #include <unicode/utf16.h>
 #include <unicode/utypes.h>
+#include <unicode/urename.h>
+#include <unicode/ustring.h>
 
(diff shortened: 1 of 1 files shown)


tarekgh:

tarekgh:
@cydhaselton just double checking, does urename.h and ustring.h exist in the older ICU versions too? I am asking to make sure will not break the builds when having older versions

tarekgh:
LGTM

cydhaselton:
@jkotas Would it be appropriate to add @janvorli to CODE_OWNERS.txt for corefx-related changes?

cydhaselton:
@tarekgh It is, at least in 58.2 (https://fossies.org/diffs/icu4c/58_2-src_vs_59rc-src/).

janvorli:
@cydhaselton @tarekgh I've verified that the headers are in the minimum supported version 50.1 too, so this change is good.

PR #11329, opened 2017-05-02
Refactor arm codegen

BruceForstall:
For arm32/arm64, move a lot of code from genCodeForTreeNode() into separate functions. A few functions are identical or close to it, and are moved to codegenarmarch.cpp. Many are relatively similar and should be considered for merging to codegenarmarch.cpp at some point.

codegenxarch.cpp could at some point improve readability by using the same functions, and moving code out of genCodeForTreeNode().

### DIFF
--- a/src/jit/codegenarm.cpp
+++ b/src/jit/codegenarm.cpp
@@ -647,25 +647,8 @@ void CodeGen::genCodeForTreeNode(GenTreePtr treeNode)
         break;
 
         case GT_RETURNTRAP:
-        {
-            // this is nothing but a conditional call to CORINFO_HELP_STOP_FOR_GC
-            // based on the contents of 'data'
-
-            GenTree* data = treeNode->gtOp.gtOp1->gtEffectiveVal();
-            genConsumeIfReg(data);
-            GenTreeIntCon cns = intForm(TYP_INT, 0);
-            emit->emitInsBinary(INS_cmp, emitTypeSize(TYP_INT), data, &cns);
-
-            BasicBlock* skipLabel = genCreateTempLabel();
-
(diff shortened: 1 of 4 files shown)


BruceForstall:

BruceForstall:
@dotnet/arm64-contrib @dotnet/arm32-contrib PTAL
cc @dotnet/jit-contrib

BruceForstall:

src/jit/codegenarm.cpp
```diff
@@ -482,142 +430,20 @@ void CodeGen::genCodeForTreeNode(GenTreePtr treeNode)
...
-                }
-            }
-            else
-            {
-                emit->emitIns_R_S(ins_Move_Extend(targetType, treeNode->InReg()), size, targetReg, varNum, offs);
-            }
-        }
-            genProduceReg(treeNode);
+        case GT_LCL_VAR:
+            genCodeForLclVar(treeNode->AsLclVar());
```
CarolEidt: I'm just a little curious why you moved the `GT_LCL_VAR` case down here - was it for consistency across architectures?

CarolEidt:
:shipit:

CarolEidt:
LGTM, though I only did a rough job of matching the code that was moved - I am trusting that it was a simple cut-and-paste.

src/jit/codegenarm.cpp
BruceForstall: Yes, arm32/arm64 had things in different orders. I'd like to keep making things more similar until perhaps they can share genCodeForTreeNode() itself.

BruceForstall:
> I am trusting that it was a simple cut-and-paste.

Essentially. I did use more specific types instead of GenTree* so that changed a few things (no need for `gtOp.` for example).

PR #11333, opened 2017-05-02
Enable LoadFrom to pickup assemblies next to requesting assembly

gkhanna79:
@jkotas PTAL

### DIFF
--- a/src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
+++ b/src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
@@ -23,37 +23,35 @@ namespace System.Reflection
         private static Assembly LoadFromResolveHandler(object sender, ResolveEventArgs args)
         {
             Assembly requestingAssembly = args.RequestingAssembly;
-            Assembly requestedAssembly = null;
-
+            
             // Requesting assembly for LoadFrom is always loaded in defaultContext - proceed only if that
             // is the case.
-            if (AssemblyLoadContext.Default == AssemblyLoadContext.GetLoadContext(requestingAssembly))
+            if (AssemblyLoadContext.Default != AssemblyLoadContext.GetLoadContext(requestingAssembly))
+                return null;
+
+            // Get the path where requesting assembly lives and check if it is in the list
+            // of assemblies for which LoadFrom was invoked.
+            bool fRequestorLoadedViaLoadFrom = false;
+            string requestorPath = Path.GetFullPath(requestingAssembly.Location);
+            if (string.IsNullOrEmpty(requestorPath))
+                return null;
+
+            lock(s_syncLoadFromAssemblyList)
             {
-                // Get the path where requesting assembly lives and check if it is in the list
-                // of assemblies for which LoadFrom was invoked.
-                bool fRequestorLoadedViaLoadFrom = false;
-                string requestorPath = Path.GetFullPath(requestingAssembly.Location);
-                if (!string.IsNullOrEmpty(requestorPath))
-                {
-                    lock(s_syncLoadFromAssemblyList)
-                    {
-                        fRequestorLoadedViaLoadFrom = s_LoadFromAssemblyList.Contains(requestorPath);
-                    }
-                }
-
-                // Requestor assembly was loaded using loadFrom, so look for its dependencies
-                // in the same folder as it.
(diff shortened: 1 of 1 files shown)


src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
```diff
@@ -15,11 +15,76 @@ namespace System.Reflection
...
+        private static object s_syncLoadFromAssemblyList = new object();
+
+        private static Assembly LoadFromResolveHandler(object sender, ResolveEventArgs args)
+        {
+            Assembly requestingAssembly = args.RequestingAssembly;
+            Assembly requestedAssembly = null;
+
+            // Requesting assembly for LoadFrom is always loaded in defaultContext - proceed only if that
+            // is the case.
+            if (AssemblyLoadContext.Default == AssemblyLoadContext.GetLoadContext(requestingAssembly))
```
jkotas: Nit: I think early outs are better for this kind of code (http://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code has a good explanation why).

```
   if (AssemblyLoadContext.Default != AssemblyLoadContext.GetLoadContext(requestingAssembly))
       return null;

   string requestorPath = Path.GetFullPath(requestingAssembly.Location);
   if (string.IsNullOrEmpty(requestorPath))
       return null;
...

   if (!fRequestorLoadedViaLoadFrom)
       return null;
...
```

jkotas:
cc @davidfowl Does this do what you would expect?

src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
gkhanna79: Fixed.

src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
```diff
@@ -15,11 +15,76 @@ namespace System.Reflection
...
+
+                // Requestor assembly was loaded using loadFrom, so look for its dependencies
+                // in the same folder as it.
+                if (fRequestorLoadedViaLoadFrom)
+                {
+                    // Form the name of the assembly using the path of the assembly that requested its load.
+                    AssemblyName requestedAssemblyName = new AssemblyName(args.Name);
+                    string requestedAssemblyPath = Path.Combine(Path.GetDirectoryName(requestorPath), requestedAssemblyName.Name+".dll");
+
+                    requestedAssembly = AssemblyLoadContext.Default.LoadFromAssemblyPath(requestedAssemblyPath);
```
davidfowl: This returns the already loaded assembly if it's already loaded right?

davidfowl:
@jkotas yes this is good.

src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
gkhanna79: Yes, it will.

gkhanna79:
Fixes https://github.com/dotnet/corefx/issues/18791

src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
```diff
@@ -15,11 +15,74 @@ namespace System.Reflection
...
+            if (!fRequestorLoadedViaLoadFrom)
+                return null;
+
+            // Requestor assembly was loaded using loadFrom, so look for its dependencies
+            // in the same folder as it.
+            // Form the name of the assembly using the path of the assembly that requested its load.
+            AssemblyName requestedAssemblyName = new AssemblyName(args.Name);
+            string requestedAssemblyPath = Path.Combine(Path.GetDirectoryName(requestorPath), requestedAssemblyName.Name+".dll");
+
+            return AssemblyLoadContext.Default.LoadFromAssemblyPath(requestedAssemblyPath);
```
jkotas: @gkhanna79 Do this also need to add the `requestedAssemblyPath` to `s_LoadFromAssemblyList` ?

gkhanna79: I think it should - perhaps we should use Assembly.LoadFrom itself and it will endup doing the right thing? What do you think?

PR #11335, opened 2017-05-02
[Local GC] Fix a build break in feature branch caused by obsolete error

swgillespie:
The CI didn't catch this since it was still in flux when the PR causing this was merged.


swgillespie:
@dotnet-bot help

Issue #2305, opened 2015-12-11
Generate asmparse.c using a standard bision/yacc


kyulee1:
asmparse.c (for ilasm) is currently prebuilt using an internal version of yacc. 
This also caused many compilations warnings on x-platforms.

sivarv:
@kyulee1 Is this needed for RC2 or can be moved to future?

ranma42:
Would it be ok if the parser was generated using [byacc](https://en.wikipedia.org/wiki/Berkeley_Yacc) instead of Bison? It generates a parser that almost works, except for requiring `realloc()`.
I experimented with injecting a (kind of hackish, as it is based on `new[]`/`delete[]`, but simple and working) C++ reimplementation of `realloc`, but better options might be possible.

I would be happy to try and work on this issue.

RussKeldorph:
@ranma42 Sorry your response fell through the cracks.  byacc seems like a reasonable choice.  If you're still interested, I can check on the license (if any).  The most important thing is to get the source into a state where anyone can compile asmparse.y with public tools.  After that, we may want to figure out how to integrate the tools with the build and delete the generated prebuilt/asmparse.cpp.  The choice of tool should take the latter possibility into account.

ranma42:
I am still interested and I have gathered some more data and questions :)
I tried to generate the parser using several tools:
 - `btyacc` can generate an `asmparse.cpp` that compiles without any other intervention on the grammar or to the other files (as shown in https://github.com/ranma42/coreclr/commit/3dcf6dfd4e18eafbbf0419ee2154ac5427595cbd).
 - `byacc` can generate a parser that compiles if provided with an implementation of `realloc` (which is currently intentionally made unavailable)
 - `bison` generates a parser that does not compile for non-trivial reasons (such as conflicts in the `define`s)
 - `bisonc++` rejects the grammar

All of tools are available as packages for ubuntu, but they might not be as convenient to get and install in other environments.

If the current direction (`btyacc`) looks promising, I can work on completing the integration, otherwise I can revert to `byacc` by adding and documenting the hackery needed to let it get some kind of `realloc`.

Regarding the integration of `btyacc`, I wonder if it makes sense to build it as part of the compilation process instead of adding it as an external requirement.

RussKeldorph:
@ranma42 Awesome! The fact that btyacc can be used without changes is great news.  At the very least people will be able to test changes to asmparse.y before submitting a PR.

[...]

PR #11350, opened 2017-05-02
Fix some static analysis warnings

RussKeldorph:
Most fixes are just addressing the use of (signed) enum-typed variables
as array indices.  Casting to unsigned allows us to cheaply include the
lower bound checks in the existing upper bound checks.  I would prefer
to force the underlying types of enumerations to be unsigned, but that
is a relatively new C++ feature and could have broader consequences than
I want to risk at this point.

The one other fix to asmparse.cpp just suppresses a warning that,
while technically valid, is not causing a real problem.  Perhaps this
will get better if/when #2305 is addressed.

### DIFF
--- a/src/ilasm/prebuilt/asmparse.cpp
+++ b/src/ilasm/prebuilt/asmparse.cpp
@@ -4,12 +4,13 @@
 
 #line 2 "asmparse.y"
 
- // Licensed to the .NET Foundation under one or more agreements.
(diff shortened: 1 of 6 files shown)


RussKeldorph:

RussKeldorph:
Remaining arm failures are known and would pass if rerun--assuming no other known failure is hit.

RussKeldorph:
@dotnet/jit-contrib

PR #11358, opened 2017-05-03
[Local GC] Fix some handle table interface violations in COMDependentHandle

swgillespie:
I found this interface violation when trying to build the GC without using VM headers (it manifests as a linker error). I didn't see any equivalent to `Set/GetDependentHandleSecondary` on the interface yet, so this PR adds and implements it.

cc @adityamandaleeka @Maoni0 @jkotas PTAL?


swgillespie:

swgillespie:
thanks!

PR #11342, opened 2017-05-02
Fix load from

gkhanna79:
@jkotas PTAL

PR #11359, opened 2017-05-03
Place precompiled SPC.dll in native assets folder.

gkhanna79:
Possible fix for https://github.com/dotnet/core-setup/pull/2255#issuecomment-298800443

### DIFF
--- a/src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/runtime.Linux.Microsoft.NETCore.Runtime.CoreCLR.props
+++ b/src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/runtime.Linux.Microsoft.NETCore.Runtime.CoreCLR.props
@@ -17,7 +17,7 @@
     <NativeBinary Include="$(BinDir)libsosplugin.so" />
     <NativeBinary Include="$(BinDir)System.Globalization.Native.so" />
     <NativeBinary Include="$(BinDir)sosdocsunix.txt" />
-    <ArchitectureSpecificLibFile Include="$(BinDir)System.Private.CoreLib.dll" />
+    <NativeBinary Include="$(BinDir)System.Private.CoreLib.dll" />
     <ArchitectureSpecificLibFile Include="$(BinDir)SOS.NETCore.dll" />
     <ArchitectureSpecificToolFile Include="$(BinDir)crossgen" />
   </ItemGroup>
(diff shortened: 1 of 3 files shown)


gkhanna79:
Ubuntu and Tizen Arm CI legs are failing with same issue as https://github.com/dotnet/coreclr/pull/11342#issuecomment-298909399.

@mmitche @hqueue @hseok-oh Can you please help look into this?

gkhanna79:
In PR https://github.com/dotnet/core-setup/pull/2255, @karajas ran into test failures for standalone apps that were not published (i.e. when running standalone apps that were output of just doing _dotnet build_).

The failures happened since, in the above case, the test app runs against nuget cache and thus, loads CoreCLR from nuget cache as well (as expected). Since CoreCLR always expects SPC.dll to be next to it, the test failed as SPC.dll was moved to be under _lib_ folder in the nuget package as opposed to _runtimes/native_ folder in my PR https://github.com/dotnet/coreclr/pull/11040.

There are two ways to address this:

1) We move SPC.dll to be under _runtimes/native_ folder in the nuget package. As a result, when the package is restored in the cache, CoreCLR will find SPC.dll next to itself and things would work. 

2) We compute the path to SPC.dll from the TPA list passed to the runtime and if SPC.dll is not found next to CoreCLR, we use the computed path (similar to how crossgen computes it for the crossgen process).

(1) is simpler and is what this current PR is about. I wanted to see if any test builds would break but we have Priority 1 test legs clean as well (the Linux Arm failures are unrelated that I am following up on seperately).

@weshaggard @ericstj Do you see any concerns with this approach?

CC @jkotas

gkhanna79:

gkhanna79:

gkhanna79:

gkhanna79:

ericstj:
I think this should be OK.  I'm not aware if anyone else expects S.P.C to be in lib (or a complete IL closure to be in LIB) but I strongly suspect not since this all the package refactoring that's been done.

gkhanna79:
> the host (since it distinguishes lib vs native)

I did check the host - it does not specialize any check for System.Private.CoreLib.dll since that is loaded directly by the runtime. That said, given that CoreLib is on TPA (because it is in lib), I will make a seperate change to the host to ensure that still remains the case (since now it will be next to CoreCLR.dll, which was the case before my original change as well).

@karajas As discussed, I will merge this and get a build going. You should be able to pick that up and not need the workaround.

gkhanna79:
@karajas I have queued off https://devdiv.visualstudio.com/DevDiv/_Build/explorer?_a=summary&buildId=719817 to get the updated CoreCLR package you can use.

weshaggard:
I'm also good with moving this to the native folder.

PR #11360, opened 2017-05-03
Revert "[ARM32/Linux] Copy tests.zip only for CI test and use x86 unittest"

RussKeldorph:
This reverts commit 2ec8b5c2cc6d68ca02eb743e679997a8f11c90aa (#11222).

I believe this should restore ARM CI job function.


RussKeldorph:
@hseok-oh FYI.  If this doesn't restore ARM CI, I will revert the revert.

Issue #5999, opened 2016-06-24
Compiler::gtUpdateSideEffects is incorrect.


pgavlin:
https://github.com/dotnet/coreclr/blob/master/src/jit/gentree.cpp#L7114

The intent of this function is to update the side effects for a tree after one of its children has been rewritten. Unfortunately, it is broken in two significant ways:
- It does not take into account side-effects from the siblings of the rewritten tree
- It does not take into account side-effects from the parent node itself

As a concrete example, consider the following tree:

```
               [000036] ------------                /--*  const     int    -1
               [000037] ---XG----U--             /--*  +_ovfl    int   
               [000034] ------------             |  |  /--*  lclVar    int    V01 loc0         
               [000035] ---XG----U--             |  \--*  +_ovfl    int   
               [000032] ------------             |     |  /--*  const     int    0
               [000033] ---XG-------             |     \--*  []        int   
               [000031] ------------             |        \--*  lclVar    ref    V02 loc1         
               [000039] -A-XG-------             *  =         int   
               [000038] D------N----             \--*  lclVar    int    V03 loc2
```

If the array index expression was hoisted to a temp and replaced with a use of the temp, the tree should become something like:

```
               [000036] ------------                /--*  const     int    -1
               [000037] ---XG----U--             /--*  +_ovfl    int   
               [000034] ------------             |  |  /--*  lclVar    int    V01 loc0         
               [000035] ---X-----U--             |  \--*  +_ovfl    int   
               [000033] ------------             |     \--*  lclVar    int    V04 tmp0         
               [000039] -A-XG-------             *  =         int   
               [000038] D------N----             \--*  lclVar    int    V03 loc2
```

In particular, note that both adds still have the exception flag set. Unfortunately, if `gtUpdateSideEffects` was called with the original array index expression and its replacement temp, it would produce the following tree:

```
               [000036] ------------                /--*  const     int    -1
               [000037] ---------U--             /--*  +_ovfl    int   
               [000034] ------------             |  |  /--*  lclVar    int    V01 loc0         
               [000035] ---------U--             |  \--*  +_ovfl    int   
               [000033] ------------             |     \--*  lclVar    int    V04 tmp0         
               [000039] -A----------             *  =         int   
               [000038] D------N----             \--*  lclVar    int    V03 loc2
```

[...]

PR #11361, opened 2017-05-03
Add missing GTF_EXCEPT when importing string length intrinsic

### LINKED ISSUE #11343: [RyuJIT] Accessing Length on a `null` string does not throw NullReferenceException
The following throws in Debug and does not throw in Release. 
Should throw regardless of optimizations.

```csharp
    class Program
    {
        {
            string s = null;
            int unused1 = s.Length; // <- should throw but gets optimized away
        }
    }
```

Tried on Desktop 32/64 bit and on netcoreapp1.1 with same results - NRE in debug, no exceptions in release.

The IL looks like the following

```
.method private hidebysig static 
	void Main (
		string[] args
	) cil managed 
{
	// Method begins at RVA 0x2050
	// Code size 8 (0x8)
	.maxstack 8
	.entrypoint

	IL_0000: ldnull
	IL_0001: callvirt instance int32 [mscorlib]System.String::get_Length()
	IL_0006: pop
	IL_0007: ret
} // end of method Program::Main
```

Note - the `callvirt` is still there and should throw when invoked on a null.

mikedn:
Fixes #11343

### DIFF
--- a/tests/src/JIT/Regression/JitBlue/GitHub_11343/GitHub_11343.csproj
+++ b/tests/src/JIT/Regression/JitBlue/GitHub_11343/GitHub_11343.csproj
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <AssemblyName>$(MSBuildProjectName)</AssemblyName>
+    <SchemaVersion>2.0</SchemaVersion>
+    <ProjectGuid>{95DFC527-4DC1-495E-97D7-E94EE1F7140D}</ProjectGuid>
+    <OutputType>Exe</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <FileAlignment>512</FileAlignment>
+    <ProjectTypeGuids>{786C830F-07A1-408B-BD7F-6EE04809D6DB};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
+    <ReferencePath>$(ProgramFiles)\Common Files\microsoft shared\VSTT\11.0\UITestExtensionPackages</ReferencePath>
+    <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\..\</SolutionDir>
+
+    <NuGetPackageImportStamp>7a9bfb7d</NuGetPackageImportStamp>
+  </PropertyGroup>
+  <!-- Default configurations to help VS understand the configurations -->
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+  </PropertyGroup>
+  <ItemGroup>
+    <CodeAnalysisDependentAssemblyPaths Condition=" '$(VS100COMNTOOLS)' != '' " Include="$(VS100COMNTOOLS)..\IDE\PrivateAssemblies">
+      <Visible>False</Visible>
+    </CodeAnalysisDependentAssemblyPaths>
+  </ItemGroup>
+  <PropertyGroup>
+    <DebugType></DebugType>
+    <Optimize>True</Optimize>
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="GitHub_11343.cs" />
+  </ItemGroup>
+  <ItemGroup>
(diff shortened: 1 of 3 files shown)


mikedn:
jit-diff summary:
```
Total bytes of diff: 110 (0.00 % of base)
    diff is a regression.
          73 : Microsoft.CodeAnalysis.CSharp.dasm (0.00 % of base)
          29 : Microsoft.CodeAnalysis.dasm (0.00 % of base)
          13 : System.Private.CoreLib.dasm (0.00 % of base)
           8 : Microsoft.CodeAnalysis.VisualBasic.dasm (0.00 % of base)
           2 : System.Reflection.Metadata.dasm (0.00 % of base)
         -15 : System.Private.Uri.dasm (-0.02 % of base)
6 total files with size differences (1 improved, 5 regressed).
          19 : Microsoft.CodeAnalysis.dasm - PeWriter:WriteDirectory(ref,ref,int,int,int,int,ref):this
          17 : Microsoft.CodeAnalysis.CSharp.dasm - DocumentationCommentCompiler:TrimmedStringStartsWith(ref,ref):bool
           8 : Microsoft.CodeAnalysis.CSharp.dasm - SyntaxTokenWithValueAndTrivia`1:SetDiagnostics(ref):ref:this
           8 : Microsoft.CodeAnalysis.CSharp.dasm - SyntaxTokenWithValueAndTrivia`1:WithLeadingTrivia(ref):ref:this
           8 : Microsoft.CodeAnalysis.CSharp.dasm - SyntaxTokenWithValueAndTrivia`1:WithTrailingTrivia(ref):ref:this
          -7 : System.Private.Uri.dasm - Uri:UnescapeDataString(ref):ref
          -6 : System.Private.Uri.dasm - Uri:Unescape(ref):ref:this
          -4 : System.Private.CoreLib.dasm - EventSource:CreateManifestAndDescriptors(ref,ref,ref,int):ref
          -3 : System.Private.Uri.dasm - Uri:get_DnsSafeHost():ref:this
          -1 : System.Private.CoreLib.dasm - String:Join(ref,ref):ref (4 methods)
41 total methods with size differences (6 improved, 35 regressed).
```

RussKeldorph:
@dotnet/jit-contrib

BruceForstall:

mikedn:
Adding `GTF_EXCEPT` is what the importer already does for `ldlen` so this should be fine. Anyway...

I went a bit through the diffs and it seems that they mostly consist of moved around code. Or more like code that's no longer moved around due to the added `GTF_EXCEPT`. Size regressions appear to be caused by change in register allocation - spills, reloads, different register choices that increase instruction size (e.g. `r8` instead of `rcx`).

There are also some unexpected improvements:
```asm
       shl      r8d, 1
       add      r8d, 2
; becomes
       lea      r8d, [2*r8+2]
```
No idea how adding `GTF_EXCEPT` causes that. It seems that it should be the other way around.

A small regression that caught my eye:
```asm
       mov      eax, dword ptr [rsp+28H]
       mov      rcx, qword ptr [(reloc)]
       mov      rcx, gword ptr [rcx]
       add      eax, dword ptr [rcx+8]
; becomes
       mov      rax, qword ptr [(reloc)]
       mov      rax, gword ptr [rax]
       mov      eax, dword ptr [rax+8]
       mov      ecx, dword ptr [rsp+28H]
       add      eax, ecx
```
It would seem that `GTF_EXCEPT` has a bigger impact than it should. I suppose that shouldn't be surprising, it's simply easier to just not touch `GTF_EXCEPT` nodes than to move them around to the extent this doesn't cause changes in observable behavior.

pgavlin:
> It would seem that `GTF_EXCEPT` has a bigger impact than it should. I suppose that shouldn't be surprising, it's simply easier to just not touch `GTF_EXCEPT` nodes than to move them around to the extent this doesn't cause changes in observable behavior.

Because it's a side effect flag, `GTF_EXCEPT` can have an outsized effect if it is set early: if a tree with `GTF_EXCEPT` is embedded in a larger tree, the `GTF_EXCEPT` flag will bubble up through the tree's ancestors. Since we cannot yet reliably remove side-effect flags in all cases (see #5999), the backend may be unnecessarily pessimistic about code motion (including the sort of code motion that is used to form RMW operations). It's also possible that the `GTF_EXCEPT` flag is blocking some transforms in `gtSetEvalOrder` or morph; it's not terribly clear without a dump.

RussKeldorph:
@mikedn Would you mind adding a regression test with the PR?

mikedn:
Yikes, should have added a test from the start. It's there now.

tests/src/JIT/Regression/JitBlue/GitHub_11343/GitHub_11343.csproj
```diff
@@ -0,0 +1,42 @@
...
+    <CodeAnalysisDependentAssemblyPaths Condition=" '$(VS100COMNTOOLS)' != '' " Include="$(VS100COMNTOOLS)..\IDE\PrivateAssemblies">
+      <Visible>False</Visible>
+    </CodeAnalysisDependentAssemblyPaths>
+  </ItemGroup>
+  <PropertyGroup>
+    <DebugType></DebugType>
+    <Optimize>True</Optimize>
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="GitHub_10940.cs" />
```
BruceForstall: @mikedn got the filename wrong here.

mikedn: Fixed. I thought it uses `$(MSBuildProjectName)` but it's only the assembly name that does that.

PR #11362, opened 2017-05-03
Fix triggering of perf correctness runs

DrewScoggins:
When I made the change to have the correctness runs fire in a PR without
the trigger phrase I accidentally did it for all PR's not just for the
smoketest runs.  This change fixes that.


DrewScoggins:
test ci please

PR #11367, opened 2017-05-03
Document Linux libnuma dependency

tmds:
cc @janvorli


tmds:
No problem. Will the 2.0 (preview) installers be built with numa support? libnuma is a native dependency then too.

Issue #11348, opened 2017-05-02
The SOS gc commands don't work on the createdump utility dumps


mikem8361:
The default createdump option to create a dump "with heaps" doesn't have all the heap memory on Linux so the various SOS gc commands like gcroot fail.  

The fix is in the /proc/$pid/maps parsing to add all the unnamed heap sections.  Currently these are not being added for "with heaps".

mikem8361:
PR #11368 fixes this.

PR #11368, opened 2017-05-03
Createdump fixes and cleanup

mikem8361:
Fixed "with heap" option to actually include all the heaps. The way /proc/$pid/maps
was parsed missed most of the RW data/heap regions.

Added -u/--full full core dump options.

Removed the useless m_memStatus variable in enummem.cpp.

Issue #11348


noahfalk:
:shipit:

PR #11372, opened 2017-05-03
Disable symbol package MyGet publish on release/*

dagood:
Adds `if ($env:SourceBranch.StartsWith("release/")) { exit }` to the symbol package publish step. This will automatically disable symbol package publishing to MyGet for `release/*` branches. This was previously done by making manual changes on the release branch, but this step doesn't need to be manual.

This is the CoreCLR part of https://github.com/dotnet/core-eng/issues/728.

### DIFF
--- a/buildpipeline/DotNet-Trusted-Publish.json
+++ b/buildpipeline/DotNet-Trusted-Publish.json
@@ -172,7 +172,7 @@
         "scriptType": "inlineScript",
         "scriptName": "",
         "arguments": "$(MyGetApiKey) $(ConfigurationGroup)",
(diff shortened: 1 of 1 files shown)


gkhanna79:
General question - where are symbols for release branches getting published?

dagood:
In all cases, symbol packages are automatically stored in the intermediate blob storage and in VSTS Drop. Other than that, automation publishes to:
 * release/1.0.0 and release/1.1.0: nowhere
 * release/2.0.0 preview1: MyGet (unintentionally--I'm going to have them removed)
 * master (future release/2.0.0 builds): archived on Microsoft official symbol server

We have upcoming work that will make it secure to get symbols from MyGet, so this mapping will change to include automatically publishing to MyGet (especially in master and future feature branch forks).

Issue #11199, opened 2017-04-25
UnmanagedMemoryStream on 32bit Linux


seanshpark:
In UnmanagedMemoryStream, besides of Initialize , `Position` can be set to `Int32_MaxValue`.
```
stream.Position = int.MaxValue;
```
[corefx:UmsWrite.cs#L115](https://github.com/dotnet/corefx/blob/master/src/System.IO.UnmanagedMemoryStream/tests/UmsWrite.cs#L115)

In Linux, user allocated memory address can be up to 0xBFFFFFFF while Win32 0x7FFFFFFF (as far as I know). So, wrap around check in `Position` can happen with normal case.
```
if (value > (long) Int32.MaxValue || _mem + value < _mem)
```
[UnmanagedMemoryStream.cs#L305](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/IO/UnmanagedMemoryStream.cs#L305)

As of result, somethimes, unit test fails,
```
   System.IO.Tests.UmsWriteTests.CannotWriteWithOverflow [FAIL]
      System.ArgumentOutOfRangeException : Stream length must be non-negative and less than 2^31 - 1 - origin.
      Parameter name: value
      Stack Trace:
            at System.IO.UnmanagedMemoryStream.set_Position(Int64 value)
            at System.IO.Tests.UmsWriteTests.CannotWriteWithOverflow()
```

seanshpark:
@jkotas , could you please give some advice on how to solve this issue?

mikedn:
> In Linux, user allocated memory address can be up to 0xBFFFFFFF while Win32 0x7FFFFFFF (as far as I know).

By default 32 bit Windows reserves 2GB for user mode but it is possible to configure it (and the application) to reserve 3GB for user mode. Also, 32 bit processes running on 64 bit Windows can get up to 4GB of user mode address space.

If the test passes on Windows it's because we're lucky or because the .NET Core host process isn't large address aware.

seanshpark:
@mikedn , thank you. I've found this; https://msdn.microsoft.com/en-us/library/windows/desktop/aa366912(v=vs.85).aspx . I think it's `Windows with 4GT`

mikedn:
[...]

PR #11376, opened 2017-05-03
Remove UnmanagedMemoryStream overflow checks.

ianhays:
A few of these checks aren't able to be hit, and one of the checks was causing issues on 32-bit systems.

resolves https://github.com/dotnet/coreclr/issues/11199

cc: @jkotas

### DIFF
--- a/src/mscorlib/src/System/IO/UnmanagedMemoryStream.cs
+++ b/src/mscorlib/src/System/IO/UnmanagedMemoryStream.cs
@@ -84,8 +84,6 @@ namespace System.IO
      */
     public class UnmanagedMemoryStream : Stream
     {
-        private const long UnmanagedMemStreamMaxLength = Int64.MaxValue;
-
         private SafeBuffer _buffer;
         private unsafe byte* _mem;
         private long _length;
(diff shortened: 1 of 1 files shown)


src/mscorlib/src/System/IO/UnmanagedMemoryStream.cs
```diff
@@ -334,10 +325,6 @@ public override long Position
                     throw new NotSupportedException(SR.NotSupported_UmsSafeBuffer);
                 if (!_isOpen) __Error.StreamIsClosed();
 
-                // Note: subtracting pointers returns an Int64.  Working around
```
jkotas: I think this should be:

```
long newPosition = (long)value - (long)_mem;
if (newPosition < 0)
    throw new ArgumentOutOfRangeException("offset", SR.ArgumentOutOfRange_UnmanagedMemStreamLength); 
if (value < _mem)
    throw new IOException(SR.IO_SeekBeforeBegin); 

Interlocked.Exchange(ref _position, newPosition);
```

to throw for the negative position case from my example; and to behave correctly for other similar cases.

jkotas: Another example:

```
static void Main() {
     var ums = new UnmanagedMemoryStream((byte*)0x40000000, 1);
     ums.PositionPointer = (byte*)0xF0000000;
     Console.WriteLine(ums.Position.ToString());
}
```

If you run it on 32-bit machine, the current implementation will produce negative position. Instead, it should produce Position == 0xB0000000. We should add this one as regression test case to corefx.

ianhays: That seems reasonable. I was under the impression that we would want to have the same behavior as netfx and wait to throw until the next operations on the UMS that operate on the invalid position. Actually throwing for a negative position is certainly more useful though.

PR #11377, opened 2017-05-03
Refactor arm32/arm64 per-node codegen into functions

BruceForstall:
These new functions can sometimes be shared, as they are identical or
virtually so.

In addition, the driver function genCodeForTreeNode() has been
made common.

### DIFF
--- a/src/jit/codegenarm.cpp
+++ b/src/jit/codegenarm.cpp
@@ -1060,12 +1060,12 @@ void CodeGen::genLeaInstruction(GenTreeAddrMode* lea)
 }
 
 //------------------------------------------------------------------------
-// genCodeForDiv: Produce code for a GT_DIV/GT_UDIV/GT_MOD/GT_UMOD node.
+// genCodeForDivMod: Produce code for a GT_DIV/GT_UDIV/GT_MOD/GT_UMOD node.
 //
 // Arguments:
 //    tree - the node
 //
-void CodeGen::genCodeForDiv(GenTreeOp* tree)
+void CodeGen::genCodeForDivMod(GenTreeOp* tree)
 {
     assert(tree->OperIs(GT_DIV, GT_UDIV, GT_MOD, GT_UMOD));
 
--- a/src/jit/codegenarm.cpp
+++ b/src/jit/codegenarm.cpp
@@ -259,6 +259,11 @@ void CodeGen::genReturn(GenTreePtr treeNode)
     GenTreePtr op1        = treeNode->gtGetOp1();
     var_types  targetType = treeNode->TypeGet();
 
+    // A void GT_RETFILT is the end of a finally. For non-void filter returns we need to load the result in the return
+    // register, if it's not already there. The processing is the same as GT_RETURN. For filters, the IL spec says the
+    // result is type int32. Further, the only legal values are 0 or 1; the use of other values is "undefined".
+    assert(!treeNode->OperIs(GT_RETFILT) || (targetType == TYP_VOID) || (targetType == TYP_INT));
+
 #ifdef DEBUG
     if (targetType == TYP_VOID)
     {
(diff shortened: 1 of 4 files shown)


BruceForstall:

BruceForstall:
@briansull @dotnet/arm32-contrib @dotnet/arm64-contrib PTAL
cc @dotnet/jit-contrib

src/jit/codegenlinear.h
```diff
@@ -173,6 +173,14 @@ void genCodeForShiftLong(GenTreePtr tree);
...
 #endif // _TARGET_XARCH_
 
+void genCodeForCast(GenTreeOp* tree);
+
+void genCodeForLclAddr(GenTree* tree);
+
+void genCodeForIndir(GenTreeIndir* tree);
+
+void genCodeForDiv(GenTreeOp* tree);
+
```
hqueue: It seems that there is a `genCodeForDivMod`defined in codegenlinear.h at https://github.com/dotnet/coreclr/blob/master/src/jit/codegenlinear.h#L23 but not used for arm and arm64.

How about merging `genCodeForDiv` with `genCodeForDivMod` ? or making one one of them is defined for armarch ?

hqueue: For example, define `genCodeForDivMod` for arm, and define `genCodeForDiv` for arm64.

src/jit/codegenarm.cpp
```diff
@@ -1566,6 +1192,22 @@ void CodeGen::genCodeForCompare(GenTreeOp* tree)
...
+// genCodeForJcc: Produce code for a GT_JCC node.
+//
+// Arguments:
+//    tree - the node
+//
+void CodeGen::genCodeForJcc(GenTreeJumpCC* tree)
+{
+    assert(compiler->compCurBB->bbJumpKind == BBJ_COND);
+
+    CompareKind  compareKind = ((tree->gtFlags & GTF_UNSIGNED) != 0) ? CK_UNSIGNED : CK_SIGNED;
```
mikedn: You could use `IsUnsigned` while you're at it.

src/jit/codegenarmarch.cpp
```diff
@@ -25,6 +25,387 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
...
+#endif // _TARGET_ARM64_
+
+    // contained nodes are part of their parents for codegen purposes
+    // ex : immediates, most LEAs
+    if (treeNode->isContained())
+    {
+        return;
+    }
+
+    switch (treeNode->gtOper)
```
mikedn: OperGet?

src/jit/codegenlinear.h
```diff
@@ -173,6 +173,14 @@ void genCodeForShiftLong(GenTreePtr tree);
 void genCodeForShiftRMW(GenTreeStoreInd* storeInd);
 #endif // _TARGET_XARCH_
 
+void genCodeForCast(GenTreeOp* tree);
+
```
mikedn: Do we need all these empty lines between declarations? They don't seem to contribute to readability.

src/jit/codegenarm.cpp
BruceForstall: When moving code around (e.g., between files), I tried not to change it at all. There are lots of places we could move to more "modern" style, like this, or using the new(-ish) OperIs() function.

I also didn't do the same refactoring on xarch, because I didn't want to touch it before the last 2.0  branch from master.

src/jit/codegenlinear.h
BruceForstall: Agreed.

src/jit/codegenlinear.h
BruceForstall: I didn't see that; thanks. I'm going to use `genCodeForDivMod` for both archs, to be consistent, and just assert no MOD happens for arm64.

src/jit/codegenarm.cpp
mikedn: Makes sense. I was under the impression that you changed some of the code because I saw `OperIs` showing up in a number of places.

PR #11379, opened 2017-05-04
[Local GC] Obtaining configuration information

swgillespie:
This is an early PR containing a work-in-progress removal of the usage of `g_pConfig` from within the GC. I'm sending this out very early in hopes of getting feedback on the direction that these changes are headed before committing to the code changes within `gc.cpp`. A summary of the changes so far:

1. Three new methods are added to `IGCToCLR`: `GetBooleanConfigValue`, `GetIntConfigValue`, and `GetSizeTConfigValue` (possibly also `GetStringConfigValue`, if necessary). Each one allows the GC to ask of the EE: "Is this configuration key set and, if so, can you give me its value?". The details of how configuration values are obtained (or even if the runtime allows for configuration at all) are left to the runtime.
2. A new class `GCConfig` is added to the GC side of the interface that will call the above three methods on behalf of the GC, depending on the configuration value requested. I think that this is better than calling the above three methods directly from the GC because it gives us a central point to hook into the process of obtaining configuration information. In particular, it would be useful for us as GC developers to have the flexibility to add our own fallback configuration mechanisms (e.g. special environment variables) when querying configuration keys that the runtime doesn't know about (e.g. ones added for the purpose of debugging or experimenting with a new GC feature).
3. Code in the GC referring to the EE's global configuration state (`g_heap_type`, `g_IGCconcurrent`, `g_pConfig`, and friends) will be redirected to use `GCConfig` instead.

All three of these are sketched out in the initial commit of this PR. Before I continued on 3), though, I was hoping to gather some feedback on this design so I don't go heading off in the wrong direction - @Maoni0 @adityamandaleeka @sergiy-k @jkotas I'd appreciate any of your thoughts!


jkotas:
I do not think you want to have enum with config values. I makes it impossible to add a new config switch without touching the VM - it is not something you want to the standalone GC. Instead, there should API that takes strings for the random config values. There is prior art in the JIT/EE interface -search for getIntConfigValue, getStringConfigValue and freeStringConfigValue. I think you want something very similar to that. It may be a good idea to change `wchar_t` to char to make it more portable, and change the type of getIntConfigValue to `int64_t` if you would like to allow passing very large numbers. I do not think you need a different APIs for different integer sizes.

It may be fine to continue using flags or dedicated APIs for very stable config settings (server/client, concurent/non-concurent, ...), but it may be as well fine to just use the string based config API for everything. Up to you.

AppDomainLeaks check you can just get rid off. It was very slow debug-only check that we have stopped using on desktop like 10 years ago.

swgillespie:
Thank you! I'll move in this direction, then.

> ... but it may be as well fine to just use the string based config API for everything. Up to you.

I personally like the sound of uniformity, so I think I'll opt for that unless anyone has any objections.

swgillespie:
In my last commit I moved towards string keys and implemented a few of the config keys that the GC uses - if there aren't any objections to this design I think I can start bringing it to the GC tomorrow and testing it in earnest. Thanks for the feedback so far!

src/vm/gcenv.ee.h
```diff
@@ -52,6 +52,10 @@ class GCToEEInterface : public IGCToCLR {
     bool ForceFullGCToBeBlocking();
     bool EagerFinalized(Object* obj);
     MethodTable* GetFreeObjectMethodTable();
+    bool GetBooleanConfigValue(const char* key, bool* value);
+    bool GetIntConfigValue(const char* key, int64_t* value);
+    bool GetStringConfigValue(const char* key, const char** value);
+    void FreeStringConfigValue(const char*key);
```
jkotas: > void FreeStringConfigValue(const char*key); 

The argument name should be `value`

jkotas:
Looks fine to me so far

src/gc/gc.cpp
```diff
@@ -30951,8 +30947,8 @@ void gc_heap::set_mem_verify (uint8_t* start, uint8_t* end, uint8_t b)
 #ifdef VERIFY_HEAP
     if (end > start)
     {
-        if ((g_pConfig->GetHeapVerifyLevel() & EEConfig::HEAPVERIFY_GC) &&
-           !(g_pConfig->GetHeapVerifyLevel() & EEConfig::HEAPVERIFY_NO_MEM_FILL))
+        if ((GCConfig::GetHeapVerifyLevel() & EEConfig::HEAPVERIFY_GC) &&
```
jkotas: You may need to cache the frequently accessed values like the GCHeapVerifyLevel. Fetching them by name each time is going to be too slow.

swgillespie: Yeah, I was thinking so too - I'll add that in the next commit.

swgillespie:
I think this last commit is ready for review - this PR has passed all of the manual validation that I've thrown at it, though I will continue to validate all of these different configuration options throughout the day today.

Some notes on this PR:

1. I left all of the options pertaining to GC Stress alone, for now. This is not blocking the Local GC project for now (since the standalone GC build does not build with `STRESS_HEAP` defined) but it will need to be addressed as a part of https://github.com/dotnet/coreclr/issues/10403.
2. I don't see any code in the EE that respects these two startup flags: https://github.com/dotnet/coreclr/blob/master/src/inc/MSCOREE.IDL#L155 and https://github.com/dotnet/coreclr/blob/master/src/inc/MSCOREE.IDL#L162 . As far as I can tell, these global variables corresponding to these two startup flags are never written to `(g_IGCHoardVM` and `g_IGCTrimCommit`): https://github.com/dotnet/coreclr/blob/master/src/vm/eeconfig.cpp#L878-L890 . Are these something that we support on CoreCLR? These configs can still be turned on using environment variables, but it seems like the startup flags are ignored.

Maoni0:
VM hoarding and Trim are used by asp.net on desktop. On CoreCLR since the hosting support is extremely limited I wouldn't be surprised if they are not used at all.

src/gc/gcconfig.cpp
```diff
@@ -0,0 +1,50 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#include "common.h"
+#include "gcenv.h"
+#include "gc.h"
+
+#define BOOL_CONFIG(name, key, default, unused_doc)            \
+  bool GCConfig::Get##name()                                   \
```
jkotas: This caching is not thread safe (on ARM for sure, on Intel it depends on the memory model implemented by the C++ compiler).

swgillespie: Is the problem that the cached flag write may be observed before the cached value write on ARM?

jkotas: > Is the problem that the cached flag write may be observed before the cached value write on ARM

Right

> maybe Intel

The compiler may reorder the reads or writes - the C++ spec allows it.

Maoni0: so why do we even need to check for cached values? We are not just getting all the configs on startup?

swgillespie: In this scheme configs are retrieved lazily upon first request. We can certainly opt to do it eagerly on startup instead if that's easier.

Maoni0: I would just do a simple load on all configs like we do today. I don't see any reason to be more complicated than that.

swgillespie: Ok, will do!

swgillespie:
I think I've addressed all of the feedback in this thread - would anyone have any objections to merging this? My last commit addresses an issue I found where HoardVM config wasn't getting propagated to the GC; it should work now when specified using startup flags.

src/vm/gcenv.ee.cpp
```diff
@@ -993,3 +993,134 @@ MethodTable* GCToEEInterface::GetFreeObjectMethodTable()
...
+    if (strcmp(key, "gcConcurrent") == 0)
+    {
+        *value = g_IGCconcurrent != 0;
+        return true;
+    }
+
+    if (strcmp(key, "GCRetainVM") == 0)
+    {
+        // GCRetainVM is strange in that it is derived from startup flags
+        // on unixes, but not on windows. This call preserves the current
```
Maoni0: > but not on windows [](start = 21, length = 19)

why not on windows? this is what the g_IGCHoardVM is for. if that's not set, then you need to get the config.

swgillespie: I think this comment is inaccurate now that I fixed the issue in my last commit. `g_pConfig->GetGCRetainVM` does get the config if `g_IGCHoardVM` isn't set.

swgillespie: It is sort of strange that Windows CoreRun uses ICLRRuntimeHost4 and Unix CoreRun doesn't, though; it means that Unix CoreRun flows through an extra layer of configuration code (translating key/value pairs into start flags) that the Windows one doesn't. I don't know how the CLI hosts the runtime so I'm not sure what it does.

Maoni0: whether ICLRRuntimeHost4 is used or not, wouldn't g_pConfig->GetGCRetainVM take care of that detail? or you mean you are not sure if g_pConfig->GetGCRetainVM is getting the value correctly?

swgillespie: > or you mean you are not sure if g_pConfig->GetGCRetainVM is getting the value correctly?

For this comment, I meant that - the EE wasn't setting `g_IGCHoardVM` at all, regardless of the startup flags. This comment can probably be deleted now that this is no longer the case.

Maoni0: right, so this comment is actually incorrect now and should be deleted.

swgillespie: Will do!

swgillespie:
I'm going to go ahead and merge this into dev/local-gc - if there are any objections to this PR, it'll be flowing into master as a part of https://github.com/dotnet/coreclr/pull/11699. Thanks for the reviews!

PR #11382, opened 2017-05-04
Make CoreCLR work properly under PaX's RANDMMAP

### LINKED ISSUE #8480: Make CoreCLR work properly on GRSEC/PAX enabled Linuxes
When GRSEC/PAX has RANDMMAP enabled, it never honors the address passed to mmap unless the MAP_FIXED option is passed in. But there are two parts of CoreCLR that rely on the fact that mmap honors the address unless there is an overlapping mapping. First part is the executable allocator in PAL that tries to  reserve memory range in close vicinity of the libcoreclr.so (within 2GB range) to make sure that jitted managed code that calls into libcoreclr.so doesn't have to use jump stubs.
The second part is the jump stub allocator. If code in one jitted module calls code in another one and the distance of the jump is larger than 2GB, jitter tries to allocate jump stub within 2GB range from the caller address.
Both of these mechanisms fail on GRSEC/PAX enabled Linux since we cannot control the address where we create mappings. The result is either coreclr_initialize failure with HR 0x80004005 or an out of memory exception. Since the mapping address picked by the kernel is random, sometimes, after many attempts, we can actually start and run fine if by a chance the mapping picked by the kernel matches the requested one.
There is a workaround for this - disabling the RANDMMAP for dotnet, corerun or the renamed coreconsole executables using the `paxctl -c -r` command. But it would be great to make it work, possibly with lower performance, even without this workaround.

kouvel:
Issues:
- The ExecutableMemoryAllocator is used to attempt to map native images into a memory range near libcoreclr so that its helper table can use short relative addresses for jumps to libcoreclr
  - RANDMMAP typically prevents mmap calls with a specific address from reserving memory at the requested address, so the executable memory allocator fails to reserve any memory. When Server GC is enabled, the large GC heap can exacerbate the issue by taking address space near libcoreclr.
- Native images are loaded far from libcoreclr, and now jump stub space needs to be allocated near the native image, but RANDMMAP typically prevents this too
- NGenReserveForJumpStubs is intended to reserve some memory near mapped native images for jump stubs, but that reservation is done with a separate mmap call in the same way as above and RANDMMAP typically prevents this too
- The JIT needs to allocate memory for code that may need to jump/call to a native image or libcoreclr, which may require jump stubs near the code that cannot be allocated. CodeHeapReserveForJumpStubs reserves space in code heap blocks without using a separate call to mmap, so this works, but without this environment variable by default there is still a good chance of failing.
- See https://github.com/dotnet/coreclr/blob/56d550d4f8aec2dd40b72a182205d0a2463a1bc9/Documentation/design-docs/jump-stubs.md for more details

Fixes #8480
- It would be ideal to fix all of the above properly, such that there would never be a need to attempt reserving memory within a certain range. Since we're running out of time for 2.0, I figured the following simpler, temporary solution that should cover most of the practical cases, may be appropriate for 2.0.
- Extended ExecutableMemoryAllocator to reserve address space even when it cannot do so near libcoreclr
- Had ClrVirtualAllocWithinRange use the executable memory allocator to reserve memory for jump stubs when the requested range is satisfied
- This covers a maximum of ~2 GB of executable code and should cover most of the practical cases. Once this space is exhausted, under RANDMMAP, native images loaded later will fail, and for jitted code the environment variable above can be used.

### DIFF
--- a/src/pal/src/include/pal/virtual.h
+++ b/src/pal/src/include/pal/virtual.h
@@ -59,8 +59,7 @@ enum VIRTUAL_CONSTANTS
     VIRTUAL_EXECUTE_READ,
     
     VIRTUAL_PAGE_SIZE       = 0x1000,
-    VIRTUAL_PAGE_MASK       = VIRTUAL_PAGE_SIZE - 1,
-    BOUNDARY_64K    = 0xffff
+    VIRTUAL_PAGE_MASK       = VIRTUAL_PAGE_SIZE - 1
 };
 
 /*++
@@ -130,10 +129,21 @@ public:
         AllocateMemory
 
         This function attempts to allocate the requested amount of memory from its reserved virtual
-        address space. The function will return NULL if the allocation request cannot
+        address space. The function will return null if the allocation request cannot
         be satisfied by the memory that is currently available in the allocator.
     --*/
     void* AllocateMemory(SIZE_T allocationSize);
+
+    /*++
+    Function:
+        AllocateMemory
+
+        This function attempts to allocate the requested amount of memory from its reserved virtual
+        address space, if memory is available within the specified range. The function will return
+        null if the allocation request cannot satisfied by the memory that is currently available in
+        the allocator.
+    --*/
+    void *AllocateMemoryWithinRange(const void *beginAddress, const void *endAddress, SIZE_T allocationSize);
 
 private:
     /*++
@@ -160,12 +170,13 @@ private:
     // that can be used to calculate an approximate location of the memory that
     // is in 2GB range from the coreclr library. In addition, having precise size of libcoreclr
     // is not necessary for the calculations.
-    const int32_t CoreClrLibrarySize = 100 * 1024 * 1024;
+    static const int32_t CoreClrLibrarySize = 100 * 1024 * 1024;
 
     // This constant represent the max size of the virtual memory that this allocator
     // will try to reserve during initialization. We want all JIT-ed code and the
     // entire libcoreclr to be located in a 2GB range.
-    const int32_t MaxExecutableMemorySize = 0x7FFF0000 - CoreClrLibrarySize;
+    static const int32_t MaxExecutableMemorySize = 0x7FFF0000;
+    static const int32_t MaxExecutableMemorySizeNearCoreClr = MaxExecutableMemorySize - CoreClrLibrarySize;
 
     // Start address of the reserved virtual address space
     void* m_startAddress;
--- a/src/pal/src/include/pal/virtual.h
+++ b/src/pal/src/include/pal/virtual.h
@@ -59,7 +59,8 @@ enum VIRTUAL_CONSTANTS
     VIRTUAL_EXECUTE_READ,
     
     VIRTUAL_PAGE_SIZE       = 0x1000,
-    VIRTUAL_PAGE_MASK       = VIRTUAL_PAGE_SIZE - 1
+    VIRTUAL_PAGE_MASK       = VIRTUAL_PAGE_SIZE - 1,
+    VIRTUAL_64KB            = 0x10000
 };
 
 /*++
@@ -177,9 +178,6 @@ private:
     // entire libcoreclr to be located in a 2GB range.
     static const int32_t MaxExecutableMemorySize = 0x7FFF0000;
     static const int32_t MaxExecutableMemorySizeNearCoreClr = MaxExecutableMemorySize - CoreClrLibrarySize;
-
-    // Start address of the reserved virtual address space
-    void* m_startAddress;
 
     // Next available address in the reserved address space
     void* m_nextFreeAddress;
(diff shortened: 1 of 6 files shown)


kouvel:
No merge, only tried "hello world" so far, need to run some tests under PaX and RANDMMAP

kouvel:
@janvorli @jkotas @gkhanna79, before I test this further, do you think this would be a reasonable approach for 2.0?

kouvel:
@BruceForstall as well ^

janvorli:
@kouvel it sounds like a reasonable approach for now to me. In the future, we could modify the native image mapping in PAL (in MAPMapPEFile) so that it would reserve virtual address space for the jump stubs together with reserving the space for the native image in a single operation. Then we would modify the VirtualReserveFromExecutableMemoryAllocatorWithinRange to use these blocks of virtual address space. 
It seems this would solve all the problems except for the co-location of libcoreclr.so and System.Private.CoreLib.dll native image, which has perf consequences, but no functional issues.

kouvel:
> It seems this would solve all the problems except for the co-location of libcoreclr.so and System.Private.CoreLib.dll native image, which has perf consequences, but no functional issues.

Yes this is one of the things that Bruce suggested as well in the doc, it would solve the native image loading issue but it would not solve the problem with jitted code needing jump stubs in default mode (without setting an environment variable).

src/pal/src/map/virtual.cpp
```diff
@@ -884,8 +886,10 @@ static LPVOID VIRTUALReserveMemory(
 
     // First, figure out where we're trying to reserve the memory and
     // how much we need. On most systems, requests to mmap must be
-    // page-aligned and at multiples of the page size.
-    StartBoundary = (UINT_PTR)lpAddress & ~BOUNDARY_64K;
+    // page-aligned and at multiples of the page size. Unlike on Windows, on
+    // Unix, the allocation granularity is the page size, so the start boundary
+    // and memory size to reserve are not aligned to 64 KB.
+    StartBoundary = (UINT_PTR)lpAddress & ~VIRTUAL_PAGE_MASK;
```
kouvel: Ok thanks for the info. I'll need to instead change the executable memory allocator to reserve blocks with 64 KB granularity.

src/pal/src/map/virtual.cpp
```diff
@@ -2079,22 +2135,56 @@ void ExecutableMemoryAllocator::TryReserveInitialMemory()
...
+    address space. The function will return null if the allocation request cannot
     be satisfied by the memory that is currently available in the allocator.
 
     Note: This function MUST be called with the virtual_critsec lock held.
 
 --*/
 void* ExecutableMemoryAllocator::AllocateMemory(SIZE_T allocationSize)
 {
-    void* allocatedMemory = NULL;
+#ifdef BIT64
```
kouvel: It seems unnecessary to do any additional work to determine that it should return null, it also makes it clear that this path applies to 64-bit systems only.

kouvel:
Made changes from above except where noted

src/pal/src/include/pal/virtual.h
```diff
@@ -178,9 +179,6 @@ class ExecutableMemoryAllocator
     static const int32_t MaxExecutableMemorySize = 0x7FFF0000;
     static const int32_t MaxExecutableMemorySizeNearCoreClr = MaxExecutableMemorySize - CoreClrLibrarySize;
 
-    // Start address of the reserved virtual address space
-    void* m_startAddress;
```
janvorli: I would prefer keeping this in for debugging purposes.

kouvel: Thanks, will fix

src/pal/src/map/virtual.cpp
```diff
@@ -2045,59 +2130,93 @@ void ExecutableMemoryAllocator::TryReserveInitialMemory()
...
+        ALIGN_DOWN(sizeOfAllocation - ((UINT_PTR)m_nextFreeAddress - (UINT_PTR)m_startAddress), VIRTUAL_64KB);
 }
 
 /*++
 Function:
     ExecutableMemoryAllocator::AllocateMemory
 
     This function attempts to allocate the requested amount of memory from its reserved virtual
-    address space. The function will return NULL if the allocation request cannot
+    address space. The function will return null if the allocation request cannot
```
janvorli: A nit: null -> nullptr

kouvel: In comments I usually just refer to "null", it's just unfortunate that the keyword had to be nullptr

kouvel:
In CoreCLR tests, I'm seeing 4 failures and 5 with server GC. None of the failures appear to be related to these changes. I'll file separate issues for the remaining failures. For now, I think this change can be merged.

gkhanna79:
Do you see the failures without your changes?

kouvel:
Let me try without server GC

kouvel:
The same tests are failing without my changes, without server GC. With server GC, CoreLib.dll fails to load half of the time so that doesn't count.

The 4 failing JIT tests allocate a fair bit of stack space and are running into a seg fault for the same reason. Not sure what is wrong yet. The GC test that failed in server GC mode was a one-off, it's passing on reruns.

PAL tests are clean. I'll plan on trying CoreFX tests once the above issues are flushed out.

Issue #11383, opened 2017-05-04
[X86 Profiling] Call of FireEtwMethodLoadVerbose_V1 from ETW::MethodLog::SendHelperEvent with NULL parameters causes segfault on X86 in memcpy


sergign60:
The following call (src/vm/eventtrace.cpp)
```
VOID ETW::MethodLog::SendHelperEvent(ULONGLONG ullHelperStartAddress, ULONG ulHelperSize, LPCWSTR pHelperName)
{
    WRAPPER_NO_CONTRACT;

    if(pHelperName)
    {
         PCWSTR szDtraceOutput1=W("");
         ULONG methodFlags = ETW::MethodLog::MethodStructs::JitHelperMethod; // helper flag set
         FireEtwMethodLoadVerbose_V1(ullHelperStartAddress, 
                                     0, 
                                     ullHelperStartAddress, 
                                     ulHelperSize, 
                                     0, 
                                     methodFlags, 
                                     NULL,
                                     pHelperName, 
                                     NULL,
                                     GetClrInstanceId());
    }
}
```
causes segmentation fault in ```memcpy``` because str is NULL
bin/obj/Linux.x86.Debug/Generated/eventprovider/lttng/eventprovhelpers.cpp
```
bool WriteToBuffer(PCWSTR str, char *&buffer, int& offset, int& size, bool &fixedBuffer)
{
	int byteCount = (PAL_wcslen(str) + 1) * sizeof(*str);

	if (offset + byteCount)
	{
		if (!ResizeBuffer(buffer, size, offset, size + byteCount, fixedBuffer))
			return false;
	}

	memcpy(buffer + offset, str, byteCount);
	offset += byteCount;
	return true;
}
```
The simple changes NULL to  W16_NULLSTRING in SendHelperEvent fix this problem
Just now I can't check this problem on the other platforms. My question is are NULL arguments in SendHelperEvent correct?

sergign60:
CC: @Dmitri-Botcharnikov

sergign60:
The another proposed way is
```
iff --git a/src/scripts/genXplatLttng.py b/src/scripts/genXplatLttng.py
index bacf034..6ff74c9 100644
--- a/src/scripts/genXplatLttng.py
+++ b/src/scripts/genXplatLttng.py
@@ -613,9 +613,9 @@ bool WriteToBuffer(PCWSTR str, char *&buffer, int& offset, int& size, bool &fixe
[...]

PR #11385, opened 2017-05-04
Fixed issue #11383

sergign60:


sergign60:
CC: @Dmitri-Botcharnikov @lucenticus

sergign60:
https://github.com/dotnet/coreclr/issues/11383

src/vm/eventtrace.cpp
```diff
@@ -6760,9 +6762,9 @@ VOID ETW::MethodLog::SendHelperEvent(ULONGLONG ullHelperStartAddress, ULONG ulHe
                                      ulHelperSize, 
                                      0, 
                                      methodFlags, 
-                                     NULL, 
+                                     W16_NULLSTRING,
```
brianrob: Can you tell me more about why NULL is not an OK value here?

sergign60: @brianrob I described the problem in #11383

ramarag: i am not entirely sure if this is necessary will leave it to @brianrob to make the call

ramarag:
LGTM, @leculver do PTAL

leculver:
LGTM as well.

src/vm/eventtrace.cpp
brianrob: @sergig60, my concern is that there are likely to be many call sites that pass NULL for strings.  Your fix already ensures that if a NULL str is passed that it short circuits and never calls memcpy.  Thus, I don't think you need to pass an empty string.

sergign60: @brianrob Ok, If you are sure, I'll return NULL's

PR #11387, opened 2017-05-04
System.IO.Path.GetPathRoot string.Empty or whitespace character string should throw ArgumentException

ViktorHofer:
Relates to https://github.com/dotnet/corefx/issues/19342


src/mscorlib/shared/System/IO/Path.Unix.cs
```diff
@@ -196,7 +196,10 @@ public static bool IsPathRooted(string path)
         public static string GetPathRoot(string path)
         {
             if (path == null) return null;
-            return IsPathRooted(path) ? PathInternal.DirectorySeparatorCharAsString : String.Empty;
+			if (string.IsNullOrWhiteSpace(path))
```
danmosemsft: indents all over this method :)

ViktorHofer: thanks, I always forget to check for that... I think VS is doing that and is ignoring the .editorconfig file at the root.

src/mscorlib/shared/System/IO./Path.Unix.cs
```diff
@@ -0,0 +1,218 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
jkotas: Accidentally added file?

danmosemsft: you have a period in your path after IO

danmosemsft: I'm guessing this file isn't part of your change

src/mscorlib/shared/System/IO/Path.Windows.cs
```diff
@@ -136,10 +136,14 @@ public static bool IsPathRooted(string path)
         // path on the current drive), "X:" (a relative path on a given drive,
         // where X is the drive letter), "X:\" (an absolute path on a given drive),
         // and "\\server\share" (a UNC path for a given server and share name).
-        // The resulting string is null if path is null.
+        // The resulting string is null if path is null. If the path is empty or
```
danmosemsft: same comment in Unix one?

src/mscorlib/shared/System/IO./Path.Unix.cs
ViktorHofer: that's absolutely fantastic. Windows doesn't show directories with a period at the end, not even in bash -,-

danmosemsft: git status :)

src/mscorlib/shared/System/IO/Path.Unix.cs
```diff
@@ -193,10 +193,15 @@ public static bool IsPathRooted(string path)
...
         }
 
+        // The resulting string is null if path is null. If the path is empty or
+        // only contains whitespace characters an ArgumentException gets thrown.
         public static string GetPathRoot(string path)
         {
             if (path == null) return null;
-            return IsPathRooted(path) ? PathInternal.DirectorySeparatorCharAsString : String.Empty;
+			if (string.IsNullOrWhiteSpace(path))
+                throw new ArgumentException(SR.Arg_PathIllegal);
```
safern: Should we add the param name in the exception?

ViktorHofer: First I had it in but then I checked with other occurrences and there the param overload isn't used.

safern: I had the doubt as I'm fixing: https://github.com/dotnet/corefx/issues/19337

cc: @JeremyKuhne @ianhays

ViktorHofer: On ReferenceSources the param isn't used: https://referencesource.microsoft.com/#mscorlib/system/io/path.cs,431

But adding it shouldn't break anything, right?

safern: Adding it shouldn't break anything. I was curious as in various places we don't have param name on desktop (or it is different) and we added it on .NET Core to have a more detailed exception.

JeremyKuhne:
I'm not sure matching desktop here is the best decision. See the other thread. #19342

ViktorHofer:

ViktorHofer:

Issue #5176, opened 2016-05-24
paltest_mapviewoffile_test1 fails when /tmp is mounted as tmpfs


ellismg:
I noticed this problem on Fedora 23 (and @pgavlin had been hitting it as well).

This test wants to `open(2)` a file in `/tmp` which on Fedora is mounted as `tmpfs`, which does not support `O_DIRECT` (`O_DIRECT` is passed because the test specifies [FILE_FLAG_NO_BUFFERING](https://github.com/dotnet/coreclr/blob/master/src/pal/tests/palsuite/filemapping_memmgt/MapViewOfFile/test1/MapViewOfFile.c#L46)).

Because `O_DIRECT` is passed, open returns `EINVA`L, which gets translated into CreateFile returning `INVALID_HANDLE_VALUE`, and the test can not set up what it needs.

ellismg:
@janvorli Thoughts on the best way to handle this?  I think this means we can't get any coverage of O_DIRECT stuff when we run by default on systems where /tmp is tmpfs.

janvorli:
@ellismg I am sorry for a late reply, I have actually written it the day before yesterday and then got distracted and didn't push the comment 
I've found an interesting thread discussing O_DIRECT vs tmpfs:
https://lkml.org/lkml/2007/1/4/55
But I have a question - I don't see the test to access /tmp directly, but rather open just "test.tmp". Do we put the test binaries into the /tmp?

ellismg:
> Do we put the test binaries into the /tmp?

When I looked at it, it seemed like the CWD when the test was run was a folder in /tmp.

janvorli:
I can see it now - we use /tmp so that we can run multiple runs of the tests in parallel.
One option to fix the problem is to modify the test to create the file /var/tmp instead of /tmp. The /var/tmp is a real filesystem and so it should not have the problem of the tmpfs. Or we may change the runpaltests.sh to use /var/tmp instead of /tmp for all tests.

ellismg:
The other option (and I'm not sure if it's preferable or not) is to just use $TEMP if it is set instead of /tmp and we could have our builds set $TEMP to some non tmpfs thing (I guess by default /var/tmp persists across reboots which may mean we should be more careful about dropping files there?)

sergiy-k:
[...]

PR #11393, opened 2017-05-04
Reenable MapViewOfFile PAL test1

janvorli:
This change reenables the test that was having problems on systems when the /tmp
is created on tmpfs file system which doesn't support direct access (unbuffered).
The fix is to explicitly use /var/tmp instead (and /data/local/tmp/ on Android).


janvorli:
Fixes #5176

janvorli:
CC: @ellismg, @gkhanna79

Issue #11381, opened 2017-05-04
Incorrect metadata on SOS.NETCore.dll


morganbr:
SOS.NETCore.dll has incorrect assembly attributes and file metadata. This causes the .NET Native compiler to think it's not a framework assembly as well as looking wrong. The incorrect attributes I spotted are:
```
[assembly: AssemblyDefaultAlias("System.Private.CoreLib")]
[assembly: AssemblyDescription("System.Private.CoreLib")]
[assembly: AssemblyTitle("System.Private.CoreLib")]
```
It also is missing this attribute, which would fix the .NET Native problem:
```
[assembly: AssemblyMetadata(".NETFrameworkAssembly", "")]
```

The file metadata has File description "System.Private.CoreLib"

mikem8361:
@morganbr msbuild is complaining about the dot "." in the above AssemblyMetadata attribute.

C:\ssd\coreclr2\src\ToolBox\SOS\NETCore\SOS.NETCore.csproj(39,62): error MSB4025: The project file could not be loaded. Name cannot begin with the '.' character, hexadecimal value 0x2E. Line 39,
position 62.

mikem8361:
Never mind.  I was using the wrong msbuild AssemblyInfoLines syntax to add this attribute.

mikem8361:
It looks like the other attributes have already been fixed in 2.0. PR #11395 adds the AssemblyMetadata attribute.

mikem8361:
Fixed with PR #11395

PR #11395, opened 2017-05-04
Fix incorrect metadata on SOS.NETCore.dll

mikem8361:
Issue #11381

All the other assembly attributes are correct in 2.0.  Added AssemblyMetadata.

### DIFF
--- a/src/ToolBox/SOS/NETCore/SOS.NETCore.csproj
+++ b/src/ToolBox/SOS/NETCore/SOS.NETCore.csproj
@@ -36,6 +36,10 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


morganbr:
@weshaggard , is this the right way to handle this? I also noticed https://github.com/dotnet/buildtools/blob/03988127f489bcd1cf58d71131ac628620d53de9/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets#L34 , which has AssemblyMetadata as an ItemGroup item.

weshaggard:
Setting IsDotNetFrameworkProductAssembly is the correct way to get this metadata added as @morganbr pointed out.

Issue #11114, opened 2017-04-20
Environment.FailFast should dump exception details to console


danmosemsft:
Environment.FailFast writes the message (and exception details if provided) to the Windows event log and passes them to Watson. Neither of these go to the developer's console nor to the XUnit logs. Now we have made Debug.Assert do FailFast, this is hindering debugging and test analysis.

We should also write to stderr so if it's a console app you see the details. This should include the message and exception details, with the callstack (hopefully that is just a string inside the exception when it reaches native code)

@jkotas any gotchas here?

danmosemsft:
It should obviously do this with the minimum of ceremony, since this is a failure path. I see the VM already has `PrintToStdErr()` but I don't know if any work is needed to format the message/stack.

jkotas:
All sounds good to me.

danmosemsft:
@ViktorHofer do you want to take a shot at this?

ViktorHofer:
Yeah I will look into it

tarekgh:
@ViktorHofer if you are not looking at this one, I can help look at it tomorrow. please let me know

ViktorHofer:
I will take a look at it tomorrow and will let you know if I need some help. Thanks!

ViktorHofer:
SystemNative::GenericFailFast in "src\classlibnative\bcltype\system.cpp" needs to print the exception message with callstack to stderr and if passed the inner exception. I will talk with @jkotas about it.

jkotas:
I think it should be done in  `EEPolicy::HandleFatalError` so that all sources of failfasts get logged. For example, we want to make sure that there is error message printed for fail fast caused by this program:

```
class My {

static unsafe int Main() {
    int * tmp = (int*)0x1234567890;
    return *tmp;
}

}
```

ViktorHofer:
I'm looking into this right now. I will stick with just printing out the error message to stderr for now.

@gkhanna79 Any idea how to get the StackTrace at that point of execution?

gkhanna79:
See DoReportForUnhandledException on how to compute the string version of the stacktrace.

ViktorHofer:
Perfect, thanks!

PR #11397, opened 2017-05-04
FailFast stderr logging with stacktrace

ViktorHofer:
Fixes https://github.com/dotnet/coreclr/issues/11114

I gave my best and hope that I didn't screw it completely up ;)

@jkotas Please take a look. Especially the part where it logs it to stderr as I'm not sure if two calls to `PrintToStdErrW` are what we want and if calling `GetUnicode` on `StackSString` is the right thing to do. Thanks!

### DIFF
--- a/src/vm/eepolicy.cpp
+++ b/src/vm/eepolicy.cpp
@@ -1129,13 +1129,14 @@ StackWalkAction LogCallstackForLogCallback(
 
     TypeString::AppendMethodInternal(str, pMD, TypeString::FormatNamespace|TypeString::FormatFullInst|TypeString::FormatSignature); 
     PrintToStdErrW(str.GetUnicode());
+    PrintToStdErrA("\n");
 
     return SWA_CONTINUE;
 }
 
 //---------------------------------------------------------------------------------------
 //
-// A woker to save managed stack trace.
+// A worker to save managed stack trace.
 //
 // Arguments:
 //    reporter - EventReporter object for EventLog
@@ -1173,13 +1174,16 @@ void LogCallstackForLogWorker()
 // Return Value:
 //    None
 //
-void DoLogForUnhandledException(PEXCEPTION_POINTERS pExceptionInfo)
+void DoLogForUnhandledException(LPCWSTR pszMessage, PEXCEPTION_POINTERS pExceptionInfo)
 {
     WRAPPER_NO_CONTRACT;
 
     Thread *pThread = GetThread();
     EX_TRY
     {
+        PrintToStdErrW((WCHAR*)pszMessage);
+        PrintToStdErrA("\n");
+
         StackSString s;
         InlineSString<80> ssErrorFormat;
         if(!ssErrorFormat.LoadResource(CCompRC::Optional, IDS_ER_UNHANDLEDEXCEPTIONINFO))
@@ -1192,6 +1196,7 @@ void DoLogForUnhandledException(PEXCEPTION_POINTERS pExceptionInfo)
         PrintToStdErrW((WCHAR*)s.GetUnicode());
         if (pThread)
         {
+            PrintToStdErrA("\n");
             LogCallstackForLogWorker();
         }
     }
@@ -1216,9 +1221,7 @@ void EEPolicy::LogFatalError(UINT exitCode, UINT_PTR address, LPCWSTR pszMessage
     // Log FailFast exception to StdErr
     if (exitCode == (UINT)COR_E_FAILFAST)
     {
-        PrintToStdErrW((WCHAR*)pszMessage);
-        PrintToStdErrA("\n"); // This will result in blank line
-        DoLogForUnhandledException(pExceptionInfo);
+        DoLogForUnhandledException(pszMessage, pExceptionInfo);
     }
 
     if(ETW_EVENT_ENABLED(MICROSOFT_WINDOWS_DOTNETRUNTIME_PRIVATE_PROVIDER_Context, FailFast))
--- a/src/vm/eepolicy.cpp
+++ b/src/vm/eepolicy.cpp
@@ -1144,7 +1144,7 @@ StackWalkAction LogCallstackForLogCallback(
 // Return Value:
 //    None
 //
-void LogCallstackForLogWorker()
+inline void LogCallstackForLogWorker()
 {
     Thread* pThread = GetThread();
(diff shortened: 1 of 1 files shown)


src/vm/eepolicy.cpp
```diff
@@ -1105,6 +1105,42 @@ void EEPolicy::HandleExitProcess(ShutdownCompleteAction sca)
...
+        if(!ssErrorFormat.LoadResource(CCompRC::Optional, IDS_ER_UNHANDLEDEXCEPTIONINFO))
+            ssErrorFormat.Set(W("exception code %1, exception address %2"));
+        SmallStackSString exceptionCodeString;
+        exceptionCodeString.Printf(W("%x"), pExceptionInfo->ExceptionRecord->ExceptionCode);
+        SmallStackSString addressString;
+        addressString.Printf(W("%p"), (UINT_PTR)pExceptionInfo->ExceptionRecord->ExceptionAddress);
+        s.FormatMessage(FORMAT_MESSAGE_FROM_STRING, (LPCWSTR)ssErrorFormat, 0, 0, exceptionCodeString, addressString);
+        LPCWSTR stackMsg = s.GetUnicode();
+
+        PrintToStdErrW((WCHAR*)pszMessage);
```
jkotas: I think the printing of `pszMessage` should be done as the very first thing. The system is potentially in unstable state when this method is called, and so the message formatting can crash. It would be nice to print at least something before the crash.

src/vm/eepolicy.cpp
```diff
@@ -1117,6 +1153,9 @@ void EEPolicy::LogFatalError(UINT exitCode, UINT_PTR address, LPCWSTR pszMessage
 
     _ASSERTE(pExceptionInfo != NULL);
 
+    // Log exception to StdErr
```
jkotas: Is this going to kick for regular unhandled exceptions? I believe we have a good message for regular unhandled exceptions already. This won't make them better.

Maybe this needs to be in a slightly different place so that it kicks in for failfasts only.

ViktorHofer:
PTAL. Is the only way to test this to fully compile it and then produce a FailFast error?

jkotas:
> Is the only way to test this to fully compile it and then produce a FailFast error?

Yes.

ViktorHofer:
```
C:\Users\vikto\Documents\src\coreclr-demo\runtime>corerun hello.exe
Asbjorn, let me go!exception code 80131623, exception address 00007FF8889067B0
```

That's not pretty 👎

ViktorHofer:
Updated result:
```
C:\Users\vikto\Documents\src\coreclr-demo\runtime>corerun hello.exe
Asbjorn, let me go!
exception code 80131623, exception address 00007FF8889067B0   at System.Environment.FailFast(System.String, System.Exception)   at Internal.Runtime.Augments.EnvironmentAugments.FailFast(System.String, System.Exception)   at System.Environment.FailFast(System.String, System.Exception)   at System.Environment.FailFast(System.String)   at Program.Main()
```

src/vm/eepolicy.cpp
```diff
@@ -1105,6 +1105,102 @@ void EEPolicy::HandleExitProcess(ShutdownCompleteAction sca)
...
+
+    TypeString::AppendMethodInternal(str, pMD, TypeString::FormatNamespace|TypeString::FormatFullInst|TypeString::FormatSignature); 
+    PrintToStdErrW(str.GetUnicode());
+
+    return SWA_CONTINUE;
+}
+
+//---------------------------------------------------------------------------------------
+//
+// A woker to save managed stack trace.
```
danmosemsft: typo, worker

ViktorHofer:
@jkotas should I leave the exception code and exception address in or remove it?

src/vm/eepolicy.cpp
```diff
@@ -1117,6 +1213,14 @@ void EEPolicy::LogFatalError(UINT exitCode, UINT_PTR address, LPCWSTR pszMessage
 
     _ASSERTE(pExceptionInfo != NULL);
 
+    // Log FailFast exception to StdErr
+    if (exitCode == (UINT)COR_E_FAILFAST)
+    {
+        PrintToStdErrW((WCHAR*)pszMessage);
```
danmosemsft: maybe put these 2 lines inside DoLogForUnhandledException?

src/vm/eepolicy.cpp
```diff
@@ -1105,6 +1105,102 @@ void EEPolicy::HandleExitProcess(ShutdownCompleteAction sca)
...
+//
+// A woker to save managed stack trace.
+//
+// Arguments:
+//    reporter - EventReporter object for EventLog
+//
+// Return Value:
+//    None
+//
+void LogCallstackForLogWorker()
```
danmosemsft: This function could be inlined?

danmosemsft:
@jkotas I assume line numbers will never appear in this stack ?

jkotas:
> @jkotas I assume line numbers will never appear in this stack ?

Right - with this low-level formatting routine.

jkotas:
> @jkotas should I leave the exception code and exception address in or remove it?

The exception code and exception address is not very useful for FailFast called from managed code case. But I think it is useful for the low-level crash in the runtime. Maybe the two cases should be handled differently.

ViktorHofer:
Each frame is now printed in a line.

```
C:\Users\vikto\Documents\src\coreclr-demo\runtime>corerun hello.exe
Asbjorn, let me go!
exception code 80131623, exception address 00007FFBC5F967B0
   at System.Environment.FailFast(System.String, System.Exception)
   at Internal.Runtime.Augments.EnvironmentAugments.FailFast(System.String, System.Exception)
   at System.Environment.FailFast(System.String, System.Exception)
   at System.Environment.FailFast(System.String)
   at Program.Main()
```

ViktorHofer:
@danmosemsft your issue is now fixed with this PR. What isn't included (yet), is Jans sample repro of a silent FatalException as this isn't a FailFast situation and has a completely different error code...

@jkotas I would tackle your repro in a separate PR as IMO it isn't connected to FailFast.

danmosemsft:
sounds fine to me.

ViktorHofer:

ViktorHofer:
ping @jkotas 

> I would tackle your repro in a separate PR as IMO it isn't connected to FailFast.

PR #11398, opened 2017-05-04
Fix clang version detection in src/pal/tools/gen-buildsys-clang.sh

janvorli:
The version detection that was added recently for the purpose of selecting
optimization options for ARM builds was not working properly when the
clang binary doesn't have version in its name. This resulted in syntax
error being reported on OSX and some Linux distros:
`/Users/janvorli/git/coreclr/src/pal/tools/gen-buildsys-clang.sh:
line 162: : /usr/bin/clang: syntax error: operand expected (error token is "/usr/bin/clang")`

### DIFF
--- a/src/pal/tools/gen-buildsys-clang.sh
+++ b/src/pal/tools/gen-buildsys-clang.sh
@@ -155,7 +155,7 @@ if [ "$build_arch" == "armel" ]; then
     cmake_extra_defines="$cmake_extra_defines -DARM_SOFTFP=1"
 fi
 
-clang_version=$(echo $CC | awk -F- '{ print $NF }')
+clang_version=$( $CC --version | head -1 | sed 's/[^0-9]*\([0-9]*\.[0-9]*\).*/\1/' )
 # Use O1 option when the clang version is smaller than 3.9
 # Otherwise use O3 option in release build
 if [[ ( ${clang_version%.*} -eq 3  &&  ${clang_version#*.} -lt 9 ) &&


janvorli:
CC: @jyoungyun, @gkhanna79

jyoungyun:
@janvorli Thank you for revising it. How about using the command below to extract clang version number from string? It would be better to use the `clang --version` command to get the correct version of clang.
```
clang_version=$( $CC --version | head -1 | sed 's/[^0-9]*\([0-9.].[0-9.]\).*/\1/' )
```

janvorli:
@jyoungyun thank you for the suggestion, it is better this way. I've just modified your regex a little to support clang version components higher than 9. LLVM has now switched the versioning to bump the major version for each new release, so it can reach 10 in the major version soon.

jyoungyun:
@janvorli mybuild.sh is uploaded together :) Thank you for taking my suggestion.

janvorli:
@jyoungyun ugh, thank you for spotting that I have accidentally added my private file. Fixed now.

gkhanna79:
Can this be merged @janvorli ?

janvorli:
@gkhanna79 I didn't get a review for it yet, but otherwise it is ready to get merged.

gkhanna79:
@adityamandaleeka Can you please review it today so that we can get this in before EOD?

adityamandaleeka:
This is fine since it's being used for just one purpose for now, but just FYI, the version number on Mac for the built in `clang` was empty before (since the command itself doesn't contain the version), but with the new method, it will return something like 8.0 or higher.

janvorli:
@adityamandaleeka good point.

PR #10536, opened 2017-03-29
Adding scenario-based tests

guhuro:
Modified runner to run Scenario Benchmarks
Added Harnessing for JITBench
Updating xunit-performance packages

PR #11400, opened 2017-05-04
Add JitBench scenario testing.

jorive:
Rebasing original https://github.com/dotnet/coreclr/pull/10536


jorive:
PTAL
@dotnet/rap-team

DrewScoggins:
We should change the directory structure of this I think.  We should have JitBench live in a directory called something like scenarioTests or something, not too attached on a name.  This will allow people to have a place to add scenario tests and have our automation automatically pick them up.

tests/src/performance/JitBench/JitBenchHarness.cs
```diff
@@ -0,0 +1,355 @@
...
+            PrintHeader("Starting SETUP");
+
+            var dotnetDirectory = Path.Combine(s_jitBenchDevDirectory, ".dotnet");
+            var dotnetProcessFileName = Path.Combine(dotnetDirectory, "dotnet.exe");
+
+            //Download and extract the repo.
+            using (var client = new HttpClient())
+            {
+                // Download the JitBench repository and extract it.
+                var url = @"https://github.com/guhuro/JitBench/archive/dev.zip"; // TODO: This should be updated to https://github.com/aspnet/JitBench/archive/dev.zip once the fork is merged.
```
DrewScoggins: We should check to see if this works now with the version in the ASP.NET repo.

tests/src/performance/JitBench/JitBenchHarness.cs
```diff
@@ -0,0 +1,355 @@
...
+
+                // This step will create s_JitBenchDevDirectory.
+                ZipFile.ExtractToDirectory(zipFile, s_temporaryDirectory);
+            }
+
+            // first dotnet-install.ps1
+            var installProcessStartInfo = new ProcessStartInfo() {
+                WorkingDirectory = s_jitBenchDevDirectory,
+                FileName = @"powershell.exe",
+                Arguments = @".\Dotnet-Install.ps1 -SharedRuntime -InstallDir .dotnet -Channel master -Architecture x64"
```
DrewScoggins: We should ensure that these are still the instructions for running the JitBench scenario.  I know there has been some churn on how the thing runs.

tests/src/performance/JitBench/JitBenchHarness.cs
jorive: In the meanwhile, I think we should clone this to one of our repos.

jorive: This repo has not taken any updates, so it should still work.

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
```diff
@@ -0,0 +1,355 @@
+using CommandLine;
```
jkotas: Missing license header

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
```diff
@@ -0,0 +1,355 @@
...
+                    parserSettings.HelpWriter = new StringWriter();
+                    parserSettings.IgnoreUnknownArguments = true;
+                });
+                var result = parser.ParseArguments<JitBenchHarnessOptions>(new string[] { "--help" });
+
+                var helpTextString = new HelpText {
+                    AddDashesToOption = true,
+                    AddEnumValuesToHelpText = true,
+                    AdditionalNewLineAfterOption = false,
+                    Copyright = "Copyright (c) Microsoft Corporation 2015",
```
jkotas: I do not think that we print MS Copyright in the open source tools.

jorive:
PTAL @dotnet/rap-team

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
```diff
@@ -0,0 +1,485 @@
...
+
+        private static void RestoreMusicStore(string workingDirectory, string dotnetFileName, IDictionary<string, string> environment)
+        {
+            var psi = new ProcessStartInfo() {
+                WorkingDirectory = workingDirectory,
+                FileName = dotnetFileName,
+                Arguments = "restore"
+            };
+
+            LaunchProcess(psi, 300000, environment);
```
adiaaida: if 30000 is some sort of timeout, we should have it as either a global variable or have it passed as a parameter, in case we want to increase/decrease it, so we only have to update in one place.

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
```diff
@@ -0,0 +1,485 @@
...
+
+            var dotnetProcessFileName = Path.Combine(s_jitBenchDevDirectory, ".dotnet", "dotnet.exe");
+            var musicStoreDirectory = Path.Combine(s_jitBenchDevDirectory, "src", "MusicStore");
+
+            RestoreMusicStore(musicStoreDirectory, dotnetProcessFileName, environment);
+            PublishMusicStore(musicStoreDirectory, dotnetProcessFileName, environment);
+
+            var psi = new ProcessStartInfo() {
+                FileName = "cmd.exe",
+                Arguments = $"/C \"{dotnetProcessFileName} MusicStore.dll 1>{MusicStoreRedirectedStandardOutputFileName}\"",
```
adiaaida: We should use ProcessStartInfo.RedirectStandardOutput here too

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
```diff
@@ -0,0 +1,485 @@
...
+                Directory.Move(Path.Combine(s_temporaryDirectory, $"JitBench-{sha1Id}"), s_jitBenchDevDirectory);
+            }
+        }
+
+        private static IDictionary<string, string> InstallSharedRuntime()
+        {
+            var psi = new ProcessStartInfo() {
+                WorkingDirectory = s_jitBenchDevDirectory,
+                FileName = @"powershell.exe",
+                Arguments = @".\Dotnet-Install.ps1 -SharedRuntime -InstallDir .dotnet -Channel master -Architecture x64"
```
jorive: Can we start with Windows first? This is a port of an old PR, and it would be nice to start getting numbers through automation :)

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
```diff
@@ -0,0 +1,485 @@
...
+
+        private static void ModifySharedFramework()
+        {
+            // Current working directory is the <coreclr repo root>/sandbox directory.
+            Console.WriteLine($"Modifying the shared framework.");
+
+            var sourcedi = new DirectoryInfo(Directory.GetCurrentDirectory());
+            var targetdi = new DirectoryInfo(
+                new DirectoryInfo(Path.Combine(s_jitBenchDevDirectory, ".dotnet", "shared", "Microsoft.NETCore.App"))
+                .GetDirectories("2.1*").First().FullName);
```
jorive: Currently, this harness downloads a fixed version of JitBench, so the version would not move behind the scenes, and I agree we should port this to the release/2.0.0 branch.

jorive: Maybe JitBench should be a benchmark under the `dotnet.exe` repo? so we would not need to patch/hack the Clr on top of the downloaded cli. Wouldn't it be more robust?

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
```diff
@@ -0,0 +1,485 @@
...
+        }
+
+        private static IDictionary<string, string> GenerateStore(IDictionary<string, string> environment)
+        {
+            // This step generates some environment variables needed later.
+            var environmentFileName = "JitBenchEnvironment.txt";
+            var psi = new ProcessStartInfo() {
+                WorkingDirectory = s_jitBenchDevDirectory,
+                FileName = "powershell.exe",
+                Arguments = $"-Command \".\\AspNet-GenerateStore.ps1 -InstallDir .store -Architecture x64 -Runtime win7-x64; gi env:JITBENCH_*, env:DOTNET_SHARED_STORE | %{{ \\\"$($_.Name)=$($_.Value)\\\" }} 1>>{environmentFileName}\""
```
jorive: Currently, there is no way to get this back from the Api/Process. This is a workaround. Maybe we should open an issue.

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
jorive: Maybe a config file? Each process have different timeout values.

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
jorive: Same as above comment.

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
adiaaida: That's fine. We'll also want to run this on x86, so we'll want to parameterize Architecture.

adiaaida: we can discover the architecture of the built components by running dumpbin /headers clrjit.dll in a vcvars env.

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
adiaaida: I saw 30000 a few times, but if most of them are different, this is fine.

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
```diff
@@ -0,0 +1,487 @@
...
+
+            LaunchProcess(psi, 300000, environment);
+        }
+
+        private static void PublishMusicStore(string workingDirectory, string dotnetFileName, IDictionary<string, string> environment)
+        {
+            var psi = new ProcessStartInfo() {
+                WorkingDirectory = workingDirectory,
+                FileName = "cmd.exe",
+                Arguments = $"/C \"{dotnetFileName} publish -c Release -f netcoreapp2.0 --manifest %JITBENCH_ASPNET_MANIFEST% /p:MvcRazorCompileOnPublish=false\""
```
dpodder: We should hoist `netcoreapp2.0` as a constant that we may need to update at the same time as the checksum of the JitBench repo, as it might change moving forward when we update the test sources.

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
```diff
@@ -0,0 +1,487 @@
...
+                }
+            }
+
+            using (var p = new Process() { StartInfo = processStartInfo })
+            {
+                p.Start();
+                if (p.WaitForExit(timeoutMilliseconds) == false)
+                {
+                    // FIXME: What about clean/kill child processes?
+                    p.Kill();
```
dpodder: Good call. Given that this is likely something we need across many scenario benchmarks in the future, though, would it make sense to add it as publicly-available utility functionality in XUnit-Performance itself? (We can deprecate it in the future if a future version of the framework picks up support for doing a clean `KillAll()`).

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
```diff
@@ -0,0 +1,487 @@
...
+                    settings.IgnoreUnknownArguments = true;
+                }))
+                {
+                    JitBenchHarnessOptions options = null;
+                    parser.ParseArguments<JitBenchHarnessOptions>(args)
+                        .WithParsed(parsed => options = parsed)
+                        .WithNotParsed(errors => {
+                            foreach (Error error in errors)
+                            {
+                                switch (error.Tag)
```
dpodder: Why do we need this custom error handling?

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
jorive: Yes, I might add it as an extension under the xUnit-Performance Api.

jorive:
@dotnet-bot test ci please

jorive:
@dotnet-bot test ci please

jorive:
@dotnet-bot test ci please

jorive:
@dotnet-bot test ci please

jorive:
@dotnet-bot test ci please

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
```diff
@@ -0,0 +1,487 @@
...
+        {
+            var psi = new ProcessStartInfo() {
+                WorkingDirectory = s_jitBenchDevDirectory,
+                FileName = @"powershell.exe",
+                Arguments = $".\\Dotnet-Install.ps1 -SharedRuntime -InstallDir .dotnet -Channel master -Architecture {s_targetArchitecture}"
+            };
+            LaunchProcess(psi, 180000);
+
+            // TODO: This is currently hardcoded, but we could probably pull it from the powershell cmdlet call.
+            return new Dictionary<string, string> { { "PATH", $"{Path.Combine(s_jitBenchDevDirectory, ".dotnet")};{psi.Environment["PATH"]}" } };
```
adiaaida: agreed

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
adiaaida: I'm actually ok with the framework/sdk moving. This is true for all of our perf testing. However, I would like us to be more general with choosing which version we are choosing to move coreclr/clrjit into, as Daniel has suggested.

jorive:
@dotnet-bot test ci please
@dotnet-bot test Windows_NT x64 illink perf

jorive:
@dotnet-bot help

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
jorive: This was brought from the original PR, and it might have been a port from a different project.

---
In reply to: [129456611](https://github.com/dotnet/coreclr/pull/11400#discussion_r129456611) [](ancestors = 129456611)

jorive:
@dotnet/rap-team ping

jorive:
@dotnet-bot test ci please

jorive:
@dotnet-bot test Windows_NT x64 Release illink

jorive:
@dotnet-bot test ci please
@dotnet-bot test Windows_NT x64 Release illink

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
```diff
@@ -0,0 +1,489 @@
...
+                new DirectoryInfo(Path.Combine(s_jitBenchDevDirectory, ".dotnet", "shared", "Microsoft.NETCore.App"))
+                .GetDirectories("*")
+                .OrderBy(s => s.Name)
+                .Last()
+                .FullName);
+
+            Console.WriteLine($"  Source : {sourcedi.FullName}");
+            Console.WriteLine($"  Target : {targetdi.FullName}");
+
+            var compiledBinariesOfInterest = new string[] {
```
brianrob: We're doing a patch of the compiled runtime here, which is something that can be fragile going forward.  We should surface this as an issue with the eng team and ask what the right way to do this is - perhaps a full-stack build.

perf.groovy
```diff
@@ -578,9 +568,9 @@ parallel(
 
             if (isPR) {
                 TriggerBuilder builder = TriggerBuilder.triggerOnPullRequest()
-                builder.setGithubContext("${os} ${arch} ILLink Perf Tests")
+                builder.setGithubContext("${os} ${arch} Perf Scenarios Tests")
```
brianrob: Nit: Expand Perf to Performance.

perf.groovy
```diff
@@ -420,7 +408,7 @@ def static getFullThroughputJobName(def project, def os, def isPR) {
 
             def osGroup = getOSGroup(os)
             def runType = isPR ? 'private' : 'rolling'
-            def benchViewName = isPR ? 'coreclr private \$BenchviewCommitName' : 'coreclr rolling \$GIT_BRANCH_WITHOUT_ORIGIN \$GIT_COMMIT'
+            def benchViewName = isPR ? 'coreclr-throughput private \$BenchviewCommitName' : 'coreclr-throughput rolling \$GIT_BRANCH_WITHOUT_ORIGIN \$GIT_COMMIT'
```
brianrob: Perhaps I missed it, but is this always a throughput run?  Also, the comment right above this for the input parameter BenchViewCommitName is no longer correct with this change.

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
jorive: I agree. This is very fragile, and I think this could optionally be a test under dotnet-cli too.

---
In reply to: [130418028](https://github.com/dotnet/coreclr/pull/11400#discussion_r130418028) [](ancestors = 130418028)

perf.groovy
jorive: Yes, this is always a TP run.

---
In reply to: [130419578](https://github.com/dotnet/coreclr/pull/11400#discussion_r130419578) [](ancestors = 130419578)

tests/src/performance/Scenario/JitBench/JitBenchHarness.cs
dpodder: There's some value in having the test here, instead of in CLI; namely, we can test the effects of changes on the runtime and JIT much more quickly, before the changes are ingested by a full-stack build. We've had to do this in a couple of other places (e.g.: optdata creation), so at a minimum I think it would be worthwhile to script this patching separately, and have a single, well-tested, CI-verified script that we could just call to produce a CLI drop out of a runtime build. If we can do something less fragile (e.g., move all such validation to a full-stack build), then I'm all for it.

brianrob: Let's file an issue and get an evaluation on it from the eng folks.  Of course, don't block this change on it.

perf.groovy
brianrob: Ok, great.  Thanks.

perf.groovy
```diff
@@ -578,9 +568,9 @@ parallel(
 
             if (isPR) {
                 TriggerBuilder builder = TriggerBuilder.triggerOnPullRequest()
-                builder.setGithubContext("${os} ${arch} ILLink Perf Tests")
+                builder.setGithubContext("${os} ${arch} Performance Scenarios Tests")
                 builder.triggerOnlyOnComment()
-                builder.setCustomTriggerPhrase("(?i).*test\\W+${os}\\W+${arch}\\W+illink\\W+perf.*")
+                builder.setCustomTriggerPhrase("(?i).*test\\W+${os}\\W+${arch}\\W+scenarios.*")
```
adiaaida: I'd like the trigger to still have perf in it, so that we know we are running perf testing.

jorive:
@dotnet-bot test ci please

PR #11403, opened 2017-05-04
Path.GetDirectoryName() throw exception when path is empty or has only white spaces

safern:
Fixes: https://github.com/dotnet/corefx/issues/19337

Tests added on: https://github.com/dotnet/corefx/pull/19343

cc: @danmosemsft

### DIFF
--- a/src/mscorlib/shared/System/IO/Path.cs
+++ b/src/mscorlib/shared/System/IO/Path.cs
@@ -76,19 +76,27 @@ namespace System.IO
         // "\\server\share").
         public static string GetDirectoryName(string path)
         {
-            if (path != null)
+            if (path == null)
             {
-                PathInternal.CheckInvalidPathChars(path);
-                path = PathInternal.NormalizeDirectorySeparators(path);
-                int root = PathInternal.GetRootLength(path);
-
-                int i = path.Length;
-                if (i > root)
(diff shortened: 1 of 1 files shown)


safern:

builds timed out.

src/mscorlib/shared/System/IO/Path.cs
```diff
@@ -76,19 +76,27 @@ public static string ChangeExtension(string path, string extension)
...
-                int i = path.Length;
-                if (i > root)
-                {
-                    while (i > root && !PathInternal.IsDirectorySeparator(path[--i])) ;
-                    return path.Substring(0, i);
-                }
+            if (string.IsNullOrWhiteSpace(path))
+            {
+                throw new ArgumentException(SR.Arg_PathIllegal, nameof(path));
+            }
```
stephentoub: It'd be better to make this something like:
```C#
if (string.IsNullOrWhitespace(path))
{
    if (path == null) return null;
    throw new ArgumentException(SR.Arg_PathIllegal, nameof(path));
}
```
That way we only do the null check on path once in the common path rather than twice.

safern: Fixed! Good thinking, thanks!

PR #11411, opened 2017-05-04
Use ReferenceEquals instead of Delegate.op_Equality for sentinel checks

justinvp:
The reference equality check is more efficient than `Delegate.op_Equality`.

cc: @stephentoub, @JeremyKuhne, @ianhays


benaadams:
Yeah `Action` `==` non-inlines via virtual `MulticastDelegate.Equals` :sad`

justinvp:
Related: https://github.com/dotnet/corefx/pull/19371

PR #11414, opened 2017-05-05
Enable build on old Linux

janvorli:
This change enables building coreclr on old Linux distros that
don't have PR_SET_PTRACER and don't have _xstate struct in the
standard sigcontext.h header.


janvorli:
@adityamandaleeka I had a wrong include in the configure.cmake for detecting the presence of the _xstate. Fixed now.

janvorli:

PR #11428, opened 2017-05-05
Fix typo in perf.groovy

DrewScoggins:
Had a typo where I was passing deafult instead of default as a flag to
the collection harness.  This fixes that.


DrewScoggins:
PTAL @jorive

jorive:
LGTM

PR #11433, opened 2017-05-06
Log Events to EventPipe on Linux

brianrob:
Adds the ability to log ETW events to the EventPipe on Linux.  This results in a binary file that contains the events and can be read using the FastSerialization format.


brianrob:
@vancem, PTAL.  This is the same code that you looked at yesterday.

@valenis, FYI.

vancem:
I have some feedback I sent Brian, but I am OK with doing that after merge.

PR #11436, opened 2017-05-07
Finish deleting dead CAS code from CoreLib

### LINKED ISSUE #9321: Remove CAS from CoreLib

jkotas:
Fixes #9321 and deletes CleanupToDoList.cs


src/mscorlib/src/System/Reflection/Emit/AssemblyBuilder.cs
```diff
@@ -38,7 +38,6 @@ namespace System.Reflection.Emit
     using System.Runtime.Serialization;
```
danmosemsft: comment at the top of this file refers to low trust code ,may need updating.

danmosemsft: Nevermind, I guess.., I see countless other remnants of what looks CAS related. Just searching for "partial.*trust", grants, "full.*trust" and the like. It will probably take a long time to see the last of that and it's best to take it in manageable bites.

danmosemsft:
@jkotas since there's a lot of churn here and it's basically cleanup, what do you think of holding off merging for a few days, to let 2.0 get out the branch? If so we should label no-merge.

jkotas:
Agree.

jkotas:
CoreCLR.dll x64 release is 110k smaller

jkotas:
@dotnet-bot test test Windows_NT x64 corefx_baseline

jkotas:

PR #11437, opened 2017-05-07
Add dynamic method unload event to Profiler API

mjsabby:
The PR contains a dynamic method unload event that is fired whenever a dynamic method is destroyed. 

This is needed because profilers can listen to dynamic method load events (via DynamicMethodJittingStarted/Finished) but previously would never be notified of destroyed dynamic methods. This can happen for example when the dynamic method gets GC'd.

In essence this is a bug fix since profilers would never receive dynamic unload events and could run out of memory.


mjsabby:
@brianrob @noahfalk @vancem please take a look. it's mostly isolated, but I needed to get thread id and timestamp so simple changes in the main event pipe infra.

src/vm/runtimehandles.cpp
```diff
@@ -2615,6 +2615,10 @@ void QCALLTYPE RuntimeMethodHandle::Destroy(MethodDesc * pMethod)
     // Fire Unload Dynamic Method Event here
     ETW::MethodLog::DynamicMethodDestroyed(pMethod);
 
+    BEGIN_PIN_PROFILER(CORProfilerTrackJITInfo());
```
noahfalk: This event feels like it would merit a new flag rather than re-using the JIT compilation flag. Maybe COR_PRF_MONITOR_DYNAMIC_FUNCTION_UNLOADS. Conceptually the lifetime of the code which backs a method could be different than the method itself, even though today its pretty close to 1:1.

We already have a COR_PRF_MONITOR_FUNCTION_UNLOADS, which is unused, but I hesitate to re-use it because we might trigger annoying compat issues and flag bits are cheap.

noahfalk:
lgtm, just waiting to hear back from Brian and for David Mason to give it the green light using some additional internal tests we've got.

src/inc/corprof.idl
```diff
@@ -2408,6 +2412,31 @@ interface ICorProfilerCallback8 : ICorProfilerCallback7
...
+]
+interface ICorProfilerCallback9 : ICorProfilerCallback8
+{
+    // This event is triggered whenever a dynamic method is garbage collected
+    // and subsequently unloaded.
+
+    HRESULT DynamicMethodUnloaded([in] FunctionID functionId);
+
+    HRESULT EventPipeEventDelivered(
+        [in] REFGUID eventGuid,
```
vancem: Brian may also bring this up (because I pushed it on him as well), I would like to avoid GUIDs as a way of identifying providers.   I would like this to be a string instead.   Brian will change the EventPipe code for this, but you should do it as well (at worst, just string-ify the GUID, but we should be able to provide you with the Provider name).

mjsabby: @vancem if Brian will make this a string, can it be his change that does it all? Or are you suggesting I just make the GUID a string regardless of his change?

src/inc/corprof.idl
```diff
@@ -2408,6 +2412,31 @@ interface ICorProfilerCallback8 : ICorProfilerCallback7
...
+        [in] DWORD eventId,
+        [in] DWORD eventVersion,
+        [in] DWORD eventThreadId,
+        [in] LARGE_INTEGER eventTimestamp,
```
vancem: I am suspect the timestamp is QueryPerformanceCounter.     This looks to be the only place we use timestamps in the profiling API, so we may be setting a precedent.   My preference is that we use something that is more universal (e.g. some UTC based tick like FILETIME)

If we don't want to do this we should add the 'GetFrequency' API to the interface but this is a poor substitute (clients need to know some sort of base, that is a pain to produce.   this is why I prefer just using FILETIME (the runtime can do the conversion.  

See __GetSystemTimeAsFileTime for a way to get the current time as a FILETIME.   This can a call to QueryPerformanceCounter gives you the information you need to convert from any QPC value to a FILETIME.   

@noahfalk

src/inc/corprof.idl
noahfalk: as a proper COM interface we can't mutate it after it ships, and there is reasonable risk that whatever is commited here will ship as-is due to the rising bug-bar for changes. I don't want to ship an API that we know will be immediately deprecated.

vancem: I suggested that to Brian, but he had a worry that I did not really understand and we ran out of time.   I believe testing was a big part of it.   

My point is that I would like us to AVOID writing MORE code (e.g. tests) that would need to be changed if this changed.   Thus if we are going to use this API, I prefer it to be the right type for the provider (string).  

But it occurs to me that the API surface really needs to be bigger here.   The profiler really needs to be able to send control messages (turn the pipe on or off).     Feels like it is incomplete.

src/inc/corprof.idl
mjsabby: I was thinking about that and figured a profiler author would be able to call the necessary stuff on their end. But I'm happy to make it a FILETIME.

mjsabby:
@noahfalk ready for your review again, it's just a delete of the eventpipe stuff.

mjsabby:
For those doing 2.0 triage, I consider this is an important bug fix in the profiler apis since today DynamicMethodJittingStarted/Finished leads to a condition where if you're maintaining a dictionary it'll continue to grow forever since we never get events for dynamic methods that get garbage collected. Essentially it is a memory leak from the profiler apis perspective.

I had previously clubbed this change with a broader change called "eventpipes" but I've since removed that completely due to risk that EventPipes might change.

Not having this change in for 2.0 will be a lopsided profiler api implementation for a critical .NET Core release that will be widely deployed for a long time to come.

mjsabby:
Another relevant note for 2.0 triage decision makers -- even though the previous API additions DynamicMethodJittingStarted/Finished were added a few months back, they were not available in .NET Core 1.1, so .NET Core 2.0 will be the first public release they go out on and not having this critical api callback will be quite confusing to the profiler authors.

PR #11439, opened 2017-05-08
Make Debugger class static

justinvp:
It's `static` in the [.NET Standard](https://github.com/dotnet/standard/blob/cc9f646354fc68a13707a82323d4032b8dbfda52/netstandard/ref/mscorlib.cs#L4982) and [.NET Core](https://github.com/dotnet/corefx/blob/40f04b15ef889d78ee4490c0950bc3a74067d639/src/System.Diagnostics.Debug/ref/System.Diagnostics.Debug.cs#L75) refs, and in [CoreRT's implementation](https://github.com/dotnet/corert/blob/d174a2e4eb8ed875e9b2b56ae564f512a4aec3e2/src/System.Private.CoreLib/src/System/Diagnostics/Debugger.cs#L9), so make it static in the CoreCLR implementation as well.

Also, minor cleanup while making changes here.

### DIFF
--- a/src/mscorlib/src/System/Diagnostics/Debugger.cs
+++ b/src/mscorlib/src/System/Diagnostics/Debugger.cs
@@ -17,6 +17,9 @@ namespace System.Diagnostics
         [MethodImpl(MethodImplOptions.NoInlining)]
         public static void Break() => BreakInternal();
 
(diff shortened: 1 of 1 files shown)


stephentoub:
cc: @weshaggard

PR #11443, opened 2017-05-08
Fix copy error in perf script

DrewScoggins:
When I merged the PR that enabled the lean runs I added another call to
our run script.  This run script had an assumption that it would only
ever be called once and as such would try and copy two files that had
the same name.  This would cause it to prompt to overwrite or not and
fail the run.  This change makes each copy unique based on the
collection flags that were used for collection.

### DIFF
--- a/tests/scripts/run-xunit-perf.cmd
+++ b/tests/scripts/run-xunit-perf.cmd
@@ -107,8 +107,12 @@ setlocal
   )
 
   rem Save off the results to the root directory for recovery later in Jenkins
-  call :run_cmd xcopy "Perf-%BENCHNAME%.xml" "%CORECLR_REPO%\Perf-%BENCHNAME%-%COLLECTION_FLAGS%.xml" || exit /b 1
-  call :run_cmd xcopy "Perf-%BENCHNAME%.etl" "%CORECLR_REPO%\Perf-%BENCHNAME%-%COLLECTION_FLAGS%.etl" || exit /b 1
+  IF EXIST "Perf-%BENCHNAME%.xml" (
+    call :run_cmd copy "Perf-%BENCHNAME%.xml" "%CORECLR_REPO%\Perf-%BENCHNAME%-%COLLECTION_FLAGS%.xml" || exit /b 1
+  )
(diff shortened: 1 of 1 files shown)


DrewScoggins:
PTAL @brianrob @jorive.

tests/scripts/run-xunit-perf.cmd
```diff
@@ -107,8 +107,12 @@ setlocal
   )
 
   rem Save off the results to the root directory for recovery later in Jenkins
-  call :run_cmd xcopy "Perf-%BENCHNAME%*.xml" "%CORECLR_REPO%\" || exit /b 1
-  call :run_cmd xcopy "Perf-%BENCHNAME%*.etl" "%CORECLR_REPO%\" || exit /b 1
+  IF EXIST "Perf-%BENCHNAME%.xml" (
```
jorive: I would suggest keeping `xcopy` with the following flags:
```cmd
xcopy.exe /VYQ
/V           Verifies the size of each new file.
/Y           Suppresses prompting to confirm you want to overwrite an
               existing destination file.
/Q           Does not display file names while copying.
```
and always append either the `COLLECTION_FLAGS` or using the `ETW_COLLECTION` values, though I am more incline to use the `ETW_COLLECTION` flag: `Perf-%BENCHNAME%-Profile_%ETW_COLLECTION%.xml`. The reason is that if the collection flags ever add a character that is invalid for a file name, this would break.

Another fun fact is that the `COPY` command uses the `+` character to append files (and the API uses that character to append providers):
```cmd
To append files, specify a single file for destination, but multiple files
for source (using wildcards or file1+file2+file3 format).
```

DrewScoggins: We talked about this and decided to switch to using the %ETW_COLLECTION% as the variable to use.  We need to continue using copy because xcopy needs to specify whether you are copying a file or a directory.  We could do this by piping echo F to the xcopy, but just using copy seemed cleaner to me.

jorive: LGTM

PR #11445, opened 2017-05-08
Add define for FEATURE_SIMD

jashook:
This fixes a build break in the desktop product.

ptal @CarolEidt @rartemev 
/cc @dotnet/jit-contrib

### DIFF
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -7444,10 +7444,18 @@ void CodeGen::genProfilingEnterCallback(regNumber initReg, bool* pInitRegZeroed)
                 continue;
             }
 
-            var_types   storeType = varDsc->lvaArgType();
-            regNumber   argReg    = varDsc->lvArgReg;
-            instruction store_ins =
-                ((storeType == TYP_SIMD8) && genIsValidIntReg(argReg)) ? INS_mov : ins_Store(storeType);
+            var_types storeType = varDsc->lvaArgType();
+            regNumber argReg    = varDsc->lvArgReg;
+
+            instruction store_ins = ins_Store(storeType);
+
+#ifdef FEATURE_SIMD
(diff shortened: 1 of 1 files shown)


jashook:
Desktop build came back clean with the fix.

src/jit/codegencommon.cpp
```diff
@@ -7418,7 +7418,7 @@ void CodeGen::genProfilingEnterCallback(regNumber initReg, bool* pInitRegZeroed)
         return;
     }
 
-#if defined(_TARGET_AMD64_) && !defined(UNIX_AMD64_ABI) // No profiling for System V systems yet.
```
CarolEidt: Since this code doesn't really depend on FEATURE_SIMD, it would seem to make sense to change the setting of `storeIns` below to something like:
```
instruction store_ins = ins_Store(storeType);
#ifdef FEATURE_SIMD
if ((storeType == TYP_SIMD8) && genIsValidIntReg(argReg))
{
    storeType = INS_mov;
}
#endif
```
and similarly for `loadIns`

jashook: Will make the change thanks

jashook:
@CarolEidt ptal with the feedback addressed. The desktop build came back clean again.

src/jit/codegencommon.cpp
```diff
@@ -7446,8 +7446,16 @@ void CodeGen::genProfilingEnterCallback(regNumber initReg, bool* pInitRegZeroed)
 
             var_types   storeType = varDsc->lvaArgType();
             regNumber   argReg    = varDsc->lvArgReg;
-            instruction store_ins =
-                ((storeType == TYP_SIMD8) && genIsValidIntReg(argReg)) ? INS_mov : ins_Store(storeType);
+
+            instruction store_ins = ins_Store(storeType);
+
+            #ifdef FEATURE_SIMD
```
mikedn: Usually preprocessing directives are not indented.

Issue #11447, opened 2017-05-08
Test `_hfa_main_testC_hfa_nd2C_d` failing on x64 under GCStress=0xC and JITStress=1


pgavlin:
e.g. https://ci.dot.net/job/dotnet_coreclr/job/master/job/jitstress/job/x64_checked_windows_nt_gcstress0xc_jitstress1/lastCompletedBuild/testReport/JIT_jit64/_hfa_main_testC_hfa_nd2C_d_hfa_nd2C_d_/_hfa_main_testC_hfa_nd2C_d_hfa_nd2C_d_cmd/

From the JIT dumps, this appears to be due to misplacement of a call to the PInvoke frame initialization IR that attempts to read the stub parameter from `r10` after it has been trashed.

cc @briansull

RussKeldorph:
@pgavlin Should this be closed?

pgavlin:
Yes.

Issue #11450, opened 2017-05-08
Represent the secret stub arg passed to the PInvoke frame initializer using a lclVar.


pgavlin:
Currently this argument is read directly from the physical register in which it is passed, which prevents the compiler from safely inserting code before the physreg read. This was the cause of #11447.

AndyAyersMS:
@RussKeldorph seems like we're unlikely to get to this for 2.1. I'm going to move it to future; if you disagree feel free to move it back.

PR #11451, opened 2017-05-08
Fix issue #11447.

pgavlin:
This issue was caused by the code for PInvoke frame initialization reading
from the stub argument register after it had been trashed by prior
instructions. This change fixes this issue by inserting the PInvoke
frame initialization at the beginning of the scratch block into which it
is inserted rather than at the end.

Note that a more correct fix here would be to represent the stub
argument as a lclVar rather than a physical register; that work is
tracked by #11450.


pgavlin:
@briansull @BruceForstall @dotnet/jit-contrib PTAL

Issue #11446, opened 2017-05-08
systemvbringup.structinregs failing on x86 under GCStress=0xc


pgavlin:
see e.g. https://ci.dot.net/job/dotnet_coreclr/job/master/job/jitstress/job/x86_checked_windows_nt_gcstress0xc/25/testReport/JIT_Methodical/_structs_systemvbringup_structinregs_structinregs_/_structs_systemvbringup_structinregs_structinregs_cmd/

From the JIT dump, it looks like this is an issue related to struct promotion: V10, which is a struct that contains a struct that contains a GC ref, is promoted to a single lclVar of type ref (rather than type struct). After this promotion, there is still IR in the function which accesses the promoted lclVar as a struct (e.g. using `lclFld`/`st.lclFld`), which confuses the incremental liveness performed during code generation and causes the lclVar to die too early.

pgavlin:
cc @CarolEidt 

@rartemev, this may also be the cause of the GC stress failure in this test on x64. Are you looking into that test at the moment?

pgavlin:
Closed by #11455.

PR #11455, opened 2017-05-08
Fix issue #11446.

pgavlin:
This issue was a GC hole that was caused by bad IR produced by
recursivelty promoting a struct with the following form:

```
struct S
{
    struct T
    {
        object o;
    };

    T t;
}
```

In this case, the recursive promotion logic created a single lclVar to
represent `S.t` and retyped it using the type of `T.o`, but did not
rewrite IR that treated the single lclVar as a struct (e.g. instances of
`lclFld/st.lclFld Vx` where `Vx` referred to the lclVar that represented
`S.t` were not rewritten an `lclVar/st.lclVar Vx`). This IR in
turn confused the incremental liveness used during code generation,
which caused the stack slot holding `o`'s value to die too early.

This change fixes codegen's incremental liveness to handle this case.

### DIFF
--- a/src/jit/lclvars.cpp
+++ b/src/jit/lclvars.cpp
@@ -1554,8 +1554,6 @@ void Compiler::lvaCanPromoteStructType(CORINFO_CLASS_HANDLE    typeHnd,
 
             if (pFieldInfo->fldSize == 0)
             {
-                // Non-primitive struct field.
-
                 // Size of TYP_BLK, TYP_FUNC, TYP_VOID and TYP_STRUCT is zero.
                 // Early out if field type is other than TYP_STRUCT.
                 // This is a defensive check as we don't expect a struct to have
@@ -1565,16 +1563,9 @@ void Compiler::lvaCanPromoteStructType(CORINFO_CLASS_HANDLE    typeHnd,
                     return;
                 }
 
-                // Try to promote byref-like structs with a single field of scalar type that is aligned at its
-                // type's natural boundary.
-
-                // Do not promote if the single field is not byref-like.
-                const DWORD fieldFlags       = info.compCompHnd->getClassAttribs(pFieldInfo->fldTypeHnd);
-                const bool  fieldIsByRefLike = (fieldFlags & CORINFO_FLG_CONTAINS_STACK_PTR) != 0;
-                if (!fieldIsByRefLike)
-                {
-                    return;
-                }
+                // Non-primitive struct field.
+                // Try to promote structs of single field of scalar types aligned at their
+                // natural boundary.
 
                 // Do Not promote if the struct field in turn has more than one field.
                 if (info.compCompHnd->getClassNumInstanceFields(pFieldInfo->fldTypeHnd) != 1)
--- a/src/jit/lclvars.cpp
+++ b/src/jit/lclvars.cpp
@@ -1554,6 +1554,8 @@ void Compiler::lvaCanPromoteStructType(CORINFO_CLASS_HANDLE    typeHnd,
 
             if (pFieldInfo->fldSize == 0)
             {
+                // Non-primitive struct field.
+
                 // Size of TYP_BLK, TYP_FUNC, TYP_VOID and TYP_STRUCT is zero.
                 // Early out if field type is other than TYP_STRUCT.
                 // This is a defensive check as we don't expect a struct to have
@@ -1563,9 +1565,16 @@ void Compiler::lvaCanPromoteStructType(CORINFO_CLASS_HANDLE    typeHnd,
                     return;
                 }
 
-                // Non-primitive struct field.
-                // Try to promote structs of single field of scalar types aligned at their
-                // natural boundary.
+                // Try to promote byref-like structs with a single field of scalar type that is aligned at its
+                // type's natural boundary.
+
+                // Do not promote if the single field is not byref-like.
+                const DWORD fieldFlags       = info.compCompHnd->getClassAttribs(pFieldInfo->fldTypeHnd);
+                const bool  fieldIsByRefLike = (fieldFlags & CORINFO_FLG_CONTAINS_STACK_PTR) != 0;
+                if (!fieldIsByRefLike)
+                {
+                    return;
+                }
 
                 // Do Not promote if the struct field in turn has more than one field.
                 if (info.compCompHnd->getClassNumInstanceFields(pFieldInfo->fldTypeHnd) != 1)
--- a/src/jit/codegenlinear.cpp
+++ b/src/jit/codegenlinear.cpp
@@ -1228,7 +1228,7 @@ void CodeGen::genConsumeRegs(GenTree* tree)
             genConsumeAddress(tree->AsIndir()->Addr());
         }
 #ifdef _TARGET_XARCH_
-        else if (tree->OperGet() == GT_LCL_VAR)
+        else if (tree->OperIsLocalRead())
         {
             // A contained lcl var must be living on stack and marked as reg optional, or not be a
             // register candidate.
--- a/src/jit/codegenxarch.cpp
+++ b/src/jit/codegenxarch.cpp
@@ -1558,6 +1558,8 @@ void CodeGen::genCodeForTreeNode(GenTreePtr treeNode)
             GenTreePtr op1 = treeNode->gtGetOp1();
             genConsumeRegs(op1);
             emit->emitInsBinary(ins_Store(targetType), emitTypeSize(treeNode), treeNode, op1);
+
+            genUpdateLife(treeNode);
         }
         break;
 
--- a/src/jit/emitxarch.cpp
+++ b/src/jit/emitxarch.cpp
@@ -3005,6 +3005,11 @@ regNumber emitter::emitInsBinary(instruction ins, emitAttr attr, GenTree* dst, G
             }
         }
 
+        if (lclField != nullptr)
+        {
+            codeGen->genUpdateLife(lclField);
+        }
+
         if (tmpDsc != nullptr)
         {
             emitComp->tmpRlsTemp(tmpDsc);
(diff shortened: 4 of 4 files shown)


pgavlin:
@russellhadley @JosephTremoulet @CarolEidt @dotnet/jit-contrib PTAL

pgavlin:
I have verified that this change does not prevent struct promotion of `Span<T>`. I am running desktop and CoreCLR asm diffs now.

JosephTremoulet:
I'm having trouble parsing "created a single lclVar to
represent `S.t` and retyped it using the type of `T.o`, but did not
rewrite IR that treated the single lclVar as a struct (e.g. instances of
`lclFld/st.lclFld Vx` where `Vx` referred to the lclVar that represented
`S.t` were not rewritten an `lclVar/st.lclVar Vx`)" and following what "none of these types are the target
of field accesses" means.  Can you give some example IR of the problem?

src/jit/lclvars.cpp
```diff
@@ -1563,9 +1565,16 @@ void Compiler::lvaCanPromoteStructType(CORINFO_CLASS_HANDLE    typeHnd,
...
+                // Try to promote byref-like structs with a single field of scalar type that is aligned at its
+                // type's natural boundary.
+
+                // Do not promote if the single field is not byref-like.
+                const DWORD fieldFlags       = info.compCompHnd->getClassAttribs(pFieldInfo->fldTypeHnd);
+                const bool  fieldIsByRefLike = (fieldFlags & CORINFO_FLG_CONTAINS_STACK_PTR) != 0;
+                if (!fieldIsByRefLike)
+                {
+                    return;
+                }
```
CarolEidt: I think I understand the reason to be conservative here, but this will also disable promoting a field that's a struct containing a single ptr-sized int. That seems unnecessarily pessimizing. Couldn't you just check below that `fieldVarType` is not `TYP_REF`?

JosephTremoulet:
Sorry, still not following -- why can't we hit the same pattern when `foo1` is a `ByReference<T>`?

pgavlin:
Oh, sorry. IIUC we won't hit that pattern b/c we use intrinsics to access the pointer in a span and therefore never directly reference the `ByReference<T>` stored in a span. See the [code in the importer for `CORINFO_INTRINSIC_{ReadOnly,}Span_GetItem`](https://github.com/dotnet/coreclr/blob/master/src/jit/importer.cpp#L3633-L3712).

JosephTremoulet:
The span indexers only [recently](https://github.com/dotnet/coreclr/pull/10910) became intrinsics, and they [are](https://github.com/dotnet/coreclr/pull/10910#issue-221121741) "optional-expand" intrinsics at that.  I'm hesitant to extrapolate (for correctness) that all uses of `ByReference<T>` fields will be performed via intrinsic rewrites.

src/jit/lclvars.cpp
pgavlin: Happily, that case seems to fall out due to "normed structs": for structs with a single pointer-sized integral field, the VM reports the type of the struct as the type of the field and we are able to perform the promotion. Here is an example:

`struct T` is normalized to `int`:
```
struct S
{
    struct T
    {
        public int i;
    }

    T t;
    int j;

    static int Main(string[] args)
    {
        S s;
        s.t.i = 42;

        if (args.Length == 0)
        {
            S s2;
            s2.t.i = 100;
            s2.j = 0;

            s = s2;
        }

        return s.t.i;
    }
}
```
```
*************** In fgPromoteStructs()

lvaTable before fgPromoteStructs
; Initial local variable assignments
;
;  V00 arg0              ref  class-hnd
;  V01 loc0           struct ( 8)  ld-addr-op
;  V02 loc1           struct ( 8)  ld-addr-op

Promoting struct local V01 (S):
lvaGrabTemp returning 3 (V03 tmp0) (a long lifetime temp) called for field V01.t (fldOffset=0x0).

lvaGrabTemp returning 4 (V04 tmp1) (a long lifetime temp) called for field V01.j (fldOffset=0x4).

Promoting struct local V02 (S):
lvaGrabTemp returning 5 (V05 tmp2) (a long lifetime temp) called for field V02.t (fldOffset=0x0).

lvaGrabTemp returning 6 (V06 tmp3) (a long lifetime temp) called for field V02.j (fldOffset=0x4).

lvaTable after fgPromoteStructs
; Initial local variable assignments
;
;  V00 arg0              ref  class-hnd
;  V01 loc0           struct ( 8)  ld-addr-op
;  V02 loc1           struct ( 8)  ld-addr-op
;  V03 tmp0              int  V01.t(offs=0x00) P-INDEP
;  V04 tmp1              int  V01.j(offs=0x04) P-INDEP
;  V05 tmp2              int  V02.t(offs=0x00) P-INDEP
;  V06 tmp3              int  V02.j(offs=0x04) P-INDEP
```

`struct T` is not normalized:
```
struct S
{
    struct T
    {
        public object i;
    }

    T t;
    int j;

    static int Main(string[] args)
    {
        S s;
        s.t.i = 42;

        if (args.Length == 0)
        {
            S s2;
            s2.t.i = 100;
            s2.j = 0;

            s = s2;
        }

        return (int)s.t.i;
    }
}
```

With this change:
```
lvaTable before fgPromoteStructs
; Initial local variable assignments
;
;  V00 arg0              ref  class-hnd
;  V01 loc0           struct ( 8)  ld-addr-op
;  V02 loc1           struct ( 8)  ld-addr-op
;  V03 tmp0              ref 
;  V04 tmp1              ref 

lvaTable after fgPromoteStructs
; Initial local variable assignments
;
;  V00 arg0              ref  class-hnd
;  V01 loc0           struct ( 8)  ld-addr-op
;  V02 loc1           struct ( 8)  ld-addr-op
;  V03 tmp0              ref 
;  V04 tmp1              ref 
```

Without this change:
```
lvaTable before fgPromoteStructs
; Initial local variable assignments
;
;  V00 arg0              ref  class-hnd
;  V01 loc0           struct ( 8)  ld-addr-op
;  V02 loc1           struct ( 8)  ld-addr-op
;  V03 tmp0              ref 
;  V04 tmp1              ref 

Promoting struct local V01 (S):
lvaGrabTemp returning 5 (V05 tmp2) (a long lifetime temp) called for field V01.j (fldOffset=0x0).

lvaGrabTemp returning 6 (V06 tmp3) (a long lifetime temp) called for field V01.t (fldOffset=0x4).

Promoting struct local V02 (S):
lvaGrabTemp returning 7 (V07 tmp4) (a long lifetime temp) called for field V02.j (fldOffset=0x0).

lvaGrabTemp returning 8 (V08 tmp5) (a long lifetime temp) called for field V02.t (fldOffset=0x4).

lvaTable after fgPromoteStructs
; Initial local variable assignments
;
;  V00 arg0              ref  class-hnd
;  V01 loc0           struct ( 8)  ld-addr-op
;  V02 loc1           struct ( 8)  ld-addr-op
;  V03 tmp0              ref 
;  V04 tmp1              ref 
;  V05 tmp2              int  V01.j(offs=0x00) P-INDEP
;  V06 tmp3              ref  V01.t(offs=0x04) P-INDEP
;  V07 tmp4              int  V02.j(offs=0x00) P-INDEP
;  V08 tmp5              ref  V02.t(offs=0x04) P-INDEP
```

pgavlin:
> I'm hesitant to extrapolate (for correctness) that all uses of `ByReference<T>` fields will be performed via intrinsic rewrites.

If that is the case, then we may need to disable this optimization entirely for 2.0.

cc @AndyAyersMS

JosephTremoulet:
> we may need to disable this optimization entirely for 2.0

Or make liveness conservative when it sees a nonsensical field use on a GC pointer local?

pgavlin:
> Or make liveness conservative when it sees a nonsensical field use on a GC pointer local?

@russellhadley and I were concerned by this option, but it would probably be relatively straightforward: a `lclFld` of a ref-typed local could be considered a full use of the lcl var and a `st.lclFld` to the same could be considered a full def. FWIW, the liveness pass gets things right--it's the incremental liveness performed during codegen that gets this wrong.

src/jit/lclvars.cpp
CarolEidt: Thanks - I'd forgotten about the normalizing

CarolEidt:
@pgavlin - it seems like the fix to codegen liveness would be reasonable, IMO

pgavlin:
Okay, I've updated the change to simply call `genUpdateLife` in binop processing when the binop uses a lclFld. Looking at this a bit closer, the bug may actually have been in the binop processing--I think I we also need to call `genUpdateLife` for lclVar uses.

CarolEidt:
Why is the liveness update not being handled by `genConsumeOperands` prior to the call to `emitInsBinary`? I don't think the latter should be doing liveness updates (though that particular method is really more codegen-like than emitter-like).
Do you have an example where the liveness is not being updated for a lclVar? It would be my guess that the incorrect update for lclFld is due to `genConsumeOperands` not handling that one correctly.

pgavlin:
> Why is the liveness update not being handled by `genConsumeOperands` prior to the call to `emitInsBinary`? I don't think the latter should be doing liveness updates (though that particular method is really more codegen-like than emitter-like).

`genConsumeOperands` is not handling the lclFld case (as you guessed correctly).

> It would be my guess that the incorrect update for lclFld is due to `genConsumeOperands` not handling that one correctly.

`emitInsBinary` is also used to process `GT_STORE_LCL_FLD`; the problem is not that the liveness of the sources is being updated incorrectly, it is that the liveness of the destination is not being updated correctly.

I'll refactor this and move the relevant code around.

pgavlin:
(FWIW, I added this code to `emitInsBinary` because there is similar code in `emitInsMov` when the destination is a lclVar)

AndyAyersMS:
There are many other methods on `Span<T>` that access the pointer field and are not handled via intrinsics. We won't have much coverage of these in CoreCLR -- you might consider GC stressing the CoreFX tests.

pgavlin:
@dotnet-bot help

PR #11456, opened 2017-05-08
Add an alternate algorithm done during LSRA

CarolEidt:
Adding an alternative approach that would be performed in LSRA and that wouldn't require adding additional proxy variables.  It would also add the notion of write-thru defs, that could be used to spill variables at def points while also keeping the register live for near uses.

### DIFF
--- a/Documentation/design-docs/eh-writethru.md
+++ b/Documentation/design-docs/eh-writethru.md
@@ -103,6 +103,39 @@ On the IR directly before SSA build:
 At end no proxy should be live across EH flow and all value updates will be
 written back to the stack location.
 
+### Alternate Algorithm: In LSRA
+
+* Add a flag to identify Intervals as "WriteThru". This would be set on all lclVars
+  considered by liveness to be exceptVars.
+* Additionally, add a flag to identify RefPositions as "WriteThru". The motivation
+  for having both, is that in the exception var case, we want to create all defs as
+  write-thru, but for other purposes we may want to make some defs write-thru
+  (i.e. they spill but the target register remains live), but not all defs for a given lclVar.
+* During liveness, mark exception vars as `lvLiveInOutOfHndlr`, but not `lvDoNotEnregister`.
+* During interval creation, if a lclVar is marked `lvLiveInOutOfHndlr`, set `isWriteThru` on the interval.
+* Set handler entry blocks as having no predecessor for register-mapping purposes.
+  - Leave the inVarToRegMaps empty (all incoming vars on stack)
+* Set the outVarToRegMap to empty for EH exit blocks.
+* During allocation, treat isWriteThru interval defs and uses differently:
+  - A def is always marked writeThru if it is assigned a register. If it doesn't get a register
+    at all, it is marked spillAfter as per usual.
+  - A use is never marked spillAfter (as the stack location is always valid at a use).
+* During resolution/writeback:
+  - Mark all isWriteThru defs with `GTF_SPILL`, as for `spillAfter`, but keep the reg assignment,
+    and the interval stays active.
+  - Assert that uses of isWriteThru intervals are never marked spillAfter
+* During `genFnProlog()`, ensure that incoming reg parameters that have register assoginments also
+  get stored to stack if they are marked lvLiveInOutOfHndlr.
+
+#### Challenges/Issues with the LSRA approach above:
+
+* Liveness currently adds all exceptVars to the live-in for blocks where `ehBlockHasExnFlowDsc` returns true.
+  This results in more "artificial" liveness than strictly entry to and exit from EH regions.
+* In some cases, write-thru may be worse, performance-wise, than always using memory, if the EH local is
+  infrequently referenced in non-EH code. This is a slightly different issue than known spill placement
+  and allocation issues, but is related (i.e. when to choose not to keep the register live, and simply
+  create the value in memory if that doesn't require a register).
+
 ## Next steps
 
 The initial prototype that produced the example bellow is currently being


CarolEidt:
@russellhadley PTAL

Documentation/design-docs/eh-writethru.md
```diff
@@ -103,6 +103,39 @@ On the IR directly before SSA build:
...
+  - A def is always marked writeThru if it is assigned a register. If it doesn't get a register
+    at all, it is marked spillAfter as per usual.
+  - A use is never marked spillAfter (as the stack location is always valid at a use).
+* During resolution/writeback:
+  - Mark all isWriteThru defs with `GTF_SPILL`, as for `spillAfter`, but keep the reg assignment,
+    and the interval stays active.
+  - Assert that uses of isWriteThru intervals are never marked spillAfter
+* During `genFnProlog()`, ensure that incoming reg parameters that have register assoginments also
+  get stored to stack if they are marked lvLiveInOutOfHndlr.
+
```
russellhadley: Does the above include inferring a non-redundant (or likely none redundant) reload insertion point?  I'm thinking about finding an insertion point that, best case, dominates all uses, next best case, does not extend the dynamic trace, and lastly is correct but partially redundant.  (I'm assuming we won't get fully redundant reloads).  I was thinking that it might be useful to add synthetic ref positions at handler entries and filter/fault/finally exits for reloads and then let the rest of the spilling logic (remat/fold/etc.) work on them as needed.

russellhadley:
Good start.  I had one follow up question about where the reloads would be computed.  Also one of us can probably take a stab at the motivation for the alternative as well as what criteria we will use to make the decision.

Documentation/design-docs/eh-writethru.md
CarolEidt: The reloads will be inserted in the eh region on demand. That is, because the inVarToRegMap will have everything on the stack, the first use encountered will cause the variable to be reloaded. A subsequent use will avoid a redundant reload as long as it is not spilled.
The reloads on eh exit will also fall out "naturally" - this time from the resolution process. Because the outVarToRegMap will have everything on the stack, reloads will be inserted to match the sucessor(s) (either at the end of the exit block, at the successor, or on a split edge).
Thus, this approach both suffers from existing allocation issues, but also benefits from improvements that will ultimately be made in better spill/reload placement and improved resolution.

CarolEidt:
@russellhadley - I have pushed my experimental branch to my repo (https://github.com/CarolEidt/coreclr/tree/EHWriteThruLSRA).

It is incomplete (i.e. incorrect) for incoming parameters, as I haven't done the work to make `genFnProlog()` home them to both stack and register.

Here's the code for your sample. Note that I didn't scrutinize it closely, so there may be more wrong with it than just the lack of homing the arg to the stack. However, it goes from 178 to 163 bytes, saving 11 bytes once you account for the additional 4 bytes to home the arg:
```
; Assembly listing for method Enreg01:Run():int:this
; optimized code
; rbp based frame
; fully interruptible
;
;  V00 this         [V00,T00] ( 12,  4   )     ref  ->  [rbp+0x10]   this class-hnd
;  V01 loc0         [V01,T01] ( 10,  2   )     int  ->  [rbp-0x1C]  
;* V02 loc1         [V02    ] (  0,  0   )     ref  ->  zero-ref    class-hnd
;  V03 tmp0         [V03,T02] (  4,  0   )     ref  ->  rsi         class-hnd
;  V04 tmp1         [V04,T03] (  2,  0   )     ref  ->  rdx        
;  V05 tmp2         [V05,T04] (  2,  0   )     ref  ->  rdi        
;  V06 OutArgs      [V06    ] (  1,  1   )  lclBlk (32) [rsp+0x00]  
;  V07 PSPSym       [V07    ] (  1,  1   )    long  ->  [rbp-0x30]   do-not-enreg[X] addr-exposed
;
; Lcl frame size = 56

G_M16307_IG01:
       push     rbp
       push     rdi
       push     rsi
       push     rbx
       sub      rsp, 56
       lea      rbp, [rsp+50H]
       mov      qword ptr [rbp-30H], rsp
       mov      rsi, rcx                          ; This is the homing of V00 (this) to rsi
       ; Here's where there's a missing store of V00 to the stack.

G_M16307_IG02:
       mov      edi, dword ptr [rsi+16]
       mov      dword ptr [rbp-1CH], edi          ; write-thru of V01 (sum)

G_M16307_IG03:
       mov      rcx, rsi                          ; use of V00 from register
       mov      edx, 97
       call     Enreg01:TryValue(int):int:this
       nop      

G_M16307_IG04:
       mov      eax, edi

G_M16307_IG05:
       lea      rsp, [rbp-18H]
       pop      rbx
       pop      rsi
       pop      rdi
       pop      rbp
       ret      

G_M16307_IG06:
       push     rbp
       push     rdi
       push     rsi
       push     rbx
       sub      rsp, 40
       mov      rbp, qword ptr [rcx+32]
       mov      qword ptr [rsp+20H], rbp
       lea      rbp, [rbp+50H]

G_M16307_IG07:
       mov      rsi, rdx
       mov      rcx, 0xD1FFAB1E
       mov      rdi, gword ptr [rcx]
       mov      ecx, dword ptr [rsi+140]
       call     System.Convert:ToString(int):ref
       mov      rdx, rax
       mov      rcx, rdi
       call     System.Console:WriteLine(ref,ref)
       mov      edi, dword ptr [rbp-1CH]          ; reload of V01
       mov      rbx, gword ptr [rbp+10H]          ; reload of V00
       add      edi, dword ptr [rbx+16]           ; Add val to V01 (sum) in edi
       add      edi, dword ptr [rsi+140]          ; Add e.x to V01
       mov      dword ptr [rbp-1CH], edi          ; write-thru of V01
       lea      rdx, bword ptr [rbx+8]            ; use of V00 from rbx
       mov      rcx, rbx                          ; use of V00 from rbx
       call     Enreg01:foo(byref):int:this
       add      edi, dword ptr [rbx+16]
       mov      dword ptr [rbp-1CH], edi          ; write-thru of V01
       lea      rax, G_M16307_IG04

G_M16307_IG08:
       add      rsp, 40
       pop      rbx
       pop      rsi
       pop      rdi
       pop      rbp
       ret      

; Total bytes of code 163, prolog size 20 for method Enreg01:Run():int:this
```

CarolEidt:
@russellhadley - I pushed a change to codegencommon.cpp that handles the incoming arg. However, in looking at the generated code, I realized that it is not handling the EH region exit case properly. I don't think that should be too hard to do.

CarolEidt:
@russellhadley - Another commit. This one adds a check for whether the block has an EH predecessor. If so, it assumes incoming vars are on the stack. This jits the sample correctly, though crossgen fails for System.Private.Corelib.dll, so there are obviously remaining issues.

JosephTremoulet:
@CarolEidt, @russellhadley, is this ready to merge?

CarolEidt:
> is this ready to merge?

Since @russellhadley authored the original document, I was waiting on him.

PR #11459, opened 2017-05-08
Implement Circular Buffering for EventPipe

brianrob:
Rather than taking a lock on each event write operation, implement a mostly lock-free circular buffering algorithm that holds the events in-memory until tracing is stopped and the contents are written to disk.


brianrob:
@vancem, PTAL.

@valenis, FYI.

brianrob:

brianrob:

src/vm/eventpipebuffermanager.cpp
```diff
@@ -0,0 +1,595 @@
...
+        THROWS;
+        GC_NOTRIGGER;
+        MODE_ANY;
+        PRECONDITION(pThread != NULL);
+        PRECONDITION(requestSize > 0);
+    }
+    CONTRACTL_END;
+
+    // Allocating a buffer requires us to take the lock.
+    SpinLockHolder _slh(&m_lock);
```
janvorli: There is a lot of stuff under the spinlock, including memory allocations. It would be great to reduce the amount of stuff done under the spinlock to the minimum necessary. It seems that the code can be relatively easily refactored to achieve that goal. Unless I am mistaken, only the m_pPerThreadBufferList->InsertTail and a portion of the code to steal a buffer from another thread needs to be synchronized.

src/vm/eventpipebuffermanager.cpp
```diff
@@ -0,0 +1,595 @@
...
+        {
+            // We're done.  There are no more events.
+            break;
+        }
+
+        // Write the oldest event.
+        pFile->WriteEvent(*pOldestInstance);
+
+        // Pop the event from the buffer.
+        pOldestContainingBuffer->PopNext(stopTimeStamp);
```
janvorli: How is the PopNext call from one thread synchronized with WriteEvent from another thread?

src/vm/eventpipe.h
```diff
@@ -43,6 +47,18 @@ class StackContents
         Reset();
     }
 
+    void CopyTo(StackContents *pDest)
+    {
+        LIMITED_METHOD_CONTRACT;
+        _ASSERTE(pDest != NULL);
+
+        memcpy(pDest->m_stackFrames, m_stackFrames, sizeof(UINT_PTR) * m_nextAvailableFrame);
```
brianrob: Yes, I will do this.

src/vm/eventpipebuffermanager.cpp
brianrob: It turns out that I will need more to happen under the spin lock - this PR doesn't include the clean-up routines, and so they will require list from the thread object to be under lock.  We could potentially try not to lock on the allocations, but we'd have to allocate pre-emptively and then free it if we didn't need it.

I'm going to be closing this PR out because I have more changes on top of these and it makes sense to review them together.  I will see if there is anything that I can do here.

src/vm/eventpipebuffermanager.cpp
```diff
@@ -0,0 +1,595 @@
...
+    // 3. Process up to the lowest most recent timestamp for the set of buffers.
+    // 4. When we get NULLs from each of the buffers on PopNext(), we're done.
+    // 5. While iterating if PopNext() == NULL && Empty() == NULL, remove the buffer from the list.  It's empty.
+    // 6. While iterating, grab the next lowest most recent timestamp.
+    // 7. Walk through the list again and look for any buffers that have a lower most recent timestamp than the next most recent timestamp.
+    // 8. If we find one, add it to the list and select its most recent timestamp as the lowest.
+    // 9. Process again (go to 3).
+    // 10. Continue until there are no more buffers to process.
+
+    // Niavely walk the circular buffer, writing the event stream in timestamp order.
```
brianrob: Fixed.

src/vm/eventpipebuffermanager.cpp
```diff
@@ -0,0 +1,595 @@
...
+    while(true)
```
brianrob: Yes, this is true.  Since this is a v1 feature and we expect event loads to be relatively low for the scenario that we're targeting, this is likely to be OK.  Going forward, I agree that we do need to look at this.

I will file an issue for this.

src/vm/eventpipebuffermanager.cpp
brianrob: PopNext is only called once tracing is disabled.  This means that if anyone is writing an event, they are writing it after the stop timestamp.  What we do is this:

- Buffers are always zero'd.
- Read as far into the buffer as we can.  However, if the timestamp is zero or after the stopTimeStamp, ignore it and don't go any further.  Thus, we won't read corrupted data or go further than the write pointer.

brianrob:
@janvorli, thank you for your comments.  I'll come back with a WIP PR that contains more changes on top of this.

PR #11465, opened 2017-05-09
Stop hoisting statics above cctors

### LINKED ISSUE #10780: NullReferenceException during 'dotnet build' on ubuntu arm32

JosephTremoulet:
Extend the logic in optLoopHoist around `CLS_VAR` nodes to ensure we don't
hoist any tree that contains a `CLS_VAR` for a field with the flag
`CORINFO_FLG_FIELD_INITCLASS` set, unless we also (identify and) hoist the
corresponding static init helper call.  The previous logic was
insufficient in that it blocked hoisting of singleton `CLS_VAR` nodes, but
not hoisting of trees that contain `CLS_VAR` nodes as sub-trees.

Add flag `GTF_FLD_INITCLASS`/`GTF_CLS_VAR_INITCLASS` so that optLoopHoist
can recall which fields are paired with static init helper calls, and only
block the hoisting of their trees.

Fixes #10780.

### DIFF
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -6057,6 +6057,11 @@ GenTreePtr Compiler::impImportStaticFieldAccess(CORINFO_RESOLVED_TOKEN* pResolve
                 // In future, it may be better to just create the right tree here instead of folding it later.
                 op1 = gtNewFieldRef(lclTyp, pResolvedToken->hField);
 
+                if (pFieldInfo->fieldFlags & CORINFO_FLG_FIELD_INITCLASS)
+                {
+                    op1->gtFlags |= GTF_FLD_INITCLASS;
+                }
+
                 if (pFieldInfo->fieldFlags & CORINFO_FLG_FIELD_STATIC_IN_HEAP)
                 {
                     op1->gtType = TYP_REF; // points at boxed object
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -6628,9 +6628,10 @@ GenTreePtr Compiler::fgMorphField(GenTreePtr tree, MorphAddrContext* mac)
                 else
 #endif // _TARGET_64BIT_
                 {
-                    // Only volatile could be set, and it maps over
-                    noway_assert((tree->gtFlags & ~(GTF_FLD_VOLATILE | GTF_COMMON_MASK)) == 0);
-                    noway_assert(GTF_FLD_VOLATILE == GTF_IND_VOLATILE);
+                    // Only volatile or classinit could be set, and they map over
+                    noway_assert((tree->gtFlags & ~(GTF_FLD_VOLATILE | GTF_FLD_INITCLASS | GTF_COMMON_MASK)) == 0);
+                    static_assert_no_msg(GTF_FLD_VOLATILE == GTF_CLS_VAR_VOLATILE);
+                    static_assert_no_msg(GTF_FLD_INITCLASS == GTF_CLS_VAR_INITCLASS);
                     tree->SetOper(GT_CLS_VAR);
                     tree->gtClsVar.gtClsVarHnd = symHnd;
                     FieldSeqNode* fieldSeq =
--- a/tests/src/JIT/Regression/JitBlue/GitHub_10780/GitHub_10780.cs
+++ b/tests/src/JIT/Regression/JitBlue/GitHub_10780/GitHub_10780.cs
@@ -1,0 +1,52 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+// Repro case for a bug involving hoisting of static field loads out of
+// loops and (illegally) above the corresponding type initializer calls.
+
+using System.Runtime.CompilerServices;
+
+namespace N
+{
+    public struct Pair
+    {
+        public int Left;
+        public int Right;
+
+        public static Pair TenFour = new Pair() { Left = 10, Right = 4 };
+    }
+
+    static class C
+    {
+        static int Sum;
+        static int Two;
+
+        // Bug repro requires a use of a Pair value; this is a small fn that takes
+        // a Pair by value to serve as that use.  Inline it aggressively so that
+        // we won't think the call might kill the static field.
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        static void Accumulate(Pair pair)
+        {
+            Sum += pair.Left + pair.Right;
+        }
+
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void SumNFourteens(int n)
+        {
+            for (int i = 0; i < n; ++i)
+            {
+                Two = 2; // Store to C.Two here is a global side-effect above which we won't hoist the static initializer (since it may throw).
+                Accumulate(Pair.TenFour);  // Hoisting the load of Pair.TenFour above the static init call is incorrect.
+            }
+        }
+
+        public static int Main(string[] args)
+        {
+            Sum = 0;
+            SumNFourteens(7);  // Now Sum = 14 * 7 = 98 (and Two = 2)
+            return Sum + Two;  // 98 + 2 = 100 on success
+        }
+    }
+}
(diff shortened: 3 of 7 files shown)


JosephTremoulet:
@briansull @hseok-oh PTAL
/cc @RussKeldorph @dotnet/jit-contrib

JosephTremoulet:
I'm kicking off asm diffs now and will report back...

JosephTremoulet:
jit-diff (x86 release --tests --frameworks) shows no diffs
Desktop superpmi shows diffs in just 6 methods, where we're now able to hoist a few more static vars because they are not `CORINFO_FLG_FIELD_INITCLASS` fields.

JosephTremoulet:
@briansull, @dotnet/jit-contrib hopefully someone can take a look today, since this is intended for 2.0

src/jit/morph.cpp
```diff
@@ -6628,9 +6628,10 @@ GenTreePtr Compiler::fgMorphField(GenTreePtr tree, MorphAddrContext* mac)
                 else
 #endif // _TARGET_64BIT_
                 {
-                    // Only volatile could be set, and it maps over
-                    noway_assert((tree->gtFlags & ~(GTF_FLD_VOLATILE | GTF_COMMON_MASK)) == 0);
-                    noway_assert(GTF_FLD_VOLATILE == GTF_IND_VOLATILE);
+                    // Only volatile or classinit could be set, and they map over
+                    noway_assert((tree->gtFlags & ~(GTF_FLD_VOLATILE | GTF_FLD_INITCLASS | GTF_COMMON_MASK)) == 0);
+                    noway_assert(GTF_FLD_VOLATILE == GTF_CLS_VAR_VOLATILE);
+                    noway_assert(GTF_FLD_INITCLASS == GTF_CLS_VAR_INITCLASS);
```
CarolEidt: I see that this is repeating an existing pattern, but it seems weird to have a `noway_assert` for a condition that is statically checkable, though I guess a `#if` and `#error` combination would probably be less readable.

BruceForstall: Use static_assert or static_assert_no_msg instead.

JosephTremoulet: Updated

src/jit/importer.cpp
```diff
@@ -6057,6 +6057,11 @@ GenTreePtr Compiler::impImportStaticFieldAccess(CORINFO_RESOLVED_TOKEN* pResolve
                 // In future, it may be better to just create the right tree here instead of folding it later.
                 op1 = gtNewFieldRef(lclTyp, pResolvedToken->hField);
 
+                if (pFieldInfo->fieldFlags & CORINFO_FLG_FIELD_INITCLASS)
+                {
+                    op1->gtFlags |= GTF_FLD_INITCLASS;
```
briansull: While this location may suffice for the case that we have here,
 I think that it may fail to set the flag on other kinds of GT_FIELD nodes. 
(Like instance fields , can they have or do they need it set?)
(Also what should happen for address of static fields?)

tests/src/JIT/Regression/JitBlue/GitHub_10780/GitHub_10780.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Runtime.CompilerServices;
+
+namespace N
+{
+    public struct Pair
+    {
+        public int Left;
+        public int Right;
+
+        public static Pair TenFour = new Pair() { Left = 10, Right = 4 };
```
briansull: I expected that this construction would reference the current value of C.Two

src/jit/importer.cpp
JosephTremoulet: I searched for other appearances of `gtNewFieldRef` (and the `GenTreeField` constructor) and didn't see any others that should need this.  Instance fields are not triggers for static initialization (the rules are in [ecma 335](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&cad=rja&uact=8&ved=0ahUKEwi65JmEjeTTAhUO3WMKHR1yDbYQFggiMAA&url=https%3A%2F%2Fwww.ecma-international.org%2Fpublications%2Fstandards%2FEcma-335.htm&usg=AFQjCNGmzI4xDYNpbHE274KyfnbDUQPD_g) I.8.9.5).  Address of static gets imported as an `IconHandleNode` (just a few lines down), not a FT_FIELD node.

tests/src/JIT/Regression/JitBlue/GitHub_10780/GitHub_10780.cs
JosephTremoulet: Yeah, the role of `C.Two` in this test is a bit confusing -- I just needed something in the loop that would keep us from hoisting the type init call (which, btw, is the call to init the statics of `Pair`, *not* the statics of `C`), and needed it to not keep registers live across the loop (lest our costing decide not to hoist the reference to `Pair.TenFour`), so writing to a static (which stops us from hoisting the call because it may throw) was what I landed on.  The bug is in reordering the call to `Pair..cctor` vs the load of`Pair.TenFour`, not vs the store of `C.Two`.

PR #11466, opened 2017-05-09
Fix static analysis issues

kouvel:


kouvel:
@janvorli @gkhanna79

src/md/ceefilegen/cceegen.cpp
```diff
@@ -47,10 +47,20 @@ HRESULT CCeeGen::CreateNewInstance(CCeeGen* & pGen) // static, public
     TESTANDRETURNMEMORY(pGen->m_peSectionMan);
 
     HRESULT hr = pGen->m_peSectionMan->Init();
-    TESTANDRETURNHR(hr);
+    if (FAILED(hr))
+    {
+        pGen->Cleanup();
+        delete pGen;
```
janvorli: Can you please use NewHolder<CCeeGen> as a local and use its Extract method to the output parameter in the success case? It is preferred to use holders in the runtime, especially when there are multiple return code paths.

src/vm/ceeload.cpp
```diff
@@ -12757,6 +12757,7 @@ void Module::LogTokenAccess(mdToken token, SectionFormat format, ULONG flagnum)
         GC_NOTRIGGER;
         MODE_ANY;
         PRECONDITION(g_IBCLogger.InstrEnabled());
+        PRECONDITION(flagnum >= 0);
```
janvorli: The flagnum is ULONG so I am not sure how can this comparison help anything. What was the issue?

src/zap/zapheaders.cpp
```diff
@@ -325,7 +325,7 @@ ZapPEExports::ZapPEExports(LPCWSTR dllPath)
 
 DWORD ZapPEExports::GetSize()
 {
-	return DWORD(sizeof(IMAGE_EXPORT_DIRECTORY) + wcslen(m_dllFileName) + 1);
+	return DWORD(sizeof(IMAGE_EXPORT_DIRECTORY) + wcslen(m_dllFileName) * sizeof(BYTE) + 1);
```
janvorli: This change looks strange. Multiplying the wcslen - length of 16 bit string -  by sizeof BYTE?

src/md/ceefilegen/cceegen.cpp
kouvel: Done

src/vm/ceeload.cpp
kouvel: You're right, I have undone the changes relevant to flagnum. The error was about tkKind.

src/zap/zapheaders.cpp
kouvel: It didn't like combining sizeof (byte count) and char length (element count). The scaling is to explicitly indicate that the intention is to use only one byte per char, as done in the Save() function below.

PR #11480, opened 2017-05-09
Fix lldbplugin to properly build with lldb 3.9.

mikem8361:


mikem8361:

src/ToolBox/SOS/lldbplugin/CMakeLists.txt
```diff
@@ -56,7 +56,7 @@ endif()
 
 # Check for LLDB library
 find_library(LLDB NAMES LLDB lldb lldb-4.0 lldb-3.9 lldb-3.8 lldb-3.7 lldb-3.6 lldb-3.5 PATHS "${WITH_LLDB_LIBS}" PATH_SUFFIXES llvm NO_DEFAULT_PATH)
-find_library(LLDB NAMES LLDB lldb lldb-4.0 lldb-3.8 lldb-3.7 lldb-3.6 lldb-3.5 PATH_SUFFIXES llvm)
+find_library(LLDB NAMES LLDB lldb lldb-4.0 lldb-3.9 lldb-3.8 lldb-3.7 lldb-3.6 lldb-3.5 PATH_SUFFIXES llvm)
```
janvorli: @mikem8361  I am sorry for breaking this when I've enabled the 4.0. But seeing this break, it seems it would be safer for the future to extract the list `lldb lldb-4.0 lldb-3.9 lldb-3.8 lldb-3.7 lldb-3.6 lldb-3.5` into a separate variable and use it in both find_library.

mikem8361: I hope it is alright if I don't add the variable because almost all the CI tests are finished and I need to get this in tonight.

janvorli: Sure, we can do it any time later. It would be just a cleanup. 
LGTM.

PR #11435, opened 2017-05-06
Re-Factor EventSource to Support Writing to EventPipe

brianrob:
Re-factor EventSource and EventProvider to allow for writing to the EventPipe on Linux.

This change includes the following functionality in addition to the refactoring work:

- Create EventPipeProvider on EventSource initialization.
- Properly plumb IsEnabled checks via callbacks/commands.
- Delete EventPipeProvider on EventSource dispose.

A subsequent PR from @helloguo will contain the rest of the EventSource functionality to enable writing to EventPipe.

PR #11481, opened 2017-05-09
Write EventSource events to EventPipe

helloguo:
This PR writes EventSource events to EventPipe.


helloguo:
@brianrob FYI

brianrob:
@vancem, PTAL.  Specifically commit https://github.com/dotnet/coreclr/pull/11481/commits/1f9947e55aa7cf0472c5140554bff25a3d135db4 is @helloguo's addition to my PR on EventSource (https://github.com/dotnet/coreclr/pull/11435).

@valenis, FYI.

brianrob:
@helloguo can you start looking into CI failures?

helloguo:
@brianrob Fixed.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -692,6 +689,35 @@ protected EventSource(EventSourceSettings settings, params string[] traits)
...
+            Debug.Assert(m_eventData != null);
+            int cnt = m_eventData.Length;
+            for (int i = 0; i < cnt; i++)
+            {
+                Int64 keywords = m_eventData[i].Descriptor.Keywords;
+                uint eventID = (uint)m_eventData[i].Descriptor.EventId;
+                uint eventVersion = m_eventData[i].Descriptor.Version;
+                uint level = m_eventData[i].Descriptor.Level;
+                // TODO: what's the value of needStack?
+                bool needStack = false;
```
vancem: needStack is not a attribute of an event (it can be removed).   Basically any event can have a stack and it is up to the consumer to decide (as long as you are still on the thread, you can take a stacks later).  

I would probably simply drop needStack.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -692,6 +689,35 @@ protected EventSource(EventSourceSettings settings, params string[] traits)
...
+                
+                // Note: it's not clear that elements in m_eventData are unique or not
+                // so that we use a list allEvents to record the unique elements. 
+                if(allEvents.Contains(eventID))
```
vancem: With the exception of ID=0 (which is an 'illegal' value), eventIDs need to be unique (EventSource does check for this because it is an easy mistake).   Thus you can rip out the allEvents (or leave it as debug-only code).    Some EventSources can have a large number of events (> 10K), so we really don't want something here that will make this N*N.

src/mscorlib/shared/System/Diagnostics/Tracing/IEventProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+            Guid* activityId,
+            Guid* relatedActivityId,
+            int userDataCount,
+            EventProvider.EventData* userData);
+
+        // Get or set the per-thread activity ID.
+        int EventActivityIdControl(UnsafeNativeMethods.ManifestEtw.ActivityControl ControlCode, ref Guid ActivityId);
+
+        // Register an EventPipeEvent handle to this EventPipeEventProvider.
+        unsafe void AddEventHandle(Int64 keywords, uint eventID, uint eventVersion, uint level, bool needStack);
```
vancem: an we renamed this to 'DefineEvent' or something like that?   It took me a while to understand the purpose.   I would also put the eventID first as it is the 'handle'.   

Is the plan to also pass information about the event parameters in this API?   It seems like we want to pass the name and 'signature description' (probably already serialized as it is easier to do this on the managed side.

src/vm/ecalllist.h
```diff
@@ -1271,6 +1271,15 @@ FCFuncStart(gEventLogger)
 FCFuncEnd()
 #endif // defined(FEATURE_EVENTSOURCE_XPLAT)
 
+#ifdef FEATURE_PERFTRACING
+FCFuncStart(gEventPipeInternalFuncs)
+    QCFuncElement("CreateProvider", EventPipeInternal::CreateProvider)
+    QCFuncElement("AddEvent", EventPipeInternal::AddEvent)
```
vancem: AddEvent -> DefineEvent or DefineEventSchema.

helloguo:
superseded by https://github.com/dotnet/coreclr/pull/11577

PR #11484, opened 2017-05-10
Disable a few long-running tests.

pgavlin:
These tests consistently time out under GC stress or heap verification.
Each disabled test is known to complete successfully given adequate
time.


pgavlin:
@RussKeldorph @jashook @dotnet/jit-contrib PTAL

PR #11486, opened 2017-05-10
[RyuJIT/ARM32] Fix assertion in 'unreached' when using emitIns_R_L

sjsinju:
Now, All of CodeGenBringUpTests have failed.

message : Assertion failed 'unreached' in 'DomainNeutralILStubClass:IL_STUB_PInvoke(ref)'

I found the problem was caused by #11377 that 'genCodeForTreeNode' is merged to codegenarmarch.cpp.

### DIFF
--- a/src/jit/codegenarmarch.cpp
+++ b/src/jit/codegenarmarch.cpp
@@ -344,7 +344,7 @@ void CodeGen::genCodeForTreeNode(GenTreePtr treeNode)
             genPendingCallLabel       = genCreateTempLabel();
             treeNode->gtLabel.gtLabBB = genPendingCallLabel;
 #if defined(_TARGET_ARM_)
-            emit->emitIns_J_R(INS_adr, EA_PTRSIZE, genPendingCallLabel, treeNode->gtRegNum);
+            emit->emitIns_J_R(INS_adr, EA_PTRSIZE, genPendingCallLabel, targetReg);
 #elif defined(_TARGET_ARM64_)
             emit->emitIns_R_L(INS_adr, EA_PTRSIZE, genPendingCallLabel, targetReg);
 #endif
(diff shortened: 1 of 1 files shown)


src/jit/codegenarmarch.cpp
```diff
@@ -343,7 +343,11 @@ void CodeGen::genCodeForTreeNode(GenTreePtr treeNode)
         case GT_LABEL:
             genPendingCallLabel       = genCreateTempLabel();
             treeNode->gtLabel.gtLabBB = genPendingCallLabel;
+#if defined(_TARGET_ARM)
```
mskvortsov: `_TARGET_ARM_` as well as `_TARGET_ARM64_` should both have an ending underscore.

src/jit/codegenarmarch.cpp
```diff
@@ -343,7 +343,11 @@ void CodeGen::genCodeForTreeNode(GenTreePtr treeNode)
...
+            emit->emitIns_J_R(INS_adr, EA_PTRSIZE, genPendingCallLabel, treeNode->gtRegNum);
```
mskvortsov: Why not simply `targetReg` instead of `treeNode->gtRegNum`?

mskvortsov: And generally speaking, as this usage of `emitIns_J_R()` was copied from `codegenlegacy.cpp`, may it be worth to refactor arm32 `emitIns_R_L()` instead? Though I'm not sure what is the difference intended between the `J_R` and `R_L` forms.

src/jit/codegenarmarch.cpp
sjsinju: Thank you. That's my mistake.

src/jit/codegenarmarch.cpp
sjsinju: @mskvortsov  I don't know deeply. But I just copied that from the original source of codegenarm.cpp. And then, The assertion has resolved. 
It seems to need more investigate deeply like what you said for using emitIns_R_L insted of emitIns_J_R().

BruceForstall:
Sorry about that. We should really fix arm32/arm64 to use the same function name in this case (probably _R_L)

PR #11488, opened 2017-05-10
Improvements to LinkBench

swaroop-sridhar:
This change has the following improvements to LinkBench

1) Move the testing to the latest version of ILLink package
   (for all benchmarks except for Roslyn, which is still on netcoreapp1.1)
2) Support for LinkBench to run individual benchmarks in the suite --
   helps local testing.
3) Support for running without clean, cloning, build etc --
   for local runs during continuous development.
4) Support to obtain the correct version(s) of DotNet CLI required to run
   the benchmarks
5) Get-Crossgen.ps1 is required to crossgen MusicStore binaries (before/after linking).
   Since it was recently dropped from JitBench repo, it is added to linkBench as an asset.
6) Add the --perf:runid parameter to generate output file names suitable
   to upload the results to BenchView.

### DIFF
--- a/tests/src/performance/linkbench/linkbench.cs
+++ b/tests/src/performance/linkbench/linkbench.cs
@@ -1,3 +1,5 @@
+using Microsoft.Xunit.Performance;
+using Microsoft.Xunit.Performance.Api;
 using System;
 using System.Diagnostics;
 using System.Collections.Generic;
@@ -8,8 +10,6 @@ using System.Text;
 using System.Globalization;
 using System.Linq;
 using System.Xml.Linq;
-using Microsoft.Xunit.Performance;
-using Microsoft.Xunit.Performance.Api;
 using Xunit;
 using Xunit.Abstractions;
 
@@ -18,7 +18,8 @@ namespace LinkBench
     public class Benchmark
     {
         public string Name;
-
+        public bool ShouldRun;
+        public string ProjFile;
         public string UnlinkedDir;
         public string LinkedDir;
         public double UnlinkedMsilSize;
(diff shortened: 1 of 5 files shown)


swaroop-sridhar:
@sbomer please review

tests/src/performance/linkbench/linkbench.cs
```diff
@@ -126,22 +128,103 @@ private double ComputeCertDiff()
...
+            try
+            {
+                AssemblyLoadContext.GetAssemblyName(file);
+            }
+            catch (Exception)
+            {
+                // We should really check only for BadImageFormatException
+                // However, constructing and throwing a BadImageFormatException 
+                // throws missingmethod exception, so this is a work-around, 
+                // until that problem is fixed.
```
sbomer: Do you know why this is, or what changed?

tests/src/performance/linkbench/linkbench.cs
```diff
@@ -155,75 +238,162 @@ public class LinkBench
...
+                "LinkBench\\corefx\\bin\\ILLinkTrimAssembly\\netcoreapp-Windows_NT-Release-x64\\trimmed"),
+            new Benchmark("Roslyn",
+                "LinkBench\\roslyn\\Binaries\\Release\\Exes\\CscCore\\win7-x64\\publish",
+                "LinkBench\\roslyn\\Binaries\\Release\\Exes\\CscCore\\win7-x64\\Linked")
+        };
+
+        static int UsageError()
+        {
+            Console.WriteLine("Usage: LinkBench [--nosetup] [--nobuild] [--perf:runid <id>] [<benchmarks>]");
+            return -4;
```
sbomer: It would be nice to add descriptions of the flags, and a list of available benchmarks.

tests/src/performance/linkbench/linkbench.cs
```diff
@@ -155,75 +238,162 @@ public class LinkBench
...
+                    {
+                        if (String.Compare(args[i], benchmark.Name, true) == 0)
+                        {
+                            benchmark.SetToRun();
+                            benchmarkSpecified = true;
+                            break;
+                        }
+                    }
+
+                    if(!benchmarkSpecified)
```
sbomer: Add a space after if for consistency

tests/src/performance/linkbench/linkbench.cs
```diff
@@ -155,75 +238,162 @@ public class LinkBench
...
+                    {
+                        Console.WriteLine("Unknown Benchmark {0}", args[i]);
+                    }
                 }
             }
 
+            // If benchmarks are not explicitly specified, run all benchmarks
+            if(!benchmarkSpecified)
```
sbomer: Add space after if

sbomer:
I think we should crossgen MusicStore binaries following the current jitbench instructions instead of keeping our own copy of Get-Crossgen.ps1.

sbomer:
Just wanted to summarize our conversation about the need for Get-Crossgen.ps1. We need it for the following reasons:
- Since the musicstore benchmark has a workaround to use an old framework (to get non-crossgen'd packages), we need to run crossgen on these assemblies manually for now. Even once we have the linker running on R2R assemblies and remove this workaround, we'll need a way to get the pre-crossgen assemblies for the size comparison.
- We need to use it to crossgen the linked assemblies for the size comparison, since the linker targets don't yet include a way to crossgen the linked assemblies.

PR #11492, opened 2017-05-10
Fix linker GUARD option

janvorli:
I have found that the `/GUARD:CF` option set in the root CMakeFiles.txt
was not being propagated to the Visual C++ project files generated
by Cmake because of a wrong casing. The CMakeFiles.txt used `/guard:cf`
but it works correctly only if it is all upper case.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -312,8 +312,8 @@ if (WIN32)
     endforeach (Definition)
   endforeach (Config)
 
-  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /guard:cf")
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /guard:cf")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /GUARD:CF")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /GUARD:CF")
 
   # Linker flags
   #


janvorli:
@gkhanna79 can you take a look please? The issue was discovered by binscope.

gkhanna79:
@janvorli What did BinSkim specifically complain about? Doing "dumpbin /headers /loadconfig" does confirm that CFG sections and tables are present.

Also, is this case-sensitivity issue unique to the linker and not the compiler (that sounds really strange)? And if we were to fix this, I think you should prep PRs for CoreFX and Core-Setup to fix the same as well for 2.0 (and also for 1.0.x and 1.1.x - though they can come in next week).

janvorli:
@gkhanna79 binskim said for all of our dlls:
```
xxx.dll does not enable the control flow guard (CFG) 
To resolve this issue, pass /GUARD:CF on both the 
and linker command lines. Binaries also require 
/DYNAMICBASE linker option in order to enable CFG.
```
It is not a compiler / linker issue, but rather a CMake one. By looking at the generated vcxproj's I have verified that there was no /GUARD:CF linker option present. After changing the /guard:cf in the root CMakeLists.txt to upper case and re-running build, the option appeared there and binskim stopped complaining.
It could be that the problem is caused by the version of Cmake that I have installed (3.7.2).

gkhanna79:
Thanks @janvorli. Can you please try with CMake 3.5.2 which is what official lab has installed?

I distinctly remember that when I made https://github.com/dotnet/core-setup/commit/6c99e7e0281ee34e555d1423bab570376456ba34#diff-ebb5817454bbf398d6df8cb6e3ad7ac0, dumpbin was showing the expected information as described at https://msdn.microsoft.com/en-us/library/windows/desktop/mt637065(v=vs.85).aspx under the `How Do I Tell That a Binary is under Control Flow Guard?` section. Can you confirm if you see that the binary in question, without any changes from you, fail the test mentioned at this page?

Also, which package did you use to validate this change against?

janvorli:
@gkhanna79 I was running binskim on the binaries built in the coreclr repo, I was not using any package. 
Now I have tried to run binskim on the latest 2.0.0 preview package and it found everything being ok, so that would confirm that it works fine with cmake 3.5.2 that we use in the lab.

I will check with cmake 3.5.2 locally to confirm that it is a problem of the cmake 3.7.2.

janvorli:
@gkhanna79 I can confirm that running the `dumpbin /headers /loadconfig .\bin\obj\Windows_NT.x64.Debug\src\dlls\mscoree\coreclr\Debug\coreclr.dll` with binaries built with cmake 3.7.2 without my change doesn't show the "Guard" under the DLL characteristics:
```
            160 DLL characteristics
                  High Entropy Virtual Addresses
                  Dynamic base
                  NX compatible
```
I am going to try with cmake 3.5.2.

janvorli:
@gkhanna79 with CMake 3.5.2, it works correctly even with the option being specified in the CMakeFiles.txt as lowercase. 
I have also tried it with the lastest cmake 3.8.1 and it behaves the same way as 3.7.2. That means that the linker guard option is ignored if it is specified in the CMakeLists.txt as lowercase. So that confirms it is cmake >= 3.7.2 specific.

gkhanna79:
Thanks for the confirmation that everything is in order, as expected, @janvorli.  I dont know if that is a bug in CMake 3.7.2+ or by design. 

Either ways, to protect ourselves from a future update of CMake, should we just go ahead and make everything uppercase (as you have in this PR, for all the master branches in the 3 repos)?

jkotas:
Newer version of CMake will translate the `/guard:cf` option to the proper msbuild option. If you build with CMake 3.8, you should see the following in the vcxproj files generated by cmake:

```
      <ControlFlowGuard>Guard</ControlFlowGuard>
...
      <LinkControlFlowGuard>Guard</LinkControlFlowGuard>
```

(https://github.com/Kitware/CMake/blob/debeb07a4005dfe4ea506041f2b2b11ee9bb2d09/Source/cmVS141CLFlagTable.h#L63 is the code that does that.)

The C++ targets in `C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.CppCommon.targets` should turn `ControlFlowGuard` into passing the right switch to the compiler. Why it is not happening?

janvorli:
@jkotas part of the problem is that for some weird reason, cmake doesn't translate the Cpp `/guard:cf` option to the `<ControlFlowGuard>` node and puts it into the `<AdditionalOptions>` node instead. As far as I understand the `C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.CppCommon.targets`, it should transform the `<ControlFlowGuard>` from the compiler options to the `<LinkControlFlowGuard>` of the linker options in case the explicit `<LinkControlFlowGuard>` was not provided in the original vcxproj. 
The interesting thing is that if I edit the original vcxproj by removing the `/guard:cf` from the  `<AdditionalOptions>` and create a  `<ControlFlowGuard>` node instead and then rebuild the project (e.g. coreclr.vcxproj), it works correctly. 
So the remaining question is why doesn't the explicitly specified `<LinkControlFlowGuard>`  in the original vcxproj kick in.

gkhanna79:
@janvorli And it (magically) works if the casing is uppercase?

janvorli:
@gkhanna79 well, if the case is uppercase, cmake doesn't recognize it and passes it in linker's `<AdditionalOptions>` instead of creating the `<LinkControlFlowGuard>` node. And so it works.

jkotas:
So this means:
- There is a bug in Visual Studio 2015 C++ msbuild targets/tasks that they do not respect LinkControlFlowGuard property
- This bug is fixed in Visual Studio 2017
- You will run into this bug if you use latest cmake + Visual Studio 2015 because of latest cmake generates LinkControlFlowGuard property

Your change is working around this bug.

Then it looks good to me.

jkotas:
It maybe nice to update the commit description to match. Also, this change should be made in all repos as Gaurav said.

janvorli:
@jkotas after instrumenting the Microsoft.CppCommon.targets a bit, I believe I have found the real reason why the `<LinkControlFlowGuard>` added by the cmake doesn't work. The issue is that cmake adds it to a wrong place. It puts it under the `<Link>` node, but it should be placed under a `<PropertyGroup>`.  The `<Link>` node actually has an attribute named `ControlFlowGuard`, not `<LinkControlFlowGuard>`. And the code in the Microsoft.CppCommon.targets sets it this way:
```xml
    <Link
....
....
        ControlFlowGuard               ="$(LinkControlFlowGuard)"
```

jkotas:
If it is the case, I do not understand why it works fine with Visual Studio 2017 + latest cmake even though LinkControlFlowGuard is under Link node there as well:

```
    <Link>
      <LinkControlFlowGuard>Guard</LinkControlFlowGuard>
    </Link>
```

jkotas:
Either way, I think that this workaround is fine. We can trace down and file the bug later.

janvorli:
@jkotas I've forgotten to mention that I've tried to modify one of the generated vcxprojs by moving the LinkControlFlowGuard tag out of the Link node to a PropertyGroup one and it fixed the issue.

As for why it works under VS 2017, I don't know. I will give it a try to see what's going on there. It is possible that for VS 2017, cmake correctly generates the ControlFlowGuard node under the ClCompile node instead of placing the /guard:cf into the AdditionalOptions. That also fixes the problem, since the LinkControlFlowGuard property is then generated from that ControlFlowGuard node by the Microsoft.CppCommon.targets .

jkotas:
> It is possible that for VS 2017, cmake correctly generates the ControlFlowGuard node

Yes, it is the case. It explains it then. 

So I agree that it is a bug in cmake. @khouzam should be able to help to get it fixed.

danmosemsft:
@janvorli should we merge this ?

janvorli:
@danmosemsft I have completely forgotten about this PR I've made. I have verified now that it doesn't break anything in VS2017 and merged it.

danmosemsft:
Thanks @janvorli

gkhanna79:
@janvorli Can you please make the same fix in CoreFX and Core-Setup as well?

janvorli:
@gkhanna79 sure!

danmosemsft:
@gkhanna79 should these go in 2.0 also?

gkhanna79:
No, this is not required for 2.0.0 since this is workaround/fix for an issue in newer CMake versions,  incase someone were to use them, which we do not support for 2.0.0 builds.

PR #11493, opened 2017-05-10
Add upload-packages.sh file

jyoungyun:
This script file is used for uploading CoreCLR nupkgs to a specific nuget or myget server.

/cc @gkhanna79 @hqueue

### DIFF
--- a/upload-packages.sh
+++ b/upload-packages.sh
@@ -1,0 +1,78 @@
+#!/usr/bin/env bash
+
+usage()
+{
+    echo "Upload the NuGet packages to the specified nuget or myget server."
+    echo "For uploading to nuget or myget server the following properties are required."
+    echo "   -FeedURL=\"NuGet or MyGet feed URL\""
+    echo "   -SymbolURL=\"NuGet or MyGet symbol server URL\""
+    echo "   -APIKey=\"NuGet or MyGet feed API key\""
+    echo "Configuration can be Release, Checked, or Debug"
+    echo "Architecture can be x64, x86, arm, armel, or arm64"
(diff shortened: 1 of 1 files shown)


jyoungyun:
@gkhanna79 Could you review this script file? When building CoreCLR in tizen build infra, it get the coreclr code of github and execute the actual build command in it (within docker). After this process, nupkg needs to be uploaded to a specific nuget or myget server, but a separate script is required for this. I can embed this script in Jenkins code, or I can put it in github code. If you agree to include it in the github code, I will make a similar PR to corefx and core-setup. How to use is as follows.
```
./upload-packages.sh -FeedURL=https://tizen.myget.org/F/dotnet-core/api/v3/index.json -SymbolURL=https://tizen.myget.org/F/dotnet-core/symbols/api/v2/package -APIKey=<secret> Release armel
```

gkhanna79:
@jyoungyun Since this is required for Tizen build infrastructure, is this better suited to be committed into one of your internal Tizen repos as opposed to the public CoreCLR repo (since it is not going to be used by it)?

Or, am I missing something?

jyoungyun:
@gkhanna79 Yes, I can upload this script in our internal Tizen repos. I suggested this script because it was not only Tizen, but also because later on it would make sence for other OS to put those nupkgs on their myget. However, it is only need in Tizen at the moment, and I can only insert it as internal to our side. If you want, I wll have a script internally. :)

gkhanna79:
Yes, for now, it will be good to keep it in Tizen repo and we can look into this for future as the need arises.

I will close this PR for now.

PR #11495, opened 2017-05-10
[RyuJIT/ARM32] Fix GT_NEG decomposition

mskvortsov:
Fixes these tests:
```
JIT/Directed/perffix/primitivevt/mixed1_cs_d
JIT/Directed/perffix/primitivevt/mixed1_cs_r
JIT/IL_Conformance/Old/Conformance_Base/ldc_neg_i8
JIT/Methodical/fp/exgen/1000w1d_cs_d
JIT/Methodical/fp/exgen/1000w1d_cs_do
JIT/Methodical/fp/exgen/1000w1d_cs_r
JIT/Methodical/fp/exgen/1000w1d_cs_ro
JIT/Methodical/fp/exgen/10w5d_cs_d
JIT/Methodical/fp/exgen/10w5d_cs_r
JIT/Methodical/fp/exgen/200w1d-02_cs_d
JIT/Methodical/fp/exgen/200w1d-02_cs_do
JIT/Methodical/fp/exgen/200w1d-02_cs_r
JIT/Methodical/fp/exgen/200w1d-02_cs_ro
JIT/Methodical/MDArray/DataTypes/long_cs_d
JIT/Methodical/MDArray/DataTypes/long_cs_r
...
```

[codegenlegacy.cpp#L14500-L14513](https://github.com/dotnet/coreclr/blob/c78891ae1cc668a77377b2b0fb03c7e1a6894b29/src/jit/codegenlegacy.cpp#L14500-L14513):
```c++
// ARM doesn't have an opcode that sets the carry bit like
// x86, so we can't use neg/addc/neg.  Instead we use subtract
// with carry.  Too bad this uses an extra register.
...
inst_RV_IV(INS_rsb, regLo, 0, EA_4BYTE, INS_FLAGS_SET);
getEmitter()->emitIns_R_R_R_I(INS_sbc, EA_4BYTE, regHi, regZero, regHi, 0);
```

cc @dotnet/arm32-contrib

### DIFF
--- a/src/jit/decomposelongs.cpp
+++ b/src/jit/decomposelongs.cpp
@@ -922,7 +922,7 @@ GenTree* DecomposeLongs::DecomposeNeg(LIR::Use& use)
     loResult->gtType     = TYP_INT;
     loResult->gtOp.gtOp1 = loOp1;
 
-    GenTree* zero     = m_compiler->gtNewZeroConNode(TYP_INT);
+    GenTree* zero = m_compiler->gtNewZeroConNode(TYP_INT);
 #if defined(_TARGET_X86_)
     GenTree* hiAdjust = m_compiler->gtNewOperNode(GT_ADD_HI, TYP_INT, hiOp1, zero);
     GenTree* hiResult = m_compiler->gtNewOperNode(GT_NEG, TYP_INT, hiAdjust);
--- a/src/jit/decomposelongs.cpp
+++ b/src/jit/decomposelongs.cpp
@@ -923,10 +923,14 @@ GenTree* DecomposeLongs::DecomposeNeg(LIR::Use& use)
     loResult->gtOp.gtOp1 = loOp1;
 
     GenTree* zero     = m_compiler->gtNewZeroConNode(TYP_INT);
+#if defined(_TARGET_X86_)
     GenTree* hiAdjust = m_compiler->gtNewOperNode(GT_ADD_HI, TYP_INT, hiOp1, zero);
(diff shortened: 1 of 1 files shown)


mikedn:
> ARM doesn't have an opcode that sets the carry bit like x86, so we can't use neg/addc/neg.

Hmm, it seems to me that ARM's `RSB reg, 0` is the same thing as X86's `NEG reg`.

mskvortsov:
These two instructions have different behaviour w.r.t. setting the carry flag.
X86:

> The CF flag set to 0 if the source operand is 0; otherwise it is set to 1.

ARMv7:
```
RSB{S}{<c>}{<q>} {<Rd>,} <Rn>, #<const>
(result, carry, overflow) = AddWithCarry(NOT(R[n]), imm32, ‘1’);

// AddWithCarry()
// ==============
(bits(N), bit, bit) AddWithCarry(bits(N) x, bits(N) y, bit carry_in)
    unsigned_sum = UInt(x) + UInt(y) + UInt(carry_in);
    signed_sum = SInt(x) + SInt(y) + UInt(carry_in);
    result = unsigned_sum<N-1:0>; // same value as signed_sum<N-1:0>
    carry_out = if UInt(result) == unsigned_sum then '0' else '1';
    overflow = if SInt(result) == signed_sum then '0' else '1';
    return (result, carry_out, overflow);
```

RussKeldorph:
Yeah, IIRC, this is one of the few places the RSC instruction comes in handy, but they sacrificed it when designing the 32-bit Thumb encodings so they could add ORN.  Apparently the latter is more useful.

RussKeldorph:
@dotnet/jit-contrib

PR #11496, opened 2017-05-10
Allow multiple SetData calls on AppDomain

AlexGhiondea:
/cc @sepidehMS

### DIFF
--- a/src/mscorlib/src/System/AppDomain.cs
+++ b/src/mscorlib/src/System/AppDomain.cs
@@ -579,19 +579,14 @@ namespace System
                 throw new ArgumentNullException(nameof(name));
             Contract.EndContractBlock();
 
-            if (name.Equals(AppDomainSetup.LoaderOptimizationKey))
-                return FusionStore.LoaderOptimization;
-            else
+            object data;
+            lock (((ICollection)LocalStore).SyncRoot)
             {
-                object data;
-                lock (((ICollection)LocalStore).SyncRoot)
-                {
-                    LocalStore.TryGetValue(name, out data);
-                }
-                if (data == null)
-                    return null;
-                return data;
+                LocalStore.TryGetValue(name, out data);
             }
+            if (data == null)
+                return null;
+            return data;
         }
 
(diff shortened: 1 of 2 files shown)


sepidehMS:
Thanks for doing this @AlexGhiondea

jkotas:
You may also want to get rid of `AppDomainSetup.Locate` and the special casing from GetData, so that the following works same way as on desktop:

```
static void Main() {
    AppDomain.CurrentDomain.SetData("APPBASE", "c:\\in the middle of nowhere");
    Console.WriteLine(AppDomain.CurrentDomain.GetData("APPBASE"));
}
```

It should print `c:\in the middle of nowhere`.

AlexGhiondea:

src/mscorlib/src/System/AppDomainSetup.cs
```diff
@@ -29,7 +29,7 @@ internal sealed class AppDomainSetup
         internal enum LoaderInformation
         {
             // If you add a new value, add the corresponding property
-            // to AppDomain.GetData() and SetData()'s switch statements,
+            // to AppDomain.pa() and SetData()'s switch statements,
```
justinvp: pa?

AlexGhiondea: *sigh*

AlexGhiondea: fixed.

AlexGhiondea: @justinvp good catch!

src/mscorlib/src/System/AppDomain.cs
```diff
@@ -590,37 +579,18 @@ public Object GetData(string name)
...
-                    lock (((ICollection)LocalStore).SyncRoot)
-                    {
-                        LocalStore.TryGetValue(name, out data);
-                    }
-                    if (data == null)
-                        return null;
-                    return data;
-                }
-            }
+            if (name.Equals(AppDomainSetup.LoaderOptimizationKey))
```
jkotas: This should be removed as well. (LoaderOptimization is not used for anything in CoreCLR.)

Issue #8974, opened 2017-01-17
Add version.txt file to CoreCLR packages


cesarbs:
CoreFX packages contain a version.txt with the commit the package was built from. This is really useful to determine if a given change was included in the package.

CoreCLR packages don't have that file, making it hard to determine if a change made it to a recent build. Would be nice to have it there too.

weshaggard:
@chcosta is this something we just haven't enabled in coreclr yet?

chcosta:
There will need to be some slight refactoring to make this work for CoreFx and CoreClr.  Currently, the code path is based on execution of a "BuildAllProjects" target.  

https://github.com/dotnet/buildtools/blob/a3c0637b9ae8d322078aff64fae97ee2dbcfecab/src/Microsoft.DotNet.Build.Tasks/PackageFiles/versioning.targets#L218

This should be refactored in some other way so that it is guaranteed to run because the BuildAllProjects target isn't ubiquitous across repos (even across BuildTools-enabled repos).  Once it is running, there also needs to be an update to the coreclr packaging to include the version.txt file; though really, we should move that step out of corefx and coreclr and try to include it in BuildTools packaging targets as well.

halter73:
FWIW, I found that coreclr.dll has the commit SHA in its "Product version". I don't know if libcoreclr.so also has the also SHA embedded somehow.

janvorli:
The libcoreclr.so has the version and hash embedded in it, so there is no need to add the version.txt. You can get it this way:
strings libcoreclr.so | grep "@(#)"

chcosta:
Thanks @janvorli .  CoreFx does a similar thing and bakes in commit hash information into it's dll's (managed dll's for sure, i'd have to look to see if it's happening in native binaries as well).  I think there's value in having a consistent way to find this information, so that you don't have to remember to check the embedded information or a txt file.  Whether that consistent way is binary embedding, a plain text file, or both; I have no preference, I just think we should try to be consistent.

danmosemsft:
@gkhanna79 @chcosta is anyone taking this on for 2.0? if not we should move it out.

danmosemsft:
MOving it out.

PR #11497, opened 2017-05-10
Add version.txt to the coreclr transport package

joperezr:
cc: @ericstj @weshaggard @gkhanna79 

Adding version.txt file to the transport CoreClr packages. This will help us get the core-setup Microsoft.NETCore.App package contain a txt file with the hashes of the packages it used when being built.

### DIFF
--- a/src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/Microsoft.NETCore.Runtime.CoreCLR.pkgproj
+++ b/src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/Microsoft.NETCore.Runtime.CoreCLR.pkgproj
@@ -18,11 +18,4 @@
   <Import Condition="'$(_packageTargetOSGroup)' != ''"  Project="$(MSBuildThisFileDirectory)runtime.$(_packageTargetOSGroup).$(MSBuildProjectName).props" />
   
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
-  <ItemGroup>
-    <!-- Add version file to packages -->
-    <File Condition="Exists('$(SyncInfoFile)')"
-          Include="$(SyncInfoFile)">
-          <SkipPackageFileCheck>true</SkipPackageFileCheck>
-    </File>
-  </ItemGroup>
 </Project>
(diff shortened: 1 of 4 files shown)


ericstj:
Any reason why we don't just do this in targets/props for all packages in coreclr?

src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/Microsoft.NETCore.Runtime.CoreCLR.pkgproj
```diff
@@ -18,4 +18,11 @@
   <Import Condition="'$(_packageTargetOSGroup)' != ''"  Project="$(MSBuildThisFileDirectory)runtime.$(_packageTargetOSGroup).$(MSBuildProjectName).props" />
   
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
+  <ItemGroup>
```
joperezr: If we want this file in all of our packages, then I can move this piece to dir.targets shared by all packages.

weshaggard: I think we should get this file included in all the packages we produce. We have issue https://github.com/dotnet/coreclr/issues/8974 which is about adding this.

weshaggard: Just put it https://github.com/dotnet/coreclr/blob/master/dir.props#L190 with the other common package files.

joperezr: I believe I can't add it there, because SyncInfoFile is not yet defined, which is why I'll add it in dir.targets

joperezr:
> Any reason why we don't just do this in targets/props for all packages in coreclr?

done.

src/.nuget/Microsoft.NETCore.Runtime.CoreCLR/Microsoft.NETCore.Runtime.CoreCLR.pkgproj
weshaggard: It will be there after the first time that target runs to generate it. This is what we do in corefx. At any rate I don't have a concern with it being where you have it.

joperezr:
Merging now since I need this change in coresetup.

PR #11505, opened 2017-05-10
Fix AssemblyName version

sepidehMS:
Fixes https://github.com/dotnet/corefx/issues/19179 by using the managed CoreRT implementation of AssemblyName formatting. Only it checkes for [`canonicalizedVersion.Revision != ushort.MaxValue`](https://github.com/dotnet/coreclr/compare/master...sepidehMS:FixAssemblyFullName?expand=1#diff-fcc2014f65447c529681c711311362b9R45) and does not use RuntimeAssemblyName.

cc: @jkotas @AtsushiKan @danmosemsft

### DIFF
--- a/src/mscorlib/System.Private.CoreLib.csproj
+++ b/src/mscorlib/System.Private.CoreLib.csproj
@@ -86,11 +86,9 @@
   <PropertyGroup Condition="'$(OsEnvironment)' == 'Unix'">
     <DebugType>portable</DebugType>
   </PropertyGroup>
-
   <PropertyGroup Condition="'$(TargetsOSX)' == 'true'">
     <DefineConstants>PLATFORM_OSX;$(DefineConstants)</DefineConstants>
   </PropertyGroup>
-
   <!-- Assembly attributes -->
   <PropertyGroup>
     <AssemblyName>System.Private.CoreLib</AssemblyName>
@@ -455,6 +453,10 @@
     <Compile Include="$(BclSourcesRoot)\System\Reflection\Emit\EnumBuilder.cs" />
     <Compile Include="$(BclSourcesRoot)\System\Reflection\Emit\TypeToken.cs" />
     <Compile Include="$(BclSourcesRoot)\System\Reflection\Emit\XXXOnTypeBuilderInstantiation.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="$(BclSourcesRoot)\System\Reflection\Runtime\Assemblies\AssemblyNameHelpers.cs" />
+    <Compile Include="$(BclSourcesRoot)\System\Reflection\Runtime\Assemblies\AssemblyNameLexer.cs" />
   </ItemGroup>
   <ItemGroup>
     <Compile Include="$(BclSourcesRoot)\System\Reflection\Metadata\AssemblyExtensions.cs" />
--- a/src/mscorlib/src/System/Reflection/Runtime/Assemblies/AssemblyNameHelpers.cs
+++ b/src/mscorlib/src/System/Reflection/Runtime/Assemblies/AssemblyNameHelpers.cs
@@ -37,7 +37,25 @@ namespace System.Reflection.Runtime.Assemblies
                 if (canonicalizedVersion.Major != ushort.MaxValue)
                 {
                     sb.Append(", Version=");
-                    sb.Append(a.Version.ToString());
+                    sb.Append(canonicalizedVersion.Major);
+
+                    if(canonicalizedVersion.Minor != ushort.MaxValue)
+                    {
+                        sb.Append('.');
+                        sb.Append(canonicalizedVersion.Minor);
+
+                        if(canonicalizedVersion.Build != ushort.MaxValue)
+                        {
+                            sb.Append('.');
+                            sb.Append(canonicalizedVersion.Build);
+
+                            if(canonicalizedVersion.Revision != ushort.MaxValue)
+                            {
+                                sb.Append('.');
+                                sb.Append(canonicalizedVersion.Revision);
+                            }
+                        }
+                    }
                 }
             }
 
--- a/src/mscorlib/src/System/Reflection/Runtime/Assemblies/AssemblyNameHelpers.cs
+++ b/src/mscorlib/src/System/Reflection/Runtime/Assemblies/AssemblyNameHelpers.cs
@@ -1,0 +1,156 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+/*============================================================
+**
+  Type:  AssemblyNameHelpers
+**
+==============================================================*/
+
+using System;
+using System.Globalization;
+using System.IO;
+using System.Text;
+using System.Collections.Generic;
+
+namespace System.Reflection.Runtime.Assemblies
+{
+    internal static class AssemblyNameHelpers
+    {
+        private const int PUBLIC_KEY_TOKEN_LEN = 8;
+
+        public static String ComputeDisplayName(AssemblyName a)
+        {
+            if (a.Name == String.Empty)
+                throw new FileLoadException();
+
+            StringBuilder sb = new StringBuilder();
+            if (a.Name != null)
+            {
+                sb.AppendQuoted(a.Name);
+            }
+
+            if (a.Version != null)
+            {
+                Version canonicalizedVersion = a.Version.CanonicalizeVersion();
+                if (canonicalizedVersion.Major != ushort.MaxValue)
+                {
+                    sb.Append(", Version=");
+                    sb.Append(a.Version.ToString());
+                }
+            }
+
+            String cultureName = a.CultureName;
+            if (cultureName != null)
+            {
+                if (cultureName == String.Empty)
+                    cultureName = "neutral";
+                sb.Append(", Culture=");
+                sb.AppendQuoted(cultureName);
+            }
+
+            byte[] pkt = a.GetPublicKeyToken();
+            if (pkt != null)
+            {
+                if (pkt.Length > PUBLIC_KEY_TOKEN_LEN)
+                    throw new ArgumentException();
+
+                sb.Append(", PublicKeyToken=");
+                if (pkt.Length == 0)
+                    sb.Append("null");
+                else
+                {
+                    foreach (byte b in pkt)
+                    {
+                        sb.Append(b.ToString("x2", CultureInfo.InvariantCulture));
+                    }
+                }
+            }
+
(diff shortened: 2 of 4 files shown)


src/mscorlib/System.Private.CoreLib.csproj
```diff
@@ -385,6 +383,8 @@
   <ItemGroup>
     <Compile Include="$(BclSourcesRoot)\System\Reflection\Assembly.CoreCLR.cs" />
     <Compile Include="$(BclSourcesRoot)\System\Reflection\AssemblyName.cs" />
+    <Compile Include="$(BclSourcesRoot)\System\Reflection\AssemblyNameHelpers.cs" />
```
weshaggard: We should probably move these under shared so they are correctly mirrored between coreclr and corert.

sepidehMS: @weshaggard These new files live [here](https://github.com/dotnet/corert/tree/893cbdc5cf61089085ebf0d7a7a81fc54495020f/src/System.Private.CoreLib/src/System/Reflection/Runtime/Assemblies) in CoreRT. Should I keep the same structure?

src/mscorlib/src/System/Reflection/AssemblyNameHelpers.cs
```diff
@@ -0,0 +1,168 @@
...
+                    {
+                        sb.Append(b.ToString("x2", CultureInfo.InvariantCulture));
+                    }
+                }
+            }
+
+            if (0 != (a.Flags & AssemblyNameFlags.Retargetable))
+                sb.Append(", Retargetable=Yes");
+
+            AssemblyContentType contentType = a.Flags.ExtractAssemblyContentType();
```
AtsushiKan: Careful now: RuntimeAssemblyName.Flags and AssemblyName.Flags don't return the same value. In particular, AssemblyName.Flags don't include the content type. You have to get that from AssemblyName.ContentType.

src/mscorlib/src/System/Reflection/AssemblyNameHelpers.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            if (a.Version != null)
+            {
+                Version canonicalizedVersion = a.Version.CanonicalizeVersion();
+                if (canonicalizedVersion.Major != ushort.MaxValue)
+                {
+                    sb.Append(", Version=");
+                    sb.Append(canonicalizedVersion.Major);
+                    sb.Append('.');
+                    sb.Append(canonicalizedVersion.Minor);
```
jkotas: The check needs to be for all parts, e.g.

```
   AssemblyName assemblyName = new AssemblyName("MyAssemblyName");
   assemblyName.Version = new Version(42, 42);
   Console.WriteLine(assemblyName.FullName);
```

should print `MyAssemblyName, Version=42.42`

src/mscorlib/src/System/Reflection/AssemblyNameHelpers.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                if (cultureName == String.Empty)
+                    cultureName = "neutral";
+                sb.Append(", Culture=");
+                sb.AppendQuoted(cultureName);
+            }
+
+            byte[] pkt = null;
+            if (0 != (a.Flags & AssemblyNameFlags.PublicKey))
+                pkt = a.GetPublicKeyToken();
```
AtsushiKan: Note that GetPublicKeyToken() can side-effect the AssemblyName (one reason why corert uses RuntimeAssemblyName for everything - it avoids all the eccentricities of AssemblyName.)

I'm not sure if this is critical, but it is an AssemblyName that you don't own.

src/mscorlib/src/System/Reflection/AssemblyNameHelpers.cs
jkotas: Actually, `sb.Append(a.Version.ToString())` may work just fine instead all this.

src/mscorlib/src/System/Reflection/AssemblyNameLexer.cs
```diff
@@ -0,0 +1,137 @@
...
+                c = _chars[_index++];
+            }
+
+            tokenString = sb.ToString();
+            if (quoteChar == 0)
+                tokenString = tokenString.Trim(); // Unless quoted, whitespace at beginning or end doesn't count.
+            return Token.String;
+        }
+
+        internal static KeyValuePair<char, String>[] EscapeSequences =
```
AtsushiKan: Including all of AssemblyNameLexer just to get EscapeSequences seems excesssive (unless you decide to throw the whole set into the shared partition.)

src/mscorlib/System.Private.CoreLib.csproj
weshaggard: They should just get moved under shared so that they mirror between the 2.

danmosemsft: yes it would be shared\system\reflection\...

When the auto-PR reflects this change into CoreRT, the PR build will break there because it will now have this type in two places. You would then go update that PR there to delete the old one to fix the PR so it can be merged.

weshaggard: Or you can move them in CoreRT first and then let them mirror to CoreCLR

danmosemsft: Note the files that are shared in coreclr are listed in \mscorlib\shared\System.Private.CoreLib.Shared.shproj not in thecproj.

jkotas: I would recommend to not worry about the sharing for ZBB. Make this work well for CoreCLR, without sharing. We can figure out the sharing with CoreRT in separate PR - it can wait after ZBB.

src/mscorlib/src/System/Reflection/AssemblyNameHelpers.cs
```diff
@@ -0,0 +1,168 @@
...
+            String cultureName = a.CultureName;
+            if (cultureName != null)
+            {
+                if (cultureName == String.Empty)
+                    cultureName = "neutral";
+                sb.Append(", Culture=");
+                sb.AppendQuoted(cultureName);
+            }
+
+            byte[] pkt = null;
```
AtsushiKan: The way you've changed this, if the AssemblyName doesn't have the full-length public key (which is the typical case), the generated fullname will be missing the PKT altogether.

sepidehMS: @AtsushiKan Isn't that the Desktop behavior as mentioned [here](https://github.com/dotnet/corefx/issues/19179)?

AtsushiKan: No, I'm talking about a case like this:

```
        public static void Main()
        {
            AssemblyName a = new AssemblyName("Foo, Version=1.0.0.0");
            a.SetPublicKeyToken(typeof(object).Assembly.GetName().GetPublicKeyToken());

            string s = a.FullName;
            Console.WriteLine(s);
        }
```

The PublicKey and PublicKeyToken are separate fields in AssemblyName. It's common to have PublicKeyToken set to an actual value while PublicKey is set to null. In such a case, the FullName is supposed to include the PublicKeyToken.

src/mscorlib/src/System/Reflection/AssemblyNameHelpers.cs
AtsushiKan: >Note that GetPublicKeyToken() can side-effect 

After offline conversation, I'll set aside this concern for ZBB. Post-ZBB, we can do a proper sharing of this stuff that should get rid of this good.

sepidehMS:
Addressed feedback. As @jkotas suggested, I will share the files with CoreRT in a separate PR post-ZBB, as that might introduce complications. I moved these files to a separate directory, though. PTAL. Thanks.

danmosemsft:
Please also take a look at our test coverage in corefx.

sepidehMS:

sepidehMS:
@danmosemsft I checked and ran the tests locally. Will modify the tests and re-enable them once this gets merged and is picked up in CoreFX.

danmosemsft:
@sepidehms yeah, I meant more shredding we have sufficient coverage for this new code...

sepidehMS:

sepidehMS:
Ready for review.

src/mscorlib/src/System/Reflection/Runtime/Assemblies/AssemblyNameHelpers.cs
```diff
@@ -0,0 +1,156 @@
...
+**
+==============================================================*/
+
+using System;
+using System.Globalization;
+using System.IO;
+using System.Text;
+using System.Collections.Generic;
+
+namespace System.Reflection.Runtime.Assemblies
```
AtsushiKan: We should move these to the System.Reflection namespace. (Ideally that should have been done last year when these files were moved out of Reflection.Core.)

src/mscorlib/src/System/Reflection/Runtime/Assemblies/AssemblyNameHelpers.cs
```diff
@@ -0,0 +1,156 @@
...
+                sb.AppendQuoted(a.Name);
+            }
+
+            if (a.Version != null)
+            {
+                Version canonicalizedVersion = a.Version.CanonicalizeVersion();
+                if (canonicalizedVersion.Major != ushort.MaxValue)
+                {
+                    sb.Append(", Version=");
+                    sb.Append(a.Version.ToString());
```
AtsushiKan: You need to continue using the canonicalizedVersion to be compatible with the desktop.

The canonicalizedVersion is what truncates the 32-bit Version components into unsigned 16 bit numbers (to match the underlying metadata structures on both N and ECMA and ensure that the AssemblyName is roundtrippable from them.)

A Version component whose lower 16-bits is 0xffff is treated as the "not specified" value and causes it and the numbers that come after it to be dropped from the output (see code sample below) It's probably better not to use Version.ToString() here and implement the logic yourself.

```
        public static void Main()
        {
            AssemblyName a = new AssemblyName("Foo");
            a.Version = new Version(1, 2, 0x1ffff, 4);
            Console.WriteLine("Version.ToString()    = " + a.Version);
            Console.WriteLine("AssemblyName.FullName = " + a.FullName);
        }

```
outputs:
 
```
 Version.ToString()    = 1.2.131071.4
 AssemblyName.FullName = Foo, Version=1.2

```

on the Desktop.

sepidehMS: @AtsushiKan Thanks for pointing that out. Fixed.

PR #11506, opened 2017-05-10
merger takes MC file names as UTF8 names.

sandreenko:
There were problems with:
1. Arabick and Japanese symbols in CLR_SH;
2. path + MC name length were bigger than 260.

We didn't see these problem before, because we moved all MC files and
renamed them with short numbers names.
Now if we want to:
1. avoid  MC files moving;
2. save information that is encoded in this names for spmi diff between
different collections we want this fix.

So this fix work with MC names as they came from "FilterDirectory" that returns "WIN32_FIND_DATAW" that represent .MC names as "_Field_z_ WCHAR  cFileName[ MAX_PATH ];"

### DIFF
--- a/src/ToolBox/superpmi/mcs/verbmerge.cpp
+++ b/src/ToolBox/superpmi/mcs/verbmerge.cpp
@@ -15,16 +15,25 @@
 // The caller must delete the returned string with delete[].
 //
 // static
-char* verbMerge::MergePathStrings(const char* dir, const char* file)
+LPTSTR verbMerge::MergePathStrings(LPCTSTR dir, LPCTSTR file)
 {
-    size_t dirlen  = strlen(dir);
-    size_t filelen = strlen(file);
+    size_t dirlen  = wcslen(dir);
+    size_t filelen = wcslen(file);
     size_t newlen  = dirlen + 1 /* slash */ + filelen + 1 /* null */;
-    char*  newpath = new char[newlen];
-    strcpy(newpath, dir);
-    strcat(newpath, DIRECTORY_SEPARATOR_STR_A);
-    strcat(newpath, file);
+    LPTSTR newpath = new WCHAR[newlen];
+    wcscpy(newpath, dir);
+    wcscat(newpath, DIRECTORY_SEPARATOR_STR_A);
+    wcscat(newpath, file);
     return newpath;
+}
+
+char* verbMerge::ConvertUTF8ToMultiByte(LPCTSTR wstr)
+{
+    unsigned int codePage   = CP_UTF8;
+    int          sizeNeeded = WideCharToMultiByte(codePage, 0, wstr, -1, NULL, 0, NULL, NULL);
+    char*        encodedStr = new char[sizeNeeded];
+    WideCharToMultiByte(codePage, 0, wstr, -1, encodedStr, sizeNeeded, NULL, NULL);
+    return encodedStr;
 }
 
 // AppendFile: append the file named by 'fileName' to the output file referred to by 'hFileOut'. The 'hFileOut'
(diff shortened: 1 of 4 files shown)


sandreenko:
The thing is that  when Cmake file was created, 
```
-remove_definitions(-DUNICODE)  
-remove_definitions(-D_UNICODE) 
```
were there for CoreCLR, but the saved project files, that are under source control in CodeGenMirroro, repo were not regenerated with this new cmake options.

src/ToolBox/superpmi/mcs/verbmerge.h
```diff
@@ -15,21 +16,23 @@ class verbMerge
...
-    static bool DirectoryFilterDirectories(WIN32_FIND_DATAA* findData);
-    static bool DirectoryFilterFile(WIN32_FIND_DATAA* findData);
-    static int __cdecl WIN32_FIND_DATAA_qsort_helper(const void* p1, const void* p2);
-    static int FilterDirectory(const char*                  searchPattern,
+    typedef bool (*DirectoryFilterFunction_t)(WIN32_FIND_DATAW*);
+    static bool DirectoryFilterDirectories(WIN32_FIND_DATAW* findData);
+    static bool DirectoryFilterFile(WIN32_FIND_DATAW* findData);
+    static int __cdecl WIN32_FIND_DATAW_qsort_helper(const void* p1, const void* p2);
+    static int FilterDirectory(LPCTSTR                      searchPattern,
                                DirectoryFilterFunction_t    filter,
```
BruceForstall: Here, and elsewhere, you should use LPCWSTR, LPWSTR, and WCHAR, not the "T" variants. (The T variants switch between CHAR and WCHAR depending on the UNICODE define. We don't want these to shift).

src/ToolBox/superpmi/mcs/verbmerge.cpp
```diff
@@ -15,42 +15,52 @@
...
 
+char* verbMerge::ConvertUTF8ToMultiByte(LPCTSTR wstr)
+{
+    unsigned int codePage   = CP_UTF8;
+    int          sizeNeeded = WideCharToMultiByte(codePage, 0, wstr, -1, NULL, 0, NULL, NULL);
+    char*        encodedStr = new char[sizeNeeded];
+    WideCharToMultiByte(codePage, 0, wstr, -1, encodedStr, sizeNeeded, NULL, NULL);
+    return encodedStr;
+}
+
```
BruceForstall: Looks like this is only used for Log output? In that case, can't you just use `%S` (capital 'S') instead of `%s` (lower-case 's') in the format specification, instead?

src/ToolBox/superpmi/mcs/verbmerge.cpp
```diff
@@ -285,25 +297,51 @@ int verbMerge::AppendAllInDir(HANDLE              hFileOut,
...
+            delete[] fileFullPath;
+
+            fileFullPath = newBuffer;
+        }
 
         // Is it zero length? If so, skip it.
         if ((findData.nFileSizeLow == 0) && (findData.nFileSizeHigh == 0))
         {
+            char* fileFullPathAsChar = ConvertUTF8ToMultiByte(fileFullPath);
+
```
BruceForstall: Unused.

src/ToolBox/superpmi/mcs/verbmerge.cpp
```diff
@@ -384,43 +422,51 @@ int verbMerge::DoWork(const char* nameOfOutputFile, const char* pattern, bool re
...
         {
             file = dir;
-            dir  = _strdup(".");
+            dir  = _wcsdup(L".");
         }
     }
     else
     {
-        const char* wildcard = strchr(lastSlash, '*');
+        LPCTSTR wildcard = wcschr(lastSlash, '*');
```
BruceForstall: shouldn't the `'*'` also have a prefix L? Same below, for the `'\0'` cases.

BruceForstall:
You can't use `L"..."`; you need to use `W("...")`. See palclr.h. L"" means something different on Linux than Windows.

sandreenko:
> You can't use L"..."; you need to use W("..."). See palclr.h. L"" means something different on Linux than Windows.

I noticed that, but I found some places where we use L"" (dasm, binder), why doesn't it crash anything?

src/ToolBox/superpmi/mcs/verbmerge.cpp
sandreenko: No, the single symbols doesn't  need implicit cast. So all 's' do not need W().

src/ToolBox/superpmi/mcs/verbmerge.cpp
sandreenko: Fixed

src/ToolBox/superpmi/mcs/verbmerge.h
sandreenko: Fixed

src/ToolBox/superpmi/mcs/verbmerge.cpp
sandreenko: I tried it and it didn't work.  _vscprintf returns 0 length for the Wchar* string.

src/ToolBox/superpmi/superpmi-shared/standardpch.h
```diff
@@ -88,6 +88,9 @@
 #ifndef DIRECTORY_SEPARATOR_STR_A
 #define DIRECTORY_SEPARATOR_STR_A "\\"
 #endif
+#ifndef DIRECTORY_SEPARATOR_STR_W
+#define DIRECTORY_SEPARATOR_STR_W W("\\")
+#endif
```
BruceForstall: Does this need to be below the definition of W, below?

src/ToolBox/superpmi/mcs/verbmerge.cpp
```diff
@@ -15,42 +15,52 @@
...
-    strcat(newpath, file);
+    LPWSTR newpath = new WCHAR[newlen];
+    wcscpy(newpath, dir);
+    wcscat(newpath, DIRECTORY_SEPARATOR_STR_W);
+    wcscat(newpath, file);
     return newpath;
 }
 
+char* verbMerge::ConvertUTF8ToMultiByte(LPCWSTR wstr)
+{
```
BruceForstall: I don't understand this. UTF8 strings should be stored in `char*`, whereas LPCWSTR should store a string of 2-byte Unicode characters (or maybe UTF-16).

src/ToolBox/superpmi/mcs/verbmerge.cpp
```diff
@@ -285,25 +297,50 @@ int verbMerge::AppendAllInDir(HANDLE              hFileOut,
...
     }
 
     for (int i = 0; i < elemCount; i++)
     {
-        const WIN32_FIND_DATAA& findData     = fileArray[i];
-        char*                   fileFullPath = MergePathStrings(dir, findData.cFileName);
+        const _WIN32_FIND_DATAW& findData     = fileArray[i];
+        LPWSTR                   fileFullPath = MergePathStrings(dir, findData.cFileName);
+
+        if (wcslen(fileFullPath) > 260) // It is too long path, use \\?\ to access it.
```
BruceForstall: Instead of 260, use MAX_PATH?

sandreenko:
@BruceForstall I got rid of ci errors and it works fine with my test case, please take another look.

PR #11507, opened 2017-05-10
EventPipe Circular Buffer Support and Ability to Start/Stop Tracing

brianrob:
This PR contains the following changes:

- Support for logging to a circular buffer owned by the EventPipe instead of using synchronous file writes.
- Very basic private APIs to support start/stop of the EventPipe from managed code.

The start/stop APIs will be expanded to allow for specification of events to enable, file path, circular buffer size, etc.


brianrob:
FYI @valenis

brianrob:
@vancem, this is ready for review.

src/mscorlib/src/System/Diagnostics/Eventing/EventPipe.cs
```diff
@@ -0,0 +1,155 @@
...
+            m_providerName = providerName;
+            m_keywords = keywords;
+        }
+
+        internal string ProviderName
+        {
+            get { return m_providerName; }
+        }
+
+        internal UInt64 Keywords
```
vancem: I probably misunderstood your question when you ask me how global levels where, but in ETW each provider can have a level specified independently (like keywords).   Thus we should add Level here and remove it in EventPipeConfiguration.

src/mscorlib/src/System/Diagnostics/Eventing/EventPipe.cs
```diff
@@ -0,0 +1,155 @@
...
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Security;
+using Microsoft.Win32;
+
+namespace System.Diagnostics.Tracing
+{
+    [StructLayout(LayoutKind.Sequential)]
+    internal struct EventPipeProviderConfiguration
```
vancem: Probably better to make it a class rather than a struct, as we will add level and later probably some key value strings to this.    In general

src/mscorlib/src/System/Diagnostics/Eventing/EventPipe.cs
brianrob: Thanks @vancem.  I've fixed this and pushed the update.

src/mscorlib/src/System/Diagnostics/Eventing/EventPipe.cs
brianrob: I tried making this a class, but interop doesn't seem to know how to marshal it.  That's why I originally made it a struct.  If there is a way to do this, let me know, and I can make the change.

src/mscorlib/src/System/Diagnostics/Eventing/EventPipe.cs
```diff
@@ -0,0 +1,155 @@
...
+            }
+            // TODO: Validate level.
+            // TODO: Consider using an enum.
+            m_outputFile = outputFile;
+            m_circularBufferSizeInMB = circularBufferSizeInMB;
+            m_level = level;
+            m_providers = new List<EventPipeProviderConfiguration>();
+        }
+
+        internal string OutputFile
```
vancem: It would be useful/helpful if you mark the methods that are ultimately intended to be public 'public' rather than internal.   By marking the class itself 'internal' you make the methods inside internal (so there is no difference today) but it makes it easy to simply expose the API by flipping the declaration on the class.   It also lets a reader understand what is implementation and what is intended to be external interface.

src/mscorlib/src/System/Diagnostics/Eventing/EventPipe.cs
```diff
@@ -0,0 +1,155 @@
...
+        [DllImport(JitHelpers.QCall, CharSet = CharSet.Unicode)]
+        [SuppressUnmanagedCodeSecurity]
+        internal static extern void Disable();
+
+        //
+        // These PInvokes are used by EventSource to interact with the EventPipe.
+        //
+        [DllImport(JitHelpers.QCall, CharSet = CharSet.Unicode)]
+        [SuppressUnmanagedCodeSecurity]
+        internal static extern IntPtr CreateProvider(Guid providerID, UnsafeNativeMethods.ManifestEtw.EtwEnableCallback callbackFunc);
```
vancem: You know I am on this kick to get rid of Guids and instead use strings as the way of identifying providers (EventSources).    It seems to me that this one is relatively low hanging fruit.

src/mscorlib/src/System/Diagnostics/Eventing/EventPipe.cs
```diff
@@ -0,0 +1,155 @@
...
+            m_providers.Add(new EventPipeProviderConfiguration(
+                providerName,
+                keywords));
+        }
+    }
+
+    internal static class EventPipe
+    {
+        internal static void Enable(EventPipeConfiguration configuration)
+        {
```
vancem: I notice that EventPipe is a singleton (it has no 'handle' that represents opened pipe).   I think that this will become a needless stumbling block if we ever need more than client.   Lifetime issues also tend to be handled sloppily when you have static singletons (the obligation to close the instance on end of use is not as clear).  Thus I believe that EventPipe be a true instance class, and that the Pinvokes take a handle.  

Note that it is OK in the short term to fail if you try to enable more than one EventPipe simultaneously (although my guess is that you won't need to do that, supporting multiple instances will all naturally 'fall out'.

vancem:
I know that you want to merge this today.   In my feedback I am suggesting changes that might take more than a few hours.  I am OK with talking about these and addressing them after checkin.

src/mscorlib/src/System/Diagnostics/Eventing/EventPipe.cs
```diff
@@ -0,0 +1,157 @@
...
+                providerName,
+                keywords,
+                loggingLevel));
+        }
+    }
+
+    internal static class EventPipe
+    {
+        internal static void Enable(EventPipeConfiguration configuration)
+        {
```
vancem: I am also conflicted about the model that is presented on enabling.  The way you have it currently you have to supply all the configuration information in one call.   This configuration is moderately complex (a list of structs that have strings inside them).   This in turn causes you to defined 'config' classes for this information and use non-trivial interop features (auto-marshaling of strings in structs).  

The alternative is in fact what ETW does.   Basically you supply configuration data as a series of calls (Enabling or disabling one provider at a time).  This is actually more flexible, and useful, and happens to also avoid the need for a 'config' structure (all the fields becomes arguments to the 'EnableProvider' call), and the non-trivial interop features (since now you call functions that take primitive or string values).   

As I said I am conflicted, but my instincts say that there was no reason to deviate from what ETW/EventListener did (having one model throughout is good).  

The more I think about it, the more I would prefer to following the ETW/EventListener pattern (you have a 'EnableProvider' calls you make).  The reason is that we will want to have EventListener be able to control EventPipes, and we don't want any unnecessary model mismatches.

src/mscorlib/src/System/Diagnostics/Eventing/EventPipe.cs
```diff
@@ -0,0 +1,157 @@
...
+        [SuppressUnmanagedCodeSecurity]
+        internal static extern IntPtr AddEvent(IntPtr provHandle, Int64 keywords, uint eventID, uint eventVersion, uint level, bool needStack);
+
+        [DllImport(JitHelpers.QCall, CharSet = CharSet.Unicode)]
+        [SuppressUnmanagedCodeSecurity]
+        internal static extern void DeleteProvider(IntPtr provHandle);
+
+        [DllImport(JitHelpers.QCall, CharSet = CharSet.Unicode)]
+        [SuppressUnmanagedCodeSecurity]
+        internal static extern unsafe void WriteEvent(IntPtr eventHandle, void* data, uint length);
```
vancem: One of the major cleanup items is to fix an issue with custom filtering and ETW (bug 105200).  TO fix this we need to be able NOT write to every session with a particular WriteEvent call.  We will be doing this by using the Filter parameter of the [EventWriteEx](https://msdn.microsoft.com/en-us/library/windows/desktop/dd392307(v=vs.85).aspx) method, and we will need to plumb that through here as well.      I debated asking you to do this plumbing now, but It is probably better to wait.   

I do want you to be aware that ultimately however, we need a parameter passed to WriteEvent that represents the set of sessions to send the event to (which we don't today, and we default to 'all.

vancem:
I quickly skimmed the EventPipeBuffer and EventPipeBufferManager.   I think a quick review is OK because ultimately it simply implements a rather boring 'WriteEvent' method (but scalable).    I am also happy with the general architecture (a per-thread linked list of buffers).   We have already gone over the threading architecture and I think it is OK.   Any bugs will not perturb the larger design.

src/mscorlib/src/System/Diagnostics/Eventing/EventPipe.cs
brianrob: Yup.  Tracked by https://github.com/dotnet/coreclr/issues/11551.

src/mscorlib/src/System/Diagnostics/Eventing/EventPipe.cs
brianrob: Filed as #11565.

src/mscorlib/src/System/Diagnostics/Eventing/EventPipe.cs
brianrob: Filed as #11566.

brianrob:
@vancem, thank you very much for the feedback!

PR #11207, opened 2017-04-25
Preserve value numbers in gtSetEvalOrder

JosephTremoulet:
There was already code to preserve the value number during global CSE, but
value numbers remain relevant through assertion prop, so this changes the
code to simply always pass the PRESERVE_VN flag.

PR #10453, opened 2017-03-24
Promote implicit-byref struct arguments

JosephTremoulet:
This change extends struct promotion to "implicit byref" parameters (i.e. struct which are passed by value at source level but for which the ABI dictates that the caller make a copy in its frame and pass the callee a pointer to the copy).  Struct promotion is how we do field-sensitive analysis, so this enables propagating facts about `Span<T>` parameters' lengths to eliminate redundant checks, and similar.

Struct promotion annotations exist on local var descriptors, which poses an issue here because the incoming struct memory is not a local of the callee.  To address this, promotion of implicit byref parameters is implemented by:
  1. Creating a local struct to represent the parameter
  2. Initializing the new local from the incoming value via a struct copy at method entry
  3. Rewriting all references to the parameter throughout the method body to instead refer to the new local

Aside from field-sensitive propagation of values etc., a beneficial effect on the generated code (that falls out of the above approach) is that the references throughout the method body are references to locals instead of pointer dereferences, so this change results in elimination of many more redundant struct field loads.

The struct copy to initialize the new local of course incurs a cost that must be weighed against the benefits of exposing the individual fields and improved redundant load elimination.  This change uses a simple heuristic that checks whether a parameter is address-exposed, and also compares the number of references to the parameter in the method body against the number of fields; it effectively avoids doing the promotion for parameters that are address-exposed or are not heavily referenced.

Avoiding the struct copy by adding the capacity to reason about promoted structs that don't reside in the current method's frame (and recognize these "restrict" pointer parameters' referents as not killed by arbitrary pointer dereferences, etc.) could alleviate the costs, as could optimizations like PDE or rematerialization to move the copies to more optimal code paths; these are left as possible future improvements.

This necessitated a change in phase ordering;  the implicit byref rewriting now happens in a few stages:
 - fgMarkImplicitByRefArgs now does nothing more than flag which args are
   implicit byrefs, and runs before struct promotion.
 - fgRetypeImplicitByRefArgs actually retypes the parameters (from struct
   to pointer), and decides which promotions are beneficial.  It inserts
   the initializing struct copies for beneficial ones, and annotates
   non-beneficial ones so their appearances will be rewritten via the
   pointer parameter.  It runs after struct promotion and address-escape
   analysis.
 - fgMorphImplicitByRefArgs is moved out of fgMarkAddressExposedLocals,
   and into fgMorphBlocks.  It rewrites any implicit byref parameters
   which have not been promoted to have the extra indirection at all their
   references.
[...]

PR #11521, opened 2017-05-11
Propagate assertions for more checked bounds

JosephTremoulet:
Generalize (and rename) the assertion/limit code for tracking array
lengths (which have until now been recognized by the VNFunc being
GT_ARR_LENGTH) to track all values that appear in the function as
length arguments to bounds chcks nodes.  Add a hashtable to identify that
set of value numbers, `m_checkedBoundVNs` to the VNStore object, and
populate it during value-numbering.  Wherever names were adjusted, the
generalization of "array length" is called "checked bound".

This allows the array bound check removal machinery to operate on span
bound checks as well.

### DIFF
--- a/src/jit/assertionprop.cpp
+++ b/src/jit/assertionprop.cpp
@@ -627,12 +627,12 @@ void Compiler::optPrintAssertion(AssertionDsc* curAssertion, AssertionIndex asse
         vnStore->vnDump(this, curAssertion->op1.bnd.vnLen);
         printf("]");
     }
-    else if (curAssertion->op1.kind == O1K_ARRLEN_OPER_BND)
+    else if (curAssertion->op1.kind == O1K_BOUND_OPER_BND)
     {
         printf("Oper_Bnd");
         vnStore->vnDump(this, curAssertion->op1.vn);
     }
-    else if (curAssertion->op1.kind == O1K_ARRLEN_LOOP_BND)
+    else if (curAssertion->op1.kind == O1K_BOUND_LOOP_BND)
     {
         printf("Loop_Bnd");
         vnStore->vnDump(this, curAssertion->op1.vn);
@@ -711,12 +711,12 @@ void Compiler::optPrintAssertion(AssertionDsc* curAssertion, AssertionIndex asse
                     printf("MT(%08X)", dspPtr(curAssertion->op2.u1.iconVal));
                     assert(curAssertion->op2.u1.iconFlags != 0);
                 }
-                else if (curAssertion->op1.kind == O1K_ARRLEN_OPER_BND)
+                else if (curAssertion->op1.kind == O1K_BOUND_OPER_BND)
                 {
                     assert(!optLocalAssertionProp);
                     vnStore->vnDump(this, curAssertion->op2.vn);
                 }
-                else if (curAssertion->op1.kind == O1K_ARRLEN_LOOP_BND)
+                else if (curAssertion->op1.kind == O1K_BOUND_LOOP_BND)
                 {
                     assert(!optLocalAssertionProp);
                     vnStore->vnDump(this, curAssertion->op2.vn);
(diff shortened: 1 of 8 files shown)


JosephTremoulet:
@briansull, @AndyAyersMS PTAL
/cc @dotnet/jit-contrib

This addresses #10050 with the exception that the value-number update that happens for compare nodes when one of their arguments is an array length that gets CSE'd does not apply in as many cases to spans, because we don't have the CSE candidate occurrence lists collected yet when we build the `optCseCheckedBoundMap`; that is left as a future possible improvement.

The `Performance/CodeQuality/Span/Indexer` test consists of several indexing patterns which we can use to compare the power of span bound check removal against array bound check removal.  The effect of this change (on top of #11207) is:

|      Sub-Test    |      Effect of this change     |     Effect of this change plus #10453    |    Effect of this change plus fix for `optCseCheckedBoundMap` issue  |
|-----------------|---------------------------------|-----------------------------------------|----------------------------------------------------------------------------------|
|     Indexer1    |     bound check removed; array parity   |     |     |
|     Indexer2   |      no change (CSE doesn't update VN)   |  bound check removed; array parity   |  bound check removed; array parity    |
|     Indexer3    |     bound checked removed; array parity   |     |     |
|     Indexer4    |     bound checked removed; array parity   |     |     |
|     Indexer5    |     no change (afraid write could clobber span length)   |  bound checked removed; array parity   |     |
|     Indexer6    |     no change (afraid write could clobber span length)   |  bound checked removed; array parity   |     |
|     ReadOnlyIndexer1    |     bound check removed; array parity   |     |     |
|     ReadOnlyIndexer2   |      no change (CSE doesn't update VN)   |  bound check removed; array parity   |  bound check removed; array parity    |
|     WriteViaIndexer1    |     no change (afraid write could clobber span length)   |  bound checks on data[idx] removed, not on data[0]   |     |
|     WriteViaIndexer2    |     no change, but already have array parity   |     |     |
|     KnownSizeArray    |     no change, but already have array parity   |     |     |
|     KnownSizeCtor    |     no change, but already have array parity   |     |     |
|     KnownSizeCtor2    |     no change, but already have array parity   |     |     |
|     SameIndex1    |     no change, but already have array parity   |     |     |
|     SameIndex2    |     no change (afraid write could clobber span length)   |  bound checked removed; array parity   |     |
|     CoveredIndex1    |     no change, but already have array parity   |     |     |
|     CoveredIndex2    |     no change, but already have array parity   |     |     |
|     CoveredIndex3    |     no change, but already have array parity   |     |     |      

No other diffs from this change (spans aren't used much in our test corpus)

src/jit/assertionprop.cpp
```diff
@@ -1788,16 +1788,16 @@ AssertionInfo Compiler::optCreateJTrueBoundsAssertion(GenTreePtr tree)
...
         return index;
     }
-    // Cases where op1 holds the condition array length and op2 is 0.
-    // Loop condition like: "i < a.len == 0"
-    // Assertion: "i < a.len == false"
-    else if (vnStore->IsVNArrLenBound(vn) &&
+    // Cases where op1 holds the condition bound and op2 is 0.
+    // Loop condition like: "i < len == 0"
+    // Assertion: "i < len == false"
+    else if (vnStore->IsVNCompareCheckedBound(vn) &&
```
briansull: The comment block reads better like this:


// Cases where op1 holds the upper bound and op2 is 0.
// Loop condition like: "i < bnd == 0"
// Assertion: "i < bnd == false"

AndyAyersMS: If you change this also change the previous comment block to match.

AndyAyersMS: And a few more below...

src/jit/valuenum.cpp
```diff
@@ -7028,6 +7062,14 @@ void Compiler::fgValueNumberTree(GenTreePtr tree, bool evalAsgLhsInd)
                 excSet = vnStore->VNPExcSetUnion(excSet, vnStore->VNPExcVal(tree->AsBoundsChk()->gtArrLen->gtVNPair));
 
                 tree->gtVNPair = vnStore->VNPWithExc(vnStore->VNPForVoid(), excSet);
+
+                // Record non-constant value numbers that are used a the length argument to bounds checks, so that
```
mikedn: are used a**s** the length

JosephTremoulet:
Updated per feedback (and to fix build issue)

JosephTremoulet:
Perf lab run came back, the indexer tests with diffs do indeed show improvement:

|     Test         |      Improvement      |
|--------------|-----------------------|
| Indexer1      |             19%             |
|  Indexer3    |              18%             |
|  Indexer4     |              7%             |
|  ReadOnlyIndexer1  |      19%        |

JosephTremoulet:
@dotnet-bot test this please

JosephTremoulet:

PR #11526, opened 2017-05-11
[RyuJIT/ARM32] Enable GT_MUL_LONG codegen

mskvortsov:
Enables codegen for the int32×int32→int64 multiplication, e.g. for the following method
```csharp
public static long LongMul(int x, int y) { return (long)x * y; }
```
the JIT now generates
```
000008  FB80 1001      smull   r1, r0, r0, r1
00000C  9100           str     r1, [sp]
00000E  9001           str     r0, [sp+0x04]
000010  9800           ldr     r0, [sp]
000012  9901           ldr     r1, [sp+0x04]
```

cc @dotnet/arm32-contrib @dotnet/jit-contrib

### DIFF
--- a/src/jit/codegenarmarch.cpp
+++ b/src/jit/codegenarmarch.cpp
@@ -189,6 +189,12 @@ void CodeGen::genCodeForTreeNode(GenTreePtr treeNode)
             genCodeForIndir(treeNode->AsIndir());
             break;
 
+#ifdef _TARGET_ARM_
+        case GT_MUL_LONG:
+            genCodeForMulLong(treeNode->AsMulLong());
+            break;
+#endif // _TARGET_ARM_
+
 #ifdef _TARGET_ARM64_
 
         case GT_MULHI:
--- a/src/jit/decomposelongs.cpp
+++ b/src/jit/decomposelongs.cpp
@@ -1554,7 +1554,8 @@ GenTree* DecomposeLongs::DecomposeRotate(LIR::Use& use)
 // these nodes, we convert them into GT_MUL_LONGs, undo the cast from int to long by
 // stripping out the lo ops, and force them into the form var = mul, as we do for
 // GT_CALLs. In codegen, we then produce a mul instruction that produces the result
-// in edx:eax, and store those registers on the stack in genStoreLongLclVar.
+// in edx:eax on x86 or in any two chosen by RA registers on arm32, and store those
+// registers on the stack in genStoreLongLclVar.
 //
 // All other GT_MULs have been converted to helper calls in morph.cpp
 //
@@ -1587,7 +1588,7 @@ GenTree* DecomposeLongs::DecomposeMul(LIR::Use& use)
 
     tree->gtOp.gtOp1 = op1->gtGetOp1();
     tree->gtOp.gtOp2 = op2->gtGetOp1();
-    tree->SetOperRaw(GT_MUL_LONG);
+    tree->SetOper(GT_MUL_LONG);
 
     return StoreNodeToVar(use);
 }
--- a/src/jit/gtlist.h
+++ b/src/jit/gtlist.h
@@ -175,7 +175,7 @@ GTNODE(LEA              , "lea"          ,GenTreeAddrMode    ,0,GTK_BINOP|GTK_EX
 // nodes such as calls, returns and stores of long lclVars.
 GTNODE(LONG             , "gt_long"      ,GenTreeOp          ,0,GTK_BINOP)
 
-// The following are nodes representing x86 specific long operators, including
+// The following are nodes representing x86/arm32 specific long operators, including
 // high operators of a 64-bit operations that requires a carry/borrow, which are
 // named GT_XXX_HI for consistency, low operators of 64-bit operations that need
 // to not be modified in phases post-decompose, and operators that return 64-bit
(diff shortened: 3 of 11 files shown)


src/jit/decomposelongs.cpp
```diff
@@ -1581,9 +1581,17 @@ GenTree* DecomposeLongs::DecomposeMul(LIR::Use& use)
     Range().Remove(op1);
     Range().Remove(op2);
 
+#if defined(_TARGET_X86_)
     tree->gtOp.gtOp1 = op1->gtGetOp1();
```
BruceForstall: Should update the header comment, since it says"edx:eax" -- say something about arm?

src/jit/decomposelongs.cpp
```diff
@@ -1581,9 +1581,17 @@ GenTree* DecomposeLongs::DecomposeMul(LIR::Use& use)
...
     tree->gtOp.gtOp2 = op2->gtGetOp1();
     tree->SetOperRaw(GT_MUL_LONG);
+#elif defined(_TARGET_ARM_)
+    GenTree* mulTree = new (m_compiler, GT_MUL_LONG) GenTreeMulLong(TYP_LONG, op1->gtGetOp1(), op2->gtGetOp1());
+    mulTree->gtFlags = tree->gtFlags;
+    Range().InsertBefore(tree, mulTree);
+    use.ReplaceWith(m_compiler, mulTree);
+    Range().Remove(tree);
+#endif
```
BruceForstall: I think you can avoid this `#ifdef` entirely by changing `SetOperRaw` to `SetOper`. I'm not sure x86 depends on the `Raw` version, or that just happens to work.

src/jit/codegenarmarch.cpp
```diff
@@ -189,6 +189,21 @@ void CodeGen::genCodeForTreeNode(GenTreePtr treeNode)
...
+            GenTreeMulLong* dst  = treeNode->AsMulLong();
+            GenTree*        src1 = dst->gtOp1;
+            GenTree*        src2 = dst->gtOp2;
+            instruction     ins  = dst->IsUnsigned() ? INS_umull : INS_smull;
+            getEmitter()->emitIns_R_R_R_R(ins, EA_4BYTE, dst->gtRegNum, dst->gtOtherReg, src1->gtRegNum,
+                                          src2->gtRegNum);
+            genProduceReg(treeNode);
+        }
+        break;
+#endif // _TARGET_ARM_
```
BruceForstall: Please create a `genCodeForMulLong` function to do this.

src/jit/gtlist.h
```diff
@@ -186,8 +186,8 @@ GTNODE(SUB_LO           , "-Lo"          ,GenTreeOp          ,0,GTK_BINOP)
 GTNODE(SUB_HI           , "-Hi"          ,GenTreeOp          ,0,GTK_BINOP)
 GTNODE(DIV_HI           , "/Hi"          ,GenTreeOp          ,0,GTK_BINOP)
 GTNODE(MOD_HI           , "%Hi"          ,GenTreeOp          ,0,GTK_BINOP)
-GTNODE(MUL_LONG         , "*long"        ,GenTreeOp          ,1,GTK_BINOP) // A mul that returns the 2N bit result of an NxN multiply. This op
-                                                                           // is used for x86 multiplies that take two ints and return a long
+GTNODE(MUL_LONG         , "*long"        ,GenTreeMulLong     ,1,GTK_BINOP) // A mul that returns the 2N bit result of an NxN multiply. This op
+                                                                           // is used for multiplies that take two ints and return a long
```
BruceForstall: x86 doesn't need this new struct. I think you should put all the GenTreeMulLong stuff (here, gtstructs.h, gentree.h, etc.) under `#ifdef _TARGET_ARM_`.

BruceForstall:
@CarolEidt this change adds GT_MUL_LONG as a special case node that requires two dst registers. The changes to LSRA look simple; are they sufficient? Do you agree with this approach?

src/jit/lsraarm.cpp
```diff
@@ -741,7 +746,7 @@ void Lowering::TreeNodeInfoInit(GenTree* tree)
     } // end switch (tree->OperGet())
 
     // We need to be sure that we've set info->srcCount and info->dstCount appropriately
-    assert((info->dstCount < 2) || tree->IsMultiRegCall());
+    assert((info->dstCount < 2) || tree->IsMultiRegCall() || (tree->OperGet() == GT_MUL_LONG));
```
CarolEidt: I suspect we may add other nodes in future that return multiple registers. In any case, for clarity I think it would be useful to have a method on `GenTree()` that indicates whether it produces multiple registers, e.g. `IsMultiRegNode()`.

CarolEidt:
> The changes to LSRA look simple; are they sufficient? Do you agree with this approach?

Yes, I think the changes are sufficient. I would like to see an abstraction (as mentioned in a comment) on `GenTree` for nodes that produce multiple registers, but I think it's the case that the multi-reg call work pretty much added the necessary support to LSRA.

---
In reply to: [300890347](https://github.com/dotnet/coreclr/pull/11526#issuecomment-300890347) [](ancestors = 300890347)

CarolEidt:
:clock1:

BruceForstall:
@mmitche I see:
```
Disk has run out of space and cannot get latest changes. Please contact netciadmin@microsoft.com. IOException: No space left on device
```
https://ci.dot.net/job/dotnet_coreclr/job/master/job/checked_ubuntu_flow_prtest/6772/

BruceForstall:

BruceForstall:
@mskvortsov Is this still [WIP], or do you want me to merge it if the tests all pass?

mskvortsov:
@BruceForstall Yes, let it be [WIP] for now, because I want to at least squash the commits before merging, but I'm already OOTO today. I'll finish this PR on Monday.

mskvortsov:
@BruceForstall I've addressed both your suggestions about SetOperRaw() and IsMultiRegNode(), rebased onto master to test the x86 change, and squashed. Please merge this.

BruceForstall:
@dotnet-bot test this please

PR #11527, opened 2017-05-11
Handle ADC in x86 unwindLazyState

adiaaida:
When we add PGO instrumentation to coreclr, the instrumentation injects
PUSH, MOV, ADD, ADC, POP into the prolog to mark the function entry. This
causes a crash because we do not currently handle ADC instructions in
unwindLazyState. This change adds support for ADC, which uses the normal
DecodeRM code we already have in place.


adiaaida:
@gkhanna79 @jkotas PTAL

Issue #11508, opened 2017-05-10
Silent bad codegen when passing and comparing ushort args


erozenfeld:
Repro:

Compile Program.cs below:

.\Tools\dotnetcli\dotnet.exe Tools\csc.exe /o+ /out:Program.exe Program.cs /r:.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Private.CoreLib.dll /r:.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Console.dll /r:.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Runtime.dll

Program.cs:
```
using System;
using System.Collections.Generic;
using System.Runtime.CompilerServices;

namespace TestApp
{
    public struct StructWithValue : IEquatable<StructWithValue>
    {
        private ushort value;

        public StructWithValue(ushort v)
        {
            value = v;
        }


        public bool Equals(StructWithValue other)
        {
            if (value.Equals (other.value))
                return true;
            else
                return false;
        }
    }
    class Program
    {
        [MethodImpl(MethodImplOptions.NoOptimization)]
        static void Main(string[] args)
        {
            var comparer = EqualityComparer<StructWithValue>.Default;
            
            for (ushort i = short.MaxValue - 1; i <= 32770; i++)
            {
                Console.WriteLine ("i = {0}", i);
                var a = new StructWithValue(i);
                var b = new StructWithValue(i);

                if (comparer.Equals(a, b))
                    continue;

                // The value of 32768 will show this message and end the loop
                // Every value > 32767 is affected...

                Console.WriteLine("Problem starts at {0}", i);

                break;

            }

        }

    }

}

```
.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\CoreRun.exe Program.exe
outputs
i = 32766
i = 32767
i = 32768
Problem starts at 32768

erozenfeld:
Codegen for passing the args to `comparer.Equals(a, b)`:
```
       488B4DE8             mov      rcx, gword ptr [rbp-18H]
       480FBF55D8           movsx    rdx, word  ptr [rbp-28H]
       4C0FBF45D0           movsx    r8, word  ptr [rbp-30H]
       488B45E8             mov      rax, gword ptr [rbp-18H]
       488B00               mov      rax, qword ptr [rax]
       488B4048             mov      rax, qword ptr [rax+72]
[...]

PR #11531, opened 2017-05-11
Re-morph lclVar nodes after simplifying (ind (addr (lclVar))).

pgavlin:
Morph simplifies (ind (addr (lclVar))) if the type of the load matches
the type of the lclVar node. This simplification is not valid, however,
if the lclVar must be normalized upon load: in that case, the lclVar
node must be normalized appropriately as part of the transformation.
This change fixes the simplification to perform this normalization if
necessary by calling `fgMorphLclVar` on the result if the result is a
lclVar node.

Fixes issue #11508.

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -13206,7 +13206,7 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
                         temp->gtDebugFlags &= ~GTF_DEBUG_NODE_MORPHED;
 #endif // DEBUG
                         const bool forceRemorph = true;
-                        temp = fgMorphLocalVar(temp, forceRemorph);
+                        temp                    = fgMorphLocalVar(temp, forceRemorph);
 #ifdef DEBUG
                         // We then set this flag on `temp` because `fgMorhpLocalVar` may not set it itself, and the
                         // caller of `fgMorphSmpOp` may assert that this flag is set on `temp` once this function
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -6099,7 +6099,7 @@ GenTreePtr Compiler::fgMorphStackArgForVarArgs(unsigned lclNum, var_types varTyp
  *  Transform the given GT_LCL_VAR tree for code generation.
  */
 
-GenTreePtr Compiler::fgMorphLocalVar(GenTreePtr tree)
+GenTreePtr Compiler::fgMorphLocalVar(GenTreePtr tree, bool force)
 {
     noway_assert(tree->gtOper == GT_LCL_VAR);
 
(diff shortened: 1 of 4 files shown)


pgavlin:
@briansull @CarolEidt @erozenfeld @dotnet/jit-contrib PTAL

src/jit/morph.cpp
```diff
@@ -13195,6 +13195,19 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
...
+                    // normalization.
+                    if (temp->OperIs(GT_LCL_VAR))
+                    {
+#ifdef DEBUG
+                        temp->gtDebugFlags &= ~GTF_DEBUG_NODE_MORPHED;
+#endif // DEBUG
+                        temp = fgMorphLocalVar(temp);
+#ifdef DEBUG
+                        temp->gtDebugFlags |= GTF_DEBUG_NODE_MORPHED;
+#endif // DEBUG
```
CarolEidt: I think this flag probably causes more problems than it solves, but given that we're not eliminating it, an explanatory comment would be good here.

src/jit/morph.cpp
```diff
@@ -13195,6 +13195,19 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
...
+                    }
+
                     return temp;
```
erozenfeld: The optimization may happen during global morph or after global morph. fgMorphLocalVar will not normalize the local unless called during global morph.

pgavlin: Good catch. Fixed.

pgavlin: ...on second thought, I can probably just remove the global morph early-out.

pgavlin: though that may be trickier than it seems, as I think that check is attempting to avoid reinserting the cast during remorphing. I will leave things as-is for now.

src/jit/morph.cpp
```diff
@@ -13195,6 +13195,23 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
...
+                    // If the result of the fold is a local var, we may need to perform further adjustments e.g. for
+                    // normalization.
+                    if (temp->OperIs(GT_LCL_VAR))
+                    {
+#ifdef DEBUG
+                        // We clear this flag on `temp` because `fgMorphLocalVar` may assert that this bit is clear
+                        // and the node in question must have this bit set (as it has already been morphed).
+                        temp->gtDebugFlags &= ~GTF_DEBUG_NODE_MORPHED;
+#endif // DEBUG
+                        temp = fgMorphLocalVar(temp, true);
```
erozenfeld: Please use a named constant or enum instead of `true`.

src/jit/morph.cpp
```diff
@@ -8522,7 +8522,7 @@ GenTreePtr Compiler::fgMorphLeaf(GenTreePtr tree)
 
     if (tree->gtOper == GT_LCL_VAR)
     {
-        return fgMorphLocalVar(tree);
+        return fgMorphLocalVar(tree, false);
```
erozenfeld: Please use a named constant or enum instead of `false`.

briansull:
looks good

PR #11535, opened 2017-05-11
Remove unneeded Serializable attributes

morganbr:
Removes unneeded Serializable attributes from enums and delegates. They can be serialized without attributes and this will remove noise from future serialization changes.

CC @danmosemsft @joshfree @stephentoub


morganbr:

PR #11542, opened 2017-05-12
Refactor xarch codegen

BruceForstall:


BruceForstall:

mikedn:
Hurray for moving code to separate functions but I'm not so sure that moving code out of codegenxarch to codegenlinear is such a good idea. The amount of code is small and it's not super complicated or anything so there's not much to gain from sharing it. Instead it makes things more complicated for anyone who tries to improve xarch or arm codegen by forcing them to change both architectures or moving the code back to arch specific files.

BruceForstall:
There is certainly a cost to changing shared code that is higher than the cost of changing non-shared code. Hopefully we have adequate cross-architecture testing (arguably we don't). Of course, front-end changes already have this burden. The same argument has been made about sharing arm32 and arm64 functions, which are much closer. These changes are also potentially just the first set towards increasing cross-arch sharing. E.g., I'd like to make genCodeForTreeNode() itself shared across all architectures; it's very close right now.

BruceForstall:
@dotnet/jit-contrib PTAL: This change refactors xarch codegen from genCodeForTreeNode() into functions, the same as arm32/arm64. And in fact, several can be shared across all architectures and have been moved into codegenlinear.cpp.

pgavlin:
I share @mikedn's concerns here. I like the cleanup that's been done inside of codegenxarch, but I am not fond of moving the code generation for specific nodes into codegenlinear (for essentially the same reasons @mikedn stated above: namely, that it works at cross-purposes to our desire to make it easier to capitalize on attributes of particular platforms).

BruceForstall:
>  that it works at cross-purposes to our desire to make it easier to capitalize on attributes of particular platforms

I don't see that. Nothing is preventing that. There is a constant tension between sharing as much as possible (good), but not sharing too much when the complexity of the shared code (via excessive `#ifdef` for example) exceeds the benefit. We have a hierarchy of sharing currently, for example for arm32: codegencommon.cpp=>codegenlinear.cpp=>codegenarmarch.cpp=>codegenarm.cpp. I'd like to pull as much as possible (and beneficial, and without increasing complexity) upstream, to remove code duplication. If it later turns out that changes need to be made that reverse the considerations that made this beneficial, then functions can be pushed back downstream.

mikedn:
> and beneficial, and without increasing complexity

And what exactly is beneficial about moving `genCodeForJcc` for example? It works fine as is and it's unlikely that it will get radically changed soon. Except maybe if I manage to get my crazy relop lowering stuff in. And ironically that's the kind of change that won't benefit from shared code, at least not in initial stages and not as long as ARM is a moving target.

Oh well, I'll add more ifdefs and call it a day. But this still strikes me as a completly unnecessary change. The JIT code sucks in many ways but I never found the lack of code sharing to be an issue, quite the contrary. Just recently I stared at `genCreateAddrMode` and its myriad of ifdefs. Moving that from codegencommon to codegenlinear and cleaning it up would probably be a more worthwhile effort.

And I only looked at `genCreateAddrMode` because I was trying to track down a lame signed/unsigned issue that makes the JIT emit a 32 bit displacement when a 8 bit one would do. That says a thing or two about the JIT codebase quality.

So unsharing code? I'll gladly do it myself. When there's nothing else left to do.

CarolEidt:
I am inclined to favor moving code into codegenlinear.cpp when it is basically the same across architectures, though I don't feel really strongly about it. This refactoring is a bit inconvenient while I am (and presumably others are) working on what turns out to be larger scope changes because we've not been checking in incrementally while things are semi-frozen.

---
In reply to: [301155034](https://github.com/dotnet/coreclr/pull/11542#issuecomment-301155034) [](ancestors = 301155034)

CarolEidt:
:shipit:

briansull:
We do want to refactor common code so that we don't have identical duplicate implementations.
So as long as there are two or more duplicate implementations of a method it makes sense to move it into a common source file.  That make bug fixes and codegen improvements easier.

If there is a particular method that someone thinks need to be keep separate because we should be generating different code for an architecture (maybe we don't today) then we can add a comment and leave that method split out.

BruceForstall:
@pgavlin @mikedn @briansull @CarolEidt 

Though I don't agree with it, to make progress, I went ahead and put identical copies of four functions that generate code in both codegenarmarch.cpp and codegenxarch.cpp. I left genCodeForCast in codegenlinear.cpp, since it doesn't itself generate code but instead calls other functions to do so.

I plan to consider moving the main genCodeForTreeNode function (with the big switch over GenTree nodes) to codegenlinear.cpp. It will require a number of `#ifdefs`, but now, with this refactoring, not too many. Any comments/concerns in advance?

BruceForstall:

mikedn:
> I plan to consider moving the main genCodeForTreeNode function (with the big switch over GenTree nodes) to codegenlinear.cpp. It will require a number of #ifdefs, but now, with this refactoring, not too many. Any comments/concerns in advance?

Sounds good to me. Are those ifdefs needed because some opers are not defined on a particular arch or because they are not needed/used? If it's the later then perhaps it's better to just add empty arch specific functions?

BruceForstall:
> Are those ifdefs needed because some opers are not defined on a particular arch or because they are not needed/used? If it's the later then perhaps it's better to just add empty arch specific functions?

I'll have to look again. Empty arch-specific functions is a good option. GT_MUL is handled with a function on xarch, and genCodeForBinary on arm. Again, could just always have genCodeForMul().

mikedn:
> Again, could just always have genCodeForMul().

Yes, and have it call genCodeForBinary on arm.

BruceForstall:
@dotnet-bot test this please

BruceForstall:

BruceForstall:
x64 debug failure: chaos timed out:
```
08:32:59   cmdLine:D:\j\workspace\debug_windows---17180f1b\bin\tests\Windows_NT.x64.Debug\JIT\Generics\Coverage\chaos65204782cs_o\chaos65204782cs_o.cmd Timed Out
```
https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_windows_nt_prtest/7135/

BruceForstall:

PR #11549, opened 2017-05-12
Only call EventActivityIdControl ETW function on Windows

brianrob:
This fixes a break introduced by https://github.com/dotnet/coreclr/pull/11435.


brianrob:
This is the only place that I saw in the corefx run.  There were a lot so I could have missed some.  EventSource for the most part has a single interface that does the ETW calls, and that was handled as part of the initial change.  This is one of the places that breaks that abstraction and was in a different file in a different directory - thus why it was missed.

stephentoub:
Ok, thanks.

PR #11558, opened 2017-05-12
[Local GC] Fix a handle table interface violation

swgillespie:
This PR addresses a handle table interface violation that arose when building the GC without VM headers.

@Maoni0 @adityamandaleeka @jkotas  @sergiy-k ptal?


swgillespie:

PR #11568, opened 2017-05-12
EventPipe: Put the serialization marker under its own IFDEF

### LINKED ISSUE #11562: Remove FastSerialization DEBUG Markers
This will cause files produced in DEBUG to differ from those in RELEASE.  If we want to keep this, consider a mechanism that allows the reader to know about it and skip it.

brianrob:
Fixes #11562.


brianrob:

PR #11569, opened 2017-05-12
[Local GC] Add async pinned handle methods to the handle interface

swgillespie:
This is a stopgap PR for `Ref_RelocateAsyncPinHandles` and `Ref_HandleAsyncPinHandles` that fixes the interface violations that occur in the EE -> GC direction. The implementations of `Ref_RelocateAsyncPinHandles` and `Ref_HandleAsyncPinHandles` in the GC still call into the EE, which will need to be addressed at some point in the future. My understanding is that this is all dead code on CoreCLR anyway (dealing with the migration of pending async pinned handles when an appdomain unloads), so even if it were addressed for CoreCLR, we wouldn't be able to test it.

@Maoni0 @adityamandaleeka (already reviewed offline) @jkotas @sergiy-k ptal?

### DIFF
--- a/src/gc/gcinterface.h
+++ b/src/gc/gcinterface.h
@@ -399,6 +399,7 @@ typedef void (* record_surv_fn)(uint8_t* begin, uint8_t* end, ptrdiff_t reloc, v
 typedef void (* fq_walk_fn)(bool, void*);
 typedef void (* fq_scan_fn)(Object** ppObject, ScanContext *pSC, uint32_t dwFlags);
 typedef void (* handle_scan_fn)(Object** pRef, Object* pSec, uint32_t flags, ScanContext* context, bool isDependent);
+typedef bool (* async_pin_enum_fn)(Object* object, void* context);
 
 // Opaque type for tracking object pointers
 #ifndef DACCESS_COMPILE
@@ -428,7 +429,7 @@ public:
 
     virtual void RelocateAsyncPinnedHandles(IGCHandleStore* pTarget) = 0;
 
-    virtual bool EnumerateAsyncPinnedHandles(bool (*callback)(Object*, void*), void* context) = 0;
+    virtual bool EnumerateAsyncPinnedHandles(async_pin_enum_fn callback, void* context) = 0;
 
     virtual ~IGCHandleStore() {};
 };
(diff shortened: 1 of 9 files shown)


src/gc/gcinterface.h
```diff
@@ -463,6 +465,8 @@ class IGCHandleManager {
     virtual Object* GetDependentHandleSecondary(OBJECTHANDLE handle) = 0;
 
     virtual Object* InterlockedCompareExchangeObjectInHandle(OBJECTHANDLE handle, Object* object, Object* comparandObject) = 0;
+
+    virtual bool HandleAsyncPinnedHandles() = 0;
```
jkotas: This is very cryptic name

jkotas:
LGTM

src/gc/gcinterface.h
swgillespie: I agree - it's cryptic enough to me that I don't really know what it does or what it should be named, though...

jkotas: This method enumerates all asyncpinned handles, and calls back the VM if the handle is pointing to overlapped that satisfies `overlapped->GetAppDomainId() != DefaultADID && overlapped->HasCompleted()`. So it basically a custom asyncpinned handle enumerator.

You will need to do something about the callback - it is the `overlapped->HandleAsyncPinHandle()` line in handletablecore.cpp. It may be best to pass the callback as a pointer+context, similar to how enumeration is done in other places on the GC/EE interface.

Once you do that, the method then can be called EnumerateCompletedAsyncPinnedHandlesOwnedByNonDefaultAppDomain. Or if you move the `overlapped->GetAppDomainId() != DefaultADID && overlapped->HasCompleted()` check into the callback as well, it can be just `EnumerateAsyncPinnedHandles`.

Also, it would be more appropriate for this method to be on IGCHandleStore because of it operates on the global GC handle store only. The VM can then call it as `GetGlobalHandleStore()->EnumerateAsyncPinnedHandles(...)`.

swgillespie: Sounds good, will do!

Am I right in thinking that this is dead code in CoreCLR, so I won't be able to test it until it ends up in Desktop?

jkotas: Right

swgillespie:
thanks!

PR #11570, opened 2017-05-12
Use full move for byte registers stores in jumps.

### LINKED ISSUE #11509: [x86][JitStressRegs=4] Methodical\Invoke\25params\25param2b_il_r\25param2b_il_r fails
https://ci.dot.net/job/dotnet_coreclr/job/master/job/jitstress/job/x86_checked_windows_nt_jitstressregs4/156/testReport/

```
D:\j\workspace\x86_checked_w---b7304e59\bin\tests\Windows_NT.x86.Checked\TestWrappers\JIT.Methodical\JIT.Methodical.XUnitWrapper.cs(95867): error : JIT_Methodical._Invoke_25params_25param2b_il_r_25param2b_il_r_._Invoke_25params_25param2b_il_r_25param2b_il_r_cmd [FAIL] [D:\j\workspace\x86_checked_w---b7304e59\tests\runtest.proj]
        
  
  Return code:      1
  Raw output file:      D:\j\workspace\x86_checked_w---b7304e59\bin\tests\Windows_NT.x86.Checked\Reports\JIT.Methodical\Invoke\25params\25param2b_il_r\25param2b_il_r.output.txt
  Raw output:
  BEGIN EXECUTION
         "D:\j\workspace\x86_checked_w---b7304e59\tests\..\bin\tests\Windows_NT.x86.Checked\Tests\Core_Root\corerun.exe" 25param2b_il_r.exe 
        Testing method of 25 parameters, mixed data type, jmp
        1
        2
        3
        4
        5
        6
        7
        8
        9
        10
        11
        12
        13
        14
        15
        16
        17
        18
        19
        20
        21
        22
        23
        24
        25
        The sum is 318
        Expected: 100
        Actual: 1
        END EXECUTION - FAILED
        FAILED
        Test Harness Exitcode is : 1
```

sandreenko:
The pr was changes to fix the issue.
It makes CodeGen to use the 4byte move for this spill.
It is legal because we always own the 4 bytes memory slot.

It could change in the future, if we implement the multifields incoming struct promotion.

Fix #11509.

### DIFF
--- a/src/jit/codegenxarch.cpp
+++ b/src/jit/codegenxarch.cpp
@@ -5406,8 +5406,9 @@ void CodeGen::genJmpMethod(GenTreePtr jmp)
         // assert should hold.
         assert(varDsc->lvRegNum != REG_STK);
 
-        var_types loadType = varDsc->lvaArgType();
-        getEmitter()->emitIns_S_R(ins_Store(loadType), emitTypeSize(loadType), varDsc->lvRegNum, varNum, 0);
+        assert(!varDsc->lvIsStructField || (compiler->lvaTable[varDsc->lvParentLcl].lvFieldCnt == 1));
+        var_types storeType = genActualType(varDsc->lvaArgType()); // We own the memory and can use the full move.
+        getEmitter()->emitIns_S_R(ins_Store(storeType), emitTypeSize(storeType), varDsc->lvRegNum, varNum, 0);
 
         // Update lvRegNum life and GC info to indicate lvRegNum is dead and varDsc stack slot is going live.
         // Note that we cannot modify varDsc->lvRegNum here because another basic block may not be expecting it.
--- a/src/jit/emitxarch.cpp
+++ b/src/jit/emitxarch.cpp
@@ -4824,7 +4824,7 @@ void emitter::emitIns_S_R(instruction ins, emitAttr attr, regNumber ireg, int va
 #ifdef _TARGET_X86_
     if (attr == EA_1BYTE)
     {
-        noway_assert(isByteReg(ireg));
+        assert(isByteReg(ireg));
     }
 #endif
     // 16-bit operand instructions will need a prefix
--- a/src/jit/emitxarch.cpp
+++ b/src/jit/emitxarch.cpp
@@ -4821,6 +4821,12 @@ void emitter::emitIns_S_R(instruction ins, emitAttr attr, regNumber ireg, int va
     UNATIVE_OFFSET sz  = emitInsSizeSV(insCodeMR(ins), varx, offs);
     insFormat      fmt = emitInsModeFormat(ins, IF_SRD_RRD);
 
+#ifdef _TARGET_X86_
+    if (attr == EA_1BYTE)
+    {
+        noway_assert(isByteReg(ireg));
+    }
+#endif
     // 16-bit operand instructions will need a prefix
     if (EA_SIZE(attr) == EA_2BYTE)
     {
--- a/src/jit/lclvars.cpp
+++ b/src/jit/lclvars.cpp
@@ -1833,7 +1833,10 @@ bool Compiler::lvaShouldPromoteStructVar(unsigned lclNum, lvaStructPromotionInfo
 
             // TODO-PERF - Implement struct promotion for incoming multireg structs
             //             Currently it hits assert(lvFieldCnt==1) in lclvar.cpp line 4417
-
+            //             Also the implementation of jmp uses the 4 byte move to store
+            //             byte parameters to the stack, so that if we have a byte field
+            //             with something else occupying the same 4-byte slot, it will
+            //             overwrite other fields.
             if (structPromotionInfo->fieldCnt != 1)
         {
             JITDUMP("Not promoting promotable struct local V%02u, because lvIsParam is true and #fields = "


sandreenko:
@dotnet-bot test Windows_NT x86 Checked jitstressregs4

src/jit/emitxarch.cpp
```diff
@@ -4821,6 +4821,12 @@ void emitter::emitIns_S_R(instruction ins, emitAttr attr, regNumber ireg, int va
     UNATIVE_OFFSET sz  = emitInsSizeSV(insCodeMR(ins), varx, offs);
     insFormat      fmt = emitInsModeFormat(ins, IF_SRD_RRD);
 
+#ifdef _TARGET_X86_
+    if (attr <= EA_2BYTE)
+    {
+        noway_assert(ireg <= REG_EBX);
```
mikedn: There's an `isByteReg` function somewhere.

sandreenko: Thanks, fixed.

sandreenko:
@dotnet-bot test Windows_NT x86 Checked jitstressregs4

sandreenko:
@dotnet-bot test Windows_NT x86 Checked jitstressregs4

sandreenko:
I changed the PR, it was originally just the check, but now it fixes the issue.

PTAL @CarolEidt, @pgavlin, @dotnet/jit-contrib.

src/jit/lclvars.cpp
```diff
@@ -1833,7 +1833,8 @@ bool Compiler::lvaShouldPromoteStructVar(unsigned lclNum, lvaStructPromotionInfo
 
             // TODO-PERF - Implement struct promotion for incoming multireg structs
             //             Currently it hits assert(lvFieldCnt==1) in lclvar.cpp line 4417
-
+            //             Also there will be a problem with byte moves, because now it uses 4byte move.
+            //             It could rewrite the other fields in a promoted struct.
```
CarolEidt: You should be more specific here and say that it is the implementation of jmp that uses the 4 byte move to store byte parameters to the stack, so that if we have a byte field with something else occupying the same 4-byte slot, it will overwrite other fields.

src/jit/codegenxarch.cpp
```diff
@@ -5406,7 +5406,7 @@ void CodeGen::genJmpMethod(GenTreePtr jmp)
         // assert should hold.
         assert(varDsc->lvRegNum != REG_STK);
 
-        var_types loadType = varDsc->lvaArgType();
+        var_types loadType = genActualType(varDsc->lvaArgType()); // We own the memory and could use the full move.
```
CarolEidt: I think that this comment should be put with the comment above (and use "can" instead of "could"). Also, you should add an assert here that, if this is a field of a promoted struct, it is the only field (`lvIsStructField` is true for this, and `lvFieldCnt` is 1 for `lvParentLcl`). That way, if someone makes the change to promote struct parameters with multiple fields and misses the comment in lclVars.cpp, we are somewhat protected.

src/jit/codegenxarch.cpp
```diff
@@ -5406,7 +5406,12 @@ void CodeGen::genJmpMethod(GenTreePtr jmp)
         // assert should hold.
         assert(varDsc->lvRegNum != REG_STK);
 
-        var_types loadType = varDsc->lvaArgType();
+        if (varDsc->lvIsStructField)
```
pgavlin: Nit: you could write this as `assert(!varDsc->lvIsStructField || (compiler->lvaTable[varDsc->lvParentLcl].lvFieldCnt == 1))`

src/jit/lclvars.cpp
```diff
@@ -1833,8 +1833,10 @@ bool Compiler::lvaShouldPromoteStructVar(unsigned lclNum, lvaStructPromotionInfo
 
             // TODO-PERF - Implement struct promotion for incoming multireg structs
             //             Currently it hits assert(lvFieldCnt==1) in lclvar.cpp line 4417
-            //             Also there will be a problem with byte moves, because now it uses 4byte move.
-            //             It could rewrite the other fields in a promoted struct.
+            //             Also the implementation of jmp uses the 4 byte move to store 
+            //             byte parameters to the stack, so that if we have a byte field
+            //             with something else occupying the same 4-byte slot, it will 
+            //             overwtire other fields.
```
RussKeldorph: typo: overwrite

src/jit/codegenxarch.cpp
```diff
@@ -4653,7 +4653,7 @@ void CodeGen::genStoreInd(GenTreePtr node)
         }
         else
         {
-            getEmitter()->emitInsMov(ins_Store(data->TypeGet()), emitTypeSize(storeInd), storeInd);
+            getEmitter()->emitInsMov(ins_Store(data->TypeGet()), emitActualTypeSize(storeInd->TypeGet()), storeInd);
```
pgavlin: I think this change is incorrect: if we are issuing an indirect store, we need to issue it with the correct type. IIUC, `emitActualTypeSize` will widen small types.

src/jit/codegenxarch.cpp
```diff
@@ -1557,7 +1557,7 @@ void CodeGen::genCodeForTreeNode(GenTreePtr treeNode)
 
             GenTreePtr op1 = treeNode->gtGetOp1();
             genConsumeRegs(op1);
-            emit->emitInsBinary(ins_Store(targetType), emitTypeSize(treeNode), treeNode, op1);
```
pgavlin: Same comment here as below: even though this is a store to the stack, it is a store to an on-stack field and must therefore be appropriately-sized.

src/jit/codegenxarch.cpp
```diff
@@ -5406,7 +5406,12 @@ void CodeGen::genJmpMethod(GenTreePtr jmp)
         // assert should hold.
         assert(varDsc->lvRegNum != REG_STK);
 
-        var_types loadType = varDsc->lvaArgType();
+        if (varDsc->lvIsStructField)
+        {
+            LclVarDsc* parentDsc = compiler->lvaTable + varDsc->lvParentLcl;
+            assert(parentDsc->lvFieldCnt == 1);
+        }
+        var_types loadType = genActualType(varDsc->lvaArgType()); // We own the memory and can use the full move.
```
pgavlin: Nit: might as well rename this to `storeType`, since we're issuing a store :)

src/jit/emitxarch.cpp
```diff
@@ -4821,6 +4821,13 @@ void emitter::emitIns_S_R(instruction ins, emitAttr attr, regNumber ireg, int va
     UNATIVE_OFFSET sz  = emitInsSizeSV(insCodeMR(ins), varx, offs);
     insFormat      fmt = emitInsModeFormat(ins, IF_SRD_RRD);
 
+    assert(attr != EA_1BYTE);
+#ifdef _TARGET_X86_
+    if (attr == EA_1BYTE)
```
pgavlin: It is a bit weird that we assert that `attr` is not `EA_1BYTE` above and then immediately handle the case where it might be with another assertion. Can you add a comment here?

src/jit/codegenxarch.cpp
CarolEidt: @sandreenko - I agree with @pgavlin ; I hadn't noticed the addition of these two changes. This is only correct to do in the case where we can guarantee that this is a single field of the struct. This code is for the general case of `GT_STORE_LCL_FLD` for which we do not have that guarantee. When we have an explicit store of a byte in the code stream, `Lowering` should have requested a byte register.

src/jit/codegenxarch.cpp
CarolEidt: I also agree here.

sandreenko:
@dotnet-bot test Windows_NT x86 Checked jitstressregs4

briansull:
Can this be an issue for x64?

sandreenko:
> Can this be an issue for x64?

Do you mean the original problem? No, it could not happen on x64 because all registers were byte addressable.

PR #11577, opened 2017-05-12
Write EventSource events with metadata to EventPipe

helloguo:
This PR writes EventSource events with metadata to EventPipe.


helloguo:
@brianrob FYI

src/vm/eventpipe.cpp
```diff
@@ -515,15 +524,51 @@ void QCALLTYPE EventPipeInternal::DeleteProvider(
...
+        // pData is an array of pointers to blobs with Ptr of objects and Size of objects
+        // and we merge them here
+        EventProviderEventData *pNativeData = (EventProviderEventData *)pData;
+        unsigned int length = 0;
+        for (int i = 0; i < dataCount; i++)
+        {
+            length += pNativeData->Size;
+        }
+
+        BYTE *buffer = new BYTE[length];
```
brianrob: Consider: A static buffer of say 100 bytes.  If the data size is greater, then allocate.  I suspect many events will be under 100 bytes and then you won't have to alloc and free on the heap.

brianrob: I'm also wondering if you wouldn't be better off producing the buffer in managed code and then passing it into native code.  By creating the set of structs and then passing them down to native code, you have to marshal the data twice.

src/vm/eventpipeprovider.cpp
```diff
@@ -158,7 +158,9 @@ EventPipeEvent* EventPipeProvider::AddEvent(INT64 keywords, unsigned int eventID
         eventID,
         eventVersion,
         level,
-        needStack);
+        needStack,
```
brianrob: Perfect.  This is exactly what I had in mind.

src/mscorlib/shared/System/Diagnostics/Tracing/EventProvider.cs
```diff
@@ -1166,6 +1167,7 @@ internal unsafe protected bool WriteEvent(ref EventDescriptor eventDescriptor, G
             status = m_eventProvider.EventWriteTransferWrapper(
                 m_regHandle,
                 ref eventDescriptor,
+                IntPtr.Zero,
```
helloguo: Because this `EventWriteTransferWrapper` is called by `WriteEventRaw` instead of `WriteEvent`. I don't think we care about `WriteEventRaw` at this point.

src/mscorlib/shared/System/Diagnostics/Tracing/EventProvider.cs
```diff
@@ -1262,6 +1265,13 @@ int IEventProvider.EventActivityIdControl(UnsafeNativeMethods.ManifestEtw.Activi
                 ControlCode,
                 ref ActivityId);
         }
+
+        // Define an EventPipeEvent handle.
+        unsafe IntPtr IEventProvider.DefineEventHandle(uint eventID, string eventName, Int64 keywords, uint eventVersion, uint level, byte *pMetadata, uint metadataLength)
+        {
+            throw new System.NotSupportedException();
+            return IntPtr.Zero;
```
helloguo: It's removed.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -1199,7 +1287,7 @@ protected unsafe void WriteEventWithRelatedActivityIdCore(int eventId, Guid* rel
                                         m_eventData[eventId].Descriptor.Task,
                                         unchecked((long)etwSessions.ToEventKeywords() | origKwd));
 
-                                    if (!m_provider.WriteEvent(ref desc, pActivityId, relatedActivityId, eventDataCount, (IntPtr)data))
+                                    if (!m_provider.WriteEvent(ref desc, IntPtr.Zero, pActivityId, relatedActivityId, eventDataCount, (IntPtr)data))
```
helloguo: Yes, fixed.

src/vm/eventpipe.cpp
helloguo: Thanks for pointing it out. The metadata serialization is moved to managed side (`EventWriteTransferWrapper` in EventPipeEventProvider.cs).

helloguo:

src/mscorlib/src/System/Diagnostics/Eventing/EventPipeEventProvider.cs
```diff
@@ -49,11 +51,40 @@ uint IEventProvider.EventUnregister(long registrationHandle)
         unsafe int IEventProvider.EventWriteTransferWrapper(
             long registrationHandle,
             ref EventDescriptor eventDescriptor,
+            IntPtr eventHanlde,
```
brianrob: Nit: eventHandle.

src/vm/eventpipe.cpp
```diff
@@ -482,22 +482,34 @@ INT_PTR QCALLTYPE EventPipeInternal::CreateProvider(
...
+    EventPipeEvent *pEvent = NULL;
+
     BEGIN_QCALL;
 
-    // TODO
+    _ASSERTE(provHandle != NULL);
+    _ASSERTE(pMetadata != NULL);
+    _ASSERTE(metadataLength >= 20);
+    EventPipeProvider *pProvider = reinterpret_cast<EventPipeProvider *>(provHandle);
+    BYTE *buffer = new BYTE[metadataLength];
```
brianrob: Instead of allocating the buffer here, you should allocate it in the EventPipeEvent ctor if an input metadata buffer is passed in.  The metadata buffer needs to live for the life of the EventPipeEvent which is longer than this call.  Then, you can free the buffer in ~EventPipeEvent.

src/vm/eventpipe.cpp
```diff
@@ -517,13 +529,19 @@ void QCALLTYPE EventPipeInternal::DeleteProvider(
...
     void *pData,
     unsigned int length)
 {
     QCALL_CONTRACT;
     BEGIN_QCALL;
 
-    // TODO
+    _ASSERTE(eventHandle != NULL);
+    EventPipeEvent *pEvent = reinterpret_cast<EventPipeEvent *>(eventHandle);
+    BYTE *buffer = new BYTE[length];
```
brianrob: You can just pass pData directly to WriteEvent here.  It has the same lifetime as buffer here, so it's just an extra allocation.

src/vm/eventpipeevent.cpp
```diff
@@ -25,6 +25,8 @@ EventPipeEvent::EventPipeEvent(EventPipeProvider &provider, INT64 keywords, unsi
     m_level = level;
     m_needStack = needStack;
     m_enabled = false;
+    m_pMetadata = pMetadata;
+    m_metadataLength = metadataLength;
```
brianrob: Per my comment above, this is where you want to allocate a new buffer and copy the buffer pointed to by pMetadata into it.  Then free the buffer in ~EventPipeEvent.

brianrob:
LGTM modulo my last comment.

@vancem, can you please review this?

brianrob:
@helloguo, I just pushed another commit to write the metadata to the file (from the pointers on the EventPipeEvent that you plumbed through).

I also just tested this and as far as I can tell, everything works.  Let's wait on @vancem for the final review.

brianrob:
@dotnet-bot test Windows_NT x64 Debug Build and Test ("cmdLine:D:\j\workspace\debug_windows---17180f1b\bin\tests\Windows_NT.x64.Debug\JIT\Generics\Coverage\chaos65204782cs_o\chaos65204782cs_o.cmd Timed Out")

src/mscorlib/shared/System/Diagnostics/Tracing/EventProvider.cs
vancem: I confirmed that WriteEventRaw is only used on the TraceLogging code path, so we can ignore this for now.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -686,6 +686,115 @@ protected EventSource(EventSourceSettings settings, params string[] traits)
...
+            {
+                uint eventID = (uint)m_eventData[i].Descriptor.EventId;
+                if (eventID == 0)
+                    continue;
+
+                string eventName = m_eventData[i].Name;
+                Int64 keywords = m_eventData[i].Descriptor.Keywords;
+                uint eventVersion = m_eventData[i].Descriptor.Version;
+                uint level = m_eventData[i].Descriptor.Level;
+                byte[] eventNameBytes = Encoding.ASCII.GetBytes(eventName);
```
vancem: Actually the default format for serializing strings in ETW (and therefore EventSource), is a null terminated unicode string.  Thus you don't need to convert to ascii (you can do a 'fixed(char* x = s)).     The same will be true for parameter names below.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -686,6 +686,115 @@ protected EventSource(EventSourceSettings settings, params string[] traits)
...
+            Debug.Assert(bufferLength >= (offset + srcLength));
+            for (int i = 0; i < srcLength; i++)
+            {
+                *(byte *)(buffer + offset + i) = *(byte *)(src + i);
+            }
+            offset += srcLength;
+        }
+
+        // Copy byte value to buffer.
+        private unsafe void WriteToBuffer(byte *buffer, uint bufferLength, ref uint offset, byte value)
```
vancem: When do you use this overload (that writes a byte?)  I dont' see it.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -686,6 +686,115 @@ protected EventSource(EventSourceSettings settings, params string[] traits)
...
+                fixed (byte *pMetadata = metadata, pEventNameBytes = eventNameBytes)
+                {
+                    uint offset = 0;
+                    WriteToBuffer(pMetadata, metadataLength, ref offset, eventID);
+                    WriteToBuffer(pMetadata, metadataLength, ref offset, pEventNameBytes, (uint)eventNameBytes.Length);
+                    WriteToBuffer(pMetadata, metadataLength, ref offset, (byte)0);
+                    WriteToBuffer(pMetadata, metadataLength, ref offset, keywords);
+                    WriteToBuffer(pMetadata, metadataLength, ref offset, eventVersion);
+                    WriteToBuffer(pMetadata, metadataLength, ref offset, level);
+                    foreach (var parameter in m_eventData[i].Parameters)
```
vancem: You need to prefix arrays with their length (otherwise the decoder does not know how many elements there are).

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
helloguo: Thanks for pointing it out. will fix it.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -686,6 +686,115 @@ protected EventSource(EventSourceSettings settings, params string[] traits)
...
+                        fixed (byte *pParameterNameBytes = parameterNameBytes)
+                        {
+                            WriteToBuffer(pMetadata, metadataLength, ref offset, pParameterNameBytes, (uint)parameterNameBytes.Length);
+                        }
+
+                        // Write terminated 0.
+                        WriteToBuffer(pMetadata, metadataLength, ref offset, (byte)0);
+                    }
+
+                    IntPtr eventHandle = m_provider.m_eventProvider.DefineEventHandle(eventID, eventName, keywords, eventVersion, level, pMetadata, metadataLength);
```
vancem: Add an assert that the expected offset is now exactly at the length of the buffer.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
helloguo: To add the terminated 0 when copying eventName and parameterName. I guess this is not needed if I use unicode string instead of ascii.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -686,6 +686,115 @@ protected EventSource(EventSourceSettings settings, params string[] traits)
             Initialize(eventSourceGuid, eventSourceName, traits);
         }
 
+#if FEATURE_PERFTRACING
+        // Generate all EventPipeEvent instances which belong to this EventProvider
+        private unsafe void GenerateEventPipeEvents()
```
vancem: Name this DefineEventPipeEvents.    It more strongly implies that this is a one time thing.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -686,6 +686,115 @@ protected EventSource(EventSourceSettings settings, params string[] traits)
             Initialize(eventSourceGuid, eventSourceName, traits);
         }
 
+#if FEATURE_PERFTRACING
+        // Generate all EventPipeEvent instances which belong to this EventProvider
```
vancem: Change comment to 
    
// Generate the serialized blobs that describe events for all strongly typed events (that is events that define strongly
// typed event methods.   Dynamically defined events (that use Write<T>) hare defined on the fly and are 
// handled elsewhere.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -686,6 +686,115 @@ protected EventSource(EventSourceSettings settings, params string[] traits)
...
+
+                    IntPtr eventHandle = m_provider.m_eventProvider.DefineEventHandle(eventID, eventName, keywords, eventVersion, level, pMetadata, metadataLength);
+                    m_eventData[i].EventHandle = eventHandle; 
+                }
+            }
+        }
+
+        // Copy src to buffer and modify the offset.
+        // Note: We know the buffer size ahead of time to make sure no buffer overflow.
+        private unsafe void WriteToBuffer(byte *buffer, uint bufferLength, ref uint offset, byte *src, uint srcLength)
```
vancem: make all of these static.

src/vm/eventpipe.cpp
```diff
@@ -482,22 +482,31 @@ INT_PTR QCALLTYPE EventPipeInternal::CreateProvider(
...
+    EventPipeEvent *pEvent = NULL;
+
     BEGIN_QCALL;
 
-    // TODO
+    _ASSERTE(provHandle != NULL);
+    _ASSERTE(pMetadata != NULL);
+    _ASSERTE(metadataLength >= 20);
+    EventPipeProvider *pProvider = reinterpret_cast<EventPipeProvider *>(provHandle);
+    pEvent = pProvider->AddEvent(keywords, eventID, eventVersion, (EventPipeEventLevel)level, (BYTE *)pMetadata, metadataLength);
```
vancem: I would recommend reordering the arguments here as well (eventID should come before keywords), because eventID really is the most important item.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -2230,7 +2338,7 @@ private unsafe void WriteEventString(EventLevel level, long keywords, string msg
                         data.Ptr = (ulong)msgStringPtr;
                         data.Size = (uint)(2 * (msgString.Length + 1));
                         data.Reserved = 0;
-                        m_provider.WriteEvent(ref descr, null, null, 1, (IntPtr)((void*)&data));
+                        m_provider.WriteEvent(ref descr, IntPtr.Zero, null, null, 1, (IntPtr)((void*)&data));
```
vancem: I believe this will causes crashes (In native code in WriteEvent() we expect EventHandles to be well defined).   This code is used when EventSource logs error messages about the EventSource itself.  

The easiest way to defined the Event with ID zero to be this event (it is already sort of used this way.  

      m_eventData[0].Handle for this.  

Note that this 'works' OK (it will no longer crash), but we should also fix CreateManifestAndDescriptors to provide a better name, that is change 
    eventData[0].Name = ""

to 
     eventData[0].Name = "EventSourceMessage"

and provide it with meta-data (you can do this by setting eventData[0].Parameters to a method that has a one string argument with name 'message' and your code will generate the right meta-data for it.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -3119,7 +3232,7 @@ private unsafe bool SendManifest(byte[] rawManifest)
                     dataDescrs[1].Size = (uint)Math.Min(dataLeft, chunkSize);
                     if (m_provider != null)
                     {
-                        if (!m_provider.WriteEvent(ref manifestDescr, null, null, 2, (IntPtr)dataDescrs))
+                        if (!m_provider.WriteEvent(ref manifestDescr, IntPtr.Zero, null, null, 2, (IntPtr)dataDescrs))
```
vancem: I have confirmed that because we don't set m_rawmanifest we never get to SendManfest and thus passing a null handle is OK here.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
vancem: I see that we don't crash (we have a check), but it also means we don't get any diagnostic information (these events are dropped).   Thus my suggestion above still stands.

src/mscorlib/src/System/Diagnostics/Eventing/EventPipeEventProvider.cs
```diff
@@ -49,11 +51,40 @@ uint IEventProvider.EventUnregister(long registrationHandle)
...
+                    return 0;
+                }
+
+                uint length = 0;
+                for (int i = 0; i < userDataCount; i++)
+                {
+                    length += userData[i].Size;
+                }
+
+                byte[] data = new byte[length];
```
vancem: We don't want any allocations on the normal logging path.  This is also another buffer copy that we can avoid.     
We can leave this in the VERY short term (today), but I would like to change thing so that

EventPipeInternal.WriteEvent takes a int userDataCount, and  EventProvider.EventData* userData.   instead of a byte* and length.    You should simply pass along these until the real work happens in EventPipeBufferManager.WriteEvent, and then do the one and only copy needed there.

src/mscorlib/src/System/Diagnostics/Eventing/EventPipeEventProvider.cs
```diff
@@ -62,5 +93,24 @@ int IEventProvider.EventActivityIdControl(UnsafeNativeMethods.ManifestEtw.Activi
...
+        // Define an EventPipeEvent handle.
+        unsafe IntPtr IEventProvider.DefineEventHandle(uint eventID, string eventName, Int64 keywords, uint eventVersion, uint level, byte *pMetadata, uint metadataLength)
+        {
+            IntPtr eventHandlePtr = EventPipeInternal.DefineEvent(m_provHandle, eventID, keywords, eventVersion, level, pMetadata, metadataLength);
+            return eventHandlePtr;
+        }
+
+        // Copy src to buffer and modify the offset.
+        // Note: We know the buffer size ahead of time to make sure no buffer overflow.
+        private unsafe void WriteToBuffer(byte *buffer, uint bufferLength, ref uint offset, byte *src, uint srcLength)
```
vancem: May not need this if you follow the advice above and do the buffer copy in native code.

vancem:
I am done with my review, but I did have things I want to see updated.

brianrob:
@vancem, thank you very much.

@helloguo, let us know if you have any questions.

src/mscorlib/src/System/Diagnostics/Eventing/EventPipeEventProvider.cs
brianrob: Agreed on this.  I have filed https://github.com/dotnet/coreclr/issues/11615 to track this.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
vancem: The overloads below are static but not this overload.  

private unsafe void WriteToBuffer(byte *buffer, uint bufferLength, ref uint offset, byte *src, uint srcLength)

brianrob:
LGTM.

Issue #11574, opened 2017-05-12
Silent bad codegen when optimizing expression


briansull:
Repro:

Compile Repro.cs below:

.\Tools\dotnetcli\dotnet.exe Tools\csc.exe /o+ /out:Program.exe Repro.cs /r:.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Private.CoreLib.dll /r:.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Console.dll
/r:.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Runtime.dll

Repro.cs

```
using System;
class Program
{
    static byte[] s_arr2;
    static byte[] s_arr3;

    static void Init()
    {
        s_arr2 = new byte[] { 0x11, 0x12, 0x13 };
        s_arr3 = new byte[] { 0x21, 0x22, 0x33 };
    }

    static void Main(string[] args)
    {
        Init();

        byte[] arr1 = new byte[] { 2 };
        byte[] arr2 = s_arr2;
        byte[] arr3 = s_arr3;

        int len = arr1.Length + arr2.Length + arr3.Length;
        int cur = 0;
        Console.WriteLine("1: cur = {0}", cur);
        cur += arr1.Length;
        Console.WriteLine("2: cur += {0}, now {1}", arr1.Length, cur);
        cur += arr2.Length;
        Console.WriteLine("3: cur += {0}, now {1}", arr2.Length, cur);
        cur += arr3.Length;
        Console.WriteLine("4: cur += {0}, now {1}", arr3.Length, cur);
        Console.WriteLine("5: len is {0}", len);
    }
}
```

Run the Repro:
.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\CoreRun.exe Repro.exe
outputs:
1: cur = 0
2: cur += 1, now 1
3: cur += 3, now 6
4: cur += 3, now 7
5: len is 7

Problem is "3: cur += 3, now 6"
The value of 6 is incorrect and should be 4

pgavlin:
This is Yet Another Remorphing Issue:

```
Considering CSE #04 [def=100, use=100, cost= 9] CSE Expression:
N006 (  3,  3) CSE #03 (def)[000037] ---X--------             /--*  arrLen    int    <l:$443, c:$442>
[...]

PR #11579, opened 2017-05-12
Fix issue #11574.

pgavlin:
This bug was yet another case of a transformation being performed during
remorphing that preserved a tree's value number without preserving its
assocated semantics. This fix disables the problematic transformations
in morph.

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -13665,7 +13665,7 @@ GenTree* Compiler::fgMorphSmpOpOptional(GenTreeOp* tree)
     GenTree*   op2  = tree->gtOp2;
     var_types  typ  = tree->TypeGet();
 
-    if (!optValnumCSE_phase && GenTree::OperIsCommutative(oper))
+    if (fgGlobalMorph && GenTree::OperIsCommutative(oper))
     {
         /* Swap the operands so that the more expensive one is 'op1' */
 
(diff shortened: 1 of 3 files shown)


pgavlin:
@briansull @erozenfeld @dotnet/jit-contrib PTAL

src/jit/morph.cpp
```diff
@@ -13665,7 +13665,7 @@ GenTree* Compiler::fgMorphSmpOpOptional(GenTreeOp* tree)
     GenTree*   op2  = tree->gtOp2;
     var_types  typ  = tree->TypeGet();
 
-    if (GenTree::OperIsCommutative(oper))
+    if (!optValnumCSE_phase && GenTree::OperIsCommutative(oper))
```
JosephTremoulet: This fix looks like it would suffer the same issue that #11207 addresses in some other code - - `optValnumCSE_phase` becomes false before we're done relying on correct value numbers; we use the value numbers in assertion prop and range check opt, which run after CSE. You could (maybe?)  craft a test case where the thing with the bad value number is used to index into an array, and get us to incorrectly remove the bonds check, even with this fix (if I'm following correctly). Not sure what would be the best adjustment, seems like you could either have the transform erase the value number or add a flag (`optValueNumbersValid`?) that's like `OptValnumCSE_phase` but stays true until we're done with value numbers...

briansull: We could change the reordering optimization to only run in the first global morph phase.

erozenfeld:
If you disable this when re-morphing, please check whether we get any diffs on core and SPMI.

src/jit/morph.cpp
```diff
@@ -13703,7 +13703,7 @@ GenTree* Compiler::fgMorphSmpOpOptional(GenTreeOp* tree)
     /* Change "((x+icon)+y)" to "((x+y)+icon)"
        Don't reorder floating-point operations */
 
-    if ((oper == GT_ADD) && !tree->gtOverflow() && (op1->gtOper == GT_ADD) && !op1->gtOverflow() &&
+    if (!optValnumCSE_phase && (oper == GT_ADD) && !tree->gtOverflow() && (op1->gtOper == GT_ADD) && !op1->gtOverflow() &&
```
briansull: While you are here I would also reorder the check that excludes floating point so it comes early.

```
    if (fgGlobalMorph && varTypeIsIntegralOrI(typ) &&
        (oper == GT_ADD) && !tree->gtOverflow() && (op1->gtOper == GT_ADD) && !op1->gtOverflow())       
```

PR #11583, opened 2017-05-12
Define the value ID_EXTRA_RELOC_BITS as 0 when HAS_TINY_DESC is 1

briansull:
Otherwise there are compilation errors when this value is used in computations later on.


briansull:
@dotnet/jit-contrib PTAL

briansull:
There is still a build failures for this:

d:\codegenmirror\src\ndp\clr\src\jit\emit.h(2266): error C2118: negative subscript

PR #11587, opened 2017-05-13
Fix String.IndexOf with empty string input

### LINKED ISSUE #11578: string.IndexOf not returning proper value when searching for an empty string
string.IndexOf is not returning the proper value when searching for an empty string and providing a start index. Works fine on .NET Framework and .NET Core 1.1.
```C#
Debug.Assert("abc".IndexOf("", 2) == 2);
```

tarekgh:
Fixes #11578


tarekgh:
@jkotas could you please have a quick look? Thanks.

tarekgh:
> Could you please also add regression test to corefx?

Yes I already has the test ready. actually the added test is how I tested the fix here.

jkotas:

PR #11589, opened 2017-05-13
Fix improper handling of GenTreeAddrMode offset

mikedn:
For unknown reasons `GenTreeAddrMode::gtOffset` is `unsigned` rather than `int`. On 32 bit hosts this doesn't really matter but on 64 bit hosts we end up zero extending instead of sign extending it, `GetTreeIndir::Offset()` does that by casting from `unsigned` to `size_t`.

It doesn't appear possible for this to cause correctness issues (because the address mode displacement is 32 bit anyway) but it's confusing and hurts CQ because the emitter can't recognize small negative displacements that can be encoded in a single byte.

It's worth noting that 
* `CodeGen::genCreateAddrMode` uses `ssize_t` internally to represent the offset but then returns it as `unsigned`
* the emitter works with either `int` or `ssize_t` displacements (e.g. `emitIns_R_ARX` & co. have `int disp` arguments, `emitNewInstrAmd` has `ssize_t dsp` argument)

Contributes to #10980

### DIFF
--- a/src/jit/gentree.h
+++ b/src/jit/gentree.h
@@ -4533,9 +4533,24 @@ struct GenTreeAddrMode : public GenTreeOp
         return gtOp2;
     }
 
-    unsigned gtScale;  // The scale factor
+    int Offset()
+    {
+        return static_cast<int>(gtOffset);
+    }
+
+    unsigned gtScale; // The scale factor
+
+#ifndef LEGACY_BACKEND
+private:
+#endif
+    // TODO-Cleanup: gtOffset should be changed to 'int' to match the getter function and avoid accidental
+    // zero extension to 64 bit. However, this is used by legacy code and initialized, via the offset
+    // parameter of the constructor, by Lowering::TryCreateAddrMode & CodeGenInterface::genCreateAddrMode.
+    // The later computes the offset as 'ssize_t' but returns it as 'unsigned'. We should change
+    // genCreateAddrMode to return 'int' or 'ssize_t' and then update this as well.
     unsigned gtOffset; // The offset to add
 
+public:
     GenTreeAddrMode(var_types type, GenTreePtr base, GenTreePtr index, unsigned scale, unsigned offset)
         : GenTreeOp(GT_LEA, type, base, index)
     {
@@ -4570,7 +4585,7 @@ struct GenTreeIndir : public GenTreeOp
     GenTree* Base();
     GenTree* Index();
     unsigned Scale();
-    size_t   Offset();
+    ssize_t  Offset();
 
     GenTreeIndir(genTreeOps oper, var_types type, GenTree* addr, GenTree* data) : GenTreeOp(oper, type, addr, data)
     {
--- a/src/jit/codegenarmarch.cpp
+++ b/src/jit/codegenarmarch.cpp
@@ -3148,7 +3148,7 @@ void CodeGen::genLeaInstruction(GenTreeAddrMode* lea)
     genConsumeOperands(lea);
     emitter* emit   = getEmitter();
     emitAttr size   = emitTypeSize(lea);
-    unsigned offset = lea->gtOffset;
+    int      offset = lea->Offset();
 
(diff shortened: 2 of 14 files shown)


mikedn:
jit-diff summary:
```
Total bytes of diff: -3311 (-0.03 % of base)
    diff is an improvement.
       -3293 : System.Private.CoreLib.dasm (-0.10 % of base)
         -12 : System.Private.Uri.dasm (-0.02 % of base)
          -6 : System.Text.Encoding.CodePages.dasm (-0.01 % of base)
3 total files with size differences (3 improved, 0 regressed), 76 unchanged.
        -312 : System.Private.CoreLib.dasm - Comparer`1:System.Collections.IComparer.Compare(ref,ref):int:this (39 methods)
        -312 : System.Private.CoreLib.dasm - EqualityComparer`1:System.Collections.IEqualityComparer.Equals(ref,ref):bool:this (48 methods)
        -156 : System.Private.CoreLib.dasm - EqualityComparer`1:System.Collections.IEqualityComparer.GetHashCode(ref):int:this (48 methods)
        -144 : System.Private.CoreLib.dasm - Dictionary`2:System.Collections.IDictionary.set_Item(ref,ref):this (29 methods)
        -144 : System.Private.CoreLib.dasm - Dictionary`2:System.Collections.IDictionary.Add(ref,ref):this (29 methods)
172 total methods with size differences (172 improved, 0 regressed), 65378 unchanged.
```

mikedn:
Review note: the whole change practically reduces to treating `gtOffset` as signed in `GenTreeIndir::Offset()`. To try to convince myself that this is actually correct I took some intermediary step such as renaming `GenTreeIndir::Offset` to `GenTreeIndir::Address` and back to `GenTreeIndir::Offset` to ensure that all its uses have been looked at. We'll need to squash this before merging.

I'll have to look into ARM specific code to ensure that it is not affected by this so WIP for now.

mikedn:
And yes, ARM specific code had similar issues. emitarm64 already treated `gtOffset` as signed and generated the expected code but lsraarmarch treated the offset as unsigned and as a result an internal integer register was unnecessarily allocated.

mikedn:

BruceForstall:
@dotnet/jit-contrib PTAL

src/jit/gentree.h
```diff
@@ -4240,9 +4240,17 @@ struct GenTreeAddrMode : public GenTreeOp
...
+    int Offset()
+    {
+        return static_cast<int>(gtOffset);
+    }
 
+    unsigned gtScale; // The scale factor
+#ifndef LEGACY_BACKEND
+private:
+#endif
+    unsigned gtOffset; // The offset to add
```
CarolEidt: I am probably missing something obvious here, but why not change the type of the field?

mikedn: Eh, like the temporary renaming of `GenTreeIndir::Offset` to `Address` this was a way to ensure that I found all it's usages. Make it private, add `Offset()` and update all uses of `gtOffset` until it builds. It's also used by legacy code and I'd rather not touch that.

danmosemsft:
@mikedn still going or waiting on review?

mikedn:
@danmosemsft Waiting on review

danmosemsft:
@CarolEidt do you have time to take another look?

src/jit/gentree.h
CarolEidt: Sorry I failed to respond to this. That strategy makes sense, though a comment to that effect would avoid confusion.

mikedn:
@CarolEidt I've added 2 cleanup comments, one to the `gtOffset` member and the other inside `genCreateAddrMode`, where it casts from `ssize_t` to `unsigned` for no good reason. I hope to clean up the later at some point, I've yet to convince myself that keeping `genCreateAddrMode` shared between ARM and XARCH is a good thing.

mikedn:
> Might be good to squash the commits before merging.

Yes, done.

Issue #11590, opened 2017-05-13
Minidumps created with new utility hang under lldb 3.9.1


mikem8361:
During my testing of the LLDB 3.9.1 Ubuntu packages, I noticed that the dumps created with the new createdump utility hung LLDB with loaded. Up to this point, I was testing these minidumps using a version LLDB built from the MS repo that had some fixes from the SQL group that must have hid this problem.

The fix to createdump is very simple: the thread state in the dump needs an stop reason signal.

mikem8361:
Done.

PR #11591, opened 2017-05-13
Fix minidump hang under shipping lldb 3.9.1

mikem8361:
The threads in the dump need a stop reason SIGABRT in this case.

Issue: #11590


lt72:
LGTM

PR #11600, opened 2017-05-14
EventPipe: Prepend variable length field sizes

### LINKED ISSUE #11563: Serialize the Number of Stack Frames in EventPipeEventInstance
Right now, the parser knows that the stack is the last item in the serialized EventPipeEventInstance.  It would be good to log the # of frames so that it can know if there is something beyond it and old parsers can skip it.

brianrob:
Fixes #11563.


brianrob:
@vancem, @valenis FYI.

brianrob:
@dotnet-bot test Tizen armel Cross Debug Build ("FAILED   - JIT/Regression/CLR-x86-JIT/V1-M09.5-PDC/b28037/b28037/b28037.sh")

PR #11613, opened 2017-05-15
Throw PNSE in ThreadPoolBoundHandle.BindHandle on Unixes

kouvel:
Functional fix for https://github.com/dotnet/corefx/issues/19785

### DIFF
--- a/src/mscorlib/src/System/Threading/ClrThreadPoolBoundHandle.cs
+++ b/src/mscorlib/src/System/Threading/ClrThreadPoolBoundHandle.cs
@@ -77,34 +77,7 @@ namespace System.Threading
         /// </remarks>
         public static ThreadPoolBoundHandle BindHandle(SafeHandle handle)
         {
-            if (handle == null)
-                throw new ArgumentNullException(nameof(handle));
-
-            if (handle.IsClosed || handle.IsInvalid)
-                throw new ArgumentException(SR.Argument_InvalidHandle, nameof(handle));
-
-            try
-            {
-                // ThreadPool.BindHandle will always return true, otherwise, it throws. See the underlying FCall
-                // implementation in ThreadPoolNative::CorBindIoCompletionCallback to see the implementation.
-                bool succeeded = ThreadPool.BindHandle(handle);
-                Debug.Assert(succeeded);
-            }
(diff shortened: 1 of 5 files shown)


kouvel:
Types in the S.T.Overlapped Unix assembly now type-forward to S.P.CoreLib, so this cleans up the one issue I found, to throw PNSE.

@alexperovich @stephentoub

src/mscorlib/src/System/Threading/ClrThreadPoolBoundHandle.cs
```diff
@@ -77,34 +77,7 @@ public SafeHandle Handle
...
-                throw new ArgumentNullException(nameof(handle));
-
-            if (handle.IsClosed || handle.IsInvalid)
-                throw new ArgumentException(SR.Argument_InvalidHandle, nameof(handle));
-
-            try
-            {
-                // ThreadPool.BindHandle will always return true, otherwise, it throws. See the underlying FCall
-                // implementation in ThreadPoolNative::CorBindIoCompletionCallback to see the implementation.
-                bool succeeded = ThreadPool.BindHandle(handle);
```
stephentoub: Should it actually be ThreadPool.BindHandle that throws the PNSE, with the rest of this method not changing?  Seems like we'd want the arg validation on handle, for example, to be consistent across platforms, even though then once you get past that on Unix you'd hit the PNSE.

kouvel: ThreadPool.BindHandle returns true for success and false for failure, so on Unix it returns false. I'll move the PNSE check after the argument validation.

kouvel: I'm fine with making ThreadPool.BindHandle throw PNSE though if that's what we want

stephentoub: > ThreadPool.BindHandle returns true for success and false for failure, so on Unix it returns false. I'll move the PNSE check after the argument validation.

Ok

kouvel:
@dotnet test Tizen armel Cross Debug Build

PR #11618, opened 2017-05-15
Fix minidump hang under shipping lldb 3.9.1

mikem8361:
Issue: #11590


mikem8361:
Don't need this until preview2.

PR #11620, opened 2017-05-15
Update Coreclr to 2.0.0 buildtools

wtgodbe:
This change brings CoreCLR up to the 2.0.0 version of buildtools. This means we no longer use any project.json files or project.json related targets, but instead list our dependencies as PackageReferences inside of .csproj files.


src/ToolBox/SOS/NETCore/SOS.NETCore.csproj
```diff
@@ -39,8 +41,25 @@
...
+    </PackageReference>
+    <PackageReference Include="System.IO.FileSystem">
+      <Version>4.0.1</Version>
+    </PackageReference>
+    <PackageReference Include="System.Runtime.InteropServices">
+      <Version>4.1.0</Version>
+    </PackageReference>
+    <PackageReference Include="System.Reflection.Metadata">
+      <Version>1.4.1</Version>
+    </PackageReference>
```
akoeplinger: this could be shortened to:

```
<PackageReference Include="Microsoft.NETCore.Platforms" Version="$(RuntimeIdGraphDefinitionVersion)" />
<PackageReference Include="System.IO.FileSystem" Version="4.0.1" />
<PackageReference Include="System.Runtime.InteropServices" Version="4.1.0" />
<PackageReference Include="System.Reflection.Metadata" Version="1.4.1" />
```

weshaggard: @akoeplinger because we aren't using the new msbuild in all cases yet so we cannot use the short form of this just yet.

build.proj
```diff
@@ -40,7 +40,8 @@
   <Target Name="Sync" BeforeTargets="Build" />
 
   <Target Name="RestoreNETCorePlatforms" AfterTargets="Build" Condition="'$(RestoreDuringBuild)'=='true'">
-    <Exec Command="$(DnuRestoreCommand) &quot;$(SourceDir).nuget/init/project.json&quot; --source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
+    <Exec Command="$(DnuRestoreCommand) $(SourceDir).nuget/init/init.csproj"
```
weshaggard: We should be switching to use DotnetRestoreCommand anywhere we directly call restore now.

dir.props
```diff
@@ -103,7 +103,8 @@
     <NugetRestoreCommand>$(NugetRestoreCommand) install</NugetRestoreCommand>
     <!-- Trim off the last slash so that nuget doesn't get confused and say there are illegal characters on
          the path. -->
-    <NugetRestoreCommand>$(NugetRestoreCommand) -OutputDirectory "$(PackagesDir.TrimEnd('\\'))"</NugetRestoreCommand>
+    <NugetRestoreCommand Condition="'$(OsEnvironment)'=='Windows_NT'">$(NugetRestoreCommand) -OutputDirectory "$(PackagesDir.TrimEnd('\\'))"</NugetRestoreCommand>
```
weshaggard: Where do you guys use NugetRestoreCommand? Hopefully no where and this can just be deleted.

dir.props
```diff
@@ -113,9 +114,10 @@
     <DotnetToolCommand Condition="'$(DotnetToolCommand)'=='' and '$(OsEnvironment)'!='Unix'">$(DotnetCliPath)dotnet.exe</DotnetToolCommand>
     <DotnetToolCommand Condition="'$(DotnetToolCommand)'=='' and '$(OsEnvironment)'=='Unix'">$(DotnetCliPath)dotnet</DotnetToolCommand>
 
-    <DnuRestoreCommand>$(DnuRestoreCommand) "$(DotnetToolCommand)"</DnuRestoreCommand>
```
weshaggard: We define DotnetToolCommand and DotnetRestoreCommand in buildtools now so you can remove these and stop the need to maintain them.

init-tools.msbuild
```diff
@@ -0,0 +1,16 @@
...
+  <PropertyGroup>
+    <TargetFramework>netcoreapp1.0</TargetFramework>
+    <EnableDefaultItems>false</EnableDefaultItems>
+    <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
+    <BaseIntermediateOutputPath>$(MSBuildThisFileDirectory)Tools/$(BuildToolsPackageVersion)</BaseIntermediateOutputPath>
+  </PropertyGroup>
+   <ItemGroup>
+    <PackageReference Include="Microsoft.DotNet.BuildTools">
+	  <Version>$(BuildToolsPackageVersion)</Version>
+	</PackageReference>
```
weshaggard: nit: indention.

tests/build.proj
```diff
@@ -28,32 +46,19 @@
...
-    <ProjectJsonFiles Include="$(ProjectDir)scripts\**\project.json" />
-  </ItemGroup>
-
-  <Target Name="UpdatePackageDependencyVersion">
-    <UpdatePackageDependencyVersion ProjectJsons="@(ProjectJsonFiles)"
-                                    PackageId="$(PackageId)"
-                                    OldVersion="$(OldVersion)"
-                                    NewVersion="$(NewVersion)" />
+  <Target Name="RestorePackage">
+    <Exec Command="$(DnuRestoreCommand) $(RestoreProj)"
```
weshaggard: DotnetRestoreCommand in general you we should not have any references to Dnu anywhere any longer.

tests/dir.props
```diff
@@ -104,9 +104,8 @@
     <DotnetToolCommand Condition="'$(DotnetToolCommand)'=='' and '$(OsEnvironment)'=='Unix'">$(DotnetCliPath)dotnet</DotnetToolCommand>    
 
     <DnuRestoreSource>@(DnuSourceList -> '--source %(Identity)', ' ')</DnuRestoreSource>
-    <DnuRestoreDirs>@(DnuRestoreDir -> '%(Identity)', ' ')</DnuRestoreDirs>
 
-    <DnuRestoreCommand>$(DnuRestoreCommand) "$(DotnetToolCommand)"</DnuRestoreCommand>
+    <DnuRestoreCommand>"$(DotnetToolCommand)"</DnuRestoreCommand>
```
weshaggard: These are all defined in BuildTools now so they can be removed.

tests/src/Common/external/external.depproj
```diff
@@ -5,19 +5,54 @@
     <!-- Given that xunit packages bring with them part of the framework, we need to specify a runtime in order to get the assets
          This RID value doesn't really matter, since the assets we are copying are not RID specific, so defaulting to Windows here
     -->
-    <ProjectJson>$(MSBuildThisFileDirectory)project.json</ProjectJson>
-    <ProjectLockJson>$(MSBuildThisFileDirectory)project.lock.json</ProjectLockJson>
+    <RestoreOutputPath>$(MSBuildThisFileDirectory)obj</RestoreOutputPath>
```
weshaggard: It would be a good idea to put this under the output path instead of in the source tree.

tests/src/JIT/Directed/TypedReference/TypedReference.csproj
```diff
@@ -27,9 +27,10 @@
   <ItemGroup>
     <Compile Include="TypedReference.cs" />
   </ItemGroup>
-  <ItemGroup>
-    <None Include="$(JitPackagesConfigFileDirectory)empty\project.json" />
-  </ItemGroup>
+  <PropertyGroup>
+    <TargetFramework>netcoreapp1.1</TargetFramework>
+    <RuntimeIdentifiers>win7-x86;win7-x64;ubuntu.14.04-x64;ubuntu.16.04-x64;ubuntu.16.10-x64;osx.10.12-x64;centos.7-x64;rhel.7-x64;debian.8-x64;fedora.24-x64</RuntimeIdentifiers>
```
weshaggard: As part of the new project system these can all be parameterized now so instead of providing a huge list like this you can pass in exactly the one you want. That helps eliminate the need for keeping these updated as well as it cuts down on unnecessary restores of all RIDs on all platforms. This can be a clean-up task but I thought I would call it out.

tests/src/Common/test_dependencies/test_dependencies.csproj
```diff
@@ -6,8 +6,37 @@
...
     <CLRTestKind>BuildOnly</CLRTestKind>
     <NugetTargetMoniker>.NETCoreApp,Version=v2.0</NugetTargetMoniker>
+    <NugetTargetMonikerShort>netcoreapp2.0</NugetTargetMonikerShort>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETCore.CoreDisTools">
+      <Version>1.0.1-prerelease-*</Version>
+    </PackageReference>
+    <PackageReference Include="Microsoft.NETCore.Platforms">
+      <Version>4.4.0-$(CoreFxExpectedPrerelease)</Version>
```
weshaggard: This version doesn't exist Microsoft.NETCore.Platforms is version 2.0.0 which is at least on of the current build failures. You can actually remove this reference completely as it will be pulled in transitively from Microsoft.Private.CoreFx.NETCoreApp.

weshaggard:
I called out the bad version in your PR which is causing the current breaks.

weshaggard:
@wtgodbe to help push this along I fixed the current build issue and merged away the conflicts.

wtgodbe:
> baseservices.compilerservices.XUnitWrapper.cs(137,33): error CS0012: The type 'Object' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e'. [D:\j\workspace\debug_windows---17180f1b\bin\tests\Windows_NT.x64.Debug\TestWrappers\baseservices.compilerservices\baseservices.compilerservices.XUnitWrapper.csproj]

@weshaggard does this have something to do with your last change? This is in the generated xunit wrapper C# code. That C# file has its .csproj generated here - https://github.com/wtgodbe/coreclr/blob/BuildToolsUpdate/tests/runtest.proj#L69. Should I make a similar addition there?

weshaggard:
>@weshaggard does this have something to do with your last change? This is in the generated xunit wrapper C# code. That C# file has its .csproj generated here - https://github.com/wtgodbe/coreclr/blob/BuildToolsUpdate/tests/runtest.proj#L69. Should I make a similar addition there?

It is a similar issue that we need to fix. I just fixed it and the restoring for the testwrapper config to handle net45.

wtgodbe:
@weshaggard looks like we have some test failures, namely in JIT/CheckProjects, and Loader/ClassLoader - https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_windows_nt_prtest/6912/console. There's also an xunit wrapper being created for Common/PerfHarness, which shouldn't be happening - I'm looking into that now. Also @DrewScoggins is looking into the perf test failures.

wtgodbe:
@gkhanna79 @weshaggard this is now updated/rebased to disable the 3 failing tests. Here is the perf test from the previous iteration - https://ci2.dot.net/job/dotnet_coreclr/job/perf/job/master/job/perf_perflab_Windows_NT_x86_smoketest_prtest/536/console

wtgodbe:
The perf job failed with:

> https://ci2.dot.net/job/dotnet_coreclr/job/perf/job/master/job/perf_perflab_Windows_NT_x86_smoketest_prtest/536/console

wtgodbe:
I've pushed a potential fix, testing it locally in parallel

PR #11623, opened 2017-05-15
Removed unused local variable.

bbartels:
I checked whether it was optimised away by Roslyn, but it doesn't seem that way.

MSIL:

[...]
ldlen        
conv.i4      
stloc.0      // taskCount
[...]
ldloc.0      // taskCount
[...]

### DIFF
--- a/src/mscorlib/src/System/Threading/Tasks/Task.cs
+++ b/src/mscorlib/src/System/Threading/Tasks/Task.cs
@@ -6011,8 +6011,7 @@ namespace System.Threading.Tasks
 
             // Make a defensive copy, as the user may manipulate the tasks array
             // after we return but before the WhenAny asynchronously completes.
-            int taskCount = tasks.Length;
-            Task[] tasksCopy = new Task[taskCount];
+            Task[] tasksCopy = new Task[tasks.Length];
             for (int i = 0; i < taskCount; i++)
             {
                 Task task = tasks[i];


src/mscorlib/src/System/Threading/Tasks/Task.cs
```diff
@@ -6011,8 +6011,7 @@ public static Task<Task> WhenAny(params Task[] tasks)
 
             // Make a defensive copy, as the user may manipulate the tasks array
             // after we return but before the WhenAny asynchronously completes.
-            int taskCount = tasks.Length;
-            Task[] tasksCopy = new Task[taskCount];
+            Task[] tasksCopy = new Task[tasks.Length];
             for (int i = 0; i < taskCount; i++)
```
jkotas: Did you meant to replace `taskCount` with `tasks.Length` here as well?

stephentoub: This will fail to compile.  You've removed the `taskCount` variable this line depends on.

jkotas:
> I checked whether it was optimised away by Roslyn

It does not matter as much what IL is generated by Roslyn. What really matters is the native code produced by RyuJIT at the end.

src/mscorlib/src/System/Threading/Tasks/Task.cs
bbartels: Oh what, how did I miss that.

bbartels: Yeah compiled it by changing that aswell, only to realize now that would do more harm in the end. My bad!

PR #11629, opened 2017-05-16
Optimized Math.Sign

pentp:
This makes `Math.Sign(decimal)` about 10x faster and `Math.Sign(sbyte/short/int/long)` about twice as fast.

### DIFF
--- a/src/mscorlib/src/System/Decimal.cs
+++ b/src/mscorlib/src/System/Decimal.cs
@@ -803,6 +803,8 @@ namespace System
         [MethodImplAttribute(MethodImplOptions.InternalCall)]
         private static extern void FCallRound(ref Decimal d, int decimals);
 
+        internal static int Sign(ref decimal d) => (d.lo | d.mid | d.hi) == 0 ? 0 : (d.flags >> 31) | 1;
+
         // Subtracts two Decimal values.
         //
         public static Decimal Subtract(Decimal d1, Decimal d2)
--- a/src/mscorlib/src/System/Math.cs
+++ b/src/mscorlib/src/System/Math.cs
@@ -660,50 +660,17 @@ namespace System
         // Sign function for VB.  Returns -1, 0, or 1 if the sign of the number
         // is negative, 0, or positive.  Throws for floating point NaN's.
         [CLSCompliant(false)]
-        public static int Sign(sbyte value)
-        {
-            if (value < 0)
-                return -1;
-            else if (value > 0)
-                return 1;
-            else
-                return 0;
-        }
-
+        public static int Sign(sbyte value) => Sign((int)value);
 
         // Sign function for VB.  Returns -1, 0, or 1 if the sign of the number
         // is negative, 0, or positive.  Throws for floating point NaN's.
-        public static int Sign(short value)
-        {
-            if (value < 0)
-                return -1;
-            else if (value > 0)
-                return 1;
-            else
-                return 0;
-        }
+        public static int Sign(short value) => Sign((int)value);
 
         // Sign function for VB.  Returns -1, 0, or 1 if the sign of the number
         // is negative, 0, or positive.  Throws for floating point NaN's.
-        public static int Sign(int value)
-        {
-            if (value < 0)
-                return -1;
-            else if (value > 0)
-                return 1;
-            else
-                return 0;
-        }
+        public static int Sign(int value) => unchecked(value >> 31 | (int)((uint)-value >> 31));
 
-        public static int Sign(long value)
-        {
-            if (value < 0)
-                return -1;
-            else if (value > 0)
-                return 1;
-            else
-                return 0;
-        }
+        public static int Sign(long value) => unchecked((int)(value >> 63 | (long)((ulong)-value >> 63)));
 
         public static int Sign(float value)
         {
@@ -727,15 +694,7 @@ namespace System
             throw new ArithmeticException(SR.Arithmetic_NaN);
         }
 
-        public static int Sign(Decimal value)
-        {
-            if (value < 0)
-                return -1;
-            else if (value > 0)
-                return 1;
-            else
-                return 0;
-        }
+        public static int Sign(decimal value) => decimal.Sign(ref value);
 
         public static long BigMul(int a, int b)
         {


mikedn:
Cool, I was thinking of doing this but it never happened.

> Math.Sign(sbyte/short/int/long) about twice as fast

For what kind of input?

omariom:
Can you guys do something with [int.CompareTo](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Int32.cs#L57)?
It is  very branchy as well :)

mikedn:
> Can you guys do something with int.CompareTo?

Funny, I'm actually experimenting with some JIT stuff that would improve that. Though currently it removes only one of the branches, removing both is a bit more complicated.

pentp:
I ran perf tests with a random sampling of numbers skewed towards smaller and positive numbers (and less decimal places). For comparison with the best-case data for the current implementation I added tests with all values set to -1.

For `sbyte` and `int` the current implementation is 25% faster for the best case (all values -1). The difference in `short`/`long` behaviour is mysterious because the assembly code is almost identical for all four methods (only difference is in the `movsx` instruction and data size, but it fits in L1... any way to inspect it w/o VTune?).
[Test code](https://gist.github.com/pentp/d9fd61a6875871822a875f02d4372d20).
``` ini

BenchmarkDotNet=v0.10.5, OS=Windows 10.0.15063
Processor=Intel Core i7-7700K CPU 4.20GHz (Kaby Lake), ProcessorCount=8
Frequency=4101567 Hz, Resolution=243.8093 ns, Timer=UNKNOWN
dotnet cli version=2.0.0-preview1-005660
  [Host] : .NET Core 4.6.25409.0, 64bit RyuJIT

Toolchain=InProcessToolchain  
```
#### Reference (all -1)
 |      Method |       Mean |      Error |     StdDev |
 |------------ |-----------:|-----------:|-----------:|
 | DecimalSign | 6,593.5 ns | 11.0956 ns |  8.6627 ns |
 |   SByteSign |   466.2 ns |  0.2550 ns |  0.2385 ns |
 |   ShortSign |   956.5 ns | 18.9021 ns | 42.6652 ns |
 |     IntSign |   468.3 ns |  0.1651 ns |  0.1544 ns |
 |    LongSign |   815.1 ns | 16.2364 ns | 41.6202 ns |
#### Reference (skewed random numbers)
 |      Method |        Mean |     Error |    StdDev |
 |------------ |------------:|----------:|----------:|
 | DecimalSign | 12,550.2 ns | 17.847 ns | 16.694 ns |
 |   SByteSign |  1,120.3 ns |  4.749 ns |  3.707 ns |
 |   ShortSign |  1,363.3 ns |  2.541 ns |  2.252 ns |
 |     IntSign |    999.3 ns | 10.126 ns |  8.455 ns |
 |    LongSign |    920.0 ns | 18.781 ns | 21.628 ns |
#### Optimized (all -1)
 |      Method |       Mean |     Error |    StdDev |
 |------------ |-----------:|----------:|----------:|
 | DecimalSign | 1,031.2 ns | 1.1569 ns | 0.9661 ns |
 |   SByteSign |   613.9 ns | 0.1033 ns | 0.0862 ns |
 |   ShortSign |   613.0 ns | 0.1186 ns | 0.1051 ns |
 |     IntSign |   613.0 ns | 0.2206 ns | 0.2064 ns |
 |    LongSign |   613.5 ns | 0.5879 ns | 0.4909 ns |
#### Optimized (skewed random numbers)
 |      Method |       Mean |      Error |    StdDev |
 |------------ |-----------:|-----------:|----------:|
 | DecimalSign | 1,354.9 ns | 10.3288 ns | 9.6615 ns |
 |   SByteSign |   614.0 ns |  0.0917 ns | 0.0813 ns |
 |   ShortSign |   613.3 ns |  0.1607 ns | 0.1424 ns |
 |     IntSign |   613.0 ns |  0.1833 ns | 0.1715 ns |
 |    LongSign |   613.7 ns |  0.6643 ns | 0.6214 ns |

mikedn:
> The difference in short/long behaviour is mysterious because the assembly code is almost identical for all four methods (only difference is in the movsx instruction and data size, but it fits in L1... any way to inspect it w/o VTune?).

When I run your tests I don't see those differences, all types require about the same time for both implementations. 

Cache doesn't sound like it should be an issue, even if the data doesn't fit in L1 the time should increase with type size but it doesn't, in "Reference (all -1)" you have fast `sbyte` and `int` and slow `short` and `long`.  And the effects of the cache should appear in the optimized case as well but they don't.

It would be interesting to see results from other CPUs. I'll try running these tests on Surface Pro 4 (Skylake CPU) later today.

As expected the new code is actually slower when the input is always negative:
``` ini

BenchmarkDotNet=v0.10.6, OS=Windows 10 Redstone 2 (10.0.15063)
Processor=Intel Core i5-4440 CPU 3.10GHz (Haswell), ProcessorCount=4
Frequency=3026471 Hz, Resolution=330.4178 ns, Timer=TSC
dotnet cli version=1.0.3
  [Host] : .NET Core 4.6.25009.03, 64bit RyuJIT

Toolchain=InProcessToolchain  

```
 |      Method |     Mean |     Error |    StdDev |
 |------------ |---------:|----------:|----------:|
 |     IntSign | 660.0 ns |  4.009 ns |  3.750 ns |
 | IntSignFast | 873.1 ns | 11.795 ns | 10.456 ns |

But it's faster when the input is always positive:

 |      Method |       Mean |    Error |   StdDev |
 |------------ |-----------:|---------:|---------:|
 |     IntSign | 1,057.0 ns | 8.606 ns | 8.050 ns |
 | IntSignFast |   873.4 ns | 3.736 ns | 3.120 ns |

so I think that this being a bit slower in the "all negative" case is not an issue.

And the fact that the optimized version requires the same time for any input is IMO preferable even if it one case it's a bit slower.

pentp:
After many different attempts to nail down the mysterious differences I finally found a repro (sort of).
When I run the following code, SByteSign is always slower and varies from 480ms to 700ms, ShortSign is always ~466ms:
```
for(int i = 0;i < 1000000;i++) t.SByteSign();
for(int i = 0;i < 1000000;i++) t.ShortSign();
```
However if I add a call to `GC.KeepAlive(null)` before the loops (or after, or anywhere really) they both run consistently ~466ms. The only difference in the codegen is `xor rcx, rcx; call GC:KeepAlive(ref)`.

Either my CPU is playing tricks on me or I'm starting to go insane...

mikedn:
Well, I managed to get to run your original benchmark on another CPU and I got this:

 |      Method |      Mean |     Error |    StdDev |
 |------------ |----------:|----------:|----------:|
 | DecimalSign | 20.660 us | 0.4114 us | 0.7626 us |
 |   SByteSign |  1.684 us | 0.0331 us | 0.0382 us |
 |   ShortSign |  1.760 us | 0.0344 us | 0.0448 us |
 |     IntSign |  1.826 us | 0.0360 us | 0.0401 us |
 |    LongSign |  1.783 us | 0.0351 us | 0.0556 us |

SByteSign is a bit faster but not by much, practically they're all the same.

> However if I add a call to GC.KeepAlive(null) before the loops (or after, or anywhere really) they both run consistently ~466ms. The only difference in the codegen is xor rcx, rcx; call GC:KeepAlive(ref).

Adding code before a loop can influence the loop code alignment and thus change the results. Adding code after shouldn't have any effect but you never know...

> Either my CPU is playing tricks on me or I'm starting to go insane

Eh, such microbenchmarks need to be taken with a grain of salt. They can be useful to validate a theory, in this case that eliminating branches makes code faster when the input isn't constant. They're not so useful for creating a theory. For example, you should not conclude from your results that casting from `sbyte` to `int` is somehow slower than casting from `short` to `int`.

pentp:
The one failing test used to be green, but now its results are missing. Is this ready for merging and I can move on to some other `decimal` methods that are suboptimal (`Abs`, `Ceiling`, `Negate`, `Remainder`)?

mikedn:
> The one failing test used to be green, but now its results are missing. 

Yeah, that was triggered by the conversation and failed. Work in progress I'd guess, I've seen those perf jobs failing before.

pentp:
Any feedback on this? Is this change too small? Are the integer `Sign` changes not worth it?
I added `unchecked` around the casts for checked builds. Anything else needs changing before merge?

danmosemsft:
@pentp oops this got overlooked.
@stephentoub are you satisfied with the benchmark data here, can you review?
@tannergooding this is a perf not correctness change but fyi.

pentp:
@stephentoub Could you take a look at this?

stephentoub:
LGTM.  @mellinoe should review as well.

PR #11631, opened 2017-05-16
Remove an unnecessary binary file

swaroop-sridhar:
Remove a binary file that was unintentionally checked in.


swaroop-sridhar:
CC: @brianrob

PR #11666, opened 2017-05-17
[x86/Linux] Port 'DacUnwindStackFrame'

parjong:
This commit resolves undefined reference on DacUnwindStackFrame from SOS plugin.

### DIFF
--- a/src/unwinder/i386/unwinder_i386.cpp
+++ b/src/unwinder/i386/unwinder_i386.cpp
@@ -131,7 +131,14 @@ OOPStackUnwinderX86::VirtualUnwind(
 
 BOOL DacUnwindStackFrame(T_CONTEXT* pContextRecord, T_KNONVOLATILE_CONTEXT_POINTERS* pContextPointers)
 {
-    return OOPStackUnwinderX86::Unwind(pContextRecord, pContextPointers);
+    BOOL res = OOPStackUnwinderX86::Unwind(pContextRecord, NULL);
+
+    if (res && pContextPointers)
+    {
+        FillContextPointers(pContextPointers, pContextRecord);
+    }
+
+    return res;
 }
 
 //---------------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


parjong:

src/unwinder/i386/unwinder_i386.cpp
```diff
@@ -72,49 +115,25 @@ OOPStackUnwinderX86::VirtualUnwind(
...
     // For x86, the value of Establisher Frame Pointer is Caller SP
     //
     // (Please refers to CLR ABI for details)
     *EstablisherFrame = ContextRecord->Esp;
     return S_OK;
 }
 
+BOOL DacUnwindStackFrame(T_CONTEXT* pContextRecord, T_KNONVOLATILE_CONTEXT_POINTERS* pContextPointers)
+{
+    return OOPStackUnwinderX86::Unwind(pContextRecord, pContextPointers);
```
janvorli: I am not sure if we want the real context pointers for the DAC unwind. For all other architectures, we just point all the context pointers to point into the pContextRecord. I guess the reason is that if we did it the way you did it here, the context pointers would be target address space addresses while the debugger code expects them to be the debugger address space ones.
@mikem8361 am I right?

mikem8361: I'm pretty sure you are correct.  But let's check with @noahfalk.

parjong: @janvorli @mikem8361 Do you mean that ``DacUnwindStackFrame``` should take ``KNONVOLATILE_CONTEXT_POINTERS``  (instead of ``KNONVOLATILE_CONTEXT_POINTERS``)?

janvorli: No, the point was that it should do what all the other platforms do in this function. That means passing NULL as the context pointers argument to the OOPStackUnwinderX86::Unwindm and rather do the following or something similar to assign them:
```C++
    if (res && pContextPointers)
    {
        for (int i = 0; i < 7; i++)
        {
            *(&pContextPointers->Edi + i) = &pContext->Edi + i;
        }
    }
```

parjong:
@janvorli Updated as suggested. Please take a look. (I extracted similar code from FillRegDisplay and use it (to reduce code duplication), but please let me know if there is something wrong).

janvorli:
@parjong, the windows x86 build is broken by this change.

PR #11672, opened 2017-05-17
Use dedicated ifdef flag for WATSON-related code

parjong:


janvorli:
@parjong there are many more places in the coreclr runtime where there is watson related code that is ifdefed out for FEATURE_PAL. Text search shows about 59 files . So I think that we should either change all of them or none. Since watson is Windows specific, it seems to me that it is actually ok to use FEATURE_PAL to ifdef it out. It is very unlikely that we would ever implement that stuff for non-windows platforms and so the ifdef seems appropriate.

parjong:
@janvorli Thank you for comment. I submitted this PR as I felt that FEATURE_PAL is used too broadly.

I tried to clean up all of Watson-related FEATURE_PAL usage (as you suggested), but it turns out that there are many ambiguous cases. I conclude that it is currently impossible for me to clean up this (as I have not much knowledge on Watson itself).

Let's leave as it is.

PR #11684, opened 2017-05-17
Remove a couple mistaken calls to popd.

pgavlin:
These calls were causing individual test runs to fail if other
directories were on the stack.


pgavlin:
@RussKeldorph PTAL

PR #11693, opened 2017-05-17
More update in /dev/defaultintf

yizhang82:
* Support non-virtual calls on interface private members correctly
* Support protected methods
* Properly handle precode
* Tentatively throw exception when seeing conflict overrides and add a test case

(This updates CoreCLR dev/defaultintf the same with the build we are showing at //build)

### DIFF
--- a/src/vm/jitinterface.cpp
+++ b/src/vm/jitinterface.cpp
@@ -5243,8 +5243,7 @@ void CEEInfo::getCallInfo(
         directCall = true;
     }
     else
-    // Force all interface calls to be interpreted as if they are virtual.
-    if (pTargetMD->GetMethodTable()->IsInterface())
+    if (pTargetMD->GetMethodTable()->IsInterface() && pTargetMD->IsVirtual())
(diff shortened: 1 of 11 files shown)


yizhang82:
@dotnet-bot Test this please

yizhang82:
@davidwrighton Sorry I forgot to ask for a review. Can you take a look and comment on it? I'll address them in my next PR.

src/vm/jitinterface.cpp
```diff
@@ -5243,8 +5243,7 @@ void CEEInfo::getCallInfo(
         directCall = true;
     }
     else
-    // Force all interface calls to be interpreted as if they are virtual.
-    if (pTargetMD->GetMethodTable()->IsInterface())
+    if (pTargetMD->GetMethodTable()->IsInterface() && pTargetMD->IsVirtual())
```
davidwrighton: This may block the ability to have base<ISomeInterface>.Method() calls.

PR #11694, opened 2017-05-17
Don't hoist IConHandle statics above cctors

### LINKED ISSUE #11689: Jit incorrectly hoists non-`ClsVar` statics
The same issue that #11465 fixed for `ClsVar` statics can also happen for `IconHandle` statics.  Here's an example that incorrectly hoists the static access above the static init helper call and thus fails on Win/x64:

```c#

namespace N
{
    struct WrappedInt
    {
        public int Value;

        public static WrappedInt Twenty = new WrappedInt() { Value = 20 };
    }
    public static class C
    {
        static int unwrap(WrappedInt wi) => wi.Value;

        static int foo(int s, int n)
        {
            for(int i = 0; i < n; ++i)
            {
                s += unwrap(WrappedInt.Twenty);
            }

            return s;
        }

        {
            return foo(20, 4);
        }
    }
}
```

JosephTremoulet:
These constants appear loop invariant, hence so do their dereferences
when a loop has no memory side-effects.  Add flag `GTF_ICON_INITCLASS`
similar to `GTF_FLD_INITCLASS`/`GTF_CLS_VAR_INITCLASS`, and use it to
prevent hoisting such references without also hoisting their static
init helper calls.

Resolves #11689

### DIFF
--- a/src/jit/gentree.h
+++ b/src/jit/gentree.h
@@ -1003,6 +1003,14 @@ public:
 #define GTF_ICON_FIELD_OFF          0x08000000 // GT_CNS_INT -- constant is a field offset
 #define GTF_ICON_SIMD_COUNT         0x04000000 // GT_CNS_INT -- constant is Vector<T>.Count
 
+#define GTF_ICON_INITCLASS          0x02000000 // GT_CNS_INT -- Constant is used to access a static that requires preceding
+                                               //               class/static init helper.  In some cases, the constant is
+                                               //               the address of the static field itself, and in other cases
+                                               //               there's an extra layer of indirection and it is the address
+                                               //               of the cell that the runtime will fill in with the address
+                                               //               of the static field; in both of those cases, the constant
+                                               //               is what gets flagged.
+
 #define GTF_BLK_VOLATILE            GTF_IND_VOLATILE  // GT_ASG, GT_STORE_BLK, GT_STORE_OBJ, GT_STORE_DYNBLK -- is a volatile block operation
 #define GTF_BLK_UNALIGNED           GTF_IND_UNALIGNED // GT_ASG, GT_STORE_BLK, GT_STORE_OBJ, GT_STORE_DYNBLK -- is an unaligned block operation
 
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -6573,6 +6573,13 @@ GenTreePtr Compiler::fgMorphField(GenTreePtr tree, MorphAddrContext* mac)
 
             GenTreePtr tlsRef = gtNewIconHandleNode(WIN32_TLS_SLOTS, GTF_ICON_TLS_HDL);
 
+            // Translate GTF_FLD_INITCLASS to GTF_ICON_INITCLASS
+            if ((tree->gtFlags & GTF_FLD_INITCLASS) != 0)
+            {
+                tree->gtFlags &= ~GTF_FLD_INITCLASS;
+                tlsRef->gtFlags |= GTF_ICON_INITCLASS;
+            }
+
             tlsRef = gtNewOperNode(GT_IND, TYP_I_IMPL, tlsRef);
 
             if (dllRef != nullptr)
@@ -6627,6 +6634,12 @@ GenTreePtr Compiler::fgMorphField(GenTreePtr tree, MorphAddrContext* mac)
                     FieldSeqNode* fieldSeq =
                         fieldMayOverlap ? FieldSeqStore::NotAField() : GetFieldSeqStore()->CreateSingleton(symHnd);
                     addr->gtIntCon.gtFieldSeq = fieldSeq;
+                    // Translate GTF_FLD_INITCLASS to GTF_ICON_INITCLASS
+                    if ((tree->gtFlags & GTF_FLD_INITCLASS) != 0)
+                    {
+                        tree->gtFlags &= ~GTF_FLD_INITCLASS;
+                        addr->gtFlags |= GTF_ICON_INITCLASS;
+                    }
 
                     tree->SetOper(GT_IND);
                     // The GTF_FLD_NULLCHECK is the same bit as GTF_IND_ARR_LEN.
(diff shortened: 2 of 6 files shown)


JosephTremoulet:
@briansull PTAL
/cc @dotnet/jit-contrib 

Still running diffs...

src/jit/morph.cpp
```diff
@@ -6560,6 +6560,13 @@ GenTreePtr Compiler::fgMorphField(GenTreePtr tree, MorphAddrContext* mac)
 
             GenTreePtr tlsRef = gtNewIconHandleNode(WIN32_TLS_SLOTS, GTF_ICON_TLS_HDL);
 
+            // Translate GTF_FLD_INITCLASS to GTF_ICON_INITCLASS
```
briansull: Also line 6468 contains a dangerous SetOper call which can cause the 
flag  GTF_FLD_INITCLASS to be interpreted as GTF_IND_REFARR_LAYOUT

src/jit/optimizer.cpp
```diff
@@ -6142,9 +6142,15 @@ bool Compiler::optHoistLoopExprsForTree(GenTreePtr        tree,
...
 
-    // Initclass CLS_VARs are the base case of cctor dependent trees.
-    bool treeIsCctorDependent = (tree->OperIs(GT_CLS_VAR) && ((tree->gtFlags & GTF_CLS_VAR_INITCLASS) != 0));
-    bool treeIsInvariant      = true;
+    // Initclass CLS_VARs and IconHandles are the base cases of cctor dependent trees.
+    // In the IconHandle case, it's of course the dereference, rather than the constant itself, that is
+    // truly dependent on the cctor.  So a more precise approach would be to separately propagate
+    // isCctorDependent and isAddressWhoseDereferenceWouldBeCctorDependent, but we don't for simplicity/throughput;
+    // the constant itself would be considered non-hoistable anyway, since optIsCSEcandidate returns
+    // false for constants.
```
briansull: `since optIsCSEcandidate returns false for constants.`

This might not continue to be true in the future.
Is it possible to flag the indirection instead of the constant?
Or are we out of flag bits on GT_IND?

src/jit/morph.cpp
JosephTremoulet: 6468 is dealing with instance fields (it's under the `if (objRef)` on line 6239 with `else` on line 6500)

src/jit/optimizer.cpp
JosephTremoulet: Yes, GT_IND is already dipping into the common bits:

```c++
#define GTF_IND_NONFAULTING   0x00000800 // An indir that cannot fault.  GTF_SET_FLAGS is not used on indirs
```

We could (maybe?) similarly "steal" `GTF_ZSF_SET` or `GTF_UNSIGNED` for a new `GT_IND` flag, though I think the extra conservativism may be warranted here... with this change as-is, we'll block hoisting static field addresses produced by `ldsflda` (if the field is marked `CORINFO_FLG_FIELD_INITCLASS` and we aren't hoisting the cctor call with it).  The importer does the marking now, but the importer doesn't have the context to know when a `GT_IND` might be fed by a prior `ldsflda`, so it seems risky to leave the `ldsflda` addresses unmarked.  Using two bools here to track cctor-dependence would let us keep the annotations on the address and so avoid the flag sharing and the `ldsflda`+indir problem, I just didn't think it was worth it with no/little current benefit.  I could be persuaded to change it though, what do you think?

JosephTremoulet:
jit-diff reports no diffs, but crossgen takes different codepaths for statics than jitting; will check spmi diffs when back online

src/jit/optimizer.cpp
briansull: I'm comfortable with this fix for now and don't want to complicate this right now by fishing for extra flag bits for the GT_IND node.

src/jit/morph.cpp
briansull: OK

src/jit/gentree.h
```diff
@@ -1007,6 +1007,14 @@ struct GenTree
...
 #define GTF_ICON_SIMD_COUNT 0x04000000 // GT_CNS_INT -- constant is Vector<T>.Count
 
+#define GTF_ICON_INITCLASS 0x02000000 // GT_CNS_INT -- Constant is used to access a static that requires preceding
+                                      //               class/static init helper.  In some cases, the constant is
+                                      //               the address of the static field itself, and in other cases
+                                      //               there's an extra layer of indirection and it is the address
+                                      //               of the cell that the runtime will fill in with the address
+                                      //               of the static field; in both of those cases, the constant
+                                      //               is what gets flagged.
+
```
briansull: This is likely to conflict with Bruce's GT Flags reformatting chnage

JosephTremoulet: Yeah, I noticed that.  I'll rebase/resolve if he merges first; I'd like to run desktop testing on this before merging.

JosephTremoulet:
@dotnet-bot test Windows_NT x64 CoreCLR Perf Tests Correctness test Windows_NT x86 CoreCLR Perf Tests Correctness

JosephTremoulet:

JosephTremoulet:
Desktop testing showed that this code had an off-by-one error that was artificially blocking the hoisting of cctor calls in some cases.  I've pushed [a fix for that](https://github.com/dotnet/coreclr/pull/11694/commits/357318f3e97df53d5c0ffc51598096a7fc0f5662) in this PR; @briansull PTAL.

With that, SPMI reports diffs in just 34 unique methods.  The ones I looked into were all attributable to an existing issue making the static var hoisting a bit more conservative than it needs to be, for which I've filed #11824.

So I think this is ready to merge (and likely port to 2.0/4.7.1) if the last change looks agreeable.

briansull:
Go ahead and check this in

hqueue:
Thanks! 
Just for the record, old issue #10780 is related to this.

PR #11698, opened 2017-05-17
LSRA throughput tweaks

BruceForstall:
Change LSRA to create Intervals only for tracked variables

Change getIntervalForLocalVar() to take a tracked variable index, not varNum.

Convert LinearScan::getVarReg() to take a tracked var index, not varNum.

This reduces instruction count of superpmi replay over minopts tests by 0.6%.

### DIFF
--- a/src/jit/lsra.cpp
+++ b/src/jit/lsra.cpp
@@ -2049,11 +2049,12 @@ void LinearScan::identifyCandidates()
 
         if (varDsc->lvTracked)
         {
-            // Create an interval for all tracked variables, even if the tracked variable is not
-            // a register candidate. This is probably just done for consistency (e.g., code that
-            // simply walks over all tracked variables don't need to check if they are register
-            // candidates, or if the interval exists). It might be possible to only allocate
-            // intervals for register candidate tracked variables.
+            // Create an interval for all tracked variables, even if the tracked variable
+            // is not a register candidate. This is probably just done for consistency
+            // (e.g., code that simply walks over all tracked variables don't need to check if
+            // they are register candidates, or if the interval exists). It might be possible to
+            // only allocate intervals for register candidate tracked variables, if all the code
+            // that depends on every tracked variable having a non-null Interval was fixed.
 
             newInt = newInterval((var_types)varDsc->lvType);
             newInt->setLocalNumber(compiler, lclNum, this);
@@ -3718,8 +3719,7 @@ void LinearScan::buildRefPositionsForNode(GenTree*                  tree,
 
             if ((tree->gtFlags & GTF_VAR_DEATH) == 0)
             {
-                VarSetOps::AddElemD(compiler, currentLiveVars,
-                                    compiler->lvaTable[tree->gtLclVarCommon.gtLclNum].lvVarIndex);
+                VarSetOps::AddElemD(compiler, currentLiveVars, varIndex);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


BruceForstall:
@CarolEidt PTAL
cc @dotnet/jit-contrib

src/jit/lsra.cpp
```diff
@@ -3529,7 +3543,10 @@ void LinearScan::buildRefPositionsForNode(GenTree*                  tree,
         // is processed, unless this is marked "isLocalDefUse" because it is a stack-based argument
         // to a call
 
-        Interval* interval        = getIntervalForLocalVar(tree->gtLclVarCommon.gtLclNum);
+        LclVarDsc* varDsc = &compiler->lvaTable[tree->gtLclVarCommon.gtLclNum];
+        assert(varDsc->lvTracked);
+        unsigned  varIndex        = varDsc->lvVarIndex;
```
CarolEidt: It would be nice to have a method on `Compiler` to do this.

src/jit/lsra.cpp
```diff
@@ -2036,14 +2037,27 @@ void LinearScan::identifyCandidates()
...
 
-        newInt->setLocalNumber(lclNum, this);
+        if (varDsc->lvTracked)
+        {
+            // Create an interval for all tracked variables, even if the tracked variable is not
+            // a register candidate. This is probably just done for consistency (e.g., code that
+            // simply walks over all tracked variables don't need to check if they are register
+            // candidates, or if the interval exists). It might be possible to only allocate
+            // intervals for register candidate tracked variables.
+
```
CarolEidt: Perhaps this should be a TODO-Throughput and/or an issue created for it?

src/jit/lsra.cpp
```diff
@@ -3644,9 +3660,12 @@ void LinearScan::buildRefPositionsForNode(GenTree*                  tree,
...
         {
             // We always push the tracked lclVar intervals
-            varDefInterval = getIntervalForLocalVar(tree->gtLclVarCommon.gtLclNum);
-            defRefType     = refTypeForLocalRefNode(tree);
-            defNode        = tree;
+            LclVarDsc* varDsc = &compiler->lvaTable[tree->gtLclVarCommon.gtLclNum];
+            assert(varDsc->lvTracked);
+            unsigned varIndex = varDsc->lvVarIndex;
+            varDefInterval    = getIntervalForLocalVar(varIndex);
+            defRefType        = refTypeForLocalRefNode(tree);
```
CarolEidt: I can see that one couldn't overload `getIntervalForLocalVar` to have a varNum variant, and even with a different name it could be confusing. But something like a `getIntervalForLocalVarNode` would be useful here and above.

src/jit/lsra.cpp
pgavlin: +1. I think that this guess is correct:
> It might be possible to only allocate intervals for register candidate tracked variables
which is simply to say that the entries in the `localVarIntervals` table for non-candidate tracked vars would be null.

BruceForstall: I should point out that I did this, and got crashes, so they key is those crashes would need to be fixed.

BruceForstall: Created #11707

src/jit/lsra.cpp
BruceForstall: It turns out varIndex is used elsewhere in both cases here, but there was another case where it wasn't, so I created `getIntervalForLocalVarNode` as suggested.

BruceForstall:

PR #11099, opened 2017-04-20
[x86/Linux] Fix pow of tryrun

seanshpark:
x86/Linux has compatible math pow()
This will fix CoreFX System.Tests.MathFTests.Pow() failures

PR #11703, opened 2017-05-18
Enable pow for arm/armel in tryrun

seanshpark:
This will fix CoreFX System.Tests.MathFTests.Pow() failures


seanshpark:
Same as in #11099

seanshpark:
@janvorli , Thank you.
And... could you please check `CoreCLR Perf Tests Correctness` failure? I've some difficulties accessing the link in `Details` maybe cause of our firewall.

janvorli:
@seanshpark yes, these tests are failing everywhere now.

PR #11709, opened 2017-05-18
[ARM32] Simplify getting aligment value when remorphing

hseok-oh:
Simplify getting aligment value in morphing phase when remorphing by using argument table.
It can be useful to get alignment value for struct morphed to FIELD_LIST in ARM32/RyuJIT.

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -3207,10 +3207,17 @@ GenTreeCall* Compiler::fgMorphArgs(GenTreeCall* call)
         }
 
         // We don't use the "size" return value from InferOpSizeAlign().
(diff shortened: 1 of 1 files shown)


hseok-oh:
cc/ @dotnet/jit-contrib

src/jit/morph.cpp
```diff
@@ -3207,10 +3207,17 @@ GenTreeCall* Compiler::fgMorphArgs(GenTreeCall* call)
         }
 
         // We don't use the "size" return value from InferOpSizeAlign().
```
RussKeldorph: Seems like this comment ought to remain next to the call to InferOpSizeAlign.

hseok-oh: Thank you. I fixed it.

PR #11711, opened 2017-05-18
pal/configure.cmake: add missing HAVE_SCHED_AFFINITY

tmds:
This is used by numa.cpp

CC @janvorli


janvorli:
@tmds thank you a lot for spotting that. I am sure that when I was testing the NUMA stuff before merging, I was stepping through the affinity related. I guess must have forgotten to git add that change.

Issue #11533, opened 2017-05-11
[Arm64] NYI lvaArgType


jashook:
Affected Tests

<pre>
C:\Users\DDJOIN\arm64_ci\tests\tests\JIT\Performance\CodeQuality\Roslyn\CscBench\CscBench.cmd 
C:\Users\DDJOIN\arm64_ci\tests\tests\managed\Compilation\Compilation\Compilation.cmd 
</pre>

Example Output

<pre>
15:32:24 TestID: CscBench.cmd_8026
15:32:24 Path  : JIT/Performance/CodeQuality/Roslyn/CscBench/CscBench.cmd
15:32:24 
15:32:24 Started: Wed May 10 15:17:09 2017
15:32:24 Started: Wed May 10 15:18:02 2017
15:32:24 Output: TEST_CMD_LINE: JIT\Performance\CodeQuality\Roslyn\CscBench> C:\Users\DDJOIN\arm64_ci\tests\tests\JIT\Performance\CodeQuality\Roslyn\CscBench\CscBench.cmd 
15:32:24 BEGIN EXECUTION
15:32:24  "C:\Users\DDJOIN\arm64_ci\tests\tests\Tests\Core_Root\corerun.exe" CscBench.exe 
15:32:24 
Assert failure(PID 3924 [0x00000f54], Thread: 2612 [0x0a34]): Assertion failed 'NYI: lvaArgType' in 'System.Threading.Tasks.Parallel:ForEachWorker(ref,ref,ref,ref,ref,ref,ref,ref,ref):struct' (IL size 262)
15:32:24     File: d:\j\workspace\arm64_cross_c---f41ee988\src\jit\lclvars.cpp Line: 3447
15:32:24     Image: C:\Users\DDJOIN\arm64_ci\tests\tests\Tests\Core_Root\CoreRun.exe
15:32:24 Expected: 100
15:32:24 Actual: 123456789
15:32:24 END EXECUTION - FAILED
15:32:24 FAILED
</pre>

jashook:
Narrowed to only hit with TYP_STRUCT with #11718. Currently this case will not happen because we copy struct args to fastTailCall.

Issue #11717, opened 2017-05-18
[amd64 Unix|arm64] fastTailCall stack fixup NYI if TYP_STRUCT 


jashook:
The Issue:

Lowering::LowerFastTailCall will fix up the stack before the jump. In the case that the caller's stack frame has a TYP_STRUCT which needs to be overwritten we would create a temp struct and pass it currently according to the windows ABI. Currently we will not execute this code path because we make several choices before lowering the tail call that keep this from being true. First, we will force create a temp object for structs passed on the stack, so we will simply use the def of the temp we made during morph instead of creating a temp later. Second, we will decide not to fast tail call on amd64 unix and arm64 much more liberally than on amd64 Windows see #2666 for more information.

The closest I could come to a repro is:

<pre>
using System;

public struct A
{
    public int a;
    public int b;
}

class TailCallStructPassing
{
    public static int bar(int count, A temp)
    {
        if (count < 100)
        {
            return count;
        }

        else
        {
            count -= 100;
            return bar(count, temp);
        }
    }

    /// Args
    ///
    /// (0) - A one
    /// (1) - A two
    /// (2) - A three
    /// (3) - A four
    /// (4) - A five
    /// (5) - A six
    /// (6) - int b
    /// (7) - int c
    /// (8) - int d
    /// (9) - int e
    /// (10) - int f
    /// (11) - int g
    /// (12) - int h
    /// (13) - int i
    /// (14) - int j
    /// (15) - A temp
    /// (16) - int k
    /// (17) - int l
    /// (18) - int m
    /// (19) - int n
    /// (20) - int o
    /// (21) - int p
    /// (22) - int q
    /// (23) - int r
    /// (24) - int s
    /// (25) - int t
    /// (26) - int u
    /// (27) - int decision
    /// (28) - int count
    /// (29) - A temp2
    /// (30) - A first
    /// (31) - int
    public static int foo(A one, A two, A three, A four, A five, A six, int b, int c, int d, int e, int f, int g, int h, int i, int j, A temp, int k, int l, int m, int n, int o, int p, int q, int r, int s, int t, int u, int decision, int count, A temp2, A first, int copy)
    {
        if (decision < 100)
        {
            return foo(six, five, four, three, two, one, b, c, u, t, s, r, q, p, o, temp, n, m, l, k, j, i, h, g, f, e, d, 500, 15, temp2, first, copy);
        }

        else
        {
            return bar(count, temp);
        }
    }

    ///
    /// Args
[...]

PR #11718, opened 2017-05-18
[*64] NYI: TYP_STRUCT, fastTailCall stack fixup

jashook:
This will add an NYI for AMD64 and Arm64 during fastTailCall stack fixup which requires
a temp. Currently this code path is very difficult to execute for the case of TYP_STRUCT
because we will have already created copies of structs that are passed on the stack
during morph. The NYI will avoid incorrect behavior in the case that we make changes in
earlier stages that will force this code path to hit.

### DIFF
--- a/src/jit/lclvars.cpp
+++ b/src/jit/lclvars.cpp
@@ -3406,6 +3406,12 @@ var_types LclVarDsc::lvaArgType()
     var_types type = TypeGet();
 
 #ifdef _TARGET_AMD64_
+#ifdef FEATURE_UNIX_AMD64_STRUCT_PASSING
+    if (type == TYP_STRUCT)
+    {
+        NYI("lvaArgType");
+    }
+#else  //! FEATURE_UNIX_AMD64_STRUCT_PASSING
     if (type == TYP_STRUCT)
     {
         switch (lvExactSize)
@@ -3443,6 +3449,12 @@ var_types LclVarDsc::lvaArgType()
                 type = TYP_BYREF;
                 break;
         }
+    }
+#endif // !FEATURE_UNIX_AMD64_STRUCT_PASSING
+#elif defined(_TARGET_ARM64_)
+    if (type == TYP_STRUCT)
+    {
+        NYI("lvaArgType");
     }
 #elif defined(_TARGET_X86_)
 // Nothing to do; use the type as is.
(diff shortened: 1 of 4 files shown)


jashook:
Note the test case will force a fixup of the stack while creating a temp variable for the ints passed in reversed order.

jashook:

jashook:
Fixed #11533.

jashook:
PTAL @RussKeldorph @sdmaclea 
/cc @dotnet/jit-contrib

jashook:
See #11717 for more information.

jashook:

jashook:

src/jit/lclvars.cpp
```diff
@@ -3444,6 +3445,17 @@ var_types LclVarDsc::lvaArgType()
                 break;
         }
     }
+#else  // FEATURE_UNIX_AMD64_STRUCT_PASSING
+    if (type == TYP_STRUCT)
+    {
+        NYI("lvaArgType");
+    }
+#endif // !FEATURE_UNIX_AMD64_STRUCT_PASSING
+#elif defined(_TARGET_ARM64_)
```
BruceForstall: Two nits:
1. It is generally preferable when an `#if` has an `#else` to put the "positive" (non-negated) condition first, to reduce potential confusion resulting from reasoning about double negatives (and potentially missing the "n" in "ifndef")
2. Coding convention states the comment on a `#else` should indicate the condition of the code that follows the `#else`, whereas the comment on the `#endif` indicates the condition of the code that precedes the `#endif`. Thus:
```
#ifdef FEATURE_UNIX_AMD64_STRUCT_PASSING
...
#else  // !FEATURE_UNIX_AMD64_STRUCT_PASSING
...
#endif // !FEATURE_UNIX_AMD64_STRUCT_PASSING
```

src/jit/lclvars.cpp
```diff
@@ -3444,6 +3445,17 @@ var_types LclVarDsc::lvaArgType()
...
+    if (type == TYP_STRUCT)
+    {
+        NYI("lvaArgType");
+    }
 #elif defined(_TARGET_X86_)
```
CarolEidt: This string of `#ifdef`s begins with a negation, which can cause confusion. I think it would be better to put the other two cases first.

jashook:
Sounds good thank you for the feedback. Will address when the arm64 testing comes back.

jashook:

jashook:

jashook:
@pgavlin have you seen this error before?

jashook:

Issue #11263, opened 2017-04-27
Delete old version of bitset::ClearD.


sandreenko:
Replace it with the new version, that doesn't allocate new buffer when it is possible or with MakeEmpty(#11233).

Issue #11236, opened 2017-04-26
Add code to verify that operations on short/long bitsets operate on the correct epoch.


pgavlin:
See the discussion in https://github.com/dotnet/coreclr/pull/11233 for more details.

These bitsets currently have no checks to ensure that the epoch in which a bitset was created matches the epoch in which it is being used. Depending on the actual representation and the operation, mismatching epochs can have a variety of harmful effects as they can cause out-of-bounds reads or writes.

@BruceForstall writes
> Ideally, we would implement a DEBUG ShortLong representation bitset that also carried epoch along with every bitset, and asserted that operations between two bitsets were in the same epoch.

Which seems like a good suggestion.

PR #11721, opened 2017-05-18
Fix BitSetUint64, clean and improve the surrounding code.

sandreenko:
In order to do #11263 and #11236 I fixed BitSetUint64, that was outdated.
I fixed several places were unnecessary coping were done.
Fixed the problem with constant arguments.
VARSET_INIT, VARSET_INIT_NOCOPY were deleted.

### DIFF
--- a/src/jit/bitsetasuint64.h
+++ b/src/jit/bitsetasuint64.h
@@ -75,6 +75,11 @@ public:
         return BitSetSupport::CountBitsInIntegral(bs);
     }
 
+    static bool IsEmptyUnion(Env env, UINT64 bs1, UINT64 bs2)
+    {
+        return (bs1 | bs2) == 0;
+    }
+
     static void UnionD(Env env, UINT64& bs1, UINT64 bs2)
     {
         bs1 |= bs2;
@@ -139,6 +144,11 @@ public:
         return (bs1 & bs2) == 0;
     }
 
+    static void LivenessD(Env env, UINT64& in, const UINT64 def, const UINT64 use, const UINT64 out)
+    {
+        in = use | (out & ~def);
+    }
+
     static bool IsSubset(Env env, UINT64 bs1, UINT64 bs2)
     {
         return ((bs1 & bs2) == bs1);
@@ -170,14 +180,13 @@ public:
 #ifdef DEBUG
     static const char* ToString(Env env, UINT64 bs)
     {
-        IAllocator* alloc          = BitSetTraits::GetDebugOnlyAllocator(env);
-        const int   CharsForUINT64 = sizeof(UINT64) * 2;
-        char*       res            = nullptr;
-        const int   AllocSize      = CharsForUINT64 + 4;
-        res                        = (char*)alloc->Alloc(AllocSize);
-        UINT64   bits              = bs;
-        unsigned remaining         = AllocSize;
-        char*    ptr               = res;
+        const int CharsForUINT64 = sizeof(UINT64) * 2;
+        char*     res            = nullptr;
+        const int AllocSize      = CharsForUINT64 + 4;
+        res                      = (char*)BitSetTraits::DebugAlloc(env, AllocSize);
+        UINT64   bits            = bs;
+        unsigned remaining       = AllocSize;
+        char*    ptr             = res;
         for (unsigned bytesDone = 0; bytesDone < sizeof(UINT64); bytesDone += sizeof(unsigned))
         {
             unsigned bits0 = (unsigned)bits;
--- a/src/jit/bitsetasuint64.h
+++ b/src/jit/bitsetasuint64.h
@@ -248,8 +248,8 @@ public:
         }
     };
 
-    typedef UINT64 ValArgType;
-    typedef UINT64 RetValType;
+    typedef const UINT64 ValArgType;
+    typedef UINT64       RetValType;
 };
 
 #endif // bitSetAsUint64_DEFINED
--- a/src/jit/codegenlegacy.cpp
+++ b/src/jit/codegenlegacy.cpp
@@ -45,7 +45,7 @@ void CodeGen::genDyingVars(VARSET_VALARG_TP beforeSet, VARSET_VALARG_TP afterSet
     unsigned   varNum;
     LclVarDsc* varDsc;
     regMaskTP  regBit;
-    VARSET_TP  VARSET_INIT_NOCOPY(deadSet, VarSetOps::Diff(compiler, beforeSet, afterSet));
+    VARSET_TP  deadSet(VarSetOps::Diff(compiler, beforeSet, afterSet));
 
     if (VarSetOps::IsEmpty(compiler, deadSet))
         return;
@@ -5627,11 +5627,9 @@ void CodeGen::genCodeForQmark(GenTreePtr tree, regMaskTP destReg, regMaskTP best
             // need to be spilled
 
             VARSET_TP VARSET_INIT(compiler, rsLiveNow, compiler->compCurLife);
-            VARSET_TP VARSET_INIT_NOCOPY(rsLiveAfter, compiler->fgUpdateLiveSet(compiler->compCurLife,
-                                                                                compiler->compCurLifeTree, tree));
+            VARSET_TP rsLiveAfter(compiler->fgUpdateLiveSet(compiler->compCurLife, compiler->compCurLifeTree, tree));
 
-            VARSET_TP VARSET_INIT_NOCOPY(regVarLiveNow,
-                                         VarSetOps::Intersection(compiler, compiler->raRegVarsMask, rsLiveNow));
+            VARSET_TP regVarLiveNow(VarSetOps::Intersection(compiler, compiler->raRegVarsMask, rsLiveNow));
 
             VARSET_ITER_INIT(compiler, iter, regVarLiveNow, varIndex);
             while (iter.NextElem(&varIndex))
@@ -12515,7 +12513,7 @@ void CodeGen::genCodeForBBlist()
         }
 #endif // DEBUG
 
-        VARSET_TP VARSET_INIT_NOCOPY(liveSet, VarSetOps::UninitVal());
+        VARSET_TP liveSet(VarSetOps::UninitVal());
 
         regMaskTP gcrefRegs = 0;
         regMaskTP byrefRegs = 0;
@@ -20623,9 +20621,9 @@ GenTreePtr Compiler::fgLegacyPerStatementLocalVarLiveness(GenTreePtr startNode, 
     MemoryKindSet memoryDef_BeforeSplit   = fgCurMemoryDef;
     MemoryKindSet memoryHavoc_BeforeSplit = fgCurMemoryHavoc;
 
-    VARSET_TP VARSET_INIT_NOCOPY(defSet_AfterThenTree, VarSetOps::MakeEmpty(this)); // These two variables will store
-                                                                                    // the USE and DEF sets after
-    VARSET_TP VARSET_INIT_NOCOPY(useSet_AfterThenTree, VarSetOps::MakeEmpty(this)); // evaluating the thenTree.
+    VARSET_TP defSet_AfterThenTree(VarSetOps::MakeEmpty(this)); // These two variables will store
+                                                                // the USE and DEF sets after
+    VARSET_TP useSet_AfterThenTree(VarSetOps::MakeEmpty(this)); // evaluating the thenTree.
 
     MemoryKindSet memoryUse_AfterThenTree   = fgCurMemoryUse;
     MemoryKindSet memoryDef_AfterThenTree   = fgCurMemoryDef;
--- a/src/jit/codegenlegacy.cpp
+++ b/src/jit/codegenlegacy.cpp
@@ -5626,7 +5626,7 @@ void CodeGen::genCodeForQmark(GenTreePtr tree, regMaskTP destReg, regMaskTP best
             // If any candidates are not alive at the GT_QMARK node, then they
             // need to be spilled
 
-            VARSET_TP VARSET_INIT(compiler, rsLiveNow, compiler->compCurLife);
+            VARSET_TP rsLiveNow(VarSetOps::MakeCopy(compiler, compiler->compCurLife));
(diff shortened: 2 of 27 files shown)


sandreenko:
Last 3 PR (Test BSShortLong, 	Test BSUInt64Class, Return the default BSShortLong) checks different reps for VARSET_TP.

sandreenko:
@dotnet/jit-contrib PTAL.
Last 4 commits will be dropped before the merge.

src/jit/bitsetasuint64.h
```diff
@@ -139,6 +144,11 @@ class BitSetOps</*BitSetType*/ UINT64,
         return (bs1 & bs2) == 0;
     }
 
+    static void LivenessD(Env env, UINT64& in, const UINT64 def, const UINT64 use, const UINT64 out)
```
pgavlin: The `LivenessD` and `IsEmptyUnion` functions do not appear to be used anywhere. If this is the case, please remove them for now.

sandreenko: There is at least one use of each.
IsEmptyUnion: codegencommon.cpp:1079,
LivenessD: liveness.cpp:1193

pgavlin: Ah, I see--I thought that these were new functions and was only looking in the changes you made.

sandreenko: They were missed for BSUInt64Class, it is why I was not able to use BSUInt64Class as VARSET_REP before.

sandreenko:
@dotnet-bot help

sandreenko:
@dotnet-bot test Windows_NT x64 CoreCLR Perf Tests

sandreenko:
@dotnet-bot test dotnet-bot test windows_nt x64 perf

sandreenko:
@dotnet-bot test dotnet-bot test windows_nt x86 perf

sandreenko:
@dotnet-bot test dotnet-bot test windows_nt x64 perf
@dotnet-bot test dotnet-bot test windows_nt x86 perf

sandreenko:
@dotnet-bot test windows_nt x64 perf
@dotnet-bot test windows_nt x86 perf

src/jit/codegenlegacy.cpp
```diff
@@ -5646,12 +5646,10 @@ void CodeGen::genCodeForQmark(GenTreePtr tree, regMaskTP destReg, regMaskTP best
             // If any candidates are not alive at the GT_QMARK node, then they
             // need to be spilled
 
-            VARSET_TP VARSET_INIT(compiler, rsLiveNow, compiler->compCurLife);
-            VARSET_TP VARSET_INIT_NOCOPY(rsLiveAfter, compiler->fgUpdateLiveSet(compiler->compCurLife,
-                                                                                compiler->compCurLifeTree, tree));
+            VARSET_VALARG_TP rsLiveNow(compiler->compCurLife);
```
BruceForstall: >VARSET_VALARG_TP [](start = 12, length = 16)

I think the VARSET_VALARG_TP is only intended to be used in a function signature, not to declare variables.

src/jit/emit.cpp
```diff
@@ -3341,23 +3341,23 @@ void emitter::emitDispIGlist(bool verbose)
 void emitter::emitDispGCinfo()
 {
     printf("Emitter GC tracking info:");
-    printf("\n  emitPrevGCrefVars(0x%p)=%016llX ", dspPtr(&emitPrevGCrefVars), emitPrevGCrefVars);
+    // printf("\n  emitPrevGCrefVars(0x%p)=%016llX ", dspPtr(&emitPrevGCrefVars), emitPrevGCrefVars);
     dumpConvertedVarSet(emitComp, emitPrevGCrefVars);
```
BruceForstall: Why is this commented out? I guess because these are bitsets that aren't necessarily pointers. You should at least leave the:
```
printf("\n  emitPrevGCrefVars ");
```

src/jit/liveness.cpp
```diff
@@ -1626,7 +1626,7 @@ void Compiler::fgComputeLifeCall(VARSET_TP& life, GenTreeCall* call)
 //    `true` if the local var node corresponds to a dead store; `false`
 //    otherwise.
 //
-bool Compiler::fgComputeLifeLocal(VARSET_TP& life, VARSET_TP& keepAliveVars, GenTree* lclVarNode, GenTree* node)
+bool Compiler::fgComputeLifeLocal(VARSET_TP& life, VARSET_VALARG_TP keepAliveVars, GenTree* lclVarNode, GenTree* node)
 {
```
BruceForstall: You converted this to a value type, not a reference type. Do you want it copied? That doesn't seem right.

src/jit/liveness.cpp
```diff
@@ -1815,19 +1815,17 @@ bool Compiler::fgComputeLifeLocal(VARSET_TP& life, VARSET_TP& keepAliveVars, Gen
...
+                             GenTreePtr       startNode,
+                             GenTreePtr       endNode,
+                             VARSET_VALARG_TP volatileVars,
+                             bool* pStmtInfoDirty DEBUGARG(bool* treeModf))
 {
     GenTreePtr tree;
     unsigned   lclNum;
 
-    VARSET_TP VARSET_INIT(this, life, lifeArg); // lifeArg is const ref; copy to allow modification.
-
```
BruceForstall: Does the caller expect lifeArg to be modified? Previously, it wasn't modified.

src/jit/liveness.cpp
```diff
@@ -1865,47 +1863,37 @@ VARSET_VALRET_TP Compiler::fgComputeLife(VARSET_VALARG_TP lifeArg,
...
 }
 
-VARSET_VALRET_TP Compiler::fgComputeLifeLIR(VARSET_VALARG_TP lifeArg, BasicBlock* block, VARSET_VALARG_TP volatileVars)
+void Compiler::fgComputeLifeLIR(VARSET_TP& life, BasicBlock* block, VARSET_VALARG_TP volatileVars)
 {
-    VARSET_TP VARSET_INIT(this, life, lifeArg); // lifeArg is const ref; copy to allow modification.
-
-    VARSET_TP VARSET_INIT(this, keepAliveVars, volatileVars);
-    VarSetOps::UnionD(this, keepAliveVars, block->bbScope); // Don't kill vars in scope
+    VARSET_TP keepAliveVars = VarSetOps::Union(this, volatileVars, block->bbScope); // Don't kill vars in scope
```
BruceForstall: Should this instead use constructor syntax:
```
    VARSET_TP keepAliveVars(VarSetOps::Union(this, volatileVars, block->bbScope)); // Don't kill vars in scope
```
?

sandreenko:
I recommend to review it by looking at one commit per time, not the whole change.

src/jit/liveness.cpp
sandreenko: It is from "Do not copy life sets in fgComputeLife." commit. 
It was used only in code like VarSetOps::AssignNoCopy(this, life, fgComputeLife(life, tree, nextColonExit, volatileVars)).
I replaced it to just update the life and save VARSET allocation.

src/jit/codegenlegacy.cpp
sandreenko: This variable is used only like function argument.

src/jit/emit.cpp
sandreenko: I am planning to write dumpVarSet and use it there in the next PR.  A function with such name existed, I found two outdates comments that referenced here, but it was deleted before tfs history.

src/jit/liveness.cpp
sandreenko: VARSET_VALARG_TP means unchangeable. It is "const VARSET_TP&".

src/jit/liveness.cpp
sandreenko: Yes, thanks, I missed that.

src/jit/codegenlegacy.cpp
BruceForstall: But the *intent* is for it to be used in a function signature. (I'm not sure the reason, honestly). In this case, why wouldn't you just use VARSET_TP?

src/jit/emit.cpp
BruceForstall: I don't understand: you can't delete the code that prints the name of the varset being displayed.

sandreenko: Yes, you are write, so is the right thing to do is just leave the 
`printf("\n  emitPrevGCrefVars ");` 
in the final change?

BruceForstall: yes.

src/jit/codegenlegacy.cpp
sandreenko: Looks like it is a naming problem.
As I understand that:
1. VARSET_VALARG_TP was created to replace "const VARSET_TP &" that was used in a function signature to pass an argument that could be read, but can't be changed. We do not need this type anywhere except arg lists.
2. If an argument can be changed, then we use "VARSET_TP &", fgComputeLifeCall is an example.

I don't see a problem in such declaration, when we create an argument not in argList, because it has a long name for example.
Compare:
```
VARSET_VALARG_TP gcrefVarsArg(getEmitter()->emitThisGCrefVars); 
getEmitter()->emitCreatePlaceholderIG(IGPT_EPILOG, block, gcrefVarsArg, gcrefRegsArg, byrefRegsArg, last); 
```
and 
`getEmitter()->emitCreatePlaceholderIG(IGPT_EPILOG, block, getEmitter()->emitThisGCrefVars, gcrefRegsArg, byrefRegsArg, last); `
I think the first variant is better.

I do not want to use VARSET_TP because there is a possibility that somebody in the future make a change to modify this variable without understanding that `compiler->compCurLife` will change too.

src/jit/emit.cpp
sandreenko: Got it, fixed.

src/jit/codegenlegacy.cpp
pgavlin: How about just using `const VARSET_TP&` explicitly here?

src/jit/liveness.cpp
```diff
@@ -1865,47 +1863,38 @@ VARSET_VALRET_TP Compiler::fgComputeLife(VARSET_VALARG_TP lifeArg,
...
-    VarSetOps::UnionD(this, keepAliveVars, block->bbScope); // Don't kill vars in scope
+    // Don't kill volatile vars and vars in scope.
+    VARSET_TP keepAliveVars(VarSetOps::Union(this, volatileVars, block->bbScope));
 
     noway_assert(VarSetOps::IsSubset(this, keepAliveVars, life));
 
     LIR::Range& blockRange      = LIR::AsRange(block);
     GenTree*    firstNonPhiNode = blockRange.FirstNonPhiNode();
-    if (firstNonPhiNode == nullptr)
+    if (firstNonPhiNode != nullptr)
```
pgavlin: I would prefer that you retain the early-return.
```
if (firstNonPhiNode == nullptr)
{
    return;
}
```
should suffice.

src/jit/codegenlegacy.cpp
sandreenko: Yes, it sounds good to me. I think I will be glad to replace all `VARSET_VALARG_TP` with `const VARSET_TP&,`. `const VARSET_TP` has the same length but it is more obvious.

src/jit/lsra.cpp
```diff
@@ -6655,7 +6655,7 @@ void LinearScan::processBlockStartLocations(BasicBlock* currentBlock, bool alloc
     VarToRegMap inVarToRegMap     = getInVarToRegMap(currentBlock->bbNum);
     bool        hasCriticalInEdge = blockInfo[currentBlock->bbNum].hasCriticalInEdge;
 
-    VARSET_TP VARSET_INIT_NOCOPY(liveIn, currentBlock->bbLiveIn);
```
pgavlin: `const VARSET_TP&`?

src/jit/lsra.cpp
```diff
@@ -6880,7 +6880,7 @@ void LinearScan::processBlockEndLocations(BasicBlock* currentBlock)
     assert(currentBlock != nullptr && currentBlock->bbNum == curBBNum);
     VarToRegMap outVarToRegMap = getOutVarToRegMap(curBBNum);
 
-    VARSET_TP VARSET_INIT_NOCOPY(liveOut, currentBlock->bbLiveOut);
+    VARSET_TP& liveOut(currentBlock->bbLiveOut);
```
pgavlin: `const VARSET_TP&`?

src/jit/lsra.cpp
sandreenko: It is a bit more complex. What was write here meant: there are two varsets, chose one and read data from that one. The problem is that it is hard to protect live_in set from a modification, moreover it will have different effects in cases of short/long sets. So I rewrote this part with pointers, now it should be better.

src/jit/lsra.cpp
sandreenko: Fixed, as the previous one.

src/jit/liveness.cpp
sandreenko: Fixed

src/jit/lsra.cpp
pgavlin: Makes sense; thanks.

CarolEidt:
I think the changes are good, overall. And I applaud the elimination of `VARSET_INIT_NOCOPY`. I also think it's good to make `VARSET_VALARG_TP` explicitly `const` (which I don't think it was, previously). However, I think it would be good to be quite explicit about that in the comments in both bitset.h and varset.h.

sandreenko:
> I think the changes are good, overall. And I applaud the elimination of VARSET_INIT_NOCOPY. I also think it's good to make VARSET_VALARG_TP explicitly const (which I don't think it was, previously). However, I think it would be good to be quite explicit about that in the comments in both bitset.h and varset.h.

I added comments about the new functions and the changed arg type.

sandreenko:
@dotnet-bot test

sandreenko:
@dotnet-bot retest this please

sandreenko:

sandreenko:
it is infrastructure

PR #11726, opened 2017-05-18
Make ActivityID guids process specific.

vancem:
This is a port a change from the Nuget version of this library.

Bascially it hashs in the process ID into the activity ID GUID.  This make collisions across processes unlikely, so you don't need to use this ID in conjunction with process ID to maintain uniqueness.  

@brianrob


danmosemsft:
@brianrob chasing up old PR's .. do you have next action here?

brianrob:
@dotnet-bot test this please

Issue #11716, opened 2017-05-18
Lots of warnings from Build.Common.props during build


janvorli:
There is a lot of warnings like the following now during coreclr build:
`home/janvorli/git/coreclr/Tools/Build.Common.props(53,3): warning MSB4011: "/home/janvorli/git/coreclr/Tools/BuildVersion.targets" cannot be imported again. It was already imported at "/home/janvorli/git/coreclr/dir.props (90,3)". This is most likely a build authoring error. This subsequent import will be ignored. [/home/janvorli/git/coreclr/src/.nuget/Microsoft.NETCore.ILDAsm/Microsoft.NETCore.ILDAsm.pkgproj]`

PR #11728, opened 2017-05-18
Get rid of double-import of Buildversion.targets

wtgodbe:
Attempts to resolve https://github.com/dotnet/coreclr/issues/11716. Build.Common.Props now imports Buildversion.targets, so there's no need to explicitly import Buildversion.targets in the same file where we import Build.Common.Props

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -87,8 +87,6 @@
     <OutputPath Condition="'$(OutputPath)' == ''">$(BinDir)</OutputPath>
   </PropertyGroup>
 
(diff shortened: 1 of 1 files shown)


wtgodbe:
CC @janvorli @gkhanna79

wtgodbe:
Local build didn't show any of the types of warning in question

gkhanna79:
> Local build didn't show any of the types of warning in question

Do you mean after this change @wtgodbe ?

wtgodbe:
@gkhanna79 yes

PR #11738, opened 2017-05-19
gdbjit symbols exported

chunseoklee:
Two symbols are exported to enable gdbjit feature. In tizen build system, these symbols are stripped during rpmbuild.
Symbols added are :
    __jit_debug_register_code
    __jit_descriptor

### DIFF
--- a/functions.cmake
+++ b/functions.cmake
@@ -86,7 +86,10 @@ function(preprocess_def_file inputFilename outputFilename)
                               PROPERTIES GENERATED TRUE)
 endfunction()
 
-function(generate_exports_file inputFilename outputFilename)
+function(generate_exports_file)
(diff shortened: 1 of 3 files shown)


janvorli:
@chunseoklee there is a problem with adding these symbols to the list on OSX. OSX linker requires the symbols in the exported symbol list to exist and it is not the case for these symbols when the FEATURE_GDBJIT is not enabled. Unfortunately, we don't have a way to have conditional exports. So we will need to add such support. It seems that the simplest way would be to modify the generate_exports_file so that it can have any number of input files. Then we can put these new symbols into a separate file and pass that file to the generate_exports_file only if the FEATURE_GDBJIT is enabled. AWK can be passed multiple files at once, so the change would be trivial.

PR #11742, opened 2017-05-19
Implement GetCurrentProcessCpuCount for Unix and merge NumberOfProcesors implementations.

tmds:
@janvorli PTAL


src/pal/src/numa/numa.cpp
```diff
@@ -599,6 +597,14 @@ GetProcessAffinityMask(
             *lpProcessAffinityMask = processMask;
             *lpSystemAffinityMask = systemMask;
         }
+        else if (errno == EINVAL)
```
janvorli: As I have mentioned before, I was thinking this cannot happen. But if I am wrong, then we should use what you have in the other place (ideally extracting the dynamic allocation to a separate function and using it at both places)

src/utilcode/util.cpp
```diff
@@ -1306,10 +1303,8 @@ int GetCurrentProcessCpuCount()
 
 #else // !FEATURE_PAL
 
-    SYSTEM_INFO sysInfo;
-    ::GetSystemInfo(&sysInfo);
-    cCPUs = sysInfo.dwNumberOfProcessors;
-    return sysInfo.dwNumberOfProcessors;
+    cCPUs = PAL_GetCurrentProcessCpuCount();
```
janvorli: Could you please remove the ifdef instead and use the same code for Windows / Unix? And remove the PALGetCurrentProcessCpuCount from pal.h

src/pal/src/misc/sysinfo.cpp
```diff
@@ -128,6 +128,52 @@ int GetNumberOfProcessors()
...
+#ifdef HAVE_SCHED_GETAFFINITY
+
+    int pid = getpid();
+    size_t setsize;
+    cpu_set_t* set;
+    int ncpus = CPU_SETSIZE;
+    int rv;
+    do
+    {
+        setsize = CPU_ALLOC_SIZE(ncpus);
```
tmds: cpu_set_t will work almost always. The kernel may support more cpus than the mask of the clibrary supports. My stock fedora x64 kernel supports up to 8192 while cpu_set_t (glibc) stores up to 1024 cpus. The man page mentions the EINVAL as a result of providing a smaller mask than the kernel. The kernel also takes into account the number of installed cpus, so as long as you don't have more than 1024, it works fine.

tmds:
@janvorli thank you for the review, I will continue working on this after the weekend.

src/gc/unix/gcenv.unix.cpp
```diff
@@ -419,16 +412,15 @@ size_t GCToOSInterface::GetLargestOnDieCacheSize(bool trueSize)
 //  specify a 1 bit for a processor when the system affinity mask specifies a 0 bit for that processor.
 bool GCToOSInterface::GetCurrentProcessAffinityMask(uintptr_t* processMask, uintptr_t* systemMask)
 {
-    // TODO(segilles) processor detection
-    return false;
+    return !!::GetProcessAffinityMask(::GetCurrentProcess(), (PDWORD_PTR)processMask, (PDWORD_PTR)systemMask);
```
tmds: GetProcessAffinityMask is defined in numa.cpp. I'm not sure how to move it here. Perhaps numa.cpp should move under gc? I will revert back to the `return false` implementation for now.

src/utilcode/util.cpp
tmds: I moved this to pal to be able to use HAVE_SCHED_GETAFFINITY. I think this can't be accessed under utilcode.

src/gc/unix/gcenv.unix.cpp
```diff
@@ -77,14 +77,7 @@ static size_t g_RestrictedPhysicalMemoryLimit = 0;
...
 {
-    // Calculate and cache the number of processors on this machine
-    int cpuCount = sysconf(_SC_NPROCESSORS_ONLN);
-    if (cpuCount == -1)
-    {
-        return false;
-    }
-
-    g_logicalCpuCount = cpuCount;
+    g_logicalCpuCount = PAL_GetLogicalCpuCountFromOS();
```
tmds: Aha. I assumed gcenv.unix.cpp would be the implementation used by coreclr on Linux, but in fact that is gc.os.cpp...
So, this is the implementation used by corert.

tmds:
@janvorli The OSX build fails with:
```
02:04:43 Undefined symbols for architecture x86_64:
02:04:43   "_PAL_GetCurrentProcessCpuCount", referenced from:
02:04:43       GetCurrentProcessCpuCount() in libutilcodestaticnohost.a(util.cpp.o)
```
I don't know why it can't find the function. A bit unexpected the function name is prefixed with an underscore... Do you know what may be the issue?

src/gc/unix/gcenv.unix.cpp
janvorli: I meant copy and paste the source here. I know it sounds strange, but I don't see a clean way that would work - the GC is being transformed to a completely standalone self-contained thing that can be built without any dependency on PAL.

src/pal/src/misc/sysinfo.cpp
janvorli: Ok, thanks for the details, it makes sense to use your new implementation then.

src/utilcode/util.cpp
janvorli: I meant just removing the ifdef and keeping only the code that was used for Windows before. The GetProcessAffinityMask that the Windows code uses now works on Unix too. It would also unify the behavior with Windows - the function GetCurrentProcessCpuCount never returns larger number than 64 for Windows. And I have just realized that the GetSystemInfo should return number <= 64 in the dwNumberOfProcessors (it means the same number we get from the GetProcessAffinityMask). The MSDN doc says: 
> dwNumberOfProcessors
The number of logical processors **in the current group**.

A CPU group contains max 64 processors. 
These two changes need to match, since there is an array allocated with size equal to the SYSTEM_INFO.dwNumberOfProcessors and indexed by value obtained from the GetCurrentProcessCpuCount.

janvorli:
@tmds The OSX fails because the PAL_GetCurrentProcessCpuCount was not added to the exports list (src/dlls/mscoree/mscorwks_unixexports.src). The underscore prefix is added to all C style functions on OSX.

src/pal/src/numa/numa.cpp
```diff
@@ -208,11 +206,16 @@ NUMASupportInitialize()
...
+        g_cpuCount = PAL_GetLogicalCpuCountFromOS();
         g_groupCount = 1;
         g_highestNumaNode = 0;
 
         AllocateLookupArrays();
+
+        for (int i = 0; i < g_cpuCount; i++)
+        {
+            g_cpuToAffinity[i].Number = i;
+        }
```
tmds: @janvorli I think this was missing.

src/gc/unix/gcenv.unix.cpp
```diff
@@ -417,18 +429,99 @@ size_t GCToOSInterface::GetLargestOnDieCacheSize(bool trueSize)
 //  A process affinity mask is a subset of the system affinity mask. A process is only allowed
 //  to run on the processors configured into a system. Therefore, the process affinity mask cannot
 //  specify a 1 bit for a processor when the system affinity mask specifies a 0 bit for that processor.
-bool GCToOSInterface::GetCurrentProcessAffinityMask(uintptr_t* processMask, uintptr_t* systemMask)
+bool GCToOSInterface::GetCurrentProcessAffinityMask(uintptr_t* lpProcessAffinityMask, uintptr_t* lpSystemAffinityMask)
```
tmds: I don't think corert (FEATURE_REDHAWK?) does numa so I added non-NUMA GetCurrentProcessAffinityMask and GetCurrentProcessCpuCount implementations.

src/pal/src/numa/numa.cpp
janvorli: Yes, you are right. Thank you for spotting this. I was rushing the original change out before the feature complete date and this is a lesson for me to not to rush things. Could you please add setting the Group member to zero?

src/gc/unix/gcenv.unix.cpp
```diff
@@ -417,18 +429,99 @@ size_t GCToOSInterface::GetLargestOnDieCacheSize(bool trueSize)
...
+
+    pmask &= smask;
+
+    int count = 0;
+    while (pmask)
+    {
+        if (pmask & 1)
+            count++;
+
+        pmask >>= 1;
```
janvorli: A nit - you can make this a bit better by changing the `pmask >>= 1` to `pmask &= (pmask - 1)`. That reduces the number of loop to the number of set bits in the mask.

tmds: nice!

src/pal/src/misc/sysinfo.cpp
```diff
@@ -94,13 +94,10 @@ SET_DEFAULT_DEBUG_CHANNEL(MISC);
...
-/*++
-Function:
-    GetNumberOfProcessors
 
-Return number of processors available for the current process
---*/
-int GetNumberOfProcessors()
+DWORD
+PALAPI
+PAL_GetLogicalCpuCountFromOS()
```
tmds: This function existed already and is used by GetLogicalCpuCountFromOS in util.cpp.

tmds:
@janvorli The OSX build fails because it can't find `GetProcessAffinityMask`. I have added that to `mscorwks_unixexports.src`. Perhaps I need to put it someplace else? The function is defined in `numa.cpp` and used in `util.cpp`.

janvorli:
@tmds you need to add it to the libmscordaccore.so exports too - src/dlls/mscordac/mscordac_unixexports.src

janvorli:

tmds:
@dotnet-bot test this please

tmds:

PR #11745, opened 2017-05-19
Delete default copy/move of RelativePointer/RelativeFixupPointer

ruben-ayrapetyan:
As value of `RelativePointer`/`RelativeFixupPointer` depends on pointer's location, it is usually not valid to perform bitwise copy of the pointer objects.
The changes remove the implicit copy/move constructors and assignment operators for the classes to catch incorrect copying at compile-time.
Bitwise copy of `RelativePointer` is added as separate method `BitwiseCopyTo` for exceptional cases, in which it is necessary.

### DIFF
--- a/src/vm/field.h
+++ b/src/vm/field.h
@@ -43,6 +43,8 @@ class FieldDesc
   protected:
     RelativePointer<PTR_MethodTable> m_pMTOfEnclosingClass;  // This is used to hold the log2 of the field size temporarily during class loading.  Yuck.
 
+    // See also: FieldDesc::InitializeFrom method
+
 #if defined(DACCESS_COMPILE)
     union { //create a union so I can get the correct offset for ClrDump.
         unsigned m_dword1;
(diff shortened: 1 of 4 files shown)


ruben-ayrapetyan:
@dotnet-bot Test Ubuntu arm Cross Release Build

ruben-ayrapetyan:

ruben-ayrapetyan:
@jkotas, Please take a look.

src/vm/field.h
```diff
@@ -85,10 +85,35 @@ class FieldDesc
     LPUTF8 m_debugName;
 #endif
 
+public:
     // Allocated by special heap means, don't construct me
-    FieldDesc() {};
+    FieldDesc() =delete;
+
+#ifndef DACCESS_COMPILE
+    FieldDesc& operator = (const FieldDesc& src)
```
jkotas: Where is this used? It does not look right that it is needed.

jkotas:
LGTM otherwise.  Thanks!

src/vm/field.h
ruben-ayrapetyan: Is it used in `src/vm/generics.cpp`:
```
if (pOldMT->GetNumStaticFields() != 0)
{
    pStaticFieldDescs = (FieldDesc*) pamTracker->Track(pAllocator->GetLowFrequencyHeap()->AllocMem(S_SIZE_T(sizeof(FieldDesc)) * S_SIZE_T(pOldMT->GetNumStaticFields())));
    FieldDesc* pOldFD = pOldMT->GetGenericsStaticFieldDescs();

    g_IBCLogger.LogFieldDescsAccess(pOldFD);

    for (DWORD i = 0; i < pOldMT->GetNumStaticFields(); i++)
    {
        pStaticFieldDescs[i] = pOldFD[i];
        pStaticFieldDescs[i].SetMethodTable(pMT);
    }
}
```

The pointer in is overwritten `SetMethodTable` just after copy assignment.

Is it better to replace the assignment operator with a `Copy` or `InitializeFrom` method?

jkotas: > Is it better to replace the assignment operator with an InitializeFrom method?

Yes, I think it would look better.

ruben-ayrapetyan: Fixed.

ruben-ayrapetyan: However, I'm thinking about moving all fields (except the pointer) to another structure like `SimplyCopyable` to avoid explicit copy of every member field. Is it a good idea?

ruben-ayrapetyan:

src/vm/field.h
jkotas: > I'm thinking about moving all fields (except the pointer) to another structure like SimplyCopyable 

I do not think it would be an improvement.

ruben-ayrapetyan:
@jkotas, Thank you!

PR #11747, opened 2017-05-19
Update signing for coreclr repo

weshaggard:
Make sign.builds consistent with corefx and only add strongname
signatures for managed assemblies and not native assemblies

cc @gkhanna79 @wtgodbe


weshaggard:
Merging before CI to unblock official builds.

PR #11750, opened 2017-05-19
Fix legality checks in SetRegOptionalForBinOp.

### LINKED ISSUE #11734: System.Runtime.CompilerServices.UnsafeTests.WriteBytesIntoInt32 failing on Ubuntu x64 under JitStressRegs=4
e.g. https://ci.dot.net/job/dotnet_coreclr/job/master/job/jitstress/job/x64_checked_ubuntu_corefx_jitstressregs4/178/

pgavlin:
It is not legal to make an operand register optional on xarch if the type
of the operand is narrower than the type of the operator and the contents
of the upper bits are not known to be zero. `SetRegOptionalForBinOp` was
not considering the case where `op1` was legal but `op2` was the preferred
register-optional operand and was illegal.

Fixes #11734.

### DIFF
--- a/src/jit/lower.h
+++ b/src/jit/lower.h
@@ -178,15 +178,17 @@ private:
     {
         assert(GenTree::OperIsBinary(tree->OperGet()));
 
-        GenTree* op1 = tree->gtGetOp1();
-        GenTree* op2 = tree->gtGetOp2();
+        GenTree* const op1 = tree->gtGetOp1();
+        GenTree* const op2 = tree->gtGetOp2();
 
-        if (tree->OperIsCommutative() && tree->TypeGet() == op1->TypeGet())
+        const bool op1Legal = tree->OperIsCommutative() && (tree->TypeGet() == op1->TypeGet());
+        const bool op2Legal = tree->TypeGet() == op2->TypeGet();
+
+        if (op1Legal)
         {
-            GenTree* preferredOp = PreferredRegOptionalOperand(tree);
-            SetRegOptional(preferredOp);
+            SetRegOptional(op2Legal ? PreferredRegOptionalOperand(tree) : op1);
         }
-        else if (tree->TypeGet() == op2->TypeGet())
+        else if (op2Legal)
         {
             SetRegOptional(op2);
         }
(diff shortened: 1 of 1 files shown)


pgavlin:

pgavlin:
@BruceForstall the only diffs I am seeing in SPMI are of the following form:

```diff
        mov      r8, gword ptr [rsp+118H]
        mov      r9, gword ptr [rsp+120H]
        mov      rcx, r8
        or       rcx, r9
-       or       rcx, qword ptr [rsp+128H]
+       mov      r10, gword ptr [rsp+128H]
+       or       rcx, r10
        jne      SHORT G_M64274_IG07
```

It would appear that this is all code that is attempting to check whether a number of GC pointers are non-null. These sorts of changes account for all ~100 diffs that are reported.

src/jit/lower.h
```diff
@@ -178,15 +178,17 @@ class Lowering : public Phase
...
         assert(GenTree::OperIsBinary(tree->OperGet()));
 
-        GenTree* op1 = tree->gtGetOp1();
-        GenTree* op2 = tree->gtGetOp2();
+        GenTree* const op1 = tree->gtGetOp1();
+        GenTree* const op2 = tree->gtGetOp2();
 
-        if (tree->OperIsCommutative() && tree->TypeGet() == op1->TypeGet())
+        const bool op1Legal = tree->OperIsCommutative() && (tree->TypeGet() == op1->TypeGet());
+        const bool op2Legal = tree->TypeGet() == op2->TypeGet();
```
CarolEidt: It would seem that the kinds of diffs you have seen, although small, could potentially be avoided by checking `genTypeSize` instead of strict equality.

pgavlin: I agree. I will experiment with this a bit.

pgavlin: Changing the legality tests to use the size rather than the type results in some minor improvements, all of which are the inverse of the diff I reported above. For example:

```diff
        cmp      r8d, 9
        jne      SHORT G_M65468_IG04
-       mov      rax, gword ptr [rsp+28H]
-       or       r9, rax
+       or       r9, qword ptr [rsp+28H]
        jne      SHORT G_M65468_IG04
        mov      eax, 264
```

CarolEidt: Excellent!

PR #11757, opened 2017-05-19
Update Unix SafeHandle to throw NotFound correctly

JeremyKuhne:
Need to match Windows semantics for missing files. This means throwing
FileNotFound only if the last segment of the path can't be found.

Related to https://github.com/dotnet/corefx/pull/19959 and https://github.com/dotnet/corefx/issues/19850.

### DIFF
--- a/src/mscorlib/shared/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs
+++ b/src/mscorlib/shared/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs
@@ -43,6 +43,7 @@ namespace Microsoft.Win32.SafeHandles
 
             if (handle.IsInvalid)
             {
+                handle.Dispose();
                 Interop.ErrorInfo error = Interop.Sys.GetLastErrorInfo();
 
                 // If we fail to open the file due to a path not existing, we need to know whether to blame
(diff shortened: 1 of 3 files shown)


JeremyKuhne:
Validated that this has the expected behavior.  That said, this change requires updating UnixFileSystem.CopyFile() in CoreFX (NonExistentPath copy test is failing). Still investigating to see if either the error semantics are either different for the Win32 CopyFile or if it has to do with the order of opening file handles (e.g. Windows may try and open destination first).

JeremyKuhne:
Heh, scratch that. For whatever reason I didn't look close enough to see that the test was conditioned for the old behavior. This looks clean outside of the if(Windows) bit in NonExistentPath. :)

JeremyKuhne:
Validated fully. `System.IO.FileSystem.Tests` needs [`File_Copy_str_str.NonExistentPath()`](https://github.com/dotnet/corefx/blob/6451bed4d4a667d9c05a41c0843504d020d283ab/src/System.IO.FileSystem/tests/File/Copy.cs#L61-L72)'s OS condition removed to take this change into CoreFX.

src/mscorlib/shared/Microsoft/Win32/SafeHandles/SafeFileHandle.Unix.cs
```diff
@@ -38,18 +39,29 @@ public SafeFileHandle(IntPtr preexistingHandle, bool ownsHandle) : this(ownsHand
...
-            // Open the file. 
-            SafeFileHandle handle = Interop.CheckIo(
-                Interop.Sys.Open(path, flags, mode),
-                path, 
-                isDirectory: enoentDueToDirectory,
-                errorRewriter: e => (e.Error == Interop.Error.EISDIR) ? Interop.Error.EACCES.Info() : e);
+            if (handle.IsInvalid)
+            {
+                Interop.ErrorInfo error = Interop.Sys.GetLastErrorInfo();
+
```
stephentoub: Ideally we'd dispose of the handle, even though it's invalid, to avoid it needing to be finalized.

JeremyKuhne: Thanks- totally overlooked that.

JeremyKuhne:
Disposed the handle and ported https://github.com/dotnet/corefx/pull/20049

JeremyKuhne:
@stephentoub Can you take a look at the update?

PR #11759, opened 2017-05-20
Fix ReadMemory error reading DSO link_map name

mikem8361:
Issue #17758

### DIFF
--- a/src/debug/createdump/crashinfo.cpp
+++ b/src/debug/createdump/crashinfo.cpp
@@ -434,18 +434,27 @@ CrashInfo::GetDSOInfo()
     }
 
     // Add the DSO link_map entries
+    ArrayHolder<char> moduleName = new char[PATH_MAX];
     for (struct link_map* linkMapAddr = debugEntry.r_map; linkMapAddr != nullptr;) {
         struct link_map map;
         if (!ReadMemory(linkMapAddr, &map, sizeof(map))) {
             return false;
         }
-        char moduleName[257] = { 0 };
+        int i = 0;
         if (map.l_name != nullptr) {
-            if (!ReadMemory(map.l_name, &moduleName, sizeof(moduleName) - 1)) {
-                return false;
+            for (; i < PATH_MAX; i++)
+            {
+                if (!ReadMemory(map.l_name + i, &moduleName[i], 1)) {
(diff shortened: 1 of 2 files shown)


src/debug/createdump/crashinfo.cpp
```diff
@@ -439,13 +439,21 @@ CrashInfo::GetDSOInfo()
...
+                    TRACE("DSO: ReadMemory link_map name %p + %d FAILED\n", map.l_name, i);
+                    break;
+                }
+                if (moduleName[i] == '\0') {
+                    break;
+                }
             }
         }
-        TRACE("DSO: link_map entry %p l_ld %p l_addr %lx %s\n", linkMapAddr, map.l_ld, map.l_addr, moduleName);
+        TRACE("DSO: link_map entry %p l_ld %p l_addr %lx %s\n", linkMapAddr, map.l_ld, map.l_addr, (char*)moduleName);
```
noahfalk: You probably want to set moduleName[MAX_LONGPATH-1] = 0 as a precaution. Otherwise, looks good.

src/debug/createdump/crashinfo.cpp
```diff
@@ -439,13 +439,21 @@ CrashInfo::GetDSOInfo()
         if (!ReadMemory(linkMapAddr, &map, sizeof(map))) {
             return false;
         }
-        char moduleName[257] = { 0 };
+        ArrayHolder<char> moduleName = new char[MAX_LONGPATH];
+        moduleName[0] = '\0';
         if (map.l_name != nullptr) {
-            if (!ReadMemory(map.l_name, &moduleName, sizeof(moduleName) - 1)) {
-                return false;
+            for (int i = 0; i < MAX_LONGPATH; i++)
```
janvorli: The MAX_LONGPATH seems still pretty short - it is defined as 1024. PATH_MAX defined in Unix headers is better - it is 4096. 
A nit - you might read the whole block at once, starting at MAX_LONGPATH, if that fails going to MAX_LONGPATH / 2 etc. That would give you a logarithmic number of calls to ReadMemory (w.r.t. to the max buffer size) instead of linear (w.r.t. to the length of the filename) ones. It seems it would be less calls in 99% of cases.

mikem8361:

PR #11762, opened 2017-05-20
removing NO_WRITE_BARRIER code

Maoni0:
NO_WRITE_BARRIER hasn't been used in ages and will not be used. 

@jkotas @sdmaclea @swgillespie PTAL. I was going to check this in after @swgillespie is done with the integration from the local_gc branch but he's on vacation. And @sdmaclea this means there's a few less lines you'll need to care about for the OS_PAGE_SIZE stuff.


swgillespie:
👍

PR #11765, opened 2017-05-20
Scale back [Serializable] on CoreCLR types

morganbr:
Removes SerializableAttribute from CoreCLR types not intended to be serializable as well as adding special handling to MulticastDelegate to prevent serializing delegates (which can't be correctly serialized cross platform/runtime).

This is the first CoreCLR portion of https://github.com/dotnet/corefx/issues/19119. 

Please do not merge until https://github.com/dotnet/corefx/pull/20035 is merged since those are the test changes that match this change.


morganbr:
CC @dotnet/corert-contrib

src/mscorlib/shared/System/DefaultBinder.cs
```diff
@@ -9,7 +9,6 @@
 namespace System
 {
     //Marked serializable even though it has no state.
```
jkotas: Comment is no longer applicable.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSourceException.cs
```diff
@@ -18,9 +18,6 @@ namespace System.Diagnostics.Tracing
     /// <summary>
     /// Exception that is thrown when an error occurs during EventSource operation.
     /// </summary>
-#if !ES_BUILD_PCL
-    [Serializable]
```
jkotas: I would think that this should stay, but under different ifdef. @brianrob ?

src/mscorlib/src/System/Collections/Generic/Comparer.cs
```diff
@@ -134,7 +132,6 @@ public override int Compare(T x, T y)
     // since we want to serialize as ObjectComparer for
     // back-compat reasons (see below).
 
-    [Serializable]
     internal sealed class Int32EnumComparer<T> : Comparer<T>, ISerializable where T : struct
```
jkotas: Do these need to remain serializable to make the `Comparer<T>` serialization work?

src/mscorlib/shared/System/UnitySerializationHolder.cs
```diff
@@ -12,7 +12,6 @@
 
 namespace System
 {
-    [Serializable]
     // Holds classes (Empty, Null, Missing) for which we guarantee that there is only ever one instance of.
```
jkotas: This whole file should be deleted.

src/mscorlib/shared/System/Reflection/MemberInfoSerializationHolder.cs
```diff
@@ -9,7 +9,6 @@
 
 namespace System.Reflection
 {
-    [Serializable]
 #if CORECLR
```
jkotas: This whole file should be deleted.

src/mscorlib/src/System/Collections/Generic/Comparer.cs
stephentoub: I would expect so. I seem to remember adding these explicitly in response to a bug, and adding tests for them in corefx.

src/mscorlib/shared/System/DefaultBinder.cs
morganbr: Removed

src/mscorlib/src/System/Collections/Generic/Comparer.cs
morganbr: Looks plausible. I've added them.

src/mscorlib/shared/System/UnitySerializationHolder.cs
morganbr: I'll include that with general ISerializable cleanup (I've updated https://github.com/dotnet/corefx/issues/19119 to include that)

src/mscorlib/shared/System/Reflection/MemberInfoSerializationHolder.cs
morganbr: I'll include that with general ISerializable cleanup (I've updated dotnet/corefx#19119 to include that)

AtsushiKan: If you get rid of this file, you can get rid of

  src\System\Reflection\MemberSerializationStringGenerator.cs

too. There are versions of this file in both CoreCLR and CoreRT.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSourceException.cs
vancem: I believe that the EventSourceException is Serializable because we require all Exceptions to be serializable on the Desktop runtime to support Exceptions traveling across AppDomain boundaries.     

EventSource is unusual, in that at least at one time we tried to keep the source code the same on Desktop and CoreCLR (however because shiproom would wanted only changes that were in response to a particular request to change the desktop, the code bases are NOT in sync today).    Still it may still be useful to keep open the option of a 'bulk update' (I think it is still true that we COULD simply copy the contents to the desktop and it would be 'correct'.   Removing [Serializable] for the desktop build would break that. 

I looked and we don't have a #ifdef that means' Desktop.    The closest thing we have is 'not coreCLR or Project N or not Portable' 

#if  !CORECLR && !ES_BUILD_PN && !ES_BUILD_PCL

This is my recommendation.   Simply surround the [serialable] with this instead.

brianrob: A couple of things to be aware of:

1. This is shared source so it will sync to CoreRT and ProjectN as well.  You'll want to make sure that you have the proper !CORERT ifdef as well if you care about CoreRT.
2. EventSource tests are located in CoreFx.  Before you submit, you should request that CoreFX tests be run against this change.  These are the trigger phrases you want to use (which will automatically request the runs):

morganbr: Thanks. I've added @vancem 's suggested ifdef along with !CORERT.

@brianrob , I've been testing this against CoreFX and modified tests appropriately so that it works. That change was https://github.com/dotnet/corefx/pull/20035.

morganbr:

src/mscorlib/shared/System/Diagnostics/Tracing/EventSourceException.cs
brianrob: Great.  Thanks.

brianrob:
FYI, the perf correctness failures are tracked by https://github.com/dotnet/coreclr/issues/11856.

morganbr:
@brianrob , thanks. I'll go ahead and merge then since previous CIs succeeded and #11856 hasn't been fixed yet.

karelz:
Ported to 2.0.0 in #12020

PR #11767, opened 2017-05-20
Fix package Publish in official builds

gkhanna79:
Package publishing in official builds of master branch was failing for Windows because it looked for packages under AnyOS instead of Windows_NT. Publishing non-Windows works.

It appears that this is because ___BuildOS_ property was not being set for package publishing (https://github.com/dotnet/coreclr/blob/master/config.json#L622) while it was being set for package building (https://github.com/dotnet/coreclr/blob/master/config.json#L660).

Surprisingly, release/2.0.0 Windows builds work even though they have the same issue!

With this change, the publish step appears to be fixed and problem does not repro locally.

@weshaggard @wtgodbe @chcosta Can you PTAL?

### DIFF
--- a/config.json
+++ b/config.json
@@ -622,6 +622,7 @@
       "defaultValues": {
         "toolName": "msbuild",
         "settings": {
+          "__BuildOS": "default",
           "Project": "./src/publish.proj",
           "MsBuildFileLogging": "/flp:v=detailed;LogFile=publish-packages.log"
         }


gkhanna79:
Merging the change since it does not affect CI to unblock official build.

weshaggard:
>Surprisingly, release/2.0.0 Windows builds work even though they have the same issue!

The change looks good but it would be nice to understand why this is unique to master.

gkhanna79:
I agree - @wtgodbe Can you help understand why master needed this and if we need to port this to 2.0.0 as well?

wtgodbe:
How long has this issue been present in master?

gkhanna79:
It seems atleast since after the BT update.

PR #11780, opened 2017-05-22
Use sigInst.methInstCount != 0 to detect generic methods

MichalStrehovsky:
`methInstCount` is the thing that RyuJIT uses to detect generic virtual
methods in another place. `CORINFO_CALLCONV_GENERIC` reporting was added
because nobody noticed `methInstCount`. We already report
`methodInstCount` in CoreRT - we can delete a couple of lines on the
CoreRT side by switching to this scheme.


MichalStrehovsky:
@sandreenko PTAL

Issue #11729, opened 2017-05-18
Fix broken performance test infrastructure


DrewScoggins:
The check-in that moved build tools forward to 2.0, broke all of our performance tests.

AndyAyersMS:
Not sure what specifically you meant, but it looks like the xunit.performance.api got rolled back to 1.0.0-beta-build0003 and so we lost the profile-off ability and other recent improvements.

Consider the following results chart -- profile off was enabled around 5/5 and showed lower data for a time, then around 5/17 it changed and now reports similar numbers to profile on.

![image](https://cloud.githubusercontent.com/assets/10121823/26322861/9ba3f666-3ee1-11e7-81fe-246f4bb59bcf.png)

cc @jorive

RussKeldorph:
@jorive @adiaaida @DrewScoggins Has this been fixed?

RussKeldorph:
Closing stale.  @billwert FYI

PR #11793, opened 2017-05-22
Bump the version of the xUnit-Performance package

jorive:
- In a previous PR this was reverted and broke the Profile On/Off perf runs

### DIFF
--- a/tests/src/Common/external/external.depproj
+++ b/tests/src/Common/external/external.depproj
@@ -21,11 +21,21 @@
     <PackageReference Include="Microsoft.CodeAnalysis.Compilers">
       <Version>1.1.1</Version>
     </PackageReference>
-    <PackageReference Include="xunit.performance.api" Version="1.0.0-beta-build0004" />
-    <PackageReference Include="xunit.performance.core" Version="1.0.0-beta-build0004" />
-    <PackageReference Include="xunit.performance.execution" Version="1.0.0-beta-build0004" />
-    <PackageReference Include="xunit.performance.metrics" Version="1.0.0-beta-build0004" />
-    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="1.0.3-alpha-experimental" />
+    <PackageReference Include="xunit.performance.api">
+      <Version>1.0.0-beta-build0004</Version>
+    </PackageReference>
+    <PackageReference Include="xunit.performance.core">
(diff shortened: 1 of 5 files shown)


tests/src/Common/external/external.depproj
```diff
@@ -21,21 +21,11 @@
...
-    <PackageReference Include="xunit.performance.execution">
-      <Version>1.0.0-beta-build0003</Version>
-    </PackageReference>
-    <PackageReference Include="xunit.performance.metrics">
-      <Version>1.0.0-beta-build0003</Version>
-    </PackageReference>
-    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent">
-      <Version>1.0.3-alpha-experimental</Version>
-    </PackageReference>
+    <PackageReference Include="xunit.performance.api" Version="1.0.0-beta-build0004" />
```
DrewScoggins: nit:  Maybe we should match how the rest of the package references are made.  I don't care if they are all single line style or the old multi-line style.

DrewScoggins:
Once the smoketest performance tests run, we need to check the artifacts and ensure that the Off configurations do not have ETL files or any of the PMC metrics

jorive:
@dotnet/rap-team PTAL

jorive:
This changes fixes #11729

AndyAyersMS:
Do we need to update the reference in the top-level `dependencies.props` ?

Seems wrong that we have this information repeated in so many places in the tree (though admittedly it used to be worse).
```
    <StaticDependency Include="@(XunitPerformanceDependency)">
      <Version>1.0.0-alpha-build0040</Version>
    </StaticDependency>
```

tests/src/Common/external/external.depproj
jorive: Sure, I can do that.

jorive:
@AndyAyersMS I like the idea of cleaning the `dependencies.prop` file, which to be honest, I did not know it existed. It does look out of date too and it would be a good idea to avoid duplicating this information across different files (which is unmanageable).
Having said that, I would prefer deferring it for a future PR.
Thoughts?

DrewScoggins:
@dotnet-bot help

DrewScoggins:

DrewScoggins:
LGTM

AndyAyersMS:
Changes LGTM too, but please follow up with the investigation into `dependencies.props`.

PR #11794, opened 2017-05-22
Remove a couple mistaken calls to popd.

pgavlin:
These calls were causing individual test runs to fail if other
directories were on the stack.


gkhanna79:
@wtgodbe This is the fix for R2R failures.

wtgodbe:
Awesome, thanks Pat!

pgavlin:
The debug failure is a test timeout that we've been seeing occasionally.

Issue #11733, opened 2017-05-18
 System.Linq.Expressions.Tests.BinaryPowerTests.CheckFloatPowerTest(useInterpreter: False) failing on Ubuntu x64 with JitStress=2


pgavlin:
e.g. https://ci.dot.net/job/dotnet_coreclr/job/master/job/jitstress/job/x64_checked_ubuntu_corefx_jitstress2/176/

PR #11795, opened 2017-05-22
Port d520bc1 from master: Do not reverse operands for user-call intrinsics.

pgavlin:
Maintaining reversed execution order for such nodes is problematic when
these intrinsics are rewritten into calls during rationalize.

The original commit fixed issue #11733, which was a case of silent bad code
generation due to the problem described above occurring at a use of `Math.Pow`.


pgavlin:

PR #11807, opened 2017-05-22
Update Unix SafeHandle to throw NotFound correctly (#11757)

JeremyKuhne:
Port for issue https://github.com/dotnet/corefx/issues/20040. CoreFX part is https://github.com/dotnet/corefx/pull/20134.

* Update Unix SafeHandle to throw NotFound correctly

Need to match Windows semantics for missing files. This means throwing
FileNotFound only if the last segment of the path can't be found.

* Dispose handle and trim ending separator properly


JeremyKuhne:
#11757 is the master fix.

PR #11813, opened 2017-05-23
Bump the version of the xUnit-Performance package (#11793)

jorive:
- In a previous PR this was reverted and broke the Profile On/Off perf runs

### DIFF
--- a/tests/src/Common/PerfHarness/PerfHarness.csproj
+++ b/tests/src/Common/PerfHarness/PerfHarness.csproj
@@ -1,11 +1,14 @@
 <Project Sdk="Microsoft.NET.Sdk">
+
   <PropertyGroup>
     <OutputType>exe</OutputType>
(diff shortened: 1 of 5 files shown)


gkhanna79:
Does this fix the perf test failures in https://github.com/dotnet/coreclr/pull/11809?

jorive:
@gkhanna79 No, it does not. This fixes the version of the xUnit-Performance API we use, which enable us to run the benchmarks with and without ETW profiler and do benchmark analysis.

jorive:
@dotnet-bot test Windows_NT x64 perf

jorive:
@dotnet-bot test Windows_NT x86 perf

DrewScoggins:
@dotnet-bot test Windows_NT x64 perf
@dotnet-bot test Windows_NT x86 perf

Issue #11820, opened 2017-05-23
[x86/Linux] In checked mode, TC fails associated with CORINFO_HELP_OVERFLOW


ragmani:
$ ./corerun coreclr-unittest/JIT/Regression/CLR-x86-JIT/V1.2-M01/b00722/b00722/b00722.exe
Segmentation fault (core dumped)

$ ./corerun coreclr-unittest/JIT/Regression/CLR-x86-JIT/V1.2-M01/b04319/b04319/b04319.exe
Segmentation fault (core dumped)

ragmani:
In both TCs, the sementation fault occurs as follows:
```
Program received signal SIGSEGV, Segmentation fault.
0xf7443f98 in CHECK (this=<optimized out>) at /home/parjong/projects/dotnet/coreclr/src/inc/check.h:110
110                   , m_condition (NULL)
(gdb) x/4i $pc
=> 0xf7443f98 <JIT_Overflow()+632>:     movaps %xmm0,-0x98(%ebp)
   0xf7443f9f <JIT_Overflow()+639>:     movl   $0x0,-0x88(%ebp)
   0xf7443fa9 <JIT_Overflow()+649>:     lea    -0x98(%ebp),%esi
   0xf7443faf <JIT_Overflow()+655>:     mov    %esi,(%esp)
(gdb) bt
#0  0xf7443f98 in CHECK (this=<optimized out>) at /home/parjong/projects/dotnet/coreclr/src/inc/check.h:110
#1  JIT_Overflow () at /home/parjong/projects/dotnet/coreclr/src/vm/jithelpers.cpp:5400
#2  0xf4b9f6c1 in ?? ()
#3  0xf4b9f4ba in ?? ()
#4  0xf75286a1 in CallDescrWorkerInternal () from /home/parjong/projects/dotnet/dotnet-overlay/Linux.x86.Checked.Checked/libcoreclr.so
...
```

ragmani:
Here is the related il of b00722:
```
; Assembly listing for method ILGEN_0x195e59cd:Method_0x24dd30a(double,int,long,int,int,byte,int,int,char,ubyte,byte,long,int,double,int):int
; Emitting BLENDED_CODE for generic X86 CPU
; optimized code
; ebp based frame
; partially interruptible
; Final local variable assignments
;
;* V00 arg0         [V00    ] (  0,  0   )  double  ->  zero-ref
;* V01 arg1         [V01    ] (  0,  0   )     int  ->  zero-ref
[...]

PR #11821, opened 2017-05-23
[x86/Linux] fix TC fails associated with CORINFO_HELP_OVERFLOW

ragmani:
add emitting "and esp, 0xfffffff0" instruction before emitting "call CORINFO_HELP_OVERFLOW".
This PR from #11820

### DIFF
--- a/src/jit/codegenxarch.cpp
+++ b/src/jit/codegenxarch.cpp
@@ -5153,6 +5153,12 @@ void CodeGen::genCallInstruction(GenTreeCall* call)
             {
                 fPossibleSyncHelperCall = true;
             }
+#if defined(UNIX_X86_ABI)
+            if (helperNum == CORINFO_HELP_OVERFLOW)
+            {
+                inst_RV_IV(INS_AND, REG_SPBASE, ~(STACK_ALIGN - 1), EA_PTRSIZE);
+            }
+#endif
         }
         else
         {


wateret:
This looks good but may affect stack walking. Could you please run GC stress test? You can run with option `--gcstresslevel=3` for `runtest.sh`.

wateret:
@ragmani I ran GC stress test it looks like there is only a regression.

- JIT/Regression/JitBlue/GitHub_4044/GitHub_4044/GitHub_4044.sh

I guess that this test case causes overflow inside filter funclet.

By the way it is interesting that these two tests pass. These are the ones that I'm working on, I need to take a look further.

- JIT.Methodical.Boxing.seh._relfilter._relfilter
- JIT.Methodical.eh.basics.trythrowexcept_r.trythrowexcept_r

ragmani:
Thank you for testing with GC stress.
By the way, this test case still fails in an environment built with master branch.
- JIT/Regression/JitBlue/GitHub_4044/GitHub_4044/GitHub_4044.sh
The fail seems to be unrelated to this PR.

CarolEidt:
Why is the alignment not handled by the call to `genAlignStackBeforeCall(call)` at the beginning of this method?

ragmani:
The function can not be used because the stack level is not known there.
The value of the stack level is always zero, since the call is called after the jump.
So I filled the stack with the and operation.
Currently, @parjong is looking at another solution.

CarolEidt:
It's still unclear to me why this case wouldn't or shouldn't be handled by `genAlignStackBeforeCall(call)`. In any case, I don't think it's correct to just adjust the stack without making a corresponding change to `genStackLevel`.

parjong:
@CarolEidt The following comment in ``Compiler::fgAddCodeRef`` would explain why ``genAlignStackBeforeCall(call)`` does not work in this case.
```
        // If different range checks happen at different stack levels,
        // they can't all jump to the same "call @rngChkFailed" AND have
        // frameless methods, as the rngChkFailed may need to unwind the
        // stack, and we have to be able to report the stack level.
        //
        // The following check forces most methods that reference an
        // array element in a parameter list to have an EBP frame,
        // this restriction could be removed with more careful code
        // generation for BBJ_THROW (i.e. range check failed).
        //
        // For Linux/x86, we possibly need to insert stack alignment adjustment
        // before the first stack argument pushed for every call. But we
        // don't know what the stack alignment adjustment will be when
        // we morph a tree that calls fgAddCodeRef(), so the stack depth
        // number will be incorrect. For now, simply force all functions with
        // these helpers to have EBP frames. It might be possible to make
        // this less conservative. E.g., for top-level (not nested) calls
        // without stack args, the stack pointer hasn't changed and stack
        // depth will be known to be zero. Or, figure out a way to update
        // or generate all required helpers after all stack alignment
        // has been added, and the stack level at each call to fgAddCodeRef()
        // is known, or can be recalculated.
```
The use of  ``push/pop`` in x86 makes it impossible to compute the exact stack level if there are multiple jumps to this exception throwing block. 

Could you let me know where is the best place to adjust stack level? I find that ``CodeGen::genAdjustStackLevel`` has a corresponding code for ESP frames (with only one jump).

So, is it better to revise ``CodeGen::genAdjustStackLevel`` to restore SP based on FP?

BruceForstall:
Have we agreed that this is an incorrect fix?

It appears another attempt to fix this is #12504

CarolEidt:
> Have we agreed that this is an incorrect fix?

I wasn't comfortable that this was a fix as opposed to a workaround, and I don't have sufficient in-depth knowledge of the stack management to provide more insight.

> It appears another attempt to fix this is #12504

Yes, and it looks much more promising. Thanks @BruceForstall for reviewing it.

ragmani:
I close it because it is reflected in the other PR( #12504 ) more clearly fixed.

PR #11853, opened 2017-05-23
Partially remove relocations from Module section of NGEN-ed images

ruben-ayrapetyan:
Related issue: #10380

@Dmitri-Botcharnikov @gbalykov

### DIFF
--- a/src/debug/daccess/nidump.cpp
+++ b/src/debug/daccess/nidump.cpp
@@ -4470,14 +4470,14 @@ void NativeImageDumper::TraverseNgenHash(DPTR(HASH_CLASS) pTable,
     }
 
     DisplayWriteFieldPointer(m_pModule,
-                             DPtrToPreferredAddr(pTable->m_pModule),
+                             DPtrToPreferredAddr(pTable->GetModule()),
                              HASH_CLASS, MODULE);
 
     // Dump warm (volatile) entries.
     DisplayWriteFieldUInt(m_cWarmEntries, pTable->m_cWarmEntries, HASH_CLASS, MODULE);
     DisplayWriteFieldUInt(m_cWarmBuckets, pTable->m_cWarmBuckets, HASH_CLASS, MODULE);
     DisplayWriteFieldAddress(m_pWarmBuckets,
-                             DPtrToPreferredAddr(pTable->m_pWarmBuckets),
+                             DPtrToPreferredAddr(pTable->GetWarmBuckets()),
                              sizeof(HASH_ENTRY_CLASS*) * pTable->m_cWarmBuckets,
                              HASH_CLASS, MODULE);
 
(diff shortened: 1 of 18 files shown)


ruben-ayrapetyan:
@jkotas, Please, take a look.

jkotas:
Is this change worth it? There is just one of these sections in the image so the impact is low, and unless you remove all relocations from it you are still going to pay the private page for it.

jkotas:
The part of the change for EEClass looks fine.

ruben-ayrapetyan:
@jkotas, Thank you for review.

The relocations, which were not removed, in most cases are located on the same page, as the `Module` structure. The `Module` structure is modified after being loaded, so the page anyway becomes private.

These changes remove relocations for the following fields:
- `LookupMapBase::pTable`
- `LookupMapBase::pIndex`
- `EEClassOptionalFields::m_pVarianceInfo`
- `DelegateEEClass::m_pInvokeMethod`
- `DelegateEEClass::m_pBeginInvokeMethod`
- `DelegateEEClass::m_pEndInvokeMethod`
- `EEClass::m_pGuidInfo`
- `NgenHashTable::m_pModule`
- `NgenHashTable::m_pWarmBuckets`
- `NgenHashTable::m_sHotEntries.m_pEntries`
- `NgenHashTable::m_sHotEntries.m_pBuckets`
- `NgenHashTable::m_sColdEntries.m_pEntries`
- `NgenHashTable::m_sColdEntries.m_pBuckets`
- `TypeVarTypeDesc::m_pModule`

As far as I understand, the several left fields are those, which exist only in debug build, and also the one in release mode - `LookupMapBase::hotItemList`, which seem to be not fixed up (because there is no profile used).

Is it better to also change the `hotItemList` to relative?

Currently, the changes completely remove Module section's relocations from about 80% of pages on set of images, which was referenced in https://github.com/dotnet/coreclr/issues/11363#issuecomment-299180256). However, the section is sometimes located on same page with other sections, so about 25 of the pages can still contain relocations.

On application, which is referenced in https://github.com/dotnet/coreclr/issues/10380#issuecomment-291146637, the following memory consumption changes occur for mappings of images:
- Rss: 8832 -> 8648 (2% improvement)
- Private_Dirty: 3020 -> 2708 (10% improvement)
- Shared: 5352 -> 5416 (64 kilobytes of Private_Dirty moved there)
- Private_Clean:  460 -> 524 (64 kilobytes of Private_Dirty moved there)

// The base for measurements is currently another than was used in #10380
// Fix of #10382 increased count of relocated images, so the absolute values of memory consumption
// became greater, and so absolute difference related to relocations removal also became greater.

jkotas:
> the changes completely remove Module section's relocations from about 80% of pages

I guess you meant something else than `CORCOMPILE_SECTION_MODULE` ... that's what confused me.

> The relocations, which were not removed, in most cases are located on the same page, as the Module structure

Right. And the Module structure will be pretty much always written into anyway, so it should not matter that it has relocations.

LookupMapBase should be inline in the Module structure, and so the relocations inside should not matter either.

ruben-ayrapetyan:
> I guess you meant something else than CORCOMPILE_SECTION_MODULE

I meant `ModuleSection` (`ZapVirtualSectionType`), which includes `CORCOMPILE_SECTION_MODULE` and `CORCOMPILE_SECTION_MODULE_COLD` (`CorCompileSection`).

> LookupMapBase should be inline in the Module structure, and so the relocations inside should not matter either.

There is `LookupMapBase::pNext` field that seems to be used for pointing to non-inlined map. I missed it at first, because there is no `Fixup` call for the field - only `ZeroPointerField`. Is the field always zero in ngen-ed images?
If so, I think it is better to remove changes for the structure, and otherwise its better to update `hotItemList` and `pNext`.
What do you think about this?

jkotas:
> the field always zero in ngen-ed images?

Correct.

>  I think it is better to remove changes for the structure,

Agree.

ruben-ayrapetyan:
@jkotas, Thank you.

The changes for `LookupMapBase` were removed.

jkotas:
Thanks

PR #11854, opened 2017-05-23
StreamingContext Context equality fix

ViktorHofer:
StreamingContext currently compares its Context with an equality operator which isn't wise as most types are implementing an equal override but no equality operator.


stephentoub:
I'm not convinced the original was wrong.  The original is validating that the _additionalContext contains the same object.  This new version is validating that _additionalContext contains an object whose Equals method says it matches.  What makes the new semantic better than current one?

ViktorHofer:
I thought we are defining equality of StreamingContext as equal Context objects. I'm not sure what the guideline is for such wrapper objects. Same applies to Lazy<T> and WeakReference. 

Thinking about it again I would now also tend towards same objects instead of equal objects.

stephentoub:
Given that and that core matches desktop, let's leave it as is.  Thanks.

PR #11864, opened 2017-05-24
[x86/Linux] Thread-safe UMThunkMarshInfo::RunTimeInit

parjong:
It is possible that RunTimeInit changes UMThunkMarshInfo object while SetupArguments is running, which causes assert failure in SetupArguments.

Here is the detailed scenario of how assert failure happens:
 - Thread T1 and T2 invoke RunTimeInit on the same UMThunkMarshInfo object.
 - Thread T1 invokes UMThunkMarshInfo::SetupArguments (after returned from UMThunkMarshInfo::RunTimeInit).
 - Thread T2 updates m_cbStackArgSize as 0 just before RunTimeInit is running.
 - Thread T1 reads m_cbStackArgSize.
 - Thread T1 hits assertion (pDst == pCurDst) failure.

This commit attempts to prevent this potential assertion failure via setting m_cbStackArgSize only after its final value is ready.

### DIFF
--- a/src/vm/dllimportcallback.cpp
+++ b/src/vm/dllimportcallback.cpp
@@ -1382,9 +1382,9 @@ VOID UMThunkMarshInfo::RunTimeInit()
     UINT16 cbRetPop = 0;
 
     //
-    // m_cbStackArgSize represents the number of arg bytes for the MANAGED signature
+    // cbStackArgSize represents the number of arg bytes for the MANAGED signature
     //
-    m_cbStackArgSize = 0;
+    UINT32 cbStackArgSize = 0;
 
     int offs = 0;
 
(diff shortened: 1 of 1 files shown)


parjong:
This seems to resolve FX unittest failure reported in https://github.com/dotnet/coreclr/pull/11165#issuecomment-296847534.

parjong:
JIT_Generics._Coverage_chaos65204782cs_o_chaos65204782cs_o_._Coverage_chaos65204782cs_o_chaos65204782cs_o_cmd failed, but it seems irrelevant.

parjong:

janvorli:
@jkotas - is this race expected or is it an indication of a problem somewhere else? I assume the former, but wanted to be sure.

jkotas:
Yes, it is expected. The EE datastructures generally assume that you get from uninitialized to initialized atomically.

PR #11865, opened 2017-05-24
Disable IEnumerable.GetEnumerator on RCW through IDispatch

yizhang82:
CoreCLR doesn't really support IDispatch and this is probably one of the more obscure cases that we don't block it properly. When calling IEnumerable.GetEnumerator, we will try 3 possibilities (IBindableIterable, IIterable<T>, and IDispatch.Invoke with DISPID_NEWENUM). When we go through the last one we end up going through EnumerableToDispatchMarshaler code (which is implemented in C++/CLI for whatever reason) using ComTypes.IEnumerable and both of these are currently missing. We can revisit IDispatch support at some point in the future but for now it's good to be consistent.

This makes sure we throw a good error message:
 
Exception caught from 'GetEnumerator()': IDispatch and IDispatchEx are not supported
StackTrace:
   at System.Collections.IEnumerable.GetEnumerator()
   at COMTest.Program.Main(String[] args) 

@jkotas PTAL

### DIFF
--- a/src/vm/mngstdinterfaces.cpp
+++ b/src/vm/mngstdinterfaces.cpp
@@ -994,8 +994,12 @@ FCIMPL1(Object*, StdMngIEnumerable::GetEnumerator, Object* refThisUNSAFE)
 
     if (retVal == NULL)
     {
+#ifndef FEATURE_CORECLR
         // classic COM interop scenario
         retVal = ObjectToOBJECTREF((Object*)GetEnumeratorWorker(args));
+#else
+        COMPlusThrow(kPlatformNotSupportedException, IDS_EE_ERROR_IDISPATCH);
+#endif // FEATURE_CORECLR
     }
 
     GCPROTECT_END();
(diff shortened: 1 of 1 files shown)


src/vm/mngstdinterfaces.cpp
```diff
@@ -994,8 +994,12 @@ FCIMPL1(Object*, StdMngIEnumerable::GetEnumerator, Object* refThisUNSAFE)
 
     if (retVal == NULL)
     {
+#ifndef FEATURE_CORECLR
```
jkotas: Delete it instead of ifdefing - we have deleted all FEATURE_CORECLR ifdefs in CoreCLR and the code protected by them.

yizhang82:
sounds good. #ifdef removed. Thanks

danmosemsft:
@yizhang82 what is the status of this one?
@dotnet-bot test this please (stalled)

yizhang82:
@danmosemsft Oops. Forgot about this one. I've tested it and it's good. Once the PR is green I'll merge it. I probably forgot about it because one of the checks stalled and I've decided to wait a bit more...

PR #10772, opened 2017-04-06
Direct-mapping of IL-assembly images without writeable sections

ruben-ayrapetyan:
Before the changes (applies only to the Unix-like platforms):

- If file alignment sections of IL-only assemblies is less than size of virtual page, then the pages are not directly mapped to memory, but instead are copied to anonymous memory.

With the changes (applies only to the Unix-like platforms):

- IL-only assemblies without writeable sections are directly mapped to memory in flat mode (i.e. without separate maps per section).

This means that physical memory for the assemblies is allocated only once with the changes, even if the assemblies are loaded into multiple processes, so memory consumption reduces.

Related issue: #10693 

<br/>
Testing:

- CI
  - Default: Successful
  - Additionally, Ubuntu x64 Checked Build and Test (Jit - CoreFx): Successful 
- CoreCLR test suite: Successful (on Windows, with these changes enabled for testing, although they are supposed to be Unix only)
- Memory consumption: please, see below

Memory consumption of Puzzle.Tizen application (see #10380)

Part | Before | After | Difference
------------ | ------------- | ------------- | -------------
Private (per-application) | 20244 kB | 19552 kB | 692 kB less
Shared (one copy for all applications) | 22132 kB | 22750 kB | 618 kB more
Rss | 42376 kB | 42302 kB | 74 kB less

Per-application part is 692 kB less, which is 10.3% of CoreCLR's per-application memory consumption part.

PR #11867, opened 2017-05-24
[x86/Linux] Reentrant PEImage::Load

### LINKED ISSUE #11866: [x86/Linux] m_pLayouts[IMAGE_LOADED] == NULL assert failure
I observed the following assert failure while running FX unittest:
```

Discovering: System.Dynamic.Runtime.Tests
Discovered:  System.Dynamic.Runtime.Tests
Starting:    System.Dynamic.Runtime.Tests

Assert failure(PID 18638 [0x000048ce], Thread: 18653 [0x48dd]): m_pLayouts[IMAGE_LOADED] == NULL
    File: /var/jenkins/workspace/CLR/x86_Chk_RT/repo/code/src/vm/peimage.cpp Line: 1107
    Image: /var/jenkins/workspace/FX/FX_x86_Checked_Test/working/overlay/dotnet
```

parjong:
This commit resivse PEImage::Load as a re-entrant method in order to fix #11866.

### DIFF
--- a/src/vm/peimage.cpp
+++ b/src/vm/peimage.cpp
@@ -1096,6 +1096,7 @@ void PEImage::Load()
 {
     STANDARD_VM_CONTRACT;
 
+    // Performance optimization to avoid lock acquisition
     if (HasLoadedLayout())
     {
         _ASSERTE(GetLoadedLayout()->IsMapped()||GetLoadedLayout()->IsILOnly());
@@ -1104,7 +1105,9 @@ void PEImage::Load()
 
     SimpleWriteLockHolder lock(m_pLayoutLock);
 
-    if (m_pLayouts[IMAGE_LOADED] != NULL)
+    // Re-check after lock is acquired as HasLoadedLayout here and the above line
+    // may return a different value in multi-threading environment.
+    if (HasLoadedLayout())
     {
         return;
     }
--- a/src/vm/peimage.cpp
+++ b/src/vm/peimage.cpp
@@ -1104,7 +1104,10 @@ void PEImage::Load()
 
     SimpleWriteLockHolder lock(m_pLayoutLock);
 
-    _ASSERTE(m_pLayouts[IMAGE_LOADED] == NULL);
+    if (m_pLayouts[IMAGE_LOADED] != NULL)
+    {
+        return;
+    }
 
 #ifdef PLATFORM_UNIX
     if (m_pLayouts[IMAGE_FLAT] != NULL


gkhanna79:
Is this a regression from the mapped layout changes made few weeks ago?

CC @janvorli

parjong:
#10772 seems to be related.

jkotas:
cc @ruben-ayrapetyan

parjong:

src/vm/peimage.cpp
```diff
@@ -1104,7 +1104,10 @@ void PEImage::Load()
 
     SimpleWriteLockHolder lock(m_pLayoutLock);
 
-    _ASSERTE(m_pLayouts[IMAGE_LOADED] == NULL);
+    if (m_pLayouts[IMAGE_LOADED] != NULL)
+    {
+        return;
+    }
```
ruben-ayrapetyan: The `HasLoadedLayout` called above is the following:
```inline BOOL PEImage::HasLoadedLayout()                                                                                 
{                                                                                                                      
    LIMITED_METHOD_CONTRACT;                                                                                           
    SUPPORTS_DAC;                                                                                                      
    return m_pLayouts[IMAGE_LOADED]!=NULL;                                                                             
}
```

It actually performs the same check.

I think there is a multithreading race condition for value of the `m_pLayouts[IMAGE_LOADED]`.

So, moving the lock acquisition above the `HasLoadedLayout` check should also fix the issue.

This implementations is also correct, and can be a performance optimization, because it doesn't perform lock acquisition if layout is already loaded.

I propose to implement the new `if (m_pLayouts[IMAGE_LOADED] != NULL)` with `if (HasLoadedLayout())`, and to add a comment which describes that this is a performance optimization, like the following:

```
void PEImage::Load()
{
    STANDARD_VM_CONTRACT;

    if (HasLoadedLayout())
    {
        _ASSERTE(GetLoadedLayout()->IsMapped()||GetLoadedLayout()->IsILOnly());
        return;
    }

    SimpleWriteLockHolder lock(m_pLayoutLock);

    // The condition could be changed before lock was acquired, so repeating the check
    // This is a performance optimization to avoid acquiring lock in case layout is already loaded
    if (HasLoadedLayout())
    {
       return;
    }

   _ASSERTE(m_pLayouts[IMAGE_LOADED] == NULL);
```

or to move the lock acquisition above the first check.

jkotas: Well, the discussion about the memory models are always fun. What can happen in theory vs. what happens in practice.

> In case, compiler can prove this, the value can be cached, which can break semantics of this code

There is no way for the compiler to prove this. Also, the lock has barriers in it - even if the compiler was somehow able to reason about the whole lock code, the barriers inside the lock have to prevent it.

If you are worried about this, you can fetch the value outside the lock using `VolatileLoad` that should prevent any dangerous optimizations that may happen here in theory.

ruben-ayrapetyan: @jkotas, Thank you for explanation.
So, the `Volatile<LONG> m_RWLock;` in `SimpleRWLock` actually have barrier semantics. I incorrectly considered it as standard `volatile` value, while actually it is special implementation with additional barrier semantics.

jkotas:

PR #11875, opened 2017-05-24
Fix compile with disabled FEATURE_READYTORUN option

BredPet:
Signed-off-by: Petr Bred <bredpetr@gmail.com>

### DIFF
--- a/src/vm/eventpipe.cpp
+++ b/src/vm/eventpipe.cpp
@@ -107,7 +107,7 @@ void EventPipe::Shutdown()
 
 void EventPipe::Enable(
     LPCWSTR strOutputPath,
-    uint circularBufferSizeInMB,
+    unsigned int circularBufferSizeInMB,
     EventPipeProviderConfiguration *pProviders,
     int numProviders)
 {
--- a/src/inc/corcompile.h
+++ b/src/inc/corcompile.h
@@ -1295,11 +1295,13 @@ class ICorCompilePreloader
             CORINFO_METHOD_HANDLE method, 
             CORINFO_METHOD_HANDLE duplicateMethod) = 0;
 
+#ifdef FEATURE_READYTORUN_COMPILER
     // Returns a compressed encoding of the inline tracking map 
     // for this compilation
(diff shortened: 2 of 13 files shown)


BredPet:
@sergign60 
@Dmitri-Botcharnikov 
@jkotas please review

Documentation/building/linux-instructions.md
```diff
@@ -237,3 +237,72 @@ If you want to focus on the speed optimization for high-end devices, you have to
  SET (CLR_DEFINES_DEBUG_INIT              DEBUG _DEBUG _DBG URTBLDENV_FRIENDLY=Checked BUILDENV_
 ```
 
+Additional configuring features for ARM/Linux
+=============================================
+
+Disable READYTORUN
+------------------
```
jkotas: I do not think this should be documented here like this. It will get out of date very quickly. If you have a good reason for compiling with this disabled often, make it option for the CMake scripts.

src/vm/eventpipe.cpp
```diff
@@ -107,7 +107,7 @@ void EventPipe::Shutdown()
 
 void EventPipe::Enable(
     LPCWSTR strOutputPath,
-    uint circularBufferSizeInMB,
+    unsigned int circularBufferSizeInMB,
```
jkotas: This looks unrelated to the change. Could you please submit separate PR for it?

Documentation/building/linux-instructions.md
BredPet: Ok

src/vm/eventpipe.cpp
BredPet: The chain is 
`eventpipe.cpp -> common.h -> ceeload.h -> (IF FEATURE_READYTORUN -> readytoruninfo.h -> nativeformatreader.h -> uint)`
Of course, I can still separate, but it seems to me that this is a direct binding.

jkotas: I see. It was not obvious to me - it is fine to keep as part of this.

BredPet:
@jkotas please review

PR #11876, opened 2017-05-24
Merge remaining common arm32/arm64 lowering functions

hqueue:
- Merge `Lowering::IsCallTargetInRange(void* addr)`
- Merge `Lowering::IsContainableImmed(GenTree* parentNode, GenTree* childNode)`
- Define `Lowering::isRMWRegOper(GenTreePtr tree)` for arm32 which is never defined and used yet

After this commit, lowerarm.cpp and lowerarm64.cpp contains only `isRMWRegOper()` which is never used in arm32/arm64. We can now remove lowerarm.cpp and lowerarm64.cpp completely later if these files are not necessary anymore.


hqueue:
Although we can already remove lowerarm.cpp and lowerarm64.cpp, I just leave it, because ryujit/arm32 is being implemented and we don't know whether we will need them again or not.

What do you think of it ? If you think we can remove them, I will update this PR.

BruceForstall:

BruceForstall:
cc @dotnet/arm64-contrib @dotnet/arm32-contrib

src/jit/lowerarmarch.cpp
```diff
@@ -30,6 +30,123 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
...
+    // Currently, we'll be generating calls using blr and manually loading an absolute
+    // call target in a register using a sequence of load immediate instructions.
+    //
+    // As you can expect, this is inefficient and it's not the recommended way as per the
+    // ARM64 ABI Manual but will get us getting things done for now.
+    // The work to get this right would be to implement PC-relative calls, the bl instruction
+    // can only address things -128 + 128MB away, so this will require getting some additional
+    // code to get jump thunks working.
+    return true;
+#elif _TARGET_ARM_
```
BruceForstall: This should be
```
#elif defined(_TARGET_ARM_)
```

src/jit/lowerarmarch.cpp
```diff
@@ -30,6 +30,123 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
...
+
+            case GT_EQ:
+            case GT_NE:
+            case GT_LT:
+            case GT_LE:
+            case GT_GE:
+            case GT_GT:
+                if (childNode->IsIntegralConst(0))
+                {
+#ifdef _TARGET_ARM_
```
BruceForstall: No need for `#ifdef` since the `NYI_ARM` already functions as an `#ifdef`

src/jit/lowerarmarch.cpp
```diff
@@ -30,6 +30,123 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
...
+        switch (parentNode->OperGet())
+        {
+            default:
+                return false;
+
+            case GT_ADD:
+            case GT_SUB:
+#ifdef _TARGET_ARM64_
+                if (emitter::emitIns_valid_imm_for_add(immVal, size))
+#elif _TARGET_ARM_
```
BruceForstall: => `#elif defined(_TARGET_ARM_)`

src/jit/lowerarmarch.cpp
```diff
@@ -30,6 +30,123 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
...
+                if (emitter::emitIns_valid_imm_for_cmp(immVal, size))
+                    return true;
+                break;
+#endif
+            case GT_AND:
+            case GT_OR:
+            case GT_XOR:
+#ifdef _TARGET_ARM64_
+                if (emitter::emitIns_valid_imm_for_alu(immVal, size))
+#elif _TARGET_ARM_
```
BruceForstall: again =>`#elif defined`

hqueue:
@dotnet-bot please test Tizen armel Cross Release Build

hqueue:
@dotnet-bot please test Windows_NT x64 Release Priority 1 Build and Test

src/jit/lowerarmarch.cpp
```diff
@@ -30,6 +30,122 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
...
+            default:
+                return false;
+
+            case GT_ADD:
+            case GT_SUB:
+#ifdef _TARGET_ARM64_
+                if (emitter::emitIns_valid_imm_for_add(immVal, size))
+#elif defined(_TARGET_ARM_)
+                if (emitter::emitIns_valid_imm_for_add(immVal, INS_FLAGS_DONT_CARE))
+#endif
```
pgavlin: I find this sort of ifdefing pretty hard to read. Can you change this to something like
```
#ifdef _TARGET_ARM64_
                return emitter::emitIns_valid_imm_for_add(immVal, size);
#elif defined(_TARGET_ARM_)
                return emitter::emitIns_valid_imm_for_add(immVal, INS_FLAGS_DONT_CARE);
#endif
```

pgavlin:
I pretty strongly dislike the way that the `#ifdefs` in `Lowering::IsContainableImmed` affect the control flow, especially [here](https://github.com/dotnet/coreclr/pull/11876/files#diff-68b10496f6119a8a5fbaf9f51a633787R121), where the cases fall through on one architecture but do not fall though on the other. IMO it would be better to leave these as independent functions.

You could try refactoring the common bits into a new function--perhaps named `Lowering::isARMContainableImmed` to which the architecture-dependent functions delegate if they do not make an architecture-dependent decision.

src/jit/lowerarmarch.cpp
hqueue: @pgavlin Looks better :) I updated this and similar other cases too.

hqueue:
@pgavlin Thank you for the feedback.

> I pretty strongly dislike the way that the #ifdefs in Lowering::IsContainableImmed affect the control flow, especially here, where the cases fall through on one architecture but do not fall though on the other. IMO it would be better to leave these as independent functions.

I also agree that it is not a good idea to introduce different flow depending on architecture.
I had similar thought in mind when merging this.

> You could try refactoring the common bits into a new function--perhaps named Lowering::isARMContainableImmed to which the architecture-dependent functions delegate if they do not make an architecture-dependent decision.

I'm afraid introducing a new function to delegate architecture-dependent logic might be not desirable here, because I think the purpose of merging code to lowerarmarch.cpp is to expose the similarity of arm32/arm64 at the same place.


I've three alternate refactoring in mind. 
How about below refactoring ? 

Suggetion 1: Group all cases  for each architecture. (Bogus, but the differences can be found at a glance.)
```
#ifdef _TARGET_ARM64_
            case GT_EQ:
            case GT_NE:
            case GT_LT:
            case GT_LE:
            case GT_GE:
            case GT_GT:
                return emitter::emitIns_valid_imm_for_cmp(immVal, size);
                break;
            case GT_AND:
            case GT_OR:
            case GT_XOR:
                return emitter::emitIns_valid_imm_for_alu(immVal, size);
                break;
#elif defined(_TARGET_ARM_)
            case GT_EQ:
            case GT_NE:
            case GT_LT:
            case GT_LE:
            case GT_GE:
            case GT_GT:
            case GT_AND:
            case GT_OR:
            case GT_XOR:
                return emitter::emitIns_valid_imm_for_alu(immVal);
                break;
#endif
```

Suggetion 2: Also seperate cases for `_TARGET_ARM_`
```

            case GT_EQ:
            case GT_NE:
            case GT_LT:
            case GT_LE:
            case GT_GE:
            case GT_GT:
#ifdef _TARGET_ARM64_
                return emitter::emitIns_valid_imm_for_cmp(immVal, size);
#elif defined(_TARGET_ARM_)
                return emitter::emitIns_valid_imm_for_alu(immVal);             // Split cases for _TARGET_ARM_  too.
#endif
                break;
            case GT_AND:
            case GT_OR:
            case GT_XOR:
#ifdef _TARGET_ARM64_
                return emitter::emitIns_valid_imm_for_alu(immVal, size);
#elif defined(_TARGET_ARM_)
                return emitter::emitIns_valid_imm_for_alu(immVal);
#endif
                break;
```



Suggetion 3: Use __fallthrough.
```

            case GT_EQ:
            case GT_NE:
            case GT_LT:
            case GT_LE:
            case GT_GE:
            case GT_GT:
#ifdef _TARGET_ARM64_
                return emitter::emitIns_valid_imm_for_cmp(immVal, size);
                break;
#elif defined(_TARGET_ARM_)
                __fallthrough;
#endif

            case GT_AND:
            case GT_OR:
            case GT_XOR:
#ifdef _TARGET_ARM64_
                return emitter::emitIns_valid_imm_for_alu(immVal, size);
#elif defined(_TARGET_ARM_)
                return emitter::emitIns_valid_imm_for_alu(immVal);
#endif
                break;
```

What do you think of these ?

hqueue:
@dotnet-bot please test OSX10.12 x64 Checked Build and Test

pgavlin:
> What do you think of these ?

Personally, I prefer option 1 as I find it by far the easiest to read, but I would accept option 2 if others strongly prefer its density. I would not accept option 3.

hqueue:
Considering readability, I also prefer option 1.
I will update this PR today later if there is no other opinion.

hqueue:
@dotnet/arm64-contrib Please trigger arm64 CI for this PR. Thanks.

RussKeldorph:

hqueue:
@RussKeldorph Thank you for triggering CI. However `Windows_NT arm64 Cross Checked Build and Test` (https://ci.dot.net/job/dotnet_coreclr/job/master/job/arm64_cross_checked_windows_nt_prtest/244/) failed due to timeout. Can you please take a look and trigger the CI again?

jashook:

PR #11878, opened 2017-05-24
Make non-tracked lclVars contained

CarolEidt:
Make untracked lclVars contained where possible.
This causes numerous diffs, which are overwhelmingly positive, with a 1.84% improvement in jit-diff --test for x64 Windows. A few, however, are regressions in both the number and size of instructions. These are cases where an immediate is loaded into a register, which is then the target of a read-modify-write instruction. Because this modifies the register that held the immediate, it is no longer available for reuse, and in some cases this causes an extra instruction to be generate.

### DIFF
--- a/src/jit/codegenxarch.cpp
+++ b/src/jit/codegenxarch.cpp
@@ -714,12 +714,6 @@ void CodeGen::genCodeForDivMod(GenTreeOp* treeNode)
     genConsumeOperands(treeNode->AsOp());
     if (varTypeIsFloating(targetType))
     {
-        // Check that divisor is a valid operand.
-        // Note that a reg optional operand is a treated as a memory op
-        // if no register is allocated to it.
-        assert(divisor->isUsedFromReg() || divisor->isMemoryOp() || divisor->IsCnsFltOrDbl() ||
-               divisor->IsRegOptional());
-
         // Floating point div/rem operation
         assert(oper == GT_DIV || oper == GT_MOD);
 
@@ -829,7 +823,8 @@ void CodeGen::genCodeForBinary(GenTree* treeNode)
     if (!op1->isUsedFromReg())
     {
         assert(treeNode->OperIsCommutative());
-        assert(op1->isMemoryOp() || op1->IsCnsNonZeroFltOrDbl() || op1->IsIntCnsFitsInI32() || op1->IsRegOptional());
+        assert(op1->isMemoryOp() || op1->IsLocal() || op1->IsCnsNonZeroFltOrDbl() || op1->IsIntCnsFitsInI32() ||
+               op1->IsRegOptional());
 
         op1 = treeNode->gtGetOp2();
         op2 = treeNode->gtGetOp1();
--- a/src/jit/lower.cpp
+++ b/src/jit/lower.cpp
@@ -104,6 +104,26 @@ bool Lowering::IsSafeToContainMem(GenTree* parentNode, GenTree* childNode)
 }
 
 //------------------------------------------------------------------------
+// IsContainableMemoryOp: Checks whether this is a memory op that can be contained.
+//
+// Arguments:
+//    node - the node of interest.
+//
+// Notes:
+//    This differs from the isMemoryOp() method on GenTree because it checks for
+//    the case of an untracked local. Note that this won't include locals that
+//    for some reason do not become register candidates, nor those that get
+//    spilled.
+//
+// Return value:
+//    True if this will definitely be a memory reference that could be contained.
+//
+bool Lowering::IsContainableMemoryOp(GenTree* node)
+{
+    return node->isMemoryOp() || (node->IsLocal() && !comp->lvaTable[node->AsLclVar()->gtLclNum].lvTracked);
+}
+
+//------------------------------------------------------------------------
 
 // This is the main entry point for Lowering.
 GenTree* Lowering::LowerNode(GenTree* node)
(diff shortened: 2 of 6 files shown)


CarolEidt:
@dotnet-bot help

CarolEidt:
@dotnet-bot test Windows_NT x64 corefx_minopts

CarolEidt:
@dotnet/jit-contrib PTAL

CarolEidt:
BTW, it should be noted that the improvements are quite varied, and much lower on benchmarks (.18%) or just SPC.dll, which was just .22%. . The test results are probably unduly reflective of some big improvements (up to 15%) on very large tests.
There were a couple of benchmarks that showed notable improvement:
```
11 total files with size differences (11 improved, 0 regressed), 43 unchanged.
         -34 : BenchF\InvMt\InvMt\InvMt.dasm - Benchstone.BenchF.InvMt:Bench():bool
         -19 : BenchmarksGame\revcomp\revcomp\revcomp.dasm - BenchmarksGame.Revcomp:BenchInner(bool,ref)
         -18 : BenchF\Bisect\Bisect\Bisect.dasm - Benchstone.BenchF.Bisect:Bench():bool
         -15 : BenchF\MatInv4\MatInv4\MatInv4.dasm - Benchstone.BenchF.MatInv4:Bench():bool
         -10 : BenchF\Regula\Regula\Regula.dasm - Benchstone.BenchF.Regula:Bench():bool
12 total methods with size differences (12 improved, 0 regressed), 404 unchanged.
```

src/jit/codegenxarch.cpp
```diff
@@ -718,7 +718,7 @@ void CodeGen::genCodeForDivMod(GenTreeOp* treeNode)
         // Note that a reg optional operand is a treated as a memory op
         // if no register is allocated to it.
         assert(divisor->isUsedFromReg() || divisor->isMemoryOp() || divisor->IsCnsFltOrDbl() ||
-               divisor->IsRegOptional());
+               divisor->IsRegOptional() || divisor->IsLocal());
```
briansull: Those are all of the kinds that are allowed, what is left  that is not allowed?
Is there anyway to assert that we don't have that?

src/jit/lower.cpp
```diff
@@ -104,6 +104,34 @@ bool Lowering::IsSafeToContainMem(GenTree* parentNode, GenTree* childNode)
...
+//    True if this will definitely be a memory reference that could be contained.
+//
+bool Lowering::IsContainableMemoryOp(GenTree* node)
+{
+    if (node->isMemoryOp())
+    {
+        return true;
+    }
+    if (!node->IsLocal())
+    {
```
briansull: Should this be written this way instead?
(To me it reads better)
```
if (node->IsLocal())
{
   // When we have a Local Var or Fld, return true for non-tracked variables, since they always live on the stack
   return !comp->lvaTable[node->AsLclVar()->gtLclNum].lvTracked;
}
return false;
```

src/jit/codegenxarch.cpp
CarolEidt: That is a very interesting question - I suspect that, at this point, this is really a useless assert. I'll look into it a bit.

src/jit/lower.cpp
CarolEidt: Will do.

src/jit/codegenxarch.cpp
CarolEidt: I believe that this is indeed exhaustive. If it has a register, the first condition will be true. Otherwise, if it is a legal operand to divide, it will be one of:
- a spilled tree temp (in which case `isMemoryOp()` will be true
- an indirection or `GT_CLS_VAR` (in which case `isMemoryOp()` will be true
- a `RegOptional()` memory operand
- a local

I can't think of any other case - so I think this is a useless assert.

CarolEidt:
@pgavlin - could you have a look at this?

pgavlin:
Yes--I will look at this today.

src/jit/codegenxarch.cpp
```diff
@@ -714,12 +714,6 @@ void CodeGen::genCodeForDivMod(GenTreeOp* treeNode)
     genConsumeOperands(treeNode->AsOp());
     if (varTypeIsFloating(targetType))
     {
-        // Check that divisor is a valid operand.
-        // Note that a reg optional operand is a treated as a memory op
-        // if no register is allocated to it.
-        assert(divisor->isUsedFromReg() || divisor->isMemoryOp() || divisor->IsCnsFltOrDbl() ||
```
pgavlin: Why was this assert removed?

src/jit/lower.cpp
```diff
@@ -104,6 +104,36 @@ bool Lowering::IsSafeToContainMem(GenTree* parentNode, GenTree* childNode)
...
+        return true;
+    }
+    if (node->IsLocal())
+    {
+        // When we have a Local Var or Fld, return true for non-tracked variables,
+        // since they always live on the stack
+        return !comp->lvaTable[node->AsLclVar()->gtLclNum].lvTracked;
+    }
+    return false;
```
pgavlin: This could just be `return node->IsLocal() && !comp->lvaTable[node->AsLclVar()->gtLclNum].lvTracked;`

pgavlin: (or even `return node->isMemoryOp() || (node->IsLocal() && !comp->lvaTable[node->AsLclVar()->gtLclNum].lvTracked)`)

src/jit/codegenxarch.cpp
briansull: See outdated comments above or this link:
https://github.com/dotnet/coreclr/pull/11878#discussion_r118364395

pgavlin: Ah, I see it now, and I agree that the assert appears to accept any operand type. Thanks for the link!

PR #11880, opened 2017-05-24
[RyuJIT/arm32] Fix genRegSetToCond() for FP case

mskvortsov:
Makes use of more generic genJumpKindsForTree() instead of integer-only genJumpKindForOper().

Fixes `JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b63743`.

cc @dotnet/arm32-contrib


BruceForstall:

Issue #11831, opened 2017-05-23
Assert failure(PID 7660 [0x00001dec], Thread: 8932 [0x22e4]): Assertion failed 'type == lvaTable[lnum].lvType' 


RussKeldorph:
New assertion since Friday in several legs, e.g. https://ci.dot.net/job/dotnet_coreclr/job/master/job/jitstress/job/x64_checked_windows_nt_jitstress1/182/

```
D:\j\workspace\x64_checked_w---02832afb\bin\tests\Windows_NT.x64.Checked\TestWrappers\JIT.Directed\JIT.Directed.XUnitWrapper.cs(21737): error : JIT_Directed._nullabletypes_castclassvaluetype_ro_castclassvaluetype_ro_._nullabletypes_castclassvaluetype_ro_castclassvaluetype_ro_cmd [FAIL] [D:\j\workspace\x64_checked_w---02832afb\tests\runtest.proj]
        
        Assert failure(PID 7660 [0x00001dec], Thread: 8932 [0x22e4]): Assertion failed 'type == lvaTable[lnum].lvType' in 'Helper:Compare(struct,struct):bool' (IL size 8)
        
            File: d:\j\workspace\x64_checked_w---02832afb\src\jit\gentree.cpp Line: 6954
            Image: D:\j\workspace\x64_checked_w---02832afb\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\CoreRun.exe
        
        
  
  Return code:      1
```

RussKeldorph:
@dotnet/jit-contrib

pgavlin:
As in https://github.com/dotnet/corefx/pull/20025, the culprit here is https://github.com/dotnet/coreclr/pull/10453. @JosephTremoulet, can you take a look?

stephentoub:
Note that this is blocking CoreFx taking an updated CoreCLR.

PR #11883, opened 2017-05-24
Update full-struct references to promoted IBR args

### LINKED ISSUE #11408: [RyuJIT] TailcallStress needs to check more conditions
When `COMPlus_TailcallStress=1` is set, the JIT converts as many call/ret pairs as possible into tail call, which it treats as `tail.` prefixed "explicit" tail calls. It does not in this case run the legality checks that are run when checking to determine if implicit tailcall optimization can be done, specifically the code under `FEATURE_TAILCALL_OPT` in fgMorphCall() that rejects a potential tail call if any variable in the function has had its address taken, if there are pinned locals, or if there are struct promoted struct arguments.

This causes code such as this to fail:
```
    unsafe class Program
    {
        static void foo(IntPtr x)
        {
            Console.WriteLine(*(int*)x);
        }
        {
            int x = 42;
            foo((IntPtr)(&x));
        }
    }
```

### LINKED ISSUE #11814: Silent bad codegen when returning a modified struct
Repro:

Compile MutateStructArg.cs below:

.\Tools\dotnetcli\dotnet.exe Tools\csc.exe /o+ /out:Program.exe MutateStructArg.cs /r:.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Private.CoreLib.dll /r:.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Console.dll
/r:.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Runtime.dll

MutateStructArg.cs

```
using System;
class MutateStructArg
{
   public struct P
   {
      public string S;
      public int    X;
   }

   public static int Main()
   {
      P l1 = new P();
      l1.S = "Hello World";
      l1.X = 42;
      P l2 = foo(l1);
      Console.WriteLine(l2.S); // Print modified value "Goodbye World"
      if ((l2.S == "Goodbye World") && (l2.X == 100))
      {
         return 100;   // success
      }
      else
      {
         Console.WriteLine("**** Test FAILED ***");
         return 1;  // failure
      }
   }

   public static P foo(P a)
   {
      Console.WriteLine(a.S);   // Print the incoming value "Hello World"
      a.S = "Goodbye World";    // Mutate the incoming value
      a.X = 100;
      return a;                 // Copy the modified value to the return value 
   }
}
```
Run the Repro:
.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\CoreRun.exe MutateStructArg.exe
Hello World
Hello World
**** Test FAILED ***

JosephTremoulet:
Update fgMorphImplicitByRefArgs to rewrite references to struct-promoted
implicit-by-reference parameters as references to the new struct temps
created in fgRetypeImplicitByRefArgs; fgMorphStructField isn't updating
these because it's only interested in field references, and runs upstream
of fgRetypeImplicitByRefArgs where the full struct temp is created,
anyway.
Invert the sense of lvPromoted for implicit byref args in the interim
between fgRetypeImplicitByRefArgs and fgMarkDemotedImplicitByRefArgs,
since now fgMarkDemotedImplicitByRefArgs needs to update both and would
look horribly backwards the other way around.

Fixes #11814.

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -17974,11 +17974,13 @@ void Compiler::fgMarkDemotedImplicitByRefArgs()
         {
             if (varDsc->lvPromoted)
             {
-                // The parameter is simply a pointer now, so clear lvPromoted.
+                // The parameter is simply a pointer now, so clear lvPromoted.  It was left set
+                // by fgRetypeImplicitByRefArgs to communicate to fgMorphImplicitByRefArgs that
+                // appearances of this arg needed to be rewritten to a new promoted struct local.
                 varDsc->lvPromoted = false;
 
                 // Clear the lvFieldLclStart value that was set by fgRetypeImplicitByRefArgs
-                // to tell fgMorphImplicitByRefArgs how to rewrite appearances of this arg.
+                // to tell fgMorphImplicitByRefArgs which local is the new promoted struct one.
                 varDsc->lvFieldLclStart = 0;
             }
             else if (varDsc->lvFieldLclStart != 0)
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -17896,19 +17896,32 @@ void Compiler::fgRetypeImplicitByRefArgs()
                     fieldVarDsc->lvPrefReg = 0;
                 }
 
-                if (undoPromotion)
-                {
-                    // Hijack lvFieldLclStart to record the new temp number.
-                    // It will get fixed up in fgMarkDemotedImplicitByRefArgs.
-                    varDsc->lvFieldLclStart = newLclNum;
-                }
-                else
-                {
-                    // Unmark the parameter as promoted (it's a pointer now).
-                    varDsc->lvPromoted      = false;
-                    varDsc->lvFieldCnt      = 0;
-                    varDsc->lvFieldLclStart = 0;
-                }
+                // Hijack lvFieldLclStart to record the new temp number.
+                // It will get fixed up in fgMarkDemotedImplicitByRefArgs.
+                varDsc->lvFieldLclStart = newLclNum;
+                // Go ahead and clear lvFieldCnt -- either we're promoting
+                // a replacement temp or we're not promoting this arg, and
+                // in either case the parameter is now a pointer that doesn't
+                // have these fields.
+                varDsc->lvFieldCnt = 0;
+
+                // Hijack lvPromoted to communicate to fgMorphImplicitByRefArgs
+                // whether references to the struct should be rewritten as
+                // indirections off the pointer (not promoted) or references
+                // to the new struct local (promoted).
+                varDsc->lvPromoted = !undoPromotion;
+            }
+            else
+            {
+                // The "undo promotion" path above clears lvPromoted for args that struct
+                // promotion wanted to promote but that aren't considered profitable to
+                // rewrite.  It hijacks lvFieldLclStart to communicate to
+                // fgMarkDemotedImplicitByRefArgs that it needs to clean up annotations left
+                // on such args for fgMorphImplicitByRefArgs to consult in the interim.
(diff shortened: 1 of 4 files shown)


JosephTremoulet:
@briansull PTAL.
/cc @dotnet/jit-contrib, @stephentoub 

I've verified that this fixes the repro from dotnet/corefx#20025 in addittion to #11814, though it doesn't fix #11831

JosephTremoulet:

src/jit/morph.cpp
```diff
@@ -17896,19 +17896,32 @@ void Compiler::fgRetypeImplicitByRefArgs()
...
+                // Hijack lvFieldLclStart to record the new temp number.
+                // It will get fixed up in fgMarkDemotedImplicitByRefArgs.
+                varDsc->lvFieldLclStart = newLclNum;
+                // Go ahead and clear lvFieldCnt -- either we're promoting
+                // a replacement temp or we're not promoting this arg, and
+                // in either case the parameter is now a pointer that doesn't
+                // have these fields.
+                varDsc->lvFieldCnt = 0;
+
+                // Hijack lvPromoted to communicate to fgMorphImplicitByRefArgs
```
CarolEidt: Please add comments on these fields in compiler.h, even though these are transient.

src/jit/morph.cpp
```diff
@@ -17958,45 +17970,58 @@ void Compiler::fgMarkDemotedImplicitByRefArgs()
...
-            // The temp struct is now unused; set flags appropriately so that we
-            // won't allocate space for it on the stack.
-            LclVarDsc* structVarDsc     = &lvaTable[structLclNum];
-            structVarDsc->lvRefCnt      = 0;
-            structVarDsc->lvAddrExposed = false;
+        if (lvaIsImplicitByRefLocal(lclNum))
+        {
+            if (varDsc->lvPromoted)
+            {
+                // The parameter is simply a pointer now, so clear lvPromoted.
```
CarolEidt: This is something that could be confusing if viewed in isolation. You might also mention here that this field has been used temporarily to communicate between `fgRetypeImplicitByRefArgs()` and this method.

src/jit/morph.cpp
```diff
@@ -17896,19 +17896,32 @@ void Compiler::fgRetypeImplicitByRefArgs()
...
+            {
+                // The "undo promotion" path above clears lvPromoted for args that struct
+                // promotion wanted to promote but that aren't considered profitable to
+                // rewrite.  It hijacks lvFieldLclStart to communicate to
+                // fgMarkDemotedImplicitByRefArgs that it needs to clean up annotations left
+                // on such args for fgMorphImplicitByRefArgs to consult in the interim.
+                // Here we have an arg that was simply never promoted, so make sure it doesn't
+                // have nonzero lvFieldLclStart, since that would confuse fgMorphImplicitByRefArgs
+                // and fgMarkDemotedImplicitByRefArgs.
+                assert(varDsc->lvFieldLclStart == 0);
```
CarolEidt: This is really applicable to the previous checkin, but I'm wondering why you're not using `BAD_VAR_NUM` for these, since 0 is a valid varNum (not that it could ever be a valid `lvFieldLclStart`)

src/jit/morph.cpp
pgavlin: 👍

src/jit/morph.cpp
JosephTremoulet: That's just the way it gets initialized when the node is created, the ones with 0 for `lvFieldLclStart` are the ones I'm not touching at all.

CarolEidt: Ah I see. Thanks.

JosephTremoulet:
Comments updated.

Issue #11861, opened 2017-05-23
Issues found when building with MSVC + /permissive


xiangfan-ms:
The C++ compiler team is currently doing daily RWC (Read World Code) testing which includes CoreCLR.
We are also working with internal partners and the community to fix source issues exposed by "/permissive-".
The option isn’t finalized yet (we plan to add some large impact features into it in the incoming VS update), so we don’t suggest to adopt it yet.
In the meantime, we want to address these source issues as early as possible so that the community can adopt it when it is finalized.
We discover some source issues when we build it with "/permissive-".

Here is a simplified repro:

	#include <type_traits>

	template <typename T>
	struct S2 {
		T *GetValue();
	};

	namespace clr
	{
	#if 0
		template <typename ItfT>
		static inline
			typename std::enable_if< std::is_pointer<ItfT>::value, ItfT >::type
			SafeAddRef(ItfT pItf);
	#endif

		template < typename ItfT >
		ItfT *
			SafeAddRef(S2<ItfT> & pItf)
		{
			return SafeAddRef(pItf.GetValue());
		}
	}

	namespace clr
	{

		template <typename ItfT>
		static inline
			typename std::enable_if< std::is_pointer<ItfT>::value, ItfT >::type
			SafeAddRef(ItfT pItf)
		{
			return pItf;
		}
	}

	void SafeAddRef(void* pUnk);

	struct S {
	};

	void f(S2<S> *p)
	{
		clr::SafeAddRef(*p);
	}

The issue is that the first 'SafeAddRef' wants to call the second 'SafeAddRef'. However, the second overload hasn't been forward declared yet.
This means only the first and the third 'SafeAddRef' (found via argument dependent lookup) will participate in the overload resolution and none of them are valid candidate.

Adding the forward declaration (as shown in the repro) can fix the issue. But ideally, 'src\\inc\\holder.h' (which has the first 'SafeAddRef') should include 'src\\inc\\utilcode.h' (which has the second 'SafeAddRef').

jkotas:
Fixed by  #11886.

PR #11886, opened 2017-05-24
Fix Issue 11861

xiangfan-ms:
This is to address https://github.com/dotnet/coreclr/issues/11861.

Add a forward declaration for SafeAddRef from utilcode.h.
SafeAddRef in holder.h calls SafeAddRef in utilcode.h. The latter hasn't been declared yet, so it isn't a valid candidate during overload resolution.
It happens to work in MSVC default mode because of the lack of two phase name lookup.


jkotas:

jkotas:
Thanks!

PR #11891, opened 2017-05-24
Allow RelativePointer::SetValue usage for non-DAC builds only

gbalykov:
SetValue shouldn't be used in DAC builds, so unused based versions of SetValue are completely removed and SetValue without base is rewritten for non DAC builds.


gbalykov:
@jkotas 

Could you, please, take a look?

gbalykov:
@jkotas, Thank you!

PR #11894, opened 2017-05-24
Never enregister type-punned regStruct lclVars.

### LINKED ISSUE #11804: Uninitialized stack value instead of register returned by Unsafe.As<Vector<int>, int>(ref vector)
Using Unsafe to reinterpret a vector can load from the vector's stack memory even if the logical value has been mutated in registers and has not yet been stored. For example:
```cs
        static int UnsafeGrab()
        {
            var vector = new Vector<int>(4);
            return Unsafe.As<Vector<int>, int>(ref vector);
        }
        static int IndexerGrab()
        {
            var vector = new Vector<int>(4);
            return vector[0];
        }
```
IndexerGrab returns 4 as expected, while UnsafeGrab returns an uninitialized value like -1894710053.

UnsafeGrab assembly:
```
 mov         eax,4  
 movd        xmm0,eax  
 pshufd      xmm0,xmm0,0  
 mov         eax,dword ptr [rsp]  ;<--
 mov         eax,eax
 ret  
```
IndexerGrab assembly:
```
 mov         eax,4  
 movd        xmm0,eax  
 pshufd      xmm0,xmm0,0  
 movd        eax,xmm0
 ret  
```
Tested on NETCore.App 2.0.0-preview2-25309-07.

pgavlin:
Type-punned lclVars--i.e. lclVars that are accessed via
`(INDIR Tx (ADDR (LCLVAR Ty V)))`--are currently considered
enregisterable if they are SIMD-typed, have not been the accessed by
field, and have not otherwise been declared unenregisterable. This
appears to be an oversight, as the JIT is not able to generate correct
code for the type-punned access if the lclVar is in fact enregistered.

This change removes the special case for these SIMD-typed lclVars.

Fixes #11804.

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -13202,26 +13202,14 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
                         varTypeIsStruct(tempTyp) || (tempTyp == TYP_BLK) || (tempTyp == TYP_LCLBLK);
                     const unsigned varSize = useExactSize ? varDsc->lvExactSize : genTypeSize(temp);
 
+                    // Make sure we do not enregister this lclVar.
+                    lvaSetVarDoNotEnregister(lclNum DEBUGARG(DNER_LocalField));
+
                     // If the size of the load is greater than the size of the lclVar, we cannot fold this access into
                     // a lclFld: the access represented by an lclFld node must begin at or after the start of the
                     // lclVar and must not extend beyond the end of the lclVar.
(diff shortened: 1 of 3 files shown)


pgavlin:

CarolEidt:
Did you see any diffs? I would guess that in most cases there are checks elsewhere that would cause these not to be enregistered in the non-SIMD case.

pgavlin:
> Did you see any diffs?

I did not, no.

> I would guess that in most cases there are checks elsewhere that would cause these not to be enregistered in the non-SIMD case.

That's correct. AFAICT `lvRegStruct` is only true for SIMD-typed lclVars that are not accessed by field.

PR #11901, opened 2017-05-24
Make containedness explicit

CarolEidt:
Eliminate the use of `GTF_REG_VAL` in RyuJIT and reuse the bit to mark nodes as contained.
Abstract `GTF_REG_VAL` for legacy backend.
Important note about `GTF_REG_VAL`: This is used in `LEGACY_BACKEND` in most cases to record, during code generation, whether code has been generated to produce the value in a register. It is checked in various places to determine the type of code to generate.
In the RyuJIT backend, it was only set on lclVar nodes, and then only served as a redundant indicator that the var is in a register. Changing the places where it is checked to assert that it is zero produced no diffs. This also pointed out numerous areas of dead code. Note also that the RyuJIT backend always knows, based on containment and register allocation decisions, whether it is generating operations from the stack or registers.
Also, the updating of gc register sets is taken care of by `genConsumeReg()` and `genProduceReg()`.

### DIFF
--- a/src/jit/codegenxarch.cpp
+++ b/src/jit/codegenxarch.cpp
@@ -1331,10 +1331,9 @@ void CodeGen::genReturn(GenTreePtr treeNode)
                 // If it already has a home location, use that. Otherwise, we need a temp.
                 if (genIsRegCandidateLocal(op1) && compiler->lvaTable[op1->gtLclVarCommon.gtLclNum].lvOnFrame)
                 {
-                    // Store local variable to its home location, if necessary.
-                    if ((op1->gtFlags & GTF_REG_VAL) != 0)
+                    if (compiler->lvaTable[op1->gtLclVarCommon.gtLclNum].lvRegNum != REG_STK)
                     {
-                        op1->gtFlags &= ~GTF_REG_VAL;
+                        op1->gtFlags |= GTF_SPILL;
                         inst_TT_RV(ins_Store(op1->gtType,
                                              compiler->isSIMDTypeLocalAligned(op1->gtLclVarCommon.gtLclNum)),
                                    op1, op1->gtRegNum);
@@ -1551,6 +1550,7 @@ void CodeGen::genCodeForReturnTrap(GenTreeOp* tree)
     GenTree* data = tree->gtOp1;
     genConsumeRegs(data);
     GenTreeIntCon cns = intForm(TYP_INT, 0);
+    cns.SetContained();
     getEmitter()->emitInsBinary(INS_cmp, emitTypeSize(TYP_INT), data, &cns);
 
     BasicBlock* skipLabel = genCreateTempLabel();
@@ -3600,6 +3600,7 @@ void CodeGen::genLockedInstructions(GenTreeOp* treeNode)
     // all of these nodes implicitly do an indirection on op1
     // so create a temporary node to feed into the pattern matching
     GenTreeIndir i = indirForm(type, addr);
+    i.SetContained();
     getEmitter()->emitInsBinary(ins, emitTypeSize(type), &i, data);
 
     if (treeNode->gtRegNum != REG_NA)
@@ -3940,6 +3941,7 @@ GenTreeIndir CodeGen::indirForm(var_types type, GenTree* base)
 {
     GenTreeIndir i(GT_IND, type, base, nullptr);
     i.gtRegNum = REG_NA;
+    i.SetContained();
     // has to be nonnull (because contained nodes can't be the last in block)
     // but don't want it to be a valid pointer
     i.gtNext = (GenTree*)(-1);
@@ -4234,11 +4236,8 @@ void CodeGen::genCodeForLclAddr(GenTree* tree)
     var_types targetType = tree->TypeGet();
     regNumber targetReg  = tree->gtRegNum;
 
-    // Address of a local var.  This by itself should never be allocated a register.
-    // If it is worth storing the address in a register then it should be cse'ed into
-    // a temp and that would be allocated a register.
+    // Address of a local var.
     noway_assert(targetType == TYP_BYREF);
-    noway_assert(!tree->InReg());
 
     inst_RV_TT(INS_lea, targetReg, tree, 0, EA_BYREF);
     genProduceReg(tree);
@@ -4275,7 +4274,7 @@ void CodeGen::genCodeForLclFld(GenTreeLclFld* tree)
     unsigned varNum = tree->gtLclNum;
     assert(varNum < compiler->lvaCount);
 
-    getEmitter()->emitIns_R_S(ins_Move_Extend(targetType, tree->InReg()), size, targetReg, varNum, offs);
+    getEmitter()->emitIns_R_S(ins_Move_Extend(targetType, false), size, targetReg, varNum, offs);
 
     genProduceReg(tree);
 }
@@ -4295,18 +4294,11 @@ void CodeGen::genCodeForLclVar(GenTreeLclVar* tree)
 
     bool isRegCandidate = compiler->lvaTable[tree->gtLclNum].lvIsRegCandidate();
 
-    if (isRegCandidate && !(tree->gtFlags & GTF_VAR_DEATH))
-    {
-        assert(tree->InReg() || (tree->gtFlags & GTF_SPILLED));
-    }
-
     // If this is a register candidate that has been spilled, genConsumeReg() will
     // reload it at the point of use.  Otherwise, if it's not in a register, we load it here.
 
-    if (!tree->InReg() && !(tree->gtFlags & GTF_SPILLED))
+    if (!isRegCandidate && !(tree->gtFlags & GTF_SPILLED))
     {
-        assert(!isRegCandidate);
-
 #if defined(FEATURE_SIMD) && defined(_TARGET_X86_)
         // Loading of TYP_SIMD12 (i.e. Vector3) variable
         if (tree->TypeGet() == TYP_SIMD12)
@@ -4334,7 +4326,6 @@ void CodeGen::genCodeForStoreLclFld(GenTreeLclFld* tree)
 
     var_types targetType = tree->TypeGet();
     noway_assert(targetType != TYP_STRUCT);
-    noway_assert(!tree->InReg());
     assert(!varTypeIsFloating(targetType) || (targetType == tree->gtOp1->TypeGet()));
 
 #ifdef FEATURE_SIMD
@@ -4432,6 +4423,7 @@ void CodeGen::genCodeForStoreLclVar(GenTreeLclVar* tree)
             {
                 op1->gtRegNum = REG_NA;
                 op1->ResetReuseRegVal();
+                op1->SetContained();
             }
 
             if (!op1->isUsedFromReg())
@@ -4985,8 +4977,7 @@ void CodeGen::genCallInstruction(GenTreeCall* call)
                                                                        .eightByteSizes[iterationNum]));
                 if (putArgRegNode->gtRegNum != argReg)
                 {
-                    inst_RV_RV(ins_Move_Extend(putArgRegNode->TypeGet(), putArgRegNode->InReg()), argReg,
-                               putArgRegNode->gtRegNum);
+                    inst_RV_RV(ins_Move_Extend(putArgRegNode->TypeGet(), false), argReg, putArgRegNode->gtRegNum);
                 }
             }
         }
@@ -4997,7 +4988,7 @@ void CodeGen::genCallInstruction(GenTreeCall* call)
             genConsumeReg(argNode);
             if (argNode->gtRegNum != argReg)
             {
-                inst_RV_RV(ins_Move_Extend(argNode->TypeGet(), argNode->InReg()), argReg, argNode->gtRegNum);
+                inst_RV_RV(ins_Move_Extend(argNode->TypeGet(), false), argReg, argNode->gtRegNum);
             }
         }
 
@@ -6445,7 +6436,7 @@ void CodeGen::genIntToIntCast(GenTreePtr treeNode)
         {
             noway_assert(srcSize < EA_PTRSIZE);
 
-            ins = ins_Move_Extend(srcType, castOp->InReg());
+            ins = ins_Move_Extend(srcType, false);
 
             /*
                 Special case: ins_Move_Extend assumes the destination type is no bigger
@@ -6476,7 +6467,7 @@ void CodeGen::genIntToIntCast(GenTreePtr treeNode)
         }
         else
         {
-            ins = ins_Move_Extend(dstType, castOp->InReg());
+            ins = ins_Move_Extend(dstType, false);
         }
     }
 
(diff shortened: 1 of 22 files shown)


src/jit/gentree.h
```diff
@@ -1909,15 +1916,55 @@ struct GenTree
...
+        return (GetRegTag() != GT_REGTAG_NONE) ? gtRegNum : REG_NA;
+    }
+#endif
+
+    static bool IsContained(unsigned flags)
+    {
+        return ((flags & GTF_CONTAINED) != 0);
+    }
+
+    void SetAsContained()
```
mikedn: nit: `SetAs` sounds a bit odd, perhaps `SetContained` or `MakeContained`?

src/jit/gentree.h
```diff
@@ -1909,15 +1916,55 @@ struct GenTree
...
+    {
+        gtFlags |= GTF_CONTAINED;
+    }
+
+    void ResetContained()
+    {
+        gtFlags &= ~GTF_CONTAINED;
+        gtLsraInfo.regOptional = false;
```
mikedn: I wonder how would this work if `gtLsraInfo` is moved out of `GenTree`.

src/jit/target.h
```diff
@@ -44,15 +44,23 @@
...
+//
+// Each register list in register.h must declare REG_STK as the last value.
+// In the following enum declarations, the following REG_XXX are created beyond
+// the "real" registers:
+//    REG_STK          - Used to indicate something evaluated onto the stack.
+//    ACTUAL_REG_COUNT - The number of physical registers. (same as REG_STK).
+//    REG_COUNT        - The number of physical register + REG_STK. This is the count of values that may
+//                       be assigned during register allocation.
+//    REG_NA           - Used to indicate that a register is either not yet assigned or not required.
+//
```
mikedn: Comments about the newly added `REG_UNK`, `REG_OPT` and `REG_OPTUSE`?

src/jit/lsraxarch.cpp
```diff
@@ -281,12 +281,12 @@ void Lowering::TreeNodeInfoInit(GenTree* tree)
...
                 if (cmpOp1->IsSIMDEqualityOrInequality() && (cmpOp2->IsIntegralConst(0) || cmpOp2->IsIntegralConst(1)))
                 {
-                    // We always generate code for a SIMD equality comparison, but the compare
-                    // is contained (evaluated as part of the GT_JTRUE).
+                    // We always generate code for a SIMD equality comparison, but the compare itself produces no value.
                     // Neither the SIMD node nor the immediate need to be evaluated into a register.
                     l->clearOperandCounts(cmp);
                     l->clearDstCount(cmpOp1);
                     l->clearOperandCounts(cmpOp2);
+                    cmpOp2->SetAsContained();
```
mikedn: Hmm, it seems that even after this change we still can't move this logic out of `TreeNodeInfoInit` easily.

src/jit/lower.cpp
```diff
@@ -43,6 +43,7 @@ void Lowering::MakeSrcContained(GenTreePtr parentNode, GenTreePtr childNode)
 {
     assert(!parentNode->OperIsLeaf());
     assert(childNode->canBeContained());
+    childNode->SetAsContained();
 
     int srcCount = childNode->gtLsraInfo.srcCount;
```
mikedn: So we still depend on `gtLsraInfo` being initialized so we can't move containment logic from `TreeNodeInfoInit` to lower?

src/jit/gentree.h
CarolEidt: So ... this is one step on the way. As I mentioned in https://github.com/dotnet/coreclr/blob/master/Documentation/design-docs/lsra-throughput.md#representing-containedness my original plan was to use a `GTF_TREE_ROOT` flag but then I realized that I should just make it an explicit `GTF_CONTAINED` flag. The next steps (in some order) are to:
* replace `gtLsraInfo.regOptional` with `REG_OPT` and `REG_OPT_USE` values for `gtRegNum`.
* move the setting of these into `Lowering`
* remove `gtLsraInfo` out of `GenTree`, since it can now be determined without any context from parent nodes

src/jit/lsraxarch.cpp
CarolEidt: Actually, this is really resetting containedness on descendent nodes, but is complicated in the current model because we're determining containedness and setting source counts at the same time. Eventually this will just call `ResetContained()` on the descendent nodes, since there won't need to be any fixing up of source counts.

src/jit/target.h
CarolEidt: Ah, I hadn't intended to include them yet. But `REG_UNK` would be an explicit way to represent that a register is required but not yet assigned, and `REG_OPT` and `REG_OPTUSE` would replace `gtLsraInfo.regOptional`. After `Lowering` `REG_NA` would indicate that no register is required. I'm not sure it's necessary from a functionality perspective, but I think it would make the model cleaner.

src/jit/lower.cpp
CarolEidt: Not yet, but this is a step on the way.

src/jit/gentree.h
CarolEidt: `SetContained` sounds good, and more appropriate with `ResetContained`

src/jit/gentree.h
mikedn: Thanks for the link, it looks like I missed or completely forgot about that document

CarolEidt:
@dotnet-bot test Ubuntu gcstress0x3
@dotnet-bot test Ubuntu gcstress0xc_minopts_heapverify1
@dotnet-bot test Windows_NT x86 Checked gcstress0xc_minopts_heapverify1

CarolEidt:
@dotnet-bot test Windows_NT gcstress0xc_jitstress1
@dotnet-bot test Windows_NT x86 Checked gcstress0x3
@dotnet-bot test Ubuntu gcstress0x3

CarolEidt:
@dotnet/jit-contrib PTAL

CarolEidt:
I am doing some GCstress testing in the interest of thoroughness, though I'm confident that the removal of `GTF_REG_VAL` for RyuJIT will not affect GC reporting. The first batch I triggered were all failing in master (I should have checked first).

pgavlin:
> The first batch I triggered were all failing in master (I should have checked first).

Yes, it looks like there are widespread problems with the external disassembler at the moment. AFAICT from Kusto this is still ongoing.

cc @swaroop-sridhar

swaroop-sridhar:
Looks like libcoredistools.so wasn't downloaded/available for the GCStress 0xC runs.

The way the runtime scripts fetch libcoredistools.so changed recently, so this could be related: https://github.com/dotnet/coreclr/commit/798387ee90d7536d5a620af8482100ba76f6061a 
@wtgodbe: Was libcoredistools.so dependency tested with that change?

RussKeldorph:
Win/x64 GCStress=C runs seem to be working.  It's Win/x86 and Ubuntu/x64 that are broken.  The earliest job I can find with the "missing disassembler" error was [May 17, 2:16:09 PM](https://ci.dot.net/job/dotnet_coreclr/job/master/job/jitstress/job/x86_checked_windows_nt_gcstress0xc_jitstress1/25/).  The last job without it was [May 16, 6:40:07 AM](https://ci.dot.net/job/dotnet_coreclr/job/master/job/jitstress/job/x86_checked_windows_nt_gcstress0xc/27/).  I believe that window does indeed correspond to #11620 as @swaroop-sridhar suspects.

CarolEidt:
@dotnet-bot test Windows_NT gcstress0xc_jitstress1

CarolEidt:
@pgavlin @BruceForstall @dotnet/jit-contrib PTAL
The Win64/GCStress=C tests have failures that correlate with recent or intermittent failures in master.
The other GCStress testing appears to be broken.

wtgodbe:
@swaroop-sridhar I did not test libcoredistools change explicitly, I guess I assumed it would be done by the default CI suite. I think the thing to do would be to move that dependency into its own .csproj, and attempt to do its restore in a similar fashion to how it used to be done (explicitly as part of build.sh/build.cmd, rather than in the test dependency package restore pass). If you'd like I can put together a PR for that later today.

pgavlin:
@wtgodbe pulling it down as a package should be fine--let's move this discussion to #12011.

CarolEidt:
> Any assembly diffs?

No assembly diffs.

CarolEidt:
> I think that I would have preferred to #ifdef the body of InReg etc. rather than #ifdef the uses, but it doesn't look like it's possible to do that without introducing behavior that differs between release and checked builds that use the RyuJIT backend.

That was the first path I went down, but I came to the conclusion that it was misleading.

src/jit/gentree.cpp
```diff
@@ -15469,34 +15477,39 @@ bool GenTree::canBeContained() const
...
+    return false;
+#else // !LEGACY_BACKEND
     assert(IsLIR());
+    bool returnValue = ((gtFlags & GTF_CONTAINED) != 0);
 
+#ifdef DEBUG
     if (!canBeContained() || gtHasReg())
     {
-        return false;
+        assert(returnValue == false);
```
CarolEidt: My preference would be to leave this as-is. Otherwise, I would probably change the name of the variable to something like `isMarkedContained` - otherwise `!returnValue` seems to imply "there is no return value" rather than "the return value is false".

src/jit/gentree.h
```diff
@@ -598,9 +598,9 @@ struct GenTree
 
 private:
     union {
-        // NOTE: After LSRA, one of these values may be valid even if GTF_REG_VAL is not set in gtFlags.
-        // They store the register assigned to the node. If a register is not assigned, _gtRegNum is set to REG_NA
+        // These store the register assigned to the node. If a register is not assigned, _gtRegNum is set to REG_NA
         // or _gtRegPair is set to REG_PAIR_NONE, depending on the node type.
+        // For the LEGACY_BACKEND,these are valid only if GTF_REG_VAL is set in gtFlags.
         regNumberSmall _gtRegNum;  // which register      the value is in
```
BruceForstall: Unrelated, but we need to do a better job of extracting regPair stuff (e.g., regPairNoSmall) from the RyuJIT build. E.g., here arm32 regNumberSmall is `char` but the unused `regPairNoSmall` is `short`, so we're (possibly) wasting space.

src/jit/gentree.h
```diff
@@ -1909,15 +1916,56 @@ struct GenTree
...
+    {
+        return ((flags & GTF_CONTAINED) != 0);
+    }
+
+    void SetContained()
+    {
+        gtFlags |= GTF_CONTAINED;
+    }
+
+    void ResetContained()
```
BruceForstall: >ResetContained [](start = 9, length = 14)

`ClearContained()` instead?

src/jit/gentree.h
```diff
@@ -1909,15 +1916,56 @@ struct GenTree
...
+    {
+        return (GetRegTag() != GT_REGTAG_NONE) ? gtRegNum : REG_NA;
+    }
+#endif
+
+    static bool IsContained(unsigned flags)
+    {
+        return ((flags & GTF_CONTAINED) != 0);
+    }
+
```
BruceForstall: Should there also be:
```
    bool IsContained() const
    {
        return ((gtFlags & GTF_CONTAINED) != 0);
    }

```

src/jit/codegenxarch.cpp
```diff
@@ -6620,26 +6611,6 @@ void CodeGen::genIntToIntCast(GenTreePtr treeNode)
...
-                // Move the value into a byte register
-                noway_assert(!"Signed byte convert from non-byte-addressable register");
-            }
-
-            /* Generate "mov targetReg, castOp->gtReg */
-            if (targetReg != sourceReg)
-            {
-                inst_RV_RV(INS_mov, targetReg, sourceReg, srcType, srcSize);
-            }
-        }
```
BruceForstall: I don't understand why this is deleted. If there's some reason I don't understand, then isn't it true that `ins` can never be `INS_AND`, so the code below doesn't need to check for that?

BruceForstall: So InReg() shouldn't be checked. But maybe castOp should be checked for being `!contained`?

---
In reply to: [120245924](https://github.com/dotnet/coreclr/pull/11901#discussion_r120245924) [](ancestors = 120245924)

src/jit/gentree.cpp
```diff
@@ -10538,10 +10544,12 @@ void Compiler::gtDispRegVal(GenTree* tree)
         }
     }
 
-    if (tree->gtFlags & GTF_REG_VAL)
+#ifdef CODEGEN_LEGACY
+    if (tree->InReg())
```
BruceForstall: => `LEGACY_BACKEND`?

src/jit/gentree.cpp
```diff
@@ -15469,34 +15477,39 @@ bool GenTree::canBeContained() const
...
-    if (IsSIMDEqualityOrInequality())
+    else if (IsSIMDEqualityOrInequality())
     {
-        return false;
+        assert(!isMarkedContained);
     }
 
-#if !defined(LEGACY_BACKEND) && !defined(_TARGET_64BIT_)
+#if !defined(_TARGET_64BIT_)
     if (OperGet() == GT_LONG)
```
BruceForstall: The below code is now only under `DEBUG` which creates a debug/release diff. Seems like everything in this DEBUG section should be a simple assert.

src/jit/gentree.h
BruceForstall: Ok, I see there already is a (changed) `isContained()` -- confusingly with lower-case leading `i`.

---
In reply to: [120244670](https://github.com/dotnet/coreclr/pull/11901#discussion_r120244670) [](ancestors = 120244670)

BruceForstall:
It's a bit unfortunate you touched so much `LEGACY_BACKEND` code, since, e.g., we don't have an x86 `LEGACY_BACKEND` build working right now. Oh, well.

src/jit/gentree.h
CarolEidt: Yes, that was its existing name.

src/jit/codegenxarch.cpp
CarolEidt: This code was dead, since `InReg()` always returned zero. Doing anything else but deleting this code would change behavior, which I was specifically avoiding.

src/jit/gentree.cpp
CarolEidt: Ouch - yes. Believe it or not, I had actually tested LEGACY_BACKEND on an earlier version of these changes, but that was before I'd change `InReg()` from always returning false for RyuJIT backend, to eliminating it.

src/jit/gentree.cpp
CarolEidt: Yes, it shoud be an assert.

CarolEidt: As it turns out, the code and comment are actually incorrect. It can be contained and not be the last node in the block. Since it's an assert I've gone ahead and added a `gtGetParent` check.

mikedn: > It can be contained and not be the last node in the block

Do you mean uncontained? AFAIR this code only checked `gtNext` because if it's not last it would have been decomposed.

CarolEidt: > Do you mean uncontained?

Oops, yes, I meant uncontained. It will also not be decomposed if it has no parent, so checking `gtNext` is insufficient to cover the uncontained cases.

mikedn: I think I understand what happened here, I wrote this code before LIR, when `gtNext == nullptr` meant "last statement node", not "last BB node".

CarolEidt:
> It's a bit unfortunate you touched so much LEGACY_BACKEND code, since, e.g., we don't have an x86 LEGACY_BACKEND build working right now. Oh, well.

I got `LEGACY_BACKEND` "sort of" working on desktop, and ran diffs over the frameworks. There are zero diffs, though both base and diff have 1001 asserts.

CarolEidt:
@pgavlin @BruceForstall - I believe I've incorporated all the feedback. Could you have another look?

src/jit/gentree.h
CarolEidt: Done.

CarolEidt:
@dotnet/jit-contrib ping - could I get a review or two on this, please? I know this is a large set of changes, but I'd love to get this in before I have another painful merge.

CarolEidt:
Many thanks @pgavlin !

PR #11910, opened 2017-05-25
Coretype variables renamed back to netfx counterpart for reflection based serialization

ViktorHofer:
Relates to https://github.com/dotnet/corefx/issues/19119

This PR will be merged into the 2.0 branch. See issue above for more details.

cc @krwq

### DIFF
--- a/src/mscorlib/shared/System/Char.cs
+++ b/src/mscorlib/shared/System/Char.cs
@@ -26,7 +26,7 @@ namespace System
         //
         // Member Variables
         //
-        private char _value;
+        private char m_value; // Do not rename (binary serialization)
 
         //
         // Public Constants
@@ -102,7 +102,7 @@ namespace System
         // Calculate a hashcode for a 2 byte Unicode character.
         public override int GetHashCode()
         {
-            return (int)_value | ((int)_value << 16);
+            return (int)m_value | ((int)m_value << 16);
         }
 
         // Used for comparing two boxed Char objects.
@@ -113,13 +113,13 @@ namespace System
             {
                 return false;
             }
-            return (_value == ((Char)obj)._value);
+            return (m_value == ((Char)obj).m_value);
         }
 
         [System.Runtime.Versioning.NonVersionable]
         public bool Equals(Char obj)
         {
-            return _value == obj;
+            return m_value == obj;
         }
 
         // Compares this object to another object, returning an integer that
@@ -140,13 +140,13 @@ namespace System
                 throw new ArgumentException(SR.Arg_MustBeChar);
             }
 
-            return (_value - ((Char)value)._value);
+            return (m_value - ((Char)value).m_value);
         }
 
         [Pure]
         public int CompareTo(Char value)
         {
-            return (_value - value);
+            return (m_value - value);
         }
 
         // Overrides System.Object.ToString.
@@ -154,14 +154,14 @@ namespace System
         public override String ToString()
         {
             Contract.Ensures(Contract.Result<String>() != null);
-            return Char.ToString(_value);
+            return Char.ToString(m_value);
         }
 
         [Pure]
         public String ToString(IFormatProvider provider)
         {
             Contract.Ensures(Contract.Result<String>() != null);
-            return Char.ToString(_value);
+            return Char.ToString(m_value);
         }
 
         //
@@ -477,47 +477,47 @@ namespace System
 
         char IConvertible.ToChar(IFormatProvider provider)
         {
-            return _value;
+            return m_value;
         }
 
         sbyte IConvertible.ToSByte(IFormatProvider provider)
         {
-            return Convert.ToSByte(_value);
+            return Convert.ToSByte(m_value);
         }
 
         byte IConvertible.ToByte(IFormatProvider provider)
         {
-            return Convert.ToByte(_value);
+            return Convert.ToByte(m_value);
         }
 
         short IConvertible.ToInt16(IFormatProvider provider)
         {
-            return Convert.ToInt16(_value);
+            return Convert.ToInt16(m_value);
         }
 
         ushort IConvertible.ToUInt16(IFormatProvider provider)
         {
-            return Convert.ToUInt16(_value);
+            return Convert.ToUInt16(m_value);
         }
 
         int IConvertible.ToInt32(IFormatProvider provider)
         {
-            return Convert.ToInt32(_value);
+            return Convert.ToInt32(m_value);
         }
 
         uint IConvertible.ToUInt32(IFormatProvider provider)
         {
-            return Convert.ToUInt32(_value);
+            return Convert.ToUInt32(m_value);
         }
 
         long IConvertible.ToInt64(IFormatProvider provider)
         {
-            return Convert.ToInt64(_value);
+            return Convert.ToInt64(m_value);
         }
 
         ulong IConvertible.ToUInt64(IFormatProvider provider)
         {
-            return Convert.ToUInt64(_value);
+            return Convert.ToUInt64(m_value);
         }
 
         float IConvertible.ToSingle(IFormatProvider provider)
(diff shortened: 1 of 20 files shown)


src/mscorlib/shared/System/Char.cs
```diff
@@ -26,7 +26,7 @@ public struct Char : IComparable, IComparable<Char>, IEquatable<Char>, IConverti
         //
         // Member Variables
         //
-        private char _value;
+        private char m_value; // Do not rename (binary serialization)
```
jkotas: I do not think it is necessary to rename these to allow binary serialization.

Binary serializers have to treat primitive types like char or byte in special way. They cannot depend on the fields in them. Otherwise, they would end up with infinite recursion.

jkotas: Do you have a case where the current names break something?

morganbr: The primitives shouldn't need renames since the serialization format special-cases them. The other changes should still be necessary.

krwq:
@stephentoub I've made the EqualityComparers serialize like desktop PTAL. cc: @ViktorHofer

src/mscorlib/src/System/Collections/Generic/EqualityComparer.cs
```diff
@@ -401,6 +402,14 @@ internal override int LastIndexOf(T[] array, T value, int startIndex, int count)
             }
             return -1;
         }
+
+        public void GetObjectData(SerializationInfo info, StreamingContext context)
```
stephentoub: The type needs to implement ISerializable as well, no?  Same for the other one.

krwq: yes, I was just going to send patch with this 😉

krwq:

krwq:

krwq:
@dotnet-bot test this please

src/mscorlib/shared/System/Char.cs
ViktorHofer: We were hitting a few issues and after the rename they were gone. Could be related to something else, needs more investigation. Due to the fact that we need to finish this for 2.0 we (@morganbr @danmosemsft @stephentoub @krwq) decided to keep the changes on the Primitives.

src/mscorlib/src/System/Collections/Generic/EqualityComparer.cs
ViktorHofer: Good catch @krwq! 
cc @morganbr

krwq: @ViktorHofer thank @stephentoub :wink:

src/mscorlib/shared/System/Char.cs
jkotas: Could you please share details about the issues that you are hitting that are "fixed" by this?

src/mscorlib/src/System/Collections/Hashtable.cs
```diff
@@ -66,6 +66,7 @@ namespace System.Collections
     // 
     [DebuggerTypeProxy(typeof(System.Collections.Hashtable.HashtableDebugView))]
     [DebuggerDisplay("Count = {Count}")]
+    [Serializable]
```
jkotas: This is not going to work because of it will conflict with public Hashtable in corefx. When deserializing payload from desktop, the binary serializer will not know whether it needs to create this internal Hashtable or public Hashtable in corefx.

What is this needed for? If it is really needed, I think it needs to be fixed by moving the Hashtable from corefx to corelib.

stephentoub: More generally, are we actually exposing this Hashtable from any APIs, e.g. are we returning this from APIs that return IDictionary?  If so, we probably want to move the "real" Hashtable to CoreCLR anyway, to handle existing code that casts the IDictionary to Hashtable.

src/mscorlib/shared/System/Char.cs
ViktorHofer: I will follow up on your question later...

src/mscorlib/src/System/Collections/Hashtable.cs
ViktorHofer: Yes we are in at least a few cases I just found out. After my build is finished I can tell you the concrete types that are using it.

jkotas:
The merge of master did not seem to get done right. The diff shows many unrelated changes that makes it hard to review. Could you please clean it up?

ViktorHofer:
Yes will do.

src/mscorlib/src/System/Collections/Hashtable.cs
jkotas: > More generally, are we actually exposing this Hashtable from any APIs, e.g. are we returning this from APIs that return IDictionary?

We have been engineering around this problem (e.g. Environment.GetVariables). I agree that we should finally bite the bullet and move Hashtable to CoreLib if there is any reason for it.

ViktorHofer:
done

src/mscorlib/src/System/Collections/Hashtable.cs
ViktorHofer: Is there anything I need to watch out for when moving it from corefx to coreclr?

ViktorHofer: TimeZone is currently not serializable because of it: https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/CurrentSystemTimeZone.Cache.cs#L33

ViktorHofer: How is it possible that StringDictionary which lives in corefx uses the internal Hashtable from coreclr?
https://github.com/dotnet/corefx/blob/master/src/System.Collections.Specialized/src/System/Collections/Specialized/StringDictionary.cs#L21

```
  Deserialization of type System.Collections.Specialized.StringDictionary failed
     System.Runtime.Serialization.Formatters.Tests.BinaryFormatterTests.ValidateTfmHashes(original: [System.Collections.DictionaryEntry, System.Collections.DictionaryEntry], tfmBase64Hashes: [\"AAEAAAD/////AQAAAAAAAAAMAgAAAGFTeXN0ZW0uQ29sbGVjdG\"..., \"AAEAAAD/////AQAAAAAAAAAMAgAAAElTeXN0ZW0sIFZlcnNpb2\"...]) [FAIL]
        System.Runtime.Serialization.SerializationException : Type 'System.Collections.Hashtable' in Assembly 'System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e' is not marked as serializable.
```

morganbr: StringDictionary should be using CoreFX's Hashtable.

morganbr: If TimeZone is the only reason for moving HashTable, we could also drop serialization on TimeZone. When I added it to the list, I didn't realize it's obsolete. We do have data showing it used in serialization, but it's rare (~0.001%).

src/mscorlib/src/System/Collections/Generic/EqualityComparer.cs
```diff
@@ -479,5 +488,12 @@ internal override int LastIndexOf(T[] array, T value, int startIndex, int count)
...
+        {
+            // The LongEnumEqualityComparer does not exist on 4.0 so we need to serialize this comparer as ObjectEqualityComparer
```
danmosemsft: is 4.0  relevant to us?

stephentoub: It doesn't matter. Because this code exists in 4.7, the type in 4.7  doesn't have a deserialization ctor, which means core has to do the same dance. The comment could be tweaked if desired, though.

danmosemsft:
Merging in this state to unblock 2.0 porting. I"ll follow up with @jkotas offline and we can rename the primitive type members back if necessary.

Issue #11503, opened 2017-05-10
Timing issue in some baseservices tests


RussKeldorph:
There is a timing issue in tests when multiple tests share a test binary via the `<CLRTestProjectToRun>` property and when the tests are run with crossgen (`r2r` jobs).

I believe the problem is that even though the `RunCrossGen` logic attempts to prevent two projects from `crossgen`ing the same binary at the same time, one project can have [moved the binary](https://github.com/dotnet/coreclr/blob/c9f469e5fc99a498f46bd2d60fb198b7f7f1c22b/tests/src/CLRTest.CrossGen.targets#L48) while another project is attempting to execute it, resulting in the "not found" message below.  If we must support this "binary reuse" mechanism in the test harness, rather than locking, I'd recommend a safer mechanism that only writes to the project's own folder, copying if necessary.  I realize that has its own complications if a binary depends on its containing folder, but perhaps this is a lesser evil.

Incidentally, while you're in here, consider addressing #11502 as well.

Example of failures:
https://ci.dot.net/job/dotnet_coreclr/job/master/job/x64_checked_ubuntu_pri1r2r_tst/1705/#showFailuresLink

```
MESSAGE:
BEGIN EXECUTION export CDPATH=. /bin/sh -c ../CheckAddInt/CheckAddInt.sh BEGIN EXECUTION /mnt/j/workspace/dotnet_coreclr/master/x64_checked_ubuntu_pri1r2r_tst/bin/tests/Windows_NT.x64.Checked/Tests/coreoverlay/corerun CheckAddInt.exe /start:0 /add:100 Managed assembly not found: No such file or directory Expected: 100 Actual: 255 END EXECUTION - FAILED Expected: 0 Actual: 1 END EXECUTION - FAILED
+++++++++++++++++++
STACK TRACE:
```

RussKeldorph:
@gkhanna79 @wtgodbe FYI

gkhanna79:
This failure is not present in https://mc.dot.net/#/product/netcore/master/source/official~2Fcoreclr~2Fmaster~2F/type/test~2Ffunctional~2Fr2r~2Fcli~2F/build/20170509.02.

@RussKeldorph Is there something different in CI to trigger this? how consistently does this repro?

wtgodbe:
[...]

PR #11915, opened 2017-05-25
Fix timing issues in RunOnly tests

wtgodbe:
Attempts to resolve https://github.com/dotnet/coreclr/issues/11503 and https://github.com/dotnet/coreclr/issues/11502

CC @gkhanna79 @RussKeldorph

### DIFF
--- a/tests/src/CLRTest.Execute.Bash.targets
+++ b/tests/src/CLRTest.Execute.Bash.targets
@@ -262,6 +262,9 @@ $(BashLinkerTestCleanupCmds)
       ]]></BashCLRTestLaunchCmds>
       <BashCLRTestLaunchCmds Condition="'$(CLRTestKind)' == 'RunOnly'"><![CDATA[
 $(BashCLRTestLaunchCmds)
+if [ ! -z ${RunCrossGen+x} ]%3B then
+  CheckForCrossgenMarker
+fi
 echo export CDPATH="$%28dirname "$0")"
 export CDPATH="$%28dirname "$0")"
 echo /bin/sh -c  $(InputAssemblyName) 
@@ -356,8 +359,20 @@ ReleaseLock()
         return 3
     fi
 }
+
+CheckForCrossgenMarker()
+{
+    echo "in CheckForCrossgenMarker"
+    if [ ! -f "$MarkerFile" ]%3B then
+        sleep 1
+        CheckForCrossgenMarker
+    fi
+    echo "exiting CheckForCrossgenMarker"
+}
+
 cd "$%28dirname "$0")"
 LockFile="lock"
+MarkerFile="$(BuildProjectRelativeDir)\crossgen.mrk"
 
 
 # The __TestEnv variable may be used to specify a script to source before the test.
--- a/tests/src/CLRTest.CrossGen.targets
+++ b/tests/src/CLRTest.CrossGen.targets
@@ -57,7 +57,8 @@ if [ ! -z ${RunCrossGen+x} ]%3B then
           fi
         fi 
         ReleaseLock       
-    fi        
+    fi
+    touch $(BuildProjectRelativeDir)\crossgen.mrk        
 fi        
         ]]>        
       </CrossgenBashScript>
@@ -93,6 +94,7 @@ if defined RunCrossGen (
             Exit /b 1
         )
     )
+    copy /y nul $(BuildProjectRelativeDir)\crossgen.mrk
 ) 
         ]]>
       </CrossgenBatchScript>
--- a/tests/src/CLRTest.Execute.Batch.targets
+++ b/tests/src/CLRTest.Execute.Batch.targets
@@ -132,6 +132,15 @@ Exit /b 2
 :ReleaseLock
 if exist %lockFolder% rd /s /q %lockFolder%
 Exit /b 0
+
+
+:CheckForCrossgenMarker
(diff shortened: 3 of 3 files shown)


wtgodbe:
@dotnet-bot test test Ubuntu Release pri1r2r

wtgodbe:
Also @ramarag PTAL

tests/src/CLRTest.Execute.Bash.targets
```diff
@@ -356,8 +359,20 @@ ReleaseLock()
         return 3
     fi
 }
+
+CheckForCrossgenMarker()
+{
+    echo "in CheckForCrossgenMarker"
+    if [ ! -f "$MarkerFile" ]%3B then
+        sleep 1
+        CheckForCrossgenMarker
```
RussKeldorph: Unbounded recursion seems unnecessary.  Wouldn't a simple `for` or `while` loop be better.

RussKeldorph:
@wtgodbe Can you explain how this works?  For `CheckForCrossgenMarker` exit, I expect something to delete `MarkerFile`, but I don't see that happening?  Do I misunderstand?

Also, I'm not following how this avoids a race condition.  It seems like the simplest thing would be to expand the `TakeLock` logic to just cover both crossgen'ing and execution of the test.  Maybe `CLRTestBa*PostCommands` would be useful for that.  Not pretty, but perhaps better than now.

Also, how does this address #11502?  Seems like you're still modifying the test binaries.

ramarag:
if this is only meant for Runonly tests, i would suggest getting rid of this category altogether, they are causing more grief than benefits. Last i saw there were only a few remaining.

RussKeldorph:
Ah, I misunderstood at least one thing: it looks like each test waits until MarkerFile is created and nothing needs to delete it.  In that case, shouldn't only the test that did the crossgen create MarkerFile?  I'm not following how that's an exclusive task.

RussKeldorph:
@ramarag
> Last i saw there were only a few remaining.

```
C:\git\coreclr\tests\src>findstr -spn CLRTestProjectToRun *proj|findstr -v GCSimulator|wc -l
         52

C:\git\coreclr\tests\src>findstr -spn CLRTestProjectToRun *proj|findstr GCSimulator|wc -l
        430
```

:-/

wtgodbe:
@RussKeldorph it turns out that only the Build-Only tests attempt to crossgen the .exe file, while the run-only tests just attempt to Corerun whatever .exe is present - so only 1 project will create the marer file, then all of the run-only's wait for it to exist before executing

wtgodbe:
I had thought of surrounding the call to corerun with TakeLock, but then we lose parallelism on executing these tests

wtgodbe:
I'm also not sure it's possible to not modify test binaries at all - at the very least, we need to replace the original .exe with the crossgen'd exe at some point, meaning that during that operation there will be a period where there is no .exe present in the test output dir. This attempts to circumvent that problem by ensuring that nobody tries to execute the test .exe during or before Crossgen

ramarag:
@RussKeldorph

I believe the gcsimulator tests do not use the normal test infra to run them, @swgillespie am i correct ?

wtgodbe:
I agree we should get rid of those tests if possible, and can file an issue against myself to do so for 2.1, but we need to get these tests passing for 2.0.0, which means getting this change in today.

wtgodbe:
@dotnet-bot test this please

wtgodbe:
@dotnet-bot test test Ubuntu Release pri1r2r

wtgodbe:

RussKeldorph:
> it turns out that only the Build-Only tests attempt to crossgen the .exe file

In that case, I don't know why the TakeLock construct is necessary.  By "Build-Only" Are you referring to `<CLRTestKind>`?  I guess with that you could use that to ensure no race conditions, but it seems fragile since you're relying on the tests themselves to to conform to a vague contract.  For example, the tests under `baseservices\threading\paramthreadstart` use `RunOnly` but not `BuildOnly`.  Would those need to be fixed to conform to your assumption, or is `BuildOnly` implied somewhere?  (Sorry for asking so many questions; I haven't had a chance to dig through every line of this.)

> I'm also not sure it's possible to not modify test binaries at all - at the very least, we need to replace the original .exe with the crossgen'd exe at some point

We could do all the writing to a temp folder.  I suspect the vast majority of tests don't have a dependency on their own path location.  For the tests that do, we should give them a way to refer to their original location or copy necessary dependencies to their "working folder."  At any rate, I'd prefer not to close #11502 this this PR.  That should only be closed if we really stop writing to the tests folder or we decide it's not worth fixing.

RussKeldorph:
> is BuildOnly implied somewhere?

I suspect this this mechanism works as long as, among a group of tests that share a binary, there is at most one that is *not* `RunOnly`, so a single `BuildOnly` or a single Build-and-Run is fine.  If that's the explanation, then Looks Ok To Me.  Not great, but ok.

swgillespie:
@ramarag 

> I believe the gcsimulator tests do not use the normal test infra to run them, @swgillespie am i correct ?

GCSimulator tests use the normal test infra, it skips all tests not marked with the GCSimulator test attribute.

wtgodbe:
@dotnet-bot test this please

wtgodbe:
@dotnet-bot test test Ubuntu Release pri1r2r

PR #11917, opened 2017-05-25
Clear GTF_CALL from non-calls in rationalize

JosephTremoulet:
Since rationalization hoists comma antecedents out into separate
statements, any side-effects that the antecedent had might no longer apply
to the comma's ancestor nodes.  Moreover, in LIR, side-effect flags need
only describe the immediate node; summary flags are computed when Ranges
are constructed.  So, update rationalize to clear the `GTF_CALL` flag from
any node that is not a call.  Failure to do so gives an
overly-conservative view of side-effects, which can inhibit dead store
removal.

### DIFF
--- a/src/jit/rationalize.cpp
+++ b/src/jit/rationalize.cpp
@@ -970,10 +970,19 @@ Compiler::fgWalkResult Rationalizer::RewriteNode(GenTree** useEdge, ArrayStack<G
             node->gtFlags &= ~GTF_ALL_EFFECT;
         }
     }
-    else if (!node->OperIsStore())
+    else
     {
-        // Clear the GTF_ASG flag for all nodes but stores
-        node->gtFlags &= ~GTF_ASG;
+        if (!node->OperIsStore())
+        {
+            // Clear the GTF_ASG flag for all nodes but stores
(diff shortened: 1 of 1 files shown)


JosephTremoulet:
@pgavlin PTAL; this implements your [suggestion](https://github.com/dotnet/coreclr/pull/10946#issuecomment-294004794) from #10946.
/cc @dotnet/jit-contrib


jit-diff release --frameworks --tests shows just one diff, improved deadcode elimination:

```
Found files with textual diffs.
Summary:
(Note: Lower is better)
Total bytes of diff: -4 (0.00 % of base)
    diff is an improvement.
          -4 : JIT\IL_Conformance\Old\Conformance_Base\starg_i\starg_i.dasm (-10.00 % of base)
1 total files with size differences (1 improved, 0 regressed), 6513 unchanged.
          -4 : JIT\IL_Conformance\Old\Conformance_Base\starg_i\starg_i.dasm - _starg:main(ref):int
1 total methods with size differences (1 improved, 0 regressed), 314382 unchanged.
```

JosephTremoulet:
test this please

PR #11919, opened 2017-05-25
Add XUnit results for Windows_NT x86 jobs.

pgavlin:
Just what it says on the tin.

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -1504,9 +1504,7 @@ def static calculateBuildCommands(def newJob, def scenario, def branch, def isPR
                         }
 
                         if (!isBuildOnly) {
-                            if (architecture == 'x64' || !isPR) {
-                                Utilities.addXUnitDotNETResults(newJob, 'bin/**/TestRun*.xml')
-                            }
+                            Utilities.addXUnitDotNETResults(newJob, 'bin/**/TestRun*.xml', true)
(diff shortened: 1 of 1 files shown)


pgavlin:
@mmitche PTAL

pgavlin:
@dotnet-bot test ci please

pgavlin:
cc @russellhadley @RussKeldorph

pgavlin:
(and @jashook)

jashook:
We will need to make sure that this will not cause job failures for arm64 windows. I believe it will not, but it will be worth explicitly checking post checkin.

russellhadley:
Speaking of multiple target.  Will will just start getting arm64 and arm32 results from this too?  Is it really only x86?

jashook:
Arm64 Windows will not come online with this, we still need to enable xunit archiving for the jobs. Once arm64 unix is online though it will be enabled.

pgavlin:
> We will need to make sure that this will not cause job failures for arm64 windows
> ...
> Speaking of multiple target. Will will just start getting arm64 and arm32 results from this too? Is it really only x86?

The affected code is guarded on Windows_NT + {x86,x64} + !buildOnly, so it really should be just x86 jobs.

jashook:
I see makes sense.

PR #11920, opened 2017-05-25
Remove the check that requires an instrumented flags for SPC when collecting IBC data

### LINKED ISSUE #11918: Crossgen /tuning requires that System.Private.Corelib.dll be a /tuning image as well
This requirement should not be enforced for CoreCLR.

To Repro:

crossgen /tuning  /Platform_Assemblies_Paths . System.Runtime.dll
Microsoft (R) CoreCLR Native Image Generator - Version 4.5.22220.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Could not load file or assembly '.\System.Private.CoreLib.dll'. The native image could not be loaded, because it was generated for use by a different version of the runtime. (Exception from HRESULT: 0x80131059)
Error: compilation failed for "System.Runtime.dll" (0x80131059)

briansull:
This check is not needed on CoreCLR and blocks some valuable usage cases.

Fixes #11918

### DIFF
--- a/src/vm/pefile.cpp
+++ b/src/vm/pefile.cpp
@@ -1232,17 +1232,6 @@ BOOL PEAssembly::CheckNativeImageVersion(PEImage *peimage)
 
     CorCompileConfigFlags configFlags = PEFile::GetNativeImageConfigFlagsWithOverrides();
 
-    if (IsSystem())
-    {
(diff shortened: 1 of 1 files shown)


briansull:
@jkotas @dpodder PTAL

jkotas:
@dotnet-bot test this please

PR #11922, opened 2017-05-26
EETypeHashTable lookups can deadlock with GC during comparison with unrestored persisted TypeHandles

jkotas:
Port stress fix from desktop. It affect complex scenario with fragile NGen only that do not really exist on CoreCLR today, but it does not hurt to have the fix in CoreCLR just in case.


jkotas:

PR #11923, opened 2017-05-26
Allowing larger segment size for SOH

Maoni0:
in preparation for enabling it for NoGC region. the main change here is to make sure we do an ephemeral promote if the new ephemeral size cannot fit into a normal sized seg.

### DIFF
--- a/src/gc/gc.cpp
+++ b/src/gc/gc.cpp
@@ -2667,6 +2667,9 @@ size_t gc_heap::eph_gen_starts_size = 0;
 heap_segment* gc_heap::segment_standby_list;
 size_t        gc_heap::last_gc_index = 0;
 size_t        gc_heap::min_segment_size = 0;
+size_t        gc_heap::soh_segment_size = 0;
+size_t        gc_heap::min_loh_segment_size = 0;
+size_t        gc_heap::segment_info_size = 0;
 
 #ifdef GC_CONFIG_DRIVEN
 size_t gc_heap::time_init = 0;
@@ -2963,8 +2966,12 @@ gc_heap::dt_low_ephemeral_space_p (gc_tuning_point tp)
             
             dprintf (GTC_LOG, ("h%d: plan eph size is %Id, new gen0 is %Id", 
                 heap_number, plan_ephemeral_size, new_gen0size));
-            ret = ((size_t)(heap_segment_reserved (ephemeral_heap_segment) - (heap_segment_mem (ephemeral_heap_segment))) <
-                    (plan_ephemeral_size + new_gen0size));
+
+            // If we were in no_gc_region we could have allocated a larger than normal segment,
+            // and the next seg we allocate will be a normal sized seg so if we can't fit the new
+            // ephemeral generations there, do an ephemeral promotion.
+            ret = ((soh_segment_size - segment_info_size) < (plan_ephemeral_size + new_gen0size));
+
             break;
         }
         default:
@@ -4374,9 +4381,6 @@ void virtual_free (void* add, size_t size)
                  size, (size_t)add, (size_t)((uint8_t*)add+size)));
 }
 
-// We have a few places that call this but the seg size doesn't change so call it
-// once and save the result.
-// TODO: move back after we do this.
 static size_t get_valid_segment_size (BOOL large_seg=FALSE)
 {
     size_t seg_size, initial_seg_size;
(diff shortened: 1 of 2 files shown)


Maoni0:
@adityamandaleeka PTAL

src/gc/gc.cpp
```diff
@@ -2963,8 +2966,12 @@ gc_heap::dt_low_ephemeral_space_p (gc_tuning_point tp)
             
             dprintf (GTC_LOG, ("h%d: plan eph size is %Id, new gen0 is %Id", 
                 heap_number, plan_ephemeral_size, new_gen0size));
-            ret = ((size_t)(heap_segment_reserved (ephemeral_heap_segment) - (heap_segment_mem (ephemeral_heap_segment))) <
-                    (plan_ephemeral_size + new_gen0size));
+
+            // While we were in the no gc region we could have allocated a larger than normal segment,
```
adityamandaleeka: > While we were in the no gc region [...]

This sounds like we are assuming that we know we got here after being in a no-GC region. Is that always the case?

I just did a quick search and it looks like we only call `dt_low_ephemeral_space_p` with `tuning_deciding_promote_ephemeral` if the pause mode is _not_ `pause_no_gc` (in `gc_heap::expand_heap`). I think I'm missing something here.

src/gc/gc.cpp
```diff
@@ -9928,6 +9926,15 @@ HRESULT gc_heap::initialize_gc (size_t segment_size,
 #endif //BACKGROUND_GC
 #endif //WRITE_WATCH
 
+#ifdef BACKGROUND_GC
+    //leave the first page to contain only segment info
```
adityamandaleeka: minor nit: comment formatting isn't consistent.

src/gc/gcpriv.h
```diff
@@ -3002,6 +3002,17 @@ class gc_heap
     PER_HEAP_ISOLATED
     size_t min_segment_size;
 
+    // For SOH we always allocate segments of the same
+    // size unless no gc region requires larger ones.
```
adityamandaleeka: Please change 'no gc region' to 'NoGCRegion'. It will be way clearer when parsing this comment later :smile:.

src/gc/gc.cpp
Maoni0: We got here after we've *exited* from the no gc region mode, now we need to handle the seg we allocated "While we were in the no gc region".

src/gc/gcpriv.h
Maoni0: the actual implementation is all named something no_gc_region, that's why I referred to it as no gc region. I guess I can change it to no_gc_region if that seems better.

adityamandaleeka: Yes, that works too. The way it is now, it sounds like something totally different.

adityamandaleeka:
The a comment on `get_valid_segment_size` talks about having places where that function is called unnecessarily. Since you've removed all (?) of those places, we can probably remove the part of the comment now, right? It also says "TODO: move back after we do this"--not sure what that means.

src/gc/gc.cpp
adityamandaleeka: But we also end up here if there was a heap expansion without ever having encountered a no-gc region, right?

Maoni0: yeah, but that doesn't cause us to ever create a segment that's larger than the default.

Maoni0:
I will remove the comments for get_valid_segment_size.

src/gc/gcpriv.h
Maoni0: I am curious - now that you know the  implementation is called no_gc_region, does this still sound like something totally different?

Maoni0:
@dotnet-bot test Ubuntu x64 Formatting

Maoni0:

src/gc/gcpriv.h
adityamandaleeka: Yes. I knew what this comment was trying to say (we are, after all, in a PR about making changes in preparation for no-gc region), but my point is that it's ambiguous and needlessly hard to parse. Your suggestion about adding the underscores is fine.

src/gc/gc.cpp
adityamandaleeka: OK. As with the other comment, I think this needs to be rephrased in order to make it clear outside of the context of this PR. Can you rephrase this to something like "in cases where we were in a no-gc region..."?

"While we were in the no gc region" implies that we had to have been in a no-gc region, which is not necessarily the case.

Maoni0:
updated commit to address all comment feedback.

adityamandaleeka:
LGTM, thanks.

PR #11925, opened 2017-05-26
extension nullable to check a box has no value or has a default value.

tldzyx:


danmosemsft:
@tldzyx hello, thanks for this. Is there an issue # driving this? New API would require api approval, plus changes in corefx repo also.
I will close this, please open a proposal issue in the corefx repo. If there is one, please reopen.

PR #11941, opened 2017-05-26
Fix failures in ETW logging on 4GB aware 32 bit processes

vancem:
We incorrectly cast a 32 bit pointer to a 64 bit poitner using as a SIGNED integer.
If this is a 32 bit process that is using more than 2GB of memory this can result in
sign rather than zero extension.  This makes the poitner invalid at the OS level
and causes the OS API to fail.

We disovered this in Visual Studio when debugging large (Rosyln) scenarios.
There were numerous failures which causes sever slowdowns becasue the
EventSource logged OutputDebugString events when the OS API failed.

The fix is to use unsigned extension.   Note that I have confirmed that casting
from a IntPtr or a void* to a ulong does zero extension (that is it uses the target type
to determine whether to use sign or zero extension).

To be useful for Visual Studio, this needs to be ported to the desktop runtime.

### DIFF
--- a/src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
+++ b/src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
@@ -1040,7 +1040,7 @@ namespace System.Diagnostics.Tracing
             /// Address where the one argument lives (if this points to managed memory you must ensure the
             /// managed object is pinned.
             /// </summary>
-            public IntPtr DataPointer { get { return (IntPtr)m_Ptr; } set { m_Ptr = unchecked((long)value); } }
+            public IntPtr DataPointer { get { return (IntPtr)m_Ptr; } set { m_Ptr = unchecked((ulong)value); } }
             /// <summary>
             /// Size of the argument referenced by DataPointer
             /// </summary>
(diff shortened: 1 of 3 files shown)


vancem:
@sharwell @brianrob - Brian, we should make a bug to port this back to Desktop.

brianrob:
I have filed a bug to port this to desktop.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -1040,7 +1040,7 @@ protected internal struct EventData
             /// Address where the one argument lives (if this points to managed memory you must ensure the
             /// managed object is pinned.
             /// </summary>
-            public IntPtr DataPointer { get { return (IntPtr)m_Ptr; } set { m_Ptr = unchecked((long)value); } }
+            public IntPtr DataPointer { get { return (IntPtr)m_Ptr; } set { m_Ptr = unchecked((ulong)value); } }
```
sharwell: ❓ Are you sure this fixes the issue? I'm still seeing sign-extension of the underlying value in the C# Interactive window.

```
> IntPtr.Size
4
> var dataPointer = (IntPtr)unchecked((int)0xdeadf00d);
> dataPointer.ToString("x16")
"00000000deadf00d"
> ((ulong)dataPointer).ToString("x16")
"ffffffffdeadf00d"
```

I would have expected to see the following here:

```csharp
unsafe
{
  m_Ptr = (ulong)(void*)value;
}
```

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -1056,14 +1056,14 @@ protected internal struct EventData
             /// <param name="reserved">Value for reserved: 2 for per-provider metadata, 1 for per-event metadata</param>
             internal unsafe void SetMetadata(byte* pointer, int size, int reserved)
             {
-                this.m_Ptr = (long)(ulong)(UIntPtr)pointer;
+                this.m_Ptr = (ulong)pointer;
```
sharwell: 💭 The C# Interactive window doesn't allow unsafe code, but I'm expecting this to work based on the following (emphasis mine):

> Mappings between pointers and integers are implementation-defined. However, on 32\* and 64-bit CPU architectures with a linear address space, conversions of pointers to or from integral types **typically behave exactly like conversions of `uint` or `ulong` values, respectively, to or from those integral types**.

vancem: I have tested this with a small app, byte* acts like an unsigned, and is zero extended.

src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
vancem: Ack.   You are correct.   Good catch.     I had performed an experiment much like your own but was not paying enough attention and was running 64 bit when I did the experiment (were it worked).   I will open another PR to fix this and more carefully validate it.

PR #11945, opened 2017-05-27
[WIP] try VARSET_TP iterator without arguments.

sandreenko:
Use return arg to keep both values.
Check `if (m_bits != 0) `, because it can be scheduled before ` hasBit = BitScanForward(&nextBit, m_bits); `.
Local runs showed on System.Private.Corelib crossgen (Time: total value from the JitTimeLogFile)

before: 4614 ms, after 4445 ms. It is an average for 20 runs.

As usual delete two more defines.

### DIFF
--- a/src/jit/assertionprop.cpp
+++ b/src/jit/assertionprop.cpp
@@ -169,16 +169,16 @@ void Compiler::optAddCopies()
         bool isDominatedByFirstBB = false;
 
         BlockSetOps::Iter iter(this, varDsc->lvRefBlks);
-        unsigned          blkNum = 0;
-        while (iter.NextElem(&blkNum))
+        unsigned          bbNum = 0;
+        while (iter.NextElem(&bbNum))
         {
-            /* Find the block 'blkNum' */
+            /* Find the block 'bbNum' */
             BasicBlock* block = fgFirstBB;
-            while (block && (block->bbNum != blkNum))
+            while (block && (block->bbNum != bbNum))
             {
                 block = block->bbNext;
             }
-            noway_assert(block && (block->bbNum == blkNum));
+            noway_assert(block && (block->bbNum == bbNum));
 
             bool     importantUseInBlock = (varDsc->lvIsParam) && (block->getBBWeight(this) > paramAvgWtdRefDiv2);
             bool     isPreHeaderBlock    = ((block->bbFlags & BBF_LOOP_PREHEADER) != 0);
@@ -215,7 +215,7 @@ void Compiler::optAddCopies()
 #ifdef DEBUG
             if (verbose)
             {
-                printf("        Referenced in BB%02u, bbWeight is %s", blkNum, refCntWtd2str(block->getBBWeight(this)));
+                printf("        Referenced in BB%02u, bbWeight is %s", bbNum, refCntWtd2str(block->getBBWeight(this)));
 
                 if (isDominatedByFirstBB)
                 {
@@ -322,16 +322,16 @@ void Compiler::optAddCopies()
 
             /* We have already calculated paramImportantUseDom above. */
             BlockSetOps::Iter iter(this, paramImportantUseDom);
-            unsigned          blkNum = 0;
-            while (iter.NextElem(&blkNum))
+            unsigned          bbNum = 0;
+            while (iter.NextElem(&bbNum))
             {
-                /* Advance block to point to 'blkNum' */
+                /* Advance block to point to 'bbNum' */
                 /* This assumes that the iterator returns block number is increasing lexical order. */
-                while (block && (block->bbNum != blkNum))
+                while (block && (block->bbNum != bbNum))
                 {
                     block = block->bbNext;
                 }
-                noway_assert(block && (block->bbNum == blkNum));
+                noway_assert(block && (block->bbNum == bbNum));
 
 #ifdef DEBUG
                 if (verbose)
--- a/src/jit/assertionprop.cpp
+++ b/src/jit/assertionprop.cpp
@@ -168,7 +168,8 @@ void Compiler::optAddCopies()
 
         bool isDominatedByFirstBB = false;
 
-        BLOCKSET_ITER_INIT(this, iter, varDsc->lvRefBlks, blkNum);
+        BlockSetOps::Iter iter(this, varDsc->lvRefBlks);
+        unsigned          blkNum = 0;
         while (iter.NextElem(&blkNum))
         {
             /* Find the block 'blkNum' */
@@ -320,8 +321,8 @@ void Compiler::optAddCopies()
 #endif
 
             /* We have already calculated paramImportantUseDom above. */
-
-            BLOCKSET_ITER_INIT(this, iter, paramImportantUseDom, blkNum);
+            BlockSetOps::Iter iter(this, paramImportantUseDom);
+            unsigned          blkNum = 0;
             while (iter.NextElem(&blkNum))
             {
                 /* Advance block to point to 'blkNum' */
--- a/src/jit/compiler.cpp
+++ b/src/jit/compiler.cpp
@@ -8297,7 +8297,8 @@ void dumpConvertedVarSet(Compiler* comp, VARSET_VALARG_TP vars)
     pVarNumSet            = (BYTE*)_alloca(varNumSetBytes);
     memset(pVarNumSet, 0, varNumSetBytes); // empty the set
 
-    VARSET_ITER_INIT(comp, iter, vars, varIndex);
+    VarSetOps::Iter iter(comp, vars);
+    unsigned        varIndex = 0;
     while (iter.NextElem(&varIndex))
     {
         unsigned varNum = comp->lvaTrackedToVarNum[varIndex];
--- a/src/jit/bitvec.h
+++ b/src/jit/bitvec.h
@@ -37,20 +37,4 @@ typedef BitSetShortLongRep BitVec;
 typedef BitVecOps::ValArgType BitVec_ValArg_T;
 typedef BitVecOps::RetValType BitVec_ValRet_T;
 
-// Initialize "_varName" to "_initVal."  Copies contents, not references; if "_varName" is uninitialized, allocates a
-// set for it (using "_traits" for any necessary allocation), and copies the contents of "_initVal" into it.
-#define BITVEC_INIT(_traits, _varName, _initVal) _varName(BitVecOps::MakeCopy(_traits, _initVal))
-
-// Initializes "_varName" to "_initVal", without copying: if "_initVal" is an indirect representation, copies its
-// pointer into "_varName".
-#define BITVEC_INIT_NOCOPY(_varName, _initVal) _varName(_initVal)
-
-// The iterator pattern.
-
-// Use this to initialize an iterator "_iterName" to iterate over a BitVec "_bitVec".
-// "_bitNum" will be an unsigned variable to which we assign the elements of "_bitVec".
-#define BITVEC_ITER_INIT(_traits, _iterName, _bitVec, _bitNum)                                                         \
-    unsigned        _bitNum = 0;                                                                                       \
-    BitVecOps::Iter _iterName(_traits, _bitVec)
-
 #endif // _BITVEC_INCLUDED_
--- a/src/jit/blockset.h
+++ b/src/jit/blockset.h
@@ -58,12 +58,4 @@ typedef BitSetShortLongRep BlockSet;
 typedef BlockSetOps::ValArgType BlockSet_ValArg_T;
 typedef BlockSetOps::RetValType BlockSet_ValRet_T;
 
-// The iterator pattern.
-
-// Use this to initialize an iterator "_iterName" to iterate over a BlockSet "_blockSet".
-// "_blockNum" will be an unsigned variable to which we assign the elements of "_blockSet".
-#define BLOCKSET_ITER_INIT(_comp, _iterName, _blockSet, _blockNum)                                                     \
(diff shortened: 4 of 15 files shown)


sandreenko:
@dotnet-bot test perf_throughput_perflab_Windows_NT_x64

sandreenko:
@dotnet-bot test perf throughput perflab Windows NT x64

sandreenko:
@dotnet-bot test perf throughput perflab Windows_NT x64

sandreenko:

sandreenko:
dotnet-bot test windows_nt x64 perf_throughput

sandreenko:
@dotnet-bot test windows_nt x64 perf_throughput

sandreenko:
dotnet-bot test windows_nt x64 throughput

sandreenko:
@dotnet-bot test windows_nt x64 throughput

sandreenko:
@dotnet-bot test windows_nt x64 throughput

sandreenko:
Updated results (an average for 25 runs)
```
old 4421
new 3987
```
so it is near 9% advantage according to JitCompilation time report (all time, ms).

[results.xlsx](https://github.com/dotnet/coreclr/files/1076103/results.xlsx)

src/jit/gentree.h
```diff
@@ -149,6 +149,18 @@ struct InlineCandidateInfo;
 
 typedef unsigned short AssertionIndex;
 
+//------------------------------------------------------------------------
+// GetAssertionIndex: return 1-based AssertionIndex from 0-based int index.
```
sandreenko: I do not see why we can't make AssertionIndex  0-based with -1 as NO_ASSERTION_INDEX. It will allows us to delete this confusing conversions, that are easy to miss.

sandreenko:
@dotnet-bot test windows_nt x64 throughput

sandreenko:
PTAL @dotnet/jit-contrib

BruceForstall:
Should this still be marked [WIP]? (If you get positive reviews will you merge it?)

What platform/architecture are the timings for? (x64 release Windows? What about x86 release Windows?)

src/jit/gentree.h
BruceForstall: Why do we do this? Is it to pack Assertion Numbers in the assertion bitmaps? Thus, bit 0 == assertion# 1, bit 1 == assertion # 2, etc.? We should stop doing this and just "burn" bit 0, so we don't need to do this math. Thus, bit0 == unused, bit 1 == assertion# 1, bit 2 == assertion #2, etc. We do it this way for BasicBlock numbers already. It's simpler and doesn't require this logic. There will possibly be diffs as one fewer assertion would be allowed in the bitmaps (assuming those are fixed sized)

---
In reply to: [122082167](https://github.com/dotnet/coreclr/pull/11945#discussion_r122082167) [](ancestors = 122082167)

src/jit/assertionprop.cpp
```diff
@@ -2852,16 +2852,15 @@ GenTreePtr Compiler::optAssertionProp_LclVar(ASSERT_VALARG_TP assertions, const
     }
 
     BitVecOps::Iter iter(apTraits, assertions);
-    unsigned        index = 0;
-    while (iter.NextElem(&index))
+    for (int i = iter.NextElem(); i != -1; i = iter.NextElem())
     {
```
BruceForstall: Another pattern would be:
```
int i;
while ((i = iter.NextElem()) != -1)
```
which, IMO, might be preferable, to avoid duplicating `iter.NextElem()`.

BruceForstall:
Can you explain why it's faster? 9% seems hard to believe.

Assuming you are measuring locally, you should build both baseline and diff release builds without PGO, to avoid PGO data degradation effects (although perhaps PGO retraining would make your change faster than without it):
```
--- a/pgosupport.cmake
+++ b/pgosupport.cmake
@@ -36,8 +36,8 @@ function(add_pgo TargetName)
         # If we don't have profile data availble, gracefully fall back to a non-PGO opt build
         if(EXISTS ${ProfilePath})
             if(WIN32)
-                set_property(TARGET ${TargetName} APPEND_STRING PROPERTY LINK_FLAGS_RELEASE        " /LTCG /USEPROFILE:PGD=${ProfilePath}")
-                set_property(TARGET ${TargetName} APPEND_STRING PROPERTY LINK_FLAGS_RELWITHDEBINFO " /LTCG /USEPROFILE:PGD=${ProfilePath}")
+                # set_property(TARGET ${TargetName} APPEND_STRING PROPERTY LINK_FLAGS_RELEASE        " /LTCG /USEPROFILE:PGD=${ProfilePath}")
+                # set_property(TARGET ${TargetName} APPEND_STRING PROPERTY LINK_FLAGS_RELWITHDEBINFO " /LTCG /USEPROFILE:PGD=${ProfilePath}")
             else(WIN32)
                 if(UPPERCASE_CMAKE_BUILD_TYPE STREQUAL RELEASE OR UPPERCASE_CMAKE_BUILD_TYPE STREQUAL RELWITHDEBINFO)
                     if(HAVE_LTO)
```

src/jit/bitset.h
```diff
@@ -147,8 +147,8 @@ FORCEINLINE unsigned BitSetSupport::CountBitsInIntegral<unsigned>(unsigned c)
 // In addition to implementing the method signatures here, an instantiation of BitSetOps must also export a
 // BitSetOps::Iter type, which supports the following operations:
 //      Iter(BitSetValueArgType):        a constructor
-//      bool NextElem(unsigned* pElem):  returns true if the iteration is not complete, and sets *pElem to the next
-//                                       yielded member.
+//      int NextElem():  returns true the next yielded member if the iteration is not complete,
```
BruceForstall: "returns true the next yielded member" => "returns the next yielded member"

BruceForstall:
Did you verify no diffs? How? What scenarios/builds/architectures?

src/jit/bitsetasshortlong.h
```diff
@@ -513,11 +513,10 @@ class BitSetOps</*BitSetType*/ BitSetShortLongRep,
 #endif
 
                 // If there's a bit, doesn't matter if we're short or long.
-                if (hasBit)
+                if (m_bits != 0)
                 {
```
BruceForstall: It would be useful to add a comment here about why you're checking `m_bits != 0` instead of `hasBits`.

BruceForstall: maybe you could also `assert(hasBit)` 

---
In reply to: [122267573](https://github.com/dotnet/coreclr/pull/11945#discussion_r122267573) [](ancestors = 122267573)

src/jit/bitsetasuint64inclass.h
```diff
@@ -295,16 +295,15 @@ class BitSetUint64Iter
...
         static const unsigned UINT64_SIZE = 64;
 
         if ((m_bits & 0x1) != 0)
         {
-            *pElem = m_bitNum;
             m_bitNum++;
             m_bits >>= 1;
-            return true;
+            return m_bitNum++;
         }
```
BruceForstall: This should be just `return m_bitNum`. Currently, m_bitNum is getting incremented twice. I prefer the increment before the `return`, and not using `++` in the `return` clause.

BruceForstall: ok, never mind, since you need to return the before-incremented value.

But, you need to delete the `m_bitNum++` line above

---
In reply to: [122268396](https://github.com/dotnet/coreclr/pull/11945#discussion_r122268396) [](ancestors = 122268396)

src/jit/compiler.cpp
```diff
@@ -8297,10 +8297,10 @@ void dumpConvertedVarSet(Compiler* comp, VARSET_VALARG_TP vars)
     pVarNumSet            = (BYTE*)_alloca(varNumSetBytes);
     memset(pVarNumSet, 0, varNumSetBytes); // empty the set
 
-    VARSET_ITER_INIT(comp, iter, vars, varIndex);
-    while (iter.NextElem(&varIndex))
+    VarSetOps::Iter iter(comp, vars);
+    for (int i = iter.NextElem(); i != -1; i = iter.NextElem())
     {
```
BruceForstall: IMO, you should leave the descriptive `varIndex` name and not replace it with `i`. Same for other cases where you've done this.

russellhadley:
It'd be nice to see the the lab CoreCLR-througput job for this.  I hate quoting numbers off a dev box for things we want to merge.

src/jit/bitset.h
```diff
@@ -279,7 +279,7 @@ class BitSetOps
     class Iter {
       public:
         Iter(Env env, BitSetValueArgType bs) {}
-        bool NextElem(unsigned* pElem) { return false; }
+        int NextElem() { return -1; }
     };
```
BruceForstall: Instead of `int` I wonder if this should be `unsigned` and then there should be a `const unsigned BitSetOps::Iter::Done = (unsigned)-1` value that callers compare against, instead of comparing against -1. I'm worried that by changing the "value return" type from `unsigned` to `int`, you've introduced a ton of signed/unsigned type mismatches in the code (perhaps hidden because that warning is suppressed).

src/jit/gentree.h
BruceForstall: @briansull Couldn't we change AssertionProp to stop doing this +1/-1 dance everywhere, and simply burn bit 0, as I've suggested here?

briansull: No we shouldn't waste a bit

src/jit/assertionprop.cpp
sandreenko: but in this case we define I in the external area of visibility.

sandreenko:
> Should this still be marked [WIP]? (If you get positive reviews will you merge it?)
> What platform/architecture are the timings for? (x64 release Windows? What about x86 release Windows?)
It is WIP because I have the test failures and I want to discuss invalid values for iterator and assertions here before the final review.

> I love the throughput improvement, although I'm very skeptical it can be as good as reported.
I think the improvement is significant only in crossgen where we have dominators and more bitsets operations in O(n^2) algorithms.

I am planning to fix the failures and do others tests.

src/jit/assertionprop.cpp
BruceForstall: true, but I'm not worried about that. Especially if a more descriptive name than `i` is used.

src/jit/bitset.h
sandreenko: Fixed

src/jit/bitsetasshortlong.h
sandreenko: Fixed

src/jit/bitsetasuint64inclass.h
sandreenko: Fixed

src/jit/assertionprop.cpp
pgavlin: I would prefer that we not use assignment as an expression. If we want to avoid the visual noise of the repeated calls to `NextElem()`, we should consider using `C++11 iterators.

pgavlin:
FWIW, I see no improvement in instructions retired as measured by `callgrind` when crosgen'ing S.P.C.dll on Linux. Indeed, there is a very small increase in instructions retired (about .03%).

pgavlin:
I am double-checking now that PGO isn't enabled for the baseline.

pgavlin:
I believe that I have confirmed that neither build of the JIT I used for the IR measurement above was built using PGO.

pgavlin:
Also, this is not to say that we should not take this change, just to help quantify the performance impact. I hope that we can take it, as I think that it's a nice improvement over the current API.

src/jit/bitset.h
sandreenko: Fixed, but maybe there is a better way to declare it, rather than `static const unsigned Done = (unsigned)-1;` and use with `BlockSetOps::Iter::Done;`

sandreenko:
I fixed bugs and now the results are the same.
I like that we do not return value in input arguments and do not have defines, but I don't like very long for expressions and BlockSetOps::Iter::Done that is defined in each iterator.

Also we have not decided could we get rid of GetAssertionIndex ? For example with the same` ILEGAL =(unsigned char)-1`.

sandreenko:

@dotnet-bot  test  Windows_NT x64 Debug Build and Test

BruceForstall:
> I hope that we can take it, as I think that it's a nice improvement over the current API.

I disagree somewhat. I prefer the existing:
```
int i;
while (iter.NextElem(&i))
```

over the proposed:
```
for (int i = iter.NextElem(); i != BitVecOps::Iter::Done; i = iter.NextElem())
```

So if there is no performance improvement, I wouldn't be in favor of this change.

pgavlin:
> > I hope that we can take it, as I think that it's a nice improvement over the current API.
> I disagree somewhat. I prefer the existing:
>
> ```
> int i;
> while (iter.NextElem(&i))
> ```
> over the proposed:
> 
> ```
> for (int i = iter.NextElem(); i != BitVecOps::Iter::Done; i = iter.NextElem())
> ```
>
> So if there is no performance improvement, I wouldn't be in favor of this change.

Ah, I was referring to the removal of the macro for constructing these iterators. I don't really feel one way or the other about the iteration API itself. It's likely that the performance ceiling for the original API is higher than the proposed API in any case, so perhaps we should retain that aspect.

sandreenko:
I will update PR with the 
```
int i;
while (iter.NextElem(&i))
```
without defines and with GetAssertionIndex commit.

sandreenko:

sandreenko:
Useful parts were merged as separate PRs.

PR #11956, opened 2017-05-29
Add MAY_THROW_DBZ_ON_OVF

parjong:
This commit revises exceptionhandling.cpp to use MAY_THROW_DBZ_ON_OVF. This change will make it easy to search the related code.

### DIFF
--- a/src/vm/exceptionhandling.cpp
+++ b/src/vm/exceptionhandling.cpp
@@ -17,6 +17,10 @@
 #include "eventtrace.h"
 #include "virtualcallstub.h"
 
+#if defined(_TARGET_X86_) || defined(_TARGET_AMD64_)
+#define MAY_THROW_DBZ_ON_OVF
+#endif
+
 #if defined(_TARGET_X86_)
 #define USE_CURRENT_CONTEXT_IN_FILTER
 #endif // _TARGET_X86_
@@ -4796,7 +4800,7 @@ VOID DECLSPEC_NORETURN DispatchManagedException(PAL_SEHException& ex, bool isHar
     throw std::move(ex);
 }
 
-#if defined(_TARGET_AMD64_) || defined(_TARGET_X86_)
+#ifdef MAY_THROW_DBZ_ON_OVF
 
 /*++
 Function :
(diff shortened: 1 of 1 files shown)


src/vm/exceptionhandling.cpp
```diff
@@ -17,6 +17,10 @@
 #include "eventtrace.h"
 #include "virtualcallstub.h"
 
+#if defined(_TARGET_X86_) || defined(_TARGET_AMD64_)
```
janvorli: @parjong the overflow is always reported as division by zero on amd64 / x86 Unix. I don't think that this change makes the code clearer, since all the code is in one file close to each other and commented in detail. On the contrary, with your change, someone reading the code has to lookup the definition of the MAY_THROW_DBZ_ON_OVF to find out that it is only related to x86 and amd64, which is the key thing here.
In general, I prefer  to not to create new defines unless the related ifdefs are scattered over multiple places and it is difficult to figure out what is some part of code related to easily. One example of a reasonable usage of a new define was the one you have added for the way of calling funclets.

parjong: @janvorli Thank you for comment. In general, I prefer to remove duplication, and thus 
this PR attempts to remove duplication in ifdef condition, and also make it explicit that the code in ``HandleHardwareException`` is related with the code for ``IsDivByZeroAnIntegerOverflow`` from the following structure:
```
#if defined(_TARGET_X86_) || defined(_TARGET_X64_)
...
#endif 
...
BOOL HandleHardwareException(PAL_SEHException* ex)
{
    ...
#if defined(_TARGET_X86_) || defined(_TARGET_X64_)
    ...
#endif 
    ...
}
```
This code seems to be fine as it as since it is currently only for X86 and X64, and it is unlikely that other architectures needs this code, or I would like to rename ``MAY_THROW_DBZ_ON_OVF`` into some other flag such as ``OVF_IS_SIGANLLED_AS_DBZ``. Could you share your opinion?

Feel free to close this PR if you think that the current code is better.

janvorli: @parjong I would prefer to close this PR. I think it is easier to reason about ifdefs indicating this is architecture specific rather than about a new define.

Issue #11782, opened 2017-05-22
[RyuJIT/arm32] Assertion failed 'spillTempSize <= lvaGetMaxSpillTempSize()'


mskvortsov:
```
#11 0xb0c6c8b2 in noWayAssertBodyConditional (cond=0xb0f13557 "spillTempSize <= lvaGetMaxSpillTempSize()", file=0xb0f11f7c "/home/mskvortsov/git/coreclr/src/jit/lclvars.cpp", line=6408) at /home/mskvortsov/git/coreclr/src/jit/error.cpp:522
#12 0xb0d14ce8 in Compiler::lvaAllocateTemps (this=0x964c0, stkOffs=-48, mustDoubleAlign=true) at /home/mskvortsov/git/coreclr/src/jit/lclvars.cpp:6408
#13 0xb0d1384a in Compiler::lvaAssignVirtualFrameOffsetsToLocals (this=0x964c0) at /home/mskvortsov/git/coreclr/src/jit/lclvars.cpp:5888
#14 0xb0d129b2 in Compiler::lvaAssignFrameOffsets (this=0x964c0, curState=Compiler::FINAL_FRAME_LAYOUT) at /home/mskvortsov/git/coreclr/src/jit/lclvars.cpp:4489
#15 0xb0c31cea in CodeGen::genFinalizeFrame (this=0x98484) at /home/mskvortsov/git/coreclr/src/jit/codegencommon.cpp:8280
#16 0xb0c31242 in CodeGen::genGenerateCode (this=0x98484, codePtr=0xbeffca64, nativeSizeOfCode=0xbeffcea0) at /home/mskvortsov/git/coreclr/src/jit/codegencommon.cpp:3033
```
```
JIT/Directed/perffix/primitivevt/mixed1_cs_do
JIT/Directed/perffix/primitivevt/mixed1_cs_ro
JIT/Methodical/Boxing/seh/_dbgtry_cs
JIT/Methodical/Boxing/seh/_reltry_cs
JIT/Methodical/fp/exgen/200w1d-01_cs_d
JIT/Methodical/fp/exgen/200w1d-01_cs_r
JIT/Methodical/fp/exgen/3w1d-01_cs_d
JIT/Methodical/fp/exgen/3w1d-01_cs_r
JIT/Methodical/fp/exgen/3w1d-02_cs_d
JIT/Methodical/fp/exgen/3w1d-02_cs_r
JIT/Methodical/MDArray/GaussJordan/classarr_cs_do
[...]

PR #11957, opened 2017-05-29
[RyuJit/ARM32] Adjust alignment for TYP_DOUBLE in tmpPreAllocateTemps

sjsinju:
In #11782,

Assertion failed 'spillTempSize <= lvaGetMaxSpillTempSize()' has been occurred in code generation phase.

Adjusting alignment for TYP_DOUBLE in tmpPreAllocateTemps is needed.

### DIFF
--- a/src/jit/regset.cpp
+++ b/src/jit/regset.cpp
@@ -3320,6 +3320,14 @@ void Compiler::tmpPreAllocateTemps(var_types type, unsigned count)
         tmpCount++;
         tmpSize += size;
 
+#ifdef _TARGET_ARM_
+        if (type == TYP_DOUBLE)
+        {
+            // Adjust tmpSize in case it needs alignment
+            tmpSize += TARGET_POINTER_SIZE;
+        }
+#endif // _TARGET_ARM_
+
         TempDsc* temp = new (this, CMK_Unknown) TempDsc(-((int)tmpCount), size, type);
 
 #ifdef DEBUG
--- a/src/jit/regset.cpp
+++ b/src/jit/regset.cpp
@@ -3323,7 +3323,8 @@ void Compiler::tmpPreAllocateTemps(var_types type, unsigned count)
 #ifdef _TARGET_ARM_
         if (type == TYP_DOUBLE)
         {
-            // Adjust tmpSize in case it needs alignment
+            // Adjust tmpSize to accommodate possible alignment padding.
+            // Note that at this point the offsets aren't yet finalized, so we don't yet know if it will be required.
             tmpSize += TARGET_POINTER_SIZE;
         }
 #endif // _TARGET_ARM_


sjsinju:
@dotnet-bot test Ubuntu x64 Formatting

src/jit/regset.cpp
```diff
@@ -3320,6 +3320,14 @@ void Compiler::tmpPreAllocateTemps(var_types type, unsigned count)
         tmpCount++;
         tmpSize += size;
 
+#ifdef _TARGET_ARM_
+        if (type == TYP_DOUBLE)
+        {
+            // Adjust tmpSize in case it needs alignment
+            tmpSize += TARGET_POINTER_SIZE;
```
hqueue: Is it always necessary?  What if tmpSize is already aligned ?

RussKeldorph:
@dotnet/jit-contrib @dotnet/arm32-contrib

src/jit/regset.cpp
pgavlin: This looks like it's a port from the `LEGACY_BACKEND` code for ARM32 in `tmpGetTemp`. Is that correct?

I would be interested in better understanding the behavior here as well.

cc @CarolEidt who may have some additional insight into codegen temp management.

CarolEidt: I believe that this should parallel the code in `Compiler::lvaAllocateTemps()` in lclvars.cpp, in which the tmpSize is padded only if necessary:
```
            if (mustDoubleAlign && (tempType == TYP_DOUBLE)) // Align doubles for x86 and ARM
            {
                noway_assert((compLclFrameSize % TARGET_POINTER_SIZE) == 0);

                if (((stkOffs + preSpillSize) % (2 * TARGET_POINTER_SIZE)) != 0)
                {
                    spillTempSize += TARGET_POINTER_SIZE;
                    lvaIncrementFrameSize(TARGET_POINTER_SIZE);
                    stkOffs -= TARGET_POINTER_SIZE;
                }
                // We should now have a double-aligned (stkOffs+preSpillSize)
                noway_assert(((stkOffs + preSpillSize) % (2 * TARGET_POINTER_SIZE)) == 0);
            }
```

sjsinju: @hqueue @pgavlin @CarolEidt I investigated what you guys said. But I think the size was not decided exactly in the register allocation phase yet. That's because there was no decided stack offset. So we might have an only just estimate of tmpSize in the preallocation of temp phase.

In this case(TYP_DOUBLE), since the offset would be aligned with multiples of (2*TARGET_POINTER_SIZE), considering the case of aligning by additional TARGET_POINTER_SIZE, It seems to need reserving the TARGET_POINTER_SIZE of additional temp size for aliging beforehand(The final decision of the stack offset is made in Compiler::lvaAllocateTemps. It is only a temporary space for estimating the size. It may not be necessary.).

hqueue: > It seems to need reserving the TARGET_POINTER_SIZE of additional temp size for aliging beforehand(The final decision of the stack offset is made in Compiler::lvaAllocateTemps. It is only a temporary space for estimating the size. It may not be necessary.).

I see, I got your point. So it's just a estimation for max size and I also agree that we don't know exact offset at the register allocation phase yet.

LGTM

CarolEidt: This makes sense now. I would request a comment such as:
```
// Adjust tmpSize to accommodate possible alignment padding.
// Note that at this point the offsets aren't yet finalized, so we don't yet know if it will be required.
```

sjsinju: @CarolEidt OK. I would apply that comment, Thank you.

hqueue:

PR #11962, opened 2017-05-29
Partially remove relocations from Class section of NGEN-ed images

ruben-ayrapetyan:
Partially remove relocations from Class section (ZapVirtualSectionType) of NGEN-ed images

On application, which is referenced in #10380 (comment), the following memory consumption changes occur for mappings of images:

  Rss: 8604 -> 8588 (0.1% improvement)
  Private_Dirty: 2708 -> 2572 (5% improvement)
  Private_Clean: 504 -> 544
  Shared_Clean: 5392 -> 5472

// The base for measurements is another than was used in #10380
// Fix of #10382 increased count of relocated images, so the absolute values of memory consumption
// became greater, and so absolute difference related to relocations removal also became greater.

Related issue: #10380

@Dmitri-Botcharnikov @gbalykov

### DIFF
--- a/src/debug/daccess/nidump.cpp
+++ b/src/debug/daccess/nidump.cpp
@@ -4745,7 +4745,7 @@ void NativeImageDumper::TraverseTypeHashEntry(void *pContext, PTR_EETypeHashEntr
              * all that much harm here (bloats m_discoveredMTs though,
              * but not by a huge amount.
              */
-            PTR_MethodTable mt(ptd->m_TemplateMT.GetValue());
+            PTR_MethodTable mt(ptd->GetTemplateMethodTableInternal());
             if (isInRange(PTR_TO_TADDR(mt)))
             {
                 m_discoveredMTs.AppendEx(mt);
@@ -6201,7 +6201,7 @@ void NativeImageDumper::TypeDescToString( PTR_TypeDesc td, SString& buf )
         if( td->IsArray()  )
         {
             //td->HasTypeParam() may also be true.
-            PTR_MethodTable mt = ptd->m_TemplateMT.GetValue();
+            PTR_MethodTable mt = ptd->GetTemplateMethodTableInternal();
             _ASSERTE( PTR_TO_TADDR(mt) );
             if( CORCOMPILE_IS_POINTER_TAGGED(PTR_TO_TADDR(mt)) )
             {
@@ -8793,7 +8793,7 @@ void NativeImageDumper::DumpTypeDesc( PTR_TypeDesc td )
     {
         PTR_ParamTypeDesc ptd(td);
         DisplayStartVStructure( "ParamTypeDesc", TYPEDESCS );
-        WriteFieldMethodTable( m_TemplateMT, ptd->m_TemplateMT.GetValue(), 
+        WriteFieldMethodTable( m_TemplateMT, ptd->GetTemplateMethodTableInternal(),
                                ParamTypeDesc, TYPEDESCS );
         WriteFieldTypeHandle( m_Arg, ptd->m_Arg, 
                               ParamTypeDesc, TYPEDESCS );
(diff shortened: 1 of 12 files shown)


ruben-ayrapetyan:

ruben-ayrapetyan:
@jkotas, Please, take a look.

src/vm/amd64/JitHelpers_InlineGetThread.asm
```diff
@@ -147,6 +147,11 @@ align 16
         jmp     JIT_Box
 NESTED_END JIT_BoxFastMP_InlineGetThread, _TEXT
 
+RELATIVE_PTR_LOAD macro Reg, BaseReg, Offset
```
jkotas: Adding more instructions to allocation helpers is always unfortunate. 

Since you are touching this, it would be better to pass TemplateMT as the argument to the allocation helper from the JITed code. It would make the code even faster than what it is today.

ruben-ayrapetyan: I'm currently working on this, and plan to open pull request for the change early next week.

- the new `TemplateMT` argument is added to `JIT_NewArr1` and other array allocation helpers. It seems that the `ArrayTypeDesc` argument can't be removed, because `JIT_NewArr1` uses it, and other helpers pass it to the `JIT_NewArr1`;
- the argument is embedded to JIT code as constant value, which is figured out during import, in case no runtime lookup is required for the `ArrayTypeDesc`;
- in cases, which require runtime lookup of the value, the `TemplateMT` value is also figured out in runtime by loading it from `ArrayTypeDesc` in JIT code

Could you, please, check whether the described direction is right?

jkotas: There should not be a new argument to the allocation helpers. The templateMT argument should be passed to the allocation helpers instead of the current ArrayTypeDesc argument.

jkotas: Historic context: TemplateMTs used to be shared between different ArrayTypeDescs in the past. It is not the case anymore, except for array of pointers and generics (https://github.com/dotnet/coreclr/blob/b258792e59b09060f54e0c9bbd31edc3e67d1ae8/src/vm/typedesc.cpp#L216). It should not be a problem in this case. There is 1:1 mapping between ArrayTypeDesc and TemplateMT for all practical purposes.

ruben-ayrapetyan: Thank you very much for your comments.

There is a prototype implementation in #12369, which perhaps supports all cases, with except of generics.

Specifically, it doesn't support `System.__Canon[]` arrays - the case requires runtime lookup from the generic type descriptor and signature.

As far as I see, the `JIT_GenericHandleClass` invocation is not generated by JIT, because method tables for `System.__Canon[]` is not considered to be canonical by `th.IsCanonicalSubtype` (and so, by `CEEInfo::embedGenericHandle`).

Is it good idea to add the case to `TypeHandle::IsCanonicalSubtype`?

jkotas: > the case requires runtime lookup from the generic type descriptor and signature.

This should be done via `ELEMENT_TYPE_NATIVE_ARRAY_TEMPLATE_ZAPSIG`.

> Is it good idea to add the case to `TypeHandle::IsCanonicalSubtype`?

I do not see a problem with it.

ruben-ayrapetyan:

ruben-ayrapetyan:
@dotnet-bot test Windows_NT x86 corefx_baseline

ruben-ayrapetyan:
Depends on #12369. Please, don't merge before the #12369.

src/vm/amd64/JitHelpers_InlineGetThread.asm
ruben-ayrapetyan: Fixed.

jkotas:

PR #11963, opened 2017-05-29
Remove relocations from SECTION_MethodDesc for ngened images for fields accessed from jit code

gbalykov:
This pull request removes relocations from MethodDesc section for ngened images.This is second part of changes, which affects jit code. First part #11394 

Related issue: #11363

Next fields are replaced with RelativePointers for Linux ARM in this pull request
```
NDirectMethodDesc::ndirect.m_pWriteableData
InstantiatedMethodDesc::m_pPerInstInfo
```

Next fields are replaced with RelativePointers for all platforms in the first pull request #11394 
```
InstantiatedMethodDesc::m_pDictLayout
InstantiatedMethodDesc::m_pWrappedMethodDesc
MethodDesc::m_pszMethodName
MethodImpl::pImplementedMD
MethodImpl::pdwSlots
NDirectMethodDesc::ndirect.m_pImportThunkGlue
NDirectMethodDesc::ndirect.m_pszEntrypointName
NDirectMethodDesc::ndirect.m_pszLibName
StoredSigMethodDesc::m_pSig
```

After the changes, SECTION_MethodDesc completely disappears from the table shown in https://github.com/dotnet/coreclr/issues/11363#issuecomment-299180256.

Estimation on Tizen Xamarin GUI Application (specifically, Puzzle.Tizen - see tables at #10380 (comment)) is about 268 kb (4%) savings of per-application CoreCLR's memory part in Fragile mode.

### DIFF
--- a/src/ToolBox/superpmi/superpmi-shared/methodcontext.cpp
+++ b/src/ToolBox/superpmi/superpmi-shared/methodcontext.cpp
@@ -1573,6 +1573,7 @@ void MethodContext::recGetCallInfo(CORINFO_RESOLVED_TOKEN* pResolvedToken,
             value.stubLookup.runtimeLookup.indirections = (DWORD)pResult->stubLookup.runtimeLookup.indirections;
             value.stubLookup.runtimeLookup.testForNull  = (DWORD)pResult->stubLookup.runtimeLookup.testForNull;
             value.stubLookup.runtimeLookup.testForFixup = (DWORD)pResult->stubLookup.runtimeLookup.testForFixup;
+            value.stubLookup.runtimeLookup.indirectFirstOffset = (DWORD)pResult->stubLookup.runtimeLookup.indirectFirstOffset;
             for (int i                                    = 0; i < CORINFO_MAXINDIRECTIONS; i++)
                 value.stubLookup.runtimeLookup.offsets[i] = (DWORDLONG)pResult->stubLookup.runtimeLookup.offsets[i];
         }
@@ -1583,6 +1584,7 @@ void MethodContext::recGetCallInfo(CORINFO_RESOLVED_TOKEN* pResolvedToken,
             value.stubLookup.runtimeLookup.indirections = (DWORD)0;
             value.stubLookup.runtimeLookup.testForNull  = (DWORD)0;
             value.stubLookup.runtimeLookup.testForFixup = (DWORD)0;
+            value.stubLookup.runtimeLookup.indirectFirstOffset = (DWORD)0;
             for (int i                                    = 0; i < CORINFO_MAXINDIRECTIONS; i++)
                 value.stubLookup.runtimeLookup.offsets[i] = (DWORDLONG)0;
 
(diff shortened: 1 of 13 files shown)


gbalykov:
@jkotas,

How do you think, for which configurations should this be enabled?

jkotas:
Linux ARM. For others (x64 in particular) - the JIT/perf team would need to take look.

gbalykov:
@jkotas 

In order to change fields, which are accessed from JIT code, to RelativePointers for Linux ARM only, I would like to use

```cpp
#if defined(PLATFORM_UNIX) && defined(_TARGET_ARM_)
```

However, I'm not sure about cross-compilation between Windows and Unix in this case.
Would it be better to add new define `TARGET_PLATFORM_UNIX`, which now would be defined as `PLATFORM_UNIX`, and would highlight the Windows/Unix cross-compilation differences?

janvorli:
@gbalykov it is highly unlikely we would add cross build native code on Unix for Windows or vice versa. For native build, you need all of the headers of the Unix libraries and the Linux libraries. That would be difficult to set up on Windows and there doesn't seem to be a benefit.

gbalykov:
@janvorli,

Is it unsafe to use

```cpp
#if defined(PLATFORM_UNIX) && defined(_TARGET_ARM_)
```

for cross compilation on Unix for fields of MethodDesc?

As @jkotas mentioned in https://github.com/dotnet/coreclr/pull/11394#issuecomment-299319152, the condition breaks cross compilation if used in JIT/EE interface. Is it also the case for the MethodDesc structure?

jkotas:
>  cross build native code on Unix for Windows or vice versa. For native build, you need all of the headers of the Unix libraries and the Linux libraries

This is not about cross build. This is about building crossgen that runs on Windows, but produces code for Unix. You do not need headers of the Unix libraries, etc. for that.

Today, the Windows-hosted Unix-targeting crossgen is limited to debugging the JIT: You can debug Unix JIT on Windows. However, it does not produce valid native images, and it would be quite a bit of work to make it work.

gbalykov:
@janvorli @jkotas 
Thanks for your comments. 

So, as we understand, this way

```cpp
#if defined(PLATFORM_UNIX) && defined(_TARGET_ARM_)
```

can be used for fields of MethodDesc and it doesn't break cross compilation, because `_TARGET_ARM_` is passed both to jit and crossgen. We will try to update this pull request accordingly.

gbalykov:
We want to use
```cpp
#if defined(PLATFORM_UNIX) && defined(_TARGET_ARM_)
```
only at declaration of 

```
NDirectMethodDesc::ndirect.m_pWriteableData
InstantiatedMethodDesc::m_pPerInstInfo
```

So, we want to add isRelative flag for RelativePointer,FixupPointer,RelativeFixupPointer and PlainPointer
```cpp
static constexpr bool isRelative = true;
```
which will be used instead of macro in all other places. Besides, we want to replace fields with PlainPointer in case `!PLATFORM_UNIX || !_TARGET_ARM_`, which will allow to simplify getters/setters for fields.

gbalykov:

gbalykov:
@jkotas 

Could you, please, take a look?

We've added ReadPointer functions, which could help to avoid mistakes like in #12004.

What do you think about adding new method, which will be overloaded on FixupPointerField and FixupRelativePointerField in the same way as ReadPointer?

This will allow to replace next code

```cpp
if (decltype(InstantiatedMethodDesc::m_pPerInstInfo)::isRelative)
{
    image->FixupRelativePointerField(this, offsetof(InstantiatedMethodDesc, m_pPerInstInfo));
 }
else
{
    image->FixupPointerField(this, offsetof(InstantiatedMethodDesc, m_pPerInstInfo));
}    
```
with 

```cpp
image->FixupPlainOrRelativePointerField(this, &InstantiatedMethodDesc::m_pPerInstInfo);
```

jkotas:
>What do you think about additionally adding new method,

Sounds fine to me. The VM changes look fine to me as well.

cc @dotnet/jit-contrib for the review of JIT changes.

MichalStrehovsky:
Should this kind of change invalidate the JitInterface GUID? It changed the contract for `CORINFO_RUNTIME_LOOKUP`. I don't particularly care anymore since #12071 is going to rev the GUID soon anyway, but it would have saved me scratching my head for 5 minutes on the CoreRT side.

PR #11978, opened 2017-05-30
Remove Debian Armel

gkhanna79:
Confirmed with @jyoungyun that Debian.8-armel is no longer required. Removing it from the identity package of CoreCLR.

@ericstj @weshaggard PTAL.

CC @Petermarcu @RussKeldorph

### DIFF
--- a/src/.nuget/dir.props
+++ b/src/.nuget/dir.props
@@ -112,9 +112,6 @@
 
   <ItemGroup Condition="$(SupportedPackageOSGroups.Contains(';Linux;'))">
     <OfficialBuildRID Include="alpine.3.4.3-x64" />
-    <OfficialBuildRID Include="debian.8-armel">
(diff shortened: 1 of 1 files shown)


src/.nuget/dir.props
```diff
@@ -112,9 +112,6 @@
 
   <ItemGroup Condition="$(SupportedPackageOSGroups.Contains(';Linux;'))">
     <OfficialBuildRID Include="alpine.3.4.3-x64" />
```
weshaggard: Why not remove the other non-portable RID's as well? Similar to what I'm doing in corefx?

gkhanna79: I will do that as part of a seperate PR.

weshaggard: sounds good.

PR #11980, opened 2017-05-30
Expose Enumerator members out of EnvironmentAugments.

AtsushiKan:
This syncs up EnvironmentAugment's surface area with
CoreRT's and unblocks the next step (which is
to eliminate the double dictionary by having
corefx call EnumerateEnvironmentVariables()
and eliminating GetEnvironmentVariables() permanently.)

(Note: This is unrelated to the other "future" item
of adopting the CoreRT versions of these files
through the shared partition. That requires some
careful testing to avoid regressing x-plat functionality
on CoreCLR so I'm leaving that up-for-grabs for
someone who actually works on Unix. We'll also want
to wait until those files have had some time to
settle down in CoreRT.)

### DIFF
--- a/src/mscorlib/src/System/Environment.cs
+++ b/src/mscorlib/src/System/Environment.cs
@@ -497,19 +497,6 @@ namespace System
             return GetEnvironmentVariableCore(variable, target);
         }
 
-        public static IDictionary GetEnvironmentVariables()
-        {
-            // separated from the EnvironmentVariableTarget overload to help with tree shaking in common case
-            return GetEnvironmentVariablesCore();
-        }
-
-        internal static IDictionary GetEnvironmentVariables(EnvironmentVariableTarget target)
-        {
-            ValidateTarget(target);
-
-            return GetEnvironmentVariablesCore(target);
-        }
-
         public static void SetEnvironmentVariable(string variable, string value)
         {
             ValidateVariableAndValue(variable, ref value);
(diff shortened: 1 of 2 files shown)


AtsushiKan:

AtsushiKan:

src/mscorlib/src/System/Environment.cs
```diff
@@ -701,20 +690,19 @@ private static IDictionary GetEnvironmentVariablesCore(EnvironmentVariableTarget
             }
             else
             {
-                throw new ArgumentException(SR.Format(SR.Arg_EnumIllegalVal, (int)target));
+                throw new ArgumentOutOfRangeException(nameof(target), target, SR.Format(SR.Arg_EnumIllegalVal, target));
```
jkotas: `EnumIllegalVal` is `ArgumentException` everywhere. Why change it `ArgumentOutOfRangeException` here?

AtsushiKan: This matches what ValidateTargets() throws (and what the CoreFx tests expect.)

Others that are throwing ArgumentException for a bad Target are wrong too but they're unreachable since someone calls ValidateTargets() first. I'm keeping the changes minimal on this one since I expect this to be replaced through the shared partition eventually.

src/mscorlib/src/System/Environment.cs
```diff
@@ -672,19 +614,66 @@ private static string GetEnvironmentVariableCore(string variable, EnvironmentVar
...
+            // CreateProcess page (null-terminated array of null-terminated strings).
+            // Note the =HiddenVar's aren't always at the beginning.
+
+            // Copy strings out, parsing into pairs and inserting into the table.
+            // The first few environment variable entries start with an '='.
+            // The current working directory of every drive (except for those drives
+            // you haven't cd'ed into in your DOS window) are stored in the 
+            // environment block (as =C:=pwd) and the program's exit code is 
+            // as well (=ExitCode=00000000).
+
```
jkotas: Change the CoreRT implementation to match this one? This one looks easier to follow.

I do not think it will be easy to use Span in the CoreRT one because Span is stack-only and cannot be used in iterators.

jkotas:
LGTM otherwise

src/mscorlib/src/System/Environment.cs
jkotas: I will be non-trivial for somebody to reverse engineer this once they get to it. I think it is useful to keep the code the same where possible, even if it was not moved to the shared partition yet...

PR #11985, opened 2017-05-30
[WIP - Do not merge] Default to portable build in dev and official builds

gkhanna79:
This change defaults the dev and official builds to be portable, removes the distro specific RIDs and pipeline build entries. This change does not include CI changes - I will have those as seperate commit.

@wtgodbe @ramarag Is there anything in test infrastructure (Helix or non-Helix) that relies on RID specific builds and could break?

@weshaggard @wtgodbe @chcosta PTAL

CC @RussKeldorph

### DIFF
--- a/build-packages.sh
+++ b/build-packages.sh
@@ -3,16 +3,15 @@
 usage()
 {
     echo "Builds the NuGet packages from the binaries that were built in the Build product binaries step."
-    echo "Usage: build-packages -BuildArch -BuildType [-portable]"
+    echo "Usage: build-packages -BuildArch -BuildType"
     echo "BuildArch can be x64, x86, arm, arm64 (default is x64)"
     echo "BuildType can be release, checked, debug (default is debug)"
-    echo "-portable - build for Portable Distribution"
     echo
     exit 1
 }
 
 __ProjectRoot="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-__PortableBuild=0
+__IsPortableBuild=1
 
 # Use uname to determine what the OS is.
 OSName=$(uname -s)
@@ -71,8 +70,9 @@ while :; do
         __Arch=$(echo $1| cut -d'=' -f 2)
         ;;
 
-        -portableBuild)
-            __PortableBuild=1
+        -PortableBuild=false)
+            unprocessedBuildArgs="$unprocessedBuildArgs $1"
+            __IsPortableBuild=0
             ;;
         *)
         unprocessedBuildArgs="$unprocessedBuildArgs $1"
@@ -81,7 +81,7 @@ while :; do
 done
 
 # Portable builds target the base RID
-if [ $__PortableBuild == 1 ]; then
+if [ $__IsPortableBuild == 1 ]; then
     if [ "$__BuildOS" == "Linux" ]; then
         export __DistroRid="linux-$__Arch"
     elif [ "$__BuildOS" == "OSX" ]; then
--- a/build.cmd
+++ b/build.cmd
@@ -97,9 +97,6 @@ set __BuildPackages=1
 set __BuildNativeCoreLib=1
 set __RestoreOptData=1
 
-REM Is this a portable build?
-set __IsPortableBuild=
-
 :Arg_Loop
 if "%1" == "" goto ArgsDone
 
@@ -116,8 +113,6 @@ if /i "%1" == "arm64"               (set __BuildArchArm64=1&set processedArgs=!p
 if /i "%1" == "debug"               (set __BuildTypeDebug=1&set processedArgs=!processedArgs! %1&shift&goto Arg_Loop)
 if /i "%1" == "checked"             (set __BuildTypeChecked=1&set processedArgs=!processedArgs! %1&shift&goto Arg_Loop)
 if /i "%1" == "release"             (set __BuildTypeRelease=1&set processedArgs=!processedArgs! %1&shift&goto Arg_Loop)
-
-if /i "%1" == "-portable"             (set __IsPortableBuild=-portable&set processedArgs=!processedArgs! %1&shift&goto Arg_Loop)
 
 REM All arguments after this point will be passed through directly to build.cmd on nested invocations
 REM using the "all" argument, and must be added to the __PassThroughArgs variable.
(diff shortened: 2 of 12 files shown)


build-packages.sh
```diff
@@ -3,16 +3,16 @@
 usage()
 {
     echo "Builds the NuGet packages from the binaries that were built in the Build product binaries step."
-    echo "Usage: build-packages -BuildArch -BuildType [-portable]"
+    echo "Usage: build-packages -BuildArch -BuildType [-nonportable]"
     echo "BuildArch can be x64, x86, arm, arm64 (default is x64)"
     echo "BuildType can be release, checked, debug (default is debug)"
-    echo "-portable - build for Portable Distribution"
+    echo "-nonportable - build for Distro-specific RID"
```
weshaggard: I considered changing this to non-portable in corefx but instead I just kept the portable flag and made our config.json always pass it unless you pass portable=false.  I'm not sure if you want to follow a similar pattern here as well.

gkhanna79:

gkhanna79:

build-packages.sh
janvorli: I think that it would be ideal to keep the option consistent between our repos. I'd prefer having less of those moments when I have to scan the long list of options in the help to find one that corresponds to another option in other repo.

weshaggard: I agree with consistency but we could make them consistent by changing them all to nonportable as well. I'm OK with either option.

gkhanna79:
@ellismg PTAL as well.

build-packages.sh
janvorli: I agree, that's what I meant.

gkhanna79: Thinking about it, I think passing "-PortableBuild=false" should suffice.  I have removed this.

build.cmd
```diff
@@ -117,7 +114,7 @@ if /i "%1" == "debug"               (set __BuildTypeDebug=1&set processedArgs=!p
 if /i "%1" == "checked"             (set __BuildTypeChecked=1&set processedArgs=!processedArgs! %1&shift&goto Arg_Loop)
 if /i "%1" == "release"             (set __BuildTypeRelease=1&set processedArgs=!processedArgs! %1&shift&goto Arg_Loop)
 
-if /i "%1" == "-portable"             (set __IsPortableBuild=-portable&set processedArgs=!processedArgs! %1&shift&goto Arg_Loop)
+if /i "%1" == "-nonportable"             (set __IsPortableBuild=-nonportable&set processedArgs=!processedArgs! %1&shift&goto Arg_Loop)
```
weshaggard: You are still using nonportable in various out scripts.

gkhanna79: Yes, I am cleaning that up as well.

gkhanna79:
@weshaggard PTAL. In addition to removing the "nonportable" flag and relying on "PortableBuild" to set the mode, I have also made PortableBuild a boolean property instead of an integer one.

I will merge this if green to get an official build going.

@wtgodbe Let me know if you see any Helix/official build related concern with this.

gkhanna79:
@ellismg PTAL as well.

gkhanna79:
I will merge this and get a build going.

src/.nuget/dir.props
```diff
@@ -26,8 +26,12 @@
     <SupportedPackageOSGroups Condition="'$(SupportedPackageOSGroups)' == ''">Windows_NT;OSX;Linux</SupportedPackageOSGroups>
     <SupportedPackageOSGroups>;$(SupportedPackageOSGroups);</SupportedPackageOSGroups>
 
+    <!-- Identify OS family based upon the RuntimeOS, which could be distro specific (e.g. osx.10.12) or 
+         portable (e.g. osx). 
+    -->
     <_runtimeOSVersionIndex>$(RuntimeOS.IndexOfAny(".-0123456789"))</_runtimeOSVersionIndex>
     <_runtimeOSFamily Condition="'$(_runtimeOSVersionIndex)' != '-1'">$(RuntimeOS.SubString(0, $(_runtimeOSVersionIndex)))</_runtimeOSFamily>
+    <_runtimeOSFamily Condition="'$(_runtimeOSVersionIndex)' == '-1'">$(RuntimeOS)</_runtimeOSFamily>
```
gkhanna79: If __DistroID is passed as a portable RID (which build.sh does), then runtimeOSVersionIndex is -1 while RuntimeOS family corresponds to the OS part of the portable RID> Hence, I added this check.

@weshaggard Let me know if you have any concerns with it.

gkhanna79:
Queued off https://devdiv.visualstudio.com/DevDiv/_Build/explorer?_a=summary&buildId=777284

config.json
```diff
@@ -462,12 +462,6 @@
             "RestoreDuringBuild": false
           }
         },
-        "portable": {
```
weshaggard: Don't you still need this to allow someone to pass portable=false?

src/.nuget/dir.props
weshaggard: I think that is fine as long as the places using _runtimeOSFamily can handle this being a full RID.

src/.nuget/dir.props
```diff
@@ -111,48 +115,22 @@
...
-    <OfficialBuildRID Include="debian.8-x64" />
-    <OfficialBuildRID Include="fedora.23-x64" />
-    <OfficialBuildRID Include="fedora.24-x64" />
     <OfficialBuildRID Include="linux-x64" />
     <OfficialBuildRID Include="linux-arm">
       <Platform>arm</Platform>
     </OfficialBuildRID>
-    <OfficialBuildRID Include="opensuse.42.1-x64" />
-    <OfficialBuildRID Include="rhel.7-x64" />
     <OfficialBuildRID Include="tizen.4.0.0-armel">
```
weshaggard: I think it is probably worth putting a comment on this RID similar to what we did in core-setup and corefx.

build.sh
```diff
@@ -674,8 +673,8 @@ while :; do
             __CrossBuild=1
             ;;
             
-        -portable)
-            __PortableBuild=1
+        -portablebuild=false)
```
weshaggard: Do we need to follow-up with tizen so they know to pass this parameter now?

config.json
gkhanna79: We can specify -PortableBuild=false explicitly on the command line.

build.sh
gkhanna79: Yes, I will let them know as well.

gkhanna79:
The queued build failed for the Test build due to known issues but the pipe build definitions built the product as expected for portable RID.

I will commence a PR to port this to 2.0.0.

PR #11986, opened 2017-05-30
Skip cast nodes for the tail call check.

### LINKED ISSUE #11983: CoreRT shows a problem with inlined tail calls.
Run ILC in Release mode with Debug RyuJit with such test:
```
using System;
using System.Reflection;

internal class ConstructorAttribute : Attribute
{
}

internal class Program
{
    private bool GetAttributeConstructor(ConstructorInfo c)
    {
        return CustomAttributeExtensions.IsDefined(c, typeof(ConstructorAttribute), true);
    }

    {
        new Program().GetAttributeConstructor(typeof(Program).GetTypeInfo().GetConstructor(Array.Empty<Type>()));
    }
}
```
Set ngenDump for GetAttributeConstructor, you will see:
```
Invoking compiler for the inlinee method Internal.LowLevelLinq.LowLevelEnumerable:Any(ref):bool :
IL to import:
IL_0000  02                ldarg.0
IL_0001  6f 83 00 00 0a    callvirt     0xA000083
IL_0006  0a                stloc.0
IL_0007  06                ldloc.0
IL_0008  6f d3 1d 00 06    callvirt     0x6001DD3
IL_000d  2a                ret
```
and 
```
 
***** BB01, stmt 7
               [000076] ------------             *  stmtExpr  void  (IL   ???...  ???)
               [000073] --C---------             |  /--*  cast      int <- bool <- int
               [000067] --C-G-------             |  |  \--*  call      int    Internal.LowLevelLinq.LowLevelEnumerable.Any
               [000068] ------------ arg0        |  |     +--*  const(h)  long   0x427DA0 method
               [000066] ------------ arg1        |  |     \--*  lclVar    ref    V07 tmp5
               [000075] -AC---------             \--*  =         bool
               [000074] D------N----                \--*  lclVar    bool   V05 tmp3

***** BB01, stmt 8
               [000025] ------------             *  stmtExpr  void  (IL   ???...  ???)
               [000024] --C---------             \--*  return    int
               [000023] --C---------                \--*  cast      int <- bool <- int
               [000077] ------------                   \--*  lclVar    int    V05 tmp3
```

Call 67 will be marked as tail call, but the morph condition for the return statement doesn't expect a cast and will blow up. 

It works fine with the release mode of the jit.
I am not yet sure about the right fix, but there is strange code in fgMorphCall:
```
            // Peel off casts
            while (treeWithCall->gtOper == GT_CAST)
            {
                noway_assert(!treeWithCall->gtOverflow());
                treeWithCall = treeWithCall->gtGetOp1();
            }
```
It is inside the `#ifdef Debug ` branch, that initially contained only assert checks, but some real code was added recently (#9405). @AndyAyersMS Is it right that we peel casts only in debug and only for the call node, but not for the return statement?

cc @MichalStrehovsky @dotnet/jit-contrib.

sandreenko:
Fix #11983

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -8137,8 +8137,16 @@ GenTreePtr Compiler::fgMorphCall(GenTreeCall* call)
             if (stmtExpr->gtOper == GT_ASG && info.compRetType != TYP_VOID)
             {
                 noway_assert(stmtExpr->gtGetOp1()->OperIsLocal());
+
+                GenTreePtr treeWithLcl = retExpr->gtGetOp1();
+                while (treeWithLcl->gtOper == GT_CAST)
+                {
+                    noway_assert(!treeWithLcl->gtOverflow());
+                    treeWithLcl = treeWithLcl->gtGetOp1();
+                }
+
                 noway_assert(stmtExpr->gtGetOp1()->AsLclVarCommon()->gtLclNum ==
-                             retExpr->gtGetOp1()->AsLclVarCommon()->gtLclNum);
+                             treeWithLcl->AsLclVarCommon()->gtLclNum);
             }
 
             fgRemoveStmt(compCurBB, nextMorphStmt);


sandreenko:
@dotnet-bot test Windows_NT Release pri1r2r
cc @dotnet/jit-contrib

PR #11990, opened 2017-05-30
Return 0 on success from GC Reliability Framework scripts

swgillespie:
The Reliability Framework returns 100 on success and 99 on failure, which is what previous internal test harnesses expected. Jenkins, however, expects 0 for success and 1 on failure, so it was reporting RF jobs as having failed despite them succeeding (and returning 100).

This should make the Windows RF job green (Ubuntu is still red: https://github.com/dotnet/coreclr/issues/11981)

@Maoni0 @adityamandaleeka PTAL? I tested both of these locally and they seemed to work OK.

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -1470,8 +1470,6 @@ def static calculateBuildCommands(def newJob, def scenario, def branch, def isPR
                         else if (isGcReliabilityFramework(scenario)) {
                             buildCommands += "tests\\runtest.cmd ${runtestArguments} GenerateLayoutOnly"
                             buildCommands += "tests\\scripts\\run-gc-reliability-framework.cmd ${arch} ${configuration}"
-                            Utilities.addArchival(newJob, "stdout.txt")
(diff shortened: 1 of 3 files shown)


swgillespie:
@dotnet-bot test Windows_NT x64 gc_reliability_framework

swgillespie:
@dotnet-bot test Windows_NT Release gc_reliability_framework

swgillespie:
Well, apparently that didn't work. I clearly don't understand powershell pipes well enough to do this correctly so I'm going to stop saving standard out. It seems to be useless (jenkins saves standard out anyway).

I also see that Jenkins only saves the "archival" of a build if the build succeeds. I was trying to use this to save the RF logs for failed builds, but Jenkins doesn't bother saving the logs if the build failed, which is unfortunate since that's the only time you'd want to look at the logs. I'm just going to remove the archival for now.

swgillespie:
@dotnet-bot test Windows_NT Release gc_reliability_framework
@dotnet-bot test CI please

Issue #11396, opened 2017-05-04
[RyuJIT][LSRA] Improve efficiency of VarToRegMaps


CarolEidt:
For large methods, where the efficiency is of most concern, it may be more efficient to keep a map from registers (which are likely to be densely populated) to variables rather than the reverse, as is currently done. This may impact efficiency in other areas, so requires investigation.

CarolEidt:
See the TODO-throughput comments in `LinearScan::getTempRegForResolution()` and `LinearScan::resolveEdge()`.

CarolEidt:
Also, consider abstracting the mapping of variables to registers.

Issue #11339, opened 2017-05-02
[RyuJIT][CQ] LSRA: Investigate separating cross-block lclVars from intra-block lclVars


CarolEidt:
Much of LSRA work involves dealing with sets of vars that are live at block boundaries.
We could improve efficiency by ensuring that those sets are dense, and using a smaller set.
We would only need a larger set for the liveness set that's used during traversals.
This could be beneficial if 1) there are a large number of vars that are only live within a single block, and 2) the cost of the extra sorting and updating of the "current" copy of live vars doesn't adversely impact throughput.
Also, it may be possible to reuse tracked numbers for different basic blocks, to keep the sets small.  The Intervals could still be different for the different lclNums with the same BB-local tracked var number.

CarolEidt:
In addition to ensuring that the sets of live lclVars at block boundaries are dense, we should ensure that non-register-candidates that are tracked (e.g. tracked structs that are not promoted) are not mixed in with the register candidates. Further, if the liveness info for non-register-candidates doesn't impact GC reporting, there is no reason to make those tracked after the optimization phases have completed.

mikedn:
Sort of related question: would it help LSRA to eliminate all SDSU intra-block lclvars?

CarolEidt:
> Sort of related question: would it help LSRA to eliminate all SDSU intra-block lclvars?

Yes, I think it would help (at a minimum it would eliminate the extra `RefPosition` associated with assigning the value. The question would be whether eliminating them would be more cost than the potential benefit.

PR #11994, opened 2017-05-30
Streamline LSRA for non-enregistered lclVars

CarolEidt:
- Don't create any lclVar intervals if no lclVars are ever enregistered.
- Don't create intervals for tracked lclVars that are not register candidates.
- Don't create any `RegToVarMap`s when no lclVars are ever enregistered.
- Don't do edge resolution if no lclVars are enregistered.
- Eliminate non-candidate lclVars from the sets used by LSRA.

Also:
- Change VarToRegMap to use regNumberSmall.
- Enhance `COMPLus_JitLsraStats` and `COMPlus_JitTimeLogCsv` dumping.

### DIFF
--- a/src/jit/compphases.h
+++ b/src/jit/compphases.h
@@ -22,6 +22,8 @@
 //     true.
 
 // clang-format off
+//                 enumName                      stringName                        shortName hasChildren measureIR
+//                                                                                                   parent
 CompPhaseNameMacro(PHASE_PRE_IMPORT,             "Pre-import",                     "PRE-IMP",  false, -1, false)
 CompPhaseNameMacro(PHASE_IMPORTATION,            "Importation",                    "IMPORT",   false, -1, true)
 CompPhaseNameMacro(PHASE_POST_IMPORT,            "Post-import",                    "POST-IMP", false, -1, false)
--- a/src/jit/lsra.cpp
+++ b/src/jit/lsra.cpp
@@ -1148,7 +1148,8 @@ LinearScan::LinearScan(Compiler* theCompiler)
     dumpTerse = (JitConfig.JitDumpTerseLsra() != 0);
 #endif // DEBUG
 
-    availableIntRegs = (RBM_ALLINT & ~compiler->codeGen->regSet.rsMaskResvd);
+    enregisterLocalVars = ((compiler->opts.compFlags & CLFLG_REGVAR) != 0) && compiler->lvaTrackedCount > 0;
+    availableIntRegs    = (RBM_ALLINT & ~compiler->codeGen->regSet.rsMaskResvd);
 
 #if ETW_EBP_FRAMED
     availableIntRegs &= ~RBM_FPBASE;
@@ -1809,11 +1810,17 @@ void LinearScan::doLinearScan()
 
 void LinearScan::recordVarLocationsAtStartOfBB(BasicBlock* bb)
 {
+    if (!enregisterLocalVars)
+    {
+        return;
+    }
     JITDUMP("Recording Var Locations at start of BB%02u\n", bb->bbNum);
     VarToRegMap map   = getInVarToRegMap(bb->bbNum);
     unsigned    count = 0;
 
-    VARSET_ITER_INIT(compiler, iter, bb->bbLiveIn, varIndex);
+    VarSetOps::AssignNoCopy(compiler, currentLiveVars,
+                            VarSetOps::Intersection(compiler, registerCandidateVars, bb->bbLiveIn));
+    VARSET_ITER_INIT(compiler, iter, currentLiveVars, varIndex);
     while (iter.NextElem(&varIndex))
     {
         unsigned   varNum = compiler->lvaTrackedToVarNum[varIndex];
@@ -1930,20 +1937,26 @@ void LinearScan::identifyCandidatesExceptionDataflow()
 
 bool LinearScan::isRegCandidate(LclVarDsc* varDsc)
 {
-    // Check to see if opt settings permit register variables
-    if ((compiler->opts.compFlags & CLFLG_REGVAR) == 0)
+    // We shouldn't be called if opt settings do not permit register variables.
+    assert((compiler->opts.compFlags & CLFLG_REGVAR) != 0);
+
+    if (!varDsc->lvTracked)
     {
         return false;
     }
+
+#if !defined(_TARGET_64BIT_)
+    if (varDsc->lvType == TYP_LONG)
+    {
+        // Long variables should not be register candidates.
+        // Lowering will have split any candidate lclVars into lo/hi vars.
+        return false;
+    }
+#endif // !defined(_TARGET_64BIT)
 
     // If we have JMP, reg args must be put on the stack
 
     if (compiler->compJmpOpUsed && varDsc->lvIsRegArg)
-    {
-        return false;
-    }
-
-    if (!varDsc->lvTracked)
     {
         return false;
     }
@@ -1962,16 +1975,25 @@ bool LinearScan::isRegCandidate(LclVarDsc* varDsc)
 
 void LinearScan::identifyCandidates()
 {
-
-    // Initialize the sets of lclVars that are used to determine whether, and for which lclVars,
-    // we need to perform resolution across basic blocks.
-    // Note that we can't do this in the constructor because the number of tracked lclVars may
-    // change between the constructor and the actual allocation.
-    VarSetOps::AssignNoCopy(compiler, resolutionCandidateVars, VarSetOps::MakeEmpty(compiler));
-    VarSetOps::AssignNoCopy(compiler, splitOrSpilledVars, VarSetOps::MakeEmpty(compiler));
-
-    if (compiler->lvaCount == 0)
+    if (enregisterLocalVars)
     {
+        // Initialize the set of lclVars that are candidates for register allocation.
+        VarSetOps::AssignNoCopy(compiler, registerCandidateVars, VarSetOps::MakeEmpty(compiler));
+
+        // Initialize the sets of lclVars that are used to determine whether, and for which lclVars,
+        // we need to perform resolution across basic blocks.
+        // Note that we can't do this in the constructor because the number of tracked lclVars may
+        // change between the constructor and the actual allocation.
+        VarSetOps::AssignNoCopy(compiler, resolutionCandidateVars, VarSetOps::MakeEmpty(compiler));
+        VarSetOps::AssignNoCopy(compiler, splitOrSpilledVars, VarSetOps::MakeEmpty(compiler));
+
+        // We set enregisterLocalVars to true only if there are tracked lclVars
+        assert(compiler->lvaCount != 0);
+    }
+    else if (compiler->lvaCount == 0)
+    {
+        // Nothing to do. Note that even if enregisterLocalVars is false, we still need to set the
+        // lvLRACandidate field on all the lclVars to false if we have any.
         return;
     }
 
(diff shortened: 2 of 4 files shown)


CarolEidt:
This results in a significant reduction of time spent in LSRA in MinOpts, as measured by JitTimeLogCsv. Here are the results for running JitTimeLogCsv against a SuperPmi collection of the x86 tests run with minopts:
image
Note that the percentage time for code generation went up (because the overall time went down), but the time spent there went down as well, presumably because `recordVarLocationsAtStartOfBB()` doesn't have to do anything.

CarolEidt:

sandreenko:
@CarolEidt could you please run perf and throughput jobs here?

> dotnet-bot test windows_nt x64 throughput
> dotnet-bot test windows_nt x86 throughput
> dotnet-bot test windows_nt x86 perf

src/jit/lsra.h
```diff
@@ -847,11 +847,17 @@ class LinearScan : public LinearScanInterface
     };
 
 private:
+    bool enregisterLocalVars()
+    {
+        return localVarIntervals != nullptr;
```
pgavlin: I think I'd be more comfortable just using a boolean field that is set in the constructor for this. The code in `identifyCandidates` in particular reads pretty strangely as it must set `localVarIntervals` before calling this method (and I would have expected it to use this method to decide whether or not to allocate `localVarIntervals` in the first place). To be concrete, something like:

```
private:
    bool m_enregisterLocalVars;
    bool enregisterLocalVars() const
    {
        return m_enregisterLocalVars;
    }

     ...

    LinearScan()
    {
        m_enregisterLocalVars = ((compiler->opts.compFlags & CLFLG_REGVAR) == 0) && compiler->lvaTrackedCount > 0;
    }
```

src/jit/lsra.cpp
```diff
@@ -4169,7 +4231,9 @@ BasicBlock* getNonEmptyBlock(BasicBlock* block)
 // insertZeroInitRefPositions: Handle lclVars that are live-in to the first block
 //
 // Notes:
-//    For each lclVar that is live-in to the first block:
+//    Prior to calling this method, 'currentLiveVars' must be set to the set of register
```
pgavlin: If this is simply to eliminate a redundant bit set intersection, I'd love to see an assert that verifies the contents of `currentLiveVars`.

src/jit/lsra.cpp
```diff
@@ -4845,55 +4914,64 @@ void LinearScan::buildIntervals()
 #ifdef DEBUG
 void LinearScan::dumpVarRefPositions(const char* title)
 {
-    printf("\nVAR REFPOSITIONS %s\n", title);
-
-    for (unsigned i = 0; i < compiler->lvaCount; i++)
+    if (enregisterLocalVars())
```
pgavlin: I'd just early-return here.

src/jit/lsra.cpp
```diff
@@ -6663,23 +6745,45 @@ bool LinearScan::isSecondHalfReg(RegRecord* regRec, Interval* interval)
...
+    VarToRegMap inVarToRegMap = getInVarToRegMap(currentBlock->bbNum);
     bool        hasCriticalInEdge = blockInfo[currentBlock->bbNum].hasCriticalInEdge;
 
-    const VARSET_TP* liveIn = &currentBlock->bbLiveIn;
+    VarSetOps::AssignNoCopy(compiler, currentLiveVars, VarSetOps::Intersection(compiler, registerCandidateVars, currentBlock->bbLiveIn));
 #ifdef DEBUG
     if (getLsraExtendLifeTimes())
     {
-        liveIn = &compiler->lvaTrackedVars;
+        VarSetOps::AssignNoCopy(compiler, currentLiveVars, compiler->lvaTrackedVars);
```
pgavlin: Does this need to be `lvaTrackedVars \intersect registerCandidateVars`?

src/jit/lsra.cpp
```diff
@@ -2475,16 +2504,47 @@ VarToRegMap LinearScan::getOutVarToRegMap(unsigned int bbNum)
...
+//
+// Arguments:
+//    bbVarToRegMap   - the map of interest
+//    trackedVarIndex - the lvVarIndex for the variable
+//    reg             - the register to which it is being mapped
+//
+// Return Value:
+//    None
+//
+void LinearScan::setVarReg(VarToRegMap bbVarToRegMap, unsigned int trackedVarIndex, regNumber reg)
```
pgavlin: We should probably turn `VarToRegMap` into a class and either moving these methods onto that class or giving it index operators with the same effect.

src/jit/lsra.cpp
```diff
@@ -6893,22 +6997,22 @@ void LinearScan::processBlockEndLocations(BasicBlock* currentBlock)
     assert(currentBlock != nullptr && currentBlock->bbNum == curBBNum);
     VarToRegMap outVarToRegMap = getOutVarToRegMap(curBBNum);
 
-    const VARSET_TP* liveOut = &currentBlock->bbLiveOut;
+    VarSetOps::AssignNoCopy(compiler, currentLiveVars, VarSetOps::Intersection(compiler, registerCandidateVars, currentBlock->bbLiveOut));
 #ifdef DEBUG
     if (getLsraExtendLifeTimes())
     {
-        liveOut = &compiler->lvaTrackedVars;
+        VarSetOps::AssignNoCopy(compiler, currentLiveVars, compiler->lvaTrackedVars);
```
pgavlin: Does this need to be `lvaTrackedVars \intersect registerCandidateVars`?

src/jit/lsra.cpp
```diff
@@ -9769,6 +9902,7 @@ void LinearScan::resolveEdge(BasicBlock*      fromBlock,
                              ResolveType      resolveType,
                              VARSET_VALARG_TP liveSet)
 {
+    VarToRegMap entryVarToRegMap = inVarToRegMaps[compiler->fgFirstBB->bbNum];
```
pgavlin: Is this actually used in this method? I can't seem to find it.

src/jit/compphases.h
```diff
@@ -22,6 +22,8 @@
 //     true.
 
 // clang-format off
+//                 enumName                      stringName                        shortName hasChildren measureIR
+//                                                                                                   parent
```
CarolEidt: I find it easier to have an aligned header than to count the fields and map them back.

src/jit/lsra.cpp
```diff
@@ -1809,11 +1809,16 @@ void LinearScan::doLinearScan()
...
+    if (!enregisterLocalVars())
+    {
+        return;
+    }
     JITDUMP("Recording Var Locations at start of BB%02u\n", bb->bbNum);
     VarToRegMap map   = getInVarToRegMap(bb->bbNum);
     unsigned    count = 0;
 
-    VARSET_ITER_INIT(compiler, iter, bb->bbLiveIn, varIndex);
+    VarSetOps::AssignNoCopy(compiler, currentLiveVars, VarSetOps::Intersection(compiler, registerCandidateVars, bb->bbLiveIn));
```
CarolEidt: I assume you mean the ration of reg-candidate lclVars to tracked lclVars (the difference between tracked and untracked is illustrated by the improvements Bruce got with #11698. I think there are a non-trivial number of tracked lclVars that are not reg candidates (e.g. non-enregisterable structs that are not promoted). I think that your suggestion could be considered as part of #11339.

src/jit/lsra.cpp
CarolEidt: That might be useful, but not as part of this PR. I would think that might be addressed by #11396

src/jit/lsra.cpp
CarolEidt: I generally avoid early returns where the additional indentation doesn't impair readability, but that's a matter of preference.

src/jit/lsra.cpp
CarolEidt: Done.

src/jit/lsra.cpp
CarolEidt: Yes I believe it should be.

CarolEidt: Actually, it should just use `registerCandidateVars` (which is a strict subset of `lvaTrackedVars`)

src/jit/lsra.cpp
CarolEidt: Ditto.

src/jit/lsra.cpp
CarolEidt: No, it isn't - and I don't even recall why I added it.

src/jit/lsra.h
CarolEidt: I'm fine with adding a boolean.

CarolEidt: Though note that `LinearScan` doesn't use the `m_` prefixes, and I will keep the existing style (which I prefer). Should we choose to change it, I would propose that we do it all or none.

CarolEidt:

CarolEidt:
@dotnet-bot test Windows_NT jitstressregs0x80
@dotnet-bot test Windows_NT x86 corefx_jitstressregs3
@dotnet-bot test Ubuntu x64 corefx_jitstressregs0x80

CarolEidt:
@pgavlin - I've incorporated/responded to your feedback. PTAL.
@BruceForstall - could you also have a look?
@sandreenko - the perf tests don't seem to have minOpts measurements (which is what this PR is primarily focused on), but the non-noisy results were positive.

src/jit/lsra.cpp
pgavlin: Seems fine.

pgavlin (approved):
Any assembly diffs?

CarolEidt:
Zero assembly diffs

pgavlin:
> Zero assembly diffs

Including with MinOpts, or with the usual opt levels?

CarolEidt:
> Including with MinOpts, or with the usual opt levels?

Just the usual opt levels (desktop and coreclr).

src/jit/lsra.h
```diff
@@ -852,6 +852,7 @@ class LinearScan : public LinearScanInterface
     Interval* getIntervalForLocalVar(unsigned varIndex)
     {
         assert(varIndex < compiler->lvaTrackedCount);
+        noway_assert(localVarIntervals[varIndex] != nullptr);
         return localVarIntervals[varIndex];
```
BruceForstall: Does this really need to be a noway_assert? It looks like this is a hot function, and this adds code to the release build path.

CarolEidt: No, I don't think it really does. I'll change it to a regular assert.

src/jit/lsra.cpp
```diff
@@ -1930,8 +1937,10 @@ void LinearScan::identifyCandidatesExceptionDataflow()
 
 bool LinearScan::isRegCandidate(LclVarDsc* varDsc)
 {
-    // Check to see if opt settings permit register variables
-    if ((compiler->opts.compFlags & CLFLG_REGVAR) == 0)
+    // We shouldn't be called if opt settings do not permit register variables.
+    assert((compiler->opts.compFlags & CLFLG_REGVAR) != 0);
+
+    if (!varDsc->lvTracked)
     {
```
BruceForstall: Unrelated: it seems like this code from identifyCandidates() should be here:
```
#if !defined(_TARGET_64BIT_)
        if (varDsc->lvType == TYP_LONG)
        {
            // Long variables should not be register candidates.
            // Lowering will have split any candidate lclVars into lo/hi vars.
            varDsc->lvLRACandidate = 0;
            continue;
        }
#endif // !defined(_TARGET_64BIT)
```
as:
```
#if !defined(_TARGET_64BIT_)
        if (varDsc->lvType == TYP_LONG)
        {
            // Long variables should not be register candidates.
            // Lowering will have split any candidate lclVars into lo/hi vars.
            return false;
        }
#endif // !defined(_TARGET_64BIT)
```

src/jit/lsra.cpp
```diff
@@ -1962,6 +1966,8 @@ bool LinearScan::isRegCandidate(LclVarDsc* varDsc)
 
 void LinearScan::identifyCandidates()
 {
+    // Initialize the set of lclVars that are candidates for register allocation.
+    VarSetOps::AssignNoCopy(compiler, registerCandidateVars, VarSetOps::MakeEmpty(compiler));
```
BruceForstall: Do this, and the other two bitsets below (resolutionCandidateVars, splitOrSpilledVars) need to be initialized even if `lvaCount ==0` (condition below)?

src/jit/lsra.cpp
```diff
@@ -12205,6 +12371,10 @@ void LinearScan::verifyFinalAllocation()
             VARSET_ITER_INIT(compiler, iter, currentBlock->bbLiveIn, varIndex);
             while (iter.NextElem(&varIndex))
             {
+                if (localVarIntervals[varIndex] == nullptr)
+                {
+                    continue;
+                }
                 regNumber regNum                  = getVarReg(inVarToRegMap, varIndex);
```
BruceForstall: Here (and below/elsewhere), shouldn't you use the API, i.e.:
```
Interval* interval                = getIntervalForLocalVar(varIndex);
if (interval == nullptr)
{
    continue;
}
```

src/jit/lsra.cpp
CarolEidt: Done.

src/jit/lsra.cpp
CarolEidt: No, they don't. I've made that change as well.

src/jit/lsra.cpp
CarolEidt: No - because if it is not a candidate, that method will assert. I think I had missed a couple of places which I've now updated, but this one (and the others like it) can't use that. I've added an assert, though.

justmehlc76 (commented):
PITFRUMP ("◇smart_move◇") [1991 NiseMac] + [1992 NiseMac] = +9500

CarolEidt:
@BruceForstall I've made the changes you suggested, which required adding some additional checks for `enregisterLocalVars`. Could you take a look?

BruceForstall (approved):

PR #11998, opened 2017-05-31
Remove forward declarations

parjong:
Some debugging functions have a forward declaration, but these forward declarations seems to be not necessary. 

This commit attempts to remove these declarations.


parjong:
@janvorli Please take a look :)

janvorli:
@parjong, I don't see much benefit from removing forward declarations. Our contribution guidelines say "DON'T make PRs for style changes." This one seems to me to be one of those.

parjong:
I usually fell uncomfortable when there is unnecessary code duplication (function declarations in this case), and attempt to minimize it, but it is just my preference.

I'll close if you feel that this PR is just for style changes.

PR #12002, opened 2017-05-31
[x64/Linux] Use correct argument registers in InterpreterStub

parjong:
This commit allows us to interpret Main in a simple "Hello, World" application in x64/Linux.

### DIFF
--- a/src/vm/amd64/cgencpu.h
+++ b/src/vm/amd64/cgencpu.h
@@ -268,6 +268,14 @@ struct CalleeSavedRegistersPointers {
 #define THIS_kREG kRCX
 #endif
 
+#ifdef PLATFORM_UNIX
+#define ARG1_kREG kRDI
+#define ARG2_kREG kRSI
+#else // PLATFORM_UNIX
+#define ARG1_kREG kRCX
+#define ARG2_kREG kRDX
+#endif // !PLATFORM_UNIX
+
 #ifdef UNIX_AMD64_ABI
 
 #define NUM_FLOAT_ARGUMENT_REGISTERS 8
--- a/src/vm/amd64/cgencpu.h
+++ b/src/vm/amd64/cgencpu.h
@@ -263,18 +263,16 @@ struct CalleeSavedRegistersPointers {
 #ifdef UNIX_AMD64_ABI
 #define THIS_REG RDI
 #define THIS_kREG kRDI
+
+#define ARGUMENT_kREG1 kRDI
+#define ARGUMENT_kREG2 kRSI
 #else
 #define THIS_REG RCX
 #define THIS_kREG kRCX
-#endif
 
-#ifdef PLATFORM_UNIX
-#define ARG1_kREG kRDI
-#define ARG2_kREG kRSI
-#else // PLATFORM_UNIX
-#define ARG1_kREG kRCX
-#define ARG2_kREG kRDX
-#endif // !PLATFORM_UNIX
+#define ARGUMENT_kREG1 kRCX
+#define ARGUMENT_kREG2 kRDX
+#endif
 
 #ifdef UNIX_AMD64_ABI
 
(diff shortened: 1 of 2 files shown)


src/vm/interpreter.cpp
```diff
@@ -1431,8 +1431,8 @@ CorJitResult Interpreter::GenerateInterpreterStub(CEEInfo* comp,
         sl.X86EmitPopReg(kEBP);
         sl.X86EmitReturn(static_cast<WORD>(argState.callerArgStackSlots * sizeof(void*)));
 #elif defined(_AMD64_)
-        // EDX has "ilArgs" i.e., just the point where registers have been homed.
-        sl.X86EmitIndexLeaRSP(kEDX, static_cast<X86Reg>(kESP_Unsafe), 8);
+        // Pass "ilArgs", i.e. just the point where registers have been homed, as 2nd arg
+        sl.X86EmitIndexLeaRSP(kRSI, static_cast<X86Reg>(kESP_Unsafe), 8);
```
janvorli: This is not correct, kRSI is the 2nd argument on Unix only. On Windows, it is the kEDX that was there before. The best way would be to introduce ARGUMENT_kREG1 / ARGUMENT_kREG2 defines and put them to cgencpu.h for amd64 next to the THIS_REG definition.

src/vm/interpreter.cpp
```diff
@@ -1450,10 +1450,10 @@ CorJitResult Interpreter::GenerateInterpreterStub(CEEInfo* comp,
 #endif
         {
             // For a non-ILStub method, push NULL as the StubContext argument.
-            sl.X86EmitZeroOutReg(kRCX);
-            sl.X86EmitMovRegReg(kR8, kRCX);
+            sl.X86EmitZeroOutReg(kRDX);
+            sl.X86EmitMovRegReg(kR8, kRDX);
         }
-        sl.X86EmitRegLoad(kRCX, reinterpret_cast<UINT_PTR>(interpMethInfo));
+        sl.X86EmitRegLoad(kRDI, reinterpret_cast<UINT_PTR>(interpMethInfo));
```
janvorli: Again, this is not correct for Windows.

src/vm/interpreter.cpp
```diff
@@ -1450,10 +1450,10 @@ CorJitResult Interpreter::GenerateInterpreterStub(CEEInfo* comp,
 #endif
         {
             // For a non-ILStub method, push NULL as the StubContext argument.
-            sl.X86EmitZeroOutReg(kRCX);
-            sl.X86EmitMovRegReg(kR8, kRCX);
+            sl.X86EmitZeroOutReg(kRDX);
```
janvorli: Please use the same register as the one used in the X86EmitRegLoad below

src/vm/interpreter.cpp
parjong: @janvorli Thanks you for comment. I wasn't aware of Linux/Windows ABI difference on x64. I plan to revise PR as you suggested, but 2nd argument is passed via ``EDX`` (instead of ``RDX``)?

janvorli: No, it should be kRDX, I guess kEDX and kRDX just have the same value and so there was no problem.

The amd64 calling convention is very different between windows and Unix. Windows use just four general purpose registers - RCX, RDX, R8, R9 for arg passing, while Unix uses six - RDI, RSI, RDX, RCX, R8 and R9. Moreover, the rules for passing structs in registers and on stack are completely different between the platforms.

parjong:
I'll revise as you suggested. I though that ``ARGUMENT_kREG1`` should be defined as ``kRDX``, but ``kEDX`` (instead of ``kRDX``) should be used here due to some restriction that I am not aware of.

janvorli:
@parjong kRDX and kEDX constants have the same value. I guess the fact that the kEDX was used here seems like a copy paste error or something like that.

src/vm/amd64/cgencpu.h
```diff
@@ -268,6 +268,14 @@ struct CalleeSavedRegistersPointers {
 #define THIS_kREG kRCX
 #endif
 
+#ifdef PLATFORM_UNIX
```
janvorli: Can you please put it into the previous ifdef? Also, the UNIX_AMD64_ABI is a better ifdef for this. And last, x86 uses ARGUMENT_REG1 notation, so it would be nice to use the ARGUMENT_kREG1 as I've suggested before.

PR #12004, opened 2017-05-31
Fix issue with RelativePointer instead of RelativeFixupPointer

gbalykov:
Fix issue introduced with #11394 with RelativePointer instead of RelativeFixupPointer


gbalykov:
@jkotas, please, take a look

PR #12010, opened 2017-05-31
Post-Serialization cleanup fixes to Reflection

AtsushiKan:
This is based on the following principles:

- We should not be making changes to instance/virtual methods
in the abstract base types for Reflection. These are apis
in themselves to third party classes that subclass these types.
Even if the Runtime chooses not to make its own implementations
of these type serializable, third party classes have the right
to make that choice for themselves and build off the base class
serialization infrastructure that was shipped before.

RuntimeAssembly and RuntimeModule still override with PNSE
so we'll get the desired exception for those cases.

- The Runtime's own implementations of these types will not be
serializable. Removing all [Serializable] attributes and
serialization interfaces. Runtime types are internal types and
thus have no duty to implement ISerializable if the public
abstract base type does not.

### DIFF
--- a/src/mscorlib/src/System/Reflection/RuntimeConstructorInfo.cs
+++ b/src/mscorlib/src/System/Reflection/RuntimeConstructorInfo.cs
@@ -6,12 +6,11 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.Contracts;
 using System.Globalization;
-using System.Runtime.Serialization;
 using RuntimeTypeCache = System.RuntimeType.RuntimeTypeCache;
 
 namespace System.Reflection
 {
-    internal sealed class RuntimeConstructorInfo : ConstructorInfo, ISerializable, IRuntimeMethodInfo
+    internal sealed class RuntimeConstructorInfo : ConstructorInfo, IRuntimeMethodInfo
     {
         #region Private Data Members
         private volatile RuntimeType m_declaringType;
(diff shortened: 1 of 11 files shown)


morganbr:
@AtsushiKan , would these changes need to be included in .NET Core 2.0 or is the PNSE behavior acceptable for that?

src/mscorlib/src/System/Reflection/RuntimeConstructorInfo.cs
```diff
@@ -462,18 +461,5 @@ public override Object Invoke(BindingFlags invokeAttr, Binder binder, Object[] p
             return RuntimeMethodHandle.InvokeMethod(null, null, sig, true);
         }
         #endregion
-
-        #region ISerializable Implementation
-        public void GetObjectData(SerializationInfo info, StreamingContext context)
-        {
-            throw new PlatformNotSupportedException();
```
morganbr: Is there any concern that calling GetObjectData on Runtime*Infos will throw NotImplemented, which is different than every other type we don't support serializing?

AtsushiKan: How would you call GetObjectData() on a ConstructorInfo? The method doesn't exist anywhere in the contract type hierarchy (or even in the implementation stack for that matter.)

GetObjectData() only exists on Module and Assembly, and the Runtime classes for those do override with PNSE.

AtsushiKan:
I think the parts that are in the shared partition should be in .NET Core 2.0. 

The internal Runtime level stuff, I don't care about.

AtsushiKan:
@morganbr - do you have any further concerns before I merge?

src/mscorlib/src/System/Reflection/RuntimeConstructorInfo.cs
morganbr: Got it. That sounds reasonable (somebody could have been casting to ISerializable on NetFX, but given that the interface isn't on the base types, that's past the promise of NS 2.0)

morganbr:
@AtsushiKan , per @danmosemsft , please take this through Shiproom separately if you'd like it to be in 2.0.

AtsushiKan:
@morganbr @danmosemsft 

What's the process for Shiproom? 

An additional wrinkle is that I can't actually create a commit for the changes I want to go into release/2.0 because at this moment in time, release/2.0 has the shared files exactly the way I want them. My commit simply reverts your changes to those three files - and those changes are still in PR https://github.com/dotnet/coreclr/pull/12020.

PR #12011, opened 2017-05-31
Fix TestNugetRuntimeId.

pgavlin:
This property was always being set to `OS-x64`, which was causing
libraries that targeted the wrong architecture to be deployed into
CORE_ROOT on x86. This change fixes this setting to use
`OS-$(__BuildArch)`.

### DIFF
--- a/tests/src/Common/test_dependencies/test_dependencies.csproj
+++ b/tests/src/Common/test_dependencies/test_dependencies.csproj
@@ -30,7 +30,7 @@
     <TargetFramework>netcoreapp2.0</TargetFramework>
     <TargetFrameworkIdentifier>.NETCoreApp</TargetFrameworkIdentifier>
     <PackageTargetFallback>$(PackageTargetFallback);dnxcore50;netcoreapp1.1;portable-net45+win8</PackageTargetFallback>
-    <RuntimeIdentifiers>win-x64;win-x86;win7-x86;win7-x64;ubuntu.14.04-x64;ubuntu.16.04-x64;ubuntu.16.10-x64;osx.10.12-x64;osx-x64;centos.7-x64;rhel.7-x64;debian.8-x64;fedora.24-x64;linux-x64</RuntimeIdentifiers>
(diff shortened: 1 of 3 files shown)


pgavlin:
@dotnet-bot test x86 Windows_NT Checked gcstress0xc

pgavlin:

pgavlin:
@wtgodbe PTAL. I believe that this should fix the issues we've been seeing in x86 GC stress wherein an x64-targeting copy of coredistools.dll is being deployed to CORE_ROOT. I am hopeful that this also fixes the similar issues we're seeing with this job on Linux, but I'm not certain.

cc @RussKeldorph

pgavlin:
@dotnet-bot test Ubuntu x64 Checked gcstress0xc

pgavlin:
Hm, it doesn't look like this works as expected. I'll keep digging.

pgavlin:
Okay, so this is at least part of the solution. It appears that we may also need to update the [Microsoft.NETCore.CoreDisTools metapackage](https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.CoreDisTools) to include references to the latest versions of the various runtime packages.

pgavlin:
cc @swaroop-sridhar

swaroop-sridhar:
Yes, my read of the failure is that the test failed because it couldn't find libcoredistools.so.
We obviously cannot use coredistools.dll on Linux. So, it has to be restored explicitly.

wtgodbe:
OK, I'll work on putting together a changeset to do that.

pgavlin:

pgavlin:
@dotnet-bot 

pgavlin:
@dotnet-bot test Windows_NT Checked gcstress0xc

pgavlin:
@dotnet-bot test Windows_NT x86 gcstress0xc

pgavlin:
@dotnet-bot test Windows_NT gcstress0xc

pgavlin:
@dotnet-bot 
test Windows_NT x86 Checked gcstress0xc
test Windows_NT gcstress0xc

pgavlin:
@wtgodbe we'll still want this change--can you take a quick look?

pgavlin:
OS X and NT x64 failure are infra and what appears to be a flaky test.

PR #12017, opened 2017-05-31
Remove GCX_PREEMP call when allocating a buffer.

### LINKED ISSUE #12014: MusicStore Hangs When Tracing Enabled

brianrob:
Fixes #12014.

This fixes a deadlock that can occur if tracing is enabled when the first background GC is run.  On first background GC, the following steps occur:

1. The thread invoking the GC sets a flag marking the GC as started.
2. The thread invoking the GC creates a new background GC thread and waits for it to enter its thread proc.
3. The new background GC thread calls Thread::HasStarted which writes an event stating that the thread was created.
4. Because this is the first event written on this thread, we a new buffer is allocated to hold the event.  During this, we switch to preemptive mode.
5. After allocation, the background GC thread attempts to switch back to cooperative mode, but gets stuck because a GC has been started.

Thus we, have a deadlock because the background GC thread is waiting on the GC that it is about to do to be completed before it switches back to cooperative mode.

The fix for this is to not switch to preemptive mode when allocating buffers.  The cost here is relatively low.

### DIFF
--- a/src/vm/eventpipebuffermanager.cpp
+++ b/src/vm/eventpipebuffermanager.cpp
@@ -283,7 +283,12 @@ bool EventPipeBufferManager::WriteEvent(Thread *pThread, EventPipeEvent &event, 
     // Check to see if we need to allocate a new buffer, and if so, do it here.
     if(allocNewBuffer)
     {
-        GCX_PREEMP();
+        // We previously switched to preemptive mode here, however, this is not safe and can cause deadlocks.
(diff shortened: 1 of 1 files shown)


brianrob:
I am re-running some tests against release/2.0.0 with this change and will merge once this is done.

Issue #12015, opened 2017-05-31
Fix failures in ETW logging on 4GB aware 32 bit processes (second try)


vancem:
PR #11941 was a first attempt at a fix, but it was not correct.  

@sharwell notices this through review (good job), but the fix was already merged.   This issue tracks following up on the correct fix and testing.

vancem:
The good news is the fix is trivial.  

            public unsafe IntPtr DataPointer { get { return (IntPtr)m_Ptr; } set { m_Ptr = unchecked((ulong)(void*)value); } }

Need to do testing however.  

@brianrob

sharwell:
:memo: I think the getter needs to be updated as well, as can be shown by the following:

```
> ulong ptr = 0xdeadf00d;
> (IntPtr)ptr
Arithmetic operation resulted in an overflow.
  + System.IntPtr.op_Explicit(long)
```

For that, I think you need this:

```csharp
return (IntPtr)(void*)m_Ptr;
```

vancem:
This fix has been merged in PR #12022, we can close the issue.

PR #12022, opened 2017-05-31
Additional fix for ETW failure in 32 process on 64 bit machines.

vancem:
The commit associated with Pull Request
    Fix failures in ETW logging on 4GB aware 32 bit processes #11941

Was incorrect.   Some long were changed to ulong, but the critical casting an IntPtr to ulong still does
sign extension and not zero extension.   Need to cast to a void* first.  This fixes this.

@sharwell  @brianrob 

Fixes issue #12015

### DIFF
--- a/src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
+++ b/src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs
@@ -1134,7 +1134,7 @@ namespace System.Diagnostics.Tracing
             /// Address where the one argument lives (if this points to managed memory you must ensure the
             /// managed object is pinned.
             /// </summary>
-            public IntPtr DataPointer { get { return (IntPtr)m_Ptr; } set { m_Ptr = unchecked((ulong)value); } }
+            public unsafe IntPtr DataPointer { get { return (IntPtr)m_Ptr; } set { m_Ptr = unchecked((ulong)(void*)value); } }
             /// <summary>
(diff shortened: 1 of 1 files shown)


vancem:
This is not particularly easy thing to test.  

I did ad hoc testing using the following program 

[EtwFailureTest.cs.txt](https://github.com/dotnet/coreclr/files/1042996/EtwFailureTest.cs.txt)

On a stock 32 bit runtime if you turn on ETW logging (e.g. PerfView /Providers=*MySource collect) AND you modify the 32 bit dotnet.exe  with editbin /LargeAddressAware dotnet.exe) then the program above will fail with Write exception without the fix, and with the fix it runs to completion.  

Thus to allow this to be tested automatically we would first need a fix to the dotnet.exe to become large address aware.   We also have to figure out what happens on Linux.    Once the harnessing issues are dealt with adding the test is not a big deal The normal EventSource harness in the CoreFX repo does do ETW collection and will naturally notice a failed event write).   

I view having a regression test for this to be low priority as the likelihood of this every regressing is very low (after all we got this first without even FINDING the bug, and it is not likely to change.   I think it is sufficient that we have unit tested it.     If anyone feels otherwise, they can make an issue to track adding an actual regression test.  

@brianrob  when we port this fix to Desktop, be sure to include both commits (you may wish to make a note in the Desktop bug).

vancem:
I just saw your feedback on the getter.  I have pushed a fix.

PR #12025, opened 2017-06-01
Skip argTable assignment for split struct when remorphing

hseok-oh:
- Skip argTable no. of register & stack slot assignment for split struct when remorphing
- Modify assertion check condition for structs morphed to GT_FIELD_LIST

cc/ @dotnet/arm32-contrib

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -4147,8 +4147,8 @@ GenTreeCall* Compiler::fgMorphArgs(GenTreeCall* call)
                             NYI_ARM("Struct split between integer registers and stack");
 #endif // !LEGACY_BACKEND
                             // This indicates a partial enregistration of a struct type
-                            assert((isStructArg) || argx->OperIsFieldList() ||
-                                   argx->OperIsCopyBlkOp() || (argx->gtOper == GT_COMMA && (args->gtFlags & GTF_ASG)));
+                            assert((isStructArg) || argx->OperIsFieldList() || argx->OperIsCopyBlkOp() ||
+                                   (argx->gtOper == GT_COMMA && (args->gtFlags & GTF_ASG)));
                             unsigned numRegsPartial = size - (intArgRegNum - MAX_REG_ARG);
                             assert((unsigned char)numRegsPartial == numRegsPartial);
                             call->fgArgInfo->SplitArg(argIndex, numRegsPartial, size - numRegsPartial, reMorphing);
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -1207,10 +1207,12 @@ fgArgTabEntryPtr fgArgInfo::AddStkArg(unsigned   argNum,
     return curArgTabEntry;
 }
 
+#ifdef DEBUG
 void fgArgInfo::RemorphReset()
 {
     nextSlotNum = INIT_ARG_STACK_SLOT;
 }
+#endif
 
 fgArgTabEntry* fgArgInfo::RemorphRegArg(
     unsigned argNum, GenTreePtr node, GenTreePtr parent, regNumber regNum, unsigned numRegs, unsigned alignment)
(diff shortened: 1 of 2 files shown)


RussKeldorph:
@dotnet/jit-contrib

src/jit/morph.cpp
```diff
@@ -1367,7 +1367,7 @@ void fgArgInfo::RemorphStkArg(
     nextSlotNum += numSlots;
 }
 
-void fgArgInfo::SplitArg(unsigned argNum, unsigned numRegs, unsigned numSlots)
+void fgArgInfo::SplitArg(unsigned argNum, unsigned numRegs, unsigned numSlots, bool isReMorph)
```
CarolEidt: It is quite unfortunate that this method has the side-effect of updating the stateful field `nextSlotNum`. Otherwise I would suggest avoiding the call to this method when remorphing.

hseok-oh: @CarolEidt `RemorphStkArg` also updates `nextSlotNum`. But in remorphing, `nextSlotNum` seems to be used for assertion check only. IMO, we can modify to skip updating `nextSlotNum` using `DEBUG` flag and `reMorphing` check.

hseok-oh:
@JosephTremoulet This assertion check may block struct argument morphed to `GT_FIELD_LIST` on RyuJIT/ARM32, but there is no correctness issue yet because legacyJIT/ARM32 does not use `GT_FIELD_LIST`. And some existing tests can check this.

hseok-oh:

hseok-oh:
@CarolEidt I've updated to skip reset & updating `nextSlotNum` when it is remorphing and release build. PTAL

CarolEidt:
I'm not totally in favor of the change to avoid updating `nextSlotNum` in release build only. I don't think it is a significant cost to keep it, and since it is a field that is actually used in release builds, I think it is risky to make its behavior different for release and debug/checked. The alternative might be to avoid updating it at all when remorphing, but I'm not sure that's a better option. @JosephTremoulet do you have other thoughts on this?
However, with the change to protect the call to `SplitArg()` with `!reMorphing`, you don't need that extra argument to `SplitArg()`.

JosephTremoulet:
> @JosephTremoulet do you have other thoughts on this?

No, but I'll second the opinion that the risk from having `nextSlotNum` diverge in debug vs. release seems worse than the entanglement of `SplitArg` side-effecting `nextSlotNum`.

We could consider making it the caller's responsibility to update `nextSlotNum` after calling `SplitArg`, and then not call `SplitArg` when remorphing, but IMO it's better to leave the call in and have the assertion that the arg tab entry lines up.

We could also consider specializing rather than parameterizing `SplitArg`, i.e. have something with a name like `VerifySplitArg` that gets called when remorphing, but IMO it's not worth the trouble or the code duplication.

So I'd be inclined to keep the original fix.

hseok-oh:
@JosephTremoulet @CarolEidt I also agree to make same behavior for release and debug/checked. I've updated to return to original fix, and remove `isReMorphing` argument by using `argsComplete`. 
PTAL

hseok-oh:

src/jit/morph.cpp
```diff
@@ -1319,7 +1319,9 @@ void fgArgInfo::RemorphStkArg(
         }
     }
 
+#ifdef DEBUG
```
JosephTremoulet: You should undo this, too, I think.

hseok-oh: Thanks. Fixed it.

src/jit/morph.cpp
```diff
@@ -1383,10 +1383,18 @@ void fgArgInfo::SplitArg(unsigned argNum, unsigned numRegs, unsigned numSlots)
...
+        assert(curArgTabEntry->isSplit == true);
+        assert(curArgTabEntry->numRegs == numRegs);
+        assert(curArgTabEntry->numSlots == numSlots);
+    }
+    else
+    {
+        curArgTabEntry->isSplit  = true;
+        curArgTabEntry->numRegs  = numRegs;
+        curArgTabEntry->numSlots = numSlots;
+    }
```
CarolEidt: Very good - this works nicely.

Issue #11997, opened 2017-05-31
[Question] Who uses libSystem.Globalization.Native.a?


parjong:
CLR build produces both ``System.Globalization.Native.so`` and ``libSystem.Globalization.Native.a``, but it seems that ``libSystem.Globalization.Native.a`` is unnecessary. 

Is it necessary? I find that dotnet shared framework ubuntu package also includes this file.

janvorli:
The CoreRT project uses it.

tarekgh:
https://github.com/dotnet/corert/blob/e4db93bb495383782b4e1e6c5181356b72cbff88/src/BuildIntegration/Microsoft.NETCore.Native.Unix.props#L45

https://github.com/dotnet/corert/blob/cb947515370bdc09f99abd9fcc176c8cca8d2960/buildscripts/build-managed.sh#L99

https://github.com/dotnet/corert/blob/fd9268c6727ad30f1e8573c52ead16a77bdc713f/src/packaging/packages.targets#L188

@parjong I am closing the issue as the question answered by @janvorli but feel free to reply back with more questions if you have any.

PR #12026, opened 2017-06-01
Add a option to exclude libSystem.Globalization.Native.a from artifacts

parjong:
As discussed in #11997, libSystem.Globalization.Native.a is necessary for Core RT, but not necessary for CLR. 

This commit adds a build option to exclude libSystem.Globalization.Native.a from CLR artifacts. This commit does not change default behavior.


parjong:

janvorli:
@parjong why is the fact that we build the library a problem for you?
CC: @gkhanna79

parjong:
I just want to slightly reduce the size of binaries (about 700Kb?). Actually, this commit does not prevent build, it just prevent installation.

jkotas:
The actual content of the installation is controlled by the content of the nuget packages. This file is not packaged into CoreCLR nuget package. The extra files in binaries directory should not matter.

Why does the size of files in binaries directory matter to you?

parjong:
We are currently integrating .NET Core into Tizen which uses its own build/package system. That system does not use NuGet packages, and accesses these binaries directly (that is why I proposed this PR at first).

However, @jyoungyun confirms that there is a manually managed list of files to be included in the package, and thus this change is not necessary.

PR #12036, opened 2017-06-01
[Local GC] Add GetThread to the GC/EE interface

swgillespie:
This fixes a linker error when building the GC as a shared library.

### DIFF
--- a/src/gc/env/gcenv.ee.h
+++ b/src/gc/env/gcenv.ee.h
@@ -50,6 +50,7 @@ public:
     static void EnablePreemptiveGC(Thread * pThread);
     static void DisablePreemptiveGC(Thread * pThread);
     static bool TrapReturningThreads();
+    static Thread* GetThread();
 
     static gc_alloc_context * GetAllocContext(Thread * pThread);
     static bool CatchAtSafePoint(Thread * pThread);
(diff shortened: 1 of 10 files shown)


jkotas:
It it pretty unfortunate how many GCToEEInterface indirect calls the mode flips compile into now:

Consider the following fragment - it will compile into 3 indirect calls with local GC. Instead, it can be just a single call (both GetThread and IsPreemptiveGCDisabled check can be done as part EnablePreemptiveGC):

```
current_thread = GCToEEInterface::GetThread(); 
cooperative_mode = gc_heap::enable_preemptive (current_thread);
```

I think we should make IsPreemptiveGCDisabled, EnablePreemptiveGC, DisablePreemptiveGC, GetAllocContext, CatchAtSafePoint to always operate on current thread, and get rid of the thread argument on them. They are always used on the current thread and it will be cheaper to fetch the current thread internally in these methods.

Otherwise looks good.

swgillespie:
@jkotas Thanks! Created issue https://github.com/dotnet/coreclr/issues/12043 to track that for a follow-up PR.

swgillespie:
Rebased against master to fix a merge conflict.

swgillespie:

swgillespie:
thanks!

Issue #12018, opened 2017-05-31
OpCodeGen.pl script fails on Windows due to syntax errors


4creators:
Environment:

Windows 10 Pro with several lang packs, active en-US
Command prompt VS 2017 v15.2
fresh clone of coreclr git repo
installed Strawberry perl v5.24.0 in user path 

Reported error:

```
e:\src\ms\coreclr\coreclr.git\src\inc>perl OpCodeGen.pl
Missing braces on \o{} at OpCodeGen.pl line 58, within string
Missing braces on \o{} at OpCodeGen.pl line 87, within string
Missing braces on \o{} at OpCodeGen.pl line 105, within string
Missing braces on \o{} at OpCodeGen.pl line 123, within string
Missing braces on \o{} at OpCodeGen.pl line 141, within string
Execution of OpCodeGen.pl aborted due to compilation errors.
```

danmosemsft:
Thanks @4creators . Would you like to offer a PR to fix?

4creators:
@danmosemsft Sure, will do it

4creators:
@danmosemsft  I have made a PR #12040 with some comments. Problem seems to be more complex and deeper than it seems from the surface. Pls look at the PR comments.

4creators:
Fixed with commit 116a6cc7dd0ff0912251f7d1febed36460ea11cd

PR #12040, opened 2017-06-01
Rewrite of OpCodeGen.pl script to generate correctly IL OpCodes

4creators:
related C# code: FlowControl.cs, OpCodes.cs, OpCodeTypes.cs, OperandType.cs, StackBehaviour.cs from opcode.def file. Commit comprises OpCodeGen.pl generator code as well as generated with this code files: FlowControl.cs, OpCodes.cs, OpCodeTypes.cs, OperandType.cs, StackBehaviour.cs.

This PR should partially fix issue #12018 

Unsolved problems for future code auto generation are related to "hand made" changes to some of the above C# files, and changes introduced to opcode.def file. 

1. Removal of obsolete opcodes from opcode.def breaks ability to generate several of the above enums due to lack of source data. Current solution is to hard code obsolete opcodes and ObsoleteAttribute into OpCodeGen.pl script. Links to problematic code:
  
[FlowControl.Phi](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Reflection/Emit/FlowControl.cs#LN27-28)

[OpcodeType.Annotation](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Reflection/Emit/OpcodeType.cs#LN22-23)

[OperandType.InlinePhi](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Reflection/Emit/OperandType.csLN28-29)

One of the possible solutions to this problem would be to create opcode_obsolete.def file with obsolete opcode data. 

2. In the case of enums StackBehaviour and  OperandType there are some hand edits. Namely there is discontinuity in numbering of enum members which goes directly from InlineR = 7 to InlineSig = 9,

[OperandType.InlineR - OperandType.InlineSig](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Reflection/Emit/OperandType.csLN#30-31)

Another discountinuity arises by throwing despite having all memebers sorted by name Popref_popi_pop1 = 28, at the end of enum. During sort it goes on position No 13.

[StackBehaviour.Popref_popi_pop1](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Reflection/Emit/StackBehaviour.csLN#35-50)

My proposal for a solution to this problem would be to restore numbering to sorting order in line with the [current docs](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.stackbehaviour?view=netframework-4.7). It may affect to a limited extent code behaviour and backward compatibility but since almost nobody replaces enumeration labels with hard coded numbers this should work.

The last problem which I see is the current usage pattern of generated code which is generated once outside of the build and there are no tests for verifying if generator is capable to work properly. The solution could be to force code generation during every build providing preconditions are met (i.e. new version of opcode.def file). Since this would mean taking a new dependency on perl for building dotnet I would suggest porting the code to C#.


danmosemsft:
@dotnet/jit-contrib who presumably own this

4creators:

pgavlin:
@danmosemsft I'm not sure who owns this, but I don't think that it's the JIT team: this is `Reflection.Emit`-related code in the BCL.

@stephentoub do you know who might be an appropriate reviewer?

danmosemsft:
@jkotas do you know anything about this file

jkotas:
> It may affect to a limited extent code behaviour and backward compatibility but since almost nobody replaces enumeration labels with hard coded numbers this should work.

The C# compiler replaces enumeration labels with numbers. Changing the numbers associated with the labels would be unacceptable breaking change.

jkotas:
> @jkotas do you know anything about this file

This PR is mostly about Reflection.Emit. cc @DnlHarvey @AtsushiKan

PR #12041, opened 2017-06-01
Add support for flat vtables in RyuJIT

MichalStrehovsky:
CoreRT needs one less indirection to get to the virtual method.

On CoreCLR a virtual method call resolution goes like:

```
this -> MethodTable -> chunk(?) -> slot
```

On CoreRT, it's more flat:

```
this -> EEType -> slot
```


MichalStrehovsky:
@dotnet/jit-contrib PTAL

PR #12047, opened 2017-06-02
Fixed determintaion of need to make code that contains tail.call loop fully interuptible

rartemev:
It possible to have partial interruptible code without any reachable safepoint if two methods invoke each other using tail.call creating a loop.

```
void A() {
  if (C.stop == false) {
    B();   // Tail.Call
  } else {
    Foo(); // safepoint
    return;
  }
}

void B() {
  if (C.stop == false) {
    A();   // Tail.Call
  } else {
    Bar(); // safepoint
    return;
  }
}
```

In this case we have a loop between A() and B() so both methods have to be fully interruptible.

The PR fixes the algorithm that detects such cases, but only for caller without any recursive analysis .

Tests like JIT\Methodical\tailcall_v4\hijacking might hang on due to this bug.

Originally the problem is related to #4122

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -18488,8 +18488,7 @@ void Compiler::fgSetBlockOrder()
 
 #if FEATURE_FASTTAILCALL
 #ifndef JIT32_GCENCODER
-        if (block->endsWithTailCallOrJmp(this, true) && !(block->bbFlags & BBF_GC_SAFE_POINT) &&
-            optReachWithoutCall(fgFirstBB, block))
+        if (block->endsWithTailCallOrJmp(this, true) && optReachWithoutCall(fgFirstBB, block))
         {
             // We have a tail call that is reachable without making any other
             // 'normal' call that would have counted as a GC Poll.  If we were
--- a/src/jit/optimizer.cpp
+++ b/src/jit/optimizer.cpp
@@ -3289,7 +3289,7 @@ void Compiler::optUnrollLoops()
 
 /*****************************************************************************
  *
- *  Return non-zero if there is a code path from 'topBB' to 'botBB' that will
+ *  Return false if there is a code path from 'topBB' to 'botBB' that might
  *  not execute a method call.
  */
 
--- a/src/jit/compiler.h
+++ b/src/jit/compiler.h
@@ -6759,6 +6759,7 @@ public:
     static fgWalkPreFn CountSharedStaticHelper;
     static bool IsSharedStaticHelper(GenTreePtr tree);
     static bool IsTreeAlwaysHoistable(GenTreePtr tree);
+    static bool IsGcSafePoint(GenTreePtr tree);
 
     static CORINFO_FIELD_HANDLE eeFindJitDataOffs(unsigned jitDataOffs);
     // returns true/false if 'field' is a Jit Data offset
--- a/src/jit/compiler.hpp
+++ b/src/jit/compiler.hpp
@@ -4022,6 +4022,31 @@ inline bool Compiler::IsTreeAlwaysHoistable(GenTreePtr tree)
     }
 }
 
+inline bool Compiler::IsGcSafePoint(GenTreePtr tree)
+{
+    if (tree->IsCall())
+    {
+        GenTreeCall* call = tree->AsCall();
+        if (!call->IsFastTailCall())
+        {
+            if (call->gtCallType == CT_INDIRECT)
+            {
+                return true;
+            }
+            else if (call->gtCallType == CT_USER_FUNC)
+            {
+                if ((call->gtCallMoreFlags & GTF_CALL_M_NOGCCHECK) == 0)
+                {
+                    return true;
+                }
+            }
+            // otherwise we have a CT_HELPER
+        }
+    }
+
+    return false;
+}
+
 //
 // Note that we want to have two special FIELD_HANDLES that will both
 // be considered non-Data Offset handles
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -8293,25 +8293,9 @@ NO_TAIL_CALL:
     // having a call (for computing full interruptibility).
     CLANG_FORMAT_COMMENT_ANCHOR;
 
-#ifdef _TARGET_AMD64_
-    // Amd64 note: If this is a fast tail call then don't count it as a call
-    // since we don't insert GC-polls but instead make the method fully GC
-    // interruptible.
-    if (!call->IsFastTailCall())
-#endif
+    if (IsGcSafePoint(call))
     {
-        if (call->gtCallType == CT_INDIRECT)
-        {
-            compCurBB->bbFlags |= BBF_GC_SAFE_POINT;
-        }
-        else if (call->gtCallType == CT_USER_FUNC)
-        {
-            if ((call->gtCallMoreFlags & GTF_CALL_M_NOGCCHECK) == 0)
-            {
-                compCurBB->bbFlags |= BBF_GC_SAFE_POINT;
-            }
-        }
-        // otherwise we have a CT_HELPER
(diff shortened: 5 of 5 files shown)


rartemev:
@dotnet-bot test Windows_NT jitstressregs0x80
@dotnet-bot test Windows_NT x86 corefx_jitstressregs3
@dotnet-bot test Ubuntu x64 corefx_jitstressregs0x80

rartemev:
@dotnet/jit-contrib please review.

src/jit/optimizer.cpp
```diff
@@ -3304,7 +3304,7 @@ bool Compiler::optReachWithoutCall(BasicBlock* topBB, BasicBlock* botBB)
 
     // We can always check topBB and botBB for any gc safe points and early out
 
-    if ((topBB->bbFlags | botBB->bbFlags) & BBF_GC_SAFE_POINT)
+    if ((topBB->bbFlags & BBF_GC_SAFE_POINT) || (!endsWithTailCall && (botBB->bbFlags & BBF_GC_SAFE_POINT)))
```
CarolEidt: It is not clear to me why `BBF_GC_SAFE_POINT` would even be set on a block that ends with a tail call.
Also, since you are here, you should change these to check the value, e.g. `((topBB->bbFlags & BBF_GC_SAFE_POINT) != 0)` etc.

src/jit/optimizer.cpp
```diff
@@ -3293,7 +3293,7 @@ void Compiler::optUnrollLoops()
  *  not execute a method call.
  */
 
-bool Compiler::optReachWithoutCall(BasicBlock* topBB, BasicBlock* botBB)
+bool Compiler::optReachWithoutCall(BasicBlock* topBB, BasicBlock* botBB, bool endsWithTailCall /* = false*/)
```
CarolEidt: The function header comment on this method not very informative. It appears that `true` is the conservative answer, based on the code below, but it would be good to modify the comment to make that clear, e.g. "Return false only if it can be proven that there is a code path from 'topBB' to 'botBB' that will not execute a method call or hit a BBF_GC_SAFE_POINT."

src/jit/optimizer.cpp
briansull: If the block has additional normal calls that preceed the final tail call then it is appropriate that the block be flagged with  `BBF_GC_SAFE_POINT `

src/jit/optimizer.cpp
```diff
@@ -3293,7 +3293,7 @@ void Compiler::optUnrollLoops()
  *  not execute a method call.
```
briansull: Change the comment from  "Return non-zero" to "Return false"
And reword,  "will not" to "could possibly not"

src/jit/optimizer.cpp
briansull: If the block has additional normal calls that preceed the final tail call then it is appropriate that the block be flagged with `BBF_GC_SAFE_POINT`

briansull: It also should be an error if the sole call in the block is a block ending tail call and we have the flag `BBF_GC_SAFE_POINT` set.

But perhaps we don't yet enforce that invariant, as we may start off considering the call to be a normal call and set the flag, then later on we may opportunistically convert it into a tail call.  We would need to re-walk the statements to see if there is still a different call site to keep the flag set correctly.

src/jit/optimizer.cpp
```diff
@@ -3332,7 +3332,7 @@ bool Compiler::optReachWithoutCall(BasicBlock* topBB, BasicBlock* botBB)
 
             // Does this block contain a gc safe point?
 
-            if (curBB->bbFlags & BBF_GC_SAFE_POINT)
+            if ((curBB->bbFlags & BBF_GC_SAFE_POINT) && !endsWithTailCall && (curBB != botBB))
```
briansull: You change seem to be we can't trust the `BBF_GC_SAFE_POINT` flag when the block end with a tail call.
I don't like going down that route.

I think that we should explore the option of enforcing the invariant that having  `BBF_GC_SAFE_POINT` means that we have one or more normal non-tail callsites in the block.

src/jit/flowgraph.cpp
```diff
@@ -1877,6 +1877,48 @@ void Compiler::fgComputeReachabilitySets()
...
+                    }
+                }
+
+                if ((block->bbFlags & BBF_GC_SAFE_POINT) != 0)
+                {
+                    break;
+                }
+            }
+        }
+        fgComputeReachabilitySets();
```
briansull: Add a comment explaining what  ` fgComputeReachability()` does to the BBF_GC_SAFE_POINT flags

src/jit/compiler.cpp
```diff
@@ -4594,6 +4594,8 @@ void Compiler::compCompile(void** methodCodePtr, ULONG* methodCodeSize, JitFlags
     // this point is responsible for calling fgSetStmtSeq() to keep the
     // nodes properly linked.
     // This can create GC poll calls, and create new BasicBlocks (without updating dominators/reachability).
+
+    fgUpdateSafepointFlags();
```
briansull: You should add a comment explaining why you placed this call here:

// This needs to be done before we call fgSetBlockOrder because that method can call  fgCreateGCPolls()
//  which uses the BBF_GC_SAFE_POINT flag to determine if we need to insert GC Poll helper calls.

rartemev:
@pgavlin @briansull please take a look

src/jit/flowgraph.cpp
```diff
@@ -1877,6 +1877,48 @@ void Compiler::fgComputeReachabilitySets()
...
+                            if (call->IsTailCallViaHelper())
+                            {
+                                block->bbFlags |= BBF_GC_SAFE_POINT;
+                            }
+                        }
+                        else
+                        {
+                            block->bbFlags |= BBF_GC_SAFE_POINT;
+                        }
+                    }
```
briansull: You should instead pull out lines 1896- 1907 into a new method:
bool fgIsGcSafePoint (or something similar)
I would make it return true if the call is a GC Safe point and then set the BBF_GC_SAFE_POINT here and in morph.cpp

Then you should call the new method from here and from morph.cpp line 8296

Note that your code is setting  BBF_GC_SAFE_POINT for all helper calls and that isn't correct.
In particular calls to the GC Write Barrier helpers are not GC safe points.
Conversely with our existing code it is also not exactly correct to say that all helper calls are not GC safe points, 
but is it conservatively safe to do so.

We should eventually add a new method to HelperCallProperties that returns true if the helper call is a GC safe point.   However you probably shouldn't address that with this check in.  Opening a new GitHib issue to do so in a follow on checkin would probably be best.

here is the code from Morph that your new method should closely match:

```
#ifdef _TARGET_AMD64_
    // Amd64 note: If this is a fast tail call then don't count it as a call
    // since we don't insert GC-polls but instead make the method fully GC
    // interruptible.
    if (!call->IsFastTailCall())
#endif
    {
        if (call->gtCallType == CT_INDIRECT)
        {
            compCurBB->bbFlags |= BBF_GC_SAFE_POINT;
        }
        else if (call->gtCallType == CT_USER_FUNC)
        {
            if ((call->gtCallMoreFlags & GTF_CALL_M_NOGCCHECK) == 0)
            {
                compCurBB->bbFlags |= BBF_GC_SAFE_POINT;
            }
        }
        // otherwise we have a CT_HELPER
    }
...

rartemev:
@briansull 
Actually I noticed that condition `if (!call->IsFastTailCall())` has been ignored for arm64 and FastTailCall's also was considered as safepoints that is not correct. Removing `#ifdef AMD64` macro fixes the problem for arm64. So I just extracted the gc check to the separated method and got rid of PP check.

Should I remove it or add addition condition for ARM's?

briansull:
That ifdef does look sketchy, I would probably remove it, but I will have to look into it more carefully.

rartemev:
ping

PR #12050, opened 2017-06-02
[RyuJIT/ARM32] Enable passing large split struct argument

hseok-oh:
This enables passing split struct larger than 16 bytes.
To support splitted struct, it defines new GenTree type - GenTreePutArgSplit.
GenTreePutArgSplit is similar with GenTreePutArgStk,
but it is used for splitted struct only
and it has additional field to save register information.
GenTreePutArgSplit node is generated in lower phase.

related issue: #10722 
cc/ @dotnet/arm32-contrib


BruceForstall:
cc @dotnet/jit-contrib

src/jit/gentree.h
```diff
@@ -4845,6 +4856,86 @@ struct GenTreePutArgStk : public GenTreeUnOp
...
+                           slotNum PUT_STRUCT_ARG_STK_ONLY_ARG(numSlots),
+                           putIncomingArgArea,
+                           callNode)
+        , gtNumRegs(numRegs)
+        , gtIsHfa(isHfa)
+    {
+    }
+
+    // not consider HFA yet
+    regNumber gtOtherRegs[MAX_REG_ARG - 1];
```
CarolEidt: This comment is not clear. Should it be "HFA args are not yet handled"?

src/jit/gentree.h
```diff
@@ -4845,6 +4856,86 @@ struct GenTreePutArgStk : public GenTreeUnOp
...
+    //---------------------------------------------------------------------------
+    // GetRegNumByIdx: get ith register allocated to this struct argument.
+    //
+    // Arguments:
+    //     idx   -   index of the struct
+    //
+    // Return Value:
+    //     Return regNumber of ith register of this struct argument
+    //
+    regNumber GetRegNumByIdx(unsigned idx) const
```
CarolEidt: I would spell out "Index" in both the method name and the parameter name.

src/jit/lower.cpp
```diff
@@ -808,6 +813,37 @@ GenTreePtr Lowering::NewPutArg(GenTreeCall* call, GenTreePtr arg, fgArgTabEntryP
     isOnStack = info->regNum == REG_STK;
 #endif // !FEATURE_UNIX_AMD64_STRUCT_PASSING
 
+#ifdef _TARGET_ARM_
+    // Strcut can be splitted into register(s) and stack in ARM
```
CarolEidt: Typos: I suggest "Struct can be split between register(s) and stack on ARM".

src/jit/lsra.cpp
```diff
@@ -4070,6 +4075,15 @@ void LinearScan::buildRefPositionsForNode(GenTree*                  tree,
             useCandidates  = allRegs(registerType);
         }
 
+#ifdef _TARGET_ARM_
+        if (tree->OperIsPutArgStruct())
+        {
+            // get i-th candidate
+            currCandidates = genFindLowestReg(candidates);
```
CarolEidt: This relies on the fact that the arg regs are used in order. It would be good to add an assert that this is the same register as returned by `GetRegNumByIdx` (or `GetRegNumByIndex` if you change it).

src/jit/morph.cpp
```diff
@@ -1571,6 +1586,8 @@ void fgArgInfo::ArgsComplete()
                 // Spill multireg struct arguments that have Assignments or Calls embedded in them
                 curArgTabEntry->needTmp = true;
             }
+#ifndef _TARGET_ARM_
+            // Not tested yet
```
CarolEidt: This should be some form of `TODO` comment, and there should presumably be an associated `NYI`?

src/jit/gentree.h
hseok-oh: But GenTreeCall, GenTreeCopyOrReload also uses same name. Isn't it better to use same name?

src/jit/gentree.h
```diff
@@ -1289,6 +1289,13 @@ struct GenTree
         return OperIsStoreBlk(OperGet());
     }
 
+#ifdef _TARGET_ARM_
+    bool OperIsPutArgStruct() const
```
hseok-oh: Thanks. I'll fix it.

src/jit/morph.cpp
hseok-oh: If we uses NYI, legacyJIT/arm32 will be not works. I think It's better to remain it as `TODO` comment.

hseok-oh:
@CarolEidt I've update your feedback. PTAL
- Reduce the clutter of #ifdef 
- Rename `PutArgStruct` as `PutArgSplit`
- Fix comments

src/jit/morph.cpp
hseok-oh: ~~If we uses NYI, legacyJIT/arm32 will be not works.~~
Sorry, It's my misunderstand. But It doesn't seem important point for correctness. So anyway, I think It's better to remain it as TODO comment.

src/jit/lsra.cpp
hseok-oh: We cannot compare registers return by `GetRegNumByIdx` here because it is not initialized here (LSRA buildInterval phase). It is initialized at LSRA resolve phase.

src/jit/gentree.h
CarolEidt: IMO there are a number of existing names that aren't ideal, but I'm OK with being consistent or switching.

src/jit/lsra.cpp
CarolEidt: Ah, I was confused. Instead, how about asserting that the register is `genRegArgNext` of the previous register, if any?

src/jit/codegenarmarch.cpp
```diff
@@ -880,6 +886,216 @@ void CodeGen::genPutArgReg(GenTreeOp* tree)
     genProduceReg(tree);
 }
 
+#ifdef _TARGET_ARM_
+//---------------------------------------------------------------------
+// genPutArgStruct - generate code for a GT_PUTARG_STRUCT node
```
CarolEidt: This needs to be change to reflect the new name.

src/jit/codegenarmarch.cpp
```diff
@@ -880,6 +886,216 @@ void CodeGen::genPutArgReg(GenTreeOp* tree)
...
+
+    // Either varNode or addrNOde must have been setup above,
+    // the xor ensures that only one of the two is setup, not both
+    assert((varNode != nullptr) ^ (addrNode != nullptr));
+
+    BYTE* gcPtrs = nullptr;
+
+    unsigned gcPtrCount; // The count of GC pointers in the struct
+    int      structSize;
+    bool     isHfa;
```
CarolEidt: You should be able to get these values from the node. In inherits `gtNumberReferenceSlots` and `getArgSize()`, and it would make sense to add an `isHfa` flag (probably to `GenTreePutArgStk` to avoid extra padding).

src/jit/codegenarmarch.cpp
```diff
@@ -880,6 +886,216 @@ void CodeGen::genPutArgReg(GenTreeOp* tree)
...
+
+    // This is the varNum for our load operations,
+    // only used when we have a multireg struct with a LclVar source
+    unsigned varNumInp = BAD_VAR_NUM;
```
CarolEidt: I'm not sure what the p is intended to imply in `varNumInp` is not very meaningful. Perhaps `srcVarNum` or `varNumIn` would be better?

src/jit/codegenarmarch.cpp
```diff
@@ -880,6 +886,216 @@ void CodeGen::genPutArgReg(GenTreeOp* tree)
...
+        argOffsetOut += TARGET_POINTER_SIZE;  // We stored 4-bytes of the struct
+        assert(argOffsetOut <= argOffsetMax); // We can't write beyound the outgoing area area
+        remainingSize -= TARGET_POINTER_SIZE; // We loaded 4-bytes of the struct
+        structOffset += TARGET_POINTER_SIZE;
+        nextIndex += 1;
+    }
+
+    while (remainingSize > 0)
+    {
+        {
```
CarolEidt: Why the extra block here?

src/jit/codegenarmarch.cpp
```diff
@@ -880,6 +886,216 @@ void CodeGen::genPutArgReg(GenTreeOp* tree)
...
+                loadType = TYP_UBYTE;
+            }
+            else if (loadSize == 2)
+            {
+                loadType = TYP_USHORT;
+            }
+            else
+            {
+                // Need to handle additional loadSize cases here
+                noway_assert(loadSize == 4);
```
CarolEidt: Should these be an `NYI` instead?

src/jit/codegenarmarch.cpp
```diff
@@ -1557,7 +1773,21 @@ void CodeGen::genCallInstruction(GenTreeCall* call)
...
             {
-                inst_RV_RV(ins_Move_Extend(argNode->TypeGet(), argNode->InReg()), argReg, argNode->gtRegNum);
+#ifdef _TARGET_ARM_
+                if (curArgTabEntry->isSplit)
+                {
+                    for (unsigned idx = 0; idx < curArgTabEntry->numRegs; idx++)
+                    {
+                        regNumber regDst = (regNumber)((unsigned)argReg + idx);
+                        regNumber regSrc = (regNumber)((unsigned)argNode->gtRegNum + idx);
+                        inst_RV_RV(ins_Move_Extend(argNode->TypeGet(), argNode->InReg()), regDst, regSrc);
```
CarolEidt: `InReg()` is no longer used for the RyuJIT backend, in which the code generator should always know whether a node is in a register or not. I think you need `true` here.

src/jit/codegenarmarch.cpp
hseok-oh: I think this assertion is needless for ARM32. This assertion comes form the implementation of `CodeGen::genPutArgStk. I'll remove this and fix the assertion in `CodeGen::genPutArgStk` later.

src/jit/codegenarmarch.cpp
hseok-oh: I'll fix it. Thanks.

src/jit/codegenarmarch.cpp
hseok-oh: You're right. I'll update it.

hseok-oh:
@CarolEidt I've applied your feedback. PTAL
- Rename `varNumInp` to `srcVarNum`
- Fix comment: `genPutArgStruct` to `genPutArgSplit`
- Update to not use `InReg()`
- Get values from the node for code generation: `gcPtr`, `gcPtrCount`, `structSize`, `isHfa`
- Remove useless code

src/jit/codegenarmarch.cpp
hseok-oh: I've update to get these value from the node.

src/jit/gentree.h
hseok-oh: I'll make new PR to fix this.

hseok-oh:

hseok-oh:

src/jit/gtlist.h
```diff
@@ -295,6 +295,9 @@ GTNODE(PINVOKE_PROLOG   , GenTree            ,0,GTK_LEAF|GTK_NOVALUE)   // pinvo
 GTNODE(PINVOKE_EPILOG   , GenTree            ,0,GTK_LEAF|GTK_NOVALUE)   // pinvoke epilog seq
 GTNODE(PUTARG_REG       , GenTreeOp          ,0,GTK_UNOP)               // operator that places outgoing arg in register
 GTNODE(PUTARG_STK       , GenTreePutArgStk   ,0,GTK_UNOP|GTK_NOVALUE)   // operator that places outgoing arg in stack
+#ifdef _TARGET_ARM_
+GTNODE(PUTARG_SPLIT     , GenTreePutArgSplit ,0,GTK_UNOP)               // operator that places outgoing arg in registers with stack (splitted struct in ARM32)
```
BruceForstall: To be precise, shouldn't this be:
```
#if defined(_TARGET_ARM_) && !defined(LEGACY_BACKEND)
```
?

BruceForstall: >splitted [](start = 134, length = 8)

Grammatical note: do not use the word "splitted" (not a correct English word). Instead, simply use "split".

src/jit/gtstructs.h
```diff
@@ -96,6 +96,9 @@ GTSTRUCT_1(PhiArg      , GT_PHI_ARG)
 GTSTRUCT_1(StoreInd    , GT_STOREIND)
 GTSTRUCT_N(Indir       , GT_STOREIND, GT_IND, GT_NULLCHECK, GT_BLK, GT_STORE_BLK, GT_OBJ, GT_STORE_OBJ, GT_DYN_BLK, GT_STORE_DYN_BLK)
 GTSTRUCT_1(PutArgStk   , GT_PUTARG_STK)
+#ifdef _TARGET_ARM_
+GTSTRUCT_1(PutArgSplit , GT_PUTARG_SPLIT)
```
BruceForstall: ==>
```
#if !defined(LEGACY_BACKEND) && defined(_TARGET_ARM_)
```
?

src/jit/gentree.h
```diff
@@ -4873,6 +4882,9 @@ struct GenTreePutArgStk : public GenTreeUnOp
     unsigned gtNumSlots;             // Number of slots for the argument to be passed on stack
     unsigned gtNumberReferenceSlots; // Number of reference slots.
     BYTE*    gtGcPtrs;               // gcPointers
+#ifdef _TARGET_ARM_
+    bool gtIsHfa; // isHfa
+#endif
```
BruceForstall: The comment here is not useful (being a restatement of the member name). Is there something more descriptive that would be useful?

BruceForstall: Is this actually required? Couldn't you use `Compiler::IsHfa(GenTreePtr tree)` to determine if this node is an HFA?

src/jit/gentree.h
```diff
@@ -4887,6 +4899,94 @@ struct GenTreePutArgStk : public GenTreeUnOp
 #endif
 };
 
+#ifdef _TARGET_ARM_
+// Represent the struct argument: value is splitted in register(s) and stack
```
BruceForstall: =>
```
#if defined(_TARGET_ARM_) && !defined(LEGACY_BACKEND)
```

src/jit/gentree.h
```diff
@@ -4887,6 +4899,94 @@ struct GenTreePutArgStk : public GenTreeUnOp
...
+        , gtNumRegs(numRegs)
+    {
+#ifdef FEATURE_PUT_STRUCT_ARG_STK
+        gtIsHfa = isHfa;
+#endif
+    }
+
+    // HFA args is not yet handled
+    regNumber gtOtherRegs[MAX_REG_ARG - 1];
+
```
BruceForstall: Can you add a small comment here about what this is, and its relationship to gtRegNum?

src/jit/gentree.h
```diff
@@ -4887,6 +4899,94 @@ struct GenTreePutArgStk : public GenTreeUnOp
...
+                           callNode)
+        , gtNumRegs(numRegs)
+    {
+#ifdef FEATURE_PUT_STRUCT_ARG_STK
+        gtIsHfa = isHfa;
+#endif
+    }
+
+    // HFA args is not yet handled
+    regNumber gtOtherRegs[MAX_REG_ARG - 1];
```
BruceForstall: >regNumber [](start = 4, length = 9)

This should probably be `regNumberSmall`

src/jit/codegenarmarch.cpp
```diff
@@ -1621,7 +1769,21 @@ void CodeGen::genCallInstruction(GenTreeCall* call)
             genConsumeReg(argNode);
             if (argNode->gtRegNum != argReg)
```
CarolEidt: I think that this line should be moved to the `else` clause below, and in the `isSplit` case, each register needs to be consumed.

src/jit/gentree.h
CarolEidt: The call to `IsHfa` involves a JIT/EE interface call, so I think it's reasonable to maintain a flag on the node.

src/jit/lower.cpp
```diff
@@ -809,6 +808,40 @@ GenTreePtr Lowering::NewPutArg(GenTreeCall* call, GenTreePtr arg, fgArgTabEntryP
     isOnStack = info->regNum == REG_STK;
 #endif // !FEATURE_UNIX_AMD64_STRUCT_PASSING
 
+#ifdef _TARGET_ARM_
+    // Strcut can be split into register(s) and stack on ARM
```
CarolEidt: Typo: Struct

src/jit/morph.cpp
```diff
@@ -1557,6 +1565,14 @@ void fgArgInfo::ArgsComplete()
                 {
                     prevArgTabEntry->needPlace = true;
                 }
+#if defined(_TARGET_ARM_) && !defined(LEGACY_BACKEND)
+                else if (prevArgTabEntry->isSplit)
+                {
+                    // TODO: We need to test it on LegacyJIT/ARM32.
```
CarolEidt: These should be something like `TODO-Arm`

src/jit/gentree.h
BruceForstall: You can also mention that this is similar to GetRegNumByIdx/SetRegNumByIdx in GenTreeCall

---
In reply to: [122343854](https://github.com/dotnet/coreclr/pull/12050#discussion_r122343854) [](ancestors = 122343854)

src/jit/codegenarmarch.cpp
```diff
@@ -1621,7 +1769,21 @@ void CodeGen::genCallInstruction(GenTreeCall* call)
...
             {
-                inst_RV_RV(ins_Move_Extend(argNode->TypeGet(), true), argReg, argNode->gtRegNum);
+#ifdef _TARGET_ARM_
+                if (curArgTabEntry->isSplit)
+                {
+                    for (unsigned idx = 0; idx < curArgTabEntry->numRegs; idx++)
+                    {
```
BruceForstall: Could you add an assert:
```
assert(curArgTabEntry->numRegs >= 1);
```
?

src/jit/lower.cpp
BruceForstall: >Strcut [](start = 7, length = 6)

Typo: "Strcut" => "Struct"

src/jit/lower.cpp
```diff
@@ -809,6 +808,40 @@ GenTreePtr Lowering::NewPutArg(GenTreeCall* call, GenTreePtr arg, fgArgTabEntryP
...
+    if (info->isSplit)
+    {
+        if (arg->OperGet() != GT_OBJ)
+        {
+            NYI_ARM("Lowering: Oper for struct argument is not GT_OBJ");
+        }
```
BruceForstall: This is an NYI and not an assert?

src/jit/lower.cpp
```diff
@@ -809,6 +808,40 @@ GenTreePtr Lowering::NewPutArg(GenTreeCall* call, GenTreePtr arg, fgArgTabEntryP
...
+        putArg->AsPutArgSplit()->setGcPointers(numRefs, gcLayout);
+
+        JITDUMP("new node is : ");
+        DISPNODE(putArg);
+        JITDUMP("\n");
+
+        return putArg;
+    }
+#endif
+
```
BruceForstall: Couldn't the end of this be:
```
    else
#endif // _TARGET_ARM_
```

in which case, you wouldn't need to include this in your `if`:
```
        if (arg->gtFlags & GTF_LATE_ARG)
        {
            putArg->gtFlags |= GTF_LATE_ARG;
        }
        else
        {
            info->node = putArg;
        }

...

        JITDUMP("new node is : ");
        DISPNODE(putArg);
        JITDUMP("\n");

        return putArg;
    }
#endif
```
?

(it looks like `updateArgTable` is always `true` so you could delete it.)

src/jit/morph.cpp
```diff
@@ -1439,6 +1439,14 @@ void fgArgInfo::ArgsComplete()
             continue;
 #endif
         }
+#if defined(_TARGET_ARM_) && !defined(LEGACY_BACKEND)
+        // Not tested on LegacyJIT/ARM32 yet
+        else if (curArgTabEntry->isSplit)
```
BruceForstall: This will never be used on arm32 `LEGACY_BACKEND` -- we shouldn't be touching that code.

src/jit/lsraarmarch.cpp
```diff
@@ -718,6 +732,57 @@ void Lowering::TreeNodeInfoInitPutArgStk(GenTreePutArgStk* argNode, fgArgTabEntr
     }
 }
 
+#ifdef _TARGET_ARM_
+//------------------------------------------------------------------------
+// TreeNodeInfoInitPutArgStruct: Set the NodeInfo for a GT_PUTARG_STRUCT node
+//
+// Arguments:
+//    argNode - a GT_PUTARG_STRUCT node
+//
```
BruceForstall: This comment needs to be updated (it uses "struct" instead of "split" several places.)

src/jit/morph.cpp
```diff
@@ -1557,6 +1565,14 @@ void fgArgInfo::ArgsComplete()
                 {
                     prevArgTabEntry->needPlace = true;
                 }
+#if defined(_TARGET_ARM_) && !defined(LEGACY_BACKEND)
+                else if (prevArgTabEntry->isSplit)
+                {
+                    // TODO: We need to test it on LegacyJIT/ARM32.
+                    // It have worked well on LegacyJIT/ARM32 in spite of we did not set this.
+                    prevArgTabEntry->needPlace = true;
```
BruceForstall: Can this comment be removed? We shouldn't be touching `LEGACY_BACKEND`.

src/jit/morph.cpp
```diff
@@ -1579,6 +1595,8 @@ void fgArgInfo::ArgsComplete()
                 // Spill multireg struct arguments that have Assignments or Calls embedded in them
                 curArgTabEntry->needTmp = true;
             }
+#ifndef _TARGET_ARM_
+            // TODO: This implementation is not tested on ARM32
             else
```
BruceForstall: Instead of `#ifdef`ing this out, can you make it a `NYI_ARM("multireg struct arguments");` ?

src/jit/morph.cpp
```diff
@@ -3415,11 +3434,7 @@ GenTreeCall* Compiler::fgMorphArgs(GenTreeCall* call)
                     size = (unsigned)(roundUp(info.compCompHnd->getClassSize(argx->gtArgPlace.gtArgPlaceClsHnd),
                                               TARGET_POINTER_SIZE)) /
                            TARGET_POINTER_SIZE;
-                    if (isHfaArg)
-                    {
-                        hasMultiregStructArgs = true;
-                    }
-                    else if (size > 1 && size <= 4)
+                    if (isHfaArg || size > 1)
                     {
```
BruceForstall: Why does this need to change? Is this a bug in existing `LEGACY_BACKEND`? Or a new requirement for RyuJIT? (To not check `size <= 4`)

src/jit/codegenarmarch.cpp
```diff
@@ -885,6 +891,148 @@ void CodeGen::genPutArgReg(GenTreeOp* tree)
     genProduceReg(tree);
 }
 
+#ifdef _TARGET_ARM_
+//---------------------------------------------------------------------
+// genPutArgSplit - generate code for a GT_PUTARG_STRUCT node
+//
+// Arguments
+//    tree - the GT_PUTARG_STRUCT node
+//
```
BruceForstall: here, and above, `GT_PUTARG_STRUCT` => `GT_PUTARG_SPLIT`

src/jit/gtlist.h
hseok-oh: I'll fix it.

hseok-oh: I'll fix it.

src/jit/gtstructs.h
hseok-oh: I'll fix it.

src/jit/gentree.h
hseok-oh: I'll remove this comment.

hseok-oh: I agree with @CarolEidt

src/jit/gentree.h
hseok-oh: I'll add comment.

src/jit/gentree.h
hseok-oh: Thanks. I'll fix it.

src/jit/codegenarmarch.cpp
```diff
@@ -885,6 +891,148 @@ void CodeGen::genPutArgReg(GenTreeOp* tree)
...
+    unsigned varNumOut    = compiler->lvaOutgoingArgSpaceVar;
+    unsigned argOffsetMax = compiler->lvaOutgoingArgSpaceSize;
+    unsigned argOffsetOut = treeNode->gtSlotNum * TARGET_POINTER_SIZE;
+
+    // Put on stack first
+    unsigned nextIndex     = treeNode->gtNumRegs;
+    unsigned structOffset  = nextIndex * TARGET_POINTER_SIZE;
+    int      remainingSize = structSize - structOffset;
+
+    // remainingSize is always multiple of TARGET_POINTER_SIZE
```
hseok-oh: I'll add a assertion.

src/jit/codegenarmarch.cpp
```diff
@@ -885,6 +891,148 @@ void CodeGen::genPutArgReg(GenTreeOp* tree)
...
+        remainingSize -= TARGET_POINTER_SIZE; // We loaded 4-bytes of the struct
+        structOffset += TARGET_POINTER_SIZE;
+        nextIndex += 1;
+    }
+
+    // Set registers
+    structOffset = 0;
+    for (unsigned idx = 0; idx < treeNode->gtNumRegs; idx++)
+    {
+        baseReg        = treeNode->GetRegNumByIdx(idx);
```
hseok-oh: Good idea. I'll change this variable name.

src/jit/codegenarmarch.cpp
hseok-oh: Thanks. I'll move this clause and implement `genConsumeReg`.

src/jit/lower.cpp
hseok-oh: I'll fix it.

src/jit/morph.cpp
hseok-oh: I'll remove this comment because I did not found any problem on LegacyJIT/ARM32.

src/jit/codegenarmarch.cpp
hseok-oh: I'll add this assertion.

src/jit/lower.cpp
hseok-oh: This NYI is for FIELD_LIST struct. I'll be enabling that later.

src/jit/lower.cpp
hseok-oh: I'll fix it.

src/jit/morph.cpp
hseok-oh: I'll remove this comment

src/jit/lsraarmarch.cpp
hseok-oh: I'll fix it.

src/jit/morph.cpp
hseok-oh: I'll remove this comment.

src/jit/morph.cpp
hseok-oh: As I commented (https://github.com/dotnet/coreclr/pull/12050#discussion_r120794799), I think it's better to remain it as TODO comment.

src/jit/morph.cpp
hseok-oh: Finally, there is no change because `fgMorphMultiregStructArg` skip morphing when size > 4.

src/jit/codegenarmarch.cpp
hseok-oh: I'll fix it

hseok-oh:
@CarolEidt @BruceForstall  I've applied your feedback. PTAL

- Fix some comments: 
`genPutArgSplit`, `GenTreePutArgStk`, `GenTreePutArgSplit`, `NuwPutArg`, `ArgComplete`
- Add assertion check in `genPutArgSplit`, `genCallInstruction`
- Rename variable: `baseReg`
- Change flag for `GenTreePutArgSplit`: `_TARGET_ARM` && `!LEGACY_BACKEND`
- Change type of `gtOtherRegs` in `GenTreePutArgSplit`
- Remove duplicated code: `NewPutArg`
- Implement spill flag & restore for `GenTreePutArgSplit`'s `genConsumeReg`

hseok-oh:
Example 
```
using System;

public struct IntStruct
{
    public int num1;
    public int num2;
    public int num3;
    public int num4;
    public int num5;
    public int num6;
}

public class BringUpTest
{
    const int Pass = 100;
    const int Fail = -1;

    public static int StructAdd(int a, int b, IntStruct rp, int c, int d)
    {   
        return rp.num1 + rp.num2 + rp.num3 + c + d + a + b + rp.num4 + rp.num5 + rp.num6;
    }

    public static int Main()
    {   
        IntStruct a = new IntStruct();

        a.num1 = 11;
        a.num2 = 12;
        a.num3 = 13;
        a.num4 = 14;
        a.num5 = 15;
        a.num6 = 16;

        int y = StructAdd(1, 2, a, 3, 4);
        if (y != 91) return Fail;
        return Pass;
    }
}
```

Generated Code (init struct & put argument)
```
Added IP mapping: 0x0008 STACK_EMPTY (G_M4347_IG02,ins#6,ofs#16)
Generating: N011 (  5,  6) [000011] -------------                IL_OFFSET void   IL offset: 0x8 REG NA
Generating: N013 (  1,  1) [000008] -------------        t8 =    CNS_INT   int    11 REG r3 $82
IN0007:             movs    r3, 11
                                                              /--*  t8     int
Generating: N015 (  5,  6) [000010] UA-----------             *  STORE_LCL_FLD int    V00 loc0         ud:3->0[+0] Fseq[num1] NA REG NA
IN0008:             str     r3, [sp+0x18]       // [V00 loc0]
Added IP mapping: 0x0011 STACK_EMPTY (G_M4347_IG02,ins#8,ofs#20)
Generating: N017 (  5,  6) [000017] -------------                IL_OFFSET void   IL offset: 0x11 REG NA
Generating: N019 (  1,  1) [000014] -------------       t14 =    CNS_INT   int    12 REG r3 $83
IN0009:             movs    r3, 12
                                                              /--*  t14    int
Generating: N021 (  5,  6) [000016] UA-----------             *  STORE_LCL_FLD int    V00 loc0         ud:4->0[+4] Fseq[num2] NA REG NA
IN000a:             str     r3, [sp+0x1c]       // [V00 loc0+0x04]
Added IP mapping: 0x001A STACK_EMPTY (G_M4347_IG02,ins#10,ofs#24)
Generating: N023 (  5,  6) [000023] -------------                IL_OFFSET void   IL offset: 0x1a REG NA
Generating: N025 (  1,  1) [000020] -------------       t20 =    CNS_INT   int    13 REG r3 $84
IN000b:             movs    r3, 13
                                                              /--*  t20    int
Generating: N027 (  5,  6) [000022] UA-----------             *  STORE_LCL_FLD int    V00 loc0         ud:5->0[+8] Fseq[num3] NA REG NA
IN000c:             str     r3, [sp+0x20]       // [V00 loc0+0x08]
Added IP mapping: 0x0023 STACK_EMPTY (G_M4347_IG02,ins#12,ofs#28)
Generating: N029 (  5,  6) [000029] -------------                IL_OFFSET void   IL offset: 0x23 REG NA
Generating: N031 (  1,  1) [000026] -------------       t26 =    CNS_INT   int    14 REG r3 $85
IN000d:             movs    r3, 14
                                                              /--*  t26    int
Generating: N033 (  5,  6) [000028] UA-----------             *  STORE_LCL_FLD int    V00 loc0         ud:6->0[+12] Fseq[num4] NA REG NA
IN000e:             str     r3, [sp+0x24]       // [V00 loc0+0x0c]
Added IP mapping: 0x002C STACK_EMPTY (G_M4347_IG02,ins#14,ofs#32)
Generating: N035 (  5,  6) [000035] -------------                IL_OFFSET void   IL offset: 0x2c REG NA
Generating: N037 (  1,  1) [000032] -------------       t32 =    CNS_INT   int    15 REG r3 $86
IN000f:             movs    r3, 15
                                                              /--*  t32    int
Generating: N039 (  5,  6) [000034] UA-----------             *  STORE_LCL_FLD int    V00 loc0         ud:7->0[+16] Fseq[num5] NA REG NA
IN0010:             str     r3, [sp+0x28]       // [V00 loc0+0x10]
Added IP mapping: 0x0035 STACK_EMPTY (G_M4347_IG02,ins#16,ofs#36)
Generating: N041 (  5,  6) [000041] -------------                IL_OFFSET void   IL offset: 0x35 REG NA
Generating: N043 (  1,  1) [000038] -------------       t38 =    CNS_INT   int    16 REG r3 $87
IN0011:             movs    r3, 16
                                                              /--*  t38    int
Generating: N045 (  5,  6) [000040] UA-----------             *  STORE_LCL_FLD int    V00 loc0         ud:8->0[+20] Fseq[num6] NA REG NA
IN0012:             str     r3, [sp+0x2c]       // [V00 loc0+0x14]
Added IP mapping: 0x003E STACK_EMPTY (G_M4347_IG02,ins#18,ofs#40)
Generating: N047 ( 39, 30) [000058] -------------                IL_OFFSET void   IL offset: 0x3e REG NA
Generating: N049 (  1,  1) [000045] -------------       t45 =    CNS_INT   int    3 REG r3 $88
IN0013:             movs    r3, 3
                                                              /--*  t45    int
Generating: N047 ( 39, 30) [000058] -------------                IL_OFFSET void   IL offset: 0x3e REG NA
Generating: N049 (  1,  1) [000045] -------------       t45 =    CNS_INT   int    3 REG r3 $88
IN0013:             movs    r3, 3
                                                              /--*  t45    int
Generating: N051 (???,???) [000086] -------------             *  PUTARG_STK [+0x10] int    REG NA
IN0014:             str     r3, [sp+0x10]       // [V01 OutArgs+0x10]
Generating: N053 (  1,  1) [000046] -------------       t46 =    CNS_INT   int    4 REG r3 $89
IN0015:             movs    r3, 4
                                                              /--*  t46    int
Generating: N055 (???,???) [000087] -------------             *  PUTARG_STK [+0x14] int    REG NA
IN0016:             str     r3, [sp+0x14]       // [V01 OutArgs+0x14]
Generating: N057 (  3,  2) [000044] -c-----N-----       t44 =    LCL_VAR_ADDR byref  V00 loc0         u:9 NA (last use) REG NA
                                                              /--*  t44    byref
Generating: N059 (  9,  7) [000051] -c-----------       t51 = *  OBJ(24)   struct REG NA $102
                                                              /--*  t51    struct
Generating: N061 (???,???) [000088] -------------       t88 = *  PUTARG_SPLIT struct REG r2
IN0017:             ldr     r3, [sp+0x20]       // [V00 loc0+0x08]
IN0018:             str     r3, [sp]    // [V01 OutArgs]
IN0019:             ldr     r3, [sp+0x24]       // [V00 loc0+0x0c]
IN001a:             str     r3, [sp+0x04]       // [V01 OutArgs+0x04]
IN001b:             ldr     r3, [sp+0x28]       // [V00 loc0+0x10]
IN001c:             str     r3, [sp+0x08]       // [V01 OutArgs+0x08]
IN001d:             ldr     r3, [sp+0x2c]       // [V00 loc0+0x14]
IN001e:             str     r3, [sp+0x0c]       // [V01 OutArgs+0x0c]
IN001f:             ldr     r2, [sp+0x18]       // [V00 loc0]
IN0020:             ldr     r3, [sp+0x1c]       // [V00 loc0+0x04]
Generating: N063 (  1,  1) [000042] -------------       t42 =    CNS_INT   int    1 REG r0 $8a
IN0021:             movs    r0, 1
                                                              /--*  t42    int
Generating: N065 (???,???) [000089] -------------       t89 = *  PUTARG_REG int    REG r0
Generating: N067 (  1,  1) [000043] -------------       t43 =    CNS_INT   int    2 REG r1 $8b
IN0022:             movs    r1, 2
                                                              /--*  t43    int
Generating: N069 (???,???) [000090] -------------       t90 = *  PUTARG_REG int    REG r1
Generating: N071 (  3,  7) [000091] -------------       t91 =    CNS_INT(h) int    0x744F4930 ftn REG lr
IN0023:             movw    lr, 0x4930
IN0024:             movt    lr, 0x744f
                                                              /--*  t91    int
Generating: N073 (  6,  9) [000092] -------------       t92 = *  IND       int    REG lr
IN0025:             ldr     lr, [lr]
                                                              /--*  t88    struct arg2 r2,r3
                                                              +--*  t89    int    arg0 in r0
                                                              +--*  t90    int    arg1 in r1
                                                              +--*  t92    int    control expr
Generating: N075 ( 35, 26) [000047] --C-G--------       t47 = *  CALL      int    BringUpTest.StructAdd $202
Call: GCvars=00000000 {}, gcrefRegs=0000 {}, byrefRegs=0000 {}
IN0026:             blx     lr          // BringUpTest:StructAdd(int,int,struct,int,int):int

```

src/jit/codegenarmarch.cpp
```diff
@@ -1617,11 +1766,31 @@ void CodeGen::genCallInstruction(GenTreeCall* call)
...
+                }
+            }
+            else
+#endif
+            {
+                regNumber argReg = curArgTabEntry->regNum;
+                if (argNode->gtRegNum != argReg)
+                {
+                    {
+                        inst_RV_RV(ins_Move_Extend(argNode->TypeGet(), true), argReg, argNode->gtRegNum);
```
BruceForstall: Why are there two scopes here?

src/jit/gentree.h
```diff
@@ -4887,6 +4899,165 @@ struct GenTreePutArgStk : public GenTreeUnOp
...
+    // First reg of struct is aloways given by gtRegNum.
+    // gtOtherRegs holds the other reg numbers of struct.
+    // HFA args is not yet handled.
+    regNumberSmall gtOtherRegs[MAX_REG_ARG - 1];
+
+    // GTF_SPILL or GTF_SPILLED flag on a multi-reg struct node indicates that one or
+    // more of its result regs are in that state.  The spill flag of each of the
+    // return register is stored in the below array.
+    unsigned gtSpillFlags[MAX_REG_ARG];
+
```
BruceForstall: This looks the same as the `FEATURE_MULTIREG_RET` code in GenTreeCall. Can you adopt the recent change I made there, converting gtSpillFlags from an array of 4 32-bit unsigned elements (for arm32), namely 16 bytes, to a single 8-bit `unsigned char`? The key is that only data you need to store is 2 bits, for GTF_SPILL and GTF_SPILLED.

src/jit/gentree.h
```diff
@@ -4887,6 +4899,165 @@ struct GenTreePutArgStk : public GenTreeUnOp
...
+    // specified by its index.
+    //
+    // Arguments:
+    //    idx  -  Position or index of the register
+    //
+    // Return Value:
+    //    Returns GTF_* flags associated with.
+    unsigned GetRegSpillFlagByIdx(unsigned idx) const
+    {
+        assert(idx < MAX_RET_REG_COUNT);
```
BruceForstall: >MAX_RET_REG_COUNT [](start = 21, length = 17)

This should be:
```
assert(idx < MAX_REG_ARG);
```

src/jit/codegenarmarch.cpp
```diff
@@ -1617,11 +1766,31 @@ void CodeGen::genCallInstruction(GenTreeCall* call)
...
             genConsumeReg(argNode);
-            if (argNode->gtRegNum != argReg)
+#ifdef _TARGET_ARM_
+            if (curArgTabEntry->isSplit)
             {
-                inst_RV_RV(ins_Move_Extend(argNode->TypeGet(), true), argReg, argNode->gtRegNum);
+                assert(curArgTabEntry->numRegs >= 1);
+                for (unsigned idx = 0; idx < curArgTabEntry->numRegs; idx++)
+                {
+                    regNumber regDst = (regNumber)((unsigned)curArgTabEntry->regNum + idx);
```
BruceForstall: nit: instead of `regDst`, I would use `argReg`, since that is used elsewhere in the function to convey the same concept.

src/jit/gentree.cpp
```diff
@@ -1742,6 +1748,24 @@ regMaskTP GenTree::gtGetRegMask() const
...
+            GenTree*            tree     = const_cast<GenTree*>(this);
+            GenTreePutArgSplit* splitArg = tree->AsPutArgSplit();
+            unsigned            regCount = splitArg->gtNumRegs;
+
+            resultMask = RBM_NONE;
+            for (unsigned i = 0; i < regCount; ++i)
+            {
+                regNumber reg = splitArg->GetRegNumByIdx(i);
+                if (reg != REG_NA)
+                {
```
BruceForstall: Can this reg ever be REG_NA for split args? Or can you instead use `assert(reg != REG_NA)`?

src/jit/gentree.cpp
```diff
@@ -11833,6 +11860,41 @@ void Compiler::gtGetArgMsg(
...
+                    if (curArgTabEntry->numRegs == 1)
+                    {
+                        sprintf_s(bufp, bufLength, "arg%d %s out+%02x%c", argNum, compRegVarName(argReg),
+                                  (curArgTabEntry->slotNum) * TARGET_POINTER_SIZE, 0);
+                    }
+                    else
+                    {
+                        regNumber otherRegNum = (regNumber)((unsigned)(argReg) + curArgTabEntry->numRegs - 1);
+                        char      seperator   = (curArgTabEntry->numRegs == 2) ? ',' : '-';
```
BruceForstall: >seperator [](start = 34, length = 9)

typo: seperator => separator

src/jit/gentree.cpp
```diff
@@ -11833,6 +11860,41 @@ void Compiler::gtGetArgMsg(
...
+                        sprintf_s(bufp, bufLength, "arg%d %s%c%s out+%02x%c", argNum, compRegVarName(argReg), seperator,
+                                  compRegVarName(otherRegNum), (curArgTabEntry->slotNum) * TARGET_POINTER_SIZE, 0);
+                    }
+                }
+                else
+                {
+                    regNumber curReg = (regNumber)((unsigned)(argReg) + listCount);
+                    if (((unsigned)(curReg)) <= MAX_ARG_REG_COUNT)
+                    {
```
BruceForstall: It seems wrong to cast a `regNumber` to `unsigned` and then compare it against a count. Should this instead by `if (listCount <= MAX_ARG_REG_COUNT)`?

src/jit/gentree.h
```diff
@@ -4887,6 +4899,165 @@ struct GenTreePutArgStk : public GenTreeUnOp
...
+                           putIncomingArgArea,
+                           callNode)
+        , gtNumRegs(numRegs)
+    {
+        gtIsHfa = isHfa;
+        ClearOtherRegs();
+        ClearOtherRegFlags();
+    }
+
+    // First reg of struct is aloways given by gtRegNum.
```
CarolEidt: Typo: aloways => always

src/jit/lower.cpp
```diff
@@ -809,112 +808,166 @@ GenTreePtr Lowering::NewPutArg(GenTreeCall* call, GenTreePtr arg, fgArgTabEntryP
...
+            NYI_ARM("Lowering: Oper for struct argument is not GT_OBJ");
         }
+
+        putArg = new (comp, GT_PUTARG_SPLIT)
+            GenTreePutArgSplit(arg, info->slotNum PUT_STRUCT_ARG_STK_ONLY_ARG(info->numSlots), info->numRegs,
+                               info->isHfaRegArg, call->IsFastTailCall(), call);
+
+        // Set GC Pointer info
+        BYTE*    gcLayout = new (comp, CMK_Codegen) BYTE[info->numSlots + info->numRegs];
+        unsigned numRefs  = comp->info.compCompHnd->getClassGClayout(arg->gtObj.gtClass, gcLayout);
```
CarolEidt: If you have a `GT_OBJ` node here, you can get the gc info from the `GenTreeObj` node, rather than having to make another call across the JIT\EE interface. And perhaps you could simply construct the layout for the `GT_FIELD_LIST` case once that is handled.

src/jit/codegenarmarch.cpp
hseok-oh: My mistake. I'll fix it

src/jit/gentree.h
hseok-oh: I'll rebase it and adopt recent change.

src/jit/gentree.cpp
hseok-oh: I'll change this to use assert.

src/jit/gentree.cpp
hseok-oh: For example, If `a(int a, int b, struct c)` and `sizeof(struct c)` is 8 byte, `struct c` uses `r2` and `r3`. So we need to compare MAX_ARG_REG_COUNT with the last register, not `listcount`.

src/jit/lower.cpp
hseok-oh: Thanks. I'll fix it.

src/jit/gentree.cpp
BruceForstall: Here's an alternative:
```
unsigned regArgNum = genMapIntRegNumToRegArgNum(argReg) + listCount;
if (regArgNum < MAX_REG_ARG)
{
    regNumber curReg = genMapIntRegArgNumToRegNum(regArgNum);
    sprintf_s(bufp, bufLength, "arg%d m%d %s%c", argNum, listCount, compRegVarName(curReg), 0);
}
else
{
    unsigned stackSlot = regArgNum - MAX_REG_ARG;
    sprintf_s(bufp, bufLength, "arg%d m%d out+%s%c", argNum, listCount, stackSlot, 0);
}
```

Notes:
1. `MAX_REG_ARG` is the maximum number of register arguments (for arm, that's r0, r1, r2, r3); `MAX_ARG_REG_COUNT` is the maximum number of argument registers that can be occupied by a single call argument.
2. The code above doesn't require casting between indices or offsets and registers.
3. I think the code before had a bug, by using `<=` instead of `<` -- it would allow curReg of 4, which would be r4, but the max argument register is r3.


---
In reply to: [122913222](https://github.com/dotnet/coreclr/pull/12050#discussion_r122913222) [](ancestors = 122913222)

hseok-oh: OK. I understood it. Thank you for kind comment. I'll fix it.

hqueue: @hseok-oh If so, than how about useing `REG_ARG_LAST` instead of `MAX_ARG_REG_COUNT` ?
Becasue `MAX_ARG_REG_COUNT` means maximum register count for a single argument, not regsiter number.
 as stated in `src/jit/target.h` below.
```
  #define MAX_ARG_REG_COUNT             4  // Maximum registers used to pass a single argument in multiple registers. (max is 4 floats or doubles using an HFA)
```

hqueue: @hseok-oh In addition, how about using `REG_ARG_LAST` instead of `MAX_ARG_REG_COUNT` ?
Becasue `MAX_ARG_REG_COUNT` means maximum register count for a single argument, not register number.
 as stated in `src/jit/target.h` below.
```
  #define MAX_ARG_REG_COUNT             4  // Maximum registers used to pass a single argument in multiple registers. (max is 4 floats or doubles using an HFA)
```

hseok-oh: @hqueue I fixed it based on @BruceForstall 's comment to consider HFA.

src/jit/lower.cpp
hseok-oh: @CarolEidt It seems that gc info is not initialized in this case. I got assertion failure when I tried to get gc info from `GT_OBJ`. So we need to generate gc info here.

hseok-oh:
@CarolEidt @BruceForstall I've applied your feedback. PTAL.
- Rebase
- Update managing spillFlag
- Implement spill & restore
- Fix typos and rename variables
- Fix bug related to print gentree

src/jit/codegenarmarch.cpp
```diff
@@ -1617,11 +1766,29 @@ void CodeGen::genCallInstruction(GenTreeCall* call)
         }
         else
         {
-            regNumber argReg = curArgTabEntry->regNum;
             genConsumeReg(argNode);
```
CarolEidt: It seems that you are still consuming only the first register here. Perhaps you need a `genConsumePutArgSplit` to consume all the register operands?

hseok-oh: I updated `genUnspillRegIfNeeded` (https://github.com/hseok-oh/coreclr/blob/b5d86a6cec83522e6dcb784076f9136d81547129/src/jit/codegenlinear.cpp#L1007-L1035) called by `genConsumeReg`. 
And I added `gtSpillFlags` field in `GenTreePutArgSplit`. (https://github.com/hseok-oh/coreclr/blob/b5d86a6cec83522e6dcb784076f9136d81547129/src/jit/gentree.h#L4969)
To handle this flag, I also updated `genProduceReg` in `codegenlinear.cpp`, `rsSpillTree` and `rsUnspillInPlace` in `regset.cpp`. Is there more function to update to consume all registers?

CarolEidt: The problem is that there are multiple registers, so `genConsumeReg()` needs to be called for each. That should take care of calling `genUnspillRegIfNeeded()`. The problem is that if you only call `genConsumeReg()` for the first register, then the liveness won't be updated for the other registers.

src/jit/lower.cpp
CarolEidt: The gc info should have been initialized when the `GT_OBJ` is created, unless it is known to have zero gc refs. That is, `IsGCInfoInitialized()` should return `true`, and `gtGcPtrs` should be non-null if `gtGcPtrCount` is non-zero. If that is not the case, then it is a bug.

src/jit/codegenarmarch.cpp
hseok-oh: We can call `CodeGen::genConsumeReg(GenTree* tree)` to consume register for each argument . On other architectures, if argument is multi-reg struct (or long), it will be morphed to `FieldList` and each field has its `PutArgReg` node, and we can call `genConsumeReg()` for each `PutArgReg` node. But we don't morph to `FieldList` all multi-reg struct arg on arm32. For example, some parts of very large size struct can be passed using registers and IMO it makes too complex to morph this struct. So I defined new node type `PutArgSplit`, and it has multi-reg targets with stack slot. I updated this feature similar to multi-reg return.

CarolEidt: I'm not sure what you're saying here. This code is not correct as-is, because multiple registers are being used, but you are not calling `genConsumeReg` for each one.

hseok-oh: I cannot understand why we need to update the liveness when `genConsumeReg()` is called from `genCallInstruction()`. I thought that we should update liveness in `genConsumeReg()` when the argument `tree` of `genConsumeReg()` represents local variable (In other words, `tree->OperIsLocal() == true`). 
When `genConsumeReg()` is called from `genCallInstruction()`, `tree` does not represent local variable (it always has `GT_PUTARG_REG` as oper in general, and `GT_PUTARG_SPLIT` for split struct on arm32). As a result, updating livenesss in `genConsumeReg()` is skipped. So  I modified only `genUnspillRegIfNeeded()`.
Please tell me what I misunderstand.

CarolEidt: If it is the case that a `fgArgTabEntry` with `isSplit == true` will always have a `GT_PUTARG_SPLIT`, then you should assert it here. In that case, please add a comment to indicate that a `genConsumeReg` is not needed, and move the `genConsumeReg` above to the else clause below here. Otherwise it makes no sense that you have consumed just a single register for a node that has multiple registers.

---
In reply to: [123937346](https://github.com/dotnet/coreclr/pull/12050#discussion_r123937346) [](ancestors = 123937346)

CarolEidt: If it is the case that a `fgArgTabEntry` with `isSplit == true` will always have a `GT_PUTARG_SPLIT`, then you should assert it here. In that case, please add a comment to indicate that a `genConsumeReg` is not needed, and move the `genConsumeReg` above to the else clause below here. Otherwise it makes no sense that you have consumed just a single register for a node that has multiple registers.

hseok-oh: Now I understand that you wanted to define special `genConsumeReg` function for `PutArgSplit` such as `genConsumeAddress`, `genConsumeRegs`, or `genConsumePutStructArgStk`, right? I also it's good approach because `PutArgSplit` consume register(s), and it will be better to handle liveness in this special function if need.

CarolEidt: Yes. I think you are correct in that, if we always expect a `GT_PUTARG_SPLIT`, then it is sufficient to ensure that any spill or unspill is handled as needed.  But I think it would be best to handle this in a consistent way - e.g. with something like a `genConsumePutArgSplit`.

hseok-oh: I defined new function `genConsumeArgSplitStruct`. PTAL

src/jit/regset.cpp
```diff
@@ -1535,8 +1538,15 @@ void RegSet::rsSpillTree(regNumber reg, GenTreePtr tree, unsigned regIdx /* =0 *
         ReturnTypeDesc* retTypeDesc = call->GetReturnTypeDesc();
         treeType                    = retTypeDesc->GetReturnRegType(regIdx);
     }
+#ifdef _TARGET_ARM_
+    else if (tree->OperIsPutArgSplit())
+    {
+        splitArg = tree->AsPutArgSplit();
+        treeType = splitArg->GetRegType(regIdx);
+    }
     else
```
CarolEidt: This line is now under `#ifdef _TARGET_ARM_`, so for the non-arm case what is supposed to be the else clause of the `if (tree->IsMultiRegCall())` is unconditionally executed, which could possibly explain the x86 test failure.

hseok-oh:
@CarolEidt I've update
- Fix bug in regset.cpp
- Add comments in morph.cpp's NYI_ARM
- Fix comments' typo in lsraarmarcp.cpp
PTAL

CarolEidt:
@BruceForstall - I think this is ready to merge.

hseok-oh:
@CarolEidt @BruceForstall Thank you for review!

Issue #12029, opened 2017-06-01
UnmanagedType.CustomMarshaler not working for struct fields in 2.0.0-preview1-002111-00


dlech:
I'm trying to marshal a struct using a custom marshaller for one of the fields, but it is failing with an "unknown error".

Sample Code:

```
using System;
using System.Runtime.InteropServices;

namespace CustomMarshalTest
{
    
    struct TestStruct
    {
        [MarshalAs(UnmanagedType.CustomMarshaler, MarshalTypeRef = typeof(Custom))]
        public string Value;
    }

    class Custom : ICustomMarshaler
    {
        static Custom instance = new Custom();

        public static ICustomMarshaler GetInstance(string cookie) => instance;

        public void CleanUpManagedData(object ManagedObj)
        {
        }

        public void CleanUpNativeData(IntPtr pNativeData)
        {
            if (pNativeData != IntPtr.Zero) {
                Marshal.FreeCoTaskMem(pNativeData);
            }
        }

        public int GetNativeDataSize()
        {
            return IntPtr.Size;
        }

        public IntPtr MarshalManagedToNative(object ManagedObj)
        {
            var str = ManagedObj as string;
            if (str == null) {
                return IntPtr.Zero;
            }
            return Marshal.StringToCoTaskMemUni(str);
        }

        public object MarshalNativeToManaged(IntPtr pNativeData)
        {
            if (pNativeData == IntPtr.Zero) {
                return null;
            }
            return Marshal.PtrToStringUni(pNativeData);
        }
    }

    class Program
    {
        static void Main(string[] args)
        {
            try {
                Console.WriteLine("Hello World!");
                var s = new TestStruct { Value = "mfg" };
                var ptr = Marshal.AllocCoTaskMem(IntPtr.Size);
                Marshal.StructureToPtr<TestStruct>(s, ptr, false);
                var x = Marshal.PtrToStructure<TestStruct>(ptr);
                Marshal.FreeCoTaskMem(ptr);
            } catch (Exception ex) {
                Console.WriteLine(ex);
            }
        }
    }
}
```

Steps to repeat:

```
dotnet new console
# paste code into Program.cs
dotnet restore
dotnet run
```

Result:

```
Hello World!
System.TypeLoadException: Cannot marshal field 'Value' of type 'CustomMarshalTest.TestStruct': Unknown error.
   at System.Runtime.InteropServices.Marshal.StructureToPtr(Object structure, IntPtr ptr, Boolean fDeleteOld)
[...]

PR #12051, opened 2017-06-02
Fix a bug in LoadNativeStringResource to honor the contract properly when the buffer is not big enough

yizhang82:
Fixes bad error message in https://github.com/dotnet/coreclr/issues/12029. The code in https://github.com/dotnet/coreclr/blob/master/src/vm/fieldmarshaler.cpp#L4255-L4256 relies on proper reporting of ERROR_INSUFFICIENT_BUFFER and setting the size - otherwise it'll show the default fallback error "Unknown error". 
LoadNativeStringResource now aligns with what we do in windows. 
@jkotas PTAL


jkotas:
LGTM 

@janvorli may want to take a look as well since he has done the original implementation of some of these parts.

PR #12053, opened 2017-06-02
Do not copy bin subdirectories to test core_root

mikedn:
As far as I can tell there's no need to copy various bin subdirectories (PDB, .nuget, IL etc.) to core_root when running tests and they have hundreds of megabytes.

### DIFF
--- a/build-test.cmd
+++ b/build-test.cmd
@@ -286,7 +286,7 @@ if exist "%CORE_ROOT%" rd /s /q "%CORE_ROOT%"
 if exist "%CORE_ROOT_STAGE%" rd /s /q "%CORE_ROOT_STAGE%"
 md "%CORE_ROOT%"
 md "%CORE_ROOT_STAGE%"
-xcopy /s "%__BinDir%" "%CORE_ROOT_STAGE%"
+xcopy "%__BinDir%" "%CORE_ROOT_STAGE%"
 
 
 if defined __BuildAgainstPackagesArg ( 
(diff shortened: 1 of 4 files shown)


mikedn:
Hmm, looks like runtest.proj depends on the .nugget directory somehow: https://github.com/dotnet/coreclr/blob/master/tests/runtest.proj#L362. For what it's not clear, tests run fine. Do we even need that .json file anymore?

mikedn:
tests.zip produced by a Windows debug build is 300 megabytes smaller:

https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_windows_nt_prtest/7518/artifact/bin/tests/
https://ci.dot.net/job/dotnet_coreclr/job/master/job/debug_windows_nt_prtest/7519/artifact/bin/tests/

BruceForstall:
@wtgodbe PTAL

wtgodbe:
I'm 99% sure it won't, but there's a chance that the against-packages test build will be broken by this (that scenario isn't tested in CI). I'm running that build locally now, assuming that succeeds, LGTM.

wtgodbe:
Actually, never mind, I thought that your change was in build-test, not runtest. You may want to make a similar change here - https://github.com/dotnet/coreclr/blob/master/build-test.cmd#L289

wtgodbe:
Local build succeeded with your change & the change I mentioned above

mikedn:
So if it works does that mean that the json file generate by runtest.proj is no longer needed?

wtgodbe:
We don't even restore project.json's anymore (we use .csproj's instead), so we should be able to get rid of that generated project.json.

mikedn:
I've updated build-test.cmd and removed the JSON building target. 

As far as I can tell the OSX build failure is not related to these changes, just some Jenkins issue.

mikedn:
@wtgodbe Good to go?

wtgodbe:
Looks fine to me, pending:

PR #12055, opened 2017-06-02
[WIP] Simplify long relop lowering

mikedn:


mikedn:
jit-diff summary:
```
Total bytes of diff: -9399 (-0.09 % of base)
    diff is an improvement.
          66 : System.Private.Uri.dasm (0.09 % of base)
       -4426 : System.Private.CoreLib.dasm (-0.16 % of base)
       -2427 : System.Reflection.Metadata.dasm (-3.06 % of base)
        -472 : Microsoft.CodeAnalysis.VisualBasic.dasm (-0.03 % of base)
        -435 : Microsoft.CodeAnalysis.dasm (-0.07 % of base)
        -319 : System.Net.Http.dasm (-0.19 % of base)
35 total files with size differences (34 improved, 1 regressed), 44 unchanged.
          91 : System.Reflection.Metadata.dasm - TypeDefTableReader:FindTypeContainingMethod(int,int):int:this
          91 : System.Reflection.Metadata.dasm - TypeDefTableReader:FindTypeContainingField(int,int):int:this
          87 : System.Private.Uri.dasm - Uri:CheckKnownSchemes(int,char,byref):bool
          76 : System.Private.CoreLib.dasm - UmAlQuraCalendar:ConvertGregorianToHijri(struct,byref,byref,byref)
          51 : System.Linq.Parallel.dasm - DoubleMinMaxAggregationOperatorEnumerator`1:MoveNextCore(byref):bool:this
        -830 : System.Reflection.Metadata.dasm - MetadataReader:InitializeTableReaders(struct,ubyte,ref,ref):this
        -548 : System.Reflection.Metadata.dasm - PEHeader:.ctor(byref):this
        -255 : System.Private.CoreLib.dasm - GenericArraySortHelper`1:DownHeap(ref,int,int,int) (18 methods)
        -227 : Microsoft.CodeAnalysis.VisualBasic.dasm - ExpressionEvaluator:PerformCompileTimeBinaryOperation(char,byte,ref,ref,ref):ref
        -213 : System.Reflection.Metadata.dasm - MetadataReader:InitializeStreamReaders(byref,ref,byref,byref,byref):this
772 total methods with size differences (681 improved, 91 regressed), 64719 unchanged.
```

mikedn:
It's tempting to change the `JTRUE(long relop)` "decomposition" to use the same approach that's used by all other long relops, the JIT code is simpler and the generated code always contains a single conditional branch instead of 2 or 3.

However, there are some ugly CQ regressions here in there. For example:
```asm
; before
       cmp      eax, 0xD1FFAB1E
       ja       SHORT G_M2396_IG05
       jb       SHORT G_M2396_IG06
       test     ecx, ecx
       jbe      SHORT G_M2396_IG06
; after
       mov      dword ptr [esp+3CH], eax
       xor      eax, eax
       mov      dword ptr [esp+38H], eax
       mov      eax, 0xD1FFAB1E
       mov      dword ptr [esp+34H], eax
       mov      eax, dword ptr [esp+38H]
       cmp      eax, ecx
       mov      eax, dword ptr [esp+34H]
       sbb      eax, dword ptr [esp+3CH]
       jae      SHORT G_M2396_IG06
```

@CarolEidt @pgavlin @BruceForstall Thoughts about this?

pgavlin:
> However, there are some ugly CQ regressions here and there. For example:

This particular regression appears to be due to the RA, though my guess is that there are some tweaks we could make to the IR that could shorten the lifetimes of some of the temps and make the RA's job easier. Do you have the IR for this particular case, and are the other regressions similar?

pgavlin:
Thinking about this a little more, the updated version seems like it will always require additional registers relative to the current expansion for `JTRUE`, as the new expansion needs to write to a register. With that in mind, this might not be a great choice for situations in which register pressure is already rather high (e.g. x86).

cc @russellhadley

mikedn:
> Do you have the IR for this particular case
```
N004 (  3, 10) [000214] ------------      t214 =    const     int    0
               [000314] ------------      t314 =    const     int    0x7FF00000
                                                 /--*  t214   int    
                                                 +--*  t213   int    
               [000316] ----G-------             *  cmp       void  
                                                 /--*  t314   int    
                                                 +--*  t312   int    
               [000317] ------------      t317 = *  -Hi       int   
N006 ( 19, 34) [000196] ----G----U--                jcc       void   cond=>=
```
Might work better if I move those constants around.

> and are the other regressions similar

Sort of, all the large ones I've looked at involve RA one way or another. There may be smaller ones where RA is fine but the new approach simply generates larger code, I've yet to find an example in diffs.

> Thinking about this a little more, the updated version seems like it will always require additional registers relative to the current expansion for JTRUE, as the new expansion needs to write to a register.

Yes, that's a potential issue.

CarolEidt:
I agree with @pgavlin that some tweaking of the IR could reduce register lifetimes. However, it seems that there is also opportunity for making the 0 contained in the cmp.

mikedn:
> However, it seems that there is also opportunity for making the 0 contained in the cmp.

That doesn't work in this particular example because we have `cmp 0, reg` and not `cmp reg, 0`.

mikedn:
It looks like by moving constants around and doing a bunch of other tweaks I can increase the code size improvement from 9399 to 11552 bytes.

Regressions still exists. One example that doesn't involve RA:
```asm
; before
       cmp      ecx, 0xD1FFAB1E
       jne      SHORT G_M45178_IG03
       cmp      edx, 0xD1FFAB1E
       je       G_M45178_IG14
; after
       mov      ecx, eax
       sub      ecx, 0xD1FFAB1E
       mov      esi, edx
       sub      esi, 0xD1FFAB1E
       or       ecx, esi
       je       G_M45178_IG04
```
There's nothing that can be done about this, it trades branches for registers.

And another example that appears to involve RA:
```asm
; before
       sar      eax, 31
       cmp      dword ptr [ebp-58H], eax
       ja       SHORT G_M28149_IG09
       jb       SHORT G_M28149_IG10
       mov      eax, dword ptr [ebp-30H]
       cmp      dword ptr [ebp-34H], eax
       jbe      SHORT G_M28149_IG10
G_M28149_IG09:
       call     [Throw:OutOfBounds()]
G_M28149_IG10:
       mov      eax, bword ptr [ebp-50H]
       mov      edx, dword ptr [ebp-24H]
       add      edx, dword ptr [eax]
       movzx    eax, byte  ptr [edx]
       movzx    edx, byte  ptr [edx+1]
       shl      edx, 8
       or       eax, edx
       movzx    ecx, ax
       jmp      SHORT G_M28149_IG15
G_M28149_IG11:

; after
       sar      eax, 31
       mov      dword ptr [ebp-5CH], eax
       mov      eax, dword ptr [ebp-58H]
       cmp      dword ptr [ebp-3CH], eax
       mov      eax, dword ptr [ebp-5CH]
       sbb      eax, dword ptr [ebp-54H]
       jae      SHORT G_M28149_IG10
       call     [Throw:OutOfBounds()]
       mov      dword ptr [ebp-14H], edi
G_M28149_IG09:
       mov      eax, bword ptr [ebp-4CH]
       mov      edi, dword ptr [ebp-24H]
       add      edi, dword ptr [eax]
       mov      eax, edi
       mov      dword ptr [ebp-28H], eax
       movzx    eax, byte  ptr [eax]
       mov      dword ptr [ebp-54H], eax
       mov      edi, dword ptr [ebp-28H]
       movzx    eax, byte  ptr [edi+1]
       shl      eax, 8
       movzx    edi, dword ptr [ebp-54H]
       or       edi, eax
       movzx    edi, di
       jmp      G_M28149_IG13
G_M28149_IG10:
       mov      dword ptr [ebp-14H], edi
       jmp      SHORT G_M28149_IG09
G_M28149_IG11:
```
It's kind of strange that the actual compare code isn't significantly worse, it even uses the same number of registers - 1. But the code that follows the compare, well, it's bad.

mikedn:
jit-diff after improvement:
```
Total bytes of diff: -11585 (-0.11 % of base)
    diff is an improvement.

       -5304 : System.Private.CoreLib.dasm (-0.19 % of base)
       -2427 : System.Reflection.Metadata.dasm (-3.06 % of base)
        -584 : Microsoft.CodeAnalysis.VisualBasic.dasm (-0.03 % of base)
        -529 : Microsoft.CodeAnalysis.dasm (-0.09 % of base)
        -349 : System.Net.Http.dasm (-0.20 % of base)
        -320 : Microsoft.CSharp.dasm (-0.10 % of base)
        -306 : System.Runtime.Numerics.dasm (-0.54 % of base)
        -258 : System.Threading.Tasks.Dataflow.dasm (-0.22 % of base)
        -214 : System.IO.MemoryMappedFiles.dasm (-1.99 % of base)
        -172 : System.IO.Compression.dasm (-0.26 % of base)

36 total files with size differences (36 improved, 0 regressed), 43 unchanged.

          91 : System.Reflection.Metadata.dasm - TypeDefTableReader:FindTypeContainingMethod(int,int):int:this
          91 : System.Reflection.Metadata.dasm - TypeDefTableReader:FindTypeContainingField(int,int):int:this
          76 : System.Private.CoreLib.dasm - UmAlQuraCalendar:ConvertGregorianToHijri(struct,byref,byref,byref)
          47 : System.Private.Uri.dasm - Uri:CheckKnownSchemes(int,char,byref):bool
          47 : System.Security.Cryptography.Algorithms.dasm - DES:IsSemiWeakKey(ref):bool
          40 : System.Reflection.Metadata.dasm - MemoryBlock:BinarySearchReferenceRange(int,int,int,int,bool,byref,byref):this
          27 : System.Private.CoreLib.dasm - UnmanagedMemoryStream:Write(ref,int,int):this
          20 : System.Private.CoreLib.dasm - LongEnumEqualityComparer`1:IndexOf(ref,long,int,int):int:this
          17 : Microsoft.CodeAnalysis.dasm - Compilation:ConstructModuleSerializationProperties(ref,ref,struct):ref:this
          16 : System.Private.CoreLib.dasm - PathHelper:TryExpandShortFileName(byref,ref):ref

        -830 : System.Reflection.Metadata.dasm - MetadataReader:InitializeTableReaders(struct,ubyte,ref,ref):this
        -548 : System.Reflection.Metadata.dasm - PEHeader:.ctor(byref):this
        -255 : System.Private.CoreLib.dasm - GenericArraySortHelper`1:DownHeap(ref,int,int,int) (18 methods)
        -242 : System.Private.CoreLib.dasm - ReaderWriterLockSlim:TryEnterReadLockCore(struct):bool:this
        -221 : Microsoft.CodeAnalysis.VisualBasic.dasm - ExpressionEvaluator:PerformCompileTimeBinaryOperation(char,byte,ref,ref,ref):ref
        -213 : System.Reflection.Metadata.dasm - MetadataReader:InitializeStreamReaders(byref,ref,byref,byref,byref):this
        -132 : Microsoft.CSharp.dasm - ExpressionBinder:isConstantInRange(ref,ref,bool):bool
        -114 : Microsoft.CodeAnalysis.VisualBasic.dasm - OverloadResolution:FoldIntegralCharOrDateTimeBinaryOperator(int,ref,ref,ref,ref,byref,byref):ref
        -114 : System.Private.CoreLib.dasm - GenericArraySortHelper`1:PickPivotAndPartition(ref,int,int):int (18 methods)
        -112 : Microsoft.CodeAnalysis.dasm - SwitchIntegralJumpTableEmitter:CreateBucketLabels(struct):ref:this

794 total methods with size differences (739 improved, 55 regressed), 64696 unchanged.
```

mikedn:
In the first regression I showed the poor CQ was caused by the swapping of compare operands. Constants moved from op2 to op1 and thus they have to be loaded in registers. This can be avoided, instead of transforming `x <= 3` into `3 >= y` we can transform it into `x < 4` so the constant stays in op2. Now we get:
```asm
;before
       cmp      eax, 0xD1FFAB1E
       ja       SHORT G_M2396_IG05
       jb       SHORT G_M2396_IG06
       test     ecx, ecx
       jbe      SHORT G_M2396_IG06

; after
       cmp      ecx, 1
       sbb      eax, 0xD1FFAB1E
       jb       SHORT G_M2396_IG06
```

I don't think I can improve this further. Now it's up to the team to decide what to do with this.

pgavlin:
Do the regressions remain similar to the regressions in your original run?

mikedn:
The other 2 regression examples are unchanged. In particular, nothing I've tried had any effect on the top 2 method regressions - `TypeDefTableReader:FindTypeContainingMethod(int,int):int:this` and `TypeDefTableReader:FindTypeContainingField(int,int):int:this`.

pgavlin:
Thanks.

@CarolEidt @BruceForstall I think this is promising. It seems that we understand the regressions and IMO they are worth the code size improvements and reduction in branches. All that said, I still need to review the code itself :)

Thoughts?

CarolEidt:
> It seems that we understand the regressions and IMO they are worth the code size improvements and reduction in branches. 

I would agree. I'll do a review as well.

Issue #11970, opened 2017-05-30
using daily build nuget package


tmds:
https://github.com/dotnet/coreclr/blob/master/Documentation/workflow/OfficalAndDailyBuilds.md
describes the use of the Microsoft.NETCore.Runtime.CoreCLR NuGet package at https://dotnet.myget.org/gallery/dotnet-core.

I'm unable to make an application that consumes this package.

```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
 <packageSources>
    <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json"/>
 </packageSources>
</configuration>
```

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.NETCore.Runtime.CoreCLR" Version="2.1.0-preview1-25330-01"/>
  </ItemGroup>
</Project>
```

```
using System;

namespace consoleapp
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello World!");
        }
    }
}
```

```
$ dotnet restore -r linux-x64
$ dotnet publish -r linux-x64 --self-contained true
```
The version in the publish folder is:
```
$ strings bin/Debug/netcoreapp2.0/linux-x64/publish/libcoreclr.so | grep '@(#)'
@(#)Version 4.6.25302.01 built by: root-/tmp/tmp46857c8743c94ff2ab1fbff79bf87042.exec.cmd: line 2: hostname: command not found. Commit Hash: ea50f78e431f325c4382c21e9b91c3e76898ba1e
```
While the version in the NuGet package is:
```
$ strings ~/.nuget/packages/runtime.linux-x64.microsoft.netcore.runtime.coreclr/2.1.0-preview1-25330-01/runtimes/linux-x64/native/libcoreclr.so | grep '@(#)'
@(#)Version 4.6.25330.01 built by: root-/tmp/tmp8694902894184f4b92358bf5e3aab61b.exec.cmd: line 2: hostname: command not found. Commit Hash: ee454e11d0c73820dde10ab710f9d6eccd4471f1
```

tmds:
I updated the csproj file to set a RuntimeFrameworkVersion.
```
[...]

PR #12056, opened 2017-06-02
Describe how to use daily by setting RuntimeFrameworkVersion and publishing self-contained.

tmds:
Improve doc on consuming daily (https://github.com/dotnet/coreclr/issues/11970)

@weshaggard @joperezr

### DIFF
--- a/Documentation/workflow/OfficalAndDailyBuilds.md
+++ b/Documentation/workflow/OfficalAndDailyBuilds.md
@@ -12,11 +12,36 @@ and you can expect to see new versions roughly three months.   However it is als
 Team publishes **daily builds** of all sorts of packages including those built by the CoreCLR and CoreFX 
 repositories.  You can see what is available from
 
- * <https://dotnet.myget.org/gallery/dotnet-core>, and in particular you can see the builds of CoreCLR at 
- * <https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.Runtime.CoreCLR>.   
- 
+ * <https://dotnet.myget.org/gallery/dotnet-core>, and in particular you can see the builds of 
+ * CoreCLR at <https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.Runtime.CoreCLR>
+ * NETCore.App at <https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.App>
+
 Thus if your goal is just to get the latest bug fixes and features, you don't need to build CoreCLR yourself you 
-can simply add <https://dotnet.myget.org/F/dotnet-core/api/v3/index.json> to your Nuget Feed list. 
+can simply add <https://dotnet.myget.org/F/dotnet-core/api/v3/index.json> to your Nuget Feed list and set the
+`RuntimeFrameworkVersion` in your project file to a `Microsoft.NETCore.App` version. You need to restore
+and publish your application so it includes the runtime (`self-contained`).
+
+```
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+ <packageSources>
+    <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json"/>
+ </packageSources>
+</configuration>
+```
+```
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <RuntimeFrameworkVersion>2.0.0-preview2-*/RuntimeFrameworkVersion>
+  </PropertyGroup>
+</Project>
+```
+```
+$ dotnet restore -r linux-x64
+$ dotnet publish -r linux-x64 --self-contained true
+```
 
 ## Package Version Numbers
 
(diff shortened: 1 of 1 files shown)


Documentation/workflow/OfficalAndDailyBuilds.md
```diff
@@ -12,11 +12,36 @@ and you can expect to see new versions roughly three months.   However it is als
...
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <RuntimeFrameworkVersion>2.0.0-preview2-*/RuntimeFrameworkVersion>
+  </PropertyGroup>
+</Project>
+```
+```
+$ dotnet restore -r linux-x64
+$ dotnet publish -r linux-x64 --self-contained true
```
weshaggard: @eerhardt is the --self-contained true option necessary now?

Documentation/workflow/OfficalAndDailyBuilds.md
```diff
@@ -12,11 +12,36 @@ and you can expect to see new versions roughly three months.   However it is als
 Team publishes **daily builds** of all sorts of packages including those built by the CoreCLR and CoreFX 
 repositories.  You can see what is available from
 
- * <https://dotnet.myget.org/gallery/dotnet-core>, and in particular you can see the builds of CoreCLR at 
- * <https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.Runtime.CoreCLR>.   
- 
+ * <https://dotnet.myget.org/gallery/dotnet-core>, and in particular you can see the builds of 
+ * CoreCLR at <https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.Runtime.CoreCLR>
```
joperezr: NIT: is this extra bullet point intentional? I ask since it seems like it is a continuation from the first one

Documentation/workflow/OfficalAndDailyBuilds.md
joperezr: I thought that this was because of a bug, and it would eventually get fixed. Also, when I recently did this, I didn't pass in the `--self-contained` argument to the `dotnet` command, but instead set a property on my csproj. You can see an example [here](https://github.com/dotnet/corefx/blob/master/src/SharedFrameworkValidation/RestoreSDKProject/RestoreSDKProject.csproj#L8)
```xml
    ...
    <RuntimeIdentifier>$(_PublishRID)</RuntimeIdentifier>
    <SelfContained>true</SelfContained>
  </PropertyGroup>
  ...
```

Documentation/workflow/OfficalAndDailyBuilds.md
```diff
@@ -12,11 +12,36 @@ and you can expect to see new versions roughly three months.   However it is als
...
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <RuntimeFrameworkVersion>2.0.0-preview2-*/RuntimeFrameworkVersion>
+  </PropertyGroup>
+</Project>
+```
+```
+$ dotnet restore -r linux-x64
```
joperezr: NIT: given that we are not talking about linux specifically, I would just change the `-r linux-x64` with `-r RID` and then have a link somewhere in this same doc telling you the possible RIDs you could use.

Documentation/workflow/OfficalAndDailyBuilds.md
eerhardt: Passing `--self-contained true` it isn't necessary if:

1. The project is `OutputType=Exe`
2. You set a non-blank `RuntimeIdentifier`

If these two conditions are true, the `SelfContained` property defaults to `true`.  See https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.RuntimeIdentifierInference.targets#L102-L110

@joperezr - the example you gave didn't have `OutputType=Exe`, which is why explicitly setting `SelfContained` was necessary.

Documentation/workflow/OfficalAndDailyBuilds.md
eerhardt: You can also just put the `<RuntimeIdentifier>` into the .csproj, and then you don't need to pass it on the command line twice.

Documentation/workflow/OfficalAndDailyBuilds.md
tmds: This is intentional.

tmds:
@eerhardt @joperezr @weshaggard I've updated the doc based on your feedback. Can you approve when ok for you? The last one to approve can then merge the PR.

weshaggard:
LGTM Thanks @tmds for updating the docs.

Documentation/workflow/OfficalAndDailyBuilds.md
tmds: @eerhardt If you have a large solution where some projects have an RID set. Suppose you want to use a different RID. If you specify that as a property to msbuild, then projects which didn't have an rid set (-> publishing not selfcontained), will now publish self-contained. Is that right? Isn't that an issue?

eerhardt: That's a general issue/feature with any MSBuild property being set on the command line and building multiple projects. You'd get similar behavior if you said `dotnet publish -f netcoreapp2.0`.  All the projects would be published for `netcoreapp2.0`, even if they had `netcoreapp1.0` or `net461` in the project file.

When you specify a property on the command line, it applies to all projects being built in that command.

I guess I don't see how `RuntimeIdentifier` and `SelfContained` are any different.

PR #12057, opened 2017-06-02
Yield thread when using SleepEx(0, false) on Linux

helloguo:
This PR makes the calling thread to relinquish the CPU when calling SleepEx(0, false) so that SleepEx(0, false) behaves same on Windows and Linux.

### DIFF
--- a/src/pal/src/include/pal/corunix.hpp
+++ b/src/pal/src/include/pal/corunix.hpp
@@ -1098,6 +1098,10 @@ namespace CorUnix
             ) = 0;
 
         virtual
+        void
+        YieldThread() = 0;
+
+        virtual
         PAL_ERROR
         AbandonObjectsOwnedByThread(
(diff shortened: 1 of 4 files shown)


kouvel:
Thanks!

kouvel:
I think we can just call sched_yield directly there without having YieldThread, that is what is done elsewhere. Could consider exposing a PAL_YieldThread in the future if necessary for usage outside the PAL. I'll update it.

kouvel:
Actually it looks like SwitchToThread is the equivalent that is already exposed for usage outside the PAL

helloguo:
> I think we can just call sched_yield directly there without having YieldThread, that is what is done elsewhere. 

@kouvel You are right. The reason I put a wrapper here is because all the pthread related APIs are in IPalSynchronizationManager class.

helloguo:
@kouvel thanks for your update.

PR #12059, opened 2017-06-02
TypeForwardedFrom attributes on coretypes for NetFX serialization

ViktorHofer:
Relates to https://github.com/dotnet/corefx/pull/20599

cc @krwq


src/mscorlib/shared/System/StringComparer.cs
```diff
@@ -170,6 +171,7 @@ public int GetHashCode(object obj)
     }
 
     [Serializable]
+    [System.Runtime.CompilerServices.TypeForwardedFrom("mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089")]
     internal sealed class CultureAwareComparer : StringComparer
```
jkotas: Does the serialization on .NET Core still works with this change? These internal types do not exist in mscorlib on .NET Core. I suspect that we may need to add these internal types to the mscorlib façade in corefx somehow.

src/mscorlib/shared/System/Diagnostics/Tracing/StubEnvironment.cs
```diff
@@ -93,6 +93,7 @@ public static Tuple<T1> Create<T1>(T1 item1)
     }
 
     [Serializable]
+    [System.Runtime.CompilerServices.TypeForwardedFrom("mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089")]
```
jkotas: This is not necessary here - this code is not compiled into CoreCLR CoreLib.

ViktorHofer: Thanks!

src/mscorlib/shared/System/StringComparer.cs
ViktorHofer: I'm currently building offline... Will keep you updated

weshaggard: We currently have no way to add these internal types into the mscorlib facade. We use the shipped .NET Framework mscorlib assembly and only type forward public types. We cannot even type forward these types to S.Private.CoreLib without internals visible to. I would just exclude these internal ones unless we know they are needed for a specific reason.

src/mscorlib/src/System/Resources/LooselyLinkedResourceReference.cs
```diff
@@ -27,7 +27,7 @@ namespace System.Resources {
     using System.Globalization;
 
     [Serializable]
-
+    [System.Runtime.CompilerServices.TypeForwardedFrom("mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089")]
```
ViktorHofer: not sure if we need that here? cc @morganbr

src/mscorlib/src/System/Runtime/MemoryFailPoint.cs
```diff
@@ -440,6 +440,7 @@ private void Dispose(bool disposing)
 
 #if _DEBUG
         [Serializable]
+        [System.Runtime.CompilerServices.TypeForwardedFrom("mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089")]
```
ViktorHofer: not sure if we need that here? cc @morganbr

src/mscorlib/src/System/Resources/LooselyLinkedResourceReference.cs
ViktorHofer: not sure if we need that here? cc @morganbr

src/mscorlib/shared/System/StringComparer.cs
ViktorHofer: Here are the internal types listed. Without them we can't serialize Comparers, EqualityComparers, not even Exception.

```
CultureAwareComparer 
OrdinalComparer
OrdinalIgnoreCaseComparer
GenericComparer<T>
NullableComparer<T>
ObjectComparer<T>
Int32EnumComparer<T>
UInt32EnumComparer<T>
Int64EnumComparer<T>
UInt64EnumComparer<T>
GenericEqualityComparer<T>
NullableEqualityComparer<T>
ObjectEqualityComparer<T>
NonRandomizedStringEqualityComparer
ByteEqualityComparer 
EnumEqualityComparer<T>
SByteEnumEqualityComparer<T>
ShortEnumEqualityComparer<T>
LongEnumEqualityComparer<T>
ListDictionaryInternal 
DictionaryNode 
__RestrictedErrorObject <-- Used for exception data
```

stephentoub: > Without them we can't serialize Comparers, EqualityComparers

Which also means we fail to serialize many collections, right?

src/mscorlib/src/System/TimeZoneInfo.cs
```diff
@@ -29,6 +29,7 @@ internal enum TimeZoneInfoOptions
     };
 
     [Serializable]
+    [System.Runtime.CompilerServices.TypeForwardedFrom("mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089")]
     public sealed partial class TimeZoneInfo : IEquatable<TimeZoneInfo>, ISerializable, IDeserializationCallback
```
justinvp: `TimeZoneInfo` was originally in System.Core in .NET 3.5, but was type-forwarded to mscorlib in .NET 4.0. In the [reference source](https://github.com/Microsoft/referencesource/blob/90b323fe52bec428fe4bd5f007e9ead6b265d553/mscorlib/system/timezoneinfo.cs#L57), it has `[TypeForwardedFrom("System.Core, Version=3.5.0.0, Culture=Neutral, PublicKeyToken=b77a5c561934e089")]`. Does that need to match here?

src/mscorlib/shared/System/StringComparer.cs
weshaggard: Actually it doesn't mean we cannot serialize them it just means that we won't be able to deserialize across .NET Core and .NET Framework. I've not been up to speed with the latest thinking in this area but is that a goal for all the types we plan to support for binary serialization?

src/mscorlib/src/System/TimeZoneInfo.cs
weshaggard: Yes it should match.

src/mscorlib/shared/System/StringComparer.cs
stephentoub: > Actually it doesn't mean we cannot serialize them it just means that we won't be able to deserialize across .NET Core and .NET Framework

Right, that's what I meant.  That's the context of this TypeForwardedFrom change.

> is that a goal for all the types we plan to support for binary serialization?

Yes, for any types we ship as [Serializable] in .NET Core 2.0.

src/mscorlib/src/System/TimeZoneInfo.cs
justinvp: Same with `AdjustmentRule` and `TransitionTime`.

ViktorHofer: thanks!

src/mscorlib/shared/System/StringComparer.cs
ViktorHofer: cc @danmosemsft

weshaggard: If this is something we need to do then we are going to need to make a number of changes to support it. We are going to need to start manually creating and maintaining an mscorlib ref in corefx that contains these internal types. We are also going to need to give InternalsVisibileTo access to mscorlib from System.Private.CoreLib. With those changes we can technically support this but I would like to avoid doing that if at all possible.

danmosemsft: Wes is your concern the amount of work, the risk for 2.0, or the long term maintenance? As @stephentoub mentioned, it's a goal for 2.0 to support roundtripping. Viktor can work on it, and we still have time, it seems to me.

weshaggard: I'm mostly concerned with the long term maintenance and the general notion of exposing internals. I think the general binary serialization support is already going limit the set of changes we can make so I will just have to get over that. 

My concerns shouldn't block moving forward on this I just wanted to make sure folks understand what we are doing. @ViktorHofer I can help you figure out what needs to happen to accomplish this. Perhaps we can come up with a solution that doesn't require us to maintain an entire mscorlib ref assembly in corefx to expose these. We can probably make some changes to GenFacades to accomplish this.

stephentoub: > I think the general binary serialization support is already going limit the set of changes we can make so I will just have to get over that

Yes :frowning:  That's partially why we trimmed the support back to a core set of types that, from the data we have available, seems to represent the most important types used in this contexts.

ViktorHofer: Okay, let's discuss what needs to be done offline or in another thread.

src/mscorlib/src/System/Collections/ListDictionaryInternal.cs
```diff
@@ -505,6 +506,7 @@ public void Reset()
         }
 
         [Serializable]
+        [System.Runtime.CompilerServices.TypeForwardedFrom("mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089")]
```
weshaggard: I don't believe TypeForwardedFrom does anything useful for nested types.

ViktorHofer: thanks!

ViktorHofer:
@weshaggard I added the change here and will close https://github.com/dotnet/coreclr/pull/12083

ViktorHofer:
@jkotas @stephentoub PTAL

src/mscorlib/src/System/Resources/LooselyLinkedResourceReference.cs
```diff
@@ -27,7 +27,6 @@ namespace System.Resources {
     using System.Globalization;
 
     [Serializable]
```
jkotas: Or even better - remove the whole file. It is unused.

ViktorHofer: I don't think it gets into core as the it contains a conditional compilation: #if LOOSELY_LINKED_RESOURCE_REFERENCE. Should I still remove it?

ViktorHofer: @jkotas okay

Issue #11513, opened 2017-05-10
[Local GC] Undefined symbol when building standalone: g_TrapReturningThreads


swgillespie:
```
  "_g_TrapReturningThreads", referenced from:
      SVR::WaitLongerNoInstru(int) in gcsvr.cpp.o
      SVR::GCHeap::SafeToRestartManagedThreads() in gcsvr.cpp.o
      WKS::WaitLongerNoInstru(int) in gcwks.cpp.o
```

This is read in hot loops and it may not be possible to introduce an indirection here.

PR #12067, opened 2017-06-02
[Local GC] Add TrapReturningThreads to the GC/EE interface

swgillespie:
This fixes another linker error (https://github.com/dotnet/coreclr/issues/11513) that manifests when building standalone. This approach is the same as the one in https://github.com/dotnet/coreclr/pull/12036.

Of note is that at least one call to `GCToEEInterface::TrapReturningThreads` lies on a code path that is called every 8th spin of a spin lock. It seems to me that it should be okay to add an indirection on a spin lock path since we are likely about to yield to the scheduler or go to sleep anyway.

### DIFF
--- a/src/gc/env/gcenv.ee.h
+++ b/src/gc/env/gcenv.ee.h
@@ -49,6 +49,7 @@ public:
     static bool IsPreemptiveGCDisabled(Thread * pThread);
     static void EnablePreemptiveGC(Thread * pThread);
     static void DisablePreemptiveGC(Thread * pThread);
+    static bool TrapReturningThreads();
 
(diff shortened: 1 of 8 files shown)


swgillespie:

swgillespie:
@Maoni0 @jkotas @adityamandaleeka @sergiy-k PTAL? Should be pretty similar to my `GetThread` PR.

swgillespie:
@dotnet-bot test OSX10.12 x64 Checked Build and Test (GitHub having trouble today)

swgillespie:
thanks!

PR #12071, opened 2017-06-03
new JitEE interface method: expandRawHandleIntrinsic

sandreenko:
It is implemented only in CoreRT and used in importer.

Delete isDelegateCreationAllowed because it is unused and I was waiting for a JitEEinterface change 
to delete it.


src/jit/importer.cpp
```diff
@@ -6675,6 +6703,12 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
             call = impIntrinsic(newobjThis, clsHnd, methHnd, sig, pResolvedToken->token, readonlyCall,
                                 (canTailCall && (tailCall != 0)), &intrinsicID);
 
+            if (compInlineResult->IsFailure())
```
sandreenko: It is because now we call impLookupToTree inside impIntrinsic, that can set Failure if it is compIsForInlining() and we need runtime lookup.

sandreenko:
I didn't notice any order in these files (and code convention too), so I just added my method right after the last added resolveVirtualMethod.

src/jit/importer.cpp
```diff
@@ -3714,6 +3714,34 @@ GenTreePtr Compiler::impIntrinsic(GenTreePtr            newobjThis,
             break;
         }
 
+        case CORINFO_INTRINSIC_GetRawHandle:
+        {
+            assert(IsTargetAbi(CORINFO_CORERT_ABI)); // Only CoreRT supports it.
+            CORINFO_RESOLVED_TOKEN resolvedToken;
+            resolvedToken.hMethod      = method;
```
sandreenko: @MichalStrehovsky Is it necessary to set hMethod here? It is usually an output argument for resolvedToken.

sandreenko:
PTAL @dotnet/jit-contrib, @MichalStrehovsky. @jkotas.

src/jit/importer.cpp
MichalStrehovsky: You can null it out. We can fetch that from token/tokenScope (and in fact, we need do it from the token anyway if the method is canonical (and it pretty much always will be canonical)).

src/inc/corinfo.h
```diff
@@ -2111,13 +2121,6 @@ class ICorStaticInfo
             BOOL                        *pfIsOpenDelegate /* is the delegate open */
             ) = 0;
 
-    // Determines whether the delegate creation obeys security transparency rules
```
jkotas: While you are on it you can also delete `getAddrModuleDomainID` ... I have noticed earlier today that it is also unused.

jkotas:
LGTM

src/jit/importer.cpp
```diff
@@ -3714,6 +3714,34 @@ GenTreePtr Compiler::impIntrinsic(GenTreePtr            newobjThis,
...
+
+            CORINFO_GENERICHANDLE_RESULT embedInfo;
+            info.compCompHnd->expandRawHandleIntrinsic(&resolvedToken, &embedInfo);
+
+            GenTreePtr eeTypePtrOfNode = impLookupToTree(&resolvedToken, &embedInfo.lookup,
+                                                         gtTokenToIconFlags(memberRef), embedInfo.compileTimeHandle);
+            if (eeTypePtrOfNode == nullptr)
+                return nullptr;
+
+            unsigned eeSlot = lvaGrabTemp(true DEBUGARG("eeTypePtrOf"));
```
MichalStrehovsky: Nit: could you remove the references to `EETypePtrOf`?

While the original motivation for this is to allow having a `EETypePtr EETypePtrOf<T>()` method that intrinsically expands to _"raw value of the TypeHandle of T"_, I'm already thinking of using this for `IntPtr DefaultConstructorOf<T>()` that expands to _"pointer to the default constructor of type T"_ (to be used in the implementation of `Activator.CreateInstance<T>`). The API surface is general purpose to allow adding other interesting cases without requiring RyuJIT to learn about them.

sandreenko:
I think I do not understand the purpose jit\ICorJitInfo_API_names.h, @AndyAyersMS do you know that? Why is not resolveVirtualMethod there?

src/jit/importer.cpp
```diff
@@ -3714,6 +3714,34 @@ GenTreePtr Compiler::impIntrinsic(GenTreePtr            newobjThis,
...
+            impAssignTempGen(eeSlot, eeTypePtrOfNode, clsHnd, (unsigned)CHECK_SPILL_NONE);
+            GenTreePtr lclVar     = gtNewLclvNode(eeSlot, TYP_I_IMPL);
+            GenTreePtr lclVarAddr = gtNewOperNode(GT_ADDR, TYP_I_IMPL, lclVar);
+            var_types  resultType = JITtype2varType(sig->retType);
```
MichalStrehovsky: Would it be possible to add some kind of assert that the return type of the method is interchangeable with TYP_I_IMPL? (either an IntPtr, or a pointer-sized struct that is ideally not a GC pointer).

src/inc/corinfo.h
sandreenko: Thanks, done.

src/jit/importer.cpp
sandreenko: Cool, done.

src/jit/importer.cpp
sandreenko: I am not sure about a GC pointer check.
I added noway_assert that the return tree  has the same size as TYP_I_IMPL.

src/jit/importer.cpp
sandreenko: Ok, is rawPointer an appropriate name?

sandreenko:
@dotnet-bot test Windows_NT arm64 Checked pri1r2r

sandreenko:
@dotnet-bot test Windows_NT arm64 Checked pri1r2r

sandreenko:
@dotnet-bot test Windows_NT amd64Checked pri1r2r

sandreenko:
@dotnet-bot test Windows_NT amd64 Checked pri1r2r

sandreenko:
@dotnet-bot test Windows_NT Checked pri1r2r

BruceForstall:
Please port these changes to the desktop and build at least x86/x64/arm64 checked/release there before merging this (to ensure there are no desktop-specific build problems that will ensue when this is mirrored).

src/jit/importer.cpp
```diff
@@ -3714,6 +3714,38 @@ GenTreePtr Compiler::impIntrinsic(GenTreePtr            newobjThis,
             break;
         }
 
+        case CORINFO_INTRINSIC_GetRawHandle:
+        {
+            assert(IsTargetAbi(CORINFO_CORERT_ABI)); // Only CoreRT supports it.
+            CORINFO_RESOLVED_TOKEN resolvedToken;
```
BruceForstall: Seems like this should be a noway_assert

src/jit/ICorJitInfo_API_wrapper.hpp
```diff
@@ -189,17 +189,6 @@ BOOL WrapICorJitInfo::isCompatibleDelegate(
     return temp;
 }
 
-BOOL WrapICorJitInfo::isDelegateCreationAllowed(
-            CORINFO_CLASS_HANDLE        delegateHnd,
```
BruceForstall: Shouldn't you add a new case for expandRawHandleIntrinsic() in this file?

src/jit/ICorJitInfo_API_names.h
```diff
@@ -20,7 +20,6 @@ DEF_CLR_API(getUnmanagedCallConv)
 DEF_CLR_API(pInvokeMarshalingRequired)
 DEF_CLR_API(satisfiesMethodConstraints)
 DEF_CLR_API(isCompatibleDelegate)
-DEF_CLR_API(isDelegateCreationAllowed)
 DEF_CLR_API(isInstantiationOfVerifiedGeneric)
```
BruceForstall: Shouldn't you add a new case for expandRawHandleIntrinsic() in this file?

BruceForstall:
When this is merged, please be prepared to immediately do a desktop SuperPMI recollection.

src/jit/ICorJitInfo_API_names.h
sandreenko: I am not sure, I asked @AndyAyersMS in the message above why there is no resolveVirtualMethod there. Looks like we do not test code under MEASURE_CLRAPI_CALLS 1.

BruceForstall: This was created by Peter Kukol to measure JIT-EE interface overhead. It's probably reasonable to keep it working as close as possible, even if we don't actively build with it regularly.

sandreenko: Ok, I fixed that for both methods. I checked that it is possible to build VM with this flag on.

src/jit/ICorJitInfo_API_wrapper.hpp
sandreenko: Fixed

src/jit/importer.cpp

sandreenko:
> Please port these changes to the desktop and build at least x86/x64/arm64 checked/release there before merging this (to ensure there are no desktop-specific build problems that will ensue when this is mirrored).

Done for checked, it is in progress for release.

PR #12082, opened 2017-06-05
Fix crash due to lack of PSPSym in the CoreRT ABI

BredPet:
Signed-off-by: Petr Bred <bredpetr@gmail.com>

### DIFF
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -10162,11 +10162,12 @@ void CodeGen::genCaptureFuncletPrologEpilogInfo()
 #endif // DEBUG
 
         assert(PSP_slot_CallerSP_offset < 0);
-        assert(compiler->lvaPSPSym != BAD_VAR_NUM);
-        assert(PSP_slot_CallerSP_offset == compiler->lvaGetCallerSPRelativeOffset(compiler->lvaPSPSym)); // same offset
-                                                                                                         // used in main
-                                                                                                         // function and
-                                                                                                         // funclet!
+        if (compiler->lvaPSPSym != BAD_VAR_NUM)
+        {
+            assert(PSP_slot_CallerSP_offset ==
+                   compiler->lvaGetCallerSPRelativeOffset(compiler->lvaPSPSym)); // same offset used in main
+                                                                                 // function and funclet!
+        }
     }
 }
 


BredPet:
@Dmitri-Botcharnikov 
@sergign60 
@jkotas please review

src/jit/codegencommon.cpp
```diff
@@ -10162,11 +10162,14 @@ void CodeGen::genCaptureFuncletPrologEpilogInfo()
 #endif // DEBUG
 
         assert(PSP_slot_CallerSP_offset < 0);
-        assert(compiler->lvaPSPSym != BAD_VAR_NUM);
-        assert(PSP_slot_CallerSP_offset == compiler->lvaGetCallerSPRelativeOffset(compiler->lvaPSPSym)); // same offset
-                                                                                                         // used in main
-                                                                                                         // function and
-                                                                                                         // funclet!
+        if (!compiler->IsTargetAbi(CORINFO_CORERT_ABI)) // PSPSym is not used by the CoreRT ABI
```
BruceForstall: I guess PSP_slot_CallerSP_offset is just garbage in CoreRT? And the only interesting thing here is that lvaPSPSym is undefined for CoreRT?

jkotas:
It would be nice to make this look same way as amd64: https://github.com/dotnet/coreclr/blob/master/src/jit/codegencommon.cpp#L10436

jkotas:
You can also make the arm version early out for `if (!compiler->ehAnyFunclets())` so that it is easier to diff with amd64.

src/jit/codegencommon.cpp
BredPet: lvaPSPSym is initialized with a BAD_VAR_NUM in lclvars.cpp+76 and doesn't change because in lclvars.cpp+3871:
```
// PSPSym and LocAllocSPvar are not used by the CoreRT ABI                                                                                                                               
if (!IsTargetAbi(CORINFO_CORERT_ABI))
{
#if FEATURE_EH_FUNCLETS
    if (ehNeedsPSPSym())
    {
        lvaPSPSym            = lvaGrabTempWithImplicitUse(false DEBUGARG("PSPSym"));
        LclVarDsc* lclPSPSym = &lvaTable[lvaPSPSym];
        lclPSPSym->lvType    = TYP_I_IMPL;
    }
#endif // FEATURE_EH_FUNCLETS
```

BredPet:
@jkotas ehAnyFunclets check already exists in that place

PR #12091, opened 2017-06-05
Disable tail call stress in GH_11689 if ZapDisable is enabled.

pgavlin:
Tail call stress does not mix well with ZapDisable due to the issues
described in #11408.

### DIFF
--- a/tests/src/JIT/Regression/JitBlue/GitHub_11689/GitHub_11689.csproj
+++ b/tests/src/JIT/Regression/JitBlue/GitHub_11689/GitHub_11689.csproj
@@ -35,13 +35,16 @@
     <Compile Include="GitHub_11689.cs" />
   </ItemGroup>
   <PropertyGroup>
+    <!-- NOTE: tailcall stress should be reenabled under zapdisable when #11408 is fixed -->
     <CLRTestBatchPreCommands><![CDATA[
 $(CLRTestBatchPreCommands)
-set COMPlus_TailcallStress=1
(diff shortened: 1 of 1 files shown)


pgavlin:
@JosephTremoulet @dotnet/jit-contrib PTAL

pgavlin:
@JosephTremoulet @dotnet/jit-contrib PTAL

pgavlin:
@RussKeldorph we ought to take this to 2.0 as well. Are we still in tell mode for test-only changes?

RussKeldorph:
@pgavlin No, I think we need to ask now, though not with the full template.  @Petermarcu heads up.

JosephTremoulet:
LGTM.  Are we supposed to be doing this for all new tests, or is there something about this particular test?

pgavlin:
> LGTM. Are we supposed to be doing this for all new tests, or is there something about this particular test?

I assume that this test requires tail call stress in order to repro its issue... is that not the case?

The problem with TailCallStress and ZapDisable is that TailCallStress does not perform sufficient legality checking which currently breaks some framework code (specifically, a call that passes the address of a local var is converted into a tailcall and the runtime asserts). The missing legality checks are tracked by #11408.

JosephTremoulet:
Oh, I see now. No, the issue had nothing to do with tail calls, I just didn't notice that when copying another test's csproj to create this one. And I'm pretty sure I copied this one from one I added earlier, also not tail call related... I'll do a sweep for superfluous tail call stress setting.

PR #12096, opened 2017-06-05
Port commit 128aabe from master: disable tail call stress in GH_11689 if ZapDisable is enabled.

pgavlin:
Tail call stress does not mix well with ZapDisable due to the issues
described in #11408.


pgavlin:
@dotnet-bot test Windows_NT x64 Release Priority 1

pgavlin:

PR #12102, opened 2017-06-06
build.sh: fall back to portable Linux when Linux host has unknown rid

tmds:


tmds:
CC @ellismg

janvorli:
There is a small issue - the portable tools that we use don't have the fix for openssl on debian 9 (dotnet/corefx#19070), which means the build now fails on Debian 9 or Ubuntu 17.04 unless you explicitly add options to disable all managed parts of the build.
@ellismg how complicated would it be to move to tools version that already has that fix now?

ellismg:
Is `initHostDistroRid()` not needed anymore?  It looks like the variables it sets up are never consumed anymore.  Can we just delete it?

> @ellismg how complicated would it be to move to tools version that already has that fix now?

It should be as simple as updating the `DotNetCLIVersion.txt` file in the root of the repository to a dotnet that has the fix.

tmds:
@ellismg @janvorli I updated the PR to invert the logic. Instead of maintaining a list of supported, it now contains a list of unsupported RIDs.

janvorli:
@tmds I don't understand the motivation for inverting the polarity of the distro list. It made sense before, since the supported distros are the ones for which we have the tools package  on our servers. Flipping the polarity means that for the distros that are not listed there, we would fail to download the tools, since there is no such file for their RID.

tmds:
@janvorli I was assuming the same thing but then after @ellismg 's comment I had a closer look in the init-tools.sh script and saw it downloads the generic linux package.
The host rid in build.sh is only used to check if it can compile the managed bits (isMSBuildOnNETCoreSupported). So now I am changing that function so this defaults to true for unknown rids.

janvorli:
@tmds ah, I didn't know that the tools were already modified that way. Makes sense then.

tmds:
@ellismg if this is ok for you too, can you merge the PR?

Issue #11929, opened 2017-05-26
[RyuJIT/armel] CodeGenBringUpTests Assertion failed 


wateret:
A sub-issue for #11822 

Some tests fail with below assert failure.

```
               Assert failure(PID 17906 [0x000045f2], Thread: 17906 [0x45f2]): Assertion failed 'inArgMask & RBM_FLTARG_REGS' in 'BringUpTest:method_2I2D4S(int,int,double,double,struct,struct,struct,struct):int' (IL size 271)

                   File: /opt/code/github/hqueue/coreclr/src/jit/regalloc.cpp Line: 709
```

List of the tests fail with above message in `JIT/CodeGenBringUpTests`

```
FAILED   - [ 157]JIT/CodeGenBringUpTests/struct16args/struct16args.sh
```

alpencolt:
There is the same issue on `JIT/CodeGenBringUpTests/FPVar/FPVar.exe` on softfp mode.
I will look to it.

alpencolt:
Fixed by #12105 and #12005

PR #12005, opened 2017-05-31
[RyuJIT/armel] Support putting floating-point args

wateret:
- Make Lowering and LSRA be aware of armel argument push convention.
- Implement codegen for GT_COPY that was newly created from `LowerArg()`.

Fix #11928

PR #12105, opened 2017-06-06
[RyuJIT/armel] Fix for correct passing and returning float values.

alpencolt:
Float pointing arguments should be passed using common purpose register
on soft mode. Returning value from function should be stored into r0.

Fix for  #11929
@wateret could you take a look to this change?
+@dotnet/arm32-contrib

### DIFF
--- a/src/jit/codegenarm.cpp
+++ b/src/jit/codegenarm.cpp
@@ -308,7 +308,20 @@ void CodeGen::genReturn(GenTreePtr treeNode)
             genConsumeReg(op1);
 
             regNumber retReg = varTypeIsFloating(treeNode) ? REG_FLOATRET : REG_INTRET;
-            if (op1->gtRegNum != retReg)
+            if (varTypeIsFloating(treeNode) && (compiler->opts.compUseSoftFP || compiler->info.compIsVarArgs))
+            {
+                if (targetType == TYP_FLOAT)
+                {
+                    getEmitter()->emitIns_R_R(INS_vmov_f2i, EA_4BYTE, REG_INTRET, op1->gtRegNum);
+                }
+                else
+                {
+                    assert(targetType == TYP_DOUBLE);
+                    getEmitter()->emitIns_R_R_R(INS_vmov_d2i, EA_8BYTE, REG_INTRET, REG_NEXT(REG_INTRET),
+                                                op1->gtRegNum);
+                }
+            }
+            else if (op1->gtRegNum != retReg)
             {
                 inst_RV_RV(ins_Move_Extend(targetType, true), retReg, op1->gtRegNum, targetType);
             }


src/jit/codegenarm.cpp
```diff
@@ -308,7 +308,11 @@ void CodeGen::genReturn(GenTreePtr treeNode)
             genConsumeReg(op1);
 
             regNumber retReg = varTypeIsFloating(treeNode) ? REG_FLOATRET : REG_INTRET;
-            if (op1->gtRegNum != retReg)
+            if (varTypeIsFloating(treeNode) && compiler->opts.compUseSoftFP)
+            {
+                getEmitter()->emitIns_R_R(INS_vmov_f2i, EA_4BYTE, REG_INTRET, op1->gtRegNum);
+            }
+            else if (op1->gtRegNum != retReg)
             {
```
BruceForstall: It seems like this might not be sufficient. E.g., registerfp.cpp for the `LEGACY_BACKEND` case has:
```
            if (compiler->info.compIsVarArgs || compiler->opts.compUseSoftFP)
            {
                if (tree->TypeGet() == TYP_FLOAT)
                {
                    inst_RV_RV(INS_vmov_f2i, REG_INTRET, REG_FLOATRET, TYP_FLOAT, EA_4BYTE);
                }
                else
                {
                    assert(tree->TypeGet() == TYP_DOUBLE);
                    inst_RV_RV_RV(INS_vmov_d2i, REG_INTRET, REG_NEXT(REG_INTRET), REG_FLOATRET, EA_8BYTE);
                }
            }
```
so, it handles both float and double, and also considers the varargs case (which I guess we don't have for .NET Core, but doesn't hurt to check anyway, in case we need to implement it at some point)

alpencolt: Right, I will fix it.

src/jit/codegencommon.cpp
```diff
@@ -4282,7 +4282,14 @@ void CodeGen::genFnPrologCalleeRegArgs(regNumber xtraReg, bool* pXtraRegClobbere
...
             // Bingo - add it to our table
-            regArgNum = genMapRegNumToRegArgNum(varDsc->lvArgReg, regType);
+            if (!compiler->opts.compUseSoftFP)
+            {
+                regArgNum = genMapRegNumToRegArgNum(varDsc->lvArgReg, regType);
+            }
+            else
+            {
+                regArgNum = genMapIntRegNumToRegArgNum(varDsc->lvArgReg);
+            }
```
BruceForstall: Can you explain why this is necessary? In particular, this function (genFnPrologCalleeRegArgs) is only called if `regState->rsCalleeRegArgMaskLiveIn` is non-empty. For the `compUseSoftFP` case, shouldn't the floating-point register file have an empty set for `regState->rsCalleeRegArgMaskLiveIn`?

src/jit/lsra.cpp
```diff
@@ -4301,7 +4301,7 @@ void LinearScan::updateRegStateForArg(LclVarDsc* argDsc)
             isFloat = true;
         }
 
-        if (isFloat)
+        if (isFloat && !compiler->opts.compUseSoftFP)
         {
```
BruceForstall: Can't you hoist this into the isFloat definition, e.g.:
```
        bool isFloat = isFloatRegType(argDsc->lvType)
                        && !compiler->opts.compUseSoftFP
#ifndef _TARGET_AMD64_
                        && !compiler->info.compIsVarArgs
#endif
                        ;
```

I also presume that with compUseSoftFP, lvIsHfaRegArg() should always be false.

src/jit/codegencommon.cpp
alpencolt: `regState->rsCalleeRegArgMaskLiveIn` isn't empty for floating-point register in `compUseSoftFP` mode. And  compiler crashes inside `genMapRegNumToRegArgNum` call.
So if it should be empty bug in another place.

src/jit/regalloc.cpp
```diff
@@ -706,7 +706,10 @@ regNumber Compiler::raUpdateRegStateForArg(RegState* regState, LclVarDsc* argDsc
 
     if (regState->rsIsFloat)
     {
-        noway_assert(inArgMask & RBM_FLTARG_REGS);
+        if (!opts.compUseSoftFP)
+        {
+            noway_assert(inArgMask & RBM_FLTARG_REGS);
```
BruceForstall: This doesn't seem right. It seems like the caller should never call this function when both `regState->rsIsFloat` and `opts.compUseSoftFP` are true.

src/jit/codegencommon.cpp
BruceForstall: Right. For SoftFP, the function has no incoming floating-point register arguments, since all FP arguments are passed as integer arguments (as if using the varargs ABI). So it seems that the mask of incoming FP argument registers (namely, `regState->rsCalleeRegArgMaskLiveIn`) should be empty.

src/jit/lsra.cpp
wateret: @BruceForstall Yes I'm sure that with compUseSoftFP, lvIsHfaRegArg() is always false since FEATURE_HFA is disabled for ARM_SOFTFP.

wateret:
@alpencolt It looks like these changes are only for callee side, aren't they? You may have already noticed, I am working on passing float arguments in #12005.

alpencolt:
@wateret I saw #12005 that's why I mentioned you. We can merge changes if they do the same but different ways. This pull request is mostly for fixing tests reported in #11929.

Changes in `codegenarm.cpp` for caller side (we take value from `r0` and put to `sN`).

src/jit/codegencommon.cpp
alpencolt: This change was unnecessary. Only changes in `lsra.cpp` were needed.
Please review updated commit.

alpencolt:
Rebased and fixed according to #12005.

PR #12113, opened 2017-06-06
Stop extraneously setting tail call stress

JosephTremoulet:
Tests GitHub_10780 and GitHub_11689 are repro cases for issues that don't
have anything to do with tail calls; the bit of code in their csproj's
that sets COMPlus_TailCallStress was an accidental copy-paste inclusion
from GitHub_11408.  Remove that extra code.

### DIFF
--- a/tests/src/JIT/Regression/JitBlue/GitHub_10780/GitHub_10780.csproj
+++ b/tests/src/JIT/Regression/JitBlue/GitHub_10780/GitHub_10780.csproj
@@ -34,19 +34,6 @@
   <ItemGroup>
     <Compile Include="GitHub_10780.cs" />
   </ItemGroup>
-  <PropertyGroup>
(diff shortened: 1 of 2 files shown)


JosephTremoulet:
@pgavlin PTAL.  This is follow-up from discussion in #12091.  I searched but didn't find extraneous tail call stress in any other tests.

pgavlin:
Does 11408 contain the clause to disable tailcall stress if ZapDisable is set?

JosephTremoulet:
> Does 11408 contain the clause to disable tailcall stress if ZapDisable is set?

No, but 11408 is [excluded](https://github.com/dotnet/coreclr/blob/73287a13f4028835182126b93f3935b6e99cb0aa/tests/issues.targets#L166-L168) entirely (and [flagged](https://github.com/dotnet/coreclr/blob/73287a13f4028835182126b93f3935b6e99cb0aa/tests/testsFailingOutsideWindows.txt#L60) as failing outside windows).

PR #3888, opened 2016-03-23
Remove special casing of platform assemblies

jkotas:

PR #12117, opened 2017-06-06
Restore logic for when to use satellite resource lookup under AppX

jkotas:
Remove unused cache of Assembly flags while I was on it

### DIFF
--- a/src/mscorlib/src/System/HResults.cs
+++ b/src/mscorlib/src/System/HResults.cs
@@ -71,7 +71,6 @@ namespace System
         internal const int COR_E_INVALIDOPERATION = unchecked((int)0x80131509);
         internal const int COR_E_INVALIDPROGRAM = unchecked((int)0x8013153a);
         internal const int COR_E_KEYNOTFOUND = unchecked((int)0x80131577);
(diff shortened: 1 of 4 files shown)


tarekgh:
LGTM.

tarekgh:
@jkotas do we need to include this for v2? I am not sure of that as this is UWP issue so would be OK to have the fix in UWP6. right?

jkotas:
I think so

tarekgh:
https://github.com/dotnet/coreclr/pull/3888

jkotas:
@dotnet-bot test this please

danmosemsft:
@safern could this be the cause of the resourcel oad failures that just started happening in the coreclr appx runs?

jkotas:
If you are not doing the PRI resource reprocessing for tests that happens for regular AppX apps, this will certainly break them.

You can workaround it by including the tests in `PLATFORM_RESOURCE_ROOTS`. Or if you need a quick fix, change `bool fUseSatelliteAssemblyResourceLookupUnderAppX` to be unconditionally true like what it was before this fix. I will reintroduce the bug that this change fixed though.

PR #12122, opened 2017-06-06
Remove use of gtEffectiveVal in RyuJIT codegen

BruceForstall:
This is not required in RyuJIT at the codegen phase: COMMA and
value-producing NOPs are no longer in place.

### DIFF
--- a/src/jit/codegenarm.cpp
+++ b/src/jit/codegenarm.cpp
@@ -1054,7 +1054,7 @@ void CodeGen::genCodeForStoreLclFld(GenTreeLclFld* tree)
     // Ensure that lclVar nodes are typed correctly.
     assert(!varDsc->lvNormalizeOnStore() || targetType == genActualType(varDsc->TypeGet()));
 
-    GenTreePtr  data = tree->gtOp1->gtEffectiveVal();
+    GenTreePtr  data = tree->gtOp1;
     instruction ins  = ins_Store(targetType);
     emitAttr    attr = emitTypeSize(targetType);
     if (data->isContainedIntOrIImmed())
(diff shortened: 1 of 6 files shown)


BruceForstall:

BruceForstall:

russellhadley:
Is there a checker in place that asserts this?  Or an assert in the accessor?  (I want to make sure that this just isn't an implicit convention that bakes in fragility if the code changes)  Also it kind of seems as we diverge the rules of the IR between Tree IR and LIR that we should have some way to check what state we're in so we can assert different things.  Might be worth raising the priority on this given it can get lost in a long tail of little clean up changes.

BruceForstall:
> Is there a checker in place that asserts this?

no.

> Also it kind of seems as we diverge the rules of the IR between Tree IR and LIR that we should have some way to check what state we're in so we can assert different things. 

We do have `BasicBlock::IsLIR()` and `GenTree::IsLIR()`

BruceForstall:
@pgavlin PTAL
cc @dotnet/jit-contrib

mikedn:
> Is there a checker in place that asserts this?
>> no.

AFAIR `GT_COMMA` triggers an `unreached` in `TreeNodeInfoInit` and an `assert` in `genCodeForTreeNode`.

pgavlin:
> > Is there a checker in place that asserts this?
> no.

There is actually a checker that asserts this: `LIR::Range::CheckLIR` ensures that all nodes in the range being checked are LIR nodes, and commas and value-producing NOPs are not LIR nodes.

russellhadley:
@pgavlin, Perfect.  That's more what I thought.  Is it worth inserting a check of this at codegen time to ensure that no utility got called out of turn and added something?  I think this code tops out in genCodeForBBList  - the check might go there.   There still is the annoying meta point of knowing when you have to call gtEffectiveVal or not.

pgavlin:
> Is it worth inserting a check of this at codegen time to ensure that no utility got called out of turn and added something? I think this code tops out in genCodeForBBList - the check might go there.

Yeah, we could run the check before generating code for each basic block. That would be a trivial change to make.

russellhadley:
@BruceForstall can you open the issue for the checker?

CarolEidt:
> @BruceForstall can you open the issue for the checker?

So that @pgavlin can close it with #12138?

russellhadley:
Ahh.  Sorry. Missed the email.  Thanks.

PR #12124, opened 2017-06-07
Simplify genRegArgNext()

BruceForstall:
There is no need to check the register type before calling REG_NEXT.

### DIFF
--- a/src/jit/regset.cpp
+++ b/src/jit/regset.cpp
@@ -3572,39 +3572,30 @@ regNumber genRegArgNext(regNumber argReg)
 {
     assert(isValidIntArgReg(argReg) || isValidFloatArgReg(argReg));
 
-    regNumber result = REG_NA;
+    switch (argReg)
+    {
 
 #ifdef _TARGET_AMD64_
 #ifdef UNIX_AMD64_ABI
-    // Windows X64 ABI:
-    //     REG_EDI, REG_ESI, REG_ECX, REG_EDX, REG_R8, REG_R9
-    //
-    if (argReg == REG_ARG_1) // REG_ESI
-    {
-        result = REG_ARG_2; // REG_ECX
-    }
-    else if (argReg == REG_ARG_3) // REG_EDX
-    {
-        result = REG_ARG_4; // REG_R8
-    }
-    else
-#else  // Windows ABI
-    // Windows X64 ABI:
-    //     REG_ECX, REG_EDX, REG_R8, REG_R9
-    //
-    if (argReg == REG_ARG_1) // REG_EDX
-    {
-        result = REG_ARG_2; // REG_R8
-    }
-    else
-#endif // UNIX or Windows ABI
-#endif // _TARGET_AMD64_
-    {
-        // Otherwise, we just iterate 'argReg' using REG_NEXT.
-        result = REG_NEXT(argReg);
-    }
 
-    return result;
+        // Linux x64 ABI: REG_EDI, REG_ESI, REG_ECX, REG_EDX, REG_R8, REG_R9
+        case REG_ARG_1:       // REG_ESI
+            return REG_ARG_2; // REG_ECX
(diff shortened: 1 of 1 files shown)


BruceForstall:

BruceForstall:
@CarolEidt PTAL
cc @dotnet/jit-contrib

src/jit/regset.cpp
```diff
@@ -3570,49 +3570,38 @@ bool genIsProperRegPair(regPairNo regPair)
...
 regNumber genRegArgNext(regNumber argReg)
 {
+    assert(isValidIntArgReg(argReg) || isValidFloatArgReg(argReg));
+
     regNumber result = REG_NA;
 
-    if (isValidFloatArgReg(argReg))
+#ifdef _TARGET_AMD64_
+#ifdef UNIX_AMD64_ABI
+    // Windows X64 ABI:
```
pgavlin: Might as well correct this comment while you're here...

src/jit/regset.cpp
```diff
@@ -3570,49 +3570,38 @@ bool genIsProperRegPair(regPairNo regPair)
...
-            result = REG_ARG_2; // REG_R8
-        }
+    // Windows X64 ABI:
+    //     REG_ECX, REG_EDX, REG_R8, REG_R9
+    //
+    if (argReg == REG_ARG_1) // REG_EDX
+    {
+        result = REG_ARG_2; // REG_R8
+    }
+    else
```
pgavlin: Please do not split the `else` statement from its block using `#ifdef`. It seems like this would be better off rewritten as a switch in any case:

```
switch (argReg)
{
#ifdef _TARGET_AMD64_
#ifdef UNIX_AMD64_ABI
    case REG_ARG_1:
        return REG_ARG_2;

    case REG_ARG_3:
        return REG_ARG_4;
#else
    case REG_ARG_1:
        return REG_ARG_2;
#endif

    default:
        return REG_NEXT(argReg);
}
```

BruceForstall:
Excellent ideas; I rewrote using `switch`.

BruceForstall:

src/jit/regset.cpp
```diff
@@ -3570,52 +3570,32 @@ bool genIsProperRegPair(regPairNo regPair)
...
-        // Windows X64 ABI:
-        //     REG_ECX, REG_EDX, REG_R8, REG_R9
-        //
-        if (argReg == REG_ARG_1) // REG_EDX
-        {
-            result = REG_ARG_2; // REG_R8
-        }
-#endif // UNIX or Windows ABI
+
+        // Linux x64 ABI: REG_EDI, REG_ESI, REG_ECX, REG_EDX, REG_R8, REG_R9
```
janvorli: @BruceForstall The order here is incorrect. The ECX and EDX should be swapped.

janvorli: And should is say RCX / RDX, ... instead?

BruceForstall: Interesting; it was completely broken for Linux x64. But it looks like the JIT never uses this function for x86/x64; it's only used for arm/arm64. Still, I fixed the Linux x64 case.

BruceForstall:
x64 release pri 1 failing with:
```
11:25:26 D:\j\workspace\x64_release_w---0575cb46\Tools\dotnetcli\sdk\2.0.0-preview1-005724\NuGet.targets(97,5): error : Unable to resolve 'runtime.win7-x64.Microsoft.NETCore.CoreDisTools (>= 1.0.1-prerelease-00002)' for '.NETCoreApp,Version=v2.0 (win7-x64)'. [D:\j\workspace\x64_release_w---0575cb46\tests\src\Common\test_dependencies\test_dependencies.csproj] [D:\j\workspace\x64_release_w---0575cb46\tests\build.proj]
```
https://ci.dot.net/job/dotnet_coreclr/job/master/job/x64_release_windows_nt_pri1_prtest/7633/

Issue #12098, opened 2017-06-06
lldb-3.9 cannot load libsosplugin.so


micli:
## Environment
dotnet --info
.NET Command Line Tools (2.0.0-preview1-005867)

Product Information:
 Version:            2.0.0-preview1-005867
 Commit SHA-1 hash:  37d826d763

Runtime Environment:
 OS Name:     debian
 OS Version:  8
 OS Platform: Linux
 RID:         debian.8-x64
 Base Path:   /opt/dotnet/sdk/2.0.0-preview1-005867/

Microsoft .NET Core Shared Framework Host
Version  : 2.0.0-preview1-002091-00
Build    : fa62d01e00eaec93f228548f2ee00cd81c148066

lldb-3.9 which was installed from apt.llvm.org

#### using offical .NET Core 2.0 Preview
cd /opt/dotnet/shared/Microsoft.NETCore.App/2.0.0-preview1-002091-00
lldb-3.9
(lldb) plugin load libsosplugin.so
**error: this file does not represent a loadable dylib**

#### using custom build .NET Core 2.0 source code clone from github.com/dotnet/coreclr
cd /home/micl/dotnet/coreclr/bin/Product/Linux.x64.Debug
lldb-3.9
(lldb) plugin load libsosplugin.so
**Segmentation fault**

Neither of them was not loaded correctly.

janvorli:
@micli that's expected. The lldb plugin that's present in the package is for lldb 3.6 and it cannot work with other versions since it is statically linked to the lldb libraries of that specific version. You'll need to build coreclr yourself with lldb-3.9 dev package installed to get plugin that works with that.

stephentoub:
@janvorli, I actually get the same error with lldb-3.6 on my Ubuntu 16.10 VM.

janvorli:
I have looked into that and the issue is that the libsosplugin.so in our portable package is not portable itself due to the different SONAME of liblldb.so on Fedora based distros and others. Since out portable build is built on Fedora based distro, the plugin cannot resolve the dependency.
I am trying to figure out if we can make it portable.
CC: @gkhanna79, @Petermarcu

gkhanna79:
Assigning to @mikem8361 who owns SOS.

CC @lt72

janvorli:
[...]

PR #12125, opened 2017-06-07
Fix portable build sos plugin problems

mikem8361:
Removing the explicit reference to liblldb. Since the lldb program has already loaded this lib, our will now load regardless of the distro and version of lldb.

Issue #12098.

### DIFF
--- a/src/ToolBox/SOS/lldbplugin/CMakeLists.txt
+++ b/src/ToolBox/SOS/lldbplugin/CMakeLists.txt
@@ -54,21 +54,6 @@ if(NOT ENABLE_LLDBPLUGIN)
     return()
 endif()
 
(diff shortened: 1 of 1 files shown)


ellismg:
By my reading (and I could be wrong) if we are on Linux and LLDB isn't installed, we won't error out now (where as we would previously).  Is that the correct behavior?

mikem8361:
I'll check to sure but if the lldb dev package isn't install we should still fail because LLDB.h won't be found (except on OSX and arm).

ellismg:
> I'll check to sure but if the lldb dev package isn't install we should still fail because LLDB.h won't be found (except on OSX and arm).

Ahh, that makes sense.

PR #12140, opened 2017-06-07
Enable CPU Group on Linux AMD64

helloguo:
This PR enables CPU Groups on Linux AMD64 when `COMPlus_GCNumaAware` and `COMPlus_GCCpuGroup` are set to 1. It fixes the low CPU utilization mentioned in https://github.com/dotnet/coreclr/issues/12141

### DIFF
--- a/src/utilcode/util.cpp
+++ b/src/utilcode/util.cpp
@@ -1288,7 +1288,15 @@ int GetCurrentProcessCpuCount()
     // the case where there are more than 64 processors, so we will return a
     // maximum of 64 here.
     if (count == 0 || count > 64)
+#if defined(_TARGET_AMD64_) && defined(FEATURE_PAL)
+    {
+        count = GetCurrentProcessActiveCpuCount(GetCurrentProcess());
+        if (count == 0 || count > 1024)
+            count = 1024;
+    }
+#else
         count = 64;
+#endif
 
     cCPUs = count;
             
--- a/src/utilcode/util.cpp
+++ b/src/utilcode/util.cpp
@@ -1266,11 +1266,6 @@ int GetCurrentProcessCpuCount()
     if (cCPUs != 0)
         return cCPUs;
 
-#if defined(_TARGET_AMD64_) && defined(FEATURE_PAL)
-    int count = GetCurrentProcessActiveCpuCount(GetCurrentProcess());
-    if (count == 0 || count > 1024)
-        count = 1024;
-#else
     DWORD_PTR pmask, smask;
 
     if (!GetProcessAffinityMask(GetCurrentProcess(), &pmask, &smask))
@@ -1294,7 +1289,6 @@ int GetCurrentProcessCpuCount()
     // maximum of 64 here.
     if (count == 0 || count > 64)
         count = 64;
-#endif
 
     cCPUs = count;
             
--- a/src/utilcode/util.cpp
+++ b/src/utilcode/util.cpp
@@ -1266,6 +1266,11 @@ int GetCurrentProcessCpuCount()
     if (cCPUs != 0)
         return cCPUs;
 
+#if defined(_TARGET_AMD64_) && defined(FEATURE_PAL)
+    int count = GetCurrentProcessActiveCpuCount(GetCurrentProcess());
+    if (count == 0 || count > 1024)
+        count = 1024;
+#else
     DWORD_PTR pmask, smask;
 
     if (!GetProcessAffinityMask(GetCurrentProcess(), &pmask, &smask))
@@ -1288,13 +1293,6 @@ int GetCurrentProcessCpuCount()
     // the case where there are more than 64 processors, so we will return a
     // maximum of 64 here.
     if (count == 0 || count > 64)
-#if defined(_TARGET_AMD64_) && defined(FEATURE_PAL)
-    {
-        count = GetCurrentProcessActiveCpuCount(GetCurrentProcess());
-        if (count == 0 || count > 1024)
-            count = 1024;
-    }
-#else
         count = 64;
 #endif
 
--- a/src/dlls/mscoree/mscorwks_unixexports.src
+++ b/src/dlls/mscoree/mscorwks_unixexports.src
@@ -104,6 +104,10 @@ GlobalMemoryStatusEx
 VirtualQuery
 WideCharToMultiByte
 WriteFile
+GetLogicalProcessorInformationEx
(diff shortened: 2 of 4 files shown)


helloguo:
@janvorli PTAL

src/utilcode/util.cpp
```diff
@@ -1288,7 +1288,15 @@ int GetCurrentProcessCpuCount()
     // the case where there are more than 64 processors, so we will return a
```
janvorli: @helloguo please see the comment here: 
The GC also cannot (currently) handle the case where there are more than 64 processors, so we will return amaximum of 64 here. So the limitation is intentional and is the same for Windows and Unix.

janvorli: CC @Maoni0 - or is this comment outdated now and we can support more? Should we fix the windows code in that case too?

janvorli: @Maoni0 the GetCurrentProcessCpuCount is called only when the CPU groups are not enabled. And this function returns max 64 on Windows too. I think that without CPU groups enabled, the 64 is still the limit for this specific function. For CPU groups, GC code uses CPUGroupInfo::GetNumActiveProcessors() instead to get the cpu count. That will work for Unix too later.
I have recently added NUMA support for Unix, including the CPU groups, but it is not enabled yet since I didn't have enough time to fully test it in the 2.0 timeframe, especially on machines with a lot of processors. I have also decided later to make it runtime optional based on whether the libnuma is installed or not.

Maoni0: We are saying the same thing - without enabling CPU groups you always get 64 CPUs max right now, on Windows or not.

It sounds like you said you already implemented CPU groups on Linux even though Linux probably doesn't have the concept of CPU groups. So if we enabled your code, it would work out the same way on Windows and on Linux. Thus no need for this PR. Is that correct?

helloguo:
@janvorli @Maoni0 Thanks for your input. One question: Let's say one Linux machine has only one socket but this socket has more than 64 logical cores. Will the the CPU group (let's assume we have CPU group on Linux) be enabled? If it is not enabled and `GetCurrentProcessCpuCount` is called, what should the the return value of `GetCurrentProcessCpuCount`?

Maoni0:
There are 2 approaches - I can see us picking either one of them:

1) make it work the same way on Windows and on Linux, as in even though Linux has no processor groups (which is really just something that Windows did to compensate for the 64-bit number it used to represent the mask....), we implement it ourselves so the code that looks at # of processors and etc looks the same on Linux and Windows. It sounds like this is what Jan did - he can verify.

Of course we can look at this from the opposite way which is we make GetCurrentProcessCpuCount always return the true # of processors on Windows and Linux so the CPU group code is completely transparent to the callers in GC (and threadpool). 

I am open to making the CPUGroup config by default enabled. It seems like a reasonable assumption that customers with >64 processor machines usually want to take advantage of all processors instead of having to specify a configs to achieve that.

2) On Linux we don't require the user to enable the CPUGroup configs and GetCurrentProcessCpuCount  returns the true number of processors, which is what you did. We need to make sure its callers are handling it properly. For example, we need to make sure something we are doing on Windows that cover > 64 cases is also done when GetCurrentProcessCpuCount returns > 64 on Linux.

janvorli:
@helloguo the reason for creating the CPU groups on Linux is that GC and few other things in the runtime work with this model and in CoreCLR, GC runs on top of the runtime that is written as if there were windows below and the PAL emulates the necessary functions. We diverge from this only in cases where emulating the Windows functionality would not be possible or it would cause perf and other issues. One such example is the runtime suspension where we use signals on Unix and SuspendThread on Windows. We used to simulate SuspendThread using signals, but it was too clunky and error prone, so we have moved away from such model.
The CPU groups don't seem to be such a case, it is all about organizing the cpus in certain ways. If we got rid of that and exposed a different model based e.g. purely on NUMA nodes, then the code in GC and runtime would have to be split for Unix and Windows.

helloguo:
@janvorli @Maoni0 Thanks for explanation. It's good to know that Jan has started CPUGroups work on Linux. When is that going to be enabled for Linux? I can help for testing on my machine if there is a need.

janvorli:
@helloguo if you can test it, it would be awesome. Basically, it is code complete and all merged in. To enable that, you need to have numa-dev library installed when building the project and then set two env variables before launching the application. These are:
```
COMPlus_GCNumaAware
COMPlus_GCCpuGroup 
```
Both should be set to 1.

helloguo:
@janvorli After setting the two env variables, it doesn't work on my machine. I trace it down a little bit. `InitCPUGroupInfo` https://github.com/dotnet/coreclr/blob/master/src/utilcode/util.cpp#L1030 does not work probably.  `InitCPUGroupInfoAPI` https://github.com/dotnet/coreclr/blob/master/src/utilcode/util.cpp#L1047 returns false so that `InitCPUGroupInfo `returns before setting `m_enableGCCPUGroups`. `InitCPUGroupInfoAPI` https://github.com/dotnet/coreclr/blob/master/src/utilcode/util.cpp#L861 returns false because `m_pGetLogicalProcessorInformationEx`  https://github.com/dotnet/coreclr/blob/master/src/utilcode/util.cpp#L879 is `NULL `, which suggests `GetProcAddress `does not find this method in PAL layer.

I'm working on the fix right now. Any suggestion to move forward?

janvorli:
@helloguo Have you rebuilt coreclr with the libnuma devel package installed? I have tested that on my local box with 4 numa nodes and it was working.

helloguo:
@janvorli  I think so. But I will clean my build and give it another try.

janvorli:
You'll need to delete the bin folder completely. CMake would not pick the numa presence otherwise

helloguo:
@janvorli The reason it does not work on my machine is that `CPUGroupInfo::InitCPUGroupInfoAPI()` return false. `GetProcAddress` cannot find `GetLogicalProcessorInformationEx`, `SetThreadGroupAffinity`, `GetThreadGroupAffinity` and `GetCurrentProcessorNumberEx`. So I add these methods in mscorwks_unixexports.src. Along with setting `COMPlus_GCNumaAware` and `COMPlus_GCCpuGroup`, the CPUGroups on Linux works for me now.

janvorli:
@helloguo thank you for fixing that. Since I was stepping through all of the numa code in runtime and PAL when I was working on the support, I must have just forgotten to add changes in the mscorwks_unixexports.src when I was committing it.

PR #12142, opened 2017-06-07
Convert some very common noway_assert to simple assert

BruceForstall:
With these few changes, I measured a JIT instruction count reduction
of 0.37% of SuperPMI over the tests, and 0.17% for MinOpts.

Related to #10421

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -886,8 +886,8 @@ void Compiler::fgRemoveReturnBlock(BasicBlock* block)
 
 flowList* Compiler::fgGetPredForBlock(BasicBlock* block, BasicBlock* blockPred)
 {
-    noway_assert(block);
-    noway_assert(blockPred);
+    assert(block);
+    assert(blockPred);
     assert(!fgCheapPredsValid);
 
     flowList* pred;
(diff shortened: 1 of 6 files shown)


BruceForstall:
@briansull PTAL
cc @dotnet/jit-contrib

src/jit/flowgraph.cpp
```diff
@@ -886,8 +886,8 @@ void Compiler::fgRemoveReturnBlock(BasicBlock* block)
 
 flowList* Compiler::fgGetPredForBlock(BasicBlock* block, BasicBlock* blockPred)
 {
-    noway_assert(block);
-    noway_assert(blockPred);
+    assert(block);
```
pgavlin: Nit: might as well fix these to `asssert(block != nullptr)` etc. while you're here. Non-blocking, though.

src/jit/morph.cpp
```diff
@@ -14873,8 +14872,8 @@ GenTreePtr Compiler::fgMorphToEmulatedFP(GenTreePtr tree)
 
 GenTreePtr Compiler::fgMorphTree(GenTreePtr tree, MorphAddrContext* mac)
 {
-    noway_assert(tree);
-    noway_assert(tree->gtOper != GT_STMT);
+    assert(tree);
```
pgavlin: Ditto

src/jit/flowgraph.cpp
BruceForstall: For this change I was focused on just deleting 6 characters per line :-)

BruceForstall:

PR #12144, opened 2017-06-07
delete BLOCKSET_INIT_NOCOPY and BLOCKSET_INIT.

sandreenko:
BLOCKSET_INIT was unused.
BLOCKSET_INIT_NOCOPY was replaced by the implicit constructor.


sandreenko:
PTAL @dotnet/jit-contrib . 
The same change as was merged for bitsets.

PR #12145, opened 2017-06-07
Modify iteration behavior some benchmarks

AndyAyersMS:
This change updates the iteration behavior for the following benchmarks, so that
they generally run for about 1 second per iteration:
* Inline/InlineGCStruct
* Inline/NoThrowInline
* Span/IndexerBench/Indexer4
* Functions/MathTests/*

It also removes the length 1, 10, and 1000 variants of the Span/SpanBench
tests, leaving just the length 100 variants. Analysis has shown that there
is no need to test varying lengths here. Iteration counts for these tests
have been left as is and will be fixed subsequently.

Partially addresses #11654.


AndyAyersMS:
@JosephTremoulet PTAL
cc @dotnet/jit-contrib

PR #12150, opened 2017-06-07
[release/2.0.0] Pick up some infrastructure fixes from master

ellismg:


compileoptions.cmake
```diff
@@ -23,8 +23,10 @@ if (CLR_CMAKE_PLATFORM_UNIX)
   # after hitting just about 20 errors.
   add_compile_options(-ferror-limit=4096)
 
-  # All warnings that are not explicitly disabled are reported as errors
-  add_compile_options(-Werror)
+  if (CLR_CMAKE_WARNINGS_ARE_ERRORS)
```
gkhanna79: I think the define is considered defined for CMake whether you set its value to OFF or ON. You will need to have a more specific check here.

gkhanna79: @janvorli PTAL as well.

janvorli: @gkhanna79 this "if" doesn't check if the symbol is defined, but rather its value. Condition is true if the variable is 1, ON, YES, TRUE, Y, or a non-zero number. False if the variable is 0, OFF, NO, FALSE, N, IGNORE, NOTFOUND, the empty string, or ends in the suffix -NOTFOUND

gkhanna79: Thanks for correcting me @janvorli.

CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ if(CORECLR_SET_RPATH)
 endif(CORECLR_SET_RPATH)
 
 OPTION(CMAKE_ENABLE_CODE_COVERAGE "Enable code coverage" OFF)
+OPTION(CLR_CMAKE_WARNINGS_ARE_ERRORS "Warnings are errors" ON)
```
janvorli: @ellismg there is one additional change #12106 that modifies the CMAKE_ENABLE_CODE_COVERAGE to CLR_CMAKE_ENABLE_CODE_COVERAGE which would be nice to pick with this.

ellismg: @janvorli Thanks.  @omajid mentioned that as well, I'll pick that up as well.

ellismg:
Per Shiproom, we should take this after Preview 2 Escrow ends.  I've added `*NO MERGE*` for now and once the branch is open, I will merge.

ellismg:
@dotnet-bot test this please

ellismg:
This was previously approved and the branch is open now, so I'm merging it.

PR #12155, opened 2017-06-08
Add lclVar semantic checks to CheckLIR.

pgavlin:
Specifically, ensure that an unaliasable lclVar is not redefined between the
point at which a use appears in linear order and the point at which that
use is consumed by its user. This ensures that it is always safe to treat
a lclVar use as happening at the user (rather than at the lclVar node).

### DIFF
--- a/src/jit/lir.cpp
+++ b/src/jit/lir.cpp
@@ -1573,7 +1573,7 @@ bool LIR::Range::CheckLIR(Compiler* compiler, bool checkUnusedValues) const
             AliasSet::NodeInfo operandInfo(compiler, operand);
             if (operandInfo.IsLclVarRead())
             {
-                int count;
+                int        count;
                 const bool removed = unconsumedLclVarReads.TryRemove(operandInfo.LclNum(), &count);
                 assert(removed);
 
--- a/src/jit/lir.cpp
+++ b/src/jit/lir.cpp
@@ -4,6 +4,7 @@
 
 #include "jitpch.h"
 #include "smallhash.h"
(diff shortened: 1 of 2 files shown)


pgavlin:

pgavlin:
Note: I originally attempted to check a stronger property, namely that no possible redefinition of a lclVar V0 occurs between a read of that lclVar and the user of the read. "Possible redefinitions" included stores to any location that may alias V0 (including indirect stores to arbitrary addresses). Unfortunately, this property did not hold.

src/jit/lir.cpp
```diff
@@ -1556,6 +1557,47 @@ bool LIR::Range::CheckLIR(Compiler* compiler, bool checkUnusedValues) const
...
+        {
+            int count = 0;
+            unconsumedLclVarReads.TryGetValue(nodeInfo.LclNum(), &count);
+            unconsumedLclVarReads.AddOrUpdate(nodeInfo.LclNum(), count + 1);
+        }
+
+        if (nodeInfo.IsLclVarWrite())
+        {
+            int unused;
+            assert(!unconsumedLclVarReads.TryGetValue(nodeInfo.LclNum(), &unused));
```
CarolEidt: This code could use some comments in general, but especially at the assert, because it's non-obvious. If someone runs into this assert, it would be good to have a comment to help them understand, e.g. "This lclVar has just been written; there had better not be any unconsumed reads of that lclVar from previous nodes."

pgavlin: Good suggestion.

PR #12158, opened 2017-06-08
Added new crossgen flag /warnings

briansull:
We can now disable several annoying compilations warnings messages from crossgen.

With this checkin only the most common warning:

"Implicit boxing for calls to constrained methods not supported" is suppressed.

To see this warning you will now need to add the /warnings flags to crossgen.cpp

### DIFF
--- a/src/zap/zapinfo.cpp
+++ b/src/zap/zapinfo.cpp
@@ -2131,7 +2131,10 @@ void ZapInfo::getCallInfo(CORINFO_RESOLVED_TOKEN * pResolvedToken,
         if (pResult->thisTransform == CORINFO_BOX_THIS)
         {
             // READYTORUN: FUTURE: Optionally create boxing stub at runtime
-            m_zapper->Warning(W("ReadyToRun: Implicit boxing for calls to constrained methods not supported\n"));
+            if (doCompilationWarnings())
+            {
+                m_zapper->Warning(W("ReadyToRun: Implicit boxing for calls to constrained methods not supported\n"));
+            }
             ThrowHR(E_NOTIMPL);
         }
     }
--- a/src/inc/corcompile.h
+++ b/src/inc/corcompile.h
@@ -1825,4 +1825,11 @@ inline bool IsReadyToRunCompilation()
 #endif
 }
 
+extern bool g_fCompilationWarnings;
+
+inline bool doCompilationWarnings()
+{
+    return g_fCompilationWarnings;
+}
+
 #endif /* COR_COMPILE_H_ */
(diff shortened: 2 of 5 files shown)


briansull:
I have to add a fix for:

The /CreatePDB switch cannot be used with other switches, except /lines and the various path switches.

:-(

src/zap/zapinfo.cpp
```diff
@@ -2131,7 +2131,10 @@ void ZapInfo::getCallInfo(CORINFO_RESOLVED_TOKEN * pResolvedToken,
         if (pResult->thisTransform == CORINFO_BOX_THIS)
         {
             // READYTORUN: FUTURE: Optionally create boxing stub at runtime
-            m_zapper->Warning(W("ReadyToRun: Implicit boxing for calls to constrained methods not supported\n"));
+            if (doCompilationWarnings())
```
jkotas: Should this rather be in `Zapper::Print` where we filter out all warnings?

Or if you just want to get rid of this particular warning only - I would rather just delete it.

briansull: Warning just calls Print:

```
void Zapper::Warning(LPCWSTR format, ...)
{
    // Preserve last error so we can still do ThrowLastError() after printing the message
    DWORD err = GetLastError();

    va_list args;
    va_start(args, format);

    Print(CORZAP_LOGLEVEL_WARNING, format, args);

    va_end(args);

    SetLastError(err);
}

void Zapper::Info(LPCWSTR format, ...)
{
    va_list args;
    va_start(args, format);

    Print(CORZAP_LOGLEVEL_INFO, format, args);

    va_end(args);
}
```

jkotas: Print does not just Print. It filters based on `CorZapLogLevel`:

https://github.com/dotnet/coreclr/blob/master/src/zap/zapper.cpp#L1756

briansull: I see, I could disable every warning by making thischange in Print.
I did consider that.  
But I thought that I would just disable the less important one that we are hitting today and leave the other one.

briansull: If we delete this warning, that would mean to me, that we don't think that we should ever address this.  I'm not sure what the perf cost is for this item

jkotas: > If we delete this warning, that would mean to me, that we don't think that we should ever address this

The GitHub issues are the best way to track work items ... adding a comment line switch won't help us to track it.

If you would like to still have a way to see this warning somehow, you can change it to `CORZAP_LOGLEVEL_INFO` so that it gets printed with verbose output only. I just do not think it is worth it to have a new special command line switch for it.

briansull:
I will pursue the approach of simply delete the warning(s)

Issue #12080, opened 2017-06-05
[RyuJIT/ARM32] Assertion failed '!"Instruction cannot be encoded" by Ins_add


sjsinju:
```
JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b54611/b54611/b54611.sh
```

```
FAILED   - [5883][   6s]JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b54611/b54611/b54611.sh
               BEGIN EXECUTION
               /home/dotnet_jenkins/workspace/hqueue/Run_coreclr_tests_Checked_ryujit_nyi_all/runtime/tests/Tests/coreoverlay/corerun b54611.exe
               
               Assert failure(PID 6899 [0x00001af3], Thread: 6899 [0x1af3]): Assertion failed '!"Instruction cannot be encoded"' in 'ILGEN_0x142726a2:Method_0x6a72bf0e(long):int' (IL size 589)
               
                   File: /opt/code/github/hqueue/coreclr/src/jit/emitarm.cpp Line: 2451
                   Image: /home/dotnet_jenkins/workspace/hqueue/Run_coreclr_tests_Checked_ryujit_nyi_all/runtime/tests/Tests/coreoverlay/corerun
               
               ./b54611.sh: line 240:  6899 Aborted                 (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" $ExePath $CLRTestExecutionArguments
               Expected: 100
               Actual: 134
               END EXECUTION - FAILED
```

sjsinju:
Merged #12162

PR #12162, opened 2017-06-08
[RyuJIT/ARM32] Fix assertion when using invalid imm for INS_add

sjsinju:
#12080 

`JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b54611/b54611/b54611.sh`

Assertion failed '!"Instruction cannot be encoded" has been occurred.

INS_add couldn't process with the invalid imm of -273 on this test.
So in this PR, If it get the value that cannot be processed, moving to register that after changing status would be ran so that it can be processed.

The result is below
```
IN00b1:             mvn     r3, 272
IN00b2:             adds    r3, r3, r0
```

### DIFF
--- a/src/jit/emitarm.cpp
+++ b/src/jit/emitarm.cpp
@@ -7773,7 +7773,17 @@ regNumber emitter::emitInsTernary(instruction ins, emitAttr attr, GenTree* dst, 
 
     if (intConst != nullptr)
     {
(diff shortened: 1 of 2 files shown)


mikedn:
Doesn't that assert imply that `Lowering::IsContainableImmed` is somehow broken and needs fixing?

sjsinju:
@mikedn Yes. I think you are right. Let me check again. Thank you.

sjsinju:
@dotnet-bot  test Ubuntu arm Cross Release Build

BruceForstall:
@dotnet/arm32-contrib

PR #12079, opened 2017-06-05
[RyuJIT/ARM32] Fix assertion when struct argument passing less than 4 byte

hseok-oh:
Fix assertion when struct argument is less than 4 byte and promoted

Related issue: #11850 

cc/ @dotnet/arm32-contrib

PR #12163, opened 2017-06-08
[RyuJIT/ARM32] Fix comment and ifdef condition

hseok-oh:
- Add comment for #elif and #endif
- Modify to apply previous fix for RyuJIT/ARM32 only (not for LegacyJIT/ARM32)

Related PR: #12079


hseok-oh:
@BruceForstall PTAL

Issue #12154, opened 2017-06-07
Crossgen: Assertion failed 'varNum < comp->lvaCount' in 'Program:CallUnsafeOnCompleted(ref,byref)' (IL size 25)


MichalStrehovsky:
```csharp
internal class Program
{
    struct ConfiguredTaskAwaiter<T> : ICriticalNotifyCompletion
    {
        public void OnCompleted(Action continuation)
        {
            
        }

        public void UnsafeOnCompleted(Action continuation)
        {
        }
    }

    [MethodImpl(MethodImplOptions.NoInlining)]
    static void CallUnsafeOnCompleted<TAwaiter>(Action continuation, ref TAwaiter awaiter)
            where TAwaiter : ICriticalNotifyCompletion
    {
        try
        {
            awaiter.UnsafeOnCompleted(continuation);
        }
        catch (Exception)
        {
        }
    }

    private static void Main(string[] args)
    {
        ConfiguredTaskAwaiter<object> a;
        CallUnsafeOnCompleted(null, ref a);
    }
}
```

When crossgen'd with `COMPLUS_NGenDump=CallUnsafeOnCompleted` the above program asserts with the assert in bug title.

NGenDump for the asserting method: [test.txt](https://github.com/dotnet/coreclr/files/1059539/test.txt)

RussKeldorph:
@dotnet/jit-contrib

PR #12172, opened 2017-06-08
Don't dump an uninitialized bitset in LSRA.

pgavlin:
LSRA was unconditionally dumping `fpCalleeSaveCandidateVars`, which is
not initialized if we are not enregistering any lclVars. This was
causing an assertion with dumping enabled.


pgavlin:

Fixes #12154

pgavlin:
OS X failure is a widespread infrastructure issue.

PR #12173, opened 2017-06-08
[Local GC] No longer need to initialize SVR dacvars separately

swgillespie:
Now that all server dacvars go through the `GcDacVars` struct (https://github.com/dotnet/coreclr/pull/9255 - despite the name this also includes Server GC), we don't need to initialize them here. Furthermore, since there are no `DEFINE_DACVAR_SVR` macros in `dacvars.h` anymore, this function has no effect anyway today.


swgillespie:

swgillespie:
@noahfalk @mikem8361 PTAL? this should all be dead code.

swgillespie:
thanks for the reviews!

Issue #11914, opened 2017-05-25
Managed tests build fails with VS2017


JosephTremoulet:
Builds fail with:

> D:\Source\coreclr\Tools\Microsoft.CSharp.Core.targets(106,11): error MSB4064: The "OverrideToolHost" parameter is not s
> upported by the "Csc" task. Verify the parameter exists on the task, and it is a settable public instance property. [D:
> \Source\coreclr\tests\src\CoreMangLib\cti\system\reflection\emit\DynMethodJumpStubTests\DynMethodJumpStubTests.csproj]
> D:\Source\coreclr\Tools\Microsoft.CSharp.Core.targets(67,5): error MSB4063: The "Csc" task could not be initialized wit
> h its input parameters.  [D:\Source\coreclr\tests\src\CoreMangLib\cti\system\reflection\emit\DynMethodJumpStubTests\Dyn
MethodJumpStubTests.csproj]

which is the same error message from #10056 (which was fixedwith #10430).

Builds pass at 075a2d6f and fail at aad6903; this appears to be a result of [this change](https://github.com/dotnet/coreclr/pull/10532/files#diff-6178796b5428b89bb33ad7bf80143a79R40) from #10532.

JosephTremoulet:
@chcosta / @weshaggard can the same approach from #10430 be applied here, and what would that look like?

/cc @rartemev

ViktorHofer:
+1 happens for me all the time now...

JosephTremoulet:
> Builds pass at 075a2d6 and fail at aad6903; this appears to be a result of this change from #10532.

I just noticed that #10532 was reverted in #11790, including removing the `OverrideToolHost` line from `test\dir.props`, but that the build still fails with the same error at a16a8a852.  Reverting the change didn't fix things, which would seem to indicate either that #11790 was a partial revert or that something else in the interim also introduced `OverrideToolHost` somewhere.  @RussKeldorph, #11790 wasn't a partial revert, was it?

RussKeldorph:
@JosephTremoulet It was intended to be a complete revert.  There were no merge conflicts, so I assumed it was complete, but I confess I did not verify it line for line.

JosephTremoulet:
It looks like I may have somehow done the original bisecting incorrectly; I'm seeing inconsistent results trying to drill in further.  I'm running again now with full nuget cache clear and repo git clean between builds, will report back.  @ViktorHofer, if you have more info on when you do/don't see this, that could help.

[...]

PR #12175, opened 2017-06-08
Use roslyn common props in test build

wtgodbe:
Attempts to resolve https://github.com/dotnet/coreclr/issues/11914

@JosephTremoulet I haven't been able to repro that failure locally using my VS2017 command prompt, but I suspect this change might fix it - https://github.com/dotnet/coreclr/pull/10430, which you referenced in the issue, only applied to the product build, and not the test build. This change should bring in the same change to the test build. Can you try running a local build with these changes to see if the error goes away?

CC @ViktorHofer @briansull


wtgodbe:
@weshaggard can you take a look at the errors in the Windows test build? I don't know enough about compilers to diagnose what's happening here

JosephTremoulet:
FWIW, I've verified that on a machine with only VS2017 installed, which runs into the "TaskOverride" error message w/o this patch, applying this patch brings it to the same slew of csc errors showing up in the CI runs here.

briansull:
Any new status on this? @weshaggard @wtgodbe

weshaggard:
I'm going to spend some time digging into this today. I'm currently trying to get a local repro.

weshaggard:
@wtgodbe Your change is the correct fix but in doing so we are hitting compiler bug https://github.com/dotnet/roslyn/issues/14785. So we will either need to workaround that bug or pick-up a newer compiler with the fix. I suspect it will be easiest to workaround it for now. I'm going to see if I can do that.

weshaggard:
Assuming full CI is green we should merge this and file a tracking bug to undo the tests that were commented out. 

cc @pgavlin in case he has concerns with the commented out JIT tests.

pgavlin:
Please file an issue to track re-enabling these tests and reference that issue in a comment somewhere in each changed file (e.g. at the top of the file or near the first test that is commented out).

weshaggard:
I've filed https://github.com/dotnet/coreclr/issues/12233 to track the test changes and I will add a link to that in the commit for this PR before merging, just waiting to ensure CI is green before pushing that commit update.

wtgodbe:
CI is green, so I'll go ahead and merge. Thanks @weshaggard for getting this green!

briansull:
Note that after a sync of this change, I needed to do a full clean to get this working:

git clean -d -x -f

ViktorHofer:
Should this be moved into the release/2.0.0 branch? I just pulled it down and wanted to build it and faced them same issue again.

PR #12180, opened 2017-06-08
Pull Performance Run Dependencies from Release/2.0.0

brianrob:
Prior to this change, performance run dependencies (tests and corefx bits) were pulled from master.  This resulted in testing an improper configuration (release/2.0.0 CoreCLR with master perf tests and CoreFX bits).

This change pulls these dependencies from release/2.0.0.

### DIFF
--- a/tests/scripts/perf-prep.sh
+++ b/tests/scripts/perf-prep.sh
@@ -73,7 +73,7 @@ if [ $throughput -eq 1 ]; then
     cd _/fx
 
     # Checkout the specific commit we want
-    git checkout f0b9e238c08f62a1db90ef0378980ac771204d35
+    git checkout cbb9ee1c0f33803a8a71488efe910ca4da70dd6d
 
     # Build
     ./build.sh -release
--- a/tests/scripts/perf-prep.sh
+++ b/tests/scripts/perf-prep.sh
@@ -80,11 +80,11 @@ if [ $throughput -eq 1 ]; then
 else
     # Set up the copies
     # Coreclr build containing the tests and mscorlib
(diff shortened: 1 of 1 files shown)


jorive:
We should also change this line `git checkout f0b9e238c08f62a1db90ef0378980ac771204d35` on the TP runs.

brianrob:
Done.  This moves the throughput runs up to the latest CoreFX commit.

jorive:
Is it possible to get it from somewhere within the repo, without hardcoding it? Maybe someone in the infrastructure team will know.

brianrob:
It does not appear to be available within the repo.  I was planning to separate these two issues because I wanted to avoid hardcoding the commit hash.

Dependencies.props contains a commit hash in the versions repo that contains a set of packages to be consumed.  It does not contain any information about the commit hash.  In order to get the commit hash, we would likely need to download one of these packages and then fetch the commit hash out of it unless there is a better way.

@dagood, do you know if there is a way to get the commit hash of the last known good CoreFX commit that has been tested with CoreCLR?  So far, I've only found the list of packages in the versions repo.

dagood:
You're correct, commit hash isn't in dotnet/versions, and currently the best way to get it is to extract a package and look at version.txt. We definitely want to get commit hash into dotnet/versions, there just hasn't been time.

brianrob:
Thanks @dagood.  @jorive, I think we should hardcode the commit now to unblock TP runs, and then we should put a ref-count on getting the commit hash into dotnet/versions if we need it, or more likely just setup the layout for TP runs the same way that we do for other perf runs.  What do you think?

jorive:
@brianrob Makes sense :)

PR #12181, opened 2017-06-08
DateTime serialization fixed, added attention labels to serialized fields

ViktorHofer:
I went through all serializable types in coreclr again and double-checked all variables and typeforwards with netfx. I noticed that DateTime serialization was partially broken because in the ISerializable routine a different name was used. I fixed that and also added attention labels to not rename serialized fields. Also removed unnecessary ISerializable implementation in Comparer.

**CultureAwareComparer serialization is partially broken because implementation is different.**
Netfx: https://referencesource.microsoft.com/#mscorlib/system/stringcomparer.cs,137 
Core: https://github.com/dotnet/coreclr/blob/master/src/mscorlib/shared/System/StringComparer.cs#L178

Should we just ignore it or write a custom deserialization logic with ISerializable? The mapping would be easy from the Boolean value to the enum value (1 = IgnoreCase, 0 = None).

cc @stephentoub @danmosemsft @morganbr

### DIFF
--- a/src/mscorlib/shared/System/StringComparer.cs
+++ b/src/mscorlib/shared/System/StringComparer.cs
@@ -177,13 +177,13 @@ namespace System
         private readonly CompareInfo _compareInfo; // Do not rename (binary serialization)
         private readonly bool _ignoreCase; // Do not rename (binary serialization)
 
-        private CompareOptions Options => _ignoreCase ? CompareOptions.IgnoreCase : CompareOptions.None;
-
         internal CultureAwareComparer(CultureInfo culture, bool ignoreCase)
         {
             _compareInfo = culture.CompareInfo;
             _ignoreCase = ignoreCase;
         }
+
+        private CompareOptions Options => _ignoreCase ? CompareOptions.IgnoreCase : CompareOptions.None;
 
         public override int Compare(string x, string y)
         {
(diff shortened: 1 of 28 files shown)


morganbr:
Sadly, we can't add ISerializable to CultureAwareComparer because that breaks DataContractSerializer. You probably need to add the field from desktop to Core, mark the other fields as NonSerialized, and add an OnDeserializ(ing/ed) callback to handle the new field.

ViktorHofer:
@morganbr I see... I think I found a workaround, PTAL

morganbr:
@ViktorHofer LGTM

src/mscorlib/shared/System/StringComparer.cs
```diff
@@ -175,27 +175,29 @@ public int GetHashCode(object obj)
     internal sealed class CultureAwareComparer : StringComparer
     {
         private readonly CompareInfo _compareInfo; // Do not rename (binary serialization)
-        private readonly CompareOptions _options; // Do not rename (binary serialization)
+        private readonly bool _ignoreCase; // Do not rename (binary serialization)
+
+        private CompareOptions Options => _ignoreCase ? CompareOptions.IgnoreCase : CompareOptions.None;
```
justinvp: Nit: It'd be nice if the property was defined after the constructor.

ViktorHofer: Thanks

PR #12185, opened 2017-06-09
[RyuJIT/ARM32] Select proper temp register when resolving edges

### LINKED ISSUE #11851: [RyuJIT/arm32] Assertion failed 'isDoubleReg(reg1)'
At emitarm.cpp Line: 2083

Example:
```
Assert failure(PID 4252 [0x0000109c], Thread: 11220 [0x2bd4]): Assertion failed 'isDoubleReg(reg1)' in 'FractalPerf.Julia:Render():double:this' (IL size 160)

    File: c:\gh\coreclr\src\jit\emitarm.cpp Line: 2083
    Image: c:\brucefo\tests\Windows_NT.arm.Checked\Tests\Core_Root\CoreRun.exe
```

Tests with this assert:
```
JIT\Performance\CodeQuality\FractalPerf\FractalPerf\FractalPerf.cmd
```

hqueue:
Fix #11851 

For ARM32, we need double temporary register in addition to integer
and float temporary register when resolving edges between basic blocks.

### DIFF
--- a/src/jit/lsra.cpp
+++ b/src/jit/lsra.cpp
@@ -9885,7 +9885,7 @@ void LinearScan::resolveEdge(BasicBlock*      fromBlock,
 #endif // !_TARGET_XARCH_
     regNumber tempRegFlt = REG_NA;
 #ifdef _TARGET_ARM_
-    // ARM32 requires double temp register in addition to interger and float temp registers.
+    // ARM32 requires double temp register in addition to integer and float temp registers.
     regNumber tempRegDbl = REG_NA;
 #endif
     if ((compiler->compFloatingPointUsed) && (resolveType != ResolveSharedCritical))
@@ -9893,7 +9893,7 @@ void LinearScan::resolveEdge(BasicBlock*      fromBlock,
         tempRegFlt = getTempRegForResolution(fromBlock, toBlock, TYP_FLOAT);
 
 #ifdef _TARGET_ARM_
-        // ARM32 requires double temp register in addition to interger and float temp registers.
+        // ARM32 requires double temp register in addition to integer and float temp registers.
         tempRegDbl = getTempRegForResolution(fromBlock, toBlock, TYP_DOUBLE);
 #endif
     }
--- a/src/jit/lsra.cpp
+++ b/src/jit/lsra.cpp
@@ -9884,17 +9884,19 @@ void LinearScan::resolveEdge(BasicBlock*      fromBlock,
         (resolveType == ResolveSharedCritical) ? REG_NA : getTempRegForResolution(fromBlock, toBlock, TYP_INT);
 #endif // !_TARGET_XARCH_
     regNumber tempRegFlt = REG_NA;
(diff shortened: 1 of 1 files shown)


BruceForstall:
@dotnet/arm32-contrib

src/jit/lsra.cpp
```diff
@@ -9884,9 +9884,18 @@ void LinearScan::resolveEdge(BasicBlock*      fromBlock,
         (resolveType == ResolveSharedCritical) ? REG_NA : getTempRegForResolution(fromBlock, toBlock, TYP_INT);
 #endif // !_TARGET_XARCH_
     regNumber tempRegFlt = REG_NA;
+#ifdef _TARGET_ARM_
+    // ARM32 requires double temp register in addition to interger and float temp registers.
```
CarolEidt: Typo "Interger" => "integer"

src/jit/lsra.cpp
```diff
@@ -9884,9 +9884,18 @@ void LinearScan::resolveEdge(BasicBlock*      fromBlock,
...
+    regNumber tempRegDbl = REG_NA;
+#endif
     if ((compiler->compFloatingPointUsed) && (resolveType != ResolveSharedCritical))
     {
         tempRegFlt = getTempRegForResolution(fromBlock, toBlock, TYP_FLOAT);
+
+#ifdef _TARGET_ARM_
+        // ARM32 requires double temp register in addition to interger and float temp registers.
+        tempRegDbl = getTempRegForResolution(fromBlock, toBlock, TYP_DOUBLE);
```
CarolEidt: The resolution uses only one register at a time, so it would be better to reserve *only* a `TYP_DOUBLE` register on arm, and then its lower half can be used when resolving a `TYP_FLOAT`.

hqueue: > The resolution uses only one register at a time, so it would be better to reserve only a TYP_DOUBLE register on arm, and then its lower half can be used when resolving a TYP_FLOAT.

When considering code structure, I also think the suggested approach is better.
However I'm afraid it may affect code quality sometimes when we need `TYP_FLOAT` temp but there  there is a free `TYP_FLOAT` register and no free `TYP_DOUBLE` register. I've looked into `getTempRegForResolution()` and it does not guarantee reserving a temp register.
What do you think of it?

src/jit/lsra.cpp
hqueue: Thanks !

src/jit/lsra.cpp
CarolEidt: If you checked for a `TYP_DOUBLE` first, and it was found, then you could save its lower half as the `TYP_FLOAT` temp. If not, then you could check for a single register for `TYP_FLOAT`

hqueue:
@CarolEidt Please take a look. Thanks!

src/jit/lsra.cpp
hqueue: I see. I've updated the PR.

PR #12188, opened 2017-06-09
Allow users to set the number of build processes

parjong:
This commit allows users to override the number of build process via setting NumProc environment variable.

There is no change in default behavior.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -312,19 +312,6 @@ build_native()
         exit 1
     fi
     
-    if [[ -z ${NumProc} ]]; then
-        # Get the number of processors available to the scheduler
-        # Other techniques such as `nproc` only get the number of
-        # processors available to a single process.
-        if [ `uname` = "FreeBSD" ]; then
-            NumProc=`sysctl hw.ncpu | awk '{ print $2+1 }'`
-        elif [ `uname` = "NetBSD" ]; then
-            NumProc=$(($(getconf NPROCESSORS_ONLN)+1))
-        else
-            NumProc=$(($(getconf _NPROCESSORS_ONLN)+1))
-        fi
-    fi
-
     # Build
     if [ $__ConfigureOnly == 1 ]; then
         echo "Finish configuration & skipping $message build."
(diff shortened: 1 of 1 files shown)


janvorli:
@parjong I like the idea of having this configurable, I've already felt like I would like something like that in the past. But it would be great if you could go one step further and make it a command line option. I hope it would be just a few additional lines of code (creating a new variable, e.g. __buildNumProcLimit or something like that,  setting it based on a new command line option and then using it to limit the number of processors used for the build). Would you mind doing such change?

parjong:
@janvorli I added ``numproc`` option. I just used NumProc as it seems that it is not necessary to introduce an additional variable.

build.sh
```diff
@@ -816,6 +819,15 @@ while :; do
         msbuildonunsupportedplatform)
             __msbuildonunsupportedplatform=1
             ;;
+        numproc)
+            if [ -n "$2" ]; then
+              NumProc="$2"
```
janvorli: Could you please change this to __numProc to match other options extracted from the command line? And to match the style of initialization, it would be great to move the extraction of the default value (getting the processor count) from the build_native to the place where all the command line option variables are initialized.

parjong:
@janvorli I updated PR as suggested :)

build.sh
```diff
@@ -922,6 +909,21 @@ if [ $__CrossBuild == 1 ]; then
     fi
 fi
 
+if [[ -n "$__NumProc" ]]; then
+    NumProc="$__NumProc"
```
janvorli: I am sorry for not being clear, this is not what I meant. The main goal was to use `__NumProc` everywhere. I don't see a reason for having two variables for the same thing. Also, if you moved this code before the loop parsing the arguments (which is what I meant by putting it to the place where the other option variables are initialized), then you don't need the `if [[ -n "$__NumProc" ]]; then` here - the __NumProc would be set to the extracted number of processors by default and the command line option would just override the value.

PR #12198, opened 2017-06-09
Fix for checking enregisterLocalVars to display fpCalleeSaveCandidateVars

kyulee1:
fpCalleeSaveCandidateVars should be used under enregisterLocalVars.
Or it contains garbage value.

### DIFF
--- a/src/jit/lsra.cpp
+++ b/src/jit/lsra.cpp
@@ -2335,7 +2335,7 @@ void LinearScan::identifyCandidates()
     if (VERBOSE)
     {
         printf("\nFP callee save candidate vars: ");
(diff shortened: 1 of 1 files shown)


kyulee1:

CarolEidt:
> Isn't this #12172?

Yes, it seems to be. Odd that it thinks there are no conflicts

RussKeldorph:
> Yes, it seems to be. Odd that it thinks there are no conflicts

I can understand no conflicts since the change exactly matches @pgavlin's.  The interesting thing is that GitHub appears to be showing the diff relative to @kyulee1's merge-base, not the master that existed when determining whether there are conflicts.  I had always assumed the latter base would be used, but I confess I had never validated that assumption.

PR #12200, opened 2017-06-09
Fix fast tail call lowering bug.

### LINKED ISSUE #12037: Assertion in jit LSRA
Compile Program.cs below:

.\Tools\dotnetcli\dotnet.exe Tools\csc.exe /o+ out:Program.exe Program.cs /r:.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Private.CoreLib.dll /r:.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Console.dll /r:.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Runtime.dll

```
using System;
using System.Diagnostics;

namespace omg
{
    public class TestRyujitBug
    {
        public static void Main()
        {
            TopFunction(new object(), 0L, 0xBEEF, new TypedDouble(1.0), new TypedDouble(2.0), new TypedDouble(3.0), false, true);
        }

        public static void TopFunction(object parameters, long? nullable,
            double doubleArg, TypedDouble typedDouble1, TypedDouble typedDouble2, TypedDouble typedDouble3, bool shortCircuit, bool toggle)
        {
            if (shortCircuit)
            {
                return;
            }
            double param = toggle ? 19.0 : 17.0;

            InlinedFunction(doubleArg, param, typedDouble1.Value, typedDouble2.Value, typedDouble3.Value);
        }

        public static void InlinedFunction(double doubleArg, double param, double typedDoubleArg1, double typedDoubleArg2, double typedDoubleArg3)
        {
            Console.WriteLine("{0} {1} {2} {3} {4}", doubleArg, param, typedDoubleArg1, typedDoubleArg2, typedDoubleArg3);
        }

        public struct TypedDouble
        {
            public TypedDouble(double value)
            {
                Value = value;
            }
            public readonly double Value;
        }
    }
}
```
The run the resulting program:

.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\CoreRun.exe Program.exe

Assert failure(PID 8592 [0x00002190], Thread: 11128 [0x2b78]): Assertion failed '!interval->isLocalVar' in 'omg.TestRyujitBug:TopFunction(ref,struct,double,struct,struct,struct,bool,bool)' (IL size 58)

    File: c:\coreclr\src\jit\lsra.cpp Line: 531
    Image: c:\coreclr\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\CoreRun.exe

Running with release coreclr results in silent bad codegen. The output is 

`48879 19 1 48879 3`
instead of
`48879 19 1 2 3`

erozenfeld:
Lowering::LowerFastTailCall has code that handles the case when the caller and the callee both have parameters passed on the stack and the callee has register arguments that are computed from the caller's stack-passed arguments. In that case the stack arguments of the callee are set up in the caller's incoming argument area so they override the caller's arguments. Lowering::LowerFastTailCall creates temps for any caller's arguments passed on the stack that are used to compute callee's register arguments. That code had a bug where it was converting GT_LCL_FLD nodes into GT_LCL_VAR nodes. That's problematic both for the case of a single-field struct with a double field and for multi-field structs.
In the former case we were getting LSRA asserts in checked builds and SBCG in release builds. In the latter case we were getting SBCG both in checked and release builds.

The fix is not to convert GT_LCL_FLD nodes into GT_LCL_VAR nodes and mark the new temps as not-enregisterable if the local is not-enregisterable.

An alternative fix would be to create multiple register temps for each GT_LCL_FLD or GT_LCL_VAR targeting the local but that would complicate the code and I decided to go with a simpler solution for this uncommon case.

Fixes #12037.

### DIFF
--- a/tests/src/JIT/Regression/JitBlue/GitHub_12037/GitHub_12037.cs
+++ b/tests/src/JIT/Regression/JitBlue/GitHub_12037/GitHub_12037.cs
@@ -1,0 +1,109 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+public class TailCallOptTest
+{
+    public static int Main()
+    {
+        bool res1 = Caller1(new object(), 0L, 0xBEEF, new TypedDouble(1.0), new TypedDouble(2.0), new TypedDouble(3.0));
+        bool res2 = Caller2(new object(), 0L, 0xBEEF, new TypedDouble(1.0), new TwoInts(3, 5), new TypedDouble(3.0));
+        return (res1 && res2) ? 100 : 0;
+    }
+
+    // In this test typedDouble2 is passed to Caller1 on the stack. Then typedDouble2.Value is passed to Callee1 in a register.
+    // Since Calee1 also has a stack argument (typedDouble3.Value) and the call is dispatched as a fast tail call,
+    // it's set up in the incoming argument area of Caller1. JIT lowering code needs to ensure that typedDouble2
+    // in that area is not overwritten by typedDouble3.Value before typedDouble2.Value is computed. The JIT had a bug in that code
+    // because typedDouble2.Value was represented as GT_LCL_FLD and it was incorrectly converted into GT_LCL_VAR resulting in type
+    // mismatches (long vs. double since the struct is passed as a long but its only field is double).
+    public static bool Caller1(object parameters, long l,
+        double doubleArg, TypedDouble typedDouble1, TypedDouble typedDouble2, TypedDouble typedDouble3)
+    {
+        double param = 19.0;
+
+        Console.Write("Let's ");
+        Console.Write("Discourage ");
+        Console.Write("Inlining ");
+        Console.Write("Of ");
+        Console.Write("Caller ");
+        Console.Write("Into ");
+        Console.WriteLine("Main.");
+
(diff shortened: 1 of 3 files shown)


erozenfeld:
6 methods had diffs (expected regressions) in desktop SPMI (total of 50 bytes).

tests/src/JIT/Regression/JitBlue/GitHub_12037/GitHub_12037.cs
```diff
@@ -0,0 +1,109 @@
...
+        bool res2 = Caller2(new object(), 0L, 0xBEEF, new TypedDouble(1.0), new TwoInts(3, 5), new TypedDouble(3.0));
+        return (res1 && res2) ? 100 : 0;
+    }
+
+    // In this test typedDouble2 is passed to Caller1 on the stack. Then typedDouble2.Value is passed to Callee1 in a register.
+    // Since Calee1 also has a stack argument (typedDouble3.Value) and the call is dispatched as a fast tail call,
+    // it's set up in the incoming argument area of Caller1. JIT lowering code needs to ensure that typedDouble2
+    // in that area is not overwritten by typedDouble3.Value before typedDouble2.Value is computed. The JIT had a bug in that code
+    // because typedDouble2.Value was represented as GT_LCL_FLD and it was incorrectly converted into GT_LCL_VAR resulting in type
+    // mismatches (long vs. double since the struct is passed as a long but its only field is double).
```
CarolEidt: Thanks for the great comment that explains what this test is doing!

erozenfeld:

PR #12209, opened 2017-06-09
make REG_VIRTUAL_STUB_PARAM depended on Abi.

sandreenko:
Add VirtualStubParam that is saved in compiler and must be init in
compCompile.

### DIFF
--- a/src/jit/compiler.h
+++ b/src/jit/compiler.h
@@ -6704,7 +6704,7 @@ public:
 #endif
     };
 
-    VirtualStubParamInfo* virtualStubParam;
+    VirtualStubParamInfo* virtualStubParamInfo;
 
     inline bool IsTargetAbi(CORINFO_RUNTIME_ABI abi)
     {
--- a/src/jit/compiler.h
+++ b/src/jit/compiler.h
@@ -6633,6 +6633,79 @@ public:
 #endif
     }
 
+    //------------------------------------------------------------------------
+    // VirtualStubParam: virtual stub dispatch extra parameter (slot address).
+    //
+    // It represents Abi and target specific registers for the parameter.
+    //
+    class VirtualStubParam
+    {
+    public:
+        VirtualStubParam(bool isCoreRTABI)
+        {
+#if defined(_TARGET_X86_)
+            reg = REG_EAX;
+            rbm = RBM_EAX;
+#elif defined(_TARGET_AMD64_)
+            if (isCoreRTABI)
+            {
+                reg = REG_R10;
+                rbm = RBM_R10;
+            }
+            else
+            {
+                reg = REG_R11;
+                rbm = RBM_R11;
+            }
+#elif defined(_TARGET_ARM_)
+            reg = REG_R4;
+            rbm = RBM_R4;
+#elif defined(_TARGET_ARM64_)
+            reg = REG_R11;
+            rbm = RBM_R11;
+#else
+#error Unsupported or unset target architecture
+#endif
+
+#ifdef LEGACY_BACKEND
+#if defined(_TARGET_X86_)
+            predict = PREDICT_REG_EAX;
+#elif defined(_TARGET_ARM_)
+            predict = PREDICT_REG_R4;
+#else
+#error Unsupported or unset target architecture
+#endif
+#endif // LEGACY_BACKEND
+        }
+
+        regNumber GetReg() const
+        {
+            return reg;
+        }
+
+        _regMask_enum GetRbm() const
+        {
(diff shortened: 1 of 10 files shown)


sandreenko:
PTAL @MichalStrehovsky,  @dotnet/jit-contrib.

src/jit/target.h
```diff
@@ -310,6 +310,79 @@ typedef unsigned short regPairNoSmall; // arm: need 12 bits
...
 
+//------------------------------------------------------------------------
+// VirtualStubParam: virtual stub dispatch extra parameter (slot address).
+//
+// It represents Abi and target specific registers for the parameter.
+//
+class VirtualStubParam
+{
+public:
+    VirtualStubParam(bool IsCoreRTABI)
```
jkotas: Nit: Start with lower case?

src/jit/target.h
```diff
@@ -310,6 +310,79 @@ typedef unsigned short regPairNoSmall; // arm: need 12 bits
...
+        rbm = REG_R11;
+#else
+#error Unsupported or unset target architecture
+#endif
+
+#ifdef LEGACY_BACKEND
+#if defined(_TARGET_X86_)
+        predict = PREDICT_REG_EAX;
+#elif defined(_TARGET_AMD64_)
+        noway_assert(!IsCoreRTABI);
```
jkotas: Do we need this - does the legacy backend works for AMD64 at all?

sandreenko: Nop, deleted.

src/jit/target.h
sandreenko: Fixed.

jkotas:
Looks good to me. 

Allocating the class that holds the register numbers feels like too much ceremony for this to me. I will  leave it to others to comment about that.

src/jit/compiler.h
```diff
@@ -6633,6 +6633,79 @@ class Compiler
 #endif
     }
 
+    //------------------------------------------------------------------------
+    // VirtualStubParam: virtual stub dispatch extra parameter (slot address).
+    //
+    // It represents Abi and target specific registers for the parameter.
+    //
+    class VirtualStubParam
```
pgavlin: Nit: I think this is better named `VirtualStubParamInfo` or similar.

src/jit/compiler.h
```diff
@@ -6633,6 +6633,79 @@ class Compiler
...
+#endif
+#endif // LEGACY_BACKEND
+        }
+
+        regNumber GetReg() const
+        {
+            return reg;
+        }
+
+        _regMask_enum GetRbm() const
```
pgavlin: Nit: `s/GetRbm/GetRegMask/`

sandreenko: Fixed, what does rbm mean?

src/jit/compiler.h
sandreenko: Fixed

src/jit/compiler.h
```diff
@@ -6633,6 +6633,79 @@ class Compiler
...
+    private:
+        regNumber     reg;
+        _regMask_enum regMask;
+
+#ifdef LEGACY_BACKEND
+        rpPredictReg predict;
+#endif
+    };
+
+    VirtualStubParamInfo* virtualStubParam;
```
pgavlin: Should have pointed this one out, too: `s/virtualStubParam/virtualStubParamInfo/`

sandreenko: Fixed

src/jit/compiler.h
```diff
@@ -6633,6 +6633,79 @@ class Compiler
...
+            predict = PREDICT_REG_EAX;
+#elif defined(_TARGET_ARM_)
+            predict = PREDICT_REG_R4;
+#else
+#error Unsupported or unset target architecture
+#endif
+#endif // LEGACY_BACKEND
+        }
+
+        regNumber GetReg() const
```
sandreenko: do we need any __forceinline directives here? This part doesn't affect performance, but as template for the future changes maybe it will be nice to state it here.

mikedn: What for? I don't think you'll ever find a C/C++ compiler that doesn't inline this.

sandreenko: Ok, sounds reasonable.

sandreenko:
@russellhadley I am not sure do we keep CodeGen-EE Contract in ProjectN OneNote up-to-date? Should we change it after this PR?

russellhadley:
@sandreenko Keeping it up to date sounds good to me.

Issue #12137, opened 2017-06-07
Incorrect result from Math.Round method


vkvenkat:
The Math.Round method is expected to round a double-precision floating-point value to the nearest integral value based on the description in https://msdn.microsoft.com/en-us/library/wyk4d9cy(v=vs.110).aspx.
So, Math.Round(0.50000000000000011102230246251565404236316680908203) should return the nearest integer, 1. Instead it returns 0 on Windows and Ubuntu.

We end up calling the COMDouble::Round method in [floatdouble.cpp:174](https://github.com/dotnet/coreclr/blob/72c57d9d9fd2d4cd8c96f124d132d316c8ef6013/src/classlibnative/float/floatdouble.cpp#L174), where the rounding happens.  

Here is the repro. The issue is with rounding the 5th element (0.50000000000000011102230246251565404236316680908203) of the _values_ array below:

```
using System;
using System.Runtime.InteropServices;

[StructLayout(LayoutKind.Explicit)] 
public struct SampleUnion
{
    [FieldOffset(0)] public ulong  ul;
    [FieldOffset(0)] public double d;
    public void Display()
    {
        Console.WriteLine("The stored value is: 0x{0:X}", ul);
        Console.WriteLine("Default rounding    {0} --> {1}", d, Math.Round(d));
        Console.WriteLine("Half to even        {0} --> {1}", d, Math.Round(d, 0, MidpointRounding.ToEven));
        Console.WriteLine("Half away from zero {0} --> {1}", d, Math.Round(d, 0, MidpointRounding.AwayFromZero));
    }
}

public class Example
{
   public static void Main()
   {
      double[] values = { -.5, -.50000000000000011102230246251565404236316680908203, -.50000000000000022204460492503130808472633361816406,
                          0.5, 0.50000000000000011102230246251565404236316680908203, 0.50000000000000022204460492503130808472633361816406,
                          1.5, 1.50000000000000022204460492503130808472633361816406,
                          2.5, 2.5000000000000004440892098500626161694526672363281 };
      SampleUnion su = new SampleUnion();

      foreach (double value in values) {
         su.d = value;
         su.Display();

      }
   }
}
```

PR #12210, opened 2017-06-09
Fix a COMDouble::Round() issue

hanblee:
fixes https://github.com/dotnet/coreclr/issues/12137

### DIFF
--- a/tests/src/CoreMangLib/cti/system/math/mathround3.cs
+++ b/tests/src/CoreMangLib/cti/system/math/mathround3.cs
@@ -13,6 +13,8 @@ public class MathRound3
         TestLibrary.TestFramework.LogInformation("[Positive]");
         retVal = PosTest1() && retVal;
         retVal = PosTest2() && retVal;
+        retVal = PosTest3() && retVal;
+        retVal = PosTest4() && retVal;
 
         retVal = NegTest1() && retVal;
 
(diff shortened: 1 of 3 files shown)


hanblee:
@tannergooding PTAL

hanblee:
@tannergooding BTW, if you are not the right person to review this, it would be great if you would please pull in the right person(s). Thanks!

tannergooding:
Taking a look now.

tannergooding:
It is important to note that the MSDN [page](https://msdn.microsoft.com/en-us/library/wyk4d9cy(v=vs.110).aspx) for `Math.Round(double)` states:
> Rounds a double-precision floating-point value to the nearest integral value.

However, the actual behavior, which is on the general `Math.Round` [page](https://msdn.microsoft.com/en-us/library/system.math.round(v=vs.110).aspx) is:
> Rounds a double-precision floating-point value to the nearest integer, and rounds midpoint values to the nearest even number 

The secondary portion (rounding midpoint values to the nearest even number) is important for just this scenario and the exact behavior taken here.

Additionally, as an FYI, the original (unmodified) code from Desktop was https://github.com/hanblee/coreclr/commit/08786f20e89eb5f518d8d25f3e7f886f69d994ea#diff-884cab80438864e5dc66c7e4c1a3c723L467, which is identical behavior to implementation currently checked in.

When completely "unrolled" the current implementation is doing
```C++
(floor(x + 0.5) == (x + 0.5)) && (fmod((x + 0.5), 2.0) != 0)
```
and the newly proposed implementation is doing:
```C++
(x == (floor(x) + 0.5)) && (fmod(floor(x + 0.5), 2.0) != 0)
```

The previous algorithm would end up comparing "precisely 1" vs "precisely 1".
The new algorithm ends up comparing 0.5 vs 0.50000000000000011, which is what allows this fix.

This is definitely an improvement as it should only do the midpoint rounding calculation when x is exactly midpoint (as best as can be represented by IEEE double-precision floating-point values, anyways).

tannergooding:
It would be great if you could add a couple tests covering this scenario (CoreFX is the primary repo for managed tests, but there are some tests in CoreCLR as well: https://github.com/dotnet/coreclr/tree/32f0f9721afb584b4a14d69135bea7ddc129f755/tests/src/CoreMangLib/cti/system/math)

tannergooding:
FYI. @janvorli, @AndyAyersMS, @jkotas who have reviewed these types of changes previously.

hanblee:
> It would be great if you could add a couple tests covering this scenario 

Good point. Will do.

tests/src/CoreMangLib/cti/system/math/mathround3.cs
```diff
@@ -91,6 +93,48 @@ public bool PosTest2()
...
+
+        // Test based on https://github.com/dotnet/coreclr/issues/12137
+        TestLibrary.TestFramework.BeginScenario("PosTest4: Verify Round(System.double) when decimal part of arg is very close to 0.5 .");
+
+        double doubleVal = 0.50000000000000011102230246251565404236316680908203;
+        double expectedVal = 1.0;
+
+        if (Math.Round(doubleVal) != expectedVal)
+        {
+            Console.WriteLine("actual value = " + Math.Round(doubleVal));
```
tannergooding: Could you use the `G17` format specifier when printing here and below?
```
> double doubleVal = 0.50000000000000011102230246251565404236316680908203;
> Console.WriteLine(doubleVal)
0.5
> Console.WriteLine("{0:G17}", doubleVal)
0.50000000000000011
```

hanblee: Done

tannergooding:
Thanks! This will need someone from the CoreCLR team to give final approval before it can be merged.

hanblee:
Got it!

hanblee:
@janvorli PTAL

PR #12215, opened 2017-06-11
Remove serializable attribute from OrdinalIgnoreCaseComparer

ViktorHofer:
This type doesn't exist in Desktop.

### DIFF
--- a/src/mscorlib/shared/System/StringComparer.cs
+++ b/src/mscorlib/shared/System/StringComparer.cs
@@ -6,6 +6,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Diagnostics.Contracts;
+using System.Runtime.Serialization;
 
 namespace System
 {
@@ -252,9 +253,8 @@ namespace System
         public override int GetHashCode() => nameof(OrdinalComparer).GetHashCode();
     }
 
-    [Serializable]
-    [System.Runtime.CompilerServices.TypeForwardedFrom("mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089")]
-    internal sealed class OrdinalIgnoreCaseComparer : StringComparer
+    [Serializable]    
+    internal sealed class OrdinalIgnoreCaseComparer : StringComparer, ISerializable
     {
         public override int Compare(string x, string y) => string.Compare(x, y, StringComparison.OrdinalIgnoreCase);
 
@@ -276,5 +276,11 @@ namespace System
         // Equals/GetHashCode methods for the comparer itself. 
         public override bool Equals(object obj) => obj is OrdinalIgnoreCaseComparer;
         public override int GetHashCode() => nameof(OrdinalIgnoreCaseComparer).GetHashCode();
+
+        public void GetObjectData(SerializationInfo info, StreamingContext context)
+        {
+            info.SetType(typeof(OrdinalComparer));
+            info.AddValue("_ignoreCase", true);
+        }
     }
 }
--- a/src/mscorlib/shared/System/StringComparer.cs
+++ b/src/mscorlib/shared/System/StringComparer.cs
@@ -16,7 +16,7 @@ namespace System
     {
         private static readonly CultureAwareComparer s_invariantCulture = new CultureAwareComparer(CultureInfo.InvariantCulture, false);
         private static readonly CultureAwareComparer s_invariantCultureIgnoreCase = new CultureAwareComparer(CultureInfo.InvariantCulture, true);
-        private static readonly OrdinalComparer s_ordinal = new OrdinalComparer();
+        private static readonly OrdinalCaseSensitiveComparer s_ordinal = new OrdinalCaseSensitiveComparer();
         private static readonly OrdinalIgnoreCaseComparer s_ordinalIgnoreCase = new OrdinalIgnoreCaseComparer();        
 
         public static StringComparer InvariantCulture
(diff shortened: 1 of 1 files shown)


jkotas:
BTW: Have you looked into whether GetEqualityComparerForSerialization needs to be added back(https://github.com/Microsoft/referencesource/search?utf8=%E2%9C%93&q=GetEqualityComparerForSerialization&type=)? It is solving similar sort of problems.

ViktorHofer:
I see. That means I can set the type to OrdinalComparer, which is the corresponding type on Desktop.

ViktorHofer:
@jkotas I haven't yet found a case where GetEqualityComparerForSerialization needs to be ported back. What was it exactly used for?

src/mscorlib/shared/System/StringComparer.cs
```diff
@@ -276,5 +276,11 @@ public override int GetHashCode(string obj)
         // Equals/GetHashCode methods for the comparer itself. 
         public override bool Equals(object obj) => obj is OrdinalIgnoreCaseComparer;
         public override int GetHashCode() => nameof(OrdinalIgnoreCaseComparer).GetHashCode();
+
+        public void GetObjectData(SerializationInfo info, StreamingContext context)
+        {
+            info.SetType(typeof(OrdinalComparer));
+            info.AddValue("_ignoreCase", true);
```
ViktorHofer: @jkotas This will work when we serialize from Core to Desktop but not the other way around because OrdinalComparer in Core doesn't use the _ignoreCase value. We need some routine to deserialize to OrdinalIgnoreCaseComparer instead of OrdinalComparer if _ignoreCase is true.

I found some info about creating a custom SerializationBinder but this is a) a hassle and b) won't work because we don't have the _ignoreCase value at that time to decide to which type to deserialize.

jkotas: I think that the best way to deal with this is to add back the bool `_ignoreCase` back to OrdinalComparer. It will make the OrdinalComparer serialize perfectly. And then do the performance optimization that saved the if-check by inheriting from the `OrdinalCompared`.

jkotas:
> GetEqualityComparerForSerialization 

Dictionary flips its true equality comparer to NonRandomizedStringEqualityComparer as a performance optimization: https://github.com/dotnet/coreclr/blob/cdb827b6cf72bdb8b4d0dbdaec160c32de7c185f/src/mscorlib/src/System/Collections/Generic/Dictionary.cs#L98

This flips has to be undone for serialization.

What is the minimum desktop version that you are targeting? If the minimum desktop version that you are targeting is guaranteed to have NonRandomizedStringEqualityComparer, you may be fine.

ViktorHofer:
> What is the minimum desktop version that you are targeting

netfx461

I can't find any information of it on Api Catalogue as the type isn't exposed publicly. I can find it on  Reference sources for netfx47: https://referencesource.microsoft.com/#mscorlib/system/collections/generic/equalitycomparer.cs,310

src/mscorlib/shared/System/StringComparer.cs
ViktorHofer: If I understand you correctly you mean:

1. add the _ignoreCase field OrdinalComparer
2. remove the _ignoreCase field from OrdinalIgnoreCaseComparer
3. inheriting OrdinalIgnoreCaseComparer from OrdinalComparer

Is that right? If yes then I don't think that will solve the issue. When Core deserializes the core/netfx blob it will create an object of type OrdinalComparer (even if the _ignoreCase field is true) and sets the _ignoreCase fields value. The OriginalIgnoreCaseComparer type will never be used. Am I missing something?

jkotas: - Make OrdinalComparer to be exactly like what it used to be in desktop.
- Inherit OrdinalCaseSensitiveComparer and OrdinalIgnoreCaseComparer that have the optimized implementations
   - In OrdinalCaseSensitiveComparer.GetObjectData and OrdinalIgnoreCaseComparer.GetObjectData, do info.SetType(typeof(OrdinalComparer)) so that they serialize into know desktop type

Ssame pattern as what is used for EnumEqualityComparer. It also has more optimized version and serializes as the less optimized version.

danmosemsft:
@ViktorHofer talk to @AlexGhiondea to get help figuring out when an internal type appeared in what version. Probalby easiest if he points you to the correct drop share so you can look at the binary.

ViktorHofer:
@danmosemsft thanks. I just downloaded the referencesources for netfx461 and found the type. It is surrounded by #if FEATURE_CORECLR so we should be safe.

ViktorHofer:
@jkotas It took me a while to understand that you meant to introduce a new specialized type. I guess it should be fine now. PTAL

src/mscorlib/shared/System/StringComparer.cs
```diff
@@ -228,7 +229,85 @@ public override int GetHashCode()
 
     [Serializable]
     [System.Runtime.CompilerServices.TypeForwardedFrom("mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089")]
-    internal sealed class OrdinalComparer : StringComparer 
+    internal class OrdinalComparer : StringComparer 
+    {
+        private bool _ignoreCase;
```
jkotas: This should have constructor that initializes `_ignoreCase` accordingly.

src/mscorlib/shared/System/StringComparer.cs
```diff
@@ -244,17 +323,22 @@ public override int GetHashCode(string obj)
...
 #endif
             }
-            return obj.GetHashCode();
+            return TextInfo.GetHashCodeOrdinalIgnoreCase(obj);
         }
 
         // Equals/GetHashCode methods for the comparer itself. 
-        public override bool Equals(object obj) => obj is OrdinalComparer;
-        public override int GetHashCode() => nameof(OrdinalComparer).GetHashCode();
+        public override bool Equals(object obj) => obj is OrdinalCaseSensitiveComparer;
```
jkotas: This should just use the parent implementation, not override.

The current impl is not symmetric: OrdinalComparer.Equals(OrdinalCaseSensitiveComparer) is true;  OrdinalCaseSensitiveComparer.Equals(OrdinalComparer) is false.

ViktorHofer: Currently `OrdinalIgnoreCaseComparer` and `OrdinalCaseSensitiveComparer` do not inherit from `OrdinalComparer` but from `StringComparer`. In the example you mentioned `EnumEqualityComparer  doesn't inherit from `ObjectEqualityComparer` either.

I don't think they should inherit `OrdinalComparer` as we are loosing optimizations with it and also increasing memory footprint because of the _ignoreCase field.

Do we need the inheritance just because of the Equality check?

src/mscorlib/shared/System/StringComparer.cs
ViktorHofer: It doesn't have one in netfx (this is the port from it). The type doesn't implement ISerializable so why would we need one?

src/mscorlib/shared/System/StringComparer.cs
jkotas: > increasing memory footprint because of the _ignoreCase field

This does not matter. These are singletons. And the extra field will actually fit into the default padding that each object has, so it won't actually increase the memory footprint at all.

> Do we need the inheritance just because of the Equality check?

It is the easiest way to make the Equality check work.

src/mscorlib/shared/System/StringComparer.cs
jkotas: This comment was related to the other one.

ViktorHofer:
Thanks Jan for the feedback! I hope now it's correct.

src/mscorlib/shared/System/StringComparer.cs
```diff
@@ -244,18 +283,72 @@ public override int GetHashCode(string obj)
...
+
+        public override int GetHashCode(string obj)
+        {
+            if (obj == null)
+            {
+#if CORECLR
+                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.obj);
+#else
+                throw new ArgumentNullException(nameof(obj));
+#endif
```
stephentoub: Do we ever compile this code without CORECLR set?  I thought we'd previously removed all such defines usage, but maybe I'm wrong?

ViktorHofer: I'm not sure. It was already in before my changes (https://github.com/ViktorHofer/coreclr/blob/1a32d7f2c1be31741c3e71fe35eb2fd22664d12c/src/mscorlib/shared/System/StringComparer.cs#L236)

jkotas: > Do we ever compile this code without CORECLR set

Yes, we do - in CoreRT. Regular exception throwing works better in ProjectN (because of the toolchain recognizes and does special optimizations for) than the ThrowHelper hack (that the toolchain does not recognize). It makes folks to sometime put the ThrowHelper under ifdefs when moving code into the shared corelib partition. We are not consistent about it though. I think that the ifdef can be removed in this case without any harm.

Related:
https://github.com/dotnet/roslyn/issues/15929
https://github.com/dotnet/corert/issues/1832

stephentoub: I see. Thanks.

PR #12104, opened 2017-06-06
Fix for CoreRT issue https://github.com/dotnet/corert/issues/3672 (armel tizen)

sergign60:

PR #12232, opened 2017-06-12
[Arm32] Fix legacy backend build when FEATURE_READYTORUN_COMPILER is not defined

jashook:


jashook:
@BruceForstall @jkotas ptal
/cc @dotnet/jit-contrib @sergign60

jashook:
The break originally came in with #12104

src/jit/codegenlegacy.cpp
```diff
@@ -19201,12 +19201,14 @@ regMaskTP CodeGen::genCodeForCall(GenTreeCall* call, bool valUsed)
                 {
                     noway_assert(helperNum != CORINFO_HELP_UNDEF);
 
+                    #ifdef FEATURE_READYTORUN_COMPILER
```
jkotas: Nit: I think that the convention is to start the ifdefs at the start of the line

src/jit/codegenlegacy.cpp
```diff
@@ -19201,12 +19201,14 @@ regMaskTP CodeGen::genCodeForCall(GenTreeCall* call, bool valUsed)
...
                     if (call->gtEntryPoint.addr != NULL)
                     {
                         accessType = call->gtEntryPoint.accessType;
                         addr       = call->gtEntryPoint.addr;
                     }
                     else
+                    #endif FEATURE_READYTORUN_COMPILER
```
BruceForstall: This should be:
```
#endif // FEATURE_READYTORUN_COMPILER
```

jashook:

PR #12184, opened 2017-06-09
Enable legacy jit for x86

rartemev:
Fix and enable legacy jit for x86

PR #12237, opened 2017-06-12
[ARM] Set local variables as addr-exposed (for LEGACY_BACKEND)

### LINKED ISSUE #12218: [ARM/Linux] System.Threading.Tasks.Tests test failure
ARM/Linux shows the following error while running Core FX System.Threading.Tasks.Tests:
```

Discovering: System.Threading.Tasks.Tests
Discovered:  System.Threading.Tasks.Tests
Starting:    System.Threading.Tasks.Tests
   System.Threading.Tasks.Tests.TaskAwaiterTests.AwaiterAndAwaitableEquality [FAIL]
      Assert.NotEqual() Failure
      Expected: Not System.Runtime.CompilerServices.ConfiguredTaskAwaitable+ConfiguredTaskAwaiter
      Actual:   System.Runtime.CompilerServices.ConfiguredTaskAwaitable+ConfiguredTaskAwaiter
      Stack Trace:
         /var/jenkins/workspace/FX/FX_Managed_Release/repo/code/src/System.Threading.Tasks/tests/System.Runtime.CompilerServices/TaskAwaiterTests.cs(235,0): at System.Threading.Tasks.Tests.TaskAwaiterTests.Awaiter
AndAwaitableEquality()
   System.Threading.Tasks.Tests.TaskSchedulerTests.GetScheduledTasksForDebugger_DebuggerAttached_ReturnsTasksFromCustomSchedulers [SKIP]
      Condition(s) not met: \"DebuggerIsAttached\"
   System.Threading.Tasks.Tests.TaskSchedulerTests.GetTaskSchedulersForDebugger_DebuggerAttached_ReturnsAllSchedulers [SKIP]
      Condition(s) not met: \"DebuggerIsAttached\"
Finished:    System.Threading.Tasks.Tests

=== TEST EXECUTION SUMMARY ===
   System.Threading.Tasks.Tests  Total: 778, Errors: 0, Failed: 1, Skipped: 2, Time: 44.293s
```

parjong:
Value numbering currently does not take account of struct updates via copy, which results in #12218.
 
This commit sets local variables in copy block as addr-exposed even for legacy backend, which makes value numbering correct and thus fixes  #12218.

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -8963,7 +8963,6 @@ GenTreePtr Compiler::fgMorphOneAsgBlockOp(GenTreePtr tree)
                 }
                 else
                 {
-#ifndef LEGACY_BACKEND
 
                     // The source argument of the copyblk can potentially
                     // be accessed only through indir(addr(lclVar))
@@ -8973,6 +8972,7 @@ GenTreePtr Compiler::fgMorphOneAsgBlockOp(GenTreePtr tree)
                     // we don't delete it as a dead store later on.
                     unsigned lclVarNum                = lclVarTree->gtLclVarCommon.gtLclNum;
                     lvaTable[lclVarNum].lvAddrExposed = true;
+#ifndef LEGACY_BACKEND
                     lvaSetVarDoNotEnregister(lclVarNum DEBUGARG(DNER_AddrExposed));
 
 #else  // LEGACY_BACKEND
@@ -10343,7 +10343,6 @@ GenTreePtr Compiler::fgMorphCopyBlock(GenTreePtr tree)
 
             tree = gtNewAssignNode(gtNewLclvNode(addrSpillTemp, TYP_BYREF), addrSpill);
 
-#ifndef LEGACY_BACKEND
             // If we are assigning the address of a LclVar here
             // liveness does not account for this kind of address taken use.
             //
@@ -10361,7 +10360,6 @@ GenTreePtr Compiler::fgMorphCopyBlock(GenTreePtr tree)
                     lvaSetVarDoNotEnregister(lclVarNum DEBUGARG(DNER_AddrExposed));
                 }
             }
-#endif // !LEGACY_BACKEND
         }
 
     _AssignFields:
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -8963,6 +8963,7 @@ GenTreePtr Compiler::fgMorphOneAsgBlockOp(GenTreePtr tree)
                 }
                 else
                 {
+#ifndef LEGACY_BACKEND
 
                     // The source argument of the copyblk can potentially
                     // be accessed only through indir(addr(lclVar))
@@ -8972,7 +8973,6 @@ GenTreePtr Compiler::fgMorphOneAsgBlockOp(GenTreePtr tree)
                     // we don't delete it as a dead store later on.
                     unsigned lclVarNum                = lclVarTree->gtLclVarCommon.gtLclNum;
                     lvaTable[lclVarNum].lvAddrExposed = true;
-#ifndef LEGACY_BACKEND
                     lvaSetVarDoNotEnregister(lclVarNum DEBUGARG(DNER_AddrExposed));
 
 #else  // LEGACY_BACKEND


parjong:

hqueue:

BruceForstall:
@dotnet/arm32-contrib

BruceForstall:
@parjong Do you intend for this to be temporary, to fix an existing bug? Will you add a GitHub issue to fix this? Does this bug affect both RyuJIT and `LEGACY_BACKEND`?

@briansull @JosephTremoulet PTAL

parjong:
@BruceForstall I intended this PR as a temporary workaround to fix the existing bug. The proper way would be to fix value numbering, but I currently have no idea on how to fix that. 

I'm not sure about RyuJIT.

\CC @hqueue  @hseok-oh Could you let me know the current status? RyuJIT could compile the example code in #12218?

hseok-oh:
@parjong @BruceForstall There is no bug on RyuJIT/ARM32 related to #12218, because it generate correct value number in this case. You can find the implementation related to this at https://github.com/dotnet/coreclr/blob/0d122b4cdb7beaf8fcae466e14134d1706b3e4c1/src/jit/morph.cpp#L10346-L10364
```
#ifndef LEGACY_BACKEND
            // If we are assigning the address of a LclVar here
            // liveness does not account for this kind of address taken use.
            //
            // We have to mark this local as address exposed so
            // that we don't delete the definition for this LclVar
            // as a dead store later on.
            //
            if (addrSpill->OperGet() == GT_ADDR)
            {
                GenTreePtr addrOp = addrSpill->gtOp.gtOp1;
                if (addrOp->IsLocal())
                {
                    unsigned lclVarNum                = addrOp->gtLclVarCommon.gtLclNum;
                    lvaTable[lclVarNum].lvAddrExposed = true;
                    lvaSetVarDoNotEnregister(lclVarNum DEBUGARG(DNER_AddrExposed));
                }
            }
#endif // !LEGACY_BACKEND
```

BruceForstall:
Should that code be enabled for `LEGACY_BACKEND`? Or, should your fix here be under:
```
#if !(defined(LEGACY_BACKEND) && defined(_TARGET_ARM_))
```

parjong:
@BruceForstall I am currently testing the following patch:
```
diff --git a/src/jit/morph.cpp b/src/jit/morph.cpp
index 215b45f..92d2f6e 100644
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -8888,7 +8888,6 @@ GenTreePtr Compiler::fgMorphOneAsgBlockOp(GenTreePtr tree)
                 }
                 else
                 {
-#ifndef LEGACY_BACKEND

                     // The source argument of the copyblk can potentially
                     // be accessed only through indir(addr(lclVar))
@@ -8898,6 +8897,7 @@ GenTreePtr Compiler::fgMorphOneAsgBlockOp(GenTreePtr tree)
                     // we don't delete it as a dead store later on.
                     unsigned lclVarNum                = lclVarTree->gtLclVarCommon.gtLclNum;
                     lvaTable[lclVarNum].lvAddrExposed = true;
+#ifndef LEGACY_BACKEND
                     lvaSetVarDoNotEnregister(lclVarNum DEBUGARG(DNER_AddrExposed));

 #else  // LEGACY_BACKEND
@@ -10268,7 +10268,6 @@ GenTreePtr Compiler::fgMorphCopyBlock(GenTreePtr tree)

             tree = gtNewAssignNode(gtNewLclvNode(addrSpillTemp, TYP_BYREF), addrSpill);

-#ifndef LEGACY_BACKEND
             // If we are assigning the address of a LclVar here
             // liveness does not account for this kind of address taken use.
             //
@@ -10286,7 +10285,6 @@ GenTreePtr Compiler::fgMorphCopyBlock(GenTreePtr tree)
                     lvaSetVarDoNotEnregister(lclVarNum DEBUGARG(DNER_AddrExposed));
                 }
             }
-#endif // !LEGACY_BACKEND
         }

     _AssignFields:

```

This patch resolves the above issue, and there is no regressions in FX innerloop result. I am currently running CLR Pri 2 tests (I'll let you know when the result is ready).

parjong:
@BruceForstall Is there any difference between legacy JIT and Ryu JIT in terms of value numbering and dead code elimination? I think that we should enable this code for both JITs if there is no difference.

mikedn:
The legacy JIT doesn't do value numbering. This also means that its CSE is completely different from RyuJIT's CSE which is based on VN.

BruceForstall:
@mikedn I think you're thinking of JIT32 (not public), not the `LEGACY_BACKEND` path which we here refer to as "legacy jit". We sometimes refer to JIT32 or JIT64 (on desktop CLR) as "compat jit".

@JosephTremoulet @briansull Can you answer @parjong 's question?

mikedn:
Ah, legacy backend + new frontend? I had the impression that we don't build that frankenstein. Is this what the current ARM JIT is?

BruceForstall:
> Ah, legacy backend + new frontend? I had the impression that we don't build that frankenstein. Is this what the current ARM JIT is?

Yes, that is the current ARM JIT. Note that there is a RyuJIT/arm32 port ongoing, currently built as protojit.dll (on arm) and protononjit.dll (cross-compiling altjit on Windows x86).

Note also that the x86 `LEGACY_BACKEND` path was recently resurrected with https://github.com/dotnet/coreclr/pull/12184 (as legacyjit.dll), to allow for throughput measurements/comparisons with RyuJIT/x86.

parjong:
@dotnet/arm32-contrib Please take a look

src/jit/morph.cpp
```diff
@@ -8973,6 +8972,7 @@ GenTreePtr Compiler::fgMorphOneAsgBlockOp(GenTreePtr tree)
                     // we don't delete it as a dead store later on.
                     unsigned lclVarNum                = lclVarTree->gtLclVarCommon.gtLclNum;
                     lvaTable[lclVarNum].lvAddrExposed = true;
+#ifndef LEGACY_BACKEND
```
briansull: Why are you removing this call to lvaSetVarDoNotEnregister here?  Setting lvaAddrExposed implies that we also cannot enregister this variable, so we call call lvaSetVarDoNotEnregister with a reason and it sets lvDoNotEnregister

BruceForstall:
>                     GenTree* srcAddr;

The code now looks like:
```
#ifndef LEGACY_BACKEND
                    lvaSetVarDoNotEnregister(lclVarNum DEBUGARG(DNER_AddrExposed));

#else  // LEGACY_BACKEND
                    lvaSetVarDoNotEnregister(lclVarTree->gtLclVarCommon.gtLclNum DEBUGARG(DNER_LocalField));
#endif // LEGACY_BACKEND
```
Don't these lines do precisely the same thing? So, the code can simply be:
```
                    lvaSetVarDoNotEnregister(lclVarNum DEBUGARG(DNER_AddrExposed));
```
?

So, previously, `LEGACY_BACKEND` would NOT set `lvAddrExposed = true`, based on the comment:
```
                    // The source argument of the copyblk can potentially
                    // be accessed only through indir(addr(lclVar))
                    // or indir(lclVarAddr) in rational form and liveness
                    // won't account for these uses. That said,
                    // we have to mark this local as address exposed so
                    // we don't delete it as a dead store later on.
```
but you are enabling it for `LEGACY_BACKEND` as well, so the comment needs to be updated, right? The comment about "rational form" doesn't apply to `LEGACY_BACKEND`, for example.

---
Refers to: src/jit/morph.cpp:8981 in da604d6. [](commit_id = da604d613749aa73700d86c1326a06a00cbe3df6, deletion_comment = False)

BruceForstall:
@CarolEidt Do you recall anything about this case? In particular, w.r.t. the comment I highlighted below that cause RyuJIT only (not `LEGACY_BACKEND`) to set `lvAddrExposed = true`?

parjong:
@BruceForstal The change in line 8975 is not necessary to resolve this issue, but I introduced it as I think that these lines should be aligned with  the changes in line 10346. If it is not necessary, it would be better to exclude the changes in line 8975 from this PR.

CarolEidt:
I am a bit mystified about why this was originally done only for the RyuJIT backend.  I don't currently have access to further past history, but it's been under `#ifndef LEGACY_BACKEND` since the code has been on github. It certainly seems like it should apply to `LEGACY_BACKEND` as well.

CarolEidt:
This code was added as a bug fix nearly 4 years ago, and it wasn't clear to me, based on the discussion of the bug fix, why it would be limited to the RyuJIT backend. The address of the lclVar is clearly being taken. So, I think this is the correct fix.

BruceForstall:

hqueue:
> This code was added as a bug fix nearly 4 years ago, and it wasn't clear to me, based on the discussion of the bug fix, why it would be limited to the RyuJIT backend. 

@CarolEidt Are `Compiler::fgMorphOneAsgBlockOp(GenTreePtr tree)` and `GenTreePtr Compiler::fgMorphCopyBlock(GenTreePtr tree)` fixed for a same bug together at the time?
Or are they separate independet fixes ?

CarolEidt:
> Are `Compiler::fgMorphOneAsgBlockOp(GenTreePtr tree)` and `GenTreePtr Compiler::fgMorphCopyBlock(GenTreePtr tree)` fixed for a same bug together at the time?
Or are they separate independet fixes ?

They were not fixed at the same time. I haven't tracked down the source of the similar code in `fgMorphOneAsgBlockOp()`, but it was made later. I suspect that it also needs to be the same for legacy and non-legacy backend. These inconsistencies may date back to a time when value numbering was only being done in conjunction with RyuJIT backend - though I'm not certain about that.

PR #12238, opened 2017-06-13
Pull CoreCLR build from RHEL7.2 in perf runs

dpodder:
Our Ubuntu perf testing has been creating Linux CoreCLR builds by creating the build locally. This fits the local developer scenario, but doesn't actually test the bits that ship; the 'official' product configuration is generated by building on RHEL 7.2 and packaging those bits for various distros, including Ubuntu.

In order to more accurately measure end-user-perceived performance, we should do the same thing official builds do: build on RHEL7.2, then copy those bits onto other OSes to test perf on those OSes.

This PR makes that happen by structuring Linux perf and throughput jobs as flow jobs, with builds up front on a RHEL VM, before the bits are dropped onto Ubuntu perf test machines.

### DIFF
--- a/perf.groovy
+++ b/perf.groovy
@@ -339,6 +339,11 @@ def static getFullPerfJobName(def project, def os, def isPR) {
         "{ build(params + [PRODUCT_BUILD: b.build.number], '${getFullPerfJobName(project, os, isPR)}') }"
     }
     def newFlowJob = buildFlowJob(Utilities.getFullJobName(project, "perf_linux_flow", isPR, '')) {
+        if (isPR) {
+            parameters {
+                stringParam('BenchviewCommitName', '\${ghprbPullTitle}', 'The name that you will be used to build the full title of a run in Benchview.  The final name will be of the form <branch> private BenchviewCommitName')
+            }
+        }
         buildFlow("""
 // First, build the bits on RHEL7.2
 b = build(params, '${fullBuildJobName}')
@@ -466,6 +471,11 @@ def static getFullThroughputJobName(def project, def os, def isPR) {
         "{ build(params + [PRODUCT_BUILD: b.build.number], '${getFullThroughputJobName(project, os, isPR)}') }"
     }
     def newFlowJob = buildFlowJob(Utilities.getFullJobName(project, "perf_throughput_linux_flow", isPR, '')) {
+        if (isPR) {
+            parameters {
+                stringParam('BenchviewCommitName', '\${ghprbPullTitle}', 'The name that you will be used to build the full title of a run in Benchview.  The final name will be of the form <branch> private BenchviewCommitName')
+            }
+        }
         buildFlow("""
 // First, build the bits on RHEL7.2
 b = build(params, '${fullBuildJobName}')
--- a/perf.groovy
+++ b/perf.groovy
@@ -366,10 +366,32 @@ parallel(
 
 } // isPR
 
+def static getFullThroughputJobName(def os) {
+    return "perf_throughput_${os}"
+}
+
 // Create the Linux/OSX/CentOS coreclr test leg for debug and release and each scenario
 [true, false].each { isPR ->
-    ['Ubuntu14.04'].each { os ->
-        def newJob = job(Utilities.getFullJobName(project, "perf_throughput_${os}", isPR)) {
+    def fullBuildJobName = Utilities.getFullJobName(project, 'perf_throughput_linux_build', isPR)
+    def architecture = 'x64'
+    def configuration = 'Release'
+
+    // Build has to happen on RHEL7.2 (that's where we produce the bits we ship)
+    ['RHEL7.2'].each { os ->
+        def newBuildJob = job(fullBuildJobName) {
+            steps {
+                shell("./build.sh ${architecture} ${configuration}")
+            }
+        }
+        Utilities.setMachineAffinity(newBuildJob, os, 'latest-or-auto')
+        Utilities.standardJobSetup(newBuildJob, project, isPR, "*/${branch}")
+        Utilities.addArchival(newBuildJob, "bin/Product/**")
+    }
+
+    // Actual perf testing on the following OSes
+    def throughputOSList = ['Ubuntu14.04']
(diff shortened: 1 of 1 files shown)


dpodder:
@dotnet-bot test ci please

perf.groovy
```diff
@@ -341,15 +408,18 @@ def static getOSGroup(def os) {
                 }
             }
             def osGroup = getOSGroup(os)
-            def architecture = 'x64'
-            def configuration = 'Release'
             def runType = isPR ? 'private' : 'rolling'
             def benchViewName = isPR ? 'coreclr private \$BenchviewCommitName' : 'coreclr rolling \$GIT_BRANCH_WITHOUT_ORIGIN \$GIT_COMMIT'
 
             steps {
                 shell("bash ./tests/scripts/perf-prep.sh --throughput")
```
brianrob: perf-prep.sh will clone CoreFX and build it, so this should come from the Redhat build.  I believe @adiaaida was going to look into pulling this from binaries instead of building.

perf.groovy
```diff
@@ -260,15 +282,18 @@ def static getOSGroup(def os) {
...
-            def architecture = 'x64'
-            def configuration = 'Release'
             def runType = isPR ? 'private' : 'rolling'
             def benchViewName = isPR ? 'coreclr private \$BenchviewCommitName' : 'coreclr rolling \$GIT_BRANCH_WITHOUT_ORIGIN \$GIT_COMMIT'
 
             steps {
                 shell("./tests/scripts/perf-prep.sh")
                 shell("./init-tools.sh")
-                shell("./build.sh ${architecture} ${configuration}")
+                copyArtifacts(fullBuildJobName) {
```
brianrob: Just making sure I understand the logic here.  Does this just pull from the latest completed RHEL build which happens to be the one we triggered?  You can get a handle to the build that you triggered and then pull the artifacts from that.  I recommend doing this.

Here's an example of how this happens in xunit-performance: https://github.com/Microsoft/xunit-performance/blob/master/netci.groovy#L42

dpodder: Makes sense, thanks for catching that! I'll make the change to match what we do in xunit-performance. The logic was ported from our flow job in optimization, so I think we want to apply a fix to do the same build-number-passing there as well.

perf.groovy
dpodder: Ah, good call, I missed that it was also doing a build and not just downloading bits. It looks like it's already grabbing artifacts on perf, so I just need to modify the throughput flow to archive/download `_/fx/**`, right?

brianrob: Yes, if you want to maintain the "build CoreFX" semantics then you'll want to build them on RHEL and then archive the binaries.

I'm not sure exactly which binaries are needed, but you should be able to find this by looking at what gets used in the throughput scripts, or by asking @adiaaida.

perf.groovy
brianrob: Yes, I think that's a good plan - makes sure that you don't have any race conditions.

perf.groovy
DrewScoggins: I don't think that we want that to come from the RedHat build.  We just do that building of CoreFX to get binaries to feed into the throughput tests.  If we change the commit that we are grabbing from we will need to make sure that we update baselines on those tests.

brianrob: I don't think building on RHEL should have an impact should it?  Our goal is to be as close to shipping as possible, which happens on RHEL.  Ideally, we don't even build CoreFX separately - we use the test layout logic that gets used in functional runs.

As far as the baseline, we did actually just update it last week because it was old.  What we should do to keep from having the stale baseline problem is take a set of binaries and make them the "curated set of binaries" to be used on throughput and don't depend on using framework binaries.  That way, we don't have to worry about any framework binaries that get treated specially or types that might get forwarded into System.Private.CoreLib since S.P.C will continue to move forward alongside the runtime used to run TP tests.

adiaaida: This is just building the test artifacts. I don't think this needs to be done in the RHEL build run, and should probably stay with the rest of the test run. But Brian is right, I am  planning on removing this and using the same test binaries that are used on Windows (from a package of some sort).

dpodder:
@dotnet-bot test ci please

dpodder:
@dotnet-bot test ci please

perf.groovy
dpodder: Thinking it over--I think I'd like to avoid changing this for now, and keep it as currently written in this PR. As long as we don't actually execute any of the corefx binaries built here, where they get built shouldn't matter to perf. I'd be inclined to make the change if it was to be kept long-term, but I'd rather defer to @adiaaida's planned fix instead.

dpodder:
@dotnet-bot test ci please

dpodder:
@dotnet-bot test ci please

dpodder:
@dotnet-bot test ci please

dpodder:
@dotnet-bot test ci please

dpodder:
@brianrob @adiaaida @jorive All testing shows green now :smile: Once CI passes, this is ready for a squash-and-merge.

FYI, I made a few more substantive changes mixed in with the various silly fixes--let me know if you have any concerns about them:
* Add `verbose` to the CoreCLR build. This gives us Clang/LLVM command lines in the logs, so we'll be able to look back at a particular run in Jenkins and confirm how it was built (PGO/nonPGO), repro build breaks more easily, etc.
* I had to add extra bits (PAL/native tests) to be archived from the bin folder for "perf" runs; they are not necessary, and therefore left out of, "throughput" build archiving. (Good call @adiaaida for hinting that I may have to do this!)
* I modified the flow jobs to include and forward the `BenchViewCommitName` param, so that manually triggered runs' titles can still be edited like before.

perf.groovy
adiaaida: I think that my change would probably to have another job, on Windows, that does the nuget install step that we do on Windows now, to pull down the TP benchmarks package, archive the binaries, and then have the test job grab the binaries from that leg. It would be in parallel to the build.  OR someone can teach me how the build system works so we could have it pull down the package somehow.

PR #12245, opened 2017-06-13
Replace System.Private.CoreLib with new version in TP jobs

adiaaida:
SPC and the jit/coreclr are tightly coupled. With the changes to
crossgen into the dll, rather than the ni.dll, the throughput jobs are
now using the non-crossgen'd version of SPC on TP runs, which led to a
regression in TP on Linux because our comparison was not the same. This
change updates testing to replace the System.Private.CoreLib in the
throughput directory with the System.Private.CoreLib built by the
coreclr build.

### DIFF
--- a/tests/scripts/run-throughput-perf.py
+++ b/tests/scripts/run-throughput-perf.py
@@ -301,6 +301,9 @@ def main(args):
     crossgen_path = os.path.join(bin_path,crossgen)
     jit_path = os.path.join(bin_path, jit)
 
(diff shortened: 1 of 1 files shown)


adiaaida:
@dotnet-bot test Ubuntu14.04 throughput please

dpodder:
The failing perf test leg should be clean once CI picks up the perf.groovy changes I just merged in #12238. I also need to fix doing optimized builds on Ubuntu14.04 now that count merging *actually* does something, but that's a separate bug.

adiaaida:
@dotnet-bot help

PR #12248, opened 2017-06-13
Skip PGO, and warn, if clang is too old

dpodder:
Clang/LLVM cannot merge profile data generated by a newer version of itself. Training currently requires 3.6, but we don't want to completely break the build for 3.5. I'm adding a version check to allow non-PGO release builds on 3.5, along with a warning message. I'm also updating the LTO detection to print a warning message if it causes PGO to be disabled.


dpodder:
@janvorli PTAL - This should be a small and simple change, I'd also like to take it to shiproom for 2.0 once it gets merged.

@swgillespie This change will let Ubuntu 14.04 release builds pass again, albeit with a warning message. Getting fully-optimized release builds requires clang3.6 or later, but we need to make sure it's available on the VM images first.

swgillespie:
Awesome, thank you!

@dotnet-bot test Ubuntu Release longgc

pgosupport.cmake
```diff
@@ -42,10 +42,16 @@ function(add_pgo TargetName)
                 set_property(TARGET ${TargetName} APPEND_STRING PROPERTY LINK_FLAGS_RELWITHDEBINFO " /LTCG /USEPROFILE:PGD=${ProfilePath}")
             else(WIN32)
                 if(UPPERCASE_CMAKE_BUILD_TYPE STREQUAL RELEASE OR UPPERCASE_CMAKE_BUILD_TYPE STREQUAL RELWITHDEBINFO)
-                    if(HAVE_LTO)
-                        target_compile_options(${TargetName} PRIVATE -flto -fprofile-instr-use=${ProfilePath} -Wno-profile-instr-out-of-date)
-                        set_property(TARGET ${TargetName} APPEND_STRING PROPERTY LINK_FLAGS " -flto -fuse-ld=gold -fprofile-instr-use=${ProfilePath}")
-                    endif(HAVE_LTO)
+                    if(NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 3.6)
```
janvorli: I wonder - do we use PGO on OSX too? I am asking since the clang version numbering on OSX is completely different from other platforms. For example, on my current Apple machine, it reports "8.1.0.8020042".

dpodder: Good catch on OSX versioning of Clang.

At the moment, we don't produce profile data on/for OSX, so this code path will not execute at all unless on Linux (the `if(EXISTS ${ProfilePath})` check can only possibly return true on Windows and Linux). This will change if we decide to produce packages for OSX in the future, at which point OSX would just treat Clang as >= 3.6 and assume PGO support.

I'd prefer to keep this as-is for now, since the failure case causes a build break later on, and we'll catch it when testing packages with OSX support. (I'm also not aware of plans to add OSX profdata yet.) If we do, we can refactor the logic to consider OS. Does that seem reasonable?

janvorli: Ok, that makes sense.

PR #12253, opened 2017-06-13
Don't close passed in FileStream handles in constructor

JeremyKuhne:
As the finalizer will run when throwing we need to be
careful to make sure we don't put the passed-in handle
in the member field until we're sure we've succeeded.

Additionally we need to make sure any wrapping SafeHandle
we create on a passed in IntPtr isn't disposed.

See related https://github.com/dotnet/corefx/issues/20797.

### DIFF
--- a/src/mscorlib/shared/System/IO/FileStream.Unix.cs
+++ b/src/mscorlib/shared/System/IO/FileStream.Unix.cs
@@ -205,6 +205,8 @@ namespace System.IO
                 // Lazily-initialize whether we're able to seek, tested by seeking to our current location.
                 _canSeek = Interop.Sys.LSeek(fileHandle, 0, Interop.Sys.SeekWhence.SEEK_CUR) >= 0;
             }
+
+            return _canSeek.Value;
         }
 
         private long GetLengthInternal()
--- a/src/mscorlib/shared/System/IO/FileStream.Unix.cs
+++ b/src/mscorlib/shared/System/IO/FileStream.Unix.cs
@@ -189,7 +189,10 @@ namespace System.IO
         public override bool CanSeek => CanSeekCore(_fileHandle);
 
         /// <summary>Gets a value indicating whether the current stream supports seeking.</summary>
-        /// <remarks>Separated out of CanSeek to enable making non-virtual call to this logic.</remarks>
+        /// <remarks>
+        /// Separated out of CanSeek to enable making non-virtual call to this logic.
+        /// We also pass in the file handle to allow the constructor to use this before it stashes the handle.
+        /// </remarks>
         private bool CanSeekCore(SafeFileHandle fileHandle)
         {
             if (fileHandle.IsClosed)
--- a/src/mscorlib/shared/System/IO/FileStream.Unix.cs
+++ b/src/mscorlib/shared/System/IO/FileStream.Unix.cs
@@ -105,15 +105,12 @@ namespace System.IO
         }
 
         /// <summary>Initializes a stream from an already open file handle (file descriptor).</summary>
-        /// <param name="handle">The handle to the file.</param>
-        /// <param name="bufferSize">The size of the buffer to use when buffering.</param>
-        /// <param name="useAsyncIO">Whether access to the stream is performed asynchronously.</param>
-        private void InitFromHandle(SafeFileHandle handle)
+        private void InitFromHandle(SafeFileHandle handle, FileAccess access, bool useAsyncIO)
         {
-            if (_useAsyncIO)
+            if (useAsyncIO)
                 _asyncState = new AsyncState();
 
-            if (CanSeekCore) // use non-virtual CanSeekCore rather than CanSeek to avoid making virtual call during ctor
+            if (CanSeekCore(handle)) // use non-virtual CanSeekCore rather than CanSeek to avoid making virtual call during ctor
                 SeekCore(0, SeekOrigin.Current);
         }
 
@@ -189,25 +186,21 @@ namespace System.IO
         }
 
         /// <summary>Gets a value indicating whether the current stream supports seeking.</summary>
-        public override bool CanSeek => CanSeekCore;
+        public override bool CanSeek => CanSeekCore(_fileHandle);
 
         /// <summary>Gets a value indicating whether the current stream supports seeking.</summary>
         /// <remarks>Separated out of CanSeek to enable making non-virtual call to this logic.</remarks>
-        private bool CanSeekCore
+        private bool CanSeekCore(SafeFileHandle fileHandle)
         {
-            get
+            if (fileHandle.IsClosed)
             {
-                if (_fileHandle.IsClosed)
-                {
-                    return false;
-                }
+                return false;
+            }
 
-                if (!_canSeek.HasValue)
-                {
-                    // Lazily-initialize whether we're able to seek, tested by seeking to our current location.
-                    _canSeek = Interop.Sys.LSeek(_fileHandle, 0, Interop.Sys.SeekWhence.SEEK_CUR) >= 0;
-                }
-                return _canSeek.Value;
+            if (!_canSeek.HasValue)
+            {
+                // Lazily-initialize whether we're able to seek, tested by seeking to our current location.
+                _canSeek = Interop.Sys.LSeek(fileHandle, 0, Interop.Sys.SeekWhence.SEEK_CUR) >= 0;
             }
         }
 
--- a/src/mscorlib/shared/System/IO/FileStream.cs
+++ b/src/mscorlib/shared/System/IO/FileStream.cs
@@ -136,21 +136,7 @@ namespace System.IO
             _exposedHandle = true;
             _bufferLength = bufferSize;
 
-#if DEBUG
-            bool hadBinding = handle.ThreadPoolBinding != null;
-
-            try
-            {
-#endif
-                InitFromHandle(handle, access, isAsync);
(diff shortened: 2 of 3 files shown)


src/mscorlib/shared/System/IO/FileStream.Unix.cs
```diff
@@ -189,25 +186,21 @@ private static Interop.Sys.OpenFlags PreOpenConfigurationFromOptions(FileMode mo
         }
 
         /// <summary>Gets a value indicating whether the current stream supports seeking.</summary>
-        public override bool CanSeek => CanSeekCore;
+        public override bool CanSeek => CanSeekCore(_fileHandle);
 
         /// <summary>Gets a value indicating whether the current stream supports seeking.</summary>
         /// <remarks>Separated out of CanSeek to enable making non-virtual call to this logic.</remarks>
-        private bool CanSeekCore
+        private bool CanSeekCore(SafeFileHandle fileHandle)
```
stephentoub: It'd be good to add a comment about why this is being passed in, so that someone in the future doesn't "simplify" it back

JeremyKuhne: Done, thanks.

src/mscorlib/shared/System/IO/FileStream.cs
```diff
@@ -81,8 +81,31 @@ public FileStream(IntPtr handle, FileAccess access, bool ownsHandle, int bufferS
...
+            catch
+            {
+                // We don't want to take ownership of closing passed in handles
+                // *unless* the constructor completes successfully.
+                GC.SuppressFinalize(safeHandle);
+
+                // This would also prevent Close from being called, but is unnecessary
+                // as we've removed the object from the finalizer queue.
+                //
+                // safeHandle.SetHandleAsInvalid();
```
stephentoub: Alternatively you could probably do:
```C#
catch
{
    if (ownsHandle) safeHandle.SetHandleAsInvalid();
    safeHandle.Dispose();
    throw;
}
```

src/mscorlib/shared/System/IO/FileStream.cs
```diff
@@ -110,13 +133,36 @@ public FileStream(SafeFileHandle handle, FileAccess access, int bufferSize, bool
...
+
+            try
+            {
+#endif
+                InitFromHandle(handle, access, isAsync);
+#if DEBUG
+            }
+            catch
+            {
+                Debug.Assert(hadBinding || handle.ThreadPoolBinding == null, "We should never error out with a ThreadPoolBinding we've added");
```
jkotas: This should be in the Windows specific impl

src/mscorlib/shared/System/IO/FileStream.cs
```diff
@@ -81,8 +81,31 @@ public FileStream(IntPtr handle, FileAccess access, bool ownsHandle, int bufferS
...
+                // as we've removed the object from the finalizer queue.
+                //
+                // safeHandle.SetHandleAsInvalid();
+                throw;
+            }
+
+            _access = access;
+            _useAsyncIO = isAsync;
+
+            // As the handle was passed in, we must set the handle field at the very end to
```
jkotas: I do not understand why you are doing this. I think it would be simpler and easier to understand if this is implemented like this:

```
public FileStream(IntPtr handle, FileAccess access, bool ownsHandle, int bufferSize, bool isAsync) 
{
  try 
  { 
       ValidateAndInitFromHandle(safeHandle, access, bufferSize, isAsync); 
   }
   catch
   {
       // We don't want to take ownership of closing passed in handles 
       // *unless* the constructor completes successfully. 
       GC.SuppressFinalize(safeHandle);
       _fileHandle = null;
   }
}
```

src/mscorlib/shared/System/IO/FileStream.cs
JeremyKuhne: Figured setting the flag was more direct? I'm not hard set either way.

src/mscorlib/shared/System/IO/FileStream.cs
JeremyKuhne: Those fields are set as readonly. I thought slightly awkward was better than removing the readonly.

src/mscorlib/shared/System/IO/FileStream.cs
JeremyKuhne: Whoops- thanks :)

src/mscorlib/shared/System/IO/FileStream.cs
jkotas: Comment would be useful. It was not obvious to me that all the awkwardness is just because of readonly.

JeremyKuhne: Will do

JeremyKuhne:

src/mscorlib/shared/System/IO/FileStream.cs
danmosemsft: Why the Dispose in @stephentoub 's suggestion? Doesn't SetHandleAsInvalid do the job?

stephentoub: Because if we do own the handle we want to make sure it's released.

src/mscorlib/shared/System/IO/FileStream.cs
```diff
@@ -110,13 +135,24 @@ public FileStream(SafeFileHandle handle, FileAccess access, int bufferSize, bool
...
 
-            InitFromHandle(handle);
+            InitFromHandle(handle, access, isAsync);
+        }
+
+        public FileStream(SafeFileHandle handle, FileAccess access, int bufferSize, bool isAsync)
+        {
+            ValidateAndInitFromHandle(handle, access, bufferSize, isAsync);
+
+            // Note: Cleaner to set the following fields in ValidateAndInitFromHandle,
```
sharwell: 💭 `out` parameters would be an option right?

src/mscorlib/shared/System/IO/FileStream.cs
JeremyKuhne: > Because if we do own the handle we want to make sure it's released.

The assertion this change is making is that we don't own the handle if we haven't finished constructing. Dispose() won't close the handle if it is set as invalid.

stephentoub: > The assertion this change is making is that we don't own the handle if we haven't finished constructing. Dispose() won't close the handle if it is set as invalid.

I'm not convinced that's valid for the IntPtr ctor.  The calling code is handing off a raw IntPtr to the FileStream saying "it's now your responsibility to close this, nothing else will".

That's very different from the ctors that take a SafeHandle, where the SafeHandle itself knows whether it owns the IntPtr, and where worst case it'll have its finalizer handle the cleanup.

src/mscorlib/shared/System/IO/FileStream.cs
JeremyKuhne: > out parameters would be an option right?

You could, but it would probably require a comment to explain it. You'd still have to set the field to the passed-in value, which would require passing those three arguments twice (one argument for the input and one for the output. I added the comment per @jkotas feedback.

src/mscorlib/shared/System/IO/FileStream.cs
JeremyKuhne: > The calling code is handing off a raw IntPtr to the FileStream saying "it's now your responsibility to close this, nothing else will".

My argument on this one is that people don't expect objects that aren't fully constructed to have ownership of closing passed in objects, period.

> That's very different from the ctors that take a SafeHandle, where the SafeHandle itself knows whether it owns the IntPtr, and where worst case it'll have its finalizer handle the cleanup.

I agree that there is a safety net there that isn't in place with the IntPtr, but it feels off to me to not be consistent. That said, we have depreciated the IntPtr constructors... @jkotas, @terrajobst, @weshaggard: Do you have any feedback on the right pattern here? There are likely other places where we have destructors that need hardened against constructor exceptions- I'd like to vet the approach.

stephentoub: > My argument on this one is that people don't expect objects that aren't fully constructed to have ownership of closing passed in objects, period.

Do you have examples from real usage this is based on?  Otherwise how do you know this change isn't introducing a leak?

JeremyKuhne: >  Otherwise how do you know this change isn't introducing a leak?

You don't know. I suppose if we're closing it deliberately and immediately now we eliminate the other part of the problem, which is random timing on the closing of the IntPtr (which, incidentally, the SafeFileHandle case would still be random if you aren't handling your errors).

I'm ok with "we never close passed in SafeHandles if we don't finish constructing, but we will close IntPtr handles as there is no fallback for cleanup if you don't handle exceptions properly" if we have general consensus. I just want to make sure we do have consensus as I'm still on the line here.

JeremyKuhne: To be clear: The "I've got a handle, try this, if it fails, try something else." scenario is the one I'm particularly concerned about. There are a number of indirect entries into creating a file stream- it isn't too outlandish to expect such a scenario.

I really want to make sure that works with SafeHandle usage, but I can understand deviating with IntPtr, even though I'm a bit torn on it.

JeremyKuhne:
With additional internal discussions we've agreed that taking ownership of a handle should never happen until construction successfully completes.

Issue #12146, opened 2017-06-07
crossgen warning - Vararg calling convention not supported. while compiling method String.Concat


briansull:
We should either Fix or Hide this warning

To Repro:

crossgen System.Private.CoreLib.dll

briansull:
Fixed with #12255

PR #12255, opened 2017-06-13
Removed the unused vararg version of String Concat

briansull:
The __arglist version of varargs is not supported on CoreCLR

Fixes issue #12146 - crossgen warning - Vararg calling convention not supported.

Added a new test case for String Concat

### DIFF
--- a/tests/src/JIT/Directed/Misc/Concat/ConcatTest.cs
+++ b/tests/src/JIT/Directed/Misc/Concat/ConcatTest.cs
@@ -1,0 +1,66 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+internal class ConcatTest
+{
+    static string strA = "A";
+    static string strB = "B";
+    static string strC = "C";
+    static string strD = "D";
+    static string strE = "E";
+
+    static string strAB      = "AB";
+    static string strABC     = "ABC";
+    static string strABCD    = "ABCD";
+    static string strABCDE   = "ABCDE";
+    static string strABCDx2  = "ABCDABCD";
+
+    static int iReturn = 100;
+
+    static public int Main()
+    {
+        iReturn = 100;
+        try
+        {
+            string result;
+
+            result = string.Concat(strA, strB);
+            CheckResult(result, strAB);
+
+            result = string.Concat(strA, strB, strC);
+            CheckResult(result, strABC);
+
+            result = string.Concat(strA, strB, strC, strD);
+            CheckResult(result, strABCD);
+
+            result = string.Concat(strA, strB, strC, strD, strE);
+            CheckResult(result, strABCDE);
+
+            result = string.Concat(strA, strB, strC, strD, strA, strB, strC, strD);
+            CheckResult(result, strABCDx2);
+
+            Console.WriteLine("Passed all tests.");
+        }
+        catch (Exception e)
+        {
+            Console.WriteLine("Failed {0}", e.StackTrace);
+
+            iReturn = 99;
+        }
+
+        return iReturn;
+    }
+
+    static void CheckResult(string result, string expected)
+    {
+        if (result != expected)
+        {
+            Console.WriteLine("FAILED: result was '" + result +
+                              "', expected was '" + expected + "'");
+            iReturn++;
+        }
+    }
+}
--- a/tests/src/JIT/Directed/Misc/Concat/ConcatTest.csproj
+++ b/tests/src/JIT/Directed/Misc/Concat/ConcatTest.csproj
@@ -29,7 +29,7 @@
     <DefineConstants>$(DefineConstants);CORECLR</DefineConstants>
   </PropertyGroup>
   <ItemGroup>
-    <Compile Include="gettypetypeofmatrix.cs" />
+    <Compile Include="ConcatTest.cs" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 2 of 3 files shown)


tests/src/JIT/Directed/Misc/Concat/ConcatTest.cs
```diff
@@ -0,0 +1,66 @@
...
+    static int iReturn = 100;
+
+    static public int Main()
+    {
+        iReturn = 100;
+        try
+        {
+            string result;
+
+            result = string.Concat(strA, strB);
```
jkotas: What's the purpose of this test?

briansull: I wanted to see the code generated for Concat(strA, strB, strC, strD) to confirm that the vararg version isn't being used.

briansull: Current when we try to crossgen a vararg method such as this one we hit this codepath.
Which causes crossgen to issue this warning. 
We also would hit this code path when using the JIT.

vm\jitinterface.cpp line 520
```
        if ((isCallConv(sigRet->callConv, IMAGE_CEE_CS_CALLCONV_VARARG)) ||
            (isCallConv(sigRet->callConv, IMAGE_CEE_CS_CALLCONV_NATIVEVARARG)))
        {
            // This signature corresponds to a method that uses varargs, which are not supported.
             COMPlusThrow(kInvalidProgramException, IDS_EE_VARARG_NOT_SUPPORTED);
        }
```

jkotas: > I wanted to see the code generated for Concat(strA, strB, strC, strD) to confirm that the vararg version isn't being used.

That's fine. But does it warrant to add it as a new test?

There are number of other tests that exercise `String.Concat`.

jkotas: Also, this test would not ever call the `__vararg` version even if it was there.

briansull: I didn't see any in the JIT tests and there were really only an handful of them else where:
(I found some here: src\CoreMangLib\cti\system\string)
But none of these tests actually tried calling Concat with more than 3 arguments.

It looked to me that an older version of the C# compiler would possibly create a call to the vararg version automatically when 4 or more arguments were passed to Concat.

As it is now the C# compiler transforms these calls into this:
Concat( new String[] = {"A", "B", "C", "D" } )

tests/src/JIT/Directed/Misc/Concat/ConcatTest.csproj
```diff
@@ -0,0 +1,41 @@
...
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+  </PropertyGroup>
+  <ItemGroup>
+    <CodeAnalysisDependentAssemblyPaths Condition=" '$(VS100COMNTOOLS)' != '' " Include="$(VS100COMNTOOLS)..\IDE\PrivateAssemblies">
+      <Visible>False</Visible>
+    </CodeAnalysisDependentAssemblyPaths>
+  </ItemGroup>
+  <PropertyGroup>
+    <!-- Set to 'Full' if the Debug? column is marked in the spreadsheet. Leave blank otherwise. -->
+    <DebugType></DebugType>
```
rartemev: Could you please specify DebugType, as "None" for example, because it will broke build tests on Linux otherwise.

rartemev: Could you please specify DebugType, as "None" for example, because it will broke build tests on Linux otherwise.

tests/src/JIT/Directed/Misc/Concat/ConcatTest.cs
jkotas: > there were really only an handful of them elsewhere

Note that majority of the framework test cases live in corefx repo. We do not have, nor trying to have, a comprehensive API test coverage in CoreCLR repo.

> It looked to me that an older version of the C# compiler would possibly create a call to the vararg version automatically when 4 or more arguments were passed to Concat

The C# compiler would not ever call the vararg version. You would have to use the undocumented `__arglist` keyword.

tests/src/JIT/Directed/Misc/Concat/ConcatTest.csproj
briansull: I  copied this verbatim from a sibling test in the Misc folder, so it would also have this issue.

rartemev: Probably it is excluded from build on Linux.

briansull: I updated both of the csproj  files with your request.

tests/src/JIT/Directed/Misc/Concat/ConcatTest.cs
briansull: Since this is an automatic C# compiler thing, I don't think that the API test coverage in CoreCLR repo would create a test case with source code to call to Concat directly with 5 arguments since it doesn't see such an API when it reflects over the API space for String.

jkotas: > Since this is an automatic C# compiler thing,

What is an automatic C# compiler thing?

briansull: When I write: `Concat (a,b,c,d,e)`
and it generates code to call `Concat (string[])`

PR #12267, opened 2017-06-14
Fix GetHashCode for ordinal comparer

ViktorHofer:
Fixes regression introduced here: https://github.com/dotnet/coreclr/pull/12215


danmosemsft:
@ViktorHofer buildable/product changes must have a CR unless it's work stoppage (like a build break). Stephen starts early EST and can do such a review fast in future :)

ViktorHofer:
As you know the reason I don't need to explain anything... Thanks for pointing it out, this won't happen again.

danmosemsft:
No problem at all thanks @ViktorHofer :)

PR #12281, opened 2017-06-14
Fix NullReferenceException thrown when using CurrentCulture

tarekgh:
CurrentCulture used the static field s_userDefaultCulture which used to store the current culture. We always ensure this field is initialized all the time. When we introduced the v2.0 added APIs, we have added ClearCachedData(). This method was setting s_userDefaultCulture to null which can cause any other thread to throw this NullReferenceException if it is using s_userDefaultCulture. Here is example of such stack

System.NullReferenceException : Object reference not set to an instance of an object.
Stack Trace :
   at System.Globalization.NumberFormatInfo.get_CurrentInfo()
   at System.Double.ToString(String format, IFormatProvider provider)

### DIFF
--- a/src/mscorlib/src/System/Globalization/CultureInfo.cs
+++ b/src/mscorlib/src/System/Globalization/CultureInfo.cs
@@ -967,7 +967,7 @@ namespace System.Globalization
 
         public void ClearCachedData()
         {
-            s_userDefaultCulture = null;
+            Init(); // reset the default culture values
 
             RegionInfo.s_currentRegionInfo = null;
             #pragma warning disable 0618 // disable the obsolete warning 


tarekgh:
https://github.com/dotnet/corefx/issues/20835

tarekgh:
@danmosemsft could you please have a look? Thanks

danmosemsft:
LGTm. As far as I can see the other fields set to null in that method are OK, eg., s_currentRegionInfo has some volatile get/set scheme.

tarekgh:
> As far as I can see the other fields set to null in that method are OK, eg., s_currentRegionInfo has some volatile get/set scheme.

Yes we are ok because these fields are stored in a temp variable before accessing it. here is example 

```C#
        public static RegionInfo CurrentRegion
        {
            get
            {
                RegionInfo temp = s_currentRegionInfo;
                if (temp == null)
                {
                    temp = new RegionInfo(CultureInfo.CurrentCulture._cultureData);

                    // Need full name for custom cultures
                    temp._name = temp._cultureData.SREGIONNAME;
                    s_currentRegionInfo = temp;
                }
                return temp;
            }
        }
```

danmosemsft:
We should port this to 2.0 I think.

Issue #12152, opened 2017-06-07
crossgen warning - Target-dependent SIMD vector types may not be used with ngen.


briansull:
We should either Fix or Hide this warning

To Repro:

crossgen /Platform_Assemblies_Paths . System.Numerics.Vectors.dll 
crossgen /Platform_Assemblies_Paths . Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.dll

briansull:
Fixed with #12287

PR #12287, opened 2017-06-15
Crossgen - Disable the Target-dependent SIMD vector types warning

briansull:
When a SIMD type is referenced during CEEPreloader instead of loading it
we throw a type load exception with a unique resource ID and message.
This commit changed the ICorCompileDataStore interface to allow us to pass
the unique resource ID over to the zapper from the VM.

The Zapper can then demote certain resource ID warnings down to the
lowest informational level, so that they aren't printed out during crossgen.
With this commit we currently are demoting IDS_EE_SIMD_NGEN_DISALLOWED

### DIFF
--- a/src/inc/corcompile.h
+++ b/src/inc/corcompile.h
@@ -1378,9 +1378,12 @@ class ICorCompileDataStore
     // Returns ZapImage
     virtual ZapImage * GetZapImage() = 0;
 
-    // Reports an error during preloading.  Return the error code to propagate,
-    // or S_OK to ignore the error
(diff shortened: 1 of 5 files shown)


briansull:
Fixes #12152

@jkotas PTAL

jkotas:

briansull:
@CarolEidt FYI, this will eliminate the 210 warnings that we get when we 
crossgen System.Numerics.Vectors.dll

CarolEidt:
@briansull thanks for doing this. Sorry for the delay in reviewing.

PR #12292, opened 2017-06-15
Disable Portable RID build for Tizen

hqueue:
We have to disable portable RID build for Tizen armel regardless of `clang` option.

/cc: @gkhanna79 @jyoungyun @seanshpark

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -853,15 +853,15 @@ if [[ $__ClangMajorVersion == 0 && $__ClangMinorVersion == 0 ]]; then
             __ClangMinorVersion=6
         fi
 
-        if [[ "$__BuildArch" == "armel" ]]; then
-            # Armel cross build is Tizen specific and does not support Portable RID build
-            __PortableBuild=0
-        fi
-
     else
         __ClangMajorVersion=3
         __ClangMinorVersion=5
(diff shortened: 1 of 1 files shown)


seanshpark:
This is needed when clang option is provided to build.sh

build.sh
```diff
@@ -853,17 +853,17 @@ if [[ $__ClangMajorVersion == 0 && $__ClangMinorVersion == 0 ]]; then
...
-            __PortableBuild=0
-        fi
-
     else
         __ClangMajorVersion=3
         __ClangMinorVersion=5
     fi
 fi
 
+if [[ "$__BuildArch" == "armel" ]]; then
```
gkhanna79: Instead of this, why dont you pass "-portablebuild=false" to the build script for armel  build?

hqueue: Yes, we can use that option explicitly.

However it seems that currently this option is turned off only when we don't use `clangX.Y` option and we want to make this option not dependent on `clangX.Y` option.

Furthermore  if you don't mind, it would be convenient to build coreclr with simple build command, i.e. `build.sh armel cross`, because this option is not supported for Tizen yet.

PR #12297, opened 2017-06-15
Avoid duplicated computations with DateTime.GetDatePart

stephentoub:
DateTime.GetDatePart is showing up in some workloads as a hot spot (it's used to get the Year/Month/Day of a DateTime).  While this PR doesn't improve the perf of an individual access, there are a variety of places within corelib where all three of Year/Month/Day are accessed together, which results in three separate calls to GetDatePart and ends up redoing most of the same computations each time.  This PR adds an alternate version of GetDatePart that allows all three to be computed and returned rather than just returning one of them, and then uses that helper in various places throughout corelib where all three are needed.  This ends up improving the throughput of various operations, e.g. ToString("o"), ToString("r"), AddMonths, AddYears, etc. by 5-15% depending on the operation.  I also noticed that GregorianCalendar had its own copy of this code, and in an internal virtual method, but nothing overrides this, so I simply deleted the duplicate code and changed GregorianCalendar to just use the exposed properties (Year/Month/Day) on DateTime.

cc: @tarekgh

(@Petermarcu, FYI, as this modifies the FastFormatRfc1123 and FastFormatRoundtrip functions you previously added.)

### DIFF
--- a/src/mscorlib/shared/System/DateTime.cs
+++ b/src/mscorlib/shared/System/DateTime.cs
@@ -453,9 +453,7 @@ namespace System
         {
             if (months < -120000 || months > 120000) throw new ArgumentOutOfRangeException(nameof(months), SR.ArgumentOutOfRange_DateTimeBadMonths);
             Contract.EndContractBlock();
-            int y = GetDatePart(DatePartYear);
-            int m = GetDatePart(DatePartMonth);
-            int d = GetDatePart(DatePartDay);
+            GetDatePart(out int y, out int m, out int d);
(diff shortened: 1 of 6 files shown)


tarekgh:
LGTM. thanks @stephentoub for optimizing this part and removing the duplication.

Petermarcu:
I assume we're already using C#7 features other places?

stephentoub:
> I assume we're already using C#7 features other places?

Yes

Petermarcu:
LGTM, thanks for doing it.

PR #12313, opened 2017-06-15
Remove static for function pointer

kyulee1:
qsort with static compare functions crashes with ALTJIT.
It appears due to _guard_check_icall_fptr in Windows 10.
Simply dropping static keyword works.

### DIFF
--- a/src/jit/compiler.cpp
+++ b/src/jit/compiler.cpp
@@ -6162,12 +6162,12 @@ void Compiler::compInitVarScopeMap()
     }
 }
 
-static int __cdecl genCmpLocalVarLifeBeg(const void* elem1, const void* elem2)
+int __cdecl genCmpLocalVarLifeBeg(const void* elem1, const void* elem2)
 {
     return (*((VarScopeDsc**)elem1))->vsdLifeBeg - (*((VarScopeDsc**)elem2))->vsdLifeBeg;
 }
 
-static int __cdecl genCmpLocalVarLifeEnd(const void* elem1, const void* elem2)
+int __cdecl genCmpLocalVarLifeEnd(const void* elem1, const void* elem2)
 {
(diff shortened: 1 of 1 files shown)


kyulee1:

CarolEidt:
I take it that this is addressing what appears to be a compiler bug. If so, a bug should be filed, and comments should be added here accordingly, and an issue filed to remove these when possible. It would be a shame to pass an unnecessary this parameter indefinitely, especially to a sort.

kyulee1:
I don't think it's a compiler but I'm not sure either what exactly the problem is with this Window's feature within ntdll --  this could be a user error.
Introducing #ifdef to guard for Windows seems overkill since this is only instance where we use static C compare function for qsort within RyuJit. To align with other cases, I would simply drop them.

jkotas:
> I don't think it's a compiler 

It does look like a compiler bug...

kyulee1:
I got a cutdown binary which also crashes intermittently in VS2015. It appears it's related to /guard:cf and maybe due to the changed way or bug to export guard CF function table. Using VS2017 doesn't cause this problem. I'm asking around whether this is a known issue in between VS2015 vs VS2017. Regardless it's not worth filing a bug but we may force to use VS2017 for coreclr build, or enforce /guard:cf- to disable the feature for the future bug like this.

BruceForstall:
@CarolEidt in this case, `static` is just about file visibility: there is no `this` pointer to these functions.

CarolEidt:
> in this case, static is just about file visibility: there is no this pointer to these functions.

Of course; in that case it seems perfectly reasonable to make this change.

kyulee1:
FYI. I was talking to native compiler folks, and finally confirmed that this is an existing incremental link issue that drops static function in CFG function table. So, this workaround is valid. The related bug is being tracked internally.

PR #12314, opened 2017-06-15
[Local GC] Move handle creation/deletion profiler callbacks out of the GC

swgillespie:
This PR adds `DiagHandleCreated` and `DiagHandleDestroyed` to `GCToEEInterface`, where the handle table can call into the EE when a handle is created or destroyed. This allows us to move some code referring to profiling internals out of the GC and into the EE.

One thing that I am unsure about is the use of these two functions in `HndCreateHandles` and `HndDestroyHandles`, which create or destroy handles in bulk. With this PR it is now more expensive to report multiple handles since the existing implementation loops over a list of handles before pinning the profiler, while this implementation loops on the GC side and makes a (possibly indirect) call to the EE for every handle, which pins and unpins the profiler every time. If it's common to create and destroy large amounts of handles using these handles, this approach may not be good. Feedback welcomed on that!


swgillespie:
@Maoni0 @adityamandaleeka @jkotas PTAL, thoughts?

jkotas:
The VM can call the profiling interface when it after it creates the handle, or before it destroys it. I do not think it needs to be a callback.

jkotas:
Also, it is odd to have handle-related methods here where all handle related-methods are on the handle manager interface. But that's a secondary point.

swgillespie:
Ok! I'll see what I can do with that.

jkotas:
> If it's common to create and destroy large amounts of handles using these handles, this approach may not be good. 

These methods do not seem to be used at all. You can put them under `#if 0` at least.

swgillespie:
In the last three commits I went with what Jan suggested and moved the profiler callback to places in the VM where handles are created. I also created a new file, `gchandleutilities.cpp`, where I put some handle-specific code (the actual profiler callback) and moved some handle code that was previously in `gcheaputilities.cpp` to it.

I did a search of the Desktop codebase for `HndCreateHandles` and `HndDestroyHandles` and did not see any uses their either, so I removed it. I'd be happy to put it under `#if 0` if that's better.

src/vm/gchandleutilities.h
```diff
@@ -441,6 +480,7 @@ inline void DestroyWinRTWeakHandle(OBJECTHANDLE handle)
     }
 
     GCHandleUtilities::GetGCHandleManager()->DestroyHandleOfType(handle, HNDTYPE_WEAK_WINRT);
+    DiagHandleDestroyed(handle);
```
jkotas: You should call  these before the handle is destroyed.

swgillespie: Will do!

Maoni0:
it would be a bit nicer to not have to duplicate the calls to DiagHandleCreated/Deleted in each type but it's trivial so I don't have a strong opinion - feel free to do it or not do it :)

swgillespie:
@Maoni0 do you mean something like collapsing all of these helper functions to all call a singular helper function that calls `DiagHandleCreated/Deleted`?

Maoni0:
yep, so you don't need to call Diag* every time in the functions that accept different types of handles.

swgillespie:
I originally had all of these helpers call through a common function to avoid some code duplication but I got some feedback not to do that ([here](https://github.com/dotnet/coreclr/pull/11092#discussion_r112355532)). I'm happy to do it now, especially since now we have a little more code duplication than we did before (where we were duplicating an OOM check).

Maoni0:
I dunno if Jan suggested that purely based on aesthetics or he specifically didn't want the EH code in the common function. It's up to you - my rule of thumb is aesthetic changes will unlikely please everyone so the code owner should just make a decision :)

jkotas:
> Jan suggested that purely based on aesthetics 

Yes, it was suggested based on aesthetics based on what it looked at that time. Now that there is more code getting added, I agree that having common method may be better.

swgillespie:
Sounds good! I like the common method approach, so I'll do that in the next commit.

swgillespie:
@Maoni0 does this last commit look OK to you?

swgillespie:
thanks!

Issue #12224, opened 2017-06-12
Some tests that catch `NullReferenceException` hang on Linux under GCStress=0xC


pgavlin:
The following tests exhibit this behavior:
- tests/src/Interop/MarshalAPI/ReadWrite/ReadWriteByte.csproj
- tests/src/Interop/MarshalAPI/ReadWrite/ReadWriteInt16.csproj
- tests/src/Interop/MarshalAPI/ReadWrite/ReadWriteInt32.csproj
- tests/src/Interop/MarshalAPI/ReadWrite/ReadWriteInt64.csproj
- tests/src/Interop/MarshalAPI/ReadWrite/ReadWriteIntPtr.csproj
- tests/src/JIT/Regression/CLR-x86-JIT/V1-M10/b08944/b08944b.csproj

These tests are being temporarily disabled while the hang is investigated. It appears that the runtime is entering an infinite loop when searching for an appropriate exception handler when an AV in a callee is turned into an NRE.

cc @janvorli

pgavlin:
@janvorli can you take a look at this?

janvorli:
@pgavlin sure!

gkhanna79:
Are we tracking this for 2.0?

janvorli:
I don't think it needs to be tracked for 2.0. It is an issue with GC stress 0xC only and from what I've seen so far, it seems to be caused by some bad interaction between the handling of the SIGSEGV caused by the intentional access violation in the test and the GCStress 0xC access violation handling for the inserted invalid instructions that are used to trigger GC after each jitted instruction.

janvorli:
@gkhanna79 I've found the cuprit and fixed it locally. But I have found that the issue can also happen when e.g. a NullReferenceException happens in catch handler of another NullReferenceException. It is a problem caused by the Linux kernel reusing storage for the signal context. 

So having found that, my answer to your question now changes - I believe we should fix it for 2.0.

What happens is below:
* hardware exception happens
* signal handler is called and it receives a pointer to the signal context (context of the hardware exception)
* we propagate that exception in runtime, find a catch handler, but that catch handler code causes hardware exception again.
* signal handler is called and it receives a pointer to the signal context - the same location as before
* we do first pass of the exception handling. When we hit the frame of our common signal handler function, we don't use the libunwind to go further, but we instead get pointer to the signal context that we have stored to a local  when we were executing the common signal handler and set the unwound context to that context.
[...]

PR #12316, opened 2017-06-16
Fix chained hardware exception handling on Unix

janvorli:
There is an issue when hardware exception occurs while handling another hardware exception. In such case,
the exception unwinding ends up in an infinite loop. It is caused by the kernel reusing the same location
for signal handler context.

The fix is to use the windows style context local variable in the common_signal_handler that contains
the right context - it is the original signal context converted to windows style context.

### DIFF
--- a/src/pal/src/exception/seh-unwind.cpp
+++ b/src/pal/src/exception/seh-unwind.cpp
@@ -261,18 +261,13 @@ BOOL PAL_VirtualUnwind(CONTEXT *context, KNONVOLATILE_CONTEXT_POINTERS *contextP
 
 #ifndef __APPLE__
     // Check if the PC is the return address from the SEHProcessException in the common_signal_handler. 
-    // If that's the case, extract its local variable containing the native_context_t of the hardware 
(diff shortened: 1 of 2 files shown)


janvorli:
Fixes #12224

janvorli:
CC: @gkhanna79

janvorli:
@dotnet-bot test CentOS7.1 x64 Release Priority 1 Build and Test

janvorli:
@dotnet-bot test OSX10.12 x64 Release Priority 1 Build and Test

janvorli:
@dotnet-bot test OSX10.12 pri1

janvorli:
@dotnet-bot test RHEL7.2 pri1

janvorli:
@gkhanna79, @jkotas, @Petermarcu I've found there is still a problem, the fix needs to be slightly different. I am reverting the commit and will send another PR soon.

janvorli:
Hmm, revert doesn't work for some reason. So I will fix it in a separate PR.

PR #12320, opened 2017-06-16
Compact the size of GenTreeCall nodes

BruceForstall:
The data stored in GenTreeCall under `#ifdef FEATURE_MULTIREG_RET`
(defined for x86 (RyuJIT), amd64/Linux, arm32, arm64) was poorly
packed. Improving this reduces x86 GenTreeCall node size from 124 to 112 bytes.
Since GenTreeCall is the largest node, it also reduces the size of all
"large" nodes.


BruceForstall:
@CarolEidt @pgavlin PTAL
cc @dotnet/jit-contrib

BruceForstall:

PR #12321, opened 2017-06-16
Convert size measurements for nodes to unsigned __int64

BruceForstall:
I found these numbers overflowing a size_t on x86 with a long
SuperPMI run.


BruceForstall:
@CarolEidt @pgavlin PTAL
cc @dotnet/jit-contrib

CarolEidt:
> I found these numbers overflowing a size_t on x86 with a long
SuperPMI run.

Wow!

PR #12324, opened 2017-06-16
Port to release/2.0.0: Fix chained hardware exception handling on Unix (#12316)

janvorli:
There is an issue when hardware exception occurs while handling another hardware exception. In such case,
the exception unwinding ends up in an infinite loop. It is caused by the kernel reusing the same location
for signal handler context.

The fix is to use the windows style context local variable in the common_signal_handler that contains
the right context - it is the original signal context converted to windows style context.


janvorli:

janvorli:
I've found an issue in the fix, so I am closing this PR. I'll send a new one soon.

PR #12328, opened 2017-06-16
LoadFromResolve event handler should return null 

gkhanna79:
Since LoadFromResolve event handler is wired upto AppDomain.AssemblyResolve event, it should return null when an assembly cannot be found for any reason as AssemblyResolve event expects that (See https://msdn.microsoft.com/en-us/library/system.appdomain.assemblyresolve(v=vs.110).aspx).

Fixes https://github.com/dotnet/corefx/issues/21095

@jkotas PTAL

### DIFF
--- a/src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
+++ b/src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
@@ -52,7 +52,19 @@ namespace System.Reflection
             string requestedAssemblyPath = Path.Combine(Path.GetDirectoryName(requestorPath), requestedAssemblyName.Name+".dll");
 
             // Load the dependency via LoadFrom so that it goes through the same path of being in the LoadFrom list.
-            return Assembly.LoadFrom(requestedAssemblyPath);
+            Assembly resolvedAssembly = null;
+
+            try
+            {
+                resolvedAssembly = Assembly.LoadFrom(requestedAssemblyPath);
+            }
+            catch(FileNotFoundException)
+            {
+                // Catch FileNotFoundException when attempting to resolve assemblies via this handler to account for missing assemblies.
+                resolvedAssembly = null;
+            }
+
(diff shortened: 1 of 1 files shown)


src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
```diff
@@ -52,7 +52,20 @@ private static Assembly LoadFromResolveHandler(object sender, ResolveEventArgs a
...
 
             // Load the dependency via LoadFrom so that it goes through the same path of being in the LoadFrom list.
-            return Assembly.LoadFrom(requestedAssemblyPath);
+            Assembly resolvedAssembly = null;
+
+            try
+            {
+                resolvedAssembly = Assembly.LoadFrom(requestedAssemblyPath);
+            }
+            catch(Exception)
```
jkotas: Should it be catching FileNotFoundExceptions only?

src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
```diff
@@ -52,7 +52,20 @@ private static Assembly LoadFromResolveHandler(object sender, ResolveEventArgs a
...
+            {
+                // Catch all exceptions when attempting to resolve assemblies via this handler
+                // since AssemblyResolve event would expect a null to be returned instead of an exception.
+                resolvedAssembly = null;
```
jkotas: Nit: You do not need to set it to null here.

src/mscorlib/src/System/Reflection/Assembly.CoreCLR.cs
gkhanna79: I thought about doing so but the event callback expects a null for any scenario that does not return a valid assembly. Hence, I added a generic catch.

gkhanna79:
@jkotas Updated to catch only FNF.

PR #12330, opened 2017-06-16
Fix CompareInfo.Version on Windows

tarekgh:
We call Windows to get the sort version. We call Windows passing the structure NlsVersionInfoEx but Windows expect we have to initialize the size of this structure which we are missing.

### DIFF
--- a/src/mscorlib/src/System/Globalization/CompareInfo.Windows.cs
+++ b/src/mscorlib/src/System/Globalization/CompareInfo.Windows.cs
@@ -469,6 +469,7 @@ namespace System.Globalization
             Debug.Assert(!_invariantMode);
 
             Interop.Kernel32.NlsVersionInfoEx nlsVersion = new Interop.Kernel32.NlsVersionInfoEx();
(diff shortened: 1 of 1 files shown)


tarekgh:
@danmosemsft @krwq could you please have a look?

krwq:
Looks good to me! Thank you @tarekgh! could we add a tiny test case checking if this is not null and not all zeros? that would help us avoid similar errors in the future

tarekgh:
> could we add a tiny test case checking if this is not null and not all zeros? that would help us avoid similar errors in the future

I am planning to add a test in corefx when we get the updated coreclr containing the fix

danmosemsft:
@dotnet-bot test osx10.12 x64 checked build and test (@mmitche IO error, I took box offline https://ci.dot.net/job/dotnet_coreclr/job/master/job/checked_osx10.12_tst_prtest/3202/consoleFull)

tarekgh:
Test OSX10.12 x64 Checked Build and Test please

danmosemsft:
@dotnet/dnceng another IO issue.

PR #12331, opened 2017-06-16
Add regression test for issue #12224

janvorli:
This change adds regression test for EH getting stuck in an infinite
loop when NullReferenceException happens inside a handler of another
NullReferenceException.

### DIFF
--- a/tests/src/Regressions/coreclr/GitHub_12224/test12224.cs
+++ b/tests/src/Regressions/coreclr/GitHub_12224/test12224.cs
@@ -1,0 +1,43 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+using System;
+using System.Threading;
+
+public class Test12224
+{    
+    // Regression test for EH getting stuck in an infinite loop when NullReferenceException
+    // happens inside a handler of another NullReferenceException.
+    static void ExecuteTest(object context)
+    {
+        string s = null;
+        try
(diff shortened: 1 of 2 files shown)


janvorli:
FYI: @Petermarcu

tests/src/Regressions/coreclr/GitHub_12224/test12224.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    public static int Main()
+    {
+        Thread thread = new Thread(new ParameterizedThreadStart(Test12224.ExecuteTest));
+        thread.IsBackground = true;
+        thread.Start(null);
+
+        // Give the thread 30 seconds to complete (it should be immediate). If it fails
+        // to complete within that timeout, it has hung.
+        bool terminated = thread.Join(new TimeSpan(0, 0, 30));
```
gkhanna79: By terminated, do we mean to say "not timed out"?

janvorli: Yes, the MSDN doc says about the Join return value:
true if the thread terminated; false if the thread has not terminated after the amount of time specified by the timeout parameter has elapsed.

janvorli:
@dotnet-bot test RHEL7.2 pri1

janvorli:
@dotnet-bot test RHEL7.2 pri1

Petermarcu:
Once this is green, go ahead and send the updated details in mail with both of these changes and we can get them in.

janvorli:

PR #12335, opened 2017-06-16
Remove unnecessary bitvec allocation in assertion prop

mikedn:
`optVnNonNullPropCurStmt` is called for each node and every time it allocates an empty `BitVec` because `optNonNullAssertionProp_Call` and `optAssertionProp_Ind` require it, they are also used in other code paths that do provide a non-empty `BitVec`.

### DIFF
--- a/src/jit/assertionprop.cpp
+++ b/src/jit/assertionprop.cpp
@@ -3524,12 +3524,12 @@ AssertionIndex Compiler::optAssertionIsNonNullInternal(GenTreePtr op, ASSERT_VAL
     // If local assertion prop use lcl comparison, else use VN comparison.
     if (!optLocalAssertionProp)
     {
-        ValueNum vn = op->gtVNPair.GetConservative();
-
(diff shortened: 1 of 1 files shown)


sandreenko:
LGTM, cc @dotnet/jit-contrib .

but as I see from the code the function MayBeUninit actually check that it is uninit.
```
    // Returns "true" iff "bs" may be the uninit value.
    static bool MayBeUninit(BitSetValueArgType bs);
```
should be
```
// Returns "true" iff "bs" is the uninit value.
static bool IsUninit(BitSetValueArgType bs);
```

mikedn:
> but as I see from the code the function MayBeUninit actually check that it is uninit.

I think that the reason it is called "MayBe" rather than "Is" is that for short vectors uninit and empty are indistinguishable. This uninit thing is IMO a bit bizarre but I'm taking advantage of it not only to avoid allocations (that can also be avoided by caching an empty bitvec somewhere) but to also avoid unnecessary `IsEmpty` checks which are somewhat expensive for long vectors.

PR #12343, opened 2017-06-17
support spmi collection for r2r calls.

### LINKED ISSUE #11702: SuperPMI should support ReadyToRun
When crossgen'ing assemblies with ReadyToRun, the following NYIs show up in SuperPMI:
* getReadyToRunHelper
* GetReadyToRunDelegateCtorHelper

There may be others.

sandreenko:
Please review the second commit.
The first is just a formatting.

Fix #11702.


sandreenko:
cc @dotnet/jit-contrib

src/ToolBox/superpmi/superpmi-shared/methodcontext.cpp
```diff
@@ -2335,50 +2336,86 @@ void MethodContext::recGetReadyToRunHelper(CORINFO_RESOLVED_TOKEN* pResolvedToke
...
+    GetReadyToRunHelper->Add(key, value);
 }
-void MethodContext::dmpGetReadyToRunHelper(DWORDLONG key, DWORD value)
+
+void MethodContext::dmpGetReadyToRunHelper(Agnostic_GetReadyToRunHelper_TOKENin  key,
+                                           Agnostic_GetReadyToRunHelper_TOKENout value)
 {
-    LogError("getReadyToRunHelper NYI");
+    LogError("dmpGetReadyToRunHelper NYI");
 }
```
BruceForstall: Can you implement this?

src/ToolBox/superpmi/superpmi-shared/methodcontext.cpp
```diff
@@ -2335,50 +2336,86 @@ void MethodContext::recGetReadyToRunHelper(CORINFO_RESOLVED_TOKEN* pResolvedToke
...
+    GetReadyToRunDelegateCtorHelper->Add(key, value);
 }
-void MethodContext::dmpGetReadyToRunDelegateCtorHelper(DWORDLONG key, DWORD value)
+
+void MethodContext::dmpGetReadyToRunDelegateCtorHelper(Agnostic_GetReadyToRunDelegateCtorHelper_TokenIn value,
+                                                       CORINFO_LOOKUP                                   key)
 {
-    LogError("getReadyToRunDelegateCtorHelper NYI");
+    LogError("dmpgetReadyToRunDelegateCtorHelper NYI");
 }
```
BruceForstall: Can you implement this?

BruceForstall:
Don't you need to add something to `enum mcPackets`?

How have you tested this? Have you collected a crossgen of the framework assemblies, and then done a playback?

The implementation looks fine to me, but I don't understand the details of those APIs.

src/ToolBox/superpmi/superpmi-shared/methodcontext.cpp
sandreenko: If honest I want to delete all dump methods because:
1) We don't use them;
2) We don't test them;
3) They shows information in non-useful format.

I can imagine scenario when dmp methods could be useful, but not in the current variant.

To implement them I will have to write like tons of printf with all fields inside CORINFO_RESOLVED_TOKEN* and CORINFO_LOOKUP_KIND* and etc. So I prefer to keep it unimplemented for now.

In the future we will add dump functions for our types and just call them here,

sandreenko:
> Don't you need to add something to enum mcPackets?
I am not familiar with it, what does it stand for? I don't see any usages.
However, R2R are already defined there:
```
Packet_GetReadyToRunHelper                           = 150, // Added 10/10/2014
Packet_GetReadyToRunDelegateCtorHelper               = 157, // Added 3/30/2016 
```

src/ToolBox/superpmi/superpmi-shared/methodcontext.cpp
BruceForstall: I do use them sometimes (mcs -dump). The printfs are not that hard/bad to write, normally.

sandreenko:

sandreenko:
@BruceForstall could you look again?

I checked several libraries, for example for Microsoft.CodeAnalysis.CSharp.dll there are 9854 methods,
before there were 4515 failed, now 1612.

I didn't add dump for the part of the value for ctor delegate, because I am wondering that somewhere should already exist method that dumps CORINFO_LOOKUP, does it?

sandreenko:
Updated with another cleaning, duplicates were deleted from GetCallInfo and EmbedGenericHandle.
Dump for CORINFO_LOOKUP was added.

PR #12344, opened 2017-06-17
Really fix chained hardware exception handling on Unix

janvorli:
This is a real fix for the issue. When making the previous fix, I have not realized that the windows style context that we generate in the common_signal_handler and that I was using instead of the unix signal context is modified when IL_Rethrow is called.
So the right fix is to create a local copy of the original context that nothing would modify and that would always point to the exception that caused the specific common_signal_handler call.


janvorli:

janvorli:
@dotnet-bot test OSX10.12 pri1

janvorli:
@dotnet-bot test RHEL7.2 pri1

PR #12345, opened 2017-06-17
Fix perf.groovy

swaroop-sridhar:
Fix perf.groovy so that ILLink perf tests can be triggered.


swaroop-sridhar:
@dotnet-bot test ci please

swaroop-sridhar:
@dotnet-bot test Windows_NT x64 ILLink Perf

adiaaida:
@swaroop-sridhar it is expected that the il link job would still fail with a groovy change. The triggered job uses the old version, not the new version. Feel free to simply check this in.

PR #12346, opened 2017-06-17
Encapsulate the lclVar table in a class.

pgavlin:
This change replaces the open-coded lclVar table with a class, `LclVarTable`. This class provides bounds-checked indexing, C++11-compatible iteration, etc.

As this change is rather large, I've divided it into three logical parts:
- The first commit moves some lclVar-related code into a new header, `lclvars.h`.
- The second commit adds the `LclVarTable`.
- Each successive commit updates a logical slice of the JIT, starting with common/utility code and then moving from the frontend to the code generator.


src/jit/lclvars.h
```diff
@@ -650,3 +650,47 @@ class LclVarDsc
...
+
+    LclVarTable(LclVarTable&& other) = delete;
+    LclVarTable(const LclVarTable& other) = delete;
+
+    unsigned Count() const
+    {
+        return m_size;
+    }
+
+    LclVarDsc& operator[](const int index) const
```
mikedn: Why `int` and not `unsigned`?

pgavlin: I actually changed that to `unsigned` in d2eb569 :)

pgavlin: 5a1e7d5, rather

src/jit/assertionprop.cpp
```diff
@@ -68,8 +65,10 @@ void Compiler::optAddCopies()
         return;
     }
 
-    for (lclNum = 0, varDsc = lvaTable; lclNum < lvaCount; lclNum++, varDsc++)
+    for (unsigned lclNum = 0; lclNum < lvaCount; lclNum++)
     {
+        LclVarDsc* varDsc = &lvaTable[lclNum];
```
mikedn: Should perhaps `operator[]` return `LclVarDsc*` rather than `LclVarDsc&` because most of the JIT code uses pointers to lclvars?

pgavlin: I considered this, but not until I had already gone down the rabbit hole of `&lvaTable[lclNum]`. I don't feel particularly strongly either way, so if `LclVarDsc*` seems preferable, I'm happy to make that change.

mikedn: To be clear - it's only preferable because so much code already uses pointers and converting that code to use references may not be an option. Otherwise I see no reason to use `LclVarDsc*` instead of `LclVarDsc&` anywhere where nulls aren't needed.

pgavlin: That's a fair point.

One other change I had considered was to add `LclVarDsc* operator+(unsigned index)` to `LclVarTable`, which would avoid the need to update most of these sites in any case.

I think I'd prefer to keep things as-is, though, as I find references generally preferable to pointers given the nullability issue. My intent is to gradually replace the various uses of `LclVarDsc*` with `LclVarDsc&`, so this should eventually be a non-issue.

pgavlin:

pgavlin:
@dotnet-bit test Tizen armel Cross Release Build

src/jit/lclvars.cpp
```diff
@@ -193,45 +193,27 @@ void Compiler::lvaInitTypeRef()
...
-    }
-
-    lvaTable         = (LclVarDsc*)compGetMemArray(lvaTableCnt, sizeof(*lvaTable), CMK_LvaTable);
-    size_t tableSize = lvaTableCnt * sizeof(*lvaTable);
-    memset(lvaTable, 0, tableSize);
-    for (unsigned i = 0; i < lvaTableCnt; i++)
-    {
-        new (&lvaTable[i], jitstd::placement_t()) LclVarDsc(this); // call the constructor.
-    }
+    lvaTable = LclVarTable(*this, info.compLocalsCount = info.compArgsCount + info.compMethodInfo->locals.numArgs);
```
sandreenko: Is not 
```
info.compLocalsCount = info.compArgsCount + info.compMethodInfo->locals.numArgs;
lvaTable = LclVarTable(*this, info.compLocalsCount); 
```
better?

src/jit/lclvars.cpp
```diff
@@ -7264,15 +7200,13 @@ void Compiler::lvaDispVarSet(VARSET_VALARG_TP set, VARSET_VALARG_TP allVars)
     {
         if (VarSetOps::IsMember(this, set, index))
         {
-            unsigned   lclNum;
-            LclVarDsc* varDsc;
-
             /* Look for the matching variable */
-
-            for (lclNum = 0, varDsc = lvaTable; lclNum < lvaCount; lclNum++, varDsc++)
+            unsigned lclNum;
```
sandreenko: don't we need an init value here or check that we found the matching variable?

src/jit/lclvars.h
```diff
@@ -650,3 +650,101 @@ class LclVarDsc
...
+        {
+            ++m_lclVarDsc;
+            return *this;
+        }
+    };
+
+    LclVarTable();
+    LclVarTable(Compiler& compiler, unsigned initialSize);
+
+    LclVarTable& operator=(const LclVarTable& other);
```
sandreenko: Why can it be a move assignment? As I see we only use it to init from constructor or from inline lvaTable, that we don't use after.

src/jit/lclvars.cpp
pgavlin: Yes, it is--good catch.

src/jit/lclvars.h
pgavlin: We use this to assign to an inlinee compiler's lclVar table from the parent compiler's lclVar table. Making that a move assignment would invalidate the parent's lclVar table, which we do not want to do.

src/jit/lclvars.cpp
pgavlin: Yes, either of these is fine.

BruceForstall:
It would be useful to do calgrind/pin measurements to ensure no throughput regression. And a lab throughput run, too.

I assume there are no codegen diffs (e.g., SuperAsmDiffs)

src/jit/lclvars.h
```diff
@@ -0,0 +1,750 @@
...
+        {
+            printf("%s", getRegName(lvRegNum));
+        }
+    }
+#endif // DEBUG
+
+}; // class LclVarDsc
+
+class LclVarTable final
+{
```
BruceForstall: I'm not fond of all the references. There are a zillion functions in the JIT that take a `Compiler*`; I don't see why these need to take a `Compiler&` instead. It also seems like `operator[]` should return a `LclVarDsc*`.

pgavlin: > I'm not fond of all the references. There are a zillion functions in the JIT that take a `Compiler*`; I don't see why these need to take a `Compiler&` instead.

Defensive coding: it's much, much more difficult to accidentally manufacture a null `&` than a null `*`. If you feel strongly about this, though, I'm happy to change these to `Compiler*`s.

>  It also seems like `operator[]` should return a `LclVarDsc*`.

That's two votes for that option (@mikedn) mentioned the same. I'll make the change.

On a related topic, I had suggested in that discussion that we could add a `LclVarDsc* operator+(unsigned index)` to `LclVarTable`, which would avoid the need to make most of the edits in these files. I shied away from that because I felt that it made the underlying behavior less clear, but I'm not entirely opposed to adding it. Thoughts?

BruceForstall: I'm not terribly fond of any operator overloading. I'd consider renaming `operator[]` to `GetLocal(unsigned index)`

---
In reply to: [122849081](https://github.com/dotnet/coreclr/pull/12346#discussion_r122849081) [](ancestors = 122849081)

src/jit/lclvars.h
```diff
@@ -0,0 +1,750 @@
...
+        return Iterator(&m_array[m_count]);
+    }
+
+    IteratorPair<Iterator> LclVars(unsigned start, unsigned end) const
+    {
+        assert(start <= end);
+        assert(end <= m_count);
+        return MakeIteratorPair(Iterator(&m_array[start]), Iterator(&m_array[end]));
+    }
+
```
BruceForstall: How does the performance of:
```
    unsigned lclNum = 0;
    for (LclVarDsc *varDsc = lvaTable; lclNum < lvaCount; lclNum++, varDsc++)
```
compare to:
```
    for (LclVarDsc* varDsc : lvaTable)
```
?

I like the readability of the new code, but the old code seems hard to beat for throughput (and there is no "magic" to think about).

Also, have you debugged these in Visual Studio? What happens when you're on the "for" statement and "step into"? Seems like you'll end up in a swamp of iterator helpers...

src/jit/codegenarmarch.cpp
```diff
@@ -1968,6 +1964,8 @@ void CodeGen::genJmpMethod(GenTreePtr jmp)
             continue;
         }
 
+        const unsigned varNum = compiler->lvaTable.GetLclNum(varDsc);
+
```
BruceForstall: If you left the iteration alone (i.e., old style), you wouldn't need to do this. So is it worth it to change the iteration?

src/jit/codegencommon.cpp
```diff
@@ -2668,13 +2665,9 @@ void CodeGen::genExitCode(BasicBlock* block)
...
 
-            unsigned   varNum;
-            LclVarDsc* varDsc;
-
             /* Figure out which register parameters hold pointers */
 
-            for (varNum = 0, varDsc = compiler->lvaTable; varNum < compiler->lvaCount && varDsc->lvIsRegArg;
-                 varNum++, varDsc++)
+            for (LclVarDsc* varDsc : compiler->lvaTable)
             {
```
BruceForstall: You dropped the `&& varDsc->lvIsRegArg` clause

src/jit/codegenarmarch.cpp
pgavlin: The old code needed to do `varDsc = compiler->lvaTable + varNum;`, so in this case it's a wash. I think the `foreach` style is a bit more bulletproof, hence the change.

src/jit/codegencommon.cpp
pgavlin: Yes, good catch.

src/jit/lclvars.cpp
```diff
@@ -7307,3 +7241,111 @@ void Compiler::lvaDispVarSet(VARSET_VALARG_TP set, VARSET_VALARG_TP allVars)
...
+    Grow(compiler, initialSize < 8 ? 16 : initialSize * 2);
+    m_count = initialSize;
+}
+
+LclVarTable& LclVarTable::operator=(const LclVarTable& other)
+{
+    m_count = other.m_count;
+    m_capacity = other.m_capacity;
+    m_array = other.m_array;
+
```
BruceForstall: Does this function ever get called? And it expects to share the `m_array`?

pgavlin: Yes, believe it or not--we share the lclVar table between a root compiler and its inlinee compilers.

src/jit/lclvars.h
BruceForstall: I guess I'm ok with `operator[]` returning a `&` since then the calling code makes lvaTable "look" more like an array, e.g. 'LclVarDsc* tmp = &lvaTable[lclNum]`.

I actually like not having the `operator+` and forcing all callers to look the same (and changing all of them to do so)

pgavlin:
> It would be useful to do calgrind/pin measurements to ensure no throughput regression. And a lab throughput run, too.

callgrind reports an improvement of about a hundredth of a percent with this change.

> I assume there are no codegen diffs (e.g., SuperAsmDiffs)

That is correct.

src/jit/lclvars.cpp
BruceForstall: maybe it would be worthwhile to add a comment to that effect.

src/jit/lclvars.h
```diff
@@ -0,0 +1,750 @@
...
+                                // (if lvIsParam is true)
+#if OPT_BOOL_OPS
+    unsigned char lvIsBoolean : 1; // set if variable is boolean
+#endif
+    unsigned char lvRngOptDone : 1; // considered for range check opt?
+    unsigned char lvLoopInc : 1;    // incremented in the loop?
+    unsigned char lvLoopAsg : 1;    // reassigned  in the loop (other than a monotonic inc/dec for the index var)?
+    unsigned char lvArrIndx : 1;    // used as an array index?
+    unsigned char lvArrIndxOff : 1; // used as an array index with an offset?
+    unsigned char lvArrIndxDom : 1; // index dominates loop exit
```
CarolEidt: Since you're here, it would be good to be consistent with describing this booleans. To me, when the comment has a "?" at the end it sounds as if the author wasn't sure. I would either phrase it as an assertion, like this one "index dominates loop exit" or as a complete question "Does the index dominate the loop exit".

src/jit/lclvars.h
```diff
@@ -0,0 +1,750 @@
...
+    {
+        return Iterator(&m_array[0]);
+    }
+
+    Iterator end() const
+    {
+        return Iterator(&m_array[m_count]);
+    }
+
+    IteratorPair<Iterator> LclVars(unsigned start, unsigned end) const
```
CarolEidt: >LclVars [](start = 27, length = 7)

I don't feel strongly about this, and I know that it is common usage to use a plural to imply an iterator, but I prefer a more explicit name, e.g. `LclVarIterator`

karelz:
What is status of this PR? 2.5 months no update :(

pgavlin:
Still in progress, essentially.

karelz:
If you don't work on it actively (and don't plan to do so in next 1-2 weeks), I would suggest to close the PR, until you have time to continue the work on it. Thanks!

karelz:

Issue #12355, opened 2017-06-18
Errors in multiple MSBuild properties initialization of System.Private.CoreLib.csproj in VS2017


4creators:
### Building from Visual Studio 2017 command line
According to my tests it affects only builds performed from Visual Studio 2017 command line. Builds done from Visual Studio 2015 command line are not affected.

System.Private.CoreLib.csproj build is invoked via $(RepoRoot)/src/build.proj together with SOS project. During build at the initial stage of property evaluation in System.Private.CoreLib.csproj when majority of properties were not evaluated project [imports Microsoft.Net.Compilers.props at line 4](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/System.Private.CoreLib.csproj#L4):
```
<Import Project="..\..\Tools\net46\roslyn\build\Microsoft.Net.Compilers.props" Condition="'$(OS)'=='Windows_NT'" />
```
Next on [line 7 project imports $(RepoRoot)/dir.props](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/System.Private.CoreLib.csproj#L7):
```
<Import Project="..\..\dir.props" /> 
```
which itself imports [Microsoft.Net.Compilers.props at line 186](https://github.com/dotnet/coreclr/blob/master/dir.props#L186)
```
 <Import Project="$(RoslynPropsFile)" Condition="'$(RunningOnUnix)'!='true' and Exists('$(RoslynPropsFile)') and '$(UseRoslynCompilers)'!='false'" />
```
Resulting double import of Microsoft.Net.Compilers.props file leads to MSBuild warning MSB4011:
```
e:\src\ms\coreclrbug\dir.props(186,3): warning MSB4011: "e:\src\ms\coreclrbug\Tools\net46\roslyn\build\Microsoft.Net.Compilers.props" cannot be imported again. It was already imported at "e:\src\ms\coreclrbug\src\mscorlib\System.Private.CoreLib.csproj (4,3)". This is most likely a build authoring error. This subsequent import will be ignored. 
```
The most important part of this message `This subsequent import will be ignored` indicates that evaluation of MSBuild properties in Microsoft.Net.Compilers.props file will be performed only once before all definitions and property values declared in $(RepoRoot)/dir.props will be introduced into evaluation scope. As a result any property which was evaluated in Microsoft.Net.Compilers.props before $(RepoRoot)/dir.props were imported and was not explicitly overriden will retain it's value. Some of this properties include IntermediateOutputPath.

[...]

PR #12356, opened 2017-06-18
Fix errors in MSBuild properties initialization in System.Private.CoreLib.csproj

4creators:
Fixes issue #12355

Project System.Private.CoreLib.csproj imports directly Microsoft.Net.Compilers.props
at line# 4 before including dir.props from repo root. $(RepoRoot)/dir.props imports
Microsoft.Net.Compilers.props at line# 186. As a result of this sequence of duplicated
project imports and lack of safeguards against importing the same .props file twice
MSBuild reports that Microsoft.Net.Compilers.props file is included twice with warning
that second import will be ignored - in this case not evaluated. This results
in initialization of Microsoft.Net.Compilers.props values to defaults because MSBuild
evaluates Microsoft.Net.Compilers.props only once before $(RepoRoot)/dir.props are included.
Removing import of Microsoft.Net.Compilers.props from System.Private.CoreLib.csproj
and moving $(RepoRoot)/dir.props file import to top of System.Private.CoreLib.csproj
solves these problems by ensuring expected order of evaluation of Microsoft.Net.Compilers.props
file by MSBuild after all project wide values are initialized in $(RepoRoot)/dir.props.

Above described configuration problems affect only builds with Visual Studio 2017 both from developer command prompt and from IDE.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-
+  <Import Condition="Exists('..\dir.props')" Project="..\dir.props" />
+  
   <!--
     $(OS) is set to Unix/Windows_NT. This comes from an environment variable on Windows and MSBuild on Unix.
   -->
--- a/src/mscorlib/System.Private.CoreLib.csproj
+++ b/src/mscorlib/System.Private.CoreLib.csproj
@@ -1,10 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <!-- Always use latest Roslyn compiler -->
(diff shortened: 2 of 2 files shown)


dir.props
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-
+  <Import Condition="Exists('..\dir.props')" Project="..\dir.props" />
+
```
4creators: Added to enable inclusion of dir.props file located in directory above repo root. This enables coordinated changes to build configurations of multiple dotnet projects build side by side. It is used in [corefx](https://github.com/dotnet/corefx/blob/master/dir.props#L3), [standard](https://github.com/dotnet/standard/blob/master/dir.props#L3), [core-setup](https://github.com/dotnet/core-setup/blob/master/dir.props#L3), [corert](https://github.com/dotnet/corert/blob/master/dir.props#L2) and other repos.

4creators:
@jkotas Is it possible to merge this commit - without this fix it is impossible to use of VS2017 for work on System.Privet CoreLib. This creates serious build problems (when building from IDE) as described in #12355

jkotas:
Thanks!

PR #12358, opened 2017-06-19
[RyuJIT/ARM32] Assign 2 numslots to gtLsraInfo.srcCount in TYP_DOUBLE

### LINKED ISSUE #11849: [RyuJIT/arm32] Assertion failed 'info->numSlots == 1'
At lsraarmarch.cpp Line: 686

Example:
```
Assert failure(PID 13364 [0x00003434], Thread: 5500 [0x157c]): Assertion failed 'info->numSlots == 1' in 'ldarg_s_r8:test_float64():int' (IL size 2310)

    File: c:\gh\coreclr\src\jit\lsraarmarch.cpp Line: 686
    Image: c:\brucefo\tests\Windows_NT.arm.Checked\Tests\Core_Root\CoreRun.exe
```

Tests with this assert:
```
JIT\IL_Conformance\Old\directed\ldarg_s_r8\ldarg_s_r8.cmd
```

jyoungyun:
Fixed #11849 issue

This issue is related to passing more than 8 double arguments as function.
The double argument uses 2 numSlots but this assersion always requires only one numSlots except TYP_STRUCT and GT_FIELD_LIST.

Tracing with gdb:
```
#10 0x70de8944 in Lowering::TreeNodeInfoInitPutArgStk (this=0x7effc290, argNode=0xe5a38, info=0xc1c3c)
    at /opt/code/src/jit/lsraarmarch.cpp:694
694	in /opt/code/src/jit/lsraarmarch.cpp
(gdb) p info->numSlots 
$4 = 2
(gdb) p *putArgChild
$8 = {_vptr$GenTree = 0x70fdf194 <vtable for GenTreeDblCon+8>, gtOper = genTreeOps::GT_CNS_DBL, gtType = var_types::TYP_DOUBLE, 
  gtOperSave = genTreeOps::GT_NONE, gtCSEnum = 0 '\000', gtLIRFlags = 0 '\000', gtAssertionInfo = {m_isNextEdgeAssertion = 0, 
    m_assertionIndex = 0}, gtCostsInitialized = true, _gtCostEx = 3 '\003', _gtCostSz = 4 '\004', 
  gtRegTag = GenTree::GT_REGTAG_REG, {_gtRegNum = 49 '1', _gtRegPair = 49}, gtFlags = 0, gtDebugFlags = 2, gtVNPair = {
    m_liberal = 256, m_conservative = 256}, gtRsvdRegs = 0, gtLsraInfo = {loc = 5, srcCandsIndex = 1 '\001', 
    dstCandsIndex = 1 '\001', internalCandsIndex = 0 '\000', _srcCount = 0 '\000', _dstCount = 1 '\001', 
    _internalIntCount = 2 '\002', _internalFloatCount = 0 '\000', isLocalDefUse = 0 '\000', isLsraAdded = 0 '\000', 
    isDelayFree = 0 '\000', hasDelayFreeSrc = 0 '\000', isTgtPref = 0 '\000', regOptional = 0 '\000', 
```

But even though this patch is merged, it still occurs a below assersion. This issue will be traced with new issue.
```
Assert failure(PID 19461 [0x00004c05], Thread: 19461 [0x4c05]): Assertion failed '((consume == 0) && (produce == 0)) || (ComputeAvailableSrcCount(tree) == consume)' in 'ldarg_s_r8:test_float64():int' (IL size 2310)

    File: /opt/code/src/jit/lsra.cpp Line: 3599
    Image: /home/pi3/Downloads/ryujit/overlay-0615-fixed/corerun
```

/cc @hqueue @hseok-oh @wateret @sjsinju

### DIFF
--- a/src/jit/lsraarmarch.cpp
+++ b/src/jit/lsraarmarch.cpp
@@ -711,12 +711,13 @@ void Lowering::TreeNodeInfoInitPutArgStk(GenTreePutArgStk* argNode, fgArgTabEntr
             MakeSrcContained(argNode, putArgChild);
         }
     }
+#ifdef _TARGET_ARM_
     else if (putArgChild->TypeGet() == TYP_DOUBLE) {
-        // We consume all of the items in the TYP_DOUBLE
-        // TYP_DOUBLE uses 2 numSlots
-        argNode->gtLsraInfo.srcCount = info->numSlots;
+        // Even though TYP_DOUBLE uses 2 numSlots, we consume 1 srcCount in TYP_DOUBLE
+        argNode->gtLsraInfo.srcCount = 1;
         putArgChild->SetContained();
     }
+#endif
     else
     {
         // We must not have a multi-reg struct
--- a/src/jit/lsraarmarch.cpp
+++ b/src/jit/lsraarmarch.cpp
@@ -711,17 +711,15 @@ void Lowering::TreeNodeInfoInitPutArgStk(GenTreePutArgStk* argNode, fgArgTabEntr
             MakeSrcContained(argNode, putArgChild);
         }
     }
-#ifdef _TARGET_ARM_
-    else if (putArgChild->TypeGet() == TYP_DOUBLE) {
-        // Even though TYP_DOUBLE uses 2 numSlots, we consume 1 srcCount in TYP_DOUBLE
-        argNode->gtLsraInfo.srcCount = 1;
-        putArgChild->SetContained();
-    }
-#endif
     else
     {
+#ifdef _TARGET_ARM_
+        // We must not have a multi-reg strcut; double uses 2 slots and isn't a multi-reg struct
+        assert((info->numSlots == 1) || ((info->numSlots == 2) && (putArgChild->TypeGet() == TYP_DOUBLE)));
+#else // !_TARGET_ARM_
         // We must not have a multi-reg struct
         assert(info->numSlots == 1);
+#endif // !_TARGET_ARM_
     }
 }
 
--- a/src/jit/lsraarmarch.cpp
+++ b/src/jit/lsraarmarch.cpp
@@ -711,6 +711,12 @@ void Lowering::TreeNodeInfoInitPutArgStk(GenTreePutArgStk* argNode, fgArgTabEntr
             MakeSrcContained(argNode, putArgChild);
         }
     }
+    else if (putArgChild->TypeGet() == TYP_DOUBLE) {
+        // We consume all of the items in the TYP_DOUBLE
+        // TYP_DOUBLE uses 2 numSlots
+        argNode->gtLsraInfo.srcCount = info->numSlots;
+        putArgChild->SetContained();
+    }
     else
     {
         // We must not have a multi-reg struct
--- a/src/jit/lsraarmarch.cpp
+++ b/src/jit/lsraarmarch.cpp
@@ -714,9 +714,9 @@ void Lowering::TreeNodeInfoInitPutArgStk(GenTreePutArgStk* argNode, fgArgTabEntr
     else
     {
 #ifdef _TARGET_ARM_
-        // We must not have a multi-reg strcut; double uses 2 slots and isn't a multi-reg struct
+        // We must not have a multi-reg struct; double uses 2 slots and isn't a multi-reg struct
         assert((info->numSlots == 1) || ((info->numSlots == 2) && (putArgChild->TypeGet() == TYP_DOUBLE)));
-#else // !_TARGET_ARM_
+#else  // !_TARGET_ARM_
         // We must not have a multi-reg struct
         assert(info->numSlots == 1);
 #endif // !_TARGET_ARM_


hseok-oh:
cc/ @dotnet/arm32-contrib

hqueue:
@jyoungyun please add header "[RyuJIT/ARM32]" to the tile :)

src/jit/lsraarmarch.cpp
```diff
@@ -711,6 +711,12 @@ void Lowering::TreeNodeInfoInitPutArgStk(GenTreePutArgStk* argNode, fgArgTabEntr
             MakeSrcContained(argNode, putArgChild);
         }
     }
+    else if (putArgChild->TypeGet() == TYP_DOUBLE) {
```
wateret: I think this `else if` block should be wrapped in `#ifdef _TARGET_ARM_` since this file is for both ARM and ARM64.

jyoungyun: Yes, right. I will add next patch for adding `_TARGET_ARM_` tag and changing logic.

jyoungyun:
Next assersion:
```
Assert failure(PID 21848 [0x00005558], Thread: 21848 [0x5558]): Assertion failed '!isMarkedContained' in 'ldarg_s_r8:test_float64():int' (IL size 2310)

    File: /opt/code/src/jit/gentree.cpp Line: 15484
    Image: /home/pi3/Downloads/ryujit/overlay-0619-fixed/corerun
```

src/jit/lsraarmarch.cpp
```diff
@@ -711,6 +711,13 @@ void Lowering::TreeNodeInfoInitPutArgStk(GenTreePutArgStk* argNode, fgArgTabEntr
             MakeSrcContained(argNode, putArgChild);
         }
     }
+#ifdef _TARGET_ARM_
+    else if (putArgChild->TypeGet() == TYP_DOUBLE) {
+        // Even though TYP_DOUBLE uses 2 numSlots, we consume 1 srcCount in TYP_DOUBLE
+        argNode->gtLsraInfo.srcCount = 1;
+        putArgChild->SetContained();
```
BruceForstall: Looks like `argNode->gtLsraInfo.srcCount = 1;` is unnecessary, since that was already done above.

Why do `TYP_DOUBLE` nodes need to get special treatment, marked as contained?

Is this issue really a matter of loosening the assert below, maybe changing it to:
```
#ifdef _TARGET_ARM_
        // We must not have a multi-reg struct; double uses 2 slots and isn't a multi-reg struct.
        assert((info->numSlots == 1) || ((info->numSlots == 2) && (putArgChild->TypeGet() == TYP_DOUBLE)));
#else // !_TARGET_ARM_
        // We must not have a multi-reg struct
        assert(info->numSlots == 1);
#endif // !_TARGET_ARM_
```

jyoungyun: @BruceForstall I've been looking at the ryujit code for the first time, so I can get the wrong code. :( I faced an issue because I marked as contained so I'm trying to fix it as you said. Thank you for friendly comments!

src/jit/lsraarmarch.cpp
```diff
@@ -711,17 +711,15 @@ void Lowering::TreeNodeInfoInitPutArgStk(GenTreePutArgStk* argNode, fgArgTabEntr
...
-    else if (putArgChild->TypeGet() == TYP_DOUBLE) {
-        // Even though TYP_DOUBLE uses 2 numSlots, we consume 1 srcCount in TYP_DOUBLE
-        argNode->gtLsraInfo.srcCount = 1;
-        putArgChild->SetContained();
-    }
-#endif
     else
     {
+#ifdef _TARGET_ARM_
+        // We must not have a multi-reg strcut; double uses 2 slots and isn't a multi-reg struct
```
BruceForstall: typo: strcut => struct

BruceForstall:
@jyoungyun does your latest change fix the assert you're trying to fix? Since the current proposed change only changes an existing assert, but doesn't change any non-assert behavior, it seems like it can't fix the original assert.

CarolEidt:
> Since the current proposed change only changes an existing assert, but doesn't change any non-assert behavior, it seems like it can't fix the original assert.

I gather that the original assert was `assert(info->numSlots == 1);` which is incorrect for `TYP_DOUBLE`. So I think this would address that issue. 

We can't do `argNode->gtLsraInfo.srcCount = info->numSlots;` because for a `TYP_DOUBLE` there should be only one source, which is a double register (although they are aliased to two floats, we still handle it in register allocation as a single source).

hqueue:
@jyoungyun arm cross CI failures are due to infra, and it is now fixed. When you submit PR to fix formatting, arm cross CI will work without problem.

jyoungyun:
@BruceForstall Yes, when I change the assersion phrase, it works well. :) This patch does not change any behavior of double operation. So I think it still has an issues to fix more but it works well so I was embarrassed. :)
@CarolEidt I fixed the formatting and one typo issue. This patch is ready for merge. Please take a look.

PR #12360, opened 2017-06-19
[RyuJIT/armel] Support `double` argument passing

### LINKED ISSUE #12293: [RyuJIT/armel] Support passing double type register args
Parent Issue: #11822

`float` type variables were covered by #12005, but we don't have a trivial solution for `double` types. Since that there is no multi-reg support for GT_PUTARG_REG and GT_COPY. I'm going to implement this according to https://github.com/dotnet/coreclr/pull/12005#pullrequestreview-41574710.

wateret:
Fix #12293

### DIFF
--- a/src/jit/gentree.cpp
+++ b/src/jit/gentree.cpp
@@ -6998,6 +6998,39 @@ GenTree* Compiler::gtNewBlkOpNode(
     return result;
 }
 
+//------------------------------------------------------------------------
+// gtNewPutArgReg: Creates a new PutArgReg node.
+//
+// Arguments:
+//    type - The actual type of the argument
+//    arg  - The argument node
+//
+// Return Value:
+//    Returns the newly created PutArgReg node.
+//
+// Notes:
+//    The node is generated as GenTreeMultiRegOp on armel, as GenTreeOp on all the other archs
+//
+GenTreePtr Compiler::gtNewPutArgReg(var_types type, GenTreePtr arg)
+{
+    assert(arg != nullptr);
+
+    GenTreePtr node = nullptr;
+#if !defined(LEGACY_BACKEND) && defined(_TARGET_ARM_)
+    // A PUTARG_REG could be a MultiRegOp on armel since we could move a double register to two int registers.
+    if (opts.compUseSoftFP)
+    {
+        node = new (this, GT_PUTARG_REG) GenTreeMultiRegOp(GT_PUTARG_REG, type, arg, nullptr);
+    }
+    else
+#endif
+    {
+        node = gtNewOperNode(GT_PUTARG_REG, type, arg);
+    }
+
+    return node;
+}
+
 /*****************************************************************************
  *
  *  Clones the given tree value and returns a copy of the given tree.
--- a/src/jit/gentree.h
+++ b/src/jit/gentree.h
@@ -1637,6 +1637,9 @@ public:
 #ifdef FEATURE_SIMD
             case GT_SIMD:
 #endif // !FEATURE_SIMD
+#if !defined(LEGACY_BACKEND) && _TARGET_ARM_
+            case GT_PUTARG_REG:
+#endif // !defined(LEGACY_BACKEND) && _TARGET_ARM_
                 return true;
             default:
                 return false;
@@ -3818,16 +3821,17 @@ struct GenTreeCmpXchg : public GenTree
 };
 
 #if !defined(LEGACY_BACKEND) && defined(_TARGET_ARM_)
-struct GenTreeMulLong : public GenTreeOp
+struct GenTreeMultiRegOp : public GenTreeOp
 {
     regNumber gtOtherReg;
 
-    GenTreeMulLong(var_types type, GenTreePtr op1, GenTreePtr op2) : GenTreeOp(GT_MUL_LONG, type, op1, op2)
+    GenTreeMultiRegOp(genTreeOps oper, var_types type, GenTreePtr op1, GenTreePtr op2)
+        : GenTreeOp(oper, type, op1, op2), gtOtherReg(REG_NA)
     {
     }
 
 #if DEBUGGABLE_GENTREE
-    GenTreeMulLong() : GenTreeOp()
+    GenTreeMultiRegOp() : GenTreeOp()
     {
     }
 #endif
@@ -5694,7 +5698,7 @@ inline bool GenTree::IsMultiRegNode() const
     }
 
 #if !defined(LEGACY_BACKEND) && defined(_TARGET_ARM_)
-    if (gtOper == GT_MUL_LONG || OperIsPutArgSplit())
+    if (gtOper == GT_MUL_LONG || gtOper == GT_PUTARG_REG || gtOper == GT_COPY || OperIsPutArgSplit())
     {
         return true;
     }
--- a/src/jit/lower.cpp
+++ b/src/jit/lower.cpp
@@ -1008,7 +1008,7 @@ GenTreePtr Lowering::NewPutArg(GenTreeCall* call, GenTreePtr arg, fgArgTabEntryP
 #endif // FEATURE_MULTIREG_ARGS
 #endif // not defined(FEATURE_UNIX_AMD64_STRUCT_PASSING)
             {
-                putArg = comp->gtNewOperNode(GT_PUTARG_REG, type, arg);
+                putArg = comp->gtNewPutArgReg(type, arg);
             }
         }
         else
@@ -1224,12 +1224,13 @@ void Lowering::LowerArg(GenTreeCall* call, GenTreePtr* ppArg)
     {
 
 #ifdef _TARGET_ARMARCH_
-        // For vararg call, reg args should be all integer.
+        // For vararg call or on armel, reg args should be all integer.
         // Insert a copy to move float value to integer register.
         if ((call->IsVarargs() || comp->opts.compUseSoftFP) && varTypeIsFloating(type))
         {
-            var_types  intType = (type == TYP_DOUBLE) ? TYP_LONG : TYP_INT;
-            GenTreePtr intArg  = comp->gtNewOperNode(GT_COPY, intType, arg);
+            var_types intType = (type == TYP_DOUBLE) ? TYP_LONG : TYP_INT;
+
+            GenTreePtr intArg = new (comp, GT_COPY) GenTreeCopyOrReload(GT_COPY, intType, arg);
 
             info->node = intArg;
             ReplaceArgWithPutArgOrCopy(ppArg, intArg);
--- a/src/jit/lsra.cpp
+++ b/src/jit/lsra.cpp
@@ -134,11 +134,17 @@ void lsraAssignRegToTree(GenTreePtr tree, regNumber reg, unsigned regIdx)
         tree->gtRegNum = reg;
     }
 #if defined(_TARGET_ARM_)
-    else if (tree->OperGet() == GT_MUL_LONG)
+    else if (tree->OperGet() == GT_MUL_LONG || tree->OperGet() == GT_PUTARG_REG)
     {
         assert(regIdx == 1);
-        GenTreeMulLong* mul = tree->AsMulLong();
-        mul->gtOtherReg     = reg;
+        GenTreeMultiRegOp* mul = tree->AsMultiRegOp();
+        mul->gtOtherReg        = reg;
+    }
+    else if (tree->OperGet() == GT_COPY)
+    {
+        assert(regIdx == 1);
+        GenTreeCopyOrReload* copy = tree->AsCopyOrReload();
+        copy->gtOtherRegs[0]      = reg;
     }
     else if (tree->OperGet() == GT_PUTARG_SPLIT)
     {
@@ -4139,6 +4145,9 @@ void LinearScan::buildRefPositionsForNode(GenTree*                  tree,
     // push defs
     LocationInfoList locationInfoList;
     LsraLocation     defLocation = currentLoc + 1;
+#ifdef ARM_SOFTFP
+    regMaskTP remainingUseCandidates = useCandidates;
+#endif
     for (int i = 0; i < produce; i++)
     {
         regMaskTP currCandidates = candidates;
@@ -4160,7 +4169,15 @@ void LinearScan::buildRefPositionsForNode(GenTree*                  tree,
             currCandidates = genFindLowestReg(candidates);
             candidates &= ~currCandidates;
         }
-#endif
+#ifdef ARM_SOFTFP
+        // If oper is GT_PUTARG_REG, set bits in useCandidates must be in sequential order.
+        else if (tree->OperGet() == GT_PUTARG_REG || tree->OperGet() == GT_COPY)
+        {
+            useCandidates = genFindLowestReg(remainingUseCandidates);
+            remainingUseCandidates &= ~useCandidates;
+        }
+#endif // ARM_SOFTFP
+#endif // _TARGET_ARM_
 
         if (interval == nullptr)
         {
--- a/src/jit/lsraarm.cpp
+++ b/src/jit/lsraarm.cpp
@@ -727,6 +727,35 @@ void Lowering::TreeNodeInfoInit(GenTree* tree)
             info->internalIntCount = 1;
             break;
 
+        case GT_COPY:
+            info->srcCount = 1;
+            info->dstCount = 1;
+#ifdef ARM_SOFTFP
+            // This case currently only occurs for double types that are passed as TYP_LONG;
+            // actual long types would have been decomposed by now.
+            if (tree->TypeGet() == TYP_LONG)
+            {
+                info->dstCount = 2;
+            }
+#endif
+            break;
+
+        case GT_PUTARG_REG:
+#ifdef ARM_SOFTFP
+            // This case currently only occurs for double types that are passed as TYP_LONG;
+            // actual long types would have been decomposed by now.
+            if (tree->TypeGet() == TYP_LONG)
+            {
+                info->srcCount = 2;
+            }
+            else
+#endif
+            {
+                info->srcCount = 1;
+            }
+            info->dstCount = info->srcCount;
+            break;
+
         default:
 #ifdef DEBUG
             char message[256];
@@ -744,7 +773,6 @@ void Lowering::TreeNodeInfoInit(GenTree* tree)
         case GT_CLS_VAR_ADDR:
         case GT_IL_OFFSET:
         case GT_CNS_INT:
-        case GT_PUTARG_REG:
         case GT_PUTARG_STK:
         case GT_LABEL:
         case GT_PINVOKE_PROLOG:
@@ -752,7 +780,6 @@ void Lowering::TreeNodeInfoInit(GenTree* tree)
         case GT_SETCC:
         case GT_MEMORYBARRIER:
         case GT_OBJ:
-        case GT_COPY:
         case GT_PUTARG_SPLIT:
             info->dstCount = tree->IsValue() ? 1 : 0;
             if (kind & (GTK_CONST | GTK_LEAF))
--- a/src/jit/lsraarmarch.cpp
+++ b/src/jit/lsraarmarch.cpp
@@ -403,7 +403,17 @@ void Lowering::TreeNodeInfoInitPutArgReg(
     info.srcCount++;
 
     // Set the register requirements for the node.
-    const regMaskTP argMask = genRegMask(argReg);
+    regMaskTP argMask = genRegMask(argReg);
+#ifdef ARM_SOFTFP
+    // If type of node is `long` then it is actually `double`.
+    // The actual `long` types must have been transformed as a field list with two fields.
+    if (node->TypeGet() == TYP_LONG)
+    {
+        info.srcCount++;
+        assert(genRegArgNext(argReg) == REG_NEXT(argReg));
+        argMask |= genRegMask(REG_NEXT(argReg));
+    }
+#endif // ARM_SOFTFP
     node->gtLsraInfo.setDstCandidates(m_lsra, argMask);
     node->gtLsraInfo.setSrcCandidates(m_lsra, argMask);
 
--- a/src/jit/codegenarm.cpp
+++ b/src/jit/codegenarm.cpp
@@ -1800,7 +1800,7 @@ void CodeGen::genStoreLongLclVar(GenTree* treeNode)
     {
         assert((op1->gtFlags & GTF_MUL_64RSLT) != 0);
 
-        GenTreeMulLong* mul = op1->AsMulLong();
+        GenTreeMultiRegOp* mul = op1->AsMultiRegOp();
 
         // Stack store
         getEmitter()->emitIns_S_R(ins_Store(TYP_INT), emitTypeSize(TYP_INT), mul->gtRegNum, lclNum, 0);
@@ -1818,8 +1818,9 @@ void CodeGen::genStoreLongLclVar(GenTree* treeNode)
 // Return Value:
 //    None.
 //
-void CodeGen::genCodeForMulLong(GenTreeMulLong* node)
+void CodeGen::genCodeForMulLong(GenTreeMultiRegOp* node)
 {
+    assert(node->OperGet() == GT_MUL_LONG);
     genConsumeOperands(node);
     GenTree*    src1 = node->gtOp1;
     GenTree*    src2 = node->gtOp2;
--- a/src/jit/codegenarmarch.cpp
+++ b/src/jit/codegenarmarch.cpp
@@ -191,7 +191,7 @@ void CodeGen::genCodeForTreeNode(GenTreePtr treeNode)
 
 #ifdef _TARGET_ARM_
         case GT_MUL_LONG:
-            genCodeForMulLong(treeNode->AsMulLong());
+            genCodeForMulLong(treeNode->AsMultiRegOp());
             break;
 #endif // _TARGET_ARM_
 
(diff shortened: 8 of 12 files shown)


src/jit/lsra.cpp
```diff
@@ -4148,6 +4161,14 @@ void LinearScan::buildRefPositionsForNode(GenTree*                  tree,
             useCandidates  = allRegs(registerType);
         }
 
+#ifdef ARM_SOFTFP
```
wateret: This part is kind of pair with lin 421 at laraarmarch.cpp, but I am not sure about this change. @caroleidt this is the part that I was wondering from the our last conversation #12293.

src/jit/lsraarmarch.cpp
```diff
@@ -403,13 +403,22 @@ void Lowering::TreeNodeInfoInitPutArgReg(
...
+        argMask |= genRegMask(REG_NEXT(argReg));
+    }
+#endif // ARM_SOFTFP
     node->gtLsraInfo.setDstCandidates(m_lsra, argMask);
     node->gtLsraInfo.setSrcCandidates(m_lsra, argMask);
 
     // To avoid redundant moves, have the argument operand computed in the
     // register in which the argument is passed to the call.
-    node->gtOp.gtOp1->gtLsraInfo.setSrcCandidates(m_lsra, m_lsra->getUseCandidates(node));
+    node->gtOp.gtOp1->gtLsraInfo.setDstCandidates(m_lsra, m_lsra->getUseCandidates(node));
```
wateret: To make the other changes(the one with my comment) work, I thought we need this change. Maybe I am still getting something wrong.

wateret:
@dotnet/jit-contrib @CarolEidt Please take a look.

src/jit/gentree.h
```diff
@@ -1628,6 +1628,9 @@ struct GenTree
 #ifdef FEATURE_SIMD
             case GT_SIMD:
 #endif // !FEATURE_SIMD
+#ifdef _TARGET_ARM_
+            case GT_PUTARG_REG:
+#endif
                 return true;
```
BruceForstall: `GT_PUTARG_REG` is defined as `GTK_UNOP`, so I don't see why this is necessary. (i.e., it's not a binary node in the first place)

src/jit/gentree.h
```diff
@@ -3803,16 +3806,17 @@ struct GenTreeCmpXchg : public GenTree
...
 #if !defined(LEGACY_BACKEND) && defined(_TARGET_ARM_)
-struct GenTreeMulLong : public GenTreeOp
+struct GenTreeMultiRegOp : public GenTreeOp
 {
     regNumber gtOtherReg;
 
-    GenTreeMulLong(var_types type, GenTreePtr op1, GenTreePtr op2) : GenTreeOp(GT_MUL_LONG, type, op1, op2)
+    GenTreeMultiRegOp(var_types type, GenTreePtr op1, GenTreePtr op2)
+        : GenTreeOp(GT_MUL_LONG, type, op1, op2), gtOtherReg(REG_NA)
     {
```
CarolEidt: I would assume that, if you changing this from `GenTreeMulLong` to `GenTreeMultiRegOp`, you expect to use this more generally than just for `GT_MUL_LONG`. If that's not the case, then this should be kept as `GenTreeMulLong`. If that is the case, then you shouldn't have an implicit oper on the constructor.

CarolEidt:
>     if (IsMultiRegCall())

Not directly related to this PR, but it would seem that this should return true for `GT_COPY` for `FEATURE_MULTIREG_RET`.

---
Refers to: src/jit/gentree.h:5471 in e35a840. [](commit_id = e35a840b793160b0f9047c71338931b564dafbad, deletion_comment = False)

BruceForstall:
Can you give an example of the IR (JitDump) before your change? I thought the code already forced all double to long in fgMorphArgs for armel, so I'm not sure why we need special handling for double.

src/jit/lower.cpp
```diff
@@ -960,7 +960,16 @@ GenTreePtr Lowering::NewPutArg(GenTreeCall* call, GenTreePtr arg, fgArgTabEntryP
 #endif // FEATURE_MULTIREG_ARGS
 #endif // not defined(FEATURE_UNIX_AMD64_STRUCT_PASSING)
         {
-            putArg = comp->gtNewOperNode(GT_PUTARG_REG, type, arg);
+#ifdef _TARGET_ARM_
+            if (comp->opts.compUseSoftFP && type == TYP_DOUBLE)
+            {
+                putArg = new (comp, GT_PUTARG_REG) GenTreeMultiRegOp(type, arg, nullptr);
```
CarolEidt: Would this also be required for passing `TYP_LONG`? Also, I think you need to be consistent in using `GenTreeMultiRegOp` for `GT_PUTARG_REG`, or you should use a different oper (e.g. `GT_PUTARG_MULTIREG`), since the JIT (mostly) depends on the oper being a "tag" for the `GenTree` type. That is, a given oper should always be on the same `GenTree` type.

src/jit/lsra.cpp
```diff
@@ -3596,7 +3602,14 @@ void LinearScan::buildRefPositionsForNode(GenTree*                  tree,
     int consume = info.srcCount;
     int produce = info.dstCount;
 
+#ifndef ARM_SOFTFP
     assert(((consume == 0) && (produce == 0)) || (ComputeAvailableSrcCount(tree) == consume));
+#else
+    // TODO-ARM-Cleanup: we need to change assert condition.
+    // For example, count and dst count of `vmov.d2i` do not match.
+    // The below call is temporally added just to supress warning message.
+    ComputeAvailableSrcCount(tree);
```
hqueue: Please take a look at [armel CI failure](https://ci.dot.net/job/dotnet_coreclr/job/master/job/armel_cross_release_tizen_prtest/3570/console).
You can use it with `DEBUG` defined or in `assert`, because `ComputeAvailableSrcCount()` is defined only when `DEBUG` is defined.

src/jit/lower.cpp
```diff
@@ -1175,12 +1184,29 @@ void Lowering::LowerArg(GenTreeCall* call, GenTreePtr* ppArg)
...
         if ((call->IsVarargs() || comp->opts.compUseSoftFP) && varTypeIsFloating(type))
         {
-            var_types  intType = (type == TYP_DOUBLE) ? TYP_LONG : TYP_INT;
-            GenTreePtr intArg  = comp->gtNewOperNode(GT_COPY, intType, arg);
+            isSoftFPDoubleRegArg = (comp->opts.compUseSoftFP && type == TYP_DOUBLE && info->regNum != REG_STK);
+            var_types  intType   = (type == TYP_DOUBLE) ? TYP_LONG : TYP_INT;
+
+#ifdef _TARGET_ARM_
+            // We need gtOtherRegs[] for the case - copy from a double register to two int registers
+            GenTreePtr intArg = new (comp, GT_COPY) GenTreeCopyOrReload(GT_COPY, intType, arg);
```
CarolEidt: You don't need to `ifdef` this. On other targets, although `GenTreeCopyOrReload` is derived from `GenTreeUnOp` it has no additional fields. This is how LSRA generates these nodes.

src/jit/lsra.cpp
```diff
@@ -3596,7 +3602,14 @@ void LinearScan::buildRefPositionsForNode(GenTree*                  tree,
     int consume = info.srcCount;
     int produce = info.dstCount;
 
+#ifndef ARM_SOFTFP
     assert(((consume == 0) && (produce == 0)) || (ComputeAvailableSrcCount(tree) == consume));
+#else
+    // TODO-ARM-Cleanup: we need to change assert condition.
```
CarolEidt: Instead of `TODO-ARM-Cleanup` this should be `TODO-ARM-Bug`

src/jit/lsra.cpp
CarolEidt: At this point, the node has been created, so the `gtOtherRegs` values should be correct, right?  I think you could assert `tree->gtOtherRegs[i] == genRegNumFromMask(currCandidates)` below. It would also be good to add a comment that this code, and the code in lsraarmarch.cpp rely on the fact that the registers are in sequential order.

---
In reply to: [122643874](https://github.com/dotnet/coreclr/pull/12360#discussion_r122643874) [](ancestors = 122643874)

src/jit/lsra.cpp
```diff
@@ -10355,6 +10376,23 @@ void TreeNodeInfo::Initialize(LinearScan* lsra, GenTree* node, LsraLocation loca
     else
     {
         dstCandidates = genRegMask(node->gtRegNum);
+#ifdef ARM_SOFTFP
+        regNumber reg2 = REG_NA;
+        if (node->OperGet() == GT_PUTARG_REG)
```
CarolEidt: Shouldn't this include `GT_MUL_LONG` as well?

src/jit/lsraarm.cpp
```diff
@@ -727,6 +727,43 @@ void Lowering::TreeNodeInfoInit(GenTree* tree)
             info->internalIntCount = 1;
             break;
 
+        case GT_COPY:
+        case GT_PUTARG_REG:
+            // TODO-ARM-Cleanup : the logic was copied from general case,
+            // Only the exceptional case was added - when src count should be 2.
+            info->dstCount = tree->IsValue() ? 1 : 0;
+            if (kind & (GTK_CONST | GTK_LEAF))
```
CarolEidt: Both of these are known to be `GTK_UNOP`, so this is unnecessary (`GTK_SMPOP` includes `GTK_UNOP`).  I would suggest adding a method to GenTreeMultiRegOp such as `GetRegCount()`, and then you could simply do:
```
info->srcCount = tree->AsGenTreeMultiRegOp()->GetRegCount();
```
You could also assert based on type that `info->srcCount` has the right value.

src/jit/lsraarmarch.cpp
```diff
@@ -403,13 +403,22 @@ void Lowering::TreeNodeInfoInitPutArgReg(
...
     // Set the register requirements for the node.
-    const regMaskTP argMask = genRegMask(argReg);
+    regMaskTP argMask = genRegMask(argReg);
+#ifdef ARM_SOFTFP
+    // If type of node is `long` then it is actually `double`.
+    // The actual `long` types must have been transformed as a field list with two fields.
+    if (node->TypeGet() == TYP_LONG)
+    {
+        info.srcCount++;
+        argMask |= genRegMask(REG_NEXT(argReg));
```
CarolEidt: This would be a good place to assert that `genRegArgNext(argReg == REG_NEXT(argReg)`, and note that the register allocator depends on the arg registers being in sequential order.

src/jit/lower.cpp
CarolEidt: I realize now that `TYP_LONG` is passed using `GT_FIELD_LIST`, so disregard the first question. However, the other comment still applies.

---
In reply to: [123120550](https://github.com/dotnet/coreclr/pull/12360#discussion_r123120550) [](ancestors = 123120550)

CarolEidt:
:clock1:

src/jit/gentree.h
wateret: I have changed `GT_PUTARG_REG` to be `GenTreeMultiRegOp`(a subclass of `GenTreeOp`) for ARM. That's why it was necessary.

src/jit/gentree.h
wateret: Oh I have missed that, I'll fix it.

src/jit/lower.cpp
wateret: It looks like with FEATURE_MULTIREG_RET, there is an additional field. In gentree.h,

```cpp
4923 struct GenTreeCopyOrReload : public GenTreeUnOp
4924 {
4925 #if FEATURE_MULTIREG_RET
4926     // State required to support copy/reload of a multi-reg call node.
4927     // First register is is always given by gtRegNum.
4928     //
4929     regNumberSmall gtOtherRegs[MAX_RET_REG_COUNT - 1];
4930 #endif
```

I wanted to use `gtOtherRegs[0]` as the second register when we copy double register.

src/jit/lsra.cpp
wateret: I'll fix it. Maybe It'd be nicer to change assert condition for ARM.

src/jit/lsra.cpp
wateret: @CarolEidt IMHO `gtOtherRegs`  values are invalidated from Lowering phaseby `node->gtClearReg(comp)`.

src/jit/lsra.cpp
wateret: Yes I think so. However there was no code for `GT_MUL_LONG` here so I'm not sure if I should add that. I'm not sure if there is a case that GT_MUL_LONG node is actually generated. So I left it out.

src/jit/lsraarm.cpp
wateret: @CarolEidt I just tried that but there is no easy way for the tree node to get register count. The reason is kind of similar to https://github.com/dotnet/coreclr/pull/12360/files#r123158463, that is `gtRegNum` and `gtOtherReg` are invalidated from Lowering phase.

wateret:
@BruceForstall JitDump(Main function) of `DblAdd.exe`: [dump_dbladd_armel.txt](https://github.com/dotnet/coreclr/files/1090963/dump_dbladd_armel.txt)
The problem shows at LSRA, not morphing.

src/jit/lower.cpp
CarolEidt: > It looks like with FEATURE_MULTIREG_RET, there is an additional field. In gentree.h

Right. What I meant was that we should always use `GenTreeCopyOrReload` here even without `FEATURE_MULTIREG_RET`, in which case there is no additional cost relative to `GenTreeUnOp`, and it ensures oper/struct consistency for the `GenTree` nodes.

src/jit/lsra.cpp
CarolEidt: If you don't include that node because you don't expect it, then perhaps in the else clause you could assert that it is not a multireg node.

wateret:
@CarolEidt Added if condition for `GT_MUL_LONG` in lsra.cpp:10379 and removed the unnecessary part(1e977ab).

src/jit/lower.cpp
```diff
@@ -960,7 +960,16 @@ GenTreePtr Lowering::NewPutArg(GenTreeCall* call, GenTreePtr arg, fgArgTabEntryP
 #endif // FEATURE_MULTIREG_ARGS
 #endif // not defined(FEATURE_UNIX_AMD64_STRUCT_PASSING)
         {
-            putArg = comp->gtNewOperNode(GT_PUTARG_REG, type, arg);
+#ifdef _TARGET_ARM_
+            if (comp->opts.compUseSoftFP && type == TYP_DOUBLE)
+            {
+                putArg = new (comp, GT_PUTARG_REG) GenTreeMultiRegOp(GT_PUTARG_REG, type, arg, nullptr);
```
CarolEidt: In order to reduce `#ifdef`s here, it would be good to add a `Compiler::gtNewPutArgReg()` method.

src/jit/lower.cpp
```diff
@@ -1179,8 +1188,14 @@ void Lowering::LowerArg(GenTreeCall* call, GenTreePtr* ppArg)
...
         if ((call->IsVarargs() || comp->opts.compUseSoftFP) && varTypeIsFloating(type))
         {
-            var_types  intType = (type == TYP_DOUBLE) ? TYP_LONG : TYP_INT;
-            GenTreePtr intArg  = comp->gtNewOperNode(GT_COPY, intType, arg);
+            var_types intType    = (type == TYP_DOUBLE) ? TYP_LONG : TYP_INT;
+
+#ifdef _TARGET_ARM_
+            // We need gtOtherRegs[] for the case - copy from a double register to two int registers
+            GenTreePtr intArg = new (comp, GT_COPY) GenTreeCopyOrReload(GT_COPY, intType, arg);
+#else  // !_TARGET_ARM_
```
CarolEidt: This should not be `#ifdef _TARGET_ARM_`. It should always use `GenTreeCopyOrReload()`

src/jit/lsra.cpp
CarolEidt: I see. There should at least be a comment about the assumption that the arg registers are in sequential order.

src/jit/lsraarmarch.cpp
CarolEidt: This should be `setSrcCandidates()` because `node` is using its op1 as a source, and setting the requirements on the candidates accordingly.

src/jit/lower.cpp
wateret: I see. I'll update it.

src/jit/lsraarmarch.cpp
wateret: Then does it mean that the code "lsra.cpp:4163"(in the loop for pushing produces) should be done in consume loop?

CarolEidt: Yes. This is a somewhat tricky case that I don't think we've handled before - that is, where we are consuming sources from a node that produces multiple registers, and where they have fixed reg requirements. This would need to be handled near line 4013 in your version of lsra.cpp, where it sets `fixedAssignment` assuming that there will only be a single register in that case.

wateret:
@CarolEidt Please take a look at the last commit. However it does not work. 

It crashes with:

```
buildIntervals second part ========

NEW BLOCK BB01
<RefPosition #0   @0   RefTypeBB BB01 regmask=[]>
at start of tree, map contains: {  }
  N003.                    IL_OFFSET IL offset: 0x0 REG NA
  consume=0 produce=0
     op

Def candidates [allInt], Use candidates [allInt]
at start of tree, map contains: {  }
  N005. t5              =  CNS_DBL   1.0000000000000000 REG NA
  consume=0 produce=1
t5 = op
Interval  1: RefPositions {} physReg:NA Preferences=[allInt]
<RefPosition #1   @5   RefTypeDef <Ivl:1 internal> CNS_DBL BB01 regmask=[allInt]>
Interval  2: RefPositions {} physReg:NA Preferences=[allInt]
<RefPosition #2   @5   RefTypeDef <Ivl:2 internal> CNS_DBL BB01 regmask=[allInt]>

<RefPosition #3   @5   RefTypeUse <Ivl:1 internal> CNS_DBL BB01 regmask=[allInt] last>
<RefPosition #4   @5   RefTypeUse <Ivl:2 internal> CNS_DBL BB01 regmask=[allInt] last>
Def candidates [allDouble], Use candidates [allDouble]
Interval  3: RefPositions {} physReg:NA Preferences=[allDouble]
<RefPosition #5   @6   RefTypeDef <Ivl:3> CNS_DBL BB01 regmask=[allDouble]>
<RefPosition #5   @6   RefTypeDef <Ivl:3> CNS_DBL BB01 regmask=[allDouble]>
at start of tree, map contains: { N005. CNS_DBL -> (6.N005) }
  N007. t7              =  COPY
  consume=2 produce=1
t7 = op
t6 <RefPosition #6   @7   RefTypeFixedReg <Reg:r0 > BB01 regmask=[r0]>
<RefPosition #7   @7   RefTypeUse <Ivl:3> BB01 regmask=[allDouble] last>

Assert failure(PID 21637 [0x00005485], Thread: 21637 [0x5485]): Assertion failed 'iterator != tree->OperandsEnd()' in 'BringUpTest:Main():int'
(IL size 62)

    File: /opt/code/github/hqueue/coreclr/src/jit/lsra.cpp Line: 3947
    Image: /opt/usr/home/owner/dotnet/ttt/corerun
```

Actually I don't understand what is going on in the consume loop(the one that starts with `for (int useIndex = 0; useIndex < consume; useIndex++)`). Could you please explain what is being done from that loop? It looks way more complicated than the produce loop.

CarolEidt:
> Could you please explain what is being done from that loop? It looks way more complicated than the produce loop.

It is indeed more complicated than the produce loop. This is because the current node has to match its expected source operands to the previously encountered def nodes in `operandDefs`, and this in turn is made somewhat more complicated by the fact that a single operand may indirectly represent multiple results, if it is contained. So, the outer loop is looping over the number of expected registers used. The first "if" does some bookkeeping on the map of previously seen nodes (`operandDefs`) and then advances to the next operand. The following code goes on to take care of any special handling or constraints on the use.

The assert you are seeing is because `operandDefs` is empty, but it has only found one of two expected source operands. You can see this because the dump indicates that it is handling a `COPY` node that expects to consume 2 sources, but the "at start of tree" message shows a single node:
```
at start of tree, map contains: { N005. CNS_DBL -> (6.N005) }
  N007. t7              =  COPY
  consume=2 produce=1
t7 = op
t6 <RefPosition #6   @7   RefTypeFixedReg <Reg:r0 > BB01 regmask=[r0]>
<RefPosition #7   @7   RefTypeUse <Ivl:3> BB01 regmask=[allDouble] last>
```
So, either the `CNS_DBL` node needs to also produce two registers, or the `COPY` needs to consume just one.

wateret:
@CarolEidt With your comment, the last commit looked totally wrong so I removed that commit and added fixes. It also reflects your previous review comments.

wateret:
@dotnet-bot Test Tizen armel Cross Release Build Please

src/jit/gentree.cpp
```diff
@@ -7621,6 +7621,39 @@ GenTree* Compiler::gtNewBlkOpNode(
     return result;
 }
 
+//------------------------------------------------------------------------
+// gtNewPutArgReg: Creates a new Obj node.
```
CarolEidt: This should say "Creates a new PutArgReg node." (or "GT_PUTARG_REG" node, or something like that).

src/jit/gentree.cpp
```diff
@@ -7621,6 +7621,39 @@ GenTree* Compiler::gtNewBlkOpNode(
...
+//
+// Arguments:
+//    type - The actual type of the argument
+//    arg  - The argument node
+//
+// Return Value:
+//    Returns a node representing the struct value at the given address.
```
CarolEidt: This comment also needs to be updated.

src/jit/lsra.cpp
```diff
@@ -3596,7 +3602,13 @@ void LinearScan::buildRefPositionsForNode(GenTree*                  tree,
     int consume = info.srcCount;
     int produce = info.dstCount;
 
+#ifndef ARM_SOFTFP
     assert(((consume == 0) && (produce == 0)) || (ComputeAvailableSrcCount(tree) == consume));
+#else
+    assert(((consume == 0) && (produce == 0)) || ((consume == 1) && (produce == 2) && tree->OperGet() == GT_COPY &&
```
CarolEidt: This is not correct. It should always be the case that either the node consumes and produces no registers, or the latter condition holds. If this doesn't hold, then you have the wrong count on one of the nodes. It looks like the code below sets the `srcCount` to 1 and the `dstCount` to 2 for this case of `GT_COPY`, so why is this special case needed?

src/jit/lsraarm.cpp
```diff
@@ -727,6 +727,31 @@ void Lowering::TreeNodeInfoInit(GenTree* tree)
...
+            if (tree->TypeGet() == TYP_LONG) // Actually was `double` type
+            {
+                info->dstCount = 2;
+            }
+#endif
+            break;
+
+        case GT_PUTARG_REG:
+#ifdef ARM_SOFTFP
+            if (tree->TypeGet() == TYP_LONG) // Actually was `double` type
```
CarolEidt: How is an actual long type parameter passed? Wouldn't that also be `TYP_LONG`? Or would it have been decomposed into to `TYP_INT` parameters?

src/jit/lsraarmarch.cpp
```diff
@@ -403,7 +403,17 @@ void Lowering::TreeNodeInfoInitPutArgReg(
     info.srcCount++;
 
     // Set the register requirements for the node.
-    const regMaskTP argMask = genRegMask(argReg);
+    regMaskTP argMask = genRegMask(argReg);
```
CarolEidt: In looking through the code, I can't see where this method (`Lowering::TreeNodeInfoInitPutArgReg()`) is called. It appears that for arm64 it is handled with the default case, and for arm it is handled directly in `Lowering::TreeNodeInfoInit()`.

src/jit/gentree.cpp
wateret: Sorry I missed that.

src/jit/lsraarm.cpp
wateret: As far as I can see, `TYP_LONG`s must have been decomposed like you said.

src/jit/lsra.cpp
wateret: It is only for last drafts, now we don't need that. Thanks!

src/jit/lsraarmarch.cpp
wateret: It is called from `Lowering::TreeNodeInfoInitCall(GenTreeCall* call)`. In `Lowering::TreeNodeInfoInitPutArgReg()`, not only it sets the PutArgReg node's candidates but also modifies the Call node's srcCount(`TreeNodeInfo& info` is the call node's info).

src/jit/lsraarm.cpp
CarolEidt: It might be worth an extra comment

CarolEidt: (e.g. "This case currently only occurs for double types that are passed as TYP_LONG; actual long types would have been decomposed by now."

src/jit/lsraarmarch.cpp
CarolEidt: I see. We may need to modify that in future, because once we are doing the containment analysis in `Lowering`, and then doing `TreeNodeInfoInit` on the fly, we need to be able to set all the counts and candidates in a forward walk (without modifying a node's info from its parent). But this works for now.

CarolEidt:
@wateret - This looks good; thanks for the additional comments. It seems that there are now conflicts that need to be resolved and then we can merge this.

wateret:
@CarolEidt I just resolved conflicts and squashed commits to clean up the commit message. Thank you for your review :D

wateret:
@dotnet-bot Test Tizen armel Cross Debug Build please

PR #12362, opened 2017-06-19
Enable MemoryFailPoint functionality on Unix

hughbe:
As per comment in https://github.com/dotnet/coreclr/issues/7807

Corefx has a disabled test this issue

### DIFF
--- a/src/mscorlib/src/System/Runtime/MemoryFailPoint.cs
+++ b/src/mscorlib/src/System/Runtime/MemoryFailPoint.cs
@@ -158,7 +158,6 @@ namespace System.Runtime
                 throw new ArgumentOutOfRangeException(nameof(sizeInMegabytes), SR.ArgumentOutOfRange_NeedNonNegNum);
             Contract.EndContractBlock();
 
-#if !FEATURE_PAL // Remove this when CheckForAvailableMemory is able to provide legitimate estimates
             ulong size = ((ulong)sizeInMegabytes) << 20;
             _reservedMemory = size;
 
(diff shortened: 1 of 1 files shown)


janvorli:

janvorli:

janvorli:

hughbe:
@janvorli are you happy with this? thanks

danmosemsft:
@hughbe can you resolve conflicts?
@janvorli do you sign off?

janvorli:
@danmosemsft I am reviewing it now.

janvorli:
@danmosemsft, @hughbe we still cannot enable this for Unix. There is one additional thing that would not work. The VirtualQuery implementation in the PAL never returns size of the free region (always returns 0 for them), since it has no knowledge about free regions ranges.  That means that the MemFreeAfterAddress function would end up running in an infinite loop.

danmosemsft:
@janvorli is this feasible to fix,? if so I will open an issue.

janvorli:
@danmosemsft I think it would be feasible, so creating an issue for it sounds good. We could possibly augment the VirtualQuery implementation in PAL to read and parse the /proc/self/maps memory map when the address is not in the allocated blocks we keep track of (and the equivalent of that on OSX). 
If that would cause perf issues in other places where VirtualQuery is used, we could create a new PAL function that would be used in the MemFreeAfterAddress managed function only and PInvoke into that.

karelz:
What is status of this PR? What are the next steps to move it forward? Is it blocked on something?

janvorli:
@karelz this PR should be closed. We have #13560 to make the necessary fixes so that it could work on Unix.

PR #12365, opened 2017-06-19
Fixing CompareInfo AssemblyMode.Full serialization between Core <--> Desktop

ViktorHofer:
Relates to https://github.com/dotnet/corefx/pull/21223

@tarekgh I'm not sure which steps are exactly necessary to make it fully functional. This is just a proposal for you. Please correct me if I'm missing something.

I will create a port for 2.0 after feedback round finished.

cc @stephentoub

### DIFF
--- a/src/mscorlib/src/System/Globalization/CompareInfo.cs
+++ b/src/mscorlib/src/System/Globalization/CompareInfo.cs
@@ -75,8 +75,8 @@ namespace System.Globalization
         // _invariantMode is defined for the perf reason as accessing the instance field is faster than access the static property GlobalizationMode.Invariant
         [NonSerialized]
         private readonly bool _invariantMode = GlobalizationMode.Invariant;
-
-        private int culture; // Do not rename (binary serialization)
+        
+        private int culture; // Do not rename (binary serialization). The fields sole purpose is to support Desktop serialization.
 
         internal CompareInfo(CultureInfo culture)
         {
--- a/src/mscorlib/src/System/Globalization/CompareInfo.cs
+++ b/src/mscorlib/src/System/Globalization/CompareInfo.cs
@@ -76,6 +76,8 @@ namespace System.Globalization
         [NonSerialized]
         private readonly bool _invariantMode = GlobalizationMode.Invariant;
 
+        private int culture; // Do not rename (binary serialization)
+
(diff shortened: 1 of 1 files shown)


tarekgh:
@ViktorHofer I am not sure if we are really want to support Whidbey serialization in net core. This is change may help a little in some of Whidbey cases but still there is other cases will be broken anyway. For example, on Windows 10, there are many cultures introduced with LCID equal to 0x1000 (named as custom culture ID). if you try to serialize and deserilaize such cultures between net core and Whidbey it'll fail. 

If we are not going to have a complete solution for Whidbey I would suggest it is better to say we don't support serialization with Whidbey.

ViktorHofer:
@danmosemsft can answer that better

danmosemsft:
@tarekgh What is the support for serializing CompareInfo between 4.x and 2.0 ? Can we do that much?

Sounds like if we don't want to support it, we at least need this field so that serialization doesn't fail always.

tarekgh:
> What is the support for serializing CompareInfo between 4.x and 2.0 ? Can we do that much?

It is broken for the same reason

> Sounds like if we don't want to support it, we at least need a dummy field so that serialization doesn't fail always.

I am not sure why this is happening if the serialized stream coming from full desktop v4+or net core. on net core we handle the deserialization manually so I don't expect this fail

src/mscorlib/src/System/Globalization/CompareInfo.cs
```diff
@@ -76,6 +76,8 @@ public partial class CompareInfo : IDeserializationCallback
         [NonSerialized]
         private readonly bool _invariantMode = GlobalizationMode.Invariant;
 
+        private int culture; // Do not rename (binary serialization)
```
tarekgh: please add a comment that this field is there only to make the full framework happy with the serialization

tarekgh:
LGTM.

ViktorHofer:
Thanks Tarek!

ViktorHofer:
@tarekgh Added the comment. Please sign-off

tarekgh:
LGTM. thanks @ViktorHofer

ViktorHofer:
Isn't there a timeout for the OSX build?

PR #12368, opened 2017-06-19
[CoreRT armel] fix for PINVOKE calls in CoreRT

sergign60:


sergign60:
CC: @jkotas @Dmitri-Botcharnikov @BredPet @mskvortsov

sergign60:
@jkotas please review

src/jit/codegencommon.cpp
```diff
@@ -8687,8 +8687,10 @@ void CodeGen::genFnProlog()
     {
         excludeMask |= RBM_PINVOKE_FRAME;
 
+#ifndef _ARM_
         assert((!compiler->opts.ShouldUsePInvokeHelpers()) || (compiler->info.compLvFrameListRoot == BAD_VAR_NUM));
```
jkotas: This does not look right. `compLvFrameListRoot` should not be allocated at all with ShouldUsePInvokeHelpers.

sergign60: @kotas Yes, you're right. I've made the mistake

Issue #11476, opened 2017-05-09
[x86, R2R] Jit/Regressions/JitBlue/DevDiv_278523 test failure


sandreenko:
This test contains such IL sequence:
```
// This sequence of IL repros the issue. Note that the `ldc.i8` is necessary (rather than an `ldc.i4` that is
// implicitly converted to a long byt the `add`).
ldloca.s V_1  <- native int
ldc.i8 4          <- int64
add                <- native int + int64
```

On x86 this sequence is not compiled and executed in CoreRun, because the method that contains it (Test64Bit) is called under condition that pointer size is 64 bits.

However, CrossGen has to compile it and it blows up on the assert that address should have native int size.

ECMA says that it is " an invalid CIL instruction" ([ecma](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-335.pdf#page328), page 328, Table III.2: Binary Numeric Operations).

Jit64 and RyuJit64 process it as correct sequence and execute this code without any problems.

I will add PEVerify and jit32 results soon.

sandreenko:
PTAL @dotnet/jit-contrib .
What is a contract for " an invalid CIL instruction"? Should we throw an exception? Could we execute such code? Could we execute it differently on different architects?

sandreenko:
PEVerify output:
```
[IL]: Error: [DevDiv_278523.exe : C::Test32Bit][offset 0x0000000C][found address of value 'T'] Expected numeric type on the stack.
[IL]: Error: [DevDiv_278523.exe : C::Test64Bit][offset 0x00000014][found address of value 'T'] Expected numeric type on the stack.
2 Error(s) Verifying DevDiv_278523.exe
```
So it shows the error for native int + int64.
Why does it complaint about native int + int32?

sandreenko:
jit32 ngen fails with the same assert.

CarolEidt:
> Why does it complaint about native int + int32?

It is a managed pointer (the address of a local), and although it is not invalid IL, it is unverifiable to add a managed pointer and an integer (so peverify appropriately gives an error).

> What is a contract for " an invalid CIL instruction"? Should we throw an exception?

We should invoke the `BADCODE` macro, which causes an exception to be thrown (by the VM I believe).

> Could we execute such code? Could we execute it differently on different architects?

In theory invalid IL should be independent of target architecture, but for historical reasons we have generally permitted IL like this that was previously permitted by the legacy 64-bit JIT.

[...]

PR #12374, opened 2017-06-19
Fixed arith operations with 'native int' and 'long' according to ECMA-335

rartemev:
Changes to fix the issue #11476

### DIFF
--- a/src/jit/importer.cpp
+++ b/src/jit/importer.cpp
@@ -11280,6 +11280,19 @@ void Compiler::impImportBlockCode(BasicBlock* block)
                 op2 = impPopStack().val;
                 op1 = impPopStack().val;
 
+#ifdef FEATURE_CORECLR
+                if (IsOperArithmetic(oper))
+                {
+                    const var_types tiOp1 = op1->TypeGet();
+                    const var_types tiOp2 = op2->TypeGet();
+
+                    if (tiOp1 == TYP_BYREF && (tiOp2 != TYP_BYREF && tiOp2 != TYP_INT))
+                    {
+                        BADCODE("Address plus non-int32 is not allowed. See ECMA-335 @ Table III.2");
+                    }
+                }
+#endif // FEATURE_CORECLR
+
 #if !CPU_HAS_FP_SUPPORT
                 if (varTypeIsFloating(op1->gtType))
                 {
--- a/src/jit/compiler.h
+++ b/src/jit/compiler.h
@@ -3098,6 +3098,7 @@ public:
     static bool IsTargetIntrinsic(CorInfoIntrinsics intrinsicId);
     static bool IsMathIntrinsic(CorInfoIntrinsics intrinsicId);
     static bool IsMathIntrinsic(GenTreePtr tree);
+    static bool IsOperArithmetic(genTreeOps oper);
 
 private:
     //----------------- Importing the method ----------------------------------
--- a/src/jit/compiler.hpp
+++ b/src/jit/compiler.hpp
@@ -4142,6 +4142,12 @@ inline bool Compiler::IsGcSafePoint(GenTreePtr tree)
     return false;
 }
 
+inline bool Compiler::IsOperArithmetic(genTreeOps oper)
+{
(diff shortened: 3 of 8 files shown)


rartemev:
@dotnet/jit-contrib please review

src/jit/importer.cpp
```diff
@@ -11226,6 +11226,23 @@ void Compiler::impImportBlockCode(BasicBlock* block)
 #endif // _TARGET_64BIT_
                 }
 
+#ifdef FEATURE_CORECLR
+                if ((oper == GT_ADD) || (oper == GT_SUB) || (oper == GT_MUL) || (oper == GT_DIV) || (oper == GT_MOD))
```
pgavlin: What about the unsigned versions of these operators? Also, do we need to take logical operators into account?

src/jit/importer.cpp
```diff
@@ -11226,6 +11226,23 @@ void Compiler::impImportBlockCode(BasicBlock* block)
...
+                {
+                    const typeInfo& tiOp1 = impStackTop(1).seTypeInfo;
+                    const typeInfo& tiOp2 = impStackTop().seTypeInfo;
+
+                    // check for structure too because typeof(ldloca struct) is struct
+                    if (tiOp1.IsNativeIntType() || tiOp1.IsStruct())
```
pgavlin: Nit: `if ((tiOp1.IsNativIntType() || tiOp1.IsStruct()) && tiOp2.IsType(TI_LONG))`

src/jit/importer.cpp
rartemev: I thought about that but since only those operation mentioned in the table, I am checking only for them.

How to determine that operation is unsigned?

src/jit/importer.cpp
rartemev: Fixed

src/jit/importer.cpp
pgavlin: > How to determine that operation is unsigned?

I think you may only be missing `GT_UMOD` and `GT_UDIV`. We might not even have those operators at this point, so it may not be necessary in any case.

CarolEidt: It doesn't make a lot of sense (i.e. less likely that it something that we would encounter), but for completeness, we should add `GT_UMOD` and `GT_UDIV`, as well as the logical operators. We can have  `GT_UMOD` and `GT_UDIV` in the importer, as they correspond to the IL operators `rem.un` and `div.un`.

pgavlin: In that case, it may be worth creating a static version of `GenTree::OperIsArithmetic` (which includes udiv/umod and the logical ops) and using that for the check.

rartemev: Done

rartemev:

src/jit/importer.cpp
```diff
@@ -11226,6 +11226,20 @@ void Compiler::impImportBlockCode(BasicBlock* block)
 #endif // _TARGET_64BIT_
                 }
 
+#ifdef FEATURE_CORECLR
+                if (IsOperArithmetic(oper))
+                {
+                    const typeInfo& tiOp1 = impStackTop(1).seTypeInfo;
+                    const typeInfo& tiOp2 = impStackTop().seTypeInfo;
+
+                    // check for structure too because typeof(ldloca struct) is struct
```
briansull: I don't see how the type of ldloca(struct) can be a struct, it would have to be a byref,

And in any case I don't think we want to fold the special case for struct (if is is needed) into the TYP_LONG case.
Make it a separate BADCODE check.

briansull: You also will need a test case for the struct case if you include that change

rartemev: Actually DevDiv_278523 test case tests exactly you are asking, (ldloca(struct) + ldc.i8)

rartemev:

rartemev:
@dotnet-bot test CI

rartemev:
@dotnet-bot test FreeBSD x64 Checked Build

danmosemsft:
@rartemev what is the status of this old PR?

rartemev:
@danmosemsft I had more important tasks so I am going to back to this problem soon.

PR #12382, opened 2017-06-20
Please include this line to enable build

denezt:
find_path(LLDB_H "lldb/API/LLDB.h" PATHS "/usr/lib/llvm-4.0/include")


hqueue:
@dotnet-bot
test Ubuntu16.04 arm Cross Debug Build

ellismg:
LGTM.  Thanks for the patch!  Since we don't yet do any builds with LLVM 4.0, did you run into any other issues?

denezt:
@ellismg Thanks for adding that pull request. Yes, I encounter so many problems and unclear error messages while using building `coreclr` on Ubuntu 17.04. Then it finally built after finding the version of LLVM my distro uses which is LLVM-4.0.

PR #12387, opened 2017-06-20
Fix systemMask on Linux

helloguo:
This PR fixes `systemMask`, which comes from `DWORD_PTR systemMask = GetFullAffinityMask(g_cpuCount); `on Linux. The issue was that the `systemMask` returns `0` instead of `0xFFFF` on a machine with 16 logical cores. In consequence, `GetCurrentProcessCpuCount` returns 64, which is supposed to be 16.

### DIFF
--- a/src/pal/src/numa/numa.cpp
+++ b/src/pal/src/numa/numa.cpp
@@ -146,7 +146,7 @@ NUMASupportInitialize()
         int numaNodesCount = numa_max_node() + 1;
 
         g_possibleCpuCount = numa_num_possible_cpus();
-        g_cpuCount = PAL_GetLogicalCpuCountFromOS();
+        g_cpuCount = 0;
         g_groupCount = 0;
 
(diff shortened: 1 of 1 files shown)


helloguo:
@janvorli PTAL

src/pal/src/numa/numa.cpp
```diff
@@ -143,7 +145,8 @@ NUMASupportInitialize()
         struct bitmask *mask = numa_allocate_cpumask();
         int numaNodesCount = numa_max_node() + 1;
 
-        g_cpuCount = numa_num_possible_cpus();
+        g_possibleCpuCount = numa_num_possible_cpus();
+        g_cpuCount = PAL_GetLogicalCpuCountFromOS();
```
janvorli: Can you please change the g_cpuCount computation so that it is initialized to zero here and nodeCpuCount is added to that in the loop below? While I assume it would get the same value, I would prefer getting all the information from the numa here for consistency reasons.

PR #12389, opened 2017-06-20
[Local GC] Remove a number of inclusions of handle table private headers from the VM

swgillespie:
This PR removes a number of occurrences where the VM was including GC headers. In most cases, nothing from these headers were being used. For the things that were being used from these headers:
* Uses of `Ref_TraceRefCountedHandles` were refactored to go through `IGCHandleManager::TraceRefCountedHandles`, which I added to the `IGCHandleManager` interface. This was not detected earlier because it only is used when `FEATURE_EVENT_TRACE` is defined.
* The `OBJECTREF_TO_UNCHECKED_OBJECTREF` and `UNCHECKED_OBJECTREF_TO_OBJECTREF` macros were moved to `gcinterface.h`, because they are used by both the VM and the GC. I don't think this is ideal. It seems like we should be moving away from using `OBJECTREF` when inside the GC.
* `GCHandleValidatePinnedObject` was renamed to `ValidatePinnedObject` and its prototype was moved out of `objecthandle.h` and onto `gchandleutilities.h`. It's implementation still lives in `marshalnative.cpp` [here](https://github.com/dotnet/coreclr/blob/master/src/vm/marshalnative.cpp#L595). Once https://github.com/dotnet/coreclr/pull/12314 lands, I think `gchandleutilities.cpp` would be a much better place for it.

### DIFF
--- a/src/debug/daccess/dacimpl.h
+++ b/src/debug/daccess/dacimpl.h
@@ -14,6 +14,11 @@
 #ifndef __DACIMPL_H__
 #define __DACIMPL_H__
 
+// This header include will need to be removed as part of GitHub#12170.
+// The only reason it's here now is that this header references the GC-private
+// structure "HandleTableMap".
+#include "../../gc/objecthandle.h"
+
 #if defined(_TARGET_ARM_) || defined(FEATURE_CORESYSTEM) // @ARMTODO: STL breaks the build with current VC headers
 //---------------------------------------------------------------------------------------
 // Setting DAC_HASHTABLE tells the DAC to use the hand rolled hashtable for
--- a/src/vm/gchandleutilities.h
+++ b/src/vm/gchandleutilities.h
@@ -33,6 +33,10 @@ private:
 void ValidateObjectAndAppDomain(OBJECTREF objRef, ADIndex appDomainIndex);
 void ValidateHandleAssignment(OBJECTHANDLE handle, OBJECTREF objRef);
 
+// Check that the supplied object is valid to put in a pinned handle,
+// throwing an exception if not.
+void ValidatePinnedObject(OBJECTREF obj);
+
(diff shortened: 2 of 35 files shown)


src/debug/daccess/dacimpl.h
```diff
@@ -14,6 +14,11 @@
 #ifndef __DACIMPL_H__
 #define __DACIMPL_H__
 
+// This header include will need to be removed as part of GitHub#12170.
+// The only reason it's here now is that this header references the GC-private
+// structure "HandleTableMap".
+#include "../../gc/objecthandle.h"
```
swgillespie: I will be submitting a follow-up PR for this in a day or two - it's a little too large and off-topic to include as a part of this PR.

swgillespie:
@Maoni0 @sergiy-k @adityamandaleeka @jkotas PTAL?

src/vm/gchandleutilities.h
```diff
@@ -33,6 +33,10 @@ class GCHandleUtilities
 void ValidateObjectAndAppDomain(OBJECTREF objRef, ADIndex appDomainIndex);
 void ValidateHandleAssignment(OBJECTHANDLE handle, OBJECTREF objRef);
 
+// Check that the supplied object is valid to put in a pinned handle,
+// throwing an exception if not.
+void ValidatePinnedObject(OBJECTREF obj);
```
jkotas: Shouldn't this rather be in marshalnative.h since the implementation is in marshalnative.cpp; and it implements arbitrary interop limitation - nothing fundamental for the GC handle table?

swgillespie: Ah, I did not realize it was enforcing an interop restriction - will do!

swgillespie:

swgillespie:
thanks!

PR #12395, opened 2017-06-20
Fix for a bug in fgMorphRecognizeBoxNullable.

### LINKED ISSUE #12392: Assert in JIT register allocator followed by an infinite loop
Compile Program.cs below:

.\Tools\dotnetcli\dotnet.exe Tools\csc.exe /o+ /out:Program.exe Program.cs /r:.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Private.CoreLib.dll /r:.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Console.dll /r:.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Runtime.dll

```
using System;
using System.Reflection;

namespace Test
{
    class Program
    {

        {
            test t = new test();
            if (!t.testmethod())
            {
                Console.WriteLine("SUCCESS");
            }
            else
            {
                Console.WriteLine("FAILURE");
            }
        }
    }

    class test
    {
        Setting<bool?> supportInteractive = Setting.ForBool(null);

        public bool testmethod()
        {
            return this.supportInteractive.HasValue;
        }
    }

    public class Setting<T>
    {
        Setting()
        {
        }

        public Setting(T value)
        {
            this.value = value;
        }

        public bool HasValue
        {
            get
            {
                if (this.value != null)
                {
                    Type t = this.value.GetType();
                    if (t.IsGenericType)
                    {
                        if (t.GetGenericTypeDefinition() == typeof(Nullable<>))
                        {
                            PropertyInfo hasValueProperty = t.GetProperty("HasValue");
                            bool result = (bool)hasValueProperty.GetValue(this.value, null);
                            return result;
                        }
                    }
                }
                return null != this.value;
            }
        }

        T value;
        public T Value
        {
            get { return this.value; }
            set { this.value = value; }
        }
    }

    public class Setting
    {
        Setting()
        {
            ;
        }

        public static Setting<bool?> ForBool(string parameter)
        {
            if (null == parameter)
            {
                return new Setting<bool?>(null);
            }
            Setting<bool?> setting = new Setting<bool?>(bool.Parse(parameter));
            return setting;
        }
    }
}

```
Then run the resulting program:
.\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\CoreRun.exe Program.exe

```
Assert failure(PID 19884 [0x00004dac], Thread: 21184 [0x52c0]): Assertion failed '((consume == 0) && (produce == 0)) || (ComputeAvailableSrcCount(tree) == consume)' in 'Test.Setting`1[Nullable`1][System.Nullable`1[System.Boolean]]:get_HasValue():bool:this' (IL size 111)

    File: c:\coreclr\src\jit\lsra.cpp Line: 3599
    Image: c:\coreclr\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\CoreRun.exe
```

erozenfeld:
fgMorphRecognizeBoxNullable doesn't work correctly when called during re-morphing. In particular it can't handle late args in the call to CORINFO_HELP_BOX_NULLABLE. The test case results in several asserts (the first one in lsra) followed by an infinite loop in the jit.

The reason the optimization is not performed during global morph is that only patterns with GT_EQ and GT_NE are handled. Some versions of csc generate

...
ldnull
cgt.un
ret

for

this C# line

null != this.value

so we get GT_GT instead.

The fix has two parts:

1. Don't attempt to perform the optimization when called during re-morph and the struct parameter is a late arg.
2. Call fgMorphRecognizeBoxNullable for GT_GT nodes with GTF_UNSIGNED set. This allows the optimization to fire during global morph.

Fixes #12392

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -10585,32 +10585,49 @@ GenTree* Compiler::fgMorphForRegisterFP(GenTree* tree)
     return tree;
 }
 
+//--------------------------------------------------------------------------------------------------------------
+// fgMorphRecognizeBoxNullable:
+//   Recognize this pattern:
+//
+//   stmtExpr  void  (IL 0x000...  ???)
+//     return    int
+//             CNS_INT     ref    null
+//         EQ/NE/GT        int
+//             CALL help ref    HELPER.CORINFO_HELP_BOX_NULLABLE
+//                 CNS_INT(h)  long   0x7fed96836c8 class
+//                 ADDR      byref
+//                     FIELD struct value
+//                         LCL_VAR ref V00 this
+//
+//   which comes from this code:
+//
+//      return this.value==null;
+//
+//   and transform it into
+//
+//   stmtExpr  void  (IL 0x000...  ???)
+//     return    int
+//             CNS_INT     ref    null
+//         EQ/NE/GT        int
+//             IND bool
+//                 ADDR      byref
+//                     FIELD struct value
+//                         LCL_VAR ref V00 this
+//
+// Arguments:
+//       compare - Compare tree to optimize.
+//
+// return value:
+//       A tree that has a call to CORINFO_HELP_BOX_NULLABLE optimized away if the pattern is found;
+//       the original tree otherwise.
+//
+
 GenTree* Compiler::fgMorphRecognizeBoxNullable(GenTree* compare)
 {
     GenTree*     op1 = compare->gtOp.gtOp1;
     GenTree*     op2 = compare->gtOp.gtOp2;
     GenTree*     opCns;
     GenTreeCall* opCall;
-
-    // recognize this pattern:
-    //
-    // stmtExpr  void  (IL 0x000...  ???)
-    //     return    int
-    //             const     ref    null
-    //         ==        int
-    //             call help ref    HELPER.CORINFO_HELP_BOX_NULLABLE
-    //                 const(h)  long   0x7fed96836c8 class
-    //                 addr      byref
-    //                     ld.lclVar struct V00 arg0
-    //
-    //
-    // which comes from this code (reported by customer as being slow) :
-    //
-    // private static bool IsNull<T>(T arg)
-    // {
-    //    return arg==null;
-    // }
-    //
 
     if (op1->IsCnsIntOrI() && op2->IsHelperCall())
     {
@@ -10637,8 +10654,17 @@ GenTree* Compiler::fgMorphRecognizeBoxNullable(GenTree* compare)
         return compare;
     }
 
-    // replace the box with an access of the nullable 'hasValue' field which is at the zero offset
-    GenTree* newOp = gtNewOperNode(GT_IND, TYP_BOOL, opCall->gtCall.gtCallArgs->gtOp.gtOp2->gtOp.gtOp1);
+    // Get the nullable struct argument
+    GenTree* arg = opCall->gtCall.gtCallArgs->gtOp.gtOp2->gtOp.gtOp1;
+
+    // Check for cases that are unsafe to optimize and return the unchanged tree
+    if (arg->IsArgPlaceHolderNode() || arg->IsNothingNode() || ((arg->gtFlags & GTF_LATE_ARG) != 0))
+    {
+        return compare;
+    }
+
+    // Replace the box with an access of the nullable 'hasValue' field which is at the zero offset
+    GenTree* newOp = gtNewOperNode(GT_IND, TYP_BOOL, arg);
 
     if (opCall == op1)
     {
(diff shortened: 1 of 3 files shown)


erozenfeld:
@briansull @CarolEidt @dotnet/jit-contrib PTAL

src/jit/morph.cpp
```diff
@@ -10638,7 +10638,12 @@ GenTree* Compiler::fgMorphRecognizeBoxNullable(GenTree* compare)
     }
 
     // replace the box with an access of the nullable 'hasValue' field which is at the zero offset
```
briansull: I would move this comment down to right before line 10646

src/jit/morph.cpp
```diff
@@ -10638,7 +10638,12 @@ GenTree* Compiler::fgMorphRecognizeBoxNullable(GenTree* compare)
...
-    GenTree* newOp = gtNewOperNode(GT_IND, TYP_BOOL, opCall->gtCall.gtCallArgs->gtOp.gtOp2->gtOp.gtOp1);
+    GenTreePtr arg = opCall->gtCall.gtCallArgs->gtOp.gtOp2->gtOp.gtOp1;
```
briansull: I would add a comment here stating the we check for the following three cases and early out, because it is unsafe to optimize.

src/jit/morph.cpp
```diff
@@ -11496,7 +11501,16 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
                         }
                     }
                 }
-                fgMorphRecognizeBoxNullable(tree);
+
+                __fallthrough;
+
+            case GT_GT:
+
+                if ((oper != GT_GT) || ((tree->gtFlags & GTF_UNSIGNED) != 0))
```
briansull: I'm not sure I understand this check. 
It appears that we have to exclude an unsigned GT_GT but everything else is OK.

briansull: Maybe get rid of the fall through and duplicate the call might make this safer and clearer,

erozenfeld: We allow unsigned GT_GT as well as GT_EQ and GT_NE. Other comparison operators are not allowed on reference types according to ECMA-335.

src/jit/morph.cpp
erozenfeld: Done.

src/jit/morph.cpp
erozenfeld: Done.

src/jit/morph.cpp
```diff
@@ -10585,33 +10585,50 @@ GenTree* Compiler::fgMorphForRegisterFP(GenTree* tree)
...
+//                         LCL_VAR ref V00 this
+//
+// Arguments:
+//       compare - Compare tree to optimize.
+//
+// return value:
+//       A tree that has a call to CORINFO_HELP_BOX_NULLABLE optimized away if the pattern is found;
+//       the original tree otherwise.
+//
+
```
CarolEidt: Awesome! Thanks!

src/jit/morph.cpp
erozenfeld: I prefer to keep the __fallthrough. Otherwise I'll have to duplicate the call as well as the two lines after it.

erozenfeld: I added a comment to make the check easier to understand.

src/jit/morph.cpp
```diff
@@ -11496,7 +11522,17 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
...
                     }
                 }
-                fgMorphRecognizeBoxNullable(tree);
+
+                __fallthrough;
+
+            case GT_GT:
+
+                // Try to optimize away calls to CORINFO_HELP_BOX_NULLABLE for GT_EQ, GT_NE, and unsigned GT_GT.
+                if ((oper != GT_GT) || ((tree->gtFlags & GTF_UNSIGNED) != 0))
```
mikedn: `((tree->gtFlags & GTF_UNSIGNED) != 0)` -> `tree->IsUnsigned()`

mikedn:
Note that we have code that transforms `GT_GT.un(x, null)` into `GT_NE(x, null)`, see #9454. Presumably that code runs too late to help with this.

erozenfeld:
Yes, fgMorphRecognizeBoxNullable runs when pre-processing a node and the code that transforms GT_GT.un(x, null) into GT_NE(x, null) runs when post-processing a node.

src/jit/morph.cpp
```diff
@@ -10637,8 +10654,17 @@ GenTree* Compiler::fgMorphRecognizeBoxNullable(GenTree* compare)
         return compare;
     }
 
-    // replace the box with an access of the nullable 'hasValue' field which is at the zero offset
-    GenTree* newOp = gtNewOperNode(GT_IND, TYP_BOOL, opCall->gtCall.gtCallArgs->gtOp.gtOp2->gtOp.gtOp1);
+    // Get the nullable struct argument
+    GenTreePtr arg = opCall->gtCall.gtCallArgs->gtOp.gtOp2->gtOp.gtOp1;
```
mikedn: I'd use `GenTree*`, if only because couple of lines below we also have `GenTree*`. In general, this `GenTreePtr` thing should IMO be avoided. It's pointless and inconsistent.

src/jit/morph.cpp
erozenfeld: Done.

src/jit/morph.cpp
erozenfeld: I agree, consistency is good. Changed to GenTree*.

CarolEidt: > In general, this GenTreePtr thing should IMO be avoided. It's pointless and inconsistent.

I totally agree!

erozenfeld:
10 unique methods with diffs in desktop SPMI. All improvements eliminating calls to CORINFO_HELP_BOX_NULLABLE. Total of 155 bytes.

PR #12402, opened 2017-06-21
Spmi second cleaning

sandreenko:
Use only fields that actually matter as keys for maps.


sandreenko:
@dotnet-bot test 
Tizen armel Cross Release Build

sandreenko:
I thought about it,  I think that special types are better, because it more obvious rather than the previous zero-init code. 
However, I would prefer to write information about actual used fields somewhere in jitEEinterface.

For example:
 JitEEInterface call has argument A with 10 fields, but use only 8. The another 2 fields doesn't affect the result of this call,
As I understand there are two cases here:
1. but contain reasonable information and there is no harm to save this fields as keys.
2. but they are not initialized and contain garbage (for example CallInfo.verSig if flags & CORINFO_CALLINFO_VERIFICATION == 0), in this case we can't save this fields as keys.

Does actually the second variant exist? Don't we have a contract that, for example, all unused fields in JitEE transactions are zero-init?
If it doesn't we can always save all fields as keys and don't create special types.

@jkotas do you know about the initialization of unused fields in JitEEInterface calls?

BruceForstall:
I don't know much about these complex JIT-EE interface calls, but it certainly would be useful if they were better documented!

sandreenko:
Size changes:
Base mch for clr spmi test 32.4Mb  -> 30.8Mb. ~ -5%.

jkotas:
> Does actually the second variant exist?

It exists in full framework.

> Don't we have a contract that, for example, all unused fields in JitEE transactions are zero-init?

There is not a contract like that. 

IIRC, getCallInfo used to zero out everything originally, but then it was taken out as throughout optimization.

> it certainly would be useful if they were better documented!

Agree.

sandreenko:
Thanks.

PR #12404, opened 2017-06-21
Align Linux TP benchmarks with Windows TP benchmarks

adiaaida:
On Windows, we download a set of assemblies that we have uploaded to
MyGet. On Linux, we were building corefx each run to get the benchmarks.
This change updates Linux to pull down the same benchmarks Windows uses,
using curl to grab the nuget package from MyGet.

Note: after this commit, we will have data for fewer assemblies on
Linux, but we will now have the same set for both Windows and Linux.
Some data in Benchview may be affected.

### DIFF
--- a/tests/scripts/perf-prep.sh
+++ b/tests/scripts/perf-prep.sh
@@ -64,19 +64,17 @@ python3.5 --version
 python3.5 ./tests/scripts/Microsoft.BenchView.JSONFormat/tools/machinedata.py
 
 if [ $throughput -eq 1 ]; then
-    # Clone corefx
-    if [ -d "_" ]; then
-        rm -r -f _
+    # Download throughput benchmarks
+    if [ -d "Microsoft.Benchview.ThroughputBenchmarks.x64.Windows_NT" ]; then
+        rm -r -f Microsoft.Benchview.ThroughputBenchmarks.x64.Windows_NT
     fi
-    mkdir _
-    git clone https://github.com/dotnet/corefx.git _/fx
-    cd _/fx
+    mkdir Microsoft.Benchview.ThroughputBenchmarks.x64.Windows_NT
+    cd Microsoft.Benchview.ThroughputBenchmarks.x64.Windows_NT
 
-    # Checkout the specific commit we want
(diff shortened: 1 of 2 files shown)


adiaaida:
@dotnet-bot test ci please
@dotnet-bot test linux throughput flow please

adiaaida:
/cc: @dotnet/rap-team 

I expect the Linux flow job to fail, as I am modifying both scripting and groovy, and the assemblies will be found in a different location. Testing is mostly for making sure the curl, unzip, etc. aspects of the scripting work as expected.

tests/scripts/perf-prep.sh
```diff
@@ -64,19 +64,17 @@ python3.5 --version
 python3.5 ./tests/scripts/Microsoft.BenchView.JSONFormat/tools/machinedata.py
 
 if [ $throughput -eq 1 ]; then
-    # Clone corefx
-    if [ -d "_" ]; then
-        rm -r -f _
+    # Download throughput benchmarks
+    if [ -d "Microsoft.Benchview.ThroughputBenchmarks.x64.Windows_NT" ]; then
+        rm -r -f Microsoft.Benchview.ThroughputBenchmarks.x64.Windows_NT
```
dpodder: Looks like we hardcode `x64` here (and further below), but still parametrize `${architecture}` in perf.groovy. Just want to make sure that is intentional?

adiaaida: Right now we don't have architecture parameterized in this script. Additionally, I *think* that we could remove the architecture all together, and only have one package for all of the TP assemblies, but I haven't tested that.

All that to say: yes, this was intentional, but if you think that I should parameterize it, I will.

dpodder: Gotcha. Yes, I think it would be better to either parametrize it out, *or* get rid of it altogether (if possible we should ensure we use the same input set for all platforms, so that we can compare x64-vs-x86 just like windows-vs-linux). But, no need to hold up this PR for that change, as long as the code behaves as intended (it makes mostly sense, since we only have x64 on Linux).

PR #12406, opened 2017-06-21
[release/2.0.0] Update trusted RHEL build Dockertag to support LTO (Port #12396)

dpodder:
The new RHEL image was generated using the same Dockerfile as the previous, but has LLVM recompiled with gold support enabled, which is required to do a build with -flto (a prerequisite for PGO).

Cherry-pick bed557d59ab60ede4204d6186ca7daf57eb7869c from master


dpodder:
Opened #12407 to track the status of this merge.

PR #12422, opened 2017-06-22
Make lvaSetClass dump diffable

CarolEidt:


CarolEidt:
@dotnet/jit-contrib PTAL - tiny dump change

Issue #11349, opened 2017-05-02
[RyuJIT][LSRA] Take register encoding sizes into account


CarolEidt:
Consider preferring smaller encoding sizes for high ref count, and larger encoding sizes for small ref count.
(This issue was migrated from the desktop bug tracking system.)

mikedn:
A related example I happened to stumble upon:
```C#
class Num { public int x, y, z, w; }
static Dictionary<int, Num> Test(int x, int y)
{
    return new Dictionary<int, Num> { { 42, new Num { x = x, y = y, z = 4, w = 2 } } };
}
```
generates:
```asm
E8C78B815F           call     CORINFO_HELP_NEWSFAST
4C8BC0               mov      r8, rax
41897008             mov      dword ptr [r8+8], esi
4189780C             mov      dword ptr [r8+12], edi
41C7401004000000     mov      dword ptr [r8+16], 4
41C7401402000000     mov      dword ptr [r8+20], 2
488BCB               mov      rcx, rbx
BA2A000000           mov      edx, 42
41B902000000         mov      r9d, 2
E84969A35E           call     System.Collections.Generic.Dictionary`2[Int32,__Canon][System.Int32,System.__Canon]:TryInsert(int,ref,ubyte):bool:this
```
Here LSRA was too eager to move to `r8` (presumably selected due to ABI requirements). If the move to `r8` was delayed until the actual call the code would have been 4 bytes smaller.

PR #12423, opened 2017-06-22
Mark lvDoNotEnregister lclVars as contained

CarolEidt:
Even if a lclVar is tracked, it may not be a register candidate. When determining whether a lclVar should be contained or RegOptional, take that into account.
In the interest of making this as accurate as possible, mark lclVars early as lvDoNotEnregister when they meet criteria that will later disqualify them from a register.

### DIFF
--- a/src/jit/lower.cpp
+++ b/src/jit/lower.cpp
@@ -108,20 +108,39 @@ bool Lowering::IsSafeToContainMem(GenTree* parentNode, GenTree* childNode)
 // IsContainableMemoryOp: Checks whether this is a memory op that can be contained.
 //
 // Arguments:
-//    node - the node of interest.
-//
-// Notes:
-//    This differs from the isMemoryOp() method on GenTree because it checks for
-//    the case of an untracked local. Note that this won't include locals that
-//    for some reason do not become register candidates, nor those that get
-//    spilled.
+//    node        - the node of interest.
+//    useTracked  - true if this is being called after liveness so lvTracked is correct
 //
 // Return value:
 //    True if this will definitely be a memory reference that could be contained.
 //
-bool Lowering::IsContainableMemoryOp(GenTree* node)
+// Notes:
+//    This differs from the isMemoryOp() method on GenTree because it checks for
+//    the case of doNotEnregister local. This won't include locals that
+//    for some other reason do not become register candidates, nor those that get
+//    spilled.
+//    Also, if we call this before we redo liveness analysis, any new lclVars
+//    introduced after the last dataflow analysis will not yet be marked lvTracked,
+//    so we don't use that.
+//
+bool Lowering::IsContainableMemoryOp(GenTree* node, bool useTracked)
 {
-    return node->isMemoryOp() || (node->IsLocal() && !comp->lvaTable[node->AsLclVar()->gtLclNum].lvTracked);
+#ifdef _TARGET_XARCH_
+    if (node->isMemoryOp())
+    {
+        return true;
+    }
+    if (node->IsLocal())
+    {
+        if (!m_lsra->enregisterLocalVars)
+        {
+            return true;
+        }
+        LclVarDsc* varDsc = &comp->lvaTable[node->AsLclVar()->gtLclNum];
+        return (varDsc->lvDoNotEnregister || (useTracked && !varDsc->lvTracked));
+    }
+#endif // _TARGET_XARCH_
+    return false;
 }
 
 //------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


CarolEidt:

CarolEidt:
@dotnet-bot help

CarolEidt:
@dotnet-bot test Windows_NT x64 perf
@dotnet-bot test Windows_NT x86 perf
@dotnet-bot test Windows_NT x64 throughput
@dotnet-bot test Windows_NT x86 throughput

CarolEidt:
@dotnet/jit-contrib PTAL
The original motivation for this change was that, in `Lowering`, where I am moving the containment analysis, the `lvTracked` field is not accurate because liveness has not yet been done for any new lclVars introduced by the optimizer, rationalizer or `Lowering`. However, this is also more accurate because a lclVar can be tracked but still not enregisterable. So this is a net improvement in code generation. However, in order to rely on this, I needed to mark a few addditional cases as `lvDoNotEnregister` which were previously not marked as such, but simply not marked `lvLRACandidate` in `LinearScan`. 

Like the recent change to make untracked lclVars contained, this has many diffs. They are nearly all positive, with a few size regressions. For jit-diff of frameworks+tests:

- x86 has 10477911 fewer bytes (.71%), 26438 improved, 49 regressed
- x64 has 1081122 fewer bytes (.69%), 27291 improved, 14 regressed

I'm still investigating the regressions. So far they fall into a couple of categories:

- Making things contained results in slightly different register allocation, which is usually better (also fewer `RefPositions`) but occasionally causes regressions.
- In one case the fact that fields of dependently promoted structs are marked early causes a cse to be created that wasn't previously, due to different costing of nodes, and this winds up changing register allocation in an unfortunate way.

src/jit/lower.cpp
```diff
@@ -108,20 +108,39 @@ bool Lowering::IsSafeToContainMem(GenTree* parentNode, GenTree* childNode)
...
 //
-// Return value:
-//    True if this will definitely be a memory reference that could be contained.
-//
-bool Lowering::IsContainableMemoryOp(GenTree* node)
+bool Lowering::IsContainableMemoryOp(GenTree* node, bool useTracked)
 {
-    return node->isMemoryOp() || (node->IsLocal() && !comp->lvaTable[node->AsLclVar()->gtLclNum].lvTracked);
+#ifdef _TARGET_XARCH_
+    if (node->isMemoryOp())
```
BruceForstall: Can you explain why there is `#ifdef _TARGET_XARCH_` in the new code, but not in the old?

CarolEidt: There is code in `LowerCompare()` that uses this for compare narrowing, and it's really only useful to do this for x86/x64. The other code that uses this is in arch-specific `TreeNodeInfoInit` methods.

src/jit/lower.cpp
```diff
@@ -108,20 +108,39 @@ bool Lowering::IsSafeToContainMem(GenTree* parentNode, GenTree* childNode)
...
 //    spilled.
+//    Also, if we call this before we redo liveness analysis, any new lclVars
+//    introduced after the last dataflow analysis will not yet be marked lvTracked,
+//    so we don't use that.
 //
-// Return value:
-//    True if this will definitely be a memory reference that could be contained.
-//
-bool Lowering::IsContainableMemoryOp(GenTree* node)
+bool Lowering::IsContainableMemoryOp(GenTree* node, bool useTracked)
```
pgavlin: Minor nit: do you think that callers will always pass a constant argument for the `useTracked` parameter (I think that's the case in this change)? If so, you might consider making it a template parameter.

CarolEidt: I find it easier to read without the template syntax, and I'm reasonably confident that the compiler will "do the right thing". That said, I should probably have mentioned that this is likely to be temporary, since eventually all the containment analysis will be moving to `Lowering` where we can't rely on `lvTracked`. The mitigation is to make things contained in `TreeNodeInfoInit` if they are lclVars marked `RegOptional`. The only fly in the ointment is that the semantics are not identical (as I mentioned before), since for the latter we don't always check `IsSafeToContainMem` even though I think we should. However that's a topic/change for another PR.

CarolEidt:
Having analyzed all of the regressions, they fall into the following buckets:
- General register allocation diffs, including purely encoding size diffs (#11349) as well as diffs involving spill (nothing fundamental; these diffs go both directions)
- Register allocation interferes with constant reuse.
- Add of a containable constant with a non-last-use lclVar. This results in an extra copy, since the lclVar reg can't be overwritten. Opened issue #12462 for this.
- Previously, some compares were getting narrowed by `LowerCompare()` because `lvTracked` was being used by `IsContainableMemoryOp()`, and many compiler temps were not yet marked `lvTracked` because liveness had not yet been re-run. Using `~lvDoNotEnregister` instead results in the narrowing no longer being done. However, even when it is a register, a smaller encoding is available for eax (a special case of #11349), as well as the more general encoding size issue.

Issue #12341, opened 2017-06-16
createdump utility: missing program header, sections and other info


mikem8361:
The symbol acquisition tool will need all the loaded native modules' ELF program headers and sections to compute the proper symbol store key from the build ID. 

The managed assemblies loaded in the process needed to be added to the dump (the PE header and debug directory) so the symbol acquisition tool can download the symbols for those modules too.

The .eh_frame section isn't part of the dump created either and it is needed by the out of context unwinding support.

PR #12425, opened 2017-06-23
Add ELF program and section headers to core dump.

mikem8361:
Issue #12341

This allows the build id for the modules be extracted properly.

Add .eh_frame, .eh_frame_hdr and some other important sections to core dump.

Add managed module names to core dump so the symbol acquisition
tool can get symbol store keys for them.

Add the non memory/file backed memory regions to core dump so debuggers
like lldb can check if an address is "code" or not. Unwinding doesn't
work when there isn't a module/.so file present.

### DIFF
--- a/src/debug/createdump/crashinfo.cpp
+++ b/src/debug/createdump/crashinfo.cpp
@@ -501,7 +501,7 @@ CrashInfo::GetDSOInfo()
 }
 
 inline bool
-NameCompare(char* name, const char* sectionName)
+NameCompare(const char* name, const char* sectionName)
 {
     return strncmp(name, sectionName, strlen(sectionName) + 1) == 0;
 }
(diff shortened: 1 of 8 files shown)


src/debug/createdump/memoryregion.h
```diff
@@ -36,23 +47,57 @@ struct MemoryRegion
...
+
+    const uint32_t Permissions() const { return m_flags & MEMORY_REGION_FLAG_PERMISSIONS_MASK; }
+    const uint32_t Flags() const { return m_flags; }
+    const bool IsBackedByMemory() const { return (m_flags & MEMORY_REGION_FLAG_MEMORY_BACKED) != 0; }
     const uint64_t StartAddress() const { return m_startAddress; }
     const uint64_t EndAddress() const { return m_endAddress; }
     const uint64_t Size() const { return m_endAddress - m_startAddress; }
     const uint64_t Offset() const { return m_offset; }
-    const char* FileName() const { return m_fileName; }
+    char* FileName() const { return m_fileName; }
```
janvorli: I wonder why was this change needed. There doesn't seem to be any newly added call to this method in the current PR and I assume the original version with the `const char*` return was ok before.

src/debug/createdump/crashinfo.cpp
```diff
@@ -454,18 +488,134 @@ CrashInfo::GetDSOInfo()
...
+            return false;
+        }
         linkMapAddr = map.l_next;
     }
 
     return true;
 }
 
+inline bool
+NameCompare(char* name, const char* sectionName)
```
janvorli: A nit - the first parameter could be `const`

mikem8361:
I undid the change and made all references to the filename “const char *” (constructors, other functions, etc.).

Issue #12383, opened 2017-06-20
[Question][JIT] Why "(cond)?0:1" can convert to "cond"? Confuse about then node and else node in qmark.


303248153:
I'm really confuse with this code:
https://github.com/dotnet/coreclr/blob/v2.0.0-preview1/src/jit/morph.cpp#L12539

``` c++
                // If we have (cond)?0:1, then we just return "cond" for TYP_INTs
                //
                // Don't do this optimization for ARM: we always require assignment
                // to boolean to remain ?:, since we don't have any way to generate
                // this with straight-line code, like x86 does using setcc (at least
                // after the IT instruction is deprecated).

                if (genActualType(op1->gtOp.gtOp1->gtType) == TYP_INT && genActualType(typ) == TYP_INT &&
                    thenNode->gtOper == GT_CNS_INT && elseNode->gtOper == GT_CNS_INT)
                {
                    ival1 = thenNode->gtIntCon.gtIconVal;
                    ival2 = elseNode->gtIntCon.gtIconVal;

                    // Is one constant 0 and the other 1?
                    if ((ival1 | ival2) == 1 && (ival1 & ival2) == 0)
                    {
                        // If the constants are {1, 0}, reverse the condition
                        if (ival1 == 1)
                        {
                            gtReverseCond(op1);
                        }

                        // Unmark GTF_RELOP_JMP_USED on the condition node so it knows that it
                        // needs to materialize the result as a 0 or 1.
                        noway_assert(op1->gtFlags & (GTF_RELOP_QMARK | GTF_RELOP_JMP_USED));
                        op1->gtFlags &= ~(GTF_RELOP_QMARK | GTF_RELOP_JMP_USED);

                        DEBUG_DESTROY_NODE(tree);
                        DEBUG_DESTROY_NODE(op2);

                        return op1;
                    }
```

This code will convert "cond ? 0 : 1" to "cond", and "cond ? 1 : 0" to "!cond", which I think it should be opposite, why?

mikedn:
Hmm, this code does appear to be broken but I don't think it ever runs in RyuJIT. Qmarks are only generated when importing certain code sequences and they won't match this pattern. This code should probably be ifdefed out, only when LEGACY_BACKEND is defined you're likely to encounter this pattern.

303248153:
Is it really appear to be broken?
Because I just saw the comment about ThenNode and ElseNode, I'm not so sure what's this code actually mean.

mikedn:
[...]

PR #12429, opened 2017-06-23
Remove QMARK related dead code

mikedn:
Qmarks are created during import and then converted to normal flow control after global morph. 
Morphing attempts are pointless because the patterns generated by the importer aren't recognized anyway. One morph attempt appears to be broken too, see #12383.

The importer + `fgInstrumentMethod` + `fgInstrumentMethod` generate the following patterns:
* `cond ? void : asg`
* `cond ? void : call`
* `cond ? 0 : call`
* `cond ? lcl : (cond ? 0 : call)`

GenTreeQmark's liveness sets are obviously useless, the relevant code in liveness.cpp is already ifdefed out.

### DIFF
--- a/src/jit/morph.cpp
+++ b/src/jit/morph.cpp
@@ -12549,6 +12549,7 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
 
                 noway_assert(tree->OperKind() & GTK_RELOP);
 
+#ifdef LEGACY_BACKEND
                 /* Check if the result of the comparison is used for a jump.
                  * If not then only the int (i.e. 32 bit) case is handled in
                  * the code generator through the (x86) "set" instructions.
@@ -12576,8 +12577,10 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
 
                     return tree;
                 }
+#endif // LEGACY_BACKEND
                 break;
 
+#ifdef LEGACY_BACKEND
             case GT_QMARK:
 
                 /* If op1 is a comma throw node then we won't be keeping op2 */
@@ -12698,6 +12701,7 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
 #endif // !_TARGET_ARM_
 
                 break; // end case GT_QMARK
+#endif                 // LEGACY_BACKEND
 
             case GT_MUL:
 
--- a/src/jit/compiler.cpp
+++ b/src/jit/compiler.cpp
@@ -1895,7 +1895,9 @@ void Compiler::compInit(ArenaAllocator* pAlloc, InlineInfo* inlineInfo)
 #endif // DEBUG
 #endif // MEASURE_MEM_ALLOC
 
+#ifdef LEGACY_BACKEND
         compQMarks = nullptr;
+#endif
     }
     else
     {
@@ -1911,7 +1913,9 @@ void Compiler::compInit(ArenaAllocator* pAlloc, InlineInfo* inlineInfo)
 #endif // DEBUG
 #endif // MEASURE_MEM_ALLOC
 
+#ifdef LEGACY_BACKEND
         compQMarks = new (this, CMK_Unknown) ExpandArrayStack<GenTreePtr>(getAllocator());
+#endif
     }
 
 #ifdef FEATURE_TRACELOGGING
--- a/src/jit/compiler.h
+++ b/src/jit/compiler.h
@@ -7754,11 +7754,13 @@ public:
     bool compQmarkRationalized;    // Is it allowed to use a GT_QMARK/GT_COLON node.
     bool compUnsafeCastUsed;       // Does the method use LDIND/STIND to cast between scalar/refernce types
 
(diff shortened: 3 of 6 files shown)


BruceForstall:
@dotnet/jit-contrib

pgavlin:
LGTM, though it might be better to update the title to "disable QMark-related code when using the RyuJIT backend".

src/jit/morph.cpp
```diff
@@ -12576,8 +12577,10 @@ GenTreePtr Compiler::fgMorphSmpOp(GenTreePtr tree, MorphAddrContext* mac)
 
                     return tree;
                 }
+#endif // LEGACY_BACKEND
                 break;
 
+#ifdef LEGACY_BACKEND
             case GT_QMARK:
```
BruceForstall: Is this really dead for RyuJIT? We still have qmarks from importer until the end of morph when we call `fgExpandQmarkNodes()`. Are there any diffs?

mikedn: Those are some morph attempts that don't seem to match any of the patterns generated by the importer. See the updated PR description.

No diffs.

mikedn:
I wonder if we shouldn't go further and just generate the necessary flow control in the importer. Generating qmarks and then converting them to flow control seems like an unnecessary waste of time for a JIT.

BruceForstall:
> I wonder if we shouldn't go further and just generate the necessary flow control in the importer. Generating qmarks and then converting them to flow control seems like an unnecessary waste of time for a JIT.

Seems reasonable, but any creator of qmarks (fgInstrumentMethod?) would also need to introduce flow. Maybe a few years back when we were trying to get rid of qmarks, the current state was considered a reasonable landing point. I think there was some concern about needing to rediscover qmark-like-flow to get good CQ on x86 (e.g., cmov?). If we can eliminate `LEGACY_BACKEND` it might make things easier.

AndyAyersMS:
@mikedn I've had similar thoughts but we'd need to make sure that we don't break tree-based pattern matches that happen downstream. Matching patterns across flow would be way more complex given that we don't have anything cheap connecting defs & uses cross statement.

I've also been thinking of going in the other direction -- that we should defer expansions for many of these lookup sequences and just express them as new high-level operators. This might allow us to cheaply run some of these optimizations directly in the importer (eg `typeof(T) == typeof(int)`) and then do on the fly constant prop to short-circuit importation of unreachable blocks. For heavily type-cased generic code (which is increasingly common) this would give a nice TP boost as much of the code in these method would never be examined by the jit at all.

mikedn:
> I think there was some concern about needing to rediscover qmark-like-flow to get good CQ on x86 (e.g., cmov?)

Yes but it looks like all patterns generated by the importer aren't actually suitable for CMOV (they involve calls or indirs). Though I have to double check this, I'm pretty sure I've seen JIT32 generating CMOV in one specific case.

> I've also been thinking of going in the other direction -- that we should defer expansions for many of these lookup sequences and just express them as new high-level operators. 

Right. I've experimented with if-conversion in lowering and I was wondering if it wouldn't be better to do it earlier thinking that I may enable other optimizations. And reduce the block count.

BruceForstall:

russellhadley:
Is this part of a drive to get conditional operators (setcc, cmov on x86/64, c{sel,set,inv,neg,inc} on arm64)?  Historically qmark caused problems for conditional side-effects, particularly in the RA (from what I'm told), but if we only formed cases where we could hoist the source expressions it seems like forming them early in the importer is the best (throughput) case since recognizing them in the optimizer is hard to do cheaply.

mikedn:
> Is this part of a drive to get conditional operators (setcc, cmov on x86/64, c{sel,set,inv,neg,inc} on arm64)? 

Nope, it's due to #12383. I started looking at this code and figured out that it's in fact useless, at least when LEGACY_BACKEND is not defined. Some of it (the broken code in particular) is likely useless when LEGACY_BACKEND is defined (otherwise it means that JIT32 probably has a bug).

> Historically qmark caused problems for conditional side-effects, particularly in the RA (from what I'm told)

Attempting to form arbitrary qmark trees and have to deal with liveness and side effects seems overkill to me. If side effects are present then it's very likely that it won't be possible to generate CMOV/CSEL so in this respect qmark are useless. The only reason to generate such qmarks would be to help with pattern matching as @AndyAyersMS already suggested.

> but if we only formed cases where we could hoist the source expressions it seems like forming them early in the importer is the best (throughput) case since recognizing them in the optimizer is hard to do cheaply.

I'm not sure why it would be less hard to recognize them in the importer. Are you thinking that we should only recognize certain IL patterns rather than doing actual if-conversion of BBs? Maybe, but we may find that we still do a non-0 amount of work and miss various cases.

russellhadley:
> I'm not sure why it would be less hard to recognize them in the importer. Are you thinking that we should only recognize certain IL patterns rather than doing actual if-conversion of BBs? Maybe, but we may find that we still do a non-0 amount of work and miss various cases.

It should be easier for simple cases to recognize in the importer and avoid creating basic blocks.   For sure this will miss various cases that we might want to recognize later in the optimizer (especially at a higher tier) but the benefit of recognizing the simple cases early makes this attractive.  @AndyAyersMS I'm thinking this fits with your road map.

PR #12432, opened 2017-06-23
Fix package download path.

crummel:
This was a BuildTools change that leaves the $BuildConfiguration/pkg part off of the package download path.

### DIFF
--- a/tests/dir.props
+++ b/tests/dir.props
@@ -66,7 +66,7 @@
 
    <ItemGroup> 
      <!-- Need to escape double forward slash (%2F) or MSBuild will normalize to one slash on Unix. -->
-     <DotnetSourceList Include="$(PackagesDir)AzureTransfer\$(__BuildType)\pkg\" Condition="'$(OverwriteCoreClrPackageVersion)' == 'true'" />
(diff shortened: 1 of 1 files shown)


crummel:
cc @danmosemsft

crummel:
Merging with the Mac failure as this only affects the Windows test build.

wtgodbe:
@crummel I may have been premature - did you test the workflow locally?

wtgodbe:
That line is only hit during the pipebuild - we run sync.cmd against a specific azure blob, then try to restore packages from what we downloaded during sync.cmd. Those packages wind up in $(PackagesDir)AzureTransfer\$(__BuildType)\pkg\ - why remove the '$(__BuildType)\pkg\' suffix?

crummel:
@wtgodbe Yes, I had this issue reproducing locally and this change fixed the issue.  This BuildTools commit changed the download behavior so the path is trimmed: https://github.com/dotnet/buildtools/commit/4f918b5f338f0cac04b29cd4bfe17acfd5d8e1b9.  I upgraded the BuildTools version to fix a different issue and unknowingly brought that change in.

PR #12435, opened 2017-06-23
Check for TestEnv before running test

jashook:
This handles the case that an incorrect path was passed for the TestEnv
location. Instead of running something possibly undesired, return an error
and warn that the TestEnv was incorrectly passed.


jashook:
Will test the change locally once the build finishes.

karelz:
What is status of this PR? 2 months of no update :( ...

karelz:

PR #12439, opened 2017-06-24
Update compLongUsed in rationalize.

pgavlin:
This allows us to skip decompisition if long-typed values are not used by a method.

### DIFF
--- a/src/jit/rationalize.cpp
+++ b/src/jit/rationalize.cpp
@@ -983,6 +983,11 @@ Compiler::fgWalkResult Rationalizer::RewriteNode(GenTree** useEdge, ArrayStack<G
             // Clear the GTF_CALL flag for all nodes but calls
             node->gtFlags &= ~GTF_CALL;
         }
+
+        if (node->TypeGet() == TYP_LONG)
+        {
+            comp->compLongUsed = true;
+        }
     }
 
     assert(isLateArg == ((use.Def()->gtFlags & GTF_LATE_ARG) != 0));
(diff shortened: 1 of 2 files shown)


pgavlin:

The combination of this PR, #12440, and #12441 show about a 5-6% throughput improvement on Bruce's x86 MinOpts SPMI dataset when the JIT is built without PGO.

pgavlin:
`jit-diff` reports zero diffs across the framework and test assemblies.

src/jit/rationalize.cpp
```diff
@@ -983,6 +983,11 @@ Compiler::fgWalkResult Rationalizer::RewriteNode(GenTree** useEdge, ArrayStack<G
             // Clear the GTF_CALL flag for all nodes but calls
             node->gtFlags &= ~GTF_CALL;
         }
+
+        if (node->TypeGet() == TYP_LONG)
+        {
+            comp->compLongUsed = true;
```
CarolEidt: Wow, so this already existed but was not kept up-to-date?

CarolEidt:
> The combination of this PR, #12440, and #12441 show about a 5-6% throughput improvement on Bruce's x86 MinOpts SPMI dataset when the JIT is built without PGO.

Nice!

src/jit/rationalize.cpp
pgavlin: Yup. It's set in the importer and lclVar initialization and used in `rpPredictRegUse`.

BruceForstall:
> The combination of this PR, #12440, and #12441 show about a 5-6% throughput improvement on Bruce's x86 MinOpts SPMI dataset when the JIT is built without PGO.

awesome!

PR #12440, opened 2017-06-24
Maintain `LIR::IsUnusedValue` in the backend.

pgavlin:
This flag indicates whether or not the SDSU temp produced by a node is
ever read. This information is needed by the register allocator, but is
also useful in other parts of the compiler (for example, `TryGetUse` can
return immediately if this flag is set). Setting this flag properly in
rationalize and maintaining it through decomposition and lowering allows
us to remove an IR walk immediately prior to LSRA.

### DIFF
--- a/src/jit/lir.cpp
+++ b/src/jit/lir.cpp
@@ -972,10 +972,18 @@ void LIR::Range::InsertAtEnd(Range&& range)
 // Arguments:
 //    node - The node to remove. Must be part of this range.
 //
-void LIR::Range::Remove(GenTree* node)
+void LIR::Range::Remove(GenTree* node, bool markOperandsUnused)
 {
     assert(node != nullptr);
     assert(Contains(node));
+
+    if (markOperandsUnused)
+    {
+        for (GenTree* operand : node->Operands())
+        {
+            operand->gtLIRFlags |= Flags::IsUnusedValue;
+        }
+    }
 
     GenTree* prev = node->gtPrev;
     GenTree* next = node->gtNext;
--- a/src/jit/lir.cpp
+++ b/src/jit/lir.cpp
@@ -1188,7 +1188,7 @@ bool LIR::Range::TryGetUse(GenTree* node, Use* use)
     // Don't bother looking for uses of nodes that are not values.
     // If the node is the last node, we won't find a use (and we would
     // end up creating an illegal range if we tried).
-    if (node->IsValue() && (node != LastNode()))
+    if (node->IsValue() && ((node->gtLIRFlags & Flags::IsUnusedValue) == 0) && (node != LastNode()))
     {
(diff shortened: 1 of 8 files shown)


pgavlin:

The combination of this PR, #12439, and #12441 show about a 5-6% throughput improvement on Bruce's x86 MinOpts SPMI dataset when the JIT is built without PGO.

pgavlin:
Ping.

src/jit/lir.cpp
```diff
@@ -972,11 +972,19 @@ void LIR::Range::InsertAtEnd(Range&& range)
 // Arguments:
 //    node - The node to remove. Must be part of this range.
 //
-void LIR::Range::Remove(GenTree* node)
+void LIR::Range::Remove(GenTree* node, bool markOperandsUnused)
```
BruceForstall: Please add `markOperandsUnused` to the header comment

src/jit/lir.cpp
```diff
@@ -972,11 +972,19 @@ void LIR::Range::InsertAtEnd(Range&& range)
...
     assert(node != nullptr);
     assert(Contains(node));
 
+    if (markOperandsUnused)
+    {
+        for (GenTree* operand : node->Operands())
+        {
+            operand->gtLIRFlags |= Flags::IsUnusedValue;
+        }
+    }
```
BruceForstall: Maybe you could add this as a small GenTree helper to avoid duplicating this both here and in liveness: `node->MarkOperandsUnused()`. Yeah, it's small; your choice.

CarolEidt: I think it would be useful to have methods to set and clear the unused value flag, e.g. `LIR::Flags:SetUnusedValue()` and `LIR::Flags:ClearUnusedValue()` or something like that. It makes the intent clearer than the and-ing and or-ing.

src/jit/lir.cpp
```diff
@@ -1180,7 +1188,7 @@ bool LIR::Range::TryGetUse(GenTree* node, Use* use)
     // Don't bother looking for uses of nodes that are not values.
     // If the node is the last node, we won't find a use (and we would
     // end up creating an illegal range if we tried).
-    if (node->IsValue() && (node != LastNode()))
+    if (node->IsValue() && ((node->gtLIRFlags & Flags::IsUnusedValue) == 0) && (node != LastNode()))
```
CarolEidt: And here you could use something like a `LIR::Flags:IsUnusedValue()`

src/jit/lir.cpp
pgavlin: This is a good idea. I expect that we'll need to do this more frequently if we become more aggressive about DCE in the backend.

pgavlin:
I will prepare a follow-up PR for the suggested changes to the `UnusedValue` APIs.

PR #12441, opened 2017-06-24
Avoid the use edge iterator in `TryGetUse`.

pgavlin:
`TryGetUse` does not require the ordering semantics provided by the use
edge iterator. Replace use edge iteration with manual inspection of each
node type.

### DIFF
--- a/src/jit/gentree.cpp
+++ b/src/jit/gentree.cpp
@@ -6246,15 +6246,310 @@ bool GenTree::TryGetUse(GenTree* def, GenTree*** use)
     assert(def != nullptr);
     assert(use != nullptr);
 
-    for (GenTree** useEdge : UseEdges())
+    switch (OperGet())
     {
-        if (*useEdge == def)
+        // Leaf nodes
+        case GT_LCL_VAR:
+        case GT_LCL_FLD:
+        case GT_LCL_VAR_ADDR:
+        case GT_LCL_FLD_ADDR:
+        case GT_CATCH_ARG:
+        case GT_LABEL:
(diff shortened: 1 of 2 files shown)


pgavlin:

The combination of this PR, #12439, and #12440 show about a 5-6% throughput improvement on Bruce's x86 MinOpts SPMI dataset when the JIT is built without PGO.

src/jit/gentree.cpp
```diff
@@ -6246,15 +6246,310 @@ bool GenTree::TryGetUse(GenTree* def, GenTree*** use)
     assert(def != nullptr);
     assert(use != nullptr);
 
-    for (GenTree** useEdge : UseEdges())
+    switch (OperGet())
+    {
+        // Leaf nodes
```
CarolEidt: Is the switch more efficient than checking for `GTK_LEAF`, `GTK_UNOP`, etc? That would certainly be cleaner, and then you could have cases for any nodes that require special handling.

pgavlin: Yes--checking the operator kind has proven to be a slower approach in the past (e.g. we initially used kinds in the operand iterator and needed to change that to use a similar switch for efficiency reasons).

pgavlin: Let's discuss the utility and efficiency of operator kinds in #12443 and leave this as-is for now.

pgavlin: Just to clarify, by "leave this as-is" I mean "leave this PR as-is and hopefully get it merged" :)

PR #12452, opened 2017-06-26
[Linux][GDB-JIT] Fix memory leak

kbaladurin:
Use NewArrayHolder for pVars in LocalsInfo class

### DIFF
--- a/src/vm/gdbjit.h
+++ b/src/vm/gdbjit.h
@@ -3,7 +3,7 @@
 // See the LICENSE file in the project root for more information.
 //*****************************************************************************
 // File: gdbjit.h
-// 
+//
 
 //
 // Header file for GDB JIT interface implemenation.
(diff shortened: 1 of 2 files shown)


kbaladurin:
@Dmitri-Botcharnikov @ayuckhulk

src/vm/gdbjit.h
```diff
@@ -140,7 +141,7 @@ class LocalsInfo
     NewArrayHolder< NewArrayHolder<char> > localsName;
     NewArrayHolder<Scope> localsScope;
     ULONG32 countVars;
-    ICorDebugInfo::NativeVarInfo *pVars;
+    NewArrayHolder<ICorDebugInfo::NativeVarInfo> pVars;
```
janvorli: A nit - now that this is no longer a pointer, could you please rename it to "vars" so that the name matches other array holders in this class? LGTM otherwise.

kbaladurin: Thanks for note. I renamed "pVars" to "vars".

PR #12457, opened 2017-06-26
Refactor the rationalizer's instrinsic rewriter.

pgavlin:
Change its two constituent static methods into instance methods.


pgavlin:

src/jit/rationalize.cpp
```diff
@@ -207,10 +195,10 @@ void Rationalizer::RewriteNodeAsCall(GenTree**             use,
     }
     else
     {
-        // Update the linear oder start of "root" if treeFirstNode
+        // Update the linear oder start of "m_statement" if treeFirstNode
```
BruceForstall: Might as well fix the typo: `oder`

Issue #12170, opened 2017-06-08
[Local GC] Handle table and the DAC


swgillespie:
This is an issue that @adityamandaleeka ran into a little while ago. The DAC has deep knowledge of the GC's handle table:
* It is aware of the global handle table map instance, `g_HandleTableMap`
* It calls `GetDependentHandleSecondary` on handles in the target process
* It calls `HndGetHandleExtraInfo` on handles in the target process

Today, this is not a problem; the GC DAC build compiles the above two functions into the DAC and the DAC is free to call them. For a standalone GC, this is more problematic; as it stands today the DAC is not permitted to dispatch virtually on the target's `IGCHeap` instance, which it would need to do for these two functions. Even if it could, it would not be calling the `DACCESS_COMPILE` versions of these functions, which could result in strange behavior.

This commit (https://github.com/swgillespie/coreclr/commit/1eef14c62f5fd74e8d51815e87af1aa0c2452062) is a start - it fixes the first bullet point and ifdef's out the bottom two bullet points. We'll need to come up with some sort of strategy `GetDependentHandleSecondary` and `HndGetHandleExtraInfo` - we can allow the debugger to dispatch virtually on the debugee's `IGCHeap` instance, duplicate the code, or do some other solution.

This issue is the last remaining use of the GC DAC build and, once it is complete, the GC DAC build can be removed completely which should be a nice DAC binary size win and compile time improvement.

swgillespie:
I had an idea about this today - as long as we link in a GC to the CoreCLR DLL, we do not *have* to eliminate our dependency on `GetDependentHandleSecondary` and `HndGetHandleExtraInfo`, as long as a standalone GC does not modify those functions. In theory, we can use the linked-in DAC-ized versions of these functions, even if the debugee is running with a standalone GC.

If we do this, we would include these two functions in the DAC API surface area of a standalone GC and would need to increment a major version number if we changed their behavior. That seems to me like it would be a rare occurrence, though.

jkotas:
> long as we link in a GC to the CoreCLR DLL

I think you want to make this explicit, like refactor the function into gcinterface.inl. I do think you want to link a whole GC with a fingers crossed that it will work out.

swgillespie:
Ok! (cc @sergiy-k, who I talked about this with)

To that end, here's the call graph of the two offending functions:

[...]

PR #12458, opened 2017-06-26
[Local GC] Refactor usage of GC-internal functions in the DAC

swgillespie:
Partially addresses/works around https://github.com/dotnet/coreclr/issues/12170. Not quite ready for review, testing the CI and doing debugger test + manual validation right now.

### DIFF
--- a/src/debug/daccess/daccess.cpp
+++ b/src/debug/daccess/daccess.cpp
@@ -29,6 +29,7 @@
 #endif
 
 #include "dwbucketmanager.hpp"
+#include "gcinterface.dac.h"
 
 // To include definiton of IsThrowableThreadAbortException
 // #include <exstatecommon.h>
@@ -7932,8 +7933,6 @@ STDAPI OutOfProcessExceptionEventDebuggerLaunchCallback(__in PDWORD pContext,
 
 // DacHandleEnum
 
-// TODO(Local GC) - The DAC should not include GC headers
-#include "../../gc/handletablepriv.h"
 #include "comcallablewrapper.h"
 
 DacHandleWalker::DacHandleWalker()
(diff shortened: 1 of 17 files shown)


swgillespie:

swgillespie:
reviving this now that some dust has settled - @Maoni0 @adityamandaleeka @jkotas @sergiy-k PTAL? (Also @noahfalk for the DAC changes?) This approach has worked for me locally - I'm in the process of running the debugger test suite with these changes (unfortunately I'm having some trouble getting it running but working through it). 

This PR explicitly tags the six functions whose symbol names and semantics are known to the DAC using the `GC_DAC_VISIBLE` and `GC_DAC_VISIBLE_NO_MANGLE` macros. These symbols are exposed to the dac using the `debug/daccess/gcinterface.dac.h` header. Ultimately, this means that the DAC needs to link against DAC-ized versions of these functions, which is not optimal but seems to be the simplest solution for now that allows standalone GC work to continue. Fortunately, as a part of this PR, the set of files in the `gc` directory that need to be built as part of the DAC build has been reduced to only `handletable.cpp`, `handletablescan.cpp`, `handletablecore.cpp`, and `objecthandle.cpp`. Notably, it's not necessary to build the GC itself.

The rest of this PR proceeds similarly to https://github.com/dotnet/coreclr/pull/9255, where DAC-only clones of structs are added to `gcinterface.dac.h` and used by the DAC to interface with the "real" structs in the debuggee. I have tried to move all fields accessed by the DAC to the start of these structs, except for one where a comment specifically indicated that this should not be done for performance reasons.

swgillespie:
@dotnet-bot test this please

swgillespie:
My remaining debugger test failure is failing on `master` so this PR is now free of test failures.

swgillespie:
thanks!

PR #12460, opened 2017-06-26
Rephrase rationalize as a single IR walk.

pgavlin:
This walk performs both pre- and post-order visits. The pre-order visit
is responsible for rewriting intrinsics as user calls where necessary;
the post-order visit is resonsible for converting HIR nodes into LIR
nodes.

I've presented these changes as two commits for clarity: the first commit updates the intrinsic rewriter to use the LIR utilities and the second merges the two IR walks.

The reason we need the pre-pass is that `fgMorphArgs` expects to be able to create and manipulate HIR (including assignment nodes). If we wanted to avoid the pre-/post-pass split, we could instead dump all arguments to intrinsics into lclVars, then pass those lclVars as arguments to the call. This might speed things up further but risks impacting RA.

This change causes no diffs on S.P.CoreLib. I am currently attempting a SuperPMI run.

### DIFF
--- a/src/jit/lir.h
+++ b/src/jit/lir.h
@@ -8,6 +8,7 @@
 class Compiler;
 struct GenTree;
 struct BasicBlock;
+class Rationalizer;
 
 class LIR final
 {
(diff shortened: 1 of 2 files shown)


pgavlin:
Per callgrind, this reduces the number of instructions retired on x64 Linux when running `crossgen S.P.CoreLib` by 0.7%.

pgavlin:

pgavlin:
SuperPMI reports no diffs.

PR #12473, opened 2017-06-27
[RyuJIT/ARM32][WIP] Update LSRA with helpers

hqueue:
Update LSRA with helper functions
- Introduce `updatePreviousInterval()`
- Update LSRA with helpers (`updateAssignedInterval()` and `updatePreviousInterval()`)


hqueue:
@CarolEidt  PTAL
This is a PR to clean up LSRA code for ARM32 with helper functions.
There are some remaining cases where helpers can not be applied.
I think there are more better ways to improve code and I will update this PR with reviews for a while.

PR #12474, opened 2017-06-27
Remove unnecessary update of RegRecord in LSRA

hqueue:
We don't need to clear assignedInterval of regRec here,
because it was already cleared by checkAndClearInterval() above at https://github.com/dotnet/coreclr/pull/12474/files#diff-93fb9832a03a9783e2a415b581fcb45cR6453.

### DIFF
--- a/src/jit/lsra.cpp
+++ b/src/jit/lsra.cpp
@@ -6480,6 +6480,9 @@ void LinearScan::unassignPhysReg(RegRecord* regRec, RefPosition* spillRefPositio
     {
         // This must have been a temporary copy reg, but we can't assert that because there
         // may have been intervening RefPositions that were not copyRegs.
+
+        // reg->assignedInterval has already been set to nullptr by checkAndClearInterval()
+        assert(regRec->assignedInterval == nullptr);
         return;
     }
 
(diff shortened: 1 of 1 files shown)


hqueue:
I'm very cautious about this change, because it affects every targets. Please let me know if I made a mistake.

BruceForstall:
@CarolEidt @pgavlin PTAL
cc @dotnet/jit-contrib

src/jit/lsra.cpp
```diff
@@ -6480,7 +6480,6 @@ void LinearScan::unassignPhysReg(RegRecord* regRec, RefPosition* spillRefPositio
     {
         // This must have been a temporary copy reg, but we can't assert that because there
         // may have been intervening RefPositions that were not copyRegs.
-        regRec->assignedInterval = nullptr;
```
pgavlin: Rather than removing this entirely we should replace it with `assert(regRec->assignedInterval == nullptr);`.

CarolEidt: Also, a comment would be useful here explaining why we know it has already been set to nullptr.

---
In reply to: [124342695](https://github.com/dotnet/coreclr/pull/12474#discussion_r124342695) [](ancestors = 124342695)

CarolEidt:
:clock1:

hqueue:
@pgavlin @CarolEidt  PTAL. I've update the PR with your suggestions. Thanks.

PR #12476, opened 2017-06-27
EventPipe: fix memory leaks

kbaladurin:
Fix lsan errors

### DIFF
--- a/src/vm/eventpipebuffermanager.cpp
+++ b/src/vm/eventpipebuffermanager.cpp
@@ -33,6 +33,49 @@ EventPipeBufferManager::EventPipeBufferManager()
 #endif // _DEBUG
 }
 
+EventPipeBufferManager::~EventPipeBufferManager()
+{
+    CONTRACTL
+    {
+        THROWS;
+        GC_TRIGGERS;
+        MODE_ANY;
+    }
+    CONTRACTL_END;
+
+    if(m_pPerThreadBufferList != NULL)
+    {
+        SListElem<EventPipeBufferList*> *pElem = m_pPerThreadBufferList->GetHead();
+        while(pElem != NULL)
+        {
+            SListElem<EventPipeBufferList*> *pCurElem = pElem;
+
+            EventPipeBufferList *pThreadBufferList = pCurElem->GetValue();
+            if (!pThreadBufferList->OwnedByThread())
+            {
+                Thread *pThread = NULL;
+                while ((pThread = ThreadStore::GetThreadList(pThread)) != NULL)
+                {
+                    if (pThread->GetEventPipeBufferList() == pThreadBufferList)
+                    {
+                        pThread->SetEventPipeBufferList(NULL);
+                        break;
+                    }
+                }
+
+                // We don't delete buffers themself because they can be in-use
+                delete(pThreadBufferList);
+            }
+
+            pElem = m_pPerThreadBufferList->GetNext(pElem);
+            delete(pCurElem);
+        }
+
+        delete(m_pPerThreadBufferList);
+        m_pPerThreadBufferList = NULL;
+    }
+}
+
 EventPipeBuffer* EventPipeBufferManager::AllocateBufferForThread(Thread *pThread, unsigned int requestSize)
 {
     CONTRACTL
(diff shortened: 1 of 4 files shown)


src/vm/eventpipeconfiguration.cpp
```diff
@@ -404,7 +419,9 @@ void EventPipeConfiguration::DeleteDeferredProviders()
             delete(pProvider);
         }
 
+        SListElem<EventPipeProvider*> *pCurElem = pElem;
```
brianrob: This should not happen unconditionally.  GetDeleteDeferred() is true if the provider was unloaded during tracing.  If it wasn't unloaded, then it can be used in a subsequent session, and the element should not be removed from the list.

src/vm/eventpipeconfiguration.cpp
```diff
@@ -35,12 +35,26 @@ EventPipeConfiguration::~EventPipeConfiguration()
...
     {
         delete(m_pEnabledProviderList);
         m_pEnabledProviderList = NULL;
     }
 
+    SListElem<EventPipeProvider*> *pElem = m_pProviderList->GetHead();
+    while(pElem != NULL)
+    {
+        SListElem<EventPipeProvider*> *pCurElem = pElem;
+        pElem = m_pProviderList->GetNext(pElem);
```
brianrob: I think you'll also want to delete the provider itself as well as the list element.

kbaladurin: No, m_pProviderList contains global providers (for example, EventPipeProviderDotNETRuntime) and deleting them leads to uaf.

src/vm/eventpipeconfiguration.cpp
kbaladurin: Thanks for note. I've fixed it.

kbaladurin:
@brianrob, thanks for review. I tested it using simple C# programs that used to test gdbjit. To run them I used coreclr compiled with asan. Are there special tests for validation this changes?

karelz:
What is status of this PR? What is the next step?
There hasn't been any update for last 2 months :( ...

src/vm/eventpipebuffermanager.cpp
```diff
@@ -32,6 +32,23 @@ EventPipeBufferManager::EventPipeBufferManager()
...
+{
+    CONTRACTL
+    {
+        THROWS;
+        GC_TRIGGERS;
+        MODE_ANY;
+    }
+    CONTRACTL_END;
+
+    if(m_pPerThreadBufferList != NULL)
```
brianrob: If you're going to delete the list, you should also delete the list items (though not the buffers themselves - these could possibly be in-use).

src/vm/eventpipebuffermanager.cpp
```diff
@@ -184,7 +201,7 @@ EventPipeBufferList* EventPipeBufferManager::FindThreadToStealFrom()
                 pOldestContainingList = pCandidate;
             }
             // Otherwise, to replace the existing candidate, this candidate must have an older timestamp in its oldest buffer.
-            else if((pOldestContainingList->GetHead()->GetMostRecentTimeStamp().QuadPart) > 
+            else if((pOldestContainingList->GetHead()->GetMostRecentTimeStamp().QuadPart) >
```
brianrob: Nit: Can you please remove the whitespace only changes?  It looks like there are more non-memleak changes.

brianrob:
@kbaladurin, @karelz sorry for the delay.  This one slipped off my radar.

src/vm/eventpipebuffermanager.cpp
kbaladurin: I've added deletion of the list items too.

src/vm/eventpipebuffermanager.cpp
kbaladurin: I've removed whitespace changes.

kbaladurin:
@brianrob thank you for review! I've updated PR.

brianrob:
@kbaladurin.  Thanks!  I've done a bit of manual testing as well, and this looks good.

kbaladurin:
@brianrob thank you!

PR #12478, opened 2017-06-27
Fix bad range check in TextInfo.IndexOfStringOrdinalIgnoreCase

AtsushiKan:
Breaks:

   int idx = "\uabcd".IndexOf("", 1, StringComparison.OrdinalIgnoreCase);

should return 1, returns -1 instead.

(propagated fix from CoreRT)


tarekgh:
LGTM

AtsushiKan:
@dotnet-bot Test OSX10.12 x64 Checked Build and Test

Issue #9253, opened 2017-02-01
Remove Microsoft.NETCore.Platforms from test_runtime\project.json


wtgodbe:
We shouldn't need https://github.com/dotnet/coreclr/blob/master/tests/src/Common/test_runtime/project.json#L3 this entry here, as the entry in test_dependencies/project.json should take care of this dependency - removing that line from test_runtime causes build breaks that look like:

> error : One or more packages are incompatible with .NETCoreApp,Version=v1.1 (ubuntu.16.10-x64). [D:\j\workspace\debug_windows---17180f1b\tests\build.proj]

So something must be overwriting the Microsoft.NETCore.Platforms from test_depencies/project.json.

wtgodbe:
CC @ramarag @dagood

am11:
@wtgodbe, project.json files were removed as part of CLI 2.0 work, so the link in first post is invalidated. Should:

* https://github.com/dotnet/coreclr/blob/c4876c4/tests/scripts/project.json and https://github.com/dotnet/coreclr/blob/c4876c4/tests/debugger_tests/ScriptGenerator/project.json, be removed?
* the commented out code in test csproj files like:
  ```xml
  <!--  <PropertyGroup>
      <ProjectJson>$(JitPackagesConfigFileDirectory)threading+thread\project.json</ProjectJson>
  ```
  also be removed?

wtgodbe:
I would think the scripts project.json files can be removed, unless they are explicitly being restored somewhere - @smile21prc do you know if those are still in use?

As for the commented out code, that can definitely be removed.

wtgodbe:
https://github.com/dotnet/coreclr/blob/master/tests/src/Common/test_runtime/test_runtime.csproj#L13-L15 here is the equivalent thing in the new .csproj

am11:
I was just browsing the code and found few of these remaining occurrences of project.json. Also found one in main readme:

> There are already versions of the CoreFX packages published on https://www.nuget.org/ so you can have your test application's `project.json` specify the CoreCLR you built and it will naturally pull anything else it needs ..

Perhaps, `project.json` in that text can be replaced by `project file` (without formatting)?

PR #12484, opened 2017-06-27
Cleanup some instances of project.json from the repo

am11:
As per the discussion with @wtgodbe on #9253.

With this, there are still 16 matches across 6 files for `project.json` and `project.lock.json` combined. I didn't touched those as I am not sure about the suitable replacements. Those include `helixperftasks.targets`, `dir.props` (two files), `dir.common.props` and `runtest.proj`.


danmosemsft:
Thank you for doing this!

PR #12486, opened 2017-06-27
Pin GC threads to corresponding processors on Linux

helloguo:
This PR fixes `GetThreadGroupAffinity` and `SetThreadGroupAffinity`, adds `SetThreadAffinityMask` API and pins GC threads to corresponding processors on Linux.

### DIFF
--- a/src/pal/src/numa/numa.cpp
+++ b/src/pal/src/numa/numa.cpp
@@ -595,7 +595,7 @@ SetThreadAffinityMask(
 
     if (st == 0)
     {
-        for (int i = 0; i < g_possibleCpuCount; i++)
+        for (int i = 0; i < min(8 * sizeof(KAFFINITY), g_possibleCpuCount); i++)
         {
             if (CPU_ISSET(i, &prevCpuSet))
             {
@@ -620,17 +620,17 @@ SetThreadAffinityMask(
 
     st = pthread_setaffinity_np(thread, sizeof(cpu_set_t), &cpuSet);
 
-    if (st == -1)
+    if (st != 0)
     {
-        switch (errno)
+        switch (st)
         {
         case EINVAL:
             // There is no processor in the mask that is allowed to execute the
             // process
             SetLastError(ERROR_INVALID_PARAMETER);
             break;
-        case EPERM:
-            SetLastError(ERROR_ACCESS_DENIED);
+        case ESRCH:
+            SetLastError(ERROR_INVALID_HANDLE);
             break;
         default:
             SetLastError(ERROR_GEN_FAILURE);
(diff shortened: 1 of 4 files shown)


helloguo:
@janvorli PTAL

src/pal/src/numa/numa.cpp
```diff
@@ -527,6 +554,103 @@ SetThreadGroupAffinity(
...
+
+    int st = pthread_getaffinity_np(thread, sizeof(cpu_set_t), &prevCpuSet);
+
+    if (st == 0)
+    {
+        for (int i = 0; i < g_possibleCpuCount; i++)
+        {
+            if (CPU_ISSET(i, &prevCpuSet))
+            {
+                prevMask |= ((KAFFINITY)1) << i;
```
janvorli: C++ standard says;
> If the value of the right operand is negative or is greater or equal to the number of bits in the promoted left operand, the behavior is undefined.

You could use min(8 * sizeof(KAFFINITY), g_possibleCpuCount) as the loop limit.

janvorli:
@Maoni0 I am leaving for my vacation in the morning, could you please merge the change in case @helloguo addresses my comment after I'll leave?

helloguo:
@janvorli @Maoni0 Since the change in src/vm/gcenv.os.cpp will impact GC on Linux, do you have any suggestions in terms of performance test?

janvorli:
@helloguo I think we don't currently have any performance test that would test this stuff. I am planning to do extensive testing at some point in August. The 2.1 release will focus on performance, so this aligns well with that. My preliminary thinking is that I would use some asp.net benchmark running on a machine with a lot of cores.

src/pal/src/numa/numa.cpp
```diff
@@ -527,6 +554,103 @@ SetThreadGroupAffinity(
...
+        {
+            CPU_SET(cpu, &cpuSet);
+        }
+        cpu++;
+        dwThreadAffinityMask >>= 1;
+    }
+
+    st = pthread_setaffinity_np(thread, sizeof(cpu_set_t), &cpuSet);
+
+    if (st == -1)
```
adityamandaleeka: I haven't checked this myself yet, but are you sure this is the right way to handle errors for this function? Based on the man pages I'm seeing (for example, [this one](https://linux.die.net/man/3/pthread_setaffinity_np)), it looks like this function just returns the error code directly, so you'd just check for a nonzero return value and switch based on that value rather than `errno`.

helloguo: You are right. will fix it.

Maoni0:
@janvorli sure

helloguo:

helloguo:
@Maoni0 @adityamandaleeka PTAL

PR #12487, opened 2017-06-27
Add a new, faster tree walker.

pgavlin:
This walker replaces the function-pointer-based walkers with a more
modern design centered around the curiously-recurring-template pattern.

These changes are staged as follows:
- The first commit adds the new walker
- The second commit reimplements the existing `fgWalkTree` functions in
   terms of the new walker
- The rest of the commits reimplement particularly hot tree walks in terms of
   the new walker

All told, this reduces the number of instructions retired when running
`crossgen S.P.C.dll` on Linux/x64/Release by 1-2%.


mikedn:
Cool! You may want to keep an eye on the walker code size but I don't think we have that many visitor uses to risk code explosion.

pgavlin:
> Cool! You may want to keep an eye on the walker code size but I don't think we have that many visitor uses to risk code explosion.

So far the code size changes are minimal. The baseline Windows/x86/release build is 679kB; the diff x86/release build is 680kB. Definitely something to watch out for, though :)

cc @BruceForstall @CarolEidt for a preview of this work.

pgavlin:
This is now ready for review.

cc @dotnet/jit-contrib

mikedn:
> All told, this reduces the number of instructions retired when running
crossgen S.P.C.dll on Linux/x64/Release by 1-2%.

On Windows the improvement may be even higher because it avoids the `/guard:cf` overhead associated with indirect calls.

src/jit/compiler.h
```diff
@@ -9405,6 +9391,523 @@ inline LclVarDsc::LclVarDsc(Compiler* comp)
...
+                    }
+                }
+
+                break;
+            }
+
+            // Binary nodes
+            default:
+            {
+                GenTreeOp* const op = node->AsOp();
```
CarolEidt: Since `AsOp()` will also succeed for both `GTK_UNOP` nodes as well as anything derived from `GenTreeOp`, it would be good to assert here that you actually have a `GTK_BINOP`.

src/jit/compiler.h
```diff
@@ -9405,6 +9391,523 @@ inline LclVarDsc::LclVarDsc(Compiler* comp)
 {
 }
 
+template <typename TVisitor>
```
CarolEidt: This deserves a header comment, e.g. restrictions on `DoLclVarsOnly` and the walk order.

src/jit/compiler.h
pgavlin: Done.

src/jit/compiler.h
pgavlin: Done.

pgavlin:
@dotnet-not test Tizen armel Cross Releae Build

pgavlin:

src/jit/compiler.h
```diff
@@ -9405,6 +9391,579 @@ inline LclVarDsc::LclVarDsc(Compiler* comp)
 {
 }
 
+//---------------------------------------------------------------------------------------------------------------------
+// GenTreeVisitor: a flexible tree walker implemented using the curiosly-recurring-template pattern.
```
BruceForstall: typo: curiosly

PR #12496, opened 2017-06-27
Fix debug assert

JeremyKuhne:
Can't access handle field here

### DIFF
--- a/src/mscorlib/shared/System/IO/FileStream.Windows.cs
+++ b/src/mscorlib/shared/System/IO/FileStream.Windows.cs
@@ -616,7 +616,7 @@ namespace System.IO
         // internal position
         private long SeekCore(SafeFileHandle fileHandle, long offset, SeekOrigin origin, bool closeInvalidHandle = false)
         {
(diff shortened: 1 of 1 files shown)


src/mscorlib/shared/System/IO/FileStream.Windows.cs
```diff
@@ -616,8 +616,8 @@ public override long Seek(long offset, SeekOrigin origin)
         // internal position
         private long SeekCore(SafeFileHandle fileHandle, long offset, SeekOrigin origin, bool closeInvalidHandle = false)
         {
-            Debug.Assert(!_fileHandle.IsClosed && _canSeek, "!_handle.IsClosed && _parent.CanSeek");
-            Debug.Assert(origin >= SeekOrigin.Begin && origin <= SeekOrigin.End, "origin>=SeekOrigin.Begin && origin<=SeekOrigin.End");
+            Debug.Assert(!fileHandle.IsClosed && _canSeek, "!fileHandle.IsClosed && -canSeek");
```
vitek-karas: Should be _canSeek I assume.

JeremyKuhne: Yeah, missing a shift. :)

PR #12497, opened 2017-06-27
Fix remove in list

kyulee1:
`remove` uses `erase` to delete the entry in list.
Since `erase` returns the next iterator, the iterator shouldn't be increased
when the entry is found and deleted.


kyulee1:
@dotnet/jit-contrib PTAL

PR #12498, opened 2017-06-27
spmi add R2R field to getFieldInfo

sandreenko:
There was a missed field with R2R data in Agnostic_GetFieldInfo.
The fix decreases number of failed methods during a replay from 1612 to 162.
162 is the number that we were not able to crossgen, maybe in this case we should not create an mc file for such methods in the future.


sandreenko:
cc @CarolEidt @dotnet/jit-contrib

sandreenko:
It is an additional fix to #11702.

sandreenko:
No, do you mean desktop spmi files? No, because we use spmi.exe from the drop location that doesn't have this change.

PR #12500, opened 2017-06-27
Update throughput exclusion list

adiaaida:
There are many assemblies in our throughput benchmark package that are
reference or forwarding assemblies. There is no jit work in these
assemblies, so we want to exclude them.

This change also adds an iterations parameter so users can specify a
specific number of iterations for testing.

### DIFF
--- a/tests/scripts/run-throughput-perf.py
+++ b/tests/scripts/run-throughput-perf.py
@@ -34,18 +34,49 @@ import csv
 
 # List of dlls we want to exclude
 dll_exclude_list = {
-    'Windows_NT': [
+    'All': [
         # Require Newtonsoft.Json
         "Microsoft.DotNet.ProjectModel.dll",
         "Microsoft.Extensions.DependencyModel.dll",
         # Require System.Security.Principal.Windows
         "System.Net.Requests.dll",
         "System.Net.Security.dll",
-        "System.Net.Sockets.dll"
+        "System.Net.Sockets.dll",
+        # Moving target. Not a reliable measurement
+        "System.Private.CoreLib.dll",
+        # Reference and forwarding assemblies
+        "System.AppContext.dll",
+        "System.Diagnostics.Contracts.dll",
+        "System.Dynamic.Runtime.dll",
+        "System.Globalization.dll",
+        "System.Globalization.Calendars.dll",
+        "System.IO.dll",
+        "System.IO.FileSystem.Primitives.dll",
+        "System.Reflection.dll",
+        "System.Reflection.Emit.dll",
+        "System.Reflection.Emit.ILGeneration.dll",
+        "System.Reflection.Emit.Lightweight.dll",
+        "System.Reflection.Extensions.dll",
+        "System.Reflection.Primitives.dll",
+        "System.Resources.ResourceManager.dll",
+        "System.Runtime.Handles.dll",
+        "System.Runtime.Loader.dll",
+        "System.Runtime.Serialization.Json.dll",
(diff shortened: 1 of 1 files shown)


adiaaida:
@dotnet/rap-team PTAL

adiaaida:
@dotnet-bot test Windows_NT x64 throughput please
@dotnet-bot test Linux throughput flow please

tests/scripts/run-throughput-perf.py
```diff
@@ -140,6 +173,9 @@ def validate_arg(arg, check):
     validate_arg(arch, lambda item: item in valid_archs[os_group])
     validate_arg(build_type, lambda item: item in valid_build_types)
     validate_arg(run_type, lambda item: item in valid_run_types)
+    validate_arg(iterations, lambda item: isinstance(item, int) or item.isdigit())
```
jorive: > validate_arg(iterations, lambda item: isinstance(item, int) or item.isdigit()) [](start = 3, length = 79)

Should we validate the it is greater than zero?

tests/scripts/run-throughput-perf.py
```diff
@@ -250,7 +287,7 @@ def runIterations(dll_name, dll_path, iterations, crossgen_path, jit_path, assem
     # Time.clock() returns seconds, with a resolution of 0.4 microseconds, so multiply by the multiplier to get milliseconds
     multiplier = 1000
 
-    for iteration in range(0,iterations):
+    for iteration in range(0,iterations + 1):
```
jorive: >1 [](start = 42, length = 1)

This is the warm up run, right?

tests/scripts/run-throughput-perf.py
```diff
@@ -166,6 +202,7 @@ def validate_arg(arg, check):
     log(' os_group: %s' % os_group)
     log(' build_type: %s' % build_type)
     log(' run_type: %s' % run_type)
+    log(' iterations: %s' % iterations)
```
dpodder: Is `%s` correct for integers? Or is iterations a string?

tests/scripts/run-throughput-perf.py
dpodder: Going backwards from my earlier question: if iterations is a string, does "+ 1" work for it? (Or is it an int?)

adiaaida: yes

tests/scripts/run-throughput-perf.py
adiaaida: %s can be used with strings (this has been tested)

tests/scripts/run-throughput-perf.py
adiaaida: iterations is an int.

tests/scripts/run-throughput-perf.py
adiaaida: isdigit will return false if the the thing being checked is anything other than a series of 0-9 (ie, only digits. It will return false if iterations passed as a negative number). The only time where item is not a string at this point is the default of 5.

jorive: what if the user pass `-iterations 0`?

adiaaida: It would run for 1 iteration. I can change that.

adiaaida:
@dpodder @jorive this has been updated. Python has the ability to specify the type of the parameter, so I now specify it as an int, which gets rid of a lot of the weirdness around the iterations parameter.

adiaaida:
@dpodder @jorive PTAL

PR #12504, opened 2017-06-28
[x86/Linux] Adjust SP in throwing helper block

### LINKED ISSUE #11820: [x86/Linux] In checked mode, TC fails associated with CORINFO_HELP_OVERFLOW

parjong:
CodeGen::genAdjustStackLevel currently adjusts SP only for throwing helper blocks in ESP-framed methods, which results in #11820 for EBP-framed methods.

This PR revises CodeGen::genAdjustStackLevel to adjust SP even for throwing helper blocks in EBP-framed methods in order to fix #11820.

This PR also enforces JIT to inline exception throwing calls inside funclet region to prevent potential unwind issues.

### DIFF
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -1647,9 +1647,11 @@ void CodeGen::genAdjustStackLevel(BasicBlock* block)
 #if defined(UNIX_X86_ABI)
     if (isFramePointerUsed() && compiler->fgIsThrowHlpBlk(block))
     {
-        int offset = compiler->compCalleeRegsPushed * REGSIZE_BYTES + compiler->compLclFrameSize;
-        getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_SPBASE, REG_FPBASE, -offset);
-        SetStackLevel(0);
+        // x86/Linux requires stack frames to be 16-byte aligned, but SP may be unaligned
+        // at this point if a jump to this block is made in the middle of pushing arugments.
+        //
+        // Here we restore SP to prevent potential stack alignment issue.
+        getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_SPBASE, REG_FPBASE, -genSPtoFPdelta());
     }
 #endif
 
--- a/src/jit/codegencommon.cpp
+++ b/src/jit/codegencommon.cpp
@@ -1642,6 +1642,14 @@ void CodeGen::genAdjustStackLevel(BasicBlock* block)
 {
 #if !FEATURE_FIXED_OUT_ARGS
     // Check for inserted throw blocks and adjust genStackLevel.
+#if defined(UNIX_X86_ABI)
+    if (isFramePointerUsed() && compiler->fgIsThrowHlpBlk(block))
+    {
+        int offset = compiler->compCalleeRegsPushed * REGSIZE_BYTES + compiler->compLclFrameSize;
+        getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_SPBASE, REG_FPBASE, -offset);
+        SetStackLevel(0);
+    }
+#endif
 
     if (!isFramePointerUsed() && compiler->fgIsThrowHlpBlk(block))
     {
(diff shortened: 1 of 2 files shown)


parjong:

BruceForstall:
cc @dotnet/jit-contrib

src/jit/codegencommon.cpp
```diff
@@ -2698,7 +2708,14 @@ void CodeGen::genExitCode(BasicBlock* block)
 
 void CodeGen::genJumpToThrowHlpBlk(emitJumpKind jumpKind, SpecialCodeKind codeKind, GenTreePtr failBlk)
 {
-    if (!compiler->opts.compDbgCode)
+    bool useThrowHlpBlk = !compiler->opts.compDbgCode;
+
+#if defined(UNIX_X86_ABI) && FEATURE_EH_FUNCLETS
+    // Inline exception-throwing code in funclet to make it possible to unwind funclet frames.
+    useThrowHlpBlk = useThrowHlpBlk && !siInFuncletRegion;
+#endif // UNIX_X86_ABI && FEATURE_EH_FUNCLETS
```
BruceForstall: Instead of using `!siInFuncletRegion`, which uses a variable specific to the scope info code, you should use `(compiler->funCurrentFunc()->funKind != FUNC_ROOT)`, which is more generally used.

src/jit/codegencommon.cpp
```diff
@@ -1642,6 +1642,16 @@ void CodeGen::genAdjustStackLevel(BasicBlock* block)
 {
 #if !FEATURE_FIXED_OUT_ARGS
     // Check for inserted throw blocks and adjust genStackLevel.
+    CLANG_FORMAT_COMMENT_ANCHOR;
+
+#if defined(UNIX_X86_ABI)
+    if (isFramePointerUsed() && compiler->fgIsThrowHlpBlk(block))
+    {
+        int offset = compiler->compCalleeRegsPushed * REGSIZE_BYTES + compiler->compLclFrameSize;
+        getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_SPBASE, REG_FPBASE, -offset);
```
BruceForstall: Instead, use:
```
int offset = genSPtoFPdelta();
```
or, simply put it in-line:
```
 getEmitter()->emitIns_R_AR(INS_lea, EA_PTRSIZE, REG_SPBASE, REG_FPBASE, -genSPtoFPdelta());
```

Can you please also add a comment about why this is required for Linux/x86?

src/jit/codegencommon.cpp
```diff
@@ -1642,6 +1642,16 @@ void CodeGen::genAdjustStackLevel(BasicBlock* block)
...
+        SetStackLevel(0);
+    }
```
BruceForstall: This isn't necessary; the stack level is already set to zero on entry to this function.

BruceForstall:
Can you summarize why this is required for EBP framed methods? The code in fgAddCodeRef indicates that the only reason we need to adjust the stack in these throw helpers is to unwind the stack properly in those throw helper blocks. And, for EBP frames, we don't need to know the stack level to do that.

parjong:
@BruceForstal The problem that we want to address here is stack alignment issue (yes, unwinder may be happy even with the current implementation).

The current implementation simply inserts a jump to this block without adjusting SP, and thus SP is not 16-byte aligned if the jump is made in the middle of pushing arguments,
In this case, will causes segmentation fault while running exception throwing helpers.

BruceForstall:
Ok, so in `CodeGen::genAdjustStackLevel()`, if `compiler->fgIsThrowHlpBlk(block)` is true, for Linux/x86 we have two cases:
1. `isFramePointerUsed()` -- In this case, we don't know what the stack level is on entry to the block. It could be anything. So you have chosen to re-establish ESP at the same offset from EBP as it was at the end of the prolog. This seems reasonable, since we know that level will be aligned.
2. `!isFramePointerUsed()` -- the existing code, where there will be exactly one place that calls this throw helper block, and so we know exactly what the stack level is, so we can adjust it appropriately.

parjong:
Yes, that is exactly what this PR does.

src/jit/flowgraph.cpp
```diff
@@ -17776,6 +17776,10 @@ BasicBlock* Compiler::fgAddCodeRef(BasicBlock* srcBlk, unsigned refData, Special
 #if defined(UNIX_X86_ABI)
         codeGen->setFrameRequired(true);
         codeGen->setFramePointerRequiredGCInfo(true);
+        add->acdStkLvl = 0;
+#ifdef DEBUG
+        add->acdDstBlk->bbTgtStkDepth = 0;
+#endif
 #else  // !defined(UNIX_X86_ABI)
```
BruceForstall: Are these changes necessary? It doesn't look like `CodeGen::genAdjustStackLevel()` requires them.

parjong: ``fgThrowHlpBlkStkLevel`` called from ``CodeGen::genAdjustStackLevel`` uses this, but it seems that this is not relevant  for EBP frame cases. I'll check and update PR if it is really irrelevant.

BruceForstall:
Ignoring x64 chaos test failure

Issue #12169, opened 2017-06-08
Spin Count in GC


helloguo:
In gc.cpp, there are a few places using the number of processor to get the spin count. e.g.

```
                int spin_count = 4096 * g_num_processors;
                for (int j = 0; j < spin_count; j++)
                {
                    if (color != join_struct.lock_color)
                    {
                        break;
                    }
                    YieldProcessor();           // indicate to the processor that we are spinning
                }
```

Since the number of processors are not necessarily same as the number of GC threads (e.g. https://github.com/dotnet/coreclr/pull/6104 adds System.GC.HeapCount to specify GC thread counts by config), it makes more sense for me use the number of GC threads to calculate spin count.

Should I fix it?

helloguo:
cc @Maoni0

Maoni0:
That's intentional. There can be threads running on other processors too. We don't assume only GC threads are running.

helloguo:
@Maoni0 `g_num_processors `has different values on Windows v.s. Ubuntu. On my machine (2 sockets, 88 logic cores in total), `g_num_processors  `is 44 for Windows Server 2012 R2 but 88 for Ubuntu 14.04.

swgillespie:
That number should come straight from the PAL: https://github.com/dotnet/coreclr/blob/master/src/vm/gcenv.os.cpp#L687.

helloguo:
Windows and Ubuntu should return same value (88 in this case), right?

swgillespie:
I would think so, yeah; if they aren't that would imply that something has gone awry in PAL's GetSystemInfo: https://github.com/dotnet/coreclr/blob/master/src/pal/src/misc/sysinfo.cpp#L100

helloguo:
PAL works well for Ubuntu, which gives 88. I would suspect Windows `GetSystemInfo` does something unexpected.

swgillespie:
hmm... @janvorli, do you have any thoughts on this?

janvorli:
MSDN doc on the `SYSTEM_INFO::dwNumberOfProcessors` says:
[...]

PR #12525, opened 2017-06-28
Use the number of GC threads to calculate spin count

helloguo:
Based on the discussion of https://github.com/dotnet/coreclr/issues/12169, this PR uses the number of GC threads to calculate spin count in certain places (`join()`, `r_join()`, `gc_thread_function()`) in GC.

### DIFF
--- a/src/gc/gc.cpp
+++ b/src/gc/gc.cpp
@@ -749,7 +749,7 @@ public:
             if (color == join_struct.lock_color)
             {
 respin:
-                int spin_count = 4096 * g_num_processors;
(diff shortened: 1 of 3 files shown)


helloguo:
@Maoni0 PTAL

Maoni0:
There's no need to add a new variable. We already have n_heaps which is the # of heaps for Server GC. All the places you changed were Server GC only (ie, inside #ifdef MULTIPLE_HEAPs), so you can just use n_heaps.

helloguo:
@Maoni0 Thanks for pointing it out. Fixed.

Issue #11828, opened 2017-05-23
Update the publish definition to match what is in release/2.0.0 branch


weshaggard:
Need to port https://github.com/dotnet/coreclr/commit/ba86da4a6c821b0f042e105a72fd901c7f7b94c3 and https://github.com/dotnet/coreclr/commit/8ab4fb4260bdf1dafdf4b050bcc01102437f875d to master once the builds become stable.

weshaggard:
Fixed with https://github.com/dotnet/coreclr/pull/12526

PR #12526, opened 2017-06-28
Enable SourceBranch to be passed from the pipeline parameters

jcagme:
sets SkipBranchAndVersionOverrides=false so SourceBranch and SourceVersion
are correctly passed through the pipeline file.

set SkipBranchAndVersionOverrides=true for the publish definitions as
they don't build from the SourceBranch and instead use a different
tools branch to do their work.

### DIFF
--- a/buildpipeline/pipelines.json
+++ b/buildpipeline/pipelines.json
@@ -3,7 +3,8 @@
   "Definitions": {
     "Path": ".",
     "Type": "VSTS",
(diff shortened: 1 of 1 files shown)


weshaggard:
Looks like this address https://github.com/dotnet/coreclr/issues/11828.

jcagme:
The Ubuntu arm Cross Release check shows as failed but looking at the details there were no identified problems. Merging

Issue #12128, opened 2017-06-07
Need a way to build a RELEASE build without PGD data


BruceForstall:
If I want to build a RELEASE build on Windows without PGD data (that is, specifically preventing PGD data from being used), is there a way to do so? Ideally, I could pass something like "nopgo" to build.cmd.

It looks like right now I need to hack pgosupport.cmake to command out or remove these two lines:

```
set_property(TARGET ${TargetName} APPEND_STRING PROPERTY LINK_FLAGS_RELEASE        " /LTCG /USEPROFILE:PGD=${ProfilePath}")
set_property(TARGET ${TargetName} APPEND_STRING PROPERTY LINK_FLAGS_RELWITHDEBINFO " /LTCG /USEPROFILE:PGD=${ProfilePath}")        
```

?

dpodder:
At the moment the easiest way is to add `skiprestoreoptdata` (and make sure the packages weren't already restored: `./clean.sh -p` or `clean.cmd -p`.

There are plans to add a switch like `pgo:off` (exact syntax TBD). I can work on this and get it wrapped up pretty quickly once 2.0 validation winds down.

PR #12528, opened 2017-06-28
Add a flag to build.{sh,cmd} to disable PGO.

pgavlin:
This is helpful when performing JIT throughput measurements.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -624,7 +624,7 @@ echo     for the specified platform ^(FreeBSD, Linux, NetBSD, OS X or Windows,
 echo     respectively^).
 echo     add nativemscorlib to go further and build the native image for designated mscorlib.
 echo toolset_dir ^<dir^> : set the toolset directory -- Arm64 use only. Required for Arm64 builds.
-echo nopgooptimize - do not use profile guided optimizations.
+echo nopgooptimize: do not use profile guided optimizations.
 echo pgoinstrument: generate instrumented code for profile guided optimization enabled binaries.
 echo ibcinstrument: generate IBC-tuning-enabled native images when invoking crossgen.
 echo configureonly: skip all builds; only run CMake ^(default: CMake and builds are run^)
--- a/build.cmd
+++ b/build.cmd
@@ -79,6 +79,7 @@ set __BuildStandaloneGC="-DFEATURE_STANDALONE_GC=0"
 set __BuildStandaloneGCOnly="-DFEATURE_STANDALONE_GC_ONLY=0"
 
 set __PgoInstrument=0
+set __PgoOptimize=1
 set __IbcTuning=
 
 REM __PassThroughArgs is a set of things that will be passed through to nested calls to build.cmd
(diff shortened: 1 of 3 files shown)


pgavlin:
@dpodder @BruceForstall PTAL

BruceForstall:
Related to #12128

build.cmd
```diff
@@ -138,6 +139,7 @@ if /i "%1" == "skipbuildpackages"   (set __BuildPackages=0&set processedArgs=!pr
 if /i "%1" == "skiprestoreoptdata"  (set __RestoreOptData=0&set processedArgs=!processedArgs! %1&shift&goto Arg_Loop)
 if /i "%1" == "usenmakemakefiles"   (set __NMakeMakefiles=1&set __ConfigureOnly=1&set __BuildNative=1&set __BuildNativeCoreLib=0&set __BuildCoreLib=0&set __BuildTests=0&set __BuildPackages=0&set processedArgs=!processedArgs! %1&shift&goto Arg_Loop)
 if /i "%1" == "pgoinstrument"       (set __PgoInstrument=1&set processedArgs=!processedArgs! %1&shift&goto Arg_Loop)
+if /i "%1" == "nopgooptimize"       (set __PgoOptimize=0&set processedArgs=!processedArgs! %1&shift&goto Arg_Loop)
 if /i "%1" == "ibcinstrument"       (set __IbcTuning=/Tuning&set processedArgs=!processedArgs! %1&shift&goto Arg_Loop)
```
BruceForstall: You need to add this to the "usage" message, around line 630

build.sh
```diff
@@ -730,6 +731,11 @@ while :; do
             __PgoInstrument=1
             ;;
 
+        nopgooptimize)
+            __PgoOptimize=0
```
BruceForstall: Add to the usage message, around line 40

build.cmd
pgavlin: Done.

build.sh
pgavlin: Done.

build.cmd
```diff
@@ -158,6 +160,8 @@ if [!processedArgs!]==[] (
 
 :ArgsDone
 
+if %__PgoOptimize%=0 set __RestoreOptData=0
```
dpodder: Wrong syntax--you need double-equals. To keep with script conventions, I recommend either
```batch
if "%__PgoOptimize%" == "0" set __RestoreOptData=0
```
or
```batch
if [%__PgoOptimize%] == [0] set __RestoreOptData=0
```

build.cmd
```diff
@@ -620,6 +624,7 @@ echo     for the specified platform ^(FreeBSD, Linux, NetBSD, OS X or Windows,
 echo     respectively^).
 echo     add nativemscorlib to go further and build the native image for designated mscorlib.
 echo toolset_dir ^<dir^> : set the toolset directory -- Arm64 use only. Required for Arm64 builds.
+echo nopgooptimize - do not use profile guided optimizations.
```
BruceForstall: super-nit: the cmd file standard seems to be ": ", not " - " (like in the build.sh file).

pgavlin:
> I had previously discussed with @janvorli to refactor this to be a single mode to support multiple states, e.g.: `pgo:instrument`, `pgo:force`, `pgo:auto` (default), or `pgo:off`. Could we go with the latter name for the new switch, so that we don't have to introduce a breaking change later? Or is `nopgooptimize` preferred, and should we add `forcepgooptimize` later?

I'm not too concerned with making a breaking change later. This change is meant to ease some of the throughput investigations the JIT team is currently performing, so I'd prefer to leave it as-is and get it checked in sooner.

build.cmd
pgavlin: Fixed.

dpodder:
Sounds good. LGTM.

pgavlin:
@dotnet-bot Tizen armel Cross Debug Build

pgavlin:

PR #12535, opened 2017-06-28
Support DispatcherQueues when getting WinRT dispatcher

adityamandaleeka:
This is a port of work we did on Desktop to support [DispatcherQueues](https://docs.microsoft.com/en-us/uwp/api/windows.system.dispatcherqueue) that are being introduced in Windows. It will ensure that when we post task continuations on dedicated DispatcherQueue threads, the continuations will run on the same thread.

The CoreFX side of this change is in https://github.com/dotnet/corefx/pull/21678 and should go in before this change.

@yizhang82 @sergiy-k @kouvel @jkotas

### DIFF
--- a/src/vm/synchronizationcontextnative.cpp
+++ b/src/vm/synchronizationcontextnative.cpp
@@ -17,6 +17,7 @@
 #ifdef FEATURE_APPX
 #include <roapi.h>
 #include <windows.ui.core.h>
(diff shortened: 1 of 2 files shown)


adityamandaleeka:
@dotnet-bot test Tizen armel Cross Release Build (unrelated failure)

PR #12536, opened 2017-06-28
Make ResourceManager netstandard 2.0 API behave correctly

tarekgh:
In netstandard 2.0 we have enabled the APIs like GetObject, GetStream,..etc. these are not functioning correctly in UWP apps because was assuming we always uses PRI files. The changes here is to initialize the managed resources too and let these APIs behave as if we are running inside desktop app.

### DIFF
--- a/src/mscorlib/src/System/Resources/ResourceManager.cs
+++ b/src/mscorlib/src/System/Resources/ResourceManager.cs
@@ -438,18 +438,16 @@ namespace System.Resources
         // security check in each constructor prevents it.
         private void CommonAssemblyInit()
         {
-            if (_bUsingModernResourceManagement == false)
-            {
-                UseManifest = true;
+            // Now we can use the managed resources even when using PRI's to support the APIs GetObject, GetStream...etc.
+            UseManifest = true;
 
-                _resourceSets = new Dictionary<String, ResourceSet>();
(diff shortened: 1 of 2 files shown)


tarekgh:
@morganbr @zamont could you please have a look?

src/mscorlib/src/System/Resources/ResourceManager.cs
```diff
@@ -465,7 +463,15 @@ public virtual String BaseName
         public virtual bool IgnoreCase
         {
             get { return _ignoreCase; }
-            set { _ignoreCase = value; }
+            set 
+            { 
+                if (_PRIonAppXInitialized && value)
```
morganbr: Is this more broadly breaking than we had discussed? This would mean that a NS2.0 library isn't allowed to set IgnoreCase to true. I thought the idea was that this should only happen if there were PRI resources matching BaseName.

tarekgh:
@morganbr I found that PRI resource lookup is case insensitive so I have removed the check and the throwing too.

tarekgh:
@dotnet-bot help

tarekgh:

zamont:
LGTM

tarekgh:
Test Tizen armel Cross Release Build please

tarekgh:
Looked at the failures and are not related and contacted the owners too. so I am merging this one

PR #12540, opened 2017-06-29
Add a generic operand visitor to `GenTree`.

pgavlin:
This visitor accepts any invokable argument (e.g. lambdas, functors,
etc.) with the following signature:

    `GenTree::VisitResult Visitor(GenTree* operand)`

This visitor is typically more efficient than the operand iterator
due to better opportunities for inlining and shorter dynamic path
lengths when deciding which operands need to be visited.

As noted below, these changes give about a .6% decrease in instructions retired as reported by callgrind S.P.C.dll on Linux/x64/Release.

### DIFF
--- a/src/jit/compiler.hpp
+++ b/src/jit/compiler.hpp
@@ -4754,7 +4754,7 @@ void GenTree::VisitOperands(TVisitor visitor)
         case GT_IL_OFFSET:
             return;
 
-        // Standard unary operators
+        // Unary operators with an optional operand
         case GT_NOP:
         case GT_RETURN:
         case GT_RETFILT:
@@ -4764,6 +4764,7 @@ void GenTree::VisitOperands(TVisitor visitor)
             }
             __fallthrough;
 
+        // Standard unary operators
         case GT_STORE_LCL_VAR:
         case GT_STORE_LCL_FLD:
         case GT_NOT:
--- a/src/jit/compiler.hpp
+++ b/src/jit/compiler.hpp
@@ -4928,7 +4928,8 @@ void GenTree::VisitOperands(TVisitor visitor)
             {
                 return;
             }
-            if ((call->gtCallLateArgs != nullptr) && (call->gtCallLateArgs->VisitListOperands(visitor)) == VisitResult::Abort)
+            if ((call->gtCallLateArgs != nullptr) &&
+                (call->gtCallLateArgs->VisitListOperands(visitor)) == VisitResult::Abort)
             {
                 return;
             }
--- a/src/jit/compiler.hpp
+++ b/src/jit/compiler.hpp
@@ -4711,6 +4711,287 @@ unsigned Compiler::GetSsaNumForLocalVarDef(GenTreePtr lcl)
     }
 }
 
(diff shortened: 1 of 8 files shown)


mikedn:
> This is quite a bit faster than the operand iterator itself.

How fast and why? Presumably because the visitor is usually small and thus likely to be inlined?

CarolEidt:
I think it would be better to have the visitor function return a more meaningful value - akin to the `fgWalkResult` - it could still be a simple two-valued enum so as not to add complexity, but I think it would be easier to follow the logic than with a boolean where it is easy to lose track of what true and false mean.

pgavlin:
> > This is quite a bit faster than the operand iterator itself.
>
> How fast and why? Presumably because the visitor is usually small and thus likely to be inlined?

These changes give about a .6% decrease in instructions retired as reported by `callgrind S.P.C.dll` on Linux/x64/Release. I believe that it is because of better inlining and tighter code.

> I think it would be better to have the visitor function return a more meaningful value - akin to the `fgWalkResult` - it could still be a simple two-valued enum so as not to add complexity, but I think it would be easier to follow the logic than with a boolean where it is easy to lose track of what true and false mean.

That makes sense--I will make that change.

pgavlin:
This is now ready for review.

pgavlin:
I should note that the diffs in `lsra.cpp` are presented as two commits for the sake of clarity. The first refactors the loop that builds operand use positions and the second replaces the operand iterator with the operand visitor.

src/jit/compiler.hpp
```diff
@@ -4711,6 +4711,287 @@ unsigned Compiler::GetSsaNumForLocalVarDef(GenTreePtr lcl)
...
+        case GT_NOP:
+        case GT_RETURN:
+        case GT_RETFILT:
+            if (this->AsUnOp()->gtOp1 == nullptr)
+            {
+                return;
+            }
+            __fallthrough;
+
+        case GT_STORE_LCL_VAR:
```
briansull: I would put the comment here:

` // Standard unary operators`

And above the comment could
 // Optional unary operators

src/jit/gentree.h
```diff
@@ -2131,6 +2131,35 @@ struct GenTree
...
+    //
+    //     unsigned operandCount = 0;
+    //     node->VisitOperands([&](GenTree* operand) -> GenTree::VisitResult)
+    //     {
+    //         operandCount++;
+    //         return GenTree::VisitResult::Continue;
+    //     });
+    //
+    // This function is generally more efficient that the operand iterator and should be preferred over that API for
+    // hot code.
```
BruceForstall: Seems like it would be helpful to make a statement here about *why* it is more efficient.

src/jit/compiler.hpp
pgavlin: Done.

src/jit/gentree.h
pgavlin: Done.

pgavlin:
Pri1 failures are infrastructural (paging file too small).

PR #12373, opened 2017-06-19
[CoreRT armel] fix for PINVOKE calls in CoreRT

sergign60:

PR #12550, opened 2017-06-29
[CoreRT armel] refix for PINVOKE calls in CoreRT #12373

sergign60:


sergign60:
CC: @Dmitri-Botcharnikov @BredPet @mskvortsov 
@jkotas 
Sorry, but we need to re-do it.

sergign60:
https://github.com/dotnet/corert/pull/4021

sergign60:
@dotnet-bot  Ubuntu arm Cross Release Build test

src/jit/codegenlegacy.cpp
```diff
@@ -19041,7 +19041,10 @@ regMaskTP CodeGen::genCodeForCall(GenTreeCall* call, bool valUsed)
                             indCallReg = regSet.rsGrabReg(RBM_ALLINT); // Grab an available register to use for the CALL
                                                                        // indirection
                             instGen_Set_Reg_To_Imm(EA_HANDLE_CNS_RELOC, indCallReg, (ssize_t)addr);
-                            getEmitter()->emitIns_R_R_I(INS_ldr, EA_PTRSIZE, indCallReg, indCallReg, 0);
+#ifdef _TARGET_ARM_
+                            if (!compiler->opts.ShouldUsePInvokeHelpers())
```
jkotas: This does not look right. Why do you want to skip the indirection for PInvokeHelpers?

sergign60: @dotnet-bot Windows_NT x64 Debug Build and Test test

sergign60:

src/jit/codegenlegacy.cpp
jkotas: The implementation of getAddressOfPInvokeFixup in CoreRT is wrong as well.

You should fix this by calling `getAddressOfPInvokeFixup` here and then use `lookup.accessType` to figure out the number of indirections to use - like it is done in RyuJIT backend.

sergign60: @jkotas Sorry I did not catch you quite well. RyuJIT doesn't use ```getAddressOfPInvokeFixup``` and ```getAddressOfPInvokeFixup``` does not have ```CORINFO_CONST_LOOKUP lookup``` argument as ```getAddressOfPInvokeTarget``` does

jit/lower.cpp (RyuJIT)
```
        CORINFO_CONST_LOOKUP lookup;
        comp->info.compCompHnd->getAddressOfPInvokeTarget(methHnd, &lookup);

        void* addr = lookup.addr;
        switch (lookup.accessType)
```

We can fix 
```
private void* getAddressOfPInvokeFixup(CORINFO_METHOD_STRUCT_* method, ref void* ppIndirection)
```
to return 
```
return &pLookup.addr;
```
instead of
```
return pLookup.addr;
```

jkotas: Sorry, I meant to say:

You should fix this by calling `getAddressOfPInvokeTarget` here and then use `lookup.accessType` to figure out the number of indirections to use - like it is done in RyuJIT backend.

> return &pLookup.addr;

I do not think that this would work as written - it would return address of the `pLookup` local variable.

src/jit/codegenlegacy.cpp
```diff
@@ -19032,16 +19032,31 @@ regMaskTP CodeGen::genCodeForCall(GenTreeCall* call, bool valUsed)
...
+
+                        void* addr = lookup.addr;
+
+                        _ASSERTE(addr != NULL);
+
+                        bool direct;
+#if defined(_TARGET_ARM_)
+                        direct = compiler->codeGen->validImmForBL((ssize_t)addr);
+#else
+                        direct           = true;
```
jkotas: Nit - the indentation looks off.

src/jit/codegenlegacy.cpp
```diff
@@ -19032,16 +19032,31 @@ regMaskTP CodeGen::genCodeForCall(GenTreeCall* call, bool valUsed)
...
+                        if (lookup.accessType == IAT_VALUE || lookup.accessType == IAT_PVALUE)
                         {
 #if CPU_LOAD_STORE_ARCH
                             // Load the address into a register, indirect it and call  through a register
                             indCallReg = regSet.rsGrabReg(RBM_ALLINT); // Grab an available register to use for the CALL
                                                                        // indirection
                             instGen_Set_Reg_To_Imm(EA_HANDLE_CNS_RELOC, indCallReg, (ssize_t)addr);
-                            getEmitter()->emitIns_R_R_I(INS_ldr, EA_PTRSIZE, indCallReg, indCallReg, 0);
+
+                            if (lookup.accessType == IAT_PVALUE || !direct)
```
jkotas: The `|| !direct` part of the condition does not look right. Is there an extra indirection inserted somewhere when for the `!direct` case?

sergign60: @jkotas I meant the condition
```
 if (lookup.accessType == IAT_PVALUE || (lookup.accessType == IAT_VALUE && direct == false))
```

src/jit/codegenlegacy.cpp
sergign60: @jkotas I got the following patch:
format.patch
```
iff --git a/src/jit/codegenlegacy.cpp b/src/jit/codegenlegacy.cpp
index f50847b..7f8f72d 100644
--- a/src/jit/codegenlegacy.cpp
+++ b/src/jit/codegenlegacy.cpp
@@ -19026,41 +19026,41 @@ regMaskTP CodeGen::genCodeForCall(GenTreeCall* call, bool valUsed)
                             regTracker.rsTrackRegTrash(indCallReg);
                         }
 
                         emitCallType = emitter::EC_INDIR_R;
                     }
                     else
                     {
                         noway_assert(callType == CT_USER_FUNC);
 
                         CORINFO_CONST_LOOKUP lookup;
                         compiler->info.compCompHnd->getAddressOfPInvokeTarget(methHnd, &lookup);
 
                         void* addr = lookup.addr;
 
                         _ASSERTE(addr != NULL);
 
                         bool direct;
 #if defined(_TARGET_ARM_)
                         direct = compiler->codeGen->validImmForBL((ssize_t)addr);
 #else
-                        direct = true;
+                        direct           = true;
 #endif
 
                         if (lookup.accessType == IAT_VALUE || lookup.accessType == IAT_PVALUE)
                         {
 #if CPU_LOAD_STORE_ARCH
                             // Load the address into a register, indirect it and call  through a register
                             indCallReg = regSet.rsGrabReg(RBM_ALLINT); // Grab an available register to use for the CALL
                                                                        // indirection
                             instGen_Set_Reg_To_Imm(EA_HANDLE_CNS_RELOC, indCallReg, (ssize_t)addr);
 
                             if (lookup.accessType == IAT_PVALUE || !direct)
                                 getEmitter()->emitIns_R_R_I(INS_ldr, EA_PTRSIZE, indCallReg, indCallReg, 0);
 
                             regTracker.rsTrackRegTrash(indCallReg);
                             // Now make the call "call indCallReg"
 
                             getEmitter()->emitIns_Call(emitter::EC_INDIR_R,
                                                        methHnd,                       // methHnd
                                                        INDEBUG_LDISASM_COMMA(sigInfo) // sigInfo
                                                        NULL,                          // addr
```

jkotas: ok

sergign60:

sergign60:
@dotnet-bot FreeBSD x64 Checked Build test

sergign60:

src/jit/codegenlegacy.cpp
sergign60: @jkotas this fix is under debugging now. So don't ask this question please

sergign60:
https://github.com/dotnet/corert/pull/4134

sergign60:
@BredPet 
@jkotas  please review again

sergign60:

sergign60:
@dotnet-bot test Windows_NT x86 CoreCLR Perf Tests Correctness please

sergign60:
@dotnet-bot  test Tizen armel Cross Release Build please

sergign60:

sergign60:

sergign60:

sergign60:
@dotnet-bot help

src/jit/codegenlegacy.cpp
```diff
@@ -19036,16 +19036,31 @@ regMaskTP CodeGen::genCodeForCall(GenTreeCall* call, bool valUsed)
...
+                        CORINFO_CONST_LOOKUP lookup;
+                        compiler->info.compCompHnd->getAddressOfPInvokeTarget(methHnd, &lookup);
+
+                        void* addr = lookup.addr;
+
+                        _ASSERTE(addr != NULL);
+
+                        bool direct;
+#if defined(_TARGET_ARM_)
+                        direct = compiler->codeGen->validImmForBL((ssize_t)addr);
```
jkotas: `validImmForBL` only make sense for `lookup.accessType == IAT_VALUE`. It should be only called for `lookup.accessType == IAT_VALUE`.

src/jit/codegenlegacy.cpp
```diff
@@ -19036,16 +19036,31 @@ regMaskTP CodeGen::genCodeForCall(GenTreeCall* call, bool valUsed)
...
+                        if (lookup.accessType == IAT_VALUE || lookup.accessType == IAT_PVALUE)
                         {
 #if CPU_LOAD_STORE_ARCH
                             // Load the address into a register, indirect it and call  through a register
                             indCallReg = regSet.rsGrabReg(RBM_ALLINT); // Grab an available register to use for the CALL
                                                                        // indirection
                             instGen_Set_Reg_To_Imm(EA_HANDLE_CNS_RELOC, indCallReg, (ssize_t)addr);
-                            getEmitter()->emitIns_R_R_I(INS_ldr, EA_PTRSIZE, indCallReg, indCallReg, 0);
+
+                            if (lookup.accessType == IAT_PVALUE || (lookup.accessType == IAT_VALUE && !direct))
```
jkotas: `(lookup.accessType == IAT_VALUE && !direct)` is not right. If this condition was ever true, it would result into silent bad codegen, If you do not want to deal with it with in legacy backend  because of it is currently unreachable, add `noway_assert(lookup.accessType == IAT_VALUE && compiler->codeGen->validImmForBL((ssize_t)addr))` here instead. This will make sure that it won't generate bad code silently.

src/jit/codegenlegacy.cpp
```diff
@@ -19047,16 +19047,32 @@ regMaskTP CodeGen::genCodeForCall(GenTreeCall* call, bool valUsed)
...
+                        CORINFO_CONST_LOOKUP lookup;
+                        compiler->info.compCompHnd->getAddressOfPInvokeTarget(methHnd, &lookup);
+
+                        void* addr = lookup.addr;
+
+                        _ASSERTE(addr != NULL);
+
+                        bool direct;
+#if defined(_TARGET_ARM_)
+                        if (lookup.accessType == IAT_VALUE)
```
jkotas: You can move this into the `noway_assert` and get rid of the `direct` local variable.

src/jit/codegenlegacy.cpp
```diff
@@ -19047,16 +19047,32 @@ regMaskTP CodeGen::genCodeForCall(GenTreeCall* call, bool valUsed)
...
+                        if (lookup.accessType == IAT_VALUE || lookup.accessType == IAT_PVALUE)
                         {
 #if CPU_LOAD_STORE_ARCH
                             // Load the address into a register, indirect it and call  through a register
                             indCallReg = regSet.rsGrabReg(RBM_ALLINT); // Grab an available register to use for the CALL
                                                                        // indirection
                             instGen_Set_Reg_To_Imm(EA_HANDLE_CNS_RELOC, indCallReg, (ssize_t)addr);
-                            getEmitter()->emitIns_R_R_I(INS_ldr, EA_PTRSIZE, indCallReg, indCallReg, 0);
+
+                            noway_assert(lookup.accessType == IAT_VALUE && direct);
```
jkotas: It would be useful to have a short comment in front of the `noway_assert`.

sergign60:
@jkotas I tried to copy the logic from lower.cpp file, I see there such code

line 3362-...
```
        void* addr = lookup.addr;
        switch (lookup.accessType)
        {
            case IAT_VALUE:
                if (!IsCallTargetInRange(addr))
                {
                    result = AddrGen(addr);
                }
                else
                {
                    // a direct call within range of hardware relative call instruction
                    // stash the address for codegen
                    call->gtDirectCallAddress = addr;
```
So I conclude from here that when !IsCallTargetInRange that is ```comp->codeGen->validImmForBL((ssize_t)addr);``` we need to add additional addr taking. And tests fails show this, I think

src/jit/codegenlegacy.cpp
```diff
@@ -19077,21 +19088,22 @@ regMaskTP CodeGen::genCodeForCall(GenTreeCall* call, bool valUsed)
                         }
                         else
                         {
+                            _ASSERTE(lookup.accessType == IAT_PPVALUE);
```
jkotas: Nit: The JIT code is using regular `assert(...)`. `_ASSERTE` is used in VM.

jkotas:
> we need to add additional addr taking

It is not additional addr taking what would be required here. It would be additional address generation into scratch register that would be required to make the `IAT_VALUE` case that does not fit work.

> And tests fails show this, I think

Changing the `noway_assert` to something like this should fix the problem:

```
#if defined(_TARGET_ARM_) 
    // Legacy backend does not handle the `IAT_VALUE` case that does not fit. It is not reachable currently
    // from any front end so just check for it via noway_assert.
    noway_assert(lookup.accessType != IAT_VALUE && compiler->codeGen->validImmForBL((ssize_t)addr); 
#endif
```

src/jit/codegenlegacy.cpp
```diff
@@ -20853,22 +20867,21 @@ GenTreePtr Compiler::fgLegacyPerStatementLocalVarLiveness(GenTreePtr startNode,
                 // This ensures that the block->bbVarUse will contain
                 // the FrameRoot local var if is it a tracked variable.
 
-                if (tree->gtCall.IsUnmanaged() || (tree->gtCall.IsTailCall() && info.compCallUnmanaged))
-                {
-                    /* Get the TCB local and mark it as used */
-
-                    noway_assert(info.compLvFrameListRoot < lvaCount);
+                if (info.compLvFrameListRoot < lvaCount)
```
jkotas: The same check in the non-legacy backend looks like this:

```
                if (!opts.ShouldUsePInvokeHelpers())
                {
                    /* Get the TCB local and mark it as used */

                    noway_assert(info.compLvFrameListRoot < lvaCount);
...
```

I would be nice to keep it the same.

jkotas: https://github.com/dotnet/coreclr/blob/master/src/jit/liveness.cpp#L364

jkotas:
LGTM (modulo the one last comment).

src/jit/codegenlegacy.cpp
sergign60: @jkotas I've tried this variant before but got some fails on testing. Ok, I'll try it again.

sergign60:

src/jit/codegenlegacy.cpp
```diff
@@ -19047,16 +19047,30 @@ regMaskTP CodeGen::genCodeForCall(GenTreeCall* call, bool valUsed)
...
+
+                        void* addr = lookup.addr;
+
+                        assert(addr != NULL);
+
+#if defined(_TARGET_ARM_)
+                        // Legacy backend does not handle the `IAT_VALUE` case that does not
+                        // fit. It is not reachable currently from any front end so just check
+                        // for it via noway_assert.
+                        noway_assert(lookup.accessType != IAT_VALUE && compiler->codeGen->validImmForBL((ssize_t)addr));
```
jkotas: I am sorry - it is still not right. It needs to be:
`noway_assert(lookup.accessType != IAT_VALUE || compiler->codeGen->validImmForBL((ssize_t)addr));`

sergign60:

src/jit/codegenlegacy.cpp
```diff
@@ -20853,22 +20868,23 @@ GenTreePtr Compiler::fgLegacyPerStatementLocalVarLiveness(GenTreePtr startNode,
                 // This ensures that the block->bbVarUse will contain
                 // the FrameRoot local var if is it a tracked variable.
 
-                if (tree->gtCall.IsUnmanaged() || (tree->gtCall.IsTailCall() && info.compCallUnmanaged))
-                {
-                    /* Get the TCB local and mark it as used */
+                if (!opts.ShouldUsePInvokeHelpers())
```
jkotas: Nit: It would be nice to have `{` / `}` around this. (I am surprised that the formatter does not complain about this.)

sergign60:
@jkotas Thanks for your help!

src/jit/codegenlegacy.cpp
```diff
@@ -19047,16 +19047,30 @@ regMaskTP CodeGen::genCodeForCall(GenTreeCall* call, bool valUsed)
...
+                        void* addr = lookup.addr;
+
+                        assert(addr != NULL);
+
+#if defined(_TARGET_ARM_)
+                        // Legacy backend does not handle the `IAT_VALUE` case that does not
+                        // fit. It is not reachable currently from any front end so just check
+                        // for it via noway_assert.
+                        noway_assert(lookup.accessType != IAT_VALUE || compiler->codeGen->validImmForBL((ssize_t)addr));
+#endif
```
BruceForstall: Change `compiler->codeGen->validImmForBL((ssize_t)addr)` to `arm_Valid_Imm_For_BL((ssize_t)addr)`

BruceForstall: Also, I think this should just be a normal assert, unless you can tell me that a MinOpts compile will never hit this.

---
In reply to: [127329034](https://github.com/dotnet/coreclr/pull/12550#discussion_r127329034) [](ancestors = 127329034)

src/jit/codegenlegacy.cpp
```diff
@@ -19047,16 +19047,30 @@ regMaskTP CodeGen::genCodeForCall(GenTreeCall* call, bool valUsed)
...
                         {
 #if CPU_LOAD_STORE_ARCH
                             // Load the address into a register, indirect it and call  through a register
                             indCallReg = regSet.rsGrabReg(RBM_ALLINT); // Grab an available register to use for the CALL
                                                                        // indirection
                             instGen_Set_Reg_To_Imm(EA_HANDLE_CNS_RELOC, indCallReg, (ssize_t)addr);
-                            getEmitter()->emitIns_R_R_I(INS_ldr, EA_PTRSIZE, indCallReg, indCallReg, 0);
+
+                            if (lookup.accessType == IAT_PVALUE)
+                                getEmitter()->emitIns_R_R_I(INS_ldr, EA_PTRSIZE, indCallReg, indCallReg, 0);
```
BruceForstall: coding convention: add { } around `if` block (the `emitIns_R_R_I` line)

sergign60:
@BruceForstall  please review

sergign60:

sergign60:

sergign60:

sergign60:

Issue #12477, opened 2017-06-27
[RyuJIT] Recognize and optimize constant set membership tests


mikedn:
It would be useful if the JIT recognized tests like:
* `if ('0' <= c && c <= '9')` - can be reduced to a single branch + compare + subtract
* `if (c == ' ' || c == '\t' || c == '\r' || c == '\n')` - can be reduced to 2 branches + compare + bit test (BT instruction)

The second variant is sometimes implemented using a switch/jump table (in which case is much more simpler to recognize):
```C#
switch (c) {
case ' ':
case '\t':
case '\r':
case '\n':
    return true;
default:
    return false;
}
```

The reduction in number of conditional branches to 1 or 2 can make other optimizations more effective - if-conversion in particular.

Similar issue in the Roslyn repository: https://github.com/dotnet/roslyn/issues/20375

The first variant can be implemented by managed compilers but getting good code for the second is problematic if the target arch bitness is not known. Also, the BT instruction is not available in IL so it has to be simulated with shift/and/compare which the JIT has to recognize.

mikedn:
An experimental implementation shows a ~2700 bytes improvement for the first variant and a ~170 bytes improvement for the second variant. A proper implementation may do better but I don't expect it to be significantly better.

Example codegen for the first variant:
```asm
; before
       cmp      eax, 16
       jl       SHORT G_M38823_IG03
       cmp      eax, 18
       jle      SHORT G_M38823_IG05
; after
       lea      edx, [rax-16]
       cmp      edx, 2
       jbe      SHORT G_M38823_IG05
```
Example codegen for the second variant:
```asm
; before
       cmp      ecx, 0x1F41
       je       G_M36321_IG13
       cmp      ecx, 0x1F42
       je       G_M36321_IG13
       cmp      ecx, 0x1F49
       je       G_M36321_IG13
; after
       mov      eax, ecx
       sub      eax, 0x1F41
       cmp      eax, 8
       ja       G_M36321_IG05
       mov      r9d, 259
       bt       r9, rcx
       jb       G_M36321_IG14
```

BruceForstall:
@dotnet/jit-contrib

mikedn:
Switch to BT conversion yields a ~3700 bytes improvement. Example codegen:
```asm
; before
       cmp      eax, 3
       ja       SHORT G_M9431_IG07
       mov      eax, eax
       lea      rdx, [reloc @RWD00]
[...]

PR #12552, opened 2017-06-29
Use BT in switch lowering

mikedn:
Contributes to #12477

### DIFF
--- a/src/jit/lower.cpp
+++ b/src/jit/lower.cpp
@@ -721,39 +721,40 @@ GenTree* Lowering::LowerSwitch(GenTree* node)
     }
     else
     {
-        // Lower the switch into an indirect branch using a jump table:
-        //
-        // 1. Create the constant for the default case
-        // 2. Generate a GT_GE condition to compare to the default case
-        // 3. Generate a GT_JTRUE to jump.
-        // 4. Load the jump table address into a local (presumably the just
-        //    created constant for GT_SWITCH).
-        // 5. Create a new node for the lowered switch, this will both generate
-        //    the branch table and also will be responsible for the indirect
-        //    branch.
-
-        JITDUMP("Lowering switch BB%02u: using jump table expansion\n", originalSwitchBB->bbNum);
-
-        GenTree* switchValue = comp->gtNewLclvNode(tempLclNum, tempLclType);
-#ifdef _TARGET_64BIT_
-        if (tempLclType != TYP_I_IMPL)
-        {
-            // Note that the switch value is unsigned so the cast should be unsigned as well.
-            switchValue = comp->gtNewCastNode(TYP_I_IMPL, switchValue, TYP_U_IMPL);
-            switchValue->gtFlags |= GTF_UNSIGNED;
-        }
-#endif
-        GenTreePtr gtTableSwitch =
-            comp->gtNewOperNode(GT_SWITCH_TABLE, TYP_VOID, switchValue, comp->gtNewJmpTableNode());
-        /* Increment the lvRefCnt and lvRefCntWtd for temp */
+        // At this point the default case has already been handled and we need to generate a jump
+        // table based switch or a bit test based switch at the end of afterDefaultCondBlock. Both
+        // switch variants need the switch value so create the necessary LclVar node here.
+        GenTree*    switchValue      = comp->gtNewLclvNode(tempLclNum, tempLclType);
+        LIR::Range& switchBlockRange = LIR::AsRange(afterDefaultCondBlock);
         tempVarDsc->incRefCnts(blockWeight, comp);
+        switchBlockRange.InsertAtEnd(switchValue);
 
-        // this block no longer branches to the default block
-        afterDefaultCondBlock->bbJumpSwt->removeDefault();
+        // Try generating a bit test based switch first,
+        // if that's not possible a jump table based switch will be generated.
+        if (!TryLowerSwitchToBitTest(jumpTab, jumpCnt, targetCnt, afterDefaultCondBlock, switchValue))
+        {
+            JITDUMP("Lowering switch BB%02u: using jump table expansion\n", originalSwitchBB->bbNum);
+
+#ifdef _TARGET_64BIT_
+            if (tempLclType != TYP_I_IMPL)
+            {
+                // SWITCH_TABLE expects the switch value (the index into the jump table) to be TYP_I_IMPL.
+                // Note that the switch value is unsigned so the cast should be unsigned as well.
+                switchValue = comp->gtNewCastNode(TYP_I_IMPL, switchValue, TYP_U_IMPL);
+                switchValue->gtFlags |= GTF_UNSIGNED;
+                switchBlockRange.InsertAtEnd(switchValue);
+            }
+#endif
+
+            GenTree* switchTable = comp->gtNewJmpTableNode();
+            GenTree* switchJump  = comp->gtNewOperNode(GT_SWITCH_TABLE, TYP_VOID, switchValue, switchTable);
+            switchBlockRange.InsertAfter(switchValue, switchTable, switchJump);
+
+            // this block no longer branches to the default block
+            afterDefaultCondBlock->bbJumpSwt->removeDefault();
+        }
+
         comp->fgInvalidateSwitchDescMapEntry(afterDefaultCondBlock);
-
-        LIR::Range& afterDefaultCondBBRange = LIR::AsRange(afterDefaultCondBlock);
-        afterDefaultCondBBRange.InsertAtEnd(LIR::SeqTree(comp, gtTableSwitch));
     }
 
     GenTree* next = node->gtNext;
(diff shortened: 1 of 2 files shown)


mikedn:
jit-diff summary:
```
Total bytes of diff: -3712 (-0.03% of base)
    diff is an improvement.
       -1403 : Microsoft.CodeAnalysis.CSharp.dasm (-0.07% of base)
       -1041 : Microsoft.CodeAnalysis.VisualBasic.dasm (-0.05% of base)
        -326 : Microsoft.CodeAnalysis.dasm (-0.04% of base)
        -257 : System.Linq.Expressions.dasm (-0.04% of base)
        -150 : System.Private.CoreLib.dasm (0.00% of base)
23 total files with size differences (23 improved, 0 regressed), 56 unchanged.
           2 : Microsoft.CodeAnalysis.VisualBasic.dasm - VBSemanticModel:GetSemanticSymbols(struct,ref,int,byref,byref):struct:this
           1 : Microsoft.CodeAnalysis.VisualBasic.dasm - Scanner:ScanXmlElementInXmlDoc(int):ref:this
         -66 : Microsoft.CodeAnalysis.CSharp.dasm - ErrorFacts:IsWarning(int):bool
         -56 : Microsoft.CodeAnalysis.CSharp.dasm - LocalRewriter:GetIntPtrConversionMethod(ref,ref):int
         -52 : Microsoft.CodeAnalysis.VisualBasic.dasm - ErrorFacts:IsWarning(int):bool
         -44 : System.Linq.Expressions.dasm - TypeUtils:IsImplicitNumericConversion(ref,ref):bool
         -39 : Microsoft.CodeAnalysis.CSharp.dasm - MessageIDExtensions:RequiredVersion(int):int
264 total methods with size differences (262 improved, 2 regressed), 65297 unchanged.
```

Sample diff:
```asm
; before
       cmp      eax, 3
       ja       SHORT G_M9431_IG07
       mov      eax, eax
       lea      rdx, [reloc @RWD00]
       mov      edx, dword ptr [rdx+4*rax]
       lea      rcx, G_M9431_IG02
       add      rdx, rcx
       jmp      rdx
; after
       cmp      eax, 3
       ja       SHORT G_M9431_IG07
       mov      edx, 3
       bt       edx, eax
       jb       SHORT G_M9431_IG05
```

mikedn:
@dotnet/jit-contrib Do you want this? It's perhaps a bit more "exotic" than other stuff I did so I thought I'd ask before finishing it.

pgavlin:
cc @CarolEidt @BruceForstall @russellhadley 

Just a heads up that we might be a bit slow to respond over the next couple days: a good chunk of the team is out at the moment due to Independence Day.

pgavlin:
@mikedn am I correct that this change currently only generates `GT_BT` (despite supporting other variants in the code generator and emitter)?

mikedn:
> am I correct that this change currently only generates GT_BT (despite supporting other variants in the code generator and emitter).

Yes, this only really needs `GT_BT`, the rest are used in other experiments I've done. Maybe we should keep just `GT_BT` for now (though I'd keep the emitter changes as they are, modifying the emitter turned out to be nightmarish and probably took most of the time).

mikedn:
This might need some tweaks to improve fg layout, in some cases it's pretty lame:
```asm
       cmp      ebp, 5
       ja       SHORT G_M3761_IG04
       mov      eax, 39
       bt       eax, ebp
       jb       SHORT G_M3761_IG05
       jmp      SHORT G_M3761_IG07 ; should not be needed
G_M3761_IG04:
       cmp      eax, 0xCEC8
       je       SHORT G_M3761_IG06
       jmp      SHORT G_M3761_IG07
G_M3761_IG05:
       mov      ebx, 3
       jmp      SHORT G_M3761_IG07
G_M3761_IG06:
       mov      ebx, 1
G_M3761_IG07:
```

JosephTremoulet:
> @dotnet/jit-contrib Do you want this?

Yes.

4creators:
Bit test instructions on all platforms belong to the group of problematic opcodes from performance perspective, however, they allow for size optimized codes. My worry is that this optimization may regress performance. 

For more info on assembly optimizations one can check [Optimizing subroutines in assembly language: An optimization guide for x86 platforms by Agner Fog - latest version from 2017 May 02](http://www.agner.org/optimize/optimizing_assembly.pdf) where bit test instructions are desribed in `Chapter 16 Problematic Instructions` as follows:

```
16.6 Bit test (all processors)

BT, BTC, BTR, and BTS instructions should preferably be replaced by instructions like TEST, AND, 
OR, XOR, or shifts on older processors. Bit tests with a memory operand should be avoided 
on Intel processors. BTC, BTR, and BTS use 2 μops on AMD processors. Bit test instructions are 
useful when optimizing for size.
```

Is it possible to verify performance impact of proposed optimizations?

4creators:
@mikedn This does not hold true for newest AMD processors - Ryzen, Epyc and future relases based on the same microarchitecture. This implementation uses BTC, BTR and BTS instructions for which experimentally determined parameters for Ryzen silicon are as follows:
```
| Instruction   | Operands | Ops | Latency | Reciprocal throughput |
|---------------|----------|-----|---------|-----------------------|
| BTC, BTR, BTS |   r,r/i  |  2  |    2    |          0.5          |
| BTC, BTR, BTS |    m,i   |  4  |         |           2           |
| BTC, BTR, BTS |    m,r   |  8  |         |           3           |
```
Source: [Instruction tables: Lists of instruction latencies, throughputs and micro-operation breakdowns for Intel, AMD and VIA CPUs, Anders Fog, 2017-05-02]

For this case important is only first row which indicates that processor throughput would be 2 instructions per clock cycle, uOps 2, latency 2, what is half of the value one could achieve with this silicon. I have no experimental data for EPYC processors and currently no access to silicon to test them but I would bet the instruction performance would be identical. Since some of this optimizations will pop in hot paths this would have negative impact on performance having negative impact even on Azure (MSFT plans to deploy EPYC)..

mikedn:
> This implementation uses BTC, BTR and BTS instructions

This implementation uses BT.

karelz:
@mikedn what is status of this PR? What is the next step? Code review? More work on the PR?
There hasn't been any update for 1.5 months ...

mikedn:
> There hasn't been any update for 1.5 months ...

Actually I rebased and fixed bugs earlier today :smile:. I'm stuck trying to finish another PR, I'll get back to this soon enough.

karelz:
Perfect, happy to hear it is active :)

mikedn:
Some perf numbers (from a Core i5-4400 Haswell) for:
```C#
for (int i = 0; i < iter; i++)
{
    switch (x)
    {
    case 0:
    case 1:
    case 3:
    case 4:
    case 6:
        s += 42;
        break;
    default:
        s += 44;
        break;
    }
}
```
where `x` is 1, 2, 1000 or a random sequence of 0 and 2.

Jump table (original) version:

Method |      Mean |     Error |    StdDev |
----------- |-----------|-----------|-----------|
Const_1 | 127.60 us | 0.8110 us | 0.7190 us |
Const_2 |  92.00 us | 0.0822 us | 0.0769 us |
Const_1000 |  61.43 us | 0.1166 us | 0.1090 us |
Random_0_2 | 601.31 us | 0.9332 us | 0.8272 us |

BT version:

Method |      Mean |     Error |    StdDev |
----------- |-----------|-----------|-----------|
Const_1 |  78.71 us | 1.3642 us | 1.2761 us |
Const_2 |  61.37 us | 0.1202 us | 0.1066 us |
Const_1000 |  61.38 us | 0.1077 us | 0.1007 us |
Random_0_2 | 484.33 us | 0.5566 us | 0.5206 us |

mikedn:
It seems that there's not a lot that can be done about the cases where a `jmp` is generated after the `bt`/`jc` sequence. In many cases the `jmp` cannot be avoided at all, at best it may be moved to a different path. In some cases it should be possible to move the `jmp` target block to avoid the need for `jmp` but this turns out to be somewhat problematic to do in lowering, you either risk skip lowering a block or lower a block twice.

Of course, it is also possible to disable this optimization when a `jmp` is needed. However, in an example I looked at it turns out that even with `jmp` in there the code is still smaller and sometimes faster than the original code.
```C#
int x = 0;
int y = 0;
while (x++ < iter) {
    switch (E) {
    case 9:
    case 10:
    case 13:
    case 32:
        continue;
    case 11:
    case 12:
        break;
    default:
        break;
    }
    y++;
}
return y;
```

```asm
; before
       418BD0               mov      edx, r8d
       4C8D053C000000       lea      r8, [reloc @RWD00]
       458B0490             mov      r8d, dword ptr [r8+4*rdx]
       4C8D0DB5FFFFFF       lea      r9, G_M1752_IG02
       4D03C1               add      r8, r9
       41FFE0               jmp      r8
; after
       BA13000000           mov      edx, 19
       440FA3C2             bt       edx, r8d
       8BD1                 mov      edx, ecx
       720D                 jb       SHORT G_M1752_IG06
; note that this happens to be a rather unfortunate example as this mov
; is inserted by LSRA, in other cases there's nothing between jb and jmp
       8BCA                 mov      ecx, edx
       EB05                 jmp      SHORT G_M1752_IG05
```
When E is always 11 (so the `jmp` path is always taken) these version have the same performance. When E varies such that the `jmp` path is randomly taken the BT version is faster.

mikedn:
Given the above I think this is ready for review. But it has 2 commits copied from #13626 so that one should be merged first so I can rebase this one.

mikedn:
Ubuntu x64 Checked Build and Test failed to due to some filesystem issue:
```
java.io.IOException: remote file operation failed: /mnt/j/workspace/dotnet_coreclr/master/checked_ubuntu_prtest at hudson.remoting.Channel@896f472:ubuntu1404-20170821-45c232: hudson.util.IOException2: There are some problems during the conversion into JUnit reports: 
21:21:32 	at hudson.FilePath.act(FilePath.java:1014)
21:21:32 	at hudson.FilePath.act(FilePath.java:996)
21:21:32 	at org.jenkinsci.plugins.xunit.XUnitProcessor.performTests(XUnitProcessor.java:145)
```

mikedn:
x86 diff shows that the added `jmp` can sometimes have a negative impact on register allocation:
```asm
       mov      ecx, 51
       bt       ecx, eax
       jb       SHORT G_M65019_IG03
       movzx    esi, byte  ptr [ebp+08H] ; LSRA inserted stuff
       mov      edi, gword ptr [ebp-10H]
       mov      gword ptr [ebp-10H], edi
       mov      dword ptr [ebp+08H], esi
       jmp      SHORT G_M65019_IG08
```
I don't know what causes this yet so I think I'll stop handling this case for now to get this in. The optimization is still reasonably useful even without this case.

src/jit/lower.cpp
```diff
@@ -710,39 +710,35 @@ GenTree* Lowering::LowerSwitch(GenTree* node)
...
-        // 5. Create a new node for the lowered switch, this will both generate
-        //    the branch table and also will be responsible for the indirect
-        //    branch.
-
-        JITDUMP("Lowering switch BB%02u: using jump table expansion\n", originalSwitchBB->bbNum);
-
-        GenTree* switchValue = comp->gtNewLclvNode(tempLclNum, tempLclType);
-#ifdef _TARGET_64BIT_
-        if (tempLclType != TYP_I_IMPL)
+        GenTree*    switchValue      = comp->gtNewLclvNode(tempLclNum, tempLclType);
```
CarolEidt: Since the previous if/clauses are rather large, it would be good to add a comment here about what is going to be done.

src/jit/lower.cpp
```diff
@@ -754,6 +750,154 @@ GenTree* Lowering::LowerSwitch(GenTree* node)
...
+            bitTable |= (size_t(1) << bitIndex);
+        }
+        else if (bbCase0 == nullptr)
+        {
+            bbCase0 = jumpTable[bitIndex];
+        }
+        else if (jumpTable[bitIndex] != bbCase0)
+        {
+            // The jump table contains more than 2 targets.
+            return false;
```
CarolEidt: The targetCount was checked above; does this case happen if one of the non-default cases has the same target as the default? If so, it would be good to state that here; otherwise it might seem like this should be an assert.

src/jit/lower.cpp
```diff
@@ -710,39 +710,35 @@ GenTree* Lowering::LowerSwitch(GenTree* node)
...
-
-        GenTree* switchValue = comp->gtNewLclvNode(tempLclNum, tempLclType);
-#ifdef _TARGET_64BIT_
-        if (tempLclType != TYP_I_IMPL)
+        GenTree*    switchValue      = comp->gtNewLclvNode(tempLclNum, tempLclType);
+        LIR::Range& switchBlockRange = LIR::AsRange(afterDefaultCondBlock);
+        tempVarDsc->incRefCnts(blockWeight, comp);
+        switchBlockRange.InsertAtEnd(switchValue);
+
+        if (!TryLowerSwitchToBitTest(jumpTab, jumpCnt, targetCnt, afterDefaultCondBlock, switchValue))
```
briansull: Add a comment explaining that `TryLowerSwitchToBitTest ` will try to change the switch into a BitTest

If it cannot we have to lower the switch here

briansull: Eventually we should refactor this so that we have three methods:

TryLowerSwitchToCompareAndBranch
TryLowerSwitchToBitTest
LowerSwitchToJumpTable

BruceForstall:
@dotnet-bot test this please

BruceForstall:

src/jit/lower.cpp
mikedn: > The targetCount was checked above; does this case happen if one of the non-default cases has the same target as the default?

Yes. The targetCount check above is an optimization, it could be removed but then we'd scan more jump tables for no reason.

mikedn:
Rebased and added/updated comments.

Failures in CentOS and ARM64 builds appear to be unrelated:

https://ci.dot.net/job/dotnet_coreclr/job/master/job/checked_centos7.1_tst_prtest/829/testReport/JIT.Performance.CodeQuality.Span/SpanBench/SpanBench/

https://ci.dot.net/job/dotnet_coreclr/job/master/job/arm64_cross_debug_windows_nt_tst_prtest/111/artifact/bin/tests/Windows_NT.arm64.Debug/Smarty.run.0/Smarty.0.fail.smrt/*view*/

As far as I can tell failures reported by previous jitstress builds were caused by unrelated issues that have been fixed. Let's try again:

sandreenko:
@dotnet-bot test Windows_NT amd64 Checked jitstress1
@dotnet-bot test Windows_NT amd64 Checked jitstress2

some jobs were renamed recently.

sandreenko:
@dotnet-bot test Windows_NT x64 Checked jitstress1
@dotnet-bot test Windows_NT x64 Checked jitstress2
some jobs were renamed recently.

PR #12556, opened 2017-06-29
Make libnuma soft dependency

### LINKED ISSUE #12192: Make CoreCLR dependency on libnuma on Unix soft
Currently, when CoreCLR is built on a machine with libnuma dev package installed, it gets explicit dependency on libnuma. So it won't run on a Linux without this library installed. I would like to eliminate this hard dependency and make it soft by loading the libnuma via dlopen if it is present, but working fine without any numa support if the library is not installed.

am11:
Load libnuma dynamically, so it is not linked at compile time and
hence avoid making it a hard dependency at runtime.

Fixes #12192

### DIFF
--- a/src/pal/src/numa/numa.cpp
+++ b/src/pal/src/numa/numa.cpp
@@ -35,6 +35,9 @@ SET_DEFAULT_DEBUG_CHANNEL(NUMA);
 #endif
 
 #include <pthread.h>
+#include <dlfcn.h>
+
+#include "numashim.h"
 
 using namespace CorUnix;
 
@@ -71,6 +74,10 @@ int g_possibleCpuCount = 0;
 int g_groupCount = 0;
 // The highest NUMA node available
 int g_highestNumaNode = 0;
+// Is numa available
+bool g_numaAvailable = false;
+
+void* numaHandle = nullptr;
 
 static const int MaxCpusPerGroup = 8 * sizeof(KAFFINITY);
 static const WORD NO_GROUP = 0xffff;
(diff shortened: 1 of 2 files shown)


src/pal/src/numa/numa.cpp
```diff
@@ -140,71 +142,91 @@ BOOL
...
     {
-        struct bitmask *mask = numa_allocate_cpumask();
-        int numaNodesCount = numa_max_node() + 1;
-
-        g_possibleCpuCount = numa_num_possible_cpus();
-        g_cpuCount = 0;
-        g_groupCount = 0;
+        numaHandle = dlopen("libnuma.so.1", RTLD_LAZY);
+    }
+    else if (numaHandle != 0)
```
janvorli: This "else if" is incorrect, it needs to be just `if`, otherwise the handle created by the second attempt would be ignored.

src/pal/src/numa/numa.cpp
```diff
@@ -140,71 +142,91 @@ BOOL
...
+    {
+#define PER_FUNCTION_BLOCK(fn) \
+    fn##_ptr = (decltype(fn)*)dlsym(numaHandle, fn); \
+    if (fn##_ptr == NULL) { fprintf(stderr, "Cannot get symbol " #fn " from libnuma\n"); abort(); }
+FOR_ALL_NUMA_FUNCTIONS
+#undef PER_FUNCTION_BLOCK
 
-        for (int i = 0; i < numaNodesCount; i++)
+        if ((*numa_available)() != -1)
```
janvorli: This should be just numa_available(), there is no need to dereference. The same for all other functions.

src/pal/src/numa/numashim.h
```diff
@@ -0,0 +1,49 @@
...
+    PER_FUNCTION_BLOCK(mbind_ptr)
+
+// Declare pointers to all the used numa functions
+#define PER_FUNCTION_BLOCK(fn) extern decltype(fn)* fn##_ptr;
+FOR_ALL_NUMA_FUNCTIONS
+#undef PER_FUNCTION_BLOCK
+
+// Redefine all calls to numa functions as calls through pointers that are set
+// to the functions of libnuma in the initialization.
+#define numa_available() numa_available()
```
janvorli: These are incorrect except of the mbind. All of them should have the _ptr suffix like the mbind does.
Moreover, the numa_num_possible_cpus is defined as mbind_ptr.

janvorli (changes requested):

src/pal/src/numa/numa.cpp
am11: @janvorli, thanks for the review. I have made the fixes and re-pushed the changes. Can you please take a look?

janvorli (approved):
LGTM, thank you a lot for making the change!

janvorli:
@jkotas, could you please merge it after the CI tests complete?

src/pal/src/numa/numa.cpp
```diff
@@ -140,71 +147,92 @@ BOOL
...
     {
-        struct bitmask *mask = numa_allocate_cpumask();
-        int numaNodesCount = numa_max_node() + 1;
-
-        g_possibleCpuCount = numa_num_possible_cpus();
-        g_cpuCount = 0;
-        g_groupCount = 0;
-
-        for (int i = 0; i < numaNodesCount; i++)
+        numaHandle = dlopen("libnuma.so.1", RTLD_LAZY);
```
am11, line 153: @janvorli, I was packaging a .NET application in a unikernel (bootable image) isolation, and found that if we want to include libnuma in the published disk image, we need to include one of these "symlinks" in addition to the actual binary: `libnuma.so.1.0.0`. In .NET 6's singlefilehost, libnuma (and liblttng-ust) are not statically compiled (yet, and maybe will not be in the future, since these two are soft dependencies?).

Since their ABI version hasn't changed in years (it is 1.0.0 everywhere I have looked), would adding a probing path with reordering: https://github.com/dotnet/runtime/commit/b60f1fabf1e07655350be4f1f36a4647edd0cbac be an acceptable change?

PR #12557, opened 2017-06-29
Implement a /verbose flag to show the verbose output from crossgen

briansull:
Crossgen generates lots of extra output that is logged using:
` Zapper::Info(LPCWSTR format, ...)`

But it can't be displayed unless m_verbose is true
```
    case CORZAP_LOGLEVEL_INFO:
        if (!m_pOpt->m_verbose)
            return;
        break;
```
This  commit allows us to see this output from crossgen by using a new /verbose option

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -455,6 +455,7 @@ if %__BuildNativeCoreLib% EQU 1 (
         type %__CrossGenCoreLibLog%
         goto CrossgenFailure
     )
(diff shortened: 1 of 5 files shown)


BruceForstall:
Can you add `/verbose` to the help message in `PrintUsageHelper()`?

briansull:
I also had to convert the last place where printf() was directly used in the VM. (in methodtable.cpp)

These printf() get turned into logging info statements and if we actually try to print the contents 
the C++ runtime asserts because you opened (or set) stdout to be in wide-char mode  (crossgen does this) 
and now you are trying to print 8-bit ascii.

PR #12562, opened 2017-06-30
Move JIT_TrialAlloc declaration into jitinterfacex86.cpp

parjong:
It would be better to move JIT_TrialAlloc declaration int jitinterfacex86.cpp as JIT_TrialAlloc is referenced only inside jitinterfacex86.cpp.

### DIFF
--- a/src/vm/i386/jitinterfacex86.cpp
+++ b/src/vm/i386/jitinterfacex86.cpp
@@ -34,6 +34,39 @@
 #define MON_DEBUG 1
 #endif
 
+#ifdef _TARGET_X86_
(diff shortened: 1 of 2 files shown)


src/vm/i386/jitinterfacex86.cpp
```diff
@@ -34,6 +34,39 @@
 #define MON_DEBUG 1
 #endif
 
+#ifdef _TARGET_X86_
```
jkotas: I do not think this ifdef is needed here.  This file is only compiled for x86. You can fix it together with the bug that you are looking at.

Issue #12560, opened 2017-06-30
x86 Server GC PreGrow write barrier does not work when heap is mapped to a low VA


swgillespie:
A bug in the write barrier initialization handshake in the Server GC startup path results in `g_ephemeral_low` never getting bashed into the write barrier on startup on x86, which results in heap corruption when the managed heap ends up in virtual addresses below `0x0F0F0F0F` and card table writes are mistakenly omitted. The managed heap does not usually end up that low in the VA space, but it can when running in nonstandard environments.

This was introduced by https://github.com/dotnet/coreclr/commit/04d6bd105ade5f6189a15a6dbb59b082613429a1 - will have a PR out addressing this soon.

cc @jkotas @Maoni0 @gkhanna79 @sergiy-k

leecow:
@swgillespie can you confirm that everything is covered and this can be closed?

swgillespie:
Yep, with the merge of https://github.com/dotnet/coreclr/pull/12584 this can be closed.

PR #12563, opened 2017-06-30
Fix an issue where g_ephemeral_low would never get bashed into the write...

swgillespie:
...barrier when starting up with Server GC (x86 only).

The problem is triggered by this bit of code:

```c++
void gc_heap::adjust_ephemeral_limits ()
{
    ephemeral_low = generation_allocation_start (generation_of (max_generation - 1));
    ephemeral_high = heap_segment_reserved (ephemeral_heap_segment);

    dprintf (3, ("new ephemeral low: %Ix new ephemeral high: %Ix",
                 (size_t)ephemeral_low, (size_t)ephemeral_high))

#ifndef MULTIPLE_HEAPS
    // This updates the write barrier helpers with the new info.
    stomp_write_barrier_ephemeral(ephemeral_low, ephemeral_high);
#endif // MULTIPLE_HEAPS
}
```

The change that triggered this problem added the `#ifndef MULTIPLE_HEAPS`. This is because, at the time, it was thought that `ephemeral_low` was not being used by any Server GC write barriers; since there is no single ephemeral generation, it does not make sense to check them all. amd64 does not makes use of `g_ephemeral_low` when using server GC at all, but x86 unfortunately does. Since `stomp_write_barrier_ephemeral` will not be called when running with Server GC, the GC will complete its initialization without ever calling `StompWriteBarrierEphemeral`, which is what signals the runtime to bash `g_ephemeral_low` (and possibly also `g_ephemeral_high`, if necessary) into the write barrier.

Previously, the runtime would be bashing `1` as `g_ephemeral_low`, so that all non-null pointers would have their card table bits set in the write barrier. However, since it was never initialized, the write barrier still has the sentinel value `0x0F0F0F0F` as `g_ephemeral_low` and will only set card table bits for pointers over that value. Windows rarely assigns virtual addresses as low as `0x0F0F0F0F` to managed heap segments, so this does not manifest when running on vanilla Windows. However, if the kernel does place managed heap segments below that virtual address, their card bits will never be set and the heap will be corrupted on the next GC.

This PR remedies the startup problem by explicitly calling `StompWriteBarrierEphemeral` from `GCToEEInterface::StompWriteBarrier`, as part of the `WriteBarrierOp
WriteBarrierOp::Initialize` code path, which is called once from `GCHeap::Initialize`. 

Fixes https://github.com/dotnet/coreclr/issues/12560.

### DIFF
--- a/src/vm/gcenv.ee.cpp
+++ b/src/vm/gcenv.ee.cpp
@@ -923,10 +923,13 @@ void GCToEEInterface::StompWriteBarrier(WriteBarrierParameters* args)
         VolatileStore(&g_highest_address, args->highest_address);
         ::StompWriteBarrierResize(true, false);
 
-        // g_ephemeral_low/high aren't needed for the write barrier stomp, but they
-        // are needed in other places.
+        // StompWriteBarrierResize does not necessarily bash g_ephemeral_low
+        // usages, so we must do so here. This is particularly true on x86,
+        // where StompWriteBarrierResize will not bash g_ephemeral_low when
+        // called with the parameters (true, false), as it is above.
         g_ephemeral_low = args->ephemeral_low;
         g_ephemeral_high = args->ephemeral_high;
+        ::StompWriteBarrierEphemeral(true);
         return;
     case WriteBarrierOp::SwitchToWriteWatch:
 #ifdef FEATURE_USE_SOFTWARE_WRITE_WATCH_FOR_GC_HEAP


swgillespie:
cc @Maoni0 @jkotas @sergiy-k @adityamandaleeka PTAL? Does this look like the right fix?

adityamandaleeka:
I glanced at this code and it looks like this will make us call `FlushInstructionCache` twice for that region (once in `StompWriteBarrierResize` and once in `StompWriteBarrierEphemeral`). Perhaps it would be better to reorganize things a bit so that we get the desired update of `g_ephemeral_low` but the flush only happens once?

swgillespie:
It's true, but we were doing this before and unfortunately we may be a little pressed for time for this particular fix. Perhaps that's something we can revisit in a follow-up PR?

jkotas:
This probably the minimal fix, but I agree with @adityamandaleeka that it would be nice to reorganize this:

```
        g_lowest_address = args->lowest_address;
        g_highest_address = args->highest_address;

        g_ephemeral_low = args->ephemeral_low;
        g_ephemeral_high = args->ephemeral_high;

        ::StompWriteBarrierResize(true, false);
        ::StompWriteBarrierEphemeral(true);

        FlushProcessWriteBuffers(); // I do not think that this is needed at all during init
```

jkotas:
> Perhaps that's something we can revisit in a follow-up PR?

Agree.

adityamandaleeka:
@swgillespie Yup, we can do it separately 👍

swgillespie:
Thanks for the quick reviews! Will open a PR against release/2.0.0 shortly.

swgillespie:
Follow-up issue for addressing the refactoring that Aditya and Jan suggested: https://github.com/dotnet/coreclr/issues/12586

Issue #11781, opened 2017-05-22
[RyuJIT/arm32] Assertion failed 'foundStructArg'


mskvortsov:
```
$ COMPlus_AltJit=BringUpTest:Main \
gdb --args clr-debug/corerun \
tests-release/JIT/Directed/TypedReference/TypedReference/TypedReference.exe
...
#9  0xb0c6c1cc in assertAbort (why=0xb0f1dcd5 "foundStructArg", file=0xb0f1cef7 "/home/mskvortsov/git/coreclr/src/jit/morph.cpp", line=4727) at /home/mskvortsov/git/coreclr/src/jit/error.cpp:287
#10 0xb0d4bda6 in Compiler::fgMorphMultiregStructArgs (this=0x93f10, call=0x97bc8) at /home/mskvortsov/git/coreclr/src/jit/morph.cpp:4727
#11 0xb0d45654 in Compiler::fgMorphArgs (this=0x93f10, call=0x97bc8) at /home/mskvortsov/git/coreclr/src/jit/morph.cpp:4423
#12 0xb0d58608 in Compiler::fgMorphCall (this=0x93f10, call=0x97bc8) at /home/mskvortsov/git/coreclr/src/jit/morph.cpp:8362
#13 0xb0d435d0 in Compiler::fgMorphTree (this=0x93f10, tree=0x97bc8, mac=0x0) at /home/mskvortsov/git/coreclr/src/jit/morph.cpp:15005
#14 0xb0d5d64e in Compiler::fgMorphStmts (this=0x93f10, block=0x973c8, mult=0xbeffc337, lnot=0xbeffc336, loadw=0xbeffc335) at /home/mskvortsov/git/coreclr/src/jit/morph.cpp:15804
#15 0xb0d5e00a in Compiler::fgMorphBlocks (this=0x93f10) at /home/mskvortsov/git/coreclr/src/jit/morph.cpp:16089
#16 0xb0d601cc in Compiler::fgMorph (this=0x93f10) at /home/mskvortsov/git/coreclr/src/jit/morph.cpp:17291
```
```
GC/API/GC/Collect_Forced_1
GC/API/GC/Collect_Forced_2
GC/API/GC/Collect_Forced_3
JIT/Directed/TypedReference/TypedReference
JIT/IL_Conformance/Old/Conformance_Base/refs
JIT/Methodical/flowgraph/dev10_bug679008/fgloop
JIT/Methodical/flowgraph/dev10_bug679053/regionLive
[...]

PR #12568, opened 2017-06-30
[RyuJIT/ARM32] Fix using fgMorphMultiregStructArg for mkrefany

hseok-oh:
Block using fgMorphMultiregStructArg function for mkrefany.
Morphing for mkrefany struct is already done in fgMorphArgs function.

Related issue: #11781 
cc/ @dotnet/arm32-contrib


BruceForstall:
cc @dotnet/jit-contrib

Issue #6573, opened 2016-08-02
Intermittent SIGSEGVs in some CoreCLR tests on ARM32 Linux Release


janvorli:
I have noticed SIGSEGV happening in some CI runs in JIT/Regression/CLR-x86-JIT/V1-M11-Beta1/b37131/b37131 (http://dotnet-ci.cloudapp.net/job/dotnet_coreclr/job/master/job/arm_emulator_cross_release_ubuntu_prtest/1033/) and JIT/Regression/CLR-x86-JIT/V1-M11-Beta1/b37131/b37131 (http://dotnet-ci.cloudapp.net/job/dotnet_coreclr/job/master/job/arm_emulator_cross_release_ubuntu_prtest/1040/)

danmosemsft:
Happened again
`https://ci.dot.net/job/dotnet_coreclr/job/master/job/arm_emulator_cross_debug_ubuntu_prtest/1499/consoleFull#2096352037c7af6a29-b465-404a-b249-90a3911e4354`

```
07:54:00 FAILED   - JIT/Regression/CLR-x86-JIT/V1-M11-Beta1/b37131/b37131/b37131.sh
07:54:00                BEGIN EXECUTION
07:54:00                /bindings/tmp/arm32_ci_temp/coreclr/Windows_NT.x64.Debug/Tests/coreoverlay/corerun b37131.exe
07:54:00                qemu: uncaught target signal 4 (Illegal instruction) - core dumped
07:54:00                ./b37131.sh: line 117: 59294 Illegal instruction     (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" b37131.exe $CLRTestExecutionArguments
07:54:00                Expected: 100
07:54:00                Actual: 132
07:54:00                END EXECUTION - FAILED
```

dagood:
Repro in https://github.com/dotnet/coreclr/pull/11786 https://ci.dot.net/job/dotnet_coreclr/job/release_2.0.0/job/armel_cross_release_tizen_prtest/92/
```
FAILED   - JIT/Regression/VS-ia64-JIT/V1.2-M01/b12263/b12263/b12263.sh
               BEGIN EXECUTION
               /home/coreclr/bin/tests/Windows_NT.x64.Release/Tests/coreoverlay/corerun b12263.exe
               qemu: uncaught target signal 4 (Illegal instruction) - core dumped
[...]

Issue #6298, opened 2016-07-15
ARM segmentation fault: JIT/Regression/CLR-x86-JIT/V1-M11-Beta1/b31878/b31878/b31878.sh


swgillespie:
See http://dotnet-ci.cloudapp.net/job/dotnet_coreclr/job/master/job/arm_emulator_cross_debug_ubuntu_prtest/511/console:

```
12:14:41 FAILED   - JIT/Regression/CLR-x86-JIT/V1-M11-Beta1/b31878/b31878/b31878.sh
12:14:41                BEGIN EXECUTION
12:14:41                /home/coreclr/Windows_NT.x64.Debug/Tests/coreoverlay/corerun b31878.exe
12:14:41 
               ./b31878.sh: line 117: 10078 Segmentation fault      (core dumped) $_DebuggerFullPath "$CORE_ROOT/corerun" b31878.exe $CLRTestExecutionArguments
12:14:41                Expected: 100
12:14:41                Actual: 139
12:14:41                END EXECUTION - FAILED
```

RussKeldorph:
@jashook Looks different than #6289.

parjong:
@jashook @RussKeldorph After PR #6021 is merged, there are many regressions in CoreCLR ARM (softp) Release build:

Here is the result from bdfce9ed7fb:
```
=======================
     Test Results
=======================
# CoreCLR Bin Dir  : 
# Tests Discovered : 9870
# Passed           : 8953
# Failed           : 579
# Skipped          : 338
=======================
```

Here is the result from bdfce9ed7fb without PR #6021 (revert PR #6021 manually):
```
=======================
     Test Results
=======================
# CoreCLR Bin Dir  : 
# Tests Discovered : 9870
# Passed           : 8953
# Failed           : 579
# Skipped          : 338
=======================
```

[[bdfce9ed7fb][TC 387d9fc0a Release] [include-aab8856ce03].txt](https://github.com/dotnet/coreclr/files/368532/bdfce9ed7fb.TC.387d9fc0a.Release.include-aab8856ce03.txt)
[[bdfce9ed7fb][TC 387d9fc0a Release] [revert-aab8856ce03].txt](https://github.com/dotnet/coreclr/files/368533/bdfce9ed7fb.TC.387d9fc0a.Release.revert-aab8856ce03.txt)

Interestingly, b31878 is passed on my side.

swgillespie:
[...]

PR #12572, opened 2017-06-30
[RyuJIT/armel] Passing stack double arguments

### LINKED ISSUE #11927: [RyuJIT/armel] CodeGenBringUpTests Assertion failed 'info->numSlots == 1'
A sub-issue for #11822 

Some tests fail with below assert failure.

```
               Assert failure(PID 31095 [0x00007977], Thread: 31095 [0x7977]): Assertion failed 'info->numSlots == 1' in 'BringUpTest:StaticCalls():int' (IL size 688)

                   File: /opt/code/github/hqueue/coreclr/src/jit/lsraarmarch.cpp Line: 681
```

List of the tests fail with above message in `JIT/CodeGenBringUpTests`

```
FAILED   - [  25]JIT/CodeGenBringUpTests/DblArea/DblArea.sh
FAILED   - [  28]JIT/CodeGenBringUpTests/DblAvg6/DblAvg6.sh
FAILED   - [  30]JIT/CodeGenBringUpTests/DblCall2/DblCall2.sh
FAILED   - [  31]JIT/CodeGenBringUpTests/DblDist/DblDist.sh
FAILED   - [  34]JIT/CodeGenBringUpTests/DblFillArray/DblFillArray.sh
FAILED   - [  39]JIT/CodeGenBringUpTests/DblRoots/DblRoots.sh
FAILED   - [  81]JIT/CodeGenBringUpTests/InstanceCalls/InstanceCalls.sh
FAILED   - [ 128]JIT/CodeGenBringUpTests/OpMembersOfStructLocal/OpMembersOfStructLocal.sh
FAILED   - [ 135]JIT/CodeGenBringUpTests/StaticCalls/StaticCalls.sh
```

wateret:
Fix #11927

### DIFF
--- a/src/jit/codegenarmarch.cpp
+++ b/src/jit/codegenarmarch.cpp
@@ -555,6 +555,16 @@ void CodeGen::genPutArgStk(GenTreePutArgStk* treeNode)
         {
             genConsumeReg(source);
             emit->emitIns_S_R(storeIns, storeAttr, source->gtRegNum, varNumOut, argOffsetOut);
+            if (compiler->opts.compUseSoftFP && targetType == TYP_LONG)
+            {
+                // This case currently only occurs for double types that are passed as TYP_LONG;
+                // actual long types would have been decomposed by now.
+                assert(source->IsCopyOrReload());
+                regNumber otherReg = (regNumber)source->AsCopyOrReload()->GetRegNumByIdx(1);
+                assert(otherReg != REG_NA);
+                argOffsetOut += EA_4BYTE;
+                emit->emitIns_S_R(storeIns, storeAttr, otherReg, varNumOut, argOffsetOut);
+            }
         }
         argOffsetOut += EA_SIZE_IN_BYTES(storeAttr);
         assert(argOffsetOut <= argOffsetMax); // We can't write beyound the outgoing area area
--- a/src/jit/lsraarmarch.cpp
+++ b/src/jit/lsraarmarch.cpp
@@ -738,8 +738,22 @@ void Lowering::TreeNodeInfoInitPutArgStk(GenTreePutArgStk* argNode, fgArgTabEntr
     else
     {
 #ifdef _TARGET_ARM_
+
+#ifndef ARM_SOFTFP
         // We must not have a multi-reg struct; double uses 2 slots and isn't a multi-reg struct
         assert((info->numSlots == 1) || ((info->numSlots == 2) && (putArgChild->TypeGet() == TYP_DOUBLE)));
+#else  // ARM_SOFTFP
+        // The `double` types have been transformed to `long` on armel.
+        if ((info->numSlots == 2) && (putArgChild->TypeGet() == TYP_LONG))
+        {
+            argNode->gtLsraInfo.srcCount = 2;
+        }
+        else
+        {
+            assert(info->numSlots == 1);
+        }
+#endif // ARM_SOFTFP
+
 #else  // !_TARGET_ARM_
         // We must not have a multi-reg struct
         assert(info->numSlots == 1);
--- a/src/jit/lsraarmarch.cpp
+++ b/src/jit/lsraarmarch.cpp
@@ -739,20 +739,19 @@ void Lowering::TreeNodeInfoInitPutArgStk(GenTreePutArgStk* argNode, fgArgTabEntr
     {
 #ifdef _TARGET_ARM_
 
-#ifndef ARM_SOFTFP
-        // We must not have a multi-reg struct; double uses 2 slots and isn't a multi-reg struct
-        assert((info->numSlots == 1) || ((info->numSlots == 2) && (putArgChild->TypeGet() == TYP_DOUBLE)));
-#else  // ARM_SOFTFP
+#ifdef ARM_SOFTFP
         // The `double` types have been transformed to `long` on armel.
-        if ((info->numSlots == 2) && (putArgChild->TypeGet() == TYP_LONG))
+        const bool isDouble = (info->numSlots == 2) && (putArgChild->TypeGet() == TYP_LONG);
+        if (isDouble)
         {
             argNode->gtLsraInfo.srcCount = 2;
         }
-        else
-        {
-            assert(info->numSlots == 1);
-        }
-#endif // ARM_SOFTFP
+#else  // !ARM_SOFTFP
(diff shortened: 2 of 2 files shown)


BruceForstall:

src/jit/codegenarmarch.cpp
```diff
@@ -555,6 +555,16 @@ void CodeGen::genPutArgStk(GenTreePutArgStk* treeNode)
         {
             genConsumeReg(source);
             emit->emitIns_S_R(storeIns, storeAttr, source->gtRegNum, varNumOut, argOffsetOut);
+            if (compiler->opts.compUseSoftFP && targetType == TYP_LONG)
+            {
+                // This case currently only occurs for double types that are passed as TYP_LONG;
+                // actual long types would have been decomposed by now.
+                assert(source->IsCopyOrReload());
+                regNumber otherReg = (regNumber)source->AsCopyOrReload()->gtOtherRegs[0];
+                assert(otherReg != REG_NA);
```
BruceForstall: Instead, can you use:
```
                regNumber otherReg = source->AsCopyOrReload()->GetRegNumByIdx(1);
```
? Seems better to not touch gtOtherRegs directly.

wateret: Thank you. I'll fix it.

wateret:
@dotnet-bot Test Windows_NT x64 Debug Build and Test
@dotnet-bot Ubuntu x64 Checked Build and Test
@dotnet-bot OSX10.12 x64 Checked Build and Test
@dotnet-bot FreeBSD x64 Checked Build

wateret:
@dotnet-bot Test this please

src/jit/codegenarmarch.cpp
```diff
@@ -555,6 +555,16 @@ void CodeGen::genPutArgStk(GenTreePutArgStk* treeNode)
         {
             genConsumeReg(source);
             emit->emitIns_S_R(storeIns, storeAttr, source->gtRegNum, varNumOut, argOffsetOut);
+            if (compiler->opts.compUseSoftFP && targetType == TYP_LONG)
```
CarolEidt: Shouldn't this be under `#ifdef ARM_SOFTFP`?

src/jit/lsraarmarch.cpp
```diff
@@ -738,8 +738,22 @@ void Lowering::TreeNodeInfoInitPutArgStk(GenTreePutArgStk* argNode, fgArgTabEntr
...
+        if ((info->numSlots == 2) && (putArgChild->TypeGet() == TYP_LONG))
+        {
+            argNode->gtLsraInfo.srcCount = 2;
+        }
+        else
+        {
+            assert(info->numSlots == 1);
+        }
+#endif // ARM_SOFTFP
+
```
CarolEidt: As a rule, when you have both an if and else clause, you should refactor to avoid the double negative (i.e. in this case put the `ARM_SOFTFP` code first). However, the above could be greatly simplified as:
```
#ifdef ARM_SOFTFP
        // The `double` types have been transformed to `long` on armel.
        var_types doubleType = TYP_LONG;
#else
        var_types doubleType = TYP_DOUBLE:
#endif
        assert((info->numSlots == 1) || ((info->numSlots == 2) && (putArgChild->TypeGet() == doubleType)));
```

src/jit/codegenarmarch.cpp
wateret: `compUseSoftFP` was introduced when we first ported armel. The value is TRUE when `ARM_SOFTFP` is defined, FALSE otherwise(for all architectures). Since then I use `compUseSoftFP` if applicable. However some modules like LSRA which do not hold `compiler` object I just use `#ifdef ARM_SOFTFP`.

src/jit/lsraarmarch.cpp
wateret: @CarolEidt I first tried something like your code, but updating srcCount was needed for ARM_SOFTFP. So here's what I've got. Would this look better than the current one? Or would there be a better implementation?

```cpp
#ifdef _TARGET_ARM_

#ifdef ARM_SOFTFP
        // The `double` types have been transformed to `long` on armel.
        var_types doubleType = TYP_LONG;
#else
        var_types doubleType = TYP_DOUBLE:
#endif

        // We must not have a multi-reg struct; double uses 2 slots and isn't a multi-reg struct
        const bool isDouble = (info->numSlots == 2) && (putArgChild->TypeGet() == doubleType);
        assert((info->numSlots == 1) || isDouble);

#ifdef ARM_SOFTFP
        if (isDouble)
        {
            argNode->gtLsraInfo.srcCount = 2;
        }
#endif

#else  // !_TARGET_ARM_
        // We must not have a multi-reg struct
        assert(info->numSlots == 1);
#endif // !_TARGET_ARM_
```

CarolEidt: > updating srcCount was needed for ARM_SOFTFP

I see; in that case, it's probably better to set `isDouble` conditionally instead, and move the srcCount update into the same `#ifdef`. How about this:
```
#ifdef _TARGET_ARM_

#ifdef ARM_SOFTFP
        // The `double` types have been transformed to `long` on armel.
        const bool isDouble = (info->numSlots == 2) && (putArgChild->TypeGet() == TYP_LONG);
        if (isDouble)
        {
            argNode->gtLsraInfo.srcCount = 2;
        }
#else
        const bool isDouble = (info->numSlots == 2) && (putArgChild->TypeGet() == TYP_DOUBLE);
#endif

        // We must not have a multi-reg struct; double uses 2 slots and isn't a multi-reg struct
        assert((info->numSlots == 1) || isDouble);

#else  // !_TARGET_ARM_
        // We must not have a multi-reg struct
        assert(info->numSlots == 1);
#endif // !_TARGET_ARM_
```

wateret: @CarolEidt Thanks! That looks much better.

src/jit/codegenarmarch.cpp
CarolEidt: >  However some modules like LSRA which do not hold compiler object I just use #ifdef ARM_SOFTFP.

`LinearScan` has a `compiler` field, for future reference.

wateret:

wateret:
@CarolEidt arm/armel test fails intermittently, and it seems like it is qemu related issue.
Related Issues : #6573 #6298

PR #12573, opened 2017-06-30
Update gtDispRegVal() for multi register nodes

wateret:


wateret:
```
                                                              /--*  t1     double
Generating: N007 (???,???) [000035] -------------       t35 = *  COPY      long   REG r0
                                                              /--*  t35    long
Generating: N009 (???,???) [000036] -------------       t36 = *  PUTARG_REG long   REG r0
IN0005:             vmov.d2i r0, r1, d8
Generating: N011 (  1,  1) [000002] -------------        t2 =    CNS_DBL   double 1.0000000000000000 REG f8 $41
```

On armel, when we copy or put arg it only showed the first register, but it now shows all the registers assigned for the nodes.


```
                                                              /--*  t1     double
Generating: N007 (???,???) [000035] -------------       t35 = *  COPY      long   REG r0,r1
                                                              /--*  t35    long
Generating: N009 (???,???) [000036] -------------       t36 = *  PUTARG_REG long   REG r0,r1
IN0005:             vmov.d2i r0, r1, d8
Generating: N011 (  1,  1) [000002] -------------        t2 =    CNS_DBL   double 1.0000000000000000 REG f8 $41
```

BruceForstall:

src/jit/gentree.cpp
```diff
@@ -9943,6 +9943,28 @@ void Compiler::gtDispRegVal(GenTree* tree)
...
+            }
+            printf(",%s", compRegVarName(reg));
+        }
+    }
+#endif
+
+#if !defined(LEGACY_BACKEND) && defined(_TARGET_ARM_)
+    if (tree->OperGet() == GT_PUTARG_REG)
+    {
+        printf(",%s", compRegVarName((regNumber)tree->AsMultiRegOp()->gtOtherReg));
```
BruceForstall: When is gtOtherReg valid? It looks like it is only valid with `GT_PUTARG_REG` when `opts.compUseSoftFP` is true. It seems like you could also display it for `GT_MUL_LONG` for arm.

Maybe there needs to be a:
```
bool OperIsMultiReg()
{
  return (gtOper == GT_MUL_LONG) || ((gtOper == GT_PUTARG_REG) && opts.compUseSoftFP);
}
```

src/jit/gentree.cpp
```diff
@@ -9943,6 +9943,28 @@ void Compiler::gtDispRegVal(GenTree* tree)
...
+        {
+            regNumber reg = (regNumber)tree->AsCopyOrReload()->gtOtherRegs[i];
+            if (reg == REG_NA)
+            {
+                break;
+            }
+            printf(",%s", compRegVarName(reg));
+        }
+    }
+#endif
```
BruceForstall: Can you explain the scenario where this code kicks in but the code above under `tree->IsCopyOrReloadOfMultiRegCall()` does not? Is that scenario ARM specific?

src/jit/gentree.cpp
wateret: You're right. I'll fix it.

src/jit/gentree.cpp
wateret: > Can you explain the scenario where this code kicks in but the code above under tree->IsCopyOrReloadOfMultiRegCall() does not?

It looks like that `IsCopyOrReloadOfMultiRegCall()` is for the CopyOrReload node and its op1 is **multiple return registers**. 

> Is that scenario ARM specific?

This is ARMEL specific scenario. On hardfp, it will use d0, d1, ...(which are treated as one register) so it would not have the same problem with armel.

wateret:

wateret:
@dotnet-bot test please

wateret:
@dotnet-bot Help

wateret:
@dotnet-bot test this please

BruceForstall:

PR #12580, opened 2017-06-30
Disable sync.cmd -n for Windows binaries

wtgodbe:
@chcosta this is the change I mentioned on your PR earlier today

### DIFF
--- a/src/syncAzure.proj
+++ b/src/syncAzure.proj
@@ -21,5 +21,6 @@
     <Error Condition="'$(RuntimeId)' == '' and '$(PublishTestNativeBins)' == 'true'" Text="Missing property RuntimeId." />
   </Target>
 
-  <Target Name="Build" DependsOnTargets="ValidateRequiredProperties;DownloadBlobsFromAzureTargets" />
+  <!-- We don't want to run sync on test native binaries for Windows, since those don't get published -->
+  <Target Name="Build" DependsOnTargets="ValidateRequiredProperties;DownloadBlobsFromAzureTargets" Condition="'$(PublishTestNativeBins)' != 'true' or !$(RuntimeId.ToLower().StartsWith('win'))"/>
 </Project>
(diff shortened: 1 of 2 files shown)


src/syncAzure.proj
```diff
@@ -10,6 +10,9 @@
     <BlobNamePrefix>$(__BlobNamePrefix)</BlobNamePrefix>
     <DownloadDirectory Condition="'$(PublishTestNativeBins)' != 'true'">$(PackagesDir)AzureTransfer</DownloadDirectory>
     <DownloadDirectory Condition="'$(PublishTestNativeBins)' == 'true'">$(PackagesDir)TestNativeBins\$(RuntimeId)</DownloadDirectory>
+
+    <!-- We don't want to run sync on test native binaries for Windows, since those don't get published -->
+    <SyncingNativeOnWindows Condition="'$(PublishTestNativeBins)' == 'true' and $(RuntimeId.ToLower().StartsWith('win'))">true</SyncingNativeOnWindows>
```
chcosta: Do you want to use `BuildOS` for determining current OS?

chcosta: Is the intent of `SyncingNativeOnWindows` to allow a hook so that you could override its value and force the target below to evaluate, or is it just to capture the two condition cases specified?  If it's the former, then that makes sense.  If it's the latter (because overriding this isn't really valid), then I'd just move the condition into the target below.

wtgodbe: No, because we could be running on Windows but restoring for Ubuntu/OSX/Whatever. RuntimeId defines what platform we're restoring for, which is what we care about here.

wtgodbe: It's the latter - I wanted it available at the time "Build" gets defined so that Build could be conditioned on it - I don't want it to be an error condition, since the point of this is to not exit with failure. Are you suggesting just changing the Condition for "Build" to use the two properties that define SyncingNativeOnWindows? I could get behind that

wtgodbe:
@dotnet-bot test this please

src/syncAzure.proj
chcosta: Yes, that's what I'm suggesting.

wtgodbe: Updated

PR #12581, opened 2017-06-30
PGI: Load pgort<ver>.dll from the VS native tools env; do not `install` it

### LINKED ISSUE #12347: pgoinstrument builds don't work on VS2017+
`pgosupport.cmake` tries to find the runtime used to run instrumented binaries (`pgort<ver>.dll`) and install (copy) it alongside the product binaries when building PGI, in order to avoid introducing additional external dependencies for an instrumented build. It attempts to be smart about finding the right instance of pgort, but unfortunately VS2017 changed conventions for not only the version numbering, but also the location of the file. As a result, `pgoinstrument` builds fail due to an error that they can't find `pgort150.dll`.

Either pgort detection needs to be hardened to work with VS2017+ in addition to VS2015, or the logic needs to be removed altogether (which would make instrumented builds work only in a native tools environment).

dpodder:
On Windows, PGO instrumented builds (`build.cmd release <arch> pgoinstrument`) introduce a runtime dependency on `pgort<ver>.dll` for instrumented binaries. This DLL is distributed alongside the C++ compiler, and is made available via the native tools environment that ships with Visual Studio.

Previously, we were using cmake to find and "install" this binary alongside the product when doing an instrumented build, so that the resulting bin\Product drop is free of any added external dependencies. However, this approach is fragile, and despite a best effort to make the implementation work across multiple VS releases, it already broke with VS 2017.

To fix support for pgoinstrument on VS 2017, and to harden the implementation for future releases of VS, I'm removing the custom cmake install logic for the pgort DLL. Instead, we fall back to the officially supported method: load the correct (native tools) environment before invoking any command that uses an instrumented binary. This happens in one place in the build today--loading the JIT to crossgen System.Private.CoreLib.dll.

Fixes #12347.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -540,6 +540,19 @@ set PATH=%PATH%;%WinDir%\Microsoft.Net\Framework64\V4.0.30319;%WinDir%\Microsoft
 if %__BuildNativeCoreLib% EQU 1 (
     echo %__MsgPrefix%Generating native image of System.Private.CoreLib for %__BuildOS%.%__BuildArch%.%__BuildType%
 
+    REM Need VS native tools environment for the **target** arch when running instrumented binaries
+    if %__PgoInstrument% EQU 1 (
+        set __VCExecArch=%__BuildArch%
+        if /i [%__BuildArch%] == [x64] set __VCExecArch=amd64
+        echo %__MsgPrefix%Using environment: "%__VCToolsRoot%\vcvarsall.bat" !__VCExecArch!
+        call                                 "%__VCToolsRoot%\vcvarsall.bat" !__VCExecArch!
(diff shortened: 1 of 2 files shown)


karelz:
What is status of this PR? 2 months no update :( ...

dpodder:
@karelz It's been on hold to avoid breaking the opt data training automation. There's a small bit of work that has to happen there, and that has been deferred next to 2.0 release priorities.

I hope to be able to wrap it up in the next week or two, and then this can be merged. I'll remove the no merge label when infrastructure is ready for this to go in.

dpodder:
This is now unblocked and ready for review. @karelz @janvorli PTAL
/cc: @dotnet/rap-team

PR #12583, opened 2017-06-30
Get rid of 'goto' in src\jit\hashbv.cpp

rartemev:
Refactored hashbv.cpp, got rid of 'goto' and refreshed code a little.

### DIFF
--- a/src/jit/hashbv.cpp
+++ b/src/jit/hashbv.cpp
@@ -1964,58 +1964,52 @@ indexType hashBvIterator::nextBit()
     // printf("in nextbit for bv:\n");
     // this->bv->dump();
 
-    if (!this->currNode)
+    if (this->currNode == nullptr)
     {
         this->nextNode();
     }
 
-top:
-
-    if (!this->currNode)
+    while (this->currNode != nullptr)
     {
-        return NOMOREBITS;
-    }
-
-more_data:
-    if (!this->current_data)
-    {
-        current_element++;
-        // printf("current element is %d\n", current_element);
-        // reached the end of this node
-        if (current_element == (indexType) this->currNode->numElements())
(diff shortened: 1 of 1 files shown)


src/jit/hashbv.cpp
```diff
@@ -1964,58 +1964,51 @@ indexType hashBvIterator::nextBit()
...
             else
             {
-                current_data >>= 1;
-                current_base++;
+                while (current_data != 0)
+                {
+                    current_data >>= 1;
+                    current_base++;
+
+                    if ((current_data & 1) != 0)
```
pgavlin: I think that this refactoring is incorrect. Previously, this test was done before right-shifting `current_data`, but now it is done after the shift.

BruceForstall:
@rartemev Is there a reason you're making these changes besides general "cleanliness"? E.g., better performance, looking forward to more restructuring, etc.? The reason I ask is this code hasn't been changed in a long time and I don't really see any reason to touch it now.

rartemev:
@BruceForstall The reason is I think that eliminate goto from source code is a good idea anyway. So I grep'ed for goto in src/jit* directory, took a random file and fixed it. I've never heard about 'cleanliness'. If I did something wrong I may close PR.

src/jit/hashbv.cpp
rartemev: fixed

BruceForstall:
> The reason is I think that eliminate goto from source code is a good idea anyway.

I think there are many things we could do to improve the source, but I don't think searching for gotos and eliminating them is necessarily a good use of time.

In addition, from what I can tell, this function currently isn't used, so I don't think we'll get any test coverage from any changes to it.

PR #12585, opened 2017-06-30
Improve fgAddRefPred

### LINKED ISSUE #12582: JIT throughput: fgAddRefPred
`Compiler::fgAddRefPred()` shows up high on some profiles, in particular:

```
        // TODO-Throughput: This search is quadratic if you have many jumps
        // to the same target.   We need to either not bother sorting for
        // debuggable code, or sort in optFindNaturalLoops, or better, make
        // the code in optFindNaturalLoops not depend on order.

        flowList** listp = &block->bbPreds;
        while (*listp && ((*listp)->flBlock->bbNum < blockPred->bbNum))
        {
            listp = &(*listp)->flNext;
        }
```
before this is:
```
    flowList* flow = fgGetPredForBlock(block, blockPred);
```
which walks the entire preds list looking for a match.

However, we should be able to merge these two walks:
1. In the case where the pred exists, we can early exit when the bbNum in the list is higher than the bbNum we are looking for, since the list is sorted.
2. This is the same exit condition for the 2nd loop to add it to the sorted list.

BruceForstall:
Only do one search through the predecessor list looking for an
existing predecessor edge, or looking for the proper location to
insert a new predecessor edge. Previously, we traversed the list
twice (or part of twice) for new edges. For new edges, stop the
search when we reach the correct location in sorted order.

This reduces x86 release minopts instruction count by 0.42% and
full opts instruction count by 0.33% (of a superpmi replay of
the tests).

Fixes #12582

### DIFF
--- a/src/jit/flowgraph.cpp
+++ b/src/jit/flowgraph.cpp
@@ -1044,10 +1044,27 @@ flowList* Compiler::fgAddRefPred(BasicBlock* block,
 
     assert(!fgCheapPredsValid);
 
-    flowList* flow = fgGetPredForBlock(block, blockPred);
+    flowList* flow;
 
-    if (flow)
+    // Keep the predecessor list in lowest to highest bbNum order. This allows us to discover the loops in
+    // optFindNaturalLoops from innermost to outermost.
+    //
+    // TODO-Throughput: This search is quadratic if you have many jumps to the same target. We need to either
+    // not bother sorting for debuggable code, or sort in optFindNaturalLoops, or better, make the code in
+    // optFindNaturalLoops not depend on order. This also requires ensuring that nobody else has taken a
+    // dependency on this order. Note also that we don't allow duplicates in the list; we maintain a flDupCount
+    // count of duplication. This also necessitates walking the flow list for every addition.
+
+    flowList** listp = &block->bbPreds;
+    while (*listp && ((*listp)->flBlock->bbNum < blockPred->bbNum))
     {
+        listp = &(*listp)->flNext;
+    }
+
+    if ((*listp != nullptr) && ((*listp)->flBlock == blockPred))
+    {
+        // The predecessor block already exists in the flow list; simply add to its duplicate count.
+        flow = *listp;
         noway_assert(flow->flDupCount > 0);
         flow->flDupCount++;
     }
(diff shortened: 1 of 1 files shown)


BruceForstall:
cc @dotnet/jit-contrib

src/jit/flowgraph.cpp
```diff
@@ -1044,10 +1044,27 @@ flowList* Compiler::fgAddRefPred(BasicBlock* block,
...
     assert(!fgCheapPredsValid);
 
-    flowList* flow = fgGetPredForBlock(block, blockPred);
+    flowList* flow;
 
-    if (flow)
+    // Keep the predecessor list in lowest to highest bbNum order. This allows us to discover the loops in
+    // optFindNaturalLoops from innermost to outermost.
+    //
+    // TODO-Throughput: This search is quadratic if you have many jumps to the same target. We need to either
```
pgavlin: Perhaps I'm missing something, but this search no longer seems quadratic...

src/jit/flowgraph.cpp
```diff
@@ -1044,10 +1044,27 @@ flowList* Compiler::fgAddRefPred(BasicBlock* block,
...
+    // not bother sorting for debuggable code, or sort in optFindNaturalLoops, or better, make the code in
+    // optFindNaturalLoops not depend on order. This also requires ensuring that nobody else has taken a
+    // dependency on this order. Note also that we don't allow duplicates in the list; we maintain a flDupCount
+    // count of duplication. This also necessitates walking the flow list for every addition.
+
+    flowList** listp = &block->bbPreds;
+    while (*listp && ((*listp)->flBlock->bbNum < blockPred->bbNum))
```
pgavlin: Nit: ``while ((*listp != nullptr) ...`

pgavlin:
LGTM aside from a comment that appears stale and a small nit.

src/jit/flowgraph.cpp
CarolEidt: I don't think this method itself was ever quadratic. I assume that the comment refers to its usage in setting up all the preds for all the blocks.

BruceForstall: Correct, what's quadratic is adding N predecessor edges to a block, requiring searching the predecessor list for each one.

BruceForstall:
@dotnet-bot test CROSS Check

BruceForstall:
@mmitche What is CROSS Check? Anyway, it failed with:
```
17:10:13 D:\j\workspace\CROSS_check_p---6e5c5740>powershell -Command "wget '****' -OutFile cross-integration.zip" 
17:10:17 wget : The term 'Invoke-WebRequest' is not recognized as the name of a cmdlet, 
17:10:17 function, script file, or operable program. Check the spelling of the name, or 
17:10:17 if a path was included, verify that the path is correct and try again.
17:10:17 At line:1 char:1
17:10:17 + wget 
17:10:17 'https://crossdist.blob.core.windows.net/debug/cross.zip?sv=2015-04-05&sr=c ...
17:10:17 + ~~~~
17:10:17     + CategoryInfo          : ObjectNotFound: (Invoke-WebRequest:String) [], C 
17:10:17    ommandNotFoundException
17:10:17     + FullyQualifiedErrorId : CommandNotFoundException
17:10:17  
```
so I retried

BruceForstall:

PR #12587, opened 2017-07-01
Automatic build runtime layout when skipping tests

yizhang82:
When you are iterating on a fix/feature to make a test pass, it's nice to avoid build all the tests (which takes forever) but you still want your runtime changes to go into the runtime layout directory bin\tests\Windows_NT.x64.Debug\Tests\Core_Root. This makes it so that we would build runtime layout even when you specify skiptests.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -501,7 +501,17 @@ if %__BuildTests% EQU 1 (
         REM buildtest.cmd has already emitted an error message and mentioned the build log file to examine.
         exit /b 1
     )
-)
+) else (
+    echo %__MsgPrefix%Skipping test build. Proceeding to building runtime layout for %__BuildOS%.%__BuildArch%.%__BuildType%
+
+    echo "%__ProjectDir%\tests\runtest.cmd %__BuildArch% %__BuildType% %__UnprocessedBuildArgs% GenerateLayoutOnly"
+    @call %__ProjectDir%\tests\runtest.cmd %__BuildArch% %__BuildType% %__UnprocessedBuildArgs% GenerateLayoutOnly
+
+    if not !errorlevel! == 0 (
+        REM runtest.cmd has already emitted an error message and mentioned the build log file to examine.
+        exit /b 1
+    )    
+) 
 
 REM =========================================================================================
 REM ===


jkotas:
@dotnet-bot test this please

jkotas:
@dotnet-bot test this please

mikedn:
Doesn't `runtest.cmd generatelayoutonly` do the same thing?

yizhang82:
@mikedn Why two steps (build+runtest) when you only need one? 
@janvorli @jkotas Can I get a quick review?

mikedn:
> Why two steps (build+runtest) when you only need one?

Well, this change appears to generate layout unconditionally. That wastes some time when not needed.

yizhang82:
@mikedn Comparing to the build itself (which incrementally compiles/links multiple projects), crossgen System.Private.CoreLib, packaging everything, this additional step is just a bunch of copy and barely noticeable. If you have a particular scenario in mind that this additional step become a problem, let me know.

mikedn:
> Comparing to the build itself (which incrementally compiles/links multiple projects), crossgen System.Private.CoreLib, packaging everything

corelib and packaging can be skipped (I always skip packaging since it's pretty much pointless in a local build). And when you're working in certain areas of the CLR incremental builds can be pretty fast. For example JIT related builds are much faster than VM related builds.

> this additional step is just a bunch of copy and barely noticeable

Well, it does copy a few hundred megabytes so it takes a bit of time (currently it copies the PDB directory which is quite large in debug and checked builds).

> If you have a particular scenario in mind that this additional step become a problem, let me know.

Let's see how it goes as is. We can always add a skiplayout option later.

yizhang82:
> corelib and packaging can be skipped (I always skip packaging since it's pretty much pointless in a local build). And when you're working in certain areas of the CLR incremental builds can be pretty fast. For example JIT related builds are much faster than VM related builds.

@mikedn That's a fair point. I suppose I'm biased since I never change the JIT :) Let see how this goes. Personally I would always use the CORE_ROOT directory since it has the full layout and is most convenient. But other people might have a different workflow. 

> We can always add a skiplayout option later.

Agreed.

jkotas:
Make a matching change on Unix as well? It would be nice to maintain consistent experience between Windows and Unix.

cc @wtgodbe @rartemev

yizhang82:
@jkotas As far as I can tell, we need to fix linux workflow first since we don't really build the managed tests there. I know there are folks working on this but there might be some performance problem (from msbuild unable to perform parallel builds on linux, IIRC?) so it is not enabled yet. We need to fix that first.

BruceForstall:
This change is going to break, or at least cause duplication in, existing scripting and automation.

Also, it's not clear why this is only done if "skiptests" is passed. If you actually do the test build, you still don't get a "runtest.cmd GenerateLayoutOnly" run to generate the layout. Seems like you would want to do the layout generation unconditionally.

This is also going to duplicate work if your process is to run "build.cmd & tests\runtest.cmd", as you'll generate layouts twice.

yizhang82:
@BruceForstall I can see your point for the duplication and it's not clear to me where the clear divide should be (since build.cmd already builds the test...). In my opinion that once I ran builld.cmd all the assets should be in a state that I can run the tests individually, hence the building test layout unconditionally (otherwise why are we even building the test in build.cmd). If there is possible duplication, we should fix the scripts to be smart enough to skip them as needed (for example, runtest.cmd should skip/incrementally building tests if they are already built). Of course, some duplicate work are more difficult to avoid than others.

yizhang82:
@BruceForstall Also, I think we build the runtime layouts when runtest.cmd builds the tests. But if I skip it, the layouts don't get generated. So my change effectively makes layout generation unconditional.

yizhang82:
I'm reverting this for now - looks like it's building official build when it passes -skiprestore:

```
2017-07-06T23:11:37.0946295Z ##[command]build.cmd x86 Checked skiptests skipbuildpackages -OfficialBuildId=20170706-04 -skiprestore -Priority=0
2017-07-06T23:17:55.9531903Z "C:\Program Files (x86)\MSBuild\14.0\Bin\MSBuild.exe" "E:\A\_work\245\s\tests\runtest.proj" /p:GenerateRuntimeLayout=true /nologo /nodeReuse:false  /p:Platform=x86 /maxcpucount /fileloggerparameters:Verbosity=normal;LogFile="E:\A\_work\245\s\tests\..\bin\Logs\Tests_GenerateRuntimeLayout_Windows_NT__x86__Checked.log";Append /fileloggerparameters1:WarningsOnly;LogFile="E:\A\_work\245\s\tests\..\bin\Logs\Tests_GenerateRuntimeLayout_Windows_NT__x86__Checked.wrn" /fileloggerparameters2:ErrorsOnly;LogFile="E:\A\_work\245\s\tests\..\bin\Logs\Tests_GenerateRuntimeLayout_Windows_NT__x86__Checked.err" /consoleloggerparameters:Summary /verbosity:minimal
2017-07-06T23:17:56.0782984Z E:\A\_work\245\s\tests\publishdependency.targets(48,5): error : Lock file E:\A\_work\245\s\tests\src\Common\test_runtime\obj\project.assets.json couldn't be found. Run a NuGet package restore to generate this file. [E:\A\_work\245\s\tests\runtest.proj]
2017-07-06T23:17:56.0939391Z 
2017-07-06T23:17:56.0939391Z E:\A\_work\245\s\tests\publishdependency.targets(48,5): error : Lock file E:\A\_work\245\s\tests\src\Common\test_runtime\obj\project.assets.json couldn't be found. Run a NuGet package restore to generate this file. [E:\A\_work\245\s\tests\runtest.proj]
2017-07-06T23:17:56.0939391Z RUNTEST: Error: msbuild failed. Refer to the log files for details:
2017-07-06T23:17:56.0939391Z     E:\A\_work\245\s\tests\..\bin\Logs\Tests_GenerateRuntimeLayout_Windows_NT__x86__Checked.log
2017-07-06T23:17:56.0939391Z     E:\A\_work\245\s\tests\..\bin\Logs\Tests_GenerateRuntimeLayout_Windows_NT__x86__Checked.wrn
2017-07-06T23:17:56.0939391Z     E:\A\_work\245\s\tests\..\bin\Logs\Tests_GenerateRuntimeLayout_Windows_NT__x86__Chec
```

PR #12590, opened 2017-07-01
Implement a new approach for SIMD8/LONG interactions.

pgavlin:
SIMD8 values need to be converted to longs under a small number of
situations on x64/Windows:
- SIMD8 values are passed and returned as LONGs
- SIMD8 values may be stored to a LONG lclVar

Currently, LSRA performs some gymnastics when building use positions in
order to ensure that registers are properly allocated. This change is a
stab at a different approach: rather than pushing this work onto the RA,
lowering inserts `GT_BITCAST` nodes where necessary to indicate
that the source long- or SIMD8-typed value should be retinterpreted as
a SIMD8- or long-typed value as necessary. The RA performs one specific
optimization wherein it retypes stores of `GT_BITCAST` nodes to
non-register-candidate local vars with the type of the cast's operand
and preferences the cast to its source interval.

This approach trades slightly larger IR for some functions that
manipulate SIMD8 values for tighter code in buildRefPositions.

### DIFF
--- a/src/jit/lsra.cpp
+++ b/src/jit/lsra.cpp
@@ -2242,6 +2242,7 @@ void LinearScan::identifyCandidates()
                     varDsc->lvLRACandidate = 0;
                 }
                 break;
+
             // TODO-1stClassStructs: Move TYP_SIMD8 up with the other SIMD types, after handling the param issue
             // (passing & returning as TYP_LONG).
             case TYP_SIMD8:
@@ -3749,10 +3750,13 @@ void LinearScan::buildRefPositionsForNode(GenTree*                  tree,
     assert(!tree->OperIsAssignment());
     if (tree->OperIsLocalStore())
     {
-        if (isCandidateLocalRef(tree))
+        GenTreeLclVarCommon* const store = tree->AsLclVarCommon();
+        assert((consume > 1) || (regType(store->gtOp1->TypeGet()) == regType(store->TypeGet())));
+
+        LclVarDsc* varDsc = &compiler->lvaTable[store->gtLclNum];
+        if (isCandidateVar(varDsc))
         {
             // We always push the tracked lclVar intervals
-            LclVarDsc* varDsc = &compiler->lvaTable[tree->gtLclVarCommon.gtLclNum];
             assert(varDsc->lvTracked);
             unsigned varIndex = varDsc->lvVarIndex;
             varDefInterval    = getIntervalForLocalVar(varIndex);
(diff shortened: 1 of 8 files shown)


pgavlin:
@dotnet-bot
test FreeBSD x64 Checked Build 
test Ubuntu arm64 Cross Debug Build
test Windows_NT arm64 Cross Debug Build

pgavlin:
@dotnet-bot
test FreeBSD x64 Checked Build
test Ubuntu arm64 Cross Debug Build
test Windows_NT arm64 Cross Debug Build

pgavlin:
@dotnet-bot
test FreeBSD x64 Checked Build
test Ubuntu arm64 Cross Debug Build
test Windows_NT arm64 Cross Debug Build

pgavlin:
@CarolEidt

I think that this is in good enough shape for a conceptual review. I'm not 100% confident in the choice of IR node, as it might be better to invent something new (perhaps `GT_REINTERPRET`) rather than using `CAST` nodes, but I do believe that we ought to insert explicit IR to represent the SIMD -> scalar copy/conversion/what have you. I'll have TP measurements for this change a later this morning; I expect this change to put a small dent in the number of instructions retired by `buildRefPositionsForNode`.

Thoughts?

pgavlin:
@CarolEidt measuring instructions retired while crossgen'ing S.P.C.dll on *nix indicates about a .25% improvement in throughput for that target. Although I haven't measured it, the improvement should be larger on x86 as that target does not need to perform any IR transformations.

pgavlin:

CarolEidt:
> it might be better to invent something new (perhaps GT_REINTERPRET) rather than using CAST nodes

I agree - `GT_CAST` implies more than just copying the bits. `GT_REINTERPRET` is too long, IMO. How about simply `GT_MOVE`?

src/jit/lsra.cpp
```diff
@@ -3742,18 +3743,20 @@ void LinearScan::buildRefPositionsForNode(GenTree*                  tree,
 
     bool isSpecialPutArg = false;
 
-    assert(!tree->OperIsAssignment());
```
CarolEidt: I don't necessarily object to removing it, but it is unrelated to this change.

src/jit/lsra.cpp
```diff
@@ -3742,18 +3743,20 @@ void LinearScan::buildRefPositionsForNode(GenTree*                  tree,
...
     if (tree->OperIsLocalStore())
     {
-        if (isCandidateLocalRef(tree))
+        GenTreeLclVarCommon* const store = tree->AsLclVarCommon();
+        assert((consume > 1) || (regType(store->gtOp1->TypeGet()) == regType(store->TypeGet())));
+
+        LclVarDsc* varDsc = &compiler->lvaTable[store->gtLclNum];
+        if (isCandidateVar(varDsc))
```
CarolEidt: Again, these changes seem to be unrelated to this change, and I'd rather not see them conflated.

pgavlin:
> I agree - `GT_CAST` implies more than just copying the bits. `GT_REINTERPRET` is too long, IMO. How about simply `GT_MOVE`?

The thing that I worry about with `GT_MOVE` is that to me it implies pretty concrete semantics--i.e. moving a value from one location to another--and I think that we might want to eventually insert these nodes pretty early on (e.g. in morph or the importer) where those semantics don't entirely make sense. The semantics of the `GT_REINTERPRET` node would be "take these bits of this type and reinterpret them as this other type", which seem applicable at all levels.

mikedn:
> GT_REINTERPRET is too long

`GT_RECAST` ?

CarolEidt:
> GT_RECAST ?

Or perhaps just `GT_RETYPE`?

pgavlin:
FWIW, `GT_REINTERPRET` is not that long: `GT_PUTARG_*`, `GT_MEMORYBARRIER`, `GT_RETURNTRAP`, `GT_PINVOKE_*`, etc. are all of similar length. I'd prefer to stick with `REINTERPRET` if possible, as the IMO it's pretty clear re: its intended semantics. Out of the other options presented, `GT_RETYPE` seems best so far.

src/jit/lsra.cpp
pgavlin: These are related: we need `store` for the `REINTERPRET` case below, and it makes sense to share it with the common case.

pgavlin:

pgavlin:
This is now ready for review. I've used `GT_REINTERPRET` for the proposed node name, as I feel that it remains the clearest option.

pgavlin:

pgavlin:

CarolEidt:
> This is now ready for review. I've used GT_REINTERPRET for the proposed node name, as I feel that it remains the clearest option.

I strongly prefer `GT_RETYPE`. I think it is just as clear, and simply because we have existing long names, I don't think it makes it any more attractive.

pgavlin:
> I strongly prefer `GT_RETYPE`. I think it is just as clear

I chose `GT_REINTERPRET` in part because of prior art that uses this term for similar (though not entirely identical) operations, namely C++'s `reinterpret_cast`. Ideally, we would rename `GT_CAST` to `GT_CONVERT` and use `GT_CAST` for the proposed node, as the familiar meaning of these terms aligns even more closely with the desired behaviors.

That said, I'm fine with changing the name to `GT_RETYPE` if I am outvoted.

CarolEidt:
> I chose GT_REINTERPRET in part because of prior art that uses this term for similar (though not entirely identical) operations, namely C++'s reinterpret_cast.

I was never a fan of that name, either. Note that, in C++, the use of `reinterpret_cast` is perhaps intentionally ugly, because it is a somewhat exceptional thing to do. Unlike that case, this is not a subversion of typing; it is simply a case of typing the register to which it will be assigned.

> Ideally, we would rename GT_CAST to GT_CONVERT and use GT_CAST for the proposed node, as the familiar meaning of these terms aligns even more closely with the desired behaviors.

I see no need to do that either. Casting usually implies conversion, hence the reason for C++'s use of a distinguished term for the non-conversion cast.

JosephTremoulet:
We could follow some [prior art](http://llvm.org/docs/LangRef.html#bitcast-to-instruction) and call it `GT_BITCAST`

pgavlin:
> We could follow some prior art and call it `GT_BITCAST`

I love it. Done.

